PR #19, opened 2015-10-27
Fixes a bug with trailing newlines in toolchain path.

norcalli:
I was getting the following error when running cargo.

    error: not a directory: '/Users/ashkan/.multirust/toolchains/stable
    '

and noticed that the quote goes to the next line, so I added a trim to the
toolchain constructor for the only part that seemed to be problematic. Possibly
not the best solution, but it fixed my issue.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -19,6 +19,7 @@ pub struct Toolchain<'a> {
 
 impl<'a> Toolchain<'a> {
 	pub fn from(cfg: &'a Cfg, name: &str) -> Self {
+        let name = name.trim_matches('\n');
 		Toolchain {
 			cfg: cfg,
 			name: name.to_owned(),


Diggsey:
Hi @norcalli, thanks for the pull request. However, I'd prefer to fix this at the source.

I've pushed a new version which trims newlines when reading the default toolchain. Could you tell me if that fixes the problem?

Also, could you show me what you did to get this to occur, so I can add a regression test?

norcalli:
It does seem to have fixed the problem after running `cargo build`, then `rm ~/.multirust/bin/multirust` and finally `./target/release self install`, which I hope is the appropriate way to reinstall.

The exact steps I took to recreate this was to simply pull the `lettre` crate into my `~/Workspace/projects/lettre` folder and then attempt to run cargo after installing multirust-rs. I was using the regular `multirust` before, so perhaps it has something to do with that.

Diggsey:
@norcalli Ah thanks, it was exactly as I'd suspected: the original multirust puts a newline in the "default" file, whereas multirust-rs doesn't. Simply trimming the result after reading that file was sufficient.

(Yep, that's the correct way to reinstall from source - you can also just do `multirust self update` to magically download and install updated binaries!)

PR #27, opened 2015-10-28
Build Linux binaries in an old CentOS 5 image

alexcrichton:
This should help the binaries be maximally compatible with older Linux releases
as they'll depend on a very old glibc. This also touches up the linkage to
openssl to be static on Linux and dynamic on OSX.

The docker image used here is currently the same one that we're using in
rust-lang/rust for the production of the nightly compilers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ git:
   submodules: false
 
 sudo: required
+services:
+  - docker
 install:
   - git config --global user.email "diggsey@googlemail.com"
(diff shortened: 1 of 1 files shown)


alexcrichton:
It also looks like this [doesn't slow down the CI all that much](https://travis-ci.org/alexcrichton/multirust-rs/builds/87929389) as well!

Diggsey:
Awesome! I seem to spend ages fiddling with the CI build scripts, so I'm very glad I didn't have to figure this one out :stuck_out_tongue:

PR #39, opened 2015-12-17
Update openssl and hyper

jwilm:
The currently released version of multirust-rs is incompatible with other packages on crates.io which require newer hyper/openssl. This updates hyper and openssl to the latest versions. Consumers requiring the old versions can still depend on `0.0.4`, and those of us that require newer versions can use `0.0.5` :wink:.

Thanks!


Diggsey:
This passed on appveyor now, thanks!

PR #45, opened 2016-01-26
Write tests for rust_install::component

brson:
Basic tests for installation and uninstallation.

Fixed some bugs. Added validation of package installer versions.
Added validation of installed metadata format, bringing it closer
inline with what rust-installer produces. Added permission setting
on unix.

Tabs vs. spaces is screwy here. I'll submit a follow up that rustfmts master, merges master back into new, and rustfmts new.

I'm planning on writing tests for file system transactions next, then investigating the manifest code.

It's possible this branch will start diverging a lot from the master branch soon, and it'll get harder to keep them in sync. I'm not sure yet whether it's best to try to maintain continuity with the v1 manifest world, or consider this a new product and jettison the legacy stuff.


alexcrichton:
You may want to make sure that these are running on both appveyor/travis, I think the top-level Cargo tests are being run but perhaps not the `rust-install` ones?

It also may make sense to start out these tests as integration tests in `tests/` as they require lots of filesystem interaction, but that's just a stylistic choice on your behalf really

rust-install/src/component/components.rs
```diff
@@ -36,22 +40,23 @@ impl<'a> ChangeSet<'a> {
...
-	pub fn new(prefix: InstallPrefix) -> Option<Self> {
-		if utils::is_file(prefix.manifest_file(COMPONENTS_FILE)) {
-			Some(Components { prefix: prefix })
-		} else {
-			None
-		}
-	}
-	pub fn init(prefix: InstallPrefix) -> Result<Self> {
-		try!(utils::write_file("components", &prefix.manifest_file(COMPONENTS_FILE), ""));
-		Ok(Components { prefix: prefix })
```
Diggsey: Could you explain why `init` was removed? The `new` -> `open` and Option -> Result changes make sense, although the other "InstallPrefix wrappers" for lack of a better term should probably be refactored in the same way (Configuration and Manifestation). `init` was required for initializing an fresh install.

rust-install/src/component/package.rs
```diff
@@ -77,6 +99,68 @@ impl Package for DirectoryPackage {
 	}
 }
 
+// On Unix we need to set up the file permissions correctly so
+// binaries are executable and directories readable.
```
Diggsey: Is this strictly necessary? Can't the permissions be set correctly when the package is built? If not, can this be done as part of `copy_file`/`copy_dir`: I'm wary of doing such a blanket permissions change, and I'd also like to avoid modifying the installation source in any way.

rust-install/src/errors.rs
```diff
@@ -41,6 +43,10 @@ pub enum Error {
 	ExtensionNotFound(rust_manifest::Component),
 	InvalidChangeSet,
 	NoGPG,
+    InstallerVersion(String),
+    InstalledMetadataVersion(String),
+    WalkDirForPermissions(walkdir::Error),
+    SetPermissions(io::Error),
```
Diggsey: I generally prefer to use error names to describe the problem, rather than the action during which the problem occurred. (eg. BadInstallerVersion) although I concede that this can result in lengthier error names. (It makes it a lot easier for the caller to handle specific error-cases, and the error messages to the user can be a lot friendlier)

rust-install/src/install.rs
```diff
@@ -16,16 +16,21 @@ use std::env;
 
 #[cfg(not(windows))]
 const REL_MANIFEST_DIR: &'static str = "lib/rustlib";
+
+// FIXME: (brson) I don't think this should be bin/ even though
+// windows puts its libraries in bin.  rust-installer uses lib/ no
+// matter what platform, so using bin here makes the metadata used by
+// multirust-rs incompatible with rust-installer.
```
Diggsey: I'm happy for that to change, it's simply the case that currently, rust on windows places rustlib in `bin/`.

rust-install/src/component/components.rs
brson: Huh, I thought I left a comment about this. It's because initializing the `components` file before any components are written, and outside of a transaction, could leave the installation in a partial state where nothing is installed but there is a rustlib/components file. Both 'components' and 'rust-installer-version' are initialized inside the transaction that installs the first component.

brson: I didn't change Configuration and manifestation because I haven't started learning that code yet, but I can also change them today.

rust-install/src/component/package.rs
brson: For the time being it is, yes. There is a [bug with Rust's tarballs](https://github.com/rust-lang/rust/issues/25479) that causes them not to extract with the correct permissions. I don't know how to fix it, but even if we do it will take some time to roll out.

rust-install/src/errors.rs
brson: I'll try to think of more descriptive error names.

rust-install/src/install.rs
brson: I want the location of rustlib to be consistent across platforms, and @alexcrichton tells me that there was a change recently for windows to put rustlib in lib instead of bin, so this should actually be sorted out upstream.

In any case, multirust's metadata is completely independent from rustc's so there's no chance of actual conflicts if they *do* use two different rustlibs.

brson:
Thanks for the review @Diggsey

rust-install/src/component/components.rs
Diggsey: No problem - it was meant more as a note that we'll have to do it than as a criticism of the PR!

brson:
[Here's the rustfmt PR against master](https://github.com/Diggsey/multirust-rs/pull/47). I've updated this PR to move the tests to the tests directory and run rustfmt over it. I'm going to continue writing tests for the transaction code now and will keep appending commits to this PR, then I will move on to learning the manifest code.

Diggsey:
Incredible progress! I think @alexcrichton is right though: none of the `rust_install` tests seem to be running on travis.

brson:
Hm, the history on this PR looks weird with some other folks commits. I must have merged something wrong.

brson:
Uh oh. rustfmting everything made the combined diff useless...

brson:
OK, I've addressed all the feedback. I'm not going to push more work to this branch because of the big diff already.

The appveyor build is having trouble linking openssl. Here are some of the errors:

```
libopenssl_sys-26ea9bf6206dbf1a.rlib(bio_ndef.obj) : warning LNK4099: PDB 'libMT.pdb' was not found with 'libopenssl_sys-26ea9bf6206dbf1a.rlib(bio_ndef.obj)' or at 'C:\projects\multirust-rs\target\debug\libMT.pdb'; linking object as if no debug info
libopenssl_sys-26ea9bf6206dbf1a.rlib(asn_mime.obj) : warning LNK4099: PDB 'libMT.pdb' was not found with 'libopenssl_sys-26ea9bf6206dbf1a.rlib(asn_mime.obj)' or at 'C:\projects\multirust-rs\target\debug\libMT.pdb'; linking object as if no debug info
libopenssl_sys-26ea9bf6206dbf1a.rlib(bio_asn1.obj) : warning LNK4099: PDB 'libMT.pdb' was not found with 'libopenssl_sys-26ea9bf6206dbf1a.rlib(bio_asn1.obj)' or at 'C:\projects\multirust-rs\target\debug\libMT.pdb'; linking object as if no debug info
libopenssl_sys-26ea9bf6206dbf1a.rlib(bio_b64.obj) : warning LNK4099: PDB 'libMT.pdb' was not found with 'libopenssl_sys-26ea9bf6206dbf1a.rlib(bio_b64.obj)' or at 'C:\projects\multirust-rs\target\debug\libMT.pdb'; linking object as if no debug info
libopenssl_sys-26ea9bf6206dbf1a.rlib(cryptlib.obj) : error LNK2019: unresolved external symbol __imp_GetUserObjectInformationW referenced in function OPENSSL_isservice
libopenssl_sys-26ea9bf6206dbf1a.rlib(cryptlib.obj) : error LNK2019: unresolved external symbol __imp_MessageBoxW referenced in function OPENSSL_showfatal
libopenssl_sys-26ea9bf6206dbf1a.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_CreateCompatibleBitmap referenced in function readscreen
libopenssl_sys-26ea9bf6206dbf1a.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_DeleteObject referenced in function readscreen
libopenssl_sys-26ea9bf6206dbf1a.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_GetDeviceCaps referenced in function readscreen
libopenssl_sys-26ea9bf6206dbf1a.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_GetDIBits referenced in function readscreen
libopenssl_sys-26ea9bf6206dbf1a.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_GetObjectW referenced in function readscreen
libopenssl_sys-26ea9bf6206dbf1a.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_GetDC referenced in function readscreen
libopenssl_sys-26ea9bf6206dbf1a.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_ReleaseDC referenced in function readscreen
C:\projects\multirust-rs\target\debug\deps\libshell32-d4cbd87c7e8d7b84.rlib : fatal error LNK1107: invalid or corrupt file: cannot read at 0x3120
 
Could not compile `multirust-rs`. 
```

It's using openssl from http://www.npcglib.org/~stathis/downloads/openssl-1.0.2d-vs2015.7z

alexcrichton:
I think that may have to do with some of the [old support](https://github.com/brson/multirust-rs/blob/master/appveyor.yml#L57-L62) in appveyor.yml. I believe this can now be fixed by emitting a `.cargo/config` that looks like:


```rust
[target.x86_64-pc-windows-msvc.openssl]
rustc-link-search = ['C:\OpenSSL\openssl-1.0.2d-vs2015\lib64']
rustc-link-libs = ['static=ssleay32MT', 'static=libeay32MT']
include = ['C:\OpenSSL\openssl-1.0.2d-vs2015\include']

[target.i686-pc-windows-msvc.openssl]
rustc-link-search = ['C:\OpenSSL\openssl-1.0.2d-vs2015\lib']
rustc-link-libs = ['static=ssleay32MT', 'static=libeay32MT']
include = ['C:\OpenSSL\openssl-1.0.2d-vs2015\include']

# ... etc
```

I think the GNU triples may also need some things there as well. Basically to statically link OpenSSL on Windows I think you'll want to use `.cargo/config` to override `openssl-sys`'s build script

alexcrichton:
I've [opened a PR](https://github.com/brson/multirust-rs/pull/1), but it looks like the `cli.bat` script still has some errors

brson:
Thanks alex.

brson:
OK, I tried to fix appveyor by adding `cargo build --release` as a test step so that the multirust bin ends up in the right place. If that's ok then this is good to go.

Diggsey:
\o/

brson:
Thanks Diggsey!

PR #47, opened 2016-01-28
Run rustfmt over everything

brson:
This most importantly converts tabs to spaces :)


Diggsey:
I'm OK with this being merged in whenever. However, it may screw up the diff against "new" and make merging it back in somewhat difficult.

brson:
I'll do a merge into new immediately.

PR #49, opened 2016-02-01
Tests for transactions and manifest parsing

brson:
Here are tests for the Transaction type and Manifest serialization.

I fixed some bugs along the way: copy_dir failing on windows, manifests not serializing correctly, probably others. Brought the manifest definition up to date with what I think it is going to look like.

I also reorganized some of the code I was working off of to read top-down because I find it easier to understand, so e.g. I put constructors first in impls and the `Transaction` type ahead of the `ChangedItem` type. This makes the diff harder to read. Sorry.

Now I'm working to understand how these pieces tie together to install components from the dist server, and writing integration tests with a local mock dist server.

### DIFF
--- a/rust-manifest/src/manifest.rs
+++ b/rust-manifest/src/manifest.rs
@@ -1,3 +1,15 @@
+//! Rust distribution v2 manifests.
+//!
+//! This manifest describes the distributable artifacts for a single
+//! release of Rust. They are toml files, typically downloaded from
+//! e.g. static.rust-lang.org/dist/channel-rust-nightly.toml. They
+//! describe where to download, for all platforms, each component of
+//! the a release, and their relationships to each other.
+//!
+//! Installers use this info to customize Rust installations.
+//!
+//! See tests/channel-rust-nightly-example.toml for an example.
+
 use errors::*;
 use toml;
 use utils::*;
@@ -7,20 +19,7 @@ use std::collections::HashMap;
 pub const SUPPORTED_MANIFEST_VERSIONS: [&'static str; 1] = ["2"];
 pub const DEFAULT_MANIFEST_VERSION: &'static str = "2";
 
-#[derive(Clone, Debug)]
-pub struct RequiredPackage {
-    pub url: String,
-    pub hash: String,
-}
-
-#[derive(Clone, Debug)]
-pub struct Diff {
-    pub packages: Vec<RequiredPackage>,
-    pub to_install: Vec<Component>,
-    pub to_uninstall: Vec<Component>,
-}
-
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct Manifest {
     pub manifest_version: String,
     pub date: String,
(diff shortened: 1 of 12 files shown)


rust-manifest/src/manifest.rs
```diff
@@ -48,21 +48,47 @@ pub struct Component {
...
 
-        Ok(result)
+        result
+    }
+
+    pub fn parse(data: &str) -> Result<Self> {
+        let mut parser = toml::Parser::new(data);
+        let value = try!(parser.parse().ok_or_else(move || Error::Parsing(parser.errors)));
+
+        Self::from_toml(value, "")
```
alexcrichton: In terms of ease of converting TOML -> Rust structures you may also want to look into using the `RustcEncodable` and `RustcDecodable` support. The support in TOML should provide nice error messages as well in terms of "expected a table but found a string". FWIW the `Decodable` support is what Cargo uses to deserialize a manifest into a Rust structure.

I've also found the Rust "schema" as an easy way to understand how the document is structured as well.

alexcrichton:
In terms of CI, I noticed that the OSX CI on Travis is allowed to fail, I wonder if now that it's green it could be changed to "expected to pass"?

rust-manifest/src/manifest.rs
Diggsey: The new manifest format isn't compatible with the automatic serialization code for TOML, or at least it wasn't at the time of writing.

rust-manifest/tests/channel-rust-nightly-example.toml
```diff
@@ -0,0 +1,52 @@
...
+    hash = "..."
+[pkg.rust-std]
+  version = "rustc 1.3.0 (9a92aaf19 2015-09-15)"
+  [pkg.rust-std.target.x86_64-unknown-linux-gnu]
+    available = true
+    url = "example.com"
+    hash = "..."
+[pkg.rust-docs]
+  version = "rustc 1.3.0 (9a92aaf19 2015-09-15)"
+  [pkg.rust-docs.x86_64-unknown-linux-gnu]
```
Diggsey: Shouldn't this be `rust-docs.target.x86_64-unknown-linux-gnu`? If so, shouldn't it cause the tests to fail?

brson: Wow, good eye! It doesn't look like there are tests exercising this value. I'll update it and expand the tests a bit.

brson:
@alexcrichton FYI I hated how much boilerplate there was in these tests, but found it *very hard* to extract the boilerplate because it contains lifetimes. I felt like this was a major refactoring fail for Rust.

brson:
Rebased and updated to address feedback.

alexcrichton:
lgtm

Diggsey:
@brson I've also found effective code reuse difficult around lifetimes - macros seem to be the only way to accomplish it most of the time (hence the NotifyHandler nastiness, rust just doesn't seem to like closures with parameters with contravariant lifetimes...).

Anyway, it all looks good!

PR #51, opened 2016-02-05
Re-implement and test Manifestation::update

brson:
Re-implement and test Manifestation::update

This is the function that keeps an installation updated on
a distribution channel.

`update` takes a manifest describing a release of Rust
(which may be either a freshly-downloaded one, or the
same one used for the previous install), as well as lists
off extension components to add and remove.

From that it schedules a list of components to uninstall
and to uninstall to bring the installation up to date.
It downloads the components' packages. Then in a Transaction
uninstalls old packages and installs new packages,
writes the distribution manifest to "rustlib/multirust-dist.toml"
and a configuration containing the component name-target
pairs to "rustlib/multirust-config.toml".

This implementation is simpler than the existing one,
mostly because it doesn't reuse the manifest to represent
what is installed, but instead stores a list of installed
components. It also doesn't perform transformations on
the mamnifest itself to calculate the install set; rather
all the calculations are done over lists of Component.

Move MockInstallerBuilder into rust-install.

### Next steps

I'm starting to get the lay of the land and have a more clear idea how to get to the end point. What I'm thinking of doing next is.

* Wire the new installation code into the program and write integration tests of the binary itself, focusing on error and notification cases.
* Replace v1 sh/msi installation code with new installation code, remove msi installer, move windows rustlib from bin to lib. Not planning on creating a transition plan; instead with the rustup rebranding we'll just start from scratch, not use the `.multirust` directory.
* Add component selection to the cli.
* Modify the install code further to not maintain compatibility with rust-install's metadata, instead using multirust-rs's own
* Refactor crate organization a bit: combine rust_install and rust_manifest into multirust_dist (I like 'dist' better here since it encompasses interop with the distribution servers, which I consider separate from the installers, leaves room for a separate multirust_install crate). Possibly reorginize how the bin and lib are layed out - right now their sources share a single src directory, I might want to separate them into multirust and multirust-cli or something.
* Rebrand to rustup: change all the internal names.
* Change the installation procedure to install to `$CARGO_HOME/bin` and from the `PATH` variable, change how cargo's home directory is treated internally.
* Work on the rustup.rs website.

That will get us more or less to the point where we can think about deploying rustup.rs and getting feedback.

After that we can look into NDK installation, a windows GUI installer, and Visual Studio integration.

Let me know how this plan looks to you. I'm most apprehensive about how to do a smooth transition from the worlds of multirust/multirust-rs/rustup.sh to rustup.rs. If you have much free time we may be able to divide up some of the work.


rust-install/src/manifest/mod.rs
```diff
@@ -22,181 +32,299 @@ impl Changes {
...
+    pub fn uninstall(&self, temp_cfg: &temp::Cfg, notify_handler: NotifyHandler) -> Result<()> {
+        let prefix = self.installation.prefix();
+
+        let mut tx = Transaction::new(prefix.clone(), temp_cfg, notify_handler);
+
+        // Read configuration and delete it
+        let rel_config_path = prefix.rel_manifest_file(CONFIG_FILE);
+        let ref config_str = try!(utils::read_file("dist config", &prefix.path().join(&rel_config_path)));
+        let config = try!(Config::parse(config_str));
+        try!(tx.remove_file("dist config", rel_config_path));
```
brson: FIXME: This also needs to delete the manifest

rust-install/src/component/package.rs
```diff
@@ -168,14 +168,35 @@ pub struct TarPackage<'a>(DirectoryPackage, temp::Dir<'a>);
...
+        // $pkgname-$version-$target. Skip that directory when
+        // unpacking.
+        try!(unpack_without_first_dir(&mut archive, &*temp_dir));
 
         Ok(TarPackage(try!(DirectoryPackage::new(temp_dir.to_owned())), temp_dir))
     }
 }
 
+fn unpack_without_first_dir<R: Read>(archive: &mut tar::Archive<R>, path: &Path) -> Result<()> {
+    let entries = try!(archive.entries().map_err(Error::ExtractingPackage));
```
nikomatsakis: Unrelated: this would be another use case for rust-lang/rfcs#243. You could write:

```rust
catch {
    let entries = archive.entries()?;
    ...
}.map_err(Error::ExtractingPackage)
```

rust-manifest/src/config.rs
```diff
@@ -9,19 +10,7 @@ pub const DEFAULT_CONFIG_VERSION: &'static str = "1";
 #[derive(Clone, Debug)]
 pub struct Config {
     pub config_version: String,
-    pub remote: Option<ConfigRemote>,
-    pub install: ConfigInstall,
```
Diggsey: Are we getting rid of these kind of configuration options for now? I think the ability to specify a custom remote is potentially very useful.

brson: We don't have to, no. I felt it was appropriate to delete them because they don't do anything now, and I didn't think it would help get the new manifests working.

Can you say more about its purpose?

rust-install/src/manifest/mod.rs
```diff
@@ -8,6 +10,14 @@ use install::InstallPrefix;
 use openssl::crypto::hash::{Type, Hasher};
 use itertools::Itertools;
 
+pub const DIST_MANIFEST: &'static str = "multirust-dist.toml";
```
brson: This should probably be named like the channel manifest, maybe `multirust-channel-manifest.toml`.

rust-manifest/src/config.rs
Diggsey: OK, no problem - can re-add them later as required if the intent was not to remove them permanently. The remote tells it how to update, ie. where to find newer versions of the manifest. For non-custom toolchains managed by multirust, it's not strictly necessary since it can figure it out from the toolchain name. However, for custom toolchains, or independent installations, it needs to know what it should track (eg. i686-pc-windows-msvc beta) and where to look for it (eg. `http://static.rust-lang.org/dist/`)

By allowing an arbitrary URL, it allows a great deal of flexibility - you could quite easily set up a custom rust distribution with extra extensions, or have a local dist server set up to avoid needing internet access.

Diggsey: I am slightly dubious about reusing `config.toml` to store the current installation state - it was intended just to be a specification of how the installation should be configured (eg. paths, remote, etc.) if that makes sense?

brson: > I am slightly dubious about reusing config.toml to store the current installation state - it was intended just to be a specification of how the installation should be configured (eg. paths, remote, etc.) if that makes sense?

I am thinking of sinking this information into `Components` though, replacing the flat `components` file with toml, since this information doesn't need to be duplicated by both subsystems. Would that arrangement work better for you?

Diggsey: Yep, that sounds great!

brson:
@Diggsey I've pushed a commit that address the build issues with this PR. I'm still working on the refactoring we discussed - moving the component list into the installer, and I'd like more time to do that in a follow up. Can we merge this when it passes?

Diggsey:
32-bit linux failed, but it was a problem with the travis script :unamused:

PR #55, opened 2016-02-07
Implement fmt::Debug on all types

kamalmarhubi:
This vastly increases ease of experimentation with the libraries.

### DIFF
--- a/rust-install/src/component/package.rs
+++ b/rust-install/src/component/package.rs
@@ -22,7 +22,7 @@ use std::fs::File;
 pub const INSTALLER_VERSION: &'static str = "3";
 pub const VERSION_FILE: &'static str = "rust-installer-version";
 
-pub trait Package {
+pub trait Package: fmt::Debug {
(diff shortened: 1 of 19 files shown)


rust-install/src/component/package.rs
```diff
@@ -31,6 +32,10 @@ pub trait Package {
                    -> Result<Transaction<'a>>;
 }
 
+pub trait PackageDebug: Package + fmt::Debug {}
+impl<T: Package + fmt::Debug> PackageDebug for T {}
```
Diggsey: Instead of introducing `PackageDebug`, can we just have `Package` imply `Debug`.

Diggsey:
LGTM with that one change!

kamalmarhubi:
Do you want me to squash?

rust-install/src/component/package.rs
kamalmarhubi: Done.

Diggsey:
No, it's fine thanks!

PR #56, opened 2016-02-08
Use OsStr::new() instead of UFCS for string literals

kamalmarhubi:


kamalmarhubi:
coverage looks a tiny bit over conservative... minus 0 isn't a big change!

Diggsey:
Thanks, coverage should be a little more relaxed now!

PR #60, opened 2016-02-09
Implement error::Error on all error types

kamalmarhubi:
This lets the `try!` macro work when using `Box<Error>` as a return
type.

### DIFF
--- a/rust-manifest/src/errors.rs
+++ b/rust-manifest/src/errors.rs
@@ -1,5 +1,6 @@
 use toml;
 
+use std::error;
 use std::fmt::{self, Display};
 
 #[derive(Debug)]
@@ -11,6 +12,25 @@ pub enum Error {
     TargetNotFound(String),
     MissingRoot,
     UnsupportedVersion(String),
+}
+
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        use self::Error::*;
+        match *self {
+            Parsing(_) => "error parsing manifest",
+            MissingKey(_) => "missing key",
+            ExpectedType(_, _) => "expected type",
+            PackageNotFound(_) => "package not found",
+            TargetNotFound(_) => "target not found",
+            MissingRoot => "manifest has no root package",
+            UnsupportedVersion(_) => "unsupported manifest version",
+        }
+    }
+
+    fn cause(&self) -> Option<&error::Error> {
+        None
+    }
 }
 
 impl Display for Error {
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,4 +1,5 @@
 use std::path::Path;
+use std::error;
 use std::fmt::{self, Display};
 
 use rust_install::{self, utils, temp};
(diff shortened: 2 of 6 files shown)


rust-manifest/src/errors.rs
```diff
@@ -13,6 +14,25 @@ pub enum Error {
...
+            ExpectedType(_, _) => "expected type",
+            PackageNotFound(_) => "package not found",
+            TargetNotFound(_) => "target not found",
+            MissingRoot => "manifest has no root package",
+            UnsupportedVersion(_) => "unsupported manifest version",
+        }
+    }
+
+    fn cause(&self) -> Option<&error::Error> {
+        None
```
kamalmarhubi: there's an opportunity here to return the first parse error for the `Parsing` variant. easily done if you think that makes sense.

Diggsey:
Why not use the existing `Display` implementation for the error description?

kamalmarhubi:
Lifetimes: return type of `Error::description()` is `&str` with the same lifetime as the error; using the `Display::fmt()` implementation necessitates creating a string which can't live long enough. This made me quite sad.

Diggsey:
I'm still not sure that this is the best way to do this - what are the downsides to creating a wrapper type which uses the display implementation on construction, eg.
```rust
struct BoxedError<E> {
    inner: E,
    message: String,
}
```

As long as the required `From` implementations exist, it should still be fairly seamless?

kamalmarhubi:
@Diggsey I'll give it a shot!

kamalmarhubi:
Thinking about this, I still think the underlying error types should implement `std::error::Error`. 
The cause is lost with this wrapping, since we don't have an `Error` bound on `E`.

As an example, `std` handles it as I did here. Eg, `ffi::NulError`: https://github.com/rust-lang/rust/blob/9cfa1916fa06fa5fa809d6d7807a6f978e35c67d/src/libstd/ffi/c_str.rs#L348-L354

I'm still going to give the suggestion a go though, and see if it works in my use case.

kamalmarhubi:
Take a look at #65 for a take on the wrapper approach.

Diggsey:
@brson Would be glad to hear your thoughts on this!

brson:
I think the approach in this patch - duplicating error messages for every variant - is unfortunately the idiomatic way to do it, so +1.

src/errors.rs
```diff
@@ -118,6 +119,35 @@ impl<'a> Display for Notification<'a> {
...
+        use self::Error::*;
+        match *self {
+            Install(ref e) => error::Error::description(e),
+            Utils(ref e) => error::Error::description(e),
+            Temp(ref e) => error::Error::description(e),
+            UnknownMetadataVersion(_) => "unknown metadata version",
+            InvalidEnvironment => "invalid environment",
+            NoDefaultToolchain => "no default toolchain configured",
+            PermissionDenied => "permission denied",
+            ToolchainNotInstalled(_) => "toolchain '{}' is not installed",
```
bluss: there's a stray {} in the string here.

kamalmarhubi: ah nice catch. Fixed.

kamalmarhubi:
@Diggsey let me know if you want me to make this a PR against master rather than new. Also I'm happy to rebase once it's got the go ahead.

kamalmarhubi:
I'm confused by the build error. It is testing against a merge commit that is not included in this PR, and that I did not create.

Diggsey:
I have no idea, looks like a github bug, given that it's happening to both travis and appveyor - try merging from master so that it's completely up to date.

kamalmarhubi:
Just rebased onto new and force pushed. I also tweaked a couple of descriptions. I also got rid of `_` matches so that new variants result in exhaustiveness errors so that proper `cause()` and `description()` implementations get added.

kamalmarhubi:
Actually this needs another few minutes, so don't review just yet.

kamalmarhubi:
ok, I think this is good to review now

Diggsey:
Great, thanks!

PR #62, opened 2016-02-13
fixed some clippy warnings

llogiq:
I left some warnings about inner items (in this case functions) after statements, because the functions are only there to distinguish between target OS and are near their call site. However, it would be possible to quell the warnings and move the functions out after the calling function.


Diggsey:
Great, thanks!

llogiq:
My pleasure.

PR #65, opened 2016-02-18
Add a wrapper type to allow converting errors to Box<Error>

kamalmarhubi:
This makes it easier to use multirust and associated crates as
libraries. This does not implement `error::Error` directly on the error
types to avoid having to almost-duplicated the `Dispaly::fmt()`
implementation in `Error::description()`. Instead, it adds a wrapper
that uses the `Display` impl to create and store the description. For
more discussion, see
  https://github.com/Diggsey/multirust-rs/pull/60

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 /.project
 /target
 /Cargo.lock
+/errors/target
+/errors/Cargo.lock
 /rust-install/target
 /rust-install/Cargo.lock
 /rust-manifest/target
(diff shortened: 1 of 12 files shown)


kamalmarhubi:
I will say I prefer the approach in #60. I think it's more idiomatic Rust error handling than this, as it uses the `Error` trait more appropriately.

Diggsey:
Closing this in favour of #60

PR #69, opened 2016-02-19
Make multirust-rs work with both v1 and v2 manifests

brson:
This wires the cli to the new installer code for both v1 and v2
distribution manifests. v1 installations are upgraded to v2
automatically.

I copied test cases from multirust.sh into tests/cli-v1.rs,
tests/cli-v2.rs and tests/cli-misc.rs. Many of them unrelated to
getting installation from the distribution server working are still
unported and marked #[ignore]. There is a lot of setup necessary for
these tests and they are quite slow. I've been testing with --release.

The metadata format is sloppy right now.

* rustlib/rust-installer-version - Still the same as rust-installer
* rustlib/components - Components has its own component list that is
  the same as rust-installer, and I think needed for v1 manifest
  compatibility
* rustlib/manifest-$component - The component manifest, per rust-installer
* rustlib/multirust-config.toml - The multirust distribution
  configuration, which currently contains its own component/target
  list per the v2 manifests, and is versioned.
* rustlib/multirust-channel-manifest.toml - A stored copy of the
  distribution manifest.

This does a metadata version bump, but doesn't add the upgrade code yet.

Next steps are probably:

* Port remaining tests
* Add upgrade code
* Add a MULTIRUST_ENABLE_EXPERIMENTAL env var that is required before installing from a v2 manifest. This is so that we can merge the v2 code back into master without worrying about whether it really works with the live manifests.
* Merge into master

At that point we've got feature parity with master, but with an entirely new backend.

After that

* Turn on v2 manifests when they are live.
* Do some refactoring to clean up the messes I've made. Remove old install code that's now unused.
* Add extension installation to the cli.
* Review the UI and installation experience. Make sure error cases work as expected.
* Move bins to `$CARGO_HOME` with upgrade.
* Rebrand to rustup and publish www.rustup.rs. Leave compatible multurust command for a while.


brson:
cc @Diggsey @alexcrichton

aturon:
\o/

rust-install/src/mock/clitools.rs
```diff
@@ -0,0 +1,392 @@
...
+/// that captures the current directory from the environment, and
+/// will race without the mutex.
+pub fn change_dir(path: &Path, f: &Fn()) {
+    lazy_static! {
+        static ref LOCK: Mutex<()> = Mutex::new(());
+    }
+    let _g = LOCK.lock();
+    let cwd = env::current_dir().unwrap();
+    env::set_current_dir(path).unwrap();
+    let _g = scopeguard::guard((), move |_| env::set_current_dir(&cwd).unwrap());
```
alexcrichton: I'd personally feel that defining a shim here with a `Drop` implementation would be easier than pulling in a whole crate to do this job, but either way's probably fine.

If this function stays it may also be best to return an RAII guard instead of taking a closure (helps indentation, etc)

rust-install/src/mock/mod.rs
```diff
@@ -1,3 +1,10 @@
+//! Mocks for testing
+//!
+//! FIXM: This belongs in some other crate
+
```
alexcrichton: Could this module be `#![test]`? I'd just want to statically prevent multirust from calling `change_dir` which is super dangerous outside of controlled environments.

rust-install/src/mock/clitools.rs
```diff
@@ -0,0 +1,392 @@
...
+    for channel in &["nightly", "beta", "stable"] {
+        change_channel_date(url, channel, date);
+    }
+}
+
+pub fn expect_ok(config: &Config, args: &[&str]) {
+    expect_stdout_ok(config, args, "");
+}
+
+pub fn expect_err(config: &Config, args: &[&str], expected: &str) {
```
alexcrichton: We may want to consider unifying the test suites here eventually with Cargo as lots of this is already implemented there as well, but not that critical for now

alexcrichton:
Nice!

rust-install/src/mock/clitools.rs
brson: Yeah, I'd like to do a pass over both programs to see what can be factored out.

rust-install/src/manifest/mod.rs
```diff
@@ -240,6 +241,89 @@ impl Manifestation {
...
+                     new_manifest: &[String],
+                     update_hash: Option<&Path>,
+                     temp_cfg: &temp::Cfg,
+                     notify_handler: NotifyHandler) -> Result<Option<String>> {
+        // If there's already a v2 installation then something has gone wrong
+        if try!(self.read_config()).is_some() {
+            return Err(Error::ObsoleteDistManifest);
+        }
+
+        let url = new_manifest.iter().find(|u| u.contains(&self.target_triple));
```
Diggsey: We should be specifically looking for `target_triple + ".tar.gz"` here, otherwise it will just find whichever file happens to be listed first for that target.

Diggsey: (This is what's causing the failures on appveyor)

rust-install/src/dist.rs
```diff
@@ -268,3 +273,84 @@ pub fn download_hash(url: &str, cfg: DownloadCfg) -> Result<String> {
...
+    let (manifest_file, manifest_hash) = if let Some(m) = manifest_dl { m } else { return Ok(None) };
+    let manifest_str = try!(utils::read_file("manifest", &manifest_file));
+    let manifest = try!(ManifestV2::parse(&manifest_str));
+
+    Ok(Some((manifest, manifest_hash)))
+}
+
+fn dl_v1_manifest<'a>(download: DownloadCfg<'a>,
+                      toolchain: &ToolchainDesc) -> Result<Vec<String>> {
+    let root_url = if let Some(ref d) = toolchain.date {
```
Diggsey: This functionality is already implemented on `ToolchainDesc` - can just call that instead.

rust-install/src/dist.rs
```diff
@@ -268,3 +273,84 @@ pub fn download_hash(url: &str, cfg: DownloadCfg) -> Result<String> {
...
+        format!("{}/{}", download.dist_root, d)
+    } else {
+        format!("{}", download.dist_root)
+    };
+    let manifest_url = format!("{}/channel-rust-{}", &root_url, toolchain.channel);
```
Diggsey: Same here

rust-install/src/install.rs
```diff
@@ -107,15 +99,20 @@ impl<'a> InstallMethod<'a> {
                 }
             }
             InstallMethod::Dist(toolchain, update_hash, dl_cfg) => {
-                if let Some((installer, hash)) = try!(dist::download_dist(toolchain,
```
Diggsey: `download_dist` is now completely unused, so we can get rid of it.

rust-install/src/dist.rs
```diff
@@ -268,3 +273,84 @@ pub fn download_hash(url: &str, cfg: DownloadCfg) -> Result<String> {
...
+            }
+        }
+        Err(e) => Err(e)
+    }
+}
+
+fn dl_v2_manifest<'a>(download: DownloadCfg<'a>,
+                      update_hash: Option<&Path>,
+                      toolchain: &ToolchainDesc) -> Result<Option<(ManifestV2, String)>> {
+    let ref manifest_url = if let Some(ref d) = toolchain.date {
```
Diggsey: Can use existing ToolchainDesc methods here too, but just appending ".toml"

rust-install/src/dist.rs
```diff
@@ -268,3 +273,84 @@ pub fn download_hash(url: &str, cfg: DownloadCfg) -> Result<String> {
...
+    let trip = try!(toolchain.target_triple().ok_or_else(|| Error::UnsupportedHost(toolchain.full_spec())));
+
+    let manifestation = try!(Manifestation::open(prefix.clone(), &trip));
+
+    let changes = Changes {
+        add_extensions: add.to_owned(),
+        remove_extensions: remove.to_owned(),
+    };
+
+    match dl_v2_manifest(download, update_hash, toolchain) {
```
Diggsey: Would be nice to have a notification for which type of manifest is going to be used, (can just add a TODO for now)

rust-install/src/mock/mod.rs
```diff
@@ -9,32 +16,35 @@ pub struct MockInstallerBuilder {
 
 // A component name, the installation commands for installing files
 // (either "file:" or "dir:") and the file paths and contents.
-pub type MockComponent = (&'static str, Vec<MockCommand>, Vec<(&'static str, String)>);
+pub type MockComponent = (String, Vec<MockCommand>, Vec<(String, Vec<u8>)>);
```
Diggsey: There are a few places where `MockInstallerBuilder`s are a constructed which haven't been updated to pass in `String`s instead of `&str`s. (See the travis build failures)

Diggsey:
Great! If you could just fix the two things causing the CI failures then I'm happy to merge this in - you can treat my other comments as just TODOs before this all gets merged back to master.

rust-install/src/mock/clitools.rs
brson: I don't think turning this into a guard will help readability - the indentation actually is *good* for readability because it indicates that within a particular block the directory is different. Without the extra block it's less obvious the environment has changed.

rust-install/src/mock/mod.rs
brson: Doesn't seem like it can. The tests don't resolve if I put `#[cfg(test)]` on it.

rust-install/src/manifest/mod.rs
brson: Good catch.

rust-install/src/dist.rs
brson: The existing method on `ToolchainDesc` does somewhat less than this function. I also don't think the toolchain descriptor should have the responsibility to download things from the internet, but should be a simple data type. Do you mind if I just delete the original method?

brson: Oh, I see you are referring to the `manifest_url` method. Yes, I'll use it.

rust-install/src/dist.rs
brson: Will do.

brson:
I've added a few more commits: the `MULTIRUST_ENABLE_EXPERIMENTAL` var, making installation from explicit stable versions ("1.0.0") work, and addressing feedback.

The only thing I didn't do was change `change_dir` per @alexcrichton because I don't think it will be an improvement.

brson:
Thanks for the reviews.

Diggsey:
Travis is still failing, but I'll merge this in and try to sort it out on `new`.

Issue #71, opened 2016-02-23
Progress and speed indication for downloads


crumblingstatue:
It's useful to know approximately how much time is left until the download is finished, and at what speed we are downloading at.

The original shell script version of multirust provides this information.

crumblingstatue:
I have 2 unsubmitted implementations of this.

1. https://github.com/crumblingstatue/multirust-rs/commit/90af77e06bdf5fd2590325d2fb3755ee4a497aa8
    This one naively implemented it using `print!` inside rust_install, then I realized rust_install is meant to be a library, so no user facing output should go there. I abandoned this branch.

2. https://github.com/crumblingstatue/multirust-rs/commit/b32ad1083826fde95ccb4857d66524849174446d
    Trying to build an interface for library clients to subscribe to download events, but it doesn't fit well into the current architecture, so I ended up using Rc<RefCell<T>>. I'm not yet sure if it's even possible to implement this without interior mutability with the current architecture.

Advice on how to go about this would be appreciated.

Diggsey:
Hi, I agree this would be a useful feature! It should be fairly straightforward to implement using the existing Notification system - I'd recommend adding a new enum variant called DownloadProgress or similar, which carries the progress information. The caller can then handle notifications of this kind as they wish.

crumblingstatue:
One problem with using the notification system is that `Notification` implements `Display`, but download progress isn't necessarily something you want to "Display". For example, a GUI client might want to set a progress bar to some value instead of printing it with {}.

In fact, not even multirust-rs will use it, because I'm planning to use `Terminal::delete_line` and `Terminal::carriage_return` in the implementation.

But if it's okay, then I will just provide "fake" Display implementations for it to satisfy the constraints.

Diggsey:
Thanks for implementing this!

Diggsey:
Actually, I'll leave this open to track the missing "time remaining" feature.

crumblingstatue:
I'm working on

crumblingstatue:
(Stupid touch devices and their inaccuracies) 
I'm working on this. 
I'm going to use the time crate, since we already implicitly depend on it

crumblingstatue:
We should try to resolve #76 before closing this issue.

vks:
I think this can be closed. #76 is orthogonal.

PR #73, opened 2016-02-24
Download progress take 3

crumblingstatue:
This implements the progress indication part of #71.

Since I don't know how to measure download speed without adding
another dependency (`time`), this only implements download progress display,
but not speed (which is less important).

# Library
The library part just adds the various notifications for the download events.
`Notification` implements `Display`, but download events aren't supposed to be
printed out normally, so the `Display` implementations for them aren't utilized right now.

# Client
To avoid having to implement architectural changes, I have once again opted for using `Rc` and friends. It shouldn't be a performance problem, but maybe someone has a better idea.

# Other
I also ran rustfmt in a follow-up commit to clean up after myself, which also affected some unrelated code.

Maybe the frequency of printing out the download events could be lessened (maybe once per second), but again, this would probably require something like the `time` crate.

### DIFF
--- a/rust-install/src/utils/mod.rs
+++ b/rust-install/src/utils/mod.rs
@@ -22,6 +22,12 @@ pub enum Notification<'a> {
     CopyingDirectory(&'a Path, &'a Path),
     RemovingDirectory(&'a str, &'a Path),
     DownloadingFile(&'a hyper::Url, &'a Path),
+    /// Received the Content-Length of the to-be downloaded data.
+    DownloadContentLengthReceived(u64),
+    /// Received some data.
+    DownloadDataReceived(usize),
+    /// Download has finished.
+    DownloadFinished,
     NoCanonicalPath(&'a Path),
 }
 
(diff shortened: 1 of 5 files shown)


crumblingstatue:
Uh-Oh! It looks like rustfmt actually messed up some code, making it no longer compile.
Gotta investigate.

crumblingstatue:
One way to avoid `Rc` and friends in the client would be to encode all necessary information in the notification itself, and construct a new `Term` every time we handle the notification.

For the former, I imagine something like `DownloadProgress(total, content_len)`. However, this would require assuming that the content_len is always available (which is not always the case?). We could also pass an optional content_length every time, but that doesn't reflect the truth either, since you only get the content_length once at the beginning, so you only need to check for absence once.

For the latter, I'm not sure how expensive it is constructing a new `Term` every time. Right now the implementation stores it using `Rc<RefCell>`.

crumblingstatue:
This is what the current implementation looks like in action: https://asciinema.org/a/3oayn89jweqvbj6t7oi8f0zzr

brson:
Looks good to me. The Rc is fine - this doesn't need to be hyper-optimized.

multirust-rs should probably not do this when it is not outputting to a tty - it will print lines and lines of updates. There's some tty detection code in [my patch](https://github.com/Diggsey/multirust-rs/pull/69) so maybe it can be fixed up later.

Diggsey:
@crumblingstatue Great work, thanks!

WRT `Rc`, it should be possible to store the download status directly as a capture in the notification handler (the notification system should accept any `FnMut`, and `FnMut`s can capture mutable value with move semantics).

However, as @brson rightly points out, it really doesn't matter in this case! Any minor improvements can be made later, so I'll merge this in now.

PR #74, opened 2016-02-25
Lock all cli integration tests

brson:
These change the directory, which can't really be avoided.
The lock not around the setup though, which is the most expensive part.


brson:
Not sure where the failure is coming from.

Diggsey:
@brson Don't worry - the failure is just because of the code coverage, which only occurs on 64-bit linux. I'm just working on a fix.

PR #79, opened 2016-02-27
Next

brson:
Add tests for `--link-local`, `--copy-local`, and `--installer`. Add a warning to `upgrade-data` about removing toolchains. Bump the version. Remove MSI code. Disable broken download progress display on windows. Disable download progress when there's no tty.

After this we should be able to merge back into master.


Diggsey:
@brson There's no need to disable the progress indicator entirely on windows: it does work in `cmd`, it just doesn't in the MinGW shell.

Diggsey:
I sent you a PR which fixes the merge conflicts and CI failures

brson:
I'll look into detecting mingw.

Diggsey:
@brson There's no need: alternative terminals such as mintty appear as pipes to the program, so the existing `stderr_isatty` function will return false already.

brson:
@Diggsey of course you are right!

Diggsey:
@brson However, for future reference I asked on IRC and it's possible to detect `mintty` or other terminals by examining the name of the Named Pipe which is connected to stdout/stderr.

PR #82, opened 2016-02-29
Merge new into master

brson:
I think it's time. I'd suggest testing this manually to verify that everything still works before merging. r? @Diggsey 

Next steps:

* Add component selection
* Clean up UI
* Move bins to `$CARGO_HOME`
* Rebrand as rustup


brson:
cc @alexcrichton

alexcrichton:
dat diff

Diggsey:
Ugh, probably won't get time to review and test this properly for a couple of days :frowning:

Any idea why the 32-bit windows-gnu build is taking so long? We may have to disable some of the tests for that target or it's never going to produce any binaries...

brson:
@Diggsey I don't know why it would be slower than the 64-bit build, no. There may be a few optimizations I can do to the test suite to speed it up.

brson:
I submitted another PR against new that should help with test times: https://github.com/Diggsey/multirust-rs/pull/84

Diggsey:
Normally running a new version with no arguments will give you an interactive prompt with the option to update. This allows you to easily download and install from the browser. However, because of the metadata version bump, it will just print out "metadata version out of date" and then immediately exit with no option to update. On windows this results in a console window flashing up and disappearing before it can be read, which isn't great.

Other than that it seems to be working as expected!

brson:
I'll look into the issue with the metadata version and updates.

@Diggsey can you explain to me how deployment of multirust-rs works?

Diggsey:
Upon merging to master, the CI services will run the test suite. In the event that all tests pass for a given target, the release binaries for that target will be committed to the `multirust-rs-binaries` repo which is a submodule of this one. The latest binaries for a given platform are therefore located at a fixed github URL (see the links in the README).

The `multirust self update` command checks these same URLs to see if the hash is newer, and if so it downloads and installs the newer verison.

In short, there is no deployment process beyond merging to master and getting the tests to pass. (This may need to change in future to avoid breakage - it is the way it is purely for convenience while I was regularly deploying minor updates)

brson:
Thanks @Diggsey. I'll follow this up with another PR soon that fixes the issue you found.

PR #84, opened 2016-03-01
Next

brson:
Add `list-targets`, `add-target`, `remove-target` commands

`list-target <toolchain>` lists all the rust-std components available
for an installed toolchain.

`add-target <toolchain> <target>` adds the rust-std component
for a given target to an installed toolchain.

`remove-target` does the opposite.

This code mostly lives in `multirust_rs::toolchain`. I think it probably should live in `rust_install::dist`, but haven't gotten done with the refactoring.

I removed some of the work done by the cli tests, and on linux here the tests go from 39s to 23s. On windows the wins should be even better.

I'll mostly be doing refactoring and cleanup for the rest of the week.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 rust-install = { version = "0.0.5", path = "rust-install" }
+rust-manifest = { version = "0.0.5", path = "rust-manifest" }
 clap = "1.4.5"
 regex = "0.1.41"
 rand = "0.3.11"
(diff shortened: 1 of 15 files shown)


brson:
cc @alexcrichton @aturon

alexcrichton:
:pineapple: 

Another way to perhaps help CI builds is to  not run `cargo build --release`, that probably takes quite awhile.

Diggsey:
@brson Wow, impressive results, the runtime on appveyor has practically halved!

@alexcrichton That's necessary otherwise it won't build the binaries for download :wink:

PR #90, opened 2016-03-09
Clarify MSVC 2015 is required

nodakai:
I just confirmed `multirust-rs.exe` required `VCRUNTIME140.dll` but not sure what was required for MSVC `rustc.exe`.  Please adjust the wording accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,9 @@ Add [multirust-rs](https://crates.io/crates/multirust-rs) or [rust-install](http
 ### Installing from binaries
 
 - [Windows GNU 64-bit installer](https://github.com/Diggsey/multirust-rs-binaries/raw/master/x86_64-pc-windows-gnu/multirust-rs.exe)
-- [Windows MSVC 64-bit installer](https://github.com/Diggsey/multirust-rs-binaries/raw/master/x86_64-pc-windows-msvc/multirust-rs.exe)
+- [Windows MSVC 64-bit installer](https://github.com/Diggsey/multirust-rs-binaries/raw/master/x86_64-pc-windows-msvc/multirust-rs.exe)<sup>[\*1](#vs2015)</sup>
 - [Windows GNU 32-bit installer](https://github.com/Diggsey/multirust-rs-binaries/raw/master/i686-pc-windows-gnu/multirust-rs.exe)
-- [Windows MSVC 32-bit installer](https://github.com/Diggsey/multirust-rs-binaries/raw/master/i686-pc-windows-msvc/multirust-rs.exe)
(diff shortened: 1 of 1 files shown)


Diggsey:
@nodakai MSVC 2015 is not required to run `multirust-rs.exe`, even for the MSVC builds: the DLL requirement you are seeing is only the MSVC runtime library, which can be obtained separately. (Through both windows update and the microsoft download site I believe) and will likely come preinstalled on future versions of windows.

MSVC is only required to compile rust programs for an `-msvc` target, and even then, any reasonably recent version will suffice.

The wording should still be updated to reflect this. It's a future goal to have an .msi installer for multirust-rs which will be able to handle installing this dependency automatically in future.

nodakai:
Well, I thought if someone already had an older VC, they should have known they could install newer VC runtimes separately...

> It's a future goal to have an .msi installer for multirust-rs which will be able to handle installing this dependency automatically in future.

I don't think a non-Admin user can install `VCRUNTIME140.dll` to `C:\SYSTEM32`, so how about bundling it together with `multirust.exe` ?

- http://stackoverflow.com/questions/19808262/can-i-bundle-msvcrt-dlls-with-my-app-without-installing-the-distributable-vc-p
- https://msdn.microsoft.com/en-us/library/windows/desktop/ms682586(v=vs.85).aspx#search_order_for_desktop_applications

alexcrichton:
It's definitely the intention that `multirust-rs.exe` is as versatile as possible, so if it doesn't run somewhere that's a bug to fix rather than something to document I believe.

Did the executable not run for you on Windows?

nodakai:
@alexcrichton To be clear, `VCRUNTIME140.dll` is required only for MSVC Multirust-rs.

image

I think it's also OK to just document the prerequisites because we cannot bundle MSVC without which MSVC Rust is useless.

But we shouldn't expect Windows to be shipped together with the VC runtime by default:
- https://blogs.msdn.microsoft.com/oldnewthing/20140411-00/?p=1273 Windows is not a Microsoft Visual C/C++ Run-Time delivery channel

alexcrichton:
Hm no this definitely shouldn't depend on any DLLs that aren't available by default (multirust does nothing fancy with the system). We should figure out how to rid ourselves of the dependency most likely.

Diggsey:
@alexcrichton Anything compiled with MSVC will depend on this DLL, there's not really anything we can do about that... Maybe if all the dependencies were pure rust with no C/C++ code linked in, and if libstd could be implemented without depending on the C runtime library, then it could be possible, but short of that I don't see how?

brson:
Cargo's windows binaries don't need this dll, so we should do whatever Cargo does. Perhaps that means using -gnu builds.

brson:
Since we don't have an immediate solution to the technical problem, I'm just going to merge this.

Issue #83, opened 2016-03-01
The command 'multirust doc' should open the main index 


ytain:
As I was coming back to learning the Rust language, and wanted to view the documentation that came with the stable/nightly toolchain, I was confused that I was presented with the documentation for the crate std, instead of the main index where I was looking for the language syntax to see what has changed since the last time I played with it.

My suggestion is to make 'multirust doc' command to open the main index, and the options given to this command like --std, --book, --nomicon etc., would open instead the specific pages instead of the main page.

I know that the multirust project can't help doing some trickery to inject some kind of backlink to the main index, in those pages opened by the options given to the command 'multirust doc', so changing the initial behavior is better for a beginner.

Another solution is to tell the multirust to set the doc command to open the particular page on default by using some command options that sets some internal variables that persists.

What do you think?

crumblingstatue:
I'm the original author of multirust-doc.

My [original proposal](https://github.com/brson/multirust/issues/55#issuecomment-103044816) was mainly made fit for my own convenience, as I was accessing the libstd documentation often, and I wanted a super quick way to do that.

But since the libstd documentation is not the only piece of documentation bundled with Rust, it doesn't make sense to open it by default, instead of the index, which contains links to *all* documentation.

I agree with the proposal to open the index by default, and have various flags for quickly accessing specific docs (`--std`, `--book`, `--nomicon`).

If no one objects, I will submit a pull request implementing this behavior.

crumblingstatue:
One thing to note: We shouldn't unnecessarily diverge in behavior from [brson/multirust](https://github.com/brson/multirust), so if this behavior gets implemented here, it should also be implemented in brson/multirust.

I'm willing to implement it in both places.

crumblingstatue:
Okay, this should be solved by #92.

brson:
Thanks @crumblingstatue

PR #92, opened 2016-03-10
Improve interface of multirust-doc

crumblingstatue:
Open the documentation index by default, and give access to more specific
docs through flags.

Ref #83

### DIFF
--- a/src/multirust-cli/cli.rs
+++ b/src/multirust-cli/cli.rs
@@ -206,13 +206,16 @@ r"Installs multirust for the current user.
 r"Opens the documentation for the currently active toolchain with the
 default browser.
 
-By default, it opens the API documentation for the standard library.
+By default, it opens the documentation index. Use the various flags to
+open specific pieces of documentation.
 "
                 )
-                .arg(Arg::with_name("all").long("all").help(r"
-             Opens the documentation overview page, which
-             gives access to all the installed documentation.
-             "))
+                .arg(Arg::with_name("book").long("book").help("The Rust Programming Language book"))
+                .arg(Arg::with_name("reference").long("reference").help("Rust language reference"))
+                .arg(Arg::with_name("std").long("std").help("Standard library API documentation"))
+                .arg(Arg::with_name("nomicon").long("nomicon").help("The Rustonomicon book"))
+                .arg(Arg::with_name("error-index").long("error-index").help("Compiler Error Index"))
+                .arg_group(ArgGroup::with_name("page").add_all(&["book", "reference", "std", "nomicon", "error-index"]))
         )
         .subcommand(
             SubCommand::with_name("which")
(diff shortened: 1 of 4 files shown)


brson:
While I do think this should open the index by default, I'm not comfortable encoding all these resources directly into multirust: the exact composition of the Rust docs is fairly fluid and will definitely change again.

I actually think the `doc` command is a bit of a misfeature for multirust - I'd much rather have a `cargo doc` subcommand that was much more aggressive about indexing and finding Rust documentation.

Can we make the minimal change of just opening the index instead of std?

crumblingstatue:
> I'm not comfortable encoding all these resources directly into multirust: the exact composition of the Rust docs is fairly fluid and will definitely change again.

Which items do you think are the most likely to change? I think that at least the book, and the standard library documentation won't go anywhere, plus I'd like to have the most convenient way possible for opening the standard library docs, so I at least would like to keep the `--std` option if nothing else.

> I actually think the doc command is a bit of a misfeature for multirust - I'd rather have a cargo doc subcommand that was much more aggressive about indexing and finding Rust documentation.

Cargo already has a `doc` subcommand, but it is for generating documentation, not for opening the bundled Rust documentation.

Also, is/should Cargo be aware of where Rust is installed? Doesn't it just call the `rustc` command in `PATH`? Cargo would need to be aware of this information in order to find the docs.

steveklabnik:
Yes, `cargo doc` is already about generating docs, and `cargo doc --open` opens the current crate's docs. I would also assume that `multirust` would be for rust's docs itself. I guess we could in theory make it more general, but it feels like a weird separation of concerns to me: `cargo` is about managing the current crate, `multirust` is about managing my Rust installation.

> Which items do you think are the most likely to change? I think that at least the book, and the standard library documentation won't go anywhere, plus I'd like to have the most convenient way possible for opening the standard library docs, so I at least would like to keep the --std option if nothing else.

Yes, I'm not aware of any intention to change at least `/std` and `/book`. The way the book gets generated may get pulled out-of-tree, but I would think that we wouldn't want to break all of these links by changing them, they've gotten a lot of google juice over the years.

Diggsey:
> it feels like a weird separation of concerns to me

I think @brson's rationale is that multirust is not tied to a specific rust version, whereas cargo is, so if the location of the documentation *were* to change, it would be trivial to tell cargo about it, whereas multirust would have to know about both the old and new locations.

However, I think it's a problem that can be solved when (if) we come to it: for example, we could just add a file to the doc component which stores a mapping from "logical name" to file path, so that when we encounter `--[logical name]` it can be resolved to a path without multirust needing to be intelligent about the rust version, beyond "does it have this file? If not, use the old defaults".

crumblingstatue:
Alright, only `--std` and `--book` are present now, and the tests pass. Anything else?

brson:
Thanks @crumblingstatue!

PR #93, opened 2016-03-10
Don't spawn a process for various operations

alexcrichton:
Use Rust!

### DIFF
--- a/src/multirust-utils/src/errors.rs
+++ b/src/multirust-utils/src/errors.rs
@@ -99,7 +99,7 @@ pub enum Error {
     CopyingDirectory {
         src: PathBuf,
         dest: PathBuf,
-        error: raw::CommandError,
+        error: io::Error,
     },
     CopyingFile {
         src: PathBuf,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Er, updating for a few other places that don't really need to spawn a process

Diggsey:
I definitely agree we should move towards more pure rust solutions. I'm slightly worried we'll run into similar issues as `fs::remove_dir` where the rust version is more fragile, or not as good at handling edge-cases as the OS-specific commands.

I feel like eventually this sort of functionality should be extracted into a separate crate, which can focus properly on robustness and can handle any edge-cases, or platform specificness - it should probably make its way into std eventually.

Could you add a test for the copy function? @brson's done amazing work getting the test coverage up this high :stuck_out_tongue:

alexcrichton:
Isn't this transitively tested by all the transaction tests? I also disagree that anything about the standard library is fragile, all of the functions in kernel32 are behaving exactly as documented. The underlying commands do some crazy set of operations to "get it to work" which is pretty dubious.

alexcrichton:
Do you have edge cases in mind? Do you have specific things you'd like to test?

Diggsey:
Some thoughts:
- For symlinks/hardlinks/junctions, need to decide what the expected behaviour is (should the target or the link itself be copied) and test that the actual behaviour matches that
- Test the expected behaviour when the target directory already exists, error or merge?
- Are file/folder attributes copied?

brson:
Closing this. We'll deal with it later.

Issue #95, opened 2016-03-11
Install command in Readme does not work


Gyscos:
The Readme mention this command:

```
git clone --depth 1 https://github.com/Diggsey/multirust-rs.git multirust-rs && cd multirust-rs && cargo run --release self install -a
```

Cargo thinks `-a` is for him, and complains about that. I guess adding a `--` might help?

```
cargo run --release -- self install -a
```

Gyscos:
Fixed in 302a059.

PR #97, opened 2016-03-13
#95 Fix install from source documentation.

waynr:
Adds separator between cargo and multirust commands to prevent cargo from taking the `-a` argument.


brson:
Thanks!

PR #106, opened 2016-03-15
Self-install and update

brson:
This isn't ready to merge yet, but it's ready for review.

[Here's what it looks like in action](http://showterm.io/31222e257b5e0b94dc7cd#fast).

There is a lot in this patch. Most of it is related to self-
installation and self-update.

It moves all the self-update code to multirust_cli::self_update, adds
tests to tests/cli-self-update.rs, and makes everything real solid.

Installation is hypothetically simple, as multirust is distributed as
a single binary, and installation mostly requires copying it into
place. There are some tricky bits though, mostly because of
workarounds to self-delete an exe on Windows.

During install (as `multirust-setup`):

* copy the self exe to $CARGO_HOME/bin
* hardlink rustc, etc to *that*
* update the PATH in a system-specific way
* run the equivalent of `multirust default stable`

During upgrade (`multirust self upgrade`):

* download multirust-setup to $CARGO_HOME/bin/multirust-setup
* run multirust-setup with appropriate flags to indicate
  this is a self-upgrade
* multirust-setup copies bins and handlinks into place. On windows
  this happens *after* the upgrade command exits successfully.

During uninstall (`multirust self uninstall`):

* Delete `$MULTIRUST_HOME`.
* Delete everything in `$CARGO_HOME`, including
  the multirust binary and its hardlinks

Deleting the running binary during uninstall is tricky
and racy on Windows.

The last step of uninstallation is to delete *the running binary*,
multirust.exe and the CARGO_HOME that contains it. On Unix, this
works fine. On Windows you can't delete files while they are open,
like when they are running.

Here's what we do:
- Copy multirust to a temporary file in
  CARGO_HOME/../multirust-gc-$random.exe.
- Open the gc exe with the FILE_FLAG_DELETE_ON_CLOSE and
  FILE_SHARE_DELETE flags. This is going to be the last
  file to remove, and the OS is going to do it for us.
  This file is opened as inheritable so that subsequent
  processes created with the option to inherit handles
  will also keep them open.
- Run the gc exe, which waits for the original multirust
  process to close, then deletes CARGO_HOME. This process
  has inherited a FILE_FLAG_DELETE_ON_CLOSE handle to itself.
- Finally, spawn yet another system binary with the inherit handles
  flag, so *it* inherits the FILE_FLAG_DELETE_ON_CLOSE handle to
  the gc exe. If the gc exe exits before the system exe then at
  last it will be deleted when the handle closes.

This is the DELETE_ON_CLOSE method from
http://www.catch22.net/tuts/self-deleting-executables

... which doesn't actually work because Windows won't really
delete a FILE_FLAG_DELETE_ON_CLOSE process when it exits.

.. augmented with this SO answer
http://stackoverflow.com/questions/10319526/understanding-a-self-deleting-program-in-c

The `.multirust` dir no longer lives in `AppData\Local` on Windows.
While that may be a better place for it, this change makes
multirust's behavior more consistent with Cargo's.

To transition from bins being in `.multirust/bin` to `.cargo/bin`,
multirust deletes those in the former on initial upgrade, when invoked
as `multirust-rs`. This includes those from the old `AppData`
location. It does *not* try to clean up the previous `PATH`
modifications, nor does it delete the `.multirust` directory.

The `multirust self install` command doesn't exist as that
functionality is entirely in `multirust-setup`.

### Still todo on this PR

* Need to wait for input after install on windows to avoid the cmd console closing.

### Next

* Change the self-update server to static.rust-lang.org
* Add the rustup frontend
* Setup rustup.rs

r? @alexcrichton cc @Diggsey


alexcrichton:
Just starting to take a look, but looks like AppVeyor is having problems and Travis is missing?

Cargo.toml
```diff
@@ -27,29 +27,36 @@ itertools = "0.4.1"
 time = "0.1.34"
 tempdir = "0.3.4"
 libc = "0.2.0"
+rand = "0.3.11"
+scopeguard = "0.1.2"
 
 [target.x86_64-pc-windows-gnu.dependencies]
```
alexcrichton: I suspect multirust only compiles on nightly, right? In that case you may want to try giving the new `[target.'cfg(windows)'.dependencies]` syntax a spin to avoid this duplication.

src/multirust-cli/common.rs
```diff
@@ -0,0 +1,134 @@
...
+    println!("");
+
+    r
+}
+
+fn read_line() -> String {
+    let stdin = std::io::stdin();
+    let stdin = stdin.lock();
+    let mut lines = stdin.lines();
+    lines.next().unwrap().unwrap()
```
alexcrichton: Maybe worth putting a TODO here to handle this error eventually? Seems like something that'll definitely be tripped at some point

src/multirust-cli/common.rs
```diff
@@ -0,0 +1,134 @@
...
+            }
+            NotificationLevel::Error => {
+                err!("{}", n);
+            }
+        }
+    }))
+
+}
+
+pub fn run_inner<S: AsRef<OsStr>>(mut command: Command,
```
alexcrichton: Another nightly feature you may want to give a spin here is to use `CommandExt::exec` on Unix perhaps?

src/multirust-cli/main.rs
```diff
@@ -52,818 +50,44 @@ fn main() {
...
-fn remove_toolchain_args(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
-    try!(get_toolchain(cfg, m, false)).remove()
-}
-
-fn default_(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
-    let toolchain = try!(get_toolchain(cfg, m, true));
-    if !try!(common_install_args(&toolchain, m)) {
-        if !toolchain.is_custom() {
-            try!(toolchain.install_from_dist_if_not_installed());
+    // The namem of arg0 determines how the program is going to behave
```
alexcrichton: "namem"

src/multirust-cli/main.rs
```diff
@@ -52,818 +50,44 @@ fn main() {
...
+    let arg0 = env::args().next().map(|a| PathBuf::from(a));
```
alexcrichton: Perhaps `args_os` to be maximally compatible?

src/multirust-cli/multirust_mode.rs
```diff
@@ -0,0 +1,366 @@
+use clap::ArgMatches;
```
alexcrichton: oh man this file was a pleasure to read

src/multirust-cli/main.rs
Diggsey: Unfortunately clap doesn't support invalid unicode last time I checked, so it wouldn't help.

src/multirust-cli/proxy_mode.rs
```diff
@@ -0,0 +1,28 @@
+use common::{run_inner, set_globals};
+use multirust::{Cfg, Result, Error};
+use multirust_utils::utils;
+use std::env;
+use std::path::PathBuf;
+
+pub fn main() -> Result<()> {
+    try!(::self_update::cleanup_self_updater());
+
+    let arg0 = env::args().next().map(|a| PathBuf::from(a));
```
alexcrichton: Perhaps this could thread through the OsStr to down below? Probably doesn't matter here anyway but may matter someday and is arguably "best practice"

src/multirust-cli/proxy_mode.rs
```diff
@@ -0,0 +1,28 @@
+use common::{run_inner, set_globals};
+use multirust::{Cfg, Result, Error};
+use multirust_utils::utils;
+use std::env;
+use std::path::PathBuf;
+
+pub fn main() -> Result<()> {
+    try!(::self_update::cleanup_self_updater());
```
alexcrichton: This'll be executed for all rustc/cargo invocations, right? Along those lines we probably don't want to do *too* much work in the setup here. I doubt this does much, however, but curious what you think?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+//! * hardlink rustc, etc to *that*
+//! * update the PATH in a system-specific way
+//! * run the equivalent of `multirust default stable`
+//!
+//! During upgrade (`multirust self upgrade`):
+//!
+//! * download multirust-setup to $CARGO_HOME/bin/multirust-setup
+//! * run multirust-setup with appropriate flags to indicate
+//!   this is a self-upgrade
+//! * multirust-setup copies bins and handlinks into place. On windows
```
alexcrichton: "handlinks"

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+//!   this happens *after* the upgrade command exits successfully.
```
alexcrichton: It's not immediately clear to me why this needs to be sequenced like this on Windows, but it's probably explained below when I get to it...

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+    let ref cfg = try!(common::set_globals(verbose));
+
+    // If this is a fresh install (there is no default yet)
+    // then install stable and make it the default.
+    if try!(cfg.find_default()).is_none() {
+        let stable = try!(cfg.get_toolchain("stable", false));
+        try!(stable.install_from_dist());
+        try!(cfg.set_default("stable"));
+    } else {
+        info!("updating existing installation");
```
alexcrichton: This doesn't seem to then update something, but perhaps I'm missing somethign as well?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+                               &mut si, &mut pi);
+
+        if r == 0 {
+            let err = io::Error::last_os_error();
+            return Err(Error::WindowsUninstallMadness(err));
+        }
+
+        // Sleep long enough to let the OS open the other process and increase
+        // its file handle count so we don't delete it before its run
+        thread::sleep(Duration::from_millis(100));
```
alexcrichton: holy cow, surely windows is better than this?


The SO answer at least didn't have this, but I guess your testing brought up that it should probably exist?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+    }
+
+    Ok(())
+}
+
+/// Run by multirust-gc-$num.exe to delete CARGO_HOME
+#[cfg(windows)]
+pub fn complete_windows_install() -> Result<()> {
```
alexcrichton: Should this be called "complete_windows_uninstall"?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+        let this_pid = GetCurrentProcessId();
+        while entry.th32ProcessID != this_pid {
+            let success = Process32Next(snapshot, &mut entry);
+            if success == 0 {
+                let err = io::Error::last_os_error();
+                return Err(Error::WindowsUninstallMadness(err));
+            }
+        }
+
+        let parent_id = entry.th32ParentProcessID;
```
alexcrichton: Wouldn't it be easier to pass this on the command line?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+        let parent = OpenProcess(SYNCHRONIZE, 0, parent_id);
+        if parent == INVALID_HANDLE_VALUE {
+            let err = io::Error::last_os_error();
+            return Err(Error::WindowsUninstallMadness(err));
+        }
+
+        defer! {{ let _ = CloseHandle(parent); }}
+
+        // Wait for our parent to exit
+        let res = WaitForSingleObject(parent, 1000);
```
alexcrichton: This is waiting at most 1 second, but shouldn't it wait indefinitely?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+        defer!{{ let _ = CloseHandle(gc_handle); }}
+
+        let mut si = mem::zeroed::<STARTUPINFOW>();
+        si.cb = mem::size_of::<STARTUPINFOW>() as DWORD;
+        let mut pi = mem::zeroed::<PROCESS_INFORMATION>();
+
+        // Create the gc exe process, which will inherit a handle to
+        // its own file. It will wait for this process to exit then
+        // delete CARGO_HOME.
+        let r = CreateProcessW(gc_exe_win.as_ptr(),
```
alexcrichton: I think that `Command` should be usable here, right?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+        si.cb = mem::size_of::<STARTUPINFOW>() as DWORD;
+        let mut pi = mem::zeroed::<PROCESS_INFORMATION>();
+
+        // Launch the new process without stdio so it can't print to the console
+        si.dwFlags = STARTF_USESTDHANDLES;
+        si.hStdInput = ptr::null_mut();
+        si.hStdOutput = ptr::null_mut();
+        si.hStdError = ptr::null_mut();
+
+        let r = CreateProcessW(ptr::null(),
```
alexcrichton: Like above, is `Command` usable here?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+    assert!(methods.len() == 1 && methods[0] == PathUpdateMethod::Windows);
+
+    use winreg::RegKey;
+    use winapi::*;
+    use user32::*;
+    use std::ptr;
+
+    let root = RegKey::predef(HKEY_CURRENT_USER);
+    let environment = try!(root.open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+                           .map_err(|_| Error::PermissionDenied));
```
alexcrichton: I may just be getting lost and confused in the error handling here, but it seems weird that the original error is getting thrown away and it's being overwritten with a "permission denied" error?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+
+    let old_path: String = environment.get_value("PATH").expect("non-unicode PATH");
```
alexcrichton: Should this return a `Result`?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+
+    let mut new_path = try!(get_cargo_home()).join("bin").to_string_lossy().to_string();
```
alexcrichton: Perhaps this could return an error instead of doing a lossy conversion?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+    let existing_rcfiles = rcfiles.into_iter()
+        .filter_map(|f|f)
+        .filter(|f| f.exists());
+
+    existing_rcfiles.map(|f| PathUpdateMethod::RcFile(f)).collect()
+}
+
+fn shell_export_string() -> Result<String> {
+    let path = format!("{}/bin", try!(canonical_cargo_home()));
+    Ok(format!(r#"export PATH="{}:$PATH""#, path))
```
alexcrichton: Could you add a comment why we prepend instead of append here?

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,1000 @@
...
+
+    let old_path: String = environment.get_value("PATH").expect("non-unicode PATH");
+
+    let mut new_path = try!(get_cargo_home()).join("bin").to_string_lossy().to_string();
+    if old_path.contains(&new_path) {
+        return Ok(());
+    }
+
+    new_path.push_str(";");
+    new_path.push_str(&old_path);
```
alexcrichton: Just confirming, but this is intentional that Windows *appends* where Unix *prepends* to PATH?

src/multirust-utils/src/raw.rs
```diff
@@ -355,16 +355,50 @@ pub fn remove_dir(path: &Path) -> io::Result<()> {
...
+// Again because remove_dir all doesn't delete write-only files on windows,
+// this is a custom implementation, more-or-less copied from cargo.
+// cc rust-lang/rust#31944
+// cc https://github.com/rust-lang/cargo/blob/master/tests/support/paths.rs#L52-L80
+fn rm_rf(path: &Path) -> io::Result<()> {
+    if path.exists() {
+        for file in fs::read_dir(path).unwrap() {
+            let ref file = try!(file).path();
+
+            if file.is_dir() {
```
alexcrichton: I've actually seen quite noticeable speed improvements in the past from using `is_dir` from the `DirEntry` rather than running another stat call here, perhaps the `is_dir` could be loaded from `try!(file.file_type()).is_dir()`?

alexcrichton:
Mega bonus points for all the platform specific hacks here, nice!

src/multirust-cli/common.rs
brson: I'll just fix it.

src/multirust-cli/common.rs
brson: Yes. I wasn't going to do it yet though. This is just refactored existing code.

brson: Oh, I'll just do it.

src/multirust-cli/proxy_mode.rs
brson: I looked into it and it looks difficult enough that I don't think it's worth it now.

src/multirust-cli/proxy_mode.rs
brson: In most cases all it's going to do is stat a single file. I could also only do this in multirust mode, but I think it'll be uncommon to run `multirust` soon after doing an update.

src/multirust-cli/self_update.rs
brson: multirust-setup can't replace the running exe until it exits on Windows. In my experiments I also couldn't mess with the hardlinks until the running exe exits.

src/multirust-cli/self_update.rs
brson: It's just indicating whether setup is in 'install mode' or 'upgrade mode'. The `installation` here is not referring to rustc but to rustup. If there weren't a bunch of other status messages printed by the other branch it might contain the line `info!("doing a fresh install")`.

src/multirust-cli/self_update.rs
brson: I have not witnessed this in testing, but I've seen several comments (including in the catch 22 post) stating that the OS will not atomically bump the file reference count after launching the process - you have to give it time.

brson: I'll add a note that it's cargo-culted.

src/multirust-cli/self_update.rs
brson: lol yeah

src/multirust-cli/self_update.rs
brson: We discussed this, and while yeah, it might be, I don't want to rewrite it. The worse part is the process ID race, which we also agreed not to fix.

src/multirust-cli/self_update.rs
brson: I'll change it.

src/multirust-cli/self_update.rs
brson: Yes, I'll change it.

src/multirust-cli/self_update.rs
brson: Yes.

src/multirust-cli/self_update.rs
brson: I'd prefer to circle back later and overhaul the error handling globally than try to address this now.

src/multirust-cli/self_update.rs
brson: Because of how errors are defined it's again just easier for now to punt instead of defining another error variant for this case.

src/multirust-cli/self_update.rs
brson: Ditto.

src/multirust-cli/self_update.rs
brson: Yes.

src/multirust-cli/self_update.rs
brson: Not at all. Good catch.

brson: Wait, both windows and unix are prepending.

src/multirust-utils/src/raw.rs
brson: Ok.

brson:
I've updated the PR as indicated.

I'll do a full pass over error handling next week perhaps.

src/multirust-cli/common.rs
```diff
@@ -0,0 +1,154 @@
...
+                error: multirust_utils::raw::CommandError::Io(e),
+            }.into())
+        }
+    }
+}
+
+#[cfg(unix)]
+fn exec(mut command: Command) -> io::Result<i32> {
+    use std::os::unix::process::CommandExt;
+    Err(command.exec())
```
alexcrichton: from our discussion tonight, if you wanna switch this back to *not* exec then I'd also be fine with that.

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,969 @@
...
+        }
+
+        defer!{{ let _ = CloseHandle(gc_handle); }}
+
+        try!(Command::new(gc_exe).spawn()
+             .map_err(|e| Error::WindowsUninstallMadness(e)));
+        
+        // The catch 22 article says we must sleep here to give
+        // Windows a chance to bump the processes file reference
+        // count. acrichto though is in disbelief and *demanded* that
```
alexcrichton: I'd be fine if you want to leave my github username here as well so I can be maximally harassed

src/multirust-cli/self_update.rs
```diff
@@ -0,0 +1,969 @@
...
+    // Tell other processes to update their environment
+    unsafe {
+        SendMessageTimeoutA(HWND_BROADCAST,
+                            WM_SETTINGCHANGE,
+                            0 as WPARAM,
+                            "Environment\0".as_ptr() as LPARAM,
+                            SMTO_ABORTIFHUNG,
+                            5000,
+                            ptr::null_mut());
+    }
```
alexcrichton: Oh I realized this unsafe block is duplicated with the one above, so maybe worth making a helper function? Doesn't seem too important.

alexcrichton:
r=me whenever this is ready, exciting!

brson:
OK, I'm crossing my fingers and merging this.

PR #107, opened 2016-03-15
Run tests in release mode

brson:
Try to avoid CI timeouts.

Also remove cli.sh/cli.bat. Everything here is covered
in the Rust tests.


alexcrichton:
r=me

PR #117, opened 2016-03-17
Don't consider HOME when searching for user home dir on windows

brson:
HOME is only set under msys, and it's set to something wrong.

Also fixes the bin path that is deleted during upgrade,
from FOLDERID_Profile to FOLDERID_LocalAppData.


brson:
This also sprinkles in some additional fixes.

brson:
cc @alexcrichton I merged this to get the fixes in but you should be aware of the change to CARGO_HOME

alexcrichton:
lgtm

PR #120, opened 2016-03-17
Prettify list-targets

brson:
Now the target for the host arch will display "(default)" and
other installed targets "(installed)".

r? @Diggsey


Diggsey:
LGTM with the merge conflict fixed :)

PR #122, opened 2016-03-17
Put process in a job on windows to kill children

brson:
It is so annoying to not have this.

cc @alexcrichton This should be in its own crate, but doesn't on its own seem worth the effort. I'm anticipating a few weeks of further refactoring next quarter.

I also didn't write tests, because bleh and you've already got this code tested in cargo. Please don't make me write tests.


alexcrichton:
r=me

PR #125, opened 2016-03-18
Fix generation of compatible linux bins on travis

brson:
Fixes https://github.com/rust-lang-nursery/multirust-rs/issues/123


alexcrichton:
I mentioned on IRC, but perhaps we could just run all tests inside the docker container? That would avoid the need for installing all the packages on the host as well. If that ends up being difficult though r=me, CI is always just a bunch of hacks :P

brson:
OK, I've sunk everything into docker.

alexcrichton:
r=me

PR #126, opened 2016-03-18
Fix setting of CARGO_HOME

brson:
The env_var::set_path method is for inserting a value into PATH,
so the value of CARGO_HOME ends up being crazy if CARGO_HOME is
already set.


brson:
I'm just merging this because it's resulting in horrible brokenness locally.

PR #130, opened 2016-03-18
Run travis on the auto branch

alexcrichton:
Also set up webhooks for bors

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,10 @@ git:
 branches:
   only:
     - master
-    - new
+    - auto
+
+notifications:
+  webhooks: http://buildbot.rust-lang.org/homu/travis
 
 sudo: required
 services:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
 branches:
   only:
     - master
-    - new
+    - auto
 
 install:
   - ps: |


alexcrichton:
@bors: retry

@brson looks like the multirust-rs test times are insanely long on appveyor unfortunately :(

I think there's an option for "cancel any builds if a branch is force pushed", and perhaps that should be checked off?

alexcrichton:
Good lord appveyor is slow

Diggsey:
Yeah it only builds one target at a time, so it's already 4x slower than travis from the get-go...

brson:
@alexcrichton It doesn't seem like anything is happening here. Which builds is homu waiting for? Are they the same builds github lists the status of here?

brson:
(I just retriggered the appveyor build of this pr)

alexcrichton:
I'm not sure what's up with this... So the two builds are:

* Appveyor - https://ci.appveyor.com/project/brson/multirust-rs/build/1.0.86
* Travis - https://travis-ci.org/rust-lang-nursery/multirust-rs/builds/117016630

The builds on the PR are different because they're not the merged commit on the `auto` branch, they're distinct.

Maybe bors didn't get the notification that appveyor finished? I'll try to investigate.

cc @Manishearth

Manishearth:
Did you enable the status webhook in settings?

Manishearth:
This definitely means that some build completion event didn't get sent to homu. Probably a webhook issue.

alexcrichton:
I don't understand how this wouldn't be receiving notifications when it's configured the same way as libc. I see a bunch of these in the homu logs, though, so maybe that's related?

```
Traceback (most recent call last):
  File "/home/rustbuild/homu/venvdir/bin/bottle.py", line 862, in _handle
    return route.call(**args)
  File "/home/rustbuild/homu/venvdir/bin/bottle.py", line 1732, in wrapper
    rv = callback(*a, **ka)
  File "/home/rustbuild/homu/homu/server.py", line 381, in github
    realtime=True,
  File "/home/rustbuild/homu/homu/main.py", line 404, in parse_commands
    if 'authzfail' in res.text:
UnboundLocalError: local variable 'res' referenced before assignment
```

alexcrichton:
Yeah the "force retry" isn't working here due to that error...

Manishearth:
Force doesn't yet work on travis and causes that error. I need to fix that.

brson:
@alexcrichton If you haven't can you rebase this against master? Builds were still uploading to multirust-rs-binaries as of 11 hours ago.

alexcrichton:
@bors: r+ 

let's... try again maybe?

bors just may not work for now...

PR #132, opened 2016-03-18
Tell cargo where rustc is when printing the version info

### LINKED ISSUE #131: Multirust should check PATH for ~/.cargo/bin on self update
I had `~/.multirust/bin` in my path, but not `~/.cargo/bin`. Trying to self update gives me:

```
$ multirust self update
downloading file from: 'https://github.com/Diggsey/multirust-rs-binaries/raw/master/x86_64-unknown-linux-gnu/multirust-rs.sha256'
downloading file from: 'https://github.com/Diggsey/multirust-rs-binaries/raw/master/x86_64-unknown-linux-gnu/multirust-rs'
Installing...
info: updating existing installation

stable revision:

rustc 1.7.0 (a5d1e7a59 2016-02-29)
Could not execute process `rustc -vV` (never executed)

Caused by:
  No such file or directory (os error 2)
(failed to run cargo)
```

If I add `~/.cargo/bin` to my path, I get an error about metadata, and then I can install the toolchains.

brson:
Cargo runs rustc during --version, and rustc may not necessarily be on
the `PATH`. during proxy operation multirust does not manipulate the
`PATH` variable itself but expects it to be set correctly externally.

Fixes #131, which is fallout from me removing the setting of `PATH` from multirust-rs as well the recent updater changes.

This is hard to test automatically because it relies on behavior of the real cargo that doesn't exist in the mocks.

r? @Diggsey cc @Luthaf

### DIFF
--- a/src/multirust-cli/common.rs
+++ b/src/multirust-cli/common.rs
@@ -131,6 +131,11 @@ pub fn show_tool_versions(toolchain: &Toolchain) -> Result<()> {
         if utils::is_file(&cargo_path) {
             let mut cmd = Command::new(&cargo_path);
             cmd.arg("--version");
+            // cargo invokes rustc during --version, this
+            // makes sure it can find it since it may not
+            // be on the `PATH` and multirust does not
+            // manipulate `PATH`.
+            cmd.env("RUSTC", rustc_path);
             toolchain.set_ldpath(&mut cmd);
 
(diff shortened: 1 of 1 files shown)


Luthaf:
Your reactivity is awesome!

Diggsey:
:+1:

PR #135, opened 2016-03-19
Whitespace fixes

brson:
Movement of println's in one bit of code caused regressions in
the display of other bits of code.


alexcrichton:
r=me

Issue #129, opened 2016-03-18
rustup CLI design


brson:
Edit: [Here's the interface as of 3/20](https://github.com/rust-lang-nursery/multirust-rs/issues/129#issuecomment-199054182). [Here's a comparison to pyenv](https://github.com/rust-lang-nursery/multirust-rs/issues/129#issuecomment-199047085).

Here's what I'm thinking on the rustup UI.

A few main principles.

* The changes are mostly going to be reorganizing the existing
  commands. The current output is generally good enough.

* The most important thing is updating the channels, so the simple
  command `rustup` will just update all the things.

* Custom toolchain management will be disconflated from
  release channel toolchain management. The `--copy-local`,
  `--link-local`, and `--installer` features will be dealt with differently.
  `rustup toolchain link <path>` will link in development toolchains
  (and they will automatically get a fallback cargo).

* Commands that work on channels use the current override by default,
  so you don't have to e.g. specify the toolchain when adding a target.
  Such commands will probably have `--channel` flags.

* We're going to try to use more sub-sub-commands, like the "multirust
  self" command, to free up the root namespace for more features and bury lesser features.

### TLDR

This is the important stuff

* `rustup`. Upgrades all tracking toolchains and self-upgrades,
  tracking toolchain statuses.
* `rustup default <toolchain>`. Sets the default toolchain, installing it if it isn't already. Same as `multirust default` but without the custom toolchain parts.
* `rustup update [toolchain]`. Updates the current toolchain or a given toolchain, installing if it isn't already. Mostly the same as `multirust update` but without custom toolchain parts.
* `rustup toolchain link`. Symlinks a toolchain. Useful for development. The same as `multirust update --link-local`.
* `rustup toolchain remove`
* `rustup toolchain list`. Lists the installed toolchains in some useful way.
* `rustup target list/add/remove`. Modify target architectures for a toolchain.
* `rustup override add <toolchain>`. Sets the toolchain for cwd. Same as `multirust override <toolchain>`.
* `rustup override remove`
* `rustup run <toolchain> [args ..]`. Runs a command with the environment of a given toolchain. Same as `multirust`.
* `rustup self uninstall`
* `rustup self update`
* `rustup self upgrade-data`

## Basic usage

After installation, the user already has a toolchain installed
("stable" by default). Many users don't need to interact with rustup
again until the next version of Rust is released. At that time they
will *upgrade Rust* by writing

```sh
$ rustup
```

[...]

PR #136, opened 2016-03-20
Implement rustup CLI

brson:
This adds the `rustup` command as described in #129.

Major changes:

* A bare `rustup` updates all tracked channels and self-updates.
* All the `--copy-local`, `--link local` and `--installer` options
  are gone. There is just `rustup channel link`, which covers the
  most common reason for creating custom toolchains.
* `add-target`, `list-targets`, and `remove-targets` are all under
  the `target` subcommand, and all operate on the current toolchain
  instead of needing to specify it explicitly.
* Override comands are all under the `override` subcommand.
* Lesser commands for modifying toolchains are under the `toolchain`
  subcommand, but I decided `multirust update` and `multirust update
  nightly` were useful enough to stay its own subcommand.
* The UI for updating all channels is changed from `multirust-rs`
  slightly: both the update status and channel revision are displayed
  in the same place.
* I've removed green from 'info' logging and used it only for
  successful updates to make it more impactful.

Note there are several ways to update now: `rustup` updates
everything, `rustup update` updates the current toolchain, and `rustup
update [toolchain]` updates a specific toolchain.

`--help` text is still missing.

Next steps are:

* get this live and get some internal feedback
* change `multirust-setup` to `rustup-setup`
* add deprecation warnings to `multirust`
* move metadata dir to `~/.rustup`
* rewrite readme and rustup.rs
* set up rustup.rs, win.rustup.rs and sh.rustup.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 multirust-dist = { version = "0.0.5", path = "src/multirust-dist" }
 multirust-utils = { version = "0.0.5", path = "src/multirust-utils" }
-clap = "1.4.5"
+clap = "2.2.1"
 regex = "0.1.41"
 openssl = "0.7.2"
 hyper = "0.7.0"
--- a/src/multirust-cli/cli.rs
+++ b/src/multirust-cli/cli.rs
@@ -1,7 +1,7 @@
 
 use clap::*;
 
-pub fn get() -> App<'static, 'static, 'static, 'static, 'static, 'static> {
+pub fn get() -> App<'static, 'static> {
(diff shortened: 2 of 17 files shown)


alexcrichton:
Oh man I'm excited!

Can you remind me what the self update workflow looks like? If this is merged and published, and then I `multirust self update`, will I get `rustup` as a command as well linked into place?

brson:
@alexcrichton Yes, when this lands then after the next self update everybody will have a `rustup` command.

alexcrichton:
:pineapple: 

r=me

brson:
There's a lot more to do here and parts of the CLI are going to change, but I'm merging this to get the forward progress.

PR #143, opened 2016-03-24
Add rustup-setup.sh

brson:
This is a script that can be curled to begin installation.

We'll host this as https://sh.rustup.rs via our nginx proxy.

Most of the boilerplate is copied from rustup.sh.

### DIFF
--- a/rustup-setup.sh
+++ b/rustup-setup.sh
@@ -1,0 +1,211 @@
+#!/bin/sh
+# Copyright 2016 The Rust Project Developers. See the COPYRIGHT
+# file at the top-level directory of this distribution and at
+# http://rust-lang.org/COPYRIGHT.
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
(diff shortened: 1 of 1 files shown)


rustup-setup.sh
```diff
@@ -0,0 +1,211 @@
...
+
+    local _url="$RUSTUP_UPDATE_ROOT/$_arch/multirust-setup$_ext"
+
+    local _dir="$(ensure mktemp -d)"
+    local _file="$_dir/rustup-setup$_ext"
+
+    if [ -e /usr/bin/printf ]; then
+	# printf is a shell builtin and won't interpret \x in posix
+	# (sh) mode, so using the actual printf exe if it exists
+	/usr/bin/printf "\x1b[36;1minfo\x1b[0m: downloading installer\n"
```
brson: This prints `info` in cyan to match the exe's behavior and make the output seamless.

alexcrichton:
r=me

maybe this is the last one we'll ever write?

PR #154, opened 2016-03-26
Remove Diggory's name from `rustup --help`

brson:
As an official project it's not appropriate to single out
individuals here.

cc @Diggsey end of an era.


Diggsey:
Maybe there should be a link to the github somewhere in the --help text instead?

brson:
@Diggsey yes, we're going to need more --help and can find a way to give credit in a less intrusive way for sure. Linking to somewhere else is a good idea.

PR #156, opened 2016-03-26
Add `rustup show`

brson:
This shows the active toolchain and the reason for selecting it.

In rbenv this is `rbenv version`, but we use the word 'toolchain'
in rustup. Unfortunately the `rustup toolchain` namespace is
taken with functions that specifically deal with toolchains,
whereas this command deals with toolchains, overrides and defaults.
That's why this isn't `rustup toolchain`.


Gyscos:
So it'd be a subset of `rustup status`?

brson:
@Gyscos if `rustup status` ever exists, yes. Right now I'm doubtful about `rustup status`.

alexcrichton:
Oops sorry thought this was an issue the first time I saw the email

PR #157, opened 2016-03-26
Deploy to s3. Add simple website.

brson:
Name the setup program `rustup-setup.exe`. Modify travis and appveyor
configs to deploy it from master to dev-static.rust-lang.org, and from
the stable branch to static.rust-lang.org.

The directory structure under s3 is

```
rustup/
  www/ - the (simple) website
  rustup-setup.sh - the curlable script
  dist/$target/
    rustup-setup$exe
    rustup-setup$exe.sha256
```

The shell script and website are deployed from x86_64-linux.

Self upgrades are directed to

> https://static.rust-lang.org/rustup/dist/$target/

The rustup-setup.sh script is accessible via sh.rustup.rs,
one of the windows bins at win.rustup.rs, and the website
at www.rustup.rs.

The website is really simple. It does platform detection
and says how to use sh.rustup.rs or win.rustup.rs.

The website is not necessarily final, but I'm probably
not going to think much more about it before asking
for testing and feedback.

Once this is deployed I will do a one-time deployment of the
bins to the old upgrade location to provide continuity.

This probably shouldn't be merged yet because it's late and I should think harder about whether it's really ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,10 @@ branches:
 sudo: required
 services:
   - docker
+
 install:
+  - sudo pip install awscli
+  - aws configure set preview.cloudfront true
   # Set git user for pushing binaries
   - git config --global user.email "diggsey@googlemail.com"
   - git config --global user.name "Diggory Blake (via Travis CI)"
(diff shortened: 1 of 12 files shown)


www/_site/index.html
```diff
@@ -0,0 +1,56 @@
...
+  </p>
+</div>
+
+<div id="platform-instructions-unknown">
+  <p>I don't recognize your platform.</p>
+  <p>
+    rustup supports Windows, Linux, and Mac OS X. If you <em>are</em>
+    on one of these platforms and are seeing this then either
+    JavaScript is disabled or this website is drunk and
+    <a href="https://github.com/Diggsey/multirust-rs/issues/new">should be reported.</a>
```
alexcrichton: This and the link below may want to be updated to the nursery

www/_site/rustup.js
```diff
@@ -0,0 +1,35 @@
+function detect_platform() {
```
alexcrichton: checked in by accident?

brson:
All fixed.

brson:
I'm going to try to push this through today.

PR #160, opened 2016-03-26
Put the target triple at the end of toolchain names

brson:
$channel-$date-$triple instead of $triple-$channel-$date. This
emphasizes that the channel is the most important information
here (it's the only thing that's required). Putting the
component of the name with potentially unbounded values at the end
makes parsing potentially saner over the long term, makes it
simpler in the future to parse toolchains without hard-coding
the exact architectures supported.

I added one smoke test for multi-host installs, but it needs a lot more attention still.

r? @Diggsey

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,13 +95,13 @@ SUBCOMMANDS:
 
 Standard toolchain names have the following form:
 ```
-[<arch>-][<os>-][<env>-]<channel>[-<date>]
+<channel>[-<date>][-<arch>][-<os>][-<env>]
 
+<channel>	= stable|beta|nightly
+<date>		= YYYY-MM-DD
 <arch>		= i686|x86_64
 <os>		= pc-windows|unknown-linux|apple-darwin
 <env>		= gnu|msvc
-<channel>	= stable|beta|nightly
-<date>		= YYYY-MM-DD
 ```
 
 Any combination of optional parts are acceptable.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -137,16 +137,16 @@ multirust-rs:
...
 - For the current directory, use a 32-bit beta build instead:
-	`multirust override i686-beta`
+	`multirust override beta-i686`
 
 - For the current directory, use a nightly from a specific date:
 	`multirust override nightly-2015-04-01`
 
 - Combine these:
-	`multirust override i686-msvc-nightly-2015-04-01`
+	`multirust override msvc-nightly-2015-04-01-i686`
```
Diggsey: This one isn't quite right :stuck_out_tongue:

src/multirust-dist/src/dist.rs
```diff
@@ -21,26 +21,26 @@ pub const UPDATE_HASH_LEN: usize = 20;
 
 #[derive(Debug)]
 pub struct ToolchainDesc {
-    pub arch: Option<String>,
-    pub os: Option<String>,
-    pub env: Option<String>,
     // Either "nightly", "stable", "beta", or an explicit version number
```
Diggsey: Is this comment accurate? Does that mean "1.2" is a valid toolchain name?

Diggsey:
This seems like a good idea.

r=me with the corrected toolchain name.

src/multirust-dist/src/dist.rs
brson: Right now the regex is very simple. It accepts these two forms: `1.9.0` or `0.10.0`, with exactly those numbers of digits. The next stable release will publish channel manifests for the version number, and it should just work to install `rustup update 1.8.0`, add/remove targets from it and everything.

brson:
Updated to address feedback and fix tests. Thanks for the review!

PR #161, opened 2016-03-26
Rename MULTIRUST_ env vars to RUSTUP_. Leave compatibility code.

brson:
Pretty self explanatory.

### DIFF
--- a/tests/cli-self-update.rs
+++ b/tests/cli-self-update.rs
@@ -5,7 +5,6 @@ extern crate multirust_utils;
 #[macro_use]
 extern crate lazy_static;
 extern crate tempdir;
-#[macro_use]
 extern crate scopeguard;
 
 #[cfg(windows)]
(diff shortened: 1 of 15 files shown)


alexcrichton:
r=me, but looks like tests are failing?

brson:
I was getting crashes accessing a `String` inside `defer!` on both MSVC targets, so I replaced all uses with `scopeguard::guard`, converting the captures to explicit arguments. Perhaps scopeguard contains some unsoundness cc @bluss @alexcrichton.

tests/cli-self-update.rs
```diff
@@ -38,8 +37,8 @@ pub fn setup(f: &Fn(&Config)) {
 
         // An windows these tests mess with the user's PATH. Save
         // and restore them here to keep from trashing things.
-        let ref saved_path = get_path();
```
bluss: This may be totally baseless, but could it be related to this uncommon use of `ref` here?

bluss:
Thank you for the ping! Scopeguard is rather simple and pure safe rust, so it would be a soundness bug in rustc.

source: https://github.com/bluss/scopeguard/blob/master/src/lib.rs

tests/cli-self-update.rs
brson: I believe I also converted it to a value binding and still got the crash, but can't be sure.

PR #164, opened 2016-03-27
Cargo fallback

brson:


alexcrichton:
appveyor builds seem to always be canceled recently? weird...

PR #167, opened 2016-03-27
More informative infos for target installation

brson:
When installing cross-target packages, indicate the target triple
in the download and installation notifications.


brson:
Error is spurious I think. Related to windows uninstallation.

PR #178, opened 2016-03-28
Produce more builds of rustup on Travis

alexcrichton:
This commit adds a few new targets for the rustup binary to follow suit with
Cargo and the compiler. This should help go that "last mile" in running rustc on
these platforms hopefully!

* `arm-unknown-linux-gnueabi`
* `arm-unknown-linux-gnueabihf`
* `armv7-unknown-linux-gnueabihf`
* `x86_64-unknown-freebsd`

All builds happen entirely within the standard `rust-slave-linux-cross`
container that Cargo and the compiler use, and all builds of rustup now compiler
OpenSSL locally as well (to follow suit with Cargo).


alexcrichton:
cc @japaric

ci/run.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -ex
+
+cargo build --release --target $TARGET
+cargo build --release -p multirust-dist --target $TARGET
+
+if [ -z "$SKIP_TESTS" ]; then
+  cargo test --release --target $TARGET
+fi
```
brson: Why does this run `cargo build --release -p multirust-dist`? `multirust-dist` has tests that should be run, so I expect to see `cargo test --release -p multirust-dist`.

Also if this is building every build with `--target` then that affects deployment. Probably the prepare-deploy-travis script needs to change to copy the files from the target-specific directory. It looks though like the docker config is overwriting `CARGO_TARGET_DIR` only on linux, so the artifacts are going to be in different places depending on platform.

.travis.yml
```diff
@@ -1,76 +1,83 @@
...
+           DOCKER=alexcrichton/rust-slave-dist:2015-10-20b
+
+    # Cross builds happen in the `rust-slave-linux-cross` image to ensure that
+    # we use the right cross compilers for these targets. That image should
+    # bundle all the gcc cross compilers to enable us to build OpenSSL
+    - os: linux
+      rust: nightly
+      env: TARGET=arm-unknown-linux-gnueabi
+           DOCKER=alexcrichton/rust-slave-linux-cross:2016-03-17b
+           SKIP_TESTS=1
```
brson: We could run these tests couldn't we?

ci/run.sh
alexcrichton: Oh oops I mis-transcribed the original script!

Also good point about the dist scripts, I'll double check to make sure it all checks out

.travis.yml
alexcrichton: I think? I've had lots of trouble in the past with QEMU's user-space emulation (e.g. not run a full emulator) with programs that use threads. In theory though we could just run the whole binary in an emulator (similar to how the libc crate is working). Think we should try that before landing here? It'd certainly make rustup the most heavily tested project after libc :P

alexcrichton:
Ok, I think the dist piece should be fixed. I threw in `CARGO_TARGET_DIR` to cover against if this moves around projects in the future, but I don't think it's strictly necessary (shouldn't affect dist scripts).

I wonder if perhaps the prepare dist scripts could run on all PRs? And then only run the actual upload on success?

brson:
> I wonder if perhaps the prepare dist scripts could run on all PRs? And then only run the actual upload on success?

FBM. Why do you want that?

I'll merge this tomorrow when I'll be around to watch the bots.

alexcrichton:
Ah yeah I just figure it'll help weed out bugs if we run as much as possible during CI. I would have caught the bug about forgetting to update the before-deploy scripts if they were running on CI, so maybe it'll help someone in the future.

brson:
If this works on master I'll push it to stable.

Issue #149, opened 2016-03-25
Keep only unique installed toolchain with (optionally) aliases instead of duplicating folders


pravic:
The following downloads the same toolchain - `i686-pc-windows-msvc-nightly` - multiply times:
```
multirust default nightly
multirust default msvc-nightly
multirust default i686-msvc-nightly
```

brson:
Thank for the report. This is intended future work. The toolchain names need to be normalized internally.

Issue #137, opened 2016-03-20
Add feature to set the default host, distribute only a single windows binary


brson:
To make the installation simple, and possibly to solve the issue of needing VC dlls, I think we should have all windows users install the same binary (probably i686-pc-windows-gnu?). We'll pick the smartest host by default, so even though we're distributing i686-pc-windows-gnu binaries rustup might default to x86_64-pc-windows-msvc.

To use a different host it may be sufficient to just use the existing support and specify `msvc` or `gnu` in all the toolchain names. If not we can add a way to set the default host.

This will require normalizing toolchain names.

Diggsey:
#421 !

brson:
Thanks @Diggsey.

PR #184, opened 2016-03-29
Resolve toolchain names early

Diggsey:
This will mean that after installing eg. `nightly-msvc`, selecting `nightly-pc-windows-msvc` will use the existing toolchain.


brson:
Sweet! Basic approach looks good to me. Once you've got it ironed out can you also add a test, perhaps to `tests/cli-misc`? An easy way to test is probably to run `rustup default nightly`, then check that `rustup default nightly-$host` is a no-op.

Diggsey:
It should be mostly finished now and I've added the test you suggested. It does break all of the tests which check the toolchain name in the output though, which is most of the tests. Changing them all to dynamically figure out the toolchain name based on the host is going to be *tedious*... A job for tomorrow!

Diggsey:
Fixes #189, #149 
Paves the way for #137

Diggsey:
Also to note: I put the symlink tests behind a feature flag on windows, which is explicitly enabled on appveyor. This allows the tests to pass locally by default, and you can pass `--features symlinks` if running as an administrator.

brson:
Damn, nice work updating all those tests.

I'm not sure how I feel about printing the complete normalized toolchain name in all these UI locations, but it does fix https://github.com/rust-lang-nursery/multirust-rs/issues/189 so I'm happy to land it and see what it looks like in practice.

r=me when the builds pass.

PR #196, opened 2016-03-29
Unused imports

brson:


alexcrichton:
I tend to like to put `#![deny(warnings)]` in crates to prevent these from landings (merges seem to always produce unused imports), but otherwise sgtm.

PR #197, opened 2016-03-29
Bump version to 0.1.1

brson:
I want to try bumping every time we push to stable to deploy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,19 +1,19 @@
 [root]
 name = "multirust-rs"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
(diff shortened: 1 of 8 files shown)


alexcrichton:
I *think* that this may need to ensure that `Cargo.lock` is updated as well now that we check it in. If I did it right the `Cargo.lock` is mounted read-only in the docker containers so when Cargo modifies it an error should get generated.

brson:
@alexcrichton how do I update `Cargo.lock`? `cargo update` does nothing.

alexcrichton:
Hm either that or a `cargo build` should suffice, but you may need to rebase on master to tell git that you're tracking changes to Cargo.lock now?

brson:
Done.

PR #200, opened 2016-03-29
Don't try to do cloudfront invalidation

brson:
It's not working. I'll revisit it some other time.

r? @alexcrichton


brson:
Fixes the build on master cc https://travis-ci.org/rust-lang-nursery/multirust-rs/builds/119351963

PR #201, opened 2016-03-29
Disable coverage reporting on travis

brson:
Also move the deployment prep to the build stage because if
'after_success' fails it does not fail the build.


alexcrichton:
@bors: r+

In... theory I think it'll work this time?

PR #223, opened 2016-03-31
Clean up help text for rustup self update

peschkaj:
This cleans up the help text for `rustup self update` to remove an extraneous "and".


brson:
Thanks!

PR #228, opened 2016-03-31
Restore a missing #[test]

brson:


alexcrichton:
Appveyor routinely takes 2 hours per build and has no parllelism. Appveyor also tests the PR, the merge commit on auto, and also the merge commit on master. That's 6 hours of testing for one PR, which means 4 PRs can land in one day. With 8 PRs in the queue *today alone* this will take at least 48 hours to land everything (much less run dist builds on appveyor).

I'm just gonna merge what's green on travis, there's no way we can reasonably block on appveyor right now it's so slow.

Diggsey:
We should disable all but one target for PRs and master, and disable builds of the auto branch entirely. It's less thorough but it's better than no testing at all on appveyor.

PR #232, opened 2016-03-31
On windows install, ask to press enter to continue

brson:
The wait_for_keypress method actually reads a whole line
for reasons I don't understand, so just change the UI
to ask for enter instead.

Supercedes https://github.com/rust-lang-nursery/multirust-rs/pull/231

### DIFF
--- a/src/multirust-cli/common.rs
+++ b/src/multirust-cli/common.rs
@@ -28,20 +28,11 @@ pub fn confirm(question: &str, default: bool) -> Result<bool> {
     Ok(r)
 }
 
-fn read_line() -> Result<String> {
+pub fn read_line() -> Result<String> {
     let stdin = std::io::stdin();
     let stdin = stdin.lock();
     let mut lines = stdin.lines();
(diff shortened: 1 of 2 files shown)


Diggsey:
`stdin` is always line-buffered when it comes from a terminal. To detect any key press you have to access the terminal directly, not via stdin (hence why I used to call out to `pause` which implements this functionality). We really need a more powerful cross-platform term library for this kind of stuff.

alexcrichton:
@bors: r+

Do we have recourse for not spawning a new terminal if run from a terminal as well? (or does that already happen?)

brson:
@alexcrichton So far I have not been able to reproduce spawning an extra terminal. It always runs in the same term for me.

Diggsey:
I can't reproduce it either, I've tried in both cmd and msys shell

alexcrichton:
Oh sorry, disregard me then. I thought someone reported that error somewhere else, but so many fixes have gone in I'm probably just getting confused.

PR #237, opened 2016-03-31
Allow host triple to be configured independently of the build triple

Diggsey:
This allows setting the `RUSTUP_OVERRIDE_HOST_TRIPLE` environment variable at build time to set a different host triple from the build triple.

This will allow us to release versions of rustup which use 64-bit or msvc toolchains by default, even if rustup itself was compiled for 32-bit gnu.

### DIFF
--- a/src/multirust-cli/self_update.rs
+++ b/src/multirust-cli/self_update.rs
@@ -481,7 +481,7 @@ fn delete_multirust_and_cargo_home() -> Result<()> {
 
         try!(Command::new(gc_exe).spawn()
              .map_err(|e| Error::WindowsUninstallMadness(e)));
-        
+
         // The catch 22 article says we must sleep here to give
         // Windows a chance to bump the processes file reference
         // count. acrichto though is in disbelief and *demanded* that
@@ -817,7 +817,7 @@ pub fn prepare_update() -> Result<Option<PathBuf>> {
     }
 
     // Get host triple
-    let triple = dist::get_host_triple();
+    let triple = dist::TargetTriple::from_host();
 
     let update_root = env::var("RUSTUP_UPDATE_ROOT")
         .unwrap_or(String::from(UPDATE_ROOT));
--- a/src/multirust-dist/src/dist.rs
+++ b/src/multirust-dist/src/dist.rs
@@ -81,11 +81,16 @@ impl TargetTriple {
     }
 
     pub fn from_host() -> Self {
-        let (arch, os, env) = get_host_triple_pieces();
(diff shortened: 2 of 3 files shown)


brson:
This is a good solution to the problem of host defaults on Windows.

Another thing we could do is have an option to set the default, like `rustup set-host <triple>`, not that I particularly want to add another command for this.

Do we know that we can't get rid of the msvc dll dependency the msvc builds have? If we could then we wouldn't need to do this.

What I don't want to do is give newbies installation options. On Windows we have to say 'download this one exe and that's all you need'. If we provide other binaries for people to find that's fine, but having to explain ABIs to newbies before they've even installed Rust is super bad.

I'd like to believe that it's fine to just default to i686-pc-windows-gnu, that `rustup default stable-msvc` is not too burdensome for msvc-oriented developers to deal with, and no extra accommodations need be made for alternate-host installs. But I also doubt that's a position that will not stand.

This is somewhat related to the UI issues and what we print for the toolchain name. Now that we're normalizing toolchain names the UI displays them in full. We might instead display a shortened name without the targets, but only when the target == host. That way new and casual users don't have to care about which architecture their on until it matters, and the UI is cleaner. (We might instead put the target in the update status message where it currently also displays the version).

Diggsey:
> Do we know that we can't get rid of the msvc dll dependency the msvc builds have? If we could then we wouldn't need to do this.

This may be possible in future, but not any time soon AFAICT.

> Another thing we could do is have an option to set the default, like rustup set-host <triple>, not that I particularly want to add another command for this.

Yes, I also want such functionality. It may be sufficient to switch to a toml-based config file and allow the user to edit that directly.

> What I don't want to do is give newbies installation options. On Windows we have to say 'download this one exe and that's all you need'. If we provide other binaries for people to find that's fine, but having to explain ABIs to newbies before they've even installed Rust is super bad.

We should definitely provide both 32-bit and 64-bit options, users are used to making that decision, as it's an option for pretty much any binary release. The wording should be chosen to make it clear that the decision is based on your OS's bit-ness and doesn't restrict which toolchains you can install.

My preference is to also provide -msvc host builds, compiled with -gnu to avoid the DLL dependency, also available but much less prominent, so that you can get them if you know what you're looking for.

Personally I think we should add this functionality even if we don't end up using it in our own releases: the host triple is logically different from the build triple and we shouldn't tie them together. I can easily imagine someone building rustup from source wanting to customize this.

brson:
I agree this change is reasonable independent of other concerns.

@bors r+

Issue #218, opened 2016-03-31
Allow the user to tell rustup-setup which toolchain to install initially


jimmycuadra:
Right now when you run `rustup-setup` on a fresh machine, it automatically installs the stable toolchain. If the user knows they are not going to be using stable, this is a waste of time and bandwidth. `rustup-setup` should take some sort of option to tell it which toolchain to install.

brson:
Probably `--default-toolchain`. `rustup-setup.sh` needs to also be modified to pass arguments through.

Related to https://github.com/rust-lang-nursery/multirust-rs/issues/170

jimmycuadra:
#245 resolved this. Thanks!

PR #245, opened 2016-04-01
Add --default-toolchain option to rustup-setup

magni-:
WIP for #218 

This doesn't touch `rustup-setup.sh` yet.

### DIFF
--- a/src/multirust-cli/setup_mode.rs
+++ b/src/multirust-cli/setup_mode.rs
@@ -21,13 +21,19 @@ pub fn main() -> Result<()> {
              .help("Enable verbose output"))
         .arg(Arg::with_name("no-prompt")
              .short("y")
-             .help("Disable confirmation prompt."));
+             .help("Disable confirmation prompt."))
+        .arg(Arg::with_name("default-toolchain")
+             .long("default-toolchain")
+             .takes_value(true)
+             .possible_values(&["stable", "beta", "nightly"])
+             .help("Choose a default toolchain to install"));
 
     let matches = cli.get_matches();
     let no_prompt = matches.is_present("no-prompt");
     let verbose = matches.is_present("verbose");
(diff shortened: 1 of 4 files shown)


src/multirust-cli/setup_mode.rs
```diff
@@ -21,13 +21,20 @@ pub fn main() -> Result<()> {
              .help("Enable verbose output"))
         .arg(Arg::with_name("no-prompt")
              .short("y")
-             .help("Disable confirmation prompt."));
+             .help("Disable confirmation prompt."))
+        .arg(Arg::with_name("default-toolchain")
+             .short("d")
```
brson: Can we leave off the short option for now? Most flags don't have short options now, and I'd like to think more generally about how they should be used before adding them in bits and pieces.

src/multirust-cli/setup_mode.rs
```diff
@@ -21,13 +21,20 @@ pub fn main() -> Result<()> {
...
+             .long("default-toolchain")
+             .takes_value(true)
+             .possible_values(&["stable", "beta", "nightly"])
```
brson: This is probably ok for now. There are other possible values, but few are likely to want them as their default.

src/multirust-cli/setup_mode.rs
brson: I guess the other option we're likely to add here soon is something like `--no-modify-path`. Not sure how that will affect short flags.

brson:
lgtm except I'd like to remove `-d` for now

src/multirust-cli/setup_mode.rs
magni-: Sounds good :+1:

src/multirust-cli/setup_mode.rs
magni-: Yeah I went with those three to start, but this could (should?) be extended to cover any build number in the future

PR #252, opened 2016-04-01
Make `TargetTriple` a newtype String

### LINKED ISSUE #248: Can't update because of 'invalid target triple'
**If you are using rustup you are going to hit this. You can still run `rustup self update` to update rustup; you just won't be able to update your toolchains until this is fixed. I'll get a fix out as soon as I can.**

Today when I update I see

```
error: invalid target triple: 'i586-pc-windows-msvc'
```

`i586-pc-windows-msvc` is a new host platform. For some reason this string is being converted internally to a `TargetTriple`, and that conversion only supports a fixed set up targets.

The target triple logic needs to be changed to support more-or-less arbitrary strings.

brson:
Target triples are just arbitrary strings with some rough heuristics for parsing. The Rust build servers can send us builds for any target, not just those we know about, so we can't assume much about what they are. This patch converts `TargetTriple` to a simple newtype string so that parsing a manifest containing unknown triples works.

Fixes #248.

r? @Diggsey

### DIFF
--- a/src/multirust-cli/multirust_mode.rs
+++ b/src/multirust-cli/multirust_mode.rs
@@ -324,7 +324,7 @@ fn add_target(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
     let toolchain = try!(cfg.get_toolchain(toolchain, false));
     let new_component = Component {
         pkg: "rust-std".to_string(),
-        target: try!(TargetTriple::from_str(target)),
+        target: TargetTriple::from_str(target),
     };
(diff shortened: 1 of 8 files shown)


brson:
Failure looks spurious.

Diggsey:
r=me, although it would be nice to have some kind of validation - maybe that belongs in an independent crate which can match llvm's target triple parsing...

Diggsey:
The only failure seems spurious, so I'm just going to merge this already... Should I be waiting for @bors to do something instead?

PR #258, opened 2016-04-02
dist: correctly detect the ARM Linux environment

japaric:


japaric:
r? @brson 

should fix the (b) problem I mentioned [here](https://botbot.me/mozilla/rust-tools/2016-04-02/?msg=63361905&page=1) (disclaimer: untested). As for (a), it's okay if rustup install the toolchain for `arm-unknown-linux-gnueabihf` by default, if it can later be changed to the faster armv7 one.

brson:
bors is still confused

PR #259, opened 2016-04-02
Adds formatting to strip default UNC verbatim paths on Windows machines.

peschkaj:
This should clean up output on Windows machines to remove default UNC verbatim path (`\\?`) on Windows machines. The UNC path will still be present if a rust override is set up a remote share since `util::format_path_for_display` only checks for the `\\?\` literal.

### DIFF
--- a/tests/cli-exact.rs
+++ b/tests/cli-exact.rs
@@ -117,10 +117,16 @@ r"",
 fn list_overrides() {
     setup(&|config| {
         let cwd = std::fs::canonicalize(env::current_dir().unwrap()).unwrap();
+        let mut cwd_formatted = format!("{}", cwd.display()).to_string();
+        
+        if cfg!(windows) {
+            cwd_formatted = cwd_formatted[4..].to_owned();
+        }
+        
         let trip = this_host_triple();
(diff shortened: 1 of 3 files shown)


tests/cli-exact.rs
```diff
@@ -117,10 +117,19 @@ r"",
...
     setup(&|config| {
         let cwd = std::fs::canonicalize(env::current_dir().unwrap()).unwrap();
+        let cwd_fmt = format!("{}", cwd.display()).to_string();
+        
+        let unc_present = cwd_fmt.find(r"\\?\");
+    
+        let cwd_formatted = match unc_present {
+            None => cwd_fmt,
+            Some(_) => cwd_fmt[4..].to_owned(),
+        };
```
brson: Instead of adding this logic that mirrors the implementation, can you make it so if `cfg!(windows)` it unconditionally strips the 4 bytes? That assumes the test suite will never be run on the network, but that's fine.

peschkaj: :+1: Will modify, squash, and push.

brson:
Thanks!

Issue #261, opened 2016-04-04
[win] rustup setting wrong registry type for PATH


pravic:
[Here](https://github.com/rust-lang-nursery/multirust-rs/blob/master/src/multirust-cli/self_update.rs#L688). winreg [always uses](https://github.com/gentoo90/winreg-rs/blob/master/src/types.rs#L66-L82) `REG_SZ` type while environment often contains `REG_EXPAND_SZ`.

For example:
````cmd
CARGO_HOME=...
PATH=%CARGO_HOME%\bin;%PATH%
```

So after installing rustup, `PATH` environment is broken and cannot be expanded properly.

brson:
This seems quite bad. If I understand you right at install rustup is *changing* the registry type from `REG_EXPAND_SZ` to `REG_SZ`, which breaks existing entries.

brson:
I wonder if we need to be aware of anything when reading a `REG_EXPAND_SZ` key - we don't want to 'expand' it then write back the expanded version.

brson:
Working on a patch.

PR #262, opened 2016-04-04
Don't panic when HKEY_CURRENT_USER\Environment\PATH doesn't exist

petrochenkov:
This was the reason of panic in https://github.com/rust-lang-nursery/multirust-rs/issues/181
While system-global `PATH` almost certainly exists, user-specific `PATH` has high chances of not existing.

Also, remove the `PATH` variable during uninstallation, if rustup path was the last component in it.


brson:
Oh, that's a different panic than I expected. Thanks for finding it.

I'm going to pull this into another branch to fix #261.

brson:
Superseded by https://github.com/rust-lang-nursery/multirust-rs/pull/264

PR #263, opened 2016-04-04
Use REG_EXPAND_SZ for HKCU/Environment/PATH

### LINKED ISSUE #261: [win] rustup setting wrong registry type for PATH

brson:
Fixes #261

Also throw in some compatibility code to automatically fix this
on machines we've broken.

### DIFF
--- a/src/multirust-cli/main.rs
+++ b/src/multirust-cli/main.rs
@@ -59,9 +59,9 @@ fn run_multirust() -> Result<()> {
         return Err(Error::InfiniteRecursion);
     }
 
-    // Map MULTIRUST_ env vars to RUSTUP_
+    // Do various things to clean up past messes
     // FIXME: Remove this soon to get it out of the proxy path
-    make_environment_compatible();
+    do_compatibility_hacks();
 
     // The name of arg0 determines how the program is going to behave
     let arg0 = env::args().next().map(|a| PathBuf::from(a));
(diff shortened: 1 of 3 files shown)


brson:
cc @pravic

brson:
Oh, this isn't quite ready.

brson:
Now ready.

brson:
My windows installation is suddenly broken, so I would hold off on merging this.

brson:
Definitely do not merged. I think I trashed my registry.

brson:
Somehow I filled HKCU\Environment\PATH with 900MB of garbage. I'm just going to retract this while I figure it out...

Diggsey:
Have a look at the winreg crate source for how to interact with raw values:
1) _SZ values are ucs2 because the `winreg` crate use the wide/unicode apis for accessing the registry, so you can't just stuff the data field into a &str
2) Registry vaues should be null terminated

I'd recommend implementing `FromRegValue`/`ToRegValue` for a newtype of `OsString` rather than trying to use the raw APIs directly. (The newtype's purpose being to differentiate `REG_EXPAND_SZ` from `REG_SZ` strings)

pravic:
@Diggsey `REG_EXPAND_SZ` usually used for file paths, so it is reasonable to implement `ToRegValue` for the `std::path::Path`. But it is suggestion for `winreg` crate better, I guess.

Diggsey:
@pravic It's also used for arrays of paths (which is the case here) - a newtype would therefore be both more explicit and more powerful (for example it could provide methods to expand itself based on the current environment), although I agree it's a feature that belongs in the `winreg` crate.

PR #264, opened 2016-04-05
Multiple fixes for windows registry handling

brson:
This includes @petrochenkov's fix for non-existant HKCU\Environment\PATH, as well as fixes for #261, wherein we were setting this key to be a `REG_SZ` type instead of `REG_EXPAND_SZ`.

My second commit adds tests for more scenarios brought up by @petrochenkov's handling of non-existing registry keys.

### DIFF
--- a/src/multirust-cli/main.rs
+++ b/src/multirust-cli/main.rs
@@ -59,9 +59,9 @@ fn run_multirust() -> Result<()> {
         return Err(Error::InfiniteRecursion);
     }
 
-    // Map MULTIRUST_ env vars to RUSTUP_
+    // Do various things to clean up past messes
     // FIXME: Remove this soon to get it out of the proxy path
-    make_environment_compatible();
+    do_compatibility_hacks();
 
     // The name of arg0 determines how the program is going to behave
     let arg0 = env::args().next().map(|a| PathBuf::from(a));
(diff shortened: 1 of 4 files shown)


brson:
The problem in https://github.com/rust-lang-nursery/multirust-rs/pull/263 was failure to null-terminate, which really messes things up!

Diggsey:
This is still going to bust any PATH which contains invalid unicode, which is unfortunate. When I get time I'll implement the solution using OsString which I mentioned on the previous PR.

brson:
[Here's a bug about the unicode handling](https://github.com/rust-lang-nursery/multirust-rs/issues/265).

brson:
Yeah, blanking the regkey if it's non-unicode is bad. Should probably be fixed in this pr.

src/multirust-cli/main.rs
```diff
@@ -129,3 +134,32 @@ fn make_environment_compatible() {
...
+fn fix_windows_reg_key() {
+    use winreg::RegKey;
+    use winreg::enums::RegType;
+    use winapi::*;
+
+    let root = RegKey::predef(HKEY_CURRENT_USER);
+    let environment = root.open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE);
+
+    if environment.is_err() { return }
+    let environment = environment.expect("");
```
alexcrichton: Could this be:

```rust
let environment = match environment {
    Ok(e) => e,
    Err(..) => return,
};
```

alexcrichton:
r=me with the unicode bits fixed up (that you mentioned)

src/multirust-cli/main.rs
brson: Yes, but I don't like that it takes four lines to do an early return.

brson: I'll change it anyway ... *grumble*

brson:
Updated for feedback.

Now when rustup encounters a non-unicode PATH on windows during install or uninstall it will print a warning and not attempt to modify it.

alexcrichton:
(I'd just merge this, I doubt bors will ever get around to finishing)

PR #282, opened 2016-04-06
Add PATH to ~/.profile instead of ~/.*rc

### LINKED ISSUE #211: Maybe `~/.profile` is where PATH modifications should go
@Diggsey had multirust-rs installing path modifications to `~/.profile` and I changed it because `~/.profile` is only read for login shells, and that seemed wrong since most shells are not login shells. I'm happy to change this to do the right thing, but I'd like to understand better why it's ok not to set the `PATH` for non-login shells.

[Source](https://internals.rust-lang.org/t/beta-testing-rustup-rs/3316/41?u=brson).

brson:
No transition plan for the old scheme.

Fixes #211

### DIFF
--- a/src/multirust-cli/self_update.rs
+++ b/src/multirust-cli/self_update.rs
@@ -79,19 +79,9 @@ macro_rules! pre_install_msg_unix {
     () => {
 pre_install_msg_template!(
 "This path will then be added to your `PATH` environment variable by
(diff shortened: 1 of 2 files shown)


alexcrichton:
Travis errors look legit but unrelated to this PR. I'll send a PR.

PR #289, opened 2016-04-07
Initial telemetry implementation

peschkaj:
This implements basic telemetry, as outlined in rust-lang-nursery/multirust-rs#254

- Telemetry can be enabled/disabled with `rustup telemetry on|off`
- Telemetry configuration is stored in `~/.multirust/telemetry`
- Telemetry is stored in `~/.multirust/telemetry-YYYY-MM-DD`


brson:
Thanks for the quick turnaround @peschkaj!

Here are some initial impressions:

First, telemetry will need to be in the `multirust` library, not `multirust-cli` so it will be available to other clients, e.g. IDEs. I'd suggest just taken the functionality you've implemented here and just lowering it by one 'layer', so that telemetry is measured in e.g. `Toolchain::add_component` etc.

This is going to sound mean, but I'd like you to switch from serde to rustc-serialize for the serialization, both so we don't add more nightly dependencies, but also because serde still breaks a lot. rustc-serialize should work fine for our purpose.

On where telemetry 'lives', I'd say you've got the right idea by putting the `TelemetryEvent` enum in utils - telemetry may be needed from several rustup crates and utils has no dependencies. I'd further suggest putting all telemetry stuff in its own `telemetry` module; that includes the logic for dealing with filesystem, so your 3 function on `Cfg` would just just small pass-troughs to functions in `utils::telemetry`.

I'll also go and leave some individual comments on the patch.

peschkaj:
Many thanks, none of it sounds mean at all.

I was worried about the `serde` dependency and what you say makes sense. I
think `serde` may be a leftover from when I was attempting to record
timestamps as an ISO8601 string before I punted and just used seconds since
epoch.

On Thu, Apr 7, 2016 at 1:44 PM Brian Anderson <notifications@github.com>
wrote:

> Thanks for the quick turnaround @peschkaj <https://github.com/peschkaj>!
>
> Here are some initial impressions:
>
> First, telemetry will need to be in the multirust library, not
> multirust-cli so it will be available to other clients, e.g. IDEs. I'd
> suggest just taken the functionality you've implemented here and just
> lowering it by one 'layer', so that telemetry is measured in e.g.
> Toolchain::add_component etc.
>
> This is going to sound mean, but I'd like you to switch from serde to
> rustc-serialize for the serialization, both so we don't add more nightly
> dependencies, but also because serde still breaks a lot. rustc-serialize
> should work fine for our purpose.
>
> On where telemetry 'lives', I'd say you've got the right idea by putting
> the TelemetryEvent enum in utils - telemetry may be needed from several
> rustup crates and utils has no dependencies. I'd further suggest putting
> all telemetry stuff in its own telemetry module; that includes the logic
> for dealing with filesystem, so your 3 function on Cfg would just just
> small pass-troughs to functions in utils::telemetry.
>
> I'll also go and leave some individual comments on the patch.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/rust-lang-nursery/multirust-rs/pull/289#issuecomment-207080489>
>
-- 

Jeremiah Peschka

src/multirust-cli/proxy_mode.rs
```diff
@@ -26,6 +26,10 @@ pub fn main() -> Result<()> {
 fn direct_proxy(cfg: &Cfg, arg0: &str) -> Result<()> {
     let cmd = try!(cfg.create_command_for_dir(&try!(utils::current_dir()), arg0));
     let args: Vec<_> = env::args_os().collect();
+
+    if (arg0 == "rustc" || arg0 == "rustc.exe") && cfg.telemetry_enabled() == true {
+        return run_rustc_with_telemetry(cmd, &args, &cfg);
+    }
```
brson: It looks to me like all this logic, including `common::run_inner` and `common::run_rustc_with_telemetry` will need to move into `multirust`, perhaps into something like `Cfg::run_command_for_dir`.

brson: (because telemetry needs to be done in the lib).

src/multirust-cli/rustup_mode.rs
```diff
@@ -200,6 +208,36 @@ fn default_(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
 }
 
 fn update(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
+    if cfg.telemetry_enabled() == true {
+        info!("telemetry enabled, updating with telemetry!");
```
brson: No logging here needed.

src/multirust-cli/rustup_mode.rs
```diff
@@ -200,6 +208,36 @@ fn default_(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
...
+            common::log_telemetry(te, cfg);
+            Ok(())
+        }
+        Err(e) => {
+            let te = utils::TelemetryEvent::ToolchainUpdate { toolchain: tc_name, success: false };
+            common::log_telemetry(te, cfg);
+            Err(e)
+        }
+    }
+}
```
brson: This logic can go into `Toolchain::install_from_dist` I think.

src/multirust-cli/rustup_mode.rs
```diff
@@ -280,6 +327,25 @@ fn target_add(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
...
+
+    common::log_telemetry(utils::TelemetryEvent::TargetAdd { toolchain: toolchain,
+                                                             target: target,
+                                                             success: success,
+                                                           },
+                                                           &cfg);
+
+    result
+}
+
```
brson: This logic into `Toolchain::add_component`

brson:
OK, I've given lots of feedback on what I'd consider "architectural" issues. Let's get the basic organization squared away and move from there.

peschkaj:
Thanks for the quick and thorough review! I'll get on this.

src/multirust-cli/main.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(custom_derive, plugin)]
```
brson: This shouldn't be needed without serde.

src/multirust-cli/multirust_mode.rs
```diff
@@ -326,7 +326,7 @@ fn add_target(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
         pkg: "rust-std".to_string(),
         target: TargetTriple::from_str(target),
     };
-    try!(toolchain.add_component(new_component));
+    try!(toolchain.add_component(new_component, &cfg));
```
brson: We can't pass `Cfg` to `Toolchain` - it's coupling these two types in the wrong direction. I'll try to offer suggestions how to change it.

src/multirust-cli/rustup_mode.rs
```diff
@@ -374,3 +383,9 @@ fn self_uninstall(m: &ArgMatches) -> Result<()> {
 
     self_update::uninstall(no_prompt)
 }
+
+fn telemetry(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
+    let telemetry_string = m.value_of("enabled").unwrap();
+    
+    cfg.set_telemetry(telemetry_string)
```
brson: Instead of passing a string to `set_telemetry` please convert this to a bool first. I realize you'll have to convert that bool back to something else to store in the telemetry file, but I'd rather have the types still.

src/multirust-cli/multirust_mode.rs
peschkaj: That makes sense. The only thing `Cfg` is used for in `log_telemetry` is to get the path of the `multirust_dir` for logging. Would it be more reasonable to pass a `PathBuf` through as some kind of log_destination parameter?

src/multirust/telemetry.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+#[derive(RustcDecodable, RustcEncodable, Debug)]
+struct LogMessage {
+    log_time_s: i64,
+    event: TelemetryEvent,
+}
+
+pub fn log_telemetry(event: TelemetryEvent, cfg: &Cfg) {
+    let ln = LogMessage { log_time_s: time::get_time().sec, event: event };
```
brson: `get_time` returns the local time, but I'd like to have UTC. I *think* you can get this with `now_utc` then convert to a `Timespec` with `to_timespec` to get the time since the unix epoch. I'd rather store the typed `Timespec` in `LogMessage` than seconds since it's typed and has better than seconds resolution.

src/multirust/telemetry.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    let json = json::encode(&ln).unwrap();
+
+    let now = time::now_utc();
+    let filename = format!("telemetry-{}-{:02}-{:02}", now.tm_year + 1900, now.tm_mon, now.tm_mday);
+
+    let _ = utils::append_file("telemetry", &cfg.multirust_dir.join(&filename), &json);
```
brson: Instead of putting these directly in the multirust directory can we emit the logs to `.multirust/telemetry/log-$date.json` or something like that. Keep the main directory uncluttered.

This will require putting the settings file somewhere else. The question of what to do with the telemetry settings is tricky because when running proxies we have to load that setting *fast*, but there's also a good chance that telemetry will have more configuration knobs in the future.

I *suspect* that deserializing a telemetry file every time we run rustc would be prohibitive for performance - in the common case we want to do as little deserialization of configuration as possible. Perhaps instead of storing "on" and "off" in `~/.multirust/telemetry` we can instead just use the presence or absence of a file as the flag indicating that telemetry is on. In other words, what if we had an empty file called `~/.multirust/telemetry-on`, whose presence alone indicates that telemetry is on?

src/multirust-cli/rustup_mode.rs
peschkaj: Makes sense - there's less chance of writing "potato" to a telemetry file this way.

src/multirust/telemetry.rs
peschkaj: :open_mouth: With as much as I've been using `now_utc` and various `Timespec` functions in example code I've written, I should've caught this. Will fix shortly.

src/multirust-cli/multirust_mode.rs
brson: Ok, after looking at `telemetry` it seems the reason `Cfg` needs to be passed around is to get the value of `cfg.multirust_dir`. To refactor this so `Toolchain` doesn't need to be passed `Cfgs` I suggest the following:

* Add a `telemetry::Telemetry` type that contains a `telemetry_dir` field. When `Toolchain::new` builds itself from a `Cfg`, also build a `Telemetry` type and stash it in `Toolchain`. Then convert `log_telemetry` function to a `Telemetry::log` method.

src/multirust/command.rs
```diff
@@ -0,0 +1,109 @@
...
+
+fn run_command_for_dir_without_telemetry<S: AsRef<OsStr>>(cmd: Command, args: &[S]) -> Result<()>  {
+    let output = bare_run_command_for_dir(cmd, &args);
+
+    match output {
+        Ok(out) => {
+            let _ = io::stdout().write(&out.stdout);
+            let _ = io::stdout().flush();
+            let _ = io::stderr().write(&out.stderr);
+            let _ = io::stderr().flush();
```
brson: In the case where we are not collecting telemetry, let's not capture output like this, in order to remove all the overhead we can, so keep this code path using `Command::status` instead of `Command::output`.

src/multirust/command.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    let ms = (duration.as_secs() as u64 * 1000) + (duration.subsec_nanos() as u64 / 1000 / 1000);
+
+    match output {
+        Ok(out) => {
+            let exit_code = out.status.code().unwrap_or(1);
+
+            let errors = match out.status.success() {
+                true => None,
+                _ => Some(String::from_utf8_lossy(&out.stderr).to_string()),
```
brson: Instead of capturing all of stderr, this needs to parse out strings that match `[Ennnn]` exactly and return a sorted and uniqified list of "Ennnn" strings (not integers - it's possible the "E" is significant). We can't return all of stderr because it contains identifiying info.

src/multirust/command.rs
```diff
@@ -0,0 +1,109 @@
...
+            };
+
+            let _ = io::stdout().write(&out.stdout);
+            let _ = io::stdout().flush();
+            let _ = io::stderr().write(&out.stderr);
+            let _ = io::stderr().flush();
```
brson: This simple method of capturing output is not going to be sufficient because it delays output until rustc has completed. Instead we're going to have to set up the `Command` to inherit stdout and pipe stderr to our own handle. Then we can loop over reads of our stderr pipe, parse out the error numbers, then print the buffer to the real stderr. This will have the additional benefit of not adding extra overhead to stdout since we're only capturing stderr.

brson:
OK, I left a lot of feedback and requests for change, but it's all heading in the right direction. Looking good.

src/multirust/command.rs
peschkaj: Leaving this here as a note for myself - http://stackoverflow.com/questions/31992237/how-would-you-stream-output-from-a-process

peschkaj: I implemented a partial fix for this, but after piping `stderr`, the formatting is stripped. I tried a tentative fix using a `Vec<u8>` instead of a `String`, but I ran into the same problems. I haven't been able to find an example that's similar to this, but I suspect I'm not looking hard enough or else I'm missing something.

peschkaj: It looks like the loss of color formatting is a Windows only issue - verified on Ubuntu 15.10 and some version of Fedora (see [Piping stderr strips formatting/colors][1]). 

Windows 10 does support VT100 terminal codes, but after a tentative attempt at enabling them, it looks like `rustup` isn't using the VT100 codes, but is using the older binary codes.

*Possible Fixes*
There are a few possible changes, but they only fix the issue for Windows 10 users. 
* We could use `colored` to deal with terminal colors, but that adds an additional dependency. 
* We could set the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag (see retep's comment on [Introducing Colored][2]), but this flag only exists on Windows 10 and it requires ANSI/VT100 color codes.
* Do nothing and inform Windows users that this is a quirk of `rustup` on Windows when they've enabled telemetry.

@brson I think the call on this is up to you. We can always create an issue to fix this after the 1.0 release.

[1]: https://users.rust-lang.org/t/piping-stderr-strips-formatting-colors/5380/6
[2]: https://users.rust-lang.org/t/introducing-colored-the-most-simple-way-to-add-colors-in-your-term/4968/44

peschkaj:
On further review, I think I've addressed the main issues from the last code review, all tests pass, and the only outstanding item is the loss of formatting when we pipe stderr on Windows, and this is only when telemetry is on. This may be an unfortunate side effect of piping stderr and using the Windows control sequences. The problem should be eliminated on Windows 10 and newer (see my comment on [`command.rs`][1]). 

This also might not be considered that big of an issue, I would assume that many of the people who are willing to generate telemetry are also gather telemetry on an unattended build farm, so they may not care that they're losing formatting.

[1]: https://github.com/rust-lang-nursery/rustup.rs/pull/289/commits/31c5e1baf44990015c56c515c5f9e048a77daf93#r59600418

brson:
@peschkaj Looks like I'm not going to get to this until tomorrow. Thanks for your patience.

src/multirust/config.rs
```diff
@@ -402,4 +410,51 @@ impl Cfg {
...
+    }
+
+    fn enable_telemetry(&self) -> Result<()> {
+        let work_file = try!(self.temp_cfg.new_file());
+        
+        let _ = utils::ensure_dir_exists("telemetry", &self.multirust_dir.join("telemetry"), ntfy!(&NotifyHandler::none()));
+
+        try!(utils::write_file("temp", &work_file, ""));
+
+        try!(utils::rename_file("telemetry", &*work_file, &self.multirust_dir.join("telemetry_on")));
```
brson: Just personal preference: can you call this "telemetry-on" instead of "telemetry_on"?

tests/cli-exact.rs
```diff
@@ -195,3 +195,24 @@ info: installing component 'rust-std' for '{0}'
...
+
+#[test]
+fn disable_telemetry() {
+    setup(&|config| {
+        expect_ok_ex(config, 
+                     &["rustup", "telemetry", "off"],
+                     r"",
+                     &format!("info: telemetry set to 'off'\n"));
+    });
+}
```
brson: I wonder if you could add one more smoke test: turn on telemetry, then run rustc, then check that there's a file in the telemetry folder.

src/multirust/toolchain.rs
```diff
@@ -142,11 +146,38 @@ impl<'a> Toolchain<'a> {
     }
 
     pub fn install_from_dist(&self) -> Result<UpdateStatus> {
+        // if self.cfg.telemetry_enabled() {
+        //     return self.install_from_dist_with_telemetry();
+        // }
```
brson: Please delete this dead code.

src/multirust-cli/rustup_mode.rs
```diff
@@ -160,6 +162,13 @@ pub fn cli() -> App<'static, 'static> {
                      .short("y")))
             .subcommand(SubCommand::with_name("upgrade-data")
                 .about("Upgrade the internal data format.")))
+        .subcommand(SubCommand::with_name("telemetry")
+                .about("Enable or disable rust telemetry")
+                .arg(Arg::with_name("enabled")
+                        .takes_value(true)
+                        .value_name("telemetry")
+                        .help("Set telemetry 'on' or 'off'")
+                        .possible_values(&["on", "off"])))
```
brson: Let's set this to `hidden(true)` for the time being until we've fleshed out the feature a bit more and really want people to turn it on.

brson:
OK, this looks great. I did make two comments about minor changes I'd like to do before landing.

Some thoughts about next steps:

We need to be concerned about the on-disk format and forward-compatibility. Every time we change rustup's metedata format in the future we're going to have to provide an upgrade path. Right now, while we're still iterating a lot, I've been pretty sloppy about it, but soon we're going to need to be really careful about not breaking the metadata on upgrade. This scheme we have now with the "telemetry-on" file and the "telemetry" folder will suite us for now, but will almost certainly change again in the future - this isn't a problem, just something to be aware of. The way data upgrades work presently is that the "version" file contains a number representing the format of the `~/.multirust` directory - when we need to make incompatible changes we bump that number, and write some transition code that runs during the `rustup self upgrade-data` command.

I think we will want to put a similar "version" field inside of our `LogMessage` type so that later analysis can distinguish exactly what it's looking at (this structure is going to change a lot over time). You can start it at version 1 and just bump every time we make any change.

Telemetry should *probably* limit the amount of log files it will accumulate to avoid growing without bound. Perhaps a simple way to do this would be to just check whether there are more than e.g. 100 files in the telemetry directory and delete the oldest.

Besides those small concerns, I'd say there are three avenues for further work: analysis, upload, and defining more events. The latter is probably least important - best to get everything working end-to-end to prove it out.

So I'd suggest we start thinking about what it will take to deliver the telemetry logs to us and then for us to analyze them.

Just some ideas:

We can probably start without a server at all, and just ask friends to turn on the telemetry and email them to us. We do though need a way to package up the logs in a convenient form for sending, and this form we should design to be the same whether the logs are sent manually by email or automatically over HTTPS.

To that end I'd suggest that we add functionality to convert the telemetry logs to a tarball, which can be either emailed or uploaded. We might have a `rustup telemetry package` command that stuffs all the logs into a tarball with a unique name, puts it in the current directory, prints the name of the tarball, then deletes all the old telemetry data.

The tarball will need to have a unique name so I'd suggest a scheme like `YYYY-MM-DD-$random-hex-digits.tar.gz`. The date helps with sorting and the random number just makes it unique.

With that we can start imagining how to create tooling to turn a big mess of these tarballs into useful information. I think it makes sense to put the analysis in the same crate as the rest of our telemetry code (and we'll probably end up creating a `rustup-telemetry` crate to hold it all eventually) so that we can share the code between rustup and the hypothetical analysis tools.

Again just spitballing, but we might add another CLI tool called maybe `rustup-telemetry-report`, in this same repo for doing the later analysis.

OK, that's all the ideas I've got right now. Let me know what you think.

peschkaj:
Sounds good. I'll get the fixes together by the end of Monday and respond in more detail tomorrow.

In short, this all sounds like a good plan. 

Thanks for the review.

brson:
Maybe we can just do all the reporting in rustup itself. Make it useful both for individuals and the group. So perhaps it could draw its data from the users telemetry directory or a directory of tarballs.

Oh, maybe we make the log files and the tarballs have the same naming scheme, and the analysis just works off of a directory of log files and tarballs. That seems pretty clean.

peschkaj:
I've got basic implementation of the "must haves" done.  

I've started working on the delete files, but when I try the usual test case shenanigans, the proxy'd multirust/rustup directory isn't created. Is there an incantation I should be using for this? An example is at https://gist.github.com/peschkaj/dc1211f3511320da7cfb8f0ef84bfb24

peschkaj:
Oh, should add that if we want to hold off on deleting files so we can ship the initial feature, I can pull out the code and then create multiple issues so we can track the new features separately. That does seem like a better fit to me, but I spent 8 hours in aircraft yesterday, so my decision making might be suspect.

peschkaj:
Travis appears to be drunk - `cp: cannot stat ‘target/i686-unknown-linux-gnu/release/rustup-init’: No such file or directory`

brson:
I definitely do want to land this before moving on to more features. I think we can land this right now. Just taking another look over it.

brson:
The travis failures are due to an upstream change, nbd.

brson:
> I've started working on the delete files, but when I try the usual test case shenanigans, the proxy'd multirust/rustup directory isn't created. Is there an incantation I should be using for this? An example is at https://gist.github.com/peschkaj/dc1211f3511320da7cfb8f0ef84bfb24

Do you mean that the directory represented in tessts by `config.rustupdir` isn't created (i.e. what is usually `~/.multirust`)? From reading your test case I don't see anything obviously wrong, though I would not expect you to need to write `ensure_dir_exists` since rustup should create that directory itself. Which line in the test is failing?

Please do file issues about further features so we can continue this discussion there (I'll probably forget to check on this closed PR again).

brson:
Thanks @peschkaj!

peschkaj:
:+1: happy to help

On Mon, Apr 18, 2016 at 8:57 PM Brian Anderson <notifications@github.com>
wrote:

> Thanks @peschkaj <https://github.com/peschkaj>!
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/rust-lang-nursery/rustup.rs/pull/289#issuecomment-211659470>
>
-- 

Jeremiah Peschka

messense:
@brson 

Now that `serde` reached 1.0 release and `rustc-serialize` is deprecated, should rustup switch to `serde`?

PR #293, opened 2016-04-10
Custom and interactive install

brson:
This does a few things to make installation customizable.

The `--no-modify-path` flag disables `PATH` modification. `rustup-setup.sh` passes its arguments to `rustup-setup.exe`.

During the default interactive installation you can press "a" to enter "advanced installation", which just asks you the values of the two things that can be configured (path modification and default toolchain). We need this feature because it's the only way to change the installation options without finding out the command line options and starting installation over.

[Here's what it looks like in action](http://showterm.io/47a96fc6d1e94602bbb0e#fast).

It could be a lot nicer still, but this is all the features I expect.

Fixes https://github.com/rust-lang-nursery/multirust-rs/issues/271, https://github.com/rust-lang-nursery/multirust-rs/issues/170.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,6 +405,8 @@ dependencies = [
  "tempdir 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.1.28 (registry+https://github.com/rust-lang/crates.io-index)",
  "walkdir 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winreg 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


alexcrichton:
Sounds/looks good to me! Looks like tests are failing but otherwise r=me

Victorcccc:
WeChat458edc9004ecb32ae2f1dd678ee97d51
what kind of process i could install rust ?

kinnison:
@Victorcccc That image shows several issues.  The first is that you've given an invalid value for `profile` which will result in installation later down the line.  The examples given (minimal/default/complete) are the valid values there, and I'd strongly recommend leaving it at `default` unless you have good reason to change it.  As for the error you hilighted, it suggests that your `.bash_profile` is a directory rather than a file, which `rustup` certainly isn't expecting.  If you have some unusual setup for your shell configurations then I suggest you set it to not modify PATH itself, and deal with that afterwards in whatever manner your personal configurations require.  If you continue to have problems, then filing a fresh issue, rather than replying to a pull request from almost four years ago would be more sensible.

Good luck, I hope you manage to resolve your installation problem, and welcome to Rust.

PR #310, opened 2016-04-13
Fixes for rustup-setup.sh

brson:
cc @DaveKram this should fix your issue. I'll try to get it deployed today. I'd *really* love to hear if aarch64 rustup works since you are the only one I know testing.


DaveKram:
Thanks for the response. I'll take a look when I get home from work today and let you know what I discover.

PR #312, opened 2016-04-14
Convert from multirust to rustup

brson:
Rename the crates and remove the "multirust" bin. There's certainly more to do, but this is the big stuff.


alexcrichton:
Travis seems to have disappeared, but otherwise r=me when this is ready

Issue #308, opened 2016-04-13
armv7 build doesn't default to armv7 but to arm


brson:
The `get_original_host_triple` function just uses `env::consts::ARCH` for the arch part of the target, but it should be `armv7`. Hopefully we can derive this info without setting `RUSTUP_OVERRIDE_HOST_TRIPLE`.

cc @japaric

brson:
@japaric suggested using one of the many variables Cargo sets by default, one of which is probably `TARGET`.

japaric:
It appears TARGET is only set for build scripts as [documented here](https://github.com/rust-lang/cargo/blob/master/src/doc/environment-variables.md). I'm not sure if we can use some `build.rs` magic to forward the variable to the `cargo build` process. @alexcrichton Know of any way to access TARGET during the compilation of a crate?

alexcrichton:
a bit hackish but you can

* In a build script, emit `env::var("TARGET")` into a file.
* In a crate, include it via `include_str!`

japaric:
@alexcrichton That's what I feared :P. I'll prepare a PR and let @brson decide.

PR #314, opened 2016-04-14
dist: pick the right host triple on ARMv7 devices

japaric:
Uses the "hackish" solution suggested by @alexcrichton in #308.

Check commit messages for details.

r? @brson

### DIFF
--- a/src/multirust-dist/build.rs
+++ b/src/multirust-dist/build.rs
@@ -1,0 +1,11 @@
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::path::PathBuf;
+
+fn main() {
+    let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+    let target = env::var("TARGET").unwrap();
+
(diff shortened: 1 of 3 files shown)


src/multirust-dist/build.rs
```diff
@@ -0,0 +1,22 @@
+use std::env;
+use std::fs::File;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+
+fn main() {
+    let manifest_dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+    let target = env::var("TARGET").unwrap();
+    let path = Path::new("src/TARGET");
```
alexcrichton: Oh build scripts should in theory never generate code directly into the source directory, you can throw it in `OUT_DIR` and then do something like:

```
let s = include_str!(concat!(env!("OUT_DIR"), "/target.txt")));
```

japaric:
updated to use OUT_DIR instead of CARGO_MANIFEST_DIR.

brson:
Oh my god it's horrible...

PR #323, opened 2016-04-15
Fill in copyright info in LICENSE-APACHE

thisishugo:


brson:
Thanks for the PR.

This piece of text is not supposed to be altered, though the way it's presented is confusing. This appendix section is showing an example of how to provide notification that a work uses the license without duplicating the entire license text. The way this is typically used is to copy the entire notice from the appendix to, e.g., the top of a source file, filling in the blank parts with your own copyright information. The key sentence from the included instructions is "to apply the Apache License to your work, attach the following boilerplate notice," indicating the text should be copied elsewhere.

In rustup the place where we do license notification is currently [at the end of the README](https://github.com/rust-lang-nursery/rustup.rs#license). That section could be written more elegantly indeed, though I'd be disinclined to simply copy the apache boilerplate there since we're dual-licensed.

brson:
Closing per previous comment. Thanks.

PR #324, opened 2016-04-15
Fix build status badges

SShrike:
Also, move the badges into a table and change the badge graphics to SVGs served via [shields.io](http://shields.io/)

I'm pretty certain that the AppVeyor graphic and URL point to the old repository's build status so if someone could provide me with the right URL I'll go ahead and update it.


SShrike:
I think I've found the correct AppVeyor links, or at least I hope I have.

brson:
Thanks for the fixes!

I don't love how much space this takes up before the sales pitch, but I'll live with it a while and see if it really bothers me.

SShrike:
I wasn't sure about the positioning either, perhaps it would of looked better under the sales pitch?

PR #329, opened 2016-04-17
Mentioned Visual C++ Build Tools 2015

Fireforge:
There is a Visual C++ Build Tools 2015 that installs just the C++ build tools without the full Visual Studio. It includes the necessary linker to build Rust MSVC. It's on the same downloads page, under "Tools for Visual Studio 2015". I think a lot of people would be happy to know of a more lightweight installer option, hopefully it's not confusing.
More info here: https://blogs.msdn.microsoft.com/vcblog/2016/03/31/announcing-the-official-release-of-the-visual-c-build-tools-2015/

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ toolchain] use the GNU build.
 
 MSVC builds of Rust additionally require an [installation of Visual
 Studio 2013 (or later) or the Visual C++ Build Tools 2015][vs] so
-rustc can use its linker. for Visual Studio, make sure to check
+rustc can use its linker. For Visual Studio, make sure to check
 the "C++ tools" option. No additional software installation is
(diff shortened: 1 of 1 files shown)


peschkaj:
The Visual Studio Downloads page doesn't provide a direct link to the Visual C++ Build Tools. The only source I've found with a direct download link is [the announcement blog post](https://blogs.msdn.microsoft.com/vcblog/2016/03/31/announcing-the-official-release-of-the-visual-c-build-tools-2015/). Should we also add instructions to help someone only install the VC++ Build Tools?

brson:
[Here's what the Rust website says on the subject](https://www.rust-lang.org/downloads.html#win-foot).

This is an improvement regardless if we do more so I'll take it as is.

PR #337, opened 2016-04-17
Bump to 0.1.7. Add CHANGELOG.md

brson:


brson:
I don't understand why builds are failing right now on travis. Seems like rustup.sh is allowing two toolchains to be mixed.

brson:
[Seems to be a bug in the nightly builders](https://github.com/rust-lang/rust/issues/33063).

alexcrichton:
We can trigger the builds manually tomorrow to send out new builds

Issue #340, opened 2016-04-19
Remove old telemetry files


peschkaj:
Remove old telemetry files from the rustup telemetry directory so we don't eventually fill the server with telemetry information. For an initial implementation, we can start with keeping 100 telemetry files (initially suggested in https://github.com/rust-lang-nursery/rustup.rs/pull/289)

peschkaj:
Closed by https://github.com/rust-lang-nursery/rustup.rs/issues/342

PR #342, opened 2016-04-19
Adding telemetry directory cleanup

peschkaj:
This should implement #340

### DIFF
--- a/src/rustup/telemetry.rs
+++ b/src/rustup/telemetry.rs
@@ -1,3 +1,4 @@
+use errors::*;
 use time;
 use rustup_utils::{raw, utils};
 use rustc_serialize::json;
@@ -39,7 +40,7 @@ impl Telemetry {
         Telemetry { telemetry_dir: telemetry_dir }
     }
 
-    pub fn log_telemetry(&self, event: TelemetryEvent) {
+    pub fn log_telemetry(&self, event: TelemetryEvent) -> Result<()> {
         let current_time = time::now_utc();
         let ln = LogMessage { log_time_s: current_time.to_timespec().sec,
                               event: event,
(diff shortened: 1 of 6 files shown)


src/rustup/telemetry.rs
```diff
@@ -47,6 +49,43 @@ impl Telemetry {
...
+        // If it is a new day, then attempt to clean the telemetry directory.
+        if !raw::is_file(&self.telemetry_dir.join(&filename)) {
+            self.clean_telemetry_dir();
+        }
+
         let _ = utils::append_file("telemetry", &self.telemetry_dir.join(&filename), &json);
     }
+
+    pub fn clean_telemetry_dir(&self) {
+        let contents = self.telemetry_dir.read_dir().unwrap();
```
brson: Please handle this error instead of unwrapping.

src/rustup/telemetry.rs
```diff
@@ -47,6 +49,43 @@ impl Telemetry {
...
+
+        for c in contents {
+            let x = c.unwrap();
+            let filename = x.path().file_name().unwrap().to_str().unwrap().to_owned();
+            if filename.starts_with("log") && filename.ends_with("json") {
+                telemetry_files.push(x.path());
+            }
+        }
+
+        println!("found {:?} files to delete", telemetry_files.len());
```
brson: Let's do this silently and not print anything.

src/rustup/telemetry.rs
```diff
@@ -47,6 +49,43 @@ impl Telemetry {
...
+
+        let dl: usize = telemetry_files.len() - MAX_TELEMETRY_FILES;
+        let dl = dl + 1 as usize;
+
+        telemetry_files.sort();
+        telemetry_files.dedup();
+
+        for i in 0..dl {
+            let i = i as usize;
+            let _ = fs::remove_file(&telemetry_files[i]);
```
brson: Please handle this error too.

brson:
Sweet. Looks good but for a few comments. Let me know if you have difficulties with the error handling. It can be awkward in rustup because the error types don't have automatic `From` conversions.

peschkaj:
How far up the stack should these errors go?  I can change `clean_telemetry_dir` to return a `Result<()>`. This seems like something we might want to do in the event that we can't execute `read_dir()` on the directory or that we can't delete a file from the telemetry directory. 

I was thinking that an error in `clean_telemetry_dir` should be caught in `log_telemetry` and then if an error occurs, I should use one of the `info!` macro to let the user know that we couldn't delete old log files.

brson:
@peschkaj Agh. Sorry for the delay in responding again. You can always ping me on IRC if I don't respond quickly.

I'd prefer to just bubble them all the way up to the top by returning `Result`s. That's how all other I/O errors are dealt with presently.

brson:
Thanks again!

PR #355, opened 2016-04-24
Improve download progress

nodakai:


src/rustup-cli/download_tracker.rs
```diff
@@ -121,37 +121,31 @@ impl DownloadTracker {
     }
     /// Display the tracked download information to the terminal.
     fn display(&mut self) {
-        let total_h = HumanReadable(self.total_downloaded as u64);
+        let total_h = HumanReadable(self.total_downloaded as f64);
         let sum = self.downloaded_last_few_secs
                       .iter()
-                      .fold(0usize, |a, &v| a + v);
+                      .fold(0., |a, &v| a + v as f64);
```
tamird: what's the point of making this a float? this is simple addition of integers, there's no rounding here.

src/rustup-cli/download_tracker.rs
```diff
@@ -121,37 +121,31 @@ impl DownloadTracker {
...
         let len = self.downloaded_last_few_secs.len();
         let speed = if len > 0 {
-            (sum / len) as u64
+            sum / len as f64
```
tamird: for consistency with the below i think this should be `sum as f64 / len`

nodakai: f64 / usize doesn't compile

tamird: then i guess you want `sum as f64 / len as f64`, right?

src/rustup-cli/download_tracker.rs
nodakai: Hmm, this is just a matter of taste. I just wanted `sum` to be a float and I tend to apply int-to-float casting as early as possible when I know the end result is a float.

src/rustup-cli/download_tracker.rs
nodakai: In a sense, yes. But in my version, `sum` is already of type `f64`.

src/rustup-cli/download_tracker.rs
```diff
@@ -168,20 +162,36 @@ impl DownloadTracker {
...
 impl fmt::Display for HumanReadable {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        const KIB: f64 = 1024.0;
-        const MIB: f64 = 1048576.0;
-        let size = self.0 as f64;
+        if f.alternate() {  // repurposing the alternate mode for ETA
+            let sec = self.0;
+
+            if sec <= 0. {
+                write!(f, "Unknown")
```
nodakai: Oops, the condition to print "Unknown" is now wrong...

brson:
Thanks for the improvements @nodakai, and for explaining the float math.

nodakai:
@brson Sorry I need fix the condition to display "unknown". But haven't got the time... I'll prepare another patch.

Issue #353, opened 2016-04-22
Arch and OS not detected on RPi 3 with Ubuntu MATE


Phyllostachys:
I went to www.rustup.rs to see if it would work on my RPi 3 but it reported that it couldn't detect the OS.

Phyllostachys:
Upon some more investigation, it seems that rustup.sh does properly detect armv7l and Linux, it doesn't have a source for rustc, rustc-docs, and cargo:
```
error: some components unavailable for download: 'rustc', 'rust-docs', 'cargo'
rustup: command failed: /tmp/tmp.k3xO6b3Vc4/rustup-init
```

Perhaps I jumped the gun on submitting this issue. I went ahead and made a fork, adding the support, but I don't think it is time yet to expose it.

brson:
I'm *guessing* that this compiler just wasn't published in the 1.8 stable build and this will work if you continue to install nightly with `rustup default nightly`.

Handling toolchain install failure [more gracefully](https://github.com/rust-lang-nursery/rustup.rs/issues/275) would help make this more clear.

Phyllostachys:
Added the pull request. Since it sounds like it is technically supported I figured that it wouldn't hurt.

Phyllostachys:
Since it is all merged in, closing the issue.

PR #359, opened 2016-04-26
Add armv7l support to www

Phyllostachys:
#353 

I only made a conservative change. I really thought about making it detect any Linux distro but wondered if maybe there would be some issue I didn't think about. So, I stuck with the conservative way.


brson:
Thanks!

PR #363, opened 2016-04-27
Website updates

brson:
Restyle the website to look like rust-lang.org, fix platform detection for a few cases. [Preview](http://brson.github.io/rustup.rs/www/#).

Fixes https://github.com/rust-lang-nursery/rustup.rs/issues/256


brson:
You can type "test" on the website to cycle through all the platform-specific instructions.

alexcrichton:
I'm pretty excited about this!

cc @aturon, @steveklabnik, @wycats

brson:
I put a big red warning back on there. It's still not ready to recommend without reservation.

aturon:
This looks pretty great to me.

Really the only nitpick I have is whether we should try to be more clear about what rustup offers beyond being "the ultimate way to install Rust" (e.g., talking about cross-compilation components). Though I love the *spirit* of the current phrasing :)

steveklabnik:
I am still having issues seeing the fonts. Linux/Iceweasel here. Hrm. :(

Trying to diagnose.

ashleygwilliams:
hey ya'll! you can save on requests by asking for all of the google fonts at one, e.g. https://github.com/ashleygwilliams/jsconf-2015-deck/blob/gh-pages/public/custom.css#L1

i am happy to make a PR to fix this once this lands, or you could refactor on your own :)

steveklabnik:
Changing my `font-family` to remove Fira Sans will allow this to render.

Gah.

steveklabnik:
@brson I think this is due to `@import`. The same thing _does_ work for me on the main rust page with `@font-face`.

nodakai:
I think I'm experiencing the same problem with @steveklabnik .

Google Chrome 50.0.2661.75
new-rustup-website-chrome50

Iceweasel 45.0.1
new-rustup-website-iceweasel45

Both on Debian testing x86_64

brson:
> hey ya'll! you can save on requests by asking for all of the google fonts at one, e.g. https://github.com/ashleygwilliams/jsconf-2015-deck/blob/gh-pages/public/custom.css#L1

Thanks, I'll go ahead and make the change.

> Really the only nitpick I have is whether we should try to be more clear about what rustup offers beyond being "the ultimate way to install Rust" (e.g., talking about cross-compilation components). Though I love the spirit of the current phrasing :)

I'll think about rephrasing it to show more and tell less.

> @brson I think this is due to @import. The same thing does work for me on the main rust page with @font-face.

I'll switch to @font-face.

> I think I'm experiencing the same problem with @steveklabnik .

Thanks for the screenshots. I'll try with @font-face.

brson:
I pushed another revision. This one uses `@font-face` instead of `@import`, but @ashleygwilliams since I switched away from `@import` I didn't know how to merge them into one declaration.

I changed the warning to just say "this is beta software", remove "expect breakage".

I don't yet have anything else I like better for the pitch, but I removed 'ultimate'. Now it just says the plain fact, 'rustup is an installer for the systems programming language Rust'. I'm somewhat disinclined to try to 'sell' rustup here, for two reasons: rustup is the Rust installer, and if you want Rust you've got to click that button. The user already knows they want Rust. This is how you get Rust.

Furthermore, I'm not sure there's much more that can be explained without many more words. The most import features of rustup are installation, updates, channel tracking, and installation of cross-compilation targets. So you might say "rustup is an installer for the systems programming language Rust. It keeps your installation up to date, manages multiple Rust toolchains, and makes cross-compilation easier," possibly with another link to a more complete pitch.

nodakai:
Fonts look fine now:
new-rustup-website-chrome50-1

steveklabnik:
They work for me too :+1:

brson:
Going with it as is for now.

PR #364, opened 2016-04-27
download_tracker.rs: Fix 3358cce

nodakai:
"ETA Unknown" was always displayed at the end of download.


brson:
Thanks @nodakai. As soon as I get the freebsd/netbsd errors fixed I'll deploy a new release.

Issue #346, opened 2016-04-20
Checksum failed (again)


alexbool:
I don't know if this is related to previous bugs with the same name, so filing a new one.

```
alexbool@alexbool-osx ~> rustup -v update nightly
verbose: read metadata version: '12'
verbose: updating existing install for 'nightly-x86_64-apple-darwin'
verbose: toolchain directory: '/Users/alexbool/.multirust/toolchains/nightly-x86_64-apple-darwin'
info: syncing channel updates for 'nightly-x86_64-apple-darwin'
verbose: creating temp file: /Users/alexbool/.multirust/tmp/xzj52e23ecrqyt5o_file
verbose: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-nightly.toml.sha256'
verbose: deleted temp file: /Users/alexbool/.multirust/tmp/xzj52e23ecrqyt5o_file
verbose: creating temp file: /Users/alexbool/.multirust/tmp/b6yr1ng0mairn39o_file.toml
verbose: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-nightly.toml'
verbose: deleted temp file: /Users/alexbool/.multirust/tmp/b6yr1ng0mairn39o_file.toml
error: checksum failed, expected: '3721ec0a1b3a063da1f4fe594a0f4afdefefcb6fd4c3453b848efd980e4530a9', calculated: 'dc069d0e5b653add4afee522e2f139a9f1479ca5967694e9ef351e1a5648b0b1'
```

brson:
Thanks for the report. I'd imagine this has resolved itself by now?

It's probably not the same issue we saw over the weekend. We have constant problems with our CDN invalidating our binary artifacts at different times, causing the .sha256 files to disagree with the files they correspond to. I believe what you are seeing is the effects of [this bug](https://github.com/rust-lang/rust/issues/21239).

It's super frustrating, but I've just taken another look at our CDN invalidation script and it looks right.

Even though I think this is an upstream problem, I'm going to leave this open for others to find.

brson:
After thinking about this further there is only one scenario where we should possibly be seeing these hash mismatches, and that's when comparing the manifest with its hash. Other hashes are all contained in the manifest themselves and refer to files with unique names that don't require CDN invalidation.

It may be possible to tweak the manifest format further to avoid this problem. For example, the manifest itself could be a toml file prepended with a 60 byte sha.

Could also just not validate the manifest and leave errors to happen during parsing and downloading of bins.

[...]

PR #372, opened 2016-04-28
Don't make it an error for the self-update hash to be wrong

brson:
We've got problems with the updater and its hash being out
of sync. Thes patch considers a hash failure here a temporary
problem, prints a status message and exits with success.

cc #346

r? @alexcrichton


alexcrichton:
r=me, but looks like tests are legit failing

PR #391, opened 2016-04-30
rustup-init.sh calls 'file'

brson:


brson:
I don't like this solution.

PR #392, opened 2016-04-30
Simplify detection of 32-bit Linux userland

brson:
And fail better if `file` doesn't exist.


rustup-init.sh
```diff
@@ -159,21 +159,9 @@ get_architecture() {
 
     # Detect 64-bit linux with 32-bit userland
     if [ $_ostype = unknown-linux-gnu -a $_cputype = x86_64 ]; then
-        # $SHELL does not exist in standard 'sh', so probably only exists
-        # if configure is running in an interactive bash shell. /usr/bin/env
-        # exists *everywhere*.
```
brson: There's no reason here to check the shell, if /usr/bin/env always exists.

PR #394, opened 2016-05-02
Autodetect TTY and add colors

peschkaj:
If *nix users are running on a tty (using the same checks as rustc & cargo), then colors will be displayed.

Windows users are still out in the cold. This could potentially be fixed by changing rustc to emit ANSI color codes on Windows 10 and newer - but that's a rustc thing, not a rustup thing.


src/rustup/command.rs
```diff
@@ -30,8 +30,17 @@ pub fn run_command_for_dir<S: AsRef<OsStr>>(cmd: Command,
...
     let now = Instant::now();
-
+    
     cmd.args(&args[1..]);
+    
+    if stderr_isatty() && !(&args).iter().any(|e| {
+            let e = e.as_ref().to_str().unwrap_or("");
+            e == "--color"
+        }) 
+    {
```
brson: Please pull out this expression from `(&args).iter()` onward into a local variable with a descriptive name like, `has_color_arg`.

I believe this check should be `e.starts_with("--color")` since it could see e.g. `--color=always`.

src/rustup/command.rs
```diff
@@ -132,3 +141,27 @@ fn run_command_for_dir_without_telemetry<S: AsRef<OsStr>>(mut cmd: Command, args
...
+        fn GetStdHandle(which: DWORD) -> HANDLE;
+        fn GetConsoleMode(hConsoleHandle: HANDLE,
+                          lpMode: *mut DWORD) -> BOOL;
+    }
+    unsafe {
+        let handle = GetStdHandle(STD_ERROR_HANDLE);
+        let mut out = 0;
+        GetConsoleMode(handle, &mut out) != 0
+    }
+}
```
brson: This function already exists in `rustup_cli::tty`. Can you extract that module to `rustup_utils::tty` so it can be reused here?

brson:
Looks good. Thank you.

peschkaj:
Thanks for the eyeballs! Fixed those issues, test pass locally, now to wait for appveyor and travis.

brson:
lgtm but the test failure is legit

peschkaj:
Should be fixed in the current commit. This is the reverse of the usual pain - I frequently fail to test on Linux.

PR #396, opened 2016-05-03
OpenSSL upgrades

brson:
This upgrades OpenSSL to 1.0.2h on Unix and *1.0.2g* on Windows/MSVC. The place we get our Windows builds has not yet upgraded to 'h'.


brson:
I'm going to merge this with another patch.

Issue #344, opened 2016-04-19
Switch from hyper to curl


brson:
Hyper doesn't support HTTP proxies, while curl does.

brson:
Not sure if this should block an initial release, but it does seem important.

maximih:
A lot of companies are using proxies for internet access so I think this is very important.

eminence:
the other option would be to teach hyper about proxies, but that seems like way more work than switching to curl

maximih:
FWIW, there is > 1 year old issue on hyper regarding proxy support in the client: https://github.com/hyperium/hyper/issues/531

xen0n:
I'm definitely in favor of this because of the proxy support, but the [`curl` crate](https://github.com/carllerche/curl-rust) declares an unstable API, also it didn't provide any notes on Windows support. However seeing the amount of work needed to make `hyper` support proxies it seems `curl` is the only way to go :confused:

peschkaj:
Looks like curl does work on Windows, but maybe only with MSVC?
https://github.com/killercup/cargo-edit/issues/55#issuecomment-192583891

On Mon, Apr 25, 2016, 00:16 Wang Xuerui <notifications@github.com> wrote:

> I'm definitely in favor of this because of the proxy support, but the curl
> crate <https://github.com/carllerche/curl-rust> declares an unstable API,
> also it didn't provide any notes on Windows support. However seeing the
> amount of work needed to make hyper support proxies it seems curl is the
> only way to go [image: :confused:]
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/rust-lang-nursery/rustup.rs/issues/344#issuecomment-214176336>
>
-- 

Jeremiah Peschka

brson:
@peschkaj Cargo uses libcurl, so I'm confident we can make it work on all platforms that matter.

seanmonstar:
Proxy support is easy in the async branch, I can try to add basic support to hyper's sync branch to support this feature.

Basic functionality would be: when sending the request, if `req.url.host()` does not equal the `Host` header, then send the url in the `absolute-uri` format used for proxies. Support in the request builder could look like:

```rust
client.get("http://rustup.rs/path")
    .proxy(env::var("HTTP_PROXY").unwrap())
    .send()
```

Where `HTTP_PROXY=http://our.corporate.proxy.tld`

Would this work?

seanmonstar:
[...]

PR #400, opened 2016-05-06
Various networking fixes

### LINKED ISSUE #204: Don't require a custom `.cargo/config` to build
The dependency issue with openssl (previously missing dependencies on `user32-sys` and `gdi32-sys`) has been fixed, and yet multirust still requires a custom cargo config to build.

For some reason, the dylib dependency from `gdi32-sys` does not get passed to the linker, despite it being present in the build-script output.

To reproduce this issue, install the openssl binaries exactly as the appveyor script does, and set these environment variables (assuming 64-bit):

```
DEP_OPENSSL_INCLUDE=C:\OpenSSL\openssl-1.0.2d-vs2015\include64
OPENSSL_INCLUDE_DIR=C:\OpenSSL\openssl-1.0.2d-vs2015\include64
OPENSSL_LIB_DIR=C:\OpenSSL\openssl-1.0.2d-vs2015\lib64
OPENSSL_LIBS=libeay32MT:ssleay32MT
```

Then do a clean cargo build. The expected result is a ton of warnings about missing pdb files, followed by the relevent error, a missing symbol from `gdi32.lib`.

### LINKED ISSUE #243: Build openssl on appveyor
The way we are acquiring openssl in our windows builds is sub-optimal. Can we build this ourselves like on linux?

cc @alexcrichton 

In the long term, switching from hyper to curl and rust-crypto (for sha256) is probably best because that gets is HTTP proxy support (what that entails I'm not sure), but we need to tighten this up.

brson:
* Use the platform-native TLS stacks with [native-tls](https://github.com/sfackler/rust-native-tls), removing the OpenSSL dep on Windows.
* Do TLS hostname verification, via native-tls
* Remove the direct OpenSSL dep by switching to rust-crypto for sha256
* Upgrade OpenSSL to the latest version on Unix

Fixes #243, fixes #204, #344

r? @alexcrichton cc @sfackler @seanmonstar


brson:
It appears that rust-crypto has a native component. I'll find a sha256 implementation that doesn't.

src/rustup-utils/src/lib.rs
```diff
@@ -1,13 +1,15 @@
 #![feature(core_intrinsics)] // For type_name().
 #![feature(fundamental)]
+#![feature(step_by)]
```
alexcrichton: :( I thought we wanted to go the other way

alexcrichton:
Nice! r+ from me basically, but looks like the openssl crate is still getting compiled on MinGW which is causing a build failure. Maybe something's bringing it in by accident?

src/rustup-utils/src/lib.rs
brson: Yeah, it just happened to be used by the sha code and didn't seem worth messing with, but it should be easy to remove. I'll look at it.

PR #403, opened 2016-05-06
Add `rustup doc`.

Yasushi:


brson:
Thanks for the PR. That's an embarrassing oversight!

PR #404, opened 2016-05-06
Use the sha2 crate

brson:
I went ahead and created an independent crate for sha2 in case its useful to others.


alexcrichton:
r=me

PR #415, opened 2016-05-07
Version bumps

brson:
Prepare the 0.1.10 release, which contains only a single fix to the URLs in the Cargo files. Then bump all crates to 0.1.11-pre.


brson:
Actually, I don't want to do this.

PR #418, opened 2016-05-08
Improvements to welcome screen

Diggsey:
- Removes some code duplication from `term2`
- Uses `markdown` to parse markdown-style text and display formatted output in the terminal
- Implements a word-wrap algorithm
- Shows install options before asking whether to customize
- Uses numbered choices rather than letters

The markdown crate exposes a `tokenize` method which seems to be exactly fit for purpose. However, it exposes a private type in the return value which makes it unusable. I've opened an issue about it, but for the moment I've had to fork it and use a git dependency.

r? @brson

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,6 +8,7 @@ dependencies = [
  "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "markdown 0.1.2 (git+https://github.com/Diggsey/markdown.rs.git)",
  "rand 0.3.14 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


Diggsey:
Here's what it looks like:
screenshot

brson:
This looks really cool. I love the way this emphasizes key info.

Here are the things I noticed from just running it:

* On Unix, the path to .profile in the pre-install message should be indented two spaces
* The red in WARNING looks dark here. Is there a bright variant?
* The post install "Rust is installed now" message contains a leading "#". Not sure if that's intentional, but it looks out of place because I haven't seen another one.
* The post install message might also bold "run `source $HOME/.cargo/env`".

I'll give the code a review seperately.

brson:
It looks like the markdown parsing is not applied to the post-install and pre-uninstall messages.

brson:
Otherwise this lgtm. Very nice.

brson:
r=me with the markdown processing applied to all the messages and with the bots passing.

This will break tests in cli-install-interactive for sure.

Diggsey:
That should be everything (just waiting on tests)

PR #419, opened 2016-05-08
Multiple CLI cleanups

brson:
This adds help text for most subcommands and cleans them up in various ways.

The most important thing this does is try to resolve the bikesheddy problem of command synonyms that has been bugging me. That is, should the command be 'install' or 'update'? Should all install-like commands be the same verb or not?

From the commit, this is what I'm going for:

This reorganizes the verbs used in rustup commands. The concept here
is that every subcommand uses the most appropriate 'add/remove' verb
pair for itself, but implements hidden alternate commands using
the obvious other verbs.

So now the problematic verbs are:

* `toolchain install` / `toolchain uninstall`
* `target add` / `target remove`
* `override set` / `override unset`

Additional (hidden) synonyms are:

* `toolchain update`
* `toolchain add` / `toolchain remove`
* `target add` / `target remove`

This leaves the top level `rustup update` as the only non-hidden
'update' command, which has the unique focus of updating everything.

@aturon Importantly, this means that in the README I'm showing `rustup toolchain install $toolchain` for the installation command, and `rustup update` for doing updates. I like emphasizing the 'toolchain' in early examples of install command because it's an important concept. This is slightly different than the blog post you have in your hands where I'm demonstrating the (shorter but non-existant) `rustup install` command, which I have not implemented here.


brson:
I didn't make it clear but the hidden synonyms aren't just compromise, but are needed because I changed some of the 'preferred' verbs.

alexcrichton:
I wonder if it'd be worth to also have a dedicated `rustup install` synonym for `rustup toolchain install`? That would perhaps provide a nice easy vector for newbies and would also align with other systems like nvm/rvm/etc?

kbknapp:
https://github.com/kbknapp@gmail/clap-rs/issues/469 may be applicable too. It's something I hadn't prioritized, but would be a very quick addition.

brson:
@alexcrichton I added `rustup install <toolchain>` as a hidden synonym.

alexcrichton:
Sounds good to me! Maybe some of the README could be updated as well? Other than r=me though, this all sounds good to me

brson:
Well I made `rustup install` a hidden synonym so it *works*, but not with the intent of promoting it in for use, though I do agree that `rustup install nightly` is nice and sweet, and familiar to rbenv users. Have to think about it a bit more.

alexcrichton:
Ah ok, either way sounds good to me

brson:
I went ahead and put the sweet commands in the README.

PR #420, opened 2016-05-08
Use a .toml file to store settings

Diggsey:
This doesn't bump the metadata version. Running rustup will silently convert your `version`, `default` and `overrides` files into a single `settings.toml` file.

If your metadata version is older than "12" it will first be converted to the new format, and then rustup will require you to perform a metadata upgrade exactly as it would before.

This paves the way for storing additional information in the settings file (such as the host target).


Diggsey:
Example toml file:
```toml
default_toolchain = "nightly-x86_64-pc-windows-msvc"
version = "12"

[overrides]
"\\\\?\\C:\\Users\\diggs\\workspace\\multirust-rs" = "nightly-x86_64-pc-windows-msvc"
```

src/rustup/settings.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        Self::from_toml(value, "")
+    }
+    pub fn stringify(self) -> String {
+        toml::Value::Table(self.to_toml()).to_string()
+    }
+
+    pub fn from_toml(mut table: toml::Table, path: &str) -> Result<Self> {
+        let version = try!(get_string(&mut table, "version", path));
+        if !SUPPORTED_METADATA_VERSIONS.contains(&&*version) {
```
brson: ISTM that "2" shouldn't be considered valid here since there's never been a settings.toml containing version 2.

Diggsey: The upgrade to `settings.toml` is orthogonal to the metadata version: since it's lossless it gets converted automatically, so if you upgrade from an old version of multirust to rustup, it will result in a `settings.toml` file with a version of 2, and you will then be required to do a metadata upgrade.

brson:
My only reluctance is that this commits us to deserializing a config file every time someone runs cargo or rustc, but I'm somewhat resigned to that fate. Most if this information must be read anyway so it could even be better to load it all at once. cc @alexcrichton

Can you also add the [telemetry flag](https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup/config.rs#L426)?

src/rustup/config.rs
```diff
@@ -62,10 +57,47 @@ impl Cfg {
...
+                    }
+                }
+                Ok(())
+            }));
+
+            // Failure to delete these is not a fatal error
+            let _ = utils::remove_file("version", &legacy_version_file);
+            let _ = utils::remove_file("default", &default_file);
+            let _ = utils::remove_file("overrides", &override_db);
+        }
```
brson: Can you extract this into its own function?

src/rustup/settings.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+    fn read_settings(&self) -> Result<Settings> {
+        if !utils::is_file(&self.0) {
+            try!(self.write_settings(Default::default()));
+        }
+        let content = try!(utils::read_file("settings", &self.0));
+        Settings::parse(&content)
+    }
+    pub fn with<T, F: FnOnce(&Settings) -> Result<T>>(&self, f: F) -> Result<T> {
+        f(&try!(self.read_settings()))
```
brson: I imagine this file is going to get read several times during each run. Can it instead load the file at startup and reuse it?

src/rustup/settings.rs
brson: I see. Thanks for the explanation.

brson:
Looks great. I do think it needs to take care to not read the file multiple times.

Will you also add a test that this conversion works? A simple way to do it is to modify the files in `config.rustupdir` to look like the old metadata and then test that the default and override work as expected, and that the files are deleted.

Diggsey:
I think that's everything!

alexcrichton:
> My only reluctance is that this commits us to deserializing a config file every time someone runs cargo or rustc,

Yeah unfortunately `toml-rs` isn't exactly optimized for speed right now, but something like this shouldn't be *that* slow and it should be relatively easy to speed up toml-rs wherever possible (if necessary)

brson:
@Diggsey ok, this looks fine by me. Before merging this big change and risking more breakage I want to get a build out that [fixes the windows bustage](https://github.com/rust-lang-nursery/rustup.rs/pull/434). I'll merge after that.

Diggsey:
Great!

tests/cli-misc.rs
```diff
@@ -110,6 +110,33 @@ fn upgrade_v2_metadata_to_v12() {
...
+#[test]
+fn upgrade_toml_settings() {
+    setup(&|config| {
+        rustup_utils::raw::write_file(&config.rustupdir.join("version"), "2").unwrap();
+        rustup_utils::raw::write_file(&config.rustupdir.join("default"), "beta").unwrap();
+        rustup_utils::raw::write_file(&config.rustupdir.join("overrides"),
+                                      "a;nightly\nb;stable").unwrap();
+        rustup_utils::raw::write_file(&config.rustupdir.join("telemetry-on"), "").unwrap();
+        expect_err(config, &["rustup", "default", "nightly"],
+                   "rustup's metadata is out of date. run `rustup self upgrade-data`");
```
brson: Why is this expecting an error message saying to run `upgrade-data`? Doesn't this upgrade happen silently?

I think this test should be tweaked just a little. Before writing the old metadata, first run the commands that correspond to the metadata, then write the metadata and delete the settings.toml file. Then run some `rustup` command to force the silent upgrade. Then run further `rustc` commands that verify that the metadata is intact - that `rustc --version` prints the right thing for the default and override (telemetry flag probably not worth testing explicitly).

Diggsey: This is testing an upgrade from metadata version `2`, ie. an old metadata version, for which we currently require an explicit upgrade (because it wipes the toolchains). I can add a separate test for *just* the toml upgrade, but it would be a subset of what is tested here.

Diggsey:
This is back up to date, just waiting on CI again.

Issue #423, opened 2016-05-08
All network requests on Windows 7 fail on 0.1.9.


kevinpoitra:
I was using 0.1.8 up until yesterday. I ran rustup self update, it updated to 0.1.9, and now anything that makes a network request fails. Specifically, rustup throws an error [here](https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup-utils/src/raw.rs#L247). I then tried uninstalling rustup via rustup self uninstall, and then downloaded rustup-init.exe and ran that, which also is giving the same error as it attempts to download the stable channel's current hash. [Here's the error itself](http://i.imgur.com/YR9P7fT.png).

I want to say this is due to [this rust-native-tls issue](https://github.com/sfackler/rust-native-tls/issues/3), but I'm not certain.

For reference, I'm on Windows 7.

liigo:
on Windows 10, I got this error (and i think this is also a network error, based on the os error code 10054):

```
C:\Users\liigo>rustup update
info: syncing channel updates for 'stable-i686-pc-windows-msvc'
info: syncing channel updates for 'beta-i686-pc-windows-msvc'
 72.8 KiB /  72.8 KiB (100 %)  67.6 KiB/s ETA:   0 s
info: downloading component 'rustc'
 35.0 MiB /  35.0 MiB (100 %)   2.4 MiB/s ETA:   0 s
info: downloading component 'rust-std'
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 10054, message: "\u{8fdc}\u{7a0b}\u{4e3b}\u{673a}\u{5f3a}\u{8feb}\u{5173}\u{95ed}\u{4e86}\u{4e00}\u{4e2a}\u{73b0}\u{6709}\u{7684}\u{8fde}\u{63a5}\u{3002}" } }', ../src/libcore\result.rs:785
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread '<main>' panicked at ': "PoisonError { inner: .. }"', ../src/libcore\result.rs:785
stack backtrace:
   0:   0x59fefa - <unknown>
   1:   0x57dcd3 - <unknown>
   2:   0x57f3f0 - <unknown>
   3:   0x595892 - <unknown>
   4:   0x49aa55 - <unknown>
   5:   0x4bf50c - <unknown>
   6:   0x49e6a3 - <unknown>
   7:   0x495667 - <unknown>
[...]

PR #429, opened 2016-05-10
Don't panic when the NativeSslStream lock is poisoned

brson:
This is just one small fix. More significant ones, including fixes for hanging downloads are in https://github.com/sfackler/schannel-rs/pull/1, which will automatically show up in rustup builds.

This is probably not going to resolve all the problems seen in #423, but it will improve things.


brson:
I want to get this merged and publish a new release with the schannel fixes. Then I'm going to switch Windows back to OpenSSL.

PR #431, opened 2016-05-10
Support HTTP protocol again

brson:
This means that setting

RUSTUP_DIST_ROOT=http://static.rust-lang.org/dist

and

RUSTUP_UPDATE_ROOT=http://static.rust-lang.org/rustup/dist

will workaround current Windows TLS problems.


brson:
I'm going to go ahead and merge this and get a new staging build started.

PR #434, opened 2016-05-10
Migrate to using libcurl for now instead of hyper

alexcrichton:
Unfortunately the move to rust-native-tls has had some hiccups on Windows and
Hyper also currently doesn't support features like HTTP proxies (but support's
coming soon!). For now this temporarily moves over to using libcurl to weed out
these problems until we get to a point where we're hyper's features are more
filled out. This also somewhat aligns with Cargo as well which is using libcurl
currently.

This does not use the `curl` crate on crates.io because it doesn't support
streaming downloads. Instead I've got an in-progress rewrite which is just
binding much more directly to libcurl's APIs (e.g. exposing the callback
functions directly). This branch is in a git repo currently and I hope to
upstream it to curl-rust soon as well.

### DIFF
--- a/src/rustup-utils/src/raw.rs
+++ b/src/rustup-utils/src/raw.rs
@@ -1,19 +1,22 @@
 use notifications::NotifyHandler;
 
-use std::error;
-use std::fs;
-use std::path::Path;
-use std::io;
+use std::cell::RefCell;
 use std::char::from_u32;
-use std::io::Write;
-use std::process::{Command, Stdio, ExitStatus};
+use std::error;
 use std::ffi::{OsStr, OsString};
 use std::fmt;
+use std::fs;
+use std::io::Write;
+use std::io;
+use std::path::Path;
+use std::process::{Command, Stdio, ExitStatus};
+use std::str;
 use std::thread;
 use std::time::Duration;
-use hyper::{self, Client};
+use curl::easy::Easy;
 use sha2::{Sha256, Digest};
 use errors::*;
+use url::Url;
 
 use rand::random;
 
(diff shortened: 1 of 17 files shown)


alexcrichton:
I wouldn't merge this just yet as I'm still filling out some details on the Windows side of libcurl, but hopefully those should be addressed shortly.

src/rustup-utils/src/raw.rs
```diff
@@ -152,161 +154,84 @@ pub fn tee_file<W: io::Write>(path: &Path, mut w: &mut W) -> io::Result<()> {
     }
 }
 
-pub fn download_file<P: AsRef<Path>>(url: hyper::Url,
+pub fn download_file<P: AsRef<Path>>(url: &str,
```
brson: Can you instead switch to `url::Url` instead of `str` (it's already used instead of `hyper::Url` everywhere els)? I hope to push Urls upward through the APIs to improve type safety.

brson:
Thanks @alexcrichton

alexcrichton:
Updated to thread `Path` through much farther, I'll ping once CI for curl is green

alexcrichton:
I *think* this is good to go but I'd prefer to wait for Windows CI to come along and agree. That.. may take awhile.

brson:
I killed the builds in front of it in appveyor. As soon as the first target is green I'll merge and start the build on master.

alexcrichton:
This probably wants to wait at least until the MinGW target gets green, I still have yet to see a green build of curl on appveyor  unfortunately. I *think* I've fixed it locally though.

brson:
@alexcrichton Failed on gnu.

PR #468, opened 2016-05-15
Prioritize adding path to '.bash_profile' over '.profile'.

mlschneid:
If you use '.bash_profile' on osx, '.profile' gets ignored, which prevents the changes to $PATH from being picked up. 

I don't have any experience in rust, so this code may not be idiomatic.

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -795,7 +795,16 @@ fn get_add_path_methods() -> Vec<PathUpdateMethod> {
         return vec![PathUpdateMethod::Windows];
     }
 
-    let profile = utils::home_dir().map(|p| p.join(".profile"));
+    let mut bash_profile = utils::home_dir().unwrap();
+    bash_profile.push(".bash_profile");
+
+    let profile;
+    match bash_profile.exists()
+    {
+        true => profile = utils::home_dir().map(|p| p.join(".bash_profile")),
+        _ => profile = utils::home_dir().map(|p| p.join(".profile"))
+    }
+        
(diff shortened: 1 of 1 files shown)


Diggsey:
I'm not sure how far we should go down this road, but if we're going to do this, then IMO we should add the definition to *both* files, in the event that `.bash_profile` exists.

Based on the info here: http://hayne.net/MacDev/Notes/unixFAQ.html#shellStartup

It seems that bash looks for these files, in order: `.bash_profile`, `.bash_login`, `.profile`, and it will only read *one* of those files. If we always add it to `.profile`, and then add it to any additional `.*_profile` or `.*_login` files (I don't mean actually wildcard them, we list the possibilities exhaustively) then we ensure that each process gets its PATH set exactly once, regardless of whether it is a bash shell or not.

mlschneid:
I'm terrible at git and somehow smushed and flipped commits around, but all of it together should be the fix. I think it's a bit odd to have the filenames in the code, but I don't see a config file. 

Before I do any more damage, I figured I should ask. Should I squash the commits?

brson:
Thanks for the patch.

Since there are several problems related to PATH config, I want to think a bit harder about this PATH situation before committing to more solutions. It also doesn't work on [fish shell](https://github.com/rust-lang-nursery/rustup.rs/issues/478).

I previously switched from detecting the presense of `*rc` files and modifying them to just blanket modifying the `.profile` file whether it preexists or not. Now I'm leaning toward a maximal solution of just modifying every relevant startup file we know of.

mlschneid:
No sweat captain.

PR #476, opened 2016-05-16
Compile on stable Rust

### LINKED ISSUE #58: Build on stable Rust
The main blocker for this is the use of `#[fundamental]` in the notification system. I haven't looked into it in too much detail, but I think this can be done by reducing the amount of typegineering going on. Thoughts?

alexcrichton:
This removes the usage of #[fundamental] and intrinsics::type_name in the
notifications module. The `Notifyable` trait and `Notifier` struct were also
removed in favor of just using `&Fn(Notification)` where necessary. This change
is less ergonomic when crossing crate boundaries where `Notifier` naturally
implemented `Notifyable` for multiple notification types before. Instead, now
closures must be written as:

    &|n| prev_handler(n.into())

Which is to say that when crossing crate boundaries you need to remap
notification closures and leverage a `From` implementation.

This crate does not currently compile on the stable *channel* due to the usage
of `panic::catch_unwind` in the curl crate for now, but that will get fixed with
the next stable release, this is just paving the way forward!

Closes #58


Diggsey:
Nice! This isn't a criticism, but why did you choose to explicitly call `.into()` rather than making the existing `ntfy!(x)` macro expand to `&|n| x(n.into())`?

alexcrichton:
Ah yeah that'd be possible, I just figured that the layer of a macro indirection probably wasn't worth it.

brson:
lgtm

PR #496, opened 2016-05-24
Adding TOOLCHAIN_HELP to rustup help install

peschkaj:
Adds the help from `TOOLCHAIN_HELP` to `rustup help install`. There are likely other places we can add this and get additional help about `<toolchain>` into rustup.

This should fix https://github.com/rust-lang-nursery/rustup.rs/issues/483

### DIFF
--- a/src/rustup-cli/help.rs
+++ b/src/rustup-cli/help.rs
@@ -25,7 +25,19 @@ stable, beta, and nightly toolchains from the official release
 channels, then updates rustup itself.
 
 If given a toolchain argument then `update` updates that toolchain,
-the same as `rustup toolchain update`.";
+the same as `rustup toolchain update`.
+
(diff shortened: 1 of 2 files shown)


peschkaj:
@brson Does this work for you?

brson:
OK, sorry for the delay.

I'd prefer to see the help for `update` and `install` be similar, with an INSTALL_HELP text that is substantially similar to UPDATE_HELP. If `<toolchain>` isn't sufficiently explained then it needs to be explained in both places.

Instead of explaining in full what a toolchain is perhaps these two could contain more abbreviated text with pointers to more information:

> 'toolchain' specifies a toolchain name, such as 'stable', 'nightly', or '1.8.0'. For more information see `rustup help toolchain`.

peschkaj:
:+1: will revise.

Thanks for the eyeballs.

brson:
Thanks!

PR #512, opened 2016-06-03
Filter whitespace lines from components file.

withoutboats:
Old versions of multirust may have inserted an extra new-line character
into the components file. This makes upgrade robust to files that
contain whitespace lines by not attempting to construct components with
empty names.

### DIFF
--- a/src/rustup-dist/src/component/components.rs
+++ b/src/rustup-dist/src/component/components.rs
@@ -61,6 +61,7 @@ impl Components {
         }
         let content = try!(utils::read_file("components", &path));
         Ok(content.lines()
+                  .filter(|s| s.chars().all(char::is_whitespace))
                   .map(|s| {
                       Component {
                           components: self.clone(),


brson:
Thanks. Do you know which older versions of multirust did this? I can't think of a scenario where this would arise.

withoutboats:
I downloaded multirust on this machine sometime in November/December 2015, and last updated stable quite a while ago (my best guess is I had 1.6.0, but I didn't check). Solving this didn't actually unblock me, and I ended up completely deleting my .multirust to do a fresh install. I encountered all of these issues:

* My stable toolchain was in a directory called `stable` instead of having my target included in the dir name (I solved this by renaming the directory)
* My `components` file had an extra new line in it. I solved this by removing the newline.
* After resolving that I hit [this assertion](https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup-dist/src/component/transaction.rs#L94), but I wasn't able to determine what path was failing the assertion before I gave up and decided to a totally clean install.

brson:
Hm, the rustup installer is supposed to prevent rustup from ever seeing multirust's metadata - during install it should refuse to install when it sees the outdated `.multirust` directory containing the old metadata version. This check was only added pretty recently though.

rustup itself though started using a metadata scheme that was substantially similar to multirust's but not exactly the same. This version was called "2" in both multirust and rustup, and rustup still [claims to be able to upgrade from it](https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup/config.rs#L158). I think probably that upgrade routine should no longer support upgrading from version "2".

When did you install rustup?

withoutboats:
Here's my shell log up until I got into the "arbitrarily broken" state.

```
~ > curl https://sh.rustup.rs -sSf | sh
info: downloading installer
warning: it looks like you have an existing installation of multirust
warning: rustup cannot be installed alongside multirust
warning: run `/usr/local/lib/rustlib/uninstall.sh` as root and delete `~/.multirust/version` before installing rustup
error: cannot install while multirust is installed
rustup: command failed: /var/folders/0d/swcqb0mx2j52mx0tkty_q3pxk003hv/T/rustup.HvCSN50g/rustup-init
~ > sh /usr/local/lib/rustlib/uninstall.sh
rm: /usr/local/lib/rustlib/install.log: Permission denied
install: error: failed to remove old installation log
~ > sudo sh /usr/local/lib/rustlib/uninstall.sh
Password:
install: uninstalling component 'multirust'

    multirust is uninstalled.

~ > rm .multirust/version
~ > curl https://sh.rustup.rs -sSf | sh
info: downloading installer

Welcome to Rust!

This will download and install the official compiler for the Rust programming
language, and its package manager, Cargo.

It will add the cargo, rustc, rustup and other commands to Cargo's bin
directory, located at:

  ~/.cargo/bin

This path will then be added to your PATH environment variable by modifying the
profile file located at:

  ~/.profile

You can uninstall at any time with rustup self uninstall and these changes will
be reverted.

WARNING: This is beta software.

Current installation options:

     default toolchain: stable
  modify PATH variable: yes

2) Customize installation
3) Cancel installation
1

error: toolchain 'stable' is not installed
info: caused by: not a directory: '~/.multirust/toolchains/stable-x86_64-apple-darwin'
rustup: command failed: /var/folders/0d/swcqb0mx2j52mx0tkty_q3pxk003hv/T/rustup.29G4hldf/rustup-init
```

Perhaps if 'stable' is not installed, it should suggest that if you are upgrading from multirust you should delete `.multirust` to perform a clean install?

brson:
Thanks so much! I see exactly what's wrong. The instructions for deleting `~/.multirust/version` are bogus. They should say to delete `~/.multirust` completely.

withoutboats:
@brson That makes sense :-) Closing this since it wasn't enough to solve the problem in the first place.

PR #514, opened 2016-06-03
Use Unix line endings

brson:
The .gitattributes file was using the wrong syntax.


alexcrichton:
Gah that may have been my fault :(

Issue #411, opened 2016-05-07
Make `rustup default x86_64-unknown-linux-gnu` do something smarter.


brson:
That is, naming a toolchain by *just* the target triple might do something useful. I've seen others type the wrong thing and I've done it myself. Especially since `rustup target` takes bare target triples it's easy to confuse the two.

At the least, the `default` and `update` commands should use a heuristic to detect this and offer a suggestion.

That's probably the best way to start. With `default` if we see something that looks like a bare target, then, if the default toolchain is from a release channel, instead print an explanation and say 'did you mean `nightly-$triple`.

If the `update` command, then see if there's only one installed toolchain that we can take the channel name from and say 'did you mean nightly-$tirple`.

In both cases if we can't make a guess about what they *meant*, but it does look like a target triple, we must fall back to a generic error about toolchain specs vs target triples.

moosingin3space:
I'll bite. Assuming there is only one `x86_64-unknown-linux-gnu` toolchain available, `rustup default x86_64-unknown-linux-gnu` should select that one. In the case of multiple toolchains, should the user then be prompted to choose the correct one?

brson:
@moosingin3space I'd prefer that rustup not make the selection automatically, but report an error with the correct toolchain. Likewise for multiple toolchains, print them all (up to some limit perhaps). Unfortunately, @inejge already has a PR up.

brson:
Done.

PR #516, opened 2016-06-05
'default' and 'update' heuristics for bare triples

inejge:
[For issue #411]

[Whoops, I started working on this before @moosingin3space left his/her comment. Here goes anyway, sorry if it led to wasted effort.]

This patch lets you use any recognizable target triple with __default__ or __update__. The code will scan all installed toolchains matching the given (partial) triple and offer the user to choose one, or, if there is just one found, to confirm its use. If the command is __default__, the current default toolchain is not included in the list of candidates. The user _must_ choose or confirm the inferred choice; there is no automatic operation. An empty list of candidates is an error. So is not choosing/confirming an offer.

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -4,7 +4,7 @@ use rustup::{Cfg, Toolchain, command};
 use rustup::settings::TelemetryMode;
 use errors::*;
 use rustup_dist::manifest::Component;
-use rustup_dist::dist::TargetTriple;
+use rustup_dist::dist::{TargetTriple, PartialToolchainDesc, PartialTargetTriple};
 use rustup_utils::utils;
 use self_update;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


brson:
Thanks @inejge!

I see that your logic is not quite the same as what I suggested on the issue. From my reading, for both the 'default' and 'update' cases this considers all installed toolchains with a matching triple to be candidates.

I think the logic is right for the 'update' case but needs to be different for 'default': the main reason I expect people to typo this when running 'rustup default' is because they are switching their host platform. So e.g. I start off on `nightly-x86_64-pc-windows-gnu` but want to be on `nightly-x86_64-pc-windows-msvc` so I type `rustup default x86_64-pc-windows-msvc`. If that is the most likely scenario, then we can assume they meant `nightly-x86_64-pc-windows-msvc` and suggest that. The logic for this would be: read the current default and pull out the channel name; prepend that channel name to the specified triple to get the suggested triple.

I'm also not sure I want to automatically use the channel name, nor prompt for it, though I can see why one might, and I can see why that's consistent with other areas of rustup that try to just 'do the right thing', notably the command synonyms. I've previously imagined printing out an error suggestion but forcing the user to type it correctly. Have to think about it some.

inejge:
> I think the logic is right for the 'update' case but needs to be different for 'default': the main reason I expect people to typo this when running 'rustup default' is because they are switching their host platform.

Oh, I see -- I misunderstood the example in the original comment on the issue. I'll change the logic for 'default' accordingly, just let me know whether you think that rustup should prompt for a toolchain name or error out.

inejge:
Here's a version of the patch which implements the 'default' logic as it should have in the first place, and terminates with an error in all cases. The triple provided to the subcommand doesn't have to be complete, i.e., it should be possible to enter something like `rustup default msvc` and get switched to `nightly-x86_64-pc-windows-msvc`.

brson:
Thanks @inejge! I'm so sorry I've been slow to respond to this.

PR #526, opened 2016-06-10
Reduce margin

### LINKED ISSUE #525: Installer shows extra blank lines on Windows 7
When I run rustup-init.exe on my Windows 7 system, there are a number of extra blank lines:

* Between "This path will then be added to your PATH environment variable by modifying the" and "HKEY_CURRENT_USER/Environment/PATH registry key."
* Between "You can uninstall at any time with rustup self uninstall and these changes will" and "be reverted."
* ...

rustup wrapping

It appears that if a character (including a space) is written to the last column, a blank link will be displayed.

Diggsey:
80 characters followed by a newline will result in an extra blank line in an 80 character `cmd` terminal. Since this is the default width, we should reduce the margin to 79 characters.

Fixes #525

### DIFF
--- a/src/rustup-cli/term2.rs
+++ b/src/rustup-cli/term2.rs
@@ -265,7 +265,7 @@ impl<T: Instantiable + io::Write> Terminal<T> {
     }
 
     pub fn md<S: AsRef<str>>(&mut self, content: S) {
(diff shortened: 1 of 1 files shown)


Diggsey:
This is a massive change, so review carefully :stuck_out_tongue_winking_eye:

PR #532, opened 2016-06-14
Bring back hyper with proxy support

### LINKED ISSUE #509: Reintegrate hyper
@sfackler has updated [native-tls to use new schannel bindings](https://github.com/sfackler/rust-native-tls/blob/master/Cargo.toml#L16-L17). Let's add support back for hyper, alongside curl, switchable at runtime I think.

This will involve updating the rustup-utils crate to have two different download functions, the existing one that uses curl, and the new one that uses hyper and native-tls. Look at the revert of 5fc6377aa6d6fa61d2f90ad2e8f67e4a68b83623 to figure out how to write hyper + native-tls.

If the `RUSTUP_USE_HYPER` env var is set then use hyper, otherwise use curl.

cc @seanmonstar

inejge:
Reintegrate hyper support and make it selectable at runtime by setting the environment. Add proxy support to downloads handled by hyper (also activated by setting the environment.)

Fixes #509

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,6 +104,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "cookie"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "time 0.1.35 (registry+https://github.com/rust-lang/crates.io-index)",
+ "url 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "core-foundation"
+version = "0.2.2"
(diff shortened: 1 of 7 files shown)


brson:
I retried just to see what happens.

alexcrichton:
I believe the CI error means that Cargo.lock may not be up to date and the CI builders are trying to modify the readonly copy. Perhaps a `cargo build` locally and being sure to check in that updated Cargo.lock would work?

inejge:
> I believe the CI error means that Cargo.lock may not be up to date and the CI builders are trying to modify the readonly copy.

Exactly right. Travis builds are now passing, thanks and sorry for the noise.

brson:
Thanks @inejge!

Issue #531, opened 2016-06-13
ppc64le not supported


lu-zero:
Would be possible to add builds for the platform as well?

brson:
Sure! I put a PR up at https://github.com/rust-lang-nursery/rustup.rs/pull/534

lu-zero:
Uhm, the pr is merged but rustup is still not happy:

```
curl https://sh.rustup.rs -sSf | sh
rustup: unknown CPU type: ppc64le
```

brson:
@lu-zero I posted [another PR](https://github.com/rust-lang-nursery/rustup.rs/pull/647) to try to fix. 

Until that's deployed you can download the installer directly from https://static.rust-lang.org/rustup/dist/powerpc64le-unknown-linux-gnu/rustup-init

lu-zero:
That's great, thank you :)

PR #534, opened 2016-06-15
Add powerpc builds

brson:
Completely untested. How does this look @alexcrichton?

cc #531 @lu-zero

### DIFF
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -64,6 +64,22 @@ case $TARGET in
     OPENSSL_CC=x86_64-unknown-netbsd-gcc
     OPENSSL_AR=x86_64-unknown-netbsd-ar
     ;;
+  powerpc-*-linux-*)
+    OPENSSL_OS=linux-ppc
+    OPENSSL_CC=powerpc-linux-gnu-gcc
+    OPENSSL_AR=powerpc-linux-gnu-ar
+    ;;
+  powerpc64-*-linux-*)
+    OPENSSL_OS=linux-ppc64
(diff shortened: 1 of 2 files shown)


alexcrichton:
Ah looks like the build for OpenSSL needs to be updated to build powerpc as well, but otherwise I think this should be all that's needed.

brson:
Updated the openssl builds.

alexcrichton:
Looks like the OS may be `linux-ppc` instead of `linux-powerpc` (and similar for others)

brson:
Updated again.

ci/run-docker.sh
```diff
@@ -64,6 +64,21 @@ case $TARGET in
...
     OPENSSL_AR=x86_64-unknown-netbsd-ar
     ;;
+  powerpc-*-linux-*)
+    OPENSSL_OS=linux-ppc
+    OPENSSL_CC=powerpc-linux-gnu-gcc
+    OPENSSL_AR=powerpc-linux-gnu-ar
+    ;;
+  powerpc64-*-linux-*)
+    OPENSSL_OS=linux-ppc64
+    OPENSSL_CC=powerpc64-linux-gnu-gcc
```
alexcrichton: Oh the failure now is:

```
powerpc64-linux-gnu-gcc: Command not found
```

I think that powerpc64 works by using the `powerpc-linux-gnu` compiler and then passing `-m64` as a flag

brson:
Updated again.

alexcrichton:
Hey look it works!

lu-zero:
Great !

PR #543, opened 2016-06-23
rustup-init: remove dependency on `file` command

zarvox:
Some distros don't install file(1) by default.  For the purposes of userspace
architecture detection, rather than screenscrape file(1), we can check the ELF
header directly with only tools that are included in the default coreutils.

Note that printf is a builtin in some shells (like /bin/dash on Debian), and
may be less featureful than the one from coreutils, but that the syntax here
should work on all such shells.

Tested against bash, dash, and busybox sh.


alexcrichton:
r? @brson 

Seems like a neat trick to me!

brson:
@zarvox you ninja!

PR #544, opened 2016-06-23
When listing toolchains only list directories

### LINKED ISSUE #520: `rustup toolchain list` should only consider directories as toolchains
Was pretty surprised seeing `.DS_Store` as a toolchain since I don't remember installing it ;)

```
$ rustup toolchain list
.DS_Store
```

Of course one can also trick `rustup` by making an empty directory in its `toolchains` directory, but I suppose this is by design

```
$ mkdir $RUSTUP_HOME/toolchains/foo_bar
$ rustup toolchain list
.DS_Store
foo_bar
```

ipetkov:
* Regular files in the toolchains directory will not be considered
  valid toolchain names
* If a user opens their $RUSTUP_HOME/toolchains directory with their
  file explorer, the OS may create some files there (e.g. .DS_Store on
  OS X), which we do not want listed as installed toolchains
* Fixes #520

### DIFF
--- a/src/rustup-mock/src/clitools.rs
+++ b/src/rustup-mock/src/clitools.rs
@@ -171,6 +171,17 @@ pub fn expect_stdout_ok(config: &Config, args: &[&str], expected: &str) {
     assert!(out.stdout.contains(expected), args);
 }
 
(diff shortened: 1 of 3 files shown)


brson:
Beautiful! Thanks for writing a test!

Issue #299, opened 2016-04-12
"target list" shows unavailable toolchains


brson:
On stable, for example, we don't even try to produce NetBSD binaries, so in the manifests these are listed as unailable. The UI though doesn't indicate this so it looks like you can install NetBSD.

andreastt:
I ran into the same problem this afternoon trying to add _i686-unknown-linux-musl_, which is only available on the beta- and nightly channels:

```% rustup target add --toolchain stable i686-unknown-linux-musl
error: component 'rust-std' for 'i686-unknown-linux-musl' is unavailable for download
```

brson:
This is a fairly easy fix. Every component defined in the manifest has an 'available' flag that just needs to be surfaced further.

In `Toolchain::list_components` the returned `ComponentStatus` needs an `available` field, which it can fill in similarly to how `Manifestation` detects unavailable components. For each component in the manifest it needs to look up that component's package definition to find the `available` field. Once `ComponentStatus` has the field then `common::list_targets` can filter it out.

brson:
Fixed. Thanks @Geemili

PR #553, opened 2016-06-30
Remove unavailable toolchains from `target list`

Geemili:
This should close issue #299

### DIFF
--- a/src/rustup/toolchain.rs
+++ b/src/rustup/toolchain.rs
@@ -381,8 +381,8 @@ impl<'a> Toolchain<'a> {
 
                 // Get the component so we can check if it is available
                 let component_pkg = manifest.get_package(&component.pkg)
-                    .expect(format!("manifest should contain component {}", &component.pkg));
-                let component_target_pkg = component_pkg.get(&toolchain.target)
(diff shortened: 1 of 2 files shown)


src/rustup/toolchain.rs
```diff
@@ -381,6 +382,7 @@ impl<'a> Toolchain<'a> {
                     component: component.clone(),
                     required: true,
                     installed: installed,
+                    available: targ_pkg.available,
```
brson: This is almost there, but not quite. The `TargettedPackage` represented by `targ_pkg` is the "rust" package, but you need the `TargettedPackage` for each of the components of "rust" (the "rust" package is made up of other components like "rust-std", each with their own packages).

So in both these loops over components you're going to need to take the `pkg` and `target`  fields of the component and then look up the targetted package in the manifest *for that component*. This is done the same way as the lookups for the previous `rust_pkg` and `targ_pkg` variables.

Thanks!

Geemili:
Thanks for the advice! I think I fixed it now, but I'm not sure.

Also, as a side note, how do I run rustup instead of the installer? I haven't been able to find that information anywhere.

brson:
Thanks @Geemili!

PR #560, opened 2016-07-05
Remove uses of subcommand synonyms in the examples

Armavica:
It can be confusing to see an example using a synonym of a subcommand,
for example `rustup override add 1.0.0`, when `add` does not appear as
a subcommand of `override`.


brson:
Thanks @Armavica !

PR #563, opened 2016-07-06
Add --yes as alias for -y

### LINKED ISSUE #557: Add alias `--yes` for `-y` for ease of use.
I prefer to have longer, more explicit names in configuration/documentation files; while `-y` usually means `--yes`, it seems like a good idea to provide the longer name as well.

natemara:
This should fix #557.


brson:
Thanks @natemara !

PR #570, opened 2016-07-10
Move file downloading to its own crate

brson:
All networking is now in the 'download' function, in the 'download' crate. The crate defines two features, 'curl-mode', and 'hyper-mode', both of which are enabled by default.

r? @alexcrichton


src/download/src/lib.rs
```diff
@@ -1,75 +1,38 @@
...
+    DownloadDataReceived(&'a [u8]),
 }
 
-
 /// Download via libcurl; encrypt with the native (or OpenSSl) TLS
 /// stack via libcurl
-mod curl {
-    use curl::easy::Easy;
+#[cfg(feature = "curl-mode")]
+pub mod curl {
```
alexcrichton: I'd probably recommend extracting this to its own file eventually, having all supported providers in one file may get a little cumbersome eventually

PR #572, opened 2016-07-10
Add preliminary rustls support

brson:
Builds on https://github.com/rust-lang-nursery/rustup.rs/pull/570.

This doesn't actually work yet, but it makes rustls an optional dependency. If you build with `--feature=rustls-backend` then run with `RUSTUP_USE_RUSTLS` it will use https://github.com/ctz/rustls with hyper. Next step is to write code to load the system root cert store.

At the moment this is using my fork of rustls that adds a [concurrency fix](https://github.com/ctz/rustls/pull/4).

r? @alexcrichton 

cc @ctz @briansmith


brson:
cc https://github.com/rust-lang-nursery/rustup.rs/issues/568

src/download/src/lib.rs
```diff
@@ -0,0 +1,695 @@
...
+                })
+        }
+    }
+
+    impl<T> Write for NativeSslStream<T>
+        where T: Read + Write
+    {
+        fn write(&mut self, buf: &[u8]) -> IoResult<usize> {
+            self.0.lock()
+                .map_err(|_| io::Error::new(io::ErrorKind::Other, NativeSslPoisonError))
```
alexcrichton: The `map_err` here is kinda repetitive, perhaps it could be refactored?

alexcrichton:
Looks pretty good to me, modulo error handling. I'd be somewhat wary of making this *too* general as in the limit it'd just be recreating something like curl/hyper, as they're in theory the "general abstraction" for downloading crates.

PR #573, opened 2016-07-12
Fallback to old download methods if server returns 403

### LINKED ISSUE #452: 0.1.11 fails to install new toolchain "rustup toolchain install 1.6.0"
I am receiving this error on x86_64 linux:

```
+ rustup toolchain install 1.6.0-x86_64-unknown-linux-gnu
info: syncing channel updates for '1.6.0-x86_64-unknown-linux-gnu'
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-1.6.0.toml.sha256' to '/home/vagrant/.multirust/tmp/sszoa55jy97dkvg7_file
info: caused by: http request returned an unsuccessful status code: 404
```

I assume the channel listing is only for 1.8.0 (since that's the only fixed version with that toml file) but it is overeager to use it on versions lower than 1.8.0?

brson:
Fixes #452 

Can't be tested because no network tests.

r? @alexcrichton

### DIFF
--- a/src/rustup-dist/src/dist.rs
+++ b/src/rustup-dist/src/dist.rs
@@ -518,7 +518,7 @@ pub fn update_from_dist<'a>(download: DownloadCfg<'a>,
             }
         }
         Ok(None) => return Ok(None),
-        Err(Error(ErrorKind::Utils(::rustup_utils::ErrorKind::Download404 { .. }), _)) => {
+        Err(Error(ErrorKind::Utils(::rustup_utils::ErrorKind::DownloadNotExists { .. }), _)) => {
             // Proceed to try v1 as a fallback
             (download.notify_handler)(Notification::DownloadingLegacyManifest);
         }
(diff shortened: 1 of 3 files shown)


est31:
Nice to see a fix being proposed so fast :)

jethrogb:
Why don't you just test for all 400 codes?

brson:
@jethrogb That sounds good to me if you want to submit a patch.

est31:
Can there be a new release?

masklinn:
@brson since server changes have already failed rustup.sh several time, wouldn't it be possible and simpler to generate the relevant manifests for older releases (at least the "release" channels anyway) and avoid the fallback?

brson:
@est31 I'm working on getting a new release out.

@masklinn That approach is possible, yes. It hasn't seemed worth it since this is also simple and the code already exists.

PR #578, opened 2016-07-13
Treat all HTTP client errors the same

jethrogb:
Per https://github.com/rust-lang-nursery/rustup.rs/pull/573#issuecomment-232223443


brson:
Thanks!

Issue #568, opened 2016-07-10
Implement optional support for rustls


brson:
[Rustls](https://github.com/ctz/rustls) is a pure-Rust TLS library. Just like today we support either curl or hyper + nativetls, add optional runtime support for hyper + rustls. The openssl dep is one of the most troublesome pieces of native code in Rust. It'll be nice to get rid of it.

This will require adding the dependency to `rustup_utils`, then modifying `rustup_utils::raw::download_file` to use hyper + rustls when `RUSTUP_USE_RUSTLS` environment variable is set. Follow the example of `RUSTUP_USE_HYPER`.

RustPowers:
For first, need to fix his API. To use an environment variable to support it seems to me superfluous. When rustls stabilizes, you'll be able to remove openssl dependence in the rustup.

brson:
I'm doing a bit of work on the API right now to clean it up so each HTTP implementation doesn't have to do deal with writing to file and hashing.

@RustPowers The environment variable is there to test the less mature hyper implementation. I don't expect it to stay around forever.

brson:
[This PR](https://github.com/rust-lang-nursery/rustup.rs/pull/572) adds code for a 3rd HTTPS backend, and plugs rustls into hyper. Doesn't work yet. Next step is to write code for loading the system root cert store (this might itself be a useful standalone crate).

inejge:
A brief FYI: I've been working on a PoC cross-platform CA cert loading crate, and the results are promising so far: I can enumerate and dump the system CA stores on macOS and Windows in PEM format, and have the code to distinguish between Unix-like systems (which is needed to get at their CA stores.) I'll need another day or two to get everything in shape for integration with rustup.

brson:
@inejge omg woo!

inejge:
PR opened, CA loading works but rustls connections hang after TLS negotiation, when the first user data is sent/received. (No relation to cert loading, it also happens with hardcoded values.) Will try to investigate further.

brson:
There's a very good chance the read/write code I wrote earlier is wrong since it's never been tested.

est31:
rustls has now had an audit: http://jbp.io/2020/06/14/rustls-audit.html

est31:
Made a PR for it: #2517

PR #585, opened 2016-07-15
Improve rustls CA certificate loading

inejge:
As an enhancement to #568, add a basic cross-platform CA certificate loading crate and make the download crate use it when compiled with rustls support.

The CA loader crate has seen preliminary testing on Windows, OS X, and a couple of Linux distros. It loads the system root CAs on all of those, with the caveats that a) on CentOS, the default CA bundle contains a line which is invalid UTF-8; a fix for rustls to be more robust is ready and a PR will be opened presently, b) Windows and OS X don't use any kind of local policy to filter out certificates, so all certs which can be enumerated are trusted, c) *BSDs, Android, and Solaris haven't seen any testing.

The loader crate presently lacks any documentation and its error handling is rudimentary. That is being worked on.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,6 +93,16 @@ version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "ca-loader"
+version = "0.1.0"
+dependencies = [
+ "crypt32-sys 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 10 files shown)


inejge:
The ARM build that failed did so because of an incomplete installation:
```
curl: (18) transfer closed with 15083604 bytes remaining to read
rustup: command failed: curl -# -C - -f -O https://static.rust-lang.org/dist/2016-05-10/rust-std-nightly-arm-unknown-linux-gnueabihf.tar.gz
rustup: couldn't download 'https://static.rust-lang.org/dist/2016-05-10/rust-std-nightly-arm-unknown-linux-gnueabihf.tar.gz'
```

brson:
Totally badass.

PR #586, opened 2016-07-15
Allow any toolchain to be specified as the default during rustup installation.

### LINKED ISSUE #559: Allow installing rustup without a toolchain (or allow specifying one)
I'm building some Docker containers and throwing Rust in there:

```
RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
```

This will download and install stable Rust, but I need more control over which version gets installed. Specifically, I'd like to just install rustup, and then install a specific `{stable,beta,nightly}-$DATE`:

```
RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --no-toolchains
```

I'd also be happy if I could pick a specific version straight-off:

```
RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --toolchain=beta-1999-12-31
```

I'd be happy to contribute if you have a preferred solution and can give some general pointers as to where to look.

jimmycuadra:
Instead of restricting the `--default-toolchain` argument to `rustup-init` to one of the release channel names, allow any toolchain specification, but warn the user if they do so, since this makes it easier to accidentally install rustup with an invalid toolchain name.

Fixes #559.

### DIFF
--- a/src/rustup-cli/setup_mode.rs
+++ b/src/rustup-cli/setup_mode.rs
@@ -32,7 +32,6 @@ pub fn main() -> Result<()> {
         .arg(Arg::with_name("default-toolchain")
              .long("default-toolchain")
              .takes_value(true)
-             .possible_values(&["stable", "beta", "nightly"])
              .help("Choose a default toolchain to install"))
         .arg(Arg::with_name("no-modify-path")
              .long("no-modify-path")
--- a/tests/cli-inst-interactive.rs
+++ b/tests/cli-inst-interactive.rs
@@ -140,6 +140,18 @@ fn with_non_default_toolchain() {
 }
 
 #[test]
+fn with_non_release_channel_non_default_toolchain() {
+    setup(&|config| {
+        let out = run_input(config, &["rustup-init", "--default-toolchain=nightly-2015-01-02"],
+                            "\n\n");
+        assert!(out.ok);
+
+        expect_stdout_ok(config, &["rustup", "show"], "nightly");
+        expect_stdout_ok(config, &["rustup", "show"], "2015-01-02");
+    });
+}
+
+#[test]
 fn set_nightly_toolchain() {
     setup(&|config| {
         let out = run_input(config, &["rustup-init"],


tests/cli-inst-interactive.rs
```diff
@@ -140,6 +140,18 @@ fn with_non_default_toolchain() {
...
+        let out = run_input(config, &["rustup-init", "--default-toolchain=nightly-2015-01-02"],
+                            "\n\n");
+        assert!(out.ok);
+
+        expect_stdout_ok(config, &["rustup", "show"], "nightly");
+        expect_stdout_ok(config, &["rustup", "show"], "2015-01-02");
+    });
+}
+
+#[test]
```
brson: Thanks for writing tests!

src/rustup-cli/setup_mode.rs
```diff
@@ -44,7 +43,21 @@ pub fn main() -> Result<()> {
...
+                    Custom toolchains are described with a specific syntax. \
+                    If Rust fails to install, you may want to consider installing one of the \
+                    named release channels instead: stable, beta, or nightly."
+                );
+            }
+
+            default_toolchain
+        }
+        None => "stable",
+    };
```
brson: If you don't mind, let's take this warning out. I'm afraid it will look a bit too threatening. For now let's just not do any new error handling, even though if the toolchain fails to install the failure mode is pretty messy.

Instead, I'd prefer to make the way self_update responds to a failed toolchain install more robust, so that, *if* the toolchain install fails, the installer prints a message explaining what happened and how to proceed. But that's a pretty big patch on its own and I think can be done later.

brson:
Thanks for this!

src/rustup-cli/setup_mode.rs
jimmycuadra: Good call. Reverted!

liuchong:
Got here from https://github.com/rust-lang-nursery/rustup.rs/issues/559
so, seems the "--no-toolchains" did not implemented? 😹

PR #590, opened 2016-07-17
Add details about updating rustup to README

wezm:
With the release of `0.3.0` it was unclear to me how to go about updating so I'm proposing adding the following to the README to help others who go looking for the same info in the future.


brson:
Thanks!

PR #592, opened 2016-07-18
Read/Write impl rework for rustls

inejge:
Further tweaks for #568: rustls-enabled rustup can now actually download content. I think that merging should wait, though, since I don't think I thoroughly understand the consequences of the modifications I've had to make both to rustup and rustls in order to make it happen (which is why I've switched it to a private fork for the time being.)

rustls really, really wants to be driven asynchronously, and adapting it to synchronous-style hyper involved manually inserting calls to `write_tls()` after processing input packets in the `Read` impl (without which the protocol negotiation would hang.) Also, `wants_read()` in rustls used to return `true` all the time; I modified it to track the size of its plaintext buffer, so that further `read_tls()` calls are triggered only when that buffer is empty. It seems to work, but I'm not sure what happens when the other side sends an alert, or forcibly closes the connection.

cc @ctz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,6 +83,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bitflags"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


brson:
OK, holding off on merging for further word. Thanks @inejge.

inejge:
After discussing my changes in ctz/rustls#12, I think that they can be merged. I've updated rustls to the latest master, which now supports the usage in this PR, and has a number of other fixes.

Issue #401, opened 2016-05-06
Add HTTP proxy support


brson:
With hyper 0.9.2 you can call `Client::with_http_proxy(host, port)` to set up the HTTP proxy.
I'm not sure how I want to implement this yet, whether there are standard ways to indicate a proxy should be used or not. Whatever the solution it needs to work during the initial install as well as after installation.

cc https://github.com/rust-lang-nursery/rustup.rs/issues/344
cc @seanmonstar

xen0n:
Considering the typical `rustup` invocation is done the `curl | sh` way, I think the environment variables `http_proxy` and/or `https_proxy` are preferable as the same settings are honored by curl as well. Plus it's the *de facto* standard way on Linux at least,  so it's intuitive for the user.

We just need to pull the config from the environment and parse it, as a reference [here][]'s how curl does it.

[here]: https://curl.haxx.se/libcurl/c/CURLOPT_PROXY.html

lilianmoraru:
This won't be in the initial release?

alexcrichton:
Since the switch to curl, this is actually supported in the same standard method that curl supports it now (e.g. standard env vars)

lilianmoraru:
Solved by #532?

brson:
@lilianmoraru I don't quite consider it solved yet. Two things:

* I'd like to verify that the env vars used by hyper are the same used by curl, and a quick googling did not lead me to a description of how the curl env vars work
* I'd like the proxy support documented in the README

Unfortunately I still don't know much myself about how to use this feature.

seanmonstar:
curl environment variables: https://curl.haxx.se/docs/manpage.html heading "ENVIRONMENT":

> The environment variables can be specified in lower case or upper case. The lower case version has precedence. http_proxy is an exception as it is only available in lower case.
>
> **http_proxy [protocol://]<host>[:port]**
>
> Sets the proxy server to use for HTTP.
>
> **HTTPS_PROXY [protocol://]<host>[:port]**
>
> Sets the proxy server to use for HTTPS.

hyper doesn't look for any environment variables. rustup would need to look for them, and enable the proxy settings on the Client itself.

brson:
[...]

PR #598, opened 2016-07-19
Use the env_proxy crate for proxy environment variable handling

inejge:
The crate [env_proxy](https://crates.io/crates/env_proxy) encapsulates the rules for selecting the appropriate proxy server for a target URL according to the contents of the current environment, closely following the convention used by __curl__. This PR replaces the local (and incomplete) proxy determination function with the one provided by that crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,12 +193,21 @@ version = "0.3.0"
 dependencies = [
  "ca-loader 0.1.0",
  "curl 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


inejge:
Should've referenced #401, but I wasn't aware of it. env_proxy replicates __curl__'s env var handling (as documented in the man page) faithfully. As for the README, would you prefer a short section by itself, say "Proxy support" (after "Environment variables"?), or a paragraph somewhere else? I can add the commit with the change to this PR.

brson:
@inejge This looks awesome. For the README, how about another 'working with' section, 'Working with HTTP proxies', after 'working with custom toolchains'? Can you make it a followup? I want to merge this one right now!

PR #600, opened 2016-07-20
Set system-specific dynamic loader env var for command execution

### LINKED ISSUE #588: rustup sets DYLD_LIBRARY_PATH on linux
It probably shouldn’t, as it has no effect on the loader and may be used by some program for other purposes¹.

Similarly it probably shouldn’t set LD_LIBRARY_PATH on OS X either.

¹: Not convinced? I just made one.

inejge:
Fixes #588

### DIFF
--- a/src/rustup/toolchain.rs
+++ b/src/rustup/toolchain.rs
@@ -311,8 +311,15 @@ impl<'a> Toolchain<'a> {
     pub fn set_ldpath(&self, cmd: &mut Command) {
         let new_path = self.path.join("lib");
 
-        env_var::prepend_path("LD_LIBRARY_PATH", &new_path, cmd);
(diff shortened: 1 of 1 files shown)


inejge:
AppVeyor build failure is caused by a truncated transfer:
```
info: downloading component 'rust-std' for 'x86_64-pc-windows-gnu'
error: component download failed for rust-std-x86_64-pc-windows-gnu
info: caused by: could not download file from 'https://static.rust-lang.org/dist/2016-05-10/rust-std-nightly-x86_64-pc-windows-gnu.tar.gz' to 'C:\Users\appveyor\.multirust\tmp\zvrvf3wl0p65yull_file
info: caused by: [18] Transferred a partial file
```

brson:
Thanks!

Issue #597, opened 2016-07-19
Add `no-self-update` feature


brson:
For distro packaging rustup self updates make no sense. [And currently cause mysterious errors](https://github.com/rust-lang-nursery/rustup.rs/issues/471). Add a cargo feature, `no-self-update` that makes `rustup update` not do a self update and `rustup self update` print "error: self-updates are disabled for this build of rustup. If you installed rustup from a package manager, then use it to update rustup."

steveklabnik:
Historically, with Debian, `rubygems update --system` was patched out of RubyGems by debian, causing endless strife between the two camps and a "never install ruby through `apt`" mentality by Rubyists for years.

`rustup` is not as critical as `rubygems` is, but I just wanted to make sure that you're ready for people to be using two-year old versions of `rustup`....

brson:
@steveklabnik That's a good point. It's conceivable that rustup could detect when it has a system installation and do its own self-update in a way that shadows it.

I basically think that people should not be using package managers to install rustup (the upstream installation is supposed to be dead-simple), but packagers gonna package.

Anyway, if it causes problems we'll deal with them. This is fixed.

PR #602, opened 2016-07-20
Add the "no-self-update" feature

inejge:
Cf. #597. In addition to blocking `self update`, this patch also blocks `self uninstall`, with the same rationale.


inejge:
Timeout in one Windows build (it hung at the start.)

brson:
Wow, you are on a roll! Thank you.

PR #606, opened 2016-07-20
Remove the multirust binary. Fixes #373

### LINKED ISSUE #373: Remove multirust binary
I thought I had already done this, but not so.

brson:
This finally removes the binary named `multirust` from the installation.
Instead of hardlinking to `multirust` installation now installs `rustup`
and hardlinks to that. After upgrade, on first run, rustup will silently
delete any existing multirust bin.

Remove multirust_mode module.

r? @alexcrichton


alexcrichton:
dat diff

PR #609, opened 2016-07-21
Update to error-chain 0.2.2

brson:


brson:
I don't think bors is paying attention here.

PR #614, opened 2016-07-25
Print rustup version after update

### LINKED ISSUE #542: Print new version while `rustup self update`
before or after update complete.

julienXX:
Fixes #542

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -43,6 +43,7 @@ use std::fs::{self, File};
 use std::io::Read;
 use tempdir::TempDir;
 use term2;
+use regex::Regex;
 
 pub struct InstallOpts {
     pub default_host_triple: String,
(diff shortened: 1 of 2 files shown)


src/rustup-cli/self_update.rs
```diff
@@ -1069,7 +1071,7 @@ pub fn update() -> Result<()> {
     }
     let setup_path = try!(prepare_update());
     if let Some(ref p) = setup_path {
-        info!("rustup updated successfully");
+        info!("rustup updated successfully to {}", VERSION);
```
brson: By printing `CARGO_PKG_VERSION` here rustup will be printing the version of *the currently running version of rustup*. Instead, what we want is to print the version *we just downloaded* (here represented by `setup_path`). So this is instead going to need to run the executable at `setup_path` with `--version` and parse out the new version number.

brson:
Thanks @julienXX !

julienXX:
@brson Oh okay, I should have thought about that. Thanks!
I updated the PR with getting the version of the downloaded rustup.
Do you have any advice on how I should test this? The current test pass but I don't think it's really adequate.

src/rustup-cli/self_update.rs
```diff
@@ -1069,13 +1070,22 @@ pub fn update() -> Result<()> {
...
     Ok(())
 }
 
+fn get_new_rustup_version(path: &Path) -> Result<String> {
+    let raw_version = Command::new(path).arg("--version").output().unwrap().stdout;
+    let string_version = String::from_utf8(raw_version).unwrap_or(String::new());
+    let re = Regex::new(r"\d+.\d+.\d+").unwrap();
+    let capture = re.captures(&string_version).unwrap().at(0).unwrap();
+    Ok(String::from(capture))
+}
```
brson: Since this code is on the super-important upgrade path, let's be a little more defensive. What happens if the new executable doesn't run? Seems like it could happen. What happens if it fails? What happens if the output of `--version` changes? Seems like we might want to do that (for example returning a version like "2.0.0-rc1"). Instead of unwraps here can they return errors instead? Then have the caller convert those errors to a placeholder string like "(unknown)".

An aside, if the executable fails to run that would be a bad sign - it might be worth thinking about turning that into a hard error and failing the upgrade. But that can be a different PR.

brson:
The test is good enough I think. Yeah, it could be better by having a different version number, but the effort to make that happen isn't worth it.

src/rustup-cli/self_update.rs
julienXX: Yes absolutely! I'll update the code to be more defensive. Do you feel like regexp is the way to go or could a string split be sufficient?

brson: Regexp is good!

brson: Sorry for the slow response.

julienXX:
@brson I updated the code to be more defensive. Also it now exit the process if the executable fails to run.

brson:
Looks great. Thanks @julienXX !

PR #615, opened 2016-07-25
Introduce `+TOOLCHAIN` syntax for proxies.

DanielKeep:
If the first argument to a proxy is of the form `+TOOLCHAIN`, then this
will be interpreted as the toolchain the user wants.  That is,

    cargo +nightly build

will be equivalent to

    rustup run nightly cargo build

This is purely an ergonomic feature designed to reduce typing for people
who frequently switch toolchains (particularly for testing).  It is
inspired by a similar feature in the Python launcher.

**Why `+TOOLCHAIN`?**

The obvious candidate (used by Python) of `-TOOLCHAIN` was a no-go.
The issue is that because toolchains can have arbitrary names, it is
difficult to distinguish between a toolchain and an actual argument to
the proxied program in the general case.

`@TOOLCHAIN` was also considered, as it reads somewhat naturally in
English.  However, it turns out that `@` is a meta character in
Powershell, which causes the next word to simply *disappear* from the
command line.  Presumably, there is a reason for this.

`+TOOLCHAIN` was tested and found to not need escaping with CMD,
Powershell, bash, and fish.

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,6 +405,7 @@ Command | Description
 `rustup target add arm-linux-androideabi` | Install the Android target
 `rustup target remove arm-linux-androideabi` | Remove the Android target
 `rustup run nightly rustc foo.rs` | Run the nightly regardless of the active toolchain
+`rustc +nightly foo.rs` | Shorthand way to run a nightly compiler
 `rustup run nightly bash` | Run a shell configured for the nightly compiler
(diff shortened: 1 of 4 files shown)


brson:
Cool idea. I'm not sure what I think of the syntax, but your rationale is good. Let's get a build out and see what others think of it!

Issue #490, opened 2016-05-23
Man pages


alexreg:
It would be great to have a command of the form

`rustup man [TOOLCHAIN] MAN_PAGE`

that displays the man page `MAN_PAGE` (e.g. rustc, cargo) for either the active toolchain or the specified `TOOLCHAIN` if given.

brson:
Yeah this should probably exist, though I'd like to have a solution that works reliably on Windows.

alexreg:
Do the man pages get installed on Windows? Perhaps Unix platforms could display the man page using `man`, but Windows could open an HTML-rendered version in the browser.

brson:
I think it would make sense for upstream Rust to generate HTML versions of the man pages and include them in the docs.

brson:
Based on existing commands I'd expect this to look like `rustup man <program> [--toolchain]`. This should be relatively easy to implement: add the command to `rustup_mode.rs`; lookup the toolchain dir; then just run the command and report any errors.

theindigamer:
Is this already taken? Otherwise, I'd like to do it as a first contribution.

brson:
@theindigamer It's all yours!

theindigamer:
Sorry if these questions sound very naive/elementary to you:

1. Any particular advice for getting started? I've read the `rustup_mode.rs` file and have spotted a few places to make additions such as the `App::new("rustup")` and `match matches.subcommand()`. I also need to add a function `fn man(cfg: &Cfg, m: &ArgMatches) -> Result<()>`.
2. Should the specification have `<command>` (like `rustup which`) or `<program>`? Can you give a couple of examples (with/without the toolchain)?
3. Also, there aren't actual `man` pages for `rustc`,`cargo` etc., or am I wrong about this? As I understand it, the `man` subcommand is just supposed to display the corresponding `help` menu.

sanmai-NL:
Maybe the need and the requirements for this feature could use some more thought?
I personally don't see the advantage over the usual `man` invocation. Is this supposed to become something else than the `-h`/`--help` option used in various contexts (e.g., subcommand)?

brson:
> I personally don't see the advantage over the usual man invocation.

rustup doesn't install the toolchain-specific Rust man pages to the system so running `man rustc` or `man cargo` is just an error today. This command allows the toolchains' man pages to be multiplexed like the toolchains' binaries.

[...]

Issue #332, opened 2016-04-17
Check if `~/.cargo/bin` already on PATH


chriskrycho:
During first setup, especially for users who have used multirust or rustup in the past, it's possible they'll already have `~/.cargo/bin` in their path configuration already. (It's already in my `.profile`, for example.) Instead of _always_ adding it to the end of the user's profiles, maybe check the `PATH` for it first?

brson:
I think I like this idea, though it's a little clever and I'm wary of guessing the user's intent incorrectly. I guess we would check for whether PATH contains the substring ".cargo/bin" or ".cargo\bin".

This does seem like a niche case though - most people installing rustup should not have a preexisting install of Cargo from some other source.

There's also now an interactive option to opt out of modifying PATH for those that know they don't need it.

PR #616, opened 2016-07-26
Add man subcommand.

theindigamer:
Fixes Issue #490 .


brson:
Thanks @theindigamer and sorry for the delay!

theindigamer:
Could you suggest some other issue to fix which is not too big? I can't find any with the A-easy label but I was considering fixing #332

Issue #608, opened 2016-07-21
Do something smart when user attempts to install with sudo


brson:
Running `rustup-init` under sudo will install rustup to the *user's* home directory with the files being owned by root. This is because sudo doesn't change `HOME`. Once this happens all manner of bad things will transpire.

Not sure offhand the best way to prevent this.

As mitigation, on startup rustup could check the owner of its files and see if they are owned by another user.

cc https://github.com/rust-lang/cargo/issues/2892#issuecomment-234378169

Diggsey:
Maybe compare the home directory from `/etc/passwd` of the current user (determined by `whoami`) with the value of the `HOME` environment variable. If they don't match, ask for confirmation before installing there. (Overridable via a `-y` flag or similar)

brson:
@Diggsey yeah, that sounds doable!

brson:
Fixed! Thanks again @inejge.

PR #617, opened 2016-07-26
Try detecting sudo when running rustup-init

inejge:
Re #608. Not all sudo setups keep HOME, e.g., the ones in RH-derived distros don't. For those, a sudo'd installation will not end up in user's home directory, which may be unexpected, but not ugly like rust-lang/cargo#2892. On Debian/Ubuntu and OS X the detection works as expected.

### DIFF
--- a/src/rustup-cli/setup_mode.rs
+++ b/src/rustup-cli/setup_mode.rs
@@ -1,9 +1,44 @@
 use std::env;
+use std::process;
 use self_update::{self, InstallOpts};
 use errors::*;
 use clap::{App, Arg, AppSettings};
 use rustup_dist::dist::TargetTriple;
 use common;
+
+mod sys_check {
+    #[cfg(unix)]
+    pub fn home_mismatch() -> bool {
+	extern crate libc as c;
+
+	use std::env;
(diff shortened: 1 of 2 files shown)


inejge:
The Windows build failure is spurious.

Other failures are much more interesting.

1. Linux builds triggered a "can't get user data" warning inside the sudo check, and messed up expected test output. Unless _/etc/passwd_ is nonexistent/unreadable, which is doubtful, the only other way it could've failed is if the euid under which the program is running didn't have a passwd entry. Not critical, but annoying.

2. OS X builds _massively_ failed because the test harness modifies `HOME`, confusing the sudo check. If Linux builds haven't failed at (1), they would've hit this.

Since the sudo check seems fundamentally incompatible with the way in which the test harness works, I've added an override which skips it when testing. It's a kludge; maybe a `#cfg` would be cleaner?

brson:
Looks great @inejge! I'd prefer this code be sunk into the `self_update` module, which has some other pre-flight checks. I want to do a release shortly though, so I'll do that refactoring myself. Hope you don't mind.

PR #620, opened 2016-07-28
Widen the range of entries seen as toolchains

### LINKED ISSUE #603: `rustup show` does not show installed custom toolchains
If you have a custom toolchain set up with `rustup toolchain link`, it does not appear in the list of install toolchains with `rustup show`.

inejge:
If only directory entries are chosen with `is_dir()`, the list will miss the symlinks/junctions to custom toolchains.

Fixes #603


Diggsey:
Thanks! Could you add a regression test for #603 ?

inejge:
I've added a "show" output match to the existing test for `rustup toolchain link`.

Issue #599, opened 2016-07-20
`rustup show` fails when active toolchain is custom?


brson:
Create Rustup Toolchain

```
rustup toolchain link ios $HOME/rustc-ios
```

Override current directory toolchain

```
rustup override set ios
```

```
$ rustup show
Default host: x86_64-apple-darwin
error: toolchain 'ios' does not support components
info: caused by: invalid toolchain name: 'ios'
```

[Original Report](http://stackoverflow.com/questions/38466493/unable-to-override-rustup-toolchain-for-custom-build-of-ios-toolchain)

inejge:
Notice that the error "toolchain X does not support components" is _tested for_ in `list_targets_custom_toolchain`, so in #621 I couldn't simply return an empty list from `list_components()`. When trying to list targets, this error has an unfortunate addendum "caused by: invalid toolchain name" -- should we differentiate between a custom toolchain and an invalid name somehow? (That would be a separate issue/PR.)

brson:
@inejge from your current patch in #621 it *looks* like you've solved that question?

inejge:
@brson I have, for `rustup show`; what bothered me was the misleading error message displayed for `rustup target list` when a custom toolchain is active. On second thought, one could argue that anyone working with custom toolchains should know what they're doing, and not be hopelessly confused by that message.

brson:
Fixed!

PR #621, opened 2016-07-28
Handle active custom toolchain in "show"

inejge:
Re #599. This is more complicated than it seems, because `show` tries to fetch all components of the active toolchain, and a custom toolchain a) doesn't have them, and b) doesn't even have an enforced name form. This set of patches enforces the __custom-__ prefix for custom toolchains and adds the minimum necessary logic to handle them in `show`. It also updates the README to note the new requirement.

__Edit__: whoops, wrong approach. I think I'll force push the new commits, the original ones are pointless.


inejge:
Does this perhaps need its own test?

brson:
@inejge Yes, please do add a test. There should be some existing `show` tests and custom toolchain tests to crib off of. Thanks.

inejge:
OK, I'll rebase the PR on top of 87890dbd in order to pick up the changes from #620 and group the test with the others in `fn link`, where I think it logically belongs.

brson:
Thanks again @inejge ! Going to deploy this in just a few hours.

Issue #596, opened 2016-07-19
`rustup target remove` does not clean up properly


badboy:
After removing a target, the directory sticks around

``` shell
$ cd ~/.multirust
$ rustup target add mips-unknown-linux-musl                  
info: downloading component 'rust-std' for 'mips-unknown-linux-musl'
 16.7 MiB /  16.7 MiB (100 %)   3.0 MiB/s ETA:   0 s                
info: installing component 'rust-std' for 'mips-unknown-linux-musl'
$ ls ./toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/mips-unknown-linux-musl/lib 
liballoc-e8edd0fd.rlib  ...
$ rustup target remove mips-unknown-linux-musl
$ ls ./toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/mips-unknown-linux-musl/lib
$
```

`mips-unknown-linux-musl/lib` is still there, though it is empty.

brson:
This is a bit tricky to solve. It's kinda fundamental to how the package installers work. The individual installers are designed for installing globally. So there are directories they can't just delete (like `/usr/lib`) and the installer doesn't track any difference between directories that rust 'owns' and those it doesn't.

I'd say the easiest way to fix this is to add a flag to `rustup_dist::component::components::Component::uninstall` called `remove_empty_dirs`. During uninstallation it keeps a list of directories it encounters. After uninstall it walks those directories up to the install prefix, deleting them as it goes.

The reason to do it at the end is because uninstall is going to see the same directories over and over and you don't want to keep making the readdir syscalls.

The reason to make it a flag and not just do it by default is because the installer code is *intended* to be independent of the rustup model where the toolchain dirs are owned by rustup.

thedrow:
Shouldn't this be closed since #634 is resolved?

kinnison:
@thedrow Yep, thanks for catching that.

PR #634, opened 2016-08-02
Remove empty directories after component uninstall

inejge:
Re #596. I've placed the test in _tests/cli-v2.rs_, `remove_target`; there's an almost identical one, with the same name, in _tests/cli-rustup.rs_, which I haven't modified.

__Edit__: the code does the removal pass unconditionally, as opposed to the suggestion in the issue; however, I don't see how the situation where rustup is allowed to remove the component's files, but _not_ directories, could arise. What am I overlooking?

### DIFF
--- a/src/rustup-dist/src/component/components.rs
+++ b/src/rustup-dist/src/component/components.rs
@@ -209,15 +209,15 @@ impl Component {
                 self.dirs.insert(path.clone());
                 while path.pop() {
                     if path.file_name().is_none() {
-			break;
-		    }
-		    if self.dirs.contains(&path) {
-			self.dirs.remove(&path);
-		    }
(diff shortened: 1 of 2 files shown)


brson:
@inejge thank you! Unfortunately it's the end of the day here and I'm tired. I'll try to review over the weekend.

The situation where the directory can't be removed is theoretical. The installation code is intended to be used not just in rustup but for Rust packages anywhere (e.g. the installers on the main website today are the same installation format and they install to the system). I don't feel too strongly about it.

inejge:
Ah, OK. The code should then be usable as is. I've pushed another commit with fixes for a couple of nits I've noticed in the meantime.

brson:
Thanks @inejge!

PR #645, opened 2016-08-10
Support Telemetry with lots of output

### LINKED ISSUE #639: telemetry mode hangs if there are too many errors
## Meta

```
$ rustup -V
rustup 0.5.0 (4be1012 2016-07-30)
$ rustc -V
rustc 1.12.0-nightly (0a3180baa 2016-08-03)
$ cat ~/.multirust/settings.toml 
default_host_triple = "x86_64-unknown-linux-gnu"
default_toolchain = "nightly"
telemetry = true 
version = "12"

[overrides]
```

## STR

```
$ for i in $(seq 1 10000); do echo "fn fail() {}"; done >> so-much-fail.rs
$ rustc so-much-fail.rs 
```

## Expected Result

10000 errors are printed to standard error

## Actual Result

hang, since `telemetry_rustc` (https://github.com/rust-lang-nursery/rustup.rs/blob/4be1012/src/rustup/command.rs#L51) pipes stderr and waits, without reading from it.

moved from rust-lang/rust#34535

TimNN:
This implementation first writes all output to a temporary file, before processing the output from the file. Tested on OS X.

fixes #639.


brson:
Thanks!

PR #651, opened 2016-08-11
Switch builds to stable toolchain

brson:


Boddlnagg:
See my comments in #635 about maybe updating to a more recent nightly instead.

brson:
Nightly instead.

brson:
I don't understand why cargo is trying to update the lockfile here. It doesn't do that locally.

brson:
Oh! It adds a bunch of metadata stuff!

PR #659, opened 2016-08-17
Adds component subcommand

### LINKED ISSUE #658: Add `rustup component` subcommand
The generalized version of `rustup target`. This command will allow the `rust-src` package to be installed once https://github.com/rust-lang/rust-buildbot/pull/102 lands. One complication here is that the `rust-src` package has a target equal to "*" in the manifests.

I think the way this might work is `rustup component add` defaults to the host target, which can resolve to "*". That way you can write `rustup component add rust-src`.

The command might have the same set of subcommands as `rustup target`, and e.g. `add` might look like

```
rustup.exe-component-add
Add a component to a Rust toolchain

USAGE:
    rustup.exe component add [OPTIONS] <component>

FLAGS:

OPTIONS:
        --toolchain <toolchain>
        --target <target>

ARGS:
    <target>
```

Diggsey:
Fixes #658

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -192,7 +192,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-	http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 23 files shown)


Diggsey:
Apologies for the inclusion of the whitespace changes but bad-column-width tabs are quite painful to edit. Recommend using `?w=1` to hide the whitespace changes ;)

brson:
@Diggsey can you make this commit without the reformatting? It's hard to see what's changing.

brson:
Oh, hm I see you mentioned that....

PR #661, opened 2016-08-18
Deploy MSI build artifacts

Boddlnagg:
This enables deployment of `rustup.msi` and the corresponding SHA of the embedded executable. I don't know exactly how the corresponding AppVeyor settings work, so whoever is responsible for the original setup should have a look at this, since I'm not able to test it locally.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -55,12 +55,12 @@ test_script:
   - cargo test --release --target %TARGET% %FEATURES%
   - ps: |
       if($env:BUILD_MSI) {
-        cd src\rustup-win-installer
+        pushd src\rustup-win-installer
         cargo build --release --target $env:TARGET
         cd msi
         .\build.ps1 -Target $env:TARGET
-        cd ..\..\..
-        if($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
+        popd
+        if($LastExitCode -ne 0) { exit $LastExitCode }
       }
 
 notifications:
@@ -79,6 +79,10 @@ artifacts:
     name: rustup-setup
   - path: dist\$(TARGET)\rustup-setup.exe.sha256
     name: rustup-setup-sha
+  - path: dist\$(TARGET)\rustup.msi
+    name: rustup-msi
+  - path: dist\$(TARGET)\rustup-msi.exe.sha256
+    name: rustup-msi-exe-sha
 
 deploy:
   - provider: S3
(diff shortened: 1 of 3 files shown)


brson:
The windows failure looks legit. Not sure what's going on with the unix builds, but it looks similar to something I saw building an older cargo rev recently. It *looks* like the hash of the openssl tarball changed!

I suspect the appveyor settings for 'artifacts' and 'deployment' aren't quite right but I'll have to experiment with it myself.

Thanks @Boddlnagg.

Boddlnagg:
I'm not sure if this will actually work, since the artifacts `rustup-msi` and `rustup-msi-exe-sha` will only exist for the `BUILD_MSI` configuration. But maybe AppVeyor does the right thing and just ignores the artifacts if the files in `dist` don't exist (it looks like that's what's happening for the cases where `prepare-deploy-appveyor.ps1` just exits without putting anything in `dist`).

brson:
The appveyor failure is unrelated to this pr. Looks like a race in a pre-existing test. Trying to figure it out now.

brson:
I've also fixed the travis failures in a separate PR.

Boddlnagg:
I just rebased this and removed some unrelated changes that belong to another PR. I have more changes ready, but I thing I'm going to wait until this is merged.

brson:
Thanks @Boddlnagg. Fighting fires now but haven't forgotten.

Diggsey:
@brson We need to do something with this - can I do anything to help?

brson:
@Diggsey Yeah we do.

I'm afraid I haven't looked too closely at the solution here, so I'm not sure the state of it.

If you wanted you could try to get it building again and seeing if you can get it to pass the CI.

Boddlnagg:
I can offer some help if needed, though my time is limited.

> I have more changes ready, but I think I'm going to wait until this is merged.

According to this earlier comment of mine, I also have some more changes that build upon this PR ... I am going to see what that was and can try to prepare a PR.

Boddlnagg:
@brson On another note, I think the important thing with respect to MSI build and Windows support is that someone needs to decide about what it should look like from a more general perspective. I have outlined some thoughts [here](https://github.com/rust-lang-nursery/rustup.rs/issues/253#issuecomment-243219376).

(In short: My idea and proposal is to have an MSI-based installer replace `rustup-init.exe` and power `rustup self update`, but *not* more than that. IDEs and editor plugins should provide GUIs for the toolchain management tasks themselves by wrapping `rustup`.)

brson:
@Boddlnagg thanks for the updates. Sorry for avoiding this for so long. I left a comment on the msi issue in response.

alexcrichton:
Closing in favor of https://github.com/rust-lang-nursery/rustup.rs/pull/1211 which looks like it's got some more action

PR #670, opened 2016-08-20
Fix flaky tests

brson:
r? @Diggsey 

Also filed a bug against tempdir https://github.com/rust-lang-nursery/tempdir/issues/15


Diggsey:
LGTM

PR #683, opened 2016-08-23
Fix `component remove` command and add a test for it.

Diggsey:
Just the bare minimum test I'm afraid but it's getting late :sleepy:


brson:
Thanks @Diggsey

Issue #278, opened 2016-04-06
tab completions


brson:
> I've been using it for almost a week and seems great! I only miss tab completions, that would be awesome :slight_smile: thanks Brian!

Hm, after thinking about this further, tab completions are going to be tricky. rustup does not require root to install but completions are usually installed to /etc. I really don't want to require admin just for that. I wonder if we might further frob the shell startup files (like we already to to modify PATH) to load completions.

[Source](https://internals.rust-lang.org/t/beta-testing-rustup-rs/3316/54?u=brson).

japaric:
>  I wonder if we might further frob the shell startup files (like we already to to modify PATH) to load completions.

FWIW, in my case I use `zsh` and I enable the `cargo` auto completions (that `rustup` installs) without touching the `/etc` directory by modifying the `fpath` variable in my `zshrc` as shown [here](https://github.com/japaric/.dotfiles/blob/4a5cdd594fcee65ad5030eb1c56945d13df2301b/.zshrc#L19). I also use the oh-my-zsh framework and I've noticed that I need to modify the `fpath` variable before sourcing the oh-my-zsh framework.

Disclaimer: I don't really grok zsh internals and perhaps there is a way to enable `cargo` autocompletions after the oh-my-zsh framework has been loaded.

P.S. I don't really like that CLI tools modify my zshrc, I usually undo/modify what they do :smile:. I would be happy if `rustup` just installs a `_rustup` completion file somewhere in the `.multirust` folder like it does with `_cargo`.

kbknapp:
@brson  @japaric 

I've just added the ability to [generate bash completions](http://kbknapp.github.io/clap-rs/clap/struct.App.html#method.gen_completions) for a cli with `clap` at compile time. Here's a gist of the [generated completion script](https://gist.github.com/kbknapp/09a37f2fce306573040ea0500e3b1408) for `rustup`

And a demo:

![rustup_compl](http://i.imgur.com/icbrIYe.gif)

[I have a branch on my fork](https://github.com/kbknapp/multirust-rs/tree/bash_compl) where I generated the completions, although before I'd put in a PR there are few opinionated portions that I'd rather a project maintainer look at prior

 * having to move the `rustup_mode::cli` function into a separate file in order to include via the `build.rs`. 
 * there's a strange issue where I couldn't `use help::*;` from `rustup_app.rs` so I had to `include!` it.
 * finally, the output of the completion script goes to  `OUT_DIR`

[...]

PR #689, opened 2016-08-25
Add basic autocompletion for Zsh

knight42:
Unsolved problem:

Should I hard code this file into somewhere and generate it in a proper place, say `~/.cargo/_rustup`(then ask users to copy it to /usr/share/zsh/site-functions), 

or simply let rustup copy this file to `~/.cargo/_rustup`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,19 @@ you are ready to Rust. If you decide Rust isn't your thing, you can
 completely remove it from your system by running `rustup self
 uninstall`.
 
+#### Enable tab completion for Zsh
+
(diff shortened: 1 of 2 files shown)


knight42:
part of #278

Diggsey:
I'm a little concerned about keeping this up to date with the CLI, but we can add it for now on the basis that slightly out-of-date completions are better than no completions, and our CLI aims to be backwards compatible.

However, I'd prefer not to clutter the root directory with this. `src/rustup-cli/zsh/_rustup` should be fine, and we should link to it from the README under installation instructions.

Regarding actual installation, for now just providing them in the repo is better than nothing. When we figure out the best way to install it automatically, we can always `include!` the completions file into the rustup binary.

knight42:
@Diggsey updated as you said :)

Diggsey:
Thanks! Test failure looks spurious, so merging this.

PR #693, opened 2016-08-27
Fix some unsafe code

brson:
Values in 'buf' are accessed through 'pwd' so needs to have greater scope.

r? @Diggsey


brson:
Killing these builds.

Diggsey:
Neither type has a destructor so it's actually "correct" as long as they're at the same scope level, which they are, but changing the order will make it harder to accidentally break those requirements in future so it's probably worth doing anyway.

PR #694, opened 2016-08-27
Pin builds to 2016-08-10 and bump version

brson:
Going to release this to hopefully fix crashes. Should let the CI see what it thinks before merging.

r? @Diggsey


Diggsey:
There really needs to be a better way to bump version numbers...

Anyway LGTM, r+

Issue #591, opened 2016-07-18
Rustup uses backtrace regardless of RUST_BACKTRACE


pravic:
I noticed that `rustup update` scans my HDD every time it been called.

I found that it always initializes a symbol helper (dbghelp!SymInitialize) at begin.

Setting `RUST_BACKTRACE=0` or removing `RUST_BACKTRACE` does not help.

Any explanation/recomendations about it? I tested cargo also, it does not load .pdb every time like rustup does.

Diggsey:
I suspect this is due to the use of the `error_chain` crate, which itself depends on the `backtrace` crate, and is used to capture stack traces within `Error` objects.

pravic:
@Diggsey thanks regardless of I am not so lucky to hear about that.

Any way to disable it? 

I've set `_NT_SYMBOL_PATH` to NULL before calling rustup and it helps partially: rustup scans only root directory of the current drive about PDB files.

brson:
@pravic Does this only happen when rustup ends with an error or does it also happen when rustup exits successfully? (I don't know why any backtrace code would be running on success)

 I'd like to know more about what it means to 'scan your HDD'. Why is this a problem for you?

pravic:
@brson There are no errors in regular `rustup update`, so thats why I complaining about it.

The problem is in **unnesessary** HDD access which is quiet noisy and slowly. And thats despite of incorrect behavior with RUST_BACKTRACE ignoring.

brson:
Here's the strace output of `rustup --help` for the `open` and `stat` syscalls:

```
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=38191, ...}) = 0
open("/lib/x86_64-linux-gnu/libz.so.1", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=104824, ...}) = 0
open("/lib/x86_64-linux-gnu/libdl.so.2", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=14608, ...}) = 0
open("/lib/x86_64-linux-gnu/libpthread.so.0", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0755, st_size=138744, ...}) = 0
[...]

PR #703, opened 2016-08-30
Update to error-chain 0.5.0 to allow optional backtrace. #591

squidpickles:
Thought I'd give this a try. No test failures, and it does seem to short-circuit a bunch of system calls with `RUST_BACKTRACE=0`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,9 +1,32 @@
 [root]
-name = "rustup-win-installer"
+name = "rustup"
 version = "0.6.3"
 dependencies = [
- "rustup 0.6.3",
+ "clap 2.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "download 0.3.0",
+ "error-chain 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


squidpickles:
This is a fix for #591 
Not sure about the build failures. Tests passed locally on Mac and Linux, but I don't have access to a Windows box to debug on. Is this likely a related issue?

brson:
@squidpickles Awesome! This looks great.

The linux failure is because on the bots `cargo build` is causing `Cargo.lock` to be updated, which is considered an error. And I've verified that when I build locally the lockfile also changes. Can you try locally and see if you get the same result, and if so update the lockfile.

Since the build on travis is pinned to nightly-2016-08-10 you might want to try running `rustup override set nightly-2016-08-10` to make sure you've got the same configuration.

The windows failure looks spurious.

Thanks!

squidpickles:
Looks like the `Cargo.lock` update fixed that problem. And, I've got the nightly running in that directory.  Thanks for the help with my help.

brson:
Thank you again. Sorry for leaving this for so long 😧

Issue #475, opened 2016-05-16
Does not work with ~/.cargo/bin not being the most important path


nagisa:
It manifests itself in two ways:

```
$ ~/.cargo/bin/rustup run nightly cargo build --release
   Compiling permutohedron v0.2.1
   Compiling itertools v0.4.13
   Compiling fixedbitset v0.1.0
error: the crate `std` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `core` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `collections` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `rustc_unicode` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `alloc` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `rand` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `libc` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `alloc_jemalloc` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
error: the crate `std` has been compiled with rustc 1.10.0-nightly (d91f8ab0f 2016-05-07), which is incompatible with this version of rustc [E0514]
help: consider removing the compiled binaries and recompiling with your current version of rustc
[...]

PR #707, opened 2016-09-01
making rustup prepend cargo bin to path instead of append

rjgoldsborough:
A first attempt at fixing #475 

I'm not sure if the change is this simple because I'm honestly not sure how to test this one.
But I changed the Toolchain::set_ldpath to prepend instead of append.

If you could share some advice on how to test I'm more than willing to do so :)


rjgoldsborough:
Looking at the CI build I would say it's not that simple

brson:
Looks good. Thanks @rjgoldsborough !

PR #711, opened 2016-09-06
Use released version of rustls dependency

meqif:
Currently master fails to build with:

```
Updating git repository `https://github.com/ctz/ring`
error: failed to load source for a dependency on `ring`
Caused by:
  Unable to update https://github.com/ctz/ring#9a468a7a
Caused by:
  failed to clone into: /buildslave/target/cargo-home/git/db/ring-d33dffdb083ffa66
Caused by:
  failed to authenticate when downloading repository
attempted to find username/password via git's `credential.helper` support, but failed
```

I tracked `ring` down as a dependency of `rustls` in `download`. Updating `rustls` to the latest version (0.1.2) fixes the issue.

I'm not sure how unstable `rustls`'s API is, so I took the safe bet of clamping down to the exact version instead of relaxing to 0.1. Feel free to tell me to change it, though!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,7 +178,7 @@ dependencies = [
  "lazy_static 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "native-tls 0.1.0 (git+https://github.com/sfackler/rust-native-tls.git)",
  "openssl-sys 0.7.14 (registry+https://github.com/rust-lang/crates.io-index)",
- "rustls 0.1.0 (git+https://github.com/ctz/rustls.git)",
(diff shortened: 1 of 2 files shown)


meqif:
Should I check in the changes to Cargo.lock as well?

meqif:
Without checking in the lockfile changes the Windows builds succeeded but the Travis builds failed with:

```
error: failed to write /buildslave/Cargo.lock
Caused by:
  failed to open: /buildslave/Cargo.lock
```

After adding them it seems like those build issues are solved (at the time of writing this comment, the tests are running and at least one succeeded in Travis).

I squashed the individual commits. Please let me know if you'd like me to change anything.

brson:
Thanks @meqif!

PR #723, opened 2016-09-13
Adjusted need_cmd to add uname and remove printf

raphaelcohn:
* Added `need_cmd` for `uname`
* Removed `need_cmd` for `printf` (which is always a builtin)


brson:
Thanks again!

PR #725, opened 2016-09-13
Made outputting of ANSI terminal escapes codes defensive

raphaelcohn:
I'd actually like to do a bit more of a thorough job on this script, for example, tidy up use of the default root, make need_cmd check for grep / head if required, etc depending on detected OS, add support for BusyBox wget as well as curl, etc.


brson:
Thanks!

PR #752, opened 2016-10-10
Do not interpret commas when using "rustup run"

### LINKED ISSUE #665: Comma-separated arguments seem to be split with `rustup run`
Whenever I use `rustup run` it seems to interfere with comma-separated arguments:

```rust
$ rustup run stable rustc --emit llvm-ir,link foo.rs
error: multiple input filenames provided

$ rustc --emit llvm-ir,link foo.rs
$
```

Is there perhaps something going on with the rustup option parser?

jethrogb:
Fixes #665


brson:
Thanks @jethrogb!

PR #767, opened 2016-10-21
Treat an empty `CARGO_HOME` the same as an unset `CARGO_HOME`

Diggsey:
Previously this would result in `$CARGO_HOME == $HOME`, which was *bad* when rustup tries to uninstall itself...


brson:
Thanks!

PR #772, opened 2016-10-23
Check stdout is a tty before using terminal features

kamalmarhubi:
The implementation only checked if stderr was a tty. This lead to escape
codes in output when piped from stdout. For example, this had ANSI
codes show up in `less`:

    $ rustup component list | less

while this did not:

    $ rustup component list 2>&1 | less

even though no output went to stderr.


brson:
Thanks @kamalmarhubi !

PR #773, opened 2016-10-24
Completions

kbknapp:
This commit adds a command line option to generate completion scripts, as well as updates `clap` to 2.16.0 which adds ZSH completion support.

Here's a gif of the new ZSH support and the new `rustup completions` subcommand backing it.

rustup_zsh

Specifically this commit adds 

```
$ rustup completions <shell>
```

Finally, this PR adds some additional help documentation on the `completions` command, and changes the `README.md` slightly to reflect the new additions.

Some notes about the ZSH script generated by `clap`, it supports, subcommands, flags, args, and options. The options also automatically list their `possible_values`, and will automatically *not* list any arguments they `conflict` with. This may not be too applicable to `rustup` since I don't think many of those options are used, but it's good to note for future reference.

Also, positional arguments, such as `toolchain` in a few of the subcommands is changed to uppercase ASCII by `clap` in order to differentiate.

Edit: removed verbiage about `--completions`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,7 +87,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "clap"
-version = "2.15.0"
+version = "2.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "ansi_term 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


Diggsey:
This is awesome 👍 My preference would be to stick with just `rustup completions`, since `rustup self` commands usually take the form `rustup self <subcommand> <args...>` rather than just `rustup self <args...>`

kbknapp:
That's a simple change - just let me know if you need me to change it, or any other aspects and I'd be happy to do so :wink:

brson:
Pretty amazing @kbknapp. I agree with @Diggsey . Let's just have one way to do it.

kbknapp:
I've removed the `--completions` arg. It's now just `rustup completions <SHELL>`

PR #802, opened 2016-11-06
Correctly "detect" host endianness on MIPS

xen0n:
Counter-intuitively, Linux on MIPS reports the same machine name in
`uname(2)` output, regardless of endianness. So the current check is
bogus, and will fail on little-endian boxes. However, because the endianness
is not runtime reconfigurable, binaries are guaranteed to only work on
platforms with the same endianness. Hence, we can rely on compile-time
`cfg()` checks to provide the correct host triple, despite the kernel
not providing that information.


brson:
Thanks!

PR #808, opened 2016-11-13
Remove trailing . from instructions

RHavar:
Even though the . makes sense grammatically, it can confuse users.

Here's a screenshot (macos, default terminal) with the current trailing dot:
http://i.imgur.com/5HqQBEN.png

As you can see, that can easily confuse new users into copying and pasting the command with the trailing . and not sure what went wrong.


brson:
Thanks!

PR #812, opened 2016-11-15
Recursive tool invocations should invoke the proxy, not the tool dire…

### LINKED ISSUE #809: Cargo uses "its rustc" rather than the one mandated by the `rustup toolchain`
There's this text in the README:

> Because the rust-lang/rust tree does not include Cargo, when cargo is invoked for a custom toolchain and it is not available, rustup will attempt to use cargo from one of the release channels, preferring 'nightly', then 'beta' or 'stable'.

And rustup does what it says: It uses e.g. stable Cargo if the custom toolchain doesn't include a Cargo *but* that Cargo uses e.g. stable rustc rather than the rustc in the custom toolchain.

I think this is a regression because I remember this used to "work". As in Cargo used the custom toolchain rustc.

### STR

Sorry this is a little convoluted but avoids having to bootstrap a new rustc to show tho bug:

```
# Install and switch to the nightly toolchain
$ rustup default nightly

# Create a "custom" toolchain which is actually just the nightly toolchain minus Cargo
$ mkdir -p ~/tmp && cd $_

$ cp -r $(rustc --print sysroot)/../nightly-x86_64-unknown-linux-gnu .

$ rm nightly-x86_64-unknown-linux-gnu/bin/cargo

$ rustup toolchain link custom nightly-x86_64-unknown-linux-gnu

# Install and switch to the stable toolchain
$ rustup default stable

# Remove the nightly toolchain
# (to make rustup fall back to stable Cargo rather than to the nightly Cargo)
$ rustup toolchain remove nightly

# Switch to the custom toolchain
$ rustup toolchain default custom

# Finally, show the bug
$ cargo new --bin app && cd $_

$ edit src/main.rs && cat $_
#![feature(no_core)]
fn main() {
    println!("Hello, world!");
}

# custom (nightly) rustc
$ rustc -V
rustc 1.14.0-nightly (cae6ab1c4 2016-11-05)

# stable Cargo
$ cargo -V
cargo 0.13.0-nightly (eca9e15 2016-11-01)

# Cargo uses the STABLE rustc rather than the "custom" rustc!
$ cargo build
   Compiling app v0.1.0 (file:///home/japaric/tmp/app)
error[E0554]: #[feature] may not be used on the stable release channel
 --> src/main.rs:1:1
  |
1 | #![feature(no_core)]
  | ^^^^^^^^^^^^^^^^^^^^

error: Could not compile `app`.

```

cc @brson

brson:
…ctly

The way the proxy was setting up the PATH variable contained two bugs:
first, that it allowed the toolchain path to precede the value of CARGO_HOME/bin;
but second that it didn't add CARGO_HOME/bin to the path at all. The result
was that when e.g. cargo execs rustc, it was directly execing the toolchain
rustc.

Now it execs the proxy, assuming that CARGO_HOME/bin is set up correctly,
and guaranteeing not to run the toolchain tool directly.

Fixes #809

r? @Diggsey cc @japaric @michaelwoerister

### DIFF
--- a/src/rustup-cli/proxy_mode.rs
+++ b/src/rustup-cli/proxy_mode.rs
@@ -34,9 +34,9 @@ pub fn main() -> Result<()> {
 
     // Build command args now while we know whether or not to skip arg 1.
     let cmd_args: Vec<_> = if toolchain.is_none() {
-        env::args_os().collect()
+        env::args_os().skip(1).collect()
     } else {
-        env::args_os().take(1).chain(env::args_os().skip(2)).collect()
+        env::args_os().skip(2).collect()
     };
 
     let cfg = try!(set_globals(false));
@@ -51,6 +51,6 @@ fn direct_proxy(cfg: &Cfg, arg0: &str, toolchain: Option<&str>, args: &[OsString
         None => try!(cfg.create_command_for_dir(&try!(utils::current_dir()), arg0)),
         Some(tc) => try!(cfg.create_command_for_toolchain(tc, arg0)),
     };
-    Ok(try!(run_command_for_dir(cmd, &args, &cfg)))
+    Ok(try!(run_command_for_dir(cmd, arg0, args, &cfg)))
 }
 
--- a/src/rustup/command.rs
+++ b/src/rustup/command.rs
@@ -1,8 +1,6 @@
-use std::env;
 use std::ffi::OsStr;
 use std::fs::File;
 use std::io::{self, Write, BufRead, BufReader, Seek, SeekFrom};
-use std::path::PathBuf;
 use std::process::{self, Command, Stdio};
 use std::time::Instant;
 use regex::Regex;
@@ -16,21 +14,19 @@ use telemetry::{Telemetry, TelemetryEvent};
 
 
 pub fn run_command_for_dir<S: AsRef<OsStr>>(cmd: Command,
+                                            arg0: &str,
                                             args: &[S],
                                             cfg: &Cfg) -> Result<()> {
-    let arg0 = env::args().next().map(PathBuf::from);
-    let arg0 = arg0.as_ref()
-        .and_then(|a| a.file_name())
-        .and_then(|a| a.to_str());
-    let arg0 = try!(arg0.ok_or(ErrorKind::NoExeName));
     if (arg0 == "rustc" || arg0 == "rustc.exe") && try!(cfg.telemetry_enabled()) {
-        return telemetry_rustc(cmd, args, cfg);
+        return telemetry_rustc(cmd, arg0, args, cfg);
     }
 
-    run_command_for_dir_without_telemetry(cmd, args)
+    run_command_for_dir_without_telemetry(cmd, arg0, args)
 }
 
-fn telemetry_rustc<S: AsRef<OsStr>>(mut cmd: Command, args: &[S], cfg: &Cfg) -> Result<()> {
+fn telemetry_rustc<S: AsRef<OsStr>>(mut cmd: Command,
+                                    arg0: &str,
+                                    args: &[S], cfg: &Cfg) -> Result<()> {
     #[cfg(unix)]
     fn file_as_stdio(file: &File) -> Stdio {
         use std::os::unix::io::{AsRawFd, FromRawFd};
@@ -45,7 +41,7 @@ fn telemetry_rustc<S: AsRef<OsStr>>(mut cmd: Command, args: &[S], cfg: &Cfg) -> 
 
     let now = Instant::now();
 
-    cmd.args(&args[1..]);
+    cmd.args(args);
 
     let has_color_args = args.iter().any(|e| {
         let e = e.as_ref().to_str().unwrap_or("");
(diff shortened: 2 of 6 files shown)


michaelwoerister:
Nice! `:D`

brson:
Pushed a test commit to try to understand what's going on on the bots.

brson:
OK, I think I've got the test written correctly.

brson:
Legit failures on windows.

brson:
Well, `toolchain link` seems pretty broken on Windows. I can't get it to work manually. Windows doesn't seem to understand the directory junction. I'm not sure why the rest of the `link` tests are working.

brson:
I think I'm just going to clean this up and merge the fix for unix, and figure out the windows junction problems later.

brson:
OK, I've got a variation that works on windows now.

Turned out to be _3_ bugs preventing linked toolchains from working properly on windows!

First, std's `fs::metadata` does not follow windows directory junctions as I expected, so the "does this toolchain exist" test always failed, then there was the junction bug @Diggsey [fixed](https://github.com/rust-lang-nursery/rustup.rs/issues/823), and finally, I had to hack up the fallback code to copy cargo.exe to its own directory before running it - otherwise the windows process lookup order just always runs the rustc.exe next to it and doesn't bother searching the path.

So there's some hacks here.

r? @alexcrichton

src/rustup-cli/proxy_mode.rs
```diff
@@ -34,9 +34,9 @@ pub fn main() -> Result<()> {
 
     // Build command args now while we know whether or not to skip arg 1.
     let cmd_args: Vec<_> = if toolchain.is_none() {
-        env::args_os().collect()
+        env::args_os().skip(1).collect()
     } else {
-        env::args_os().take(1).chain(env::args_os().skip(2)).collect()
+        env::args_os().skip(2).collect()
```
brson, line 39: These were incidental changes to improve error reporting in the proxies.

src/rustup/command.rs
```diff
@@ -1,8 +1,6 @@
-use std::env;
```
brson, line 1: This file is again all just incidental stuff.

brson:
Well, this doesn't pass the test suite...

Edit: oh, I updated my toolchain and the test suite no longer passes because of assumptions it contained about the relative path between test exes and rustup-init.exe, so the test failures are on my end.

src/rustup/toolchain.rs
```diff
@@ -67,7 +69,16 @@ impl<'a> Toolchain<'a> {
...
+        // HACK: linked toolchains are symlinks, and, contrary to what std docs
+        // lead me to believe `fs::metadata`, used by `is_directory` does not
+        // seem to follow symlinks on windows.
+        let is_symlink = if cfg!(windows) {
+            use std::fs;
+            fs::symlink_metadata(&self.path).map(|m| m.file_type().is_symlink()).unwrap_or(false)
+        } else {
+            false
+        };
+        utils::is_directory(&self.path) || is_symlink
```
alexcrichton, line 81: Hm I'm somewhat dubious about this change. Perhaps the error is that the call to `fs::metadata` returns `false` for `is_directory`? Couldn't function just be `self.path.exists()`?

alexcrichton: Er to elaborate, this is fine, but I feel like this is just jumping through too many hoops to conclude that something is a directory when all we're interested in is if a path exists or not.

alexcrichton (commented):
Seems ok to me, to be honest I'm not really sure what's going on here though. If it fixes things them I'm fine merging.

I guess I always figured that rustup proxies always had the highest priority in `PATH` and then from those bins we'd delegate further, but maybe that's naive? I also don't understand what a fallback is...

src/rustup/toolchain.rs
Diggsey: @alexcrichton The reason we care that it's a directory is because sometimes extra files get created (think `thumbs.db` or various dot files) and it's better if rustup doesn't confuse them for toolchains.

brson:
Failure seems bogus.

brson:
> I guess I always figured that rustup proxies always had the highest priority in PATH and then from those bins we'd delegate further, but maybe that's naive? I also don't understand what a fallback is...

That's what it should do, yes, and it used to, and this patch makes it do that again.

brson:
The fallback is when rustup uses the nightly (fallback) cargo against a custom toolchain that doesn't contain cargo.

michaelwoerister:
🎉

alexcrichton:
Just logging this because it happened, but I don't think this is something we should change.

I believe this commit [broke Cargo's test suite](https://ci.appveyor.com/project/rust-lang-libs/cargo/build/1.0.567/job/fsgg5rvwt77tqmce) unfortunately. Specifically when `cargo` is invoked it no longer has the rustc bin directory in `PATH`. This means that when Cargo during the tests compiles a dynamically linked binary (dynamically against libstd) the binary is unable to actually find the standard library.

I'm... not entirely sure how to fix this, but I'll try to figure something out.

Issue #814, opened 2016-11-16
Support MIPS/MIPS64 hosts


xen0n:
Rust's support for MIPS is getting close to mature, with the recent addition of cross-compiled MIPS host compilers it's about time we supported MIPS as a viable host platform. Here's my TODO list for that, as both a reminder for myself and hint for any future porting efforts.

* [x] ~~MIPS-specific: proper endianness handling (done, #802)~~
* [x] ~~Probe for MIPS64 n64 triples (done, #815)~~
* [x] ~~Wait for a `libc` release (0.2.18 released)~~
* [x] ~~Bump `libc` dep in `rustup` (done, #847)~~
* [x] ~~Add MIPS artifacts to channel manifest (been the case since 2016-11-17)~~
* [x] ~~Update `rustup-init.sh` (done, #825)~~
* [x] ~~Update related CI configs (not actionable by myself)~~
* [x] ~~Update `www/rustup.js` to include `Linux mips` and `Linux mips64` UA sniffing *after confirming everything works* (#866)~~

xen0n:
Oh, and also CI support. I think this needs input from Rust maintainers.

_EDIT: I believe it's being worked on #797, but seems low-priority indeed._

xen0n:
MIPS hosts are appearing since the `2016-11-17` nightly and beta, successfully installed with local-built `rustup`. Seems we only have to wait for `libc 0.2.18` before the next Rust release.

xen0n:
Considered complete! Only have to wait for the next rustup release which should have the MIPS binaries correctly produced. Closing.

xwafish:
HI, why rust still not work on mips64el ? 

>  curl https://sh.rustup.rs -sSf | sh
info: downloading installer
curl: (22) The requested URL returned error: 404 Not Found
rustup: command failed: curl -sSfL https://static.rust-lang.org/rustup/dist/mips64el-unknown-linux-gnuabi64/rustup-init -o /tmp/tmp.jZ6IYwI0sA/rustup-init

Thank you.

gnzlbg:
@xen0n I am trying to use an MSA intrinsics with MIPS/MIPS64 without any luck, did you managed to do that?

PR #825, opened 2016-11-21
Support MIPS architectures in rustup-init.sh

xen0n:
Also a prerequisite of #814, tested on MIPS64. Should have minimal impact on other supported platforms, unlike the other relevant patches touching C/Rust code.


brson:
Looks great!

PR #830, opened 2016-11-22
Rename ~/.multirust to ~/.rustup

brson:
This renames the ~/.multirust directory to ~/.rustup. The logic for the upgrade is in `do_rustup_home_upgrade`, and is defensive against errors. It amounts to moving an unrecognized `~/.rustup` folder to `~/.rustup.sh`, then moving a `~/.multirust` dir to `~/.rustup`.

If the upgrade fails then rustup will silently continue using `~/.multirust`, though this code path doesn't have tests, and is unlikely to happen...

After the upgrade, or during initial install this also sets up `~/.multirust` as a symlink to `~/.rustup` in case there are tools depending on it. This is intended to be temporary, to avoid breakage while getting it done quickly.

This needs some more manual testing before deployment but is ready for review.

r? @alexcrichton

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -228,6 +228,10 @@ pub fn install(no_prompt: bool, verbose: bool,
         if !opts.no_modify_path {
             try!(do_add_to_path(&get_add_path_methods()));
         }
+        // Create ~/.rustup and a compatibility ~/.multirust symlink.
+        // FIXME: Someday we can stop setting up the symlink, and when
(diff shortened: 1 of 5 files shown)


brson:
rustup.sh is already modified to use `~/.rustup.sh`.

alexcrichton:
r+ from me

Thanks for writing it as you did, this is some crazy logic to follow, but all makes sense to me! (and was easy to read)

brson:
Rebased.

PR #840, opened 2016-11-30
add faq entry for downloading rust source

kairosswag:
This change request is made because I have experienced it to be really hard finding out how to get the rust source code using rustup. It might be obvious for an experienced user, who knows what the components command does, but it wasn't exactly easy to google it and this might help some people who need the source for racer or similar programs.


brson:
Thanks!

PR #842, opened 2016-12-02
Default to MSVC on Windows

brson:
Instead of picking MSVC/GNU based on detection, just default to MSVC.
During install, if MSVC is not detected provide guidance.

This is more predictable and easier to explain.

r? @alexcrichton


alexcrichton:
r+ but looks like there's windows failures

brson:
Fixed.

PR #843, opened 2016-12-02
Statically linked the MSVC CRT

alexcrichton:
This PR updates a few dependencies for support with `-C
target-feature=+crt-static` and then updates CI to produce those relevant
binaries on Windows. This should produce binaries that don't require the MSVC
redistributables and instead work standalone.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,10 @@
 environment:
+  RUSTFLAGS: -Zunstable-options -Ctarget-feature=+crt-static
   matrix:
   - TARGET: i686-pc-windows-msvc
     BUILD_MSI: 1
   - TARGET: i686-pc-windows-gnu
-    MINGW_URL: https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/4.9.2/threads-win32/dwarf/i686-4.9.2-release-win32-dwarf-rt_v4-rev4.7z/download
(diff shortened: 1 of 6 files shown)


appveyor.yml
```diff
@@ -1,9 +1,10 @@
 environment:
+  RUSTFLAGS: -Zunstable-options -Ctarget-feature=+crt-static
   matrix:
   - TARGET: i686-pc-windows-msvc
     BUILD_MSI: 1
   - TARGET: i686-pc-windows-gnu
-    MINGW_URL: https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/4.9.2/threads-win32/dwarf/i686-4.9.2-release-win32-dwarf-rt_v4-rev4.7z/download
+    MINGW_URL: https://s3.amazonaws.com/rust-lang-ci
```
alexcrichton, line 7: This was a bit of an unrelated cleanup, I switched to the url that we're using for Rust and a few other projects, and AWS is generally much more reliable than sourceforge

brson:
This has some conflicts with the openssl-sys dep used by rustup's download crate.

brson:
I may spend some time cleaning up the download crate soon to unblock this.

alexcrichton:
Ok, updated the PR

brson:
openssl-sys still not working right

alexcrichton:
Pushed some updates, let's see how it goes

PR #857, opened 2016-12-15
Bump version to 1.0

brson:
r? @alexcrichton

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,22 @@
+# 1.0.0
+
+No significant changes from 0.7.0.
+
+Contributors: Alex Crichton, Andrew Koroluk, Arch, benaryorg, Benedikt Reinartz,
+Björn Steinbrink, bors, Boutin, Michael, Brian Anderson, Cam Swords, Chungmin
+Park, Corey Farwell, Daniel Keep, David Salter, Diggory Blake, Drew Fisher,
+Erick Tryzelaar, Florian Gilcher, geemili, Guillaume Fraux, Ivan Nejgebauer,
+Ivan Petkov, Jacob Shaffer, Jake Goldsborough, James Lucas, Jeremiah Peschka,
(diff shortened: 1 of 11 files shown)


steveklabnik:
🎊  itshappening.gif 🎊

brson:
@bors retry

intermittent windows io failure

flying-sheep:
will using platform-specific directories (à la https://github.com/rust-lang/rfcs/pull/1615) break semver?

if so, wouldn’t it make sense to implement them first for rustup?

brson:
@flying-sheep. I don't know the answer to the semver question specifically, but releasing rustup 1.0 doesn't mean it can't change. If nothing else rustup can have major releases.

flying-sheep:
yeah, that was what i meant. if it means doing a major releast, better get it in now and do 1.0 in two days instead of doing 2.0 in a week.

tbu-:
@alexcrichton @brson Could we get a fundamental decision in https://github.com/rust-lang/rfcs/pull/1615 first? At least: Yes, we would use platform-specific directories (maybe not these), or no, we don't want to use platform-specific directories at all.

I feel this issue is being swept under the rug.

brson:
@tbu- I'm not sure why you are linking these two issues. This PR has not much to do with https://github.com/rust-lang/rfcs/pull/1615.

If we do decide to make changes to how cargo and rustup store data, it is sure to be a difficult transition process no matter rustup's version number.

I don't think one decision affects another.

brson:
Per my previous comment I don't see rustup 1.0 and XDG as incompatible. I'm going to resume trying to land this despite the raised concerns.

The best thing to complain about for rustup 1.0 is probably security.

bors seems to have ignored by last command.

@bors: retry

Ericson2314:
On a different note, I assume std-aware cargo would eventually lead to rustup 2.0. I hope that means 2.0 in 2017 is likely.

flying-sheep:
OK, thanks @brson for addressing this, and congratulations to 1.0!

> it is sure to be a difficult transition process

we could implement an automatic migration. coupled with a major version bump, this would lead to people reading about where to find their data.

PR #859, opened 2016-12-15
Upgrade build toolchain

### LINKED ISSUE #838: Rustup-init 0.6.5 for ARMv7 causes segmentation fault
I couldn't tell where to report, so I report the issue here.

Rustup-init v0.6.5 for ARMv7 (https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-gnueabihf/rustup-init) causes segfault on Raspberry Pi 2.
I suspect this is caused by https://github.com/rust-lang/rust/issues/35843, because `patchelf --add-needed /lib/ld-linux-armhf.so.3 rustup-init` fixes the issue.

Would you recompile the binary with the latest rustc?

brson:
Upgrade to a newer Rust. I kind of want this to land after 1.0, but whatever.

May fix https://github.com/rust-lang-nursery/rustup.rs/issues/838


brson:
Updated to use a newer toolchain from https://github.com/rust-lang-nursery/rustup.rs/pull/910/files

brson:
Rebased.

Diggsey:
This happened as part of a different PR.

PR #861, opened 2016-12-16
Make executable name case insensitive

upsuper:
This is mainly for solving the issue that on Windows, filename is case-insensitive, but when rustup is called with non-lowercased name, it sometimes fails to handle correctly.

e.g. when you call `rustc.EXE` rather than `rustc.exe`, it would report
> error: command failed: 'rustc.EXE'
> info: caused by: The system cannot find the file specified. (os error 2)

and if you call it `RustUp`, it would report
> error: infinite recursion detected

Although this is mostly a Windows-only issue, I guess it doesn't harm much if we extend this behavior to all platforms, which makes the code simpler.

I tried to write a test, but it seems I cannot test the `rustc.EXE` case in test, because `rustup_mock::clitools::cmd` attaches the executable suffix unconditionally.

### DIFF
--- a/src/rustup-cli/main.rs
+++ b/src/rustup-cli/main.rs
@@ -74,26 +74,27 @@ fn run_multirust() -> Result<()> {
     let arg0 = env::args().next().map(|a| PathBuf::from(a));
     let name = arg0.as_ref()
         .and_then(|a| a.file_stem())
-        .and_then(|a| a.to_str());
+        .and_then(|a| a.to_str())
+        .map(|a| a.to_lowercase());
     match name {
-        Some("rustup") => {
+        Some(ref n) if n == "rustup" => {
             rustup_mode::main()
         }
-        Some(n) if n.starts_with("multirust-setup")||
-                   n.starts_with("rustup-setup") ||
-                   n.starts_with("rustup-init") => {
+        Some(ref n) if n.starts_with("multirust-setup")||
+                       n.starts_with("rustup-setup") ||
+                       n.starts_with("rustup-init") => {
             // NB: The above check is only for the prefix of the file
             // name. Browsers rename duplicates to
             // e.g. multirust-setup(2), and this allows all variations
             // to work.
             setup_mode::main()
         }
-        Some(n) if n.starts_with("rustup-gc-") => {
+        Some(ref n) if n.starts_with("rustup-gc-") => {
             // This is the final uninstallation stage on windows where
             // multirust deletes its own exe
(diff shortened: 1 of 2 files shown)


upsuper:
I noticed this issue because Gecko failed to detect rust. The configure script checks `rustc.EXE` and `cargo.EXE` rather than `rustc.exe` and `cargo.exe`. I have no idea whether this is broken because some code change in Gecko or the upgrade of rustup.

FWIW, the `.EXE` suffix is from Windows's environment variable `PATHEXT`, so strictly speaking it is not controlled by the build system.

upsuper:
To be clearer, the build system finds `rustc.EXE` and `cargo.EXE` (because `rustc.exe` and `cargo.exe` exist), but when it tries to execute them, the error happens and it fails to continue.

luser:
> FWIW, the .EXE suffix is from Windows's environment variable PATHEXT, so strictly speaking it is not controlled by the build system.

To clarify, we use the `which` package from pypi to locate executables in our configure script, and that honors `PATHEXT` on Windows to locate executables when given a filename without an extension.

rillian:
To clarify further, there's some suspicion that this is a Windows bug. Invoking rustc with a differently-cased filename extension seems to work on some Windows 10 releases, but not others.

rvandermeulen:
I had sorta-suspected that this broke with this month's December Win10 cumulative update released on the 13th, but haven't actually tested that theory.

luser:
I don't think this is an OS bug, I think it's just a rustup bug. In fact, I think the bug is right here:
https://github.com/rust-lang-nursery/rustup.rs/blob/17b6d2112e353606c0e6f4464484800bf67fc536/src/rustup/toolchain.rs#L293

If you invoke it as `rustc.EXE`, that check is going to fail, and so it's going to try to invoke `rustc.EXE.exe`.

EPashkin:
I agree with @luser, changing this line to `if binary_str.to_lowercase().ends_with(EXE_SUFFIX) {` solves problem.
While `PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC` 
` std::env::consts::EXE_SUFFIX` filled from https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/env.rs#L17 where it lowercased

EDIT: replace file path to link

brson:
I merged @EPashkin's patch. Thanks @upsuper .

PR #797, opened 2016-11-05
Add mips and s390x builders

brson:

PR #866, opened 2016-12-17
www: MIPS support and cleanups

xen0n:
Adds detection logic for MIPS hosts. Also cleaned up the Android probing in the meantime to probe all ~~AOSP-supported architectures~~ Android targets that is accompanied with `rust-std` builds.

Note: for better user experience, please don't merge until #797 lands and confirming the MIPS binaries are correctly published.

### DIFF
--- a/www/rustup.js
+++ b/www/rustup.js
@@ -14,10 +14,10 @@ function detect_platform() {
     var os = "unknown";
 
     if (navigator.platform == "Linux x86_64") {os = "unix";}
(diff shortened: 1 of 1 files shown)


brson:
Looks good, but waiting for a published mips build. Do you mind making an equivalent patch against https://github.com/rust-lang/rust-www/blob/master/_includes/rustup.js?

brson:
Oh, actually, since this won't publish until the next rustup release, I'm just merging this now, assuming the next release will have mips.

PR #867, opened 2016-12-17
Fix two missing quotes of download error message

xen0n:
n.b. It seems there's a similarly worded commit in the very same file... 😂


brson:
Thanks again @xen0n !

PR #869, opened 2016-12-18
Catch potential misuses of target triples in host triple position

### LINKED ISSUE #826: Unable to install mips toolchain (error reading rustc version)
```
$ rustup --version
rustup 0.6.5 (88ef618 2016-11-04)

$ rustup toolchain add beta-mips-unknown-linux-gnu
info: syncing channel updates for 'beta-mips-unknown-linux-gnu'

  beta-mips-unknown-linux-gnu unchanged - (error reading rustc version)

$ ~/.multirust/toolchains/beta-mips-unknown-linux-gnu/bin/rustc
/lib/ld.so.1: No such file or directory
```

Linux distro: Ubuntu Xenial (amd64)

### LINKED ISSUE #834: "Target not found", even though it exists
If I try to set `nightly-x86_64-unknown-linux-musl` as the default toolchain, I get an error:

    $ rustup default nightly-x86_64-unknown-linux-musl
    info: syncing channel updates for 'nightly-x86_64-unknown-linux-musl'
    error: target not found: 'x86_64-unknown-linux-musl'

But obviously the target exists because I can compile with `cargo build --target=x86_64-unknown-linux-musl` just fine.

xen0n:
Before executing the `default` and `toolchain add` actions, check if the toolchain given has a different triple than host. If so, bail out with a hint explaining the subcommands and suggesting `target add` instead.

This is based on the assumption that one can't execute binaries from a different triple; however there are niche scenarios where it is actually possible, by QEMU user emulation and similar means. We might add a `-f`
switch for forcing the old behavior if it turns out to be necessary for the relevant parties.

Fixes #826.
Fixes #834.

NOTE: Tests are missing now. I'm developing the patch on MIPS64 so the test machinery was unavailable. I'll have to sync my AMD64 box for completing this.

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -438,6 +438,7 @@ fn default_(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
     let ref toolchain = m.value_of("toolchain").expect("");
     try!(default_bare_triple_check(cfg, toolchain));
     let ref toolchain = try!(cfg.get_toolchain(toolchain, false));
+    check_target_triple_misuse(cfg, toolchain)?;
 
     let status = if !toolchain.is_custom() {
         Some(try!(toolchain.install_from_dist_if_not_installed()))
@@ -461,6 +462,7 @@ fn update(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
     if let Some(name) = m.value_of("toolchain") {
         try!(update_bare_triple_check(cfg, name));
         let toolchain = try!(cfg.get_toolchain(name, false));
+        check_target_triple_misuse(cfg, &toolchain)?;
 
         let status = if !toolchain.is_custom() {
             Some(try!(toolchain.install_from_dist()))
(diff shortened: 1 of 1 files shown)


xen0n:
Oh this broke `default_invalid_toolchain` and `update_invalid_toolchain`. It seems pretty hard to get the expected behavior back in this specific case, as `TargetTriple` is pretty much arbitrary so we can't make too many assumptions there. Should I just update the testcases to reflect the new behavior in addition to the new cases?

brson:
Thanks @xen0n.

This seems a reasonable approach, but I have a few nitpicks. First, this doesn't look like it takes into account several scenarios where hosts are commonly mixed. For example it's common to run i686-unknown-linux-gnu on x86_64-unknown-linux-gnu, same for apple-darwin. All four windows targets are common to run on x86_64 windows.

Secondly, I definitely have use cases for installing host compilers for arbitrary platforms so we will need a way to force installation, though maybe not right now.

Third, the way this reports errors by printing directly to stdout is not done anywhere else in the cli except for the installation code. I'd prefer if this could be returned as new error variant and the text printed as the error text, but I admit I don't know how that will look with such a long error message.

Sorry for taking so long to get back to you.

xen0n:
@brson Fair point. I *completely* overlooked the Windows cases when I have *exactly* installed multiple Rust's on one of my Windows tablets some time ago...

I think some sort of architecture compatibility check should be carried out, and installation should be permitted to continue if the triples are compatible. And a force switch definitely is needed.

As for the error reporting, I was trying to follow what the `default_bare_triple_check` function does, which is to provide some explanations in addition to a suggestion. And I thought some formatting can't hurt, so I added some Markdown formatting as well. (BTW the `term2` usage can be really inconsistent throughout several places in the codebase; I'd like to clean them up but currently don't have the time.)

I'm willing to take suggestions and improve this PR, but I won't have much time actually coding recently. I'll ping the PR with commits when I'm again ready to push this forward.

brson:
> (BTW the term2 usage can be really inconsistent throughout several places in the codebase; I'd like to clean them up but currently don't have the time.)

Oh yeah, there are a lot of inconsistencies :(

Diggsey:

PR #875, opened 2016-12-20
Fix browser detection for Linux ppc64 and NetBSD

brson:


brson:
cc https://github.com/rust-lang/rust-www/pull/643

PR #890, opened 2016-12-27
Add MIPS/s390x builders (with PPC64 compilation fixed)

xen0n:
This is a continuation of #797; rebased on master and applied PPC64 workaround. Let's see what the CI's are thinking.


xen0n:
I believe the FreeBSD failure is spurious; [my previous build](https://travis-ci.org/xen0n/rustup.rs/jobs/186953139) was all green while I only reworded the commit message in this PR. I'm not sure of the AppVeyor failure, but since the PR doesn't even touch the Windows side of things it should be irrelevant too.

brson:
I reran freebsd and it was fine. Thanks @xen0n !

PR #894, opened 2016-12-30
Enable auto-resume in curl command.

sam-falvo:
Address (partially) the need for resumability if the server provides
it.  See https://github.com/rust-lang-nursery/rustup.rs/issues/889

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -60,7 +60,7 @@ main() {
     fi
     
     ensure mkdir -p "$_dir"
-    ensure curl -sSfL "$_url" -o "$_file"
+    ensure curl -sSfL -C - "$_url" -o "$_file"
     ensure chmod u+x "$_file"
     if [ ! -x "$_file" ]; then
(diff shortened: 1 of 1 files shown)


brson:
Thanks for the patch. I left a [comment on the ticket about resuming downloads](https://github.com/rust-lang-nursery/rustup.rs/issues/889#issuecomment-271049779).

As written, I think this will be ineffective because the destination is a random temporary directory. To make this work I think the script would need to create a directory in ~/.rustup, and put the file there. Even still, one has to figure out what to name the file to find it later. We can't just download `rustup-init` and resume it later because the contents of that file change from release to release; so if we were to resume it it would be possible to splice two different binaries together (unless HTTP resume has some recovery mechanism to prevent that).

One way to fix that problem would be to first download rustup-init.sha256, parse the hash out, name the file after the hash. Another way to ameliorate the problem would be to validate the hash after the download.

sam-falvo:
In light of what I've learned, I think I can close and disregard this PR.  This problem definitely needs more study to be robust against concurrent link failures and version updates.  Thanks for the info!

Issue #835, opened 2016-11-25
Please publish rustup to crates.io


joshtriplett:
Please consider publishing rustup releases as a crate on crates.io. That would make it easier to package rustup releases for Linux distributions, or to install rustup via "cargo install rustup" if you have Rust and Cargo already installed via your distribution.  (There's a lot of value in having rustup packages even in a distribution that already has Rust packages, such as to install nightly or to install a cross-toolchain.)

brson:
I'm happy to make publishing rustup part of the release process. I think right now the blocker is that there are some non-crates.io deps, particularly in the download crate.

brson:
Next step is to get rid of git deps in rustup.

The rustup library may need to be broken into another crate and all the libs version numbers dropped to < 1.0.

xen0n:
FWIW `rustfmt` is properly packaged *without crate separation* by putting CLI drivers in the seemingly implicit path `src/bin`. Maybe we could do the same by refactoring `main()` out of `rustup-cli` after re-organizing everything?

What's most important though, is defining how `rustup` as a library should be invoked. Maybe the testsuite way is worth considering?

xen0n:
Also the peculiar way `rustup` currently installs itself, *hard-linking*, may be difficult to implement for Cargo installs, but I haven't looked into it. The library interface thing is actually way less relevant if all we want is just enabling `cargo install rustup`.

brson:
Oh I misunderstood this issue. I was not considering publication of the rustup _binary_, only the libs. I don't know how publishing rustup binaries via cargo would work offhand.

joshtriplett:
@brson I'd like to have the rustup binary crate available via cargo, allowing `cargo install --root=/some/path rustup` to install rustup to `/some/path/bin/rustup`.  That seems similar to installing a binary from any other crate via `cargo install`.

This would require rustup to have all its dependencies on crates.io as well.

infinity0:
Hi I'd like to bump this. Some people would like to be able to install `rustup` in Debian so they can easily get access to rust nightly, which we don't package in Debian.

This would allow people to run `aptitude install rustup` instead of `curl blah://blah/sh | sh`.

kinnison:
So while we support (to some extent) being packaged, we have no intention to release `rustup` onto crates.io at this time because we do not want to encourage anyone to think we're providing a stable API in any sense.  Also we do still have some internal crates which we don't intend to provide as an API in any sense.

[...]

PR #896, opened 2017-01-01
cargo: update to markdown 0.2

lucab:
This switches markdown dependency to the upstream crate, which now
exposes all required bits.

One step further towards #835.


brson:
Thanks!

PR #898, opened 2017-01-02
now officially recommended by rust-lang.org

radix:
The FAQ previously mentioned that rustup will be recommended by rust-lang.org in the future. The future is now!


brson:
Thanks!

PR #911, opened 2017-01-07
add environment variable RUST_DIST_URL_REPLACE for pkg mirror

hjiayz:
windows:
set RUST_DIST_URL_REPLACE=https://pkgserver.com/path/=http://pkgmirror.com/path/
linux:
export RUST_DIST_URL_REPLACE=https://pkgserver.com/path/=http://pkgmirror.com/path/

Signed-off-by: hjiayz <hjiayz@hotmail.com>

### DIFF
--- a/src/rustup-dist/src/manifestation.rs
+++ b/src/rustup-dist/src/manifestation.rs
@@ -131,6 +131,7 @@ impl Manifestation {
             notify_handler(Notification::DownloadingComponent(&component.pkg,
                                                               &self.target_triple,
                                                               component.target.as_ref()));
+															  
             let url = if altered {
                 let mut turl=url.replace(DEFAULT_DIST_SERVER, temp_cfg.dist_server.as_str());
                 for i in temp_cfg.replace_url.as_str().split(";") {
@@ -143,7 +144,6 @@ impl Manifestation {
             } else {
                 url
             };
-			println!("downloadurl {}\n",url);
 
             // Download each package to temp file
             let temp_file = try!(temp_cfg.new_file());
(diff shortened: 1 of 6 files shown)


brson:
Thanks for the patch @hjiayz, and sorry for taking so long to get back to you.

Out of curiosity, can you say more about what you will use this for? Is it to replace HTTPS with HTTP per your example?

I guess I can't see an obvious reason not to do this, except that this is the third solution yet for replacing URLs in the manifests. I'll think about it a bit and make a decision tomorrow.

Diggsey:
@brson Did you come to a decision?

brson:
@Diggsey My current inclination is that without further evidence of demand, or indication of the use cases, I'd rather not move on something like this yet.

As formulated here, this allows one to locate the binaries on some third-party server, while the manifests remain on the official server, or possibly this is combined with RUSTUP_DIST_SERVER in order to redirect _both_ manifests and the bins. Feels a bit awkward.

If the use case is for mirroring static.rust-lang.org, then I might expect a tool that does that mirroring while rewriting the manifests to contain the correct urls and rebuilding the manifest hashes. Maybe, though rewriting the manifests and changing the hashes seems pretty suspicious from a security perspective. Maybe the manifest format is just wrong and shouldn't contain absolute URLs.

The issue of mirroring static.rust-lang.org is particularly important for the Chinese Rust community, and does need a solution.

brson:
It does seem like RUSTUP_DIST_SERVER does this same replacement. That is, if you download the manifest from a mirror using RUSTUP_DIST_SERVER, rustup will also download the bins from that mirror. So this solution must be only for cases where the manifest is on a different server?

Issue #641, opened 2016-08-07
Subcommand `run` does not search PATH for command argument


murarth:
Binaries `cargo install`'d are placed in `~/.cargo/bin` (which is in my shell's `PATH`), but `rustup run <toolchain> foo` does not search `PATH`, instead assuming the named executable to be within the named toolchain's `bin` directory and requiring the full path to successfully run the command. This is a change from the behavior of `multirust`, which did search `PATH` for the named command.

brson:
Can you clarify what you mean by 'requiring the full path to successfully run the command'? rustup should know the path to the toolchain directory and always be able to find the command in the toolchain directory.

I don't know why `run` would search `~/.cargo/bin`. The command is for running binaries out of the toolchain directory.

brson:
Thanks for the report @murarth !

murarth:
I'm not suggesting that `rustup run` should search `~/.cargo/bin` specifically. Rather, I think that it should search the list of directories in the environment variable `PATH` when the named executable is not found within the toolchain. Its primary purpose may be running toolchain binaries, but it's also the most straightforward way to run a binary that is built with a non-default toolchain and affected by #350.

For example, if one uses `cargo install` to build and install [rusti](https://github.com/murarth/rusti), one must run `rustup run nightly ~/.cargo/bin/rusti` in order to run the binary. Running `rustup run nightly rusti` gives a `No such file or directory` error, as `rustup` is not searching `PATH`.

It seems sensible to me that, unless it would conflict with some other concept in `rustup`, the `run` subcommand ought to be thought of as providing the proper execution environment for any executable, not just those that make up the toolchain.

Boddlnagg:
This is basically the same as #210, although that has nothing to do with the `PATH`, but is about running a command from the local directory.

Issue #210, opened 2016-03-30
`rustup run` doesn't work for bins not in the toolchain directory


brson:
"It looks like rustup run <foo> now only works when <foo> is one of rustc or cargo. I've been able to run multirust run nightly ./build-docs.sh with the old multirust when publishing new releases of rust-phf but that now fails:"

[Source](https://internals.rust-lang.org/t/beta-testing-rustup-rs/3316/39?u=brson).

PR #917, opened 2017-01-09
Stop defaulting to $PATH searches when the binary can't be found and causing infinite recursion

Manishearth:
I don't think we should *ever* be falling back to letting the OS resolve on the basis of `$PATH`, but if others still want that functionality I can make this error happen only in the recursion case (when `RUST_RECURSION_COUNT` is 4 or something).

Currently if you have a broken toolchain (e.g. a `rustup link` toolchain to an in-progress build directory that doesn't have the executables in it yet) rustup will default to a `$PATH` search when it can't find the binary, and this will just resolve back to rustup, which causes the infinite recursion error.

Took me a while to figure out that this was a bug with rustup and not my code. Someone else hit it today as well when trying to use rust-gdb, so it might have been introduced recently, though that doesn't seem to be the case.

r? @brson


Diggsey:
The PATH fallback is important functionality in rustup - it's used whenever using a custom toolchain without cargo, and there are several other cases where you want to run a program in the environment for a particular toolchain.

Manishearth:
I don't see how that ever worked? If you have a toolchain without cargo the PATH fallback will just re-resolve to rustup and go in cycles. This works using env vars so unless we explicitly handle this it won't work.

I think we should change this code to explicitly try the default toolchain instead (and emit a warning that it did a fallback), and not do PATH-resolution at all.

Manishearth:
Or, it's only going to work for binaries that themselves don't shell out to rustup. The test that handles this does `rustup run nightly sh`, which will work.

Manishearth:
I made it check the recursion var so that `rustup run nightly sh foo` still works.

petrochenkov:
@Diggsey 
>The PATH fallback is important functionality in rustup - it's used whenever using a custom toolchain without cargo

Is this what happens when you e.g. grab Rust sources, build your own stage1 compiler from them, and use it as a custom toolchain for building Cargo-based projects?
(In this case it is indeed important.)

Manishearth:
> Is this what happens when you e.g. grab Rust sources, build your own stage1 compiler from them, and use it as a custom toolchain for building Cargo-based projects?

That's what _should_ happen, but it doesn't work right now.

Diggsey:
Was on my phone so couldn't give much detail earlier.

This is the PR that added the fallback: #822 
As you can see it fixed two separate issues, #641 and #210 - the `cargo` issue may be me mis-remembering, or it could be that `PATH` semantics have changed since then...

Thanks for making the change anyway, however I think:
- It should be consistent with how `RUST_RECURSION_COUNT` is used elsewhere (where a value of 5 is used)
- It should not be fixed at a very low value, because it will give false positives when tools call each other (eg. cargo calling rustc goes via rustup again)
- If it's going to have a default low value, it should at least be possible to override via another environment variable.

Manishearth:
It has to be less than 5; since it needs to preempt the regular recursion check. I can make it 4.

Diggsey:
Can you make it a named constant, use `<constant> - 1` in this case, and add a regression test.

Rustup has a big problem with changes to one aspect of the tool causing problems elsewhere, and the only way it's going to get better is by making dependencies more explicit, and being meticulous about adding regression tests.

brson:
This PR looks good to me, but there is a legit test failure.

Edit: thanks @Manishearth and @Diggsey

Manishearth:
The test failure arises because we're trying to test what happens when you fallback to the path in a regular install, and the test environment is not a regular install -- it has multiple rustcs (one in the path from Travis, one that goes through rustup). @diggsey said he'd look into fixing this IIRC. I don't recall what solution we'd come up with.

Diggsey:
@Manishearth sorry, lost track of this - I think I have a better way to get this to work rather than messing too much with the CI environment, I'm just testing it out now.

Diggsey:
I decided to temporarily hardlink a new proxy binary to use for the test. This seems to work for the most part, although some of the linux tests are still failing, as they don't seem to pick it up.

brson:
Yeah that looks like a good fix in the test. Nice thinking.

Diggsey:
Travis has decided not to run my build, and my local reproduction of the problem fixed itself before I even changed anything 😡 Who knows, it might be fixed now...

PR #936, opened 2017-02-01
Improve browser detection and install instructions

brson:
r? @steveklabnik


burdges:
How do you override browser detection?  I've used shell scripts for docs when using `rustup.sh`, but..

I recently tried using `rustup doc` and nothing but "firefox is already running, but is not responding" because the browser instance needs a profile to be specified on the command line. 

_Edit:_ Altering `xdg-open` will work I suppose.

PR #949, opened 2017-02-10
Add android support to rustup-init.sh

malbarbo:
We do not have android support eat, but this avoid downloading linux binaries.

For example, on my nexus4 running rustup-init.sh gives:

```
info: downloading installer
... rustup-init: not found
```

The binary for armv7-unknown-linux-gnueabihf was download. The not found error is about the linker ld-linux.so.3.


malbarbo:
Is the fail related with the PR?

Diggsey:
No, it's unrelated. Thanks!

PR #951, opened 2017-02-10
Add fallback to symlink if hardlink fails

malbarbo:
Fixes https://github.com/rust-lang-nursery/rustup.rs/issues/950


malbarbo:
@Diggsey Do you think that this is a viable strategy? Do you have any alternative?

I'm running rustc, cargo and rustup in android, now I'm working in the upstream patches.

Diggsey:
Looks fine to me - did you have some concerns?

malbarbo:
No concerns, just asking.

Issue #889, opened 2016-12-27
Support resume of partial downloads


johalun:
First of all, thanks for a good job. Rustup makes things very convinient and works well while on stable and fast internet.

However, I'm in China on business trip and trying to install Rust using Rustup from here. Other than painfully slow downloads, download is often interrupted and I have to restart from scratch. 

Having a local mirror here in China would be awesome for all the people here but that might be tricky so at least, it would be nice with support for partial downloads and the ability to resume broken download.

Thank you

sam-falvo:
I second this issue.  I'm in San Francisco Bay Area, on 12Mbps (ostensibly!) business-class cable modem connection.  The *actual* connection often drops to single-digit bits per second (!!!) for extended periods of time due to high packet loss (>48%) deep within Comcast's fabric.  In addition, as of my posting this message on December 29, 2016, it looks like AWS is also having exceptionally high packet losses as well.  I'd include a traceroute here if it weren't so transient, and if I had the knowledge that someone on the rustup team had the power to address the issue.

Features desired includes one or more of the following:

- An option to turn off the timeout.  **Alternatively,** an option to set my own timeout.  I can try to add this feature myself (I think it's just a curl option) and submit a PR, but I wanted to document the issue here for reference.
- An option to enable or disable auto-resume functionality.  I'm a little less knowledgeable about how to make this work.  curl has options to support resume, IIRC, but I think logic in the initial shell script will need to be changed to preserve downloaded files.  I just skimmed it, but in skimming, I haven't seen where to make this logic change.

johalun:
I'm not sure but resume might have to be supported by the server as well. If this is not easily done, maybe it could at least cache the downloaded files so that we don't have to re-download all files when it fails on cargo download in the end..

sam-falvo:
I tried to spend some time with the current software, and while I could find the one `curl` command that downloads the next install stage script (the one that actually downloads and installs the components), finding out where *those* install scripts are located has been something of a question mark for me.  Maybe I'm not seeing the obvious, but when I visit https://github.com/rust-lang-nursery, I do not see another candidate repository in which those scripts would reside.

Where else should I look?

Thanks.

kosinix:
Seconded. Everytime rustup fails on my slow internet connection, it has to start all over again its frustrating.  I could not install rust for that matter. Is there a viable workaround for this at the moment? Maybe download each components manually?

brson:
[...]

PR #958, opened 2017-02-19
Store downloaded files in a persistent directory until installation

jelford:
This PR should go some way to addressing #889 by caching downloaded
components in a persistent directory until they have been installed.

This way, if the download/install process is interrupted, the file
that made it can be re-used. This should help ease the pain of
intermittent connections.

### DIFF
--- a/src/download/src/lib.rs
+++ b/src/download/src/lib.rs
@@ -47,21 +47,6 @@ pub fn download(url: &Url,
     Err("no working backends".into())
 }
 
-pub fn download_to_path(url: &Url,
-                        path: &Path,
-                        callback: Option<&Fn(Event) -> Result<()>>)
-                        -> Result<()> {
-    for &backend in BACKENDS {
-        match download_to_path_with_backend(backend, url, path, callback) {
-            Err(Error(ErrorKind::BackendUnavailable(_), _)) => (),
-            Err(e) => return Err(e),
-            Ok(()) => return Ok(()),
-        }
-    }
-
-    Err("no working backends".into())
-}
-
 pub fn download_with_backend(backend: Backend,
                              url: &Url,
                              callback: &Fn(Event) -> Result<()>)
--- a/src/rustup-dist/src/dist.rs
+++ b/src/rustup-dist/src/dist.rs
@@ -8,9 +8,12 @@ use manifest::Component;
 use manifest::Manifest as ManifestV2;
 use manifestation::{Manifestation, UpdateStatus, Changes};
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
+use std::ops;
+use url::Url;
 use std::fmt;
 use std::env;
+use std::fs;
 
 use regex::Regex;
 use sha2::{Sha256, Digest};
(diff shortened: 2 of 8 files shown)


jelford:
I'd like to go further and add support for resuming partial downloads, but I'm not sure what the direction is with e.g. `rustup-dist::download`. It doesn't look used (although it is public, so not sure...? This is my first proper bit of rust, so not familiar with how these things work yet). 

I'd _like_ to refactor the logic I've put into `dist::DownloadCfg` into a new module (which I would call `download`, were it not for the presence of an existing module there), and have that encapsulate All Things Download-Related, but could do with a bit of guidance on what's appropriate / the intended direction of the existing stuff before I go too far. I think what's here already should be useful for some people (it would have been enough to help me out on the recent train journey that inspired the work), but if there's interest I'd be happy to follow up with a new PR to finish the job in the future.

brson:
Hi @jelford. This looks excellent so far. Thank you, and sorry for taking so long.

The download code in rustup is a mess. Curl is the backend we use, so for now I'd say just find that code path and tell curl to do the resume, ignore the others.

In the future I'd like to just delete that crate end use the reqwest crate (https://github.com/rust-lang-nursery/rustup.rs/issues/993).

Is there any benefit to doing this _without_ the download resumption?

jelford:
Hi @brson, thanks for taking a look, no worries about the wait. 

I agree that this would be better with partial-download resumption, but what I'd say is this: even with just support for resuming in the whole-file case, this would have helped me out a lot on the train (and a few people have suggested that as a first step in eg the linked issue) a couple of times now (LTE internet in the English countryside is... not what it could be).

I'll try to find a time to do partial downloads, but can't say how soon (I don't have allocated time for open source), so if you feel you can merge this as it is (or with some not-too-drastic changes), that might be worthwhile.

brson:
> resuming in the whole-file case

Aha. I did not think of that for some reason.

brson:
I feel pretty good about this as-is. My main concern is about garbage collection - what happens to old, unresumed downloads? Are they just going to sit on disk forever?

ISTM it _might_ be appropriate to delete the entire download directory after a successful toolchain installation. Seems like the big concern doing that would be about concurrent rustup execution (which is totally broken today anyway).

cc @Diggsey Do you have any opinions about this feature?

Diggsey:
This looks good for me - I'm not too worried about clean-up, the code already seems to clean up files that were used after a successful install, so the only left over files come from failed download attempts which are never successfully retried.

It would be good to get a regression to test on this (just make sure it outputs "reusing downloaded file" in at least one case when expected).

brson:
Oh good point. I agree this needs a test if we can. I'm not sure how to simulate a failed download locally though. That might require adding some instrumentation to rustup, like if RUSTUP_MOCK_DL_FAIL is set, fail.

For a test case, probably crib off one of the simple cases in cli-v2.

jelford:
Thanks, I'll have a look at adding a test-case, but it might have to wait until later next week. 

I think the garbage-collection situation should be okay, if one presumes that mostly people would aim to retry failed downloads anyway. It will be more of an issue with partial downloads (here, anything partial can get cleaned, so you can say "if the hash-check fails, smoke it" - whereas in the partial case you can't use the hash). The "delete everything on success" solution would certainly work, if it's a worry.

That AppVeyor failure: I don't have a windows machine, so I might need a hand to look into that.

brson:
Don't worry about AppVeyor. The windows build is flaky at the moment.

jelford:
Well, it looks like I did actually find some time. I've just pushed some tests around file re-use. I've also opened up a new branch in my repo with some changes to support partial-download resumption (only for curl, as discussed). 

I'm not sure your preferred workflow: would you like to get what's here finished off (if you have any more changes you'd like) and merged in, then I'll open a new PR for partials, or would you rather just have the whole lot in one go (I can close this PR and open a new one). 

I'd warn that I'm not sure what you'll make of the way I've done the testing for partials (added a local http-server that spins up and server files- actually the code is ripped out of a toy download manager I've been tinkering with), so it would be better not to end up with a bunch of things you don't want to merge getting attached to this one.

brson:
@jelford Let's land this first and do the rest as a followup.

PR #968, opened 2017-02-26
Fix build

nrc:
(on nightly)


nodakai:
I'm curious, how the stable compiler accepted the old version?

brson:
@nodakai rustup is pinned to an old nightly presently and not tested on stable.

Issue #31, opened 2015-10-31
Use exec() to delegate to the real cargo/rustc on Unix


alexcrichton:
Right now the `Command` API is presumably used which is cross platform (yay!) but doesn't provide the best experience on Unix. For example if I `gdb cargo` it doesn't actually gdb the Cargo executable itself but rather the shim. If, however, `exec` were used to spawn the new process then gdb I believe should naturally "just work".

Plus there's the added benefit of fewer processes sticking around!

Diggsey:
Sounds reasonable. When I fix #30 you should be able to do `multirust proxy gdb cargo` too, and it will work regardless.

nagisa:
Needs more priority. People are getting bitten by this when trying to get backtraces for ICEs.

Diggsey:
This may be problematic: https://github.com/rust-lang-nursery/rustup.rs/pull/972#issuecomment-287243676

RalfJung:
@brson 

> This does solve the problem, but I want to solve these two issues a different way, and the main reason is that rustup wants to be able to perform work after the invoked subcommand is run (e.g. recording telemetry about the results). I am not prepared to give up that capability yet.

What kind of telemetry would that be? Is anything like that actually being done? It seems like currently, there are unsolved issues affecting users caused by using fork-exec, for the benefit of some potential future feature. Or have I entirely missed some rustup functionality? (Wouldn't be the first time.^^)

Maybe another option would be to spawn some "watcher" process, exec in the main process (so signals etc. work properly), and have the "watcher" perform whatever has to be done when the subprocess terminates? Not sure however if that would actually be simpler.

EDIT: I just saw `rustup telemetry` for the first time. oO Unfortunately, the help message gives 0 indication for what exactly is covered by "telemetry", but I must have entirely missed this.
Sorry.

geofft:
> Maybe another option would be to spawn some "watcher" process, exec in the main process (so signals etc. work properly), and have the "watcher" perform whatever has to be done when the subprocess terminates? Not sure however if that would actually be simpler.

That seems like the right model, and as an example, `strace -D` does exactly this: "Run tracer process as a detached grandchild, not as parent of the tracee.  This reduces the visible effect of strace by keeping the tracee a direct child of the calling process." Compare what happens if you send a Ctrl-C to, say, `strace python` vs. `strace -D python`.

[...]

PR #972, opened 2017-02-28
Replace rustup with running command

### LINKED ISSUE #806: Ctrl+C on `cargo run`
When I Ctrl+C on cargo run and my Rust program is set up to intercept the signal, the Rustup process stops and returns control to the terminal, even though my Rust program (and indeed Cargo) may still be running. I've observed this on the latest version of Rustup.

I think it would be best for Rustup to do something similar to what `cargo run` does, and intercept Ctrl+C when running a program, and ignore it – instead, it should just check regularly to see if the program (`cargo`) has exited.

Note: this is confirmed not to be a `cargo` issue, as per https://github.com/rust-lang/cargo/issues/3281.

### LINKED ISSUE #845: cargo: exit on signal swallowed by rustup wrapper
See https://github.com/rust-lang/cargo/issues/3367 - the difference:
```
$ cargo run
   Compiling rust-test v0.1.0 (file:///home/gbr/devel/rust-test)
    Finished debug [unoptimized + debuginfo] target(s) in 0.14 secs
     Running `target/debug/rust-test`
$ target/debug/rust-test
[1]    7789 segmentation fault  target/debug/rust-test
```
The exit on signal is hidden by rustup's cargo wrapper.

KalitaAlexey:
I copied the implementation from [Cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/process_builder.rs#L92).

I don't know if I can add any tests for that.

I checked on Ubuntu and Windows.

Fixes #806
Fixes #845

### DIFF
--- a/src/rustup/command.rs
+++ b/src/rustup/command.rs
@@ -136,27 +136,21 @@ fn run_command_for_dir_without_telemetry<S: AsRef<OsStr>>(
     mut cmd: Command, arg0: &str, args: &[S]) -> Result<()>
 {
     #[cfg(unix)]
-    fn run(mut command: Command, arg0: &str) -> Result<()> {
+    fn run(mut command: Command) -> io::Error {
         use std::os::unix::process::CommandExt;
         let error = command.exec();
-        Err(error).chain_err(|| rustup_utils::ErrorKind::RunningCommand {
-            name: OsStr::new(arg0).to_owned(),
-        })
+        error
     }
 
     #[cfg(windows)]
-    fn run(mut command: Command, arg0: &str) -> Result<()> {
+    fn run(mut command: Command) -> io::Error {
         match command.status() {
             Ok(status) => {
                 // Ensure correct exit code is returned
                 let code = status.code().unwrap_or(1);
                 process::exit(code);
             }
-            Err(e) => {
-                Err(e).chain_err(|| rustup_utils::ErrorKind::RunningCommand {
-                    name: OsStr::new(arg0).to_owned(),
-                })
-            }
+            Err(e) => e
         }
     }
 
@@ -166,7 +160,9 @@ fn run_command_for_dir_without_telemetry<S: AsRef<OsStr>>(
     // when and why this is needed.
     cmd.stdin(process::Stdio::inherit());
 
-    run(cmd, arg0)
+    Err(run(cmd)).chain_err(|| rustup_utils::ErrorKind::RunningCommand {
+        name: OsStr::new(arg0).to_owned(),
+    })
 }
 
 #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


KalitaAlexey:
@birkenfeld,
Can you check if it fixes your problem?

KalitaAlexey:
@nrc,
Don't you know if @brson is on vocation?

KalitaAlexey:
@brson,
I don't know why it failed.

src/rustup/command.rs
```diff
@@ -135,24 +135,38 @@ fn telemetry_rustc<S: AsRef<OsStr>>(mut cmd: Command,
 fn run_command_for_dir_without_telemetry<S: AsRef<OsStr>>(
     mut cmd: Command, arg0: &str, args: &[S]) -> Result<()>
 {
-    cmd.args(&args);
+    #[cfg(unix)]
+    fn run(mut command: Command, arg0: &str) -> Result<()> {
+        use std::os::unix::process::CommandExt;
+        let error = command.exec();
+        Err(error).chain_err(|| rustup_utils::ErrorKind::RunningCommand {
```
nagisa: You’ll probably want to chain the error after calling run, like this:

```
run(cmd, arg0).chain_err(|| ...)
```

instead. That greatly cuts down on the amount of code needed.

nagisa: See for example similar code in [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz/blob/master/src/main.rs#L271)

KalitaAlexey: Sounds reasonable. Thanks.

brson:
Hi @KalitaAlexey.

Thanks for the fixes, and sorry for the delay.

This does solve the problem, but I want to solve these two issues a different way, and the main reason is that rustup wants to be able to perform work after the invoked subcommand is run (e.g. recording telemetry about the results). I am not prepared to give up that capability yet.

I'd instead prefer to do something more along the lines described in the op of https://github.com/rust-lang-nursery/rustup.rs/issues/806, installing a signal handler and capturing Ctrl+C. Likewise the signal-swallowing can be handled by examining how the subprocess terminated.

KalitaAlexey:
@brson,
If you know how to do that, you should do it because I don't know.
I need to think about it.

nagisa:
Theres another good reason to exec that is proper status reporting and
debugability. Namely currently the exit code is always 1 regardless of
rustc editting with code 101 or sogsegving. You cannot use gbd through the
wrapper that shadows the name of the rustc either which makes the use case
a pain. Exec solves all of these.

birkenfeld:
I agree with @nagisa - especially in that rustup-rustc/rustup-cargo should behave the same as normal rustc/cargo. This is otherwise very confusing to users who shouldn't have to care about how their toolchain was installed.

Diggsey:
I'll close this, since we're not planning on using exec - further discussion about whether this is the right decision can happen on the issue (#31)

PR #996, opened 2017-03-17
Bump

brson:
I don't intend to release this until investigating the [windows flakiness](https://github.com/rust-lang-nursery/rustup.rs/issues/995).

cc @Diggsey

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,47 @@
+# 1.1.0
+
+* [Fix browser detection for Linux ppc64 and NetBSD](https://github.com/rust-lang-nursery/rustup.rs/pull/875)
+* [Update windows info](https://github.com/rust-lang-nursery/rustup.rs/pull/879)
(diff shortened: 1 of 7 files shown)


malbarbo:
Please, consider waiting for https://github.com/rust-lang-nursery/rustup.rs/pull/1000.

Keruspe:
1.0.0 doesn't build with rust 1.16 stable (https://github.com/rust-lang-nursery/rustup.rs/pull/968) and fails with libressl too (https://github.com/rust-lang-nursery/rustup.rs/pull/1011).
I don't know of it weighs against the windows flakiness but not building against current stable is sad, imho.

brson:
> 1.0.0 doesn't build with rust 1.16 stable (#968) and fails with libressl too (#1011).
> I don't know of it weighs against the windows flakiness but not building against current stable is sad, imho.

I agree this is unfortunate. I don't see it as a reason to hold up a release though.

PR #1001, opened 2017-03-21
Remove curl usage from appveyor

Diggsey:
Appveyor removed curl from PATH, which broke our CI


Diggsey:
Issue: https://github.com/appveyor/ci/issues/1426

Diggsey:
Just going to merge this, test failure is clearly spurious and the other PRs need to include this change before they can run on appveyor.

PR #1002, opened 2017-03-21
[Don't merge] Try disabling shadow copy

Diggsey:


Diggsey:
@brson Either it's a fluke or this actually works. It's just coming up on the fifth rerun, which will be zero failures across 20 individual builds.

...and it failed the 19th build

brson:
@Diggsey that seems like a pretty good record though. Should we merge it and hope it's an improvement?

Diggsey:
No, it was just (bad) luck I think. I ran it again and it failed the very next build too. I wonder if there is another factor, such as time of day which affects how likely a failure is to occur.

PR #1005, opened 2017-03-23
RLS support

brson:
I think this is sufficient for basic support.

With this PR, rustup will create 'rls' proxies in CARGO_HOME/bin, and will do so lazily when that binary doesn't exist.

After this PR one can write

```
rustup component add rls
rustup component add rust-src
rustup component add rust-analysis
```

to get a working RLS, assuming the 'rls' package is deployed.

Next steps are to make `rustup component` accept multiple components, so RLS can be installed with just `rustup compnent add rls rust-src rust-analysis` (cc @durka).

I'd suggest that RLS have nice error handling for cases where rust-src and rust-analysis aren't available. It might suggest running the proper rustup commands.

rustup probably needs to emit a better error message when somebody tries to run `rls` without installing it. Since the proxy always exists, it will say something like "the binary 'rls' doesn't exist in toolchain X". It would be better for rustup to know which package contains 'rls' and say so.

cc @nrc @jonathandturner

### DIFF
--- a/src/rustup-cli/main.rs
+++ b/src/rustup-cli/main.rs
@@ -126,6 +126,7 @@ fn do_compatibility_hacks() {
     make_environment_compatible();
     fix_windows_reg_key();
     delete_multirust_bin();
+    add_rls_proxy();
 }
 
 // Convert any MULTIRUST_ env vars to RUSTUP_ and warn about them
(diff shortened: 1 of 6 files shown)


brson:
Hm, this has a bug in it from the changes to the test suite.

Diggsey:
Is `add_rls_proxy` strictly necessary? Won't the new proxy be automatically created anyway when rustup updates? (it always calls `install_bins`)

brson:
> Is add_rls_proxy strictly necessary? Won't the new proxy be automatically created anyway when rustup updates (it always calls install_bins)

The self-update will not on its own create the new proxy. `install_bins` is run on upgrade but it is the _old_ install bins that does not know about rls.

Diggsey:
Isn't it the new `install_bins` that gets run? The previous rustup calls the new one with `--self-replace`, and that one updates the hardlinks.

brson:
Uh, yeah I think you are right about that. Wow. I should sleep on this.

alexcrichton (approved):

brson:
OK, I pushed some hasty changes to try to fix the failing tests, and remove the lazy upgrade.

Attempting to make the test suite run bins off the PATH ended up causing all kinds of problems with the environment seeping in, so I just rolled that back. rustup is really frustrating to test...

brson:
@Diggsey r?

I've really lost perspective on this patch, just throwing crap at it and hoping it works out.

brson:
@bors retry

weird error:

```
---- multirust_upgrade_works_with_proxy stdout ----
	thread 'multirust_upgrade_works_with_proxy' panicked at 'assertion failed: fs::symlink_metadata(&multirust_dir).unwrap().file_type().is_symlink()', tests\cli-rustup.rs:739
```

PR #1008, opened 2017-03-23
Use correct name for undefined linked toolchain invocation

### LINKED ISSUE #927: Incorrect error for undefined linked toolchain invocation
Here rust2 is not defined:

```
brian@ip-10-145-43-250:/mnt2/dev/s-structured-log-rs⟫ cargo +rust2 test
error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installed
```

I don't know why that error is emitted.

topecongiro:
Closes #927.

### DIFF
--- a/src/rustup-cli/job.rs
+++ b/src/rustup-cli/job.rs
@@ -156,7 +156,6 @@ mod imp {
             let mut killed = false;
             let list = &jobs.list[..jobs.header.NumberOfProcessIdsInList as usize];
             assert!(list.len() > 0);
-            info!("found {} remaining processes", list.len() - 1);
 
             let list = list.iter().filter(|&&id| {
                 // let's not kill ourselves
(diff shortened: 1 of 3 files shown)


brson:
Thanks @topecongiro!

Can you add a test to `cli-exact.rs` that captures the output of of `cargo +bogus test` after this? You should be able to crib off the existing cases in that file.

topecongiro:
@brson I will try and rebase. Thank you!

brson:
Interesting failure here @topecongiro on windows. It looks to me like it's failing because on windows rustup is emitting a "info: found 0 remaining processes" .... which is bizarre. I think you should just delete that `info!` line from `job.rs` to make all platforms agree.

topecongiro:
@brson Thank you for pointing out! Rebased.

PR #1013, opened 2017-03-26
zsh install support

polonez:
- add path to .zprofile
- add a test for zsh


Diggsey:
Thanks for the PR, we should definitely start improving the experience for more different shells.

However, I think we should always add to `.profile` regardless, and also add to shell-specific files if we think that shell will be used. For zsh, a good heuristic might be whether `~/.zprofile` already exists, *or* whether `SHELL` is set to zsh as you are checking here.

brson:
I like @Diggsey's suggestions here. In general, I think the surest way we're going to be able to reliably set PATH is by pessimistically configuring in multiple places on Unix.

polonez:
I'll update soon, thanks.

PR #1015, opened 2017-03-28
Aggressive remove_dir_all

brson:
Let's see if this helps the flakiness.

cc @Diggsey @alexcrichton 

cc @Aaronepower this PR contains a self-contained remove_dir_all module based on winapi.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -606,6 +606,7 @@ dependencies = [
  "download 0.3.0",
  "error-chain 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


Diggsey:
Looks like we might need an aggressive `remove_file` too...

alexcrichton:
I personally forget, but can you clarify how precisely this is more aggressive again?

Diggsey:
It moves the files out of the subdirectory and up next to the directory being deleted, prior to deleting them. This means if the file deletion is delayed due to outstanding locks, then the directories can still be deleted successfully without waiting for other processes to close their file handles.

The failure here looks different - like a race condition in one of the self-updater tests possibly?

brson:
Updated to fix the linux build hopefully.

brson:
Test failure was network error and i want to deploy this in next release.

PR #986, opened 2017-03-14
add/remove multiple toolchains

durka:
Teaches the `toolchain` subcommand to accept multiple arguments for toolchains to add or remove.

Depends on #968.
Fixes #976.

Issue #1009, opened 2017-03-23
`target add` and `component add` should succeed if target/component is already installed


brson:
For automation purposes it's nice for this command to 'just work' when these commands are done redundantly. Change these commands to issue an 'info' indicating the thing is already installed and return successfully.

typesanitizer:
For cases when the target is distinct from the toolchain (say target=`x86_64-pc-windows-gnu` and toolchain=`stable-x86_64-unknown-linux-gnu`), there is an `info` issued along the lines of:

```
 info: component 'rust-std' for target 'x86_64-pc-windows-gnu' is up to date
```
Call order for issuing this info: [`bare_add_component`](https://github.com/rust-lang-nursery/rustup.rs/blob/a1287b6e897e495fb8371a2dfebf2e5f41d4365b/src/rustup/toolchain.rs#L547) -> ??? -> [`update`](https://github.com/rust-lang-nursery/rustup.rs/blob/a1287b6e897e495fb8371a2dfebf2e5f41d4365b/src/rustup-dist/src/manifestation.rs#L452) -> ??? -> `fmt` (in `notifications.rs`) which calls  [`ComponentAlreadyInstalled`](https://github.com/rust-lang-nursery/rustup.rs/blob/a1287b6e897e495fb8371a2dfebf2e5f41d4365b/src/rustup-dist/src/notifications.rs#L75).

The error only happens when the target is a part of the toolchain (say target=`x86_64-unknown-linux-gnu` and toolchain=`stable-x86_64-unknown-linux-gnu`). I think this error is coming from [here](https://github.com/rust-lang-nursery/rustup.rs/blob/a1287b6e897e495fb8371a2dfebf2e5f41d4365b/src/rustup/toolchain.rs#L567).

1. Should this error be matched and an `info` be issued in the corresponding functions `target_add` and `component_add` in [`rustup_mode.rs`](https://github.com/rust-lang-nursery/rustup.rs/blob/f9077d2fae3882e4fb45d8a35ee241911a3b4fee/src/rustup-cli/rustup_mode.rs)?
2. Should the `info` be issued using the same notification mechanism or do you just use a simple `info!` macro in the definitions for `target_add`/`component_add`?

nre-ableton:
I'm also running into this "error" and find it very confusing. I am not sure why `rustup` would fail in this case when everything is in an "ok" state. Are there any plans to fix this?

kinnison:
[...]

PR #1016, opened 2017-03-29
Add/remove multiple components+targets

durka:
Same as #986 but for components and targets. After this I don't see any more places to add `.multiple(true)`.

cc #1005


durka:
Not sure what's going on with `multi_host_smoke_test` but I can reproduce locally, will figure it out.

durka:
Another note is that all these loops exit on the first error. If you `rustup component add x y` and `x` doesn't exist (or is already installed, see #1009), it won't get to `y`. I'm not sure whether this is desired behavior.

brson:
> Another note is that all these loops exit on the first error. If you rustup component add x y and x doesn't exist (or is already installed, see #1009), it won't get to y. I'm not sure whether this is desired behavior.

Seems ok for now when there are no dependencies between components that need to be maintained.

durka:
Should be fixed now. Not really sure what I'm doing in the mock manifest code.

Issue #954, opened 2017-02-13
Document the use of toolchain link


jonhoo:
The current output of `toolchain --help` mentions that "`rustup` can also install toolchains from [..] local builds", but then does not go into further detail. `toolchain link` is the command meant for that purpose, but its `--help` output is, well, unhelpful:
```console
$ rustup toolchain link --help
rustup-toolchain-link
Create a custom toolchain by symlinking to a directory

USAGE:
    rustup toolchain link <toolchain> <path>

FLAGS:
    -h, --help    Prints help information

ARGS:
    <toolchain>
    <path>
```

In particular, it should at the very least be noted that `path` should be `/path/to/rust/build/$triple/stage1`, and `link` should be pointed to from `toolchain --help`. It'd also be useful to mention that only a `rustc` build is required, not a full `cargo` build.

brson:
Thank you for the bug report @jonhoo !

typesanitizer:
PR submitted here: https://github.com/rust-lang-nursery/rustup.rs/pull/1017

durka:
Would be great to also add some error checking, so if you try to `rustup toolchain link stage1/bin` or `stage1/bin/rustc` instead of totally mysterious "infinite recursion detected" later.

typesanitizer:
1. @durka, what are all the things that need to be checked before linking? In my build, I have a lot of `.so` files inside `stage1/lib` and `.rlib` and `.so` files in `stage1/lib/x86_64-unknown-linux-gnu/lib`.
2. I'm not sure which function should handle the error checking though: `toolchain_link` calls `install_from_dir` which calls `install` which calls `run` which basically calls `utils::symlink_dir`. Which function's responsibility would it be to handle this error? I suppose we could do it in `toolchain_link` itself by just calling `rustup_utils::utils::assert_is_file` for `$path/bin/rustc`.

durka:
At least a sanity check that bin, lib, and bin/rustc are in the expected
place would be nice.

brson:
Looks like the original issue is fixed. Thanks @theindigamer. Do we need to open another issue for the followups mentioned here cc @durka @theindigamer ?

typesanitizer:
@brson the merged PR included error checking for directories as suggested by @durka . Is there anything else that requires a follow up?

brson:
Nope! Thanks @theindigamer.

PR #1017, opened 2017-03-29
Added help for `rustup toolchain link` as documented in issue #954.

typesanitizer:
* Also renamed TOOLCHAIN_INSTALL_HELP -> INSTALL_HELP for consistency.


src/rustup-cli/help.rs
```diff
@@ -78,7 +78,23 @@ inferred, so the above could be written:
...
-custom toolchains with the `rustup toolchain link` command.";
+custom toolchains with the `rustup toolchain link` command. This can
+also be used to symlink toolchains from local builds. For more
+information see `rustup toolchain help link`.";
+
+pub static TOOLCHAIN_LINK_HELP: &'static str =
+r"
+Symlinks a toolchain from a local directory specified by 'path' and
+gives it a custom name specified by 'toolchain' if it does not name a
+standard release channel.
```
jonhoo: I'm not sure I understand what this means? The text just below seems to imply that `nightly` is a perfectly fine name to use, but the text here seems to suggest that I *can't* use a "standard release channel" name?

typesanitizer: (Okay, I don't fully understand _all_ the code but here is what I think is correct.)
What it means is that if you give it a name such as a standard release channel, say `nightly` then it will parse and understand what you are trying to say (https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup/config.rs#L388). 
You can certainly use `nightly` but that would mean that your actual `nightly` compiler (if you have one) will be replaced by the one at 'path'.
If you give it some unknown name which it doesn't recognize, say `nightly1`, then it will think that you are trying to give a custom name to the toolchain and will oblige.

typesanitizer: The call order is `toolchain_link` (in `rustup-cli/rustup-mode.rs`) -> `get_toolchain` (in `rustup/config.rs`) -> `Toolchain::from` (in `rustup/Toolchain.rs`) -> `resolve_toolchain` (in `rustup/config.rs`, linked earler) -> `PartialToolchainDesc::from_str` (in `rustup-dist/src/dist.rs`).

jonhoo: Ah, I see. Then I think the text should be clearer about this. How about:

> Symlinks a toolchain from a local directory specified by 'path' and gives it a custom name specified by 'toolchain'. If `toolchain` matches the name of a standard release channel such as `stable`, `nightly`, `1.8.0`, etc., then `path` will be used in place of the standard binaries for those toolchains. For example, if a toolchain symlink is set up called `nightly`, any directory with a `nightly` override will now also use the new symlink.

Also, I notice the text currently says
> For more information see `rustup help toolchain`
Does `rustup help toolchain` *actually* give more information at the moment?

typesanitizer: > How about: ...

Good point. Your explanation is much clearer than mine.

> Does `rustup help toolchain` actually give more information at the moment?

`rustup help toolchain` has a full explanation of the toolchain name being split as `<channel>[-<date>][-<host>]` alongwith a couple of examples and the meanings of the individual terms. This is also referred to by `rustup help install`so I thought I should follow suit.

jonhoo: In fact, I wonder if we should also repeat some, if not all, of `TOOLCHAIN_LINK_HELP` in `rustup help toolchain` as well (if that's not already the case).

typesanitizer: > if that's not already the case

Currently, `TOOLCHAIN_HELP` (used for `rustup help toolchain`) only tells what each subcommand does in a succinct manner and each subcommand is responsible for its own documentation (or lack thereof). The current documentation for `link` inside `TOOLCHAIN_HELP` is just:

> SUBCOMMANDS:
> ...
>     link      Create a custom toolchain by symlinking to a directory
> ...
> Toolchain names that don't name a channel instead can be used to name
> custom toolchains with the `rustup toolchain link` command. This can
> also be used to symlink toolchains from local builds. For more
> information see `rustup toolchain help link`.

I feel that this is sufficient information to indicate to the user whether `toolchain link` is something they want to use, although the wording could be improved to make the dual purpose (overwriting or aliasing) clearer. Just making the utility clearer doesn't require the full details; e.g. how the path is to be specified.

src/rustup-cli/help.rs
```diff
@@ -78,7 +78,23 @@ inferred, so the above could be written:
     $ rustup default stable-msvc
 
 Toolchain names that don't name a channel instead can be used to name
-custom toolchains with the `rustup toolchain link` command.";
+custom toolchains with the `rustup toolchain link` command. This can
+also be used to symlink toolchains from local builds. For more
+information see `rustup toolchain help link`.";
```
brson: These two sentences seem slightly redundant. What about something like "rustup can also manage symlinked local toolchain builds, which is often used for developing Rust itself. For more information see `rustup toolchain help link`.

src/rustup-cli/help.rs
```diff
@@ -78,7 +78,23 @@ inferred, so the above could be written:
...
+Symlinks a toolchain from a local directory specified by 'path' and
+gives it a custom name specified by 'toolchain' if it does not name a
+standard release channel.
+
+'toolchain' specifies a toolchain name, such as 'stable', 'nightly',
+or '1.8.0'. For more information see `rustup help toolchain`.
+
+'path' is specified as `/path/to/rust/build/$triple/stage1` where
+'$triple' should be replaced with the desired triple. Only a build of
+`rustc` is required to be present at the given 'path'.
```
brson: `toolchain link` can work with toolchains from any source, and this advice is specifically about linking the toolchain directly from the build directory. Can you clarify? Maybe like, "When used for development of Rust itself, toolchains can be linked directly out of the build directory, specifying `path` as ...".

brson:
Thanks so much @theindigamer. I left a few comments.

Wrt the semantics around whether its ok to reuse names of release channels, I don't recall what rustup does and am happy to merge whatever you and @jonhoo agree on.

typesanitizer:
@brson , @jonhoo 
Thanks for all the feedback. 
I tested the semantics and it turns out that you _cannot_ name a custom toolchain with a standard name such as `nightly`; `rustup` will throw an error (my bad, I didn't do this earlier).
```
$ rustup toolchain link stable build/x86_64-unknown-linux-gnu/stage2
error: invalid custom toolchain name: 'stable-x86_64-unknown-linux-gnu'
```
That said, here are all the changes I propose:

* At the end of `toolchain help`
> rustup can also manage symlinked local toolchain builds, which are
> often used to for developing Rust itself. For more information see
> `rustup toolchain help link`.";

* In `toolchain link help`
> Symlinks a toolchain from a local directory specified by 'path' and
> gives it a custom name specified by 'toolchain'.
>
> 'toolchain' is the custom name to be assigned to your toolchain. Any
> name is permitted as long as it does not fully match an initial 
> substring of a standard release channel. For example, you can use 
> the names 'latest' or '2017-04-01' but you cannot use 'stable' or 
> 'beta-i686' or 'nightly-x86_64-unknown-linux-gnu'.
> 
> 'path' specifies the directory where the binaries and libraries for
> the custom toolchain can be found. For example, when used for 
> development of Rust itself, toolchains can be linked directly out of
> the build directory. After building, you can test out different 
> compiler versions as follows:
> 
>     $ rustup toolchain list
>     stable-x86_64-unknown-linux-gnu (default)
>     $ rustup toolchain link latest-stage1 build/x86_64-unknown-linux-gnu/stage1
>     $ rustup toolchain list
>     stable-x86_64-unknown-linux-gnu (default)
>     latest-stage1
>     $ rustup toolchain default latest-stage1
>
> If you now compile a crate with default settings, the custom 
> toolchain 'latest-stage1' will be used.

(I've used four spaces for indenting the code; Github put that in a code block for display.)

If you folks think the explanation is satisfactory, I'll add a commit with the changes 
and update the PR.

jonhoo:
@theindigamer good catch! The updated text looks pretty good, though I'd suggest two changes:

 - "Symlinks a toolchain [..] and gives it a custom name specified by 'toolchain'.", and "'toolchain' is the custom name to be assigned to your toolchain" coming right after each other seems a bit redundant. You can probably remove one of them.
 - I would remove the `rustup toolchain list` invocations from the example -- at least I personally don't feel as though they add much.

typesanitizer:
Alright. Updated text for `toolchain link help` (I removed the top bit as there is already a message "Create a custom toolchain by symlinking to a directory" present):

> 'toolchain' is the custom name to be assigned to the new toolchain. 
> Any name is permitted as long as it does not fully match an initial
> substring of a standard release channel. For example, you can use
> the names 'latest' or '2017-04-01' but you cannot use 'stable' or
> 'beta-i686' or 'nightly-x86_64-unknown-linux-gnu'.
>
> 'path' specifies the directory where the binaries and libraries for
> the custom toolchain can be found. For example, when used for
> development of Rust itself, toolchains can be linked directly out of
> the build directory. After building, you can test out different
> compiler versions as follows:
>
>     $ rustup toolchain link latest-stage1 build/x86_64-unknown-linux-gnu/stage1
>     $ rustup override set latest-stage1
>
> If you now compile a crate in the current directory, the custom 
> toolchain 'latest-stage1' will be used.

EDITS:
* "your toolchain" -> "the new toolchain"
* "default" -> "override set"
* ~"the default settings"~
* fixed formatting

jonhoo:
Any particular reason you want to set the new one as `default`? Could we instead use `override set`? Also, can we `s/your toolchain/the new toolchain/`.

typesanitizer:
@jonhoo , good point. I didn't think about the `default` point. I've updated the comment. 
(I didn't want to keep pasting it over and over again and increase the visual clutter.)

jonhoo:
I think that means you can also remove " with default settings". With that change I'm happy :)

typesanitizer:
I have the bit about default settings because someone might have declared an explicit dependency on a `rustc` version like [here](https://github.com/joshtriplett/rfcs/blob/tool-dependencies/text/0000-tool-dependencies.md) (which is not a default setting) and the toolchain change might potentially conflict with that, in which case compiling wouldn't work. I'm not sure how `cargo` handles that kind of error.

EDIT: I recognize that it is just a PR which has not been merged but I thought there might be some other way (which I cannot think of) where a user has some non-default settings so that they end up some specific compiler, which will prevent the override from working.

EDIT 2: I also concede that the wording "with default settings" is kind of bad and compresses too much meaning into few words. Well, if a user has a weird build system, then they probably know if and when `override set` does not work. Done, removed.

jonhoo:
@theindigamer the latest draft of the text looks good. I see you also pushed some checks for the existence of `rustc` inside there. Is that truly the only valid path structure? Is there a particular reason for why `toolchain link` doesn't just take a `rustc`, but instead the `stage1` directory?

typesanitizer:
@jonhoo 
> Is that truly the only valid path structure?

This is a minimal check. There may be additional folders such as `etc` and `share` (your local installations of stable/nightly rust probably have these). But `bin` and `lib` are definitely required.

> Is there a particular reason for why `toolchain link` doesn't just take a `rustc`, but instead the `stage1` directory?

The `.so` files in `lib` are required for compilation; it will fail if you delete all of them.

jonhoo:
Okay, you've got my r+, even though you don't need it :p
@brson r?

src/rustup-cli/rustup_mode.rs
```diff
@@ -687,7 +688,15 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
...
-
+    let mut pathbuf = PathBuf::from(path);
+
+    pathbuf.push("lib");
+    try!(utils::assert_is_directory(&pathbuf));
+    pathbuf.pop();
+    pathbuf.push("bin");
+    try!(utils::assert_is_directory(&pathbuf));
+    pathbuf.push("rustc");
+    try!(utils::assert_is_file(&pathbuf));
```
brson: I'm ok with adding this validation, but will you sink it into the `toolchain.install_from_dir` call below? This file is purely for argument parsing.

brson:
This looks great to me. I just added one nit about where to do the new validation. Thanks @theindigamer and @jonhoo.

src/rustup-cli/rustup_mode.rs
typesanitizer: @brson Done. I've added it to the top of the function as `install_from_dir` may be used in the future with `link=false` but it would still require checking the directory structure correctly in that case.

typesanitizer: Whoops, sorry I had to change one line (remove the added dependency on `PathBuf`) so I rebased locally. (That's why the first commit link below is dead.)

brson:
Looks like there's a merge conflict now @jonhoo. Can you rebase?

jonhoo:
@theindigamer owns this, so he has to be the one to do it :)

brson:
Lol sorry @jonhoo .

typesanitizer:
@brson , @jonhoo I'm not fully sure what I'm supposed to do here... do I (locally) pull the latest version of `master`, then rebase it on top of the current branch (`fix-issue-954`) and then push `fix-issue-954` to my Github repo?

Diggsey:
@theindigamer github has instructions for resolving merge conflicts - at the bottom of this PR, there's a button to resolve conflicts online using the web interface, or a link to instructions for doing it via the command line.

typesanitizer:
@Diggsey , thanks. I didn't realize I could edit the text file on the merge conflicts page.

typesanitizer:
If I understand it correctly, the tests are failing because of two reasons:

1. Some tests like `custom_invalid_names` and `custom_invalid_names_with_archive_dates`  expect an error based on `invalid custom name`. Since I added the path checking to the very beginning, the error thrown is that directories are absent, which is unexpected for the test. This can be remedied by moving the directory error-checking to the end of `install_from_dir` (instead of the beginning).
2. Some tests like `rustup_failed_path_search`, `custom_toolchain_cargo_fallback_proxy` and `custom_toolchain_cargo_fallback_run` expect `toolchain link` to succeed silently even if the directory structure is wrong and expect other things to break later. I'm not sure what would be the best way to fix this. Should I just create fake directories and files for `bin`, `lib` and `rustc`?

Should I add a test which ought to fail with the wrong directory structure?

typesanitizer:
@brson , good to merge?

brson:
@bors r+ p=1

Yes @theindigamer looks great. Sorry for the long delay.

PR #1019, opened 2017-03-30
Indent help text

### LINKED ISSUE #940: Improve indentation of help
I find the indentation of command like `rustup override` very confusing

```
$ rustup override
rustup-override 
Modify directory toolchain overrides

USAGE:
    rustup override [SUBCOMMAND]

FLAGS:

SUBCOMMANDS:
    list     List directory toolchain overrides
    set      Set the override toolchain for a directory
    unset    Remove the override toolchain for a directory
    help     Prints this message or the help of the given subcommand(s)


Overrides configure rustup to use a specific toolchain when
running in a specific directory.

Directories can be assigned their own Rust toolchain with
`rustup override`. When a directory has an override then
any time `rustc` or `cargo` is run inside that directory,
or one of its child directories, the override toolchain
will be invoked.

To pin to a specific nightly:

    rustup override set nightly-2014-12-18

Or a specific stable release:

    rustup override set 1.0.0
$ _
```

At first glance it looks roughly like this
```
$ rustup override
      ... something indented
a lot of unindented text. a lot of text. a lot of text.
a lot of unindented text. a lot of text. a lot of text.
      ... something indented again
$ _
```

I find that extremely unreadable due to the mix of indented and unindented chunks.
I suggest to adopt the structure of a manpage, i.e. everything is indented except for headlines, which are uppercase.

The example above could look like this

```
$ rustup override
NAME:
    rustup-override

DESCRIPTION:
    Modify directory toolchain overrides

USAGE:
    rustup override [SUBCOMMAND]

FLAGS:

SUBCOMMANDS:
    list     List directory toolchain overrides
    set      Set the override toolchain for a directory
    unset    Remove the override toolchain for a directory
    help     Prints this message or the help of the given subcommand(s)

DISCUSSION:
    Overrides configure rustup to use a specific toolchain when
    running in a specific directory.

    Directories can be assigned their own Rust toolchain with
    `rustup override`. When a directory has an override then
    any time `rustc` or `cargo` is run inside that directory,
    or one of its child directories, the override toolchain
    will be invoked.

    To pin to a specific nightly:

        rustup override set nightly-2014-12-18

    Or a specific stable release:

        rustup override set 1.0.0
$ _
```

=> Much better IMO. Now I can immediately see what belongs where. :-)
What do you think about this suggestion?

babbageclunk:
Add a DISCUSSION heading to each item and indent (and rewrap) the help text.

Fixes #940.

### DIFF
--- a/src/rustup-cli/help.rs
+++ b/src/rustup-cli/help.rs
@@ -1,223 +1,236 @@
 pub static RUSTUP_HELP: &'static str =
-r"
-rustup installs The Rust Programming Language from the official
-release channels, enabling you to easily switch between stable, beta,
-and nightly compilers and keep them updated. It makes cross-compiling
-simpler with binary builds of the standard library for common platforms.
+r"DISCUSSION:
+    rustup installs The Rust Programming Language from the official
+    release channels, enabling you to easily switch between stable,
+    beta, and nightly compilers and keep them updated. It makes
+    cross-compiling simpler with binary builds of the standard library
+    for common platforms.
 
-If you are new to Rust consider running `rustup doc --book`
-to learn Rust.";
+    If you are new to Rust consider running `rustup doc --book` to
+    learn Rust.";
 
 pub static SHOW_HELP: &'static str =
-r"
-Shows the name of the active toolchain and the version of `rustc`.
+r"DISCUSSION:
+    Shows the name of the active toolchain and the version of `rustc`.
 
-If the active toolchain has installed support for additional
-compilation targets, then they are listed as well.
+    If the active toolchain has installed support for additional
+    compilation targets, then they are listed as well.
 
-If there are multiple toolchains installed then all installed
-toolchains are listed as well.";
+    If there are multiple toolchains installed then all installed
+    toolchains are listed as well.";
 
 pub static UPDATE_HELP: &'static str =
-r"
-With no toolchain specified, the `update` command updates each of the
-installed toolchains from the official release channels, then updates
-rustup itself.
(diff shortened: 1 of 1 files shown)


brson:
Thanks @babbageclunk!

@kbknapp I wonder if you can peek at this and suggest whether it's the right way. My understanding of this issue is that we are trying to imitate claps generated formatting manually. Can clap itself produce formatting like this for these freeform text bits.

kbknapp:
@brson Unfortunately this is the best way at this time. 

The `after_help` portion is designed to be free-form and thus all headers and formatting are left to users. The only thing that `after_help` *does do* is rewrap properly at word boundaries. But since rustup is already using hard newline wrapping this isn't really an issue. 

With the current PR, if someone has a terminal window with < 70 (I think?) columns the rewrapping that clap will do won't respect the hard tabs (or 4x spaces). However, someone with a terminal window that small is probably a rare-ish occurrence.

I'm open to someone adding the ability for clap to respect the starting indentation when using [`template`](https://docs.rs/clap/2.22.2/clap/struct.App.html#method.template)s, and would be willing to mentor such efforts. I just don't have time to implement it myself at the moment. This would be the best solution as then all rewrapping and indentation is handled by clap.

Diggsey:
Let's do this for now, we can always improve it later if we find a better way.

@bors r+

babbageclunk:
Oh, nice! I guess that was a flaky test?

Diggsey:
Just a network error, yeah.

PR #1020, opened 2017-03-31
Support partial downloads

### LINKED ISSUE #889: Support resume of partial downloads

jelford:
This PR should close  #889

A couple of implementation details:

Only added support for the curl backend; previously discussed that there's an intention to get rid of rustup's own download code, and the default feature-set uses curl anyway, so hopefully this is okay.

Added new testing to the download crate - while it's there, it makes sense to have a test. Since using curl's "resume" functionality, I figured it's probably fine to just file:// urls for test cases. Previously tested using a small hyper-based http server, but that feels like overkill.

For hashing files, I've set the buffer size to 2^15 - just because that's what strace tells me is used by `sha256sum` on my local PC. It seems much slower than that command though, and it's not obvious why, so maybe I've done something silly here.

Finally, and maybe most controversially, I haven't done anything about cleaning up aborted partials. I don't really know when a good time is to do this, but a couple of suggestions that I'd be happy to implement:
* Every run, just check the download cache for any files > 7 days old and smoke them
* On self-update, as that seems like a natural time for generic "maintenance" sorts of operations

I mentioned in my last PR, but the same disclaimer: I haven't written much rust, so I fully expect you will see some problems (also very happy to accept style criticisms). I accidentally ran a `rustfmt` on some things so apologies for the noise (can revert but... maybe it's worth having anyway?).


jelford:
Not sure about that travis error: "couldn't find crate for std." I don't know whether there's a way to just give it another kick?

Diggsey:
I restarted that job. The appveyor issues are legit though.

jelford:
Yes, that could definitely never work. Don't know why I put the conditional there. I can't check my work though, don't have a windows box, but think I've just pushed a fix.

Diggsey:
Thanks @jelford 
Could you revert the formatting changes from this PR - the rust style guidelines are currently in the process of being changed (away from visual indentation by the look of things), so I don't think we should reformat anything at this time, and even if we did, it should be separate from functional changes to the code.

While it would be nice to move the download code out of rustup, I'm not sure that the intent is to switch to curl though? I'm not necessarily opposed to adding this kind of feature just for the curl backend though if it will help people on poor connections - @brson thoughts?

jelford:
Thanks for the quick response @Diggsey.
I've reverted those formatting changes, sorry for the noise. 

On switching to curl, I just meant to observe that this is the default, so most people should get the benefit right away. I'm hoping 90% of this goes away as part of https://github.com/rust-lang-nursery/rustup.rs/issues/993, so I wasn't keen to spend too long getting it into the hyper-backend. 

As a side note, I only took a quick look and I don't *think* reqwest currently has support for resume as a 1st-class thing. I was thinking it's easy enough just to set the appropriate header, but need to think about the error case where the server doesn't support it. libcurl will just bomb out when it gets a response without the Content-Range header. Is that actually fine? Do all the mirrors support it? It works for me locally, but I don't know whether it's S3 everywhere or something else in different parts of the world. If not, it would be a shame to break the whole thing for what is essentially a bandwidth optimization.

Options in that case would be:
* add an environment variable to de-activate resume requests
* fail gracefully from the specific curl error and fall back to downloading the whole file.

The former is simple, but not as nice a UI.

brson:
> While it would be nice to move the download code out of rustup, I'm not sure that the intent is to switch to curl though? I'm not necessarily opposed to adding this kind of feature just for the curl backend though if it will help people on poor connections - @brson thoughts?

My intent is to switch to reqwest, assuming that reqwest uses native-tls, and someday gets support for rustls. The rustup download crate was supposed to basically be reqwest, but at this point it seems best to do the hyper/rustls transition in the ecosystem, not in rustup. Adding resume support just means we'll have to make reqwest support that too (or otherwise create a crate that does what reqwest does with resume support).

brson:
Good tests, thanks @jelford.

brson:
> Is that actually fine? Do all the mirrors support it? It works for me locally, but I don't know whether it's S3 everywhere or something else in different parts of the world. If not, it would be a shame to break the whole thing for what is essentially a bandwidth optimization.

I think we can not assume much about the servers, because people will at some point host their own rustup mirrors.

brson:
Patch looks good to me, r=me if @Diggsey is happy.

brson:
Windows flakiness

PR #1022, opened 2017-04-01
Update appveyor image

Diggsey:
Should reduce the flakiness of appveyor builds, as the new image disables windows defender.


Diggsey:
This increases the total build time by ~8 minutes because msys2 is not preinstalled on the new image, but that should be fixed once it is merged to master, assuming the appveyor file cache works as expected.

Issue #987, opened 2017-03-15
Installation failure via the script has bad error message


brson:
```
~/C/h/e/duration ❯❯❯ curl https://sh.rustup.rs -sSf | sh

info: downloading installer
warning: it looks like you have existing multirust metadata
warning: rustup cannot be installed alongside multirust
warning: delete `/Users/wycats/.multirust` before installing rustup
error: cannot install while multirust is installed
rustup: command failed: /var/folders/qn/cf7wq1s54dlfg51t4xjt371c0000gn/T/tmp.mz2CwTMt/rustup-init
```

This last line comes from rustup-init.sh and is useless. Probably the right thing to do here is just not print anything in response to an error running rustup-init.exe, though there is the risk of some failure that produces no useful error text, like e.g. rustup-init.exe isn't runnable for some reason.

Let's just take out the error reporting and see how it goes.

bitrauser:
Hello, I tried to take out the error reporting and made a PR. Is this what you intended?

brson:
Looks fixed.

PR #1023, opened 2017-04-01
Remove error reporting from rustup-init.sh

noisegul:
In response to #987 I made rustup-init.sh terminate unsuccessfully instead of printing a message in case there is an error running rustup-init.exe.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -118,9 +118,9 @@ main() {
             err "Unable to run interactively. Run with -y to accept defaults, --help for additional options"
         fi
 
-        run "$_file" "$@" < /dev/tty
+        ignore "$_file" "$@" < /dev/tty
     else
-        run "$_file" "$@"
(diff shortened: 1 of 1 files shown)


Diggsey:
Looks like `run()` is also used by `ignore()`, but I don't think the latter should exit if an error occurs.

noisegul:
I see, so it would make more sense in this case to remove the conditional block completely in order to take out error reporting without influencing the behaviour of `ignore()`?

Diggsey:
Yes, I think so.

noisegul:
Thank you for your feedback. I changed this.

lu-zero:
`run()` this way is basically a pass-through, it might make sense to not use it at all.

noisegul:
I agree, in this case removing it would also re-establish the intended documentation factor of `ignore()`, as stated in the comments.

PR #1024, opened 2017-04-01
Document installing to a custom location using CARGO_HOME and RUSTUP_HOME environment variables

### LINKED ISSUE #994: Document the usage of CARGO_HOME and RUSTUP_HOME to install to a custom location
It's a common complaint that rustup doesn't let you control where to install, but it actually does. It's just not documented. Setting CARGO_HOME and RUSTUP_HOME while running rustup-init will set those directories as desired. Note though that if one does this then they are on their own wrt ensuring their environment is set up right - those environment variables must be configured whenever the toolchain is executed, and CARGO_HOME/bin must be on the PATH.

sjmarshy:
This should Fix #994 

Add a section `Choosing where to install` to the `install` section of the README discussing the
use of `CARGO_HOME` and `RUSTUP_HOME`

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,22 @@ For `zsh`, you must then add the following line in your `~/.zshrc` before
 fpath+=~/.zfunc
 ```
 
+#### Choosing where to install
+
+`rustup` allows you to customise your installation by setting the environment
+variables `CARGO_HOME` and `RUSTUP_HOME` while running the `rustup-init`
+executable. As mentioned in the [Environment Variables] section, `RUSTUP_HOME`
+sets the root rustup folder, which is used for storing installed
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,6 +83,22 @@ For `zsh`, you must then add the following line in your `~/.zshrc` before
 fpath+=~/.zfunc
 ```
 
+#### Choosing where to install
+
+`rustup` allows you to customise your installation by setting the environment
+variables `CARGO_HOME` and `RUSTUP_HOME` while running the `rustup-init`
```
Diggsey: I think saying "while" might give the wrong impression, despite the clarification below. Maybe say "before running" instead?

Diggsey:
LGTM with that one change

Issue #1027, opened 2017-04-01
Investigate caching on travis builds


jelford:
Travis has some support for caching (https://docs.travis-ci.com/user/caching/). It's not that useful for conserving bandwidth, but I think rustup might be able to get some benefit from caching the compilation of OpenSSL static libs - which looks like it's adding quite a lot onto the build.

Since it's done inside a docker container, you'd have to use the "arbitrary folder" flavour of caching support. I think one would have to make a per-build-target folder, since while travis will give you a different cache per OS, I don't think it'll do it for each set of environment variables in the build matrix.

It coud work something like this:

    root/
        openssl/                                     # cache at this level
            i686-unknown-linux-gnu        # mount each of these to target/$TARGET/openssl/openssl-install
            x64-unknown-linux-gnu

For non-docker builds, travis has "first class" support for cargo.

jelford:
If there's appetite for something like this, I'm happy to get travis running against my own repo and try it out.

Diggsey:
@jelford Looks like travis does take into account env vars: https://github.com/travis-ci/travis-build/pull/706

Please go ahead, anything that reduces the build time would be nice!

jelford:
Oh that's great, should be able to get it going without hacking around with folders then. I'll give it a go.

jelford:
This has landed but I'll leave it open until I've had a chance to look at the effect on some builds and report back findings for others' interest.

jelford:
The numbers for this look ok to me; builds on [travis's history page](https://travis-ci.org/rust-lang-nursery/rustup.rs/builds) have gone from around the 5h30m mark to around the 4h mark. 

In terms of the cache misses I described on the PR (about half the cargo dependency builds are hitting the cache across all configurations), those don't seem to have magically gone away, so if someone with a little experience of getting travis caches to work happens to see this, I'd love to hear about what might be going on here.

PR #1028, opened 2017-04-02
Cache travis artefacts

jelford:
We can re-use the following across builds:
* compiled OpenSSL libs
* compiled crate dependencies

Use specific folder caching because we've got a custom installation of rust/cargo etc. so travis's `cargo` caching can't really be expected to get that right. Move the install location of OpenSSL to ensure that if the version is bumped, we won't get a bad cache hit and silently fail to apply the upgrade.

If the rust toolchain gets bumped, it would be worth checking what gains can be got by incremental compilation of the local crate.

In terms of the actual speed-up observed, it's difficult to be precise because travis build times seem to jump around a lot anyway, but I would say each build configuration is saving at least a few minutes. This is coming from:
* OpenSSL can get re-used basically every time; it will only need to be rebuilt when the version gets bumped (which might be what you expect for all the dependencies, but...)
* Rust dependencies seem to hit the cache... sometimes. I'm not sure quite what's going on here, but some dependencies get picked up and other don't. I set the build to verbose to see which ones, but if there's a pattern I'm not seeing it.
* I've tried to avoid caching things that just get downloaded (e.g. `CARGO_HOME`) since Travis's docs note downloading from a cache won't be much faster than downloading from whatever mirror they're coming from in the first place. Empirically I have found their caches faster that crates.io, but it inflates the size of the cache a lot and Iit seems bad to use it for a purpose they've asked people not to.

The upshot is that quite a lot of time is still spent doing the cargo build for dependencies but... less than it was, so I think that's a win. You can see an example of what I mean [here](https://travis-ci.org/jelford/rustup.rs/jobs/217883314). The crates that are getting re-built are a mix of types (`--crate-type lib / bin`), and a mix of `--out-dir`. I'm willing to give it a shrug and say maybe Travis applies some sort of heuristic when it decides what to actually bother storing, but would be interested in any pointers on what cargo's schema is for which deps it puts where (I couldn't see anything obvious on crates.io, beyond setting `CARGO_TARGET_DIR`).

The issue for this is #1027

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,7 @@ matrix:
       env: TARGET=s390x-unknown-linux-gnu
            DOCKER=alexcrichton/rust-slave-linux-cross:2016-11-11
            SKIP_TESTS=1
-    
+
     # Android use a local docker image
     - os: linux
       env: TARGET=arm-linux-androideabi
@@ -91,7 +91,7 @@ matrix:
       env: TARGET=i686-linux-android
            DOCKER=android
            SKIP_TESTS=1
-    
+
     # On OSX we want to target 10.7 so we ensure that the appropriate
     # environment variable is set to tell the linker what we want.
     - os: osx
@@ -107,6 +107,7 @@ cache:
   - directories:
     - target/$TARGET/openssl/openssl-install
     - target/$TARGET/release
+    - target/release/deps
 
 branches:
   only:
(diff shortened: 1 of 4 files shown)


brson:
Wow, thanks for taking the initiative! Looks great to me @jelford.

@bors r+

PR #1029, opened 2017-04-05
Recognize armv8l as aarch64

malbarbo:
This was report in https://www.reddit.com/r/rust/comments/631bl2/exciting_rust_coming_to_android_and_chromebooks/dfsu2da/

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -253,7 +253,7 @@ get_architecture() {
             fi
             ;;
 
-        aarch64)
+        aarch64 | armv8l)
             local _cputype=aarch64
             ;;
 


malbarbo:
It seems that there is some difference in host detection code of `rustup-init.sh` and `src/rustup-dist/src/dist.rs`. I do not understand exactly when host detection of `src/rustup-dist/src/dist.rs` is executed. Should the code in `src/rustup-dist/src/dist.rs` be updated?

cuviper:
See https://github.com/rust-lang/rust/pull/41152#discussion_r110505533 -- I'm pretty sure `armv8l` is still 32-bit.

malbarbo:
@cuviper Thank you for the clarification.

PR #1035, opened 2017-04-06
Add bash completion guidance for macOS users

bitjson:
For macOS users, one of the easiest and most popular ways to "get bash completion working" is via  [Homebrew](https://brew.sh/)'s `bash-completion` formula (which other formulas rely upon for completion support). 

E.g. many users will have installed completions for Git using [this guide](https://github.com/bobthecow/git-flow-completion/wiki/Install-Bash-git-completion#os-x--macos).

This PR adds documentation (to `rustup help completions`) for installing rustup completions for macOS users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,9 @@ gist is as simple as using one of the following:
 # Bash
 $ rustup completions bash > /etc/bash_completion.d/rustup.bash-completion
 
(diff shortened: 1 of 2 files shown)


brson:
Thanks @bitjson !

PR #1038, opened 2017-04-06
Check ZDOTDIR when adding path to .zprofile

### LINKED ISSUE #1036: ZDOTDIR not obeyed when installing to .zprofile
> If the shell is a login shell, commands are read from /etc/zprofile and then $ZDOTDIR/.zprofile.
> If ZDOTDIR is unset, HOME is used instead.
> — man zsh

Currently https://github.com/rust-lang-nursery/rustup.rs/blob/904ecf0a97b508f929233280644592d90cf9e2c5/src/rustup-cli/self_update.rs#L984 just installs to `$HOME/.zprofile`, this should check if `ZDOTDIR` is set and use that instead.

Nemo157:
Fixes #1036


brson:
Flaky windows.

PR #1040, opened 2017-04-08
Don't crash if modifying multiple profile files

brson:
Fixes https://github.com/rust-lang-nursery/rustup.rs/issues/1034

cc @rmbreak


smackysnacks:
LGTM 👍

PR #1046, opened 2017-04-09
Add better usage when running rustup doc

BrunoMacias:
Currently, running rustup doc prints nothing to the terminal. This
addition prints the available args for the doc command when the
additional args are missing.


Diggsey:
You don't currently need to pass any arguments to `rustup doc`. Won't making the argument group be required break that?

brson:
Yes @Diggsey I think so, a plain `rustup doc` today is supposed to open the documentation landing page.

I do see behavior on a headless linux machine where `rustup doc` appears to do nothing.

Diggsey:
#1125

PR #1054, opened 2017-04-12
Fix pkgconfig for openssl build

### LINKED ISSUE #1051: rustup update fails and erroneously mention no https support in curl
```
$ rustup update
info: syncing channel updates for 'stable-armv7-unknown-linux-gnueabihf'
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/home/pi/.rustup/tmp/kpgqmw7yym237tou_file'
info: syncing channel updates for 'nightly-armv7-unknown-linux-gnueabihf'
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-nightly.toml.sha256' to '/home/pi/.rustup/tmp/1g8p1l43ph77ahvz_file'
error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to '/tmp/rustup-update.c5REq4ZVU3fY/release-stable.toml'
info: caused by: [1] Unsupported protocol (Protocol "https" not supported or disabled in libcurl)
=> NOT OK

pi@kilabspi02:~ $ curl https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256
6a73d4aa8eccdfe03fdd7ddd84685c8ccba3ca6516db0475197ef40a21686a52  channel-rust-stable.toml
=> OK

pi@kilabspi02:~ $ curl https://static.rust-lang.org/dist/channel-rust-nightly.toml.sha256
2762bdca077d5a227a7950d69676956c2ed92f553d6bbe578138c607a4495f2a  channel-rust-nightly.toml
=> OK

pi@kilabspi02:~ $ curl https://static.rust-lang.org/rustup/release-stable.toml
schema-version = '1'
version = '1.2.0'
=> OK

pi@kilabspi02:~ $ curl --version
curl 7.38.0 (arm-unknown-linux-gnueabihf) libcurl/7.38.0 OpenSSL/1.0.1t zlib/1.2.8 libidn/1.29 libssh2/1.4.3 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smtp smtps telnet tftp 
Features: AsynchDNS IDN IPv6 Largefile GSS-API SPNEGO NTLM NTLM_WB SSL libz TLS-SRP 
=> curl *does* have https
```

malbarbo:
To avoid cache problems https://github.com/rust-lang-nursery/rustup.rs/pull/1028 uses a temporary install directory for openssl. When the openssl install is completed the install dir is renamed. In doing so the pkgconfig files become invalid and libcurl-sys fails to find openssl and is built without https support. This fix the pkgconfig files. Fixes https://github.com/rust-lang-nursery/rustup.rs/issues/1051.

### DIFF
--- a/ci/build-run-docker.sh
+++ b/ci/build-run-docker.sh
@@ -22,3 +22,10 @@ docker run \
   -e SKIP_TESTS=$SKIP_TESTS \
   -it $DOCKER \
   ci/run-docker.sh
+
+# check that rustup-init was built with ssl support
+# see https://github.com/rust-lang-nursery/rustup.rs/issues/1051
+if ! (nm target/$TARGET/release/rustup-init | grep Curl_ssl_version &> /dev/null); then
+  echo "Missing ssl support!!!!"
+  exit 1
+fi
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -177,6 +177,9 @@ else
    install=$final_install_path
 fi
 
+# the install dir was renamed, so we need to update pkgconfig
+sed -i -e 's/-partial//g' $install/lib/pkgconfig/*
+
 # Variables to the openssl-sys crate to link statically against the OpenSSL we
 # just compiled above
 export OPENSSL_STATIC=1


Diggsey:
@malbarbo Thanks, this looks important to fix ASAP

However, modifying pkgconfig directly with a regex seems a bit brittle, and if it breaks, we won't realise until too late.

What do you think of the following:
1) Cache the build directory instead of the install directory (the build directory can be safely renamed to retain the atomicity of the cache update, without breaking any paths) and run the `make install` command on every build (I would expect it to be relatively fast, but if that's not the case we can rethink).

2) Add a test to verify that curl was built with ssl support, whatever's easiest to implement, just so long as a regression here causes the build to fail.

malbarbo:
@Diggsey I added a test to verify that rustup-init was built with ssl support.

I don't known exactly how to do the first point and I'm afraid that I cannot investigate it right now. Feel free to continue this fix.

Before a new release, consider fixes for https://github.com/rust-lang-nursery/rustup.rs/issues/1047, https://github.com/rust-lang-nursery/rustup.rs/issues/1055 and https://github.com/rust-lang-nursery/rustup.rs/issues/1058

malbarbo:
Docker was using the cached openssl build... Let's see how it goes now.

malbarbo:
@Diggsey travis is green now.

malbarbo:
@brson Considering that this is an important fix, you may be interest.

jelford:
Fwiw a revert of #1028 would also be a reasonable option as a quick fix (it's strictly an optimization - when it works - so the only hit would be build time).

malbarbo:
Closing in favor of https://github.com/rust-lang-nursery/rustup.rs/pull/1065;

PR #1056, opened 2017-04-12
Add armv8l support

malbarbo:
Fixes https://github.com/rust-lang-nursery/rustup.rs/issues/1055

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -244,7 +244,7 @@ get_architecture() {
             fi
             ;;
 
-        armv7l)
+        armv7l | armv8l)
             local _cputype=armv7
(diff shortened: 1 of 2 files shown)


malbarbo:
Should `src/rustup-dist/src/dist.rs` be updated? It seems that it does not contains the same cases of `rustup-init.sh`.

ghost:
*In Emperor Palatine accent* DEW IT

Diggsey:
Thanks! Yes I think you'll probably need to add it here too: https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup-dist/src/dist.rs#L165

malbarbo:
Done. I also add a missing armv7l case.

PR #1063, opened 2017-04-13
Allow rustup to handle unavailable packages

erickt:
Before this, if a package was unavailable (like nightly is at the moment),
it would error out with a message like "error: missing key: 'url'" because
at the moment a few of the rust-analysis packages didn't build. This resulted
in the `channel-rust-nightly.toml` to be created with blocks like this:

```toml
[pkg.rust-analysis.target.aarch64-apple-ios]
available = false

[pkg.rust-analysis.target.aarch64-linux-android]
available = false

[pkg.rust-analysis.target.aarch64-unknown-fuchsia]
available = false

[pkg.rust-analysis.target.aarch64-unknown-linux-gnu]
available = true
hash = "be50ffa6f94770929b53bae553977cb6d78b03506f033d14a7251c7b0cdb9035"
url = "https://static.rust-lang.org/dist/2017-04-13/rust-analysis-nightly-aarch64-unknown-linux-gnu.tar.gz"
```

rustup assumed that there'd always be a `hash` and `url`, which is not
the case when packages are unavaible. This patch then just updates
rustup to handle their absence.

Closes #1063

### DIFF
--- a/src/rustup-dist/src/manifest.rs
+++ b/src/rustup-dist/src/manifest.rs
@@ -41,13 +41,17 @@ pub enum PackageTargets {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct TargetedPackage {
-    pub available: bool,
+    pub bins: Option<PackageBins>,
+    pub components: Vec<Component>,
+    pub extensions: Vec<Component>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct PackageBins {
     pub url: String,
     pub hash: String,
     pub xz_url: Option<String>,
     pub xz_hash: Option<String>,
-    pub components: Vec<Component>,
-    pub extensions: Vec<Component>,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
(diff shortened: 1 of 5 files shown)


erickt:
I forgot to mention that with this change the tests pass, and nightly now downloads. I'm not sure how to go about creating a test case for this bug though.

Diggsey:
@erickt Thanks!
The test failures on windows and linux look legitimate. WRT testing this feature - if you look in rustup-mock, you can see that it generates several scenarios. A good way to test would be to add a missing package to one of the manifests in a scenario, and then add a test which tries to install from that manifest.

brson:
Seems ok. There's some forwards-compatibility risk here by relying on the `url` and `hashes` keys, with the addition of [xz](https://github.com/rust-lang-nursery/rustup.rs/pull/1100) packages. That is, if the gz packages ever went away this logic would see the package as unavailable. I don't see the gz packages going away any time soon though.

So I'm ok accepting this package as is. Travis errors are legit though.

Edit: this comment was misreading the patch.

brson:
Hm, after reviewing this does seem to respect the 'available' flag in the manifest, not base its logic on the presence of the hash/url fields?

brson:
Oh right the original bug here is just that rustup errors in a weird way when a component is unavailable, because rcs generats manifests in a way I didn't expect. So the logic here is to avoid reading the url/hash fields when a component is unavailable to deal with the manifests that actually exist in the wild. Patch still seems fine to me.

brson:
I'm rebasing this.

brson:
Rebased and tested. Caught an error in the manifest stringification.

r? @Diggsey

Diggsey:
Test failures are legit.

PR #1065, opened 2017-04-14
Fix OpenSSL linkage by using the final install-directory in the build

jelford:
This PR addresses #1051 by avoiding moving OpenSSL's install-target directory after it's been configured. It also encorporates the regression test suggested by @malbarbo on https://github.com/rust-lang-nursery/rustup.rs/pull/1054.

It still makes sense to move directories about to avoid getting a partially-built copy of openssl, and I think it makes sense to cache the finished product rather than the src/build directory. 

I haven't been able to test the output on one of the affected platforms (although the check on symbols passes) as I don't know a good way to get artefacts out of the travis build.

### DIFF
--- a/ci/build-run-docker.sh
+++ b/ci/build-run-docker.sh
@@ -25,9 +25,7 @@ docker run \
 
 # check that rustup-init was built with ssl support
 # see https://github.com/rust-lang-nursery/rustup.rs/issues/1051
-out=$(nm target/$TARGET/release/rustup-init | grep Curl_ssl_init)
-
-if [ -z "$out" ]; then
-  echo "error: Missing ssl support!!!!" >&2
+if ! (nm target/$TARGET/release/rustup-init | grep Curl_ssl_version &> /dev/null); then
+  echo "Missing ssl support!!!!" >&2
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


malbarbo:
@jelford I checked the test it it does not print the message when it fails. The script uses `set -ex`, so if grep fails the script will fail before executing if statement. I you find appropriated, you can update to:

```bash
# check that rustup-init was built with ssl support
# see https://github.com/rust-lang-nursery/rustup.rs/issues/1051
if ! (nm target/$TARGET/release/rustup-init | grep Curl_ssl_version &> /dev/null); then
  echo "Missing ssl support!!!!"
  exit 1
fi
```

jelford:
Thanks @malbarbo , incorporated your change.

ghost:
We want this!!!

Diggsey:
@bors r+
Thanks @jelford and @malbarbo, this looks great.

I can't make a release unfortunately, that'll have to wait until @brson is around.

alexreg:
Could we get this merged soon, hopefully? It's quite annoying not to be able to update, right now.

jelford:
@Diggsey this is going to start to get more urgent with 1.17 shipping. Any timeline on when we might be able to get something pushed out? Sorry to nag.

Diggsey:
@jelford There's nothing I can do about it unfortunately. @brson is aware of the issue and is going to make a new release ASAP, but he's been away and I imagine has had a ton of stuff to catch up on since getting back.

jelford:
Okay, thanks for the update, no problem. Sorry for all this.

Issue #730, opened 2016-09-23
Add 'rustup uninstall' alias


comex:
It's confusing that there's a `rustup install` equivalent to `rustup toolchain install`, but no `rustup uninstall` to match `rustup toolchain uninstall`.

(It's also a bit weird that the documentation for `rustup install` states that it's an alias for `rustup update <toolchain>`, which in turn is documented as an alias for `rustup toolchain install` - flipping between the verbs "install" and "update".)

Not sure whether `remove` should also be aliased.

clarfonthey:
Is there going to be any progress on this?

Diggsey:
@clarcharr PRs are welcome

kbknapp:
It's a [one line addition](https://github.com/rust-lang-nursery/rustup.rs/blob/705c110e83c609e7901992058f0f58e26a623a4f/src/rustup-cli/rustup_mode.rs#L315). Could be a good first PR for someone.

PR #1073, opened 2017-04-18
Alias 'rustup toolchain uninstall' to 'rustup uninstall'

jordanhiltunen:
See issue: #730. 

This PR aliases `rustup toolchain uninstall` to `rustup uninstall`, making uninstallation more symmetrical with installation, which already views `rustup install` and `rustup toolchain install` as equivalent.

The `rustup uninstall` alias is directly patterned off of the implementation of
`rustup install`. I’ve also added a test in cli-rustup.rs to verify.

The issue linked also alludes to some confusing irregularities in the documentation (random flipping between use of 'update' and 'install'), and a softer question about whether `rustup toolchain remove` should be aliased to `rustup remove` as well.

This is my first crack at a PR here, so I was a bit sheepish to wade into those more contentious elements of the issue, but I would be glad to dig deeper into the semantic consistency of these various aliases if there's any interest, and someone thinks I'm headed in the right direction. Thanks!

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -31,6 +31,7 @@ pub fn main() -> Result<()> {
         ("show", Some(_)) => try!(show(cfg)),
         ("install", Some(m)) => try!(update(cfg, m)),
         ("update", Some(m)) => try!(update(cfg, m)),
+        ("uninstall", Some(m)) => try!(toolchain_remove(cfg, m)),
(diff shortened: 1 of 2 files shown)


Diggsey:
@jordanhiltunen Thanks! I don't think we should add further aliases (such as `rustup remove`) because we want to discourage use of those extra aliases (they only exist for backwards compatibility).

Adding the uninstall alias seems fine to me though because of the symmetry with `rustup install`, which exists for convenience rather than just backwards compatibility.

PR #1088, opened 2017-04-29
Update proxy setting section

upsuper:
There are two changes:
1. Use `https_proxy` rather than `http_proxy` because the latter doesn't work at all. All traffic from rustup seems to go via https.
2. Use SOCKS proxy as example, since it is nowadays more widely-used (thanks to GFW in China) and more powerful. Also, avoiding using two `http`s makes it more obvious that other proxy types are also supported without needing to check the manual of curl.


upsuper:
For the first change, you can test that, specifying an invalid proxy on `http_proxy` would not trigger any error, but specifying an invalid proxy on `https_proxy` would cause error immediately.

Issue #478, opened 2016-05-17
PATH support doesn't work for fish shell


brson:
It's a popular shell and doesn't read `.profile`. cc @nagisa cc https://github.com/rust-lang-nursery/rustup.rs/pull/468

nagisa:
Many linuxes should be covered by putting the environment to both `~/.pam_environment` and `~/.profile`, though it is a bag of worms because either or both can be disabled, not run or configured to be in another, non-standard location (i.e. not `$HOME/.file`).

For OS X it seems like you’d want to also inform launchd about the updated path with `launchctl setenv PATH $PATH`.

Either way this goes, fixing https://github.com/rust-lang-nursery/rustup.rs/issues/473 is another thing that would fix a lot of the problem here.

snheath:
FWIW OCaml's opam has a similar problem and handles it just by having a version of the add-stuff-to-PATH hook for each shell.  There's also the "eval (opam config env)" idiom which sets the appropriate environment variables to point to wherever opam thinks they need to point to, including PATH.

ariasuni:
rustup create `env` with `export PATH="$HOME/.cargo/bin:$PATH"`. Why not create too a file `env.fish` (like opam) with `set -gx PATH "$HOME/.cargo/bin" $PATH;`?

robey:
this still doesn't work. `~/.cargo/env` is completely wrong.

icefoxen:
I think you can just drop the `set -gx ...` command in a file in `~/.config/fish/conf.d/`, see <http://fishshell.com/docs/current/index.html#initialization>

chriswoollard:
I have had this issue a few times. It is easy to resolve.

Create a fish config file in the below directory, if it doesn't already exist.

`~/.config/fish/config.fish`

Then add the following line to the file.

`export PATH="$HOME/.cargo/bin:$PATH"`

Once you restart your terminal, the rust commands should work again.

ndac-todoroki:
Exporting in a more fisher way
```fish
set PATH $HOME/.cargo/bin $PATH
```
in `~/.config/fish/config.fish` maybe better.

mojotx:
An even more fishier way would be to use the fish_user_paths universal variable. 

Example:
`set -U fish_user_paths $HOME/.cargo/bin $fish_user_paths`

No need to mess with config files. A universal variable, once set, is permanently and persistently set across all fish shell sessions. The contents of that variable are automatically prepended to the $PATH environment variable.

[...]

PR #1090, opened 2017-05-03
Edit .bash_profile if it exists

### LINKED ISSUE #883: PATH to cargo/bin is in wrong file, should be in .bash_profile (CentOS)
CentOS is default AWS EC2 version of Linux, please consider to add path into .bash_profile also, at least if this file exist.

awelkie:
Fixes #883

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -991,6 +991,16 @@ fn get_add_path_methods() -> Vec<PathUpdateMethod> {
             let zprofile = zdotdir.map(|p| p.join(".zprofile"));
             profiles.push(zprofile);
         }
+
+        if shell.contains("bash") {
+            if let Some(bash_profile) = utils::home_dir().map(|p| p.join(".bash_profile")) {
+                // Only update .bash_profile if it exists because creating .bash_profile
+                // will cause .profile to not be read
+                if bash_profile.exists() {
+                    profiles.push(Some(bash_profile));
+                }
+            }
+        }
     }
 
     let rcfiles = profiles.into_iter().filter_map(|f|f);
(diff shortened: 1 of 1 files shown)


Diggsey:
Hi @awelkie, this looks great.

I would like to slightly alter the detection logic here, specifically:
- Modify `.zprofile` if *either* it already exists, *or* the shell is zsh
- Modify `.bash_profile` only if it exists, regardless of the current shell

IMO, this would provide the best result given each shell's profile logic.

However, I'd be happy to accept this PR as-is if you won't be able to make these changes, since it's a clear improvement over the status quo, and shouldn't cause any problems to tweak later.

brson:
@Diggsey's suggestion sounds reasonable to me. Thanks @awelkie!

awelkie:
I'll take a crack at making these changes. I would also like to add support for fish (#478). Should the logic there be to add add a file to ~/.config/fish/conf.d only if ~/.config/fish/ exists, regardless of the current shell?

Diggsey:
@awelkie Thanks!

For all shells except bash, we should set their configuration if we detect their presence at all on the system. For the moment, the two detection methods we use are `$SHELL` and whether their configuration files are detected.

Bash is different, because it will fall back to `.profile` if `.bash_profile` does not exist, so even if `$SHELL == "bash"`, we can't create `.bash_profile` if it doesn't already exist, as then bash would stop looking in `.profile`, which would break things for the user.

So for fish, we want to add a file to `~/.config/fish/conf.d` if *either* the fish config directory already exists, *or* if fish is the current shell.

awelkie:
I don't think I'll be able to get to the making the improvements for another few days, so feel free to merge this in if you want.

brson:
For this PR I agree that at least it would be better to remove reading "SHELL" and just look for the presense of .bash_profile. zsh changes can be left for some other PR.

This should really also have a test case. Crib off of "install_adds_path_to_profile".

brson:
I've updated this in #1179.

PR #1093, opened 2017-05-05
Add the rust lib dir (containing std-<hash>.dll) to the path on windows

mcgoo:
I'm not sure if this is the right thing or not, but this fixes some of https://github.com/rust-lang/cargo/issues/3394. My apologies if it undoes earlier path related fixes.

The dylib tests in the cargo testsuite work.
`cargo test` of a dylib crate works.
It does nothing for the `cargo install` case.

It adds sysroot/bin to the path on Windows which fixes finding libstd. A side effect of this is that rustc.exe is directly in the path although it is after CARGO_HOME/bin so the shim should be found first.

The behavior of prepend_path in case of a corrupt path is changed - it used to clobber the existing path with the new path, and now it will leave leave the existing path unchanged.

It leaves LD_LIBRARY_PATH set on Windows also as I believe it is used by MinGW.

### DIFF
--- a/src/rustup/env_var.rs
+++ b/src/rustup/env_var.rs
@@ -1,6 +1,5 @@
-use std::ffi::OsString;
 use std::env;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::process::Command;
 
 pub const RUST_RECURSION_COUNT_MAX: u32 = 5;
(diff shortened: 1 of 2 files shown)


brson:
Thanks for tackling this @mcgoo. This path munging code ... omg, so frustrating.

It's so hard to imagine what is going on at runtime with all the layers of indirection and environment frobbing.

A worry I have with this patch is that it might accidentally be reverting what [the original PR](https://github.com/rust-lang-nursery/rustup.rs/pull/812) that caused this was fixing - that all `cargo` invocations go through `.cargo/bin`. I _think_ this patch preserves that quality from reading it over several times.

I guess I'm willing to try this and see what happens. The three failures on appveyor seem legit. Probably some patch-munging corner case that requires things to be perfectly right in a way that is contradictory to other goals.

Diggsey:
If this regresses something, it means we need more regression tests.

The test failures on appveyor are a known flaky test, but it's surprising that three failed, let's give bors a try...
@bors r+

mcgoo:
I actually ended up with [this](https://github.com/rust-lang/cargo/pull/4006) change in cargo that fixes the path for both cross and native builds, which rustup would not be able to do easily, so anything running under cargo is fixed already.

I still think this change is correct from the standpoint of consistency with other platforms, but I don't know what it actually fixes as such anymore. `rustup run` and the debuggers I guess?

PR #1097, opened 2017-05-06
Improving description of installation process

marcogbarcellos:
I started learning and using rust and I think this description would help beginners to install rustup faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,15 @@ And it runs on all platforms Rust supports, including Windows.
 
 ## Installation
 
-Follow the instructions at [www.rustup.rs](https://www.rustup.rs). If
-that doesn't work for you there are [other installation
+Install [curl](https://curl.haxx.se/download.html), then run the following in your terminal (follow the onscreen instructions) :
+
+```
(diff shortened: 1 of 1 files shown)


Diggsey:
Hi @marcogbarcellos, thanks for the PR. However, this change does not take into account platforms where curl is not the recommended installation method, and when curl *is* required, most people should obtain it through their package manager rather than downloading from this website. This is why we link to `rustup.rs`, where the installation instructions can be provided in a platform-specific manner.

We *would* like to help beginners get started sooner - perhaps you could tell us what platform you are on, and could explain the difficulties you had with the existing installation instructions? We may be able to improve the platform-specific installation instructions in response to that.

marcogbarcellos:
sounds fair @Diggsey , no problems. I'm working on a Mac but it was not a big deal to install it, i just thought reading it on the README.md instead of going to another link would make it faster, but that's clear and easy enough, even for people beginning to code(not my case, i'm just a beginner in learning rust). Also the website is the official one following the [curl github description](https://github.com/curl/curl) Thanks!

Issue #1092, opened 2017-05-05
rustup nightly install failing


davidroeca:
Just getting back into rust and installed rustup. Tried to install the nightly build and encountered the following error:

```bash
$ rustup install nightly
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: downloading component 'rustc'
info: downloading component 'rust-std'
info: downloading component 'cargo'
info: downloading component 'rust-docs'
info: installing component 'rustc'
info: rolling back changes
error: failed to extract package
info: caused by: failed to unpack `rustc-nightly-x86_64-unknown-linux-gnu/rustc/share/man/man1/rustdoc.1` into `/home/david/.rustup/tmp/wnj79prt3m8vg6f2_dir/rustc/share/man/man1/rustdoc.1`
info: caused by: No such file or directory (os error 2)
```

Not sure if this is a nightly issue, a rustup issue, or my own lack of experience with rustup. Currently running Linux Mint 18

messense:
Same issue on macOS 11.12

est31:
I can confirm on Ubuntu 17.04. `rustup update` is broken as well. @davidroeca this is a very recent issue.

gyscos:
Here is what I get after installing the current rustup from git (to get debug info and backtraces):
```
% ~/.cargo/bin/rustup update nightly
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: downloading component 'rustc'
info: downloading component 'rust-std'
info: downloading component 'cargo'
info: downloading component 'rust-docs'
info: downloading component 'rust-analysis'
info: downloading component 'rls'
info: downloading component 'rust-src'
info: downloading component 'rust-std' for 'asmjs-unknown-emscripten'
info: downloading component 'rust-std' for 'wasm32-unknown-emscripten'
info: installing component 'rustc'
info: rolling back changes
error: failed to extract package
info: caused by: failed to unpack `rustc-nightly-x86_64-unknown-linux-gnu/rustc/share/man/man1/rustdoc.1` into `/home/gyscos/.rustup/tmp/_mffp8148_7wf8w9_dir/rustc/share/man/man1/rustdoc.1`
info: caused by: No such file or directory (os error 2)
info: backtrace:

stack backtrace:
   0:     0x55f2010a2894 - backtrace::backtrace::libunwind::trace
[...]

PR #1098, opened 2017-05-08
Ensure that intermediate directories exist when unpacking an entry

ranma42:
The `unpack` function assumes that the directory in which the file is
being extracted exists, while most `tar` tools will automatically
create the intermediate directories if they are missing.

This would have avoided #1092.

### DIFF
--- a/src/rustup-dist/src/component/package.rs
+++ b/src/rustup-dist/src/component/package.rs
@@ -199,6 +199,14 @@ fn unpack_without_first_dir<R: Read>(archive: &mut tar::Archive<R>, path: &Path)
         // Throw away the first path component
         components.next();
         let full_path = path.join(&components.as_path());
+
+        // Create the full path to the entry if it does not exist already
+        match full_path.parent() {
+            Some(parent) if !parent.exists() =>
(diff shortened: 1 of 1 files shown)


brson:
@bors r+

Note though that we still cannot remove the directory entries from the tarballs produced by Rust, because they need to be compatible with older rustups. I'm not sure how long a suitable deprecation period is.

ranma42:
Yes, I have no plans for removing them and the Rust port of rust-installer will keep them as well for the time being. Removing them was just convenient when sorting the paths.

brson:
Legit errors.

ranma42:
It is easier to fix without updating tar. Updated title and first post to get reasonable commit messages upon merging.

brson:
@bors r+

I can imagine this repeated syscall wanting to be optimized someday, but today I'm not too worried. rustup generally needs to be optimized.

ranma42:
Is the failure spurious? The non-merge AppVeyor build succeeded

Diggsey:
Yes - @bors r=brson retry

PR #1113, opened 2017-05-12
Clarify that rustup update updates rustup itself

### LINKED ISSUE #873: Clarify help for update
It's mentioned in the [docs](https://github.com/rust-lang-nursery/rustup.rs#keeping-rustup-up-to-date), but when you run `rustup --help`:

```bash
...
update         Update Rust toolchains
...
```

`rustup update` also updates rustup itself. While I think this is a good idea, I was a little surprised by the action. `rustup update --help` does explains things well.

Could consider changing the top level help to something like:

```bash
update         Update Rust toolchains and rustup
```

Since you can also do `rustup self update`, maybe this could be confusing though.

johnthagen:
Fix #873


Diggsey:
@johnthagen Thanks!

@bors r+

Issue #1103, opened 2017-05-08
remove_override_nonexistent test is flaky on windows


brson:
It might be because of `TempDir` failing to delete the directory. Upgrading to a tempdir that contains the [aggressive remove_dir_all patch](https://github.com/rust-lang-nursery/tempdir) could fix it.

cc https://github.com/rust-lang-nursery/rustup.rs/pull/1102

brson:
tempdir has been updated, but no new build published

PR #1115, opened 2017-05-12
Re-enable doc uploading on travis

mattico:
rust-lang/rust#32532 is fixed, so this might work now. Let's find out.


mattico:
AppVeyor failure is spurious, I think. Possibly related to #1103.

mattico:
Also I realize now that this doesn't actually run unless it's on the stable branch, so a green Travis run doesn't mean anything...

Diggsey:
@mattico Thanks, let's give it a try!

@bors r+

PR #1116, opened 2017-05-12
Remove unnecessary pub mod

mattico:
rust-lang/rust#18241 is fixed now, so this mod doesn't need to be pub.

Possible downside: the minimum compiler version to build is now 1.9.0 or thereabouts.


Diggsey:
Thanks!

@bors r+

PR #1117, opened 2017-05-12
Use ShellExecute rather than start.exe to open docs on windows

### LINKED ISSUE #499: Use ShellExecute to open docs on windows
rustup 0.1.12 (c6e430a 2016-05-12)
Windows 10 x64 Insider Preview build 14342

Running any of the `doc` subcommands opens a new command prompt, but does not display the docs. I've tested this from cmd.exe, powershell.exe, and the MSYS2 shell.

Let me know if you need any other info to help debug this issue.

Edit:
I've tried running the command as found in https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup-utils/src/raw.rs#L542, which works fine.

`cmd /C start https://github.com`

mattico:
Closes #499

The implementation is not too complicated, so it's a bit better than spawning a cmd.exe just to open docs. It also should fix opening docs for the few people still using a year-old insider preview build.

### DIFF
--- a/src/rustup-utils/src/raw.rs
+++ b/src/rustup-utils/src/raw.rs
@@ -372,6 +372,7 @@ pub fn open_browser(path: &Path) -> io::Result<bool> {
         use winapi;
         use std::ptr;
 
+        // FIXME: When winapi has this function, use their version
         extern "system" {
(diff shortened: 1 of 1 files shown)


Diggsey:
Thanks, this looks great!

@bors r+

PR #1119, opened 2017-05-14
Add a message when removing a component

### LINKED ISSUE #306: No info printed on `rustup target remove`
Right now a successful `rustup target remove` doesn't actually print anything, whereas most other commands print out something like "successfully did blah".

matthewjasper:
Fixes #306.
The message is (for example)
```
info: removing component 'rust-std' for 'x86_64-unknown-linux-musl'
```


Diggsey:
Thanks! It looks like some of the tests need updating?

matthewjasper:
Hmm, do we want to show the removing message when upgrading, or just when actually removing?

Diggsey:
Just keep it simple for now and always show it.

Issue #756, opened 2016-10-12
rustup tries to sync channel updates for non-updatable channels


brson:
@alexcrichton says it's trying to sync channel updates for old versions when it shouldn't.

brson:
I continue to see this error:

```
brian@ip-10-145-43-250:/mnt2/dev/quandl⟫ rustup update
info: syncing channel updates for 'stable-i686-unknown-linux-gnu'
info: syncing channel updates for 'stable-x86_64-pc-windows-gnu'
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: syncing channel updates for 'beta-x86_64-unknown-linux-gnu'
info: downloading component 'rustc'
info: downloading component 'rust-std'
info: downloading component 'cargo'
info: downloading component 'rust-docs'
info: installing component 'rustc'
info: installing component 'rust-std'
info: installing component 'cargo'
info: installing component 'rust-docs'
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: downloading component 'rustc'
info: downloading component 'rust-std'
info: downloading component 'cargo'
info: downloading component 'rust-docs'
info: installing component 'rustc'
info: installing component 'rust-std'
info: installing component 'cargo'
info: installing component 'rust-docs'
info: syncing channel updates for '1.14.0-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.15.0-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.16.0-x86_64-unknown-linux-gnu'
info: checking for self-updates

    stable-i686-unknown-linux-gnu unchanged - rustc 1.17.0 (56124baa9 2017-04-24)
     stable-x86_64-pc-windows-gnu unchanged - (rustc does not exist)
  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.17.0 (56124baa9 2017-04-24)
      beta-x86_64-unknown-linux-gnu updated - rustc 1.18.0-beta.2 (f4e8e81c3 2017-05-11)
   nightly-x86_64-unknown-linux-gnu updated - rustc 1.19.0-nightly (e40beb3af 2017-05-11)
  1.14.0-x86_64-unknown-linux-gnu unchanged - rustc 1.14.0 (e8a012324 2016-12-16)
  1.15.0-x86_64-unknown-linux-gnu unchanged - rustc 1.15.0 (10893a9a3 2017-01-19)
[...]

PR #1121, opened 2017-05-16
fix ToolchainDesc is_tracking

mikhail-m1:
issue #756.

I wonder, why is not ToolchainDesc.channel a enum?


Diggsey:
Thanks!

Rustup has evolved a lot over time, and was originally ported from a bash script, so there are certainly some hold-over design decisions that are not very "rustic". If you'd like to help rectify that, we'd be grateful for assistance with any "help-wanted" issues like this one, which should help you get familiar with the code-base.

@bors r+

PR #1124, opened 2017-05-17
fixes prompt asking for msvc even though -y is provided

### LINKED ISSUE #916: Windows confirmation prompt not disabled with -y
rustup-init.exe -y doesn't suppress all the confirmation prompts on Windows with 1.0.0.
https://github.com/rust-lang-nursery/rustup.rs/issues/912#issuecomment-271193357

pxdeu:
Fixes #916 
no_prompt was simply not being checked

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -219,7 +219,7 @@ pub fn install(no_prompt: bool, verbose: bool,
     try!(do_pre_install_sanity_checks());
     try!(do_anti_sudo_check(no_prompt));
 
-    if !try!(do_msvc_check(&opts)) {
+    if !no_prompt && !try!(do_msvc_check(&opts)) {
         info!("aborting installation");
         return Ok(());
     }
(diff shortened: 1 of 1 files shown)


Diggsey:
@pxdeu Thanks for the PR!

I think in this case, we should still perform the MSVC check, even when `-y` is passed, but in the event that it fails, just print a warning instead of prompting the user. What do you think?

pxdeu:
Sounds like a good idea. I thought about it when working on the issue. I only decided against it, because it would either slightly change the meaning of the function do_msvc_check (to not include the prompt) or add parameters to it.

Looking at it again, not including the prompt seems to be more in line with the rest of the code and the function is not called elsewhere as far as I can tell.

I can provide a new commit.

Diggsey:
LGTM @bors r+

pxdeu:
I didn't notice MSVC_MESSAGE was windows only, because I built it on windows.
Glad you have a CI in place to catch things like that.

Issue #820, opened 2016-11-18
Improve error message when override doesn't exist


est31:
Try this:

* `rustup override add nightly-2016-11-02`
* *downloads specified nightly and sets it as override*
* `rustup toolchain remove nightly-2016-11-02`
* *removes toolchain*
* `rustup show`

Gives output:
```
error: toolchain 'nightly-2016-11-02-x86_64-unknown-linux-gnu' is not installed
info: caused by: not a directory: '~/.multirust/toolchains/nightly-2016-11-02-x86_64-unknown-linux-gnu'
```

It can be fixed by removing the override but I think that should already happen when you do `rustup toolchain remove`

brson:
Good find. Thank you.

brson:
Hm, after thinking about this, I'm not sure that removing toolchains should remove overrides. Just because the toolchain doesn't exist doesn't invalidate the need for that toolchain to build a project. For now I'd prefer to just have better error messages here.

brson:
Hm, this error basic error is not horrible, but is obscured by the 'cause', which is irrelevant here. A better error though would indicate that an override requires this toolchain but it is not installed, install it with ... etc.

est31:
Or one could warn about present overrides when uninstalling.

brson:
I think I prefer to solve this by simply making a better error message, like:

```
error: override toolchain 'nightly-2016-11-02-x86_64-unknown-linux-gnu' is not installed
```

And getting rid of the causal error `info: caused by: not a directory: '~/.multirust/toolchains/nightly-2016-11-02-x86_64-unknown-linux-gnu'`, which is an implementation detail.

It would be even better if rustup suggested how to remove the override, but the current UI doesn't have a way to do that.

shiver:
I took a quick stab at resolving this. The output should look something like this:
```
$ rustup show
Default host: x86_64-unknown-linux-gnu

error: override toolchain 'nightly-2016-11-02-x86_64-unknown-linux-gnu' is not installed
```
Hopefully that is what you were after.
Thanks!

workingjubilee:
Can this be closed? @kinnison

est31:
Indeed, `rustup show` now re-downloads the nightly instead of showing an error. So I'd say it's resolved now.

PR #1128, opened 2017-05-19
#820 Supress confusing NotADirectory error and show override missing

shiver:
This is my attempt at tackling issue #820. 
Hopefully I interpreted the original request correctly and this is what you were looking for.

It is my first time dealing with error_chain, so not 100% sure I did it right.

Any suggestions you might have to improve the submission would be great.
Thanks!


Diggsey:
Thanks for the PR! It all looks good.

@bors r+

PR #1131, opened 2017-05-22
Cleanup download-related code in the rustup_dist crate

jelford:
This PR is probably easiest to read if you do it commit-wise: the first commit just removes the old, unused `download` module from rustup_dist. The second commit pulls `DownloadCfg` (and friends) out into a _new_ `download` module.

The only other change in there is that I've tried to kill some of the instances where the `notify_handler` gets needlessly passed around by moving functions to be methods on `DownloadCfg` + using the field it's already got. I suspect the only reason these weren't done before is that it was difficult to reason about the code, so hopefully this is an improvement.

There aren't any (intentional) functional changes in this, so I haven't made any changes to the tests (aside from those required to get it to compile). No rush to get this merged for the same reason.


src/rustup-dist/src/download.rs
```diff
@@ -1,95 +1,186 @@
...
             }
+        }
+
 
-            Ok(target_file)
+        let partial_file_path =
+            target_file.with_file_name(
+                target_file.file_name().map(|s| {
+                    s.to_str().unwrap_or("_")})
+                    .unwrap_or("_")
```
Diggsey, line 65: Hm, I know you didn't add this, but falling back to underscore seems like a bad idea, and the filename is supposed to be the hash anyway, so this should be an `expect` instead.

Diggsey:
This LGTM, thanks! The underscore thing I commented on should probably be fixed separately.

@bors r+

PR #1135, opened 2017-05-24
README: fix rust build dir

RalfJung:
The old text looks like a typo for me... why would there be a folder `~/rustg`? There isn't for me.


Diggsey:
Thanks! Definitely just a typo.

PR #1141, opened 2017-05-30
set_file_perms: if the file is already executable, keep it executable

### LINKED ISSUE #1140: rustup does not preserve permissions in toolchain components
The rust-src component contains some files that are marked executable, e.g. `src/jemalloc/include/jemalloc/internal/private_namespace.sh`. When I download the component manually from <https://static.rust-lang.org/dist/rust-src-nightly.tar.xz>, I can see that the permissions are correct in that file (same for the .gz). However, `~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/jemalloc/include/jemalloc/internal/private_namespace.sh` is not executable, so permissions get lost somewhere on the way.

I am trying to figure out how rustup goes about extracting the tarball, but the code is so abstract I couldn't find the right place yet...

### LINKED ISSUE #36488: can't bootstrap compiler from the rust-src component because of file permissions
All the files, even the scripts (*.sh), have their permissions set to 644.

Bootstrapping fails while executing alloc_jemalloc build script with the following error message:

```
$ chmod +x $(rustc --print sysroot)/lib/rustlib/src/rust/configure
$ $(rustc --print sysroot)/lib/rustlib/src/rust/configure --enable-rustbuild --target=arm-unknown-linux-gnueabi
$ make
<snip>
--- stderr
./config.status: line 1197: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/internal/private_namespace.sh: Permission denied
./config.status: line 1201: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/internal/private_unnamespace.sh: Permission denied
./config.status: line 1221: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/internal/public_namespace.sh: Permission denied
./config.status: line 1225: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/internal/public_unnamespace.sh: Permission denied
./config.status: line 1237: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/jemalloc_rename.sh: Permission denied
./config.status: line 1241: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/jemalloc_mangle.sh: Permission denied
./config.status: line 1245: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/jemalloc_mangle.sh: Permission denied
./config.status: line 1249: /home/japaric/.multirust/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc_jemalloc/../jemalloc/include/jemalloc/jemalloc.sh: Permission denied
<snip>
```

cc @brson @Diggsey 
### Meta

```
$ rustc -V
rustc 1.13.0-nightly (5531c314a 2016-09-12)
```

RalfJung:
This was the smallest change I could come up with to fix <https://github.com/rust-lang/rust/issues/36488>: Maintain the X bit when fixing up permissions.

However, the code is still somewhat inconsistent in that it does apply different rules when applied to a file vs. when applied to a directory -- the code recursively walking the directory still removes the X bits from all files, just like it previously ignored the "bin" directory. If you want, I can refactor this some more to be consistent here.

Fixes: #1140

### DIFF
--- a/src/rustup-dist/src/component/package.rs
+++ b/src/rustup-dist/src/component/package.rs
@@ -123,9 +123,15 @@ impl Package for DirectoryPackage {
 // but due to rust-lang/rust#25479 they don't.
 #[cfg(unix)]
 fn set_file_perms(dest_path: &Path, src_path: &Path) -> Result<()> {
-    use std::fs;
+    use std::fs::{self, Metadata};
     use std::os::unix::fs::PermissionsExt;
     use walkdir::WalkDir;
+
+    // Compute whether this entry needs the X bit
+    fn needs_x(meta: &Metadata) -> bool {
+        meta.is_dir() || // Directories need it
+        meta.permissions().mode() & 0o700 == 0o700 // If it is rwx for the user, it gets the X bit
+    }
 
     // By convention, anything in the bin/ directory of the package is a binary
     let is_bin = if let Some(p) = src_path.parent() {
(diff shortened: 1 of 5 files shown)


RalfJung:
Not sure what these traivs failures are, the logs are empty...

Diggsey:
Thanks for the PR!

Is there a reason we can't just preserve the permissions exactly as they are in the source package? Did previous packages have incorrect permissions, and we want to retain backwards compatibility?

I think it would be helpful to have tests covering the specific cases we want to support.

RalfJung:
> Is there a reason we can't just preserve the permissions exactly as they are in the source package? Did previous packages have incorrect permissions, and we want to retain backwards compatibility?

They definitely used to have incorrect permissions: <https://github.com/rust-lang/rust/issues/25479>
It looks to me like this is no longer the case (i.e., when I checked briefly, the permissions looked all right), but I am not sure and I don't want to break everything, so this change is as conservative as I could make it.

> I think it would be helpful to have tests covering the specific cases we want to support.

How would these tests look like? Are tests allowed to access the internet, so that they could download the latest stable, beta, nightly and check the permissions of some files?

Diggsey:
We have code in `src\rustup-mock\src\dist.rs` to generate mock tarballs. It should be possible to generate tarballs with both legacy and new-style permissions, and verify that the permissions are correct after installation.

I also think we should try to avoid messing with the permissions as much as possible, given that they are now correct. For example, instead of overwriting all permissions, we could preserve them, but specifically add the "execute" permission to all files from the bin directory, where it's not already present.

Diggsey:
Hm, apparantly there is precedent for only preserving the execute bit, in eg. git. - https://github.com/rust-lang-nursery/rustup.rs/issues/313#issuecomment-306302316

vi:
Yes. Although Git tree format supports all permission bits, only one bit is actually used.

RalfJung:
> We have code in src\rustup-mock\src\dist.rs to generate mock tarballs. It should be possible to generate tarballs with both legacy and new-style permissions, and verify that the permissions are correct after installation.

Is there any testcase already using that? I couldn't find any, and that makes it kind of hard to figure out how the API is supposed to be used...

EDIT: Ah, I found rustup-dist/tests/install.rs. That looks good. There even is a unix_permissions test in there. I could extend that.
However, it seems to me that this involves extending MockComponent (the one in rustup-mock/src/lib.rs, *not* the one in dist.rs) to also specify permissions for the file -- or, at the very least, the X bit. Would that be all right?

RalfJung:
All right I (a) did some refactoring in `set_file_perms` such that it also preserves X bits when it copies entire directories, and (b) extended the test suite to test this preservation. `cargo test -p rustup-dist` shows up all green.

RalfJung:
I can of course change this to just bit-OR `0o111` to the existing permissions for everything inside a `bin` directory, and otherwise leave everything untouched. For now, I just figured I'd change as little as possible because things mostly work the way they are right now.

Diggsey:
@RalfJung this looks great to me! Thanks for taking the time to add tests, it's much appreciated.

@bors r+

PR #1145, opened 2017-06-01
VS 2017 updates

brson:
Once https://github.com/rust-lang/rust/issues/42225 lands rustup should build correctly under VS 2017.

r? @Diggsey


Diggsey:
Not much to review here 😛 

@bors r+

PR #1156, opened 2017-06-09
Disable doc generation again 

brson:
Doesn't work because cargo doesn't exist on the travis image. I need to get a release out now, so just reverting.

cc @mattico


mattico:
I'm not surprised it didn't actually work, CI is a moving target. Sorry it caused you trouble!

PR #1162, opened 2017-06-12
sha2 version update

newpavlov:
Version 0.6 of `sha2` crate is independent from `rust-crypto` and should be a bit faster compared to v0.1.2.


brson:
Thanks @newpavlov !

lilianmoraru:
@brson What do you think about switching the archives(.sha256 -> .sha512) and this to SHA-512 for collision resistance and speed improvements?
This comment(and the entire issue actually) might be of interest: https://github.com/mozilla/sccache/issues/108#issuecomment-296006697

Note that ring still uses a lot of assembly and C code.

Diggsey:
@lilianmoraru The hash is calculated concurrently with the download so it's much more likely to be bottle-necked on network speed, and in terms of collision resistance there's not enough of a difference between sha256 and sha512 to make it worth-while IMO:

sha256 will require significant breakthroughs in order to achieve even a practical collision attack, and if such an attack does become practical, it is more likely to be caused by a weakness in the algorithm (which would then affect sha512 too) than by the digest length being too short.

lilianmoraru:
@Diggsey It just seems like a win-win(considering that we are in the 64-bit world for a good while - most users would be positively affected by such a simple change).

Diggsey:
But neither of them *are* wins for our users. This change will neither speed up rustup, nor improve security in a measurable way.

PR #1192, opened 2017-07-03
Teach hyper/rustls HTTP backends to resume partial downloads

### LINKED ISSUE #1181: Teach hyper/rustls HTTP backends to resume partial downloads
The curl backend can already do it, but not the two hyper backends. It should just be a matter of threading the parameter through and setting the HTTP [Range header](http://stackoverflow.com/questions/3303029/ddg#18745164).

Test by compiling with `--features=rustls` and `RUSTUP_USE_HYPER` / `RUSTUP_USE_RUSTLS`, and `--verbose`.

No way to actually write test cases for this though.

ghost:
This PR Fixes #1181

### DIFF
--- a/src/download/src/lib.rs
+++ b/src/download/src/lib.rs
@@ -40,13 +40,6 @@ fn download_with_backend(backend: Backend,
     }
 }
 
-fn supports_partial_download(backend: &Backend) -> bool {
-    match backend {
-        &Backend::Curl => true,
-        _ => false
-    }
-}
-
 pub fn download_to_path_with_backend(
     backend: Backend,
     url: &Url,
@@ -60,7 +53,7 @@ pub fn download_to_path_with_backend(
     use std::io::{Read, Write, Seek, SeekFrom};
 
     || -> Result<()> {
-        let (file, resume_from) = if resume_from_partial && supports_partial_download(&backend) {
+        let (file, resume_from) = if resume_from_partial {
             let possible_partial = OpenOptions::new()
                     .read(true)
                     .open(&path);
--- a/src/download/src/lib.rs
+++ b/src/download/src/lib.rs
@@ -35,8 +35,8 @@ fn download_with_backend(backend: Backend,
                              -> Result<()> {
     match backend {
         Backend::Curl => curl::download(url, resume_from, callback),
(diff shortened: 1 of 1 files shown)


src/download/src/lib.rs
```diff
@@ -612,7 +615,12 @@ pub mod hyper_base {
...
 
-        let mut res = try!(client.get(url.clone()).send()
+        use std::cmp;
+        let byte = cmp::max(resume_from, 0);
+        let header = Range::Bytes(vec![ByteRangeSpec::AllFrom(byte)]);
+        let req = client.get(url.clone()).header(header);
+
+        let mut res = try!(req.send()
                            .chain_err(|| "failed to make network request"));
         if res.status != self::hyper::Ok {
```
seanmonstar, line 619: A request with a `Range` header will return a `PartialContent` (a 206), so this needs to be updated to handle that status.

seanmonstar (changes requested):
This would also need to update the `supports_partial_download` to return true for these backends (if all are true, is the function even still needed?).

It'd probably be a good idea to add to the tests of the `download` crate, though it would need a server to serve a file for hyper to download, since the current tests use curl's ability to "download" from a `file://` url.

src/download/src/lib.rs
ghost: @seanmonstar thanks, I'll address this then :)

Could you give me more information about how to write the tests you mentioned?

seanmonstar: Writing the tests will be tricky. It'd require starting up an HTTP server and serving partial contents of a file. The currently depended on version of hyper doesn't allow servers to be torn down when the test is done. It might need to wait for an upgrade to that...

ghost: Oh, I see then

ghost:
@seanmonstar I addressed the things you mentioned and added two new commits.
I can squash them whenever this PR gets merged.

But for some reason AppVeyor keeps failing. From a quick search online, it appears the reason is a failure to verify a CA certificate.

alexcrichton:
Thanks for the PR and sorry for the delay in getting to this! I'm going to close this in favor of https://github.com/rust-lang-nursery/rustup.rs/pull/1222 which moves to reqwuest

PR #1206, opened 2017-07-19
Use the remove_dir_all crate instead of custom impl.

### LINKED ISSUE #1204: Replace rustup_utils::remove_dir_all with remove_dir_all crate
The remove_dir_all crate does the same thing. Remove the module from rustup_utils and replace uses with the external crate.

cc @Aaronepower

robert-goss:
Fixes #1204 by removing the remove_dir_all module and instead using an external crate for the one place it is called.


alexcrichton:
Oh gah apologies! I rebased and fixed up https://github.com/rust-lang-nursery/rustup.rs/pull/1207 before I saw this one, sorry about that!

For now I'm gonna close this in favor of https://github.com/rust-lang-nursery/rustup.rs/pull/1207, but thanks regardless for the PR here, and sorry for the delay!

robert-goss:
Thanks for the reply. Don't worry I should have been more pro-active with this.

PR #1211, opened 2017-07-21
Fix and re-enable MSI build

Boddlnagg:
This supersedes #661, though it does not enable deployment yet.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,9 +5,8 @@ environment:
   - TARGET: x86_64-pc-windows-msvc
     ALLOW_PR: 1
   - TARGET: i686-pc-windows-msvc
-  # FIXME
-  #- TARGET: i686-pc-windows-msvc
-  #  BUILD_MSI: 1
+  - TARGET: i686-pc-windows-msvc
+    BUILD_MSI: 1
   - TARGET: i686-pc-windows-gnu
     MINGW_DIR: mingw32
   - TARGET: x86_64-pc-windows-gnu
@@ -72,15 +71,8 @@ test_script:
   - cargo build --release --target %TARGET% %FEATURES%
   - cargo test --release -p rustup-dist --target %TARGET%
   - cargo test --release --target %TARGET% %FEATURES%
-  - ps: |
-      if($env:BUILD_MSI) {
-        cd src\rustup-win-installer
-        cargo build --release --target $env:TARGET
-        cd msi
-        .\build.ps1 -Target $env:TARGET
-        cd ..\..\..
-        if($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode) }
-      }
+  - if defined BUILD_MSI pushd src\rustup-win-installer && cargo build --release --target %TARGET% & popd
+  - if defined BUILD_MSI pushd src\rustup-win-installer\msi && powershell .\build.ps1 -Target %TARGET% & popd
 
 notifications:
   - provider: Webhook
--- a/src/rustup-win-installer/build.rs
+++ b/src/rustup-win-installer/build.rs
@@ -1,6 +1,7 @@
 extern crate gcc;
 
 use std::env;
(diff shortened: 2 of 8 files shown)


Boddlnagg:
I want to see if this works on AppVeyor, but there seems to be a problem with AppVeyor right now ...

Boddlnagg:
Looks like this works. I will remove the last two commits as soon as this is reviewed.

@brson @Diggsey I have some more changes for the MSI installer ready, but some questions will need to be discussed in the process as to the general direction of the MSI installer.

alexcrichton:
Thanks for this @Boddlnagg and sorry for the delay! It looks like this is just a few fixes here and there, so if you'd like to rebase and remove the temporary commits, I can r+ to land this

Boddlnagg:
@alexcrichton Thanks! I did the rebase.

appveyor.yml
```diff
@@ -5,9 +5,8 @@ environment:
   - TARGET: x86_64-pc-windows-msvc
     ALLOW_PR: 1
   - TARGET: i686-pc-windows-msvc
```
alexcrichton, line 7: Should this be removed now that BUILD_MSI is enabled?

alexcrichton (commented):
Thanks! Looks like CI is failing though?

src/rustup-win-installer/build.rs
```diff
@@ -1,13 +1,42 @@
...
+    let path = compiler.path();
+    let vs_version = if path.to_string_lossy().contains("VC\\bin") {
+        println!("cargo:warning=Guessing VS version: VS2015");
+        "VS2015"
+    } else if path.to_string_lossy().contains("VC\\Tools\\MSVC") {
+        println!("cargo:warning=Guessing VS version: VS2017");
+        "VS2017"
+    } else {
+        panic!("Unknown VS version with compiler path {:?}", path);
+    };
```
mattico: Could you use find_vs_version instead? https://docs.rs/gcc/0.3.51/gcc/windows_registry/fn.find_vs_version.html

appveyor.yml
Boddlnagg: The MSI version is nowhere near finished. I want to enable it on CI just so it won't break. Maybe it would make sense to also set `ALLOW_PR: 1` for it.

Boddlnagg:
The CI failure is weird: `error: failed to parse lock file at: C:\projects\rustup-rs\Cargo.lock`. Running with `--verbose` gives
```
Caused by:
  package `gcc 0.3.50 (registry+https://github.com/rust-lang/crates.io-index)` is specified as a dependency, but is missing from the package list
```

Other crates in the same workspace use `gcc 0.3.51` and manually changing the version in the lockfile makes the error disappear (I'm adding that to the PR). But I don't really understand what's going on. Is this expected behavior for cargo?

Edit: Ah, I see, it's missing a `checksum` line for `gcc 0.3.50`, I guess that's why it didn't work?

src/rustup-win-installer/build.rs
Boddlnagg: Thanks! I hadn't seen that function.

appveyor.yml
alexcrichton: Ah ok, that makes sense and no worries! My only worry is that this'll conflict with the other i686 builder for the dist artifacts being produced. Enabling the MSI installer just builds more artifacts, right? It shouldn't change the existing binaries?

(if so we can probably just gate on the msi build)

Boddlnagg: No artifacts are being produced for MSI currently, see https://github.com/rust-lang-nursery/rustup.rs/blob/ac5bffad7ba126b2313692c3dfce83ec91cb7fea/ci/prepare-deploy-appveyor.ps1#L9-L12

Changing that situation was part of #661, but I didn't include it here. Will do in another PR.

alexcrichton:
Oh nice! In that case...

@bors: r+

PR #1213, opened 2017-07-24
Use clap subcommand aliases

mgeisler:
Clap has support for subcommand aliases: the aliased subcommand
behaves like the main subcommand in all respects and our code only
needs to dispatch on the name of the main subcommand. See:

  https://docs.rs/clap/2.18.0/clap/struct.App.html#method.alias

The two two-level synonyms (install and uninstall) can not be handled
in this way since they dispatch to nested subcommands -- aliases can
only be made to subcommands on the same level in the hierarchy.


alexcrichton:
@bors: r+

Thanks!

PR #1222, opened 2017-08-03
Add reqwest download backend, remove hyper backends

### LINKED ISSUE #1181: Teach hyper/rustls HTTP backends to resume partial downloads

### LINKED ISSUE #1183: Add a reqwest backend to rustup
rustup currently supports a robust collection of HTTP backends: curl, hyper + native-tls and hyper + rustls. Add a fourth based on reqwest, making sure it has proxy and resume support. This will probably become the default backend.

It should be a matter of following the example in src/download, and adding `RUSTUP_USE_REQWEST` to src/utils/src/utils.rs.

cc https://github.com/rust-lang-nursery/rustup.rs/issues/1181

mattico:
Fixes #1183 
Closes #1181 (hyper backend will no longer exist)

About the lockfile changes:
I had to update native_tls from 1.0.0 to 1.0.4 for reqwest.
`cargo update -p native-tls` complained about not having a valid version of security-framework-sys.
`cargo update -p security-framework-sys` complained about not having a valid version of native-tls.
So I updated everything 🤷‍♂️  .


mattico:
Ah, doesn't work with file URLs. One moment.

mattico:
Fallout from updates:
MIPS/PPC: backtrace build failed (need to install powerpc-unknown-linux-binutils ?)
Android: net2 build failed (https://github.com/rust-lang-nursery/net2-rs/issues/63)

mattico:
https://github.com/rust-lang-nursery/net2-rs/pull/64
x86_64-linux-android now fails with the same error as the rest. I think the issue is with the docker image (https://github.com/rust-lang/rust-buildbot/tree/master/slaves/linux-cross). The configure script is picking up unprefixed objcopy rather than using the cross one.

mattico:
Rebased.

mattico:
Found a less-hacky way to fix the backtrace-sys build: just use env variables. Now to wait for travis.

mattico:
Hoping the rest of the build works\* , but we're going to hit https://github.com/sfackler/rust-security-framework/issues/43 now.

\* confused why it's not running, the opensource linux backlog is zero right now.

mattico:
Same issue with mio as with net2.

mattico:
I think that was a spurious failure.

alexcrichton:
Thanks for this, and sorry for the churn! If you'd like to rebase one last time I think we should be ready to go in terms of review and looking this over.

mattico:
No worries, @alexcrichton!

Going to try without the objcopy workaround first, since the build image changes may have fixed that issue.

src/download/src/lib.rs
```diff
@@ -690,30 +364,17 @@ pub mod curl {
...
-                    _callback: &Fn(Event) -> Result<()> )
-                    -> Result<()> {
-        Err(ErrorKind::BackendUnavailable("hyper").into())
-    }
-}
-
-#[cfg(not(feature = "rustls-backend"))]
-pub mod rustls {
+#[cfg(not(feature = "reqwest-backend"))]
+pub mod reqwest {
```
alexcrichton: Should this be `reqwest_be` or should the above be `reqwest`?

src/rustup-utils/src/utils.rs
```diff
@@ -221,12 +221,9 @@ fn download_file_(url: &Url,
     };
 
     // Download the file
-    let use_hyper_backend = env::var_os("RUSTUP_USE_HYPER").is_some();
```
alexcrichton: Mind leaving this around for a bit as an alias for using reqwest?

alexcrichton:
@bors: r+

Looks great, thanks!

mattico:
`Cargo.lock` is fun... rebased.

PR #1224, opened 2017-08-07
Clarify Windows build preferences in README.md.

mikebenfield:
Given the earlier dicussion of MSVC toolchain vs GNU toolchain on Windows, for someone not perfectly familiar with the terminology here, the current placement of the sentence "In general, this is the build of rustup one should install on Windows." confusingly may seem to imply that the *GNU toolchain* is preferred over the *MSVC toolchain*. I would request that the sentences be rearranged as in this pull request to clarify that *build* has nothing to do with which toolchain.

(Or possibly remove the sentence in question entirely? Given that the reader is being directed to install this particular build, does it make sense to also say that this build is the one that should be installed?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,13 +548,13 @@ The primary installation method, as described at
 [www.rustup.rs](https://www.rustup.rs), differs by platform:
 
 * On Windows, download and run the [rustup-init.exe built for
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #1230, opened 2017-08-11
Compile less executables in tests

alexcrichton:
Refactor a bit with more aggressive caching to compile at most one Rust program when testing.


ADAMSALAMA6667 (approved):
@rustbot 3e99a693a3b811bb1d8098dfa254f1bb543b6e05

PR #1233, opened 2017-08-14
Deploy MSI build artifacts

Boddlnagg:
This enables deployment of build artifacts for the MSI build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -93,6 +93,10 @@ artifacts:
     name: rustup-setup
   - path: dist\$(TARGET)\rustup-setup.exe.sha256
     name: rustup-setup-sha
+  - path: dist\$(TARGET)\rustup.msi
+    name: rustup-msi
+  - path: dist\$(TARGET)\rustup.msi.sha256
+    name: rustup-msi-sha
 
 deploy:
   - provider: S3
@@ -103,7 +107,7 @@ deploy:
     bucket: dev-static-rust-lang-org
     set_public: true
     region: us-west-1
-    artifact: rustup-init,rustup-init-sha,rustup-setup,rustup-setup-sha
+    artifact: rustup-init,rustup-init-sha,rustup-setup,rustup-setup-sha,rustup-msi,rustup-msi-sha
     folder: rustup
     on:
       branch: stable
--- a/ci/prepare-deploy-appveyor.ps1
+++ b/ci/prepare-deploy-appveyor.ps1
@@ -6,23 +6,30 @@ if ($env:APPVEYOR_REPO_BRANCH -eq "auto") {
    exit 0
 }
 
-# Don't do anything for MSI (yet)
-if ($env:BUILD_MSI) {
-   exit 0
-}
-
-# Copy rustup-init to rustup-setup for backwards compatibility
-cp target\${env:TARGET}\release\rustup-init.exe target\${env:TARGET}release\rustup-setup.exe
-
-# Generate hashes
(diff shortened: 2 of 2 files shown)


Boddlnagg:
Questions to be resolved before merging:
- Is it okay to enable this already even though the MSI version is not yet functional and shouldn't be used?
- Should the SHA256 be created for the .msi itself or for the embedded .exe? What is it being used for? Originally I thought (and maybe that was the case in the past) that it's used to check for updates (i.e., `self update` checks whether the hash of the currently executed binary is different from the hash of the version on the server), but the `self update` code now looks only at the `release-stable.toml` metadata file.

alexcrichton:
Thanks for this! (and sorry again for the delay!)

I think this is fine to deploy these artifacts for now, we may just not use them yet. Regardless they can serve as a good testing ground.

Right now we typically have `foo` and `foo.sha256` for checksum verification, so could the checksum be of the MSI installer rather than the exe inside?

Boddlnagg:
Thanks for the review! The hash is now derived from `rustup.msi` itself.

alexcrichton:
@bors: r-

Er actually, mind combining the i686-pc-windows-msvc builds? We ideally would just have one build which also creates MSI installers I think?

Boddlnagg:
@alexcrichton The build of `rustup.exe` itself is different, because the one built for the MSI is built with the `msi-installed` Cargo feature enabled. That's why it has been two separate builds so far. Or maybe I don't understand what you think we should do?

alexcrichton:
Ah hm ok I see. I sort of expected the MSI artifacts to just be an extra thing compiled on the side and the actual rustup executable wouldn't change at all. I figured the manifest would say whether the MSI is preferred or not?

Boddlnagg:
This should work if there's some kind of runtime configuration that tells `rustup.exe` that it has been installed via MSI and should use MSI for update and uninstall. I have to look at what exactly the "manifest" is.

Boddlnagg:
Since the MSI installer writes its install location to a registry entry (to find it on uninstall, see https://github.com/rust-lang-nursery/rustup.rs/issues/662#issuecomment-241226010), we could also check that registry key to know whether the rustup executable that's currently running has been installed by the MSI.

alexcrichton:
Ah yeah sorry that's what I figured was going to happen here, that rustup, at runtime, would select whether to install/uninstall via the MSI. I feel like we'd have some compatibility difficulties if we had the "MSI capable exe" and the "only exe capable exe".

Would it be difficult, though, to only have one build?

Boddlnagg:
I don't think it would be very difficult. We'd only have to decide on how to find out at runtime if the MSI installer is being used (the manifest? the Windows registry?) and someone has to implement it (I won't have time to do that right now). 

You can search the code for `msi-installed` to find all the locations where the feature flag is currently being used, which should be switched to runtime checks

alexcrichton:
Ok thanks for the info! I imagine that we'll likely encode this into the manifests once we get around to it.

Boddlnagg:
I don't really understand what you mean by manifest. If I read [this](https://github.com/rust-lang-nursery/rustup.rs/blob/07fd5bfd0ccc70fd2bb3d90f193cf4b5efe011f0/src/rustup-dist/src/manifest.rs#L3-L9), I get the impression that they are about configuration for individual installable tools/toolchains, and would not be the correct place for configuration of the `rustup` installation itself. But it might well be that I'm missing something.

alexcrichton:
Oh yeah I believe that's the manifest I'm thinking of (the channel manifests). My guess is that because MSI is "universally better" we'd just have rustup make the transition as soon as it seems MSIs in the manifest.

Diggsey:

Issue #253, opened 2016-04-02
Add windows gui / msi installer


brson:
To complete the Rust installation experience on Windows we want to be installing rustup via an msi.

Make a proof-of-concept rustup msi installer that embeds libmultirust. The behavior of this msi will be heavily customized - all it does is the standard rustup install, but presented in a windowsy way. To start with it can be really simple:

* On installation, present a screen that says more-or-less what the console installer says now. Have install/cancel buttons.
* After they press install, go to another window that displays whatever status updates we reasonably can as the install is progressing.
* After install show another screen that indicates success.
* Register the uninstaller with windows so it works from the add/remove programs screen.

This will require a lot of refactoring of the existing install code to get it embedded in this new context.

Our options for GUI's will be limited but we can't use something heavy. I'm thinking either something rust-centric like conrod, or just a very thin win32 wrapper.

retep998:
Depending on OpenGL for Conrod might not work very well if they don't have the GPU manufacturer drivers that provide OpenGL, or if they're using Windows in a VM.

Diggsey:
We could just use Inno Setup... It will allow us to produce a GUI installer which can call out to the command line rustup behind the scenes, and it's configured declaratively, so we won't even need any non-rust code.

brson:
@Diggsey Inno Setup does not produce msi's (I believe), and msi's are the delivery format people most expect and prefer.

Rust itself uses Inno to produce .exe installers but we don't advertise them because the .msi's are nicer.

retep998:
If we use the [Windows Installer](https://msdn.microsoft.com/en-us/library/windows/desktop/cc185688%28v=vs.85%29.aspx) stuff then that can provide a [user interface for us](https://msdn.microsoft.com/en-us/library/windows/desktop/aa367445%28v=vs.85%29.aspx).

genbattle:
If the maintainers would prefer to depend on other open source projects then [Wix](http://wixtoolset.org/) is probably the best bet (for MSI packages at least). Advanced installer is a pretty nice WYSIWYG solution I've used before, but their freeware version is fairly basic.

Building your own MSI by hand is pretty gritty.

pravic:
@brson 

> and msi's are the delivery format people most expect and prefer.
.msi is a really big pain due to windows installer cache bloat :(

vadimcn:
[...]

PR #1234, opened 2017-08-14
More work on MSI installer

Boddlnagg:
Actually implements a large part of the MSI logic (previously this was mostly stubs).

With these changes, the MSI installer will install rustup to the right location, so running the installer without setting `CARGO_HOME` will overwrite the existing installation!

A test script is included that installs rustup into a different location by setting `CARGO_HOME` (and `RUSTUP_HOME`).

This PR also enables building/testing the MSI version on AppVeyor CI (but the tests for install and self update are disabled, because they are not compatible with the MSI version, and no MSI specific tests have been added yet).

### DIFF
--- a/src/rustup-win-installer/msi/rustup.wxs
+++ b/src/rustup-win-installer/msi/rustup.wxs
@@ -8,6 +8,10 @@
     <Product Id="*" Name="rustup" Language="1033" Version="$(env.CFG_VER_MAJOR).$(env.CFG_VER_MINOR).$(env.CFG_VER_PATCH).0" Manufacturer="The Rust Project Developers" UpgradeCode="09acbb1c-7123-44ac-b2a9-4a04b28ced11">
         <Package InstallerVersion="200" Compressed="yes" InstallScope="perUser" />
 
+        <Condition Message="The Rust toolchain is only supported on Windows 7, Windows Server 2008 R2, or higher.">
+            <![CDATA[Installed OR (VersionNT >= 601)]]>
+        </Condition>
+
         <!-- TODO: How to configure updates? `AllowDowngrades` automatically removes previously installed versions, no matter what version they have -->
         <MajorUpgrade AllowDowngrades="yes" />
         
--- a/src/rustup-win-installer/msi/rustup.wxs
+++ b/src/rustup-win-installer/msi/rustup.wxs
@@ -14,8 +14,14 @@
         <!-- Specifies a single cab file to be embedded in the installer's .msi. -->
         <MediaTemplate EmbedCab="yes" CompressionLevel="high" />
 
-        <Feature Id="ProductFeature" Title="rustup" Level="1">
-            <ComponentRef Id="CompleteInstallation" />
+        <Feature Id="Rustup" Title="rustup" Absent="disallow" Display="1" AllowAdvertise="no">
+            <ComponentRef Id="RustupExe" />
+            <ComponentRef Id="RegisterProductCode" />
+            <ComponentRef Id="RegisterInstallDir" />
+        </Feature>
+        <!-- TODO: Add UI to actually let the user disable this Feature -->
+        <Feature Id="Path" Title="Add to PATH" Absent="allow" Description="Add Rust to PATH environment variable" Display="2" AllowAdvertise="no">
+            <ComponentRef Id="ModifyPathEnv" />
         </Feature>
         
         <!-- Set some metadata that will appear in the "Installed Programs" list -->
@@ -35,8 +41,6 @@
         <UIRef Id="CustomUI" />
         <WixVariable Id="WixUIDialogBmp" Value="dialogbg.bmp" />
         <WixVariable Id="WixUIBannerBmp" Value="banner.bmp" />
-        <!-- TODO: Include/generate license file -->
-        <!--<WixVariable Id="WixUILicenseRtf" Value="LICENSE.rtf" />-->
 
         <Directory Id="TARGETDIR" Name="SourceDir">
             <!-- `INSTALLLOCATION` will be set by custom action -->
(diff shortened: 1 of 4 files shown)


Boddlnagg:
A potential issue with the current approach is that the logic to determine the install location (which is being extracted into a separate crate as per https://github.com/rust-lang-nursery/rustup.rs/pull/1205) is now duplicated here in the MSI (using MSI actions instead of calling Rust code). It could be delegated to Rust code in a custom action, but in general we should aim to use custom actions as sparingly as possible where MSI provides equivalent functionality natively. I do understand that this makes maintenance harder, so this is something that needs to be discussed, not only w.r.t. this PR, but future work on the MSI in general.

alexcrichton:
Thanks for this! Sorry I'm pretty out-of-the-loop with respect to the MSI installer. Mind filling in the details for me a bit?

My impression was that the MSI installer was mostly for a better installation experience on Windows (e.g. the first time you use it) and then I believe for uninstallation as well (it shows up in the "Uninstall Programs" standard thing on Windows). Is that right? If so, is the thinking here that we'll transition to the MSI installer by default for everyone that already exists to ensure that there's a uniform uninstallation experience?

Boddlnagg:
> My impression was that the MSI installer was mostly for a better installation experience on Windows (e.g. the first time you use it) and then I believe for uninstallation as well (it shows up in the "Uninstall Programs" standard thing on Windows). Is that right?

Yes. Some discussion (with a few issues still needing to be figured out) happened in #253.

> If so, is the thinking here that we'll transition to the MSI installer by default for everyone that already exists to ensure that there's a uniform uninstallation experience?

At least that would be my thinking, yes. So at some point, we would enable a transition path from the non-MSI `rustup` to the MSI-based `rustup`, but this does not yet exist.

> Mind filling in the details for me a bit?

As you say, the goal is a better (un-)installation experience on Windows. MSI provides this, but is probably not the only way to achieve it (a custom GUI which registers the appropriate registry entries for the system-wide "Installed Programs" dialog would also work). MSI also has the advantage that it runs within a system service, so it won't suffer from the problems around deleting files that are in-use by Windows that lead to [this monster](https://github.com/rust-lang-nursery/rustup.rs/blob/b33d20a91dc189d99f3e21e3a4762d07c57801da/src/rustup-cli/self_update.rs#L814-L1012).

The "problem" with MSI is that it's a database format which is meant to be used mainly to install files and settings. So what I did (after discussing it with @brson), is to turn rustup into a DLL and use that from the MSI in the form of "custom actions":
https://github.com/rust-lang-nursery/rustup.rs/blob/b33d20a91dc189d99f3e21e3a4762d07c57801da/src/rustup-win-installer/Cargo.toml#L9

https://github.com/rust-lang-nursery/rustup.rs/blob/b33d20a91dc189d99f3e21e3a4762d07c57801da/src/rustup-win-installer/msi/rustup.wxs#L5

"Custom actions" are exported DLL functions which are called while the MSI proceeds, so we can theoretically call all the existing rustup installation logic (written in Rust) from the MSI. However, the "right"/idiomatic way to write MSIs is to have as few custom actions as possible, and use them only for very specific, small tasks, because writing correct custom actions (that e.g. also work on [rollback](https://msdn.microsoft.com/en-us/library/windows/desktop/aa371369%28v=vs.85%29.aspx)) is hard (this is why I disabled rollback completely for this MSI). MSI built-in actions are already quite powerful and definitely more robust than what we can come up with. So most of the installation logic can be written in MSI/WIX itself, e.g. adding the rustup directory to the path in a robust way is [trivial](https://github.com/rust-lang-nursery/rustup.rs/blob/b33d20a91dc189d99f3e21e3a4762d07c57801da/src/rustup-win-installer/msi/rustup.wxs#L57) (which theoretically lets us get rid of a lot of Windows-specific code in the rustup self-install code). But duplicating the logic in WIX is a maintenance burden of course ...

One thing which is currently being done by Rust code, but should (IMHO) be done by MSI built-in actions instead is the "hard-linking" of the installed binaries. There is no built-in action for hard-linking in MSI, only for duplicating files, but the problem is that hard-linking either requires admin privileges or [only works on NTFS](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365006(v=vs.85).aspx) (it's currently implemented in that way ... did anyone ever stumple upon that?). However, considering that the Rust toolchain needs plenty of disk space anyway, I wonder if it's really not a better way to just duplicate the files (brson was against that when I brought it up before).

Lastly, I want to add a personal thing: I'm probably the only one right now working on this, and I don't know if there is anyone else familiar with MSI/WIX in the Rust community. The problem is that unfortunately I won't be able to contribute much more in the future (besides the PRs that I have opened already) because I have other obligations.

Boddlnagg:
Another thing: The largest piece that is still missing after this PR is the actual setup of the initial toolchain (including the GUI that lets you select which toolchain(s) you want).
Currently the MSI will only setup the directories and install the binaries, but won't put anything inside `.rustup`. The easiest way to do that (in a first step) is probably to launch some `rustup install` command in a user-visible console window after the main installation has finished. Later it should be part of the main installation, not shoing a console window, but instead report the correct progress and status to the installer progress bar.

Boddlnagg:
And I just stumbled upon https://github.com/rust-lang-nursery/rustup.rs/issues/662 which is also relevant to the MSI.

alexcrichton:
Ok thanks for the explanation! And no worries if you've got other obligations now, thanks regardless for all your work up to this point!

This seems like a totally OK PR to land in the meantime as it's mostly just touching the installer. Could the `ALLOW_PR` bit be removed and then I think we could land this?

Boddlnagg:
Sure. I've now removed the commit concerning `ALLOW_PR`.

PR #1237, opened 2017-08-17
Renames

nrc:
This adds a `renames` section to the manifest. The idea is that we will add a rename from `rls` to `rls-preview` at the same time as renaming the rls component to rls-preview. This new feature will cause all existing users of the rls to have that component renamed to rls-preview. When the rls hits 1.0 we'll change the rename from `rls-preview` to `rls` and then everyone who already has the rls will be moved to the 1.0 name. Likewise for rustfmt, etc.

r? @alexcrichton

### DIFF
--- a/src/rustup-dist/src/manifest.rs
+++ b/src/rustup-dist/src/manifest.rs
@@ -25,6 +25,7 @@ pub struct Manifest {
     pub manifest_version: String,
     pub date: String,
     pub packages: HashMap<String, Package>,
+    pub renames: HashMap<String, String>,
 }
 
 #[derive(Clone, Debug, PartialEq)]
@@ -82,7 +83,8 @@ impl Manifest {
         Ok(Manifest {
             manifest_version: version,
             date: try!(get_string(&mut table, "date", path)),
-            packages: try!(Self::table_to_packages(table, path)),
+            packages: try!(Self::table_to_packages(&mut table, path)),
+            renames: try!(Self::table_to_renames(table, path)),
         })
     }
     pub fn to_toml(self) -> toml::Table {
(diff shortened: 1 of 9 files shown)


src/rustup-dist/src/manifest.rs
```diff
@@ -118,6 +123,28 @@ impl Manifest {
         result
     }
 
+    fn table_to_renames(mut table: toml::Table, path: &str) -> Result<HashMap<String, String>> {
+        let mut result = HashMap::new();
+        let rename_table = try!(get_table(&mut table, "rename", path));
```
alexcrichton, line 128: I think this means that this'll fail if `rename` isn't present, right?

src/rustup-dist/src/manifest.rs
```diff
@@ -118,6 +123,28 @@ impl Manifest {
...
+
+        for (k, v) in rename_table {
+            if let toml::Value::Table(mut t) = v {
+                result.insert(k.to_owned(), get_string(&mut t, "to", path)?);
```
alexcrichton, line 132: How come the level of indirection here? Why not:

```toml
[rename]
cargo-old = 'cargo'
```

(etc)

alexcrichton:
Looking good! Could this also add a `tests/*.rs` style integration test which runs through the `rustup` commands themselves to ensure that all works?

The two cases I'm thinking about are:

* First a manifest only has `rls-preview`, you install it, then we publish a manifest with a `[rename]` that *doesn't* have `rls-preview`, and your `rustup update` should work.
* Second a manifest has a `[rename]` and adding the renamed component should work

src/rustup-dist/src/manifest.rs
nrc: No, if there is no table `get_table` returns `Ok(toml::Table::new())`

src/rustup-dist/src/manifest.rs
nrc: I was basically following the convention for `pkg`, I also thought it was a little bit more clear, but I'm not attached to the current format, if you'd prefer to change.

src/rustup-dist/src/manifest.rs
alexcrichton: Ok great!

src/rustup-dist/src/manifest.rs
alexcrichton: Ok sounds good to me!

nrc:
Tests added!

PR #1238, opened 2017-08-20
improve reqwest backend

seanmonstar:
- use a single Client, yay connection pooling
- fix proxies (scheme was missing)
- fix reqwest partial downloads
- add tests (including partial downloads)
- add warning about RUSTUP_USE_HYPER var

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -233,7 +233,10 @@ dependencies = [
  "curl 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_proxy 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "reqwest 0.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "futures 0.1.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hyper 0.11.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


src/download/tests/download-curl-resume.rs
```diff
@@ -62,36 +36,48 @@ fn partially_downloaded_file_gets_resumed_from_byte_offset() {
...
+    let callback_partial = Mutex::new(false);
+    let callback_len = Mutex::new(None);
+    let received_in_callback = Mutex::new(Vec::new());
 
     download_to_path_with_backend(Backend::Curl,
                                   &from_url,
                                   &target_path,
                                   true,
                                   Some(&|msg| {
+        println!("event: {:?}", msg);
```
alexcrichton: Stray println?

alexcrichton:
Looks great to me, thanks!

src/download/tests/download-curl-resume.rs
seanmonstar: Woops, yep. Used to learn that I needed to add a `Seek` to the reqwest backend file downloader.

seanmonstar:
I noticed in using this that it would warn about the environment variable several times a run, as it tried to download several files. So I've added a static bool to only emit the warning once...

seanmonstar:
I had removed a `[replace]` for mio, as it was warning me that it wasn't needed. I'm not sure why, but I've added it back, to fix those tests on Android.

seanmonstar:
I had included downgrades in the `Cargo.lock` that I didn't even notice. I've gone through and fixed every single one, besides the absolute necessary to needed in this PR.

alexcrichton:
Tests passed, gonna merge manually.

PR #1242, opened 2017-08-22
Use `exec` on Unix to propagate signals

### LINKED ISSUE #31: Use exec() to delegate to the real cargo/rustc on Unix

### LINKED ISSUE #806: Ctrl+C on `cargo run`

alexcrichton:
This is only possible when telemetry is disabled, but almost all of the time it
is! This should help fix lots of common rustup-related issues related to signals
and such.

Closes #31
Closes #806

### DIFF
--- a/src/rustup-cli/proxy_mode.rs
+++ b/src/rustup-cli/proxy_mode.rs
@@ -53,4 +53,3 @@ fn direct_proxy(cfg: &Cfg, arg0: &str, toolchain: Option<&str>, args: &[OsString
     };
     Ok(try!(run_command_for_dir(cmd, arg0, args, &cfg)))
 }
-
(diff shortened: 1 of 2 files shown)


geofft:
Oh yay thank you!!

PR #1250, opened 2017-09-12
Automatically install override toolchain when missing.

### LINKED ISSUE #1218: Add a way to install the toolchain specified in the override file
https://github.com/rust-lang-nursery/rustup.rs/pull/1172 allows us to create a `rust-toolchain` file and specify a version of the toolchain to be used in the current directory. But I can't seem to find a way to install the specified version automatically. 
Now I do `cat rust-toolchain | xargs rustup install`, but it would be nice to have a command like `rustup install rust-toolchain` or maybe simply `rustup install` (fail if the file is not present). 

Thoughts? @brson

SimonSapin:
A typical scenario is:

* I work on a repository that uses `rust-toolchain` to pin to a specific Nightly version
* I run `git pull`, `rust-toolchain` has been changed to update to a new Rust version
* I run `cargo build`

Result before this PR (typically): rustup fails with an error like:

```
error: override toolchain 'nightly-2017-08-31' is not installed
info: caused by: the toolchain file at '/home/simon/projects/servo/rust-toolchain' specifies an uninstalled toolchain
```

A better result would be to automatically install toolchains as needed.

Closes #1218

### DIFF
--- a/src/rustup/config.rs
+++ b/src/rustup/config.rs
@@ -259,16 +259,20 @@ impl Cfg {
                 }
             };
 
-            match self.verify_toolchain(&name) {
-                Ok(t) => {
-                    Ok(Some((t, reason)))
+            match self.get_toolchain(&name, false) {
+                Ok(toolchain) => {
+                    if toolchain.exists() {
+                        Ok(Some((toolchain, reason)))
+                    } else if toolchain.is_custom() {
+                        // Strip the confusing NotADirectory error and only mention that the override
+                        // toolchain is not installed.
+                        Err(Error::from(reason_err))
+                            .chain_err(|| ErrorKind::OverrideToolchainNotInstalled(name.to_string()))
+                    } else {
+                        try!(toolchain.install_from_dist());
+                        Ok(Some((toolchain, reason)))
+                    }
                 }
-                Err(Error(ErrorKind::Utils(::rustup_utils::ErrorKind::NotADirectory { .. }), _)) => {
-                    // Strip the confusing NotADirectory error and only mention that the override
-                    // toolchain is not installed.
-                    Err(Error::from(reason_err))
-                        .chain_err(|| ErrorKind::OverrideToolchainNotInstalled(name.to_string()))
-                },
                 Err(e) => {
                     Err(e)
                         .chain_err(|| Error::from(reason_err))
--- a/tests/cli-rustup.rs
+++ b/tests/cli-rustup.rs
@@ -4,7 +4,6 @@ extern crate rustup_dist;
 extern crate rustup_utils;
 extern crate rustup_mock;
 extern crate tempdir;
-extern crate regex;
 
 use std::fs;
 use std::env::consts::EXE_SUFFIX;
@@ -16,7 +15,6 @@ use rustup_mock::clitools::{self, Config, Scenario,
                             expect_err,
                             set_current_dist_date,
                             this_host_triple};
-use regex::Regex;
 
 macro_rules! for_host { ($s: expr) => (&format!($s, this_host_triple())) }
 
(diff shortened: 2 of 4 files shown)


aturon:
cc @nrc @Diggsey @alexcrichton @withoutboats -- any of you up for reviewing?

Diggsey:
I'm not that comfortable having it automatically install the new toolchain, I can imagine that would be unacceptable to some people. I can see a few alternatives:
1) Prompt before installing the missing toolchain, overridable with `-y` or similar.
2) Fail as before, but print out the command needed to install the missing toolchain.
3) Have it be a global config option, "automatically install missing toolchains"

Would one or more of those work for you?

@alexcrichton thoughts?

SimonSapin:
Servo tries to keep up with Rust nightly, which means upgrading several times a month. I think it is important that this is as seamless as possible for regular contributors. Pulling from `master` and rebuilding should Just Work.

So I think option 2 is not particularly useful.

Option 1 seems tricky for two reasons: this code path is often used in non-interactive environments (e.g. Travis-CI) where prompting would just hang. As to the command line argument, I don’t know where it should go for proxy executables like `cargo`.

Option 3 sounds like that config should be owned by the user and Servo’s build system should not mess with it.

What do you think of an environment variable?

SimonSapin:
Should all this apply only to `rust-toolchain` (whose "source of truth" is often an upstream repository) but not other forms of overrides (which are controlled by the local user)?

nrc:
Given that the behaviour is only triggered by building a project with a toolchain file, it seems to me that the user has already opted-in. I don't think this is difference from Cargo downloading a crate for a build. At most we should get the user's consent once per project (e.g., the first time you build Servo, rustup asks if it should download a toolchains, user can then opt-in to auto-downloads or get errors, but I don't think even that is necessary.

SimonSapin:
I agree that this is similar to automatically downloading dependency crates.

@Diggsey, what do you think of that point of view, or of the environment variable approach?

SimonSapin:
CC @alexcrichton

alexcrichton:
I personally agree with @nrc that if this is only used in conjunction with `rustup.toml` then it seems reasonable to automatically download the appropriate toolchain rather than requiring an explicit update step.

In reading this, though, I had a few questions:

* Does this only trigger with `rustup.toml`? In reading the tests it looks like this happens regardless of the presence of a `rustup.toml`.
* Could a test be added explicity for this behavior?

SimonSapin:
What is `rustup.toml`? It is not mentioned in https://github.com/rust-lang-nursery/rustup.rs/blob/master/README.md

This PR is relevant when:

* A `rust-toolchain` file exists in the current directory or an ancestor,
* Or the `rustup override` command has been used to set up an override for the current directory or an ancestor,
* Or the `RUSTUP_TOOLCHAIN` environment variable is set.

Each of these cases end up going through the same code path.

alexcrichton:
@SimonSapin oh sorry, I meant the `rust-toolchain` file you pointed out to me on IRC.

It sounds like though it's already working! Sounds good to me that we'd also include overrides and the env var.

Can you add a test (or maybe there already is one?) about `rustc +beta` returning an error if the beta toolchain isn't installed? (with no other override/toolchain file/etc business). Could you also add a test which explicitly includes the `rust-toolchain` file and auto-install behavior?

SimonSapin:
I’ve pushed a new commit with the requested tests.

alexcrichton:
@bors: r+

Thanks!

SimonSapin:
Could we get a new release with this change? Can I help?

alexcrichton:
Sorry I've been slow to respond, I will try to get to a release when I get a chance

PR #1255, opened 2017-09-20
Improve clarity of "adding req. component" error

### LINKED ISSUE #1251: Confusing error when attempting to install `rust-docs` component
Ran the command `curl https://sh.rustup.rs -sSf | sh` to install rustup, then `rustup component add rust-src` which worked, then `rustup component add rust-docs` which fails with the following error:

```
error: component 'rust-docs' for target 'x86_64-apple-darwin' is required for toolchain 'stable-x86_64-apple-darwin' and cannot be re-added
```

Which is not google-able and doesn't seem (to me) to provide any clues on how to resolve the issue... any hints?

silverjam:
Fixes #1251.

### DIFF
--- a/src/rustup/errors.rs
+++ b/src/rustup/errors.rs
@@ -50,7 +50,7 @@ error_chain! {
         }
         AddingRequiredComponent(t: String, c: Component) {
             description("required component cannot be added")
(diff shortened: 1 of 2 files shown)


alexcrichton:
@bors: r+

Thanks!

PR #1257, opened 2017-09-22
Support `rustup-init --default-toolchain none`

petrochenkov:
To address this use case https://github.com/rust-lang-nursery/rustup.rs/issues/309#issuecomment-213839631.

This is especially useful when user directories have quotas into which large components (e.g. docs) do not fit.


petrochenkov:
ping @Diggsey @alexcrichton

alexcrichton:
@petrochenkov oh oops sorry, thanks for the ping!

Just to clarify, what's the behavior of the rustc/cargo/etc shims in this scenario? If you avoid installing a toolchain, what does `cargo --version` do, for example?

petrochenkov:
@alexcrichton 
```
$ cargo --version

$ rustc --version
```

alexcrichton:
Ok sounds great!

@bors: r+

PR #1261, opened 2017-09-30
Update dependencies

alanhdu:
- Use `toml 0.4` instead of `toml 0.1`
- Use `serde` instead of `rustc-serialize`
- Use `lazy_static 0.2` (over `0.1`)
- Upgrade to `error-chain 0.11` (to silence unused doc comment warnings).


alexcrichton:
@bors: r+

Thanks!

PR #1269, opened 2017-10-22
update shebang to reflect bashisms

mcandre:
Update shebang, given that `local` is not a POSIX semantic. Discovered this when trying to run the rustup installer on Oracle Solaris, where `/bin/sh` is a symlink to ksh93.

Note that the rustup.rs website should also pipe to `bash`, not `sh`.

Alternatively, we can rewrite this script in pure POSIX sh, though I leave that to cooler heads.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Copyright 2016 The Rust Project Developers. See the COPYRIGHT
 # file at the top-level directory of this distribution and at
 # http://rust-lang.org/COPYRIGHT.


mcandre:
Looks like one of the macOS CI builds stalled for this tweak, dunno why?

https://travis-ci.org/rust-lang-nursery/rustup.rs/jobs/290991285

mati865:
If changing to bash using `#!/usr/bin/env bash` should be preferred.

mcandre:
@mati865 Eh, if the shell were ash, dash, or posh, or relied on alpha features from Homebrew bash tip, then `/usr/bin/env` would definitely help to port the script to varied systems. But for common shell languages like bash, sh, ksh, zsh, the /usr/bin/env prefix just slows things down unnecessarily.

Diggsey:
Thanks for the PR!

Issue #1253, opened 2017-09-19
Show alternate install options on website


withoutboats:
https://rustup.rs tries to figure out what OS you are visiting from, and show appropriate instructions on installation. Sometimes, though rustup.rs is wrong - for example, if you are using anti-fingerprinting software, your browser might return an inaccurate response causing us to believe you are running a different OS than you are.

We should support an option on the main page to switch it to display the installation instructions for other operating systems.

jelford:
I could take this; I will have some time this week to put something together. I'm thinking of button at the bottom that says something like:

    I think you're running on $platform, but if not, <here's> the full list.

... which would just switch to `#platform-instructions-default`. It could maybe show text like `#platform-instructions-unknown` (not exactly the same) below?

For what it's worth, as a NoScript user I see the default screen, and I just assumed that's what everyone saw; I never saw anything wrong with it.

jelford:
I didn't get a chance to look at this while I was on holiday, and I don't think I'll be able to get to it now. If someone else wants to pick it up, that would seem sensible.

senden9:
I added a pull request that pick that issue up. See #1300.

PR #1270, opened 2017-10-22
Added a link to all installers to the homepage

JWorthe:
Good afternoon folks. I came across #1253 by chance, and thought it looked like something I could help with. I've added links under the Unix-specific and Windows-specific installation instructions in a smaller font that points to the 'default' set of installation instructions that lists all platforms.

While I was looking at this, I also noticed that the instructions for the 'unknown' OS didn't quite look the same as for the 'default' platform, so I updated some CSS selectors.

I'm open to any suggestions if you want me to do this slightly differently, or change the wording a bit :)


t-botz:
@nrc  Conflict fixed in https://github.com/rust-lang-nursery/rustup.rs/pull/1370
Please close this one.

PR #1271, opened 2017-10-22
[WIP] Add subcommand to check for available updates without updating

### LINKED ISSUE #1249: Add option to check for updates without updating
It'd be nice to have a command to run rustup update and check if anything would be downloaded without actually downloading anything. For example, I have a MOTD that shows me if I need to update any packages, and it'd be nice to also show a message if a new stable version of Rust is available to download.

ordovicia:
This PR adds subcommand to check for available updates for installed toolchains and rustup itself without actually updating.

This PR is WIP because no consensus is built about what name is proper for the subcommand or option.
Currently I have chosen `check` subcommand.
I guess there are many candidates:

* New subcommand
  * `check`
  * `check-update`
* Adding an option to `update` subcommand
  * `--no-download`
  * `--dry`
  * `--dry-run`
* and any other suggestions

Fixes #1249.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,6 +118,7 @@ cache:
 
 branches:
   only:
+    - check-update
     - master
     - stable
     - auto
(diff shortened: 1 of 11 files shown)


Zteve:
I would love this feature; staging updates requires knowing when they are available.

My suggestion for a command structure is a subcommand of its own (like `check` but could be called `query`) which is able to be placed before another subcommand and which announces what _that_ would do, without actually doing it.  Then the call would be `rustup query update`, or `rustup query self update`, or whatever.

This allows the possibility of extending this to other subcommands in future, giving it its own special syntax in the first instance.  Initially, `rustup check` would be an acceptable shorthand for `rustup check update`.

Options, like `--no-download` (too specific) and `--dry-run` (parochial) offer the opportunity for conflict with other options and aren't obvious enough.  Special decorators (like the current `+beta` shorthand) should be avoided, since they, too, are non-obvious.

matthiaskrgr:
Any updates on this? I would also really like to have this :)

ordovicia:
Sorry, I have done nothing recently.
This PR has many conflicts, so it may be somewhat hard to rebase on master.
Feel free to make a progress on this :smile: 

I now personally think that we should consider whether we *really* need this feature.

Diggsey:

PR #1286, opened 2017-11-05
Add a quick troubleshooting tip to the README

mark-i-m:
It took me and IT about a month to figure this out. I thought it was a problem with the version of SSL installed on the squid proxy, whereas they thought it was a problem with my setup. It turned out to be very strange behavior from `curl` and `libcurl`. I thought it might be useful to pass the tip on.


Diggsey:
Thanks for the PR - do you know if this is documented anywhere?

mark-i-m:
> Thanks for the PR - do you know if this is documented anywhere?

Not to my knowledge, but I might have missed it.

Issue #1287, opened 2017-11-06
rustup doc does not use the BROWSER environment variable


naufraghi:
I noted that `rustup doc` has an hardcoded list of commands to fire up a browser, but I discovered from the [`webbrowser.py` sources](https://github.com/python/cpython/blob/3.6/Lib/webbrowser.py#L619) that the user can set the `BROWSER` environment variable to override the command.

I find this very useful, because I usually have more than one open browser and sometime I prefer to open the documentation in a browser that is not the system default one.

I have a mostly working patch for this issue but my rust-foo is poor (very poor when strings come in) and the code is not something I'm very proud of.

If the feature is something someone else is interested in, I can open a PR for further review.

PR #1289, opened 2017-11-06
Use command in BROWSER env if present for `doc` command

naufraghi:
If the BROWSER enviroment variable is found, it is used as a list of
commands to be tested before the hardcoded ones to launch a browser.

Issue #1287


Diggsey:
Thanks for the PR!

MihirLuthra:
@naufraghi, how exactly could I utilize this feature? I tried `export BROWSER=firefox` and then `rustup doc std::iter::Iterator` but it still opens up in my default browser which is safari.

kinnison:
We now use `opener` not this change, as such you should look at how the XDG standard expects you to configure things.

PR #1290, opened 2017-11-07
Display helpful advices even with -y.

mog422:
Currently -y disables any helpful advices, but example usages of curl-based rustup installation frequently include -y (unfortunately).
Since -y is not a silent flag but a yes-to-all flag, we can always display advices instead and (on Windows) only pause unless -y is given.


Diggsey:
Thanks for the PR!

PR #1295, opened 2017-11-21
Add `rustup run --install`

### LINKED ISSUE #1293: Can we rely on proxy mode being based on argv[0]?
We’re considering replacing Servo’s rust/cargo bootstrapping code with rustup.rs: https://github.com/servo/servo/issues/11361

My tentative plan is to always use rustup, and run it with a `RUSTUP_TOOLCHAIN` environment variable that selects the appropriate date nightly or versioned stable. However we’d want to make sure we run `cargo` through rustup.rs’s proxy, and not run directly a plain Cargo that would ignore `RUSTUP_TOOLCHAIN` and use a rustc that might not be the appropriate version. (Which would likely cause a confusing build error later on.)

So, to do that we’d start it from Python with something like `subprocess.Popen(["cargo"] + cargo_args, executable="rustup")`. (And similarly for `rustc`, `rustdoc`, etc.) This makes Python look for an executable in `$PATH` named `rustup`, and execute it with `argv[0] == "cargo"`.

So this all relies on what’s probably an implementation detail of rustup.rs: the proxy executable installed at `$CARGO_HOME/bin/cargo` is exactly the same as `$CARGO_HOME/bin/rustup`, and they behave differently based on `argv[0]`.

Is this a bad idea? Is there a better way? Or does it sound fine to rely on running `rustup` with a different `argv[0]` triggering proxy mode?

SimonSapin:
This fixes #1293. See discussion there.

### DIFF
--- a/src/rustup-cli/proxy_mode.rs
+++ b/src/rustup-cli/proxy_mode.rs
@@ -49,7 +49,7 @@ pub fn main() -> Result<()> {
 fn direct_proxy(cfg: &Cfg, arg0: &str, toolchain: Option<&str>, args: &[OsString]) -> Result<()> {
     let cmd = match toolchain {
         None => try!(cfg.create_command_for_dir(&try!(utils::current_dir()), arg0)),
-        Some(tc) => try!(cfg.create_command_for_toolchain(tc, arg0)),
+        Some(tc) => try!(cfg.create_command_for_toolchain(tc, false, arg0)),
     };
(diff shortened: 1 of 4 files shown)


SimonSapin:
I’d appreciate a new release with this change. I *think* this is the last blocker for using rustup in Servo: https://github.com/servo/servo/issues/11361. Sorry I missed that more would be needed before I asked for 1.7!

PR #1298, opened 2017-11-30
Prevent `rustup update` to a toolchain without `rustc` or `cargo`.

### LINKED ISSUE #1297: rustup update nightly on macOS rustc does not exist
```bash
$ rustup update nightly
info: syncing channel updates for 'nightly-x86_64-apple-darwin'

  nightly-x86_64-apple-darwin unchanged - (rustc does not exist)
$ rustup --version
rustup 1.7.0 (813f7b7a8 2017-10-30)
```

Is nightly build for macOS down? I have yesterday's nightly installed before running `rustup update nightly` then `rustup` removed it while it can't find a newer one to install(I lost that logs on my terminal).

kennytm:
When the Rust CI is misconfigured, it is possible to produce a nightly where `rustc` or `cargo` is not available. This makes the entire toolchain unusable (for one day at least).

This PR tries to prevent the update operation from starting at all if these essential components are missing, so the user would still have a working though outdated toolchain.

Resolves #1297.

### DIFF
--- a/src/rustup-dist/src/manifestation.rs
+++ b/src/rustup-dist/src/manifestation.rs
@@ -107,6 +107,22 @@ impl Manifestation {
             return Ok(UpdateStatus::Unchanged);
         }
 
+        // Make sure we don't accidentally uninstall the essential components! (see #1297)
+        let missing_essential_components = ["rustc", "cargo"]
+            .iter()
+            .filter_map(|pkg| if final_component_list.iter().any(|c| &c.pkg == pkg) {
+                None
+            } else {
+                Some(Component {
+                    pkg: pkg.to_string(),
+                    target: Some(self.target_triple.clone()),
+                })
+            })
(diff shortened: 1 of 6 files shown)


alexcrichton:
Awesome, thanks! It looks like though that maybe one test is failing on Linux? (although it could be spurious, I haven't checked the tests here in awhile)

kennytm:
```
+ cargo test --release -p download --target x86_64-unknown-linux-gnu
error: failed to write /src/Cargo.lock
Caused by:
  failed to open: /src/Cargo.lock
Caused by:
  Read-only file system (os error 30)
```

Doesn't seem like caused by me... The [latest build in master](https://travis-ci.org/rust-lang-nursery/rustup.rs/jobs/309012535) is failing with the same error also.

alexcrichton:
Ok, I think I know what's going on

PR #1324, opened 2018-01-02
Add print-only url to rustup doc, fix #1288

### LINKED ISSUE #1288: Add an option to print only the url from `rustup doc`
In option to #1287, to easy the integration with other tools, `rust doc` can offer a `--print-only` option to print the target url without firing up the browser.

amar-laksh:
Sorry for the earlier one, this is the complete commit.


src/rustup-cli/rustup_mode.rs
```diff
@@ -788,6 +792,14 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
         "index.html"
     };
 
+    if m.is_present("print-only") {
+        let path =  toolchain.doc_path(doc_url);
+        match path {
+            Ok(v) => return Ok(println!("file:/{}", v.to_str().unwrap())),
```
Mark-Simulacrum, line 798: This probably wants to be `v.display()`.

Mark-Simulacrum (commented):
I think there's also a `.swp.` file and another file `3` that snuck into the PR on accident; could you remove those?

src/rustup-cli/rustup_mode.rs
amar-laksh: Cleared the extra files, BTW v.display() for some reason doesn't provide the exact URL. So didn't change that part. Thanks.

Diggsey, line 798: This will produce a URL like `file:/C:\Foo\bar` on windows, or possibly something even stranger, can you update this to work on windows, and also add a test for this command.

Diggsey (changes requested):

src/rustup-cli/rustup_mode.rs
amar-laksh: Sure, I'll try to work it out.

amar-laksh: BTW, shouldn't the URL be just the URL with no platform-specific prefixes? 
I mean just the 'C:\Foo\bar' or whatever the platform-specific URL is. @Diggsey ?

Diggsey: That is probably fine - at least on windows the browser is smart enough to turn a path into a `file:///` URL.

PR #1337, opened 2018-01-11
Warn when tools are missing and allow to override

### LINKED ISSUE #1277: [Feature Request] Warn users of component availability changes
Rustup should warn users of component availability changes before going on with an update.

For instance, the `rls` component was once renamed to `rls-preview` and I was quite confused then about the missing of `rls` after the update. And just yesterday, `rls-preview` was temporarily removed and guys are now struggling downgrading to the last working one (rust-lang-nursery/rls#535).

So whenever there is a new component available, it's better to inform users of it, e.g.

```
A new component `shiny-new-component` has been added to `nightly-toolchain`.
Run `rustup component add shiny-new-component` to install it.
```

And if there is an installed component that is removed, I reckon we should prompt users for a confirmation, e.g.

```
The component `dev-essential-component` has been removed from `nightly-toolchain 2017-12-12`.
Are you sure you want to continue? [Y/n]
```

nrc:
Closes #1277 

Idea here is to not update if the RLS is missing (for example). There was actually functionality to do this already, but it was not catching the RLS and Rustfmt since they were missing in an unexpected way.

The second commit adds `rustup update --force` to update even if some components are missing

r? @alexcrichton

### DIFF
--- a/src/rustup-cli/common.rs
+++ b/src/rustup-cli/common.rs
@@ -189,9 +189,9 @@ fn show_channel_updates(cfg: &Cfg, toolchains: Vec<(String, rustup::Result<Updat
     Ok(())
 }
 
-pub fn update_all_channels(cfg: &Cfg, self_update: bool) -> Result<()> {
+pub fn update_all_channels(cfg: &Cfg, self_update: bool, force_update: bool) -> Result<()> {
 
-    let toolchains = try!(cfg.update_all_channels());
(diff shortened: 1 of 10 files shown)


alexcrichton:
Looks good to me! I think though that the added test may be failing?

nrc:
OK, looks like we're passing tests now (I guess Travis is stuck or got bored or something).

PR #1341, opened 2018-01-23
Click the instruction box to select all the code for easy copying

shinglyu:
When user clicks on the code box in the index.html, the code will be selected automatically. So you can just Ctrl+C or right click to copy. Saves some manual work to select the code.

### DIFF
--- a/www/rustup.js
+++ b/www/rustup.js
@@ -113,8 +113,22 @@ function fill_in_bug_report_values() {
     nav_app.textContent = navigator.appVersion;
 }
 
+function set_click_select_all() {
+    "use strict";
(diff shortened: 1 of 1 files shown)


Diggsey:
Thanks for the PR! Personally I'm not a fan of these "click to select" behaviours - often you want to copy part of the command and having the page interfere with that is very irritating.

For me triple-click already selects the whole command, does that not work everywhere?

t-botz:
Triple click works everywhere. I am not a fan either of the "click to select" unless there is a reason for it like copying a code snippets without the line number which is not the case here.

PR #1346, opened 2018-01-27
Fall back to wget

kraai:
If curl is not available, fall back to using wget.  This can be useful on systems where curl isn't installed by default but wget is (e.g., Debian 9).

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -9,8 +9,8 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
-# This is just a little script that can be curled from the internet to
-# install rustup. It just does platform detection, curls the installer
+# This is just a little script that can be downloaded from the internet to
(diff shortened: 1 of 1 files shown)


rustup-init.sh
```diff
@@ -100,7 +107,10 @@ main() {
     fi
 
     ensure mkdir -p "$_dir"
-    ensure curl -sSfL "$_url" -o "$_file"
+    case $downloader in
+	curl) ensure curl -sSfL "$_url" -o "$_file";;
```
t-botz, line 111: I actually prefer the approach taken here :
https://github.com/rust-lang-nursery/rustup.rs/pull/1268
which doesn't require to know about curl/wget in several part of the code

t-botz:
@nrc duplicate of https://github.com/rust-lang-nursery/rustup.rs/pull/1373
please close

Issue #1353, opened 2018-02-08
Slow startup on Windows


Vlad-Shcherbina:
To reproduce:

1. Download `rustup-init.exe` from https://www.rustup.rs/.
2. Run it to install Rust.
3. Run `rustup --version`.

Expected: this command to print its version ("rustup 1.10.0 (c6a8f7c60 2018-01-25)") and exit instantly.

Actual behavior: this command takes about one second to print version and exit.

----------

To investigate, I tried building rustup from source. I ran the following commands:

```
git clone git@github.com:rust-lang-nursery/rustup.rs.git
cd rustup.rs
cargo build --release
copy .\target\release\rustup-init.exe .\rustup.exe
```

And then

```
.\rustup.exe --version
```

Locally built rustup starts up instantly. The size of the executable is 6MB (for comparison, the official download is 13MB). Could it be that the official download was produced for the wrong configuration (debug build or something)?

I used the following compiler to build it:

```
> rustc --version --verbose
rustc 1.25.0-nightly (29c8276ce 2018-02-07)
binary: rustc
commit-hash: 29c8276cee4a0eab7e0634ff25c6b47bd9f87c6c
commit-date: 2018-02-07
host: x86_64-pc-windows-msvc
release: 1.25.0-nightly
LLVM version: 4.0
```

steffengy:
Possibly related to https://github.com/rust-lang/cargo/issues/5028.
Atleast for me locally disabling windows defender makes those commands (including rustup --version) finish immediately, with it enabled it takes a few seconds.

Is that also the case on your end?

steffengy:
@alexcrichton 
Using either:
- https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe
- https://static.rust-lang.org/rustup/archive/1.10.0/x86_64-pc-windows-msvc/rustup-init.exe

as rustup.exe is faster (without the noticed delay)
Strangely those binaries are only 5.7 MB while my local rustup 1.10.0 is ~12 MB.

Did we previously ship debug builds or does rustup download another binary for self-update?

alexcrichton:
@steffengy whoa interesting! I think the difference you're seeing is MSVC vs MinGW. I think we ship `x86_64-pc-windows-gnu` binaries by default? If it's true though that MSVC is faster we should certainly change that!

steffengy:
@alexcrichton 
That doesn't seem to be the issue.
```
[...]

PR #1354, opened 2018-02-13
windows: detect architecture on website, update to matching arch (#1353)

steffengy:
Updates on windows will now switch over to host appropriate versions,
ensuring the best user experience (e.g. x86_64 builds on x64).

The website can distinguish between Win64/Win32 builds now and
a new platform can be introduced more easily.

r? @alexcrichton

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -1448,7 +1448,19 @@ pub fn prepare_update() -> Result<Option<PathBuf>> {
     }
 
     // Get build triple
-    let triple = dist::TargetTriple::from_build();
+    let build_triple = dist::TargetTriple::from_build();
+    let triple = if cfg!(windows) {
+        // For windows x86 builds seem slow when used with windows defender.
+        // The website defaulted to i686-windows-gnu builds for a long time.
+        // This ensures that we update to a version thats appropriate for users
+        // and also works around if the website messed up the detection.
+        // If someone really wants to use another version, he still can enforce
+        // that using the environment variable RUSTUP_OVERRIDE_HOST_TRIPLE.
+
+        dist::TargetTriple::from_host().unwrap_or(build_triple)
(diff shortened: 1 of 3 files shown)


steffengy:
Notably, this shouldn't be merged without a test of rustup itself yet, just first feedback & an initial CI check.

alexcrichton:
This looks good to me, thanks! What sort of testing did you have in mind for this?

steffengy:
@alexcrichton
Building rustup locally as i686 and x86_64 with 1.09 as version and testing if it downloads the right 1.10 executable when updating. Other than that I think this should be fine?

alexcrichton:
I can confirm that locally after building the i686-pc-windows-msvc version it switched to x86_64-pc-windows-msvc

src/rustup-cli/self_update.rs
```diff
@@ -1448,7 +1448,17 @@ pub fn prepare_update() -> Result<Option<PathBuf>> {
...
     // Get build triple
-    let triple = dist::TargetTriple::from_build();
+    let triple = match dist::TargetTriple::from_build() {
+        // For windows x86 builds seem slow when used with windows defender.
+        // The website defaulted to i686-windows-gnu builds for a long time.
+        // This ensures that we update to a version thats appropriate for users
+        // and also works around if the website messed up the detection.
+        // If someone really wants to use another version, he still can enforce
+        // that using the environment variable RUSTUP_OVERRIDE_HOST_TRIPLE.
+        #[cfg(windows)]
```
alexcrichton: This ended up causing a build warning on windows, could `if cfg!(windows)` be used instead?

steffengy:
@alexcrichton 
Changed that match to an if and tested using x86_64-pc-windows-msvc and i686-pc-windows-gnu (UGH I hate MSYS2) and it also works.

So good to go from my side.

PR #1355, opened 2018-02-13
Update Dependencies & Bump to 1.11.0

steffengy:
This makes proper architecture (=faster) windows builds available for **1.12.0**.

I sneaked some version updates in to ensure that we're a step closer
to a consistent state on that front.

Some open points:
- [ ] 1. Should we add a special case to self_update which 
makes the windows fixes available on **1.11.0** (this) already? 
e.g. hardcoding the version [somewhere](https://github.com/rust-lang-nursery/rustup.rs/blob/0068f0999acb923428cbd709650c57fbe04a0344/src/rustup-cli/self_update.rs#L1497) in the windows update logic introducted.
This potentially could lead to infinite looping, if we're not careful.
- [ ] 2. Any objections against the updates? Splitting them into a separate PR would be fine.
       (Not sure if CI even passes at this point, but we'll see)

r? @alexcrichton

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 1.11.0
+
+* [windows: detect architecture on website, update to matching arch][pr1354]
+
+[pr1354]: https://github.com/rust-lang-nursery/rustup.rs/pull/1354
+
+Contributors: Steffen Butzer
+
 # 1.10.0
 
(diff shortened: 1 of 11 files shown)


alexcrichton:
@bors: r+

Thanks! The updates look great to me and I think it's fine to hold off on the upgrades for another version

alexcrichton:
Deployed!

PR #1365, opened 2018-03-03
More command examples

### LINKED ISSUE #1362: Removing toolchains
I couldn't find any nice way of removing a toolchain once I'm done with it. For example, if I just downloaded a nightly version to play around with some unstable API, but it's not something I want to keep using going forward, so I'd like to just delete the nightly toolchain again.

Is this possible as-is, and I just missed it?

KasMA1990:
I had trouble finding the `toolchain uninstall` command, because I didn't realize that each subcommand had their own help pages to check on. This should hopefully make it clearer for somebody in the same situation.

Closes #1362

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,6 +519,8 @@ Command | Description
 `rustup override set nightly-2015-04-01` | For the current directory, use a nightly from a specific date
 `rustup toolchain link my-toolchain "C:\RustInstallation"` | Install a custom toolchain by symlinking an existing installation
 `rustup show` | Show which toolchain will be used in the current directory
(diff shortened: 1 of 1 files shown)


kbknapp:
Maybe it's confusing, but you can also say `rustup help <subcommand>` to see a subcommands help page, to include it's children such as `rustup help toolchain add` to see `rustup-toolchain-add`'s help message, etc.

t-botz:
@nrc More examples can't really hurt. Close or approve?

nrc:
Thanks!

PR #1370, opened 2018-03-07
Added a link to all installers to the homepage

t-botz:
re #1253

I have just resolved conflict of https://github.com/rust-lang-nursery/rustup.rs/pull/1270 and adapted to the new index page.

Here is how it looks:
otheros2

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -26,7 +26,7 @@
 <div id="platform-instructions-unix" class="instructions" style="display: none;">
   <p>Run the following in your terminal, then follow the onscreen instructions.</p>
   <pre>curl https://sh.rustup.rs -sSf | sh</pre>
-  <p class="other-platforms-help">You appear to be running Unix. Click <a class="default-platform-button" href="#">here</a> to see the installer for all supported platforms.</p>
+  <p class="other-platforms-help">You appear to be running Unix. If not, <a class="default-platform-button" href="#">display all supported installers</a>.</p>
 </div>
 
 <div id="platform-instructions-win32" class="instructions" style="display: none;">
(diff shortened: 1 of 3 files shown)


www/index.html
```diff
@@ -26,22 +26,25 @@
 <div id="platform-instructions-unix" class="instructions" style="display: none;">
   <p>Run the following in your terminal, then follow the onscreen instructions.</p>
   <pre>curl https://sh.rustup.rs -sSf | sh</pre>
+  <p class="other-platforms-help">You appear to be running Unix. Click <a class="default-platform-button" href="#">here</a> to see the installer for all supported platforms.</p>
```
nrc: Could we rephrase these links to avoid the 'click here' anti-pattern, e.g., make the link text "view all supported installers" or something like that

nrc (commented):
I think this is a good change, thanks for rebasing the PR. I'd like to fixup the links a little bit though, see comment inline.

nrc:
ping @withoutboats any comments on this?

www/index.html
t-botz: What about : "You appear to be running Unix. If not, [display all supported installers]."

nrc: Sounds good to me!

nrc:
ping @alexcrichton and @withoutboats - any thoughts? I'm inclined to merge this with my review comment addressed.

alexcrichton:
Sounds good to me!

Diggsey:
Hey @thibaultdelor, any chance you could update the PR with that change? Thanks!

t-botz:
Sure will do today

t-botz:
Here you go. I have also slightly reduced the font so it looks on one line in windows. It was looking weird otherwise.
Before:
image
After:
image

Diggsey:
Thanks, looks great!

PR #1380, opened 2018-03-17
Give correct error message if user tries to install an unavailable toolchain

emielwiedijk:
Currently if a user runs `rustup install unavailable-toolchain`, the error message is: `error: toolchain 'unavailable-toolchain' is not installed`. This is error message is weird, since the user runs rustup to install the toolchain. This commit copies the "InvalidToolchainName error from rustup-dist, and configures rustup install to use this error.


Diggsey:
Thanks for the PR!

PR #1388, opened 2018-04-06
Ignore IntelliJ project files

cyplo:
Ignore project directory added automatically when IntelliJ opens the rustup directory.


nrc:
Thanks!

Issue #1239, opened 2017-08-21
Invalid cross-device link (os error 18) when upgrading on a docker OverlayFS 


wraithan:
```
$ rustup update nightly
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: latest update on 2017-08-21, rust version 1.21.0-nightly (8c303ed87 2017-08-20)
info: downloading component 'rustc'
info: downloading component 'rust-std'
info: downloading component 'cargo'
info: downloading component 'rust-docs'
info: removing component 'rustc'
info: rolling back changes
error: could not rename component directory from '/root/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/etc' to '/root/.rustup/tmp/x5u5mnp0hhtywco8_dir/bk'
info: caused by: Invalid cross-device link (os error 18)
```

`std::fs::rename()` basically doesn't work on OverlayFS as far as I can tell by looking at other similar reports for various languages and projects hitting cross-device link errors on OverlayFS is boils down to using the rename syscall.

I'd like to propose wrapping the `std::fs::rename()` calls and if on linux detect os error 18 attempt to do a copy and delete instead. There are periodic other reports of errors like this on various platforms, the wrapper could try to handle the other OS cases too if they have a similar error code (or maybe even the same one if this is standard, I'm not sure).

Interestingly there is the bootstrap/update problem where folks who are experiencing may be unable  to update their rustup install and not be able get the update that fixes the problem once there is a solution. Those folks will need to be advised to reinstall their rustup.

If the proposed solution to the problem works for the dev team, I'll attempt to provide a PR within a week of getting the go ahead.

This is relevant because some people use a common Docker image for their CI environments that may not be updated frequently enough for beta/nightly and have `rustup update $desired_env` in their script. Which is how I found this problem.

wraithan:
Spoke with @nrc and @alexcrichton on IRC and they said this seemed reasonable. I'll put forward an implementation this week.

cyplo:
Heyo ! Any news on this one ? I encounter this bug regularly when doing builds on dockerized CI. Let me know if there is any more info I can provide.

cyplo:
Looking at the sources, there already exists a wrapper function called `utils::rename_file`, it's used by `components` and `transaction`. Would that be a good candidate here to replace every other call to `fs::rename` ?

ishitatsuyuki:
[...]

PR #1389, opened 2018-04-06
Make sure rustup uses utils::rename* consistently. 

cyplo:
This is a small refactoring tho prepare the codebase for further changes
to the renaming functionality, e.g. ability to recover from atomic
rename errors - makes sure that `fs::rename` is only used via wrappers in `utils`

See #1239.
Best to merge after #1388 

cc @nrc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 /target
 /Cargo.lock
 /.settings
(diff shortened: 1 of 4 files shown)


nrc:
lgtm, thanks!

PR #1390, opened 2018-04-08
rustfmt'ed

cyplo:
Reformatted the code using cargo fmt.
Added a separate Travis job to the build matrix that fails when code is
not properly formatted

potentially closes  #1291 as well

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,7 +8,8 @@ use std::process::Command;
 struct Ignore;
 
 impl<E> From<E> for Ignore
-    where E: Error
+where
+    E: Error,
 {
     fn from(_: E) -> Ignore {
         Ignore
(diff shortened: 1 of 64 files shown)


cyplo:
cc @nrc

nrc (commented):
This is great thank you! I've wanted to run rustfmt over Rustup recently and I think we should land this. However, we've found that running Rustfmt on CI is not a great experience yet because as Rustfmt evolves, it's formatting changes and can cause tests to fail. We're working on this on the Rustfmt side, but in the meantime I think it is better to land the formatting changes without the CI change.

cyplo:
Would pinning a version of rustfmt to use on CI help ? Diesel is doing something along those lines, if I recall correctly.

nrc:
> Would pinning a version of rustfmt to use on CI help ?

Yeah, I think this is a good stop-gap. The problem with this is that it makes it difficult for contributors to get the right version of Rustfmt - I think they have to build from source to get the right version, and this might not work if it no longer compiles with their current toolchain.

cyplo:
That's a very good point, I haven't thought of that. 
How about either merging just the formatting changes and then the Travis change we would have stored somewhere near some issue in the tracker, for later use? 
OR
I could allow this separate build job that checks the fmt compliance to fail, i.e. mark as allow failures on Travis? 

What do you think? Which one sounds better? Thanks!

nrc:
I would just land the formatting changes and hope that we can make the CI better in the future with some help from Rustfmt.

cyplo:
Alrighty, will catchup with master and remove the travis changes and ping you when done. thank you :)

cyplo:
Cool, the changes are now formatting changes only, removed the travis changes altogether from the PR
cc @nrc

nrc:
Thanks!

PR #1419, opened 2018-05-27
Remove components if they don't exist anymore during an update

emilyalbini:
At the moment, if a new release removes a component from its manifest, the `install` command works fine but the `update` command fails because that component is missing. This is just an hassle for the user, who has to remove the toolchain and reinstall it again.

This PR changes the behavior of rustup so that if a component is missing (so it's not in either the `components` or `extensions` list of the `rust` component) a warning is shown to alert the user about this and the component is simply removed. If the component is just marked as `available = false` the old behavior (erroring out) is still used, since that might be caused by a CI failure (if I understand correctly).

Part of https://github.com/rust-lang/rust/issues/51097.

r? @Mark-Simulacrum @alexcrichton 

screenshot from 2018-05-27 12-48-00

### DIFF
--- a/src/rustup-dist/src/manifestation.rs
+++ b/src/rustup-dist/src/manifestation.rs
@@ -475,6 +475,7 @@ impl Update {
             rust_target_package,
             new_manifest,
             &changes,
+            notify_handler,
         );
 
         // If this is a full upgrade then the list of components to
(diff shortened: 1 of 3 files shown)


Diggsey:
Thanks for the PR!

Looks like the tests need updating in response to these changes - could you also add tests for the new behaviour? In particular, I want to be sure that updates continue to fail if a component exists but is unavailable.

emilyalbini:
Updated, should have fixed and improved the tests.

Diggsey:
Thanks!

emilyalbini:
Can we get this released before 1.26.1 (tuesday)?

PR #1421, opened 2018-05-28
ADDED: print-only doc option

amar-laksh:


Diggsey:
Thanks, but this is still missing a test.

amar-laksh:
@Diggsey , could you help me regarding writing tests? Any links would be much appreciated?

Diggsey:
Sure, take a look in `tests/cli-misc.rs` - each test consists of calling `setup` with a predefined scenario, and then simply running `rustup` with some arguments. To test this you'll want to run with your `--print-docs` argument, and then assert that it prints out the correct filename.

Issue #1364, opened 2018-03-02
Download progress indicator doesn't show up. Needs term >= 0.5.0 ?


ynakao:
Today I noticed when upgrading Rust v1.24.1, download progress indicators didn't appear at my Arch linux PC.

![noeta](https://user-images.githubusercontent.com/6405335/36908370-5e73db12-1e7e-11e8-95bd-ea0400b0c7ba.jpg)

At macOS machine, download ETA showed up properly.

![rustup](https://user-images.githubusercontent.com/6405335/36907883-16de48b0-1e7d-11e8-94f7-ffc537e1408f.jpg)

This function is handled by [download_tracker.rs](https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup-cli/download_tracker.rs), and to write the info on the terminal `term` crate is necessary. So, I searched at `term` crate issues, found [the one](https://github.com/Stebalien/term/issues/81) which seems to cause this problem. According to the discussion, if installed `ncurses` version is above v6.1 this happens, and they released fixed `term` v0.5.0.

My Arch has `ncurses` v6.1, and I tested other machines. In Debian stable which has v6.0 download indicator worked, but Debian testing which has v6.1 it didn't work ([`ncurses-bin`](https://packages.debian.org/search?keywords=ncurses-bin)). macOS may have v5.x since there's /usr/bin/ncurses5.4-config.

Also I tried cloning this repo and build rustup localy with the change `term = "0.5"` in Cargo.toml (currently `term = "0.4.4"`) at my Arch machine. In the result, rustup worked fine displaying ETA messages.

![afterbuild](https://user-images.githubusercontent.com/6405335/36908392-6cc23ace-1e7e-11e8-9b61-2d7d3035c784.jpg)

Therefore, I think it needs to change rustup dependency information especially `term` crate version number. I'm sorry if I'm wrong.

Frederick888:
Same here. Arch Linux + Yakuake + Zsh.

killkrt:
Same here. Progress is not shown anymore on Linux, while is working perfectly on macOS and Windows.

OS: Arch Linux (kernel 4.16.6-1)
Shell: Zsh 5.5.1

Diggsey:
A PR for this would be welcome

ynakao:
I just submitted PR #1422.

milahu:
still have the same problem: download starts, but progress is not visible

```
rustup --version
rustup 1.22.1 (2020-07-08)

[...]

PR #1422, opened 2018-05-28
Update term crate dependency

ynakao:
Please also read #1364.
Download indicator doesn't show up at OSes that have newer `ncurses` package due to the `term` crate issue. To resolve this it needs to update `term` crate version number in `Cargo.toml`.


Diggsey:
Thanks!

Issue #1406, opened 2018-04-29
rustup show: add flag to show only active toolchain


TimNN:
I often want to know which toolchain is active for a certain directory (and which directory has the override). Having `rustup show` first print a few dozen lines of installed toolchains is rather annoying in that situation.

It would be nice to have a flag which would cause `rustup show` to only print the active toolchain. (For me it would even be preferable to make that the default behavior, since the list of installed toolchains is easily available via `rustup toolchain list`).

das-sein:
@Diggsey I think this is resolved by `rustup show active-toolchain`, right?

dwijnand:
Possibly resolved by #1633 (using `rustup default` instead)?

TimNN:
> @Diggsey I think this is resolved by `rustup show active-toolchain`, right?

This does not quite address my original issue: It shows the active toolchain, but not _**why**_ it is active (e.g. which directory the override is coming from, or that the default is active).



> Possibly resolved by #1633 (using `rustup default` instead)?

I don't have a build with that patch, but from the name of the command, I don't believe that this will resolve my issue: I'm interested in the _active_ toolchain, not the _default_ one.

dwijnand:
Good point.

das-sein:
@Diggsey I think I figured out what you mean. The `rustup show` command outputs the reason for the override for any override you have set, but `rustup show active-toolchain` didn't do that. I've started a PR with the more standard behavior. Let me know how I did.

Issue #1199, opened 2017-07-10
Completion instructions for bash


dns2utf8:
Hi all

I noticed that the completion file should go to a new location:
```
/usr/share/bash-completion/completions/rustup
```

Issue #450, opened 2016-05-12
Machine-readable output


fenhl:
I used to have a function in my shell that shows when the current directory is running on a multirust toolchain other than stable. The current formatting is making the port to rustup.rs difficult. An option to generate machine-readable output (preferably JSON) on relevant commands like `rustup show` and `rustup override list` would be great.

brson:
Yes indeed. Sorry for breaking your stuff, but I did make the decision to cater to human-reading first.

Since this is mostly for shell scripts I'd guess it's best to do something simple and line-oriented. The default 'smart' behavior of `show` where it decides how much to show or not would be bad.

Perhaps a scheme like

* `rustup show active-toolchain --machine-readable`
* `rustup show active-toolchain-version --machine-readable`
* `rustup toolchain list --machine-readable`

matklad:
>Since this is mostly for shell scripts

It would be useful in IntelliJ Rust as well, though I would also prefer simple text format.

dherman:
This looks like it could be useful for Neon too -- +1 to shell script-friendliness. For the time being I may end up parsing the output of `rustc --version` or `rustup toolchain list`, but a stable machine-readable output would be awesome!

alexreg:
Any progress on this?

QAston:
Still needed

camsteffen:
Found this because I wanted to do something like:

```shell
cd some/path
toolchain_name=$(rustup show active-toolchain)
cd some/other/path
RUSTUP_TOOLCHAIN=$toolchain_name cargo ...
```

Issue #387, opened 2016-04-30
Install zsh completions for rustc


japaric:
We already install zsh completions for cargo. It would be nice if rustup also installed the rustc zsh completions from [this repository](https://github.com/rust-lang/zsh-config).

nerdrew:
I like this.

brson:
Fine by me, though I'd like to understand better how they are supposed to be installed. You say we already install cargo zsh completions, but we don't do anything special for them - they're just buried in the toolchain directory. That's not so user-friendly.

freebroccolo:
@brson The completions file needs to be in a directory contained in `fpath` so probably the most direct way is to just create a symlink:

```ln -s /usr/local/share/zsh/site-functions/_rust rust-lang/zsh-config/_rust```

On OS X, I believe this is where `brew` usually links completions.

brson:
@freebroccolo Thanks for the tips.

I see two problems that need to be solved: first, the completions could theoretically be toolchain-specific (and indeed the cargo zsh completions today come with the compiler, not rustup). So a single symlink isn't sufficient because it won't change when the active toolchain changes.

Second, installing to /usr/local typically requires root, and rustup installation today does not require root, which I see as a strong advantage. I don't want to require root just to install completions.

For the first, I'd suggest that the drift in the command sets of `rustc`, `cargo` and other potential tools is small enough that rustup could just install one global completion that more-or-less works with all toolchains. That is, the completions are distributed with rustup, not the toolchains. If the completions are smart enough they may even be able to change their behavior based on the compiler version. Oh, maybe the completions installed by *rustup* just detect the toolchain and then defer to the *toolchain* completions. That sounds plausible.

For the second I'm not sure. Maybe have a command `rustup self install-completions`. Maybe an optional install-time component that runs `sudo`.

brson:
I think I'd prefer the simplicity of removing completions from the toolchain installer and just making rustup responsible for them. This has the downside that maybe the completions aren't exactly right for any particular toolchain (or the completions have to support multiple toolchains). Here's what I'd suggest the next steps are:

* Add the cargo and rustc zsh completions to the rustup binary via `include_str!()`
* Add a command `rustup self install-shell-completions` that does the copies with `sudo` on Unix (not sure about windows)
* Punt on install-time configuration of shell completions for now
* Then do the same for bash

cc @alexcrichton @nrc @aturon @japaric a strategy for distributing Rust tool shell-completions.

[...]

PR #1425, opened 2018-06-02
Include shell completion for Cargo

ricvelozo:
*   Include shell completion for Cargo on `rustup completions` (#387,  rust-lang/cargo#5596)
*   Add argument to show only the default toolchain: `rustup toolchain list --default` (#450, #1406)
*   Update the `rustup completions bash` instructions to use home directory (#1199)

### DIFF
--- a/src/rustup-cli/common.rs
+++ b/src/rustup-cli/common.rs
@@ -1,5 +1,6 @@
 //! Just a dumping ground for cli stuff
 
+use clap::ArgMatches;
 use rustup::{self, Cfg, Notification, Toolchain, UpdateStatus};
 use rustup::telemetry_analysis::TelemetryAnalysis;
 use errors::*;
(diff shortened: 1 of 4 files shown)


src/rustup-cli/rustup_mode.rs
```diff
@@ -80,11 +80,34 @@ pub fn main() -> Result<()> {
         },
         ("completions", Some(c)) => {
             if let Some(shell) = c.value_of("shell") {
+                let shell = shell.parse::<Shell>().unwrap();
+                let prefix = "~/.rustup/toolchains/$(rustup toolchain list --default)";
```
Diggsey, line 84: This prefix is not correct - rustup is not necessarily installed here, and the default toolchain isn't necessarily correct.

To get the installation directory of the active toolchain, you should use `rustc --print sysroot`

src/rustup-cli/rustup_mode.rs
```diff
@@ -80,11 +80,34 @@ pub fn main() -> Result<()> {
...
+
                 cli().gen_completions_to(
                     "rustup",
-                    shell.parse::<Shell>().unwrap(),
+                    shell,
                     &mut io::stdout(),
                 );
+
+                match shell {
```
Diggsey, line 92: Non-trivial logic like this should be pulled out into its own function.

src/rustup-cli/rustup_mode.rs
```diff
@@ -172,7 +195,15 @@ pub fn cli() -> App<'static, 'static> {
                 .setting(AppSettings::VersionlessSubcommands)
                 .setting(AppSettings::DeriveDisplayOrder)
                 .setting(AppSettings::SubcommandRequiredElseHelp)
-                .subcommand(SubCommand::with_name("list").about("List installed toolchains"))
+                .subcommand(
```
Diggsey, line 198: This change should not be necessary

src/rustup-cli/common.rs
```diff
@@ -326,24 +327,32 @@ pub fn list_components(toolchain: &Toolchain) -> Result<()> {
     Ok(())
 }
 
-pub fn list_toolchains(cfg: &Cfg) -> Result<()> {
+pub fn list_toolchains(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
```
Diggsey, line 330: Please revert the changes to this file as they should not be necessary.

Diggsey (changes requested):
Thanks for the PR! There are however a few changes needed before I'd be happy accepting it.

src/rustup-cli/rustup_mode.rs
ricvelozo: Are you requesting to remove the new `--default` arg?

src/rustup-cli/rustup_mode.rs
ricvelozo: Does to replace the `rustup toolchain list --default` for `rustc --print sysroot` solves the problem? I have no way to know the toolchain that the user is using, just the default.

ricvelozo: @Diggsey Help me to finish this, please.

Kerollmops: `rustc --print sysroot` will call the binary of the currently used toolchain.

nrc:
Closing due to lack of activity and reviewer bandwidth.

PR #1428, opened 2018-06-05
Update Visual C++ Build Tools URL

b-r-u:
landinghub.visualstudio.com domain says `Expired Account`: http://landinghub.visualstudio.com/visual-cpp-build-tools

So the [illustrated guide from shepmaster](https://github.com/rust-lang-nursery/rustup.rs/issues/1363#issuecomment-369953262) does not work like that anymore.

The new link: https://aka.ms/buildtools

@steveklabnik also mentioned the incorrect link: https://github.com/rust-lang-nursery/rustup.rs/issues/1384#issue-309349986


Diggsey:
Thanks for the PR!

PR #1432, opened 2018-06-09
Bump version number

Diggsey:


alexcrichton:
This should be fine to merge when ready, after merging to `master` you'll want to push the `stable` branch, and when that's done I can do the rigamarole of publishing the release

Diggsey:
@alexcrichton we currently have two failing bors tests (powerpc64le and netbsd) both of which appear to be caused by rustc issues. Is that going to be a problem?

mati865:
powerpc64le issue is basically: https://github.com/rust-lang/rust/issues/50499
It will be fixed in next stable (1.28).

NetBSD is more tricky, `setvbuf` should be available in system libc. Maybe bot is too old?

Diggsey:
@alexcrichton I've merged and pushed to stable anyway

alexcrichton:
Ok! It's probably fine to drop some of those platforms

alexcrichton:
Ok published now!

Diggsey:
Thanks!

PR #1435, opened 2018-06-15
An additional indication for input after the first question

wwkeyboard:
```
2) Customize installation
3) Cancel installation
> 
```

To help indicate to the user that they are expected to enter a number and press enter. It's probably my fault for trying to install rust first thing in the morning, but I sat at this screen for 2 or 3 sips of coffee waiting for it to proceed before realizing I needed to press return. Maybe this caret will help someone else avoid my delay! :smile:

### DIFF
--- a/src/rustup-cli/common.rs
+++ b/src/rustup-cli/common.rs
@@ -44,6 +44,7 @@ pub fn confirm_advanced() -> Result<Confirm> {
     println!("1) Proceed with installation (default)");
     println!("2) Customize installation");
     println!("3) Cancel installation");
+    print!(">");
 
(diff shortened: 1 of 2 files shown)


Diggsey:
Hehe, nice carrot 😛 

Thanks for the PR, looks like there are a few tests you'll need to update though!

wwkeyboard:
:astonished: I'll fix those after work today!

wwkeyboard:
I've also removed the extra space after the caret because lots of editors strip trailing whitespace when you save a file and I didn't want anyone to get caught up by inadvertently causing those tests to fail.

Diggsey:
Thanks!

PR #1437, opened 2018-06-17
Bring rustup.js and markup into alingment with rust-www

brson:
This makes rustup.js the same as on rust-www, pulls in
some of the language from rust-www, pulls in some fixes
to platform detection, fixes some css.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -31,20 +31,24 @@
 
 <div id="platform-instructions-win32" class="instructions" style="display: none;">
   <p>
-    Download and run
+    To install Rust, download and run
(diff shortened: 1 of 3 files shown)


brson:
This needs to be reviewed and merged with https://github.com/rust-lang/rust-www/pull/1128.

Diggsey:
LGTM, but maybe we should just remove `rustup.js` from one of these repos and have it link to `rustup.js` from the other?

brson:
@Diggsey seems ok to me, except for risk of possible unanticipated drift between the two.

brson:
I'm not inclined to work on this further, but doing the change @Diggsey suggested sounds fine.

PR #1449, opened 2018-06-28
Add subcommands to "rustup show" for active-toolchain and it's version

segevfiner:
This should be useful to shell scripts (via command substitution) and other programs that need this information.

Do note that the new commands currently succeed and print nothing when there is no active tool chain. I'm not sure if that's the proper behavior.

Related: #450

### DIFF
--- a/src/rustup-cli/help.rs
+++ b/src/rustup-cli/help.rs
@@ -17,6 +17,15 @@ pub static SHOW_HELP: &'static str = r"DISCUSSION:
     If there are multiple toolchains installed then all installed
     toolchains are listed as well.";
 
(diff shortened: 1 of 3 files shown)


Diggsey:
Thanks for the PR!

I'm not sure the "version" variant is strictly necessary? You can just run `rustc --version`. Getting the active toolchain name is potentially useful, although I think there is potential for misusing it to figure out the toolchain directory: the directory should be obtained using `rustc --print sysroot`.

Could you remove the version variant, and add a note to the help text for the other variant to consider using `rustc --print sysroot` instead?

Diggsey:
~Also, could you add a test for the new behaviour.~ Nevermind, I see you've already got one :)

PR #1453, opened 2018-07-01
Add --path flag to 'rustup doc'

da-x:
This is a rebase of #1311 originally by @joerivanruth. 

This change will be helpful when used by editor plugins such as [rhysd/rust-doc.vim](https://github.com/rhysd/rust-doc.vim).


Diggsey:
Thanks for the PR!

Could you add a test for this please?

PR #1461, opened 2018-07-12
Add clippy to the tools list

yaahc:
Implements https://github.com/rust-lang-nursery/rust-clippy/issues/2901

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -190,7 +190,7 @@ tools, but otherwise, install the C++ build tools before proceeding.
 "#;
 
 static TOOLS: &'static [&'static str] =
-    &["rustc", "rustdoc", "cargo", "rust-lldb", "rust-gdb", "rls"];
+    &["rustc", "rustdoc", "cargo", "rust-lldb", "rust-gdb", "rls", "cargo-clippy"];
 
 // Tools which are commonly installed by Cargo as well as rustup. We take a bit
 // more care with these to ensure we don't overwrite the user's previous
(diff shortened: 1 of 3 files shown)


src/rustup-cli/self_update.rs
```diff
@@ -190,7 +190,7 @@ tools, but otherwise, install the C++ build tools before proceeding.
 "#;
 
 static TOOLS: &'static [&'static str] =
-    &["rustc", "rustdoc", "cargo", "rust-lldb", "rust-gdb", "rls"];
+    &["rustc", "rustdoc", "cargo", "rust-lldb", "rust-gdb", "rls", "clippy-driver", "cargo-clippy"];
```
oli-obk: So if I remember the discussion on discord correctly, we don't actually need `clippy-driver` here, right?

src/rustup-win-installer/src/lib.rs
```diff
@@ -26,6 +26,8 @@ static TOOLS: &'static [&'static str] = &[
     "rls",
     "rustfmt",
     "cargo-fmt",
+    "clippy-driver",
```
oli-obk: same here

yaahc:
forgot to push before leaving for the weekend. Should be good now

alexcrichton:
Thanks! Can this also update `DUP_TOOLS` below the definition of `TOOLS`? That'll help encourage users to uninstall the `cargo install`'d version, if it exists, before adding the rustup component

oli-obk:
@alexcrichton While that made sense for rustfmt (because there was stable rustfmt), for clippy it doesn't make much sense. If you turn on the component, your manually installed clippy was probably already broken and will never work again.

I think we should just silently overwrite an existing installation

alexcrichton:
Ah yes, an excellent point!

alexcrichton:
For those following along, this is now published as rustup 1.13.0 and should be good to go!

PR #1478, opened 2018-08-08
Add --toolchain option to 'rustup doc'

### LINKED ISSUE #1477: Allow specifying rustup doc toolchain
As discussed in https://github.com/rust-lang/rust/issues/53158. If we would like to open the nightly docs with stable as default toolchain. We need to:

```
$ rustup default nightly
$ rustup doc --std
$ rustup default stable
```

As well, part of my confusions is that I thought the following works.

- `rustup +nightly doc` (cargo uses this so I thought it applies to rustup)
- `rustup doc --toolchain=nightly` (`rustup component add clippy preview --toolchain=nightly` works though)

~~Or maybe there's a quick way to switch between stable and nightly docs if both toolchains are installed?~~ Answered in https://github.com/rust-lang/rust/issues/53158#issuecomment-411291134 by @cuviper

pickfire:
Fix #1477

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -389,6 +389,12 @@ pub fn cli() -> App<'static, 'static> {
                         .long("reference")
                         .help("The Rust Reference"),
                 )
+                .arg(
+                    Arg::with_name("toolchain")
+                        .help(TOOLCHAIN_ARG_HELP)
+                        .long("toolchain")
+                        .takes_value(true),
+                )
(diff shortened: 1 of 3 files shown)


nrc (commented):
This looks good, but could you add a test please?

pickfire:
@nrc I am not sure where can I add tests for this. I checked but the shorthand is handled by somewhere else so should we even add tests for this?

pickfire:
Added tests with simple check to test if toolchain is working with nightly.

nrc:
Great, thank you

PR #1482, opened 2018-08-18
add path modification to .bashrc

mstallmo:
Hi all!
I recently installed rust for the first time on my Ubuntu 18.04 machine and noticed that $HOME/.cargo/bin gets added to the .profile and .bash_profile files but not .bashrc. This caused me a little bit of headache figuring out why new terminals couldn't find the cargo command. To fix this I've added .bashrc to the files that get updated with the new PATH when installing rustup. This action will only be performed if the .bashrc file exists and works exactly the same as #1179. 

Thanks for taking a look!

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -1175,6 +1175,14 @@ fn get_add_path_methods() -> Vec<PathUpdateMethod> {
         }
     }
 
+    if let Some(bashrc) = utils::home_dir().map(|p| p.join(".bashrc")) {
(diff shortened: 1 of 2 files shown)


pickfire:
@mstallmo Why add to `.bashrc` when it adds to `.profile` which will be sourced before `.bashrc`? There are a lot of other shells as well though.

nrc:
Sorry, the Rustup maintainers don't have the bandwidth to review at the moment, thanks for submitting a PR.

PR #1485, opened 2018-08-19
Open docs with `opener` crate

Seeker14491:
The idea is to pull the doc-opening functionality from Cargo and rustup out into its own crate, as discussed in https://github.com/rust-lang/cargo/issues/5701#issuecomment-409007567.

As I mentioned in [Cargo's pull request](https://github.com/rust-lang/cargo/pull/5888#issue-208154627), the handling of the `$BROWSER` environment variable changes on Linux.


pickfire:
@Seeker14491 Why not use [`open` crate](https://crates.io/crates/open) instead, it is quite minimal and does not depend on failure. (did 1.0 major release)

Seeker14491:
An advantage of `opener` is better error messages on Mac and Linux. Additionally, the main motivation of making the `opener` crate was to unify cargo and rustup's documentation-opening code. Using two separate crates would defeat the purpose. It's been integrated into cargo, and seems to be working well.

However, [a cargo issue](https://github.com/rust-lang/cargo/issues/6064) has been opened where it seems we've decided we want to explicitly check the `$BROWSER` environment variable for the purpose of overriding the browser used to open the docs. This should probably be implemented before this pull request is merged.

pickfire:
@Seeker14491 Ah, I see that cargo already uses `opener`. There seems to be an issue to use `$BROWSER` in `open` since 2017 though. https://github.com/Byron/open-rs/issues/8

I think both `cargo` and `rustup` uses way different crate, both are different projects, I don't see how it matters much to use different crate since the usage of `rustup` and `cargo` is very different.

Say, `rustup` uses `rustup component add --toolchain nightly` but `cargo` have `cargo +nighty`. As well, the frequency of usage between `rustup` differs quite a lot from `cargo` (it even have a fancy progress bar).

nrc:
Closing for now due to lack of review bandwidth. I would like to share more code between Cargo and Rustup though.

PR #1488, opened 2018-08-25
Add '/home' to .gitignore.

zachlute:
The README suggests creating this directory for local rustup
development, but anybody who tries to do so will immediately find a
whole lot of files in their git status. This prevents that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /.idea/
 *~
 /**/target
+/home


pickfire:
Why do we need to add `/home` to gitignore? I still don't get it. Can't it be ignored in the global git ignore file?

zachlute:
I mean, sure, it could. But then:

1. Every single person has to do it themselves.
1. They'll only know to do it after they have git telling them they have 5000 changes.
1. Some other project might actually use a /home directory for something, so adding it to the global .gitignore wouldn't work for those people.
1. It would be nice to add a note about doing that to the instructions.

On the other hand, is there some _downside_ to adding it to the .gitignore? Like, why _not_ do it? It's pretty common for projects to add their local generated data directories to it and I can't see what it would hurt.

pickfire:
@zachlute I still don't get how `/home` can be in the git repository for rust crate? Why not include `/root` as well?

For me, I manage dotfiles with `alias home='git --work-tree=$HOME --git-dir=$HOME/.home'` (probably similar case to that). I would not pollute my home directory as a git repository by shadowing the git repository as another directory.

zachlute:
The instructions in the rustup readme say:

```bash
$ cargo build
$ mkdir home
$ RUSTUP_HOME=home CARGO_HOME=home target/debug/rustup-init --no-modify-path -y
```

I followed these instructions and vscode immediately refused to perform git operations because I had too many changed files. That's what prompted this change in the first place.

pickfire:
Oh, now I see. The `home` there is not your home directory, it can be anything. For me, probably will be `/tmp/rustup.rs/home` but I don't see that as a problem though.

Instead why not switch to `.target/home` since `.target` is already ignored? (and that might be a better place to store the `home` and we could name it something else instead)

zachlute:
Sure, that's fine. Feel free to replace this with a PR that changes the instructions to suggest people do that. As a potential contributor, I followed the instructions and had a bad experience, so I attempted to resolve that for others. I certainly don't want to argue about it.

Telling me how I could have avoided having that bad experience if I'd done a different thing than directed seems to kind of miss the point, doesn't it?

pickfire:
@zachlute I feel like you could be more persuasive to start another patch for this, sorry about giving out comments that was not anticipated. (I already have one patch but still waiting for reply) T_T

.gitignore
```diff
@@ -5,3 +5,4 @@
 /.idea/
 *~
 /**/target
+/home
```
joshtriplett: Please include a newline at the end of the file.

zachlute: I'm reasonably certain this is fixed now, but I squashed the commits for sanity's sake and the web interface makes it impossible to tell that anything was actually changed. But it looks right on a fresh local clone, so...

kinnison:
This looks good (and sensible) to me :+1:

PR #1495, opened 2018-09-01
Allow using inherited RUSTUP_UPDATE_ROOT variable in rustup-init.sh.

crlf0710:
This also makes the usage of the variable consistent with rust version by moving the "/dist" part to usage site.

For people in mainland China, Amazon web hosting and cdn services are inaccessible. Using this patch, we can set the `RUSTUP_UPDATE_ROOT` variable beforehand to ask `rustup-init.sh` to download rustup-init.exe from a mirror site.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -15,7 +15,9 @@
 
 set -u
 
-RUSTUP_UPDATE_ROOT="https://static.rust-lang.org/rustup/dist"
+if [ -z "$RUSTUP_UPDATE_ROOT" ]; then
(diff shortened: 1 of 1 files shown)


lucasjinreal:
Will this merge? this is really useful for those who can not access to official website

kinnison:
This change seems harmless and, given the comments about China, useful.  I think it'd be okay to merge.

nrc (approved):
Thanks! Sorry it took a while to merge.

PR #1504, opened 2018-09-14
Preserve symlinks when installing

tromey:
The lldb-preview component includes symlinks.  Currently, rustup
changes these to be regular files while installing.  This then causes
the resulting lldb to load liblldb twice, which causes a crash at
startup, because command line options are registered twice.  This
changes rustup to preserve symlinks when installing.


tromey:
I wasn't sure if I should base this on stable perhaps?

tromey:
@nrc would you mind reviewing this or forwarding it to someone else to review?  Fixing this is the next (hopefully final) step in getting lldb to work via rustup.

nrc:
LGTM!

PR #1511, opened 2018-09-19
Fix Windows job management

### LINKED ISSUE #1493: `cargo doc --open` broken on Windows due to rustup proxy
The `cargo doc --open` command is currently broken on nightly Cargo on Windows. The implementation of this command [recently changed](https://github.com/rust-lang/cargo/pull/5888) in Cargo. This new implementation is affected by a bug in rustup's job object code on Windows. Job objects are a Windows feature that rustup uses to implement the behavior such that when ctrl-c is pressed, child processes are also killed. The existing [job object code](https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup-cli/job.rs) was copied from Cargo. I recently [fixed](https://github.com/rust-lang/cargo/pull/5887) the job code in Cargo, which was also [causing issues](https://github.com/rust-lang/cargo/issues/5598).

I intended on sending a pull request to rustup to apply the same fix, but I ran into a problem: The fix relies on a destructor to run when rustup exits, but rustup uses `std::process::exit`, which doesn't run destructors. I can think of two solutions; either move the `std::process::exit` call(s) outwards so that the calls can run after the destructors, or use a mechanism other than a destructor to run the necessary code.

**Links**
My branch with the fixed job code: https://github.com/Seeker14491/rustup.rs/tree/job
Line in rustup where the object whose destructor needs to run is created: https://github.com/rust-lang-nursery/rustup.rs/blob/22b6cdb8b000f9217af1dd5dd4e73ce032bdb25d/src/rustup-cli/proxy_mode.rs#L14

Seeker14491:
Fixes #1493

### DIFF
--- a/src/rustup-cli/job.rs
+++ b/src/rustup-cli/job.rs
@@ -25,6 +25,9 @@ pub fn setup() -> Option<Setup> {
 
 #[cfg(unix)]
 mod imp {
+    use std::env;
+    use libc;
+
     pub type Setup = ();
 
     pub unsafe fn setup() -> Option<()> {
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks!

PR #1516, opened 2018-09-24
Ignore test `uninstall_self_delete_works` on Mac OS

nrc:
r? @alexcrichton (will push to stable too, if that's ok).

### DIFF
--- a/tests/cli-self-upd.rs
+++ b/tests/cli-self-upd.rs
@@ -254,6 +254,7 @@ fn uninstall_fails_if_not_installed() {
 // does the uninstall by actually invoking the installed binary in
 // order to test that it can successfully delete itself.
 #[test]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "rustup"
-version = "1.14.0"
+version = "1.14.1"
```
alexcrichton: Oh you can leave this at 1.14.0, we never published it originally so we'll just update the stable branch and push out the real 1.14.0 release

alexcrichton:
r=me with the same version (feel free to merge to both master and stable)

Cargo.toml
nrc: ok, cool

PR #1517, opened 2018-09-26
Add x86_64-unknown-linux-musl build support

### LINKED ISSUE #1053: Promote x86_64-unknown-linux-musl from target to host toolchain
Could the musl toolchain be promoted from something people access with

`rustup target add x86_64-unknown-linux-musl`

to something that people can

`curl https://sh.rustup.rs -sSf | sh -s -- --default-host x86_64-unknown-linux-musl -y`

?

martell:
Closes #1053

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=mips64-unknown-linux-gnuabi64   }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=mips64el-unknown-linux-gnuabi64 }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=s390x-unknown-linux-gnu         }
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=x86_64-unknown-linux-musl       }
+    - { os: linux, env: SKIP_TESTS=1 TARGET=x86_64-unknown-linux-musl        }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=arm-linux-androideabi           }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=armv7-linux-androideabi         }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=aarch64-linux-android           }
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ matrix:
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=mips64-unknown-linux-gnuabi64   }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=mips64el-unknown-linux-gnuabi64 }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=s390x-unknown-linux-gnu         }
+    - { <<: *linux, env: SKIP_TESTS=1 TARGET=x86_64-unknown-linux-musl       }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=arm-linux-androideabi           }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=armv7-linux-androideabi         }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=aarch64-linux-android           }
--- a/ci/docker/x86_64-unknown-linux-musl/Dockerfile
+++ b/ci/docker/x86_64-unknown-linux-musl/Dockerfile
@@ -1,0 +1,14 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get install -y \
+  musl-dev \
+  musl-tools \
+  curl \
+  ca-certificates \
+  perl \
+  make \
+  gcc
+
+
+RUN curl -L https://github.com/mozilla/sccache/releases/download/0.2.8/sccache-0.2.8-x86_64-unknown-linux-musl.tar.gz | tar xzf - \
+  && mv ./sccache-0.2.8-x86_64-unknown-linux-musl/sccache /usr/bin/sccache && rm -rf ./sccache-0.2.8-x86_64-unknown-linux-musl
--- a/ci/docker/x86_64-unknown-linux-musl/Dockerfile
+++ b/ci/docker/x86_64-unknown-linux-musl/Dockerfile
@@ -9,6 +9,5 @@ RUN apt-get update && apt-get install -y \
   make \
   gcc
 
-
 RUN curl -L https://github.com/mozilla/sccache/releases/download/0.2.8/sccache-0.2.8-x86_64-unknown-linux-musl.tar.gz | tar xzf - \
   && mv ./sccache-0.2.8-x86_64-unknown-linux-musl/sccache /usr/bin/sccache && rm -rf ./sccache-0.2.8-x86_64-unknown-linux-musl
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -96,6 +96,13 @@ static LIST_ENVS: &[&str] = &[
     "musl",
 ];
 
+// Linux hosts don't indicate clib in uname, however binaries only
+// run on boxes with the same clib, as expected.
+#[cfg(all(not(windows), not(target_env = "musl")))]
+const TRIPLE_X86_64_UNKNOWN_LINUX: &str = "x86_64-unknown-linux-gnu";
+#[cfg(all(not(windows), target_env = "musl"))]
+const TRIPLE_X86_64_UNKNOWN_LINUX: &str = "x86_64-unknown-linux-musl";
+
 // MIPS platforms don't indicate endianness in uname, however binaries only
 // run on boxes with the same endianness, as expected.
 // Hence we could distinguish between the variants with compile-time cfg()
@@ -175,7 +182,7 @@ impl TargetTriple {
                 (_, b"aarch64") if cfg!(target_os = "android") => Some("aarch64-linux-android"),
                 (_, b"i686") if cfg!(target_os = "android") => Some("i686-linux-android"),
                 (_, b"x86_64") if cfg!(target_os = "android") => Some("x86_64-linux-android"),
-                (b"Linux", b"x86_64") => Some("x86_64-unknown-linux-gnu"),
+                (b"Linux", b"x86_64") => Some(TRIPLE_X86_64_UNKNOWN_LINUX),
                 (b"Linux", b"i686") => Some("i686-unknown-linux-gnu"),
                 (b"Linux", b"mips") => Some(TRIPLE_MIPS_UNKNOWN_LINUX_GNU),
                 (b"Linux", b"mips64") => Some(TRIPLE_MIPS64_UNKNOWN_LINUX_GNUABI64),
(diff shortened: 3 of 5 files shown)


strfry:
@martell Please regard my additional commit 54b50f5d2454a9ba86eb3230c1a20cd0b0937590 in #1527 .
The Dockerfile must describe a normal glibc-distro with musl-capabilities for this build system to work.

What else is needed to make rustc appear in x86_64-unknown-linux-musl flavor?

martell:
Hey @strfry,

First, thanks for the help :)
I thought no one was interested in this but me.

> The Dockerfile must describe a normal glibc-distro with musl-capabilities for this build system to work.

EDIT: I see the error now.
bash does not come by default on alpine.
I'll update.
```
+docker run --entrypoint bash --user 2000:2000 --volume /home/travis/.rustup/toolchains/stable-x86_64-unknown-linux-gnu:/travis-rust:ro --volume /home/travis/build/rust-lang-nursery/rustup.rs:/src:ro --volume /home/travis/build/rust-lang-nursery/rustup.rs/target:/src/target --workdir /src --env TARGET=x86_64-unknown-linux-musl --env SKIP_TESTS=1 x86_64-unknown-linux-musl ci/run-docker.sh
container_linux.go:265: starting container process caused "exec: \"bash\": executable file not found in $PATH"
docker: Error response from daemon: oci runtime error: container_linux.go:265: starting container process caused "exec: \"bash\": executable file not found in $PATH".
ERRO[0000] error waiting for container: context canceled
```

> What else is needed to make rustc appear in x86_64-unknown-linux-musl flavor?

I'm not sure, I upstreamed a few commits from musl / alpine needed changes into llvm 5.0 which was 2 releases ago now in the hopes that later we could get rustup support for alpine.
I think for build systems making minimal alpine rust containers it would be better than using the package manager.

@alexcrichton could you weigh in with some direction/feedback?

Best,
Martell

strfry:
@martell ,

I started out with just adding bash (and then later rustc and cargo), but the way this is supposed to work is that the rustc is passed in through a Docker volume from the Travis host system. And that doesn't work when the container is a different architecture than the host distro.
What finally worked was to adapt one of the other Dockerfiles and install musl-tools a as the cross-compiler. The build will try call musl-gcc at some point, seems like someone already figured this out ;)


The other part i ment is getting rust itself to compile x86_64-musl as a host architecture, so that it's provided for rustup to download.
I opened rust-lang/rust#55163 and ran into some issues.

alexcrichton:
@martell I think it's probably not worth doing this until we get a musl host rustc working, otherwise this rustup won't have anything to download! When it works though it'll be easiest to compile a static rustup here, just using cross compilation from a glibc environment to `x86_64-unknown-linux-musl` which will produce a statically linked binary

kinnison:
@martell Do you think you want to keep this PR open for now, or should we close it off and perhaps just leave #1053 to track this?

dwijnand:
Seeing as #1053 is blocked on upstream https://github.com/rust-lang/rust/issues/57439, I think it makes sense to close this PR in the meantime.  Please note that a closed PR isn't a rejected PR, the work here can be reused and resubmitted.

martell:
Sorry for the delay, I'll close this for now until the host support has been resolved.

mati865:
IIUC this needs `cfg` similar to the one for MIPS:
https://github.com/rust-lang/rustup.rs/blob/819ddda6b7fc1f18884f027f646cd8714f8c4e9a/src/dist/dist.rs#L178

EDIT: https://github.com/rust-lang/rustup.rs/blob/master/rustup-init.sh would also need to detect if host is musl based. I tested it on Void Linux and Alpine Linux and checking output of `ldd --version` for `musl` looks reasonable:

```
$ ldd --version
musl libc (x86_64)
Version 1.1.21
Dynamic Program Loader
Usage: ldd [options] [--] pathname
```

martell:
I will add this over the weekend but @mati865 you should feel free to hijack this PR if you do not want to wait.
You finished up the rustc and cargo support along with the forge stuff and I think you deserve more that anyone to be the one who enables it in rustup :)

martell:
@alexcrichton soft ping

alexcrichton:
Ah oops thanks for the ping. I'm not sure the container will work here though since it's primarily used with glibc toolchains, although you can edit .travis.yml to see the failure on the PR. Could the build be configured to do a cross build from a glibc toolchain to a musl toolchain?

.travis.yml
```diff
@@ -31,6 +31,7 @@ matrix:
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=mips64-unknown-linux-gnuabi64   }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=mips64el-unknown-linux-gnuabi64 }
     - { <<: *linux, env: SKIP_TESTS=1 TARGET=s390x-unknown-linux-gnu         }
+    - { <<: *linux, env: SKIP_TESTS=1 TARGET=x86_64-unknown-linux-musl       }
```
rye: I was slightly confused at first by the `.travis.yml` file, but I think I understand what is meant.

While this PR has an inclusion to `.travis.yml` of a new TARGET to the matrix which _does_ get expanded with `if: branch != master`, for _PR builds_ on Travis, `branch` expands to the base branch for the PR.  Obviously, `master != master` is false, so this job isn't run and hence the PR is basically testing the functionality of what's on `master`.

So to actually test these changes prior to merging, the only change required would be to (temporarily) hand-code the matrix addition to always run? That might look something like:

```suggestion
    - { os: linux, env: SKIP_TESTS=1 TARGET=x86_64-unknown-linux-musl        }
```

(My suggestion would not be ideal to ship, but would illuminate any build failures.)

martell:
Following up here, sorry things have been slow.
Finally got some time this weekend to have another look.

@alexcrichton we appear to have an issue using a glibc distro like ubuntu cross building to the musl target.

instead of a binary called ldd we have musl-ldd in these environments.
using alpine or void linux directly is much easier to work with because of this.

```bash
/# ldd --version
ldd (Ubuntu GLIBC 2.27-3ubuntu1) 2.27
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.

/# musl-ldd --version
musl libc (x86_64)
Version 1.1.19
Dynamic Program Loader
Usage: musl-ldd [options] [--] pathname
```

cross compiling also results in this with the current build system.
`error: failed to run `rustc` to learn about target-specific information`
The travis build history with the various methods should reflect this.
cc @rye

martell:
Finally got some time this weekend to have another look.

@alexcrichton I believe this should be mergeable now.
When you approve I will drop 44ddbe9 from the PR

Thanks again for the suggestion @rye.

rye:
@martell, no problem, and thank you for doing this! It at least certainly seems that [#3396.2](https://travis-ci.com/rust-lang/rustup.rs/jobs/201283144) (musl build) was successful and did use the appropriate target! Can't wait to use rustup in my Alpine environments soonish. 😄

kinnison (commented):
I'm not in a position to review the musl-specific bits, but one comment I do have is regarding the docker machine.

@lzutao put a lot of effort into allowing us to reuse the Rust compiler docker images in order to ensure that we were matching libcs etc wherever possible.  Is there a suitable rust-compiler docker for musl to save the creation of the docker from ubuntu?

tesuji:
@kinnison Yes, Rustc has [`dist-x86_64-musl` dockerfile][1]. I haven't tested it yet
but it's worth a try.

[1]: https://github.com/rust-lang/rust/blob/master/src/ci/docker/dist-x86_64-musl/Dockerfile

kinnison:
@lzutao Do you think you could look at what's been done on this PR and then see if you can come up with a proposed patch to reuse the rustc docker if appropriate?

alexcrichton:
This all looks good to me, thanks @martell!

@kinnison FWIW musl is pretty special in that it doesn't require the same docker containers as other targets. The main difference is that the musl target is by default statically linked instead of having a dependency on a dynamically linked libc. For glibc platforms you need a special toolchain for targeting a really old glibc (which is our goal in rust-lang/rust and what rustup wants to mirror), but for musl there's no need since any toolchain can produce a workable binary!

In that sense this should be good to go IMO w/o docker changes.

martell:
Awesome, I'm happy to do any docker changes as a follow up if anyone wishes.
I did actually try `FROM rust-x86_64-unknown-linux-musl` but that did not exist.
We can create one if needed. Anyone have a specific issue they want to create for this?

@alexcrichton I do not have permission to merge.
I did however remove the ci testing commit as promised.
Can you merge?

src/dist/dist.rs
```diff
@@ -96,6 +96,13 @@ static LIST_ENVS: &[&str] = &[
     "musl",
 ];
 
+// Linux hosts don't indicate clib in uname, however binaries only
+// run on boxes with the same clib, as expected.
+#[cfg(all(not(windows), not(target_env = "musl")))]
+const TRIPLE_X86_64_UNKNOWN_LINUX: &'static str = "x86_64-unknown-linux-gnu";
```
tesuji: Remove `'static`. We don't need it anymore.

src/dist/dist.rs
```diff
@@ -96,6 +96,13 @@ static LIST_ENVS: &[&str] = &[
...
+#[cfg(all(not(windows), target_env = "musl"))]
+const TRIPLE_X86_64_UNKNOWN_LINUX: &'static str = "x86_64-unknown-linux-musl";
```
tesuji: Same

tesuji (commented):
The PR might still fail because of missing sccache.

ci/docker/x86_64-unknown-linux-musl/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get install -y \
+  musl-dev \
+  musl-tools \
+  curl \
+  ca-certificates \
+  perl \
+  make \
+  gcc
```
tesuji, line 10: Install sccache here: https://github.com/mozilla/sccache/releases/tag/0.2.8
Just download and extract it in PATH

alexcrichton:
Note that I'm personally just commenting here, I'm leaving the final review up to @kinnison

kinnison:
If there's no value in using the rustc docker for now, then I'll wait on the `sccache` comment responses before I make a decision.

martell:
@kinnison done :)
Same deal as before, once approved I will remove the travis change commit.
Ahh damn, I was a few hours late to make the next rustup release :/

tesuji:
@martell You could use [ci/docker/armv7-unknown-linux-gnueabihf/Dockerfile][1] as an example to 
install sccache.

[1]: https://github.com/rust-lang/rustup.rs/blob/master/ci/docker/armv7-unknown-linux-gnueabihf/Dockerfile

kinnison:
The appveyor failure is spurious (possibly due to too many parallel tests at once?)

tesuji:
I pushed some changes to [my branch](https://github.com/lzutao/rustup.rs/tree/musl-x86_64). You could cherry-pick it to your PR.

tesuji (changes requested):
Edit `ci/cloudfront-invalidation.txt` and add musl targets.

kinnison:
> Edit `ci/cloudfront-invalidation.txt` and add musl targets.

It's not entirely clear if the cloudfront invalidation would be necessary, but it can't hurt to have it amended.

tesuji:
Ok, I take it at #1882

martell:
Thanks @lzutao

PR #1526, opened 2018-10-11
Size optimizations: Build with LTO and alloc_system

### LINKED ISSUE #982: Consider distributing release + LTO binary
The current `rustup-init` binary is 9.3M in size, while it shouldn't be so big to bootstrap (you need to download 10M *before* even seeing a message.) It seems to contain debug symbols too.

Building with release resulted in 5.0M binary, with LTO it's 4.9M (doesn't make much sense). This halves download, and makes the experience better.

mbrubeck:
This reduces the size of the rustup-init binary by 36% (from 11.8 MB to 7.5 MB) when building in release mode on my Linux system with rustc 1.31.0-nightly (2bd5993ca 2018-10-02).

Fixes #982.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,3 +77,7 @@ test = false # no unit tests
 name = "rustup-init"
 path = "src/rustup-cli/main.rs"
 test = false # no unit tests
+
+[profile.release]
+lto = true
+codegen-units = 1
--- a/src/rustup-cli/main.rs
+++ b/src/rustup-cli/main.rs
@@ -14,11 +14,6 @@
 
 #![recursion_limit = "1024"]
 
-use std::alloc::System;
-
-#[global_allocator]
-static A: System = System;
-
 #[macro_use]
 extern crate error_chain;
 extern crate rustup_dist;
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -77,3 +77,7 @@ test = false # no unit tests
 name = "rustup-init"
 path = "src/rustup-cli/main.rs"
 test = false # no unit tests
+
+[profile.release]
+lto = true
+codegen-units = 1
```
alexcrichton, line 83: This can be removed because with LTO enable it has no effect on the final file size the build will otherwise be faster if this is not set to 1

```suggestion
```

src/rustup-cli/main.rs
```diff
@@ -14,6 +14,11 @@
 
 #![recursion_limit = "1024"]
 
+use std::alloc::System;
+
+#[global_allocator]
```
alexcrichton: Could a comment be added here as to why this is done?

Cargo.toml
mbrubeck: Empirically, removing this line causes the size of the binary to increase by 644 KB (8.4%).

(This has consistently been my experience with LTO and codegen-units, so I think we need to re-examine some assumptions about LTO.  It's not a perfect fix for deoptimization from codegen-units.)

alexcrichton: Egad sorry for the delay here. This sounds fine by me though.

PR #1541, opened 2018-11-14
WIP: Profiles

nrc:


mati865:
I'm unable to try it because `rustup` depends on too old `openssl` crates but does this PR allow removing already installed components (`rust-docs`)?

nrc:
> does this PR allow removing already installed components

It doesn't work yet, but it will

PR #1547, opened 2018-11-20
Improve formatting of longer download times in download tracker

Manishearth:
Someone on twitter complained that on slow connections Rustup gives progress values like `1234 m 56 s`, which is kinda annoying to look at.

I changed it so that past 100 minutes we start counting hours, and past 2 days we start counting days. The days bit may not be necessary.

The cutoff values were taken from wget.

### DIFF
--- a/src/rustup-cli/download_tracker.rs
+++ b/src/rustup-cli/download_tracker.rs
@@ -176,10 +176,25 @@ impl fmt::Display for HumanReadable {
 
             if sec.is_infinite() {
                 write!(f, "Unknown")
-            } else if sec > 1e3 {
+            } else if sec > 48. * 3600. {
                 let sec = self.0 as u64;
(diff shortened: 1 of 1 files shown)


kinnison:
I worry that this code is becoming really quite intricate and although the change looks right, it's super-hard to track all the numbers and be certain.

I have, however, spent time looking, and can't find an existing crate which might provide this formatting for us, so I'd begrudgingly :+1: this.

Would you mind rebasing it onto current master and checking the formatting (We've run rustfmt across the codebase and while github says there's no conflicts, things might not be entirely the same)  Also that'd have the added benefit of re-kicking the CI checks since rustup.rs has moved since you supplied this patch.

Thanks,

D.

Manishearth:
The numbers are copied from wget so they should be correct.

Issue #1546, opened 2018-11-19
Gracefully handle renamed components when updating toolchains


alexcrichton:
In some local testing when we rename a component (for example from `rls-preview` to `rls`) then if the component is previously installed whenever you update it'll uninstall `rls-preview` but fail to install `rls` shortly afterwards.

This is likely a local regression in rustup from some time ago, but it's unsure where just yet!

Once this is fixed and propagated enough in rustup then we can switch the actual names in the released manifests from `rls-preview` to `rls` and such.

alexcrichton:
cc @nrc

kinnison:
Does this need to remain open or was it resolved by #1549 ?  If it's not resolved then what remains to-do?

kinnison:
I'm assuming this is all done.

PR #1549, opened 2018-11-21
Change handling of renames

nrc:
This PR makes handling of renamed components more flexible by allowing components to be added or removed based on either the 'before' or 'after' names. We always use the 'before' name when communicating with the user, but use the 'after' name internally (basically just in the manifest).

This is a backwards-compatible stop-gap. It means that for the edition we can talk about 'rls' rather than 'rls-preview'. (Assuming renames are from 'rls' to 'rls-preview', which is the current situation in the manifest).

The final commit adds an instruction for adding missing components to rustup's proxies.

Since Rustup is not distributed with Rust, this does not need to be part of the edition, timing-wise. However, it would be good to have this in the wild and tested somewhat so that new users will get the benefits when the edition is released.

r? @alexcrichton

### DIFF
--- a/src/rustup-cli/common.rs
+++ b/src/rustup-cli/common.rs
@@ -284,7 +284,7 @@ pub fn rustc_version(toolchain: &Toolchain) -> String {
 pub fn list_targets(toolchain: &Toolchain) -> Result<()> {
     let mut t = term2::stdout();
     for component in toolchain.list_components()? {
-        if component.component.pkg == "rust-std" {
+        if component.component.name_in_manifest() == "rust-std" {
             let target = component
                 .component
                 .target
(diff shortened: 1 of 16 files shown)


nrc:
Closes #1546

alexcrichton:
I'll be honest in that I don't really have a great handle on what's going on here, but this all looks largely like innocuous changes passing more parameters around and/or fallout from tweaking interfaces. In that sense the guts of the patch look good to me and I think we can probably go ahead and land.

I think there's some test failures on CI though? Otherwise r=me.

As one other thought, though, this is trying to use the "from" name in the renames to display names, right? Shouldn't it use the "to" names eventually? I think with this patch we'd start saying `rustfmt`, but then whenever we actually switch the package names it'll say `rustfmt-preview`, right?

nrc:
> Shouldn't it use the "to" names eventually?

We'll need more rustup changes once we want to switch again. We might just live with 'from' and 'to' being the wrong way around forever. But I don't think we can change this backwards compatibly.

alexcrichton:
Mk, I'm fine landing this!

PR #1552, opened 2018-11-27
Use `renames` instead of `rename` to match the actual manifest

nrc:
Kind of worrying this didn't make any difference for a year. It would explain at least why renames had some unexpected behaviour which was not picked up by the tests, renames IRL have always been ignored.

r? @alexcrichton


alexcrichton:
@bors: r+

Uh...

PR #1553, opened 2018-11-27
Return Ok status when trying to add required component

mati865:
Fixes https://github.com/rust-lang/rustup.rs/issues/441


nrc (approved):
Thanks!

PR #1554, opened 2018-11-28
More tweeks to renames

nrc:
r? @alexcrichton 

This is mostly undoing some of the changes from previous PRs (with some renames). I had mistakenly made the tests unlike the actual packages downloaded, then made the code correct for the tests not real life. The actual meaningful change here is in src/rustup-dist/src/manifestation.rs, where we use the name of the component in the manifest to refer the name of a component in a package (where it matches the manifest, not the user-visible name).

### DIFF
--- a/src/rustup-cli/common.rs
+++ b/src/rustup-cli/common.rs
@@ -284,7 +284,7 @@ pub fn rustc_version(toolchain: &Toolchain) -> String {
 pub fn list_targets(toolchain: &Toolchain) -> Result<()> {
     let mut t = term2::stdout();
     for component in toolchain.list_components()? {
-        if component.component.name_in_manifest() == "rust-std" {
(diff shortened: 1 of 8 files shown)


alexcrichton:
Oh dear.. so the tests don't seem to be covering all these cases that are coming up? Is this something we are still confident in releasing just a week before the edition?

nrc:
> so the tests don't seem to be covering all these cases that are coming up?

There is test coverage, but it relies on the mocking matching the real life manifests, which is a non-trivial issue because the mocking is so complex.

> Is this something we are still confident in releasing just a week before the edition?

My confidence is certainly less than 100%. I do think we should be able to either roll-back or fix rustup if we get bug reports in the next week. We usually do get lots of usage.

alexcrichton:
@nrc looks like you've got the `stable` branch management down, so whenever it's ready just ping me and I can do the release

Issue #745, opened 2016-10-04
Wrong `default-host` leads to panic


Boddlnagg:
Setting something like `rustup set default-host x64_86-msvc` (on Windows) leads to the following error on `rustup show` or `rustup install`:

```
thread 'main' panicked at 'host triple couldn't be converted to partial triple', ../src/libcore\option.rs:700
```

I think it shouldn't accept an invalid `default-host` in the first place ...

lazpeng:
Also happens with the initial installation, selecting a custom install and passing an "invalid" triple such as "stable-x86_64-pc-windows-gnu" also leads to panic

Bluejanis:
That wrong default host didnt even let me install rustup correctly.
I had to delete the directory and set it to gnu manually for it to work.

kinnison:
Hi, I have begun looking at this since it does seem to continue to be a problem.  I've opened PR #1578 and I would welcome any advice on there of where to add the checks needed to *cleanly* report this early and usefully to the user.

rbtcollins:
I believe https://github.com/rust-lang/rustup/pull/1578 fixed this; If a similar issue occurs again please open a new ticket.

Issue #874, opened 2016-12-20
./rustup-init.exe --default-host msvc -- panics


alexcrichton:
After executing `./rustup-init.exe --default-host msvc` it prints out info like:

```
...

Current installation options:

   default host triple: msvc
     default toolchain: stable
  modify PATH variable: yes

1) Proceed with installation (default)
2) Customize installation
3) Cancel installation

...
```

When I confirm that, though, I'm met with:

```
thread 'main' panicked at '', ../src/libcore\option.rs:715
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

canfoderiskii:
It seems the value for argument `--default-host` should be `x86_64-pc-windows-msvc` or `i686-pc-windows-msvc` .

Diggsey:
Dupe of #745

PR #1578, opened 2018-12-14
Deal cleanly with malformed default-host

kinnison:
It is better to report errors cleanly rather than panic.  If the
default-host is set badly then we can end up panicking because
we're unable to parse it.  Since default-host is user input, it's
better we don't panic.

This PR will, when complete, add in some appropriate checks for default-host
and hopefully lead the user to resolution of the problem resulting in fixing #745

So far, all I've done is to change the `.expect()` into `.ok_or()` and propagate the `Result<...>`
properly.  This removes the panic but means that we get some very odd reports from things
like `rustup show`, hence the plan to add some proper earlier checks which can report errors more usefully.

If this approach seems sane, I'd appreciate any pointers possible on where to add appropriate checks, and in addition, where to add appropriate tests.

### DIFF
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -223,6 +223,7 @@ fn canonical_cargo_home() -> Result<String> {
 /// and adding `CARGO_HOME`/bin to PATH.
 pub fn install(no_prompt: bool, verbose: bool, mut opts: InstallOpts) -> Result<()> {
     do_pre_install_sanity_checks()?;
+    do_pre_install_options_sanity_checks(&opts)?;
     check_existence_of_rustc_or_cargo_in_path(no_prompt)?;
     do_anti_sudo_check(no_prompt)?;
 
@@ -460,6 +461,30 @@ fn do_pre_install_sanity_checks() -> Result<()> {
         return Err("cannot install while rustup.sh is installed".into());
     }
 
+    Ok(())
+}
+
+fn do_pre_install_options_sanity_checks(opts: &InstallOpts) -> Result<()> {
+    // Verify that the installation options are vaguely sane
+    (|| {
+        let host_triple = dist::TargetTriple::from_str(&opts.default_host_triple);
+        let partial_channel = dist::PartialToolchainDesc::from_str(&opts.default_toolchain)?;
+        let resolved = partial_channel.resolve(&host_triple)?.to_string();
+        debug!(
+            "Successfully resolved installation toolchain as: {}",
+            resolved
+        );
+        Ok(())
+    })()
+    .map_err(|e: Box<std::error::Error>| {
+        format!(
+            "Pre-checks for host and toolchain failed: {}\n\
+             If you are unsure of suitable values, the 'stable' toolchain is the default.\n\
+             Valid host triples look something like: {}",
+            e,
+            dist::TargetTriple::from_host_or_build()
+        )
+    })?;
     Ok(())
 }
 
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -454,7 +454,7 @@ fn do_pre_install_sanity_checks() -> Result<()> {
             "delete `{}` to remove rustup.sh",
             rustup_sh_path.expect("").display()
         );
-        warn!("or, if you already rustup installed, you can run");
+        warn!("or, if you already have rustup installed, you can run");
         warn!("`rustup self update` and `rustup toolchain list` to upgrade");
         warn!("your directory structure");
         return Err("cannot install while rustup.sh is installed".into());
--- a/src/rustup-cli/self_update.rs
+++ b/src/rustup-cli/self_update.rs
@@ -468,7 +468,12 @@ fn do_pre_install_options_sanity_checks(opts: &InstallOpts) -> Result<()> {
     // Verify that the installation options are vaguely sane
     (|| {
         let host_triple = dist::TargetTriple::from_str(&opts.default_host_triple);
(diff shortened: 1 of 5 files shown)


kinnison:
At this point, I've added a `debug!()` macro, fixed a minor typo in one of the sanity check warnings, and added an explicit sanity check on the host and toolchain inputs to the installer.  This combo should handle the situation mentioned in #874 by @alexcrichton too.

My next steps will be to try and improve the situation by suggesting the build triple and the 'stable' channel if these errors come up.

kinnison:
This PR will need further work as #1583 has been merged and #1585 may get merged soon.  Also there'll be a PR about clippy lint fixes which might happen before this too.  I'll get back to it, I promise.

kinnison:
I've un-WIPped this because I finally added a commit which addresses the direct bad action in #745 

I also rebased, so hopefully this'll be good.

src/rustup-cli/self_update.rs
```diff
@@ -454,12 +454,37 @@ fn do_pre_install_sanity_checks() -> Result<()> {
...
+    })()
+    .map_err(|e: Box<std::error::Error>| {
+        format!(
+            "Pre-checks for host and toolchain failed: {}\n\
+             If you are unsure of suitable values, the 'stable' toolchain is the default.\n\
+             Valid host triples look something like: {}",
+            e,
+            dist::TargetTriple::from_host_or_build().as_ref()
+        )
+    })?;
```
nrc, line 492: I think we should move this checking to its own function - `do_pre_install_sanity_checks` is about checking the user's system, rather than the installation options

src/rustup-dist/src/dist.rs
```diff
@@ -210,6 +210,12 @@ impl TargetTriple {
     }
 }
 
+impl AsRef<str> for TargetTriple {
```
nrc: I don't think we should use `AsRef` for this as it is doing a type conversion, not just a ref/deref

src/rustup-cli/self_update.rs
```diff
@@ -454,12 +454,37 @@ fn do_pre_install_sanity_checks() -> Result<()> {
...
     }
 
+    // Verify that the installation options are vaguely sane
+    (|| {
+        let host_triple = dist::TargetTriple::from_str(&opts.default_host_triple);
+        let toolchain_to_use = if opts.default_toolchain == "none" {
+            "stable"
+        } else {
+            &opts.default_toolchain
+        };
```
nrc, line 475: This seems like it is duplicating behaviour defined elsewhere.

src/rustup/config.rs
```diff
@@ -111,7 +111,15 @@ impl Cfg {
             env_override: env_override,
             dist_root_url: dist_root,
             dist_root_server: dist_root_server,
-        })
+        };
+
+        // Run some basic checks against the constructed configuration
+        // For now, that means simply checking that 'stable' can resolve
+        // for the current configuration.
+        cfg.resolve_toolchain("stable")
```
nrc, line 119: I don't understand why we need this check. Could you explain please?

src/rustup/config.rs
```diff
@@ -470,7 +478,20 @@ impl Cfg {
     }
 
     pub fn set_default_host_triple(&self, host_triple: &str) -> Result<()> {
-        if dist::PartialTargetTriple::from_str(host_triple).is_none() {
+        if let Some(host) = dist::PartialTargetTriple::from_str(host_triple) {
```
nrc: Seems like these checks are duplicating the checks done in `resolve`?

nrc (commented):
Thanks for the PR, there are some comments inline

src/rustup-dist/src/dist.rs
kinnison: Fair.  Also there's already a Display so I've dropped this particular commit from the PR

src/rustup/config.rs
kinnison: Sure.  By doing this check here we verify that later attempts to resolve the toolchain are more likely to succeed.  It means we report the error about bad host types etc earlier in the sequence of operations.  This leads to, in my opinion, a slightly better user experience.

src/rustup/config.rs
kinnison: Yeah, for some reason I was entirely blind to how to reuse `PartialToolchainDesc::resolve()` when I wrote this check.  I've reworked it now.

src/rustup-cli/self_update.rs
kinnison: Okay, I've created a function called `do_pre_install_options_sanity_checks()` and moved the content into there, though for now I'm calling it from the same spot as `do_pre_install_sanity_checks()` unless I find something better.

src/rustup-cli/self_update.rs
kinnison: To my knowledge it's not per-se.  This is simply because using PartialToolchainDesc::resolve needs a valid toolchain name or it won't work (and none isn't valid IIRC).

It looks similar to other checks, but this time we're resolving from options and don't have a `Cfg` struct so I can't just call `set_default_host` et al, and deal with things that way.  Also the `debug!()` is meant to be useful to tell us that it's behaving right and picking the correct fully qualified toolchain name.

If I've missed something you know of, please do let me know, but I've had a fairly good grep around I think.

kinnison (commented):
Thanks for your review @nrc - I've responded to everything I think, and I'll push a new set of commits momentarily

src/rustup/config.rs
nrc: Thanks for the explanation. I wonder if this will ever give a false negatve? If we have some toolchain which is nightly-only perhaps?

nrc:
Looks good, thanks!

PR #1582, opened 2018-12-18
Fix various ShellCheck warnings

tesuji:
From [`shellcheck`](https://www.shellcheck.net/) output:

```command line
In ./rustup-init line 65:
    local _dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup)"
          ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 143:
    local _current_exe_head=$(head -c 5 /proc/self/exe )
          ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 162:
    local _current_exe_endianness="$(head -c 6 /proc/self/exe | tail -c 1)"
          ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 174:
    local _ostype="$(uname -s)"
          ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 175:
    local _cputype="$(uname -m)"
          ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 183:
    if [ "$_ostype" = Darwin -a "$_cputype" = i386 ]; then
                             ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.


In ./rustup-init line 266:
            local _cputype="$(get_endianness $_cputype "" 'el')"
                  ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 270:
            local _bitness="$(get_bitness)"
                  ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 271:
            if [ $_bitness = "32" ]; then
                 ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./rustup-init line 282:
            local _cputype="$(get_endianness $_cputype "" 'el')"
                  ^-- SC2155: Declare and assign separately to avoid masking return values.


In ./rustup-init line 303:
    if [ $_ostype = unknown-linux-gnu -a $_cputype = x86_64 ]; then
                                      ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.


In ./rustup-init line 312:
    if [ $_ostype = "unknown-linux-gnueabihf" -a $_cputype = armv7 ]; then
                                              ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.


In ./rustup-init line 345:
    if [ $? != 0 ]; then err "$1"; fi
         ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
```

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -312,7 +312,7 @@ get_architecture() {
     fi
 
     # Detect 64-bit linux with 32-bit userland for powerpc
-    if [ $_ostype = unknown-linux-gnu -a $_cputype = powerpc64 ]; then
+    if [ $_ostype = unknown-linux-gnu ] && [ $_cputype = powerpc64 ]; then
         if [ "$(get_bitness)" = "32" ]; then
             local _cputype=powerpc
         fi
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -273,7 +273,7 @@ get_architecture() {
 
         mips64)
             _bitness="$(get_bitness)"
-            if [ "$_bitness" -eq "32" ]; then
+            if [ "$_bitness" -eq 32 ]; then
                 if [ "$_ostype" = "unknown-linux-gnu" ]; then
                     # 64-bit kernel with 32-bit userland
                     # endianness suffix is appended later
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -62,10 +62,11 @@ main() {
             ;;
     esac
 
-    local _url="$RUSTUP_UPDATE_ROOT/dist/$_arch/rustup-init$_ext"
+    local _url="${RUSTUP_UPDATE_ROOT}/dist/${_arch}/rustup-init${_ext}"
 
-    local _dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup)"
-    local _file="$_dir/rustup-init$_ext"
+    local _dir
+    _dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup)"
+    local _file="${_dir}/rustup-init${_ext}"
 
     local _ansi_escapes_are_valid=false
     if [ -t 2 ]; then
@@ -106,7 +107,7 @@ main() {
     ensure chmod u+x "$_file"
     if [ ! -x "$_file" ]; then
         printf '%s\n' "Cannot execute $_file (likely because of mounting /tmp as noexec)." 1>&2
-        printf '%s\n' "Please copy the file to a location where you can execute binaries and run ./rustup-init$_ext." 1>&2
+        printf '%s\n' "Please copy the file to a location where you can execute binaries and run ./rustup-init${_ext}." 1>&2
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


rustup-init.sh
```diff
@@ -316,20 +317,20 @@ get_architecture() {
...
+    if [ "$_ostype" = "unknown-linux-gnueabihf" ] && [ "$_cputype" = armv7 ]; then
         if ensure grep '^Features' /proc/cpuinfo | grep -q -v neon; then
             # At least one processor does not have NEON.
-            local _cputype=arm
+            _cputype=arm
         fi
     fi
 
-    local _arch="$_cputype-$_ostype"
+    _arch="${_cputype}-${_ostype}"
```
ehuss, line 331: I think this converts `_arch` to a global.

rustup-init.sh
```diff
@@ -224,85 +226,84 @@ get_architecture() {
...
         mips)
-            local _cputype="$(get_endianness $_cputype "" 'el')"
+            _cputype="$(get_endianness $_cputype "" 'el')"
             ;;
 
         mips64)
-            local _bitness="$(get_bitness)"
-            if [ $_bitness = "32" ]; then
-                if [ $_ostype = "unknown-linux-gnu" ]; then
+            _bitness="$(get_bitness)"
```
ehuss, line 275: This converts `_bitness` to a global.

ehuss (commented):
I don't feel like this fixes anything specific, but otherwise don't see any other issues.

nrc:
Thanks for the review @ehuss, this can be merged with @ehuss 's points addressed

rustup-init.sh
tesuji: Thanks for reviewing @ehuss. Because we have defined `_arch` as local variable, we don't need to use `local` keyword when assigning after that. You could check that with this script:
```sh
func ()
{
  local loc_var=23       # Declared as local variable.
  echo                   # Uses the 'local' builtin.
  echo "\"loc_var\" in function = $loc_var"
  global_var=999         # Not declared as local.
                         # Therefore, defaults to global.
  echo "\"global_var\" in function = $global_var"
  loc_var=666            # No keyword local
}

func

# Now, to see if local variable "loc_var" exists outside the function.

echo
echo "\"loc_var\" outside function = $loc_var"
                                      # $loc_var outside function =
                                      # No, $loc_var not visible globally.
echo "\"global_var\" outside function = $global_var"
                                      # $global_var outside function = 999
                                      # $global_var is visible globally.
echo

exit 0
#  In contrast to C, a Bash variable declared inside a function
#+ is local ONLY if declared as such.
```
Result:
```command line
$ sh  ./local.sh 

"loc_var" in function = 23
"global_var" in function = 999

"loc_var" outside function = 
"global_var" outside function = 999
```

rustup-init.sh
tesuji: As I explained above, this doesn't covert `_bitness` to a global.

nrc:
Thanks!

rustup-init.sh
ehuss: > Because we have defined `_arch` as local variable

`_arch` is only defined as local in `main`.  It is assigned in `get_architecture`.  Using your example:

```sh
func ()
{
  func2
  echo "\"loc_var\" in function = $loc_var"
  local loc_var=23       # Declared as local variable.
  echo "\"loc_var\" in function = $loc_var"
}

func2 ()
{
  loc_var=42
  echo "\"loc_var\" in function2 = $loc_var"
}

func
# Notice `loc_var` is now global.
echo "\"loc_var\" outside function = $loc_var"

```

rustup-init.sh
ehuss: `_bitness` is never set to local anywhere.

rustup-init.sh
tesuji: I hope resolve this issue in #1604.

PR #1583, opened 2018-12-18
Run basic 2018 edition fix

kinnison:
In order to use 2018 features we need the edition to be updated.
This commit is the result of running:

cargo fix --edition -p rustup -p rustup-dist -p rustup-mock -p rustup-utils -p download

And then adding 'edition = "2018"' to each requisite Cargo.toml

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 
 name = "rustup"
 version = "1.16.0"
+edition = "2018"
 authors = [ "Diggory Blake <diggsey@googlemail.com>" ]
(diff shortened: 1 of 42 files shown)


kinnison:
As a followup if this is accepted, I intend to sort out a cargo fmt across the workspace.  If anyone knows of a good way to mechanically do this to preserve as much authorship information as possible, I'd love to know, otherwise it'll just smear my name across an awful lot of the codebase :(

nrc:
Thank you!

PR #1585, opened 2018-12-18
Reformat the entire codebase using `cargo fmt`

kinnison:
To make sure that people using editors which use rustfmt by default
on saving don't touch more lines than intended when working on the
rustup codebase, we'll take the hit to format it all once.

Please note, this depends on #1583 (as it effectively contains that commit also)

### DIFF
--- a/src/download/src/lib.rs
+++ b/src/download/src/lib.rs
@@ -10,8 +10,8 @@ extern crate lazy_static;
 #[cfg(feature = "reqwest-backend")]
 extern crate reqwest;
 
-use url::Url;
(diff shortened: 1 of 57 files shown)


nrc:
Did you do any manual fixups or is this all just running `cargo fmt`?

kinnison:
@nrc The formatting commit was pure `cargo fmt` and no manual changes.

kinnison:
All I did was rebase against current master and confirm there were no further formatting changes needed.

nrc (approved):
Thanks!

PR #1588, opened 2018-12-20
Update to use `dirs::home_dir()`

kinnison:
As of Rust 1.29.0 the `std::env::home_dir()` function is deprecated and the compiler refers us to `dirs::home_dir()` instead.

I've had a look and I think it makes sense.  In addition it definitely supports redox which might be of interest later.

There are two commits - the code change, and the update to Cargo.lock

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,6 +28,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "argon2rs"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "blake2-rfc 0.2.18 (registry+https://github.com/rust-lang/crates.io-index)",
+ "scoped_threadpool 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "arrayref"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


kinnison:
Hrm, looks like a windows related issue.  I really will have to up-my-game and sort out a windows dev box of some kind

kinnison:
Let's give that a go.  This time I unified use of `home_dir()` through rustup-utils and added `dirs` for unix.

kinnison:
For reference, this is built on top of #1585 so merging this would include that formatting run.  if you'd rather I rebase without that, just yell.

nrc (approved):
Thanks! Looks good but could you rebase please and merge the last two commits.

alexcrichton:
If we're going to land this I'd prefer to hold off, the home directory handling in rustup is very specific and finnicky, and the deprecation is to a function that behaves differently on some platforms. I'm worried this will cause breakage, so I'd like to dig in a bit more before merging.

kinnison:
@nrc I will do so, but as @alexcrichton says, we should hold off until he can confirm this is safe.  I believe it is, since it leaves rustup's Windows handling in place (which is, so far as I can tell) the finicky bit, though I am not a Windows expert and don't know the exact details behind the choices.

I'll do the rebase at some point soon, but I'm content to wait for Alex to think and either okay or deny the change :-)

alexcrichton:
Ok this looks good to me, thanks!

Issue #1557, opened 2018-12-01
Add progress bar for "installing component" step


johnthagen:
Some components, such as `rust-docs` and `rust-src` take a bit longer to install, even on SSD-equipped machines. These take even longer on Windows (#1540).

One thing that would help in these situations would be if the user could see the progress of the component install to:

1. Get a sense for how long the component will take to finish installing
2. Know that `rustup` isn't just hung up

Perhaps Cargo's [new progress bar](https://github.com/rust-lang/cargo/pull/5995/) released in stable 1.30 could be reused in some way?

```
...
info: installing component 'rust-docs'
Installing [============>                                           ] 330/1400
```

johnthagen:
At least one user reported thinking that `rustup` was stuck when it got to installing `rust-docs`: https://github.com/rust-lang/rustup.rs/issues/1458#issuecomment-405044857

PR #1593, opened 2018-12-27
rustup-dist: Use Download notifications to track install

kinnison:
People have requested some indication of progress for long-running
install steps.  This commit re-uses the download tracker logic to
provide a progress bar (the length is the compressed component
tarball but should be good enough) to provide such feedback.

Fixes: #1557

### DIFF
--- a/src/rustup-dist/src/manifestation.rs
+++ b/src/rustup-dist/src/manifestation.rs
@@ -200,13 +200,18 @@ impl Manifestation {
 
             let gz;
             let xz;
+            let notification_converter = |notification: rustup_utils::Notification| {
+                notify_handler(Notification::Utils(notification));
+            };
+            let reader =
+                utils::FileReaderWithProgress::new_file(&installer_file, &notification_converter)?;
             let package: &Package = match format {
                 Format::Gz => {
-                    gz = TarGzPackage::new_file(&installer_file, temp_cfg)?;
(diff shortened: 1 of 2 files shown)


kinnison:
All the rebase does is remove one of the fixes messages because it was over-enthusiastic.

nrc (commented):
Thanks for the PR, I think this is probably a good idea. I would move the reader from the manifestation module to somewhere else though, maybe the dist::download module or a utils module somewhere

kinnison:
I will look at the right place to move it to, and attempt to get to it later when I'm home for the evening.  I worry that the callback nesting might take me a while to get right though, so it might be a day or two before I can post depending on how awake I am when I get home.  Thanks for the confirmation that the idea itself is okay though.

kinnison:
@nrc turns out that the FileReaderWithProgress moved nicely to `rustup_utils` but I'm not entirely clear on how I might add tests for this.  Obviously I've run it by hand and am confident it does what I wanted, but do you feel it needs something under `cargo test` ?

kinnison:
Interestingly there's some formatting differences in the CI.  I'll have to double-check my rustfmt install.

kinnison:
I've rebased and hopefully that'll clear up that formatting issue.

nrc:
Thanks!

PR #1595, opened 2018-12-30
Add more detail error messages when installing with some components has failed.

h-michael:
Error message that we get when installing with some components failed may seem unkind for beginners.
Many users create issue at component's repository.
For example,
https://github.com/rust-lang/rls/issues/1186
https://github.com/rust-lang/rls/issues/641
https://github.com/rust-lang/rustfmt/issues/3244
https://github.com/rust-lang/rust-clippy/issues/2869

So I might want to add more information.
(But there may be more suitable message.)

### DIFF
--- a/src/rustup-dist/src/errors.rs
+++ b/src/rustup-dist/src/errors.rs
@@ -5,6 +5,10 @@ use std::io::{self, Write};
 use std::path::PathBuf;
 use toml;
 
+pub const TOOLSTATE_MSG: &str = "if you require these components, please install and use the latest successful build version, \
+                                which you can find at https://rust-lang-nursery.github.io/rust-toolstate, for example.\n\
+                                rustup install nightly-2018-12-27";
+
 error_chain! {
     links {
         Utils(rustup_utils::Error, rustup_utils::ErrorKind);
@@ -127,11 +131,21 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest) -> String {
         if same_target {
             let mut cs_strs = cs.iter().map(|c| format!("'{}'", c.short_name(manifest)));
             let cs_str = cs_strs.join(", ");
-            let _ = write!(buf, "some components unavailable for download: {}", cs_str);
+            let _ = write!(
+                    buf,
+                    "some components unavailable for download: {}\n{}",
+                    cs_str,
+                    TOOLSTATE_MSG,
+                );
         } else {
             let mut cs_strs = cs.iter().map(|c| c.description(manifest));
             let cs_str = cs_strs.join(", ");
-            let _ = write!(buf, "some components unavailable for download: {}", cs_str);
+            let _ = write!(
+                    buf,
+                    "some components unavailable for download: {}\n{}",
+                    cs_str,
+                    TOOLSTATE_MSG,
+                );
         }
     }
 
--- a/src/rustup-dist/src/errors.rs
+++ b/src/rustup-dist/src/errors.rs
@@ -5,9 +5,10 @@ use std::io::{self, Write};
 use std::path::PathBuf;
 use toml;
 
-pub const TOOLSTATE_MSG: &str = "if you require these components, please install and use the latest successful build version, \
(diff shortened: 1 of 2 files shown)


src/rustup-dist/src/errors.rs
```diff
@@ -127,11 +127,25 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest) -> String {
         if same_target {
             let mut cs_strs = cs.iter().map(|c| format!("'{}'", c.short_name(manifest)));
             let cs_str = cs_strs.join(", ");
-            let _ = write!(buf, "some components unavailable for download: {}", cs_str);
+            let _ = write!(
+                    buf,
+                    "some components unavailable for download: {}\n\
+                    if you want these components, please install and use latest successful build version, \
```
kinnison: You're missing a 'the' between 'use' and 'latest'

src/rustup-dist/src/errors.rs
```diff
@@ -127,11 +127,25 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest) -> String {
...
+                    which you can find out at https://rust-lang-nursery.github.io/rust-toolstate, like this.\n\
```
kinnison: I'd drop the 'out' replace the ',' with '.' and change 'like this.' to '\nTo do this, run something like:\n'

src/rustup-dist/src/errors.rs
```diff
@@ -127,11 +127,25 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest) -> String {
...
+                    rustup install nightly-2018-12-27",
+                    cs_str
+                );
```
kinnison: Same comments apply here

tests/cli-misc.rs
```diff
@@ -808,7 +808,10 @@ fn update_unavailable_rustc() {
         expect_err(
             config,
             &["rustup", "update", "nightly"],
-            "some components unavailable for download: 'rustc', 'cargo'",
+            "some components unavailable for download: 'rustc', 'cargo'\n\
+            if you want these components, please install and use latest successful build version, \
+            which you can find out at https://rust-lang-nursery.github.io/rust-toolstate, like this.\n\
+            rustup install nightly-2018-12-27",
```
kinnison: Same comments apply here.

kinnison (commented):
Over-all the concept is probably good, though I worry about directing people to the rust-toolstate page directly.

I wonder if we should consider different messages depending on the channel selected.  If this is nightly, the message you suggest would be suitable, but if this is stable or beta, that suggests something less good going on which might benefit from a different kind of message.

src/rustup-dist/src/errors.rs
nrc:
```suggestion
                    if you require these components, please install and use the latest successful build, \
```

src/rustup-dist/src/errors.rs
nrc:
```suggestion
                    which you can find at https://rust-lang-nursery.github.io/rust-toolstate, for example.\n\
```

src/rustup-dist/src/errors.rs
nrc: Could you pull the message out into a const please

nrc (approved):
lgtm with the requested changes

h-michael:
@nrc, @kinnison thanks for your review.

> Could you pull the message out into a const please

How can I use write! macro with const literal ?

nrc:
I was thinking you would use something like `write!(buf, "some components unavailable for download: {}\n{}", ..., TOOLSTATE_MSG)` where `TOOLSTATE_MSG` is the const).

h-michael:
@nrc I see, thanks!

h-michael:
@nrc I've fixed.
But test is failed. This may be caused by #1547.

h-michael:
@nrc Fixed and passed CI

nrc:
Thanks!

PR #1597, opened 2019-01-02
Support to open more documents directly in `rustup doc`

king6cong:
e.g. `rustup doc --nomicon`
Documents supported:
alloc,book,cargo,core,edition-guide,nomicon,proc_macro,reference,rust-by-example,rustc,rustdoc,std,test,unstable-book

image

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -379,20 +379,12 @@ pub fn cli() -> App<'static, 'static> {
                         .long("path")
                         .help("Only print the path to the documentation"),
                 )
-                .arg(
-                    Arg::with_name("book")
-                        .long("book")
-                        .help("The Rust Programming Language book"),
-                )
-                .arg(
-                    Arg::with_name("std")
-                        .long("std")
-                        .help("Standard library API documentation"),
-                )
-                .arg(
-                    Arg::with_name("reference")
-                        .long("reference")
-                        .help("The Rust Reference"),
+                .args(
+                    &DOCS_DATA.into_iter().map(|(name, help_msg, _)| {
+                        Arg::with_name(name)
+                            .long(name)
+                            .help(help_msg)
+                    }).collect::<Vec<_>>()
                 )
                 .arg(
                     Arg::with_name("toolchain")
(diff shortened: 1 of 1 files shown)


src/rustup-cli/rustup_mode.rs
```diff
@@ -950,17 +945,34 @@ fn override_remove(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
...
+    ("proc_macro", "A support library for macro authors when defining new macros", "proc_macro/index.html"),
+    ("reference", "The Rust Reference", "reference/index.html"),
+    ("rust-by-example", "A collection of runnable examples that illustrate various Rust concepts and standard libraries", "rust-by-example/index.html"),
+    ("rustc", "The compiler for the Rust programming language", "rustc/index.html"),
+    ("rustdoc", "Generate documentation for Rust projects", "rustdoc/index.html"),
+    ("std", "Standard library API documentation", "std/index.html"),
+    ("test", "Support code for rustc's built in unit-test and micro-benchmarking framework", "test/index.html"),
+    ("unstable-book", "The Unstable Book", "unstable-book/index.html"),
+];
+
```
kinnison, line 966: Rather than hard-coding this list, is there a way we can load it from the toolchain?  I accept that might be hard currently since we'd have to parse a partial commandline loosely to ensure we knew which toolchain to use, and then go again once we'd loaded the docs list, but it'd mean that if new docs were added, we'd automatically pick them up.

kinnison (commented):
This seems to support everything the old API did, but I'm not entirely convinced having that many hard-coded documentation sections is sensible.

src/rustup-cli/rustup_mode.rs
king6cong: Yeah, I have also considered auto-generating vs hard-coding. Actually, the flags are got by running `cd ~/.rustup/toolchains/nightly-2018-12-31-x86_64-apple-darwin/share/doc/rust/html && ls */index.html | xargs -n1 echo | cut -d'/' -f1` :)

In addiction to the toolchain choosing problem, I think these should also be considered:

1. `rust doc` flags did not always have to match up with the directory name where the docs resides, that's why the document index path isn't auto-generated by appending `/index.html` to flags. This is also how things are handled at the moment.
2. Help messages cannot be easily auto-generated due to different layouts of these docs
3. I think it may not be the case that every doc needs a `cargo doc` shortcut
4. This list does not tend to change very often. 

So at last, I thought handpicking might be a better choice

kinnison: That sounds fair, thanks for explaining it.

nrc:
cc @rust-lang/rustdoc thoughts? Seems nice to have to me, but are there potential downsides?

GuillaumeGomez:
The more the better I assume? The only known problem with docs is that it's very slow on windows. But beyond that, seems like a good thing to me.

QuietMisdreavus:
Seems more like a @rust-lang/docs thing.

I'm a fan of this. We have been adding things to the bookshelf over time, but as @king6cong mentioned, it's not so often that it will create a problem.

PR #1599, opened 2019-01-03
allow non-utf8 arguments to proxies

euclio:
Fixes rust-lang/rust#56146.

### DIFF
--- a/src/rustup-cli/proxy_mode.rs
+++ b/src/rustup-cli/proxy_mode.rs
@@ -15,7 +15,7 @@ pub fn main() -> Result<()> {
     let ExitCode(c) = {
         let _setup = job::setup();
 
-        let mut args = env::args();
+        let mut args = env::args_os();
 
         let arg0 = args.next().map(PathBuf::from);
         let arg0 = arg0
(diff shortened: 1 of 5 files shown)


src/rustup-cli/proxy_mode.rs
```diff
@@ -26,14 +26,15 @@ pub fn main() -> Result<()> {
...
-
+        let toolchain = arg1.as_ref()
+            .and_then(|arg1| arg1.to_str())
+            .and_then(|arg1| {
+                if arg1.starts_with('+') {
+                    Some(&arg1[1..])
+                } else {
+                    None
+                }
+            });
```
kinnison: Could we do the `.starts_with('+')` before the `.to_str()` and issue an error if the user has specified a non-utf8 toolchain name?

kinnison (commented):
Other than being unsure what 'proxies' are in this context, I think this change looks sensible.  Certainly preferable to a panic, though I worry if the user might be surprised if they enter a non-utf8 toolchain name and are then confused when instead of honouring it, it is silently ignored.

nrc:
This looks good, but I agree it would be better to warn the user rather than silently ignore it.

src/rustup-cli/proxy_mode.rs
euclio: Added the check and two more tests.

kinnison (approved):
This seems to do as requested, thank you for the change and the tests.

nrc:
Thanks!

PR #1604, opened 2019-01-07
Fix non-local variables in rustup-init

tesuji:
Hope this fix those issues ([1], [2]) pointed by @ehuss in #1582 .

[1]: https://github.com/rust-lang/rustup.rs/pull/1582#discussion_r245556881
[2]: https://github.com/rust-lang/rustup.rs/pull/1582#discussion_r245556864

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -22,7 +22,7 @@ fi
 #XXX: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat 1>&2 <<EOF
-rustup-init 1.0.0 (408ed84 2017-02-11)
+rustup-init 1.16.0 (beab5ac2b 2018-12-06)
 The installer for rustup
 
 USAGE:
@@ -111,8 +111,6 @@ main() {
         exit 1
     fi
 
-
-
     if [ "$need_tty" = "yes" ]; then
         # The installer is going to want to ask for confirmation by
         # reading stdin.  This script was piped into `sh` though and
@@ -175,7 +173,7 @@ get_endianness() {
 }
 
 get_architecture() {
-    local _ostype _cputype
+    local _ostype _cputype _bitness _arch
     _ostype="$(uname -s)"
     _cputype="$(uname -m)"
 
@@ -283,7 +281,6 @@ get_architecture() {
                 # only n64 ABI is supported for now
                 _ostype="${_ostype}abi64"
             fi
-
             _cputype="$(get_endianness "$_cputype" "" 'el')"
             ;;
 
@@ -378,6 +375,7 @@ ignore() {
 # This wraps curl or wget. Try curl first, if not installed,
 # use wget instead.
 downloader() {
+    local _dld
     if check_cmd curl; then
         _dld=curl
     elif check_cmd wget; then


tesuji:
Rebased on master.

nrc:
@ehuss would you mind taking a look please?

rustup-init.sh
```diff
@@ -175,9 +172,10 @@ get_endianness() {
 }
 
 get_architecture() {
-    local _ostype _cputype
+    local _ostype _cputype _bitness _arch
     _ostype="$(uname -s)"
     _cputype="$(uname -m)"
+    _bitness="$(get_bitness)"
```
ehuss: This now runs on all platforms, which will fail on some with the following error.
```
head: /proc/self/exe: No such file or directory
rustup: unknown platform bitness
```
I suspect it was only run on linux for that reason.

ehuss (changes requested):

ehuss:
At this time I don't think we want to make this many changes to the init script.  If you can narrow it down to just adding the appropriate `local` declarations, I think that would be good.  Thanks for taking a look, though!

rustup-init.sh
```diff
@@ -15,14 +15,13 @@
 
 set -u
 
-if [ -z "$RUSTUP_UPDATE_ROOT" ]; then
-    RUSTUP_UPDATE_ROOT="https://static.rust-lang.org/rustup"
-fi
+RUSTUP_UPDATE_ROOT=${RUSTUP_UPDATE_ROOT:-"https://static.rust-lang.org/rustup"}
```
lzutao: This change fixes Undefined Variables error in `RUSTUP_UPDATE_ROOT` because of setting `set -u`.

rustup-init.sh
```diff
@@ -268,23 +266,15 @@ get_architecture() {
             ;;
 
         mips)
-            _cputype="$(get_endianness $_cputype "" 'el')"
+            _cputype=$(get_endianness mips '' el)
             ;;
```
lzutao: I think that change is easier to read.

rustup-init.sh
```diff
@@ -268,23 +266,15 @@ get_architecture() {
...
-            _bitness="$(get_bitness)"
-            if [ "$_bitness" -eq 32 ]; then
-                if [ "$_ostype" = "unknown-linux-gnu" ]; then
-                    # 64-bit kernel with 32-bit userland
-                    # endianness suffix is appended later
-                    _cputype=mips
-                fi
-            else
+            if [ "$_bitness" -eq 64 ]; then
                 # only n64 ABI is supported for now
```
lzutao: Does mips64 have 16-bits?

rustup-init.sh
```diff
@@ -15,14 +15,13 @@
 
 set -u
 
-if [ -z "$RUSTUP_UPDATE_ROOT" ]; then
-    RUSTUP_UPDATE_ROOT="https://static.rust-lang.org/rustup"
-fi
+RUSTUP_UPDATE_ROOT=${RUSTUP_UPDATE_ROOT:-"https://static.rust-lang.org/rustup"}
 
 #XXX: If you change anything here, please make the same changes in setup_mode.rs
+version='rustup-init 1.16.0 (beab5ac2b 2018-12-06)'
```
dwijnand: Should we keep this local to `usage`?

nrc:
ping @lzutao do you have any update here? Could you reply to the questions inline please?

rustup-init.sh
tesuji: Maybe rename it to `rustup_version`?

tesuji: The reason I split it to `version` is to easy to spot this need to be updated. People seems to forget to update this field.

tesuji:
@nrc I think this PR make a small change to init script. Should I split this change to another PR?
Cause one important point of this change is to fix **Undefined Variables** error 
in `RUSTUP_UPDATE_ROOT` because of setting `set -u`.

rustup-init.sh
dwijnand: Ah, good reason.  I'm fine with it staying as "version".

kinnison:
My rough feeling is that we can probably safely merge this, though I would like to be reassured that it behaves sensibly on MIPS since that seems to be the crux of the change.  If you're not able to provide that reassurance then yes, making this PR only about fixing the actual bug you describe and then a separate PR can be about solving the MIPS related stuff.

tesuji:
Force push to only include local stuff fix.

Thanks @kinnison, I would open another PR for other stuffs.

kinnison (approved):
I think this now LGTM. Do you agree @dwijnand ? @nrc ?

dwijnand (approved):
yes

PR #1605, opened 2019-01-09
 self update after updating a specific toolchain 

nrc:
r? @alexcrichton 

This will make sure people who do `rustup update stable` still get a new rustup

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -132,6 +132,19 @@ pub fn cli() -> App<'static, 'static> {
                         .help(TOOLCHAIN_ARG_HELP)
                         .required(true)
                         .multiple(true),
+                )
+                .arg(
+                    Arg::with_name("no-self-update")
+                        .help("Don't perform self update when running the `rustup` command")
+                        .long("no-self-update")
+                        .takes_value(false)
+                        .hidden(true),
+                )
+                .arg(
+                    Arg::with_name("force")
+                        .help("Force an update, even if some components are missing")
+                        .long("force")
+                        .takes_value(false),
                 ),
         )
         .subcommand(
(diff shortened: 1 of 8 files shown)


alexcrichton:
r=me with green CI

src/rustup-cli/rustup_mode.rs
```diff
@@ -132,6 +132,19 @@ pub fn cli() -> App<'static, 'static> {
                         .help(TOOLCHAIN_ARG_HELP)
                         .required(true)
                         .multiple(true),
+                )
+                .arg(
+                    Arg::with_name("no-self-update")
+                        .help("Don't perform self update when running the `rustup` command")
```
kinnison, line 138: should this be:

```rust
.help("Don't perform self update when running the `rustup install` command")
```

perhaps?

src/rustup-cli/rustup_mode.rs
```diff
@@ -196,6 +209,13 @@ pub fn cli() -> App<'static, 'static> {
                                 .help(TOOLCHAIN_ARG_HELP)
                                 .required(true)
                                 .multiple(true),
+                        )
+                        .arg(
+                            Arg::with_name("no-self-update")
+                                .help("Don't perform self update when running the `rustup` command")
```
kinnison, line 215: Ditto here but with 
```rust
.help("Don't perform self update when running the `rustup toolchain install` command")
```

perhaps?

kinnison (approved):
I'm pretty OK with this in general, though the amount of tests needing `--no-self-update` I wonder if it might be better to also have a `RUSTUP_SUPPRESS_SELF_UPDATE` type environment variable which we could set by default in most of the `expect_ok()` calls.

Nothing blocking a merge IMO though.

PR #1606, opened 2019-01-09
Add miri to rustup

oli-obk:
It was just added to the distribution in https://github.com/rust-lang/rust/pull/57086

This PR enables calling `cargo miri` after installing the `miri` component

### DIFF
--- a/src/rustup-win-installer/src/lib.rs
+++ b/src/rustup-win-installer/src/lib.rs
@@ -27,6 +27,7 @@ static TOOLS: &'static [&'static str] = &[
     "rustfmt",
     "cargo-fmt",
     "cargo-clippy",
+    "cargo-miri",
 ];
 
 #[no_mangle]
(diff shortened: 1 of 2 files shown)


kinnison:
Would it make sense to have a `component_for_bin()` to map `cargo-miri` to `miri` ?

oli-obk:
I was considering that, but I felt like miri is niche enough so it won't impact too many ppl. I can change it though if anyone thinks it's worthwile

kinnison:
`component_for_bin()` is pretty much only used to suggest component names to install if something is missing, but if someone tries to run `cargo miri ...` and then doesn't get the hint, they might be upset.

I don't mind either way, just thought i'd mention it in case you hadn't already considered and discarded the idea :-D

kinnison:
Change-wise this looks good :+1: but I would be more comfortable if it were rebased so that there aren't fixup commits in the sequence.  I can do this if you don't have time, so long as someone can tell me how to update someone else's PR :-)

oli-obk:
Updated

kinnison (approved):
LGTM. @nrc I *think* this is good for merge.  I can't see any reason why it'd break anything.

nrc:
Thanks!

aleksmelnikov:
Will miri component be availiable in the next rustup release (1.17.0)?

PR #1610, opened 2019-01-10
Fixed according to warning of clippy

k-nasa:
## Fixed according to warning of clippy

- Remove unnecessary return keywords
- Remove unnecessary name assignment
- Remove static lifetime keyword


kinnison (approved):
These all look safe, and look to be part of a large clippy refactor I was intending to do at some point, so :+1: from me.

nrc:
Thanks!

Issue #1596, opened 2018-12-31
Misleading error message when a component is not found


Yatekii:
When one tries to install a component on a a toolchain which does not have that component, you get a message like this:

```
error: component 'rls' for target 'x86_64-unknown-linux-gnu' is unavailable for download
```

This message implies that the component does not exist for your architecture. But it in fact means it's not available for the currently selected toolchain, which is not mentioned in the error message.
For example in my case I had the toolchain of my PWD overridden to nightly and this `rls` was not found. The error message suggested that `rls` is not available for my architecture. But in fact it was available; just for stable.

This can be very missleading as for example in my case I knew that the component existed for my architecture and I though that something completely different went wrong. Maybe my net or the likes.

In my opinion it is absolutely never enough to imply something in error messages (as in this case it is implied that the current toolchain is meant) because unfamiliar users, which get confronted with error messages like this the most, are not aware of the implications.

There is an easy fix to this: Add the complete name of the toolchain to this error message and all is good.

fbruetting:
I also have the same issue, and adding the complete name of the toolchain doesn’t fix this. Any suggestions?

```
rustup update nightly-x86_64-unknown-linux-gnu
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: latest update on 2019-02-16, rust version 1.34.0-nightly (a9410cd1a 2019-02-15)
error: component 'rls' for target 'x86_64-unknown-linux-gnu' is unavailable for download
```

// OK, uninstalling and reinstalling nightly helped.

PR #1616, opened 2019-01-17
Thread toolchain through to error message

kinnison:
In the component-unavailable error case, we need to report the
full toolchain rather than just the component and target name.
This threads it through and reports it in the error message.

Fixes: #1596

This is not ideal right now, but it's a starting point for discussion.

Currently the error message ends up looking something like:

```
error: component 'rust-std' for target 'x86_64-unknown-linux-gnu' is unavailable for download for toolchain 'nightly-x86_64-unknown-linux-gnu'
```

Clearly the repetition of the target name in the toolchain name isn't ideal, I've been considering
reformatting the message but didn't want to do so without some further input on what might be better.

### DIFF
--- a/src/rustup-dist/src/dist.rs
+++ b/src/rustup-dist/src/dist.rs
@@ -556,6 +556,7 @@ pub fn update_from_dist_<'a>(
                 force_update,
                 &download,
                 download.notify_handler.clone(),
+                &toolchain.manifest_name(),
             )? {
                 UpdateStatus::Unchanged => Ok(None),
                 UpdateStatus::Changed => Ok(Some(hash)),
--- a/src/rustup-dist/src/errors.rs
+++ b/src/rustup-dist/src/errors.rs
@@ -107,9 +107,9 @@ error_chain! {
             description("missing package for the target of a rename")
             display("server sent a broken manifest: missing package for the target of a rename {}", name)
         }
-        RequestedComponentsUnavailable(c: Vec<Component>, manifest: Manifest) {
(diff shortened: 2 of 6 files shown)


h-michael:
How about `download for 'nightly' channel`?

kinnison:
I need to work out the right way to produce a "short" toolchain, but yes that wording migh be nicer @h-michael

kinnison:
I've managed to make it say that (and if there's a date associated, channel-date) so I'd appreciate reviews / opinions.  If noone has anything negative to say in the next day or so I'll un-WIP this.

nrc:
Looks good. We would want to fix the repetition. We should probably split the toolchain into channel and host (is it host or target, actually?) and only report the channel and host if it is different to the target

kinnison:
So I changed the toolchain to just be `channel` or `channel-date` but I suppose we properly need to distinguish the toolchain host *if* it doesn't match the host of the system running rustup.  The goal would be to ensure that the toolchain statement in the error message should match what is being used with the minimal content possible.

I think this probably needs careful thought to decide what's ideal.  What is here now is better, but perhaps not ideal.

dwijnand:
Removing the repetition seems to me like a nice-to-have of lesser value than resolving #1596, so I'm happy to accept this as is and trim it later/never.

kinnison:
I've opted to rebase this onto mater and as per @dwijnand 's observation that something is better than nothing, I commend it for review and ideally merge.

I'm not sure entirely how I'd make the change which we've identified as most desirable so I suggest that we take this for now, and consider fixing up the suggestions further after simplification has proceeded somewhat.

dwijnand (approved):

PR #1619, opened 2019-01-20
Better error message for missing binary

In-line:
Hey! I was annoyed, that simple copy-pasting from error message didn't work, so I tried to make better suggestion.

### DIFF
--- a/tests/cli-misc.rs
+++ b/tests/cli-misc.rs
@@ -380,6 +380,9 @@ fn rustup_failed_path_search() {
                 &config.customdir.join("custom-1").to_string_lossy(),
             ],
         );
+
+        expect_ok(config, &["rustup", "default", "custom"]);
+
         let broken = &["rustup", "run", "custom", "fake_proxy"];
         expect_err(
             config,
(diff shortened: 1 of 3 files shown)


tests/cli-misc.rs
```diff
@@ -394,6 +404,61 @@ fn rustup_failed_path_search() {
...
+                "custom-2"
+            ],
+        );
+
+        let broken = &["rustup", "run", "custom-1", "fake_proxy"];
+        expect_err(
+            config,
+            broken,
+            &format!(
+                "rustup component add fake_proxy{} --toolchain custom-2",
```
In-line: It seems that `expect_err` doesn't read second line of the `stderr`. Is this standard library bug? How to test second line? I'm not sure.

kinnison: It certainly *should* be reading all of stdout: https://github.com/rust-lang/rustup.rs/blob/416f4e98d45ed3ebba85f2347fbd383a0400d0d6/src/rustup-mock/src/clitools.rs#L358

kinnison: I believe the situation here is simply that `component_for_bin()` is returning None for your `fake_proxy` and as such no installation suggestion can be generated.  You might have to construct a fake for a proxy whose name is known to `component_for_bin()` such as `cargo-miri`

In-line: @kinnison I will try to fix test on the following days.

kinnison:
@In-line It looks like you're getting close, but your reformatting would be better done in the same commit, could you please amend your change to one commit (and also remove the exeext stuff since it's not needed for the component name.  If you do that I think you're golden.

I like the error message change, but we'll need to wait for at least one more maintainer to take a look.

In-line:
@kinnison I'm going to finish this PR at the end of the week. Thanks for the feedback!

dwijnand:
LGTM!

dwijnand (approved):

nrc:
Awesome, thanks for the PR!

PR #1622, opened 2019-01-23
Fix panic if stdout disappears

### LINKED ISSUE #1621: rustup panics when piping to `grep -q`
I noticed that rustup panics in the case below.

```
 RUST_BACKTRACE=1 rustup show | grep -q ^nightly
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', libstd/io/stdio.rs:700:9
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::panicking::default_hook::{{closure}}
             at libstd/sys_common/backtrace.rs:71
   2: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:227
             at libstd/panicking.rs:476
   3: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:390
   4: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:345
   5: std::io::stdio::_print
             at libstd/io/stdio.rs:700
             at libstd/io/stdio.rs:709
   6: rustup_init::rustup_mode::main
   7: rustup_init::run_rustup
   8: rustup_init::main
   9: std::rt::lang_start::{{closure}}
  10: main
  11: __libc_start_main
  12: <unknown>
```

NickeZ:
Fixes #1621

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -702,34 +702,37 @@ fn show(cfg: &Cfg) -> Result<()> {
             print_header("installed toolchains")
         }
         let default_name = cfg.get_default()?;
-        for t in installed_toolchains {
-            if default_name == t {
-                println!("{} (default)", t);
+        let mut t = term2::stdout();
+        for it in installed_toolchains {
+            if default_name == it {
+                let _ = writeln!(t, "{} (default)", it);
             } else {
-                println!("{}", t);
+                let _ = writeln!(t, "{}", it);
             }
         }
         if show_headers {
-            println!("")
-        };
+            let _ = writeln!(t, "");
+        }
     }
 
     if show_active_targets {
         if show_headers {
             print_header("installed targets for active toolchain");
         }
-        for t in active_targets {
-            println!(
+        let mut t = term2::stdout();
+        for at in active_targets {
+            let _ = writeln!(
+                t,
                 "{}",
-                t.component
+                at.component
                     .target
                     .as_ref()
                     .expect("rust-std should have a target")
             );
         }
         if show_headers {
-            println!("")
-        };
+            let _ = writeln!(t, "");
+        }
     }
 
     if show_active_toolchain {
@@ -737,32 +740,34 @@ fn show(cfg: &Cfg) -> Result<()> {
             print_header("active toolchain")
         }
 
+        let mut t = term2::stdout();
+
         match active_toolchain {
             Ok(atc) => match atc {
                 Some((ref toolchain, Some(ref reason))) => {
-                    println!("{} ({})", toolchain.name(), reason);
-                    println!("{}", common::rustc_version(toolchain));
+                    let _ = writeln!(t, "{} ({})", toolchain.name(), reason);
+                    let _ = writeln!(t, "{}", common::rustc_version(toolchain));
                 }
                 Some((ref toolchain, None)) => {
-                    println!("{} (default)", toolchain.name());
-                    println!("{}", common::rustc_version(toolchain));
+                    let _ = writeln!(t, "{} (default)", toolchain.name());
+                    let _ = writeln!(t, "{}", common::rustc_version(toolchain));
                 }
                 None => {
-                    println!("no active toolchain");
+                    let _ = writeln!(t, "no active toolchain");
                 }
             },
             Err(err) => {
                 if let Some(cause) = err.source() {
-                    println!("(error: {}, {})", err, cause);
+                    let _ = writeln!(t, "(error: {}, {})", err, cause);
                 } else {
-                    println!("(error: {})", err);
+                    let _ = writeln!(t, "(error: {})", err);
                 }
             }
         }
 
         if show_headers {
-            println!("")
-        };
+            let _ = writeln!(t, "");
+        }
     }
 
     fn print_header(s: &str) {


src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
             print_header("installed toolchains")
         }
         let default_name = cfg.get_default()?;
-        for t in installed_toolchains {
-            if default_name == t {
-                println!("{} (default)", t);
+        let mut t = term2::stdout();
+        for it in installed_toolchains {
+            if default_name == it {
+                let _ = writeln!(t, "{} (default)", it);
```
kinnison, line 708: Shouldn't we be using `?` rather than ignoring the error returns from `writeln!()` since this is just informational and so the jump out shouldn't affect matters?

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
             } else {
-                println!("{}", t);
+                let _ = writeln!(t, "{}", it);
```
kinnison, line 710: Ditto.

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
             }
         }
         if show_headers {
-            println!("")
-        };
+            let _ = writeln!(t, "");
```
kinnison, line 714: And again.

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
 
     if show_active_targets {
         if show_headers {
             print_header("installed targets for active toolchain");
         }
-        for t in active_targets {
-            println!(
+        let mut t = term2::stdout();
+        for at in active_targets {
+            let _ = writeln!(
```
kinnison, line 724: And once more

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
+                at.component
                     .target
                     .as_ref()
                     .expect("rust-std should have a target")
             );
         }
         if show_headers {
-            println!("")
-        };
+            let _ = writeln!(t, "");
```
kinnison, line 734: Here too?

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
 
+        let mut t = term2::stdout();
+
         match active_toolchain {
             Ok(atc) => match atc {
                 Some((ref toolchain, Some(ref reason))) => {
-                    println!("{} ({})", toolchain.name(), reason);
-                    println!("{}", common::rustc_version(toolchain));
+                    let _ = writeln!(t, "{} ({})", toolchain.name(), reason);
+                    let _ = writeln!(t, "{}", common::rustc_version(toolchain));
```
kinnison, line 749: These two too.

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
                 }
                 Some((ref toolchain, None)) => {
-                    println!("{} (default)", toolchain.name());
-                    println!("{}", common::rustc_version(toolchain));
+                    let _ = writeln!(t, "{} (default)", toolchain.name());
+                    let _ = writeln!(t, "{}", common::rustc_version(toolchain));
```
kinnison, line 753: And here

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
                 }
                 None => {
-                    println!("no active toolchain");
+                    let _ = writeln!(t, "no active toolchain");
```
kinnison, line 756: And here

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
                 }
             },
             Err(err) => {
                 if let Some(cause) = err.source() {
-                    println!("(error: {}, {})", err, cause);
+                    let _ = writeln!(t, "(error: {}, {})", err, cause);
```
kinnison, line 761: Here too

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
                 } else {
-                    println!("(error: {})", err);
+                    let _ = writeln!(t, "(error: {})", err);
```
kinnison, line 763: And here

src/rustup-cli/rustup_mode.rs
```diff
@@ -702,67 +702,72 @@ fn show(cfg: &Cfg) -> Result<()> {
...
                 }
             }
         }
 
         if show_headers {
-            println!("")
-        };
+            let _ = writeln!(t, "");
```
kinnison, line 769: Final one.

kinnison (changes requested):
I'm not entirely sure I'm comfortable with the number of times we have to acquire `term2::stdout` but I think I understand why it'd have to be this way.

But I don't like how we're swallowing the errors from writeln! I'd rather we just early jump out and allow things to shut down more quickly.

NickeZ:
Thanks for the review. Is it better to only acquire stout once and to propagate it into for example "print_header"?

dwijnand:
What are our chances of being able to simulate this failure in a test?  I agree that #1621 is a bug and should be fixed, but I'd prefer to find the minimal solution that swallows the least errors and panics as much as it used to.

NickeZ:
Another option is to register a handler to sigpipe and exit gracefully. I guess that is a straightforward solution if rustup already handles signals.

Rust registers the "ignore" signal handler for sigpipe by default:

https://github.com/rust-lang/rust/blob/79d8a0fcefa5134db2a94739b1d18daa01fc6e9f/src/libstd/sys/unix/mod.rs#L63-L81

On the other hand, rust std ignores all other reasonable stdout IO errors such as stdout being closed or read-only. So I'm not sure what other kinds of errors you would like to propagate from write!().

https://github.com/rust-lang/rust/blob/01af12008d63a64446a86d995e772f8a539a4202/src/libstd/io/stdio.rs#L88-L118

kinnison:
Mostly I just find `let _ = writeln!(...);` much less pleasant to see in code than `writeln!(...)?;` but I'm not going to block things if others disagree with me.  I'm still developing my particular Rust aesthetic.

NickeZ:
> What are our chances of being able to simulate this failure in a test?

The easiest reproducer is below. I don't know if that qualiefies as a "simulation of this failure".

```
rustup show | head -1
```

Actually both the current behavior and my PR is the wrong expected behavior. Current behavior exits with error code != 0, and with my PR rustup continues to run even though we technically have received SIGPIPE and should exit asap.

Would you like me to propagate the error all the way up to main and there handle it appropriately? Which is die silently with error code 0.

kinnison:
I think that #1630 is a better version of this. and I'd suggest closing this PR in favour of that.

PR #1625, opened 2019-01-27
Install some GitHub issue templates

dwijnand:
I've had good experiences using issue templates, and these are derived from the ones established in the cargo repo.  WDYT?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,26 @@
+---
+name: Bug report
+about: Create a report to help us improve
+labels: bug
+---
+
+<!-- Thanks for filing a 🐛 bug report 😄! -->
+
+**Problem**
+<!-- A clear and concise description of what the bug is. -->
+<!-- including what currently happens and what you expected to happen. -->
+
+**Steps**
+<!-- The steps to reproduce the bug. -->
+1.
+2.
+3.
+
+**Possible Solution(s)**
+<!-- Not obligatory, but suggest a fix/reason for the bug, -->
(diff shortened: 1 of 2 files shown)


dwijnand:
WDYT @kinnison?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,29 @@
...
+3.
+
+**Possible Solution(s)**
+<!-- Not obligatory, but suggest a fix/reason for the bug, -->
+<!-- or ideas how to implement the addition or change -->
+
+**Notes**
+
+Output of `rustup --version`:
+Output of `rustc --version`:
```
kinnison: I think rather than `rustc --version` it might be better to have `rustup show` output

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,29 @@
...
+
+<!-- Also, any additional context or information you feel may be relevant to the issue. -->
+<!-- (e.g rust version, OS platform/distribution/version, target toolchain(s), release channel.. -->
```
kinnison: This would be covered by the `rustup show` output

kinnison (changes requested):
Over all I like the idea of having some templates.  It might be nice to have a PR template later too but this is an excellent start.

With the changes I have suggested, I'd have no problem giving a :+1:

dwijnand:
Thanks, @kinnison.  All done.

kinnison (approved):
LGTM

nrc (approved):

nrc:
SGTM, thanks!

PR #1623, opened 2019-01-25
Add tab completion instructions for PowerShell

matt9ucci:
For Windows PowerShell and PowerShell Core (Linux/Mac/Win) users.

PR #1629, opened 2019-01-30
Add tab completion test for PowerShell

matt9ucci:
Using Pester and AppVeyor. Related to #1623.

### DIFF
--- a/src/rustup-win-installer/msi/Completion.Tests.ps1
+++ b/src/rustup-win-installer/msi/Completion.Tests.ps1
@@ -1,0 +1,36 @@
+if (Get-Command -Name rustup -ErrorAction Stop) {
+	# Enable rustup tab completion
+	rustup completions powershell | Out-String | Invoke-Expression
+}
+
+function Get-Result([string]$Text, [int]$CursorPosition = $Text.Length, [hashtable]$Options) {
+	[System.Management.Automation.CommandCompletion]::CompleteInput($Text, $CursorPosition, $Options).CompletionMatches
+}
+
+Describe 'rustup flags and subcommands' {
+	$result = Get-Result 'rustup '
+	$result.Count | Should Be 23
+	$result[00].CompletionText | Should Be 'completions'
+	$result[01].CompletionText | Should Be 'component'
+	$result[02].CompletionText | Should Be 'default'
+	$result[03].CompletionText | Should Be 'doc'
+	$result[04].CompletionText | Should Be '-h'
+	$result[05].CompletionText | Should Be 'help'
+	$result[06].CompletionText | Should Be '--help'
+	$result[07].CompletionText | Should Be 'install'
+	$result[08].CompletionText | Should Be 'override'
+	$result[09].CompletionText | Should Be 'run'
+	$result[10].CompletionText | Should Be 'self'
+	$result[11].CompletionText | Should Be 'set'
+	$result[12].CompletionText | Should Be 'show'
+	$result[13].CompletionText | Should Be 'target'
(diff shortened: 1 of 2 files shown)


dwijnand (approved):
Nice!  Thank you.

src/rustup-win-installer/msi/Completion.Tests.ps1
```diff
@@ -0,0 +1,36 @@
...
+	rustup completions powershell | Out-String | Invoke-Expression
+}
+
+function Get-Result([string]$Text, [int]$CursorPosition = $Text.Length, [hashtable]$Options) {
+	[System.Management.Automation.CommandCompletion]::CompleteInput($Text, $CursorPosition, $Options).CompletionMatches
+}
+
+Describe 'rustup flags and subcommands' {
+	$result = Get-Result 'rustup '
+	$result.Count | Should Be 23
```
kinnison: I'm worried that doing this exact test might be risky going forward.

Perhaps, if it's possible, something like a "More than 10" type test would be better, and then rather than requiring the specific ordering of content subsequent in this function, that we look for some critical words instead such as install, update, completions, --help, --verbose.  Otherwise I'm really worried that as we adjust the command line interface (which will happen, even if it's just adding stuff) it's going to be hard for any of us who don't know PowerShell to keep track.

kinnison (changes requested):
I'm grateful for the test, but the concern I raised would prevent me from approving.  The test is just too brittle as it is written.

src/rustup-win-installer/msi/Completion.Tests.ps1
dwijnand: :+1: probably better to not over-specify.

matt9ucci: I agree.

src/rustup-win-installer/msi/Completion.Tests.ps1
```diff
@@ -9,28 +9,13 @@ function Get-Result([string]$Text, [int]$CursorPosition = $Text.Length, [hashtab
...
-	$result[16].CompletionText | Should Be 'uninstall'
-	$result[17].CompletionText | Should Be 'update'
-	$result[18].CompletionText | Should Be '-V'
-	$result[19].CompletionText | Should Be '-v'
-	$result[20].CompletionText | Should Be '--verbose'
-	$result[21].CompletionText | Should Be '--version'
-	$result[22].CompletionText | Should Be 'which'
+
+	# Keep it simple because over-specify testing might be risky
+	$result.Count | Should -BeGreaterThan 10
```
matt9ucci, line 14: It is a "More than 10" type test.

src/rustup-win-installer/msi/Completion.Tests.ps1
```diff
@@ -9,28 +9,13 @@ function Get-Result([string]$Text, [int]$CursorPosition = $Text.Length, [hashtab
...
+	$result.CompletionText | Should -Contain 'completions'
+	$result.CompletionText | Should -Contain 'install'
+	$result.CompletionText | Should -Contain 'uninstall'
+	$result.CompletionText | Should -Contain 'update'
+	$result.CompletionText | Should -Contain '--help'
+	$result.CompletionText | Should -Contain '--verbose'
```
matt9ucci, line 20: `Should -Contain` can verify elements of an array ignoring order.

dwijnand:
Nice, thanks.

kinnison (approved):
LGTM.

nrc:
Thank you!

PR #1630, opened 2019-01-30
Propagate io and terminal errors upward and ignore EPIPE

### LINKED ISSUE #1621: rustup panics when piping to `grep -q`

NickeZ:
This is an alternative to: #1622

In this solution I instead ignore the EPIPE error at the highest level. I haven't contributed to rustup before, so please check the PR carefully.

Fixes #1621


kinnison:
HI @NickeZ thanks for taking the time to produce this alternative patch.  I'm sorry I was away last week so didn't get a chance to review it before now.

A brief look over suggests this is a reasonable approach, but I want to take some time to consider if there're other places where we ought to handle epipe in a similar fashion (e.g. component list)

Please bear with me for a few days.

kinnison (approved):
I like this change. :+1:

NickeZ:
Great, but can you check if the macro I have written is correct/idiomatic?

edit: It could probably just be a regular function...

NickeZ:
Is it possible to add a test that runs the equivalent of `rustup show | head -1`? Could someone show me approximately where and how to do that?

Cargo.lock
```diff
@@ -1,3 +1,5 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
```
kinnison, line 2: Is this added by a new version of Cargo?

kinnison (approved):
Yeah, I think as a function it's a bit nicer.  I didn't mind the macro at all, but this is perhaps easier for others to grok.

Are there any other parts of the codebase where handling epipe would be useful?  I'm not asking for them in this PR, it has a :+1: already from me, but perhaps once this is in you might go for a wander and see if there's other places worth checking on?

Regarding the testing for `rustup show | head -1` I don't think we have a way to easily simulate that in our tests right now, so I don't mind not having that yet, but if you think of something I'll be pleased to review another PR (or change on this one) for it.

Cargo.lock
NickeZ: It must have been. I only ran cargo fmt, check and build. I didn't manually change that file. do you want me to remove it?

NickeZ:
Do you want me to squash it or anything else before merging? To me it is good to go as it is now.

kinnison:
> Do you want me to squash it or anything else before merging? To me it is good to go as it is now.

My preference is certainly that the commits are neat and tidy, so yes, a squash would be appreciated.  

Thanks

Cargo.lock
dwijnand: > Is this added by a new version of Cargo?

Yes: https://github.com/rust-lang/cargo/pull/6548

NickeZ:
OK, squashed.

nrc:
Sorry, this needs a rebase

NickeZ:
rebased and pushed

NickeZ:
ping

Issue #1632, opened 2019-02-01
`rustup default` with no arguments should print the current default


nrc:

dwijnand:
Is that in addition to updating nightly and stable, or in alternative?

dwijnand:
(Cheeky note: had you merged #1625 and used the template I wouldn't've needed to ask, probably =D)

das-sein:
I figured the default behavior is desirable, so in my pull request I let it be.

nrc:
I'm an idiot, I meant `rustup default`, not `rustup update`, sorry for the confusion.

dwijnand:
Fixed by https://github.com/rust-lang/rustup.rs/pull/1633

tesuji:
Close it?

PR #1633, opened 2019-02-03
Print default toolchain on `rustup default` without arguments

das-sein:
Implementation of #1632 

Behavior: When `rustup default` is not provided arguments, print the default toolchain (new behavior) before commencing updates of all installed toolchains.

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -189,7 +189,7 @@ pub fn cli() -> App<'static, 'static> {
                 .arg(
                     Arg::with_name("toolchain")
                         .help(TOOLCHAIN_ARG_HELP)
-                        .required(true),
+                        .required(false),
                 ),
         )
         .subcommand(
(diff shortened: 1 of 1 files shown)


dwijnand:
Hey, @lazorgator.  Thanks for sending the PR.  Want to update the tests so we can assess how this implementation behaves?

das-sein:
Yep, sorry about that! Working on them now for both PRs.

das-sein:
@dwijnand Should be all set. I only needed to modify one test. However, I noticed that it was possible to be in a state wherein no default toolchain was set in the settings. This required modification of the `config.get_default()` behavior very slightly. It doesn't seem to affect how the rest of the software expects that method to work. Without the modification it would throw an exception about trying to `unwrap()` a None during the `rustup_self_update_exact` test.

dwijnand (approved):
Thanks, @lazorgator.  The patch seems reasonable to me.  I'd merge it if I could.  So let's see if this matches @nrc's expectations in #1632.

lazorgator:
Sorry for force push. Accidentally pushed a README change meant for the component filtering PR. As with any history rewrite, I must state that I am a bad person lol.

nrc:
I'm very sorry, but I messed up and wrote `update` when I meant `default` in #1632 . Could you change this PR to printing the same message but on the different command please?

lazorgator:
@nrc All good, will do. Should I keep this in the same PR and simply change the name of it? Also, should I keep this in a new commit or do you prefer squashed?

das-sein:
@nrc Alright, added the functionality to `rustup default` as well! If you want me to remove the functionality from `rustup update`, let me know.

dwijnand:
Perhaps it's best to remove it from `rustup update`.

das-sein:
@dwijnand @nrc Should be all set, I think. Functionality removed from `rustup update`.

dwijnand (approved):
Re-LGTM

nrc:
Awesome, thanks!

dwijnand:
Does this duplicate `rustup show active-toolchain`?

das-sein:
@dwijnand It does I think! I'll work on a PR to just make the `rustup default` behavior an alias to that command, if possible. `rustup show active-toolchain` also deals with the override display issue.

Issue #1631, opened 2019-01-31
Uninstalling default toolchain leaves rustup in bad state


johnthagen:
To reproduce:

```
$ rustup --version
rustup 1.16.0 (beab5ac2b 2018-12-06)
$ rustup install nightly-2019-01-27
$ rustup default nightly-2019-01-27
$ rustup uninstall nightly-2019-01-27
$ rustup show
Default host: x86_64-apple-darwin

installed toolchains
--------------------

stable-x86_64-apple-darwin
nightly-x86_64-apple-darwin

active toolchain
----------------

(error: toolchain 'nightly-2019-01-27-x86_64-apple-darwin' is not installed, not a directory: '/Users/user/.rustup/toolchains/nightly-2019-01-27-x86_64-apple-darwin')
```

Should the uninstall stop you from uninstalling something already set as default? Should the uninstall switch the users default back? Currently it silently works but then the user is in a bad state.

das-sein:
After playing with this, it seems the cleaner solution would be to simply prevent the de-installation of the default toolchain if it's the last toolchain available. Otherwise, there will be significant work adding checks to any function that relies on knowing the default toolchain, as well as any function that outputs the default toolchain to the user. Very messy.

I'm working on this now, but am wondering: is it too opinionated to switch the default toolchain for the user? That seems like it could lead to more confusion than letting them hit the error that their selected default isn't installed, which they would expect.

petrochenkov:
Having no toolchains is a valid situation (https://github.com/rust-lang/rustup.rs/pull/1257), I don't think it's a good idea to permit entering the "no toolchains" state in one way, but not in another.

lazorgator:
It seems the changes I'm making in my PR allow you to initialize rustup without a default toolchain just fine from my own testing, but still working on getting the repo tests to pass.

lazorgator:
Botched the previous pull request :expressionless: Made a new one. `--default-toolchain none` functionality maintained.

PR #1634, opened 2019-02-03
Prevent de-installation of last toolkit

das-sein:
Behavior: should the user attempt to `uninstall` their only remaining toolkit, an error will be produced telling them this is not possible to do.

If something like this isn't done, rustup enters a bad state as demonstrated in #1631 

Edit: This still allows for `--default-toolchain none` during install.


das-sein:
Sorry for the number of commits. First major rust project. :disappointed_relieved:

das-sein:
Very messy pull request. I'll re-open when I fix this up. Sorry again.

PR #1636, opened 2019-02-04
Prevent uninstall of only remaining toolchain

das-sein:
Behavior: should the user attempt to uninstall their only remaining toolkit, an error will be produced telling them this is not possible to do. If uninstall is allowed, rustup enters a bad state as demonstrated in #1631

This still allows for `--default-toolchain none` during install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,7 +555,7 @@ Command | Description
 `rustup override set nightly-2015-04-01` | For the current directory, use a nightly from a specific date
 `rustup toolchain link my-toolchain "C:\RustInstallation"` | Install a custom toolchain by symlinking an existing installation
 `rustup show` | Show which toolchain will be used in the current directory
-`rustup toolchain uninstall nightly` | Uninstall a given toolchain
+`rustup toolchain uninstall nightly` | Uninstall a given toolchain (you cannot uninstall the only toolchain left installed)
 `rustup toolchain help` | Show the `help` page for a subcommand (like `toolchain`) to see what actions are available on it
 
 ## Environment variables
--- a/src/rustup-cli/errors.rs
+++ b/src/rustup-cli/errors.rs
@@ -27,6 +27,10 @@ error_chain! {
             description("toolchain is not installed")
             display("toolchain '{}' is not installed", t)
         }
+        ToolchainUninstallOnly(t: String) {
+            description("attempting to uninstall only remaining toolchain")
+            display("'{}' is the only toolchain left", t)
+        }
         InvalidToolchainName(t: String) {
             description("invalid toolchain name")
             display("invalid toolchain name: '{}'", t)
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -896,6 +896,11 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
 fn toolchain_remove(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
     for toolchain in m.values_of("toolchain").expect("") {
         let toolchain = cfg.get_toolchain(toolchain, false)?;
+        let installed_toolchains = cfg.list_toolchains()?;
+        if installed_toolchains.len() == 1 && toolchain.name() == installed_toolchains[0] {
+            // Do not allow removal of toolchain if it's the only toolchain
+            return Err(ErrorKind::ToolchainUninstallOnly(toolchain.name().to_string()).into());
+        }
         toolchain.remove()?;
     }
     Ok(())
--- a/tests/cli-misc.rs
+++ b/tests/cli-misc.rs
@@ -773,12 +773,19 @@ fn toolchain_broken_symlink() {
         // We artifically create a broken symlink toolchain -- but this can also happen "legitimately"
         // by having a proper toolchain there, using "toolchain link", and later removing the directory.
         fs::create_dir(config.rustupdir.join("toolchains")).unwrap();
+        // Have to have at least one toolchain left before attempting to uninstall
+        expect_ok(config, &["rustup", "default", "stable"]);
         create_symlink_dir(
             config.rustupdir.join("this-directory-does-not-exist"),
             config.rustupdir.join("toolchains").join("test"),
         );
         // Make sure this "fake install" actually worked
-        expect_ok_ex(config, &["rustup", "toolchain", "list"], "test\n", "");
+        expect_ok_ex(
+            config,
+            &["rustup", "toolchain", "list"],
+            for_host!("stable-{0} (default)\ntest\n"),
+            "",
+        );
         // Now try to uninstall it.  That should work only once.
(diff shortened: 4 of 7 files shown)


tests/cli-v1.rs
```diff
@@ -362,8 +362,16 @@ fn remove_toolchain_then_add_again() {
     // Issue brson/multirust #53
     setup(&|config| {
         expect_ok(config, &["rustup", "default", "beta"]);
-        expect_ok(config, &["rustup", "toolchain", "remove", "beta"]);
-        expect_ok(config, &["rustup", "update", "beta", "--no-self-update"]);
+        // Can't uninstall the only toolkit left
```
kinnison: I think you mean tool*chain* here, rather than tool*kit*

tests/cli-v2.rs
```diff
@@ -462,8 +470,16 @@ fn remove_toolchain_then_add_again() {
     // Issue brson/multirust #53
     setup(&|config| {
         expect_ok(config, &["rustup", "default", "beta"]);
-        expect_ok(config, &["rustup", "toolchain", "remove", "beta"]);
-        expect_ok(config, &["rustup", "update", "beta", "--no-self-update"]);
+        // Can't uninstall the only toolkit left
```
kinnison: Ditto here, tool*chain* not tool*kit*

kinnison (changes requested):
Fundamentally looks reasonable.  Two tiny typos in comments, so not block-worthy, but...

I'd like to see a `README.md` update because while this is fixing an unpleasant corner case of behaviour, it does now prevent something which was previously possible.  It looks like there isn't a dedicated uninstallation section of the README so I'd be satisfied by something like `(you may not uninstall the only toolchain installed)` appended to the line mentioning `rustup toolchain uninstall` in the README.

My remaining concern is that if someone had rustup installed somewhere with room for only one toolchain, they'd be unable to switch from (say) nightly to stable without uninstalling rustup entirely should this change be merged.  I'm not sure how I feel about this.

das-sein:
@kinnison Excellent point. I had not even considered that scenario. I wonder if this means there should be an in-place switch command in addition to this change? Maybe there's another way to address this undesirable behavior.

README.md
```diff
@@ -530,9 +530,9 @@ If you need a more complex setup, rustup supports the convention used by
 the __curl__ program, documented in the ENVIRONMENT section of
 [its manual page][curlman].
 
-Note that some versions of `libcurl` apparently require you to drop the 
-`http://` or `https://` prefix in environment variables. For example, 
-`export http_proxy=proxy.example.com:1080` (and likewise for HTTPS). 
+Note that some versions of `libcurl` apparently require you to drop the
+`http://` or `https://` prefix in environment variables. For example,
+`export http_proxy=proxy.example.com:1080` (and likewise for HTTPS).
```
kinnison: While I dislike the trailing whitespace too, this hunk is unnecessary in the diff

README.md
```diff
@@ -555,7 +555,7 @@ Command | Description
 `rustup override set nightly-2015-04-01` | For the current directory, use a nightly from a specific date
 `rustup toolchain link my-toolchain "C:\RustInstallation"` | Install a custom toolchain by symlinking an existing installation
 `rustup show` | Show which toolchain will be used in the current directory
-`rustup toolchain uninstall nightly` | Uninstall a given toolchain
+`rustup toolchain uninstall nightly` | Uninstall a given toolchain (you cannot uninstall the only toolchain left installed)
```
kinnison, line 558: :+1:

kinnison (changes requested):
One tiny nit left to pick :D  Otherwise :+1:

kinnison:
I don't think the undesirable situation is enough to block this, but it's certainly worth filing an issue for discussion once this PR lands.

README.md
das-sein: No worries. Nits are there to pick. Why not aim for perfection.

das-sein:
@kinnison Unrelated hunks removed from commit and force-pushed to avoid creating a superfluous commit in the history.

And yeah I'm not entirely sure this is the best solution, but it should solve the undesirable state. Hopefully a discussion comes about and some more ideas come up.

kinnison:
I think the three commits might squash nicely into two.  One for the code and one for the README, if you don't mind doing that.

das-sein:
@kinnison Squashed the code commits and left the README commit

nrc:
Could you explain why you want to do this instead of fixing the behviour when the last toolchain is uninstalled? It seems reasonable to me to want to uninstall all toolchains

das-sein:
@nrc I was thinking from a user perspective that it would be more confusing to enter this bad state than it would be to not enter it. It also seemed to me a bit odd that uninstalling all the toolchains would be desirable, since without any toolchains, why have rustup installed at all? If this is totally undesirable, I can work on a proper fix for the logic that causes the bad state instead.

nrc:
>  since without any toolchains, why have rustup installed at all?

I was imagining it would be a  temporary state of affairs, possibly performed by a script?

das-sein:
@nrc Good point! I don't have any response to that, it is a scenario I hadn't considered. I could say that's dirty, but I think I agree that it should be possible for a script to be dirty without causing a bad state. I'll close this PR in a day, just in case anyone wants to chime in, and try to come up with a better solution.

src/rustup-cli/errors.rs
```diff
@@ -27,6 +27,10 @@ error_chain! {
             description("toolchain is not installed")
             display("toolchain '{}' is not installed", t)
         }
+        ToolchainUninstallOnly(t: String) {
```
In-line, line 30: Maybe `UninstallLastToolchain`?

src/rustup-cli/rustup_mode.rs
```diff
@@ -896,6 +896,11 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
 fn toolchain_remove(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
     for toolchain in m.values_of("toolchain").expect("") {
         let toolchain = cfg.get_toolchain(toolchain, false)?;
+        let installed_toolchains = cfg.list_toolchains()?;
+        if installed_toolchains.len() == 1 && toolchain.name() == installed_toolchains[0] {
+            // Do not allow removal of toolchain if it's the only toolchain
```
In-line, line 901: Maybe `Do not allow removal of the last toolchain`?

tests/cli-misc.rs
```diff
@@ -773,12 +773,19 @@ fn toolchain_broken_symlink() {
         // We artifically create a broken symlink toolchain -- but this can also happen "legitimately"
         // by having a proper toolchain there, using "toolchain link", and later removing the directory.
         fs::create_dir(config.rustupdir.join("toolchains")).unwrap();
+        // Have to have at least one toolchain left before attempting to uninstall
```
In-line, line 776: Maybe `Must have at least ....`?

Issue #1283, opened 2017-11-04
When updating "removing component" messages appear instead of "removing old component"


thedrow:
When running `rustup update` you get the following output:
```
info: removing component 'rustc'
info: removing component 'rust-std'
info: removing component 'cargo'
info: removing component 'rust-docs'
info: removing component 'rust-src'
info: removing component 'rust-analysis'
info: removing component 'rls-preview'
info: installing component 'rustc'
info: installing component 'rust-std'
info: installing component 'cargo'
info: installing component 'rust-docs'
info: installing component 'rust-src'
info: installing component 'rust-analysis'
info: installing component 'rls-preview'
```

From a UX perspective, users can be alarmed by a message as "removing component" as I was (woke up an hour ago, didn't drink coffee yet).
It's better to output `Removing previous version of component 'whatever'`.

das-sein:
@thedrow Issue resolved in the attached PR :)

PR #1639, opened 2019-02-05
When updating, show "removing old component" to avoid confusion

das-sein:
Resolution for UX issue described by #1283 

When components were uninstalled during an update previously, the message would be:
```
info: removing component '<component>'
```
This has been changed to reassure the user that only old versions are being uninstalled:
```
info: removing previous version of component '<component>'
```


kinnison (approved):
I like this idea, it's neat and clean. :+1:

nrc:
nice!

PR #1640, opened 2019-02-07
Use println!() over println!("")

dwijnand:


nrc:
Nice, I didn't even know you could do this :-)

dwijnand:
IntelliJ Rust knows and then fastmod bulk modifies :)

PR #1641, opened 2019-02-07
Use if let Some when doing nothing on None

dwijnand:


NickeZ:
Could you propagate the io error from println? i.e.  `writeln!(term2::stdout(), ...)?;`

Issue #1611, opened 2019-01-11
Project: simplification


nrc:
We could simplify Rustup a lot. It was built with some very ambitious goals and it is clear that some of things will never happen, but the code is designed to accommodate them causing unnecessary complexity. It is also backwards compatible with *very* old versions of Rust and Rustup. There is lots of opportunity to reduce complexity. Some ideas:

* remove telemetry (this has never been used and is not that useful in any case)
* remove support for v1 manifests (we should still detect and error out with a message to delete things manually or whatever)
* support fewer ways to install Rust
* use standard logging rather than notifications
* merge some of the crates (probably only rustup and rustup-dist)

cc @rust-lang/cargo for any suggestions of things we should/should not do

dwijnand:
I'd be happy to assist here.  But I've never hacked on Rustup so I'd need a little babysitting.

h-michael:
I think 

- we may delete code for compatible with multirust
- may we need `reqwest` backend download?
- if we need only `curl` support, we maybe merge `download` crate into `rustup-utils`

My suggestion may be besides the point, sorry.

alexcrichton:
I'd personally recommend removing all but the `curl` backend that @h-michael mentioned, and then I'd also look to simplify filesystem transaction handling code. IIRC it's overly complicated and can be implemented far more efficiently than it is today

kinnison:
This kind of simplification sounds sensible.  If you're interested in non-core contributions toward this then I'm very interested in helping.

seanmonstar:
I'd actually recommend removing the curl backend and using only reqwest. Is there some reason we can't have a Rust tool use more Rust and less C?

dwijnand:
@nrc @alexcrichton could you give more information regarding "support fewer ways to install Rust" and "simplify filesystem transaction handling code" please?  If it's likely to generate a back-and-forth, then perhaps in its own issue.

Meanwhile I've opened an issue dedicated to the crate merging: https://github.com/rust-lang/rustup.rs/issues/1644

dwijnand:
Also, it's probably good to spin out the curl vs reqwest discussion into a thread and come to a conclusion.

nrc:
> support fewer ways to install Rust

[...]

PR #1642, opened 2019-02-07
Remove telemetry

dwijnand:
Refs #1611


nrc (approved):
Awesome, love that diff stat :-) r+ with a rebase

PR #1645, opened 2019-02-08
Simplify host triplet passing code

dwijnand:


nrc (commented):
Looks good other than the question about the `rerun` directive

src/rustup-dist/build.rs
```diff
@@ -1,15 +1,6 @@
...
 
 fn main() {
-    let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
     let target = env::var("TARGET").unwrap();
-
-    File::create(out_dir.join("target.txt"))
-        .unwrap()
-        .write_all(target.as_bytes())
-        .unwrap();
-    println!("cargo:rerun-if-changed=build.rs");
```
nrc: Looks like my comment was lost: I thought we would need to keep this directive in order to avoid unnecessary rebuilds?

dwijnand: I was torn about whether to assume this was still necessary or whether it was legacy cruft.  Looks like I went with the latter assumption.  I'll recover it and we can dump it another day if need be.

nrc:
Thanks!

PR #1646, opened 2019-02-08
Shell completions for Cargo

yodaldevoid:
This is a cleanup of a prior PR (#1425) by @ricvelozo.

- Adds the ability to output shell completions for cargo using the command `rustup completions <shell> cargo`
- Updates the completion instructions to mention and suggest placing bash completions in the home directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ but the gist is as simple as using one of the following:
 
 ```
 # Bash
-$ rustup completions bash > /etc/bash_completion.d/rustup.bash-completion
+$ rustup completions bash > ~/.local/share/bash_completion/completions/rustup
 
 # Bash (macOS/Homebrew)
 $ rustup completions bash > $(brew --prefix)/etc/bash_completion.d/rustup.bash-completion
--- a/src/cli/errors.rs
+++ b/src/cli/errors.rs
@@ -1,8 +1,11 @@
 #![allow(dead_code)]
 
+use crate::rustup_mode::CompletionCommand;
+
 use std::io;
 use std::path::PathBuf;
 
+use clap::Shell;
 use error_chain::error_chain;
 use error_chain::error_chain_processing;
 use error_chain::{impl_error_chain_kind, impl_error_chain_processed, impl_extract_backtrace};
@@ -45,6 +48,10 @@ error_chain! {
         }
         WindowsUninstallMadness {
             description("failure during windows uninstall")
+        }
+        UnsupportedCompletionShell(shell: Shell, cmd: CompletionCommand) {
+            description("completion script for shell not yet supported for tool")
+            display("{} does not currently support completions for {}", cmd, shell)
         }
     }
 }
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -164,10 +164,12 @@ r"DISCUSSION:
 
     BASH:
 
-    Completion files are commonly stored in `/etc/bash_completion.d/`.
+    Completion files are commonly stored in `/etc/bash_completion.d/` for
+    system-wide commands, but can be stored in in
+    `~/.local/share/bash_completion/completions` for user-specific commands.
     Run the command:
 
-        $ rustup completions bash > /etc/bash_completion.d/rustup.bash-completion
+        $ rustup completions bash >> ~/.local/share/bash_completion/completions/rustup
 
     This installs the completion script. You may have to log out and
     log back in to your shell session for the changes to take affect.
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -251,7 +251,22 @@ r"DISCUSSION:
     into a separate file and source it inside our profile. To save the
     completions into our profile simply use
 
-        PS C:\> rustup completions powershell >> ${env:USERPROFILE}\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1";
+        PS C:\> rustup completions powershell >> ${env:USERPROFILE}\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1
+
+    CARGO:
+
+    Rustup can also generate a completion script for `cargo`. The script output
+    by `rustup` will source the completion script distributed with your default
+    toolchain. Not all shells are currently supported. Here are examples for
+    the currently supported shells.
+
+    BASH:
+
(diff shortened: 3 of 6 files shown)


kinnison:
Hi @yodaldevoid 

Thanks for your contribution.  Could you possibly tidy up the commit sequence (e.g. fold the fixes to formatting into the requisite commits instead of having a 'run rustfmt' commit) to make it easier to review?

I've put this PR on my list to watch but I'd like to see things a little tidier before I review.

I look forward to your changes.

D

yodaldevoid:
Hello @kinnison 

Thanks for getting to looking over this. I've tidied up the rustfmt commit, but no other commits jumped out at me as being appropriate to tidy up. Is there anything else you would like cleaned up?

I should note that I wanted to keep @ricvelozo's commits mostly intact. They could get merged together with other commits, to condense everything, but I wanted to preserve his contribution.

kinnison:
Fair enough, I can appreciate wanting to maintain that.  I'll put this on my list to review as soon as I can (likely within a day or two)

yodaldevoid:
@kinnison Just wanted to ping you to make sure this hasn't fallen by the wayside.

README.md
```diff
@@ -67,7 +67,7 @@ but the gist is as simple as using one of the following:
 
 ```
 # Bash
-$ rustup completions bash > /etc/bash_completion.d/rustup.bash-completion
+$ rustup completions bash > ~/.local/share/bash_completion/completions/rustup
```
kinnison, line 70: Does bash pick this up by default?  if so, from what version?

src/rustup-cli/rustup_mode.rs
```diff
@@ -1038,3 +1098,52 @@ fn set_default_host_triple(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+}
+
+impl FromStr for CompletionCommand {
+    type Err = String;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        match s {
+            _ if s.eq_ignore_ascii_case("rustup") => Ok(CompletionCommand::Rustup),
+            _ if s.eq_ignore_ascii_case("cargo") => Ok(CompletionCommand::Cargo),
+            _ => Err(String::from("[valid values: rustup, cargo]")),
```
kinnison: I dislike that `variants()` enumerates the possible strings, *this* `match` enumerates the strings, **and** this error string enumerates the strings, which match the completion commands.

I'd almost prefer:

```rust
static COMPLETIONS: &[(&'static str, CompletionCommand)] = &[
    ("rustup", CompletionCommand::Rustup),
    ("cargo", CompletionCommand::Cargo),
];
```

combined with a `from_str` which uses that array, and a `variants()` which maps it into a list of strings, and then this error can be constructed from that list too.  All in all, it just feels a little too denormalised for my liking.

src/rustup-cli/rustup_mode.rs
```diff
@@ -1038,3 +1098,52 @@ fn set_default_host_triple(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+
+fn output_completion_script(shell: Shell, command: CompletionCommand) -> Result<()> {
+    match command {
+        CompletionCommand::Rustup => {
+            cli().gen_completions_to("rustup", shell, &mut io::stdout());
+        }
+        CompletionCommand::Cargo => {
+            let script = match shell {
+                Shell::Bash => Some("/etc/bash_completion.d/cargo"),
+                Shell::Zsh => Some("/share/zsh/site-functions/_cargo"),
```
kinnison: What about other shells, does `cargo` only support bash and zsh?  If so, should we report an error here too if a different shell is selected?

kinnison (changes requested):
First up, thank you for submitting a PR and contributing to `rustup`.

This is not a bad PR, though there are a number of points which need cleanup. The big thing is that there's a distinct lack of tests.

Could we please have some basic tests for the new command you're adding?

Thanks,

D.

src/rustup-cli/rustup_mode.rs
yodaldevoid: `cargo` currently only has completion scripts for `bash` and `zsh`. I agree that we should report an error if a user specifies a shell that is not yet supported. I will add a message printed out on stderr.

README.md
yodaldevoid: `bash-completions` (the package that adds completions to `bash` and is a de-facto standard to the point that people don't realize that `bash` does not have this functionality built in) added this with this commit scop/bash-completion@c89dcbbd5510876f6304ef10806b00cc9fda19dc, which was included in release 2.2 on March 3rd, 2016 (https://github.com/scop/bash-completion/releases/tag/2.2).

ehuss:
FWIW, I'm proposing a new approach for completions for cargo at https://github.com/rust-lang/cargo/issues/6645.  It may not have any impact here, but just wanted to make you aware.

yodaldevoid:
@ehuss I'll keep an eye on that discussion, and when/if it gets implemented if no one else does it I'll make sure to submit a PR moving to the new system for newer versions of `cargo`.

dwijnand:
Thanks, @ehuss, I've been meaning to cross link that here.

yodaldevoid:
@kinnison Sorry about the long time since the last update. As the famous quote goes, "Life... finds a way to always get in the way of projects."

I believe I have addressed all requested changes. Please let me know if there is anything else you would like modified.

kinnison:
Thank you for making the updates, I'll endeavour to review this, likely tomorrow.

yodaldevoid:
@kinnison I hate to be a hypocrite in this regard, but I wanted to ping you about this PR as it has been some time.

kinnison:
> @kinnison I hate to be a hypocrite in this regard, but I wanted to ping you about this PR as it has been some time.

No, that's very fair, I needed a kick up the backside it seems the past two weeks have been busy and somehow this slipped off my TODO without me noticing :(

kinnison (approved):

kinnison:
This looks reasonable to me, I'm grateful for the tests too.  I wonder if the user ought to be told about how the completions script they're given may vary with the default toolchain; but ultimately I think that's not worth blocking this for.

pickfire:
Are there fish completions yet?

tesuji:
If you want to contribute, please help with https://github.com/rust-lang/cargo/issues/6645 .

PR #1654, opened 2019-02-12
Bring output of `rustup show active-toolchain` and `rustup default` into line with rest of rustup

### LINKED ISSUE #1406: rustup show: add flag to show only active toolchain

das-sein:
This resolves: #1406, and brings into line the fix for #1633 

#1406 seems to just be the result of the `rustup show active-toolchain` command not producing the same output as `rustup show`, which includes the override reason. This has been changed so that if `active-toolchain` returns an override, it states the reason for the override. The change to `rustup default` is just to bring its output into line with the rest of rustup in using the "\<toolchain\> (\<status\>)" format.

Previous behavior of `rustup show active-toolchain` while nightly override is set:

```shell
canis@latrans:~/rustup.rs$ rustup show active-toolchain
```

New behavior:

```shell
canis@latrans:~/rustup.rs$ rustup show active-toolchain
nightly-x86_64-unknown-linux-gnu (directory override for '/home/canis/rustup.rs')
```

### DIFF
--- a/src/rustup-cli/rustup_mode.rs
+++ b/src/rustup-cli/rustup_mode.rs
@@ -783,8 +783,12 @@ fn show(cfg: &Cfg) -> Result<()> {
 
 fn show_active_toolchain(cfg: &Cfg) -> Result<()> {
     let ref cwd = utils::current_dir()?;
-    if let Some((toolchain, _)) = cfg.find_override_toolchain_or_default(cwd)? {
-        writeln!(term2::stdout(), "{}", toolchain.name())?
+    if let Some((toolchain, reason)) = cfg.find_override_toolchain_or_default(cwd)? {
+        if reason.is_some() {
+            println!("{} ({})", toolchain.name(), reason.unwrap());
+        } else {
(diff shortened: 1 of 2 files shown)


kinnison (changes requested):
This looks good though I'd like to know why we don't have a test affected by half the patch?

Is there an opportunity to add a test for `rustup default` ?

das-sein:
@kinnison Good point! I did also think it odd, but didn't want to add an unnecessary test in case it was intentional. I'll work on some.

das-sein:
@kinnison Test for `show-active` with override added.

kinnison (approved):
LGTM, thanks for the test.

Issue #1648, opened 2019-02-10
In a directory with an override, display override after succesful default change


carols10cents:
## The problem I am trying to solve

I have some directories where I've set overrides and some directories where I haven't set overrides, and I don't remember which directories have overrides or not.

I'll be working on a project and want to change the Rust version I'm currently using, so I'll run:

```
$ rustup default beta
```

Which outputs:

```
info: using existing install for 'beta-x86_64-apple-darwin'
info: default toolchain set to 'beta-x86_64-apple-darwin'

  beta-x86_64-apple-darwin unchanged - rustc 1.33.0-beta.4 (635817b9d 2019-01-24)
```

So I say, "cool, I'm using beta now!" and go about `cargo build`ing, etc.

In a directory without an override, I am indeed using beta.

In a directory *with* an override, I get the same output, but I am *not* using beta (necessarily), I'm using the override. So I `cargo build` thinking I'm on beta, but I'm not.

## The solution I'd like

If I'm in a directory without an override, I'd like the current behavior of `rustup default` to stay the same.

If I'm in a directory *with* an override, I'd like the output of `rustup default` to remind me that I have an override and that I'm not using the default that I just changed. Something like:

```
$ rustup default beta
info: using existing install for 'beta-x86_64-apple-darwin'
info: default toolchain set to 'beta-x86_64-apple-darwin'

  beta-x86_64-apple-darwin unchanged - rustc 1.33.0-beta.4 (635817b9d 2019-01-24)

(currently using 1.24.1-x86_64-apple-darwin, directory override for '/Users/carolnichols/rust/carol-test')
```

If this is a personal problem of mine, I can script my way around it, but I wanted to file this and see what others thought and then potentially submit a PR if other folks have this problem too.

dwijnand:
Note we're about to make `rustup default` (without any arguments) more informative: https://github.com/rust-lang/rustup.rs/pull/1633, though I don't know how that behaves with overrides present (cc @lazorgator).

Advising the user that even after changing the default a directory override is present sounds like a good idea to me.  I think it should do it for all methods that have precedence on the default, as described in https://github.com/rust-lang/rustup.rs/blob/1.16.0/README.md#override-precedence.  So if there is a rust-toolchain file or a RUSTUP_TOOLCHAIN env var.

das-sein:
@dwijnand At present, the `rustup default` only returns the global default toolchain. Maybe it should detect if there is an override and output that along with the default? Something like:

```shell
[...]

PR #1655, opened 2019-02-12
Tell user if override is set in the environment after `rustup default`

das-sein:
Resolves: #1648

Given: an environment with stable, beta, and nightly installed, and an override for beta set on the current working directory.

Previous Behavior:

```shell
canis@latrans:~/rustup.rs$ rustup default nightly
info: using existing install for 'nightly-x86_64-unknown-linux-gnu'
info: default toolchain set to 'nightly-x86_64-unknown-linux-gnu'

  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.34.0-nightly (57d7cfc3c 2019-02-11)
```

New behavior:

```shell
canis@latrans:~/rustup.rs$ rustup default nightly
info: using existing install for 'nightly-x86_64-unknown-linux-gnu'
info: default toolchain set to 'nightly-x86_64-unknown-linux-gnu'

  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.34.0-nightly (57d7cfc3c 2019-02-11)

info: using override set by current environment
info: beta-x86_64-unknown-linux-gnu (directory override for '/home/canis/rustup.rs')

```

### DIFF
--- a/tests/cli-rustup.rs
+++ b/tests/cli-rustup.rs
@@ -250,6 +250,24 @@ info: default toolchain set to 'nightly-{0}'
 }
 
 #[test]
+fn default_override() {
+    setup(&|config| {
+        expect_ok(config, &["rustup", "update", "nightly", "--no-self-update"]);
+        expect_ok(config, &["rustup", "default", "stable"]);
+        expect_ok(config, &["rustup", "override", "set", "nightly"]);
+        expect_stderr_ok(
+            config,
+            &["rustup", "default", "stable"],
+            for_host!(
+                r"info: using existing install for 'stable-{0}'
+info: default toolchain set to 'stable-{0}'
+info: note that the toolchain 'nightly-{0}' is currently in use (directory override for"
+            ),
+        );
+    });
+}
+
+#[test]
 fn rustup_xz() {
     setup(&|config| {
         set_current_dist_date(config, "2015-01-01");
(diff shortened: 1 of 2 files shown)


kinnison (changes requested):
Could you please find if there's a test this could go in, or else add a test for this so that we don't risk losing the functionality.

Otherwise this is nice.

das-sein:
@kinnison No tests were affected, it seems, so I'll work on adding tests to cover this.

das-sein:
@kinnison Added a test.

tests/cli-rustup.rs
```diff
@@ -249,6 +249,25 @@ info: default toolchain set to 'nightly-{0}'
...
+        expect_ok(config, &["rustup", "default", "stable"]);
+        expect_ok(config, &["rustup", "override", "set", "nightly"]);
+        expect_stderr_ok(
+            config,
+            &["rustup", "default", "stable"],
+            for_host!(
+                r"info: using existing install for 'stable-{0}'
+info: default toolchain set to 'stable-{0}'
+info: using override set by current environment
+info: nightly-{0} (directory override for"
```
dwijnand: Sorry for the late review.

I'm a little concerned with the wording as to me it reads like the run-on sentence "(the) default toolchain (is) set to stable using override set by current environment" (which makes no sense) and then the line after just hangs there unexplained.

I was thinking:

```
info: using existing install for 'stable-{0}'
info: default toolchain set to 'stable-{0}'
info: note that the toolchain 'nightly-{0}' is currently in use (directory override for '{cwd}')
```

It might be a little long, so maybe best cut down to two lines.  But, what do you think?

dwijnand (changes requested):

das-sein:
@dwijnand Warning message simplified, commit squashed.

dwijnand (approved):
Thanks!

kinnison (approved):

PR #1660, opened 2019-02-13
Deprecate cURL

das-sein:
Part of the simplification project: #1611 

Deprecates cURL in favor of reqwest.

This is the preliminary step for the complete disuse of cURL by rustup; on-going discussion: #1657 

cc's from previous PR:
@dwijnand 
@brson 
@briansmith
@jnicholls
@seanmonstar

### DIFF
--- a/src/rustup-utils/src/notifications.rs
+++ b/src/rustup-utils/src/notifications.rs
@@ -64,10 +64,9 @@ impl<'a> Display for Notification<'a> {
             ResumingPartialDownload => write!(f, "resuming partial download"),
             UsingCurl => write!(f, "downloading with curl"),
             UsingReqwest => write!(f, "downloading with reqwest"),
-            UsingHyperDeprecated => f.write_str(
-                "RUSTUP_USE_HYPER environment variable is deprecated,\
-                 use RUSTUP_USE_REQWEST instead",
-            ),
+            UsingHyperDeprecated => {
+                f.write_str("RUSTUP_USE_HYPER environment variable is deprecated.")
+            }
         }
     }
 }
--- a/src/rustup-utils/src/notifications.rs
+++ b/src/rustup-utils/src/notifications.rs
@@ -22,7 +22,6 @@ pub enum Notification<'a> {
     ResumingPartialDownload,
     UsingCurl,
     UsingReqwest,
-    UsingHyperDeprecated,
 }
 
 impl<'a> Notification<'a> {
@@ -39,7 +38,7 @@ impl<'a> Notification<'a> {
             | ResumingPartialDownload
             | UsingCurl
             | UsingReqwest => NotificationLevel::Verbose,
-            UsingHyperDeprecated | NoCanonicalPath(_) => NotificationLevel::Warn,
+            NoCanonicalPath(_) => NotificationLevel::Warn,
         }
     }
 }
@@ -64,9 +63,6 @@ impl<'a> Display for Notification<'a> {
             ResumingPartialDownload => write!(f, "resuming partial download"),
             UsingCurl => write!(f, "downloading with curl"),
             UsingReqwest => write!(f, "downloading with reqwest"),
-            UsingHyperDeprecated => {
-                f.write_str("RUSTUP_USE_HYPER environment variable is deprecated.")
-            }
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -530,9 +530,16 @@ If you need a more complex setup, rustup supports the convention used by
 the __curl__ program, documented in the ENVIRONMENT section of
 [its manual page][curlman].
 
-Note that some versions of `libcurl` apparently require you to drop the 
-`http://` or `https://` prefix in environment variables. For example, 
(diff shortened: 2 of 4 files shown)


dwijnand:
I don't think `RUSTUP_USE_CURL` should be introduced and on usage it emits a warning message.  I think the old `UsingCurl` notification should remain.

das-sein:
@dwijnand Does this mean we should have the user compile themselves with the `default` feature in the Cargo.toml set to cURL if they want to use it?

dwijnand:
No, no compilation necessary, as the rustup binary crate depends on both backends/features:

https://github.com/rust-lang/rustup.rs/blob/194026d2cac3d06c8ef19dc6e23dabda5d3900c4/Cargo.toml#L23

das-sein:
How would a user enable one backend over the other without some kind of environment variable? I think I'm missing something here.

dwijnand:
Ah, I see the confusion.  I meant we _should_ introduce `RUSTUP_USE_CURL` but, if a user sets it, it shouldn't emit a deprecation warning message.  Just "downloading with curl".

lazorgator:
Makes sense! I'll apply that change soon.

das-sein:
@dwijnand Removed the deprecation warning.

dwijnand (approved):
LGTM.  Thanks!

src/rustup-utils/src/notifications.rs
```diff
@@ -20,8 +20,8 @@ pub enum Notification<'a> {
     DownloadFinished,
     NoCanonicalPath(&'a Path),
     ResumingPartialDownload,
-    UsingCurl,
     UsingReqwest,
+    UsingCurl,
```
kinnison, line 23: This delta seems odd

kinnison (dismissed):
This is good, though I'd prefer the odd bogon of a diff hunk was fixed.

src/rustup-utils/src/notifications.rs
das-sein: Yeah I'm not sure why that delta is there, since in the actual source it's in the same place lol. I'll try to see what I can do to fix it.

das-sein:
@kinnison Weird delta removed from commits.

src/rustup-utils/src/notifications.rs
```diff
@@ -64,10 +64,9 @@ impl<'a> Display for Notification<'a> {
             ResumingPartialDownload => write!(f, "resuming partial download"),
             UsingCurl => write!(f, "downloading with curl"),
             UsingReqwest => write!(f, "downloading with reqwest"),
-            UsingHyperDeprecated => f.write_str(
-                "RUSTUP_USE_HYPER environment variable is deprecated,\
-                 use RUSTUP_USE_REQWEST instead",
-            ),
+            UsingHyperDeprecated => {
```
nrc: Lets just remove this.

nrc (approved):
LGTM other than the Hyper thing

src/rustup-utils/src/notifications.rs
das-sein: Pushed!

acid-bong:
Hi there, hello. How safe would it be to drop Curl from dependencies, when building packages for different distros? 

By glancing at big distros and OSes at Repology, I found out only OpenSUSE and FreeBSD compile Rustup without Curl, and I'm currently thinking about dropping that dependency in Nixpkgs (the relevant PR is linked above)

rami3l:
@acid-bong Hello, and thanks for your interest!

As clarified in #4296, we will be dropping the cURL backend soon anyway, so I don't think it'll be extremely helpful to keep supporting it: we are currently keeping it just in case someone has a very particular system environment, and starting from the next release we'll be issuing a warning for them to report to our issue tracker. 

PS: We have shipped since v1.28.0 UA-based tracking for backend usages, but I might need to refer to @rust-lang/infra for the real proportion of users using the cURL backend... Hopefully it'll be reasonably small.

djc:
@decathorpe would be curious to hear your feedback/thoughts! And maybe @sylvestre?

decathorpe:
> @decathorpe would be curious to hear your feedback/thoughts!

I see no problems with dropping the cURL backend in favor of reqwest. The Fedora package will most likely just follow the rustup upstream preference here, if that's why you tagged me :)

sylvestre:
I think we (in Debian) will be fine :)

djc:
> > @decathorpe would be curious to hear your feedback/thoughts!
> 
> I see no problems with dropping the cURL backend in favor of reqwest. The Fedora package will most likely just follow the rustup upstream preference here, if that's why you tagged me :)

Sounds good! Are you defaulting rustup to native-tls over rustls today?

decathorpe:
> > > @decathorpe would be curious to hear your feedback/thoughts!
> > 
> > 
> > I see no problems with dropping the cURL backend in favor of reqwest. The Fedora package will most likely just follow the rustup upstream preference here, if that's why you tagged me :)
> 
> Sounds good! Are you defaulting rustup to native-tls over rustls today?

No, the only feature flag we pass for our build is to enable `no-self-update`, otherwise it follows defaults.

PR #1677, opened 2019-03-02
Add some debugging utilities to rustup-mock

dwijnand:


kinnison (approved):
I'm guessing that you encountered something which led you to want these extra bits of debug.

I have nothing against them being in.

dwijnand:
Yeah I built these while reviewing #1655, but decided to push them separately rather than tagged on there.

kinnison:
Cool, I don't see any danger in including them.

Issue #1678, opened 2019-03-03
Expose clippy-driver as an executable


Manishearth:
We're working on moving cargo-clippy into cargo (https://github.com/rust-lang/rust-clippy/issues/3837). As a part of this, `clippy-driver` will need to be exposed by rustup.

Can clippy-driver be exposed into the bin folder as a shim? It's fine if it has a different name or something

kinnison:
We can add proxies sure, all we'll need is some indication of if it'll be a "normal" proxy or if you'll need something special.  If it's normal, then just give us the name you want and we can add it.

If you do that pretty soon, we might even be able to get it into 1.17.0 which we'd like to release soon (and will then be entering a larger refactoring effort which will make 1.18.0 a longer way off).

Manishearth:
Just needs to be clippy-driver. Users should probably not invoke this themselves so feel free to give it another name, but just clippy-driver is fine.

kinnison:
See https://github.com/rust-lang/rustup.rs/pull/1679

PR #1679, opened 2019-03-04
Add `clippy-driver` as a proxy

kinnison:
The Cargo team are taking `cargo clippy` on-board and as such
will need `clippy-driver` as a general proxy.

### DIFF
--- a/src/rustup-win-installer/src/lib.rs
+++ b/src/rustup-win-installer/src/lib.rs
@@ -27,6 +27,7 @@ static TOOLS: &'static [&'static str] = &[
     "rustfmt",
     "cargo-fmt",
     "cargo-clippy",
+    "clippy-driver",
     "cargo-miri",
 ];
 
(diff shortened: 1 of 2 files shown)


kinnison:
@Manishearth Should we be changing `cargo-clippy` to no longer suggest `clippy` too, or will that be unnecessary?

Manishearth:
No, keep cargo-clippy exactly the same for now. Implementing this will be a long process and we'll need both to be around for a significant amount of time.

We'll make a PR removing cargo-clippy when we need to :smile: 

thanks!

Manishearth (approved):

dwijnand:
Fixes #1678?

kinnison:
> Fixes #1678?

Rats, you're quite right, I forgot that, I'll pop a fix up for that now.

kinnison:
Oops :D  I think that'll be good to merge once green.

Manishearth:
it's green

r? @dwijnand

dwijnand (approved):

Issue #1617, opened 2019-01-19
Improve error message for no space left to upgrade


estebank:
Moved from https://github.com/rust-lang/rust/issues/57759

> I got this: error: `failed to extract package`. Verbose gave no additional info.

CC @mitchtbaum

Nodraak:
Might be related to #330

pickfire:
I believe this is solved? At least the message now shows possibly insufficient space.

kinnison:
Yeah, agreed, thanks @pickfire

PR #1682, opened 2019-03-05
Make the error more verbose

Nodraak:
A small attempt to improve unclear errors. cf. #1617

I can branch/rebase/etc if this PR does not conform to Rust standards.

### DIFF
--- a/src/rustup-dist/src/errors.rs
+++ b/src/rustup-dist/src/errors.rs
@@ -72,7 +72,7 @@ error_chain! {
             display("component manifest for '{}' is corrupt", name)
         }
         ExtractingPackage {
-            description("failed to extract package")
+            description("failed to extract package (perhaps you ran out of disk space?)")
         }
         BadInstallerVersion(v: String) {
             description("unsupported installer version")


src/rustup-dist/src/errors.rs
```diff
@@ -72,7 +72,7 @@ error_chain! {
             display("component manifest for '{}' is corrupt", name)
         }
         ExtractingPackage {
-            description("failed to extract package")
+            description("failed to extract package (make sure you have enough disk space)")
```
kinnison: Are there other reasons why this message might propagate?  if so, then perhaps wording along the lines of ` (perhaps you ran out of disk space?)` might be safer?

kinnison (commented):
If the only cause for this error is out-of-disk-space then LGTM, otherwise maybe needs a wording tweak.

src/rustup-dist/src/errors.rs
Nodraak: I tried to trace the error to find more info, but I could not. I could not find the library or struct or piece of code that could raise it.
What should I do?

kinnison: You may find that you need to grep for `ExtractingPackage` across the entire codebase to find where it gets raised.

Nodraak: Yes, that's what I did, using Github search. I found that it is used 7 times in `rustup.rs/src/rustup-dist/src/component/package.rs`. I tried to trace the origin of the error. I first found the `error_chain` dependency, and from its documentation concluded that it is only a boilerplate/logic helper and has nothing to do with this particular error. In function `unpack_without_first_dir()`, I tried to find the `tar::Archive` object and its method `.entries()` but could not (Yes, it's called struct in Rust, I'm not very familiar yet with Rust naming, sorry).

kinnison: Right, so several of the uses in `package.rs` clearly aren't *purely* caused by an out-of-space issue.  For example, https://github.com/rust-lang/rustup.rs/blob/master/src/rustup-dist/src/component/package.rs#L289 happens at file-open which is *rarely* an out-of-space situation.

I think, given this quick look through the code, I'd very much prefer a less strong wording form.

Nodraak: Make sense, I'll edit my PR

PR #1683, opened 2019-03-06
rustup-init.sh: Fix unset variable usage

### LINKED ISSUE #1684: RUSTUP_UPDATE_ROOT not set by default
The rust install script (https://sh.rustup.rs) when executed fails with the error:
```
sh: 18: RUSTUP_UPDATE_ROOT: parameter not set
(23) Failed writing body
```

I tried it on two different ubuntu setup by executing `curl https://sh.rustup.rs -sSf | sh`. When I set the value manually it worked but as far as I understand it this variable should have been set with a default with a default value.

kinnison:
```
With `set -u` our check for `RUSTUP_UPDATE_ROOT` being empty
needed an additional tweak for it being unset.
```

Lesson learned from 1.17.0 - even when Daniel *thinks* he has checked
`rustup-init.sh` he probably needs to check it again.

/cc @alexcrichton - thanks for the catch.

Fixes #1684

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -15,9 +15,8 @@
 
 set -u
 
-if [ -z "$RUSTUP_UPDATE_ROOT" ]; then
-    RUSTUP_UPDATE_ROOT="https://static.rust-lang.org/rustup"
-fi
+# If RUSTUP_UPDATE_ROOT is unset or empty, default it.
+RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
 
 #XXX: If you change anything here, please make the same changes in setup_mode.rs
 usage() {


kinnison:
As an aside, I'd also like to consider if there's any way we can add to at least one of our travis runs, a verification of the at-least-basic behaviour of `rustup-init.sh` so that clumsy half-asleep release engineers don't miss things like this again :D

dwijnand:
Alternatively: https://github.com/rust-lang/rustup.rs/issues/1685.

dwijnand (approved):

tesuji:
This is what I was trying to solve in #1604, but it is not received enough attention from reviewer. So I removed that fix and forgot to open new PR for this problem.

dwijnand:
Ah, yes, I see that now.  You're right, @lzutao, sorry about that.

PR #1691, opened 2019-03-07
 Use markdown headings in GitHub templates

tesuji:


tesuji:
Closed to favor locally editting.

PR #1692, opened 2019-03-07
Use markdown headings in GitHub templates

tesuji:
r? @dwijnand

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,21 +6,21 @@ labels: bug
 
 <!-- Thanks for filing a 🐛 bug report 😄! -->
 
-**Problem**
+### Problem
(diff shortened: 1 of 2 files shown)


dwijnand:
So I actually like the way it is, because it get the right info out, but it reads less like compiling some government form.  It's more subtle.  As opposed to: "ok! h3 steps, go".

kinnison:
I have no preference either way in terms of the content, and thus since Dale prefers the current and without an impulse to change I naturally prefer status-quo, I think we're suggesting this PR be withdrawn.

Thank you for your efforts though.  You may find that discussing these things on our Discord channel before filing PRs will reduce wasted effort on your part.

tesuji:
@dwijnand Sometime people use bold text to emphasize things. This would collide with bug report form.

Anyway, closed in favor of #1693.

PR #1693, opened 2019-03-07
Update bug_report template

tesuji:


kinnison (commented):
I am confused, half of this appears to be a template change, the other half appears to be a partially filled out template.  What is meant here?

tesuji:
@kinnison Sorry for causing confusing, I have commentted out sample answers in template.

dwijnand:
Could we go back to the drawing board on this, please?  Before jumping into the specifics of the changes you're proposing, could you define what the problems are and/or what you're trying to achieve?

The templates as they are have already gone through multiple rounds of iterations with the team members in the cargo team and the rustup working group, where various alternatives and specifics have been considered.

For starters, this is still switching to H3 headers, which I already mentioned in #1692 I'd like to avoid.

tesuji:
@dwijnand I now see your rationale about not switching to H3 headers.

But by the way, I think a little advantage of the new template is that 
it makes bug reporter supplies their working environment before 
providing problem. Which could help us easier reproducing the issues 
they encountered.

dwijnand:
Why would reporting the working environment _before_ the problem help us more easily reproduce the issue being reported?  Those details are important (which is why the current template captures them) but feel it's better to first capture the user's problem details and then the environmental technical details.

tesuji:
I am used to this template when contributing to other open source project.
Maybe this template doesn't fit with rustup repository.

I would close this PR. Thanks for speding time review it.

PR #1697, opened 2019-03-09
Cleanup CI scripts

tesuji:
* Make these scripts more POSIX-compliant.
 * Fix ShellCheck's complain.

### DIFF
--- a/ci/prepare-deploy-travis.sh
+++ b/ci/prepare-deploy-travis.sh
@@ -2,16 +2,12 @@
 
 set -u -e
 
-if [ "$TRAVIS_PULL_REQUEST" == "true" ]; then
-    exit 0
-fi
-
-if [ "$TRAVIS_BRANCH" == "auto" ]; then
+if [ "$TRAVIS_PULL_REQUEST" = "true" ] || [ "$TRAVIS_BRANCH" = "auto" ]; then
     exit 0
 fi
 
 # Upload docs
-if [[ "$TARGET" == "x86_64-unknown-linux-gnu" && "$TRAVIS_BRANCH" == "stable" ]]; then
+if [ "$TARGET" = "x86_64-unknown-linux-gnu" ] && [ "$TRAVIS_BRANCH" = "stable" ]; then
     # FIXME rust-lang/rust#32532
     printf "not uploading docs"
     #git config --global credential.helper store;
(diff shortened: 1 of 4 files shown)


ci/run.sh
```diff
@@ -2,18 +2,18 @@
 
 set -ex
 
-echo "toolchain versions\n------------------"
+printf 'toolchain versions\n------------------\n'
```
dwijnand, line 5: Why printf instead of echo?

ci/prepare-deploy-travis.sh
```diff
@@ -24,13 +20,13 @@ if [[ "$TARGET" == "x86_64-unknown-linux-gnu" && "$TRAVIS_BRANCH" == "stable" ]]
...
+cp target/"$TARGET"/release/rustup-init target/"$TARGET"/release/rustup-setup
 
 # Generate hashes
-if [ "$TRAVIS_OS_NAME" == "osx" ]; then
-    find "target/$TARGET/release/" -maxdepth 1 -type f -exec sh -c 'shasum -a 256 -b "{}" > "{}.sha256"' \;;
+if [ "$TRAVIS_OS_NAME" = "osx" ]; then
+    find target/"$TARGET"/release/ -maxdepth 1 -type f -exec sh -c 'fn="$1"; shasum -a 256 -b "$fn" > "$fn".sha256' sh {} \;
 else
-    find "target/$TARGET/release/" -maxdepth 1 -type f -exec sh -c 'sha256sum -b "{}" > "{}.sha256"' \;;
+    find target/"$TARGET"/release/ -maxdepth 1 -type f -exec sh -c 'fn="$1"; sha256sum -b "$fn" > "$fn".sha256' sh {} \;
```
dwijnand, line 29: Could you explain this one to me, please?

The previous version looked clean, simple and correct.

dwijnand (commented):
Some nice cleanups!

There are two changes I didn't understand though, so I'd just want to touch base on those first.

Thank you for the PR, @lzutao!

ci/prepare-deploy-travis.sh
tesuji: ShellCheck warns: 
```
find "target/$TARGET/release/" -maxdepth 1 -type f -exec sh -c 'shasum -a 256 -b "{}" > "{}.sha256"' \;;
>>                                                             ^-- SC2156: Injecting filenames is fragile and insecure. Use parameters.
```

ci/run.sh
tesuji: ShellCheck warns:
```
echo "toolchain versions\n------------------"
     ^-- SC2028: echo may not expand escape sequences. Use printf.
```

panaman67:
```suggestion
echo -e "toolchain versions\n------------------"
```
WIll this work?

tesuji: It works because we are using bash, but
```bash
echo -e "toolchain versions\n------------------"
     ^-- SC2039: In POSIX sh, echo flags are undefined.
```

kinnison: I agree that printf is probably more correct.  Especially when we start dealing with BSDs which might use ksh or POSIX sh, or perhaps dash on Debian.

Alternatively we change the instructions to "curl | bash" rather than "curl | sh"

ci/prepare-deploy-travis.sh
kinnison: Is the extra `sh` on that line necessary?  I grok the rest.

dwijnand: Got it.  Thanks.
The extra `sh` is set as `$0` which isn't unused.  I understand it to be so that the argument is a `$1`.

dwijnand (approved):
LGTM now.

PR #1699, opened 2019-03-09
Treat time in seconds as an integer for download times

### LINKED ISSUE #1696: Rustup-Init displays ETA with lots of decimal places
**Problem**

I installed Rust 1.33 using Rustup-Init.exe on Windows 10 (in a Powershell window), and the ETA while installing displayed many decimal places that can be seen below. (See the last line of the picture).

Rustup_Time

**Steps**

1. Download Rustup-Init
2. Install Rust
3. Look at the ETA during installation

**Notes**

Output of `rustup-init --version`: rustup-init 1.17.0 (069c88ed6 2019-03-05)

Manishearth:
We use `/` and `%` in a way that assumes it's an integer.

fixes #1696

### DIFF
--- a/src/rustup-cli/download_tracker.rs
+++ b/src/rustup-cli/download_tracker.rs
@@ -176,29 +176,30 @@ impl fmt::Display for HumanReadable {
 
             if sec.is_infinite() {
                 write!(f, "Unknown")
-            } else if sec > 48. * 3600. {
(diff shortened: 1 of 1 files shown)


kinnison (approved):
LGTM. @dwijnand do you concur?

dwijnand (approved):
Yep

PR #1701, opened 2019-03-09
Switch to git-testament rather than old build.rs

kinnison:
Rather than the previous not-very-informative version string
which was generated by `build.rs` and then not updated unless
`build.rs` was touched, use the new `git-testament` crate which
does a more complete job at compile time via a proc-macro.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,6 +167,16 @@ version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "chrono"
+version = "0.4.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "num-integer 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num-traits 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


src/rustup-cli/common.rs
```diff
@@ -402,11 +404,22 @@ pub fn list_overrides(cfg: &Cfg) -> Result<()> {
...
+    static ref VERSION_STRING: String = {
+        let testament_string = format!("{}", TESTAMENT);
+        let pkg_version = env!("CARGO_PKG_VERSION");
+        if &testament_string[..pkg_version.len()] == pkg_version {
+            testament_string
+        } else {
+            format!("{} really {}", pkg_version, testament_string)
+        }
+    };
+}
```
nrc: I think it would be fine to create the version string from scratch every time, rather than use a lazy static - it's not going to be on a hot path and lazy static is an extra dep and complicates the code

nrc (commented):
Seems like a good idea!

src/rustup-cli/common.rs
kinnison: The issue is that clap requires a `&str` passed to `.version()` and in `rustup_mode.rs` we return a clap `App` from a function which makes lifetimes a bit hard to deal with.

nrc (approved):
Needs rebasing, feel free to merge after that

PR #1704, opened 2019-03-10
Merge rustup's crates

### LINKED ISSUE #1644: Merge some of the crates (probably only rustup and rustup-dist)
Split out from https://github.com/rust-lang/rustup.rs/issues/1611

The current crates tree is:

```
rustup-win-installer
└── rustup
    └── rustup-dist
        └── rustup-utils
            └── download

[dev-dependencies]
└── rustup-mock
```

WDYT about squashing that down to:

```
rustup-win-installer
└── rustup
```

And the only reason to keep those separate is because I can't run rustup-win-installer with its build.rs.  WDYT?

dwijnand:
Fixes #1644


dwijnand:
@nrc / @kinnison let me know if there's anything I can do to make it easier to review this (open to suggestions).  I can present this in a call, if running through it together helps (I've had success with this strategy, as both a submitter and reviewer).

ci/run.sh
```diff
@@ -11,7 +11,6 @@ rustfmt -vV
 cargo build --locked -v --release --target "$TARGET" --features vendored-openssl
 
 if [ -z "$SKIP_TESTS" ]; then
-  cargo test --release -p download --target "$TARGET" --features vendored-openssl
```
kinnison: Why remove the tests for the download crate?

tests/cli-exact.rs
```diff
@@ -1,7 +1,9 @@
 //! Yet more cli test cases. These are testing that the output
 //! is exactly as expected.
 
-use rustup_mock::clitools::{
+pub mod mock;
```
kinnison, line 4: Does this approach mean we end up spending CPU compiling mock over and over?

I'm not entirely upset if the answer is yes, but if so, could it be made a `#[cfg(test)] pub mod mock` in src/ instead?  Would that work?

kinnison (approved):
Pretty much spot on what I was expecting.

Two comments, but if they don't end up implying changes, then :+1:

ci/run.sh
dwijnand: My mistake

tests/cli-exact.rs
dwijnand: Yeah we do.  In cargo this is fixed by making one "testsuite" binary, with all tests.

I'll try the cfg test approach.

dwijnand: Doesn't work.

kinnison (approved):

PR #1707, opened 2019-03-11
Clean up README and add example for `rustup man`

tesuji:
[Rendered version](https://github.com/lzutao/rustup.rs/blob/readme/README.md).

r? @kinnison

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,13 +5,13 @@
 | Travis       | [![Travis Build Status][travis-build-status-svg]][travis-build-status]       |
 | AppVeyor     | [![AppVeyor Build Status][appveyor-build-status-svg]][appveyor-build-status] |
 
-*rustup* installs [The Rust Programming Language] from the official
+*rustup* installs [The Rust Programming Language][rustlang] from the official
 release channels, enabling you to easily switch between stable, beta,
 and nightly compilers and keep them updated. It makes cross-compiling
 simpler with binary builds of the standard library for common platforms.
 And it runs on all platforms Rust supports, including Windows.
 
-[The Rust Programming Language]: https://www.rust-lang.org
+[rustlang]: https://www.rust-lang.org
 
 * [Installation](#installation)
 * [How rustup works](#how-rustup-works)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -281,7 +281,7 @@ $ rustup install stable-msvc
 Toolchain names that don't name a channel instead can be used to name
 [custom toolchains].
 
-[MSVC-based toolchain]: https://www.rust-lang.org/downloads.html#win-foot
+[MSVC-based toolchain]: https://www.rust-lang.org/tools/install
```
dwijnand: Should we link this to https://forge.rust-lang.org/other-installation-methods.html#stable ?

README.md
```diff
@@ -456,7 +456,7 @@ $ rustup target add i686-pc-windows-msvc
 $ rustup target add i686-pc-windows-gnu
 ```
 
-[d]: https://www.rust-lang.org/downloads.html#win-foot
+[d]: https://www.rust-lang.org/tools/install
```
dwijnand: Same here. https://forge.rust-lang.org/other-installation-methods.html#stable ?

dwijnand (commented):
LGTM.  1 question (asked twice).

README.md
tesuji: I don't have any suggestions. I just open rust-lang/www.rust-lang.org#726 to request a mean to view Windows installation page without platform auto-detection.

kinnison: If https://github.com/rust-lang/www.rust-lang.org/pull/729 gets merged then https://www.rust-lang.org/tools/install?platform_override=win ought to work.

README.md
```diff
@@ -258,30 +258,30 @@ Standard release channel toolchain names have the following form:
 <host>          = <target-triple>
 ```
 
-'channel' is either a named release channel or an explicit version
-number, such as "1.8.0". Channel names can be optionally appended with
-an archive date, as in 'nightly-2014-12-18', in which case the
+`channel` is either a named release channel or an explicit version
+number, such as `1.8.0`. Channel names can be optionally appended with
+an archive date, as in `nightly-2014-12-18`, in which case the
```
nrc: I think this change is incorrect - the quoted text are quotes, not code.

nrc (commented):
Thanks for the PR, looks good. I guess we're waiting on the links issue to be resolved before merging this.

tesuji:
rust-lang/www.rust-lang.org#729 gets merged, now https://www.rust-lang.org/tools/install?platform_override=win works.

README.md
tesuji: I found it's easier to read by using backtick. If you have strong objection with it, I will change it back.

dwijnand (approved):

PR #1714, opened 2019-03-16
Support command rustup update nightly/yyyy-MM-dd

liuchong:
Implements https://github.com/rust-lang/rustup.rs/issues/1713

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -23,6 +23,7 @@ use url::Url;
 pub struct Toolchain<'a> {
     cfg: &'a Cfg,
     name: String,
+    desc: String,
     path: PathBuf,
     dist_handler: Box<dyn Fn(crate::dist::Notification<'_>) + 'a>,
 }
@@ -44,11 +45,23 @@ pub enum UpdateStatus {
 
 impl<'a> Toolchain<'a> {
     pub fn from(cfg: &'a Cfg, name: &str) -> Result<Self> {
+        let v: Vec<&str> = name.split('/').collect();
+        let name = v[0];
+
         let resolved_name = cfg.resolve_toolchain(name)?;
         let path = cfg.toolchains_dir.join(&resolved_name);
+
+        let resolved_name_dated = if v.len() > 1 {
+            let name_dated = name.to_string() + "-" + v[1];
+            cfg.resolve_toolchain(&name_dated)?
+        } else {
+            resolved_name.clone()
+        };
+
         Ok(Toolchain {
             cfg: cfg,
             name: resolved_name,
+            desc: resolved_name_dated,
             path: path.clone(),
             dist_handler: Box::new(move |n| (cfg.notify_handler)(n.into())),
         })
@@ -57,7 +70,7 @@ impl<'a> Toolchain<'a> {
         &self.name
     }
     pub fn desc(&self) -> Result<ToolchainDesc> {
-        Ok(ToolchainDesc::from_str(&self.name)?)
+        Ok(ToolchainDesc::from_str(&self.desc)?)
     }
     pub fn path(&self) -> &Path {
         &self.path
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -17,6 +17,7 @@ use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
+use regex::Regex;
 use url::Url;
 
 /// A fully resolved reference to a toolchain which may or may not exist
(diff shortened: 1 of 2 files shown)


src/toolchain.rs
```diff
@@ -44,11 +45,23 @@ pub enum UpdateStatus {
 
 impl<'a> Toolchain<'a> {
     pub fn from(cfg: &'a Cfg, name: &str) -> Result<Self> {
+        let v: Vec<&str> = name.split('/').collect();
```
kinnison: I wonder if `@` might be better than `/` for this?

src/toolchain.rs
```diff
@@ -44,11 +45,23 @@ pub enum UpdateStatus {
...
+        let name = v[0];
+
         let resolved_name = cfg.resolve_toolchain(name)?;
         let path = cfg.toolchains_dir.join(&resolved_name);
+
+        let resolved_name_dated = if v.len() > 1 {
+            let name_dated = name.to_string() + "-" + v[1];
+            cfg.resolve_toolchain(&name_dated)?
```
kinnison: While this will produce a suitable error, it won't be entirely clear that it's due to the `/stuff` part.  I wonder if this needs chaining slightly into a more descriptive error on failure.

src/toolchain.rs
```diff
@@ -44,11 +45,23 @@ pub enum UpdateStatus {
...
 impl<'a> Toolchain<'a> {
     pub fn from(cfg: &'a Cfg, name: &str) -> Result<Self> {
+        let v: Vec<&str> = name.split('/').collect();
+        let name = v[0];
+
         let resolved_name = cfg.resolve_toolchain(name)?;
         let path = cfg.toolchains_dir.join(&resolved_name);
+
+        let resolved_name_dated = if v.len() > 1 {
+            let name_dated = name.to_string() + "-" + v[1];
```
kinnison: This approach to adding the date will fail if any of the platform details are included in the part before the `/` for example, if I wanted nightly from date `YYYY-MM-DD` for platform `i686-pc-windows-gnu` then I'd plausibly type `rustup toolchain install nightly-i686-pc-windows-gnu/YYYY-MM-DD` which will result in an incorrectly formed toolchain descriptor.

kinnison (changes requested):
Firstly, thank you for wanting to contribute to `rustup` and for your work so far.

There are some basic issues with the approach you have put forward which I have hilighted in the comments in this review.

In addition, you haven't provided any tests for this, nor README updates nor information in `rustup help toolchain`.

Finally I'd ask - what will the behaviour be when you run `rustup update` subsequently to running such a 'dated' install command?  Will the chosen date be lost the moment all components are available?

Finally this feels incomplete to me since it provides for no way to update an existing toolchain install to a new date, so anyone with, for example, nightly already installed, will have to uninstall it before trying to install a toolchain of a particular date.

I hope that all the above is not too daunting for you, since this is a feature many have asked for over the past little while, so it'd be excellent to be able to merge it; but as things stand this is not ready yet.

Thanks again for your contribution.

liuchong:
> Firstly, thank you for wanting to contribute to `rustup` and for your work so far.
> 
> There are some basic issues with the approach you have put forward which I have hilighted in the comments in this review.
> 
> In addition, you haven't provided any tests for this, nor README updates nor information in `rustup help toolchain`.
> 
> Finally I'd ask - what will the behaviour be when you run `rustup update` subsequently to running such a 'dated' install command? Will the chosen date be lost the moment all components are available?
> 
> Finally this feels incomplete to me since it provides for no way to update an existing toolchain install to a new date, so anyone with, for example, nightly already installed, will have to uninstall it before trying to install a toolchain of a particular date.
> 
> I hope that all the above is not too daunting for you, since this is a feature many have asked for over the past little while, so it'd be excellent to be able to merge it; but as things stand this is not ready yet.
> 
> Thanks again for your contribution.

Thanks for the responding!

First of all, `@` to intead of `/` should be ok.

As my original thought, I was just want the `nightly@yyyy-MM-dd` just simply replace current `nightly` install, not a new toolchain. It do the very same thing like we run `rustup update nightly` at the day `yyyy-MM-dd`.
So if we run a `rustup update` again once all components are available, the new updated install will replace the dated install, just like nothing special has happened.

Next, let me check and modify the codes carefully to solve the issues and update README and help information.

Anything different options, let us discuss them, thanks!

kinnison:
Thanks for the update (though your code does not pass the tests).

I'm unsure I like the extra use of regular expressions.  Why not trivially parse off the `@stuff` and then use the existing toolchain parser, perhaps adding a method to subsequently sub-in a date?

As a further point, in a recent tools team discussion, we began looking at more generic solutions to this, so while I'll be pleased to look at further code updates, especially if they're passing CI, I'm not sure we'll want to merge this soon.

liuchong:
Oops...I didn't wait the ci finish. Is there a convenient way to make full test locally?

About regex, I'm not very sure if I should use that, but then I found it laid in the Cargo.toml, so... :joy_cat: 

> I'm unsure I like the extra use of regular expressions. Why not trivially parse off the @stuff and then use the existing toolchain parser, perhaps adding a method to subsequently sub-in a date?

But sorry...I'm not very clear with the meaning your word `@stuff`, and because I'm not very familiar with the source codes, I place the patch codes in a place which not very properly. Let me check them again. :smile_cat: 

About the solutions, I think maybe we could use a format like `channel-target@date`, `channel-part_of_target@version`, like:
```
rustup update nightly@2019-03-15
rustup update nightly-i686@2019-03-10
rustup update stable-x86_64-unknown-linux-gnu@1.33.0
```
I'm not very sure about the versioned format, especially about the `beta` like `1.34.0-beta.5`.
I've checked some links like:

> https://forge.rust-lang.org/other-installation-methods.html
> https://github.com/rust-lang/rust/issues/14796

but did not found a index of all archived rust builds of stable/beta/nightly.

If we impl the dated/versioned feature, I think the feature should only works in the `update` stage, should not write anything information about how we have used the command, because,

*a command `rustup update nightly@2019-03-15` run on day 2019-03-20, should just like a command `rustup update nightly` run on 2019-03-15;*
*a command `rustup update stable@1.32.0` run on a day when the latest stable version if larger then 1.32.0, should just like a command `rustup update stable` when the latest stable version is just 1.32.0;*

Maybe we could only impl the `dated nightly` for now. A dated version should only for nightly, isn't it?

kinnison:
You can run `cargo test` to run the test suite locally.

liuchong:
> image

So, seems the newest version resolved this issue in another way automatically, let me close it. 👀

PR #1716, opened 2019-03-17
Force highest TLS version supported

sanmai-NL:
The integrity and confidentiality of the installer script currently hinges on TLS. It is important to enforce the highest version of TLS in the instructions. Also, enforce the `https` scheme. Should redirects occur in the future, then each URL redirected to must be accessed using TLS 1.2 with HTTP, rather than allowing a plain HTTP link in the chain.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -376,9 +376,9 @@ downloader() {
     if [ "$1" = --check ]; then
         need_cmd "$_dld"
     elif [ "$_dld" = curl ]; then
-        curl -sSfL "$1" -o "$2"
+        curl --proto =https --tlsv1.2 --silent --show-error --fail --location "$1" --output "$2"
     elif [ "$_dld" = wget ]; then
-        wget "$1" -O "$2"
+        wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2"
     else
         err "Unknown downloader"   # should not reach here
     fi
--- a/www/index.html
+++ b/www/index.html
@@ -36,7 +36,7 @@
     then follow the onscreen instructions.
   </p>
   <p>If you're a Windows Subsystem for Linux user run the following in your terminal, then follow the onscreen instructions to install Rust.</p>
-  <pre>curl https://sh.rustup.rs -sSf | sh</pre>
+  <pre>curl --proto =https --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
   <p class="other-platforms-help">You appear to be running Windows 32-bit. If not, <a class="default-platform-button" href="#">display all supported installers</a>.</p>
 </div>
 


kinnison:
Hi @sanmai-NL,

Thank you for your contribution to Rustup.  Would it also make sense to adjust `rustup-init.sh` similarly?

sanmai-NL:
Yes, and https://www.rust-lang.org/learn/get-started as well.

As for backward compatibility of a similar change to `rustup-init.sh`: I cannot determine when `--tlsv1.2` was added to cURL. It could have been added in 2001 already, but I'm not sure based on the [cURL changelog](https://curl.haxx.se/changes.html). I have successfully used these CLI options for about 5 years on production systems, though.

kinnison:
If you can verify that the cURL version in RHEL 7 and later distros is suitable then please update your patch to cover rustup-init.sh and we'll probably merge it.  You'll need a separate issue opening on the website repo for that bit though.

sanmai-NL:
[Apparently RHEL 7 has an up-to-date cURL](https://curl.haxx.se/dlwiz/?type=bin&os=Linux&flav=Redhat&ver=RHEL7).

sanmai-NL:
@kinnison: Done. Also added similar options to the `wget` invocation. According to the changelog, [they support these parameters since 2014](https://fossies.org/linux/wget/ChangeLog).

tesuji:
~I afraid that we cannot use rustup in Ubuntu 14.04 (at least in Travis CI).~
Edit: Looks like I am wrong. [curl manpages of Ubuntu 14.04](https://manpages.ubuntu.com/manpages/trusty/en/man1/curl.1.html) lists those options.

sanmai-NL:
As for cURL, `--tlsv1.2` was introduced in 7.34 and `--proto` in 7.20.2 according to https://curl.haxx.se/docs/manpage.html. Ubuntu 14.04 has [an official cURL package based on 7.35](https://launchpad.net/ubuntu/trusty/+source/curl). You can choose [a more up-to-date Ubuntu edition (Xenial)](https://docs.travis-ci.com/user/reference/xenial/) on Travis CI.

kinnison:
We try to maintain an older CI baseline because it increases our reachability wrt. distributions - glibc versions are a pain.

sanmai-NL:
@kinnison: alright, but this is ready for merging no?

kinnison (approved):
LGTM, if noone objects in terms of support for other OSes, I'll merge in the next few days.

sanmai-NL:
@kinnison: ping.

kinnison:
@sanmai-NL Our working group meeting is tonight, I planned to check there before hitting merge.

tesuji:
@kinnison I've just noticed that command may not work on zsh. You have to type:
```console
curl --proto \=https --tlsv1.2 -sSf https://sh.rustup.rs
```
or 
```bash
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs
```
instead.

kinnison:
Hmm, I wonder why I didn't spot that..

*rummages*

Looks like I have `setopt no_equals` in my configs.  Well spotted.  Will you provide a PR to fix that please, and I'll merge it.

www/index.html
```diff
@@ -36,7 +36,7 @@
     then follow the onscreen instructions.
   </p>
   <p>If you're a Windows Subsystem for Linux user run the following in your terminal, then follow the onscreen instructions to install Rust.</p>
-  <pre>curl https://sh.rustup.rs -sSf | sh</pre>
+  <pre>curl --proto =https --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
```
alexcrichton, line 39: Could this be left as it was? This is just intended to be a quick one-liner copy/paste snippet and expanding with so many options can be pretty intimidating as a copy/paste snippet. Does this really add enough benefit relative to what it's displaying on the site?

rustup-init.sh
```diff
@@ -376,9 +376,9 @@ downloader() {
     if [ "$1" = --check ]; then
         need_cmd "$_dld"
     elif [ "$_dld" = curl ]; then
-        curl -sSfL "$1" -o "$2"
+        curl --proto =https --tlsv1.2 --silent --show-error --fail --location "$1" --output "$2"
```
alexcrichton, line 379: It looks like this may be [causing issues on Travis](https://travis-ci.com/rust-lang/cargo/jobs/195079192)?

www/index.html
sanmai-NL: You are expressing certain feelings towards this without backing them up with a UX study. I am sad you prioritize the length of an already technical one-liner over the clear and specified security improvement. I think it is instead a good signal that in the Rust ecosystem, security is taken seriously.

rustup-init.sh
sanmai-NL: We can't tell why without debug logging. We don't know whether it is solely because of this change or due to something else, such as the state of the infra.

kinnison: This has been resolved as safely as we know how with 1.18.1

www/index.html
kinnison: Certainly the display isn't as good as it used to be.  We *could* use web techniques to hide the extra arguments from the display but include them in the copy/paste, and provide a copy-command-to-clipboard button which puts it all there, for ease of use?

sanmai-NL: Why would that be important?

rustup-init.sh
sanmai-NL: @kinnison: Can you please explain a bit more?

www/index.html
kinnison: Currently the box requires scrolling, and people are more likely to fail to copy the whole command than not.  While I personally dislike the narrow styling, I understand why it's there so would prefer to maintain it over trying to do better myself.  If someone else could make it work somehow then that'd be good too.

PR #1725, opened 2019-03-24
Fix man proxy in FreeBSD

tesuji:
In FreeBSD, `man` command only [searches for man files in MANPATH][1]. 
Which means that FreeBSD `man` program does not accept interpret local man files.

On Bash-compatible shell, one could use this command to open rust specific manpages:

```sh
env MANPATH=$HOME/.rustup/toolchains/stable-x86_64-unknown-freebsd/share/man man rustc
```

[1]: https://www.freebsd.org/cgi/man.cgi?query=man&apropos=0&sektion=0&manpath=FreeBSD+12.0-RELEASE+and+Ports&arch=default&format=html

Locally tested on:
- [x] Ubuntu
- [x] FreeBSD 11

r? @kinnison 

**Note:** Ping me if someone needs to squash this PR.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1025,16 +1025,22 @@ fn doc(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
 }
 
 fn man(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
-    let manpage = m.value_of("command").expect("");
+    let command = m.value_of("command").unwrap();
+
     let toolchain = explicit_or_dir_toolchain(cfg, m)?;
-    let mut man_path = toolchain.path().to_path_buf();
-    man_path.push("share");
-    man_path.push("man");
-    man_path.push("man1");
-    man_path.push(manpage.to_owned() + ".1");
-    utils::assert_is_file(&man_path)?;
+    let mut toolchain = toolchain.path().to_path_buf();
+    toolchain.push("share");
+    toolchain.push("man");
+    utils::assert_is_directory(&toolchain)?;
+
+    let mut manpaths = std::ffi::OsString::from(toolchain);
+    manpaths.push(":"); // prepend to the default MANPATH list
+    if let Some(path) = std::env::var_os("MANPATH") {
+        manpaths.push(path);
+    }
     Command::new("man")
-        .arg(man_path)
+        .env("MANPATH", manpaths)
+        .arg(command)
         .status()
         .expect("failed to open man page");
     Ok(())


src/cli/rustup_mode.rs
```diff
@@ -1025,16 +1025,18 @@ fn doc(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
     let toolchain = explicit_or_dir_toolchain(cfg, m)?;
     let mut man_path = toolchain.path().to_path_buf();
     man_path.push("share");
     man_path.push("man");
-    man_path.push("man1");
-    man_path.push(manpage.to_owned() + ".1");
-    utils::assert_is_file(&man_path)?;
+    utils::assert_is_directory(&man_path)?;
+    let mut man_path = man_path.into_os_string();
+    man_path.push(":");
```
lzutao: MANPATH is separated by `:`. Man command needs it as it is.

src/cli/rustup_mode.rs
```diff
@@ -1025,16 +1025,17 @@ fn doc(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
-    let mut man_path = toolchain.path().to_path_buf();
-    man_path.push("share");
-    man_path.push("man");
-    man_path.push("man1");
-    man_path.push(manpage.to_owned() + ".1");
-    utils::assert_is_file(&man_path)?;
+    let mut manpaths = toolchain.path().to_path_buf();
+    manpaths.push("share");
+    manpaths.push("man");
+    utils::assert_is_directory(&manpaths)?;
```
tesuji: Is it necessary to check for existence of the man dir?
Can we assume that Rustup always install manpages in this location?

lzutao:
I think Daniel is tired. So I pick @dwijnand as reviewer.

src/cli/rustup_mode.rs
kinnison: I think it's not unreasonable to fail more informatively if someone types `rustup man` and the manpages moved, the error might be more helpful than a simply missing or incorrect manpage

kinnison (commented):
I think this looks good, though I'm loathe to approve because `-M` overrides the `MANPATH` environment variable which might also be set and relevant.  I'd be more comfortable with manipulating that than passing `-M`

tesuji:
@kinnison So you want to manipulate `MANPATH` env var via [`Command::env`](https://doc.rust-lang.org/std/process/struct.Command.html#method.env)?

kinnison:
I figure that if `MANPATH` is set, we have to do that to be correct (i.e. prepend our path onto `MANPATH`)   If it's not set, we could use `-M` but at that point the code is simpler to always prepend to `MANPATH`

tesuji:
@kinnison I pushed the change you requested.

kinnison (commented):
This looks reasonable to me.  Have you verified it against FreeBSD and at least one Linux?

kinnison:
If so, please rebase to clean up the history of the branch before I merge.

tesuji:
@kinnison I tested on Ubuntu. FreeBSD I only test the equivalent shell command.

kinnison (approved):
LGTM

PR #1726, opened 2019-03-25
Fix comment typo

medwards:
Just something I noticed while looking around to see if `rustup doc` does something special


kinnison (approved):
Thank you for your contribution to Rustup.

PR #1733, opened 2019-03-30
Remove header license info

srinivasreddy42:
https://github.com/rust-lang/rustup.rs/issues/1732


kinnison:
As I said on the issue, I'm unclear why this is valuable to do.

nrc:
I think it is reasonable to do, it brings the script file in line with the rust files (where the license boilerplate was removed along with that in the main Rust repo)

kinnison:
@nrc Fair enough, I'll remove my mild concern then :D

kinnison (approved):
Based on @nrc's agreement, I withdraw my mild concern.

PR #1734, opened 2019-03-30
Various fixups to website

sanmai-NL:
- Use new cURL invocation everywhere.
- Fix naming of macOS.
- Remove redundant `type` attribute of `script` element.


kinnison (approved):
LGTM.

Issue #904, opened 2017-01-06
Optimize disk access during install


brson:
Installation is noticeably slow. I believe there is redundant disk work during installation, associated with the transaction machinery.

brson:
I don't specifically know what is inefficient here, but somebody could do some stracing on Linux and see if there's anything obvious.

iamthad:
`stat` appears to get called 3 times for each file in  `set_file_perms`:

- One to get the file permissions in `needs_x` https://github.com/rust-lang/rustup.rs/blob/67d740fca643f07b38618f42a28757fa988b5ba7/src/rustup-dist/src/component/package.rs#L131
- One inside `utils::is_directory` https://github.com/rust-lang/rustup.rs/blob/67d740fca643f07b38618f42a28757fa988b5ba7/src/rustup-dist/src/component/package.rs#L141
- One to get the file permissions (again) before setting the mode https://github.com/rust-lang/rustup.rs/blob/67d740fca643f07b38618f42a28757fa988b5ba7/src/rustup-dist/src/component/package.rs#L151

This could probably be reduced to just one `stat` call.

kinnison:
Given that the majority of filesystem operations which Rustup performs are reading and writing large (10s of MBs) datasets, I think the `stat()` inefficiency is by no means the most important thing to consider for speed.

I do, however, think that we should look at consider if there's anything we can do to improve install performance on Windows.

mati865:
Installation on Linux filesystems is fast even on HDD so measuring performance there won't help much for Windows but still it'd be nice to make it even faster.

I don't know which one is more limiting NTFS or AV software but shipping less files would be the way to go here.
Nightly on my Linux machine consists of 17957 files, 16617 of them docs. It'd be significant improvement for Windows if they could be fused together into few bigger files.

iamthad:
If I understand correctly, the `set_file_perms` function walks the entire tree of files after they have been extracted to ensure their permissions are correct. This is especially costly in the docs, since there are a tremendous number of files there as @mati865 points out. I would expect tens of thousands of fewer system calls in an installation operation to be noticeable, especially e.g. on NFS.

rbtcollins:
I've done some basic analysis on  Windows 10 Surface; since OS specific code paths can make Linux traces (while convenient to grab) misleading :). My test setup had Windows Defender running normally; controlled folder access was off, no defender exclusions defined.

The high level view is that for docs we spend:
2m20s extracting ~16K files
[...]

PR #1744, opened 2019-04-09
Less copying during dist installation

rbtcollins:
Per #904 copying the contents of dists out of the extracted staging
directory and then later deleting that same staging directory consumes
60s out of a total of 200s on Windows.

Wall clock testing shows this patch reduces
`rustup toolchain install nightly` from 3m45 to 2m23 for me - including
download times etc.

I'm sure there is more that can be done, thus I'm not marking this as
a closing merge for #904.

### DIFF
--- a/src/dist/component/components.rs
+++ b/src/dist/component/components.rs
@@ -109,7 +109,16 @@ impl<'a> ComponentBuilder<'a> {
             .push(ComponentPart("dir".to_owned(), path.clone()));
         self.tx.copy_dir(&self.name, path, src)
     }
-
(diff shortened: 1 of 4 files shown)


kinnison (approved):
Over all, I think this looks good.  I like the cleanup for `dest_abs_path()` and the copy-vs-move semantics seem sound to me.

If at all possible, I'd love to see if @brson has enough memory of this code area to comment, otherwise given the tests are green I'm okay for merging this.

kinnison:
I can confirm that this branch has been working fine for me in local testing for 2 days.

nrc (approved):
LGTM, thanks!

PR #1749, opened 2019-04-10
Add test for dumb terminal

tesuji:
cc @kinnison 

Do we need to use [`isatty`] crate instead of manual hand code? But looks like 
its repo has been archived.

Test for #1747.

[`isatty`]: https://crates.io/crates/isatty`

### DIFF
--- a/tests/cli-misc.rs
+++ b/tests/cli-misc.rs
@@ -122,6 +122,17 @@ fn update_works_without_term() {
     });
 }
 
+// Issue #1738
+#[test]
+fn show_works_with_dumb_term() {
+    setup(&|config| {
+        let mut cmd = clitools::cmd(config, "rustup", &["show"]);
+        clitools::env(config, &mut cmd);
+        cmd.env("TERM", "dumb");
+        assert!(cmd.spawn().unwrap().wait().unwrap().success());
+    });
+}
+
 // Issue #140
 // Don't panic when `target`, `update` etc. are called without subcommands.
 #[test]
--- a/src/utils/tty.rs
+++ b/src/utils/tty.rs
@@ -1,45 +1,38 @@
 // Copied from rustc. atty crate did not work as expected
 #[cfg(unix)]
 pub fn stderr_isatty() -> bool {
-    unsafe { libc::isatty(libc::STDERR_FILENO) != 0 }
+    isatty(libc::STDERR_FILENO)
 }
 
 // FIXME: Unfortunately this doesn't detect msys terminals so rustup
 // is always colorless there (just like rustc and cargo).
 #[cfg(windows)]
 pub fn stderr_isatty() -> bool {
-    type DWORD = u32;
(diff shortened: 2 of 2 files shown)


tests/cli-misc.rs
```diff
@@ -122,6 +122,17 @@ fn update_works_without_term() {
     });
 }
 
+// Issue #1738
+#[test]
+fn show_works_with_dumb_term() {
```
kinnison, line 127: Unfortunately I'm not convinced this works anyway because the "output" is not a terminal when running under our test setup, it's a pipe.

tesuji: The test is failed before your patch. It was weird that if I use `cmd.output()`, the test is passed.
But with `cmd.spawn` the test is failed.

kinnison: Aah so .spawn() is leaving stdout/stderr going to the terminal.  Do you see output from rustup show or is that still swallowed somehow?  Also was it failing in CI beforehand?  I'm not entirely sure if travis et al run their tests in a pseudoterminal or not.

tesuji: From [the docs](https://doc.rust-lang.org/nightly/std/process/struct.Command.html#method.spawn), it tells:
- `spawn`: Executes the command as a child process, returning a handle to it.
By default, stdin, stdout and stderr are _inherited_ from the parent.
- `output`: Executes the command as a child process, waiting for it to finish and collecting all of its output.
By default, stdout and stderr are _captured_ (and used to provide the resulting output).

kinnison: Yes, I'd just not spotted the use of `.spawn()` rather than `.output()`.  What I'm concerned about is that this test might do nothing in Travis/Appveyor if they don't use pseudoterminals for builds (I'm not sure if they do or not)  Do they?

kinnison: Given the colour I'm going to guess they might, *BUT* this ends up with https://travis-ci.com/rust-lang/rustup.rs/builds/107899773#L5995 type test output spew which isn't great.

I guess I'm not sure if I think the messy output of the tests is worth it for adding this test, certainly I think the complexity of a cross-platform pseudoterminal library dependency addition might be too much though :(

tesuji: Here's the link that the test is failed on Travis before your patch: https://travis-ci.com/rust-lang/rustup.rs/builds/107833119#L6000

tesuji:
Rebased on master.

tests/cli-misc.rs
kinnison: OK, I accept that Travis must be testing it with a PTY, that's good.

I'm still not sure it's worth the test spew but I suppose it's very small spew so for now we'll accept it.

kinnison (approved):
Over-all it looks good to me.  I'm making the leap that `isatty` works well on all our platforms but it seems likely.

Issue #1740, opened 2019-04-08
`rustup install` is missing from `rustup -h`


PaulBone:
**Problem**
The `rustup install` sub-command is missing from the built-in documentation (`rustup -h`).  I learnt about it here: https://github.com/rust-lang/rustup.rs/blob/master/README.md#toolchain-specification.

Maybe there are other undocumented commands?

Thank you.

**Notes**

Output of `rustup --version`:
rustup 1.16.0 (beab5ac2b 2018-12-06)

Output of `rustup show`:
Default host: x86_64-unknown-linux-gnu

installed toolchains
--------------------

stable-i686-unknown-linux-gnu
stable-x86_64-unknown-linux-gnu

active toolchain
----------------

stable-x86_64-unknown-linux-gnu (default)
rustc 1.32.0 (9fda7c223 2019-01-16)

kinnison:
Hmm, the top level `rustup install` and `rustup uninstall` commands come as part of the `rustup toolchain` command, hence

```rust
                .setting(AppSettings::Hidden) // synonym for 'toolchain install'
```

in `fn cli()` in `src/cli/rustup_mode.rs`.  I think the *correct* fix here is to the `README.md` to use the `rustup toolchain install` form instead.

nrc:
`rustup install` is just an alias for `rustup update` and `update` is documented.

PaulBone:
The fix in #1750 looks good. Thanks.

PR #1750, opened 2019-04-11
README: Update to use the toolchain install subcommand

kinnison:
Previously we were using `rustup install` which is a hidden
deprecated alias for `rustup update`.  Instead we should use
`rustup toolchain install` which is the documented install
method since it has its counterpart `rustup toolchain uninstall`

This may assist in ameliorating the confusion which led to #1740

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,13 +179,13 @@ info: downloading self-updates
 ## Working with nightly Rust
 
 Rustup gives you easy access to the nightly compiler and its
-[experimental features]. To add it just run `rustup install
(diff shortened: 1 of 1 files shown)


nrc (approved):
This seems fine (feel free to merge if you like), we could also use `update` which is less typing for the user, though admittedly a less obvious name for the task, take the r+ for either version.

kinnison:
While I agree that `update` is shorter than `toolchain install` it doesn't have the pleasing symmetry with `toolchain uninstall` and as such I'd rather not encourage it.  Thanks for the r+

PR #1753, opened 2019-04-11
Correctly document rustup defaults on Windows

rbtcollins:
rustup checks the host OS for the CPU arch

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,17 +423,25 @@ of Visual Studio 2013 (or later) or the Visual C++ Build Tools
 check the "C++ tools" option. No additional software installation is
 necessary for basic use of the GNU build.
 
-By default rustup on Windows configures Rust to target the 32-bit MSVC
-ABI, that is the `i686-pc-windows-msvc` target triple. More
-specifically, the toolchains that rustup chooses to install, unless
+By default rustup on Windows configures Rust to target the MSVC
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -423,17 +423,17 @@ of Visual Studio 2013 (or later) or the Visual C++ Build Tools
...
-by default. When you write `rustup update nightly`, rustup interprets
-it as `rustup update nightly-i686-pc-windows-msvc`. You can change this
-behavior with `rustup set default-host` or during installation.
+to run on that target triple host and will target that triple by default.
+
+You can change this behavior with `rustup set default-host` or during installation.
 
 ```console
-$ rustup set default-host x86_64-pc-windows-msvc
+$ rustup set default-host x86_64-pc-windows-gnu
```
kinnison, line 444: I *think* this bit was meant to be demonstrating the change in `i686` vs. `x86_64` rather than `msvc` vs `gnu` which I think is addressed elsewhere in the README.

rbtcollins:
Please feel free to edit the pr to suit. I changed this to change the
target abi rather than arch because I suspect it's more common, and still
demonstrates setting one of the for options listed immediately below.

kinnison (approved):
I've tweaked the wording and think this is good to merge now

nrc:
Thanks both!

PR #1758, opened 2019-04-13
Remove global allocator

tesuji:
System allocator has been used by default since Rust v.1.32.


tesuji:
Closed in favor of #1759 .

PR #1759, opened 2019-04-13
Clean up dependency and code base

### LINKED ISSUE #1755: Replace deprecated dependent crates

I suggest to replace deprecated (abandoned) dependent crates like:

- [x] ~`markdown` => `pulldown_cmark`~
  Rationale: rustup currently needs to iterate through `markdown::Block`
  to print content to std{out,err} approriately. But `pulldown_cmark`
  doesn't provide equivalent methods (only `push_html` and `write_html`).

- [x] ~`ole32-sys` => `winapi`~
  Rationale: I remove `ole32-sys` from `Cargo.toml` and rustup build fine.

- [x] ~`time` => `chrono`~
  Rationale: `chrono` currently use `time` as dependency itself.

Just try to replace them. Build and fix cycle :) .

**Notes**

Nothing here.

tesuji:
This pull request:
- Remove unused dep: ole32-sys
- Remove unused dep: itertools
- Remove global allocator
- Make truly random string
- Sort deps according to alphabet order
- perf: Use lazy_static with regex

Closes #1755.


src/cli/help.rs
```diff
@@ -1,4 +1,4 @@
-pub static RUSTUP_HELP: &'static str = r"DISCUSSION:
+pub const RUSTUP_HELP: &str = r"DISCUSSION:
```
nrc: I don't think we should make this change - we might want to compare string pointers at some point and this makes that a footgun (or hopefully impossible?). In general, I think use consts when we care about the value, and statics where we care about the pointer (which I think we do for strings, etc).

src/cli/self_update.rs
```diff
@@ -1378,8 +1377,14 @@ fn get_new_rustup_version(path: &Path) -> Option<String> {
...
-    let re = Regex::new(r"\d+.\d+.\d+[0-9a-zA-Z-]*").unwrap();
-    let capture = re.captures(&version);
+    use lazy_static::lazy_static;
+    use regex::Regex;
+
+    lazy_static! {
+        static ref RE: Regex = Regex::new(r"\d+.\d+.\d+[0-9a-zA-Z-]*").unwrap();
+    }
+
+    let capture = RE.captures(&version);
```
nrc, line 1387: Have you confirmed this change is a performance win? There is some added complexity to the code (and lazy statics are not free - they use TLS and require taking a lock, I think), so this change could make no difference or even make things worse (as well as making the code a little more complicated).

src/cli/self_update.rs
```diff
@@ -622,6 +621,7 @@ fn install_bins() -> Result<()> {
 }
 
 pub fn install_proxies() -> Result<()> {
+    use same_file::Handle;
```
nrc: I don't think this change is worth making.

nrc (commented):
A few questions inline. Thanks for the PR!

src/cli/help.rs
tesuji: I think this variable is just a constant reference to a string.

src/cli/self_update.rs
tesuji: I think so, because regex docs seems to encourage this use of lazy_static: https://docs.rs/regex/1.1.5/regex/#example-avoid-compiling-the-same-regex-in-a-loop

src/cli/self_update.rs
tesuji: Done.

kinnison:
Mostly LGTM but needs nick's comments resolving.

src/cli/self_update.rs
nrc: :+1:

src/cli/help.rs
tesuji: So the `const` keyword differs with `static` that it makes this string
immutable. Everything else is the same between these keywords.

retep998: ~While the address of a `const` is is not fixed, if the constant is itself a pointer then *that* pointer **is** fixed.~ (EDIT: See my comment below)

nrc: I think (I'm actually not sure about this) that the compiler can still duplicate the data with consts. I'm not sure if that is better or worse

tesuji: I built rustup in release mode, the binary size of with this PR is the same
with master branch.

lzutao:
Rebased on master.

src/cli/help.rs
retep998: After some more investigation, the address of a const is not fixed nor are any of the pointers in the const. This can be easily demonstrated by having two dylibs use the same constant string and print different addresses for the string. The address of a static is fixed however. So if fixed addressing is *required* then use `static`, otherwise use `const`.

tesuji: I really don't get the point why we need fixed addresses of strings in binary file.

kinnison: At this point, if there's concern from at least one dev, why insist on this change blocking the rest of your cleanups?  It would make sense to me to remove this static->const change for now, to allow us a better chance to merge the rest (once the argument about `local` is dealt with)

tesuji: After read some comparisons between using const and static, I'm convinced 
that static would be better in this case.

Relevant links:
* https://doc.rust-lang.org/1.30.0/book/first-edition/const-and-static.html
* https://stackoverflow.com/questions/37877381/what-is-the-difference-between-immutable-and-const-variables-in-rust
* https://github.com/frol/rust-rfcs/blob/master/text/0246-const-vs-static.md

I will mark this conversation as resolved.

tesuji:
Is there anything I can do to improve this pull request?

kinnison (approved):
A final pass and this LGTM.  Thank you for the `libc::c_char` fix, since I was about to have to do that myself :D

PR #1760, opened 2019-04-13
Improve build time of x86_64-linux-gnu target

tesuji:
By using prebuilt perl, we could reduce build-time up to 3 minutes.

### DIFF
--- a/ci/docker/x86_64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/x86_64-unknown-linux-gnu/Dockerfile
@@ -1,9 +1,6 @@
+# Source from https://github.com/rust-lang-deprecated/rust-buildbot/blob/master/slaves/dist/Dockerfile
 FROM alexcrichton/rust-slave-dist:2015-10-20b
 USER root
 
 WORKDIR /
-RUN curl https://www.cpan.org/src/5.0/perl-5.28.0.tar.gz | tar xzf -
-WORKDIR /perl-5.28.0
-RUN ./configure.gnu
-RUN make -j$(nproc)
-RUN make install
+RUN curl -L https://github.com/lzutao/rustup.rs/releases/download/perl-5.28.0/perl-5.28.0.tar.gz | tar xzf - -C /


ci/docker/x86_64-unknown-linux-gnu/Dockerfile
```diff
@@ -1,9 +1,6 @@
...
 FROM alexcrichton/rust-slave-dist:2015-10-20b
 USER root
 
 WORKDIR /
-RUN curl https://www.cpan.org/src/5.0/perl-5.28.0.tar.gz | tar xzf -
-WORKDIR /perl-5.28.0
-RUN ./configure.gnu
-RUN make -j$(nproc)
-RUN make install
+RUN curl -L https://github.com/lzutao/rustup.rs/releases/download/perl-5.28.0/perl-5.28.0.tar.gz | tar xzf - -C /
```
tesuji, line 6: We should upload it to somewhere rather than in my own fork.

kinnison: Certainly I wouldn't like this part of the change.  I do feel like we can find a better solution where we don't use non-packaged perl at all, but I don't see what it is yet.

tesuji: It's better but not the best solution we have. Should we ping alexcrichton 
to see what he think about this?

kinnison: Yeah... @alexcrichton If you have a moment could you weigh in on a better way to do this part of `rustup` CI?  I really don't like storing built binaries of Perl like this, but I also dislike that we end up rebuilding perl an awful lot.

pickfire:
@lzutao Is it possible to use POSIX-compatible shell instead of bash? The CI script named `.bash` does seem to have POSIX-compatible code, should it be renamed to `sh`?

tesuji:
@pickfire All shell scripts should be POSIX-compliant now, except the `local` keyword.

.travis.yml
```diff
@@ -64,10 +63,13 @@ script:
         *)                   DOCKER=$TARGET ;;
       esac;
       if [ -n "$DOCKER" ]; then
-          sh ci/build-run-docker.sh "$DOCKER" "$TARGET" "$SKIP_TESTS";
```
pickfire: Since the scripts are POSIX-compatible, it could instead use `sh` instead of `bash`.

pickfire:
@lzutao Is it possible to remove to `local` keyword? Nice, the files have been renamed to `.sh`.

tesuji:
@pickfire I don't think that removing `local` keyword is necessary. The `local`
keyword protects helper functions from messing up with each others because
of global variables.

.travis.yml
tesuji: Done. Let's see if travis is green.

pickfire:
@lzutao I understand your concern but different parts of scripts are run separately and the use of `local` break running the script with `dash`.

tesuji:
IIRC, dash supports `local` keyword. Those scripts are running fine with 
`dash` on my machine and on CI.

ci/docker/x86_64-unknown-linux-gnu/Dockerfile
alexcrichton: Instead of a prebuilt perl, could this perhaps take the same strategy as https://github.com/rust-lang/rustup.rs/pull/1724 and use the prebuilt images from rust-lang/rust for deployment?

tesuji: Did you mean to use [crosss x86_64 linux dockerfile][1]? Or just build an new
docker image based on alexcrichton/rust-slave-dist:2015-10-20b and upload
it to Docker Hub?

[1]: https://github.com/rust-embedded/cross/blob/master/docker/x86_64-unknown-linux-gnu/Dockerfile

alexcrichton: No in #1724 it updates to use the exact same images as rust-lang/rust uses for its compilers. The images were mostly used for cross targets like powerpc, but x86_64-unknown-linux-gnu can be moved over as well.

kinnison: I think I'd very much prefer to use the exact same images as `rust-lang/rust` for all our platforms tbf.

tesuji: I tried a bit in my local machine. I think the image size of `rust-x86_64-unknown-linux-gnu` 
from `rust-lang/rust` is rather big (3.13GB) for project likes rustup (with download
time + extracted time, I expect build time will increase). I will measure it in my own fork's
CI.

pickfire:
Wait, I tested it on function level and it works but not on global level.

tesuji:
I think currently we only use `local` only in function's block, not on global block
as I know. @pickfire

ci/docker/x86_64-unknown-linux-gnu/Dockerfile
tesuji: I did the change in [my branch](https://github.com/lzutao/rustup.rs/tree/rlrust).
The [Travis build log](https://travis-ci.com/lzutao/rustup.rs/builds/108571027) here.

There are two blocking points:
* The build is failed, but I don't see why it's failed. It complains about
  ``linker `cc` not found``, but gcc and g++ is included in rust
  x86_64 and i686 docker image?
* The build time is increase. The average build time is 17 minutes and 30 
  seconds (with build succeed and all tests passed).
  Now the build time is at least 18 minutes (with tests uncompleted).

kinnison: It's possible that travis will cache the image now for us, reducing the cost of using a larger docker image, but the `cc` not found is a pain.

tesuji: @kinnison We use rust's images. Those images could be changed or updated so the 
cache will be invalid (correct me if I'm wrong).

kinnison: It would, but not very often.  (I think at least every 90 days but likely more often than that) so I think it won't be too bad.

tesuji: I forgot. We download the tar-compressed docker images from aws with curl (about 1GB). So I don't think Travis would (automatically) cache this file for us. We could force Travis cache this file, but it's not recommended by Travis and is inefficient.

tesuji:
Closed in favor of  #1815 .

PR #1762, opened 2019-04-14
Add `--embedded-book` flag to `rustup doc`

ryochack:
Added `rustup doc --embedded-book` flag to open The Embedded Rust Book directly.


kinnison (approved):
LGTM

PR #1767, opened 2019-04-15
Avoid calling `set_permissions` if the mode is ok

### LINKED ISSUE #1638: rustup install fails when rustup-init is owned by another user
Not sure if this is a rustup or Homebrew rustup-init issue, but I found that installing `rustup-init` using Homebrew, and then running `rustup-init` as a different user resulted in a permissions error (shown below). The result was that `rustup` would install toolchains, but never set up the binaries in `~/.cargo/bin`.

```
Current installation options:

   default host triple: x86_64-apple-darwin
     default toolchain: stable
  modify PATH variable: yes

2) Customize installation
3) Cancel installation
>

error: failed to set permissions for '/Users/p4nathan/.cargo/bin/rustup'
info: caused by: Operation not permitted (os error 1)
```

Changing the ownership of `/usr/local/bin/rustup-init` to the user running `rustup` averted the issue and everything installed as expected. Maybe I'm weird in that I want one installation of `rustup-init` but have multiple accounts on the same system using `rustup`. At least for now I can work around this.

lallotta:
Fixes #1638

### DIFF
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -389,9 +389,15 @@ pub fn make_executable(path: &Path) -> Result<()> {
             path: PathBuf::from(path),
         })?;
         let mut perms = metadata.permissions();
-        let new_mode = (perms.mode() & !0o777) | 0o755;
-        perms.set_mode(new_mode);
+        let mode = perms.mode();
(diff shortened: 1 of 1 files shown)


kinnison (approved):
LGTM

PR #1776, opened 2019-04-19
Enable shellcheck and fold Travis log

tesuji:
Split from #1760
r? @kinnison

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,9 +50,9 @@ matrix:
       os: osx
       osx_image: xcode9.2
       if: branch != master
-      env: MACOSX_DEPLOYMENT_TARGET=10.7 TARGET=i686-apple-darwin
-    - <<: *mac
       env: MACOSX_DEPLOYMENT_TARGET=10.7 TARGET=x86_64-apple-darwin NO_ADD=1
+    - <<: *mac
+      env: MACOSX_DEPLOYMENT_TARGET=10.7 TARGET=i686-apple-darwin
 
 install:
   - if [ -z "$NO_ADD" ]; then rustup target add "$TARGET"; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: required
 dist: trusty
 services:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,10 +1,15 @@
-sudo: required
-dist: trusty
+dist: xenial
```
kinnison, line 1: Does this result in us needing newer glibc on some platforms?

.travis.yml
```diff
@@ -1,10 +1,15 @@
...
 services:
   - docker
 language: rust
 
+git:
+  # FIXME: Issue #1783
+  #depth: 1
```
kinnison: Is there a significant speed benefit in shallow cloning here?

kinnison (commented):
Apart from the worry about the trusty -> xenial jump, I'm okay with this.  What's the impact of that jump in terms of supporting users?

.travis.yml
tesuji: We're using docker images to solve this problem.

.travis.yml
tesuji: Neglectable :) I just wonder if someone download zipped of rustup
and run `cargo test`, then it would be failed, which will confuse
them.

.travis.yml
kinnison: Oh right, so this is the outer bit only, I get it.

.travis.yml
kinnison: The bug report is fair, that's a problem for me to fix, but for this, I'm not sure we'd want depth:1 anyway since we *want* git history information.

kinnison:
I'm pretty much prepared to merge this, except I think I'd like it if the depth bit was removed rather than suggesting we'd be happy to unremove it once the bug is fixed.  I'd much prefer to always have full git history for the CI in case we add further diagnostics.

tesuji:
I removed the depth flag. Now Travis will clone full depth when building.

kinnison (approved):
LGTM

PR #1778, opened 2019-04-19
Fix clippy warnings and errors

tesuji:
Clippy log here: https://travis-ci.com/lzutao/rustup.rs/jobs/194219923#L573

**Note**: I categorized each kind of clippy warnings to each commit. So it would be 
easier to review if you read commit one-by-one rather than read it all
in `File changed` tab.

### DIFF
--- a/download/tests/download-curl-resume.rs
+++ b/download/tests/download-curl-resume.rs
@@ -1,5 +1,6 @@
 #![cfg(feature = "curl-backend")]
 
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
 
 use url::Url;
@@ -35,7 +36,7 @@ fn callback_gets_all_data_as_if_the_download_happened_all_at_once() {
 
     let from_url = format!("http://{}", addr).parse().unwrap();
 
-    let callback_partial = Mutex::new(false);
+    let callback_partial = AtomicBool::new(false);
     let callback_len = Mutex::new(None);
     let received_in_callback = Mutex::new(Vec::new());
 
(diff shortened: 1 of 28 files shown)


download/tests/download-curl-resume.rs
```diff
@@ -47,9 +48,8 @@ fn callback_gets_all_data_as_if_the_download_happened_all_at_once() {
         Some(&|msg| {
             match msg {
                 Event::ResumingPartialDownload => {
-                    let mut flag = callback_partial.lock().unwrap();
-                    assert!(!*flag);
-                    *flag = true;
+                    assert!(!callback_partial.load(Ordering::SeqCst));
+                    callback_partial.store(true, Ordering::SeqCst);
```
kinnison, line 52: Does this have the same side-effect of locking callback_partial for the duration of the read/assert/set sequence?

download/tests/download-reqwest-resume.rs
```diff
@@ -47,9 +48,8 @@ fn callback_gets_all_data_as_if_the_download_happened_all_at_once() {
         Some(&|msg| {
             match msg {
                 Event::ResumingPartialDownload => {
-                    let mut flag = callback_partial.lock().unwrap();
-                    assert!(!*flag);
-                    *flag = true;
+                    assert!(!callback_partial.load(Ordering::SeqCst));
+                    callback_partial.store(true, Ordering::SeqCst);
```
kinnison, line 52: Ditto, locking semantics - do they change?

kinnison (commented):
Apart from my one (probably irrelevant) concern about lock semantics in the mutex->atomicbool change, I'm happy

download/tests/download-curl-resume.rs
tesuji: [I think so](https://doc.rust-lang.org/nightly/std/sync/atomic/enum.Ordering.html#variant.SeqCst).

PR #1785, opened 2019-04-22
Enable building doc again

tesuji:
Demo is at https://lzutao.github.io/rustup.rs .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,13 @@ matrix:
       os: linux
       env: TARGET=i686-unknown-linux-gnu
       if: branch != master
+    - os: linux
+      env: TARGET=rustdoc-x86_64-unknown-linux-gnu
+      install: true
+      script: true
+      before_deploy:
+        - cargo doc --no-deps --features vendored-openssl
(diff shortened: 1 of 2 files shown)


kinnison:
I'm not anxious for this, not least because it implies content on rust-lang.github.io which is stuff we're not at all intending to support as an external API.  What's your rationale for wanting this?

tesuji:
Maybe we should ask @brson, [he was the one who disabled uploading docs](https://github.com/rust-lang/rustup.rs/commit/be86e1f7731785e0ef5718678914d16f8d8699d4).

tesuji:
Ping @brson , do you have any thoughts on this PR?

kinnison:
Until and unless we intend to improve internal code documentation to help new hackers on rustup, I don't think this is worthwhile since it might encourage people to think we support them linking to rustup code in their own projects, which we really don't want to do.

nrc:
I think I agree with @kinnison - we don't want to give the impression that using Rustup as a library is OK. It's easy enough for devs to do `cargo doc --open`, I don't see much advantage in publishing docs.

Sorry, and thanks for the PR!

PR #1786, opened 2019-04-22
ci: Limit shellcheck to x86_64-linux

tesuji:
We only need to run this once.


kinnison (approved):
This seems sane, thanks.

PR #1787, opened 2019-04-22
Fixed panic when no toolchain was configured

Glamhoth:
Fixed panic when no toolchain was configured and user run
`rustup default`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -344,8 +344,11 @@ impl Cfg {
     }
 
     pub fn get_default(&self) -> Result<String> {
-        self.settings_file
-            .with(|s| Ok(s.default_toolchain.clone().unwrap()))
+        self.settings_file.with(|s| {
+            s.default_toolchain
+                .clone()
+                .ok_or_else(|| "no default toolchain configured".into())
+        })
     }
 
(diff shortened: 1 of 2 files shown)


kinnison:
Can you describe the situation in which this happens, and would it be possible to write a test to demonstrate it no longer does?  If the testing is too hard I won't complain too much, though with the introduction of a message, it'd be good to have a test which explicitly trips it.

Glamhoth:
It happens when there are no toolchain configured in .rustup/settings.toml file, i.e. just after rustup was installed. I think it would be possible to test if running `rustup default` returns, in that case, an error.

kinnison:
OK, yes I can now reproduce the issue so I agree it needs fixing.  That's good.

Now we need to devise a test for this.

Perhaps, in `test/cli-misc.rs` you might try adding:

```rust
#[test]
fn no_panic_on_default_missing() {
    setup(&|config| {
        expect_err(
            config,
            &["rustup", "default"],
            "no default toolchain configured",
        );
    })
}
```

And see if that helps.  I can confirm that the above test would panic on current master.

Glamhoth:
Test passes after my fix.

kinnison (approved):
LGTM

PR #1790, opened 2019-04-22
Update download backend stubs to 2018 edition

### LINKED ISSUE #1789: Compilation fails with default-features = false
**Problem**

The rustup crate fails to build if either of the `curl-backend` or `reqwest-backend` features are not selected.

**Steps**
1. Create a new cargo project, and include in the manifest:

```toml
[dependencies.rustup]
version = "1.17"
git = "https://github.com/rust-lang/rustup.rs.git"
default-features = false
```

2. Run `cargo check` or `cargo build`

**Possible Solution(s)**
Update the two fallback modules in [download/src/lib.rs](https://github.com/rust-lang/rustup.rs/blob/master/download/src/lib.rs#L368) to compile under the 2018 edition.

**Notes**

I appreciate that this crate is not intended for public consumption etc. 😁 I'm playing around with building a rust playground, and wanted to use rustup internals to query available toolchains etc. If this is a Bad Idea please let me know!

I will prepare a PR, since I assume this is a simple oversight, and I'm in the code.

rustup version: 373448b032aec0054

cmyr:
- fixes #1789

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -368,7 +368,7 @@ pub mod reqwest_be {
 pub mod curl {
 
     use super::Event;
-    use errors::*;
+    use crate::errors::*;
     use url::Url;
 
     pub fn download(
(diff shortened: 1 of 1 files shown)


kinnison (approved):
These look reasonable, thank you for this.

PR #1791, opened 2019-04-23
Improve release process

emilyalbini:
Added the `local-rustup` directory to gitignore and made the publish script compatible with both Python 2 and Python 3.

r? @kinnison

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 *~
 /**/target
 /home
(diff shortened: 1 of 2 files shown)


kinnison (approved):
This looks good to me, though the python stuff is a little outside of my wheelhouse.

I'll merge after 1.18.0 is done.

PR #1792, opened 2019-04-23
Only build macOS and other Linux targets on stable branch

tesuji:


kinnison (commented):
I've used this "misfeature" before now, to check more platforms, I'm not sure I want this change.  Let me think about it.

tesuji:
In that case, you can remove `if` condition lines.

tesuji:
Ping @kinnison, what do you think again?

kinnison:
OK my thought is as follows:

PRs:

* Linux x86_64
* Linux ARM of some kind
* MacOS X x86_64
* Windows x86_64 MSVC

Master:

* Linux x86_64
* Linux ARM of some kind
* MacOS X x86_64
* Windows x86_64 MSVC
* Windows x86_64 GNU

Other branches within rustup (so stable etc) the full panel, but where `stable` is the only one doing a deploy to dev-static.

This will need appveyor changes as well as travis changes.

What do you think?

kinnison:
And now that I've seen it - all should run the centos check too

kinnison:
I think this is resolved with the recent merge to master of travis changes which includes early Windows support

Issue #1777, opened 2019-04-19
toolchain install: unchanged - (toolchain not installed)


Riateche:
**Problem**
Rustup failed to install a toolchain:
```
info: syncing channel updates for '1.34.0-x86_64-unknown-linux-gnu'
info: latest update on 2019-04-11, rust version 1.34.0 (91856ed52 2019-04-10)
info: downloading component 'rustc'
info: downloading component 'rust-std'
error: component download failed for rust-std-x86_64-unknown-linux-gnu
info: caused by: could not rename downloaded file from '/home/ri/.rustup/downloads/32e097e75f13975e24e45465f2911dd794194f2803e9f6ae96d6f6254596fa45.partial' to '/home/ri/.rustup/downloads/32e097e75f13975e24e45465f2911dd794194f2803e9f6ae96d6f6254596fa45'
info: caused by: No such file or directory (os error 2)
```
After that, it is not possible to retry the installation:
```
$ rustup toolchain install 1.34.0
info: syncing channel updates for '1.34.0-x86_64-unknown-linux-gnu'

  1.34.0-x86_64-unknown-linux-gnu unchanged - (toolchain not installed)

info: checking for self-updates
$ rustup toolchain uninstall 1.34.0
info: no toolchain installed for '1.34.0-x86_64-unknown-linux-gnu'
$ rustup update                                   
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: checking for self-updates
info: downloading self-update

  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.34.0 (91856ed52 2019-04-10)
```
I managed to fix the error by deleting the `~/.rustup/update-hashes/1.34.0-x86_64-unknown-linux-gnu` file manually.

The expected behavior is that `rustup toolchain install 1.34.0` installs the toolchain if it's not installed. 

**Steps**
I don't know how to reproduce the exact installation failure I got, but it seems that you can just do the following:

1. `rustup toolchain install 1.34.0`
2. Remove `~/.rustup/toolchains/1.34.0-x86_64-unknown-linux-gnu`.
3. `rustup toolchain install 1.34.0`

**Notes**

Output of `rustup --version`:
```
rustup 1.17.0 (069c88ed6 2019-03-05)
```
Output of `rustup show`:
```
Default host: x86_64-unknown-linux-gnu

info: syncing channel updates for '1.34.0-x86_64-unknown-linux-gnu'
installed toolchains
--------------------

[...]

PR #1801, opened 2019-04-24
Check for stray hash when installing a toolchain

Glamhoth:
#1777 
As @kinnison explained - problem is in update-hash that is still present after manual removal of a toolchain. This resulted in the inability to install it again. To fix this problem we now check if there is a stray hash, and delete if so. This is WIP as a test must be written.

### DIFF
--- a/tests/cli-v2.rs
+++ b/tests/cli-v2.rs
@@ -4,8 +4,8 @@
 pub mod mock;
 
 use crate::mock::clitools::{
-    self, expect_err, expect_not_stdout_ok, expect_ok, expect_ok_ex, expect_stderr_ok,
-    expect_stdout_ok, set_current_dist_date, this_host_triple, Config, Scenario,
+    self, expect_err, expect_not_stdout_ok, expect_ok, expect_stderr_ok, expect_stdout_ok,
+    set_current_dist_date, this_host_triple, Config, Scenario,
 };
 use std::fs;
 use std::io::Write;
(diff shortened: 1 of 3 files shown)


kinnison:
Hi @Glamhoth 

We do indeed need to try and organise a test now, because without a suitable test we be able to protect against this kind of thing causing a problem elsewhere in the future.

Do you think you're able to work out where to fit one in?

Glamhoth:
Sure, I'll try to write it today.

tests/cli-v2.rs
```diff
@@ -808,6 +809,61 @@ fn remove_target_missing_update_hash() {
...
 
+// Issue #1777
+#[test]
+fn warn_about_and_remove_stray_hash() {
+    setup(&|config| {
+        let mut hash_path = config.rustupdir.join("update-hashes");
+        fs::create_dir_all(&hash_path).expect("Unable to make the update-hashes directory");
+
+        hash_path.push(for_host!("nightly-{}"));
+        println!("Update-hash path: {}", hash_path.display());
```
kinnison: This println seems superfluous debugging.  Probably remove it now you're satisfied?

tests/cli-v2.rs
```diff
@@ -808,6 +809,61 @@ fn remove_target_missing_update_hash() {
...
+
+        let mut file = fs::File::create(&hash_path).expect("Unable to open update-hash file");
+        file.write_all(b"LEGITHASH")
+            .expect("Unable to write update-hash");
+        drop(file);
+
+        expect_ok_ex(
```
kinnison: The use of `expect_ok_ex` means you are very brittle against the exact output of the tool.

Instead I'd look to use `expect_stdout_ok` to check for the warning, and then possibly `expect_ok` to check that `rustc --version` works for the toolchain.  That way you can demonstrate that (a) the warning is emitted and (b) the toolchain got installed, without being brittle against the exact output.

kinnison (changes requested):
Pretty good over-all, thank you.

It'd be nice to make the test a little less brittle though -- see my comments.

Glamhoth:
I cleaned up the test.

kinnison (approved):
LGTM, thank you @Glamhoth

PR #1802, opened 2019-04-24
CI: Enable core dump on Linux

tesuji:
Based on rust-lang/rust#52333 .
Demo run here: https://travis-ci.com/lzutao/rustup.rs/jobs/195307167#L434

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 dist: xenial
 services:
   - docker
-language: generic
+language: minimal
+addons:
+  apt:
+    packages:
+      - gdb
 
 git:
   depth: false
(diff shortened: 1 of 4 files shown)


kinnison:
I am not sure who would be best to review this, to determine if it's correct and sufficient.  @lzutao do you recall who suggested core dumps?

tesuji:
On discord, `@pietroalbini` sugessted using rustc debug options. Maybe `@alexcrichton` or
`@kennytm` could review this PR?

kinnison:
Are any of @pietroalbini @kennytm or @alexcrichton in a position to review this change to determine if it'd provide the requisite information should travis end up segfaulting one of our tests in the future?

tesuji:
Rebased on master.

alexcrichton:
I would personally discourage this in the sense that it's make CI configuration very complicated for probably not a lot of reward, does rustup segfault that often on CI? For testing it there's not really any great way to verify visuall that this works so if it's to be merged I'd recommend triggering a segfault in a test and make sure it generates an appropriate log entry.

tesuji:
@kinnison What do you think about this? Should we keep or close this?

kinnison:
While I saw a couple of segfaults last week, I've not seen any since, so I think this was a spurious tech failure at Travis and not something we should complicate our CI for right now.  Especially since we really shouldn't see any segfaults in rustup code in the general course of PRs.  We'll have this to come back to if we decide we need it in the future. Thank you @lzutao for having a go at this in the meantime.

PR #1804, opened 2019-04-24
rustup-init.sh: Add tests for old OSes

tesuji:
cc @kinnison

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -323,6 +323,10 @@ say() {
     printf 'rustup: %s\n' "$1"
 }
 
+warn() {
+    printf 'warn: %s\n' "$1" >&2
+}
+
 err() {
     say "$1" >&2
     exit 1
@@ -359,7 +363,7 @@ ignore() {
 # This wraps curl or wget. Try curl first, if not installed,
 # use wget instead.
 downloader() {
-    local _dld
+    local _dld curl_secure_flags wget_secure_flags
     if check_cmd curl; then
         _dld=curl
     elif check_cmd wget; then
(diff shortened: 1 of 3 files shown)


rustup-init.sh
```diff
@@ -371,49 +375,72 @@ downloader() {
...
-    _cmd="$1"
-    _ok="y"
-    shift
-
-    # If we're running on OS-X, older than 10.13, then we always
-    # fail to find these options to force fallback
-    if check_cmd sw_vers; then
-        if [ "$(sw_vers -productVersion | cut -d. -f2)" -lt 13 ]; then
-            # Older than 10.13
-            echo "Warning: Detected OS X platform older than 10.13"
```
kinnison, line 405: You appear to have removed this test and not put in the equivalent anywhere else.  This test is necessary because 10.12's curl claims support for the arguments, but fails in a way which is indistinguishable from a MITM attack if they are used.

rustup-init.sh
```diff
@@ -371,49 +375,72 @@ downloader() {
...
-            _ok="n"
+check_curl_options() {
+    local flags tls_protocol
+    for tls in '--tlsv1.2' '--tlsv1.1' '--tlsv1.0' '--tlsv1'; do
```
kinnison, line 400: We do not want to do this.  We only want tlsv1.2 if we're specifying a protocol at all.  At least that's my understanding.

rustup-init.sh
```diff
@@ -371,49 +375,72 @@ downloader() {
...
+        if ! ( curl "$tls" 2>&1 > /dev/null | grep -q unknown ); then
+            tls_protocol="$tls"
+            break
         fi
+    done
+    if [ -z "$tls_protocol" ]; then
+        warn 'cannot use secure protocol to download'
+    elif [ "$tls_protocol" != "--tlsv1.2" ]; then
+        warn "fallback to TLS${tls_protocol#--tls}"
```
kinnison, line 409: I'm not sure there's a need for this fallback, similar reasoning to above.

rustup-init.sh
```diff
@@ -371,49 +375,72 @@ downloader() {
...
+    echo "$flags"
+}
 
-    for _arg in "$@"; do
-        if ! "$_cmd" --help | grep -q -- "$_arg"; then
-            _ok="n"
+check_wget_options() {
+    local flags tls_protocol
+    if ! (wget --secure-protocol 2>&1 > /dev/null | grep -q unrecognized); then
+        for tls in TLSv1_2 TLSv1_1 TLSv1; do
```
kinnison, line 423: Similar issue to above.

kinnison (changes requested):
At minimum the fix for OSX < 10.13 needs restoring

rustup-init.sh
tesuji: I think fallback is better than nothing at all.

rustup-init.sh
tesuji: I don't understand. I'd add new commits to test those macOS images listed
in https://docs.travis-ci.com/user/reference/osx/#macos-version.

kinnison: This is also meant to cope for users, so do not do anything travis-specific.

rustup-init.sh
kinnison: I'd rather let the server-side negotiation protect us at that point.

tesuji:
I will close this PR then.

PR #1806, opened 2019-04-25
Fix broken link in 1.18.1 changelog

jonas-schievink:


kinnison:
This was already fixed in master, thanks for the PR but I think I should just close it.  Do you agree?

jonas-schievink:
Sure, the changelog on master isn't currently linked anywhere either, so this doesn't matter too much.

PR #1808, opened 2019-04-25
Add listing of installed targets

onatm:
This feature adds the ability to list installed `targets` only.

I'm playing with cross-compilation a lot and I need a way to list only the installed `targets`.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -320,6 +320,23 @@ pub fn list_targets(toolchain: &Toolchain<'_>) -> Result<()> {
     Ok(())
 }
 
+pub fn list_installed_targets(toolchain: &Toolchain<'_>) -> Result<()> {
+    let mut t = term2::stdout();
+    for component in toolchain.list_components()? {
+        if component.component.short_name_in_manifest() == "rust-std" {
+            let target = component
+                .component
+                .target
+                .as_ref()
+                .expect("rust-std should have a target");
+            if component.installed {
+                writeln!(t, "{}", target)?;
+            }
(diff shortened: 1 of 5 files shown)


kinnison (changes requested):
Hi!

Thanks for this patch, it looks pretty OK but it needs a test and possibly some README updates to complete the work.

Do you think you can see how to write a test for this?

onatm:
Hey!

Thank you for the review. I believe I need to add tests to `tests/cli-exact.rs`, `tests/cli-v2.rs` and `tests/cli-rustup.rs`.

Something similar to:

https://github.com/rust-lang/rustup.rs/blob/7afe6a8b10b4e5f891e1b5232a7c362b40f5c168/tests/cli-exact.rs#L354-L371

Do you think that I am heading to the right direction?

onatm:
I think I managed to implement all the required tests. I am not sure about README updates. Could you point me the place to make documentation changes @kinnison ?

I've got a question about the number of commits: Should I leave as is or squash them into one?

kinnison (approved):
Hi,

The added tests look excellent, thank you very much!

I'm happy with the commits as they are, or you could squash the three test commits into one.

Either way, I'll look to merge this in soon, unless anyone else objects.

Thanks again,

D.

onatm:
I squashed test commits into one 🎉 

I hope this feature will be useful for others as well. Thanks for the review again.

tesuji (approved):

kinnison (approved):
The squash looks clean

PR #1809, opened 2019-04-26
Fix local bash completion directory path

lispy:
Per https://github.com/scop/bash-completion/blob/master/README.md's
"FAQ", bash looks for local completions in "bash-completion" (a "-"
instead of an "_", or  ~/.local/share/bash-completion/completions).

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ but the gist is as simple as using one of the following:
 
 ```console
 # Bash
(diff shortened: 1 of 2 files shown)


kinnison (approved):
Hi @lispy 

Thank you for this -- it seems like we missed some of the `-` vs `_` in a previous PR.  Well spotted!

tesuji (approved):

PR #1810, opened 2019-04-27
ci: Ensure rustup-init script run on CentOS 6

tesuji:
I tried to run rustup-init script on CentOS 5 docker images, but got
a strage SSL error codes. Maybe preinstalled TLS certificates on it is all
deprecated.


kinnison (approved):
This seems like an excellent idea.

PR #1814, opened 2019-04-27
ci: Run rustfmt on x86_64-linux only

tesuji:
Also removing obsolete `PATH="$HOME/rust/bin:$PATH"`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,21 +57,16 @@ install:
   # add mingw to PATH if necessary
   - if defined MINGW_DIR set PATH=C:\msys64\%MINGW_DIR%\bin;C:\msys64\usr\bin;%PATH%
 
-  # Add rustfmt support
-  - rustup component add rustfmt
-
   # let's see what we got
   - where gcc rustc cargo
   - rustc -vV
   - cargo -vV
-  - rustfmt -vV
 
 build: false
 
 test_script:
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -71,7 +71,6 @@ build: false
 test_script:
   - cargo build --release --target %TARGET% --locked
   - cargo test --release --target %TARGET%
-  - cargo fmt --all -- --check
```
kinnison, line 74: `cargo fmt` can be stymied by conditional compilation and since we have some conditional compilation into Windows, we should retain this just in case.

kinnison (changes requested):
Does this honestly save a lot of CI time?  I'd rather we ran cargo fmt everywhere just in case it catches something that conditional compilation made hard for it to spot unless it's consuming a huge quantity of time.

It takes less than a second to check on my slowest computer for example, so even if it took 5s on each of our platforms, that's hardly a lot of time.

What is your justification for this change?

tesuji:
I would say this doesn't save noticeable build time. The `rustfmt`
on Linux builds are same program (same executable, same architecture).
So I personally don't want to run it multiple times. I agree with you on
keeping rustfmt running on Windows.

kinnison:
Okay, so perhaps one rustfmt on linux, one on macos, and one on windows?

kinnison (approved):

PR #1815, opened 2019-04-27
ci: Use rustc docker image for x86_64 and i686 linux

tesuji:
cc @kinnison , @alexcrichton 
~So the effort of improving CI time is zero for now~ (#1760)
The next PR might be setting up sccache.

### DIFF
--- a/ci/docker/i686-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/i686-unknown-linux-gnu/Dockerfile
@@ -1,9 +1,1 @@
-FROM alexcrichton/rust-slave-dist:2015-10-20b
-USER root
-
-WORKDIR /
-RUN curl https://www.cpan.org/src/5.0/perl-5.28.0.tar.gz | tar xzf -
-WORKDIR /perl-5.28.0
-RUN ./configure.gnu
-RUN make -j$(nproc)
-RUN make install
+FROM rust-i686-unknown-linux-gnu
--- a/ci/docker/x86_64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/x86_64-unknown-linux-gnu/Dockerfile
@@ -1,9 +1,1 @@
-FROM alexcrichton/rust-slave-dist:2015-10-20b
-USER root
-
-WORKDIR /
-RUN curl https://www.cpan.org/src/5.0/perl-5.28.0.tar.gz | tar xzf -
-WORKDIR /perl-5.28.0
-RUN ./configure.gnu
-RUN make -j$(nproc)
-RUN make install
+FROM rust-x86_64-unknown-linux-gnu
--- a/ci/fetch-rust-docker.sh
+++ b/ci/fetch-rust-docker.sh
@@ -22,6 +22,8 @@ case "$TARGET" in
   powerpc64-unknown-linux-gnu)     image=dist-powerpc64-linux ;;
   powerpc64le-unknown-linux-gnu)   image=dist-powerpc64le-linux ;;
   s390x-unknown-linux-gnu)         image=dist-s390x-linux ;;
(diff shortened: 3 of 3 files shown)


ci/docker/i686-unknown-linux-gnu/Dockerfile
```diff
@@ -1,9 +1,2 @@
...
-USER root
-
-WORKDIR /
-RUN curl https://www.cpan.org/src/5.0/perl-5.28.0.tar.gz | tar xzf -
-WORKDIR /perl-5.28.0
-RUN ./configure.gnu
-RUN make -j$(nproc)
-RUN make install
+FROM rust-i686-unknown-linux-gnu
+RUN ln /rustroot/bin/gcc /rustroot/bin/cc
```
kinnison: Is there a nicer, perhaps package-managery way we can have this work?

kinnison (approved):
Over-all I like this a lot.  It feels right to me to be using the same docker image sources everywhere.

@alexcrichton how do you feel about this?  I'm guessing positive?

(tentative approval, but I'd like at least one other docker-conversant person to have final approval)

ci/docker/i686-unknown-linux-gnu/Dockerfile
mati865: There is no package manager involved here.
Setting environment variable `CC=/rustroot/bin/gcc` should work.

tesuji: No, setting `CC` env doesn't seem to work. Maybe `cc` crate doesn't respect
`CC` env?

tesuji:
```console
$ printenv CC CXX
clang
clang++
```

CC is set to clang and we still had the build error.

mati865: Ah, the target is explicitly configured: https://github.com/rust-lang/rustup.rs/blob/66c4d10871edc307865db031483f2bd799abbba3/ci/run.sh#L9
So it'd be `CC_x86_64-unknown-linux-gnu` for x84_64.

I don't know exact error but if it was the linker then `RUSTFLAGS="-C linker=gcc"` would help but this is getting a bit complex.

tesuji: Is `CC_x86_64-unknown-linux-gnu` standard or it's just used by cc crate?

tesuji: The error message is
```rust
     Running `rustc --crate-name semver_parser /src/target/cargo-home/registry/src/github.com-1ecc6299db9ec823/semver-parser-0.7.0/src/lib.rs --color always --crate-type lib --emit=dep-info,link -C opt-level=3 -C codegen-units=1 -C metadata=c7af5d73bcae4fa9 -C extra-filename=-c7af5d73bcae4fa9 --out-dir /src/target/release/deps -L dependency=/src/target/release/deps --cap-lints allow`
error: linker `cc` not found
  |
  = note: No such file or directory (os error 2)
```

mati865: > Is CC_x86_64-unknown-linux-gnu standard or it's just used by cc crate?

As far as I can tell it's used by `cc-rs` and Rust's bootstrap.

> The error message is

That's the linker error so `RUSTFLAGS="-C linker=gcc"` would work.

tesuji: I still think about it: Why does rustc's docker image have no `cc` hard link?
It is common that `cc` hard link to `gcc` on Linux distro.

tesuji:
Blocked by rust-lang/rust#60366

alexcrichton (approved):

tesuji:
Ok. Ready to merge :tada:

PR #1817, opened 2019-04-28
Remove duplicate std::env import

onatm:
This PR removes a duplicate import of `std::env` from `cli/self_update`.

I saw the duplicate import warning while working on #1808 and decided to tackle it in another PR


kinnison (approved):
Seems sane to me, Thank you

PR #1819, opened 2019-04-29
Update to env_proxy 0.3.1 to handle proxies on default ports

### LINKED ISSUE #1811: Tracking: env_proxy loses :80 on http proxies
A user reported on the discord that this was happening, we've tracked it through and filed https://github.com/inejge/env_proxy/issues/6 this issue is here to track that so that we can update our `Cargo.{toml,lock}` when a fix is released.

mgdm:
Resolves #1811.


kinnison (approved):
LGTM

Issue #1818, opened 2019-04-29
Shift away from `term` crate


kinnison:
The `term` crate is unmaintained.  cf. https://github.com/Stebalien/term/issues/93

It is recommended we shift to `termcolor` (https://crates.io/crates/termcolor)

Whatever the move is toward, it ought to be away from an unmaintained and troublesome crate.

kinnison:
Doing this may solve #1800

jens1o:
I'd like to take this. :)

kinnison:
@jens1o Cool, thank you, if you want to discuss things, you should come onto the Rustup discord on https://discord.gg/cKdYFVz and we can talk.

omang777:
I like this.

pickfire:
I believe this is solved?

kinnison:
> I believe this is solved?

@pickfire At this time we're still using `term` because we had issues with other options

Rudi3:
I'm not sure what exactly you need, but what about https://github.com/crossterm-rs/crossterm?

kinnison:
That looks potentially good, though I was thinking more about a crate which integrates progress bars and messages.

Rudi3:
I see, what about https://github.com/mitsuhiko/indicatif then?

kinnison:
Yeah something like Indicatif would be good.  It'll be quite a bit of work to rearrange all the output to go properly via that, but it would potentially allow some cleanup.

aloucks:
What was the issue with `termcolor`? Isn't that what cargo uses?

kinnison:
I don't recall what prevented `termcolor` from being suitable, likely something to do with how we currently (hackily) do progress bars.

PR #1820, opened 2019-04-29
Bug 1800: only open terminfo once

rbtcollins:
I did a slightly more than minimal fix here, promoting the
term::Terminal trait to be the contract that our markdown formatting
is written to, making the auto-disabling of formatting on non-tty's
and handling of non-colour terminals be clearly separated out.

The handling of corrupt or missing terminfo configuration is also
pushed down to the underlying terminal trait object rather than
having a bypass in our higher level code; this makes the upper layers
easier to reason about.

There is a modest increase in size, as the full term::Terminal trait
is now implemented for the translation layer; a decorator macro could
probably be writtent to shrink that if needed.

There is a remaining direct use of term in download_tracker that causes
a second open of the terminfo database, which should also be fixed but
it is constant in nature so a later pass can do it.

I don't have any tests for this; doing a unit test would be
uninteresting in this case - rust's types give use what we need at that
level, and for the conceptual level 'nothing else in the system calls
any term calls', unit tests are inappropriate. A good way to test this
would be to have the functional tests instrumented and make assertions
about the behaviour of those executions; then any future regressions
(e.g. if switching to a different term library) would be detected.

### DIFF
--- a/src/cli/main.rs
+++ b/src/cli/main.rs
@@ -22,6 +22,7 @@ mod download_tracker;
 mod errors;
 mod help;
 mod job;
+mod markdown;
 mod proxy_mode;
 mod rustup_mode;
 mod self_update;
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -3,6 +3,7 @@ use crate::errors::*;
 use crate::help::*;
 use crate::self_update;
 use crate::term2;
+use crate::term2::Terminal;
 use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, Shell, SubCommand};
 use rustup::dist::dist::{PartialTargetTriple, PartialToolchainDesc, TargetTriple};
 use rustup::dist::manifest::Component;
@@ -807,7 +808,7 @@ fn show(cfg: &Cfg) -> Result<()> {
         }
     }
 
-    fn print_header(t: &mut term2::Terminal<std::io::Stdout>, s: &str) -> Result<()> {
+    fn print_header(t: &mut term::StdoutTerminal, s: &str) -> Result<()> {
         t.attr(term2::Attr::Bold)?;
         writeln!(t, "{}", s)?;
         writeln!(t, "{}", iter::repeat("-").take(s.len()).collect::<String>())?;
--- a/src/cli/term2.rs
+++ b/src/cli/term2.rs
@@ -2,27 +2,102 @@
 //! that does not fail if `StdoutTerminal` etc can't be constructed, which happens
 //! if TERM isn't defined.
 
-use markdown::tokenize;
-use markdown::{Block, ListItem, Span};
+use lazy_static::lazy_static;
 use rustup::utils::tty;
 use std::io;
+use std::sync::Mutex;
 
 pub use term::color;
 pub use term::Attr;
+pub use term::Terminal;
 
-pub trait Instantiable {
-    fn instance() -> Self;
-}
+mod termhack {
+    // Things we should submit to term as improvements: here temporarily.
+    use std::collections::HashMap;
+    use std::io;
+    use term::terminfo::TermInfo;
+    #[cfg(windows)]
+    use term::WinConsole;
+    use term::{StderrTerminal, StdoutTerminal, Terminal, TerminfoTerminal};
 
-impl Instantiable for io::Stdout {
-    fn instance() -> Self {
-        io::stdout()
+    // Works around stdio instances being unclonable.
+    pub trait Instantiable {
+        fn instance() -> Self;
     }
-}
 
-impl Instantiable for io::Stderr {
-    fn instance() -> Self {
-        io::stderr()
+    impl Instantiable for io::Stdout {
+        fn instance() -> Self {
+            io::stdout()
+        }
+    }
+
+    impl Instantiable for io::Stderr {
+        fn instance() -> Self {
+            io::stderr()
+        }
+    }
+
+    /// Return a Terminal object for T on this platform.
+    /// If there is no terminfo and the platform requires terminfo, then None is returned.
+    fn make_terminal<T>(terminfo: Option<TermInfo>) -> Option<Box<Terminal<Output = T> + Send>>
+    where
+        T: 'static + io::Write + Send + Instantiable,
+    {
+        let mut result = terminfo
+            .map(move |ti| TerminfoTerminal::new_with_terminfo(T::instance(), ti.clone()))
+            .map(|t| Box::new(t) as Box<Terminal<Output = T> + Send>);
+        #[cfg(windows)]
+        {
+            result = result.or_else(|| {
+                WinConsole::new(T::instance())
+                    .ok()
+                    .map(|t| Box::new(t) as Box<Terminal<Output = T> + Send>)
+            })
+        }
+        result
+    }
+
+    fn make_terminal_with_fallback<T>(
+        terminfo: Option<TermInfo>,
+    ) -> Box<Terminal<Output = T> + Send>
+    where
+        T: 'static + io::Write + Send + Instantiable,
+    {
+        make_terminal(terminfo)
+            .or_else(|| {
+                let ti = TermInfo {
+                    names: vec![],
+                    bools: HashMap::new(),
+                    numbers: HashMap::new(),
+                    strings: HashMap::new(),
+                };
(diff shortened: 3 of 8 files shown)


rbtcollins:
I'm not sure why the tests are failing; they are passing for me :/.

kinnison:
All the failures appear to be in `cli-exact` which is ultra-sensitive to changes in output format.  it's possible there are newline or whitespace changes which are affected by your change, but I'll have to look harder to find out.  I've kicked off a test run on my laptop to see if I can spot any difference.

kinnison:
I can confirm i see the same errors on my laptop so I'll try to diagnose later today.

kinnison:
Looks like there're formatting codes present in the output even under test...

I changed print_indented in clitools.rs to debug-print things and:

```
out.stdout (3 lines):
    "\n      nightly-x86_64-unknown-linux-gnu unchanged\u{1b}(B\u{1b}[m - 1.3.0 (hash-n-2)\n    \n    "
out.stderr (1 lines):
    "info: \u{1b}(B\u{1b}[msyncing channel updates for \'nightly-x86_64-unknown-linux-gnu\'\n    "
expected.stdout (3 lines):
    "\n      nightly-x86_64-unknown-linux-gnu unchanged - 1.3.0 (hash-n-2)\n    \n    "
expected.stderr (1 lines):
    "info: syncing channel updates for \'nightly-x86_64-unknown-linux-gnu\'\n    "
```

rbtcollins:
Thanks; that will be a bug in the isatty detection glue then I guess - as the tests are passing for me under vscode on windows. Will look in ~48 hours.

src/cli/term2.rs
```diff
@@ -2,250 +2,131 @@
...
-            for _ in 0..self.indent {
-                let _ = write!(self.w, " ");
-            }
-            self.pos = self.indent;
-        }
+    /// Return a Terminal wrapping stdout, or None if a terminal couldn't be
+    /// opened.
+    #[allow(unused)]
+    pub fn stdout(terminfo: Option<TermInfo>) -> Option<Box<StdoutTerminal>> {
+        make_terminal(terminfo)
```
kinnison, line 83: Given this is unused, is it worth keeping in?

src/cli/term2.rs
```diff
@@ -2,250 +2,131 @@
...
-        }
 
-        // Write the word
-        let _ = write!(self.w, "{}", word);
-        self.pos += word_len;
+    /// Return a Terminal wrapping stderr, or None if a terminal couldn't be
+    /// opened.
+    #[allow(unused)]
+    pub fn stderr(terminfo: Option<TermInfo>) -> Option<Box<StderrTerminal>> {
+        make_terminal(terminfo)
```
kinnison, line 90: Ditto query about unused?

src/cli/term2.rs
```diff
@@ -2,250 +2,131 @@
...
 
 pub use term::color;
 pub use term::Attr;
+pub use term::Terminal;
 
-pub trait Instantiable {
-    fn instance() -> Self;
-}
+mod termhack {
+    // Things we should submit to term as improvements: here termporarily.
```
kinnison: s/termporarily/temporarily/ -- seems sad to leave this typo when you corrected the one about isatty :D

src/cli/term2.rs
```diff
@@ -256,56 +137,122 @@ macro_rules! swallow_unsupported {
...
+
+        self.0.cursor_up()
+    }
+
+    fn delete_line(&mut self) -> term::Result<()> {
+        self.0.delete_line()
+    }
+
+    fn carriage_return(&mut self) -> term::Result<()> {
+        // This might leak control chars in !isatty? needs checking.
```
kinnison, line 209: This *should* in reality only end up with `\r` at a guess.

src/cli/main.rs
```diff
@@ -22,6 +22,7 @@ mod download_tracker;
 mod errors;
 mod help;
 mod job;
+mod markdown;
```
kinnison, line 25: Just for cleanliness of commits, would it make sense to split this module out as one commit, and then do the caching functionality as a second commit?  If you think it's too much of a pain then I won't block on this, but I think I'd prefer if the entirety of `markdown.rs` was not blamed on caching terminfo :D

src/cli/rustup_mode.rs
```diff
@@ -807,7 +808,7 @@ fn show(cfg: &Cfg) -> Result<()> {
         }
     }
 
-    fn print_header(t: &mut term2::Terminal<std::io::Stdout>, s: &str) -> Result<()> {
+    fn print_header(t: &mut term::StdoutTerminal, s: &str) -> Result<()> {
```
kinnison, line 811: Should this not be `term2::StdoutTerminal` to reduce leakage of `term` api?

kinnison (changes requested):
Mostly LGTM.  A few comments / tweaks, and then I think we could merge.  I like the caching, I like the split out of the markdown stuff (particularly since we've had some discussion about moving to a different markdown crate in the future too)

src/cli/term2.rs
rbtcollins: Ehh. If we were keeping term for sure; and actually as its a closer fit to the termcolor api I think it shows the sort of change that may be needed again here, so yes.

src/cli/term2.rs
rbtcollins: ta

src/cli/term2.rs
rbtcollins: I think its ok; I think the comment is probably more helpful than someone who *does* find a problem going 'wtf' ;)

src/cli/main.rs
rbtcollins: Its a little awkward because the types were comingled; so we can't not change the md stuff. TBH I think having it in the one commit isn't going to make bisection a problem - and thats all I'd really worry about history wise.

src/cli/rustup_mode.rs
rbtcollins: I don't think so; thats a concrete type with the isatty and error silencing built in (which BTW is really showing where we abuse the API and call e.g. reset() without checking supports_reset() .... but thats a different story); anyhow - point is that print_header doesn't care about that distinction, and it was giving it the concrete type in the past that was encouraging the markdown comingling in the first place. So I want to make it clearer:

this depends on 'a terminal'

this implements our terminal specialisations.

And I think depending on term2:StdoutTerminal gets that wrong because of the current type alias.

I did have a patch where I ditched that entirely and term2:StdoutTerminal was an export of term:StdoutTerminal, and stdio() was returning a Box, but I decided that that didn't add much.

src/cli/term2.rs
kinnison: Accepted

src/cli/term2.rs
kinnison: Accepted.

src/cli/main.rs
kinnison: Okay, yeah, looking through again I agree it'd be a pain.

src/cli/rustup_mode.rs
kinnison: Aha, right, could we have an `impl` in there then (i.e. `&mut impl term::StdoutTerminal`) just to make it clear that this is a trait not a struct we're getting a borrow to?

rbtcollins: mm, uh,so actually it is a struct we're passing in, not a trait object, so I dunno. Can we just declare it bankrupt as we're about to rip out term anyway?

kinnison: Oh, argh, yeah OK

kinnison (approved):

rbtcollins:
What needs to happen to merge this?

jens1o:
I'd be also interested in this, so I can really start working on #1818 without having merge conflicts in the end :/

rbtcollins:
I have a followup that will likely conflict with 1818 as well; - the reason download_tracker didn't use term2 was that term2 was in the cli package, but the download_tracker was in the root, and I presume we don't want the root depending on the CLI.

rbtcollins:
(that branch is https://github.com/rbtcollins/rustup.rs/pull/new/tracker-term-use ) - I'll set up the PR once this merges.

kinnison:
Sorry for the delay :D

PR #1822, opened 2019-05-01
Various version statement fixes

kinnison:
This PR rolls up an addition of a hidden flag/command to dump the version testament which can be used if in the future a user receives an unexpectedly unclean version of rustup-init.

We use that in the CI pipeline to dump the version which would be displayed so that we can check it during our stable release process.

Also this includes a commit to remove a dirtying step (or rather to clean up after it) in our appveyor builds.


tesuji (approved):

kinnison:
Thanks for the review @lzutao

PR #1823, opened 2019-05-01
Suggestion for misspelled component name

### LINKED ISSUE #1816: Make suggestions based on Levenshtein distance

From time to time people have difficulty installing components because of typos.  Sometimes they're even really hard to spot, particularly for those who are dyslexic.  For example, recently someone had difficulty installing the `rls` component because they kept typing `rsl` but reading it as `rls` and so couldn't tell what was wrong.

Whenever we issue a component-not-found or target-not-found etc, consider adding a quick check of the available components/targets, sort them by levenshtein distance, and suggest the nearest one.

**Notes**

That way, instead of:

```console
$ rustup component add rsl
error: toolchain 'stable-x86_64-unknown-linux-gnu' does not contain component 'rsl' for target 'x86_64-unknown-linux-gnu'
```

We'd maybe get something like:
```console
$ rustup component add rsl
error: toolchain 'stable-x86_64-unknown-linux-gnu' does not contain component 'rsl' for target 'x86_64-unknown-linux-gnu' - Did you perhaps mean 'rls` ?
```

hrvolapeter:
Add suggestion for misspelled component name using Levenshtein distance

Closes #1816

r? @kinnison


src/toolchain.rs
```diff
@@ -674,6 +704,7 @@ impl<'a> Toolchain<'a> {
                     return Err(ErrorKind::UnknownComponent(
                         self.name.to_string(),
                         component.description(&manifest),
+                        self.get_component_suggestion(&component),
```
kinnison: I think it'd be better if this only suggested from the *installed* components, otherwise you might get confused by it suggesting "did you mean X" where X isn't installed.

src/toolchain.rs
```diff
@@ -629,6 +630,35 @@ impl<'a> Toolchain<'a> {
         }
     }
 
+    fn get_component_suggestion(&self, component: &Component) -> Option<String> {
+        use levenshtein::levenshtein;
+        const MAX_DISTANCE: usize = 3;
```
kinnison, line 637: A comment explaining why you chose `3` would be useful, so that it's not just a magical constant.

src/errors.rs
```diff
@@ -325,6 +325,14 @@ error_chain! {
     }
 }
 
+fn component_unknow_suggestion(s: &Option<String>) -> String {
```
kinnison: Typo in the function name here -- `_unknow_` should be `_unknown_`

kinnison (changes requested):
Mostly pretty good, just a couple of requests.

I also would prefer that you don't put `r? @kinnison` in the commit message since that ends up in the repo long-term.  By all means tag me in the PR message though.

Would you like to extend this to also support making a suggestion if the user tries to add/remove a target which is misspelled?

kinnison:
The changes in this PR were subsumed into #1824 and merged.  Thank you for your contribution to Rustup.

PR #1824, opened 2019-05-01
Make suggestions based on Damerau-Levenshtein distance

Glamhoth:
#1816
The only change that I'd suggest is to use Damerau-Levenshtein distance instead of Levenshtein distance, because the former adds transpositions to list of the basic operations. And as we know a lot of misspellings are made that way, even the example from issue is of this kind of error.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -303,9 +303,13 @@ error_chain! {
             description("toolchain does not support components")
             display("toolchain '{}' does not support components", t)
         }
-        UnknownComponent(t: String, c: String, s: String) {
+        UnknownComponent(t: String, c: String, s: Option<String>) {
             description("toolchain does not contain component")
-            display("toolchain '{}' does not contain component {}; did you mean '{}'?", t, c, s)
+            display("toolchain '{}' does not contain component {}{}", t, c, if let Some(suggestion) = s {
+                format!("; did you mean '{}'?", suggestion)
+            } else {
+                "".to_string()
+            })
         }
         AddingRequiredComponent(t: String, c: String) {
             description("required component cannot be added")
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -303,9 +303,9 @@ error_chain! {
             description("toolchain does not support components")
             display("toolchain '{}' does not support components", t)
         }
-        UnknownComponent(t: String, c: String) {
+        UnknownComponent(t: String, c: String, s: String) {
             description("toolchain does not contain component")
-            display("toolchain '{}' does not contain component {}", t, c)
+            display("toolchain '{}' does not contain component {}; did you mean '{}'?", t, c, s)
         }
         AddingRequiredComponent(t: String, c: String) {
             description("required component cannot be added")
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -11,7 +11,6 @@ use crate::install::{self, InstallMethod};
 use crate::notifications::*;
 use crate::utils::utils;
 
-use std::collections::HashMap;
 use std::env;
 use std::env::consts::EXE_SUFFIX;
 use std::ffi::OsStr;
@@ -19,7 +18,6 @@ use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-use strsim::damerau_levenshtein;
 use url::Url;
 
 /// A fully resolved reference to a toolchain which may or may not exist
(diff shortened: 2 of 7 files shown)


src/toolchain.rs
```diff
@@ -556,6 +557,90 @@ impl<'a> Toolchain<'a> {
         }
     }
 
+    fn get_component_suggestion(
+        &self,
+        component: &Component,
+        manifest: &Manifest,
+        only_instaled: bool,
```
kinnison: Typo here, `only_installed` rather than `only_instaled`

src/toolchain.rs
```diff
@@ -556,6 +557,90 @@ impl<'a> Toolchain<'a> {
...
+                {
+                    closest_match = short_name_distance.1.component.target().to_string();
+                }
+            }
+
+            // If suggestion is too different don't suggest anything
+            if closest_distance.0 > MAX_DISTANCE {
+                return String::new();
+            } else {
+                return format!("; did you mean '{}'", closest_match.to_string());
```
kinnison: The presence of the `"; did you mean '{}'"` here is a pain because it moves some of the message formatting away from the `Display` impl of the notification.  Better to have the impl take an Option<String> and do the right thing there IMO.

src/errors.rs
```diff
@@ -303,9 +303,9 @@ error_chain! {
             description("toolchain does not support components")
             display("toolchain '{}' does not support components", t)
         }
-        UnknownComponent(t: String, c: String) {
+        UnknownComponent(t: String, c: String, s: String) {
             description("toolchain does not contain component")
-            display("toolchain '{}' does not contain component {}", t, c)
+            display("toolchain '{}' does not contain component {}{}?", t, c, s)
```
kinnison: If the `s` parameter is `""` then there'll be a spurious `?` on the string.  Better for `s` to be an `Option<String>` and format accordingly.

src/toolchain.rs
```diff
@@ -556,6 +557,90 @@ impl<'a> Toolchain<'a> {
         }
     }
 
+    fn get_component_suggestion(
+        &self,
+        component: &Component,
+        manifest: &Manifest,
+        only_instaled: bool,
+    ) -> String {
```
kinnison: This ought to return `Option<String>` in case there isn't a suggestion to be made.

kinnison (changes requested):
Not bad, some comments in-line, and the commit history will need a good rebase toward the end (be careful to retain retep007's authorship at that point).

In addition, I'd like to see at least one test which demonstrates what happens if there's nothing worth suggesting within the heuristic distance of 3.  E.g. `rustup component add banana`.

tests/cli-v2.rs
```diff
@@ -981,13 +991,28 @@ fn add_target_suggest_best_match() {
...
         expect_ok(config, &["rustup", "default", "nightly"]);
+        expect_not_stderr_ok(
+            config,
+            &[
+                "rustup",
+                "target",
+                "remove",
+                &format!("{}a", clitools::CROSS_ARCH1)[..],
+            ],
+            &format!("did you mean '{}'", clitools::CROSS_ARCH1),
```
kinnison, line 1014: Why is this particular `expect_not_stderr_ok()` check including the suggestion, rather than just `"did you mean"` ?

tests/mock/clitools.rs
```diff
@@ -207,6 +207,16 @@ pub fn expect_not_stdout_ok(config: &Config, args: &[&str], expected: &str) {
     }
 }
 
+pub fn expect_not_stderr_ok(config: &Config, args: &[&str], expected: &str) {
```
kinnison, line 210: I'd prefer if this function were in a separate commit with a message explaining what the function is and why it's there.

i.e. something like:

```
clitools.rs: Add expect_not_stderr_ok

In order to support testing commands which may return certain
messages on `stderr` in a way in which they should *NOT*
produce such messages, add a counterpart to `expect_not_stdout_ok`.
```

kinnison (changes requested):
So very almost there.  Please fold the requested changes into the two main commits you have (i.e. where your initial implementation, and retep's changes lie.

I'll expect to see:

* commit 1: the clitools.rs update
* commit 2: your code (with fixes folded in)
* commit 3: retep's work (with fixes folded in)

If you can do that, I think this is mergeable, so remember to remove `WIP:` from the title when you think you're ready for the final review before merge :grin: 

Thanks,

D.

tests/cli-v2.rs
Glamhoth: By this I wanted to ensure, that it won't return exactly that suggestion, because some other valid suggestion can still apply. For example we have targets 'armv7r-none-eabi' and 'armv7r-none-eabihf'. If you have both installed `target remove armv7r-none-eabih` will suggest '...-eabi'. But if you have only '...-eabihf' installed it shouldn't suggest '...-eabi', but still suggest something - that is '...-eabihf'. I also did this in component_remove test.

kinnison: Aha that sounds good, I just didn't spot that the first time, thank you for explaining.

PR #1825, opened 2019-05-02
ci: Set up cache build with sccache

tesuji:
~Blocked on #1815.~

Now the first build time is a little slower, but incremental builds
in the same branch should be faster.

r? @kinnison

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,13 @@ git:
   depth: false
   quiet: true
 
+cache:
+  directories:
+    - "$HOME/.cargo"
+    - "$HOME/.cache/sccache"
+before_cache:
+  - rm -rf "$HOME/.cargo/registry"
+
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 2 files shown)


kinnison:
When you say it's a little slower on first build of a branch -- how much slower are we talking?
How many builds does it take to get that slowness back?  Is it on the second build, the third?

tesuji:
Look at build time on Travis, this PR may take about 30 seconds more than
normal builds. The second build and later should be faster ([upto 4 minutes][1]).

[1]: https://travis-ci.com/rust-lang/rustup.rs/builds/110499868

kinnison:
Does the rustc docker contain sccache then?  I couldn't see how it turned up otherwise.

tesuji:
Yes. In [this line][1] from build log, sccache is called.

[1]: https://travis-ci.com/rust-lang/rustup.rs/jobs/197478512#L490

PR #1827, opened 2019-05-02
Switch issue templates to use subheadings instead of bold text

LPGhatguy:
Minor change. Using Markdown sub-headings instead of just bold text helps sections stand out a bit more by default. I left the rest of the templates untouched.


tesuji:
I think rustup developers is happier with current templates. Similar PR
has been closed:  #1693 .

LPGhatguy:
I guess I didn't dig deeply enough, oops!

When I filed a bug, I had problems with my lines merging together with the bold text. It's not a big deal.

PR #1828, opened 2019-05-02
Add a script to automate most of the release process

emilyalbini:
The current way of making rustup releases is a bit annoying, because you have to call the `ci/sync-dist.py` script six times in total, three to deploy to dev and three to deploy to prod. With this script the release process becomes:

1. `ci/deploy.sh dev 1.0.0`
2. Test that everything on dev-static works
3. `ci/deploy.sh prod 1.0.0`

r? @kinnison 
cc @alexcrichton @Mark-Simulacrum

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+# This script is used to do rustup releases. It requires an AWS access token
+# allowed to update our S3 buckets and to invalidate CloudFront distributions.
+#
+# Usage:
+#
+# 1. Deploy the release on the dev environment:
+#    ./deploy.sh dev VERSION_NUMBER
+#
+# 2. Test everything works correctly:
+#    RUSTUP_UPDATE_ROOT=https://dev-static.rust-lang.org/rustup rustup self update
+#
+# 3. Deploy the release to the prod environment:
+#    ./deploy.sh prod VERSION_NUMBER
+
+set -euo pipefail
+IFS=$'\n\t'
+
+CI_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd )"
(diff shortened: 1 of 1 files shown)


tesuji:
You have some shellcheck warnings to solve.
```bash
In ci/deploy.sh line 17:
IFS=$'\n\t'
    ^-----^ SC2169: In dash, $'..' is not supported.
In ci/deploy.sh line 19:
CI_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd )"
                         ^---------------^ SC2169: In dash, array references are not supported.
In ci/deploy.sh line 28:
    python3 "${CI_DIR}/sync-dist.py" $@ --live-run
                                     ^-- SC2068: Double quote array expansions to avoid re-splitting elements.
In ci/deploy.sh line 31:
if [[ $# -ne 2 ]]; then
   ^------------^ SC2169: In dash, [[ ]] is not supported.
In ci/deploy.sh line 40:
        if [[ -e "${LOCAL_RUSTUP_DIR}" ]]; then
           ^----------------------------^ SC2169: In dash, [[ ]] is not supported.
In ci/deploy.sh line 41:
            read -p "The directory ${LOCAL_RUSTUP_DIR} will be removed. Continue (y/n)?" choice
            ^--^ SC2162: read without -r will mangle backslashes.
For more information:
  https://www.shellcheck.net/wiki/SC2068 -- Double quote array expansions to ...
  https://www.shellcheck.net/wiki/SC2169 -- In dash, $'..' is not supported.
  https://www.shellcheck.net/wiki/SC2162 -- read without -r will mangle backs...
```

emilyalbini:
Fixed most of the shellcheck lints. I don't think we should support dash though, this is going to be run interactively and `#!/bin/bash` is used.

Issue #1125, opened 2017-05-17
`rustup doc` appears to do nothing


Diggsey:
> I do see behavior on a headless linux machine where rustup doc appears to do nothing.

https://github.com/rust-lang-nursery/rustup.rs/pull/1046#issuecomment-299998690

nodakai:
```
$ docker run -it --name rustup-doc --security-opt seccomp:unconfined debian:jessie
# apt-get update
# apt-get -y install curl less strace
# curl https://sh.rustup.rs -sSf | sh
# PATH=$PATH:$HOME/.cargo/bin
# strace -f -s100 rustup doc |& less
```

In this case rustup was satisfied with just successfully launching `sensible-browser` when it actually complained as below and returned 1:
```
Couldn't find a suitable web browser!
Set the BROWSER environment variable to your desired browser.
```

* https://github.com/rust-lang-nursery/rustup.rs/blob/master/src/rustup-utils/src/raw.rs#L364

ghost:
```bash
$ time rustup doc

real	0m0.308s
user	0m0.290s
sys	0m0.017s
```

Exit code is 0.

According to strace, it executed
`/usr/bin/xdg-open /home/xftroxgpx/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/index.html`
The above(xdg-open command) shows `elinks` (a browser running in console) when I run it myself from `xfce4-terminal` and it seems to work ok. But with `rustup` it shows nothing.

Maybe `rustup doc` could always say the command it executes, so that if it fails, user can copy/paste and manually execute it?

**EDIT:** using [this](https://github.com/xftroxgpx/a3/blob/befd9c4ff9727f5e4369fc5390ece8da3e39723e/system/Z575/OSes/3archlinux/on_baremetal/filesystem_now/archlinux/home/xftroxgpx/build/2nonpkgs/rust.stuff/rustup.rs/fix_rustup_doc.patch) patch for my own needs.

Natim:
In my case it tried to open ``gvfs-open ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html/index.html`` however the html folder doesn't exists here.

Natim:
```
~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc$ tree
.
├── cargo
│   ├── LICENSE-APACHE
│   ├── LICENSE-MIT
[...]

PR #1830, opened 2019-05-02
Migrate browser opening logic to match Cargo

LPGhatguy:
May impact #1125.

Cargo uses the [opener](https://crates.io/crates/opener) crate for opening the user's browser. I noticed some interest in deduplicating and unifying the logic used between Rustup and Cargo in a couple places, and I think this is a reasonable part of doing that.

This PR removes about 80 lines of platform-specific code with a dependency on opener, which has no dependencies of its own. It indirectly removes an outstanding FIXME in the codebase.


kinnison:
Hi,

Thank you for your contribution.  Unfortunately the CI does not pass because your code is not formatted cleanly.

Could you please rebase your change to fold the crate version commit into your main commit, and also ensure that `cargo fmt` has been run so that the formatting is clean.

Once you've done that and pushed your branch once more, I shall review it again.

Fundamentally the change looks good, but I cannot approve it until you have done this.

Thanks again,

D.

LPGhatguy:
I formatted my change with rustfmt and went ahead and rebased + squished the commits.

LPGhatguy:
Looks like checks are green now (I had to open/close to rerun them because of a failure installing Rust on the CI machine initially).

kinnison (approved):
LGTM, and this should be easy to back out if it proves not to work, though given `cargo` uses this crate it seems likely to be fine.

Issue #1831, opened 2019-05-03
rustup update: seconds to complete output improvements


konstankinollc:
**Problem**
![Screen Shot 2019-05-03 at 3 20 01 PM](https://user-images.githubusercontent.com/38840828/57169645-d2d58a80-6dbc-11e9-9582-2c4995cdfdb5.png)

**Solution**
https://github.com/rust-lang/rustup.rs/pull/1832

**Notes**

Output of `rustup --version`: rustup 1.18.2 (a0bf3c9cb 2019-05-02)
Output of `rustup show`: 
x86_64-apple-darwin
stable-x86_64-apple-darwin (default)
rustc 1.34.1 (fc50f328b 2019-04-24)

konstankinollc:
Resolved.

PR #1832, opened 2019-05-03
Fix displaying ETA output for download tracker, make it consistent & treat output as integers.

konstankinollc:
Addresses #1831

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,6 @@ test = false # no unit tests
 [[bin]]
 name = "rustup-init"
 path = "src/cli/main.rs"
-test = false # no unit tests
 
 [profile.release]
 lto = true
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -118,6 +118,15 @@ impl DownloadTracker {
         }
         self.prepare_for_new_download();
     }
+    // we're doing modular arithmetic, treat as integer
+    pub fn from_seconds(sec: u32) -> (u32, u32, u32, u32) {
+        let d = sec / (24 * 3600);
+        let h = sec % (24 * 3600) / 3600;
+        let min = sec % 3600 / 60;
+        let sec = sec % 60;
+
+        (d, h, min, sec)
+    }
     /// Resets the state to be ready for a new download.
     fn prepare_for_new_download(&mut self) {
         self.content_len = None;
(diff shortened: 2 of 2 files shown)


kinnison:
I thought we'd already dealt with this, sigh.  What types are involved here because really all of the days/hours/minutes/seconds should be integers.

kinnison:
@rbtcollins Could you take a look and ensure that your PRs would fix this too?  Assuming so, let me know and we'll close this in favour of the more deep correction.  If not, then if I merge this, could you update your PRs to match?

rbtcollins:
My PR didn't fix this sorry, I was fixing an orthogonal problem - this change is still needed; @konstankinollc  could you please rebase your patch then @kinnison  can merge it? Thanks!

konstankinollc:
Absolutely, will do in a bit.

konstankinollc:
@kinnison ready for your review

konstankinollc:
It also addresses a bug in computing hours to download.

konstankinollc:
@kinnison can you please point me why this build fails?

konstankinollc:
Can anyone please explain why the build is failing on CI?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rustup"
-version = "1.18.2"
+version = "1.18.3"
```
kinnison: Please don't edit our version number, this is only changed during a release

src/cli/download_tracker.rs
```diff
@@ -233,3 +225,42 @@ impl fmt::Display for Size {
...
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn download_tracker_from_seconds_test() {
+        use crate::download_tracker::DownloadTracker;
+        assert_eq!(
+            DownloadTracker::from_seconds(2),
+            (0, 0, 0, 2)
+        );
```
kinnison: The reason the CI test fails is that these are not formatted in the manner rustfmt prefers.  You can run `cargo fmt` to correct things prior to committing your code.

kinnison (changes requested):
As noted, please don't update the package version in `Cargo.toml` and I've explained the CI failure for you

src/cli/download_tracker.rs
konstankinollc: Gotcha. Much appreciated for your quick response. Will make adjustments now. Thanks!

Cargo.lock
```diff
@@ -1228,7 +1228,7 @@ dependencies = [
 
 [[package]]
 name = "rustup"
-version = "1.18.2"
+version = "1.18.3"
```
kinnison: You've left this Cargo.lock change in place, so CI won't run.

Cargo.toml
```diff
@@ -94,7 +94,7 @@ test = false # no unit tests
 [[bin]]
 name = "rustup-init"
 path = "src/cli/main.rs"
-test = false # no unit tests
+test = true
```
kinnison: One final question, does this need to be set to `true` or is it sufficient to remove the line?

konstankinollc: Good point. Let me try to remove the line.

konstankinollc: Here is a suggestion. To be explicit that we welcome unit tests, and potentially it may ~help~ nudge others to write more unit tests, maybe we should keep it?

kinnison: Most of our tests are integration tests because unit tests have a tendency to increase rigidity in APIs, so while I don't want to reject tests, I also don't feel the need to massively encourage them at the unit-test level.

konstankinollc: Sounds good

konstankinollc: Squashed into one commit and removed a line.

kinnison (approved):
LGTM now, if it passes CI then it can merge IMO

PR #1834, opened 2019-05-05
Use term2 from download_tracker

rbtcollins:


kinnison (approved):
This looks okay to me.

PR #1836, opened 2019-05-07
Simplify HumanReadable

rbtcollins:
HumanReadable was unnecessarily complex; byte sized objects are not
fractional, and we have no use case for switching between size and
duration rendering on the same value.

Size is an integral human readable display type for less than GB
size objects, and Duration is a human readable up to hours of duration
display type.

Fixing this also uncovered the size mismatch between content length
and total downloaded which was previously masked by all the f64
conversions. This would have led to a runtime error after
downloading over 2G; switching content length storage to usize to
match the total download tracking will avoid this (for statically sized)
objects by failing when the content length header is received.

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -15,7 +15,7 @@ const DOWNLOAD_TRACK_COUNT: usize = 5;
 /// Tracks download progress and displays information about it to a terminal.
 pub struct DownloadTracker {
     /// Content-Length of the to-be downloaded object.
-    content_len: Option<u64>,
+    content_len: Option<usize>,
     /// Total data downloaded in bytes.
     total_downloaded: usize,
     /// Data downloaded this second.
@@ -81,8 +81,9 @@ impl DownloadTracker {
 
     /// Notifies self that Content-Length information has been received.
     pub fn content_length_received(&mut self, content_len: u64) {
-        self.content_len = Some(content_len);
+        self.content_len = Some(content_len as usize);
     }
+
     /// Notifies self that data of size `len` has been received.
     pub fn data_received(&mut self, len: usize) {
         self.total_downloaded += len;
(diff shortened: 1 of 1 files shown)


src/cli/download_tracker.rs
```diff
@@ -12,7 +12,7 @@ const DOWNLOAD_TRACK_COUNT: usize = 5;
 /// Tracks download progress and displays information about it to a terminal.
 pub struct DownloadTracker {
     /// Content-Length of the to-be downloaded object.
-    content_len: Option<u64>,
+    content_len: Option<usize>,
```
kinnison, line 18: Why reducing to `usize` here?

src/cli/download_tracker.rs
```diff
@@ -75,8 +75,9 @@ impl DownloadTracker {
 
     /// Notifies self that Content-Length information has been received.
     pub fn content_length_received(&mut self, content_len: u64) {
-        self.content_len = Some(content_len);
+        self.content_len = Some(content_len as usize);
```
kinnison, line 84: Is the cast here preferable to remaining `u64`?

kinnison (commented):
I'd like to know why the switch to `usize` from `u64` but otherwise seems okay to me.

src/cli/download_tracker.rs
rbtcollins: This was mismatched with total_downloaded. Either we promote that to u64 or demote the length to usize, and since we're dealing with small archives, not disk images, usize seemed more appropriate. This existing mismatch was invisible to you because everything was being cast to f64, but if a web server did serve a 2.1G package, rustup would have failed when total_downloaded wrapped - after downloading 2G of data.
With this change, rustup will still fail, but when the content_len header is received. Which is much sooner. So I have shifted the detection earlier in the process but not introduced any new limitations to the code.

src/cli/download_tracker.rs
rbtcollins: Yes, failing earlier.

kinnison (approved):
LGTM, thank you for the note in the commit, incase someone comes along wondering why usize

PR #1839, opened 2019-05-08
Reduce parent stat-per-file to stat-per-dir

rbtcollins:
This is an installation microoptimisation for ~ 3% performance during
unpack performance: stat() calls are not free, involving path traversal,
ACL checking and so forth.

Before: 34s consistently to install rust-docs.
After: 32s-33s to install rust-docs.

A better still approach would be to handle creation errors by creating
the containing directories just in time, but that requires replacing
the tar extraction logic, which I haven't yet figured out how to layer
in cleanly.

### DIFF
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -214,6 +214,7 @@ fn unpack_without_first_dir<R: Read>(archive: &mut tar::Archive<R>, path: &Path)
     let entries = archive
         .entries()
         .chain_err(|| ErrorKind::ExtractingPackage)?;
+    let mut checked_parents: HashSet<PathBuf> = HashSet::new();
     for entry in entries {
         let mut entry = entry.chain_err(|| ErrorKind::ExtractingPackage)?;
         let relpath = {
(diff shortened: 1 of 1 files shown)


src/dist/component/package.rs
```diff
@@ -228,8 +229,21 @@ fn unpack_without_first_dir<R: Read>(archive: &mut tar::Archive<R>, path: &Path)
 
         // Create the full path to the entry if it does not exist already
         match full_path.parent() {
-            Some(parent) if !parent.exists() => {
-                ::std::fs::create_dir_all(&parent).chain_err(|| ErrorKind::ExtractingPackage)?
+            Some(parent) => {
+                if !checked_parents.contains(parent) {
+                    checked_parents.insert(parent.clone().to_owned());
+                    // It would be nice to optimise this stat out, but the tar could be like so:
```
kinnison, line 235: One moderately neat way to do that would be to re-use the components iterator from above, joining and checking as you go; but this is sufficient for now

kinnison (approved):
Looks good to me, I'm not bothered about the concern, as you say it's unlikely to be a problem since tarballs are usually semi-sorted thanks to filesystem layout.

src/dist/component/package.rs
rbtcollins: That would result in stating every element from the top down, which is more than stating from the bottom up.

rbtcollins: (Or I misunderstood :). Thanks for merging it!)

PR #1841, opened 2019-05-09
Ensure data notifications are always received.

rbtcollins:
Files that were exact multiples of the read length would drop the
last read data notification.


kinnison (approved):
LGTM

PR #1842, opened 2019-05-09
More progress bars

rbtcollins:
The progress bar based file reader is relevant for v1 and tarball
packages but not being used - use it more broadly, which will also
improve performance due to the BufReader usage.

Also a drive by lint fix.


kinnison (approved):
LGTM, and thanks for the cleanup too

PR #1844, opened 2019-05-09
Make Y-Yes/N-No consistent and less repetitive.

konstankinollc:
To make the code more consistent, and use fewer match options, we can mimic the same functionality with `.to_lowercase()` for input str.


kinnison (approved):
LGTM

PR #1845, opened 2019-05-09
Better IO buffer when validating dist hashes

rbtcollins:
Also for a bonus, show a progress bar at the same time.


kinnison (approved):
Seems to fit.

Thanks

PR #1846, opened 2019-05-09
Remove unused package functions

rbtcollins:


kinnison (approved):
Dead code is dead.  Thanks.

PR #1847, opened 2019-05-10
Don't set mtime on toolchain files

rbtcollins:
We don't depend on file times in the toolchains for any of our
operations, so making them match the time from the tar archive isn't
particularly useful. My bench machine here dropped from 31/32s for
rust-docs to 26/27s with this change.


kinnison (approved):
I've had a rummage and I'm fairly sure this is safe, and the speedup on Windows is impressive.

PR #1853, opened 2019-05-16
Fix error-chain deprecated warns

### LINKED ISSUE #1852: workaround can be fixed
https://github.com/rust-lang/rustup.rs/blob/master/src/lib.rs#L54 is naming an issue which is the reason for some workaround.

the [issue](https://github.com/rust-lang-nursery/error-chain/issues/254) is now fixed, therefore the workaround can be removed ;)

tesuji:
Closes #1852 .


kinnison (changes requested):
The two code change commits look good, but the `Cargo.lock` update is much more wide-ranging than just `error-chain` as far as I can see.  Could you please only update the `error-chain` dependency (and things which go with it)?  That makes it easier to review.

tesuji:
Done.

kinnison (approved):

PR #1857, opened 2019-05-17
Minor clean up and fix some clippy warnings

tesuji:
r? @kinnison

### DIFF
--- a/download/src/errors.rs
+++ b/download/src/errors.rs
@@ -4,7 +4,7 @@ error_chain! {
     links { }
 
     foreign_links {
-        Io(::std::io::Error);
+        Io(std::io::Error);
         Reqwest(::reqwest::Error) #[cfg(feature = "reqwest-backend")];
     }
 
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -1,4 +1,5 @@
 //! Easy file downloading
+#![deny(rust_2018_idioms)]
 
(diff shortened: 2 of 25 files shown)


kinnison (changes requested):
Thanks, this looks to be a lot of useful cleanups; however the `rustfmt` patch in the middle makes me very sad.  Could you please go through the rebase and fold the formatting into the requisite fix patches instead?

lzutao:
Done.

tesuji:
Done.

kinnison (commented):
This looks good to me.

However before I merge - in a lot of places we have std::foo::stuff where that is (I thought) part of the prelude.  (e.g. `std::result::Result::ok` rather than `Result::ok`) - is the fully qualified path needed?

I won't block merge on changing them if not, but I would like to know.

tesuji:
Because we have shadowed origin Result with Result from `error-chain`.

Edit: Also I got this error:
```rust
error[E0631]: type mismatch in function arguments
   --> src/cli/common.rs:103:10
    |
103 |         .and_then(Result::ok)
    |          ^^^^^^^^
    |          |
    |          expected signature of `fn(std::result::Result<std::string::String, std::io::Error>) -> _`
    |          found signature of `fn(std::result::Result<_, errors::Error>) -> _`

```

kinnison (approved):

PR #1860, opened 2019-05-19
Replace link to #rust-beginners IRC channel

### LINKED ISSUE #1859: Replace link to #rust-beginners IRC on home page
Now that the [IRC Sunset](https://blog.rust-lang.org/2019/04/26/Mozilla-IRC-Sunset-and-the-Rust-Channel.html) has been announced, we should consider replacing the "Ask on #rust-beginners" link with something else. I wanted to open a separate issue from #1372 because this is a bit more urgent now that the IRC channels are going to be shutdown. 

Note that I'm not opening this to argue about the benefits/drawbacks of IRC. The channel is not going to be there anymore at some point and we need to make sure beginners have a way to get unstuck. That's all. 

I think we should replace it with two links:

- To the Rust discord and its beginners channel
- To the Rust users forum

This gives people two avenues to get help instead of one and largely deals with anyone who wouldn't like to join the discord server.

sunjay:
Fixes #1859.

Here's a screenshot of how this looks:

Screenshot from 2019-05-19 17-27-48

There's more text now than before, so it looked really bad on one line. I added line breaks at logical places to make it look a bit better.

The link to the Discord should never expire, but I added some instructions in a comment in case we ever need to regenerate it.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -125,7 +125,15 @@
 </div>
 
 <p>
-  Need help? <a href="https://client00.chat.mibbit.com/?server=irc.mozilla.org&channel=%23rust-beginners">Ask on #rust-beginners</a>.
+  <!-- To regenerate the Discord invite link:
+     1. Open discord
+     2. Click on the invite button on the side of the #beginners channel
+     3. Make sure the link is set to **never expire**
+     4. Alternatively, we can always use the link: https://discord.gg/rust-lang
+        This link does not point them to any particular channel.
+  -->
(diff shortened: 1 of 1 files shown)


kinnison (commented):
One concern I have is that this discord server requires semi-complex interactions in order to begin anyway, perhaps directing them at the main rust-lang link rather than a specific link to `#beginners` might be best.  Many people miss the bot telling them how to gain talk privs.

sunjay:
I personally prefer the more direct link (I think users will figure it out), but I totally get what you're saying as well. I've updated the link to be the more general one.

kinnison (approved):
LGTM, I'll merge this and it'll go live at the next rustup release.

PR #1868, opened 2019-05-23
Add support for `rustup target add all` to install all available targets

### LINKED ISSUE #1653: rustup target add all
Add a simple way of installing all available targets for a particular toolchain, e.g., `rustup target add --all` or similar.

gnzlbg:
Closes #1653 .

---

EDIT: for context, the godbolt machines try to provide all available targets, either by manually hardcoding the ones available for each toolchain, or by parsing the components list (which isn't super machine friendly), and installing them manually. This would simplify things. 

I suspect that Rustup integration tests could also try to install all targets for all toolchains to discover issues (I've seen the installation of available components fail in the past). Sadly I cannot find where these integration tests live.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -275,7 +275,15 @@ pub fn cli() -> App<'static, 'static> {
                     SubCommand::with_name("add")
                         .about("Add a target to a Rust toolchain")
                         .alias("install")
-                        .arg(Arg::with_name("target").required(true).multiple(true))
+                        .arg(
+                            Arg::with_name("target")
+                                .required(true)
+                                .multiple(true)
+                                .help(
+                                    "List of targets to install; \
+                                     \"all\" installs all available targets"
+                                )
+                        )
                         .arg(
                             Arg::with_name("toolchain")
                                 .help(TOOLCHAIN_ARG_HELP)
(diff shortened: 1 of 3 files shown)


kinnison:
> I suspect that Rustup integration tests could also try to install all targets for all toolchains to discover issues (I've seen the installation of available components fail in the past). Sadly I cannot find where these integration tests live.

We don't integration test against the release channels by default -- That's an interesting idea to add though.

kinnison:
> or by parsing the components list (which isn't super machine friendly)

We eventually want nice JSON interaction for this kind of thing, but that's medium to long term

src/cli/rustup_mode.rs
```diff
@@ -851,9 +859,31 @@ fn target_list(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
-    for target in m.values_of("target").expect("") {
-        let new_component = Component::new("rust-std".to_string(), Some(TargetTriple::new(target)));
+    let mut targets: Vec<String> = m
+        .values_of("target")
+        .expect("")
+        .map(ToString::to_string)
+        .collect();
+
+    if targets.contains(&"all".to_string()) {
+        targets.clear();
```
kinnison, line 873: Is it worth emitting a warning here if `targets.len() != 1` ?

kinnison (changes requested):
One UX niggle, otherwise LGTM.

src/cli/rustup_mode.rs
gnzlbg: I'm not sure, I also asked myself this, and did that initially. That way: `rustup target add x86_64-unknown-linux-gnu all` errors. 

But then I thought, what does that cmd actually mean? It reads "install target `x86_64-unknown-linux-gnu` and also all available targets". This is a weird thing to request, as in, the user could just have requested installing all available targets, and that would have the same effect, but it isn't a wrong thing to request.

So at the end I just changed to accept it, and that simplified the code a bit (no need for an error message here, etc.).

But yeah, I am still not sure if this is the right call or not.

gnzlbg: So at the end I just changed to accept it, and that simplified the code a bit (no need for an error message here, etc.)

kinnison: My gut feeling is that the invocation should be **either** `rustup target add foo-bar-gnu...` **or** `rustup target add all` but if noone else backs me up then I'm okay to back down as it were.

gnzlbg: I'm fine either way :) I used to just have an `assert!` here, to change this into an error, is there a different API that I should use ?

liuchong:
A strange requirement... for me 🤣

src/cli/rustup_mode.rs
kinnison: The function returns a `Result<()>` which comes from our error chain module I believe.  You could add a `TargetAllSpecifiedWithTargets` variant to `src/cli/errors.rs` and return that which ought to do the right thing.  Remember at that point to add a test to assert it happens :D

gnzlbg:
@liuchong you don't need to provide VMs that contain all rust targets for a particular nightly.. every night.. ? :D If you have to do this, keeping track of which targets are added and removed every night is a pain, and basically unmaintainable unless you try to parse the human readable output that rustup produces, and even then, because rustup does not tests that all targets are actually "installable", often you need to manually work situations where this fails.

src/cli/rustup_mode.rs
gnzlbg: The last commit should resolve this - thank you for the hints.

kinnison (approved):
This looks good now, thank you very much!

kinnison:
Assuming the CI passes, this can merge later today.

gnzlbg:
CI is green!

PR #1874, opened 2019-05-25
Link to rustup-components-history instead on missing component

kennytm:
The [rust-toolstate](https://rust-lang-nursery.github.io/rust-toolstate/) page records the status of the unpublished master branch, while [rustup-components-history](https://rust-lang.github.io/rustup-components-history) records the status of the published nightly builds. The latter is more accurate for rustup. Since the latter is now organized by us too, let's point the link there instead.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -9,7 +9,7 @@ use url::Url;
 
 pub const TOOLSTATE_MSG: &str =
     "If you require these components, please install and use the latest successful build version,\n\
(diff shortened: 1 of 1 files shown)


kinnison:
This change conflicts with another recently merged, please can you amend it.?

kinnison:
I've re-amended this to be a rebase, should be fine but I'll let the CI run through first.

PR #1878, opened 2019-05-27
Add a PowerShell implementation for rustup-init

fnichol:
This change adds a new PowerShell script (`rustup-init.ps1`) which
performs the same duties and behaviors as the existing Shell-based
script (`rustup-init.sh`).

It is primarily targeted at tier 1 supported Windows platforms, namely:

* `i686-pc-windows-gnu` (32-bit MinGW)
* `i686-pc-windows-msvc` (32-bit MSVC)
* `x86_64-pc-windows-gnu` (64-bit MinGW)
* `x86_64-pc-windows-msvc` (64-bit MSVC)

However, due to the recent portability of PowerShell releases on macOS
and Linux, this script also supports these targets which overlap with
PowerShell targets:

* `x86_64-apple-darwin` (64-bit OS X/macOS)
* `x86_64-unknown-linux-gnu` (64-bit Linux)

As much as possible, the behavior of the existing `rustup-init.sh` is
preserved, including:

* Accepting arguments to be passed through to
  `rustup.init`/`rustup-init.exe`
* Short-circuiting the `-h` and `--help` flags to display an inline
  usage
* Honor the `RUSTUP_UPDATE_ROOT` environment variable if set, and
  defaulting to `https://static.rust-lang.org/rustup`
* Attempting to force TLS v1.2 on Windows
* Download executable to a dedicated temporary directory under the
  system's temp path which is cleaned up on success and failure

For example, to install the latest nightly toolchain as the default in
an unattended scenario, a user can run the following in a PowerShell
session:

```powershell
.\rustup-init.ps1 -y --default-toolchain nightly
```

If `cmd.exe` is your current shell, you can still invoke the same by
running the above in a PowerShell session with:

```batch
powershell -Command ".\rustup-init.ps1 -y --default-toolchain nightly"
```

The script is written carefully to not call `exit` so that it can be
downloaded and used in much the same way as the Shell based "curl-bash"
strategy. Assuming that this script was available at
`https://ps1.rustup.rs` (or `https://pwsh.rustup.rs`,
`https://posh.rustup.rs`, etc.), and that no further arguments are
passed to the script, a user could invoke this script directly in a
PowerShell session with:

```powershell
iex ((New-Object System.Net.WebClient).DownloadString('https://ps1.rustup.rs'))
```
(For folks new to PowerShell, `iex` is an alias for the
`Invoke-Expression` cmdlet, similar to `eval` in other languages)

In order to pass arguments to the script, a slightly longer form can be
used. For example, to perform an unattended install of Rustup with no
default toolchain and no path modification, a user could run the
following in a PowerShell session:

```powershell
& ([scriptblock]::Create((New-Object System.Net.WebClient).DownloadString('https://ps1.rustup.rs'))) -y --default-toolchain none --no-modify-path
```

Yes, that's rather long, but I think that's as short a "code golf" as
you can get and it's on par with Chocolatey's `install.ps1` strategy.

---

## Additional Notes

* As I'm not as familiar with PowerShell as others, it's harder for me to attest to how far back this implementation supports as far as Windows releases go. I'm reasonably sure that this works with PowerShell 3.0, possibly even PowerShell 2.0 meaning this could support as far back as Windows 8 or 7. I would love any additional feedback if I've failed to keep things conservative and portable.
* I tested this on macOS/x86_64 10.14.4 with PowerShell 6.2.0 and Ubuntu/x86_64 18.04 in Docker (Linux 4.9.125-linuxkit) with PowerShell 6.2.1. I believe this covers the vast majority of off-Windows scenarios although any more feedback here would be great as well.
* In order to better test the "curl-bash" functionality, I have the identical code published in a Gist called [A Windows/Linux/macOS PowerShell Rustup Install Script](https://gist.github.com/fnichol/699d3c2930649a9932f71bab8a315b31) which anyone can try by running:
  ```powershell
  & ([scriptblock]::Create((New-Object System.Net.WebClient).DownloadString('https://gist.github.com/fnichol/699d3c2930649a9932f71bab8a315b31/raw/rustup-init.ps1')))
  ```
  I'm using this currently in a [CI setup](https://github.com/fnichol/mtoc/blob/0deec6953f7538169666712ee7c57f1afbeae8b4/.cirrus.yml#L90-L94) which has helped drive better consistency across Windows and non-Windows platforms (yay)

### DIFF
--- a/rustup-init.ps1
+++ b/rustup-init.ps1
@@ -1,0 +1,146 @@
+#!/usr/bin/pwsh
+
+# This is just a little script that can be downloaded from the internet to
+# install rustup. It just does platform detection, downloads the installer and
+# runs it.
+
+# XXX: If you change anything here, please make the same changes in
+# `src/cli/setup_mode.rs` and `rustup-init.sh`. This keeps the usage output of
+# the bootstrapping scripts consistent with the `rustup-init` executable.
+function Write-Usage {
+    Write-Host @'
+rustup-init 1.18.3 (435397f48 2019-05-22)
+The installer for rustup
+
+USAGE:
+    rustup-init.ps1 [FLAGS] [OPTIONS]
+
+FLAGS:
+    -v, --verbose           Enable verbose output
+    -y                      Disable confirmation prompt.
+        --no-modify-path    Don't configure the PATH environment variable
+    -h, --help              Prints help information
+    -V, --version           Prints version information
+
+OPTIONS:
+        --default-host <default-host>              Choose a default host triple
+        --default-toolchain <default-toolchain>    Choose a default toolchain to install
+'@
+}
+
+function Invoke-Main([AllowEmptyCollection()] [string[]]$Argv = @()) {
+    foreach ($arg in $Argv) {
+        if ("$arg" -eq "-h" -or "$arg" -eq "--help") {
+            Write-Usage
+            return
+        }
+    }
+
+    Install-Rustup $Argv
+}
+
+function Get-Architecture {
+    if (($IsWindows -eq $null) -or $IsWindows) {
+        $ostype = "pc-windows-msvc"
+    } elseif ($IsMacOS) {
+        $ostype = "apple-darwin"
+    } elseif ($IsLinux) {
+        $ostype = "unknown-linux-gnu"
+    } else {
+        Write-Failure "unrecognized OS type"
+    }
+
+    if ([System.IntPtr]::Size -eq 8) {
+        $cputype = "x86_64"
+    } elseif ([System.IntPtr]::Size -eq 4) {
+        $cputype = "i686"
+    } else {
+        Write-Failure "unknown CPU type"
+    }
+
+    $arch = "${cputype}-${ostype}"
+    Write-Verbose "Detected arch '$arch'"
+    $arch
+}
+
+# If RUSTUP_UPDATE_ROOT is unset, default it.
+function Get-RustupUpdateRoot {
+    $EnvVar = [System.Environment]::GetEnvironmentVariable("RUSTUP_UPDATE_ROOT")
+    if (($EnvVar -ne $null) -and (Test-Path -Path $EnvVar)) {
+        $EnvVar
+    } else {
+        "https://static.rust-lang.org/rustup"
+    }
+}
+
+function Install-Rustup([AllowEmptyCollection()] [string[]]$Argv) {
+    $arch = Get-Architecture
+    switch -Wildcard ($arch) {
+        "*windows*" { $ext = ".exe"; break }
+        default { $ext = "" }
+    }
+
+    $url = "$(Get-RustupUpdateRoot)/dist/$arch/rustup-init$ext"
+
+    $dir = New-TemporaryDirectory
+    $file = Join-Path $dir "rustup-init$ext"
+
+    try {
+        Write-Verbose "Downloading $url to $file"
+        Set-HighestEncryption
+        (New-Object System.Net.WebClient).DownloadFile($url, $file)
+
+        if ($IsMacOS -or $IsLinux) {
+            Write-Verbose "Setting execute permission on $file"
+            chmod u+x "$file"
+        }
+
+        Write-Verbose "Calling $file $Argv"
+        & "$file" @Argv
+    } finally {
+        Write-Verbose "Cleaning $dir"
+        Remove-Item "$dir" -Force -Recurse
+    }
+}
+
+function New-TemporaryDirectory {
+    $parent = [System.IO.Path]::GetTempPath()
+    [string]$name = [System.Guid]::NewGuid()
+    New-Item -ItemType Directory -Path (Join-Path $parent $name)
+}
+
+function Set-HighestEncryption {
+    # Implementation adapted from the Chocolatey installer script,
+    # see: https://chocolatey.org/install.ps1
+
+    # Attempt to set highest encryption available for SecurityProtocol.
+    # PowerShell will not set this by default (until maybe .NET 4.6.x). This
+    # will typically produce a message for PowerShell v2 (just an info message
+    # though)
+    $OldSPM = [System.Net.ServicePointManager]::SecurityProtocol
+    try {
+        # Set TLS 1.2 (3072) which is currently the highest protocol enabled on
+        # static.rust-lang.org. Favor TLS 1.3 (12288) with a TLS 1.2 fallback
+        # when 1.3 is supported. Use integers because the enumeration values
+        # for TLS 1.2 won't exist in .NET 4.0, even though they are addressable
+        # if .NET 4.5+ is installed (.NET 4.5 is an in-place upgrade).
+        [System.Net.ServicePointManager]::SecurityProtocol = 3072
+    } catch {
+        Write-Failure @'
+rustup: Unable to set PowerShell to use TLS 1.2 due to old .NET Framework
+installed. If you see underlying connection closed or trust errors, you may
+need to do one or more of the following: (1) upgrade to .NET Framework 4.5+ and
+PowerShell v3+, (2) download an alternative install method from
+https://rustup.rs/.
+'@
+    } finally {
+        [System.Net.ServicePointManager]::SecurityProtocol = $OldSPM
+    }
+}
+
+function Write-Failure([string]$Message) {
+    Write-Warning "rustup: $Message"
+    throw
+}
+
+Invoke-Main $args
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -9,7 +9,9 @@ set -u
 # If RUSTUP_UPDATE_ROOT is unset or empty, default it.
 RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
 
-#XXX: If you change anything here, please make the same changes in setup_mode.rs
+# XXX: If you change anything here, please make the same changes in
+# `src/cli/setup_mode.rs` and `rustup-init.sh`. This keeps the usage output of
+# the bootstrapping scripts consistent with the `rustup-init` executable.
 usage() {
     cat 1>&2 <<EOF
 rustup-init 1.18.3 (302899482 2019-05-22)
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -20,7 +20,9 @@ pub fn main() -> Result<()> {
         return Ok(());
     }
 
-    // XXX: If you change anything here, please make the same changes in rustup-init.sh
+    // XXX: If you change anything here, please make the same changes in `src/cli/setup_mode.rs`
+    // and `rustup-init.sh`. This keeps the usage output of the bootstrapping scripts consistent
+    // with the `rustup-init` executable.
     let cli = App::new("rustup-init")
         .version(common::version())
         .about("The installer for rustup")


kinnison:
Hi @fnichol !

Thanks for this, certainly I don't think any of us would have though to write such a beast.  I'm afraid that I'm not sure who will be best placed to review this script, but I will say that there'll be a number of things which will have to happen, some of which need infrastructure team help, before we would be able to offer a `ps1.rustup.rs` or similar.

The most obvious one being that `CONTRIBUTING.md` will need an update to indicate that this script will also need updating when we do a release.

I'll do my best to review this, but honestly I can't attest to the correctness or otherwise of the powershell itself, I'll just look for anything which stands out to me.  I'll need to find a power-shell person to review this from the perspective of "is it good powershell" before I would feel comfortable merging anything though.

Thanks again, and it's exciting that we might have such a thing merged soon.

D.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+FLAGS:
+    -v, --verbose           Enable verbose output
+    -y                      Disable confirmation prompt.
+        --no-modify-path    Don't configure the PATH environment variable
+    -h, --help              Prints help information
+    -V, --version           Prints version information
+
+OPTIONS:
+        --default-host <default-host>              Choose a default host triple
+        --default-toolchain <default-toolchain>    Choose a default toolchain to install
```
kinnison, line 27: I will say that we're close to merging profiles support which will add an extra option here, though I *hope* that's just a text change for this script.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    $arch = "${cputype}-${ostype}"
+    Write-Debug "Detected arch '$arch'"
+    return "$arch"
+}
+
+# If RUSTUP_UPDATE_ROOT is unset, default it.
+function Get-RustupUpdateRoot {
+    if (Test-Path env:\RUSTUP_UPDATE_ROOT) {
```
kinnison: is the backslash intended here?

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+function Set-HighestEncryption {
+    # Implementation adapted from the Chocolatey installer script,
+    # see: https://chocolatey.org/install.ps1
+
+    # Attempt to set highest encryption available for SecurityProtocol.
+    # PowerShell will not set this by default (until maybe .NET 4.6.x). This
+    # will typically produce a message for PowerShell v2 (just an info message
+    # though)
+    try {
+        # Set TLS 1.2 (3072), then TLS 1.1 (768), then TLS 1.0 (192), finally
```
kinnison: Please don't set TLS 1.1 or TLS 1.0 and definitely *NEVER* SSL 3.0 fallbacks.

If the minimum .NET supported version is sufficiently high that this will be awkward then instead try and set `1.2` and failing that, report the warning and continue as you do below.  .NET 4.5 is nearly seven years old, so if that's all that's needed, I'd be tempted to simply mandate it and error rather than warn below.

kinnison (changes requested):
Not bad, though the SSL 3.0 thing is giving me the heebie jeebies.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+# runs it.
+
+#XXX: If you change anything here, please make the same changes in setup_mode.rs
+function Write-Usage {
+    Write-Host @'
+rustup-init 1.18.3 (435397f48 2019-05-22)
+The installer for rustup
+
+USAGE:
+    rustup-init.exe [FLAGS] [OPTIONS]
```
diondokter: Is this correct? Shouldn't it be 'rustup-init.ps1 [FLAGS] [OPTIONS]'?

rustup-init.ps1
fnichol: Yeah, it should only be a text change here. I took this implementation cue from the `rustup-init.sh` script which intercepts calls with `-h` or `--help` and outputs this script-copied usage rather than downloading the `rustup-init` program and calling help on it. I'm guessing this was a resource/network saving move so I duplicated the spirit of it. Call it, some genuine "cargo culting" 😄

rustup-init.ps1
fnichol: It turns out, that, yes, it is intended here. In PowerShell filesystems, but also the environment variables are contained in PowerShell Drives, and the `Env:` is no different:

```powershell
> Get-PSDrive

Name           Used (GB)     Free (GB) Provider      Root                                                                                                                                                                                                                                                             CurrentLocation
----           ---------     --------- --------      ----                                                                                                                                                                                                                                                             ---------------
Alias                                  Alias
C                   0.98         18.90 FileSystem    C:\                                                                                                                                                                                                                           Users\fnichol\Projects\github.com\fnichol\mtoc\.ci
Cert                                   Certificate   \
Env                                    Environment
Function                               Function
HKCU                                   Registry      HKEY_CURRENT_USER
HKLM                                   Registry      HKEY_LOCAL_MACHINE
Variable                               Variable
WSMan                                  WSMan
```
In this case the `Test-Path` cmdlet works on any PowerShell drives, hence that backslash that looks like a mistake here, but for the `C` drive, it would be `C:\temp`. I know, I've learned some things here too

rustup-init.ps1
fnichol: Now *that* is something I can dial back. Admittedly, this was heavily inspired by Chocolatey's `install.ps1` and I suspect that they are targeting a much wider set of install targets. Considering anyone installing Rust may likely already have Chocolatey installed (and therefore could install more software), this might be overkill. In fact, it might make sense to remove that function completely and wait for any issues that would prompt taking action. Make sense?

rustup-init.ps1
fnichol: Ah right, this had me confused for a moment when I was reading the [`rustup-init.sh`](https://github.com/rust-lang/rustup.rs/blob/master/rustup-init.sh) implementation. It turns out that the usage in both scripts are copied from the output of calling the executable with `rustup-init --help`. I'm pretty sure that this was so that users could get the full set of flags and options without having to download the full binary first. The first iteration of this script was only focused on Windows installations (hence the `.exe` on the end), but it ended up being really easy to support macOS and Linux so that `.exe` gives me squinty eyes now… However, given more Windows users would see this output, I suppose that seeing `.exe` in a principle of least surprise situation?

fnichol:
@kinnison Thanks! Yes, I thought this would certainly not be a straight forward change 😄 Rather than submitting something like this out-of-the-blue which updated docs, CI, etc. I was hoping this would be a bit less invasive as a first step. I'm more than happy to help with the remaining changes necessary to push this out more broadly and I figured it would be pretty cross-cutting.

rustup-init.ps1
kinnison: Okay, that works for me.

rustup-init.ps1
kinnison: Wow, that's fun.  Thanks for the explanation.

rustup-init.ps1
kinnison: I think you should attempt to explicitly set TLS 1.2 and fail if that can't be done.

rustup-init.ps1
kinnison: Mmm, really that should be `.ps1` and I ought to ensure that `rustup-init.sh` says `rustup-init.sh` :D

kinnison:
Cool, so let's get the first thing out of the way and get something merged.  At worst people can then use `https://raw.githubusercontent.com/rust-lang/rustup.rs/master/rustup-init.ps1` rather than `https://ps1.rustup.rs` to get started.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    $platform = [System.Environment]::OSVersion.Platform
+    switch -Wildcard ($platform) {
+        # Linux and Darwin both report as Unix
+        "Unix" {
+            $ostype = uname -s
+            $cputype = uname -m
+            break
+        }
+        "Win*" {
+            $ostype = (Get-WmiObject Win32_OperatingSystem).Caption
```
glennsarti: Get-WMI is not available in PowerShell Core.
https://docs.microsoft.com/en-us/powershell/scripting/whats-new/breaking-changes-ps6?view=powershell-6#wmi-v1-cmdlets

Use `Get-CimInstance` instead

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env powershell
```
glennsarti: PowerShell Core is not available here.  `/usr/bin/pwsh` is the correct location
https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-6#uninstalling-binary-archives

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+rustup: Unable to set PowerShell to use TLS 1.2 and TLS 1.1 due to old .NET
+Framework installed. If you see underlying connection closed or trust errors,
+you may need to do one or more of the following: (1) upgrade to .NET Framework
+4.5+ and PowerShell v3, (2) use the Download of rustup-init.exe method of
+install. See https://rustup.rs/ for all install options.
+'@
+    }
+}
+
+function Write-Error([string]$Message) {
```
glennsarti: This is definitely a no-no.  Do not override inbuilt PowerShell functions such as Write-Error, Write-Host etc. etc.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env powershell
+
+# This is just a little script that can be downloaded from the internet to
+# install rustup. It just does platform detection, downloads the installer and
+# runs it.
+
+#XXX: If you change anything here, please make the same changes in setup_mode.rs
+function Write-Usage {
```
glennsarti, line 10: Warning - This is not idiomatic PowerShell.

PowerShell has a very rich, default help system, which means this is not required.  Instead should be using script parameters e.g.

```
#!/usr/bin/env powershell
<#
.Synopsis
   Installs RustUp
.DESCRIPTION
   This is just a little script that can be downloaded from the internet to install rustup. It just does platform detection, downloads the installer and runs it.
.PARAMETER DefaultHost
  The Host to install RustUp onto
.PARAMETER DefaultToolChain
  ...
  
.EXAMPLE
   ./rustup-init -DefaultHost '127.0.0.1'
#>

[CmdletBinding()]
Param(
  [Switch]$NoModiyPath,

  [Switch]$ShowVersion,

  [String]$DefaultHost,

  [ValidateSet("gnu", "msvc")]
  [String]$DefaultToolchain]
)

...
```

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    Write-Host @'
+rustup-init 1.18.3 (435397f48 2019-05-22)
+The installer for rustup
+
+USAGE:
+    rustup-init.exe [FLAGS] [OPTIONS]
+
+FLAGS:
+    -v, --verbose           Enable verbose output
```
glennsarti, line 19: Verbose is already implemented in the default cmdlet binding.  Do not re-implement
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_commonparameters?view=powershell-6

Use `$VerbosePreference -eq 'Continue'` to detect a `--version` flag instead,

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    -y                      Disable confirmation prompt.
```
glennsarti, line 20: Confirmation is already implemented in the default cmdlet binding. Do not re-implement.
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_commonparameters?view=powershell-6

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Preference_Variables?view=powershell-6#confirmpreference

Use `$ConfirmPreference` instead

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+        --no-modify-path    Don't configure the PATH environment variable
+    -h, --help              Prints help information
```
glennsarti, line 22: Help is standard in all PowerShell scripts and functions.  Do not re-implement.

Instead use the comment based help
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-6

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    -V, --version           Prints version information
+
+OPTIONS:
+        --default-host <default-host>              Choose a default host triple
+        --default-toolchain <default-toolchain>    Choose a default toolchain to install
+'@
+}
+
+function main([AllowEmptyCollection()] [string[]]$Argv = @()) {
```
glennsarti: This is not idiomatic PowerShell.

PowerShell already has a very rich parameter system.  Do not re-implement.  See comment above about using the `param(.,....)` code

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+            Write-Usage
+            return
+        }
+    }
+
+    Install-Rustup $Argv
+}
+
+function Get-Architecture {
+    $platform = [System.Environment]::OSVersion.Platform
```
glennsarti: PowerShell Core already has helpers for xplat developers.  Specifically $IsMacOS, $IsLinux and $IsWindows
https://stackoverflow.com/questions/44703646/determine-the-os-version-linux-and-windows-from-powershell

If $IsWindows is `$null` then you're using Windows PowerShell, which is obviously Windows only.

This function is basically moot.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+            $ostype = (Get-WmiObject Win32_OperatingSystem).Caption
+            $cputype = (Get-WmiObject Win32_OperatingSystem).OSArchitecture
+            break
+        }
+        default {
+            Write-Error "could not determine OS platform: $platform"
+        }
+    }
+
+    switch -Wildcard ($ostype) {
```
glennsarti: See above.  Use inbuilt helper variables

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+        "*Windows*" {
+            $ostype = "pc-windows-msvc"
+            break
+        }
+        default {
+            Write-Error "unrecognized OS type: $ostype"
+        }
+    }
+
+    switch ($cputype) {
```
glennsarti: Much easier to just use integer width, which is crossplatform and doesn't require string interpolation.

```
$is64bit = ([System.IntPtr]::Size -eq 8)
$is32bit = ([System.IntPtr]::Size -eq 4)
```

ref - https://ridicurious.com/2018/10/17/4-ways-to-find-os-architecture-using-powershell-32-or-64-bit/

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+        "x86_64" {
+            break
+        }
+        default {
+            Write-Error "unknown CPU type: $cputype"
+        }
+    }
+
+    $arch = "${cputype}-${ostype}"
+    Write-Debug "Detected arch '$arch'"
```
glennsarti: Nit: This is a verbose statement not debug.  Remember that `Write-Debug` can be used to halt script progress.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    return "$arch"
+}
+
+# If RUSTUP_UPDATE_ROOT is unset, default it.
+function Get-RustupUpdateRoot {
+    if (Test-Path env:\RUSTUP_UPDATE_ROOT) {
+        return "$env:RUSTUP_UPDATE_ROOT"
```
glennsarti: nit: return is un-necessary.  Just output the string.

nit: Wrapping it in quotes is un-necessary.  It's already a string.

rustup-init.ps1
glennsarti: nit: To be more correct. The ENV: drive is a PowerShell Provider. And you traverse providers like a filesystem.  Think inbuilt-Fuse.

rustup-init.ps1
glennsarti:
```
function Get-RustupUpdateRoot {
    if (Test-Path env:\RUSTUP_UPDATE_ROOT) {
        env:RUSTUP_UPDATE_ROOT
    } else {
        "https://static.rust-lang.org/rustup"
    }
}
```

Warning : Watch out for $null values for the ENV as Test-Path will throw

e.g.

```
C:\Source> Test-Path $null
Test-Path : Cannot bind argument to parameter 'Path' because it is null.
At line:1 char:11
+ Test-Path $null
```

glennsarti: Actually in retrospect...this whole thing won't work.   `env:RUSTUP_UPDATE_ROOT` returns an object, not a string. To get the value you need to call the `.value` method.  This is better;

```
function Get-RustupUpdateRoot {
    $UpdateRoot = [System.Environment]::RUSTUP_UPDATE_ROOT
    if (($UpdateRoot -ne $null) -and (Test-Path -Path $UpdateRoot)) {
        $UpdateRoot
    } else {
        "https://static.rust-lang.org/rustup"
    }
}
```

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    $arch = Get-Architecture
+    switch -Wildcard ($arch) {
+        "*windows*" { $ext = ".exe"; break }
+        default { $ext = "" }
+    }
+
+    $url = "$(Get-RustupUpdateRoot)/dist/$arch/rustup-init$ext"
+
+    $dir = New-TemporaryDirectory
+    $file = Join-Path "$dir" "rustup-init$ext"
```
glennsarti: nit: `"$dir"` is not necessary, just use `$dir`

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+
+    try {
+        if ("$arch" -match "windows") { Set-HighestEncryption }
```
glennsarti: nit: Use cross platform variable helpers instead

```
if (($IsWindows -eq $null) -or $IsWindows) { Set-HighestEncryption }

File: rustup-init.ps1 [hunk 20]
```diff
@@ -0,0 +1,174 @@
...
+
+        Write-Debug "Downloading $url to $file"
```
glennsarti: nit: this is verbose not debug

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+        (New-Object System.Net.WebClient).DownloadFile($url, $file)
+        if (-Not ("$arch" -match "windows")) { chmod u+x "$file" }
+
+        Write-Debug "Calling '$file $Argv'"
```
glennsarti: nit: this is verbose not debug

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+        & "$file" @Argv
+    } finally {
+        Write-Debug "Cleaning $dir"
```
glennsarti: nit: this is verbose not debug

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+# If RUSTUP_UPDATE_ROOT is unset, default it.
+function Get-RustupUpdateRoot {
+    if (Test-Path env:\RUSTUP_UPDATE_ROOT) {
+        return "$env:RUSTUP_UPDATE_ROOT"
+    } else {
+        return "https://static.rust-lang.org/rustup"
+    }
+}
+
+function Install-Rustup([AllowEmptyCollection()] [string[]]$Argv) {
```
glennsarti, line 76: This is not idomatic PowerShell.  Used named parameters, not $argv.

If you insist on this style, use the `ValueFromRemainingArguments` attribute which lumps unbounded parameters in a an array of string

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-6#valuefromremainingarguments-argument

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+    } finally {
+        Write-Debug "Cleaning $dir"
+        Remove-Item "$dir" -Force -Recurse
+    }
+}
+
+function New-TemporaryDirectory {
+    $parent = [System.IO.Path]::GetTempPath()
+    [string]$name = [System.Guid]::NewGuid()
+    New-Item -ItemType Directory -Path (Join-Path "$parent" "$name")
```
glennsarti: nit: `"$parent" "$name"` doesn't require the strings.  use `$parent $name`

rustup-init.ps1
glennsarti: Also...You should be a good citizen here.  Changing that method will affect the user AFTER you run this script.

Instead you should 1) set it, 2) do the download and then 3) restore the previous value.

rustup-init.ps1
```diff
@@ -0,0 +1,174 @@
...
+'@
+    }
+}
+
+function Write-Error([string]$Message) {
+    Write-Warning "rustup: $Message"
+    throw
+}
+
+main $args
```
glennsarti: `main` is not a recommended function name.  Functions should follow the verb-noun syntax.  Suggest 'Invoke-Main`

glennsarti (commented):
Please don't be put off with the number of comments.  This is a common thing when people try to convert bash scripts into PowerShell.

After having gone down the same path with some of our Puppet based bootstrapping scripts (I work for Puppet, Windows team) general feedback is:

1) PLEASE PLEASE PLEASE don't force powershell users into a bash shell experience.  PowerShell has a rich, VERY well known, and EXPECTED behaviour with respect to help and parameter passing.

e.g. using -h IS not a standard practice for getting help

By using idomatic PowerShell you will not need to author a usage function nor will you need to re-implement many of the things PowerShell already does, and has done, for many years

2) Your users will expect a PowerShell script to behave like a PowerShell script i.e.  `.\script -Verbose` should show the Verbose stream, not `--verbose`.  PowerShell users expect tab completion for parameter names. PowerShell users expect `Get-Help` to operate correctly on your scripts.

---

If you only choose one thing in my list of changes, _please_ modify the script to use the Advanced Functions and Cmdlet binding.  This will be meeting-your-users-where-they-are instead of forcing them to conform to a distinctly un-PowerShell syntax inside a PowerShell shell.

glennsarti:
> under the system's temp path 

Nit: Not strictly true.  System temp path is different to User temp path.  In this case you're using the users TMP -> TEMP -> USERPROFILE environment variable.

https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettemppath?view=netframework-4.8#remarks

If you use the System temp path you may run into permission issues, where the user can't write to `C:\Windows\System\Temp`

> `iex ((New-Object System.Net.WebClient).DownloadString('https://ps1.rustup.rs'))`

That's PowerShell 2 compatible.  For PS3+ you can use:

`iwr 'https://ps1.rustup.rs' | iex` which is far close to `curl .. | bash`  (iwr = Invoke-WebRequest, iex = Invoke-Expression)

glennsarti:
Also @fnichol , THANKYOU for starting this.  More cross platform tools need Windows/PowerShell ❤️ ❤️

rustup-init.ps1
kinnison: I think you've partly misunderstood - these arguments are passed to the downloaded `rustup-init.exe`

glennsarti: True, but these are also the commands that the PowerShell script is expecting. The PowerShell script should be taking the PowerShell style parameters and crafting the call to `rustup-init.exe`.  Currently it's passing verbatim (linux style parameters) with a short circuit for help.

cdhunt: Like @glennsarti said, Verbose and Confirm parameters are provided by the cmdlet binding, but you still have to implement the logic. In this case, reading the PowerShell `$ConfirmPreference` and translating that to the appropriate flag for `rustup-init.exe`. This allows a PowerShell users to set `$VerbosePreference` and `$ConfirmPreference` higher in the call stack or globally in their session without having to know the internals of `rustup-init`.

rustup-init.ps1
cdhunt: You can also add a parameter block with cmdlet binding to the script itself (as in @glennsarti's [comment above](https://github.com/rust-lang/rustup.rs/pull/1878/files#r287902297) which essentially turns the script into a function which gets you all of the built-in PowerShell parameter binding, checking, and help. This basically makes the script operates like a  function and the logic you currently have in `Install-Rustup` should be in the body of the script and you can remove `main`.

i.e.
```powershell
Get-Help rustup-init.ps1
```

rustup-init.ps1
kinnison: While I have a weak preference for ensuring that the command line syntax is the same for `rustup-init.ps1` as it will be for `rustup-init.exe` I am not motivated to block this rework if it's over-all desirability to Windows developers is high.

It's worth noting though that this is, in general, only going to be used in CI situations since most developers will simply download `rustup-init.exe` and double-click on it.

cdhunt: The `-y` looks to be the only argument that would be functionally different. Granted, the PowerShell `Confirm` parameter usage is a bit counter-intuitive.

rustup-init.ps1
fnichol: Ah, you are totally right and I forgot to qualify that

rustup-init.ps1
fnichol: Right--I was burning myself earlier when I had `exit` calls in the script. Did not work so well when you pulled that into a scriptblock for execution--I couldn't understand why my session…exited 😄 

I'll see what I can do here…

rustup-init.ps1
fnichol: Okay, I'll update that--makes sense

rustup-init.ps1
glennsarti: > It's worth noting though that this is, in general, only going to be used in CI situations since most developers will simply download rustup-init.exe and double-click on it.

Maybe initially, but I strongly disagree moving forward. People *will* find it's much easier to use a PowerShell script than go-to-website-download-exe-double-click-next-next-finish. 

Applying the same argument to the `https://sh.rustup.rs` script.  Why do you have it? Is that just for CI? nope.  Users found that useful instead of go-to-website-download-rpm/deb/dmg-run-that

rustup-init.ps1
fnichol: I should mention that I was hoping to support several older versions of PowerShell on the Windows side if it was trivial enough. The `Get-CimInstance` was one situation where I wasn't sure what version of PowerShell first shipped with `Get-CimInstance` vs. the older `Get-WmiObject`. Thank you for that link--is there a document or set of docs that would spell out these portability questions?

rustup-init.ps1
glennsarti: Trying to find some prior art for you ....

rustup-init.ps1
fnichol: Ack, you are right! That has to change

rustup-init.ps1
glennsarti: Given my comments below about using the xplat variables, this WMI call is probably unnecessary i.e.

`($null -eq $IsWindows) -or $IsWindows)` is enough to detect if you're on a Windows platform.

rustup-init.ps1
fnichol: Ah, that makes sense (I took this from another script) and it makes sense who consumes this (i.e. non-Windows systems).

rustup-init.ps1
glennsarti: I couldn't find it but basically

``` powershell
$OldSPM = [System.Net.ServicePointManager]::SecurityProtocol
try {
   <do lots of stuff here>
  [System.Net.ServicePointManager]::SecurityProtocol = <new value>

  <Download what you need>
} finally {
  [System.Net.ServicePointManager]::SecurityProtocol = $OldSPM
}
```

... or something along those lines.

rustup-init.ps1
fnichol: Yeah, I went back and forth over this. The built in help system is much better (you get autocomplete and docs for free), however this was copying the strategy used in the `rustup-init.sh` implementation. In essence, it's as thin a wrapper around the `rustup-init.exe` executable as possible. Meaning, that if and when `rustup-init.exe` gains new flags and options, then this script passes those through without dealing with them directly. The strange situation here is that we're displaying a *copy* of the help usage from `rustup-init.exe --help` (I presume) so users know what options are available on the executable without downloading it first. If we went the param route, we'd have to wire each flag and option back into the executable and account for future changes. It's one of those situations where I'd have left a comment at the top to explain: "we know this isn't remotely idiomatic, but it's a pass-through to the executable". Does that make sense?

rustup-init.ps1
fnichol: Same as comment above. Also, I learned about `$DebugPreference = 'Continue'` while working on this and it's GREEEAT!

rustup-init.ps1
fnichol: Ah, even better. That would be a `Write-Verbose` I presume?

rustup-init.ps1
glennsarti: Sure and I do understand the conundrum.  We have the same problem with our Puppet bootstrapping scripts being cross platform.

So perhaps there's a halfway solution here
1) For the options that we know conflict with PowerShell norms (specifically `-v`, `-h` and `-y`) they should be catered for and wired up appropriately.  We know that won't (99.99% sure) change.  Then you can use the Unbounded Parameters to capture the other options/commands etc. and pass verbatim.

This avoids the the churn when rustup adds new stuff.

2) You REALLY REALLY should have the documentation (`<# Synopsis...`) stuff at the top explaining that this script is a wrapper for xxxxx and copy the Write-Usage stuff into there.

glennsarti: Ultimately it would be nice to autogenerate this script (and the .sh version too) from the rust-init build process but that's WAAAAY outside the scope of this PR.

rustup-init.ps1
glennsarti: Correct.

rustup-init.ps1
fnichol: I finally understand what you're getting at here. It didn't help that I was trying to understand these variables using PowerShell 5 on Windows Server Core. Then I saw those three variables when running on macOS coming up as `True` or `False`. Tricky 😉

rustup-init.ps1
fnichol: Okay, I went for TLS 1.2 and a fail if that didn't succeed. I also used the save-try-finally-restore example above and feel much better about that.

rustup-init.ps1
fnichol: The whole `Get-Architecture` reads much better now and is a fraction of the size, thank you! I also learned that there is an implicit return in these functions which feels very Rust-y.

rustup-init.ps1
fnichol: I changed this to `Write-Failure`. Still might not the best of names but it should no longer shadow a built in cmdlet.

rustup-init.ps1
fnichol: I'll comment below about the main/params/etc

rustup-init.ps1
glennsarti: > I also learned that there is an implicit return

Yeah, sorta-kinda. PowerShell is about the pipeline, so _ANY_ output generated by the function is put onto the pipeline, not just the last statement ala Ruby, Rust.

This can trip you up when when later, but fortunately not in this script.

rustup-init.ps1
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/pwsh
+
+# This is just a little script that can be downloaded from the internet to
+# install rustup. It just does platform detection, downloads the installer and
+# runs it.
+
+#XXX: If you change anything here, please make the same changes in setup_mode.rs
```
rbtcollins: This should say setup_mode.rs and rustup-init.sh; the same comment in both setup_mode.rs and rustup-init.sh needs to be updated to point also to this new file: the command line options should be kept in sync in all three places in the intent of the comment. It could perhaps be clarified too.

rustup-init.ps1
rbtcollins: FWIW I agree with @glennsarti about the need to explain the intent in this file.

However it is not a wrapper - it is a bootstrapper; the difference is this: This script has to determine what exact precompiled rustup binary to download, then download it, and finally chain into it - after that it has no use and the user will never run it again. A wrapper we would expect to have the binary already present and be adapting it in some way, and be always used.

Does this matter? Maybe not, but if we are documenting the intent up the top for other people, I think we shouldn't confuse them with non-idiomatic use of such terms.

rustup-init.ps1
rbtcollins: I think the important thing here is that:
 - the documentation on how to use this script is clear
 - it interacts with the chained rustup-init appropriately: that is, -y is passed to rustup-init if the user selected that

I don't think it matters whether users will find the ps1 script better or worse, or more or less appealing than the website: we'll only find that out by measuring what happens after this is merged, so lets focus on making that happen in a way that is sustainable and works well in the Powershell ecosystem (which I know nothing about, and shall be refraining from adding chaos to the review because of that).

rustup-init.ps1
```diff
@@ -0,0 +1,142 @@
...
+    # Implementation adapted from the Chocolatey installer script,
+    # see: https://chocolatey.org/install.ps1
+
+    # Attempt to set highest encryption available for SecurityProtocol.
+    # PowerShell will not set this by default (until maybe .NET 4.6.x). This
+    # will typically produce a message for PowerShell v2 (just an info message
+    # though)
+    $OldSPM = [System.Net.ServicePointManager]::SecurityProtocol
+    try {
+        # Set TLS 1.2 (3072). Use integers because the enumeration values for
```
rbtcollins: I think its worth a note in here about why you are not setting TLS 1.3. (presumably because static.r-l.org isn't TLS1.3 enabled yet, but whatever the reason, please do capture it.

rustup-init.ps1
```diff
@@ -0,0 +1,142 @@
...
+        # TLS 1.2 won't exist in .NET 4.0, even though they are addressable if
+        # .NET 4.5+ is installed (.NET 4.5 is an in-place upgrade).
+        [System.Net.ServicePointManager]::SecurityProtocol = 3072
+    } catch {
+        Write-Failure @'
+rustup: Unable to set PowerShell to use TLS 1.2 due to old .NET Framework
+installed. If you see underlying connection closed or trust errors, you may
+need to do one or more of the following: (1) upgrade to .NET Framework 4.5+ and
+PowerShell v3, (2) use the Download of rustup-init.exe method of install. See
```
rbtcollins: For (2) here, as you have made this script support multiple OS's, I think you should just say "(2) Download an alternative install method from https://rustup.rs/ ."

rbtcollins (changes requested):

rbtcollins:
Thank you for this patch - I forgot to add that in the review!

rustup-init.ps1
fnichol: Good call, I updated this comment and included the Rust source and the Shell script in this change, lest they be forgotten later.

rustup-init.ps1
fnichol: I added a bit more explanation about why we're not yet setting TLS v1.3--short reason being I don't believe that it's supported on https://static.rust-lang.org currently. The comment calls this out and provides the enum value for TLS v1.3 so we can add that when appropriate.

Additionally after looking at this behavior, it seemed appropriate to set this for macOS and Linux. I tested these platforms out and they all behaved as expected.

rustup-init.ps1
fnichol: Even better, done!

fnichol:
I still have a few top comments to write in the script and to sort out the entrypoint function arguments. However, I think this is getting much closer and is a far better result than when we started. Thanks all for your care and feedback ❤️

kinnison:
@fnichol Thanks again for your efforts, I apologise for being quiet for so long, but this looks really exciting.

Once thing which might help matters a bit -- We automatically use the `rustup-init.sh` from the repo when bootstrapping Rust on Travis -- would you have a go at altering `appveyor.yml` to use `rustup-init.ps1` similarly, so we can see it being used in CI?

kinnison:
Hi @fnichol

This PR has sat idle for quite some time.  Did you lose interest in continuing the work, or perhaps simply run out of time to be able to contribute?  If so, then that's no problem, but if you think you might be able to rebase onto current master, and go over the remaining open comments and resolve them some way, then we'd still be interested in this work being merged.

Thanks,

Daniel.
(And happy new year)

PR #1880, opened 2019-05-28
CONTRIBUTING.md: Explain version numbering

kinnison:
This goes into some detail about how version numbers are constructed
so that in future if something goes squiffy with a release version
or if we're asking people to test non-release versions of `rustup`
then we can be confident of things when we get them to tell us
what version they ran.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,53 @@ submitted for inclusion in the work by you, as defined in the
 Apache-2.0 license, shall be dual licensed as in the README, without any
 additional terms or conditions.
 
+## Version numbers
+
+If you ever see a released version of rustup which has `::` in its version string
+then something went wrong with the CI and that needs to be addressed.
+
+We use `git-testament` to construct our version strings. This records, as a
+struct, details of the git commit, tag description, and also an indication
+of modifications to the working tree present when the binary was compiled.
+
+During normal development you may get information from invoking `rustup --version`
+which looks like `rustup-init 1.18.3+15 (a54051502 2019-05-26)` or even
+`rustup-init 1.18.3+15 (a54051502 2019-05-26) dirty 1 modification`.
+
+The first part is always the binary name as per `clap`'s normal operation. The
+version number is a combination of the most recent tag in the git repo, and the
+number of commits since that tag. The parenthesised information is, naturally,
+the SHA of the most recent commit and the date of that commit. If the indication
+of a dirty tree is present, the number of changes is indicated.  This combines
+adds, deletes, modifies, and unknown entries.
+
+You can request further information of a `rustup` binary with the
+`rustup dump-testament` hidden command.  It produces output of the form:
+
+```shell
+$ rustup dump-testament
+Rustup version renders as: 1.18.3+15 (a54051502 2019-05-26) dirty 1 modification
+Current crate version: 1.18.3
+Built from branch: kinnison/version-strings
+Commit info: 1.18.3+15 (a54051502 2019-05-26)
+Modified: CONTRIBUTING.md
+```
+
+This can be handy when you are testing development versions on your PC
+and cannot remember exactly which version you had installed, or if you have given
+a development copy (or instruction to build such) to a user, and wish to have them
+confirm *exactly* what they are using.
+
+Finally, we tell `git-testament` that we trust the `stable` branch to carry
+releases.  If the build is being performed when not on the `stable` branch, and
+the tag and `CARGO_PKG_VERSION` differ, then the short version string will include
+both, in the form `rustup-init 1.18.3 :: 1.18.2+99 (a54051502 2019-05-26)` which
+indicates the crate version before the rest of the commit.
+On the other hand, if the build was on the `stable` branch then regardless
+of the tag information, providing the commit was clean, the version is
+always replaced by the crate version.  The `dump-testament` hidden command can
+reveal the truth however.
+
 ## Making a release
 
 Before making a release, ensure that `rustup-init.sh` is behaving correctly,


CONTRIBUTING.md
```diff
@@ -29,6 +29,50 @@ submitted for inclusion in the work by you, as defined in the
 Apache-2.0 license, shall be dual licensed as in the README, without any
 additional terms or conditions.
 
+## Version numbers
+
+We use `git-testament` to construct our version strings. This includes into
```
nrc:
```suggestion
We use `git-testament` to construct our version strings. This includes
```

CONTRIBUTING.md
```diff
@@ -29,6 +29,50 @@ submitted for inclusion in the work by you, as defined in the
...
+the binary details of the git commit, tag description, and also an indication
```
nrc: "binary details" - do you mean the SHA?

CONTRIBUTING.md
```diff
@@ -29,6 +29,50 @@ submitted for inclusion in the work by you, as defined in the
...
+of modifications to the working tree present when the binary was compiled.
+
+This means that in normal development, you might get information from invoking
+`rustup --version` which look like `rustup-init 1.18.3+15 (a54051502 2019-05-26)`
```
nrc:
```suggestion
`rustup --version` which looks like `rustup-init 1.18.3+15 (a54051502 2019-05-26)`
```

CONTRIBUTING.md
```diff
@@ -29,6 +29,50 @@ submitted for inclusion in the work by you, as defined in the
...
+Current crate version: 1.18.3
+Built from branch: kinnison/version-strings
+Commit info: 1.18.3+15 (a54051502 2019-05-26)
+Modified: CONTRIBUTING.md
+```
+
+This can be super-handy when you are testing development versions on your PC
+and cannot remember exactly which version you had installed.
+
+Finally, we set trust by the `stable` branch.  If the `stable` branch is not
```
nrc: "Finally, we set trust by the `stable` branch" I don't know what this means

CONTRIBUTING.md
rbtcollins: I suggest dropping the 'This means'. Just say 'In normal development you may ...', as you have changes to make anyway.

CONTRIBUTING.md
```diff
@@ -29,6 +29,50 @@ submitted for inclusion in the work by you, as defined in the
...
 
+## Version numbers
+
+We use `git-testament` to construct our version strings. This includes into
+the binary details of the git commit, tag description, and also an indication
+of modifications to the working tree present when the binary was compiled.
+
+This means that in normal development, you might get information from invoking
+`rustup --version` which look like `rustup-init 1.18.3+15 (a54051502 2019-05-26)`
+or even like `rustup-init 1.18.3+15 (a54051502 2019-05-26) dirty 1 modification`.
```
rbtcollins: perhaps drop the second like here - 'or even `rustup-ini...'

CONTRIBUTING.md
```diff
@@ -29,6 +29,50 @@ submitted for inclusion in the work by you, as defined in the
...
+```shell
+$ rustup dump-testament
+Rustup version renders as: 1.18.3+15 (a54051502 2019-05-26) dirty 1 modification
+Current crate version: 1.18.3
+Built from branch: kinnison/version-strings
+Commit info: 1.18.3+15 (a54051502 2019-05-26)
+Modified: CONTRIBUTING.md
+```
+
+This can be super-handy when you are testing development versions on your PC
```
rbtcollins: Personally I find the superlative here a little jarring, but whats life without a little fun :). Just mentioning it for you to consider. I find it interesting that you do this, because I literally never install rustup when testing it, except for testing the proxies and self-update code I guess...

CONTRIBUTING.md
```diff
@@ -29,6 +29,50 @@ submitted for inclusion in the work by you, as defined in the
...
+used, and the tag and `CARGO_PKG_VERSION` differ, then the short version
+string will include both, in the form `rustup-init 1.18.3 :: 1.18.2+99 (a54051502
+2019-05-26) dirty 1 modification` which indicates the crate version before
+the rest of the commit.  If the `stable` branch was in the build then regardless
+of the tag information, providing the commit was clean, then the version is
+always replaced by the crate version.  The `dump-testament` hidden command can
+reveal the truth however.
+
+If you ever see a released version of rustup which has `::` in its version string
+then something went wrong with the CI and that needs to be addressed.
```
rbtcollins: I suggest putting this at the very start of the whole section. The rest is merely detail to read once this happens :)

rbtcollins (commented):
I'd be ok with this landing as-is, but I think it can be better too, so - up to you and Nick I guess :)

CONTRIBUTING.md
kinnison: This is a misunderstanding of a badly worded phrase "This includes-into-the-binary ..." I shall attempt to reword more nicely.

kinnison:
```markdown
We use `git-testament` to construct our version strings. This records, as a
struct, details of the git commit, tag description, and also an indication
of modifications to the working tree present when the binary was compiled.
```

CONTRIBUTING.md
kinnison: Agreed, done.

CONTRIBUTING.md
kinnison: Reworded as:

```markdown
Finally, we tell `git-testament` that we trust the `stable` branch to carry
releases.  If the build is being performed when not on the `stable` branch, and
the tag and `CARGO_PKG_VERSION` differ, then the short version string will include
both, in the form `rustup-init 1.18.3 :: 1.18.2+99 (a54051502 2019-05-26)` which
indicates the crate version before the rest of the commit.
On the other hand, if the build was on the `stable` branch then regardless
of the tag information, providing the commit was clean, the version is
always replaced by the crate version.  The `dump-testament` hidden command can
reveal the truth however.
```

CONTRIBUTING.md
kinnison: Yes, I agree.

CONTRIBUTING.md
kinnison: Dropped the superlative, added a mention of users having non-release builds:

```markdown
This can be handy when you are testing development versions on your PC
and cannot remember exactly which version you had installed, or if you have given
a development copy (or instruction to build such) to a user, and wish to have them
confirm *exactly* what they are using.
```

CONTRIBUTING.md
kinnison: I have no idea why I didn't think of that.  Sigh.  Done.

rbtcollins (approved):

PR #1885, opened 2019-06-02
Bug 1870 - UI edition

rbtcollins:
Bug 1870's fix didn't show users what was going on and could stall for up to 30 seconds while waiting. This will show info: messages when that happens; perhaps a lot of them, but better on balance I think than absolutely nothing. Further refinement is of course possible but I think this is the minimum needed to make more releases sensible to do.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -36,6 +36,7 @@ use crate::markdown::md;
 use crate::term2;
 use rustup::dist::dist;
 use rustup::utils::utils;
+use rustup::utils::Notification;
 use rustup::{DUP_TOOLS, TOOLS};
 use same_file::Handle;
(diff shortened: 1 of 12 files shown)


kinnison (changes requested):
Fundamentally looks good.

There's a formatting bogon and I think one missed spot where you need to qualify the `_` argument in a  closure, but otherwise I think this is likely good to go.

My only concern is how much extra spew it produces.  Is it possible for you to include a comment on the PR showing the increased verbosity, or are you stuck not able to trigger the issue yourself?

Once you've fixed the CI bogons, I'll approve.

kinnison (approved):

rbtcollins:
Indeed I have not encountered it

Issue #1854, opened 2019-05-16
Complete, or spurious `.partial` file can cause repeated failure to download


kinnison:
If a `.partial` file is actually complete, then any attempt to continue the download will fail because the byte-range request will be out-of-range.  There is then no recovery method within `rustup` instead the user ends up having to delete the `.partial` file and try again.

In #1243 this has been discussed, and there's at least one person for whom the partial file was corrupt, and one for whom it was complete, so we need to handle both of those cases.

Glamhoth:
I'm taking a look at this.

PR #1889, opened 2019-06-05
Removing .partial that causes repeated failure to download a component

Glamhoth:
#1854

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -42,6 +42,7 @@ pub fn download_to_path_with_backend(
     callback: Option<&dyn Fn(Event<'_>) -> Result<()>>,
 ) -> Result<()> {
     use std::cell::RefCell;
+    use std::fs::remove_file;
     use std::fs::OpenOptions;
     use std::io::{Read, Seek, SeekFrom, Write};
 
@@ -114,7 +115,10 @@ pub fn download_to_path_with_backend(
         Ok(())
     }()
     .map_err(|e| {
-        // TODO is there any point clearing up here? What kind of errors will leave us with an unusable partial?
+        // TODO: We currently clear up the cached download on any error, should we restrict it to a subset?
+        remove_file(path)
+            .chain_err(|| "cleaning up cached downloads")
+            .unwrap();
         e
     })
 }
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -68,26 +68,39 @@ impl<'a> DownloadCfg<'a> {
                 + ".partial",
         );
 
+        let partial_file_existed = partial_file_path.exists();
+
         let mut hasher = Sha256::new();
 
-        utils::download_file_with_resume(
+        if let Err(e) = utils::download_file_with_resume(
             &url,
             &partial_file_path,
             Some(&mut hasher),
             true,
             &|n| (self.notify_handler)(n.into()),
-        )?;
+        ) {
+            if partial_file_existed {
+                return Err(e).chain_err(|| ErrorKind::BrokenPartialFile);
+            } else {
+                return Err(e);
+            }
+        };
 
         let actual_hash = format!("{:x}", hasher.result());
 
         if hash != actual_hash {
             // Incorrect hash
-            Err(ErrorKind::ChecksumFailed {
-                url: url.to_string(),
-                expected: hash.to_string(),
-                calculated: actual_hash,
+            if partial_file_existed {
+                self.clean(&[hash.to_string() + &".partial".to_string()])?;
+                Err(ErrorKind::BrokenPartialFile.into())
+            } else {
+                Err(ErrorKind::ChecksumFailed {
+                    url: url.to_string(),
+                    expected: hash.to_string(),
+                    calculated: actual_hash,
+                }
+                .into())
             }
-            .into())
         } else {
             (self.notify_handler)(Notification::ChecksumValid(&url.to_string()));
 
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -362,6 +362,9 @@ error_chain! {
             description("invalid PGP key"),
             display("unable to read the PGP key '{}'", v.display())
         }
+        BrokenPartialFile {
+            description("partially downloaded file may have been damaged and was removed, please try again")
+        }
     }
 }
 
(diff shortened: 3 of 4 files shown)


rbtcollins (commented):
There is another identical code block later in the same file. I think looking for all the hash mismatch locations and fixing at once is preferrable to ongoing issues.

kinnison:
@Glamhoth Are you in a position to respond to Robert's comment?

Glamhoth:
Sorry for the delay. I pushed a commit that uses existing `clean` function. As for hash mismatches I'm not sure what exactly should I be looking for.

src/dist/download.rs
```diff
@@ -104,6 +109,10 @@ impl<'a> DownloadCfg<'a> {
             if self.download_dir.join(&used_file).exists() {
                 fs::remove_file(used_file).chain_err(|| "cleaning up cached downloads")?;
             }
+            else
+            {
+                println!("{:?}", used_file);
```
rbtcollins: Whats meant to happen here?

Glamhoth: That's a debug print I commited by a mistake, I got rid of it now.

download/src/lib.rs
```diff
@@ -115,6 +116,10 @@ pub fn download_to_path_with_backend(
     }()
     .map_err(|e| {
         // TODO is there any point clearing up here? What kind of errors will leave us with an unusable partial?
+        // TODO There is a point, but should we clear only for some errors?
```
kinnison: This TODO message seems a little odd... Perhaps:

```rust
```

src/errors.rs
```diff
@@ -336,6 +336,9 @@ error_chain! {
             description("bad path in tar")
             display("tar path '{}' is not supported", v.display())
         }
+        BrokenPartialFile {
+            description("partial file could be broken and was deleted, please retry")
```
kinnison: Perhaps: `partially downloaded file may have been damaged and was removed, please try again`

kinnison (changes requested):
@rbtcollins Could you comment on what further change it was you referred to in your previous comment?  I think something about handling partials elsewhere?

Otherwise, from my perspective, this change looks useful, though I'd like to see a test to demonstrate the codepaths firing.

Glamhoth:
In that case there is only a test left to be written, but for now I have absolutely no idea how to prepare damaged partial file for it.

Quote from @kinnison (so it won't be lost):
> Not impossible, but not something I can think more detailed about right now.  In brief, I'd, in perhaps cli-v2, have a test case where you need to look at the generated dist files, read the .sha256 for a channel (fr.ex.) write out a partial file in the right place which will be corrupted or somesuch, and then run rustup install $channel checking for the error

kinnison:
So while I don't have the details to hand... The test case will have created a set of channel data somewhere because that's used to provide a dist server for the test to update from.  The trick will be to find that, get the SHA of the channel manifest, and then use that to create a partial file for the channel manifest (since that'd be named after the SHA I think).  I will see if I can catch you on Discord to discuss further.

c3st7n:
I raised a PR against @Glamhoth's branch that I believe adds a test case for the mentioned scenario.

Glamhoth:
I merged @c3st7n PR and rebased my branch onto upstream master to resolve a conflict.

tests/dist.rs
```diff
@@ -1984,3 +1984,68 @@ fn checks_files_hashes_before_reuse() {
...
+                         prefix,
+                         download_cfg,
+                         temp_cfg| {
+        // write a corrupt partial out
+        let path = url.to_file_path().unwrap();
+        let target_hash = utils::read_file(
+            "target hash",
+            &path.join("dist/2016-02-02/rustc-nightly-x86_64-apple-darwin.tar.gz.sha256"),
+        )
+        .unwrap()[..64]
```
kinnison: This is unusual, I think I'd prefer:

```suggestion
        .unwrap().trim()
```

Assuming I've correctly understood the purpose of that `[..64]`

tests/dist.rs
```diff
@@ -1984,3 +1984,68 @@ fn checks_files_hashes_before_reuse() {
...
+            &[],
+            download_cfg,
+            temp_cfg,
+            false,
+        )
+        .unwrap_err();
+
+        match *err.kind() {
+            ErrorKind::ComponentDownloadFailed(_) => (),
+            _ => panic!(),
```
kinnison: This could at least be:
```suggestion
            k @ _ => panic!("Unexpected error: {}", k),
```

kinnison (changes requested):
The addition of the test is excellent.  I've put a couple of comments in-line.  If you could please
address those, and then rebase your branch so it doesn't carry fixup commits, this looks very close to mergeable.  Once you've done that I'll try and test against the various issues we have raised which might be mitigated by this, before I merge.

kinnison:
Also, welcome back @Glamhoth :-D

tests/dist.rs
tesuji:
```suggestion
            e => panic!("{:?}", e),
```

c3st7n: This was based on how it is done elsewhere in the same file, happy to change but wonder what you think about keeping it consistent with existing tests. Same goes for the other comment regarding use of `[..64]` instead of `.trim()`.

kinnison: I think I'd rather see newer tests looking nicer than the older ones.  A separate change to make the tests more obvious throughout could be done, though it doesn't belong here.

tests/dist.rs
c3st7n: Turns out the hash file has more just the sha in it:
`66d22b7dea5bf8af3295ef13a2ac71abe85e9ad2616eeb22623ca85e6f613ccf *rustc-nightly-x86_64-apple-darwin.tar.gz`

We could split the string on whitespace but at the same time we know it is a fixed 64 character string so I'm tempted to leave it as is?

Also, in the actual code where it reads a hash file (src/dist/download.rs - fn download_hash) it just uses a slice too.

c3st7n: Raised MR to glamhoth (https://github.com/Glamhoth/rustup.rs/pull/3) to use a contstant for the slice.

Glamhoth:
Merged https://github.com/Glamhoth/rustup.rs/pull/3 and rebased fixup commits

kinnison (approved):
This looks good to me.  Thank you all for your efforts.

wang384670111:
Because the server is full of memory, I solve this problem by cleaning up some temporary files to free up some memory.

PR #1890, opened 2019-06-05
Fix clippy warnings

tesuji:
**Note**: Please squash these commits before merge.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -95,7 +95,7 @@ impl Cfg {
         );
         let dist_root = dist_root_server.clone() + "/dist";
 
-        let cfg = Cfg {
+        let cfg = Self {
             rustup_dir,
             settings_file,
             toolchains_dir,
--- a/src/diskio/threaded.rs
+++ b/src/diskio/threaded.rs
@@ -89,7 +89,7 @@ impl<'a> Threaded<'a> {
 }
 
 impl<'a> Executor for Threaded<'a> {
-    fn dispatch(&mut self, item: Item) -> Box<dyn '_ + Iterator<Item = Item>> {
+    fn dispatch(&mut self, item: Item) -> Box<dyn Iterator<Item = Item> + '_> {
         // Yield any completed work before accepting new work - keep memory
         // pressure under control
         // - return an iterator that runs until we can submit and then submits
@@ -100,7 +100,7 @@ impl<'a> Executor for Threaded<'a> {
         })
     }
 
-    fn join(&mut self) -> Box<dyn '_ + Iterator<Item = Item>> {
+    fn join(&mut self) -> Box<dyn Iterator<Item = Item> + '_> {
         // Some explanation is in order. Even though the tar we are reading from (if
         // any) will have had its FileWithProgress download tracking
         // completed before we hit drop, that is not true if we are unwinding due to a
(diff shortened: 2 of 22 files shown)


rbtcollins:
I have  a large branch that will conflict, I think it will be easier for you to rebase and eject a couple of your clippy fixes than the other way around - please hold off on merging a couple days.

kinnison:
@lzutao Could you please rebase, and resubmit?

tesuji:
Sorry for the delay! I pushed new changes.

src/config.rs
```diff
@@ -95,7 +95,7 @@ impl Cfg {
         );
         let dist_root = dist_root_server.clone() + "/dist";
 
-        let cfg = Cfg {
+        let cfg = Self {
```
kinnison, line 98: Is this a new lint I was previously unaware of?

src/diskio/threaded.rs
```diff
@@ -115,16 +115,18 @@ impl<'a> Executor for Threaded<'a> {
...
-        self.notify_handler
-            .map(|handler| handler(Notification::DownloadFinished));
-        self.notify_handler
-            .map(|handler| handler(Notification::DownloadPushUnits("iops")));
+        if let Some(handler) = self.notify_handler {
+            handler(Notification::DownloadFinished);
+        }
+        if let Some(handler) = self.notify_handler {
+            handler(Notification::DownloadPushUnits("iops"));
+        }
```
kinnison: Why two if statements rather than one with the two handler calls in it?

src/diskio/threaded.rs
```diff
@@ -115,16 +115,18 @@ impl<'a> Executor for Threaded<'a> {
...
         let mut prev_files = self.n_files.load(Ordering::Relaxed);
-        self.notify_handler.map(|handler| {
+        if let Some(handler) = self.notify_handler {
             handler(Notification::DownloadContentLengthReceived(
                 prev_files as u64,
-            ))
-        });
+            ));
+        }
```
kinnison, line 125: In fact, move the `let mut prev_files...` above the `if let` above and it can be one `if` with three calls in it perhaps?

src/diskio/threaded.rs
```diff
@@ -139,14 +141,17 @@ impl<'a> Executor for Threaded<'a> {
...
-        self.notify_handler
-            .map(|handler| handler(Notification::DownloadFinished));
-        self.notify_handler
-            .map(|handler| handler(Notification::DownloadPopUnits));
+        if let Some(handler) = self.notify_handler {
+            handler(Notification::DownloadFinished);
+        }
+        if let Some(handler) = self.notify_handler {
+            handler(Notification::DownloadPopUnits);
+        }
```
kinnison, line 148: Ditto, why not one `if` statement rather than two?

kinnison (changes requested):
Mostly looks sane.  I only commented on the first `Typename -> Self` conversion, but I'd like to know which lint is making that desirable because to me it's less readable in some cases, though better in others.

A suggestion is present for some of your refactoring, otherwise looks good.

src/config.rs
tesuji: I just enabled [the lint](https://rust-lang.github.io/rust-clippy/master/index.html#use_self) in this PR.

kinnison: Aah I see.  This one is a *pedantic* lint and I'm not sure I completely care for it.  I won't say "no" this time, but it feels a bit overly pedantic for my liking.  Can make it harder to tell what something is if it's overused IMO.

tesuji:
Honestly, I don't know how `handler` and code around that work
so I am afraid of side effect that `handler` could cause. But if you
are OK with this, I change it now.

kinnison:
The changes look good, but the build failed.  And differently this time -- the `dist` code simply stopped by the looks of it.

tesuji:
The `dist` test is passed on my [Azure build](https://dev.azure.com/taolzu/rustup.rs/_build/results?buildId=335&view=logs&jobId=f1715f20-0e3b-5a41-557a-4a8f867c6c58&taskId=fdc67eb1-e0b2-5604-0801-0cd1f81bb2cc&lineStart=1037&lineEnd=1038&colStart=1&colEnd=1).

tesuji:
Oops! You forgot to squash this PR before merging.

kinnison:
> Oops! You forgot to squash this PR before merging.

I didn't really see a need to squash it.  Each commit was clear what it was doing.

PR #1896, opened 2019-06-12
A variety of fixes, and detection of snappy et al.

kinnison:
In order to support running `rustup` from within snaps and other package managed environments, I wrote some patches.  Along the way I fixed some bugs too.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1351,11 +1351,26 @@ fn do_remove_from_path(methods: &[PathUpdateMethod]) -> Result<()> {
 /// rustup-init is stored in `CARGO_HOME`/bin, and then deleted next
 /// time rustup runs.
 pub fn update() -> Result<()> {
-    if NEVER_SELF_UPDATE {
-        err!("self-update is disabled for this build of rustup");
-        err!("you should probably use your system package manager to update rustup");
-        process::exit(1);
+    use common::SelfUpdatePermission::*;
+    let update_permitted = if NEVER_SELF_UPDATE {
+        HardFail
+    } else {
+        common::self_update_permitted(true)?
+    };
+    match update_permitted {
+        HardFail => {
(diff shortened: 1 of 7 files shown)


src/cli/self_update.rs
```diff
@@ -1351,11 +1351,21 @@ fn do_remove_from_path(methods: &[PathUpdateMethod]) -> Result<()> {
 /// rustup-init is stored in `CARGO_HOME`/bin, and then deleted next
 /// time rustup runs.
 pub fn update() -> Result<()> {
-    if NEVER_SELF_UPDATE {
-        err!("self-update is disabled for this build of rustup");
-        err!("you should probably use your system package manager to update rustup");
-        process::exit(1);
```
nrc, line 1357: I think we should still check for this - it is the mechanism distros use.

src/config.rs
```diff
@@ -342,6 +342,10 @@ impl Cfg {
         )
     }
 
+    pub fn get_default_opt(&self) -> Result<Option<String>> {
+        self.settings_file.with(|s| Ok(s.default_toolchain.clone()))
+    }
+
```
nrc, line 348: Is `get_default` used a lot? I'd prefer to replace it rather than add another method

nrc (commented):
Looks good! A couple of minor possible improvements, please feel free to merge with or without changes

src/config.rs
kinnison: Is not used in many places, I'll shuffle that around.  Good catch.

src/cli/self_update.rs
kinnison: I can't believe I lost that check, again good catch, I'll fix that soon.

kinnison:
The appveyor failure appears to be spurious, but I'm going to try a small fix for it anyway since this PR is about fixes :D

kinnison:
Well that didn't help

PR #1897, opened 2019-06-12
Windows on Travis

kinnison:
This alters the travis configuration to add Windows MSVC builds, and also to adjust which images we build on which branches.  The intent is as follows:

PRs, master, stable:

* x86_64 linux gnu
* armv7 linux gnueabihf
* aarch64 linux android
* windows msvc
* Mac OS x86_64
* CentOS 6 smoke test

master, stable:

The above, plus:

* x86_64 linux musl
* aarch64 linux gnu
* powerpc64 linux gnu

Everything else only on stable (and staging branches)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ matrix:
 
     - &windows
       os: windows
-      env: TARGET=x86_64-pc-windows-msvc NO_ADD=1
+      env: TARGET=x86_64-pc-windows-msvc NO_ADD=1 EXE_EXT=.exe
       services: nil
       language: bash
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,35 +19,45 @@ matrix:
   include:
     # Linux builds use the `rust-slave-dist` image so we link them against a
     # "super old glibc" to ensure that it runs on as many platforms as possible.
-    - os: linux
+    # These builds always run.  PRs, master, staging, stable, etc.
+    - &linuxalways
+      os: linux
       env: TARGET=x86_64-unknown-linux-gnu NO_ADD=1
-    - &linux
+    # Most builds consuming this only run on the stable and staging branches
+    - &linuxstable
       os: linux
       env: TARGET=i686-unknown-linux-gnu
       if: branch != master
+    # These builds run on non-pull-requests, so master, staging, stable...
+    - &linuxmaster
+      os: linux
+      env: SKIP_TESTS=1 TARGET=aarch64-unknown-linux-gnu
+      if: type != pull_request
 
     # Cross builds happen in the `rust-slave-linux-cross` image to ensure that
     # we use the right cross compilers for these targets. That image should
     # bundle all the gcc cross compilers to enable us to build OpenSSL
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=arm-unknown-linux-gnueabi }
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=arm-unknown-linux-gnueabihf }
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=armv7-unknown-linux-gnueabihf }
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=aarch64-unknown-linux-gnu }
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=x86_64-unknown-freebsd }
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=powerpc-unknown-linux-gnu }
-    - { <<: *linux, env: SKIP_TESTS=1 TARGET=powerpc64-unknown-linux-gnu }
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -84,24 +84,24 @@ install:
...
+    fi
 
   # Check the formatting last because test failures are more interesting to have
   # discovered for contributors lacking some platform access for testing beforehand
   - if [ "${TARGET}" = x86_64-unknown-linux-gnu ]; then
-      shellcheck -s dash -e SC1090 -- rustup-init.sh ci/*.sh;
-      rustup component add rustfmt;
-      rustfmt -vV;
-      cargo fmt --all -- --check;
+    shellcheck -s dash -e SC1090 -- rustup-init.sh ci/*.sh;
```
tesuji: Are those reflow necessary?  IMO it is harder to see after this change.

ci/prepare-deploy-travis.sh
```diff
@@ -22,10 +22,10 @@ dest="deploy"
...
 bindest="$dest/dist/$TARGET"
 mkdir -p "$bindest/"
-cp target/"$TARGET"/release/rustup-init "$bindest/"
-cp target/"$TARGET"/release/rustup-init.sha256 "$bindest/"
-cp target/"$TARGET"/release/rustup-setup "$bindest/"
-cp target/"$TARGET"/release/rustup-setup.sha256 "$bindest/"
+cp target/"$TARGET"/release/rustup-init${EXE_EXT} "$bindest/"
+cp target/"$TARGET"/release/rustup-init${EXE_EXT}.sha256 "$bindest/"
+cp target/"$TARGET"/release/rustup-setup${EXE_EXT} "$bindest/"
+cp target/"$TARGET"/release/rustup-setup${EXE_EXT}.sha256 "$bindest/"
```
tesuji: Shellcheck should warn about this. You need to double quote those.

.travis.yml
```diff
@@ -57,51 +67,68 @@ matrix:
...
       deploy:
 
+    - &windows
+      os: windows
+      env: TARGET=x86_64-pc-windows-msvc NO_ADD=1 EXE_EXT=.exe
+      services: nil
+      language: bash
+      before_install:
+        - choco install strawberryperl
+        - export PATH="/c/Strawberry/perl/bin:$PATH"
```
tesuji: Did you install `strawberryperl` because of failing to build openssl?
On Windows we don't need to build with openssl feature. You could 
detect OS in `ci/run.sh`  and run appropriate build command .

.travis.yml
```diff
@@ -57,51 +67,68 @@ matrix:
...
+
```
tesuji, line 96: Also we might want to install sccache on Windows as well. But let's do it
as follow up PR.

.travis.yml
tesuji: Use `TRAVIS_OS_NAME` env: https://docs.travis-ci.com/user/environment-variables#default-environment-variables

.travis.yml
kinnison: This one is bad, I'll fix it.

ci/prepare-deploy-travis.sh
kinnison: Yep, I'll sort that out.

.travis.yml
kinnison: We install strawberryperl because openssl can't configure with the git-bash perl, it's too broken - why would we not build with openssl on windows?

.travis.yml
kinnison: Yeah, sccache is definitely a some-future time, though I am talking to Travis today about how to improve things for Rust people on Travis, maybe I'll suggest they offer an sccache addon.

.travis.yml
tesuji: Because Windows users want to use Windows builtin TLS library. Also
in Appveyor build file we don't use OpenSSL feature either.

kinnison: Oh?  I must have missed that - what feature do I pass to enable that, because if I pass no feature flags, it still tries to build openssl.

tesuji: The line 11 use openssl feature:
https://github.com/rust-lang/rustup.rs/blob/04039b0d13a8f5fba70dd55a9dcca0a52a41f0ad/ci/run.sh#L8-L17

You could use `TRAVIS_OS_NAME` env variable and get rid of that feature 
on Windows.

kinnison: Mmm, when I was trying to get it to work without strawberryperl I tried removing the features there but it still seemed to be building openssl anyway.  I'll try again I guess.

kinnison: I've pushed something which hopefully does that, though it ends up abusing shellcheck a bit.  We might be able to refactor it more nicely in the future

.travis.yml
kinnison: I've spoken with them and they're going to have a think, they may even contact you on Telegram.

nrc (approved):
lgtm

PR #1898, opened 2019-06-16
Snappy: Add support for snapcraft builds

kinnison:
This commit enables building snaps for the various platforms supported by snapcraft.io

Currently the snap is under my personal page, and despite building it here, we'll keep that until after all-hands where I'll have a chat with t-release people about what we might do for an official snap.  So far I've had no joy looking at official methods of distributing rustup in other channels such as flatpak.

### DIFF
--- a/.github/workflows/linux-builds-on-master.yaml
+++ b/.github/workflows/linux-builds-on-master.yaml
@@ -27,6 +27,11 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
+            snap_arch: amd64
+          - target: aarch64-unknown-linux-gnu # skip-pr
+            snap_arch: arm64 # skip-pr
+          - target: armv7-unknown-linux-gnueabihf
+            snap_arch: armhf
     steps:
       - uses: actions/checkout@v2
         with:
@@ -134,6 +139,25 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         if: github.event_name == 'push' && github.ref == 'refs/heads/stable'
+      - name: Build a snapcraft configuration file
+        run: |
+          bash ci/snapcraft.bash
+        if: matrix.snap_arch != ''
+        env:
+          DO_SNAP: 1
+          SNAP_ARCH: ${{ matrix.snap_arch }}
+          SNAP_EDGE: 1 # skip-stable
+      - uses: snapcore/action-build@v1
+        id: snapcraftbuild
+        with:
+          path: snapcraft
+        if: matrix.snap_arch != ''
+      - uses: snapcore/action-publish@v1
+        with:
+          store_login: ${{ secrets.SNAPCRAFT_TOKEN }}
+          snap: ${{ steps.snapcraftbuild.outputs.snap }}
+          release: edge # skip-pr skip-stable
+        if: matrix.snap_arch != '' # skip-pr
       - name: Clear the cargo caches
         run: |
           cargo install cargo-cache --no-default-features --features ci-autoclean
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -49,6 +49,17 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
+            snap_arch: amd64
+          - target: i686-unknown-linux-gnu # skip-pr skip-master
+            snap_arch: i386  # skip-pr skip-master
+          - target: aarch64-unknown-linux-gnu # skip-pr
+            snap_arch: arm64 # skip-pr
+          - target: armv7-unknown-linux-gnueabihf
+            snap_arch: armhf
+          - target: powerpc64le-unknown-linux-gnu # skip-pr skip-master
+            snap_arch: ppc64el # skip-pr skip-master
+          - target: s390x-unknown-linux-gnu # skip-pr skip-master
+            snap_arch: s390x # skip-pr skip-master
     steps:
       - uses: actions/checkout@v2
         with:
@@ -156,6 +167,27 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         if: github.event_name == 'push' && github.ref == 'refs/heads/stable'
+      - name: Build a snapcraft configuration file
+        run: |
+          bash ci/snapcraft.bash
+        if: matrix.snap_arch != ''
+        env:
+          DO_SNAP: 1
+          SNAP_ARCH: ${{ matrix.snap_arch }}
+          SNAP_EDGE: 1 # skip-stable
+      - uses: snapcore/action-build@v1
+        id: snapcraftbuild
+        with:
+          path: snapcraft
+        if: matrix.snap_arch != ''
+      - uses: snapcore/action-publish@v1
+        with:
+          store_login: ${{ secrets.SNAPCRAFT_TOKEN }}
+          snap: ${{ steps.snapcraftbuild.outputs.snap }}
+          release: edge # skip-pr skip-stable
+          release: beta # skip-pr skip-master
+        if: matrix.snap_arch != '' # skip-pr
+        if: matrix.snap_arch == 'not-for-prs' # skip-master skip-stable
       - name: Clear the cargo caches
         run: |
           cargo install cargo-cache --no-default-features --features ci-autoclean
--- a/.github/workflows/linux-builds-on-pr.yaml
+++ b/.github/workflows/linux-builds-on-pr.yaml
@@ -22,6 +22,9 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
+            snap_arch: amd64
+          - target: armv7-unknown-linux-gnueabihf
+            snap_arch: armhf
     steps:
       - uses: actions/checkout@v2
         with:
@@ -129,6 +132,24 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         if: github.event_name == 'push' && github.ref == 'refs/heads/stable'
+      - name: Build a snapcraft configuration file
+        run: |
+          bash ci/snapcraft.bash
+        if: matrix.snap_arch != ''
+        env:
+          DO_SNAP: 1
+          SNAP_ARCH: ${{ matrix.snap_arch }}
+          SNAP_EDGE: 1 # skip-stable
+      - uses: snapcore/action-build@v1
+        id: snapcraftbuild
+        with:
+          path: snapcraft
+        if: matrix.snap_arch != ''
+      - uses: snapcore/action-publish@v1
+        with:
+          store_login: ${{ secrets.SNAPCRAFT_TOKEN }}
+          snap: ${{ steps.snapcraftbuild.outputs.snap }}
+        if: matrix.snap_arch == 'not-for-prs' # skip-master skip-stable
       - name: Clear the cargo caches
         run: |
           cargo install cargo-cache --no-default-features --features ci-autoclean
--- a/.github/workflows/linux-builds-on-stable.yaml
+++ b/.github/workflows/linux-builds-on-stable.yaml
@@ -41,6 +41,17 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
+            snap_arch: amd64
+          - target: i686-unknown-linux-gnu # skip-pr skip-master
+            snap_arch: i386  # skip-pr skip-master
+          - target: aarch64-unknown-linux-gnu # skip-pr
+            snap_arch: arm64 # skip-pr
+          - target: armv7-unknown-linux-gnueabihf
+            snap_arch: armhf
+          - target: powerpc64le-unknown-linux-gnu # skip-pr skip-master
+            snap_arch: ppc64el # skip-pr skip-master
+          - target: s390x-unknown-linux-gnu # skip-pr skip-master
+            snap_arch: s390x # skip-pr skip-master
     steps:
       - uses: actions/checkout@v2
         with:
@@ -148,6 +159,24 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         if: github.event_name == 'push' && github.ref == 'refs/heads/stable'
+      - name: Build a snapcraft configuration file
+        run: |
+          bash ci/snapcraft.bash
+        if: matrix.snap_arch != ''
+        env:
+          DO_SNAP: 1
+          SNAP_ARCH: ${{ matrix.snap_arch }}
+      - uses: snapcore/action-build@v1
+        id: snapcraftbuild
+        with:
+          path: snapcraft
+        if: matrix.snap_arch != ''
+      - uses: snapcore/action-publish@v1
+        with:
+          store_login: ${{ secrets.SNAPCRAFT_TOKEN }}
+          snap: ${{ steps.snapcraftbuild.outputs.snap }}
+          release: beta # skip-pr skip-master
+        if: matrix.snap_arch != '' # skip-pr
       - name: Clear the cargo caches
         run: |
           cargo install cargo-cache --no-default-features --features ci-autoclean
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /**/target
 /home
 /local-rustup
+/snapcraft
--- a/ci/snapcraft.bash
+++ b/ci/snapcraft.bash
@@ -1,0 +1,125 @@
+#!/bin/bash
+
+set -efu
+
+set -o pipefail
+
+# This script encapsulates the job of preparing snapcract configuration for the
+# rustup snap.  We construct either an edge snap (builds on PRs or master)
+# or a beta snap if built from a non-master branch.
+
+# If built on master, we should publish the edge snap
+# If built on stable, we should publish the beta snap
+
+# To parameterise the build, the following environment variables are needed:
+# DO_SNAP=1  <-- without this, we don't run
+# SNAP_EDGE=1 <-- with this, we build an edge/devel snap, otherwise a release snap
+# SNAP_ARCH=... <-- without this we cannot build, this is the snap architecture we're building
+
+if [ -z "${DO_SNAP:-}" ]; then
+  echo "SNAP: Not attempting to snap"
+  exit 0
+fi
+
+if [ -z "${SNAP_ARCH:-}" ]; then
+  echo "SNAP: Unable to generate snap, architecture unset"
+  exit 1
+else
+  echo "SNAP: Constructing snap for architecture $SNAP_ARCH"
+fi
+
+if [ -n "${SNAP_EDGE:-}" ]; then
+  echo "SNAP: Generated snap will be an grade:devel snap"
+  GRADE=devel
+else
+  echo "SNAP: Generated snap will be a grade:stable snap"
+  GRADE=stable
+fi
+
+
+# This is the list of proxies which the snap will contain.  This needs to match
+# the set of aliases which the snap has been allocated.  If we add to this
+# list but don't get the aliases added, then people won't see the proxy except
+# as rustup.$PROXY.  If we have an alias which is not a supported proxy name
+# then rustup might get sad.
+
+PROXIES="cargo cargo-clippy cargo-fmt cargo-miri clippy-driver rls rustc rustdoc rustfmt"
+
+# From now on, things should be automagic
+
+VERSION=$(grep "^version" Cargo.toml | head -1 | cut -d\" -f2)
+
+rm -rf snapcraft
+
+mkdir -p snapcraft/snap
+
+cd snapcraft
+
+cat > snap/snapcraft.yaml <<SNAP
+name: rustup
+version: $VERSION
+confinement: classic
+base: core18
+architectures:
+  - build-on: [amd64]
+    run-on: [$SNAP_ARCH]
+description: |
+  Rustup is the Rust Language's installer and primary front-end.  You probably
+  want this if you want to develop anything written in Rust.
+SNAP
+
+if [ -n "${SNAP_EDGE:-1}" ]; then
+  cat >> snap/snapcraft.yaml <<SNAP
+
+  Please note, this snap is experimental and functionality cannot be guaranteed
+  to be consistent with released snaps.
+summary: "EXPERIMENTAL: The Rust Language Installer"
+grade: $GRADE
+SNAP
+else
+  cat >> snap/snapcraft.yaml <<SNAP
+summary: "The Rust Language Installer"
+grade: $GRADE
+SNAP
+fi
+
+cat >> snap/snapcraft.yaml <<SNAP
+parts:
+  rustup:
+    plugin: nil
+    build-attributes: [no-patchelf]
+    source: inputs
+    override-build: |
+      mkdir -p \$SNAPCRAFT_PART_INSTALL/bin
+      cp rustup-init \$SNAPCRAFT_PART_INSTALL/bin/rustup
+SNAP
+
+for PROXY in $PROXIES; do
+  cat >> snap/snapcraft.yaml <<SNAP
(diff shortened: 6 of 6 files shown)


.travis.yml
```diff
@@ -132,6 +147,14 @@ script:
 
 before_deploy:
   - sh ci/prepare-deploy-travis.sh
+  - |
+    if [ -n "$DO_SNAP" ]; then
+      if ! [ "$TRAVIS_BRANCH" = "stable" ]; then
```
tesuji: Use `if [ "$TRAVIS_BRANCH" != "stable" ]`

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+
+PROXIES="cargo cargo-clippy cargo-fmt cargo-miri clippy-driver rls rustc rustdoc rustfmt"
+
+# From now on, things should be automagic
+
+VERSION=$(grep "^version" Cargo.toml | head -1 | cut -d\" -f2)
+
+rm -rf snap
+
+mkdir -p snap/snap
```
tesuji: I would prefer to rename this to `mkdir -p snapbuild/snap` to avoid repeatedly name.

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+if ! [ -n "$SNAP_ARCH" ]; then
+  echo "SNAP: Unable to generate snap, architecture unset"
+  exit 1
+else
+  echo "SNAP: Constructing snap for architecture $SNAP_ARCH"
+fi
+
+if [ -n "$SNAP_EDGE" ]; then
+  echo "SNAP: Generated snap will be an grade:devel snap"
+  GRADE=devel
```
tesuji: This variable appears to be unused.

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+
+for PROXY in $PROXIES; do
+  cat >> snap/snapcraft.yaml <<SNAP
+  $PROXY:
+    command: bin/$PROXY
+SNAP
+done
+
+mkdir inputs
+cp ../target/$TARGET/release/rustup-init inputs/
```
tesuji: Double quote `"$TARGET"` to avoid word splitting.

tesuji (commented):
Honestly, I may know less about snapcraft than you. I would prefer to 
make `ci/snapcraft.sh` a bash script with `.bash` extension instead to
make use of bash features - array, also because this script is not intended
to used by other parties (just in our CI).

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+fi
+
+
+# This is the list of proxies which the snap will contain.  This needs to match
+# the set of aliases which the snap has been allocated.  If we add to this
+# list but don't get the aliases added, then people won't see the proxy except
+# as rustup.$PROXY.  If we have an alias which is not a supported proxy name
+# then rustup might get sad.
+
+PROXIES="cargo cargo-clippy cargo-fmt cargo-miri clippy-driver rls rustc rustdoc rustfmt"
```
tesuji: We may want to use bash feature - array instead.

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+# DO_SNAP=1  <-- without this, we don't run
+# SNAP_EDGE=1 <-- with this, we build an edge/devel snap, otherwise a release snap
+# SNAP_ARCH=... <-- without this we cannot built, this is the snap architecture we're building
+
+if ! [ -n "$DO_SNAP" ]; then
+  echo "SNAP: Not attempting to snap"
+  exit 0
+fi
+
+if ! hash snapcraft >/dev/null 2>/dev/null; then
```
tesuji: Try `! commmand -v snapcraft > /dev/null`.

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+  echo "SNAP: Unable to find snapcraft, cannot run"
+  exit 1
+fi
+
+if ! [ -n "$SNAP_ARCH" ]; then
```
tesuji: Try `[ -z ... ]` instead.

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# If built on master, we will publish the edge snap
+# If built on stable, we will publish the beta snap
+
+# To parameterise the build, the following environment variables are needed:
+# DO_SNAP=1  <-- without this, we don't run
+# SNAP_EDGE=1 <-- with this, we build an edge/devel snap, otherwise a release snap
+# SNAP_ARCH=... <-- without this we cannot built, this is the snap architecture we're building
+
+if ! [ -n "$DO_SNAP" ]; then
```
tesuji: This too.

.travis.yml
kinnison: Doh!  Done.

ci/snapcraft.sh
kinnison: I think I picked `snap/snap` because of how snapcraft uses `snap` but of course it's not necessary at the top level, so I'll fix that.

ci/snapcraft.sh
kinnison: Yep, well spotted.  I've fixed that.

ci/snapcraft.sh
kinnison: Done.  Thanks

ci/snapcraft.sh
kinnison: I'm not a shell person.  I'm happy if you can tell me what to do there, otherwise this can stay until you want to fix it later

ci/snapcraft.sh
kinnison: Fair.  Done.

ci/snapcraft.sh
kinnison: I have no words for why I didn't use `-z`  Fixed.

ci/snapcraft.sh
kinnison: Ditto.

kinnison:
@lzutao Thank you for the comprehensive review.  I believe I've tackled everything you hilighted (except the one thing I don't feel confident in) -- could you recheck please?

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+# DO_SNAP=1  <-- without this, we don't run
+# SNAP_EDGE=1 <-- with this, we build an edge/devel snap, otherwise a release snap
+# SNAP_ARCH=... <-- without this we cannot built, this is the snap architecture we're building
+
+if [ -z "$DO_SNAP" ]; then
+  echo "SNAP: Not attempting to snap"
+  exit 0
+fi
+
+if ! command -v snapcraft >/dev/null 2>/dev/null; then
```
tesuji: No need for `2>/dev/null` part.

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+# the set of aliases which the snap has been allocated.  If we add to this
+# list but don't get the aliases added, then people won't see the proxy except
+# as rustup.$PROXY.  If we have an alias which is not a supported proxy name
+# then rustup might get sad.
+
+PROXIES="cargo cargo-clippy cargo-fmt cargo-miri clippy-driver rls rustc rustdoc rustfmt"
+
+# From now on, things should be automagic
+
+VERSION=$(grep "^version" Cargo.toml | head -1 | cut -d\" -f2)
```
tesuji: Try `head -n 3 Cargo.toml | grep ^version | cut -d\" -f2`

ci/snapcraft.sh
```diff
@@ -0,0 +1,135 @@
...
+
+  Please note, this snap is experimental and functionality cannot be guaranteed
+  to be consistent with released snaps.
+summary: "EXPERIMENTAL: The Rust Language Installer"
+grade: $GRADE
+SNAP
+else
+  cat >> snap/snapcraft.yaml <<SNAP
+summary: "The Rust Language Installer"
+grade: stable
```
tesuji: Do you want to use `$GRADE` here too?

ci/snapcraft.sh
tesuji: Basic uses of bash array if you want to try: https://devhints.io/bash#arrays

ci/snapcraft.sh
kinnison: Okay, dropped.

ci/snapcraft.sh
kinnison: I'm not sure I approve of this, in case the version string moves in the header.  Really I ought to be parsing the `.toml` but I think this is moderately more sensible than the `head` approach.  Why are you advocating that instead?

ci/snapcraft.sh
kinnison: Yes, well spotted.  I was clearly asleep when making the updates before.

ci/snapcraft.sh
tesuji: It's just a suggestion. You could skip it.

tesuji:
r=me with or with out using bash array, unless others disagree.

ci/snapcraft.sh
kinnison: Okay, then for now I think I will, thank you.

kinnison:
Thank you @lzutao 

We won't be merging this until I've spoken to the release team about snaps, but it's good to know you're now satisfied with my awful shell code :-D

rbtcollins (approved):
Sure, why not.

ghost (approved):

kinnison:
Now that we're a bit further on in our releases, and Windows builds are no longer *always* failing; I'll give this some more thought.

kinnison:
I've started a discussion with Snap people about github actions for this kind of thing.  Depending on how that goes, I'll either rework this PR, or close it and open a fresh one for continued discussion around `rustup` and snapcraft.

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -152,6 +163,15 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         if: github.event_name == 'push' && github.ref == 'refs/heads/stable'
+      - name: Build a snapcraft configuration file
+        run: |
+          export DO_SNAP=1
+          export SNAP_ARCH=${{ matrix.snap_arch }}
+          export SNAP_EDGE=1 # skip-stable
+          sh ci/snapcraft.sh
+        if: matrix.snap_arch != ''
```
tesuji, line 173:
```suggestion
        if: matrix.snap_arch
```
here and there

kinnison: I prefer the explicit comparison since it's meant to be a string if it's set.

kinnison:
I now need to wait for an appropriate publish action to be written and we can proceed further.

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -152,6 +163,18 @@ jobs:
...
         if: github.event_name == 'push' && github.ref == 'refs/heads/stable'
+      - name: Build a snapcraft configuration file
+        run: |
+          sh ci/snapcraft.sh
+        if: matrix.snap_arch != ''
+        env:
+          DO_SNAP: 1
+          SNAP_ARCH: ${{ matrix.snap_arch }}
+          SNAP_EDGE: 1 # skip-stable
+      - uses: jhenstridge/snapcraft-build-action@master
```
jhenstridge: I've published a `v1` tag for this action.  I'd suggest using that instead of `master`.

ci/snapcraft.sh
```diff
@@ -0,0 +1,130 @@
...
+     - "-rustup-init"
+  link:
+    plugin: nil
+    build-attributes: [no-patchelf]
+    after: [acquire]
+    override-pull: |
+      cp \$SNAPCRAFT_STAGE/rustup-init .
+    override-build: |
+      mkdir -p \$SNAPCRAFT_PART_INSTALL/bin
+      cp rustup-init \$SNAPCRAFT_PART_INSTALL/bin/rustup
```
jhenstridge: Would this be simpler as a single part?  Something like:

```
parts:
  rustup:
    plugin: nil
    build-attributes: [no-patchelf]
    source: inputs
    override-build: |
      mkdir $SNAPCRAFT_PART_INSTALL/bin
      cp rustup-init $SNAPCRAFT_PART_INSTALL/bin/rustup
      ...
```

I believe that should have the same effect as what you've got.

ci/snapcraft.sh
```diff
@@ -0,0 +1,130 @@
...
+SNAP
+
+for PROXY in $PROXIES; do
+  cat >> snap/snapcraft.yaml <<SNAP
+      ln \$SNAPCRAFT_PART_INSTALL/bin/rustup \$SNAPCRAFT_PART_INSTALL/bin/$PROXY
```
jhenstridge: Is there any particular reason for using hard links over symlinks here? I'm curious to know if they get preserved as hard links in the resultiing squashfs file.

If they don't get preserved, then symlinks could give a noticeable reduction in package size.

ci/snapcraft.sh
```diff
@@ -0,0 +1,130 @@
...
+SNAP
+done
+
+cat >> snap/snapcraft.yaml << 'SNAP'
+    prime:
+     - "-rustup-init"
+
+environment:
+  RUSTUP_HOME: "$SNAP_USER_DATA/rustup"
```
jhenstridge: If you are storing a lot of data in this directory, it might be worth switching this to `$SNAP_USER_COMMON`. The `$SNAP_USER_DATA` directory is versioned, and will be copied on an upgrade of the snap.  This is done so that configuration can be rolled back if you revert an upgrade of a snap.

If you're happy to deal with data from potentially newer versions of your snap, then `$SNAP_USER_COMMON` is a better choice.  It will save disk space and make upgrades faster.

jhenstridge (commented):
I've left a few drive-by comments after the discussion on the Snapcraft forum.  I've been working on an action to publish snaps here:

https://github.com/jhenstridge/snapcraft-publish-action

I haven't quite locked down the API yet, but it is functional.  As the action depends on a secret (the Snap Store credentials), I'm not sure how easy it will be to test within a pull request.

ci/actions-templates/linux-builds-template.yaml
kinnison: Yeah, this is actually a bogon from when I was diagnosing something in a fork of my own, thanks for the spot.

ci/snapcraft.sh
kinnison: I honestly don't know.  This generator was developed when I was at snapcraft summit in Montréal and since then I've not touched it.  Once I've something which works I'll look at cleaning it up more.

ci/snapcraft.sh
kinnison: The hardlinks are there because that's what `rustup` normally does and I wasn't interested in risking things.  I believe squashfs does hard links just fine, so unless the snapcraft settings fail to preserve them we ought to be fine.  If you're worried we can try switching to symlinks once the rest is sorted.

ci/snapcraft.sh
kinnison: That sounds very sensible.  `rustup` is pretty stable in terms of the content of `~/.rustup` so I'll switch to `$SNAP_USER_COMMON`

kinnison:
@jhenstridge Thank you for your input and review -- I'll look at the publish action you've provided and I can probably test it by adding the secret to my own repo fork and trying it there to ensure it behaves OK.  I'll try and do that soon.

kinnison:
@jhenstridge I think that ought to do the trick, obviously I'm tied to a particular commit of your repo for now.  I tested it over at my own fork and it seemed to do the right thing, so now I need to wait for you to finalise a v1 of the publish action I guess, and then decide on the final comments you made on the snapcraft yaml file.

ci/snapcraft.sh
jhenstridge: I know squashfs can handle them.  I was more curious about whether snapcraft itself was preserving them as the data is copied from the part's install dir, to the stage dir, and finally to the prime dir.

This may not even be a problem in the end, so I agree that investigating this shouldn't be a blocker for landing this PR.

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -152,6 +163,27 @@ jobs:
...
+        id: snapcraftbuild
+        with:
+          path: snapcraft
+        if: matrix.snap_arch != ''
+      - uses: jhenstridge/snapcraft-publish-action@0dac40c79fd10f8f51ee149a58a47ab3a42aeee0
+        with:
+          store_login: ${{ secrets.SNAPCRAFT_TOKEN }}
+          snap: ${{ steps.snapcraftbuild.outputs.snap }}
+          release: edge # skip-pr skip-stable
+          release: beta # skip-pr skip-master
```
jhenstridge, line 188: This should just be:

    release: edge,beta

I don't think Github actions supports repeating input parameters for actions: they're passed to the action as environment variables rather than anything more structured.

ci/snapcraft.sh
kinnison: :+1:

ci/actions-templates/linux-builds-template.yaml
kinnison: Those comments are directives to a script which generates our real workflow files, on PRs, there'll be no `release`, on master builds it'll say `edge` and on stable builds `beta`

ci/snapcraft.sh
kinnison: I can confirm this cleanup works, thanks James.

ci/snapcraft.sh
kinnison: I've checked and the snap I end up with on x86_64 linux is 2.7M which is definitely not a multiple of the binary size, so either squashfs has noticed they're all the same and compressed well, or else the hardlinks worked :D

kinnison:
I believe everything is right now, and I've added the token to the repo's secrets ready for post-merge use.  If noone objects, I'll merge this soon and we can see how that goes.

.github/workflows/linux-builds-on-master.yaml
```diff
@@ -27,6 +27,11 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
+            snap_arch: amd64
+          - target: aarch64-unknown-linux-gnu # skip-pr
```
rbtcollins, line 31: Should these stay skip-pr?

.github/workflows/linux-builds-on-master.yaml
```diff
@@ -130,6 +135,25 @@ jobs:
...
         if: github.event_name == 'push' && github.ref == 'refs/heads/stable'
+      - name: Build a snapcraft configuration file
+        run: |
+          sh ci/snapcraft.sh
+        if: matrix.snap_arch != ''
+        env:
+          DO_SNAP: 1
+          SNAP_ARCH: ${{ matrix.snap_arch }}
+          SNAP_EDGE: 1 # skip-stable
+      - uses: jhenstridge/snapcraft-build-action@v1
```
rbtcollins: This is an individual developers snap action; I love James to bits, but if he leaves or gets hit by a bus this is going to break. Does Canonical not have a snap team account for this action to live under?

ci/snapcraft.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/sh
+
+set -e
```
rbtcollins: suggest -u -f and -o pipefail as well. See https://sipb.mit.edu/doc/safe-shell/ for a longer explanation than just me saying 'its a good idea'.

rbtcollins (changes requested):
One question, one requested change

.github/workflows/linux-builds-on-master.yaml
kinnison: The intent here was PRs only test building the snap, so we check the native arch and one cross arch to make sure I've not fluffed anything.  then master builds more of them because it uploads to the edge channel, and stable uploads all of them to the beta channel.  Does that make sense to you?

.github/workflows/linux-builds-on-master.yaml
kinnison: Currently they don't.  The eventual goal for `rust-lang` itself is to mirror any actions we depend on which aren't github official ones or actions-rs I think.  For now I also trust James won't go bad, and once I know what the overall plan is, we can see about mirroring those actions or whatever is needed if Canonical don't do official actions in the snapcraft group or somesuch.

ci/snapcraft.sh
kinnison: Hmm, -u will render the checks at the top redundant or else they'll need rejiggering to cope if they're set to empty strings.  I guess -f would be harmless since there's no * or ? in the script anyway, so I'll add that to future-proof things.  This is a 'sh' so on the builders it'll end up with dash so -o pipefail isn't available according to that link.  I'll add -f and note that this passes shellcheck in case that mollifies you about the others.

.github/workflows/linux-builds-on-master.yaml
rbtcollins: So master = edge, and stable == releases == beta, and actual stable channel is manual?

.github/workflows/linux-builds-on-master.yaml
rbtcollins: Yeah, I have zero concerns about James going bad :) Could we at least ask if they are willing to make an actions repo in snapcraft/ ? Or has that been discussed and decided against?

ci/snapcraft.sh
```diff
@@ -0,0 +1,123 @@
...
+# the set of aliases which the snap has been allocated.  If we add to this
+# list but don't get the aliases added, then people won't see the proxy except
+# as rustup.$PROXY.  If we have an alias which is not a supported proxy name
+# then rustup might get sad.
+
+PROXIES="cargo cargo-clippy cargo-fmt cargo-miri clippy-driver rls rustc rustdoc rustfmt"
+
+# From now on, things should be automagic
+
+VERSION=$(grep "^version" Cargo.toml | head -1 | cut -d\" -f2)
```
rbtcollins: Its things like this that pipefail is for; you can use mispipe from moreutils if you want to stay with POSIX sh, or switch to bash shell.

ci/snapcraft.sh
```diff
@@ -0,0 +1,123 @@
...
+
+# If built on master, we should publish the edge snap
+# If built on stable, we should publish the beta snap
+
+# To parameterise the build, the following environment variables are needed:
+# DO_SNAP=1  <-- without this, we don't run
+# SNAP_EDGE=1 <-- with this, we build an edge/devel snap, otherwise a release snap
+# SNAP_ARCH=... <-- without this we cannot build, this is the snap architecture we're building
+
+if [ -z "$DO_SNAP" ]; then
```
rbtcollins: fixing this for undefined safety is easy:
```
if [ -z "${DO_SNAP:-}" ]; then
```

rbtcollins:
Is the shell script shellchecked automatically?

.github/workflows/linux-builds-on-master.yaml
kinnison: That is indeed the current plan.

.github/workflows/linux-builds-on-master.yaml
kinnison: I believe James is seeking futher proof that his actions are good, before he persues officialdom.

kinnison: I have been discussing the work he did on https://forum.snapcraft.io/t/call-for-testing-github-action-for-snapcrtaft/14930 but as yet he's not indicated an intent to carry it forward further.

kinnison: /cc @jhenstridge

kinnison:
@rbtcollins Yes, shellcheck is run over all our shell scripts as part of the CI.

ci/snapcraft.sh
kinnison: Of course.  I'm so dozy when it comes to shell :D

kinnison: Done this where appropriate, thanks Robert.

ci/snapcraft.sh
kinnison: Aha, yeah I see.  I'll switch to bash and enable pipefail.

.github/workflows/linux-builds-on-master.yaml
jhenstridge: I have brought up moving these actions to the `snapcore` github organisation, both to make it easier to share maintenance and to make updating them a bit structured.  These actions started as a holiday project to learn about this part of the GH Actions system, so it hasn't been a priority during the week.

You are right that there are security concerns about trusting an action that I solely control. The publish action in particular is in a prime position to silently exfiltrate the store credentials of any project that uses it.  The build action could similarly modify the project before building it: a little more involved and likely to be noticed in the logs, but still a possibility.

Actions provided by `snapcore` could also do all this, but as it is the same org that publishes Snapcraft, it is also in a position to do all of this.

I'll try and move forward the process of moving the projects over.

kinnison:
@rbtcollins I believe your question and change were resolved.  I've updated to use the now official snapcore actions.  I think we're likely good to merge assuming the CI remains green.  Opinions?

kinnison:
I have decided to go ahead and merge this because we need to work through any problems long before we make a release, and we can only do that once it's trying to push to snapstore etc.

PR #1901, opened 2019-06-19
Add padding to left and right of install command

keysmashes:
Since the install command is longer now (#1716), it hits the edge of the box it's in, which looks ugly. This commit adds 1rem of padding to the left and right of the text, which is the same as the padding above and below.

Before:

image

After:

image


kinnison:
LGTM.

PR #1903, opened 2019-06-20
Replace deprecated tempdir with tempfile

tesuji:


kinnison:
Fundamentally I'm good with this, though there are test failures, so please resolve those before I review deeply.

tesuji:
```
failures:
---- remove_toolchain_then_add_again stdout ----
thread 'remove_toolchain_then_add_again' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 1142, kind: Other, message: "An attempt was made to create more links on a file than the file system supports." }', src\libcore\result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I'm now convinced that something on our side has messed up.

kinnison:
Hey @lzutao Could you rebase this against current master and see how it looks?

kinnison (approved):
I *think* this is okay.  I'd like a second opinion before merge.

rbtcollins (approved):
I don't like the new API so much, I have to say.

PR #1906, opened 2019-06-21
Switch to std::fs::read_to_string

tesuji:
According to [the doc][1]:  it is generally faster than reading into a string created with `String::new()`.

[1]: https://doc.rust-lang.org/nightly/std/fs/fn.read_to_string.html

### DIFF
--- a/download/tests/download-curl-resume.rs
+++ b/download/tests/download-curl-resume.rs
@@ -8,7 +8,7 @@ use url::Url;
 use download::*;
 
 mod support;
-use crate::support::{file_contents, serve_file, tmp_dir, write_file};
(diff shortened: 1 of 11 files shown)


kinnison (approved):
While I question the wisdom of needing to improve performance on any of these pathways, I do agree that it simplifies the codebase by removing two implementations of what is essentially something from `std`

kinnison:
I *think* the test failures are windows being odd.  But I'd like @rbtcollins to confirm before merging.

rbtcollins (approved):
LGTM

rbtcollins:
The CI failures I haven't investigated, but I see no platform interactions to be concerned about.

PR #1909, opened 2019-06-23
Document (some) caveats when developing Rust on Windows

bltavares:
This commit documents errors that a developer might face when compiling Rust projects on Windows 10.

It is by far not a comprehensive list of errors and solutions, but its a start, and document the errors I've faced when trying to compile https://github.com/rust-lang/www.rust-lang.org on Windows 10, on two scenarios:

- Fresh `rustup` installation, following the instructions on screen, and installing VS Build Tools 2019
- Older installation of `rustup`, using VS Build Tools 2017

It is important to also be aware that following the instructions of installing VS Build Tools 2019 might require patches on `*-sys` crates, as their `build.rs` might not work for the new locations on the new edition of MS Build Tools. For example: https://github.com/compass-rs/sass-rs/pull/49

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,49 @@ $ rustup target add i686-pc-windows-gnu
 [GCC toolchain]: https://gcc.gnu.org/
 [MinGW/MSYS2 toolchain]: https://msys2.github.io/
 
+### VS Build Tools Caveats
+
+When selecting the Build Tools only instalation, the Rust tooling will work as
+expected, but you might face a couple of issues when compiling `*-sys` 
+dependencies.
+
+`*-sys` dependencies often link to C/C++ libraries as part of their build 
+system, and for that, a set of environment variables and paths should be setup.
+
+Selecting only the Build tools instead of a complete installation of Visual Studio
+installation might not configure everything properly. Here are some indicators
+of problems on the development environment setup and some suggestion to fix them.
+
+<details>
+ <summary><code>error MSB4019: The imported project "C:\Microsoft.Cpp.Default.props" was not found. Confirm that the path in the &lt;Import&gt; declaration is correct, and that the file exists on disk.</code></summary>
+  This error indicates that the environent is not properly setup, and you have a couple of of options to try to fix it:
+
+- [Easiest] Open the `Developer Command Prompt for VS 2019` application to compile the project. This terminal will setup all the needed environment variables to let it compile.
+- Setup the environment variables (eg: `PATH`, `LIB`) as [documented on Microsoft's website](https://docs.microsoft.com/en-us/cpp/build/setting-the-path-and-environment-variables-for-command-line-builds?view=vs-2019)
+- Install the complete setup of Visual Studio 2015 or Visual Studio 2017 - not only the C++ Build tools
+
+If the errors persists, maybe the `*-sys` dependency in question requires [a similar patch](https://github.com/compass-rs/sass-rs/commit/2e8289539fcb2b11812b666b5104d94744fa93b6) to select the proper `msbuild.exe` executable from the environment.
+</details>
+
+<details>
+ <summary><code>fatal error LNK1112: module machine type 'x86' conflicts with target machine type 'x64'</code></summary>
+ This happens when the variables used setup tools for the wrong architecture, where you have a linker for x86 instead of the target x64.
+ 
+ 
+ Possible fixes:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -474,6 +474,49 @@ $ rustup target add i686-pc-windows-gnu
...
+### VS Build Tools Caveats
+
+When selecting the Build Tools only instalation, the Rust tooling will work as
+expected, but you might face a couple of issues when compiling `*-sys` 
+dependencies.
+
+`*-sys` dependencies often link to C/C++ libraries as part of their build 
+system, and for that, a set of environment variables and paths should be setup.
+
+Selecting the Build Tools instalation instead of the complete Visual Studio
```
kinnison: Perhaps "Selecting only the _Build tools_ instead of a complete installation of Visual Studio..." instead?

README.md
```diff
@@ -474,6 +474,49 @@ $ rustup target add i686-pc-windows-gnu
...
+
+`*-sys` dependencies often link to C/C++ libraries as part of their build 
+system, and for that, a set of environment variables and paths should be setup.
+
+Selecting the Build Tools instalation instead of the complete Visual Studio 
+installation might not configure everything properly. Here are some indicators
+of problems on the development environment setup and some suggestion to fix them.
+
+<details>
+ <summary><code>error MSB4019: The imported project "C:\Microsoft.Cpp.Default.props" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.</code></summary>
```
kinnison: Does `<Import>` need escaping here?

README.md
```diff
@@ -474,6 +474,49 @@ $ rustup target add i686-pc-windows-gnu
...
+
+<details>
+ <summary><code>fatal error LNK1112: module machine type 'x86' conflicts with target machine type 'x64'</code></summary>
+ This happens when the variables used setup tools for the wrong architecture, where you have a linker for x86 instead of the target x64.
+ 
+ 
+ Possible fixes:
+ - Open the `x64 Native Tools Command Prompt` instead of `Developer Command Prompt`
+ - Change the env variables to point to the correct version of the tools
+ - Use `vcvars64.bat` instead of `vsvars32.bat` to configure the environment
```
kinnison: is the `vc` vs. `vs` here correct (in the batch filenames)

kinnison (changes requested):
These looks useful, though I have no way of knowing if they're right.  Fix the changes and I'd be okay to include this in the README, though perhaps there may also be other places worth spreading this information to?

bltavares:
@kinnison I would be completely ok to put this somewhere else as well. I've thought that this could be one of the most likely places where people would search when getting started with Rust, but if there is a suggestion to put this information elsewhere I would be fine as well.

I've adopted the changes proposed.

kinnison:
I like the content now.  Could you please squash into one commit and then it can be merged.

rbtcollins (commented):
I think this belongs in one of:
 - rust-lang README.md 
 - cargo's README.md 
 - the rust book

kinnison:
@rbtcollins Any of those locations would also be good yes.  Do you object to this going into *our* readme as well?

rbtcollins:
A little yes, because I think having the content scattered around rather than in one place is misleading for folk (like I was just recently) - who finding one nugget in an official-like location may incorrectly assume that that location is authoritative and comprehensive.

kinnison:
Regarding the point about authority and comprehensiveness I understand and agree to at least some extent.  @bltavares do you think you could persue some other places which might be better for this content?  I'm happy to keep this PR here until you have had a chance to see if there's a better home for this, and then if you cannot find a better place, we'll consider whether we could house it short term with some kind of caveat to mitigate Robert's concerns.

kinnison:
Hi @bltavares 

Happy new year!

I was wondering if you might have time/energy to continue this work?  I'd still like to see this kind of thing documented, but I don't know if you made any progress getting it into any of the other locations that @rbtcollins suggested.

If you simply don't have the time/energy then that's fine, but if you could let us know that'd be helpful.

Thanks,

Daniel.

bltavares:
Hi @kinnison,

I have not yet tried to move this to either cargo, rust-lang or the book. Sorry about that.

I could try to open a similar PR on the book regarding the setup, as this seems to be a better place to group such kind of information given the discussion. I will try to do that during the weekend, and update here with the reference.

kinnison:
@bltavares This needs rebasing onto current master
@rbtcollins It looks like the book rejected it as content, should we consider taking it on?

PR #1913, opened 2019-06-26
Enable clippy on Travis

tesuji:
Let's wait what Travis say before merging.


rbtcollins (approved):
I'm guessing spurious failures. I'm not convinced all clippy lints are good :(. But sure.

kinnison:
So long as this doesn't lead to more pain than gain, I'd be okay with this.  I'd like to see us sort the spurious test failures *FIRST* though :D

PR #1914, opened 2019-06-26
Display the cargo and rustup home paths

### LINKED ISSUE #1907: Rustup does not tell users where the toolchains will be installed
**Problem**

When using the installer to install rustup, it is installed to `RUSTUP_HOME`. 

Last time I tried the installer it did not mention anything about this. That is, a user executes the script, the toolchain gets installed, then the user needs to figure out where things got installed, decide whether those directories are ok, and if not, then google whether that can be changed, and then, uninstall the toolchain, and then set `RUSTUP_HOME` to an appropriate location and reinstall Rust again.

**Steps**
1. Execute `rustup` installation script and press `Y` for everything. Nothing about where things are installed is mentioned, much less a way to change that is offered.

**Possible Solution(s)**

It's an installer. It should tell you where it is going to install the thing, ask you if its ok, and let you change it if it isn't. 

Something like:

> The Rust toolchain will be installed to ${RUSTUP_HOME}. Is this location ok? 
>
> User input: Y/n
>
> [on `n`]: Please set the `RUSTUP_HOME` environment variable to an appropriate location before running the installer (process exit with error).

would probably be enough to fix this bug, but allowing the user to choose an appropriate path would be nice and give a much better impression IMO.

> At which path should the Rust toolchain be installed? 
>
> User input: <valid path>

Installing things to an non-disclosed directory and not even suggesting where to change it always gave me a pretty bad impression. Sure, I was able to search for it online, but I don't think users should be required to do that for something as basic as setting an installation path in an installer.

---

This issue was raised as part of #618 , people have been running into this for years. That issue was closed because it is technically possible to change the installation path by setting the `RUSTUP_HOME` environment variable already. This issue is not about enabling a way to do that, but just about implementing basic installer functionality so that people don't have to do uninstall/reinstall dances or reading the whole README to figure out how to do the most fundamental thing an installer does (installing something somewhere).

skilly-lily:
During interactive installation, the user will see the cargo home
directory as well as the rustup home directory.  They will also see the
environment variables which control these paths.

Fixes #1907

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -74,6 +74,15 @@ Cargo's bin directory, located at:
 
     {cargo_home_bin}
 
+This can be modified with the CARGO_HOME environment variable.
+
+Rustup metadata and toolchains will be installed into the Rustup
+home directory, located at:
+
+    {rustup_home}
+
+This can be modified with the RUSTUP_HOME environment variable.
+
 ",
             $platform_msg,
             r#"
@@ -538,6 +547,7 @@ fn do_msvc_check(_opts: &InstallOpts) -> Result<bool> {
 fn pre_install_msg(no_modify_path: bool) -> Result<String> {
     let cargo_home = utils::cargo_home()?;
     let cargo_home_bin = cargo_home.join("bin");
+    let rustup_home = utils::rustup_home()?;
 
     if !no_modify_path {
         if cfg!(unix) {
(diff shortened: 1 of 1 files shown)


scruffystuffs:
Fixes #1907

skilly-lily:
CI builds are failing due to some windows-only hard link error.  Doesn't seem related to this change.

All errors look like this: 
```
---- warn_about_and_remove_stray_hash stdout ----
thread 'warn_about_and_remove_stray_hash' panicked at 'called `Result::unwrap()`
on an `Err` value: Os { code: 1142, kind: Other, message: "An attempt was made to
create more links on a file than the file system supports." }',
src\libcore\result.rs:997:5
```

src/cli/self_update.rs
```diff
@@ -74,6 +74,15 @@ Cargo's bin directory, located at:
 
     {cargo_home_bin}
 
+This can be modified with the CARGO_HOME envorinment variable.
+
+Rustup metadata and toolchains will be installed the Rustup home
```
rbtcollins: Grammar error here - "will be installed the" -> "will be installed into the"

rbtcollins (commented):
This looks fine to me other than the minor grammar error. @kinnison is there anything we need to do copyright wise - DCO, just accept it, etc?

src/cli/self_update.rs
skilly-lily: Fixed.  Also fixed misspelling of *environment*.

kinnison:
If you're okay with the content then I'd like to see the commits squashed, there's no need for a "grammar fix" or "spelling fix" commit.  Then it can merge.

skilly-lily:
@kinnison @rbtcollins 

- Spelling and grammar fixed
- Commits squashed

Let me know if there's anything else needed.

PR #1916, opened 2019-06-27
build: Transform most CI scripts to bash

tesuji:
We only use those scripts in CI where bash is always the default shell. 
Bash provide modern features and more convenient scripting experience
(i.e. may reduced error rate) with almost no penalty other than portability 
and unnoticeable run time. But that's acceptable for our CI.


kinnison:
I'm prepared to entertain merging this (being explicit about `bash` is good in my books) but I'd like to solve the spurious failures first if we can.

tesuji:
> I'd like to solve the spurious failures first if we can

It looks like we cannot fix it anytime soon. 
We could just merge this PR because it doesn't conflict with that goal.

kinnison:
I do not want clippy on stable -- I value *strongly* the indication that beta isn't regressing (and right now it is) and I can deal with deciding if the failure is relevant.

tesuji:
In that case: Do you want to keep clippy on beta or move it to nightly?

kinnison:
Definitely keep it on beta -- it gives us an excellent canary for whether or not there might be issues in the next stable.

tesuji:
I dropped the clippy commit.

Issue #1851, opened 2019-05-14
`x86_64-pc-windows-gnu/rustup-init.exe` should suggest `x86_64-pc-windows-gnu` as the default host triple


ghost:
If I download this file:

https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe

and run I get this result by default:

~~~
Current installation options:

   default host triple: x86_64-pc-windows-msvc
     default toolchain: stable
  modify PATH variable: yes
~~~

However if I download this file:

https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-gnu/rustup-init.exe

and run I get the same result. With GNU RustUp why is it not setting GNU ABI for
the default host triple?

cup:
https://github.com/rust-lang/rust/issues/60810

tesuji:
I think that each `rustup-init.exe` does the same thing is intended.
You have each one built on each Windows target.
By default rustup install msvc target if it find msvc installed.
You could override that behavior by using `--default-host`, example:
```bash
./rustup-init.exe --default-host=x86_64-pc-windows-gnu
```

I would like @kinnison or anyone to comment about this.

kinnison:
Windows builds of rustup *always* default to the `-msvc` variant, even if they're the `-gnu` variant copy of rustup.  This is done via https://github.com/rust-lang/rustup.rs/blob/master/src/dist/dist.rs#L148 (part of `rustup::dist::dist::TargetTriple::from_host()`.  It might make sense to detect in there if we're a `-gnu` variant and pick the `-gnu` host triple by default, though that'd be a significant departure from current behaviour so we'd have to be certain noone was relying on it.  For CI purposes, I recommend always passing `--default-host xxx` to `rustup-init` in order to be certain you're getting the right platform and `--default-toolchain yyy` to ensure you're always getting the right toolchain for your CI.

cup:
@kinnison its not for CI.

I am a new Rust user and I have found the onboarding process to be painful, between this issue and the standalone archives not actually being archives.

I dont thinks its unreasonable to expect that the `x86_64-pc-windows-gnu/rustup-init.exe` would install the GNU ABI by default.

cup:
https://github.com/rust-lang/rust/issues/60335

retep998:
[...]

PR #1917, opened 2019-06-27
build: Remove *-windows-gnu builds

tesuji:
cc #1851

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,19 +1,30 @@
 image: Visual Studio 2017
-environment:
-  RUSTFLAGS: -Ctarget-feature=+crt-static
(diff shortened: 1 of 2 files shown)


kinnison:
At the time we thought this was the right thing to do -- depending on how Sequoia goes, we might want to switch from defaulting to the MSVC builds to defaulting to the -gnu builds instead, so that we stand a chance of doing GPG verification on Windows.  I need to give this more thought.

kinnison:
On the basis that we've not made any more movement on this in quite a long time, I suggest we close this.  Do you agree?

tesuji:
If we cannot agree about removing windows-gnu builds, let's close it then.

PR #1918, opened 2019-06-27
build: Try to cache cargo registry on Travis

tesuji:
I will often close and reopen to trigger rebuild after cache.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ cache:
   directories:
     - "$HOME/.cargo"
     - "$HOME/.cache/sccache"
-    - "$HOME/.rustup"
+    - "$TRAVIS_BUILD_DIR/target"
 
 matrix:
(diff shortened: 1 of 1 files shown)


kinnison:
There's little gain in caching the registry content AIUI but really we should look at what Travis recommend for this since I worked with them to make things as fast as we could a few weeks ago.

tesuji:
The `cache: cargo` doesn't seem to work with `language: minimal`: https://travis-ci.com/rust-lang/rustup.rs/jobs/212050838

kinnison:
Yeah, `cache: cargo` is probably only going to work for `langauge: rust` but you can see in travis' rust docs what it would cache so that you can replicate the behaviour for us.

tesuji:
Apparently, time to cache ~/.rustup dir is longer than the time it saved
from download and install rust toolchain.

PR #1921, opened 2019-06-28
Copy Rustup binary into test environments

kinnison:
In order to better run on NTFS and friends, this copies the rustup binary into place in the test environment rather than hard-linking it.  We (ab)use an RwLock to achieve this safely since otherwise forked but not-yet execd children of other tests might be holding a FD open resulting in Text-File-Busy type situations.

### DIFF
--- a/tests/mock/clitools.rs
+++ b/tests/mock/clitools.rs
@@ -114,7 +114,7 @@ pub fn setup(s: Scenario, f: &dyn Fn(&mut Config)) {
     let cargo_path = config.exedir.join(format!("cargo{}", EXE_SUFFIX));
     let rls_path = config.exedir.join(format!("rls{}", EXE_SUFFIX));
 
-    hard_link(&build_path, &rustup_path).unwrap();
+    copy_binary(&build_path, &rustup_path).unwrap();
     hard_link(&rustup_path, setup_path).unwrap();
     hard_link(&rustup_path, rustc_path).unwrap();
     hard_link(&rustup_path, cargo_path).unwrap();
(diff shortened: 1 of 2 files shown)


tests/mock/clitools.rs
```diff
@@ -354,6 +354,22 @@ pub fn env(config: &Config, cmd: &mut Command) {
...
 
+use std::sync::RwLock;
+
+/// Returns the lock to be used when creating test environments.
+///
+/// Essentially we use this in `.read()` mode to gate access to `fork()`
+/// new subprocesses, and in `.write()` mode to gate creation of new test
+/// environments. In doing this we can ensure that new test environment creation
+/// does not result in ETXTBSY because the FDs in question happen to be in
+/// newly `fork()`d but not yet `exec()`d subprocesses of other tests.
```
rbtcollins, line 365: I don't understand why this is needed; copying files is a readonly - shared access ok - operation. Copying files that are in use as an executed binary should be ok as long as you open it appropriately.

rbtcollins (approved):

tests/mock/clitools.rs
kinnison: This is more that copying the binary involves a writable FD open for the target, and then if that FD is held open in a forked, but not yet exec'd child process of another test, then when the test which has now completed its copy attempts to exec that copy, the writable FD in the *other* test's forked subprocess prevents the exec.  There could be more precise locking semantics, but this doesn't degrade test performance too much and was the simplest I could come up with which didn't seem to have that race condition.

kinnison:
So, Appveyor succeeded, but Travis still died a death in the `dist` tests.  We're a bit closer at least, so I'm merging this.

PR #1923, opened 2019-06-29
build: Bump rand and flate2

tesuji:


kinnison (commented):
At first glance I don't think this will cause issues.  The `flate2` upgrade is trivial, the `rand` upgrade, while a new major version (0.6 to 0.7) looks compatible enough.

Does this reduce the complexity of our dependency tree at all?

kinnison:
I'm holding off on approving until we fix the spurious failures and then rebase.

PR #1925, opened 2019-07-01
Make error message more specific when the shell profile cannot be updated.

### LINKED ISSUE #292: Fail better if .profile file can't be modified
Installation will fail if `~/.profile` can't be modified. This may or not be the right thing to do, but it should fail gracefully, not with the default I/O error.

C-Saunders:
This is intended to close #292.

I went with the `error_chain` approach @kinnison suggested (thank you, by the way!), but am certainly open to changing that, as well as anything else (e.g. should tests be added for the other supported rc files?).

### DIFF
--- a/tests/cli-self-upd.rs
+++ b/tests/cli-self-upd.rs
@@ -345,6 +345,24 @@ fn install_does_not_add_path_to_bash_profile_that_doesnt_exist() {
 
 #[test]
 #[cfg(unix)]
+fn install_errors_when_rc_file_cannot_be_updated() {
+    setup(&|config| {
+        let rc = config.homedir.join(".bash_profile");
+        fs::File::create(&rc).unwrap();
+        let mut perms = fs::metadata(&rc).unwrap().permissions();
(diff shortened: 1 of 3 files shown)


tests/cli-self-upd.rs
```diff
@@ -343,6 +343,20 @@ fn install_does_not_add_path_to_bash_profile_that_doesnt_exist() {
     });
 }
 
+#[test]
+#[cfg(unix)]
+fn install_errors_when_rc_file_cannot_be_updated() {
+    setup(&|config| {
+        let rc = config.homedir.join(".bash_profile");
+        fs::File::create(&rc).unwrap();
```
C-Saunders, line 351: I felt like there was a lot of `unwrap`ping here, but I wasn't sure what the best way to handle that might be, or if others might even see it as a problem.

rbtcollins: Its a test; meh :)

kinnison (approved):
Looks good to me,

PR #1929, opened 2019-07-02
[wip] travis: Cache cargo registry dir

tesuji:
I will often close and reopen the PR to trigger rebuild after caching.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_cache:
       --git https://github.com/matthiaskrgr/cargo-cache \
       --branch mini \
       --no-default-features \
-      --features mini \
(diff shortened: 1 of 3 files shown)


matthiaskrgr:
Are you sure caching the entire cache is a good idea?
I'm wondering because it means everything will be cached twice, we cache the downloaded `.crate`s but also the extracted content of the `.crate`

Extraction of the crate should not be that much of the bottleneck but the extracted tars also take up a lot of space in the cache.
I would suggest only caching these directories:
````
~/.cargo/bin //installed bins
~/.cargo/git/db //bare git repos
~/.cargo/registry/index //registry index
~/.cargo/registry/cache //.crate tarballs
````

PR #1931, opened 2019-07-02
Do not list unavailable targets

kinnison:
We were mistakenly listing targets as available when they were not, resulting in confusion when running `rustup target add all`.

This change ensures that we do not do that, by ensuring we use extension targets rather than the toolchain's target, when checking availability.  We do not do this for components too at this time, since components are almost always the same arch as the rustc itself.


nrc (approved):
LGTM

kinnison:
AFAICS, the failures are the dist test timing out / crashing randomly on Windows CI - something we've not yet managed to reproduce outside of CI.

PR #1933, opened 2019-07-03
Display RUSTUP_HOME in `rustup show` and `rustup show home`

### LINKED ISSUE #1627: Query RUSTUP_HOME
Is there a way to query the rustup home from rustup ? Right now I am checking if the RUSTUP_HOME variable is defined, and using that if its the case, and otherwise using `${HOME}/.rustup`, but I'd prefer if there was a `rustup --home` command that would just print whatever `rustup` is using.

fisherdarling:
This PR patches `rustup show` to also display the RUSTUP_HOME path:
```
$ rustup show
rustup home:  /home/fisher/.rustup
```
It also adds a new subcommand, `rustup show home` that directly prints the path:
```
$ rustup show home
/home/fisher/.rustup
```

This closes #1627.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -43,6 +43,7 @@ pub fn main() -> Result<()> {
         ("dump-testament", _) => common::dump_testament(),
         ("show", Some(c)) => match c.subcommand() {
             ("active-toolchain", Some(_)) => handle_epipe(show_active_toolchain(cfg))?,
+            ("home", Some(_)) => handle_epipe(show_rustup_home(cfg))?,
             (_, _) => handle_epipe(show(cfg))?,
         },
         ("install", Some(m)) => update(cfg, m)?,
@@ -132,7 +133,11 @@ pub fn cli() -> App<'static, 'static> {
                     SubCommand::with_name("active-toolchain")
                         .about("Show the active toolchain")
                         .after_help(SHOW_ACTIVE_TOOLCHAIN_HELP),
-                ),
+                )
+                .subcommand(
+                    SubCommand::with_name("home")
+                        .about("Display the computed value of RUSTUP_HOME"),
+                )
         )
         .subcommand(
             SubCommand::with_name("install")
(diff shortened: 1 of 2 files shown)


fisherdarling:
cc @kinnison

src/cli/rustup_mode.rs
```diff
@@ -111,6 +112,12 @@ pub fn cli() -> App<'static, 'static> {
         .setting(AppSettings::VersionlessSubcommands)
         .setting(AppSettings::DeriveDisplayOrder)
         .setting(AppSettings::SubcommandRequiredElseHelp)
+        .arg(
+            Arg::with_name("home")
+                .help("Print $RUSTUP_HOME")
+                .short("h")
+                .long("home"),
+        )
```
kinnison: Why is this arg here at all?  this would be `rustup --home` which is not something I really want, nor are you really processing it anyway.

src/cli/rustup_mode.rs
```diff
@@ -132,7 +139,11 @@ pub fn cli() -> App<'static, 'static> {
                     SubCommand::with_name("active-toolchain")
                         .about("Show the active toolchain")
                         .after_help(SHOW_ACTIVE_TOOLCHAIN_HELP),
-                ),
+                )
+                .subcommand(
+                    SubCommand::with_name("home")
+                        .about("Print $RUSTUP_HOME"),
```
kinnison: `"Display the computed value of RUSTUP_HOME"` might be better here

kinnison (changes requested):
Needs a teeny bit of cleanup, and then needs a test to verify that it works properly.

src/cli/rustup_mode.rs
fisherdarling: Sorry, that was an artifact of a previous attempt. I forgot to remove it.

fisherdarling:
Since `rustup show` always prints the home directory, I created another macro `for_host_and_home!` that would also get the config's `rustupdir`. If we should do this some other way, let me know.

kinnison:
This is starting to look good -- please fold the formatting fixes into the commits which are relevant, and fold the fix to what you originally submitted into the top commit, so that there are two commits - one for the feature, and one for the test updates.

kinnison (approved):

PR #1934, opened 2019-07-03
Run tests serially on Windows CI

gnzlbg:
Give something like this a try for a week. If you stop seeing failures, then you probably have a race condition somewhere (not necessarily here, but in one of the dependencies, libtest, etc.).

r? @kinnison

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,7 +95,7 @@ matrix:
 
     - &windows
       os: windows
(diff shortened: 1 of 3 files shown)


kinnison:
I've merged something less drastic but similar to this.  We'll be keeping an eye on this as we really do need to track down what's going on.

PR #1935, opened 2019-07-03
self_update: Attempt to update without race condition

kinnison:
On Windows, there is a race condition where in self-update
we could end up unable to complete the job.  This is an
attempt at fixing that.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -630,7 +630,17 @@ fn install_bins() -> Result<()> {
     // NB: Even on Linux we can't just copy the new binary over the (running)
     // old binary; we must unlink it first.
     if rustup_path.exists() {
-        utils::remove_file("rustup-bin", &rustup_path)?;
+        if cfg!(windows) {
+            let removal_path = bin_path.join(&format!("rustup-to-delete{}", EXE_SUFFIX));
+            if removal_path.exists() {
+                utils::remove_file("rustup-to-delete", &removal_path)?;
+            }
+            fs::rename(&rustup_path, &removal_path).chain_err(|| {
+                ErrorKind::WindowsBinaryRename(rustup_path.clone(), removal_path.clone())
+            })?;
+        } else {
+            utils::remove_file("rustup-bin", &rustup_path)?;
+        }
     }
     utils::copy_file(&this_exe_path, &rustup_path)?;
     utils::make_executable(&rustup_path)?;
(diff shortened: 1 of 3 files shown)


src/cli/self_update.rs
```diff
@@ -630,7 +630,17 @@ fn install_bins() -> Result<()> {
     // NB: Even on Linux we can't just copy the new binary over the (running)
     // old binary; we must unlink it first.
     if rustup_path.exists() {
-        utils::remove_file("rustup-bin", &rustup_path)?;
+        if cfg!(windows) {
```
rbtcollins, line 633: One thing to note: the WSLv1 environment is constrained by the underlying NTFS semantics, though MS obviously have improved things substantially - just wondering if we might be better off not making this os-conditional.

src/cli/self_update.rs
```diff
@@ -630,7 +630,17 @@ fn install_bins() -> Result<()> {
...
+            let removal_path = bin_path.join(&format!("rustup-to-delete{}", EXE_SUFFIX));
+            if removal_path.exists() {
```
rbtcollins, line 635: This shouldn't be needed - https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw - pretty sure this is exposed either in the sys module or winapi - with MOVEFILE_REPLACE_EXISTING

src/cli/self_update.rs
```diff
@@ -630,7 +630,17 @@ fn install_bins() -> Result<()> {
...
+                utils::remove_file("rustup-to-delete", &removal_path)?;
+            }
+            fs::rename(&rustup_path, &removal_path).chain_err(|| {
```
rbtcollins, line 638: This will need special flags if the rustup_path may still be in use, but if its not in use, great.

rbtcollins (commented):
I'm not sure how this is race-free: it leaves a clear period while there is no exe at all? There's also no coordination with other processes so nothing preventing read handles on the files being deleted and replaced - see the discussion in the ticket I opened about updates with RLS open, which will impact this too

kinnison:
> I'm not sure how this is race-free: it leaves a clear period while there is no exe at all? There's also no coordination with other processes so nothing preventing read handles on the files being deleted and replaced - see the discussion in the ticket I opened about updates with RLS open, which will impact this too

The race condition in question was regarding the way that a process which invokes `rustup update` or `rustup self update` can see its child exit, and then start a new process of `rustup` which ends up preventing the self-updater from completing.

I'm in no way claiming this is the best way to do it, merely that it would mitigate the race, particularly in the test suite :D

src/cli/self_update.rs
kinnison: It certainly would not be a bad idea to make it uniform across all the OSes

src/cli/self_update.rs
kinnison: I wasn't certain, this sounds handy to know though.  If we decide to proceed along this path we can easily make this change.

src/cli/self_update.rs
kinnison: The approach is deliberate because `rustup_path` *is* in use at this point.  Windows permits renaming the binary so long as it doesn't get deleted things are okay.  This was the only approach I could come up with which didn't require the parent `rustup` process to exit first.

kinnison:
My feeling after revisiting this is that there must be a better way, so I'm abandoning this PR for now.

PR #1936, opened 2019-07-03
Additional sudo checks

das-sein:
An attempt to improve sudo warning sensitivity to help avoid issues like #1911 

Introduces two additional checks.

The first checks for sudo-specific environment variables, then triggers an error message like the home directory discrepancy test which requires explicit user interaction to override.

The second attempts to detect a discrepancy between the actual and effective IDs of the user executing the process, which triggers a warning message that alerts the user they may be executing the installation binary as sudo. This requires no user interaction to bypass.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -510,6 +510,32 @@ fn do_anti_sudo_check(no_prompt: bool) -> Result<()> {
         }
     }
 
+    #[cfg(unix)]
+    /// Attempt to detect sudo with environment variables defined in `man 8 sudo`.
+    /// Should act as a backup method to the more rigorous home directory detection.
+    pub fn sudo_env() -> bool {
+        if env::var_os("SUDO_GID").is_some()
+            || env::var_os("SUDO_UID").is_some()
+            || env::var_os("SUDO_USER").is_some()
+        {
+            return true;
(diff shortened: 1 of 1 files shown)


kinnison (commented):
This looks like it ought to be good, but I wonder if we might want a test for at least the environment one?

das-sein:
@kinnison I think you're right. The test cases might be a bit tricky to create, but it would be good to increase coverage as I don't think this function is presently tested at all.

kinnison:
Hi @fotocoyotl 

Have you had any further thoughts around adding a test case for this?  Setting specific environment variables before running a test of `rustup-init` ought not to be too hard.

Also, could you rebase onto current master as I've merged a CI fix to try and mitigate the windows test failures.

das-sein:
@kinnison I have, just haven't had a chance to implement them yet, sorry. Rebased.

kinnison:
Hi @latrare 

Happy new year!  I was wondering if you might have time in this new year to pick up this work and complete the remaining testing needed to make it mergeable?  If you simply don't have the time/energy/interest then that's completely fine, but if you could let us know either way, I'd appreciate it.

Thanks again for all your efforts last year with Rustup.

Daniel.

PR #1944, opened 2019-07-11
Recommend installing MSVC English language pack

crlf0710:
Update prompt to recommend user to install the English language pack.

cc https://github.com/rust-lang/rust/issues/35785


kinnison (approved):
Seems fair to me.

Issue #1928, opened 2019-07-02
CLI switch to disable download progress output


benaryorg:
### Describe the problem you are trying to solve

I want to add a new toolchain in a Docker container and suppress the download progress, since that messes with my terminal rather hard.
This applies to every other non-interactive download.

### Describe the solution you'd like

A CLI switch (`-s`/`--silent`: *curl* style, `-q`/`--quiet`: *wget* style) which disables output meant for a terminal.

### Notes

It is possible to workaround this by just piping the output anywhere (`| cat` at the end works fine):

https://github.com/rust-lang/rustup.rs/blob/2ece700231f8a1778647981df94052e5b27cbedf/src/cli/download_tracker.rs#L73

That workaround isn't viable if you have e.g. a *Dockerfile* with `execvp()`-like statements to avoid running in a shell (for the doom it shall bring upon thee).

kinnison:
This ought to be a reasonable intermediate project for someone to undertake.  I am prepared to mentor someone to do the work.

PicoJr:
Here is what I've written so far: <https://github.com/PicoJr/rustup.rs/tree/feat-1928>

Should I submit a PR?

Note: It's my first time contributing to rustup and I'm a rust beginner.

Help appreciated =)

## TODO?

- [ ] write unit test for the `--silent` option
- [ ] add `--quiet` alias

## Unit Tests

I ran `cargo test`, some tests failed with message "error: self-update is disabled for this build of rustup".

kinnison:
My preference is for `--quiet` rather than `--silent` but it looks like you're on your way to a solution.

The error message you mention suggests that you've managed to end up with the no-self-update feature enabled, or else you've got some unusual environment variables set such as `SNAP`.

PicoJr:
Ok, I'll rename the option to `--quiet`.

PicoJr:
I've submitted a PR: #1945

pickfire:
This issue should be solved.

PR #1945, opened 2019-07-14
feat: --quiet option

PicoJr:
# PR Content

issue: #1928 

add `--quiet` option. it disables progress console output.

## TODO

- [x] write unit test for this new option.
- [x] verify default behavior in `proxy_mode.rs` is the right one.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -124,6 +124,7 @@ pub fn cli() -> App<'static, 'static> {
         )
         .arg(
             Arg::with_name("quiet")
+                .conflicts_with("verbose")
                 .help("Disable progress output")
                 .short("q")
                 .long("quiet"),
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -31,7 +31,8 @@ pub fn main() -> Result<()> {
 
     let matches = cli().get_matches();
     let verbose = matches.is_present("verbose");
-    let cfg = &common::set_globals(verbose)?;
+    let quiet = matches.is_present("quiet");
+    let cfg = &common::set_globals(verbose, quiet)?;
 
     if maybe_upgrade_data(cfg, &matches)? {
         return Ok(());
@@ -120,6 +121,12 @@ pub fn cli() -> App<'static, 'static> {
                 .help("Enable verbose output")
                 .short("v")
                 .long("verbose"),
+        )
+        .arg(
+            Arg::with_name("quiet")
+                .help("Disable progress output")
+                .short("q")
+                .long("quiet"),
         )
         .subcommand(
             SubCommand::with_name("dump-testament")
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -32,6 +32,12 @@ pub fn main() -> Result<()> {
                 .help("Enable verbose output"),
         )
         .arg(
+            Arg::with_name("quiet")
+                .short("q")
+                .long("quiet")
+                .help("Disable progress output"),
+        )
+        .arg(
             Arg::with_name("no-prompt")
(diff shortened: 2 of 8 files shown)


kinnison:
> is this a reasonable default?

I'll need to properly review to be sure, but first glance suggests it's okay.
 
> I don't really understand the purpose of `proxy_mode.rs`

This is the entry point for rustup proxies.  Proxying is the mechanism by which `rustc` `cargo` etc, are really `rustup` but we use `argv[0]` to decide which command to run from inside the detected toolchain.  It *can* result in a download if the detected toolchain isn't installed I think, though in that context, I imagine defaulting to quiet is fine.

kinnison (commented):
After reviewing, I'm mildly concerned that we have `verbose` and `quiet` which are not really antonyms of one another.  But I'm also not sure of better terminology right now.

I'll await your further updates with tests etc, before I pass judgement further.  Renaming options is the easiest of things to do.

Could you also be sure to rebase against new master since I've merged a change which might make the CI more useful.

PicoJr:
> After reviewing, I'm mildly concerned that we have `verbose` and `quiet` which are not really antonyms of one another. But I'm also not sure of better terminology right now.
> 
> I'll await your further updates with tests etc, before I pass judgement further. Renaming options is the easiest of things to do.
> 
> Could you also be sure to rebase against new master since I've merged a change which might make the CI more useful.

Ok, thanks for the review =), I'll rebase and update the PR accordingly.

- [x] rebase
- [x] add unit tests

PicoJr:
## Tests

The goal of `--quiet` is to suppress progress output.

During unit tests stdout and stderr are checked against expected output.

However final stdout is the same regardless of whether or not `--quiet` option was used  (progress line is flushed out).

For the lack of a better solution I copied tests from `cli-rustup.rs`: `rustup_stable` and `toolchain_install_is_like_update` and used `--quiet` option.

It won't prove the feature works as expected (disables the progress output) but it shows the `--quiet` option can be used and produces the same output as before (except for the progress line)

> I'm open to suggestions for improving the tests

## Name

As for the option name, wget has a `--progress=type` option.

`--quiet` could be renamed as `--no-progress` ?

src/cli/setup_mode.rs
```diff
@@ -31,6 +31,12 @@ pub fn main() -> Result<()> {
                 .long("verbose")
                 .help("Enable verbose output"),
         )
+        .arg(
+            Arg::with_name("quiet")
+                .short("q")
+                .long("quiet")
+                .help("Disable progress output"),
+        )
```
kinnison, line 40: Would it make sense if quiet were to conflict with verbose?

src/cli/rustup_mode.rs
```diff
@@ -121,6 +122,12 @@ pub fn cli() -> App<'static, 'static> {
                 .short("v")
                 .long("verbose"),
         )
+        .arg(
+            Arg::with_name("quiet")
+                .help("Disable progress output")
+                .short("q")
+                .long("quiet"),
+        )
```
kinnison, line 131: Would it make sense for quiet to conflict with verbose?

kinnison (commented):
Just a query about verbose vs quiet then probably mergeable

src/cli/setup_mode.rs
PicoJr: | quiet | verbose | comment |
|--------|------------|---------------|
| Y      | Y           | does not really makes sense: the purpose of quiet (less output) is the opposite of verbose (more output) |
| Y      | N           | expected usage for quiet |
| N      | Y           | expected usage for verbose |
| N      | N           | default usage |

It would indeed make sense if quiet and verbose were conflicting options =)

kinnison (approved):
LGTM, Thank you :+1:

PR #1950, opened 2019-07-21
Add rust-gdbgui to the list of proxied tools

mtrsl:
Currently, a link for `rust-gdbgui` in Cargo's bin directory isn't created by `rustup`. This adds `rust-gdbgui` to the list of tools that are proxied by `rustup` so that a link is created and we can run `rust-gdbgui` directly.

Please let me know if the tests in [tests/cli-self-upd.rs](https://github.com/rust-lang/rustup.rs/blob/master/tests/cli-self-upd.rs) should be updated to include `rust-gdbgui` as well. I am guessing wherever a test involves `rust-gdb`, it should also involve `rust-gdbgui` in a similar way?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,6 +18,7 @@ pub static TOOLS: &[&str] = &[
     "cargo",
     "rust-lldb",
     "rust-gdb",
+    "rust-gdbgui",
     "rls",
     "cargo-clippy",
     "clippy-driver",


kinnison:
Which component provides `rust-gdbgui` ?  Is it `gdb-preview` alongside `rust-gdb` ?  If so, then you need to add an entry into `component_for_bin()` in `src/lib.rs`.

mtrsl:
Where are the components defined? Do they come from the compiler sources? I see that `gdb-preview` is used in `rustup` for `rust-gdb`, as you mentioned, but I can't find any reference to `gdb-preview` in the `rust` tree.

kinnison:
The components end up defined in the rustbuild process.  I'm not entirely sure where the channel manifests are generated from off-hand, though I think @nrc knows.

nrc:
The manifest is generated by [this code](https://github.com/rust-lang/rust/blob/master/src/tools/build-manifest/src/main.rs) and that is where all the names, etc are defined. I don't know where it runs exactly, but you just need to push a change to the Rust repo and it will be used for the next distributable.

kinnison:
Hi @rumajo 

Happy new year!  I was wondering if you might have time/energy/interest in persuing this PR again this year?  If not, then no worries, but if you could let us know either way then I'd appreciate it.

Regards,

Daniel.

mtrsl:
Hi @kinnison,

Happy new year to you as well! Apologies for the delay with this PR. I'm am still interested in pursuing it - just a matter of finding the time. I'll try to have a look over the next few days, but failing that I'll maybe hand it over.

PR #1953, opened 2019-07-27
Update inline help text for `rustup completions`

bcdady:
`$ rustup completions` will show the following updated help:

```rust
rustup-completions 
Generate tab-completion scripts for your shell

USAGE:
    rustup completions [ARGS]

...

DISCUSSION:
    Enable tab completion for Bash, Fish, Zsh, or PowerShell
    The script is output on `stdout`, allowing one to re-direct the
    output to the file of their choosing. Where you place the file
    will depend on which shell, and which operating system you are
    using. Your particular configuration may also determine where
    these scripts need to be placed.
```

### DIFF
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -155,13 +155,12 @@ pub static DOC_HELP: &str = r"DISCUSSION:
 
 pub static COMPLETIONS_HELP: &str =
 r"DISCUSSION:
-    One can generate a completion script for `rustup` that is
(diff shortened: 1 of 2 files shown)


kinnison (approved):
The text change here is minimal but does bring it inline with the README content.  Since it reads as well to me as a *NIX user as it did before; and the submitter is a Windows user, I'll be pleased to accept it as an over-all improvement on the status quo.

PR #1973, opened 2019-08-26
Bump home crate to v0.5

tesuji:
This home release removes support for multirust folder used in old version of rustup.


kinnison (approved):
Change looks good.

PR #1976, opened 2019-08-28
alias `rustup date` to update

baloo:
I keep typing `rustup date`, and it looks like it's easier to fix the code
to work around that than my brain.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -177,6 +177,7 @@ pub fn cli() -> App<'static, 'static> {
         .subcommand(
             SubCommand::with_name("update")
                 .about("Update Rust toolchains and rustup")
+                .aliases(&["date"])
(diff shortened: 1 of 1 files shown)


kinnison:
Hi, thank you for your contribution to `rustup`.

Unfortunately we're not really interested in extending the CLI surface of `rustup` in this way because it implies a requirement to support these typos for the foreseeable future, and adds a support burden for other distribution channels and the like.

I'd suggest that, if this is a very common typo for yourself, that you add a shell function to your shell's RC file, along the lines of:

```shell
rustup () {
    if test "x$1" = "xdate"; then
        shift
        command rustup update "$@"
    else
        command rustup "$@"
    fi
}
```

Again, thank you for wanting to contribute, and I'm sorry that we can't take this patch on.

baloo:
no worries :) thanks for the work @kinnison !

Issue #1982, opened 2019-09-04
New nightlies are at risk of tripping `File too big`


kinnison:
As seen in https://github.com/rust-lang/rustup.rs/issues/1970#issuecomment-527942840 we are now generating nightlies which have 170M rlibs in.

Our multithreaded unpacker assumes each thread won't deal with files > 100M and thus limits its RAM usage by number of threads.  We now seem to have files much larger than that and so we might need to up the limit at https://github.com/rust-lang/rustup.rs/blob/master/src/dist/component/package.rs#L280 and check associated numbers to re-trim things.

/cc @rbtcollins Opinions?

PR #1983, opened 2019-09-05
Update the sanity limit to 200M per file

kinnison:
This is needed because the stdlib .rlib files are now around 100M
and some targets exceed the old sanity limit.

/cc @rbtcollins


kinnison:
This will close #1982 if merged.

kinnison:
I'm going to merge this, if there's badness as a result, I trust that we'll spot it early.

PR #1988, opened 2019-09-09
Implement verbose option for "rustup toolchain list"

### LINKED ISSUE #1479: rustup toolchain: print the path the the toolchain
On my system I have:
```
$ rustup toolchain list
stage1
```

One question I might ask is: "where are these toolchains on the filesystem"? Especially for the manually linked (using `rustup toolchain link`) toolchain `stage1`.

How about `rustup toolchain list -v` prints the toolchains with their paths?

Thanks!

BeniCheni:
This PR resolves #1479.

##### local test

```sh
→ rustup toolchain list -v
stable-x86_64-apple-darwin (default)	/Users/bchen/.rustup
nightly-2019-09-05-x86_64-apple-darwin	/Users/bchen/.rustup
nightly-x86_64-apple-darwin	/Users/bchen/.rustup
```

```sh
→ rustup toolchain list --verbose
stable-x86_64-apple-darwin (default)	/Users/bchen/.rustup
nightly-2019-09-05-x86_64-apple-darwin	/Users/bchen/.rustup
nightly-x86_64-apple-darwin	/Users/bchen/.rustup
```

```sh
→ rustup toolchain list -h
rustup-toolchain-list
List installed toolchains

USAGE:
    rustup toolchain list [FLAGS]

FLAGS:
    -h, --help       Prints help information
    -v, --verbose    Enable verbose output with toolchain infomation
```

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -424,8 +424,13 @@ pub fn list_installed_components(toolchain: &Toolchain<'_>) -> Result<()> {
     Ok(())
 }
 
-pub fn list_toolchains(cfg: &Cfg) -> Result<()> {
+pub fn list_toolchains(cfg: &Cfg, is_verbose: bool) -> Result<()> {
     let toolchains = cfg.list_toolchains()?;
+    let toolchain_info = if is_verbose {
+        format!("\t{}", &cfg.rustup_dir.display())
+    } else {
+        String::from("")
+    };
 
     if toolchains.is_empty() {
         println!("no installed toolchains");
@@ -436,11 +441,11 @@ pub fn list_toolchains(cfg: &Cfg) -> Result<()> {
             } else {
                 ""
             };
-            println!("{}{}", &toolchain, if_default);
+            println!("{}{}{}", &toolchain, if_default, toolchain_info);
         }
     } else {
         for toolchain in toolchains {
-            println!("{}", &toolchain);
+            println!("{}{}", &toolchain, toolchain_info);
         }
     }
     Ok(())
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -53,7 +53,7 @@ pub fn main() -> Result<()> {
         ("default", Some(m)) => default_(cfg, m)?,
         ("toolchain", Some(c)) => match c.subcommand() {
             ("install", Some(m)) => update(cfg, m)?,
-            ("list", Some(_)) => handle_epipe(common::list_toolchains(cfg))?,
+            ("list", Some(m)) => handle_epipe(toolchain_list(cfg, m))?,
             ("link", Some(m)) => toolchain_link(cfg, m)?,
             ("uninstall", Some(m)) => toolchain_remove(cfg, m)?,
             (_, _) => unreachable!(),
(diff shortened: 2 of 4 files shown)


src/cli/common.rs
```diff
@@ -432,8 +433,14 @@ pub fn list_installed_components(toolchain: &Toolchain<'_>) -> Result<()> {
     Ok(())
 }
 
-pub fn list_toolchains(cfg: &Cfg) -> Result<()> {
+pub fn list_toolchains(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
```
kinnison: Rather than this, I'd prefer to see a wrapper function in `rustup_mode.rs` and this function take a `bool` for verbose mode.

(i.e. let's not leak our use of `clap` into this module file)

tests/cli-v1.rs
```diff
@@ -99,7 +99,23 @@ fn list_toolchains() {
...
+        expect_stdout_ok(
+            config,
+            &["rustup", "toolchain", "list", "-v"],
+            "beta-2015-01-01",
+        );
+        expect_stdout_ok(
+            config,
+            &["rustup", "toolchain", "list", "--verbose"],
+            "beta-2015-01-01",
+        );
```
kinnison: I don't really see anything in this suggesting that the verbose flag actually works, merely that it's not rejected by the command.  Could you augment the test to include at least the tab character so we can be somewhat confident that the verbosity is happening?  (Ditto that the tab is not present on a non-verbose listing)  Same applies to your cli-v2.rs changes.

kinnison (changes requested):
Mostly good, I'd like to see the use of `clap` not leaked into `common.rs` and the tests need a little more thought, but otherwise this is very close to mergeable.

Thank you very much for your contribution and I hope you have time to act on this feedback and provide an updated commit.

Please note that we don't like fixup commits in our history if we can avoid them; so either squash your updates into this commit, or be prepared to be asked to do so once we're happy the content is mergeable.

D.

BeniCheni:
Thank you for the feedback, @kinnison. Just a quick note to acknowledge the feedback. Will definitely find time to edit to update the PR.

BeniCheni:
Greeting @kinnison. Per suggestions, I've pushed the updates in a squashed commit to:

- used `toolchain_list` wrapper fn to avoid `clap` not leaked into `common.rs`
- updated `list_toolchains` tests in `cli-v1.rs` & `cli-v2.rs` to assert the tab character in the toolchain path, to ensure verbosity option are being invoked.

PTAL? Thanks.

BeniCheni:
Noticed failed tests in CI log. Will check and update accordingly.

src/cli/rustup_mode.rs
```diff
@@ -217,7 +217,17 @@ pub fn cli() -> App<'static, 'static> {
                 .setting(AppSettings::VersionlessSubcommands)
                 .setting(AppSettings::DeriveDisplayOrder)
                 .setting(AppSettings::SubcommandRequiredElseHelp)
-                .subcommand(SubCommand::with_name("list").about("List installed toolchains"))
+                .subcommand(
+                    SubCommand::with_name("list")
+                        .about("List installed toolchains")
+                        .arg(
+                            Arg::with_name("verbose")
+                                .help("Enable verbose output with toolchain infomation")
```
kinnison: Typo here `infomation` instead of `information`

kinnison (changes requested):
One tiny typo, I can fix it if you don't want to.

BeniCheni:
Thanks for catching the typo, @kinnison. Updated.

kinnison (approved):
:+1:

Issue #1821, opened 2019-04-29
ZSH Cargo autocompletion not working


andradei:
**Problem**
`cargo` tab completion is not working. I installed it the same way I did `rustup`, which works.

**Steps**
1. `rustup completions zsh cargo > /usr/local/share/zsh-completions/_cargo`
2. On my `.zshrc` I have

```
fpath=(/usr/local/share/zsh-completions $fpath)
compinit
```

3. Open a new terminal or reload `.zshrc`

**Notes**

Output of `rustup --version`: rustup 1.18.1 (462683b3a 2019-04-25)
Output of `rustup show`:

```
Default host: x86_64-apple-darwin

installed toolchains
--------------------

stable-x86_64-apple-darwin
nightly-x86_64-apple-darwin

active toolchain
----------------

stable-x86_64-apple-darwin (default)
rustc 1.34.0 (91856ed52 2019-04-10)
```

kinnison:
I imagine this is an issue with the zsh completions provided for cargo, rather than rustup itself.

I have just tried manually installing the cargo completions into my zsh setup and can confirm they seem to do nothing.

I'd suggest filing an issue against https://github.com/rust-lang/cargo referencing this issue so that we can discuss it usefully across the projects.

ehuss:
I'm not sure, but it seems to be an issue with how rustup is creating a completion file using `source`.  If I copy the actual source into `~/.zfunc/` and run `compinit -D`, it seems to work fine.

burjui:
@ehuss I didn't even know about the completion dump file and that `_cargo` that `rustup` produces does contain the actual completions, just `source`ing them instead. Indeed, copying the sourced file instead works.

Why doesn't `source`work though? It is supposed to be the equivalent of copying the contents of the sourced file into the script. Then it hit me: the actual completion file contains the `#compdef cargo` line at the start, so when you source it, it is ignored. I've always hated magic comments, and this issue is a perfect example why. Anyway, if you take `_cargo` generated by `rustup`
```zsh
source $(rustc --print sysroot)/share/zsh/site-functions/_cargo
```
and add the line to it:
```zsh
#compdef cargo
source $(rustc --print sysroot)/share/zsh/site-functions/_cargo
```
it works just fine. I feel dumb now for not going all the way with this from the start 🤦‍♂️

Anyway, submitted a PR for this: #1995

kinnison:
If anyone here can confirm the work on #1995 then I'd appreciate it.

[...]

PR #1995, opened 2019-09-15
Fix generation of Cargo completions for ZSH (issue #1821)

burjui:
When `source`ing the original completion file, the `#compdef cargo` line at the start is ignored, and ZSH does not autoload the completions, so `rustup` must put it the generated `_cargo` as well.
<sub>I guess `rustup` does not create a symlink instead, because that would be less portable.</sub>


kinnison:
I do not use zsh completions with cargo; but I can appreciate others might -- is there anyone who can confirm if this is sufficient to make things work?

ehuss:
I don't normally use zsh, but I tried this out and it seems to work (zsh 5.7.1).  Looks like it needs a rustfmt though.

burjui:
@ehuss Done.

PR #1996, opened 2019-09-16
Add Content Security Policy to the website

### LINKED ISSUE #1986: Add a Content Security Policy to rustup.rs
To get an A+ score on [the Mozilla Observatory](https://observatory.mozilla.org/analyze/rustup.rs?third-party=false) we need to add a Content Security Policy to the website.

emilyalbini:
This PR adds the following CSP to the rustup.rs website:

```
default-src 'none'; script-src 'self'; style-src 'self'; img-src 'self' https://www.rust-lang.org; font-src 'self'
```

Along with the new header I had to tweak the platform detection script to use classes instead of applying styles directly (which would've required the `unsafe-inline` CSP directive).

Fixes https://github.com/rust-lang/rustup.rs/issues/1986
r? @kinnison

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -13,7 +13,7 @@
 
 <body id="idx">
 
-<a id="platform-button" style="display: none;" href="#">
+<a id="platform-button" class="display-none" href="#">
   click or press "n" to cycle platforms
 </a>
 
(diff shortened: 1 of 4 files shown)


kinnison (commented):
Looks fundamentally good.  I'll check a bit more before I approve/merge though

kinnison (approved):
I've grepped around and it looks good.  I think you've caught all ".style." and "style="

Issue #1676, opened 2019-03-02
Check for component included without installing toolchain?


brackleian:
I guess this is a feature request rather than a bug; it may already be doable but I can't find out how.

Is it possible to check for a given component being included in a given toolchain, without actually downloading the whole toolchain? E.g. I'm using 'rls', which is in some nightly builds but not all. To get the latest, I have to work backwards from today's date. Installing the whole toolchain takes significant time, after which I have to uninstall it if it wasn't the right one.

Perhaps a textual list of included components could be stored on the toolchain host alongside each build; I don't mind having to query it with wget.

ehuss:
I'm not sure if this will help for your case, but you can check https://rust-lang-nursery.github.io/rust-toolstate/ or https://mexus.github.io/rustup-components-history/ to find when the last working build was.

norru:
Hello, 

I think it would be useful to have a look at the discussion at https://github.com/rust-lang/rls/issues/1221. There have been reports of users finding the current `rustup` behaviour not friendly in this use case, with a couple of suggestions of how this could be easily solved.

Here's an excerpt from that conversation, hope it helps:

> 
> 
>     we do expect RLS to be sometimes unavailable on the nightly channel
> 
> This user experience is not good. IMHO it would be better if either rustup only downloaded nightlies with all components I have installed or if rustup told me I'm "up-to-date". But failing is really quite bad.
> @alexheretic
> Member
> alexheretic commented on Jan 8 •
> 
>     This user experience is not good. IMHO it would be better if either rustup only downloaded nightlies with all components I have installed or if rustup told me I'm "up-to-date". But failing is really quite bad.
> 
> When the latest nightly is missing components you have installed, rustup will fail the update and tell you this, is that what you're talking about? This protects rustup users from RLS, etc disappearing and it's quite useful.
> 
> If that's working for you, do you just disagree with the message and exit code? If you have bugs/ideas for rustup I'd suggest raising an issue or looking over the discussion in rust-lang/rustup.rs.




>     rustup does the former and if it were to do the second it would be lying (it would be telling you you're up to date when you aren't).
> 
> Currently rustup does nothing as far as I understood. Does it actually download the latest possible version? In that case I was simply confused by the error message.
> 
>     rustup update --with-existing-components would have to figure out which components I have installed in nightly, figure out which ones are available in the latest N nightlies, detect, say nightly-2019-01-01 matches the criteria, then "transparently" sync the nightly "head" to that version.
> 
[...]

Issue #1501, opened 2018-09-05
option to install toolchain version IFF it has certain components


matthiaskrgr:
When dealing with continuous like travis, I can specify a toolchain I want to have
````
language: rust
rust:
  - beta
  - nightly
````

but not what components (rustfmt clippy) I want to install.
This means if I want to use clippy or rustfmt, I have to manually ````rustup component add```` it or, if the current nightly version does not ship it, try to *somehow* install it myself and hope it builds with the current rustc nightly.
It would be great to have some kind of option "install the lastest nightly that has components x and y available" which entails potentially downgrading to a earlier nightly if the current nightly does not ship the components but the later nightly does.

Perhaps ````rustup install nightly --needed-components=clippy,rustfmt```` or something like that.

RalfJung:
The title of this issue is misleading... it talks about updating when the issue is about installation.

Could someone rename it appropriately? I think this is the better alternative to https://github.com/rust-lang/rustup.rs/issues/1602 and https://github.com/rust-lang/rustup.rs/issues/1676.

matthiaskrgr:
@RalfJung updated!

najamelan:
@RalfJung 
> it talks about updating when the issue is about installation

It would be nice if `rustup update` did the same thing though. Eg. get you the latest nightly that has your components.

RalfJung:
@najamelan your comment consists only of quotes, looks like a formatting issue?

> get you the latest nightly that has your components.

AFAIK if you have a component installed, rustup will only upgrade to nighties that have that component.

najamelan:
@RalfJung Sorry, it needed a white line to work there. 

> AFAIK if you have a component installed, rustup will only upgrade to nighties that have that component.

So the point is, if I have a nightly that is one month old, with components, and I run `rustup update`, if todays nightly doesn't have the components, it won't update anything at all, even if yesterdays nightly had the required components.

So I think lots of people would like for rustup to update to the latest nightly that has that component.

The issue is very similar for installs and updates, so I think both should considered together when trying to solve this.

jonhoo:
@najamelan that will be the behavior with https://github.com/rust-lang/rustup.rs/pull/1997 :)

kinnison:
I believe this is covered by Rustup 1.20

AaronKutch:
[...]

PR #1997, opened 2019-09-16
Update to most recent viable nightly

### LINKED ISSUE #1628: Feature request: "rustup update nightly" to the last version which has all necessary components.
Instead of failing or forcing, just use not-so-up-to-date version.

Additinoal feature request: `rustup update nightly --date=2019-01-29` to update nightly to a specific version while keeping it as "nigthly", not as a separate toolchain. Should also allow downgrades.

jonhoo:
Previously, if the user had components installed on `nightly` which were not available for the latest nightly, `rustup update` would not update the user's nightly at all. With this patch, rustup will try progressively older nightlies until it finds a nightly that supports all the components the user has installed for their current nightly.

Fixes #1628.
Makes progress towards #1501.
Fixes the underlying issue in #1676.

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -7,6 +7,7 @@ use crate::dist::temp;
 use crate::errors::*;
 use crate::utils::utils;
 
+use chrono::prelude::*;
 use lazy_static::lazy_static;
 use regex::Regex;
 
@@ -600,6 +601,77 @@ fn update_from_dist_<'a>(
     prefix: &InstallPrefix,
     force_update: bool,
 ) -> Result<Option<String>> {
+    let mut toolchain = toolchain.clone();
+    let mut fetched = String::new();
+    let mut first_err = None;
+    let backtrack = toolchain.channel == "nightly" && toolchain.date.is_none();
+    loop {
+        match try_update_from_dist_(
+            download,
+            update_hash,
+            &toolchain,
+            profile,
+            prefix,
+            force_update,
+            &mut fetched,
+        ) {
+            Ok(v) => break Ok(v),
+            Err(e) => {
+                if !backtrack {
+                    break Err(e);
+                }
+
+                if let ErrorKind::RequestedComponentsUnavailable(components, ..) = e.kind() {
+                    (download.notify_handler)(Notification::SkippingNightlyMissingComponent(
+                        components,
+                    ));
+
+                    if first_err.is_none() {
+                        first_err = Some(e);
+                    }
+                } else if let Some(e) = first_err {
+                    // if we fail to find a suitable nightly, we abort the search and give the
+                    // original "components unavailable for download" error.
+                    break Err(e);
+                } else {
+                    break Err(e);
+                }
+
+                // The user asked to update their nightly, but the latest nightly does not have all
+                // the components that the user currently has installed. Let's try the previous
+                // nightlies in reverse chronological order until we find a nightly that does,
+                // starting at one date earlier than the current manifest's date.
+                //
+                // NOTE: we don't need to explicitly check for the case where the next nightly to
+                // try is _older_ than the current nightly, since we know that the user's current
+                // nightlys supports the components they have installed, and thus would always
+                // terminate the search.
+                toolchain.date = Some(
+                    Utc.from_utc_date(
+                        &NaiveDate::parse_from_str(
+                            toolchain.date.as_ref().unwrap_or(&fetched),
+                            "%Y-%m-%d",
+                        )
+                        .expect("Malformed manifest date"),
+                    )
+                    .pred()
+                    .format("%Y-%m-%d")
+                    .to_string(),
+                );
+            }
+        }
+    }
+}
+
+fn try_update_from_dist_<'a>(
+    download: DownloadCfg<'a>,
+    update_hash: Option<&Path>,
+    toolchain: &ToolchainDesc,
+    profile: Option<Profile>,
+    prefix: &InstallPrefix,
+    force_update: bool,
+    fetched: &mut String,
+) -> Result<Option<String>> {
     let toolchain_str = toolchain.to_string();
     let manifestation = Manifestation::open(prefix.clone(), toolchain.target.clone())?;
 
@@ -621,6 +693,8 @@ fn update_from_dist_<'a>(
                 explicit_add_components: profile_components,
                 remove_components: Vec::new(),
             };
+
+            *fetched = m.date.clone();
 
             return match manifestation.update(
                 &m,
--- a/tests/cli-misc.rs
+++ b/tests/cli-misc.rs
@@ -770,6 +770,23 @@ fn update_nightly_even_with_incompat() {
 }
 
 #[test]
+fn nightly_backtrack_skips_missing() {
+    clitools::setup(Scenario::Unavailable, &|config| {
+        set_current_dist_date(config, "2015-01-01");
+        expect_ok(config, &["rustup", "default", "nightly"]);
+
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-n-1");
+
+        // nightly is missing on latest
+        set_current_dist_date(config, "2015-01-02");
+
+        // update should not change nightly, and should not error
+        expect_ok(config, &["rustup", "update", "nightly", "--no-self-update"]);
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-n-1");
+    });
+}
+
+#[test]
 fn completion_rustup() {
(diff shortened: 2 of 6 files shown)


jonhoo:
As an anecdotal observation, the code for mocking a `dist` in `tests/mock/clitools.rs ` should probably be refactor separately from this change. It's currently pretty messy, and would benefit from some cleanup, and probably some macros.

jonhoo:
Cleanup in https://github.com/rust-lang/rustup.rs/pull/1998

tests/cli-misc.rs
```diff
@@ -8,7 +8,6 @@ use crate::mock::clitools::{
     expect_ok_eq, expect_ok_ex, expect_stderr_ok, expect_stdout_ok, run, set_current_dist_date,
     this_host_triple, Config, Scenario,
 };
-use rustup::errors::TOOLSTATE_MSG;
```
najamelan, line 11: Just wondering about feedback. It makes sense that the old toolstate message isn't appropriate anymore, but is there a message that says something like: "Cannot update nightly channel because no newer nightlies were found that have all the required components: x, y z" and another like "Your nighty has been updated to version xxxx-xx-xx because more recent versions don't have the required components: x, y, z"?

jonhoo: So, you do get this info-level message:
https://github.com/rust-lang/rustup.rs/blob/e2da58a59bdf34b68fa7074d4c42f11cbc5a22f9/src/dist/notifications.rs#L164-L168
as rustup searches backwards in history, though maybe we want an additional message saying which nightly was chosen in the end? Although that said, that _should_ be clear from the final install message.

As for "Cannot update nightly channel because no newer nightlies were found that have all the required components: x, y z", the only way that could happen is if the nightly you are already on is the newest one that has all components. In that case, rustup will give the same message it currently gives if you try to update and there are no updates available. The info messages from above should make it clear why any newer nightlies weren't downloaded.

So, with that all said, do you still think it's necessary to add or change any messages?

Also, to be clear that one removal was due to a previously failing test now succeeding since an older nightly had the required component :)

najamelan: I didn't see that info when I looked through the diff, so I suppose that probably has enough information to avoid confusion about what's happening.  It's maybe not the most elegant way of reporting, but that's just cosmetics. It could generate quite some output if you haven't updated in say a month. So I suppose it's balancing having a polished message against the hurdle of implementing it, which I can't really estimate how much work it would be. 

If rustup prints the version actually installed, than I don't think we need an extra message for that as long as there is something that hints at versions being skipped.

To summarize, I can definitely live with this and there are tons of not-cosmetic issues that are more worthwhile putting effort in.

Thanks alot for taking this on anyway, it will be a major quality of ux upgrade in any case.

najamelan: actually, probably it's rare to see components fail for more than 2 weeks, so the amount of output is probably quite reasonable, even after not updating for a long time.

jonhoo: Yeah, I was about to say, you will only see that message as many times as there are nightlies since your component set last built successfully on nightly, no matter how old your nightly is. I'm also fairly sure that rustup prints which nightly was actually installed at the end, so we should be good on that end.

I agree with you that better cosmetics could be nice here, though I _think_ this gives us at least all the info, even if it's not presented in the cleanest way. My inclination would be to merge this and then do cosmetic fixes in follow-up work (maybe related to #1501).

norru:
This is ace. Thank you very much!

src/dist/dist.rs
```diff
@@ -599,6 +600,87 @@ fn update_from_dist_<'a>(
...
+                    // if we fail to find a suitable nightly, we abort the search and give the
+                    // original "components unavailable for download" error.
+                    break Err(e);
+                } else {
+                    break Err(e);
+                }
+
+                // The user asked to update their nightly, but the latest nightly does not have all
+                // the components that the user currently has installed. Let's try the previous
+                // nightly (recursively) until we find a nightly that does.
```
kinnison: Strictly speaking this is iterative not recursive, but I wouldn't block merge on that nitpick alone

src/dist/dist.rs
```diff
@@ -599,6 +600,87 @@ fn update_from_dist_<'a>(
...
+                //
+                // NOTE: we don't need to explicitly check for the case where the next nightly to
+                // try is _older_ than the current nightly, since we know that the user's current
+                // nightlys supports the components they have installed, and thus would always
+                // terminate the search.
+                if let Some(ref mut date) = toolchain.date {
```
kinnison: My gut feeling is that this could potentially be neater done as something like the following (untested):

```rust
toolchain.date = Some(Utc.from_utc_date(
    &toolchain.date.unwrap_or_else(
        ||NaiveDate::parse_from_str(&fetched, "%Y-%m-%d").expect("Malformed manifest date"))
    )).pred().format("%Y-%m-%d").to_string());
```

Obviously better formatting would be needed, and I've not checked the borrows vs. moves in that.

src/dist/dist.rs
```diff
@@ -599,6 +600,87 @@ fn update_from_dist_<'a>(
...
+            force_update,
+            &mut fetched,
+        ) {
+            Ok(v) => break Ok(v),
+            Err(e) => {
+                if !backtrack {
+                    break Err(e);
+                }
+
+                if let ErrorKind::RequestedComponentsUnavailable(components, ..) = e.kind() {
```
kinnison, line 624: What if we're now requesting a dated nightly which doesn't exist?  While it's rare, sometimes we don't succeed in building a nightly and so we can't provide one for a given date.

kinnison (changes requested):
Apart from my concern over missing nightlies and a possible bit of reformatting, this looks like a nice simple elegant solution to the problem.

src/dist/dist.rs
jonhoo: How does that manifest in the errors? I can't seem to find an error that maps cleanly to that idea. Basically, how does the manifest change if a nightly is unavailable?

mati865: Current nightly is unavailable so it's perfect occasion to test it :laughing:.

edit: Thanks @lzutao, GitHub completion sometimes messes up.

src/dist/dist.rs
jonhoo: Yeah, it was split up this way from back when I hadn't added `fetched`. Simplified it even further now :)

jonhoo: Yeah, it was split up this way from back when I hadn't added `fetched`. Simplified it even further now :)

src/dist/dist.rs
jonhoo: Actually, this turns out to already just work, since it manifests as the `rustc` component missing (and thus hits the existing case). I've added a test.

tesuji: Wrong person @mati865

kinnison (approved):
LGTM :+1:  I'll merge once I have a confirm from CI.

vi:
When will it be available for `rustup self update`?

kinnison:
> When will it be available for `rustup self update`?

@vi When we're ready to release 1.20.0 -- We have active tester requests out right now, and we may put further requests out on the way to 1.20.0.  Once we're good, we'll release.  I'd guess at a small (one or two) number of weeks.

Arnavion:
@jonhoo @kinnison How would I restore the original behavior? Do I have to `rustup component remove` the components I don't mind losing before `rustup update`ing, then `rustup component add` them back? Or is there an easier way?

najamelan:
@Arnavion one option is to check https://rust-lang-nursery.github.io/rust-toolstate/ manually and not update if not all your components are there...

jonhoo:
@Arnavion I'm not sure I follow? The old behavior was to simply _fail_ the update. If you mean to _force_ the update, then you should be able to pass `--force`.

Arnavion:
@jonhoo The new behavior, AFAICT from your change, is that `rustup update` will always succeed, but leave me with a nightly that is not the latest. The original behavior was that `rustup update` would fail, which my "update Rust" script would detect and handle by removing components that I don't need (clippy, rustfmt), re-running `rustup update`, then `rustup component add`ing (rustfmt) or `cargo install`ing (clippy, since it's usually fixed in git master even if the submodule in the rust repo is broken) the components back.

Yes, `--force` seems to be what I was asking for. Thanks.

EliSnow:
@jonhoo, I was reminded of this great work today when I did a `rustup update` and saw the message `skipping nightly which is missing installed component 'rls-preview'`. Thank you!

PR #1998, opened 2019-09-16
Tidy up the mock dist server

jonhoo:
This is follow-on clean-up work from #1997 which tidies up `tests::mock::clitools::create_mock_dist_server` so that it is much clearer which versions are available with each "`Scenario`". I will mark this as ready to review and rebase once #1997 lands.

### DIFF
--- a/tests/cli-exact.rs
+++ b/tests/cli-exact.rs
@@ -26,13 +26,13 @@ fn update() {
             &["rustup", "update", "nightly", "--no-self-update"],
             for_host!(
                 r"
-  nightly-{0} installed - 1.3.0 (hash-n-2)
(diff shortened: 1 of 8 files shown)


jonhoo:
I believe the failing Windows test also fails on `master` fwiw.

kinnison:
> I believe the failing Windows test also fails on `master` fwiw.

Travis' Windows testing is still in beta (and we'll probably drop it) so I don't predicate merging on that passing so long as appveyor is good.

kinnison (approved):
This is very nice.  I particularly like the proper `Release` struct with its associated methods.

Thank you. :+1:

PR #2002, opened 2019-09-18
Correctly handle update across missing nightlies

### LINKED ISSUE #2001: rustup should also skip the day which is no release when updating nightly
**Problem**
I'm using version `rustup 1.19.0+16 (279e802a0 2019-09-17)`.
With `clippy` installed,  `rustup update nightly` should do backtrace and found a viable version at `2019-09-13`.  But as showing in [`componets history`](https://rust-lang.github.io/rustup-components-history/), there is no release at `2019-09-17`. Then `rustup` fail with  `componet unavailable`.

image

**Steps**

1. Clone and install from current master (279e802a0ad4a9eafd89b3740c87d7f58642439d)
2. Run `rustup update nightly`

**Possible Solution(s)**
 
I have add a `dbg!` and it show me this when check `2019-09-17`:

```plain
[src/dist/dist.rs:623] &e = Error(
    Msg(
        "no release found for \'nightly-2019-09-17\'",
    ),
    State {
        next_error: None,
        backtrace: InternalBacktrace {
            backtrace: None,
        },
    },
)
```

I think we should add a specific error for `no release` and also skip it when doing backtrace.

**Notes**

Output of `rustup --version`: `rustup 1.19.0+16 (279e802a0 2019-09-17)`
Output of `rustup show`:

jonhoo:
Fixes #2001.

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -564,6 +564,7 @@ pub fn update_from_dist<'a>(
     profile: Option<Profile>,
     prefix: &InstallPrefix,
     force_update: bool,
+    old_date: Option<&str>,
 ) -> Result<Option<String>> {
     let fresh_install = !prefix.path().exists();
     let hash_exists = update_hash.map(Path::exists).unwrap_or(false);
@@ -582,6 +583,7 @@ pub fn update_from_dist<'a>(
         profile,
         prefix,
         force_update,
+        old_date,
     );
 
     // Don't leave behind an empty / broken installation directory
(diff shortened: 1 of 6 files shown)


kinnison (approved):
I can confirm this works where it did not previously.  :+1:

kinnison:
I've asked the tests to re-try

jonhoo:
I pushed a change now that limit the backtracking explicitly to the date of the currently installed nightly, or alternatively the date of the first manifest ever (though I don't think that case is ever hit).

PR #2012, opened 2019-09-23
Simplify Notification::RemovingComponent invocation

### LINKED ISSUE #1993: Removing previous version of component message does not appear when updating
**Problem**

#1283 introduced a new message that is printed to the screen whenever we replace a component with a new one.

However, when running `rustup update` the `info: removing component` message still appears instead of `info: removing old component`.

Strangely enough, the [relevant tests pass](https://github.com/rust-lang/rustup.rs/pull/1639/files#diff-5bc536e3b333bb85e5fb398d8c29ec63R52). 

**Steps**

1. rustup update

**Possible Solution(s)**

It is possible that the patch in #1639 has overlooked some minor detail.
We should inspect it and see what we missed.

**Notes**

Output of `rustup --version`: rustup 1.19.0 (2af131cf9 2019-09-08)
Output of `rustup show`:
```
rustup home:  /home/omer/.rustup

installed toolchains
--------------------

nightly-2019-04-12-x86_64-unknown-linux-gnu

active toolchain
----------------

rustc 1.39.0-nightly (eb48d6bde 2019-09-12)
```

BeniCheni:
Fixes #1993.

After analyzing #1993 that led to #1283 & the source of #1639, the hypothesis is to simplify to just `Notification::RemovingComponent` invocation in "src/dist/manifestation.rs" and "src/dist/notification.rs", based on the understanding of the goal is to ultimately just display:

> info: removing previous version of component ***

By such simplification, also able to delete unneeded `RemovingOldComponent` code. Looking forward to review feedback & mentioning to update the PR, if my analysis or hypothesis is correctly.

P.S: I also tried to understand this comment from the 1st comment in #1993:

> Strangely enough, the relevant tests pass.

https://github.com/rust-lang/rustup.rs/blob/489fa585117dc0ac283f247f3c9a829fd32442af/src/dist/manifestation.rs#L133

I'm speculating that the `altered` variable checks for default dist server, and guessing that when tests run, the local env would somehow be considered truthy for the `altered` flag to call original `RemovingOldComponent` to land to the output code for "info: removing old component", and the tests would pass/

While running the "_prod dist_" build, the `altered` flag was considered falsy, and it would call the original `RemovingComponent` to land to the output code for "info: removing component".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,6 +103,7 @@ matrix:
       env: TARGET=x86_64-pc-windows-msvc NO_ADD=1 EXE_EXT=.exe
       services: nil
       language: bash
+      if: branch == disabled-for-now
 
 install:
   - sh rustup-init.sh --default-toolchain=stable -y
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,14 +7,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "aho-corasick"
-version = "0.5.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "memchr 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
-name = "aho-corasick"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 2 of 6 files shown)


thedrow (changes requested):
Thank you for working on this.

The reason for adding `RemovingOldComponent` was because you can also `rustup component remove`.
In this case, you are not removing an old component to replace it with a new one. You are simply removing it since you no longer need it.
I don't think that this behaviour is correct from a UX perspective.

BeniCheni:
Thanks, @thedrow. Will take a look and correct.

BeniCheni:
Will open a new PR.

BeniCheni:
Hi, @thedrow , created a new PR (#2016) w. correction. PTAL over in the new PR? Thanks!

PR #2018, opened 2019-09-27
Use Self when possible

tesuji:
By specifying `-W clipppy::use_self` with clippy.


kinnison (approved):
These all look sane. :+1:

PR #2020, opened 2019-09-27
Update normalize.css

tesuji:


kinnison (approved):
I won't pretend to be able to review the CSS, but will trust you've at least looked over the result

PR #1788, opened 2019-04-22
Kinnison/fix progress on windows

kinnison:
Following on from testing the 1.18.0 RC on Windows, this is needed to fix line erase

PR #2021, opened 2019-09-27
Bump term

tesuji:
An attempt to bump term dependency to v0.6 from v0.5.
It would be nice if we could have `@Skynoodle` test it.

cc #1788


Skynoodle:
Unfortunately, this brings back the garbled progress indicators on CMD and Powershell :(

image

tesuji:
Thanks @Skynoodle . I guess I would close this PR then.

PR #2025, opened 2019-09-27
build: Use uname to determine OS

tesuji:
`ci/run.bash` could run in docker environments, so there are no
TRAVIS_OS_NAME env.


kinnison (approved):
Okay, though Windows on Travis wasn't dockering.

Issue #2005, opened 2019-09-19
Support selecting additional components (and targets) at install time


kinnison:
When installing a toolchain, it might be nice to select additional components to install such that they will be immediately available.

For example, editors might auto-install toolchains if needed, and request `rls` `rust-analyzer` `rust-src` etc. without having to override the chosen profile from the user.

In addition, if you know you're going to need a particular target, it'd be nice to preinstall with that too (e.g. if I download some embedded code and know I'll need the `thumbv6m-none-eabi` target)

Perhaps something like:

`rustup toolchain install nightly -c rls -c clippy -t thumbv6m-none-eabi`

jonhoo:
Note to self to re-enable `clippy` in https://github.com/crate-ci/azure-pipelines/pull/59 once this lands.

kinnison:
This has landed and is in 1.20

PR #2026, opened 2019-09-27
Support installation with additional components and targets

kinnison:
This is meant to address #2005 though it'd be nice to get @jonhoo 's feedback because I'm far too tired to review my own stuff for sanity right now.


src/dist/dist.rs
```diff
@@ -717,11 +727,48 @@ fn try_update_from_dist_<'a>(
...
+                    not_found_components.push(Component::new(
+                        "rust-std".to_owned(),
+                        Some(triple.clone()),
+                        false,
+                    ));
+                }
+            }
+
+            if !not_found_components.is_empty() {
+                return Err(Error::from(ErrorKind::RequestedComponentsUnavailable(
```
jonhoo: I think this will mean that if you misspell a component, it will keep walking back nightlies all the way to 2014. Not sure how we avoid this, except perhaps to re-add a cutoff point for how far back we'll search when doing a fresh install?

src/dist/dist.rs
```diff
@@ -717,11 +727,48 @@ fn try_update_from_dist_<'a>(
...
+                } else {
+                    use crate::dist::manifest::Component;
+                    not_found_components.push(Component::new(component.to_string(), None, false));
+                }
+            }
+
+            for target in targets {
+                let triple = TargetTriple::new(target);
+                if let Some(component) =
+                    m.find_targeted_component("rust-std", &toolchain.target, &triple)?
```
jonhoo: Is `rustc` sufficient, or do we really want `rust-std`?

src/dist/dist.rs
```diff
@@ -717,11 +727,48 @@ fn try_update_from_dist_<'a>(
...
+            let mut changes = Changes {
                 explicit_add_components: profile_components,
                 remove_components: Vec::new(),
             };
 
+            let mut not_found_components = Vec::new();
+
+            for component in components {
+                if let Some(component) =
+                    m.find_targeted_component(component, &toolchain.target, &toolchain.target)?
```
jonhoo: Is it necessary to do the search here? Isn't it sufficient to add these to `changes`, and then `manifestation.update` will do the search (and yield the error)?

src/install.rs
```diff
@@ -74,6 +80,8 @@ impl<'a> InstallMethod<'a> {
                     prefix,
                     force_update,
                     old_date,
+                    if !exists { components } else { &[] },
```
jonhoo: So this means that we'll ignore `--component` and `--target` if passed to `rustup update`? Do we also give an error for that case, or will they just be silently ignored?

tests/cli-self-upd.rs
```diff
@@ -1258,3 +1258,27 @@ fn update_installs_clippy_cargo_and() {
...
+                "rls",
+                "-t",
+                clitools::CROSS_ARCH1,
+            ],
+        );
+        expect_stdout_ok(
+            config,
+            &["rustup", "component", "list"],
+            &format!("rust-std-{} (installed)", clitools::CROSS_ARCH1),
+        );
```
jonhoo, line 1287: This should also check for `rls`

tests/cli-v2.rs
```diff
@@ -1109,3 +1109,24 @@ fn target_list_ignores_unavailable_targets() {
...
+                "toolchain",
+                "install",
+                "nightly",
+                "-c",
+                "clippy",
+                "-t",
+                clitools::CROSS_ARCH1,
+                "--no-self-update",
+            ],
+        );
```
jonhoo, line 1130: Should this test for thee presence of the component and target afterwards?

src/dist/dist.rs
kinnison: So, when I tried with `-c foo` I got:

```
info: latest update on 2019-09-27, rust version 1.40.0-nightly (ddf43867a 2019-09-26)
info: skipping nightly which is missing installed component 'foo'
thread 'main' panicked at 'Malformed manifest date: ParseError(TooShort)', src/libcore/result.rs:1084:5
```
I think this is because the badly named component caused an early exit,  as such perhaps it'd be easiest to just break the loop if `fetched` is empty after an attempt?

kinnison: Having tried that, I think it'll be right, so I'll amend for that:

```
info: latest update on 2019-09-27, rust version 1.40.0-nightly (ddf43867a 2019-09-26)
error: component 'foo' is unavailable for download for channel x86_64-unknown-linux-gnu
```

being the result.

src/dist/dist.rs
kinnison: A target *is* `rust-std` for the given target, installed in the `host` toolchain.

src/dist/dist.rs
kinnison: Hmm, that's possible, yes.  I'll give it a go in a moment.

src/install.rs
kinnison: Only being for `install` was the intent, though I guess I could drop it back to allowing it for `update` too -- do you think that'd be useful?  If not, I'll add some kind of error/warning.

tests/cli-self-upd.rs
kinnison: Agreed

tests/cli-v2.rs
kinnison: Yeah I think it probably should.

src/install.rs
jonhoo: No, I think keeping it just for `install` is completely reasonable. Was just wondering what would happen if I tried to pass it to `update`. If I get an error, that's excellent.

kinnison: I decided update was also useful

src/dist/dist.rs
kinnison: I think this is a good approach and simplifies the change a good bit -- I'm very sad at how much I overcomplicate things.

src/dist/dist.rs
```diff
@@ -642,6 +650,10 @@ fn update_from_dist_<'a>(
                 }
 
                 if let ErrorKind::RequestedComponentsUnavailable(components, ..) = e.kind() {
+                    if fetched.is_empty() {
+                        // We failed to fetch, perhaps because the components are bad, stop now.
```
jonhoo: I'm confused as to how this check is correct. `fetched` is set before `manifestation.update` is even called, so checking whether it is empty can't actually check whether components were available. It should only be empty if the entire manifest failed to download, but that sounds like a case where we *do* want to continue iterating backwards?

src/dist/dist.rs
```diff
@@ -701,14 +713,24 @@ fn try_update_from_dist_<'a>(
...
 ) -> Result<Option<String>> {
     let toolchain_str = toolchain.to_string();
     let manifestation = Manifestation::open(prefix.clone(), toolchain.target.clone())?;
 
     // TODO: Add a notification about which manifest version is going to be used
     (download.notify_handler)(Notification::DownloadingManifest(&toolchain_str));
-    match dl_v2_manifest(download, update_hash, toolchain) {
+    match dl_v2_manifest(
+        download,
+        if components.is_empty() && targets.is_empty() {
```
jonhoo, line 753: Why is this `if` necessary?

src/dist/dist.rs
kinnison: I *think* this might be a bogon from a fiddle I was doing, I'll check, but I think it'll never fire.

src/dist/dist.rs
kinnison: If the toolchain exists already, and we're not explicitly trying to add components/targets then we want to skip any download if the hashes match -- this `if` blocks that short-circuit in the case that we have explicit things to add.  Without it we won't actually enter the manifestation update

jonhoo: Hmm... Isn't `update_hash` only `Some` if we're doing an update, in which case `components` and `targets` will both be empty, or am I missing something?

kinnison: I decided that it was better to allow components/targets on update too, hence I needed to elide update_hash in that case.

jonhoo: I guess I'm still not following how setting the `update_hash` to `None` will make a difference. Presumably it is only used to determine if a download is necessary in the first place (i.e., so that if you call `update` and the hash matches the latest manifest, it doesn't download anything). If that is the case, then setting it to `None` will only enable downloading an _older_ manifest (newer manifests will always be downloaded regardless). But `update` should never (at least in my opinion) "update" to an older nightly, even if I'm asking for a particular component or target. Beyond that, I also think that wouldn't even happen, since the backtracking code stops backtracking the moment you go beyond the current manifest date (which you would immediately).

kinnison: I believe you're correct in terms of no-backtracking as per the current algorithm.  However I think you've misunderstood the short-circuiting I was trying to describe.  In the current situation, `update`ing a channel runs thusly.

1. Download the channel manifest hash
2. *If* there is an existing hash (`update_hash` is `Some(Path)`) then compare it against the stored hash, on matching we exit "Ok, Unchanged"
3. Download the channel manifest, hash it, compare it to the downloaded hash.  If no match, we exit "Bad manifest download"
4. Continue the update with the new validated channel manifest.

What the change we are discussing does is skip 2 in the case that we're explicitly trying to add components/targets to the channel during the update.

Does that settle your concerns?  Is there something you feel I should do to make that clearer?

jonhoo: Ah, I think I understand now — the `if` is there so that if the newest nightly hasn't changed, but the user wishes to add components, we do not exit early. At which point we essentially degenerate to `rustup component add`. Okay, that seems fine. I think I'd like to see a comment there saying

> Even if manifest has not changed, we must continue to install requested components.

src/dist/dist.rs
```diff
@@ -657,6 +666,8 @@ fn update_from_dist_<'a>(
                     if first_err.is_none() {
                         first_err = Some(e);
                     }
+                    // We decrement the backtrack count only on unavailable component errors
```
jonhoo, line 673: I think it'd be good to have this comment explain _why_ we do that. In particular

> so that the limit only applies to nightlies that were indeed available, and ignores missing ones.

src/dist/dist.rs
```diff
@@ -614,6 +614,15 @@ fn update_from_dist_<'a>(
     let mut fetched = String::new();
     let mut first_err = None;
     let backtrack = toolchain.channel == "nightly" && toolchain.date.is_none();
+    // We want to limit backtracking if we do not already have a toolchain
+    let mut backtrack_limit: Option<i32> = if old_date.is_some() {
+        None
+    } else {
+        // The limit of 20 here is an arbitrary selection, let the user override it.
+        let provided = env::var("RUSTUP_BACKTRACK_LIMIT").unwrap_or_else(|_| "20".to_owned());
```
jonhoo: This is a super-nit, but I'd be inclined to set the default to either a full release cycle (6w) or a half release cycle (3w). Still arbitrary, but feels a little less arbitrary than 20. We're essentially here trading "chance of success" against "time to get an error if you misspell a component", where I think the chance of success has diminishing returns at anywhere over 2w.

kinnison: Since 20 is close to 3w, I've updated it to 21 and state it's half a release cycle.

PR #2031, opened 2019-09-30
Support toolchain keyword (e.g. "+nightly") to set override at root command

### LINKED ISSUE #1498: Allow `+toolchain` for rustup
It was surprising to me that I couldn't use the `+nightly` parameter on rustup for adding the `wasm32-unknown-unknown` target to nightly:
```
rustup +nightly target add wasm32-unknown-unknown
```
(and other combinations didn't work either). Instead I had to find the `--toolchain` parameter of `rustup target add`. It would be nice if `+nightly` would work on `rustup`, too.

BeniCheni:
Support the toolchain keyword (e.g "+nightly") to set override of the current working directory for toolchain, at the root command.

This PR should resolve #1498, as the local test is using similar test case to the reported command from the issue:

##### Test w. `rustup +nightly target add wasm32-unknown-unknown`
```
→ rustup +nightly target add wasm32-unknown-unknown
info: using existing install for 'nightly-x86_64-apple-darwin'
info: override toolchain for '/Users/bchen/github/rustup.rs' set to 'nightly-x86_64-apple-darwin'

  nightly-x86_64-apple-darwin unchanged - rustc 1.40.0-nightly (702b45e40 2019-10-01)

info: downloading component 'rust-std' for 'wasm32-unknown-unknown'
 10.9 MiB /  10.9 MiB (100 %)   7.7 MiB/s in  3s ETA:  0s
info: installing component 'rust-std' for 'wasm32-unknown-unknown'
 10.9 MiB /  10.9 MiB (100 %)  10.0 MiB/s in  1s ETA:  0s
```

##### Confirm that the target is added in my rustup toolchain nightly folder

```
→ ls -la ~/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/
total 0
drwxr-xr-x   3 bchen  CORP\Domain Users   96 Oct  6 21:54 .
drwxr-xr-x  19 bchen  CORP\Domain Users  608 Oct  6 21:54 ..
drwxr-xr-x  24 bchen  CORP\Domain Users  768 Oct  6 21:54 lib

```

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -35,6 +35,10 @@ pub fn main() -> Result<()> {
     let quiet = matches.is_present("quiet");
     let cfg = &mut common::set_globals(verbose, quiet)?;
 
+    if let Some(t) = matches.value_of("+toolchain") {
+        cfg.set_toolchain_override(&t[1..]);
+    }
+
     if maybe_upgrade_data(cfg, &matches)? {
         return Ok(());
     }
@@ -131,10 +135,15 @@ pub fn cli() -> App<'static, 'static> {
                 .long("quiet"),
         )
         .arg(
-            Arg::with_name("toolchain")
-                .help(TOOLCHAIN_ARG_HELP)
-                .long("toolchain")
-                .takes_value(true),
+            Arg::with_name("+toolchain")
+                .help("release channel (e.g. +stable) or custom toolchain to set override")
+                .validator(|s| {
+                    if s.starts_with('+') {
+                        Ok(())
+                    } else {
+                        Err("Toolchain overrides must begin with '+'".into())
+                    }
+                }),
         )
         .subcommand(
             SubCommand::with_name("dump-testament")
@@ -852,8 +861,7 @@ fn run(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
 
 fn which(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
     let binary = m.value_of("command").expect("");
-    let toolchain_provided = m.is_present("toolchain");
-    let binary_path = if toolchain_provided {
+    let binary_path = if m.is_present("toolchain") {
         let toolchain = m.value_of("toolchain").expect("");
         cfg.which_binary_by_toolchain(toolchain, binary)?
             .expect("binary not found")
--- a/src/config.rs
+++ b/src/config.rs
@@ -17,6 +17,7 @@ use crate::utils::utils;
 #[derive(Debug)]
 pub enum OverrideReason {
     Environment,
+    CommandLine,
     OverrideDB(PathBuf),
     ToolchainFile(PathBuf),
 }
@@ -25,6 +26,7 @@ impl Display for OverrideReason {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::result::Result<(), fmt::Error> {
         match self {
             Self::Environment => write!(f, "environment override by RUSTUP_TOOLCHAIN"),
+            Self::CommandLine => write!(f, "overridden by +toolchain on the command line"),
             Self::OverrideDB(path) => write!(f, "directory override for '{}'", path.display()),
             Self::ToolchainFile(path) => write!(f, "overridden by '{}'", path.display()),
         }
@@ -40,6 +42,7 @@ pub struct Cfg {
     pub download_dir: PathBuf,
     pub temp_cfg: temp::Cfg,
     pub gpg_key: Cow<'static, str>,
+    pub toolchain_override: Option<String>,
     pub env_override: Option<String>,
     pub dist_root_url: String,
     pub dist_root_server: String,
@@ -104,6 +107,7 @@ impl Cfg {
             temp_cfg,
             gpg_key,
             notify_handler,
+            toolchain_override: None,
             env_override,
             dist_root_url: dist_root,
             dist_root_server,
@@ -142,6 +146,10 @@ impl Cfg {
         })?;
         (self.notify_handler)(Notification::SetProfile(profile));
         Ok(())
+    }
+
+    pub fn set_toolchain_override(&mut self, toolchain_override: &str) {
+        self.toolchain_override = Some(toolchain_override.to_owned());
     }
 
     // Returns a profile, if one exists in the settings file.
@@ -281,7 +289,12 @@ impl Cfg {
     pub fn find_override(&self, path: &Path) -> Result<Option<(Toolchain<'_>, OverrideReason)>> {
         let mut override_ = None;
 
-        // First check RUSTUP_TOOLCHAIN
+        // First check toolchain override from command
+        if let Some(ref name) = self.toolchain_override {
+            override_ = Some((name.to_string(), OverrideReason::CommandLine));
+        }
+
+        // Check RUSTUP_TOOLCHAIN
         if let Some(ref name) = self.env_override {
             override_ = Some((name.to_string(), OverrideReason::Environment));
         }
@@ -304,6 +317,10 @@ impl Cfg {
             let reason_err = match reason {
                 OverrideReason::Environment => {
                     "the RUSTUP_TOOLCHAIN environment variable specifies an uninstalled toolchain"
+                        .to_string()
+                }
+                OverrideReason::CommandLine => {
+                    "the +toolchain on the command line specifies an uninstalled toolchain"
                         .to_string()
                 }
                 OverrideReason::OverrideDB(ref path) => format!(
--- a/tests/cli-misc.rs
+++ b/tests/cli-misc.rs
@@ -905,3 +905,79 @@ fn which() {
         );
     });
 }
+
+#[test]
+fn override_by_toolchain_on_the_command_line() {
+    setup(&|config| {
+        #[cfg(windows)]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+stable", "which", "rustc"],
+            "\\toolchains\\stable-x86_64-",
+        );
+        #[cfg(windows)]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+stable", "which", "rustc"],
+            "\\bin\\rustc",
+        );
+        #[cfg(not(windows))]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+stable", "which", "rustc"],
+            "/toolchains/stable-x86_64-",
+        );
+        #[cfg(not(windows))]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+stable", "which", "rustc"],
+            "/bin/rustc",
+        );
+        expect_ok(config, &["rustup", "default", "nightly"]);
+        #[cfg(windows)]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+nightly", "which", "rustc"],
+            "\\toolchains\\nightly-x86_64-",
+        );
+        #[cfg(windows)]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+nightly", "which", "rustc"],
+            "\\bin\\rustc",
+        );
+        #[cfg(not(windows))]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+nightly", "which", "rustc"],
+            "/toolchains/nightly-x86_64-",
+        );
+        #[cfg(not(windows))]
+        expect_stdout_ok(
+            config,
+            &["rustup", "+nightly", "which", "rustc"],
+            "/bin/rustc",
+        );
+        expect_stdout_ok(
+            config,
+            &["rustup", "+nightly", "show"],
+            "(overridden by +toolchain on the command line)",
+        );
+        expect_err(
+            config,
+            &["rustup", "+foo", "which", "rustc"],
+            "toolchain 'foo' is not installed",
+        );
+        expect_err(
+            config,
+            &["rustup", "@stable", "which", "rustc"],
+            "Invalid value for '<+toolchain>': Toolchain overrides must begin with '+'",
+        );
+        expect_stderr_ok(
+            config,
+            &["rustup", "+stable", "set", "profile", "minimal"],
+            "profile set to 'minimal'",
+        );
+        expect_stdout_ok(config, &["rustup", "default"], "nightly-x86_64-");
+    });
+}


src/cli/rustup_mode.rs
```diff
@@ -817,16 +818,20 @@ fn run(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
-        cfg.which_binary(&utils::current_dir()?, binary)?
-            .expect("binary not found")
+        if m.is_present("toolchain") {
+            let toolchain = m.value_of("toolchain").expect("");
+            cfg.which_binary_by_toolchain(toolchain, binary)?
+                .expect("binary not found")
+        } else {
+            cfg.which_binary(&utils::current_dir()?, binary)?
+                .expect("binary not found")
+        }
```
kinnison: So this behaviour means that if the user says `rustup --toolchain something which rustc --toolchain another` we silently ignore the `--another`  Would it be worth warning the user?  (If you don't think so, I'll accept it)

pickfire: I believe there should be a warning, especially if `rustup +nightly something which rustc --toolchain beta`.

BeniCheni:
Note: [0aacf5f](https://github.com/rust-lang/rustup.rs/pull/2031/commits/0aacf5fdc3ec6efbb0625e793d16e08e0bd4ea12) updated these:

- added support for release channel support (i.e. "nightly")

```
→ rustup +stable which cargo
/Users/bchen/.rustup/toolchains/stable-x86_64-apple-darwin/bin/cargo

→ rustup +nightly which cargo                 
/Users/bchen/.rustup/toolchains/nightly-x86_64-apple-darwin/bin/cargo
```

- simplified some code based on `clippy`'s suggestions

> Would it be worth warning the user?

> I believe there should be a warning

- added a warning, when both release channel arg & toolchain option are provided; a more _extreme_ test: (happy to edit the warning message to make better UX, if any suggestion)

```
→ rustup --toolchain=nightly +stable which cargo --toolchain beta
warning: Both the <release-channel> arg (e.g. "+nightly") and the "--toolchain" option are provided.
The <release-channel> arg would take precedence over the "--toolchain" option.
/Users/bchen/.rustup/toolchains/stable-x86_64-apple-darwin/bin/cargo
```

@kinnison / @pickfire , PTAL?

BeniCheni:
Note: [97e6043](https://github.com/rust-lang/rustup.rs/commit/97e6043983b64e0c7c6664021794d000d371676f) updated these:

- added support for release channel support (i.e. "nightly")

```
→ rustup +stable which cargo
/Users/bchen/.rustup/toolchains/stable-x86_64-apple-darwin/bin/cargo

→ rustup +nightly which cargo                 
/Users/bchen/.rustup/toolchains/nightly-x86_64-apple-darwin/bin/cargo

→ rustup +foo which cargo                 
error: '+foo' isn't a valid value for '<release-channel>'
	[possible values: +beta, +nightly, +stable]


USAGE:
    rustup [FLAGS] [OPTIONS] [release-channel] <SUBCOMMAND>

For more information try --help
```

- simplified some code based on `clippy`'s suggestions

- added tests

Per:
> I believe there should be a warning

> Would it be worth warning the user?

- ~added warnings: (happy to edit the warning message to make better UX, if any suggestion)~

~→ rustup +stable --toolchain=nightly which cargo           
warning: Both toolchain keyword (e.g. "+nightly") and "--toolchain" option are provided. The toolchain keyword would take precedence.
/Users/bchen/.rustup/toolchains/stable-x86_64-apple-darwin/bin/cargo~

~→ rustup --toolchain stable which rustc --toolchain=nightly
warning: Toolchain options are provided before and after the "which" subcommand. The first toolchain option would take precedence.
/Users/bchen/.rustup/toolchains/stable-x86_64-apple-darwin/bin/rustc~

Update: current version does not need warning anymore.

src/cli/rustup_mode.rs
```diff
@@ -135,6 +141,12 @@ pub fn cli() -> App<'static, 'static> {
                 .long("toolchain")
                 .takes_value(true),
         )
+        .arg(
+            Arg::with_name("release-channel")
+                .help("provide the release channel (e.g. +stable) for toolchain")
+                .index(1)
+                .possible_values(&["+stable", "+nightly", "+beta"]),
```
kinnison: The problem with this is that it doesn't support specific toolchain names, or custom linked toolchains.  This approach simply won't scale.

It may be better to transform the `+foo` into `--toolchain foo` if we can do that and still have clap parse the mutated CLI.

kinnison (changes requested):

BeniCheni:
> It may be better to transform the +foo into --toolchain foo ...

@kinnison, sounds good. Updated per suggestion. PTAL at convenience?  Thanks.

src/cli/rustup_mode.rs
```diff
@@ -130,11 +140,15 @@ pub fn cli() -> App<'static, 'static> {
                 .long("quiet"),
         )
         .arg(
-            Arg::with_name("toolchain")
+            Arg::with_name("toolchain-option")
```
pickfire: Why not just override `toolchain` instead of adding both `toolchain-option` and `toolchain-keyword`? There are supposed to be only one toolchain in effect though. I wonder if it would be better to just mutate `toolchain` if `+toolchain` is found without `--toolchain` specified, otherwise it should error out.

BeniCheni:
> Why not just override toolchain instead of adding both toolchain-option and toolchain-keyword? There are supposed to be only one toolchain in effect though.

@pickfire, there's indeed only one toolchain (either by keyword or option format) in effect. Though the previous "toolchain-*" names were probably misleading. Updated to, hopefully, less confusing names (`toolchain` & `keyword`).

> It may be better to transform the +foo into --toolchain foo ..._"

The current logic is to follow @kinnison's suggestion ☝️ . I think the assignment of `  let toolchain = if let ... else` block in "src/cli/rustup_mode.rs" already transforms "+nightly"  into toolchain.

If there's a way to omit `Arg::with_name("keyword")` setup, but still able to take +foo input from CLI, I'm _game_ to edit. Currently, I couldn't figure out any version how to receive +foo input without the "keyword" arg.

src/cli/rustup_mode.rs
```diff
@@ -79,7 +92,7 @@ pub fn main() -> Result<()> {
             (_, _) => unreachable!(),
         },
         ("run", Some(m)) => run(cfg, m)?,
-        ("which", Some(m)) => which(cfg, m)?,
+        ("which", Some(m)) => which(cfg, m, toolchain)?,
```
pickfire: Why not just override `toolchain` in `cfg` and pass that in instead? We can keep the current version that way without adding a new parameter.

kinnison: That is a very good suggestion.  If the config toolchain gets reset then all commands will benefit since it's effectively altering the default toolchain entry from the config file.  One thing which will have to be checked if we do that is that we don't end up saving the different toolchain back if we do something (not expected, but possible) along the lines of `rustup +another set profile default`  -> alters both `profile` and `default_toolchain` in `~/.rustup/settings.toml`

BeniCheni:
> pickfire
> Why not just override toolchain in cfg and pass that in instead? We can keep the current version that way without adding a new parameter.

> kinnison
> That is a very good suggestion
> then all commands will benefit since it's effectively altering the default toolchain entry from the config file.

@pickfire / @kinnison, after taking a step back and rethinking the approach, I also agree that overriding toolchain in cfg, and pass would have broader beneficial scope than just `which`. So the last push updated to just take "keyword" arg, and set override for the current dir accordingly.

I've tested w. the reported command from #1498: `rustup +nightly target add wasm32-unknown-unknown`, and am able to verify the "+nightly" keyword added the target in the correct rustup nightly toolchain path. Therefore, I think the PR is also resolving #1498 and edited the first comment of the PR w. reference. PTAL?

```
→ rustup +nightly target add wasm32-unknown-unknown
info: using existing install for 'nightly-x86_64-apple-darwin'
info: override toolchain for '/Users/bchen/github/rustup.rs' set to 'nightly-x86_64-apple-darwin'

  nightly-x86_64-apple-darwin unchanged - rustc 1.40.0-nightly (702b45e40 2019-10-01)

info: downloading component 'rust-std' for 'wasm32-unknown-unknown'
 10.9 MiB /  10.9 MiB (100 %)   7.7 MiB/s in  3s ETA:  0s
info: installing component 'rust-std' for 'wasm32-unknown-unknown'
 10.9 MiB /  10.9 MiB (100 %)  10.0 MiB/s in  1s ETA:  0s

→ ls -la ~/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/
total 0
drwxr-xr-x   3 bchen  CORP\Domain Users   96 Oct  6 21:54 .
drwxr-xr-x  19 bchen  CORP\Domain Users  608 Oct  6 21:54 ..
drwxr-xr-x  24 bchen  CORP\Domain Users  768 Oct  6 21:54 lib
```

tests/cli-misc.rs
```diff
@@ -866,8 +866,8 @@ fn add_remove_component() {
 #[test]
 fn which() {
     setup(&|config| {
-        let path_1 = config.customdir.join("custom-1");
```
pickfire: I wonder why are these changes necessary?

src/cli/rustup_mode.rs
pickfire: @kinnison Ah, I did not thought about that. Maybe that should be added to the tests as well.

@BeniCheni Wow, fast.

src/cli/rustup_mode.rs
```diff
@@ -130,10 +134,8 @@ pub fn cli() -> App<'static, 'static> {
                 .long("quiet"),
         )
         .arg(
-            Arg::with_name("toolchain")
-                .help(TOOLCHAIN_ARG_HELP)
-                .long("toolchain")
-                .takes_value(true),
+            Arg::with_name("keyword")
```
pickfire: I wonder if it would be helpful for the user experience if we use `+keyword` here instead.

kinnison: Better would be `toolchain` rather than `keyword`

tests/cli-misc.rs
kinnison: Certainly this additional leading underscore isn't needed.  However the conversion is needed afaict.

BeniCheni:
@pickfire / @kinnison, updated 👇 per previous comments, and did some test about not saving settings in default profile. PTAL at convenience? (🤞feel really close to LGTY 😺)

<details><summary>Per tests/cli-misc.rs</summary>
<p>

> pickfire
> I wonder why are these changes necessary?

> kinnison
> Certainly this additional leading underscore isn't needed. However the conversion is needed afaict.

Indeed the leading underscore was my overlook. Updated. Also indeed the custom-1 to custom-2 conversion is to support the custom toolchain in the `which` test case.

</p>
</details>

---

<details><summary>Per src/cli/rustup_mode.rs</summary>
<p>

> pickfire

> I wonder if it would be helpful for the user experience if we use `+keyword` here instead.

> kinnison
> Better would be `toolchain` rather than `keyword`

Tried out `toolchain`, but collided with subcommand `toolchain` and panicked. No worries. Ended using a common-ground name `+toolchain`. Think it's contextual since the code strips the 1st character (i.e. "+") off to apply toolchain value. Hope it's okay w. you guys.

</p>
</details>





---

<details><summary>From an earlier comment</summary>
<p>

> ... we don't end up saving the different toolchain back if we do something (not expected, but possible) along the lines of rustup +another set profile default -> alters both profile and default_toolchain in ~/.rustup/settings.toml

Tested `rustup +stable set profile default` and verified content of settings.toml to ensure the default_toolchain is still nightly in the default profile.

```
→ rustup +stable set profile default
info: using existing install for 'stable-x86_64-apple-darwin'
info: override toolchain for '/Users/bchen/github/rustup.rs' set to 'stable-x86_64-apple-darwin'

  stable-x86_64-apple-darwin unchanged - rustc 1.38.0 (625451e37 2019-09-23)

→ cat ~/.rustup/settings.toml
default_host_triple = "x86_64-apple-darwin"
default_toolchain = "nightly-x86_64-apple-darwin"
profile = "default"
version = "12"

[overrides]
"/Users/bchen/github/rustup.rs" = "stable-x86_64-apple-darwin"
```

</p>
</details>

kinnison (changes requested):
Looking good.  Could you please add a test to assert that `rustup +stable set profile minimal` doesn't affect the default toolchain?

pickfire:
@BeniCheni Nice, looks good to me. Thanks a lot for taking this on.

Just left the test that @kinnison mentioned.

BeniCheni:
>  Looking good. Could you please add a test to assert that rustup +stable set profile minimal doesn't affect the default toolchain?

@kinnison, sure thing! Added tests per the helpful suggestion. Thank you. PTAL? 🙏 

> @BeniCheni Nice, looks good to me. Thanks a lot for taking this on.

> Just left the test that @kinnison mentioned.

@pickfire, thank you! The version wouldn't have matured to the current version now, which could service all subcommed w. "+toolchain". Gratitude for your time & effort to review. Please take a final look, if you have time? 🙏

pickfire:
> @pickfire, thank you! The version wouldn't haven matured to the current version now, which could service all subcommed w. "+toolchain". Gratitude for your time & effort to review. Please take a final look, if you have time? :pray:

@BeniCheni Oh, no worries. I was initially trying to do this but I cannot find how `cargo` do the `+nightly` thing and I have no idea how to do that, I didn't know it is just as simple as
```rust
        .arg(
            Arg::with_name("toolchain")
                .help(TOOLCHAIN_ARG_HELP)
                .long("toolchain")
                .takes_value(true),
            Arg::with_name("+toolchain")
                .help("release channel (e.g. +stable) or custom toolchain to set override"),
        )
```
so I learned a lot as well. Either way, it is interesting to see how you use many different functions to piece this up.

The only thing I don't get is the need to have https://github.com/rust-lang/rustup.rs/pull/2031/files#diff-a11274be6e834e18a2579dad1b37ce8dR1173, but it does not matter much here.

BeniCheni:
> The only thing I don't get is the need to have https://github.com/rust-lang/rustup.rs/pull/2031/files#diff-a11274be6e834e18a2579dad1b37ce8dR1173, but it does not matter much here.

@pickfire, I'm guessing the `println!();` is simply to just print a new line. It's used in `override_add` where the logic  & UX `override_by_toolchain` is aligning to; thus, the same line is in `override_by_toolchain`:

https://github.com/rust-lang/rustup.rs/blob/b9e968b6656dd999eb09fb6dc8e7952677dd3fc2/src/cli/rustup_mode.rs#L1176

pickfire:
@BeniCheni Ah, I saw that there will always be a `println!` preceding `show_channel_update`, maybe I will send a patch to put that in `show_channel_update` to avoid the duplication.

src/cli/rustup_mode.rs
```diff
@@ -1153,6 +1154,29 @@ fn toolchain_remove(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+        None
+    };
+
+    let path = utils::current_dir()?;
+    toolchain.make_override(&path)?;
+
+    if let Some(status) = status {
+        println!();
+        common::show_channel_update(cfg, toolchain.name(), Ok(status))?;
+    }
```
kinnison: One difficulty here is that this is far too noisy if you already have a toolchain installed.  consider:

```console
somniloquy% rustup +nightly which rustc                                                   
info: using existing install for 'nightly-x86_64-unknown-linux-gnu'
info: override toolchain for '/home/dsilvers' set to 'nightly-x86_64-unknown-linux-gnu'

  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.40.0-nightly (032a53a06 2019-10-03)

/home/dsilvers/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rustc
somniloquy%                                                                               
```

In addition it appears to have added an override which is definitely not desirable.  We need to come up with a way to not have this override in a persistent fashion or we cannot have this patch.

kinnison (changes requested):
As it stands, this is not going to work because it stores a persistent override and is very noisy about doing so.  We need to ensure that `rustup +foo ...` doesn't affect any more than the individual execution.

src/cli/rustup_mode.rs
pickfire: Will making the changes in https://github.com/rust-lang/rustup.rs/pull/2031/files#diff-a11274be6e834e18a2579dad1b37ce8dR1162 help?

kinnison: That link doesn't quite seem to show me anything I understand as a fix for this.  The problem is the presence of the `make_override()` I think.

BeniCheni:
> ... With an ephemeral toolchain entry in Cfg, then OverrideReason can gain a CommandLine variant, and then in Cfg::find_override() Ben can introduce a check for that before testing self.env_override

Greeting @kinnison. After the merge of #2075 and the suggestion from Discord ☝️ , I was able to follow along the similar idea and pushed an edit. PTAL at next convenience?

Here's some local testing w. the edited version:

<details>
<summary>Tested w. rustup +nightly target add wasm32-unknown-unknown</summary>

```
# in ~/github/rustup.rs on git:which-toolchain-by-root-cmd
→ rustup +nightly target add wasm32-unknown-unknown
info: component 'rust-std' for target 'wasm32-unknown-unknown' is up to date

# in ~/github/rustup.rs on git:which-toolchain-by-root-cmd
→ ls -la ~/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/
total 0
drwxr-xr-x   3 bchen  ___   96 Oct 22 01:56 .
drwxr-xr-x  17 bchen  ___  544 Oct 22 01:56 ..
drwxr-xr-x  24 bchen  ___  768 Oct 22 01:56 lib
```
</details>

<details>
<summary>Tested w. rustup +beta which cargo</summary>

```
# in ~/github/rustup.rs on git:which-toolchain-by-root-cmd
→ rustup +beta which cargo                         
/Users/bchen/.rustup/toolchains/beta-x86_64-apple-darwin/bin/cargo
```
</details>

src/cli/rustup_mode.rs
```diff
@@ -131,10 +135,8 @@ pub fn cli() -> App<'static, 'static> {
                 .long("quiet"),
         )
         .arg(
-            Arg::with_name("toolchain")
-                .help(TOOLCHAIN_ARG_HELP)
-                .long("toolchain")
-                .takes_value(true),
+            Arg::with_name("+toolchain")
+                .help("release channel (e.g. +stable) or custom toolchain to set override"),
```
kinnison: We probably want a validator on this to ensure only values starting with a `+` get through

src/config.rs
```diff
@@ -281,7 +287,12 @@ impl Cfg {
     pub fn find_override(&self, path: &Path) -> Result<Option<(Toolchain<'_>, OverrideReason)>> {
         let mut override_ = None;
 
-        // First check RUSTUP_TOOLCHAIN
+        // First check toolchain override from command
+        if let Some(ref name) = self.toolchain_override {
+            override_ = Some((name.to_string(), OverrideReason::Environment));
```
kinnison: I'd like it if you added `OverrideReason::CommandLine` whose description was something like `overridden by +toolchain on the command line` since otherwise this incorrectly states that `rustup +beta show` picks the beta toolchain due to `RUST_TOOLCHAIN` being set in the environment.

src/cli/rustup_mode.rs
```diff
@@ -35,6 +35,10 @@ pub fn main() -> Result<()> {
     let quiet = matches.is_present("quiet");
     let cfg = &mut common::set_globals(verbose, quiet)?;
 
+    if let Some(t) = matches.value_of("+toolchain") {
+        cfg.set_toolchain_override(&t[1..]);
```
kinnison, line 39: Currently there's no guarantee this will be quite right, see later comment about validators.

src/cli/rustup_mode.rs
kinnison: Perhaps add:
```rust
                .validator(|s| {
                    if s.starts_with('+') {
                        Ok(())
                    } else {
                        Err("Toolchain overrides must begin with '+'".into())
                    }
                }),
```

kinnison (changes requested):
This is looking very good.  Just a couple of minor niggles.

BeniCheni:
> ... a validator on this to ensure only values starting with a + get through

> ... if you added OverrideReason::CommandLine ...

@kinnison, nice suggestion for those _UX tweaks_. Added a validator, the `OverrideReason::CommandLine` and updated tests accordingly. PTAL? Thanks!

kinnison (approved):
This appears to behave as desired. :+1:

Thank you very much for your efforts here.

PR #2042, opened 2019-10-07
Improve OS detection of rustup-init.sh

### LINKED ISSUE #1946: Incorrect detection on Debian/Buster
I set about installing `rustup` onto a Debian/buster system and saw the following:

```shell
somniloquy% curl https://sh.rustup.rs/ | sh                                    
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 10652  100 10652    0     0  60522      0 --:--:-- --:--:-- --:--:-- 60522
info: downloading installer
Warning: Detected OS X platform older than 10.13
Warning: Not forcing TLS v1.2, this is potentially less secure

Welcome to Rust!

This will download and install the official compiler for the Rust programming 
language, and its package manager, Cargo.
…
```

Seems we're incorrectly identifying Debian as OS-X somehow?

Rudi3:
An attempt to fix https://github.com/rust-lang/rustup.rs/issues/1946.

Checks if `$ARCH` contains `"darwin"` before checking the the existence of `sw_vers` and the software version of an alleged darwin os.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -9,6 +9,8 @@ set -u
 # If RUSTUP_UPDATE_ROOT is unset or empty, default it.
 RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
 
+ARCH=
+
 #XXX: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat 1>&2 <<EOF
@@ -46,17 +48,17 @@ main() {
     need_cmd rmdir
 
     get_architecture || return 1
-    local _arch="$RETVAL"
-    assert_nz "$_arch" "arch"
+    ARCH="$RETVAL"
+    assert_nz "$ARCH" "arch"
 
     local _ext=""
-    case "$_arch" in
+    case "$ARCH" in
         *windows*)
             _ext=".exe"
             ;;
     esac
 
-    local _url="${RUSTUP_UPDATE_ROOT}/dist/${_arch}/rustup-init${_ext}"
+    local _url="${RUSTUP_UPDATE_ROOT}/dist/${ARCH}/rustup-init${_ext}"
 
     local _dir
     _dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup)"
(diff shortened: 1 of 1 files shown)


Rudi3:
I'll fix the `dash` incompatibility soon!

kinnison:
Given I discovered that #1946 was a bad issue (my fault) I'm not sure we need this.

Rudi3:
Maybe not, but you may not wan't to solely rely on the existence of `sw_vers` to detect darwin as it can "easily" be fooled. `$ARCH` should be more reliable. On the other hand, it should not be **that** likely to have a fake `sw_vers`. ^^

kinnison:
Fair, I'll give this a bit more of a ponder.  Belt and braces can't hurt I suppose.

kinnison:
I'm not sure I like the amount of code churn needed to have the global `$ARCH` but I also don't like the idea of having to call `get_architecture` from multiple places.  It'd be non-trivial to thread `$_arch` through all the calls, leaving the possibility of just checking `$(uname -s)` in the `check_help_for` function.  I remain unsure which approach is best over-all because we abstracted uname checking into `get_architecture` for a reason.  I'm not going to make a call on this yet, but if you could come up with a way to minimise the change while having the same effect, I'd like to see a reduced patch.

kinnison:
I remain uncomfortable with this change.  If a reduced scope patch is produced I'll consider it, otherwise I'll close this PR soon.  Thank you for caring enough to try and fix this.

rbtcollins:
its pretty trivial really.

_arch -> downloader -> check_help_for. Its one code path, one variable.

Rudi3:
Yes, I could just pass it through like you suggested. I'll try that soon.

kinnison (commented):
This looks plausible.  I'm not in a position to fully review the change right now, but I shall endeavour to do so soon.  My gut feeling says this could be merged now.

kinnison:
I apologise for the delay, my December was crazy and then I was unwell.  :+1:

kinnison:
I've rebased your branch.  Assuming this passes CI, I'll merge.

PR #2043, opened 2019-10-07
Fix/improve human-readable units

### LINKED ISSUE #2041: `N iops / N iops (100%) M iops/s` makes no sense
Installing rustup I saw this very curious line pop up:

> 2 iops /   2 iops (100 %)   1 iops/s in  1s ETA:  0s    

but it makes no sense whatsoever as “iops” is already an unit of operations per second. So paraphrased:

> 2 I/O operations per second out of 2 I/O operations per second, 1 I/O operations per second per second in 1 second.

The units should be fixed up to represent actual whatever being counted here.

Rudi3:
Attempt to fix https://github.com/rust-lang/rustup.rs/issues/2041.

Differentiates between `B|B/s` and `IO-ops|IOPS` and formats them human-friendly.

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -1,6 +1,7 @@
 use crate::term2;
 use rustup::dist::Notification as In;
 use rustup::utils::tty;
+use rustup::utils::units::{Size, Unit, UnitMode};
 use rustup::utils::Notification as Un;
 use rustup::Notification;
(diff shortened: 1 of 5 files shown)


Rudi3:
This currently does not work with total numbers...

kinnison:
Since this touches code @rbtcollins wrote, I'd like their opinion on it.

Issue #1252, opened 2017-09-16
Automatically clean up `tmp` folder


alexcrichton:
I just realized I have a 10GB `.rustup/tmp` folder on my system full of temporary files...

alexcrichton:
Same for `.rustup/downloads`, it was 1GB large!

senden9:
I use rustup since a few months and my tmp folder is quite small with 4KB.
```
$ du --summarize --human .rustup/* 
4.0K	.rustup/downloads
4.0K	.rustup/settings.toml
188K	.rustup/tmp
1.3G	.rustup/toolchains
12K	.rustup/update-hashes
```
Maybe a bug? Maybe rustup crashed & did not clean up after the next run?

bgianfo:
Just hit this today, my temp folder was 100's of MB, and my downloads folder was in 1-2 GB range...

kinnison:
The `tmp` folder should be cleanable up with a bit of a tweak.  the `downloads` folder we have a PR in the works for.  I'll actually ask if they can add the `tmp` dir to that too, since it is just as good a location to deal with the cleanups.

PR #2046, opened 2019-10-09
Delete content of downloads dir of user home in update subcommand

### LINKED ISSUE #2038: Feature Request: Clean up download dir (partials etc) after successful full update run.
**Edited by @kinnison to reflect decision later in thread**

It'd be good to have the downloads directory cleaned up if a full `update` (i.e. not a single-toolchain update) is completed.  That way any partial files etc lying around from failed updates will get cleaned up when we're in a position to believe that all is well and they wouldn't be reused.

I'm prepared to mentor someone in doing this, and suggest you talk to me on Discord about it.

---
Original from @timw4mail

After a while of use, the toolchains and downloads take up a decent amount of disk space. 

Add a cleanup command, which would remove old downloads and toolchains, perhaps keeping just the most recent on each update channel.

BeniCheni:
Resolves #2038.

P.S.: can use some mentoring from review comments to:
- polish code pattern
- improve test to actually download a binary, run `update`and assert the deleted content of `downloads` folder of user home.

I'll keep thinking, and will push edit(s), if I could figure out a better test. But thought that I'd start the PR to collect some feedback.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -812,6 +812,9 @@ fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
         }
     } else {
         common::update_all_channels(cfg, self_update, m.is_present("force"))?;
+        info!("cleaning up downloads & tmp directories");
+        utils::delete_dir_contents(&cfg.download_dir);
+        cfg.temp_cfg.clean();
     }
 
     Ok(())
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -587,6 +587,22 @@ where
     })
 }
 
+pub fn delete_dir_contents(dir_path: &Path) {
+    let dir_contents = fs::read_dir(dir_path);
+    if let Ok(contents) = dir_contents {
+        for entry in contents {
+            if let Ok(entry) = entry {
+                let path = entry.path();
+                if path.is_dir() {
+                    remove_dir_all::remove_dir_all(path).expect("Failed to remove a dir");
+                } else {
+                    fs::remove_file(path).expect("Failed to remove a file");
+                }
+            };
+        }
+    };
+}
+
 pub struct FileReaderWithProgress<'a> {
     fh: std::io::BufReader<std::fs::File>,
     notify_handler: &'a dyn Fn(Notification<'_>),
--- a/src/dist/temp.rs
+++ b/src/dist/temp.rs
@@ -1,4 +1,5 @@
 use crate::utils::raw;
+use crate::utils::utils;
 use std::error;
 use std::fmt::{self, Display};
 use std::fs;
@@ -195,6 +196,10 @@ impl Cfg {
                 });
             }
         }
+    }
+
+    pub fn clean(&self) {
+        utils::delete_dir_contents(&self.root_directory);
     }
 }
 
--- a/tests/cli-rustup.rs
+++ b/tests/cli-rustup.rs
@@ -59,6 +59,7 @@ info: installing component 'cargo'
 info: installing component 'rust-docs'
 info: installing component 'rust-std'
 info: installing component 'rustc'
+info: cleaning up downloads & tmp directories
 "
             ),
         );
@@ -98,6 +99,7 @@ info: installing component 'cargo'
 info: installing component 'rust-docs'
 info: installing component 'rust-std'
 info: installing component 'rustc'
+info: cleaning up downloads & tmp directories
 "
             ),
         );
@@ -120,6 +122,7 @@ fn rustup_stable_no_change() {
             ),
             for_host!(
                 r"info: syncing channel updates for 'stable-{0}'
+info: cleaning up downloads & tmp directories
 "
             ),
         );
@@ -188,6 +191,7 @@ info: installing component 'cargo'
 info: installing component 'rust-docs'
 info: installing component 'rust-std'
 info: installing component 'rustc'
+info: cleaning up downloads & tmp directories
 "
             ),
         );
@@ -244,6 +248,7 @@ info: installing component 'cargo'
 info: installing component 'rust-docs'
 info: installing component 'rust-std'
 info: installing component 'rustc'
+info: cleaning up downloads & tmp directories
 "
             ),
         );
@@ -260,6 +265,7 @@ fn rustup_no_channels() {
             &["rustup", "update", "--no-self-update"],
             r"",
             r"info: no updatable toolchains installed
+info: cleaning up downloads & tmp directories
 ",
         );
     })


BeniCheni:
Note: ah, acknowledge that there's issue w. the tests failing CI. Will fix in a bit.

src/cli/rustup_mode.rs
```diff
@@ -808,9 +811,43 @@ fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
     Ok(())
 }
 
+fn delete_downloads_dir_contents() {
+    match dirs::home_dir() {
+        Some(path) => match path.to_str() {
+            Some(home_dir) => {
+                let mut downloads_dir_path = "".to_string();
+                downloads_dir_path.push_str(&home_dir);
+                downloads_dir_path.push_str(&"/.rustup/downloads".to_string());
```
timw4mail: Couldn't this just be `downloads_dir_path.push_str("/.rustup/downloads");` ?

timw4mail: OR, even simpler: 
`let downloads_dir_path = format!("{}/.rustup/downloads", home_dir);`

BeniCheni:
> OR, even simpler ...

@timw4mail, good 👁. `format!` is definitely more elegant. Updated.

src/cli/rustup_mode.rs
```diff
@@ -12,12 +12,15 @@ use rustup::utils::utils::{self, ExitCode};
 use rustup::{command, Cfg, Toolchain};
 use std::error::Error;
 use std::fmt;
+use std::fs::{self};
```
kinnison: This can simply be `use std::fs`

src/cli/rustup_mode.rs
```diff
@@ -12,12 +12,15 @@ use rustup::utils::utils::{self, ExitCode};
...
 use std::io::Write;
 use std::iter;
 use std::path::{Path, PathBuf};
 use std::process::{self, Command};
 use std::str::FromStr;
 
+extern crate dirs;
```
kinnison: Rustup is in the 2018 edition and so `extern crate` is unnecessary.

src/cli/rustup_mode.rs
```diff
@@ -808,9 +811,41 @@ fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
         common::update_all_channels(cfg, self_update, m.is_present("force"))?;
     }
 
+    delete_downloads_dir_contents();
```
kinnison: This is not where I suggested this should go.  I explicitly suggested it only be done at the end of a successful *full* update.

src/cli/rustup_mode.rs
```diff
@@ -808,9 +811,41 @@ fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+
     Ok(())
 }
 
+fn delete_downloads_dir_contents() {
+    match dirs::home_dir() {
+        Some(path) => match path.to_str() {
+            Some(home_dir) => {
+                let downloads_dir_path = format!("{}/.rustup/downloads", home_dir);
```
kinnison: Paths are managed differently in Rustup because we have some environment variable overrides possible.  The `Cfg` instance which is passed to the `update()` function has a `download_dir` member which you can use.

src/cli/rustup_mode.rs
```diff
@@ -808,9 +811,41 @@ fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+                delete_downloads_content(downloads_dir_path);
+            }
+            None => warn!("Unable to parse home directory"),
+        },
+        None => warn!("Unable to get home directory"),
+    }
+}
+
+fn delete_downloads_content(downloads_dir_path: String) {
```
kinnison: This `String` really should be a `&Path`

tests/cli-misc.rs
```diff
@@ -905,3 +905,14 @@ fn which() {
         );
     });
 }
+
+#[test]
+fn delete_downloads_dir_contents_by_update() {
```
kinnison: This test case should explicitly *NOT* clean up the downloads directory because there might be partials for updating a different toolchain which it would be bad to lose access to.

kinnison (changes requested):
Thank you for trying to solve this particular issue.  However there are a couple of fundamental things which need changing.

One is that you need to properly manage the download path by means of the `Cfg` instance which exists already, and not keep casting paths into strings.  `String` guarantees utf8 correctness, but UNIX paths are *NOT* guaranteed to be valid UTF8, hence Rust has `Pathbuf` and `Path`.

The second is that I *really* don't want to see the downloads dir cleaned up *unless* the user has just completed a *full* `rustup update` because that's the only time where we can be sure that the partials are no longer needed.

There is a call to `common::update_all_channels()` in `update()` in `rustup_mode.rs` which is where you should be adding your call to delete the downloads contents.  i.e. in `update()` *after* `common::update_all_channels()`.

If you want additional help with making these changes, just let me know.

Thanks again,

D.

kinnison:
It may also make sense to solve #1252 at the same time, by adding support for cleaning up the content of the temp directory in `dist/temp.rs`'s `impl Cfg`.  Perhaps you could split your cleanup function into `utils/utils.rs` and then use it from `update()` to clean up `cfg.download_dir` and add `cfg.temp_cfg.clean()` to ask `dist::temp::Cfg` to clean up too?  If that's too much for you then just ignore this suggestion and I'll add it once your PR is mergeable.

src/cli/rustup_mode.rs
```diff
@@ -808,9 +811,41 @@ fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
         common::update_all_channels(cfg, self_update, m.is_present("force"))?;
     }
 
+    delete_downloads_dir_contents();
+
     Ok(())
 }
 
+fn delete_downloads_dir_contents() {
+    match dirs::home_dir() {
```
tesuji:
```suggestion
    match home::home_dir() {
```

Cargo.toml
```diff
@@ -24,6 +24,7 @@ no-self-update = []
 [dependencies]
 chrono = "0.4"
 clap = "2"
+dirs = "2.0.2"
```
tesuji:
```suggestion
```

src/cli/rustup_mode.rs
tesuji:
```suggestion
```

src/cli/rustup_mode.rs
```diff
@@ -808,9 +811,41 @@ fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+
+fn delete_downloads_content(downloads_dir_path: String) {
+    info!("cleaning up downloads directory");
+    let downloads_dir = fs::read_dir(&downloads_dir_path);
+    if let Ok(dir) = downloads_dir {
+        for entry in dir {
+            if let Ok(entry) = entry {
+                let path = entry.path();
+                if path.is_dir() {
+                    fs::remove_dir_all(path).expect("Failed to remove a dir");
```
tesuji:
```suggestion
                    remove_dir_all::remove_dir_all(path).expect("Failed to remove a dir");
```

src/utils/utils.rs
```diff
@@ -587,6 +587,22 @@ where
...
 }
 
+pub fn delete_dir_contents(dir_path: &Path) {
+    let downloads_dir = fs::read_dir(&dir_path);
+    if let Ok(dir) = downloads_dir {
+        for entry in dir {
+            if let Ok(entry) = entry {
+                let path = entry.path();
+                if path.is_dir() {
+                    fs::remove_dir_all(path).expect("Failed to remove a dir");
```
tesuji: You lost it in a rebase?
```suggestion
                    remove_dir_all::remove_dir_all(path).expect("Failed to remove a dir");
```

BeniCheni:
> Perhaps you could split your cleanup function into utils/utils.rs and then use it from update() to clean up cfg.download_dir and add cfg.temp_cfg.clean() to ask dist::temp::Cfg to clean up too?

@kinnison, thanks for the insightful comments. Last commit attempted to follow the suggestion, hopefully, make the calls to clean up "downloads/" and "tmp/" dirs at the right place where after a full rustup update.

PTAL at convenience? Will promptly edit under guidance, if this version still is calling the cleanup functions at the right place.

BeniCheni:
Note: not sure why travis would fail [one of the steps in current CI instance](https://travis-ci.com/rust-lang/rustup.rs/jobs/245928151#L183) w. "The command "docker run ..." 🤔 

```
See 'docker run --help'.
The command "docker run \
  --volume "$TRAVIS_BUILD_DIR":/checkout:ro \
  --workdir /checkout \
  --rm \
  -it \
  centos:6 \
  sh ./ci/raw_init.sh;
" exited with 125.
0.55s$ shellcheck -s dash -- rustup-init.sh;
```

src/utils/utils.rs
```diff
@@ -587,6 +587,22 @@ where
     })
 }
 
+pub fn delete_dir_contents(dir_path: &Path) {
+    let downloads_dir = fs::read_dir(&dir_path);
```
kinnison: It's not necessarily a downloads_dir at this point, so perhaps just call this `dir_contents` ?

kinnison (changes requested):
The change looks good, but I don't understand why there's a `Cargo.lock` change here since you don't touch `Cargo.toml` -- could you please remove that from the diff, and rename the variable as per my comment, then I think we're almost good to merge.

BeniCheni:
> but I don't understand why there's a Cargo.lock change here

Reverted the overlook of Cargo.lock file in the PR.

> just call this `dir_contents`?

Sounds great. Last commit renamed to `dir_contents`. Thanks, @kinnison

src/utils/utils.rs
```diff
@@ -587,6 +587,22 @@ where
     })
 }
 
+pub fn delete_dir_contents(dir_path: &Path) {
+    let dir_contents = fs::read_dir(&dir_path);
```
kinnison: I've just spotted that you have an unnecessary borrow here.  `dir_path` is already a `&Path` so you are creating a `&&Path` which the compiler is then automatically dereffing.

```suggestion
    let dir_contents = fs::read_dir(dir_path);
```

kinnison (commented):
One little tweak, but I can do that on merge if you don't want to.

BeniCheni:
> One little tweak, but I can do that on merge if you don't want to.

@kinnison, pushed another update for another tweak. Thanks for catching the unnecessary borrow. 👍

kinnison (approved):

PR #2050, opened 2019-10-12
self_update: Ensure profile set on self-update

### LINKED ISSUE #2045: upgrading from non-profiles rustup to profiles rustup should auto-add profile= in settings.toml just like first time install does

if `rustup update` will pull a `rustup` which now has [profiles](https://github.com/rust-lang/rustup.rs/pull/1673) support, the `profile=` key/value in `~/.rustup/settings.toml` will not be added(presumably, I haven't [actually](https://github.com/rust-lang/rustup.rs/pull/1673#issuecomment-539541258) tested xD).

I dno.

**Notes**

A side effect of not having `profile=` set in `settings.toml` is that you can't install any specific nightly version, as seen [here](https://github.com/rust-lang/rustup.rs/pull/1673#issuecomment-539463239)

Workaround is running `rustup set profile` or `rustup set profile minimal`, the former sets `profile = default` in `~/.rustup/settings.toml` and the latter sets `profile = minimal` which seems to match the non-profiles version of `rustup` that just got upgraded.

Also note that any/all comments in `settings.toml` are removed by running `rustup set profile`, or more specifically [this code](https://github.com/rust-lang/rustup.rs/pull/1673/files#diff-6d8a5e2104954a026de2bfe2b936aea4R128-R138)

kinnison:
Fixes: #2045

This ought to resolve the issue by forcing a profile on self-update.  This is sub-optimal long-term but until I've satisfied myself that the potential use-cases for a config with no profile set are no longer relevant I refuse to force `Cfg::get_profile()` to always return a profile.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1582,6 +1582,23 @@ pub fn run_update(setup_path: &Path) -> Result<()> {
     process::exit(0);
 }
 
+/// Ensure that the configuration is good after a self-update
+///
+/// Currently the only thing we do is ensure that a profile is set
+/// since that could mess things up otherwise, and we don't really
+/// want to do a full metadata update for that.  There are potentially
+/// legitimate reasons for a user to unset profile though so we only
+/// set it on updates rather than simply ensuring we always have a
+/// profile set in `Cfg::get_profile()`
+fn ensure_config_good() -> Result<()> {
+    let cfg = common::set_globals(false, true)?;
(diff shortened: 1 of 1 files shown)


kinnison:
If we merge this, I should file an issue to think about if `Cfg::get_profile()` could be forced to always return `Profile::default()` if one isn't in the config.

kinnison:
@howaboutsynergy How does this look to you?

PR #2053, opened 2019-10-15
Remove scopeguard 0.3, miniz-sys and bump curl, openssl, tokio-reactor and flate2

tesuji:


kinnison:
Could you please confirm the command you ran to generate that lock diff?

tesuji:
That is 
* cargo update -p tokio-reactor
* cargo update -p flate2
* cargo update -p curl openssl-sys openssl openssl-src

PR #2056, opened 2019-10-15
Point to VS C++ Build tools

rofrol:
https://visualstudio.microsoft.com/downloads/ - points to `vs_Community.exe`, when run will give us this screen:

image

What we want instead is https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019 - points to `vs_BuildTools.exe`, when run gives us this screen:

image

Similar as here https://github.com/rust-lang/rustup.rs/blob/1094d2e96ced2ad21e36ffff305236f908a19a5d/src/cli/self_update.rs#L194

Previously there were some problems:
- https://github.com/rust-lang/rustup.rs/issues/1772
- https://github.com/rust-lang/rust/issues/42744

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,7 +480,7 @@ $ rustup target add i686-pc-windows-gnu
 ```
 
 [ABIs]: https://en.wikipedia.org/wiki/Application_binary_interface
-[Visual Studio]: https://visualstudio.microsoft.com/downloads/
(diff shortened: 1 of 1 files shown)


kinnison (approved):
This looks good, I'll merge it once I've resolved some issues with 1.20.0

BeniCheni (approved):
LGTM

PR #2058, opened 2019-10-15
Kinnison/fixups 1.20.1

kinnison:
This branch covers two major things to improve matters arising from 1.20

1. We now assume the default profile if there's nothing in the config -- this means that distro-provided `rustup` won't fail to install new toolchains entirely and strangely.
2. We now force the toolchain installation during `rustup-init` so that platforms missing components will at least still install to some extent.

Minor fixes:

1. `rustup toolchain install` now has `--force` where somehow we missed it off before.
2. Report when we skip a missing component because we're `--force`d


nrc (approved):
lgtm!

PR #2072, opened 2019-10-18
Add documentation about profiles

woshilapin:
I couldn't find documentation in the README about the recently released "profiles". I mostly copied all the documentation from the blog post ["Announcing Rustup 1.20.0"](https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles) mostly written by @pietroalbini (and also contributed by @nrc).

@pietroalbini @nrc please confirm you are OK with it. Do I need to cite you? Is there a good practice for doing it (format of the citation, where to put it, etc.)?

Thank you for this very nice feature by the way :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ And it runs on all platforms Rust supports, including Windows.
 [rustlang]: https://www.rust-lang.org
 
 * [Installation](#installation)
+  * [Profiles](#profiles)
+  * [Shell autocompletion](#enable-tab-completion-for-bash-fish-zsh-or-powershell)
+  * [Where to install?](#choosing-where-to-install)
 * [How rustup works](#how-rustup-works)
 * [Keeping Rust up to date](#keeping-rust-up-to-date)
 * [Working with nightly Rust](#working-with-nightly-rust)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +14,9 @@ And it runs on all platforms Rust supports, including Windows.
 [rustlang]: https://www.rust-lang.org
 
 * [Installation](#installation)
+  * [Profiles](#profiles)
```
slonopotamus, line 17: Indentation gone wrong

README.md
```diff
@@ -585,6 +618,7 @@ but the command-line `curl` command works fine, this may be the problem.
 Command                                                     | Description
 ----------------------------------------------------------- | ------------------------------------------------------------
 `rustup default nightly`                                    | Set the default toolchain to the latest nightly
+`rustup set profile minimal`                                | Set the default "profile" (see [profiles](#profiles))
```
slonopotamus, line 621: Maybe

> Set the default "profile" *to minimal* (see [profiles](#profiles))

Though it is possibly better to use `default` profile here so that people who randomly copy-paste commands from internets do not switch to `minimal` by accident.

README.md
```diff
@@ -668,6 +702,7 @@ $ curl https://sh.rustup.rs -sSf | sh -s -- --help
 $ curl https://sh.rustup.rs -sSf | sh -s -- --no-modify-path
 $ curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain nightly
 $ curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain none
+$ curl https://sh.rustup.rs -sSf | sh -s -- --profile minimal --default-toolchain nightly
```
slonopotamus, line 705: Same as above, not sure that using `minimal` in examples is a good thing.

README.md
woshilapin: No, it is intentional, these are subsections. But maybe we don't want them.

README.md
woshilapin: I agree but at the same time, people who wants the `default` profile don't need this command at all because `default` is... the default profile :smile:

slonopotamus: On the other side... switching to `nightly` is also a not very safe thing, so `minimal` is ok here

README.md
woshilapin: See [comment](https://github.com/rust-lang/rustup.rs/pull/2072#discussion_r336374302).

emilyalbini:
> please confirm you are OK with it. Do I need to cite you?

Nah it's fine, no citation is needed.

README.md
kinnison: I think the subsections are okay.

kinnison (approved):
This looks like an excellent combining of Pietro's writing into the README.  Thank you. :+1:

PR #2075, opened 2019-10-19
install: add --profile flag to override profile

### LINKED ISSUE #2004: Support selecting a profile at toolchain installation time
From time to time, people may want to override their default toolchain profile when installing a new one.

For example, I may set my toolchain profile default as `minimal` so that auto-installed toolchains from `rust-toolchain` files are as small as possible; but then I might want to explicitly install a toolchain as `complete`.  Currently this would involve setting the profile, installing the toolchain, and then resetting the profile to whatever it was before.

It'd be nicer if I could do `rustup toolchain install --profile complete 1.37.0` or similar.

jonhoo:
Previously, the profile to use when installing a new toolchain had to be
set globally through `rustup set profile`, or at rustup-init time.
However, sometimes this is inconvenient, because you only want to use a
particular profile for a specific toolchain. This patch allows users to
pass `--profile` to `rustup install` to override the profile used for
that one install.

Fixes #2004.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -33,7 +33,7 @@ pub fn main() -> Result<()> {
     let matches = cli().get_matches();
     let verbose = matches.is_present("verbose");
     let quiet = matches.is_present("quiet");
-    let cfg = &common::set_globals(verbose, quiet)?;
+    let cfg = &mut common::set_globals(verbose, quiet)?;
 
     if maybe_upgrade_data(cfg, &matches)? {
         return Ok(());
@@ -89,8 +89,8 @@ pub fn main() -> Result<()> {
             (_, _) => unreachable!(),
         },
         ("set", Some(c)) => match c.subcommand() {
-            ("default-host", Some(m)) => set_default_host_triple(&cfg, m)?,
-            ("profile", Some(m)) => set_profile(&cfg, m)?,
+            ("default-host", Some(m)) => set_default_host_triple(cfg, m)?,
+            ("profile", Some(m)) => set_profile(cfg, m)?,
             (_, _) => unreachable!(),
         },
         ("completions", Some(c)) => {
@@ -168,6 +168,13 @@ pub fn cli() -> App<'static, 'static> {
                         .help(TOOLCHAIN_ARG_HELP)
                         .required(true)
                         .multiple(true),
+                )
+                .arg(
+                    Arg::with_name("profile")
+                        .long("profile")
+                        .takes_value(true)
+                        .possible_values(Profile::names())
+                        .required(false)
                 )
                 .arg(
                     Arg::with_name("no-self-update")
@@ -257,6 +264,13 @@ pub fn cli() -> App<'static, 'static> {
                                 .help(TOOLCHAIN_ARG_HELP)
                                 .required(true)
                                 .multiple(true),
+                        )
+                        .arg(
+                            Arg::with_name("profile")
+                                .long("profile")
+                                .takes_value(true)
+                                .possible_values(Profile::names())
+                                .required(false)
                         )
                         .arg(
                             Arg::with_name("no-self-update")
@@ -772,8 +786,13 @@ fn check_updates(cfg: &Cfg) -> Result<()> {
     Ok(())
 }
 
-fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
+fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<()> {
     let self_update = !m.is_present("no-self-update") && !self_update::NEVER_SELF_UPDATE;
+    if let Some(p) = m.value_of("profile") {
+        let p = Profile::from_str(p)?;
+        cfg.set_profile_override(p);
+    }
+    let cfg = &cfg;
     if let Some(names) = m.values_of("toolchain") {
         for name in names {
             update_bare_triple_check(cfg, name)?;
@@ -1307,7 +1326,7 @@ fn set_default_host_triple(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
     Ok(())
 }
 
-fn set_profile(cfg: &Cfg, m: &ArgMatches) -> Result<()> {
+fn set_profile(cfg: &mut Cfg, m: &ArgMatches) -> Result<()> {
     cfg.set_profile(&m.value_of("profile-name").unwrap())?;
     Ok(())
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -32,6 +32,7 @@ impl Display for OverrideReason {
 }
 
 pub struct Cfg {
+    pub profile_override: Option<dist::Profile>,
     pub rustup_dir: PathBuf,
     pub settings_file: SettingsFile,
     pub toolchains_dir: PathBuf,
@@ -94,6 +95,7 @@ impl Cfg {
         let dist_root = dist_root_server.clone() + "/dist";
 
         let cfg = Self {
+            profile_override: None,
             rustup_dir,
             settings_file,
             toolchains_dir,
@@ -116,6 +118,10 @@ impl Cfg {
         Ok(cfg)
     }
 
+    pub fn set_profile_override(&mut self, profile: dist::Profile) {
+        self.profile_override = Some(profile);
+    }
+
     pub fn set_default(&self, toolchain: &str) -> Result<()> {
         self.settings_file.with_mut(|s| {
             s.default_toolchain = Some(toolchain.to_owned());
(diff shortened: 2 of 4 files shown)


src/cli/rustup_mode.rs
```diff
@@ -169,6 +169,12 @@ pub fn cli() -> App<'static, 'static> {
                         .required(true)
                         .multiple(true),
                 )
+                .arg(
+                    Arg::with_name("profile")
+                        .long("profile")
+                        .possible_values(Profile::names())
+                        .default_value(Profile::default_name()),
+                )
```
kinnison, line 178: While this looks good, Sadly this is only for the "hidden" `rustup install` command.  A little later in the file there's a `toolchain` subcommand which has the real `rustup toolchain install` command which you'll also need to add this flag to.

src/cli/rustup_mode.rs
```diff
@@ -774,6 +780,12 @@ fn check_updates(cfg: &Cfg) -> Result<()> {
 
 fn update(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
     let self_update = !m.is_present("no-self-update") && !self_update::NEVER_SELF_UPDATE;
+    let mut cfg = cfg.clone();
+    if let Some(p) = m.value_of("profile") {
+        let p = Profile::from_str(p)?;
+        cfg = cfg.with_profile(p);
+    }
+    let cfg = &cfg;
```
kinnison: I can see why you're doing this, but `Cfg` is meant to represent the filesystem configuration state more I think than the runtime state.  `toolchain.install_from_dist()` and friends (and yes it's a deep chain) would probably benefit from passing this profile through instead.

src/config.rs
```diff
@@ -31,7 +31,9 @@ impl Display for OverrideReason {
     }
 }
 
+#[derive(Clone)]
```
kinnison: I'm not 100% on if it's completely safe to do this, given `settings_file` has interior mutability.

As such this entire change to `Cfg` is a bit squiffy.

src/config.rs
```diff
@@ -116,6 +119,11 @@ impl Cfg {
         Ok(cfg)
     }
 
+    pub fn with_profile(mut self, profile: dist::Profile) -> Self {
+        self.profile_override = Some(profile);
+        self
+    }
+
```
kinnison: The pattern here doesn't match anything else in `Cfg` so it feels weird to me.

src/config.rs
```diff
@@ -145,6 +153,9 @@ impl Cfg {
     // a user upgrades from a version of Rustup without profiles to a version of
     // Rustup with profiles.
     pub fn get_profile(&self) -> Result<dist::Profile> {
+        if let Some(ref p) = self.profile_override {
+            return Ok(p.clone());
```
kinnison: `Profile` happens to be `Copy` if we were doing this.

src/dist/temp.rs
```diff
@@ -27,10 +28,11 @@ pub enum Notification<'a> {
     DirectoryDeletion(&'a Path, io::Result<()>),
 }
 
+#[derive(Clone)]
 pub struct Cfg {
     root_directory: PathBuf,
     pub dist_server: String,
-    notify_handler: Box<dyn Fn(Notification<'_>)>,
+    notify_handler: Arc<dyn Fn(Notification<'_>)>,
```
kinnison: I'm really not happy about all this incidental noise from making `Cfg` derive `Clone`

Especially how it results in noise in the dist tests too :(

tests/cli-misc.rs
```diff
@@ -568,6 +568,40 @@ fn rename_rls_remove() {
     });
 }
 
+#[test]
+fn install_with_profile() {
+    clitools::setup(Scenario::Full, &|config| {
```
kinnison: If you'd put this in `cli-v2.rs` and copied, say, `test_complete_profile_skips_missing_when_forced()` then you'd have seen `setup_complex()` which has what you wanted here, without using a different pattern.

Even if we keep this test body, I'd prefer the test were in `cli-v2.rs`

tests/cli-misc.rs
```diff
@@ -568,6 +568,40 @@ fn rename_rls_remove() {
...
+        set_current_dist_date(config, "2015-01-01");
+        expect_ok(config, &["rustup", "default", "stable"]);
```
kinnison: You don't need this really, it's just wasting time in the test installing a toolchain you do nothing with.

tests/cli-misc.rs
```diff
@@ -568,6 +568,40 @@ fn rename_rls_remove() {
...
+
+        // Installing with minimal profile should only install rustc
+        expect_ok(
+            config,
+            &[
+                "rustup",
+                "install",
```
kinnison: I'd have preferred to see `rustup toolchain install` rather than `rustup install` since the latter is a deprecated, hidden, alias.

kinnison (changes requested):
Thank you for tackling this issue.  There are a few small problems (such as not implementing the command actually asked for in the issue -- `rustup toolchain install --profile...`) and a big problem which is that I'm not comfortable with the big `Clone` change on `Cfg` and `temp::Cfg`

I'd rather see this implemented by passing through an Option<dist::Profile> to the installation method functions.  I appreciate this results in collateral `None` arguments in a bunch of places, but over-all I think it's more in keeping with the current codebase.  Perhaps in the future we'll have some more live data held in `Cfg` (I was considering it for another PR but now I feel iffy about that)

I'm open to being convinced I should stop worrying about Cfg in that way, at which point I'd rather `update()` took `&mut Cfg` rather than all that `Clone` mess though, but in the meantime I knocked up a diff for the pass-argument approach on my fork if you wanted to take a look at that approach.

kinnison:
The branch mentioned in the review is at: https://github.com/kinnison/rustup.rs/tree/kinnison/profile-on-install

jonhoo:
Yup, as you saw in the stream I was really unsure which approach would be better. I think it'd be handy to have a distinction between "active config" and "on-disk config", but agree that this PR is probably not the place to make that change. Would be happy to use your branch instead of mine -- they both fix the issue. My only comment on your version is that it's weird for `profile` to be the last argument in `common::update_all_channels` and `Cfg::update_all_channels`, but the second in `install_from_dist`. Oh, and that there's some weird formatting and lack of visual separation going on in the test. Happy to close this in favor of your change!

kinnison:
I think there may be a happy medium to be had here.  I'm not entirely comfortable with the clone and would rather just have an override profile option which we can set via a method taking `&mut Self` on `Cfg` -- that'd remove the noisy commit messing with `temp::Cfg`.

If you then moved the test into `cli-v2` where I really think it belongs, remove the other bogons, then we can start a principle of permitting ephemeral configuration data in `Cfg` on the basis that as soon as it gets beyond somewhere passing it &mut, you can no longer change it, it's frozen.

I'd rather we amended your PR than replaced it with my hacked up change (the inconsistency in argument position is in part due to me hacking on it while also watching what you did) since I do think having the precedent set now might help with some work I want to do with Ben Chen.

jonhoo:
Sure thing! Here's a variant with all comments addressed, where we keep `&mut Cfg` in the "outer" layers of the application.

src/config.rs
```diff
@@ -133,9 +131,10 @@ impl Cfg {
...
     }
 
-    pub fn set_profile(&self, profile: &str) -> Result<()> {
-        if !dist::Profile::names().contains(&profile) {
-            return Err(ErrorKind::UnknownProfile(profile.to_owned()).into());
+    pub fn set_profile(&mut self, profile: &str) -> Result<()> {
+        let p = dist::Profile::from_str(profile)?;
+        if self.profile_override.is_some() {
+            self.profile_override = Some(p)
         }
```
kinnison: I wonder, is it right to unset the profile override here?  To be fair, `set_profile()` is only called in a small number of pathways and probably it doesn't matter; but it does feel odd to see this change.

I think if anything, setting the profile explicitly should probably just unset the override (set it to `None`) -- what do you think?

tests/cli-v2.rs
```diff
@@ -73,6 +74,39 @@ fn install_toolchain_from_version() {
     });
 }
 
+#[test]
+fn install_with_profile() {
+    clitools::setup(Scenario::Full, &|config| {
```
kinnison: The scenario in use in `cli-v2.rs`'s `setup_complex` is likely sufficient and already available.  While it's not *bad* to use `clitools::setup()` here, it is at variance with the usual practice in this test file.  Having the setup functions all in one place makes it easier if we need to do something "interesting" in order to make sure Windows behaves.

tests/cli-v2.rs
```diff
@@ -73,6 +74,39 @@ fn install_toolchain_from_version() {
...
+        // Installing with minimal profile should only install rustc
```
kinnison: I wonder if we should add `rustup set profile complete` just to make it clear that the profile setting would install `cargo` too, but for the use of `--profile` in the subsequent command?

kinnison (changes requested):
The over-all delta looks much better now.  A few more comments and if you agree with them and make the change, you may as well rebase into a nicer set of commits ready for merge.

kinnison:
I am slightly confused as to how the test case failing in your CI runs is failing given your change though.

kinnison:
I worked it out and thanks to you permitting commits from maintainers, I've pushed a fix to your branch.  If you're okay with it, and it passes the tests, and you resolve the other comments, then fold it into your commits, no need to try and credit me.

tests/cli-v2.rs
jonhoo: I mean `setup_complex` is only used by one test, but sure :p

kinnison (approved):
LGTM :+1:

Once the CI completes, I'll merge.

Issue #2028, opened 2019-09-28
Tracking: Simple PGP signature verification


kinnison:
In order to implement simple signature verification for `rustup` to an extent that we're confident that it's good to proceed to thinking more about trust models, we need:

* [x] Select an OpenPGP implementation to use (rpgp)
    * [x] Develop a prototype using that implementation (done on #2077)
    * [x] Ensure the OpenPGP implementation supports all platforms we need it to
* [x] Basic verification of signatures over channel manifests (in the `.asc` files)
    * [x] Ensure that the signature-verification-failed message is good enough that users won't just re-run and/or file bugs on `rustup` which aren't actual bugs for us.
    * [ ] Switch warnings for errors, ensuring that GPG signatures are always checked and valid.
* [ ] Sign `rustup` releases
    * [ ] Ensure we have at least release signed before we deploy signature verification in a release
* [ ] Verify the signature on `rustup` releases in a basic way during self-update
* [ ] Verify the signature on the `rustup-init` downloaded by `rustup-init.sh`
    * [ ] Determine which tools we can use and how to detect them
    * [ ] Decide how we distribute the key to verify against
    * [ ] Implement the verification check

---

If anyone has ideas on what else needs doing, please comment below and I shall endeavour to keep this tracking issue up to date with the progress toward simple signature verification support.

dignifiedquire:
I maintain https://github.com/rpgp/rpgp which should be easy to use for this as well and is pure rust, without c dependencies.

kinnison:
Thanks @dignifiedquire I should indeed have a go with that in case I can get going more quickly than waiting for Sequoia to add Windows support which is currently blocking that prototype.
When you say there's no C deps, is that definite?

dignifiedquire:
> When you say there's no C deps, is that definite?

Yes, the goal is a fully portable version (it is currently deployed through an app using it on ios,android, windows, macos and various unix distros), with all crypto being either pure rust, with possible optimizations in assembly.

dignifiedquire:
Also probably of interest, the first security audit was just finished, with some minor things still on my list to fix, but no major issues found. 

If you have a sketch of what you want this feature to exactly do, I'd be happy to help fill in the rpgp specific details, as the docs are not that amazing quite yet

kinnison:
I threw together a very dodgy hack to try with Sequoia here: https://github.com/kinnison/rustup.rs/tree/signed-channels

[...]

PR #2077, opened 2019-10-22
Validate pgp signatures

dignifiedquire:
Ref #2028


src/dist/download.rs
```diff
@@ -167,7 +198,14 @@ impl<'a> DownloadCfg<'a> {
             (self.notify_handler)(Notification::ChecksumValid(url_str));
         }
 
-        // TODO: Check the signature of the file
+        // No signatures for tarballs for now.
+        if !url_str.ends_with(".tar.gz") && !url_str.ends_with(".tar.xz") {
+            if let Err(err) = self.check_signature(&url_str, &file) {
+                // TODO: this should probably be warn!, but log is not used in this part of the code
+                // at the moment.
+                println!("Invalid signature detected: {}", err);
```
kinnison: Actually let's have a notification for this since the approach here in the code is to send a notification.

src/dist/rust-signing-key.asc
```diff
@@ -0,0 +1,221 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
kinnison: Turns out this is at `src/rust-key.gpg.ascii` too, we should probably not include a second copy.  The config object carries a facility to get the key but I don't like it.  For now please drop this file and include the one already present.

src/dist/signatures.rs
```diff
@@ -0,0 +1,51 @@
...
+
+// TODO: Determine whether we want external keyring support
+
+use pgp::types::KeyTrait;
+use pgp::{Deserializable, SignedPublicKey, StandaloneSignature};
+
+use crate::errors::*;
+
+// const SIGNING_KEY_BYTES: &[u8] = include_bytes!("rust-signing-key.asc");
+const SIGNING_KEY_BYTES: &[u8] = include_bytes!("../../tests/mock/signing-key.pub.asc");
```
kinnison: Eventually we'll drop both of these, but for now change the "real" one (commented out) to pull in the `src/rust-key.gpg.ascii` file mentioned in the previous comment.

src/dist/signatures.rs
```diff
@@ -0,0 +1,51 @@
...
+
+lazy_static::lazy_static! {
+    static ref SIGNING_KEYS: Vec<SignedPublicKey> = {
+        pgp::SignedPublicKey::from_armor_many(std::io::Cursor::new(SIGNING_KEY_BYTES))
+            .map_err(squish_internal_err).unwrap()
+            .0
+            .collect::<std::result::Result<Vec<_>, _>>()
+            .map_err(squish_internal_err).unwrap()
+    };
+}
```
kinnison: When we move to configurable keys, we'll need to load these on the fly, so perhaps don't do this lazy-static?

src/dist/signatures.rs
```diff
@@ -0,0 +1,51 @@
...
+
+fn squish_internal_err<E: std::fmt::Display>(err: E) -> Error {
+    ErrorKind::SignatureVerificationInternalError(format!("{}", err)).into()
+}
+
+pub fn verify_signature(content: &str, signature: &str) -> Result<bool> {
```
kinnison: Could we please change this so that `content` is given as a `&dyn Read` rather than `&str` ?  I'm thinking ahead to verifying signatures on updates to rustup.

kinnison (changes requested):
This looks amazing as a first-cut.  We need to have tests which verify that if signatures exist which are bad, we get appropriate warnings, and then we need to think about how we're going to handle the keyring more effectively than the current pair of commented lines; but thank you and I hope my suggestions are useful.

src/dist/signatures.rs
dignifiedquire: We can always have this one be fixed and avoid the cost of loading it multiple times, while the others are pulled in dynamically

src/dist/signatures.rs
dignifiedquire: sure

dignifiedquire:
I will need to wait for a more configurable way for keys to put in the real thing unfortunately, as `cfg(test)` is not applied when running the tests in the `tests` directory (ref likely https://github.com/rust-lang/rust/issues/59168)

kinnison (commented):
This is looking pretty good now.  I'm thinking about how I want to provide key configurability.  I will hopefully have an answer for you in the next few days.

kinnison (commented):
This is starting to look like it's near to ready.

kinnison (approved):

kinnison:
I think we need to start tidying this into a series of functional commits so that we can get ready to merge.  This is looking amazing.

kinnison (approved):
That test looks :+1:

kinnison (approved):
That makes sense, here's hoping it sorts appveyor out :D

kinnison (approved):

PR #2080, opened 2019-10-23
build: use cargo-cache to clean up ~/.cargo dir

tesuji:


kinnison (approved):
I think this looks good.

PR #2083, opened 2019-10-25
Stop using bright colors

### LINKED ISSUE #433: White fonts tough to read on a white terminal
Playing around locally it looks like the installer has two headers in white fonts which end up being tough to read on a white terminal. Unfortunately I'm on a train and can't upload a screenshot right this second, but maybe this rings a bell?

carloabelli:
Bright colors often display badly for those who are using colorschemes
that use unexpected bright colors such as solarized.

Additionally, many terminals automatically display bold text in bright
colors (unless this option is explicitly disabled) particularly if bold
text is not supported.

It should be left up to the user whether they want their bold text to be
displayed in bright colors.

This should also fix #433 (if it is still an issue).

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -155,11 +155,11 @@ fn show_channel_updates(
         match result {
             Ok(UpdateStatus::Installed) => {
                 banner = "installed";
-                color = Some(term2::color::BRIGHT_GREEN);
+                color = Some(term2::color::GREEN);
             }
             Ok(UpdateStatus::Updated) => {
                 banner = "updated";
-                color = Some(term2::color::BRIGHT_GREEN);
+                color = Some(term2::color::GREEN);
             }
             Ok(UpdateStatus::Unchanged) => {
                 banner = "unchanged";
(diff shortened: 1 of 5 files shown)


kinnison:
Thank you for your contribution.  Could you please provide a couple of screenshots so that I can evaluate the visual difference between before and after for this PR?

carloabelli:
@kinnison 

For example if I use my base-16 color scheme and choose not to display bold text in bright colors:

2019-10-25T13:39:53,466307325-04:00

Then when displaying bright red, the text is actually displayed in orange:

2019-10-25T12:49:20,746254730-04:00

This is even more problematic for bright green because the text is almost invisible.

This PR results in the following colors instead

2019-10-25T12:51:17,643914764-04:00

For those terminals that automatically display bold text in bright colors (in my experience most of them) this should have no effect.

carloabelli:
Here's an example for bright green text:

Before:
2019-10-25T21:43:32,538280024-04:00

After:
2019-10-25T21:53:13,359710339-04:00

kinnison:
I can see how it makes a significant difference to readability for you, and I've tested it on my "normal" terminal and it's not sufficiently adversely affected to cause issues.  I'm okay with this as a general readability improvement commit.

carloabelli:
Thanks!

PR #2084, opened 2019-10-25
Correct "toolchain list" command verbose option

BeniCheni:
Correct the `verbose` (or `v`) option of `toolchain list` to display full path of toolchains. Updated tests.

##### current observation - only displays user's rustup directory

```
→ rustup toolchain list -v                                                     
stable-x86_64-apple-darwin (default)	/Users/bchen/.rustup
nightly-2019-09-25-x86_64-apple-darwin	/Users/bchen/.rustup
nightly-x86_64-apple-darwin	/Users/bchen/.rustup
```

##### fixed version

```
→ rustup toolchain list -v                     
stable-x86_64-apple-darwin (default)	/Users/bchen/.rustup/toolchains/stable-x86_64-apple-darwin/
nightly-2019-09-25-x86_64-apple-darwin	/Users/bchen/.rustup/toolchains/nightly-2019-09-25-x86_64-apple-darwin/
nightly-x86_64-apple-darwin	/Users/bchen/.rustup/toolchains/nightly-x86_64-apple-darwin/
```

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -8,6 +8,7 @@ use lazy_static::lazy_static;
 use rustup::utils::notify::NotificationLevel;
 use rustup::utils::utils;
 use rustup::{Cfg, Notification, Toolchain, UpdateStatus};
+use std::fs;
 use std::io::{BufRead, BufReader, ErrorKind, Write};
 use std::path::Path;
 use std::process::{Command, Stdio};
@@ -424,14 +425,28 @@ pub fn list_installed_components(toolchain: &Toolchain<'_>) -> Result<()> {
     Ok(())
 }
 
-pub fn list_toolchains(cfg: &Cfg, is_verbose: bool) -> Result<()> {
-    let toolchains = cfg.list_toolchains()?;
-    let toolchain_info = if is_verbose {
-        format!("\t{}", &cfg.rustup_dir.display())
-    } else {
-        String::from("")
+fn print_toolchain_path(cfg: &Cfg, toolchain: &str, if_default: &str, verbose: bool) -> Result<()> {
+    let toolchain_path = {
+        let mut t_path = cfg.toolchains_dir.clone();
+        t_path.push(&toolchain);
+        t_path
     };
+    let toolchain_meta = fs::symlink_metadata(&toolchain_path)?;
+    let toolchain_path = if verbose {
+        if toolchain_meta.is_dir() {
+            format!("\t{}", toolchain_path.display())
+        } else {
+            format!("\t{}", fs::read_link(toolchain_path)?.display())
+        }
+    } else {
+        String::new()
+    };
+    println!("{}{}{}", &toolchain, if_default, toolchain_path);
+    Ok(())
+}
 
+pub fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<()> {
+    let toolchains = cfg.list_toolchains()?;
     if toolchains.is_empty() {
         println!("no installed toolchains");
     } else if let Ok(Some(def_toolchain)) = cfg.find_default() {
@@ -441,11 +456,13 @@ pub fn list_toolchains(cfg: &Cfg, is_verbose: bool) -> Result<()> {
             } else {
                 ""
             };
-            println!("{}{}{}", &toolchain, if_default, toolchain_info);
+            print_toolchain_path(cfg, &toolchain, if_default, verbose)
+                .expect("Failed to list toolchains' directories");
         }
     } else {
         for toolchain in toolchains {
-            println!("{}{}", &toolchain, toolchain_info);
+            print_toolchain_path(cfg, &toolchain, "", verbose)
+                .expect("Failed to list toolchains' directories");
         }
     }
     Ok(())
--- a/tests/cli-v1.rs
+++ b/tests/cli-v1.rs
@@ -104,14 +104,31 @@ fn list_toolchains() {
             &["rustup", "toolchain", "list", "-v"],
             "(default)\t",
         );
+        #[cfg(windows)]
         expect_stdout_ok(
             config,
-            &["rustup", "toolchain", "list", "--verbose"],
-            "(default)\t",
+            &["rustup", "toolchain", "list", "-v"],
+            "\\toolchains\\nightly-x86",
+        );
+        #[cfg(not(windows))]
+        expect_stdout_ok(
+            config,
+            &["rustup", "toolchain", "list", "-v"],
+            "/toolchains/nightly-x86",
         );
         expect_stdout_ok(config, &["rustup", "toolchain", "list"], "beta-2015-01-01");
-        expect_stdout_ok(config, &["rustup", "toolchain", "list", "-v"], "\t");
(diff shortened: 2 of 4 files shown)


kinnison:
You have some strange indentation which needs fixing in your test.

src/cli/common.rs
```diff
@@ -424,14 +424,36 @@ pub fn list_installed_components(toolchain: &Toolchain<'_>) -> Result<()> {
...
+fn print_toolchain_path(cfg: &Cfg, toolchain: String, if_default: &str, verbose: bool) {
+    let path_delimiter = get_path_delimiter();
+    let toolchain_path = if verbose {
+        format!(
+            "\t{}{}{}{}",
+            cfg.toolchains_dir.display(),
+            path_delimiter,
+            &toolchain,
+            path_delimiter
+        )
```
kinnison: Surely this whole thing would be better done along the lines of:

```rust
let toolchain_path = if verbose {
    let tpath = cfg.toolchains_dir.clone();
    tpath.push(&toolchain);
    format!("\t{}", tpath.display())
}
```

Rather than adding an extra function for the delimiter?

src/cli/rustup_mode.rs
```diff
@@ -60,7 +60,9 @@ pub fn main() -> Result<()> {
         ("default", Some(m)) => default_(cfg, m)?,
         ("toolchain", Some(c)) => match c.subcommand() {
             ("install", Some(m)) => update(cfg, m)?,
-            ("list", Some(m)) => handle_epipe(toolchain_list(cfg, m))?,
+            ("list", Some(m)) => {
+                handle_epipe(common::list_toolchains(cfg, m.is_present("verbose")))?
+            }
```
kinnison: This change isn't really in keeping with the rest of this match, I'd rather leave the expansion of `m.is_present("verbose")` to within a local `toolchain_list()`

tests/cli-v1.rs
```diff
@@ -109,9 +109,55 @@ fn list_toolchains() {
...
+            config,
+            &["rustup", "toolchain", "list", "-v"],
+            "\\toolchains\\nightly-x86",
+        );
+        #[cfg(windows)]
+            expect_stdout_ok(
+            config,
+            &["rustup", "toolchain", "list", "--verbose"],
+            "\\toolchains\\nightly-x86",
+        );
```
kinnison: The `-v`/`--verbose` thing doesn't really need checking here, pick one and test that, no need for both.

tests/cli-v2.rs
```diff
@@ -153,9 +153,55 @@ fn list_toolchains() {
...
+            config,
+            &["rustup", "toolchain", "list", "-v"],
+            "\\toolchains\\nightly-x86",
+        );
+        #[cfg(windows)]
+        expect_stdout_ok(
+            config,
+            &["rustup", "toolchain", "list", "--verbose"],
+            "\\toolchains\\nightly-x86",
+        );
```
kinnison: Again, only need one of `-v` or `--verbose` in this test function.

kinnison (changes requested):
Not bad at all for a first pass.  Please simplify as directed and undo one of your changes, and ensure that the code is properly formatted so it passes the CI.

This is a distinct improvement in output.

BeniCheni:
> Rather than adding an extra function for the delimiter?

>  I'd rather leave the expansion of `m.is_present("verbose")` to within a local `toolchain_list()`

> The `-v`/`--verbose` thing doesn't really need checking here, pick one and test that, no need for both.

@kinnison, updated per suggestions, ready for your next convenience to review again. Thanks!

src/cli/common.rs
```diff
@@ -424,14 +424,20 @@ pub fn list_installed_components(toolchain: &Toolchain<'_>) -> Result<()> {
     Ok(())
 }
 
-pub fn list_toolchains(cfg: &Cfg, is_verbose: bool) -> Result<()> {
-    let toolchains = cfg.list_toolchains()?;
-    let toolchain_info = if is_verbose {
-        format!("\t{}", &cfg.rustup_dir.display())
+fn print_toolchain_path(cfg: &Cfg, toolchain: String, if_default: &str, verbose: bool) {
```
kinnison: It might be cute to also handle symlinked toolchains (`rustup toolchain link`) by reading the symlink and reporting the target rather than the location in `~/.rustup/toolchains`

This might be doable with (untested pseudorust):
```rust
let toolchain_path = {
    let mut t_path = cfg.toolchains_dir.clone();
    t_path.push(&toolchain);
    t_path
};
let toolchain_meta = std::fs::symlink_metadata(&toolchain_path)?;
let toolchain_path = if verbose {
    if toolchain_meta.is_dir() {
        format!("\t{}", toolchain_path.display())
    } else {
        format!("\t{}", std::fs::read_link(toolchain_path)?)
    }
} else {
    String::new()
}
```

You can test whatever you come up with, by doing something like `rustup toolchain link system /usr` and then `rustup toolchain list -v` -- I'd expect the `system` toolchain's path to be listed as `/usr` rather than `/home/whatever/.rustup/toolchains/system`

kinnison: Also, I think I'd prefer if this function took `toolchain` as `&str` even though in theory it's the only consumer of the `String` produced in the calling function -- just to futureproof things a little.

kinnison (changes requested):
This is looking very good, but I had an idea about linked toolchains (see comments)

BeniCheni:
> It might be cute to also handle symlinked toolchains (rustup toolchain link) by reading the symlink and reporting the target rather than the location in ~/.rustup/toolchains

> I'd prefer if this function took toolchain as &str even though in theory it's the only consumer of the String produced in the calling function -- just to futureproof things a little.

Greeting, @kinnison! Sorry about the delay. Been under the "_water_" w. life & work in last couple of days, but now I'm back and pushed an update to the PR.

Like your suggestion of symlink a lot to take the feature further! Updated the feature & added suggested test. Also updated to use `&str` instead of `String` for the `toolchain` param of the `print_toolchain_path` function.

Ready for your next convenience to review again. Thanks!

src/cli/common.rs
```diff
@@ -424,14 +425,28 @@ pub fn list_installed_components(toolchain: &Toolchain<'_>) -> Result<()> {
...
+        let mut t_path = cfg.toolchains_dir.clone();
+        t_path.push(&toolchain);
+        t_path
+    };
+    let toolchain_meta = fs::symlink_metadata(&toolchain_path)?;
+    let toolchain_path = if verbose {
+        if toolchain_meta.is_dir() {
+            format!("\t{}", toolchain_path.display())
+        } else {
+            format!("\t{:?}", fs::read_link(toolchain_path)?)
```
kinnison: Why not:
```suggestion
            format!("\t{}", fs::read_link(toolchain_path)?.display())
```
?

kinnison (changes requested):
Almost there, one little bogon to resolve...

BeniCheni:
> Almost there, one little bogon to resolve...

@kinnison, updated to address the bogon, :) Thanks.

kinnison (approved):

PR #2085, opened 2019-10-26
Rework the PGP key configuration

kinnison:
This PR adds a number of commits which work toward a better PGP key infrastructure inside Rustup.

In particular it adds (deliberately undocumented at this time) a `settings.toml` entry `pgp_keys` and the environment variable `RUSTUP_PGP_KEY`.

We add a `PgpKey` enum which indicates where a key came from, so that we can use it in future debugging and/or informational messages to the user.

Perhaps @dignifiedquire might take a look over this?


src/config.rs
```diff
@@ -33,6 +33,41 @@ impl Display for OverrideReason {
     }
 }
 
+#[derive(Debug)]
+pub enum PgpKey {
```
dignifiedquire: should probably document if the intention is for these to be public or secret keys

dignifiedquire (approved):
Nice, this should do the trick. 

Is there a good place to document this though? As currently this seems a bit hidden for a user

src/config.rs
kinnison: I agree, I'll push a change momentarily to rename that to `PgpPublicKey` to make it clearer.

kinnison:
I am deliberately hiding this for now.  Once we have proper signature checking integrated, I'll probably add something to the `README.md` about public keys, since I'll probably add a `rustup show keys` or somesuch to enumerate and show all the keys rustup knows about and where they come from.

PR #2086, opened 2019-10-26
Improve handling of --default-host

kinnison:
We were previously having somewhat surprising behaviour for `--default-host` when combined with `--default-toolchain none` or an installation where there already is a default toolchain.  This PR corrects that hopefully, making the use of those arguments less surprising.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -749,6 +749,10 @@ fn maybe_install_rust(
     let mut cfg = common::set_globals(verbose, quiet)?;
     cfg.set_profile(profile_str)?;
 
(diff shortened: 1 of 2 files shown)


kinnison:
@faern do you think this will solve your problem, the behaviour is now approximately:

If user specifies `--default-host` on the CLI, or chooses to customise it in the installer, then that is always set into the config, regardless of `--default-toolchain` or existing toolchain installations

faern:
Great. Yes I think `--default-host` should map 1:1 to what is written as the default host in the config. If that would for some reason not be possible/compatible then it should error, not silently set something else.

PR #2088, opened 2019-10-26
component: Support installation/removal of `component-target`

### LINKED ISSUE #2057: Removing components using the output of `component list` is a papercut

Doing a bit of cleanup, I check what components I have installed:

```
$ rustup component list --installed
rust-std-wasm32-unknown-emscripten
```

Oh, I don't need that component right now, let's remove it!

```
$ rustup component remove rust-std-wasm32-unknown-emscripten
error: toolchain 'nightly-x86_64-apple-darwin' does not contain component 'rust-std-wasm32-unknown-emscripten' for target 'x86_64-apple-darwin'
```

Ooooof. It's actually:

```
$ rustup component remove rust-std --target wasm32-unknown-emscripten
```

One solution would be to support `rustup component remove $component-$target`. Another would be to split up the output so that it's more obvious that those are separate pieces:

```
rust-std (wasm32-unknown-emscripten)
```

**Notes**

This is something that I've noticed for a while, so there's no immediate urgency on it, I just do some spring cleaning every few months and run into it. I might be affected more than most due to answering random Stack Overflow questions.

kinnison:
In order to reduce the surprise available when someone expects to be able to
take output of `rustup component list` and use it as input to `rustup component
add` or `rustup component remove` we now support creation of `Component`
objects by attempting to split it on '-' and see if we get a component name and
valid full target triple.

Fixes: #2057

### DIFF
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -13,7 +13,7 @@
 use crate::errors::*;
 use crate::utils::toml_utils::*;
 
-use crate::dist::dist::{Profile, TargetTriple};
+use crate::dist::dist::{PartialTargetTriple, Profile, TargetTriple};
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


kinnison:
WHile this isn't ready yet (no tests) I'd appreciate a quick nod from @shepmaster if this is appropriate.

kinnison:
I think those tests are sufficient.

src/dist/manifest.rs
```diff
@@ -496,6 +496,24 @@ impl Component {
             is_extension,
         }
     }
+
+    pub fn new_with_target(pkg_with_target: &str, is_extension: bool) -> Option<Self> {
+        for (pos, _) in pkg_with_target.match_indices('-') {
+            let pkg = &pkg_with_target[0..pos];
+            let target = &pkg_with_target[pos + 1..];
+            if let Some(partial) = PartialTargetTriple::new(target) {
+                if partial.arch.is_some() && partial.os.is_some() && partial.env.is_some() {
```
shepmaster: This specific bit feels like reaching into the guts of `PartialTargetTriple`. Without knowing *any* of the context, I'd have expected this to be a method like `partial.is_valid()`

shepmaster: Maybe even something like `fn validate(self) -> Option<TargetTriple>`?

shepmaster (approved):
The functionality all seems reasonable to me! Just minor comments since I read the code anyway.

src/dist/manifest.rs
kinnison: Partials are valid if any of the parts are available, I'm kinda using it a bit iffily since I want a full triple.

I've added a `impl std::convert::TryFrom<PartialTargetTriple> for TargetTriple` and I'm using that.  Thanks for the idea.

PR #2089, opened 2019-10-26
experiment: use github workflows

dignifiedquire:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+    - uses: actions-rs/toolchain@v1
+      with:
+          toolchain: beta
+          components: rustfmt
+          override: true
+          profile: minimal
+
+    - name: fmt
+      run: cargo fmt --all -- --check
+
```
tesuji, line 146: I think checking formatting on Travis CI is enough.
So we could remove this fmt job.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+  check_clippy:
+    name: Checking clippy
+    runs-on: ubuntu-latest
```
tesuji, line 149: It would be better if clippy runs on windows. We already run it on Ubuntu Travis
```suggestion
    runs-on: windows-latest
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - run: rustup component add clippy
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
tesuji, line 155: I'm not familiar with actions-rs. Why does it need GITHUB_TOKEN?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+        toolchain: ${{ matrix.rust }}
+        target: ${{ matrix.target }}
+        profile: minimal
+        override: true
+
+    - name: build
+      uses: actions-rs/cargo@v1
+      with:
+        command: build
+        args: --release --target ${{ matrix.target }}
```
tesuji, line 33: Here and other tess
```suggestion
        args: --release --locked --target ${{ matrix.target }}
```

.github/workflows/ci.yml
kinnison: The intent is to replace *all* of Travis and Appveyor eventually, this is just the first step.

.github/workflows/ci.yml
kinnison: I think @dignifiedquire was trying to replicate the current clippy checks which do do `--all --all-targets` which ought to be enough.

.github/workflows/ci.yml
kinnison: I think that lets it post results back directly onto the PR

.github/workflows/ci.yml
kinnison: This is definitely wanted.

.github/workflows/ci.yml
tesuji: IMO Travis is good enough for Linux and macOS builds. Travis supports 
cache feature that github actions currectly doesn't. Build queue on Travis is not
bad through we share it with other rust-lang projects.
I believe we should replace Appveyor.

kinnison: Short-term is removal of appveyor yes.  Longer term we want to move from Travis too because I don't think rust-lang gets sponsorship on it any longer.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+    name: Build and test
+    runs-on: ${{ matrix.os }}
+    env:
+      RUSTFLAGS: -Ctarget-feature=+crt-static
+      RUST_BACKTRACE: 1
+    strategy:
+      matrix:
+        os: [windows-latest]
+        rust: [stable]
+        target: [x86_64-pc-windows-msvc, i686-pc-windows-msvc, i686-pc-windows-gnu]
```
tesuji, line 16: We would want specific mingw version to be able to compile.
Take at look at my config here: 
https://github.com/lzutao/junction/blob/d21209cb8c34e0942cb4578937f0b92f76bbbfb5/.github/workflows/ci.yml#L13-L26

kinnison:
With the lack of progress on this, I'm going to close this PR.  I hope to come back to github actions in the future, having experimented a bit on my own code first perhaps.

PR #2091, opened 2019-10-26
Correct a variety of clippy lints

kinnison:
Deal with a number of clippy lints

### DIFF
--- a/tests/mock/dist.rs
+++ b/tests/mock/dist.rs
@@ -414,7 +414,7 @@ impl MockDistServer {
                 .iter()
                 .map(|v| toml::Value::String((**v).to_owned()))
                 .collect();
-            toml_profiles.insert(profile.to_string(), toml::Value::Array(array));
+            toml_profiles.insert((*profile).to_string(), toml::Value::Array(array));
         }
         toml_manifest.insert(String::from("profiles"), toml::Value::Table(toml_profiles));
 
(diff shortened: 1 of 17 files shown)


tests/mock/dist.rs
```diff
@@ -414,7 +414,7 @@ impl MockDistServer {
                 .iter()
                 .map(|v| toml::Value::String((**v).to_owned()))
                 .collect();
-            toml_profiles.insert(profile.to_string(), toml::Value::Array(array));
+            toml_profiles.insert((*profile).to_string(), toml::Value::Array(array));
```
tesuji, line 417: What about using String::from?

kinnison: Probably nothing, I'll admit I went with the recommendation from the compiler :D

tesuji: So str::to_string() use String::from under the hood: https://doc.rust-lang.org/nightly/src/alloc/string.rs.html#2179-2184.
I would file an issue in clippy in this case.

kinnison: I'll leave myself a TODO unless you want to do it.

tesuji: I don't bother me that much. Just a style.

kinnison: Fair enough, I'll bear it in mind next time clippy talks to me in that way :D

PR #2094, opened 2019-10-27
chore: use pointer::cast for safer pointer cast

tesuji:


src/utils/raw.rs
```diff
@@ -206,7 +206,7 @@ fn symlink_junction_inner(target: &Path, junction: &Path) -> io::Result<()> {
         );
 
         let mut data = [0u8; MAXIMUM_REPARSE_DATA_BUFFER_SIZE];
-        let db = data.as_mut_ptr() as *mut REPARSE_MOUNTPOINT_DATA_BUFFER;
+        let db = data.as_mut_ptr().cast::<REPARSE_MOUNTPOINT_DATA_BUFFER>();
         let buf = &mut (*db).ReparseTarget as *mut WCHAR;
```
kinnison, line 210: Why not also do this one?

kinnison (commented):
You appear to have missed (at least) one cast.  Was that deliberate?

src/utils/raw.rs
tesuji: `pointer::cast` can only be applied to pointer type. In this case `&mut (*db).ReparseTarget` is a reference.

kinnison: Aha, I understand.

PR #2097, opened 2019-10-29
rustup_mode.rs: Break up CLI generation to let rustfmt work

kinnison:
For some reason `rustfmt` was hitting a brick wall when formatting the expression which was the majority of `rustup_mode.rs`'s CLI.  This PR breaks that up so that formatting works again.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -148,7 +148,7 @@ pub fn cli() -> App<'static, 'static> {
         .subcommand(
             SubCommand::with_name("dump-testament")
                 .about("Dump information about the build")
-                .setting(AppSettings::Hidden) // Not for users, only CI
+                .setting(AppSettings::Hidden), // Not for users, only CI
         )
         .subcommand(
             SubCommand::with_name("show")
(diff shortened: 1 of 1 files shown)


tesuji:
It's best to create a minimal example and report the problem in rustfmt issue.

kinnison:
@lzutao There already is a rustfmt issue which I'm subscribed to; but the fix isn't slated until rustfmt 2.0 so for now we need the change in rustup :D

src/cli/rustup_mode.rs
```diff
@@ -245,8 +242,11 @@ pub fn cli() -> App<'static, 'static> {
                         .help(TOOLCHAIN_ARG_HELP)
                         .required(false),
                 ),
-        )
-        .subcommand(
+        );
+
+    // We break the app creation here so that rustfmt can cope
+    // If rustfmt ceases to format this block, break it up further
```
payload, line 248: Is there an issue about that in rustfmt?

kinnison: This is indeed a known issue in rustfmt though I was told it's unlikely to be resolved before 2.0

PR #2105, opened 2019-11-01
add 'Windows 10 SDK' text to README in `Working with Rust on Windows`

KaneGreen:
There are always some people asking `Why I can't compile 'hello world'?` in chatroom, while he/she doesn't install `Windows 10 SDK` when installing VS.  

There are `Windows 10 SDK` in the `src\cli\self_update.rs` file.
But `Windows 10 SDK` are not mentioned in the `README.md` file.

I hope adding this text into README will help them to install `Windows 10 SDK` and successfully compile 'hello world'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,8 +468,8 @@ toolchain] use the GNU build.
 When targeting the MSVC ABI, Rust additionally requires an [installation
 of Visual Studio 2013 (or later) or the Visual C++ Build Tools
 2019][vs] so rustc can use its linker. For Visual Studio, make sure to
-check the "C++ tools" option. No additional software installation is
-necessary for basic use of the GNU build.
+check the "C++ tools" and "Windows 10 SDK" option. No additional software
+installation is necessary for basic use of the GNU build.
 
 By default `rustup` on Windows configures Rust to target the MSVC
 ABI, that is a target triple of either `i686-pc-windows-msvc` or
@@ -740,9 +740,9 @@ platform of your choice:
 <a name="vs2019">†</a>
 MSVC builds of `rustup` additionally require an [installation of
 Visual Studio 2019 or the Visual C++ Build Tools 2019][vs]. For
-Visual Studio, make sure to check the "C++ tools" option. No
-additional software installation is necessary for basic use of
-the GNU build.
+Visual Studio, make sure to check the "C++ tools" and "Windows
+10 SDK" option. No additional software installation is necessary
+for basic use of the GNU build.
 
 [vs]: https://visualstudio.microsoft.com/downloads/
 


kinnison:
Is the `Windows 10 SDK` needed for Win7 or Win8 too?  If not, then the text needs adjustment slightly.

mzji:
The `msvc` toolchain always requires a Windows SDK installation, no matter what the target OS version is (since the final binary will be linked to several Windows system dlls directly).

kinnison:
But is it always the Win10 SDK, even for Win7?

mzji:
Different version of Windows SDK has different supported OS versions. In the case of Win10 SDK, [Microsoft document](https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk) shows that as of now, all Windows 10 SDKs do support to target Windows 7 SP1 or higher, so yeah, you **can** use Windows 10 SDK to build binaries for Windows 7 SP1.

kinnison:
Can they be *installed* on Win7 SP1 though?

mzji:
<strike>Seems like Windows 10 SDKs cannot be installed on Windows 7 SP1. Then usually we install the Windows 8.1 SDK instead, which is supported to be installed on Windows 7 SP1.</strike> Now KaneGreen showed us that Windows 10 SDKs can be installed on Windows 7 SP1.

mzji:
Another thing should be mentioned, is that only Visual Studio 2017 or lower version supports Windows 8.1 SDK; Windows 10 SDKs can only be used with Visual Studio 2019 (or higher version, maybe).

KaneGreen:
The words "Windows 10 SDK" is refer to [Rust README file].
Windows 8.1 user have "Windows 8.1 SDK" from VS 2017. And this can be installed on Windows 10 and works, too. (And "Windows 8.1 SDK" use less disk space than "Windows 10 SDK").
So, "Windows 10 SDK" should be "corresponding Windows SDK". But here I use "Windows 10 SDK" for consistency to [Rust README file].

---
If necessary, I may need some time to test if Windows 7 needs to install corresponding Windows SDK, for there is no computer with Windows 7 around me.

[Rust README file]: https://github.com/rust-lang/rust/blob/master/README.md#msvc

KaneGreen:
> Can they be _installed_ on Win7 SP1 though?

**Yes!**

I have already installed Visual Studio 2019 and "C++ tools" with "Windows 10 SDK" on Windows 7 Enterprise (SP1 64bit).
Rust (stable-x86_64-pc-windows-msvc) compiled `hello world` successfully.

Capture

------
In addition, I recommend using Visual Studio 2019 instead of any earlier version of VS. 
Because Microsoft has improved VS 2019 installer, and VS 2019 is easier to install and uninstall than previous versions.

kinnison:
Thank you, that's good enough for me.

PR #2107, opened 2019-11-02
build: remove redundant codegen-units=1

tesuji:
LTO is disabled if codegen units is 1


tesuji:
Hm, I'm thinking about another way is to remove lto option.
Because when testing locally, the binary size with this path is larger
than in master. I will try to measure the opposite way.

tesuji:
The binary size of when removing lto is the same as this PR.
Maybe something wrong with the documentation.
I will close this PR in this case.

PR #2108, opened 2019-11-03
for toolchains without cargo, set up environment to match rustc, not cargo

### LINKED ISSUE #1575: Handling of rustc-sysroot .so files differs when sysroots do not come with their own cargo
When I install miri (`cargo +nightly install --git https://github.com/solson/miri/ miri`) with the `nightly` toolchain, I can later execute `cargo +nightly miri --help` just fine.  However, when I do the same thing with a toolchain obtained via `rustup toolchain link` (a locally-compiled stage 2 sysroot), that does not work: After replacing `+nightly` by `+stage2`, all I get is
```
/home/r/.cargo/bin/cargo-miri: error while loading shared libraries: librustc_mir-2d8f89b6ce1d8c91.so: cannot open shared object file: No such file or directory
```

Interestingly, toolchains installed via https://github.com/kennytm/rustup-toolchain-install-master behave like linked ones: They also result in the above error.

What both of these toolchains (linked ones and rustup-toolchain-install-master) have in common is that they don't come with their own cargo.

I don't know how the responsibility of setting `LD_LIBRARY_PATH` (assuming that is what is missing here) is split between `rustup` and `cargo`, and I have no idea which `cargo` `rustup` is even using when the toolchain does not have its own.  But somewhere, something doesn't quite fit together.

(Calling the `cargo-miri` binary directly shows such an error in both cases, but when I am using nightly, it seems that `LD_LIBRARY_PATH` is set properly when using `cargo miri`.)

RalfJung:
This implements the suggestion by @ehuss for how to fix https://github.com/rust-lang/rustup.rs/issues/1575.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -378,7 +378,7 @@ impl<'a> Toolchain<'a> {
             src_file
         };
         let mut cmd = Command::new(exe_path);
-        self.set_env(&mut cmd);
+        primary_toolchain.set_env(&mut cmd); // set up the environment to match rustc, not cargo
         cmd.env("RUSTUP_TOOLCHAIN", &primary_toolchain.name);
         Ok(cmd)
     }


kinnison (approved):
This matches the recommendation AIUI.  Thank you Ralf.

PR #2116, opened 2019-11-10
Cleanly report when rust-docs isn't installed

### LINKED ISSUE #901: rustup doc should deal with absent doc instead of silently failing
On a new install of rustup, running `rustup doc` fails silently if the documentation is not installed: the browser does not open, and there is no error message.

Would it be OK to add an error message, or to automatically run `rustup component add rust-docs` ?

This is tangentially related to #891 — if the doc is installed by default, this issue would be moot.

kinnison:
When the user runs `rustup doc` on a toolchain which doesn't
have the `rust-docs` component installed, we now report more
nicely that the user ought to try and install the component.

Fixes #901

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1277,6 +1277,19 @@ const DOCS_DATA: &[(&str, &str, &str,)] = &[
 
 fn doc(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
     let toolchain = explicit_or_dir_toolchain(cfg, m)?;
+    for cstatus in &toolchain.list_components()? {
+        if cstatus.component.short_name_in_manifest() == "rust-docs" && !cstatus.installed {
+            info!(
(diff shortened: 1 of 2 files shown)


kinnison:
For reference the UX is:

```console
somniloquy% rustup component remove rust-docs                                                     
info: removing component 'rust-docs'
somniloquy% rustup doc                                                                            
info: `rust-docs` not installed in toolchain `stable-x86_64-unknown-linux-gnu`
info: To install, try `rustup component add --toolchain stable-x86_64-unknown-linux-gnu rust-docs`
error: unable to view documentation which is not installed
zsh: exit 1     rustup doc
somniloquy% rustup component add --toolchain stable-x86_64-unknown-linux-gnu rust-docs
 11.7 MiB /  11.7 MiB (100 %)   6.6 MiB/s in  3s ETA:  0s
 11.7 MiB /  11.7 MiB (100 %)   2.8 MiB/s in  3s ETA:  0s
somniloquy%                                              
```

Issue #2076, opened 2019-10-20
`rustup doc` can't distinguish between keywords and primitives with the same name


cstyles:
**Problem**
First of all, I just want to say that I really like the recent addition of topics to the `rustup doc` feature! While playing around with it, I noticed that if I run `rustup doc fn`, it brings up the documentation for the keyword `fn`. However, there's also a primitive `fn` and I can't figure out any way to open that documentation page specifically. Am I missing something or is it just not possible?

This would also, in theory, affect any other naming collisions, not just with keywords and primitives (e.g., if some `Trait` and `Enum` had the same name and scope).

**Steps**

1. Run `rustup doc fn`
2. Observe that the `keyword.fn.html` page is opened in a browser.

**Possible Solution(s)**
I'm not super jazzed about this idea but we could add flags for `--keyword`, `--primitive`, `--macro`, etc. to allow the user to tell `rustup` exactly what they're looking for. But it would only ever be helpful in these rare situations so I don't know if it's worth A) the effort and B) making the UI more complex. It's also worth mentioning that on the page for the keyword `fn`, the first thing the viewer will see in the left-hand side panel is a list of primitives. So it probably wouldn't be too much of an inconvenience for them to just find `fn` there and click on it.

**Notes**

Output of `rustup --version`: `rustup 1.20.0 (2019-10-15)`
Output of `rustup show`:
```
Default host: x86_64-apple-darwin
rustup home:  /Users/collin.styles/.rustup

stable-x86_64-apple-darwin (default)
rustc 1.38.0 (625451e37 2019-09-23)
```

kinnison:
I'm not sure I want to add more flags, but I also don't know how we'd distinguish them oherwise.  This is an interesting idea, but as you say it's not terribly hard for someone to navigate to the other concept.  I've re-labelled this help-wanted and enhancement rather than bug.

cstyles:
The search function for the rust docs allows you to preface a search with something like `fn:` to restrict the search to just functions. This also works with `mod`, `struct`, `enum`, `trait`, `type`, `macro`, `const`. Perhaps we can do something like that here?

https://github.com/rust-lang/rust/blob/8e0007f829661e57d008d2e908c95f6e84b04b25/src/librustdoc/html/layout.rs#L134-L145

kinnison:
Oh that is a very interesting approach yes.  This feature was originally written by @dorfsmay so perhaps they might have some input on how to do this?

dorfsmay:
[...]

PR #2119, opened 2019-11-11
Force type of keyword with colon (issue #2076)

dorfsmay:
For issue #2076


kinnison (approved):
This looks excellent, thank you Yves!

PR #2121, opened 2019-11-11
Kinnison/retry downloads

### LINKED ISSUE #1243: Rustup failed (probably because of a spurious network error), but then keeps failing consistently after that
I tried to update with `rustup update`. It failed with `error: component download failed for rust-std-x86_64-unknown-linux-musl`, and I tried again – it keeps failing:

```
~ $ rustup update
info: syncing channel updates for 'nightly-x86_64-apple-darwin'
info: latest update on 2017-08-26, rust version 1.21.0-nightly (2aeb5930f 2017-08-25)
info: downloading component 'rust-std' for 'x86_64-unknown-linux-musl'
error: component download failed for rust-std-x86_64-unknown-linux-musl

  nightly-x86_64-apple-darwin update failed - rustc 1.21.0-nightly (2c0558f63 2017-08-24)
```

Randomly trying with `rustup update nightly`, it spouts a more detailed error (I'm not sure, but the difference in the level of detail could be also considered as a bug?):

```
info: syncing channel updates for 'nightly-x86_64-apple-darwin'
info: latest update on 2017-08-26, rust version 1.21.0-nightly (2aeb5930f 2017-08-25)
info: downloading component 'rust-std' for 'x86_64-unknown-linux-musl'
error: component download failed for rust-std-x86_64-unknown-linux-musl
info: caused by: could not download file from 'https://static.rust-lang.org/dist/2017-08-26/rust-std-nightly-x86_64-unknown-linux-musl.tar.xz' to '/Users/um003415/.rustup/downloads/e6108b2fa9ad25cb303e21e216a557e11a2a633586b9856de84064fade0651fc.partial'
info: caused by: [33] Requested range was not delivered by the server (HTTP server doesn't seem to support byte ranges. Cannot resume.)
info: backtrace:

stack backtrace:
   0:        0x1095998ae - backtrace::backtrace::trace::h44539d42e8c86729
   1:        0x10959a12c - backtrace::capture::Backtrace::new::h5f3de869c0349b3d
   2:        0x1095993d6 - error_chain::make_backtrace::h34ae912053f4a126
   3:        0x109560e86 - download::curl::download::hce069bb0f14d5dd2
   4:        0x10955f240 - download::download_to_path_with_backend::h117f20f63fa75168
   5:        0x1094d8057 - rustup_utils::utils::download_file_with_resume::h523bb0f3a635163f
   6:        0x1094a2a8f - rustup_dist::download::DownloadCfg::download::h11601c9bf443ac0a
   7:        0x10949b7d1 - rustup_dist::manifestation::Manifestation::update::h03e87519f12f2de4
   8:        0x10948611f - rustup_dist::dist::update_from_dist_::h9075ed00e47841fe
   9:        0x1094852d9 - rustup_dist::dist::update_from_dist::h68c861ce5520c639
  10:        0x109448290 - rustup::toolchain::Toolchain::install::h8d5d75e7a0a26413
  11:        0x109449011 - rustup::toolchain::Toolchain::install_from_dist_inner::h271107f0a7dae968
  12:        0x109448ca3 - rustup::toolchain::Toolchain::install_from_dist::hda92f29cbfc4b7a4
  13:        0x10939b46e - rustup_init::rustup_mode::update::hdf1cfff907031852
  14:        0x109388c38 - rustup_init::rustup_mode::main::hd56e0e782a4ea323
  15:        0x1093b4370 - rustup_init::run_rustup::hc3aac5a16c6bde4a
  16:        0x1093b3a39 - rustup_init::main::h07c1d4d8cd7703cf
  17:        0x1095c6daa - __rust_maybe_catch_panic
  18:        0x1095c61b6 - std::rt::lang_start::ha1f4f04346e51ef5
```

Deleting the `.partial` file allowed the update to succeed.

I'm using macOS 10.12.6.

### LINKED ISSUE #1667: Retry downloads

### LINKED ISSUE #2071: Rustup doesn't handle HTTP 416 errors when downloading toolchains
**Problem**

While setting up a Crater agent the container running Crater was killed for *reasons* in the middle of a rustup download. After starting the container again rustup kept failing with this error message:

```
error: component download failed for rust-std-x86_64-unknown-linux-gnu
error: caused by: could not download file from 'https://static.rust-lang.org/dist/2019-09-26/rust-std-1.38.0-x86_64-unknown-linux-gnu.tar.xz' to '/workspace/rustup-home/downloads/51b334337804baeff7524e5496d396f254894d7529860d236975e9ed8fcca371.partial'
error: caused by: http request returned an unsuccessful status code: 416
```

**Steps**

Note that these steps didn't cause the original bug, but they reproduce the same symptoms.

1. Remove all partial downloads:
```
rm -f ~/.rustup/downloads/*.partial
```
2. Install a toolchain, but kill rustup in the middle of a download.
3. Increase the size of the only `.partial` file to 1 GB, way above the total size of the file:
```
fallocate -l 1G ~/.rustup/downloads/*.partial
```
4. Try to reinstall the same toolchain, it will fail with:
```
error: component download failed for rust-std-x86_64-unknown-linux-gnu
error: caused by: could not download file from 'https://static.rust-lang.org/dist/2019-01-01/rust-std-nightly-x86_64-unknown-linux-gnu.tar.xz' to '/home/pietro/.rustup/downloads/63e41b2768326b80fbf9bdef657bb3c717319538e75b52dce305bc4363dce1f6.partial'
error: caused by: http request returned an unsuccessful status code: 416
```

**Possible Solution(s)**

The [HTTP 416](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416) status code means the `content-range` requested to the server was not satisfiable, which in our case means a corrupted (or overly large like in my repro steps) partial file. Rustup could handle this better by ignoring the partial and starting to download from scratch when it receives a 416 status code.

**Notes**

Output of `rustup --version`:

```
rustup 1.20.2 (13979c968 2019-10-16)
```

Output of `rustup show`: *not relevant*

kinnison:
In an attempt to close #1722 as a way to fix #1667, fix #1243, and to fix #2071 more fully, add support for cleanly retrying downloads a fixed number of times.

Also this contains a small cleanup which means we finalise the download tracker properly on errors

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,6 +140,13 @@ than 1, it is clamped to 1 at minimum.
 This is not meant for use by users, but can be suggested in diagnosing an issue
 should one arise with the backtrack limits.
 
+### `RUSTUP_MAX_RETRIES`
+
+When downloading a file, rustup will retry the download a number of times. The
+default is 3 times, but if this variable is set to a valid usize then it is the
+max retry count. A value of `0` means no retries, thus the default of `3` will
+mean a download is tried a total of four times before failing out.
+
 ### `RUSTUP_BACKTRACE`
 
 By default while running tests, we unset some environment variables that will
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -243,11 +243,12 @@ fn download_file_(
         (Backend::Reqwest, Notification::UsingReqwest)
     };
     notify_handler(notification);
-    download_to_path_with_backend(backend, url, path, resume_from_partial, Some(callback))?;
+    let res =
+        download_to_path_with_backend(backend, url, path, resume_from_partial, Some(callback));
 
     notify_handler(Notification::DownloadFinished);
 
-    Ok(())
+    res.map_err(|e| e.into())
 }
 
 pub fn parse_url(url: &str) -> Result<Url> {
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -172,6 +172,9 @@ pub fn download_file_with_resume(
         Ok(_) => Ok(()),
         Err(e) => {
             let is_client_error = match e.kind() {
+                // Specifically treat the bad partial range error as not our
+                // fault in case it was something odd which happened.
+                ErrorKind::Download(DEK::HttpStatus(416)) => false,
                 ErrorKind::Download(DEK::HttpStatus(400..=499)) => true,
                 ErrorKind::Download(DEK::FileNotFound) => true,
                 _ => false,
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -12,6 +12,8 @@ use crate::dist::prefix::InstallPrefix;
 use crate::dist::temp;
 use crate::errors::*;
 use crate::utils::utils;
+use retry::delay::NoDelay;
+use retry::{retry, OperationResult};
 use std::path::Path;
 
 pub const DIST_MANIFEST: &str = "multirust-channel-manifest.toml";
@@ -151,6 +153,13 @@ impl Manifestation {
         let mut things_to_install: Vec<(Component, Format, File)> = Vec::new();
         let mut things_downloaded: Vec<String> = Vec::new();
         let components = update.components_urls_and_hashes(new_manifest)?;
+
+        const DEFAULT_MAX_RETRIES: usize = 3;
+        let max_retries: usize = std::env::var("RUSTUP_MAX_RETRIES")
+            .ok()
+            .and_then(|s| s.parse().ok())
+            .unwrap_or(DEFAULT_MAX_RETRIES);
+
         for (component, format, url, hash) in components {
             notify_handler(Notification::DownloadingComponent(
                 &component.short_name(new_manifest),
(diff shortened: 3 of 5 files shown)


src/utils/utils.rs
```diff
@@ -243,11 +246,12 @@ fn download_file_(
         (Backend::Reqwest, Notification::UsingReqwest)
     };
     notify_handler(notification);
-    download_to_path_with_backend(backend, url, path, resume_from_partial, Some(callback))?;
+    let res =
+        download_to_path_with_backend(backend, url, path, resume_from_partial, Some(callback));
 
     notify_handler(Notification::DownloadFinished);
```
payload, line 252: Removing the question mark operator will now call notify_handler DownloadFinished also in case there was an error. I would imagine everyone who listens to DownloadFinished has some assumption at that point which may not hold in case the error case happened. Is this a problem? I don't immediately see from the PR how this case is treated.

payload: Commit message makes it clearer. Talked in person too.

kinnison:
@payload Thanks for the review, is there anything else you wanted to query?

CONTRIBUTING.md
```diff
@@ -140,6 +140,13 @@ than 1, it is clamped to 1 at minimum.
 This is not meant for use by users, but can be suggested in diagnosing an issue
 should one arise with the backtrack limits.
 
+### `RUSTUP_MAX_RETRIES`
+
+When downloading a file, rustup will retry the download a number of times. The
+default is 3 times, but if this variable is set to a valid usize then it is the
+max retry count. A value of `0` means no retries, thus the default of `3` will
+mean a download is tried a total of four times before failing out.
```
matklad, line 148: Hm, shouldn't this go to Readme instead?

kinnison: My hope is that noone will need to tweak this, if we decide to expose it to users then it becomes a thing we have to support long term, rather than something where in an issue we can say "Please try this and see if it helps" to help decide if we need to change the defaults.

matklad: Ahhh...

I think a valid use-case might be increasing the limit for CI (where networking can get unstable). Cargo has a knob for it in `.cargo/config`, `[net] retries = 92`.

kinnison: CI is a place where people get *even more* angry if I change things.  Humans seem more prepared to change their own workflows than to change CI setups which "have always worked".

PR #2122, opened 2019-11-11
build: make clippy faster by using checking before that operation

tesuji:
It wins nearly one minute after this PR.


kinnison:
Is the expectation that `cargo check` is faster than `cargo clippy` on the dependencies?

tesuji:
Yes, that is.

kinnison:
Thanks.

PR #2126, opened 2019-11-12
rustup_mode: add toolchain install --allow-downgrade option

### LINKED ISSUE #2067: [Feature Request] Downgrading nightly toolchains
As talked on Discord with @kinnison 
I think an interesting usecase is to be able to *downgrade* your nightly feature to one that supports specific toolchains.
i.e. 
`rustup toolchain install nightly --downgrade -c rustc -c rust-std -c cargo -c rustfmt -c clippy -c rls -c rust-analysis -c miri`

This can make adding new components to your local toolchain easier (then tools like miri could add this to their docs and you won't need to uninstall latest nightly before installing this).

Another thing that this made me think of, maybe a way to *aggregate* components? 
For example, I install `rustup toolchain install nightly -c cargo` and now I want to add miri support so by doing something like `rustup toolchain install nightly --downgrade --add -c miri` It will now find me a toolchain with both clippy and miri.

payload:
In case you have an already installed nightly toolchain
and you want to install a nightly toolchain with a component
which is not available in the installed toolchain,
rustup is searching for the most recent nightly supporting
all requested components and which is not older than the currently
installed nightly.

With this option set it will also search for toolchains older
than the currently installed one.

- [x] add a test for downgrading

Should fix #2067

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -319,6 +319,12 @@ pub fn cli() -> App<'static, 'static> {
                                 .help("Force an update, even if some components are missing")
                                 .long("force")
                                 .takes_value(false),
+                        )
+                        .arg(
+                            Arg::with_name("allow-downgrade")
+                                .help("Allow rustup to downgrade the toolchain to satisfy your component choice")
+                                .long("allow-downgrade")
+                                .takes_value(false),
                         ),
                 )
                 .subcommand(
@@ -822,7 +828,12 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<()> {
                     .values_of("targets")
                     .map(|v| v.collect())
                     .unwrap_or_else(Vec::new);
-                Some(toolchain.install_from_dist(m.is_present("force"), &components, &targets)?)
+                Some(toolchain.install_from_dist(
+                    m.is_present("force"),
+                    m.is_present("allow-downgrade"),
+                    &components,
+                    &targets,
+                )?)
             } else if !toolchain.exists() {
                 return Err(ErrorKind::InvalidToolchainName(toolchain.name().to_string()).into());
             } else {
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -580,6 +580,7 @@ pub fn update_from_dist<'a>(
     profile: Option<Profile>,
     prefix: &InstallPrefix,
     force_update: bool,
+    allow_downgrade: bool,
     old_date: Option<&str>,
     components: &[&str],
     targets: &[&str],
@@ -601,6 +602,7 @@ pub fn update_from_dist<'a>(
         profile,
         prefix,
         force_update,
+        allow_downgrade,
         old_date,
         components,
         targets,
@@ -622,6 +624,7 @@ fn update_from_dist_<'a>(
     profile: Option<Profile>,
     prefix: &InstallPrefix,
     force_update: bool,
+    allow_downgrade: bool,
     old_date: Option<&str>,
     components: &[&str],
     targets: &[&str],
@@ -644,7 +647,8 @@ fn update_from_dist_<'a>(
         Some(if provided < 1 { 1 } else { provided })
     };
 
-    // We never want to backtrack further back than the nightly that's already installed.
+    // In case there is no allow-downgrade option set
+    // we never want to backtrack further back than the nightly that's already installed.
     //
     // If no nightly is installed, it makes no sense to backtrack beyond the first ever manifest,
     // which is 2014-12-20 according to
@@ -652,13 +656,15 @@ fn update_from_dist_<'a>(
     //
     // We could arguably use the date of the first rustup release here, but that would break a
     // bunch of the tests, which (inexplicably) use 2015-01-01 as their manifest dates.
-    let first_manifest = old_date
-        .map(|date| {
-            Utc.from_utc_date(
-                &NaiveDate::parse_from_str(date, "%Y-%m-%d").expect("Malformed manifest date"),
-            )
-        })
-        .unwrap_or_else(|| Utc.from_utc_date(&NaiveDate::from_ymd(2014, 12, 20)));
+    let first_manifest = Utc.from_utc_date(&NaiveDate::from_ymd(2014, 12, 20));
+    let old_manifest = old_date
+        .and_then(|date| utc_from_manifest_date(date))
+        .unwrap_or(first_manifest);
+    let last_manifest = if allow_downgrade {
+        first_manifest
+    } else {
+        old_manifest
+    };
 
     loop {
         match try_update_from_dist_(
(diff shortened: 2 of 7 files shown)


src/dist/dist.rs
```diff
@@ -644,21 +647,21 @@ fn update_from_dist_<'a>(
...
-                &NaiveDate::parse_from_str(date, "%Y-%m-%d").expect("Malformed manifest date"),
-            )
-        })
-        .unwrap_or_else(|| Utc.from_utc_date(&NaiveDate::from_ymd(2014, 12, 20)));
+    let first_manifest = Utc.from_utc_date(&NaiveDate::from_ymd(2014, 12, 20));
+    let old_manifest = if allow_downgrade || old_date == None {
+        first_manifest
+    } else {
+        utc_from_manifest_date(old_date.unwrap())
+    };
```
kinnison: I think this change alters the semantics of a malformed `old_date` from "oh well, use the first manifest" to "panic"

src/dist/dist.rs
```diff
@@ -926,3 +917,9 @@ fn dl_v1_manifest<'a>(download: DownloadCfg<'a>, toolchain: &ToolchainDesc) -> R
 
     Ok(urls)
 }
+
+fn utc_from_manifest_date(date_str: &str) -> Date<Utc> {
+    let msg = format!("Malformed manifest date: {:?}", date_str);
+    let date = NaiveDate::parse_from_str(date_str, "%Y-%m-%d").expect(&msg);
```
kinnison: Seems odd to do this work rather than the old pattern of:

```rust
foo.unwrap_or_else(||panic!(...))
```

kinnison (commented):
Some initial feedback.

src/dist/dist.rs
payload: Ah right, good catch.

src/dist/dist.rs
payload: I think `unwrap_or_else(||panic!(...))` is the same as `expect(format!(...))` but the expect is more idiomatic. I think I would use or_else to do something else than the normal panic of an unwrap.

src/dist/dist.rs
payload: I think a malformed `old_date` would have not been catched by the `unwrap_or_else` because the `expect` would have paniced. ;) But I'll fix it

src/dist/dist.rs
```diff
@@ -918,8 +922,8 @@ fn dl_v1_manifest<'a>(download: DownloadCfg<'a>, toolchain: &ToolchainDesc) -> R
...
 }
 
-fn utc_from_manifest_date(date_str: &str) -> Date<Utc> {
-    let msg = format!("Malformed manifest date: {:?}", date_str);
-    let date = NaiveDate::parse_from_str(date_str, "%Y-%m-%d").expect(&msg);
-    Utc.from_utc_date(&date)
+fn utc_from_manifest_date(date_str: &str) -> Option<Date<Utc>> {
+    NaiveDate::parse_from_str(date_str, "%Y-%m-%d")
+        .ok()
+        .and_then(|date| Some(Utc.from_utc_date(&date)))
```
payload: use map instead of and_then Some

payload:
The failed CI run ended with the message:
```---- env_override_beats_file_override stdout ----

Copying from /checkout/target/x86_64-unknown-linux-gnu/release/rustup-init to /checkout/target/x86_64-unknown-linux-gnu/tests/running-test-ryUcXF/rustup-exeTGkmi4/rustup

running "/checkout/target/x86_64-unknown-linux-gnu/tests/running-test-ryUcXF/rustup-exeTGkmi4/rustup" "set" "default-host" "x86_64-unknown-linux-gnu"

thread 'env_override_beats_file_override' panicked at 'failed to run test command: Os { code: 26, kind: Other, message: "Text file busy" }', src/libcore/result.rs:1165:5
```

I consider this flakyness of the underlying system, probably some filesystem hickup. I don't know if it is probably related to code in rustup set default-host which could be more robust against yet uncertain filesystem races. 

@kinnison what do you think of it?

kinnison:
@payload Yes that's a transient issue I only ever tend to see on Travis.  As such I've not managed to track it down to where it actually happens to add a retry in for that case.

src/dist/dist.rs
kinnison: The thing about the `.expect()` is that the `format!()` is eager, whereas in the `.unwrap_or_else()` the `panic!()` is lazy.  It feels like change for change's sake.

kinnison (commented):
For the test, I'd be tempted to add the `--allow-downgrade` variant to the end of the longer test, so that the scenario plays out in full as:

1. Install
2 Check for rls -- should not be present
3. Attempt to add rls -- this should fail
4. Attempt to add rls with downgrate permitted -- this should succeed
5. Check for rls -- should be present
6. Check hash for rustc -- should have *downgraded*

src/cli/rustup_mode.rs
```diff
@@ -319,6 +319,12 @@ pub fn cli() -> App<'static, 'static> {
                                 .help("Force an update, even if some components are missing")
                                 .long("force")
                                 .takes_value(false),
+                        )
+                        .arg(
+                            Arg::with_name("allow-downgrade")
+                                .help("Installs the most recent toolchain which supports all required components, even than this is older than the currently installed toolchain")
```
kinnison: This string is very long, perhaps:

```suggestion
                                .help("Allow rustup to downgrade the toolchain to satisfy your component choice")
```

tests/cli-v2.rs
```diff
@@ -1374,3 +1374,53 @@ fn check_pgp_keys() {
...
+                "nightly",
+                "--no-self-update",
+                "-c",
+                "rls",
+            ],
+            &format!(
+                "component 'rls' for target '{}' is unavailable for download for channel nightly",
+                trip,
+            ),
+        );
```
kinnison, line 1431: Perhaps here add a check that the `rls` component is not installed  You can do that with:

```rust
expect_component_not_executable(config, "rls");
```

tests/cli-v2.rs
```diff
@@ -1374,3 +1374,53 @@ fn check_pgp_keys() {
...
+        expect_ok(
+            config,
+            &[
+                "rustup",
+                "toolchain",
+                "install",
+                "nightly",
+                "--no-self-update",
+            ],
+        );
```
kinnison, line 1412: After this, I'd add an expect to check the rustc version is the version we expect (i.e. one without rls)

tests/cli-v2.rs
```diff
@@ -1374,3 +1374,53 @@ fn check_pgp_keys() {
...
+                "toolchain",
+                "install",
+                "nightly",
+                "--no-self-update",
+                "-c",
+                "rls",
+                "--allow-downgrade",
+            ],
+        );
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-nightly-2");
```
kinnison, line 1448: While this is useful to check that it did downgrade, that'd be clearer if you'd done the equivalent above as commented.  Also you could usefully assert that the `rls` component is now available via `expect_component_executable()`

kinnison (commented):
Pretty close to mergeable -- some more test ideas.  Nice job :+1:

tests/cli-v2.rs
payload: Although I added it I would rather see that in a different testcase which checks if toolchain install does not change anything without allow-downgrade. When I look at the subcommand toolchain install I see 7 arguments and would expect at least as many as 7 tests at the level of cli-v2.rs, but there are not. Or else I don't know yet about all the test levels. Well that's obvious :)

tests/cli-v2.rs
payload: I like the expect_component_executable check much more than the version check. Thx

tests/cli-v2.rs
payload: Yes, I add that. It is though like a sanity check to find out if our test setup is based on the right assumptions.

kinnison (commented):
This looks pretty good now.  Could you please rebase to clean up the commits?  I'd like to see a commit for the code and a commit for the tests at minimum.  Once that's done, I'll give it a final once-over but I think this *very* close to mergeable

payload:
@kinnison squashed it together. I have put the code change and its test together. When you split the test from the implementation it is less clear what belongs together. One reason to split test from implementation is of course to put the test before the implementation so you see the test failing first. Do you mean to put a separate test commit before the implementation commit?

PR #2132, opened 2019-11-17
Resolve potential future shock (x.yyy.zz)

### LINKED ISSUE #2129: Rustup is going to break in 2026
**Problem**

A clone of Alex Crichton just time-travelled back from the year 2026 to tell me to file this issue (they can't file the issue themselves or they risk causing a paradox). 

They said we just released 1.100.0, and everyone was all excited, but as soon as it was released, everyone was all sad because this happened:

```
rustup toolchain install 1.100.0
error: invalid toolchain name: '1.100.0'
```

Everyone expected rustup to download and install the 1.100.0 version, but instead they got this error message, despite the version being released.

**Steps**

1. Wait until 2026
2. Release 1.100.0
3. `rustup toolchain install 1.100.0`

**Possible Solution(s)**

One of the Alex Crichton clones in 2026 was able to quickly track down the issue to this regex: https://github.com/rust-lang/rustup/blob/9309d66429a879e3c1e919f410593627df69d32d/src/dist/dist.rs#L285-L286

Is there a reason we can't change these two regexen to `\d+\.\d+\.\d+` to be future-proof into 2130 and also support the unfortunate case that we might need to release 10 patch versions in a cycle?

I have not yet been visited by any clones from any alternate futures in which we make this fix, so it seems fine.

**Notes**

Output of `rustup --version`: The Alex Crichton clone didn't want to reveal any more information than absolutely necessary to avoid polluting the timeline
Output of `rustup show`: ditto

kinnison:
This is intended to fix #2129 which suggests we have seven years until the version'pocalypse.

cc @carols10cents @Diggsey

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -21,6 +21,25 @@ pub static DEFAULT_DIST_SERVER: &str = "https://static.rust-lang.org";
 // Deprecated
 pub static DEFAULT_DIST_ROOT: &str = "https://static.rust-lang.org/dist";
 
+// The channel patterns we support
+static TOOLCHAIN_CHANNELS: &[&str] = &[
+    "nightly",
+    "beta",
+    "stable",
+    // Allow from 1.0.0 through to 9.999.99
+    r"\d{1}\.\d{1,3}\.\d{1,2}",
+];
+
+lazy_static! {
+    static ref TOOLCHAIN_CHANNEL_PATTERN: String = format!(
+        r"^({})(?:-(\d{{4}}-\d{{2}}-\d{{2}}))?(?:-(.*))?$",
+        TOOLCHAIN_CHANNELS.join("|")
+    );
+    // Note this regex gives you a guaranteed match of the channel (1)
+    // and an optional match of the date (2) and target (3)
+    static ref TOOLCHAIN_CHANNEL_RE: Regex = Regex::new(&TOOLCHAIN_CHANNEL_PATTERN).unwrap();
+}
+
 // A toolchain descriptor from rustup's perspective. These contain
 // 'partial target triples', which allow toolchain names like
 // 'stable-msvc' to work. Partial target triples though are parsed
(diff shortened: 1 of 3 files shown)


src/dist/dist.rs
```diff
@@ -21,6 +21,25 @@ pub static DEFAULT_DIST_SERVER: &str = "https://static.rust-lang.org";
...
+    // Allow from 1.0.0 through to 9.999.99
+    r"\d{1}\.\d{1,3}\.\d{1,2}",
+];
+
+lazy_static! {
+    static ref TOOLCHAIN_CHANNEL_PATTERN: String = format!(
+        r"^({})(?:-(\d{{4}}-\d{{2}}-\d{{2}}))?(?:-(.*))?$",
+        TOOLCHAIN_CHANNELS.join("|")
+    );
+    // Note this regex gives you a guaranteed match of the channel (1)
```
Diggsey, line 38: Instead of writing this in a comment and referring to numeric capture group IDs, we could encode this in the type system, eg.

```rust
struct ParsedToolchainDesc {
    channel: String,
    date: Option<String>,
    target: Option<String>,
}

fn parse_toolchain_desc(input: &str) -> Result<ParsedToolchainDesc, Error> {
    ... local regex and extraction code ...
}
```

... and then this function could be used by both `FromStr` implementations. It would also make it easy to unit test the parsing function.

kinnison: You are quite right, and have seen through to my transparent attempt to leave nice refactoring jobs open for newcomers to the project.  If you'd rather I do this myself then I'm happy to, otherwise my intent was to file an issue describing something (perhaps in less concrete terms than your comment) to encourage someone to come and make a niceness happen as a way into the codebase.  In your view, should I make the change, or should I be filing that issue?  I'm happy either way.

payload (approved):

src/dist/dist.rs
kinnison: After two weeks I filed #2142

PR #2137, opened 2019-11-22
Fix a broken link and change links from http to https [ci skip]

giraffate:
`http://www.catch22.net/tuts/self-deleting-executables` was a broken link. The other links I fixed can use https.


kinnison:
Looks good, thanks.

Issue #2136, opened 2019-11-20
The `complete` profile really should warn the user


kinnison:
We have a lot of people who want to install `complete` despite it really being for Rust developers rather than those who develop *with* Rust.

As a stop-gap, since removing `complete` would be a bit of a pain, it'd be useful if `rustup` could warn whenever `complete` is used to install a toolchain.  That'd be fairly lightweight and could mitigate a lot of issues which might otherwise be raised.

This should be a very easy first issue for someone, and I'd be happy to mentor them on #wg-rustup on our discord.

kinnison:
If you wish to take on this work, please state so on this issue, and then find me on Discord to discuss it.  That way we won't have multiple people trying to solve the same thing at the same time.  Thanks.

kinnison:
In addition, https://github.com/rust-lang/rustup#profiles misrepresents the `complete` profile a little as being of use for general developers.  We probably need to fix that as part of any PR to resolve this issue.

Stupremee:
I can take care of that. @kinnison I can ping you in discord

Stupremee:
I guess this can be closed now

kinnison:
Closed in #2138

svend:
I found https://github.com/rust-lang/rustup#profiles confusing. I picked the complete profile, since I wanted IDE support out of the box.

> The complete profile includes all the components available through rustup, including miri and IDE integration tools (rls and rust-analysis).

PR #2138, opened 2019-11-23
Warn if complete profile is used

Stupremee:
This change was requested in #2136.
Whenever a user downloads a toolchain with the complete profile, 
a warn message will be printed.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -805,6 +805,9 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<()> {
         cfg.set_profile_override(p);
     }
     let cfg = &cfg;
+    if cfg.get_profile()? == Profile::Complete {
+        warn!("{}", common::WARN_COMPLETE_PROFILE);
+    }
     if let Some(names) = m.values_of("toolchain") {
         for name in names {
             update_bare_triple_check(cfg, name)?;
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -115,6 +115,10 @@ pub fn main() -> Result<()> {
         targets: &targets,
     };
 
+    if profile == "complete" {
+        warn!("{}", common::WARN_COMPLETE_PROFILE);
+    }
+
     self_update::install(no_prompt, verbose, quiet, opts)?;
 
     Ok(())
--- a/tests/cli-inst-interactive.rs
+++ b/tests/cli-inst-interactive.rs
@@ -3,7 +3,8 @@
 pub mod mock;
 
 use crate::mock::clitools::{
-    self, expect_stdout_ok, set_current_dist_date, Config, SanitizedOutput, Scenario,
+    self, expect_stderr_ok, expect_stdout_ok, set_current_dist_date, Config, SanitizedOutput,
+    Scenario,
 };
 use crate::mock::{get_path, restore_path};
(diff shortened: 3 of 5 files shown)


payload (commented):
There is also a `rustup set profile complete`. The warning would be printed on the next update but not on the `set` itself. Maybe also implement that set profile warns?

What do you think about small tests for this? I used to look into `cli-v2.rs` for that.

src/cli/rustup_mode.rs
```diff
@@ -805,6 +805,9 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<()> {
         cfg.set_profile_override(p);
     }
     let cfg = &cfg;
+    if cfg.get_profile()? == Profile::Complete {
+        warn!("downloading with complete profile isn't recommneded unless you are a developer of the rust language")
```
kinnison: You have a typo here:
```suggestion
        warn!("downloading with complete profile isn't recommended unless you are a developer of the rust language")
```

src/cli/setup_mode.rs
```diff
@@ -115,6 +115,10 @@ pub fn main() -> Result<()> {
         targets: &targets,
     };
 
+    if profile == "complete" {
+        warn!("downloading with complete profile isn't recommneded unless you are a developer of the rust language")
```
kinnison: This repeated typo suggests you might do well to have this as a constant string so there's no chance of mistakes in future.  You could put that usefully in `src/cli/common.rs` and then use it in both the locations.

kinnison (commented):
This is indeed the kind of change I was hoping for, so :heart: for that.

There's a typo in two places which suggests a shared static string would be best.

Also, as @payload suggested, it'd be useful to see a test for this.  So to verify the change to the toolchain installation step, we should check that warning is present in at least one test in `cli-v2.rs`.

One option is to add it to `test_complete_profile_skips_missing_when_forced()` though perhaps a small test just for this would be best.

In addition, at least one test in `cli-inst-interactive.rs` should check for the warning in the installation mode.  One option would be in `install_forces_and_skips_rls()` and another would be to create a new test just for this.

I look forward to seeing your updated PR, and thank you again.

tests/cli-inst-interactive.rs
```diff
@@ -254,3 +254,21 @@ fn install_forces_and_skips_rls() {
...
+                "toolchain",
+                "install",
+                "--profile",
+                "complete",
+                "stable",
+            ],
+            "warning: downloading with complete profile isn't recommended unless you are a developer of the rust language",
+        );
+    });
+}
```
kinnison: The `cli-inst-interactive.rs` test file is really for `rustup-init` testing.  This test belongs more in `cli-v2.rs` or similar where you already have it.  As such this should probably be to test `rustup-init --profile complete --default-toolchain stable -y` or somesuch.  (See other tests in this file)

kinnison (commented):
The test in `cli-v2.rs` looks like it ought to be okay, though you don't need the full error message, you could match against `warning: downloading with complete profile` -- that'd be enough.

tests/cli-inst-interactive.rs
Stupremee: I already changed that

kinnison (commented):
This looks excellent.  Could you please rebase and squash together the commits a little?  I'm happy with either one commit for everything, or one for the changes, and one for the tests.

If you can do that, I think we're very close to merging this.

PR #2150, opened 2019-12-10
Verify behaviour if host triplet goes missing from the channel manifests

kinnison:
Apple are removing support for i686 OSX.  As such, we need to be certain that rustup will behave gracefully should Rust remove tier 1 status from that target (or indeed any other in the future).

This PR tweaks the behaviour of the multi-arch tests and adds a scenario to model the above.

I'd appreciate @pietroalbini 's opinion on if it's sufficient for the use-case he mentioned to me.

I'd appreciate @jonhoo 's input on how to more nicely do the tweak to the channel generation which I'll admit is a bit of a hack in this PR currently.

### DIFF
--- a/tests/cli-misc.rs
+++ b/tests/cli-misc.rs
@@ -186,13 +186,10 @@ fn subcommand_required_for_self() {
 
 #[test]
 fn multi_host_smoke_test() {
-    // FIXME: Unfortunately the list of supported hosts is hard-coded,
-    // so we have to use the triple of a host we actually test on. That means
-    // that when we're testing on that host we can't test 'multi-host'.
-    let trip = this_host_triple();
-    if trip == clitools::MULTI_ARCH1 {
-        return;
-    }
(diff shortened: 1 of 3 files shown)


emilyalbini:
Yeah this seems good.

jonhoo:
I like the idea of having a builder that lets you construct a channel manifest with certain properties. Having named ones with particular features is maybe somewhat useful, but _probably_ detrimental as far as readability of tests goes.

kinnison:
@jonhoo I'm having trouble parsing your comment, perhaps because of tiredness, perhaps not.  Are you trying to say that the `HostGoesMissing` scenario variant is annoyingly test-specific?

jonhoo:
Ah, no, this wasn't a comment about this PR specifically, but more about the channel testing infrastructure in general :)

kinnison:
@jonhoo Okay thanks, I'm going to take that as tacit approval of this test then :D

PR #2152, opened 2019-12-10
Add message in case it's already up-to-date

GuillaumeGomez:


rbtcollins:
Hmm, can I ask why?
The current output is
```
$ rustup self update
```
And this combined with the exit code of 0 is pretty unambiguous. If an error occurs, its logged, if there is work to be done it is logged. I'm not against this per se, but not really for it either.

GuillaumeGomez:
Unambiguous but pretty weird. At first I wondered if it crashed somehow. At least this is now explicit, no need to check the exit code.

rbtcollins:
Well the thing is its also used in the large `rustup update` flow implicitly, and that looks like this:

```
 12.0 MiB /  12.0 MiB (100 %)   7.3 MiB/s in  1s ETA:  0s
info: installing component 'rust-std' for 'wasm32-unknown-unknown'

   stable-x86_64-unknown-linux-gnu updated - rustc 1.39.0 (4560ea788 2019-11-04)
  nightly-x86_64-unknown-linux-gnu updated - rustc 1.41.0-nightly (76a252ea9 2019-12-09)

```

And when there is nothing they show like so:

```
$ rustup update

   stable-x86_64-unknown-linux-gnu unchanged - rustc 1.39.0 (4560ea788 2019-11-04)
  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.41.0-nightly (76a252ea9 2019-12-09)

```

I don't want to get in the way here, but I wonder if what you really want is to have rustup self updates reported (or not) somehow more consistently with the toolchain information.

If you look at what we output for toolchains, just the one message is consistent, but we then do a summary at the bottom, which we don't do for rustup itself. Perhaps showing something like this would be good:

```
$ rustup update

   stable-x86_64-unknown-linux-gnu unchanged - rustc 1.39.0 (4560ea788 2019-11-04)
  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.41.0-nightly (76a252ea9 2019-12-09)
  rustup unchanged - rustup 1.20 (2019-10-21)
```

This would give more information

GuillaumeGomez:
Hum... I'm shared on this. I really like this output for example:

```text
$ rustup update
info: already up-to-date
```

Having more "state messages" would be nice. Something like:


```text
$ rustup update
info: done
info: done
info: already up-to-date
```

kinnison:
I quite like @rbtcollins's point that it'd be nice if the format mirrored how we report updates to toolchain channels.  This would require a little more plumbing but would be consistent which users may appreciate.

GuillaumeGomez:
I'll do it in the next days then.

kinnison:
@GuillaumeGomez Happy new year.  I wonder if you might have time in the coming week or so to make progress on this?

GuillaumeGomez:
I'll try to!

GuillaumeGomez:
As suggested, I re-used the code from common to "show".

GuillaumeGomez:
Updated!

src/errors.rs
```diff
@@ -362,6 +362,10 @@ error_chain! {
         BrokenPartialFile {
             description("partially downloaded file may have been damaged and was removed, please try again")
         }
+        Generic(v: String) {
+            description("generic error")
+            display("encountered error: {}", v)
+        }
```
kinnison: I think this change is a bogon.  I can't find use of this anywhere.

src/cli/self_update.rs
```diff
@@ -1435,21 +1436,24 @@ pub fn update() -> Result<()> {
...
-    if let Some(ref p) = setup_path {
-        let version = match get_new_rustup_version(p) {
-            Some(new_version) => parse_new_rustup_version(new_version),
-            None => {
-                err!("failed to get rustup version");
-                process::exit(1);
-            }
-        };
+    match prepare_update()? {
+        Some(ref setup_path) => {
```
kinnison: Does this need to remain `Some(ref setup_path)` now, or could it be `Some(setup_path)` ?

kinnison (changes requested):
One bogon, one query (not important if you choose to decline it), and then the commits need rebasing because we don't like `fmt` type commits, and also something has changed underneath you so I can't merge now even if I wanted to.

src/errors.rs
GuillaumeGomez: Seems so. Removing it!

src/cli/self_update.rs
GuillaumeGomez: To be confirmed. I check and update if needed.

GuillaumeGomez:
Updated!

kinnison (approved):
LGTM. :+1:

GuillaumeGomez:
\o/

PR #2158, opened 2019-12-14
use remove_dir_all is enough

tesuji:
cc @BeniCheni because it was your code, but feels free to ignore this ping


BeniCheni (approved):

kinnison (approved):
:+1:

PR #2159, opened 2019-12-14
make use std prelude and local imports

tesuji:
* use ? to instead of unwrap


kinnison (approved):
All these cleanups look good, thanks.

PR #2160, opened 2019-12-14
bump pgp crate to 0.4

tesuji:
This helps removing old and useless indirect dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,6 +101,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.11.0"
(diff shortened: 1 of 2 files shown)


kinnison:
Has anything changed which we rely on here?  Remember the PGP checking is entirely background and not exposed to the user yet.

tesuji:
Mostly no. Here's the change in pgp: https://github.com/rpgp/rpgp/compare/0.3.2...0.4.0
I updated PR description for the motivation of merging this PR.

kinnison:
Okay, that makes sense, thank you.

PR #2162, opened 2019-12-16
Demote deprecation

### LINKED ISSUE #2149: Demote deprecation warning to verbose info
After a lot of discussion with reddit and friends, it seems that deprecating `rustup install` and `rustup uninstall` would have consequences I'd not considered (printed books) and as such it'd be unpleasantly surprising for new users to be told the book they bought was wrong.

As such, we should demote the [deprecation warning code](https://github.com/rust-lang/rustup/blob/master/src/cli/rustup_mode.rs#L30) from `warn!()` to `verbose!()` and make it clear that the reason for the message is that until true alias support is added, these may not function quite as the `toolchain XXX` commands would.

Long-term we want to replace the deprecated subcommands with true aliases so that this can't be a problem again in the future.

kinnison:
This PR rearranges things a little to fix #2149 and tries to mitigate the ETXTBSY situation a little more.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -9,6 +9,7 @@ use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, Shell, SubCommand};
 use rustup::dist::dist::{PartialTargetTriple, PartialToolchainDesc, Profile, TargetTriple};
 use rustup::dist::manifest::Component;
 use rustup::utils::utils::{self, ExitCode};
+use rustup::Notification;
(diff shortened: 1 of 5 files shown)


kinnison:
I'd recommend reviewing this commit-by-commit because there's a bunch of cleanups in here too, to support the goal.

kinnison (commented):
Me from today thinks me from last night did an okay job.

TLATER (approved):
The `Notification::PlainVerboseMessage` workaround is a bit annoying, but after asking @kinnison on other  channels that seems like a necessary evil.

Looks good to me :)

PR #2163, opened 2019-12-17
tests: Fix up lack of for_host!() usage

kinnison:
This should correct a bug I was seeing on i686 platforms.


kinnison:
The full suite is running [here](https://github.com/kinnison/rustup/runs/352925850) and particularly you want to check that the i686 linux and mac builds pass.

PR #2173, opened 2019-12-24
Update links

JohnTitor:
- Replace https://forge.rust-lang.org/platform-support.html with https://forge.rust-lang.org/release/platform-support.html
- Replace https://github.com/rust-lang/rustup.rs with https://github.com/rust-lang/rustup

Also,

https://github.com/rust-lang/rustup/blob/7832b2ebee6468a9d88342181f8739c4c607d53f/Cargo.toml#L6

should be replaced but I'm not sure what the correct link is (https://rustup.rs?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,15 +141,15 @@ This release was made to solve two problems spotted in `1.20.0`
 - Mitchell Hynes
 - Matt Kantor
 
-[pr#1673]: https://github.com/rust-lang/rustup.rs/pull/1673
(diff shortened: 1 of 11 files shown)


kinnison:
We don't really have `rustdoc` published any more for `rustup` because it's not a library for others to use.  I'd suggest we either replace it with `https://rustup.rs/` or just drop the `documentation` entry alltogether.

JohnTitor:
I feel it's meaningless nowadays so I dropped it. (offtopic: I saw your message on Discord, no rush, take care of yourself!)

kinnison (approved):
All these look good to me, Thank you!

PR #2175, opened 2019-12-25
rename RUSTUP_FORCE_ARG0 to RUSTUP_NAME

tesuji:
I always find it hard to remember that environment variable name.
Hence the change.

cc @rbtcollins because you added that var.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ when rebuilding `rustup` itself.
 
 If you wish to install your new build to try out longer term in your home directory
 then you can run `cargo dev-install` which is an alias in `.cargo/config` which
-runs `cargo run -- --no-modify-path -y` to install your build into your homedir.
+runs `cargo run -- --no-modify-path -y` to install your build into your home dir.
 
 We use `rustfmt` to keep our codebase consistently formatted. Please ensure that
 you have correctly formatted your code (most editors will do this automatically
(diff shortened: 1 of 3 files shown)


kinnison:
This PR appears to have several changes conflated into it.  In addition, I'm not sure that changing `RUSTUP_FORCE_ARG0` to be more "friendly" is of value considering it's only meant to be for the use of developers.  I won't *block* the rename since all I'll need to do is change my shell functions (I don't type it in myself by hand each time) but I would want @rbtcollins to okay it, and I'd want the unrelated changes removed from this PR.

rbtcollins:
My recollection was that @kinnison wanted to avoid this being too easy for folk to access to avoid having it be considered part of the user interface. I don't object to the rename, though I too have no particular desire to rename it.

If we are going to rename it, I don't think _NAME is a particularly good *new* name though, since as far as environment variables go, 'NAME' is very context less and could refer to many things such as subcommand names.. We could drop the FORCE if thats what makes it hard to remember, or perhaps EXECNAME, or CMDNAME, or some such thing - but remember this isn't a subcommand name :).

tesuji:
My main motivation for it was to use rustup as rustup-init.
(Sometimes I copy my manually built rustup to another machine through scp).
I was too lazy to rename rustup to rustup-init (also to open Contributing.md).
I guess I should just rename it instead.
Thanks for your noticing.
Closing because of I don't think it worth enough.

PR #2176, opened 2019-12-26
Implement --allow-downgrade option for "component add" subcommand

### LINKED ISSUE #2146: Downgrade nightly when adding component

I have a nightly installed and want to add a component that is not available for this or any newer nightly.

With https://github.com/rust-lang/rustup/pull/2126 we have `rustup toolchain install --allow-downgrade`; this issue would be solved by adding the parallel `rustup component add --allow-downgrade`.

**Notes**

Currently available work-around: `rustup toolchain install $already_installed_channel --allow-downgrade -c added_component_name`

BeniCheni:
Implement `--allow-downgrade` option for `component add` sub-commend (e.g. `rustup component add rls --allow-downgrade`), to attempt to resolve #2146.

When `--allow-downgrade` is provide, it would run `toolchain.install_from_dist(...)`, to reuse the solution in #2126.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -5,6 +5,7 @@ use crate::self_update;
 use crate::term2;
 use crate::term2::Terminal;
 use crate::topical_doc;
+use chrono::prelude::*;
 use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, Shell, SubCommand};
 use rustup::dist::dist::{PartialTargetTriple, PartialToolchainDesc, Profile, TargetTriple};
 use rustup::dist::manifest::Component;
@@ -442,7 +443,13 @@ pub fn cli() -> App<'static, 'static> {
                                 .long("toolchain")
                                 .takes_value(true),
                         )
-                        .arg(Arg::with_name("target").long("target").takes_value(true)),
+                        .arg(Arg::with_name("target").long("target").takes_value(true))
+                        .arg(
+                            Arg::with_name("allow-downgrade")
+                                .help("Allow to downgrade the toolchain to satisfy your component choice")
+                                .long("allow-downgrade")
+                                .takes_value(false),
+                        ),
                 )
                 .subcommand(
                     SubCommand::with_name("remove")
@@ -1159,9 +1166,71 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
         let new_component = Component::new_with_target(component, false)
             .unwrap_or_else(|| Component::new(component.to_string(), target.clone(), true));
 
-        toolchain.add_component(new_component)?;
-    }
+        if m.is_present("allow-downgrade") {
+            let toolchain_date = toolchain
+                .get_manifest()
+                .ok()
+                .and_then(|m| m.map(|m| m.date))
+                .unwrap();
+            let downgrade_date = utc_from_manifest_date(&toolchain_date)
+                .unwrap_or_else(|| panic!("Malformed manifest date: {}", toolchain_date))
+                .pred();
 
+            let release_channel = match toolchain.name() {
+                s if s.starts_with("nightly") => "nightly",
+                s if s.starts_with("stable") => "stable",
+                s if s.starts_with("beta") => "beta",
+                _ => "nightly",
+            };
+            let nightly_date = format!(
+                "{}-{}",
+                &release_channel,
+                &downgrade_date.to_string()[0..10]
+            );
+            let temp_toolchain = cfg.get_toolchain(&nightly_date, false)?;
+
+            let status = if !temp_toolchain.is_custom() {
+                let components: Vec<_> = m
+                    .values_of("components")
+                    .map(|v| v.collect())
+                    .unwrap_or_else(Vec::new);
+                let targets: Vec<_> = m
+                    .values_of("targets")
+                    .map(|v| v.collect())
+                    .unwrap_or_else(Vec::new);
+                Some(temp_toolchain.install_from_dist(
+                    false,
+                    m.is_present("allow-downgrade"),
+                    &components,
+                    &targets,
+                )?)
+            } else if !temp_toolchain.exists() {
+                return Err(
+                    ErrorKind::InvalidToolchainName(temp_toolchain.name().to_string()).into(),
+                );
+            } else {
+                None
+            };
+
+            if let Some(status) = status.clone() {
+                println!();
+                common::show_channel_update(cfg, temp_toolchain.name(), Ok(status))?;
+            }
+
+            use rustup::UpdateStatus;
+            if let Some(UpdateStatus::Installed) = status {
+                temp_toolchain.make_default()?;
+                temp_toolchain.add_component(new_component)?;
+                toolchain.make_default()?;
+            } else if let Some(UpdateStatus::Unchanged) = status {
+                temp_toolchain.make_default()?;
+                temp_toolchain.add_component(new_component)?;
+                toolchain.make_default()?;
+            }
+        } else {
+            toolchain.add_component(new_component)?;
+        }
+    }
     Ok(())
 }
 
@@ -1471,4 +1540,10 @@ fn output_completion_script(shell: Shell, command: CompletionCommand) -> Result<
     }
 
     Ok(())
+}
+
+fn utc_from_manifest_date(date_str: &str) -> Option<Date<Utc>> {
+    NaiveDate::parse_from_str(date_str, "%Y-%m-%d")
+        .ok()
+        .map(|date| Utc.from_utc_date(&date))
 }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -446,7 +446,7 @@ pub fn cli() -> App<'static, 'static> {
                         .arg(Arg::with_name("target").long("target").takes_value(true))
                         .arg(
                             Arg::with_name("allow-downgrade")
-                                .help("Allow to downgrade the toolchain to satisfy your component choice")
+                                .help("Allow rustup to downgrade the toolchain to satisfy your component choice")
                                 .long("allow-downgrade")
                                 .takes_value(false),
                         ),
--- a/tests/cli-v2.rs
+++ b/tests/cli-v2.rs
@@ -1449,3 +1449,48 @@ fn install_allow_downgrade() {
         expect_component_executable(config, "rls");
     });
 }
+
+#[test]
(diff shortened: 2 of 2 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1159,7 +1166,51 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
         let new_component = Component::new_with_target(component, false)
             .unwrap_or_else(|| Component::new(component.to_string(), target.clone(), true));
 
-        toolchain.add_component(new_component)?;
+        if m.is_present("allow-downgrade") {
+            let old_date = toolchain
+                .get_manifest()
+                .ok()
+                .and_then(|m| m.map(|m| m.date));
+            let old_date = old_date.as_ref().map(|s| &**s);
```
tesuji:
```suggestion
            let old_date = old_date.as_deref();
```

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,75 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
             .unwrap_or_else(|| Component::new(component.to_string(), target.clone(), true));
 
-        toolchain.add_component(new_component)?;
-    }
+        if m.is_present("allow-downgrade") {
+            let toolchain_date = toolchain
+                .get_manifest()
+                .ok()
+                .and_then(|m| m.map(|m| m.date));
+            let toolchain_date = toolchain_date.as_deref();
```
tesuji:
```suggestion
```

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,75 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+            let downgrade_date = utc_from_manifest_date(toolchain_date.unwrap())
+                .unwrap_or_else(|| panic!("Malformed manifest date: {:?}", toolchain_date))
```
tesuji: then this should become
```suggestion
                .unwrap_or_else(|| panic!("Malformed manifest date: {}", toolchain_date))
```

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,75 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
 
-        toolchain.add_component(new_component)?;
-    }
+        if m.is_present("allow-downgrade") {
+            let toolchain_date = toolchain
+                .get_manifest()
+                .ok()
+                .and_then(|m| m.map(|m| m.date));
+            let toolchain_date = toolchain_date.as_deref();
+            let downgrade_date = utc_from_manifest_date(toolchain_date.unwrap())
```
tesuji:
```suggestion
            let downgrade_date = utc_from_manifest_date(&toolchain_date)
```

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,75 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
         let new_component = Component::new_with_target(component, false)
             .unwrap_or_else(|| Component::new(component.to_string(), target.clone(), true));
 
-        toolchain.add_component(new_component)?;
-    }
+        if m.is_present("allow-downgrade") {
+            let toolchain_date = toolchain
+                .get_manifest()
+                .ok()
+                .and_then(|m| m.map(|m| m.date));
```
tesuji: Why not unwrap here?
```suggestion
                .and_then(|m| m.map(|m| m.date))
                .unwrap();
```

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,75 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+            let toolchain_date = toolchain_date.as_deref();
+            let downgrade_date = utc_from_manifest_date(toolchain_date.unwrap())
+                .unwrap_or_else(|| panic!("Malformed manifest date: {:?}", toolchain_date))
+                .pred();
+
+            let release_channel = if toolchain.name().starts_with("nightly") {
```
tesuji: I like match better
```suggestion
            let release_channel = match toolchain.name() {
                s if s.starts_with("nightly") => "nightly",
```

kinnison:
Hi @BeniCheni 

Thank you for taking the time to implement this.  And thanks to @lzutao for some initial code review, that was super-nice of you to do given I've been taking a break :D

I shall look at this as soon as I can.

src/cli/rustup_mode.rs
```diff
@@ -442,7 +443,13 @@ pub fn cli() -> App<'static, 'static> {
                                 .long("toolchain")
                                 .takes_value(true),
                         )
-                        .arg(Arg::with_name("target").long("target").takes_value(true)),
+                        .arg(Arg::with_name("target").long("target").takes_value(true))
+                        .arg(
+                            Arg::with_name("allow-downgrade")
+                                .help("Allow to downgrade the toolchain to satisfy your component choice")
```
kinnison: Language nit:
```suggestion
                                .help("Allow rustup to downgrade the toolchain to satisfy your component choice")
```

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,71 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+            let downgrade_date = utc_from_manifest_date(&toolchain_date)
+                .unwrap_or_else(|| panic!("Malformed manifest date: {}", toolchain_date))
+                .pred();
+
+            let release_channel = match toolchain.name() {
+                s if s.starts_with("nightly") => "nightly",
+                s if s.starts_with("stable") => "stable",
+                s if s.starts_with("beta") => "beta",
+                _ => "nightly",
+            };
```
kinnison, line 1184: This kind of functionality feels like it belongs on the `Toolchain` type

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,71 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+            let nightly_date = format!(
+                "{}-{}",
+                &release_channel,
+                &downgrade_date.to_string()[0..10]
+            );
+            let temp_toolchain = cfg.get_toolchain(&nightly_date, false)?;
 
+            let status = if !temp_toolchain.is_custom() {
```
kinnison, line 1192: This test feels like it belongs at an early bailout rather than so deep into this function.

src/cli/rustup_mode.rs
```diff
@@ -1159,9 +1166,71 @@ fn component_add(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<()> {
...
+            use rustup::UpdateStatus;
+            if let Some(UpdateStatus::Installed) = status {
+                temp_toolchain.make_default()?;
+                temp_toolchain.add_component(new_component)?;
+                toolchain.make_default()?;
+            } else if let Some(UpdateStatus::Unchanged) = status {
+                temp_toolchain.make_default()?;
+                temp_toolchain.add_component(new_component)?;
+                toolchain.make_default()?;
+            }
```
kinnison, line 1229: Why are these pathways necessary?  If we passed through `install_from_dist` surely we will have added the listed components?  Also this odd changing of the default toolchain worries me, in case the inner `add_component()` calls fails, the default toolchain would have been unexpectedly altered to one which doesn't exist.  That feels wrong.

src/cli/rustup_mode.rs
```diff
@@ -1472,3 +1541,9 @@ fn output_completion_script(shell: Shell, command: CompletionCommand) -> Result<
 
     Ok(())
 }
+
+fn utc_from_manifest_date(date_str: &str) -> Option<Date<Utc>> {
+    NaiveDate::parse_from_str(date_str, "%Y-%m-%d")
+        .ok()
+        .map(|date| Utc.from_utc_date(&date))
+}
```
kinnison, line 1549: This function looks like it is a direct copy of code from elsewhere which might be better lifted into a utility function if it's not already in one.

tests/cli-v2.rs
```diff
@@ -1449,3 +1449,48 @@ fn install_allow_downgrade() {
...
+            config,
+            &["rustup", "component", "add", "rls", "--allow-downgrade"],
+        );
+        // rls was added to downgraded nightly-2019-09-13
+        // switch to the version and assert rls as executable
+        expect_stdout_ok(
+            config,
+            &["rustup", "default", "nightly-2019-09-13"],
+            "hash-nightly-2",
+        );
```
kinnison, line 1493: I don't understand where this new nightly toolchain has come from.  I would have expected with `--allow-downgrade` that the installed `nightly` toolchain was downgraded rather than that a brand new toolchain turned up.

kinnison (changes requested):
At the moment, there are a number of comments to be addressed.  Over-all I feel like the strange double-pathway through the `component add` cli handling ought to be unified into a single function which essentially behaves similarly to how `toolchain install` works -- if that means that we end up with the `Toolchain::add_component()` code becoming obsolete, then so be it.

I'd like answers to my questions, and once those are resolved, a treatment of my concern over the what the test suggests the behaviour currently is.  Once we've got through those, we can look to fixing up the code.

A very good start, but let's take some time to ensure we know exactly what behaviour we're expecting from this before we go further.

BeniCheni:
Thank you for your thorough questions @kinnison. I've reviewed and deemed that my implementation in this PR has a lot of confusing parts. I'd need to take a step back to grok what my thoughts were when I was writing this code when the analysis of other code to have a clear plan to edit and provide answers to your questions.

I'm going to look and think over thinks and come back with some sensible updates on this the rest of this week / weekend. Sorry about the delay. Will definitely ask questions on Discord if too stuck.

kinnison:
Take all the time you need @BeniCheni -- And ask all the questions you feel will help you on your way.  I look forward to your reworked branch.

kinnison:
Hey @BeniCheni 

How're things going?  If you're too busy / unable to get back to this, then please just let me know so I can look for someone else to carry your work forward.  I really want to see this functionality added to rustup.

Thanks,

Daniel.

BeniCheni:
Thanks for checking in @kinnison. Been a busy month. Unfortunately, I won’t be able to devote time in the next couple of weeks to follow up this PR. So I’m more than happy if someone wants to carry #2146 forward. I’ll respectfully close this PR. After the busy next two weeks, I’ll definitely come back to check out issues of rustup again.

PR #2182, opened 2020-01-01
mock: use fs::{write,read} and other small fixups

tesuji:
Also check for error in remove and copy functions.

### DIFF
--- a/tests/mock/clitools.rs
+++ b/tests/mock/clitools.rs
@@ -1114,8 +1114,8 @@ fn mock_bin(name: &str, version: &str, version_hash: &str) -> Vec<MockFile> {
             assert!(status.success());
             assert!(dest_path.exists());
 
-            // If we're on unix this will remove debuginfo, otherwise we just ignore
-            // the return result here
+            // Remove debug info from std/core which included in every programs,
+            // otherwise we just ignore the return result here
             if cfg!(unix) {
                 drop(Command::new("strip").arg(&dest_path).status());
             }
--- a/tests/mock/clitools.rs
+++ b/tests/mock/clitools.rs
@@ -448,7 +448,7 @@ where
                 retries -= 1;
                 if retries > 0
                     && e.kind() == std::io::ErrorKind::Other
-                    && format!("{}", e).contains("os error 26")
+                    && e.raw_os_error() == Some(26)
                 {
                     // This is a ETXTBSY situation
                     std::thread::sleep(std::time::Duration::from_millis(250));
--- a/tests/mock/clitools.rs
+++ b/tests/mock/clitools.rs
@@ -1154,8 +1154,14 @@ where
     A: AsRef<Path>,
     B: AsRef<Path>,
 {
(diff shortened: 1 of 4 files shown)


kinnison:
There are four changes here in one commit

1. Switch to `fs::{read,write}`
2. Pass `debuginfo=0` to rustc
3. Use `raw_os_error()` in `ETXTBUSY` check
4. Swallow `NotFound` errors when attempting to remove the target in `hard_link()` and `copy_file()`

Again, I believe I can follow these, but they would be better placed as separate commits.

I need to re-review to satisfy myself everything is sane, but I believe it must be or the CI would have gone mad.

tesuji:
I split the PR as multiple commits as requested. Sorry for that because I just cherry-picked
the big commit from a big branch that I worked on.

tests/mock/clitools.rs
```diff
@@ -1107,19 +1107,14 @@ fn mock_bin(name: &str, version: &str, version_hash: &str) -> Vec<MockFile> {
             let status = Command::new("rustc")
                 .arg(&source_path)
                 .arg("-C").arg("panic=abort")
+                .arg("-C").arg("debuginfo=0")
```
kinnison: One reason we stripped was because it used a lot of disk space during tests otherwise.  Is this sufficiently close to a stripped binary size so as to be equivalent in terms of disk space savings?  The test suite deletes things as it goes these days so it's less of an issue if not, but it could still present an issue for some if it's not close to as small.

tests/mock/clitools.rs
```diff
@@ -1149,21 +1149,27 @@ where
...
 where
     A: AsRef<Path>,
     B: AsRef<Path>,
 {
-    eprintln!(
-        "Copying from {} to {}",
-        a.as_ref().display(),
-        b.as_ref().display()
-    );
-    let _lock = cmd_lock().write().unwrap();
```
kinnison, line 1174: You have lost the protection of the `cmd_lock()` here which will increase the chances of race conditions hitting in the test suite on CI.

kinnison (changes requested):
Over-all pretty solid.  I'm a little unhappy that error-chain hasn't been updated to not need deprecated stuff but I suppose that's a case of minimum rust version stuff.  I would like the `cmd_lock()` thing fixing before I can accept this though.

tests/mock/clitools.rs
tesuji: I expect that the binary passing `debuginfo=0` is equivalent to the stripped binary,
but they are not.
I asked on discord. Hope the developers give use an explanation.

tesuji: The `debuginfo=0` is the default option when not specified.
We still need to strip the output binary because it contains
the debug table from std/core.

tests/mock/clitools.rs
tesuji: Done.

kinnison (approved):
LGTM now, thank you.

Issue #699, opened 2016-08-28
Reenable anti-sudo check


brson:
Disabled because it's being miscompiled. https://github.com/rust-lang-nursery/rustup.rs/pull/698

inejge:
Is there a known distro/nightly/rustup commit combination that produces a crashing `rustup-init` for 64-bit Linux? I tried with nightly-2016-08-21 and rustup pre-0.6.2 on CentOS 6, but couldn't get a segfault.

kinnison:
This looks to have been reenabled in https://github.com/rust-lang/rustup/commit/e308c567c0274ec93cd2876030c2b5316e6be2fe

PR #2183, opened 2020-01-01
Simplify sudo check on unix

tesuji:
This PR uses
* `cfg(unix)` attribute to limit sudo check on Unix platforms only.
* Move unsafe code interact with FFI to separate function: `home_dir_from_passwd`
  - Use `libc::sysconf` to get default entry size.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -470,61 +470,31 @@ fn do_pre_install_options_sanity_checks(opts: &InstallOpts) -> Result<()> {
 // sudo is configured not to change $HOME. Don't let that bogosity happen.
 #[cfg(unix)]
 fn do_anti_sudo_check(no_prompt: bool) -> Result<()> {
-    use std::ffi::OsString;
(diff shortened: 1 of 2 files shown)


kinnison:
There's a substantial set of changes in there for "Simplify" but I think I follow them.  This kind of rework, particularly when there is `unsafe` code is much more easily reviewed as a series of commits for future reference.

Could you please detail what testing you've done to ensure this is right, considering that the test code deliberately skips it in almost every case?

Also, how does this interact with work done in #1936 and does it mean that #699 is obsolete?

tesuji:
Yeah, I'll try to improve that in the future. 
About #699, I think e308c567c0274ec93cd2876030c2b5316e6be2fe enabled it again.
This PR doesn't interact with #1936 at all. More info in the rewrote PR description.

kinnison (approved):
Thank you for splitting things up and clarifying the relationships with those other items.

LGTM.

PR #2184, opened 2020-01-01
Make unsafe part smaller

tesuji:


tesuji:
Diff without whitespace: https://github.com/rust-lang/rustup/pull/2184/files?utf8=%E2%9C%93&diff=unified&w=1

kinnison:
This change appears to conflate at least three pieces of work

1. Migrate imports to top of function
2. Make `unsafe` part smaller
3. Better scope `snapshot` and `parent`

Again, I think I've managed to grok all three, but I'd very much prefer more commits to make it easier to review in future.  Because I'm not a windowsy person, I'm going to take a bit more time to satisfy myself that you've not accidentally changed any logic here, but I believe it's good.

src/cli/self_update.rs
```diff
@@ -953,21 +953,20 @@ fn delete_rustup_and_cargo_home() -> Result<()> {
     // of CARGO_HOME.
     let numbah: u32 = rand::random();
     let gc_exe = work_path.join(&format!("rustup-gc-{:x}.exe", numbah));
+    // Copy rustup (probably this process's exe) to the gc exe
+    utils::copy_file(&rustup_path, &gc_exe)?;
+    let gc_exe_win: Vec<_> = gc_exe.as_os_str().encode_wide().chain(Some(0)).collect();
+
+    // Open an inheritable handle to the gc exe marked
+    // FILE_FLAG_DELETE_ON_CLOSE. This will be inherited
+    // by subsequent processes.
```
kinnison: I'd suggest the comment needs tweaking here -- the thing it comments is the inheritance marker, but the file flag is actually later in the unsafe code.

src/cli/self_update.rs
```diff
@@ -953,21 +953,20 @@ fn delete_rustup_and_cargo_home() -> Result<()> {
...
-        let mut gc_exe_win: Vec<_> = gc_exe.as_os_str().encode_wide().collect();
-        gc_exe_win.push(0);
-
-        // Open an inheritable handle to the gc exe marked
-        // FILE_FLAG_DELETE_ON_CLOSE. This will be inherited
-        // by subsequent processes.
-        let mut sa = mem::zeroed::<SECURITY_ATTRIBUTES>();
-        sa.nLength = mem::size_of::<SECURITY_ATTRIBUTES>() as DWORD;
-        sa.bInheritHandle = 1;
-
```
kinnison: Here is where you need to leave the bit about the file flags

kinnison (changes requested):
Almost all of this looks fine, just some comment tweaks needed to clean up a moved comment.

src/cli/self_update.rs
tesuji: Done.

src/cli/self_update.rs
```diff
@@ -943,32 +953,20 @@ fn delete_rustup_and_cargo_home() -> Result<()> {
...
     let gc_exe = work_path.join(&format!("rustup-gc-{:x}.exe", numbah));
+    // Copy rustup (probably this process's exe) to the gc exe
+    utils::copy_file(&rustup_path, &gc_exe)?;
+    let gc_exe_win: Vec<_> = gc_exe.as_os_str().encode_wide().chain(Some(0)).collect();
+
+    let mut sa = SECURITY_ATTRIBUTES {
+        nLength: mem::size_of::<SECURITY_ATTRIBUTES> as DWORD,
+        lpSecurityDescriptor: ptr::null_mut(),
+        bInheritHandle: 1,
+    };
```
kinnison, line 936: You no longer have the comment explaining that this is to mean that the handle to the GC executable is inherited into the subprocess.  Are you claiming that it's obvious enough?

tesuji: Thanks, I missed it.

kinnison (approved):
OK, let's wait for CI then this can go in.  Thank you.

PR #2185, opened 2020-01-04
Use std::time primitives to replace time crate

### LINKED ISSUE #2181: Replace time crate with `std::time` primitives
`Rustup` currently uses both `time` v0.1 and `std::time` to represent a Duration. 

```console
% git grep '\btime::'
download/src/lib.rs:135:    use std::time::Duration;
download/src/lib.rs:256:    use std::time::Duration;
src/cli/download_tracker.rs:11:use time::precise_time_s;
src/cli/self_update.rs:924:    use std::time::Duration;
src/diskio/mod.rs:64:use time::precise_time_s;
src/diskio/threaded.rs:137:            sleep(std::time::Duration::from_millis(100));
src/toolchain.rs:23:use std::time::Duration;
tests/cli-self-upd.rs:287:    use std::time::Duration;
tests/mock/clitools.rs:454:                    std::thread::sleep(std::time::Duration::from_millis(250));
```

But with the evolve of `std::time`, it might be enough to only use `std::time` for those tasks.

The hard part is rewriting the `Duration` struct in:
https://github.com/rust-lang/rustup/blob/d6366cba52b00f517fa7f10f375cb91d40e02dc9/src/cli/download_tracker.rs#L223-L225
with custom Display implementation for `std::time::Duration`.

tesuji:
Closes #2181

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -7,8 +7,8 @@ use rustup::Notification;
 use std::collections::VecDeque;
 use std::fmt;
 use std::io::Write;
+use std::time::{Duration, Instant};
 use term::Terminal;
-use time::precise_time_s;
 
 /// Keep track of this many past download amounts
 const DOWNLOAD_TRACK_COUNT: usize = 5;
@@ -26,9 +26,9 @@ pub struct DownloadTracker {
     /// represent adjacent seconds; thus it may not show the average at all.
     downloaded_last_few_secs: VecDeque<usize>,
     /// Time stamp of the last second
-    last_sec: Option<f64>,
+    last_sec: Option<Instant>,
     /// Time stamp of the start of the download
-    start_sec: f64,
+    start_sec: Option<Instant>,
     /// The terminal we write the information to.
     /// XXX: Could be a term trait, but with #1818 on the horizon that
     ///      is a pointless change to make - better to let that transition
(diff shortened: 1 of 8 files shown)


src/cli/download_tracker.rs
```diff
@@ -111,13 +111,13 @@ impl DownloadTracker {
         self.total_downloaded += len;
         self.downloaded_this_sec += len;
 
-        let current_time = precise_time_s();
+        let current_time = Instant::now();
 
         match self.last_sec {
             None => self.last_sec = Some(current_time),
```
lzutao: This last_sec seems to do nothing except to save the last current time.
Should we remove it?

src/diskio/mod.rs
```diff
@@ -160,29 +160,30 @@ pub fn write_file<P: AsRef<Path>, C: AsRef<[u8]>>(
...
+            f.write_all(contents)?;
+        }
+        {
+            trace_scoped!("close", "name:": path_display);
+            drop(f);
+        }
+        Ok(())
     }
-    Ok(())
+    inner(path.as_ref(), contents.as_ref(), mode)
```
kinnison: This rework seems unrelated to the commit it's in, and not really related to this PR at all.  is it left over from other work you were doing?

kinnison (commented):
There's one change bogon, and I'd prefer to see the Cargo.lock change in a separate commit so I can more easily confirm it on the CLI, though it does look okay for now.  If you can please confirm if the change I hilighted needs to be in this series, I'd appreciate it.

src/diskio/mod.rs
tesuji: Yeah. It is unrelated. I will split it to a separate PR.

tesuji: Done.

kinnison (approved):
LGTM, thank you

PR #2190, opened 2020-01-07
GitHub actions

kinnison:
This PR introduces github actions for everything and removes the travis and appveyor configurations.

It needs careful checking to ensure everything is right.


.github/workflows/centos-fmt-clippy-on-all.yaml
```diff
@@ -0,0 +1,93 @@
...
+    branches:
+      - master
+      - stable
+  schedule:
+    - cron: "30 0 * * 1" # Every Monday at half past midnight
+
+jobs:
+  check:
+    name: Checks
+    runs-on: ubuntu-latest
```
rbtcollins, line 20: The name and file name claims to check on centos. This metadata says it checks on ubuntu. I think that this is confusing.

rbtcollins: Or if it does check on centos a comment would help

kinnison: The centos check is done in a docker in one of the actions.  fmt, clippy, and shellcheck are not done on centos.

kinnison: If you want to patch the texts to make things nicer, please file a PR, this was the culmination of two days of frustration, so it's not the prettiest it could be yet :D

rbtcollins: I'm not entirely sure I understand it all yet. What is a 'centos check' vs say a shellcheck run on centos?

kinnison: the centos check is the docker run where we validate that `rustup-init.sh` still works on centos 6 which is the oldest platform (oldest curl etc) which we support.

PR #2192, opened 2020-01-08
Unwrap Option<Path> early and use Cow::into_owned

tesuji:


kinnison:
This commit conflates two changes -- one as described (unwrapping the `Option<Path>`) and one which replaces uses of `.to_string_lossy()` with `.display().to_string()`.

tesuji:
Okay. I dropped the `.display().to_string()` changes.

kinnison:
If you could tweak the commit message of the second commit, and update the PR title, then I think this could be merged.  (You have `Cown` where you meant `Cow`, and with the two commits, the PR title is now incorrect)

kinnison (approved):
:+1:

PR #2197, opened 2020-01-09
Add component name when target is not equal to host

### LINKED ISSUE #2191: Component missing message should include target if not equal to toolchain host
As per https://github.com/rust-lang/rust/issues/67724#issuecomment-571855495 it'd be good if the component-missing message included the target of the component *iff* the target is not equal to the host for the toolchain.

i.e. `rust-std` for the toolchain's host, but `rust-std-foo-unknown-bar` for a cross-std.

LeSeulArtichaut:
Closes #2191 

I haven't been able to test this at all, but I still wanted to create this PR so someone could test for me before I find time to do it myself.

### DIFF
--- a/src/dist/notifications.rs
+++ b/src/dist/notifications.rs
@@ -1,5 +1,5 @@
 use crate::config::PgpPublicKey;
-use crate::dist::dist::TargetTriple;
+use crate::dist::dist::{TargetTriple, ToolchainDesc};
 use crate::dist::manifest::Component;
 use crate::dist::temp;
 use crate::errors::*;
@@ -31,7 +31,7 @@ pub enum Notification<'a> {
     DownloadingManifest(&'a str),
     DownloadedManifest(&'a str, Option<&'a str>),
     DownloadingLegacyManifest,
-    SkippingNightlyMissingComponent(&'a [Component]),
+    SkippingNightlyMissingComponent(&'a ToolchainDesc, &'a [Component]),
     ForcingUnavailableComponent(&'a str),
     ManifestChecksumFailedHack,
     ComponentUnavailable(&'a str, Option<&'a TargetTriple>),
@@ -72,7 +72,7 @@ impl<'a> Notification<'a> {
             | ManifestChecksumFailedHack
             | RollingBack
             | DownloadingManifest(_)
-            | SkippingNightlyMissingComponent(_)
+            | SkippingNightlyMissingComponent(_, _)
             | RetryingDownload(_)
             | DownloadedManifest(_, _) => NotificationLevel::Info,
             CantReadUpdateHash(_)
@@ -174,13 +174,19 @@ impl<'a> Display for Notification<'a> {
                 "removing stray hash found at '{}' in order to continue",
                 path.display()
             ),
-            SkippingNightlyMissingComponent(components) => write!(
+            SkippingNightlyMissingComponent(toolchain, components) => write!(
                 f,
                 "skipping nightly which is missing installed component{} '{}'",
                 if components.len() > 1 { "s" } else { "" },
                 components
                     .iter()
-                    .map(|component| component.short_name_in_manifest().to_owned())
+                    .map(|component| {
+                        if component.target.as_ref() != Some(&toolchain.target) {
+                            component.name_in_manifest().to_owned()
+                        } else {
+                            component.short_name_in_manifest().to_owned()
+                        }
+                    })
                     .collect::<Vec<_>>()
                     .join("', '")
             ),
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -689,7 +689,7 @@ fn update_from_dist_<'a>(
 
                 if let ErrorKind::RequestedComponentsUnavailable(components, ..) = e.kind() {
                     (download.notify_handler)(Notification::SkippingNightlyMissingComponent(
-                        components,
+                        &toolchain, components,
                     ));
 
                     if first_err.is_none() {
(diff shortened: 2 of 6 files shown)


kinnison:
Hi @LeSeulArtichaut 

Thanks for your contribution.  I'll look at the change soon, but without a test in place I'm not sure I can merge it.  Also, could you please rebase your work to squash the two commits together -- having a commit which is just about formatting makes it harder to reason about the history of a line of code.

LeSeulArtichaut:
Got it. I will work on this as soon as I can

kinnison:
Hi @LeSeulArtichaut 

Are you able to write some tests for this?  If not, just say and I'll look for someone else to complete the work.  If you are, however, then I'd really like to be able to merge this functionality soon.

Thanks,

Daniel.

LeSeulArtichaut:
I’m sorry, I haven’t made any progress. I’d still like to complete the work if it is ok to wait a little longer. Are there any guides on creating tests for Rustup?

kinnison:
It's alright, we can wait at least a little more :-D.  Unfortunately there aren't guides for writing tests for Rustup however I'd expect that looking at `tests/cli-v2.rs` will probably give you ideas.  I generally find a test which is similar to the one I need to write, and then work from there, either copy/paste it and change as needed, or write from scratch with the example as a template.

LeSeulArtichaut:
Alright, thanks a lot! I will try to work on this despite the few minor exams I will have to take next week.

kinnison:
Your exams should come first.  Please :D

LeSeulArtichaut:
@kinnison There you go. I hope this is enough! Please tell me if changes or refractoring are needed

src/dist/notifications.rs
```diff
@@ -174,11 +174,20 @@ impl<'a> Display for Notification<'a> {
                 "removing stray hash found at '{}' in order to continue",
                 path.display()
             ),
-            SkippingNightlyMissingComponent(components) => write!(
-                f,
-                "skipping nightly which is missing installed component '{}'",
-                components[0].short_name_in_manifest()
-            ),
+            SkippingNightlyMissingComponent(components) => {
+                let long_name = if components[0].target != TargetTriple::from_host() {
```
LeSeulArtichaut: Maybe add a comment here to explain?

kinnison:
I shall go over things tonight, thank you.

src/dist/notifications.rs
kinnison: Thinking about it, this is slightly confusing if only because on multiarch systems, `TargetTriple::from_host()` is not necessarily the same as the triple of the toolchain.  We might need to augment this particular notification with the toolchain's host triple.

kinnison (dismissed):
I think the only remaining concern is that of the multiarch scenario where, for example, I might have a  linux host, but I have a windows toolchain installed (and using wine I can use it to build stuff "natively") -- under that circumstance, we'd need the default triple to be the windows triple not the linux one.

The point where a `SkippingNightly` notification is sent has a `ToolchainDesc` and so ought to be able to pass the toolchain's triple through.  Do you think you might be able to do that, or would you prefer me to push a commit to your branch to show you what I mean, for you to integrate?

src/dist/notifications.rs
LeSeulArtichaut: So, you want to take the (user-defined) default toolchain triple as the base, instead of the current host? Or did I misunderstood you?

kinnison: The point at which you create this notification has access to a `Toolchain` instance which knows the host triple of the toolchain.  It'd be best to pass this into the notification so that it can format based on that, rather than on the host's triple.

LeSeulArtichaut: @kinnison Is 025cf35 what you meant?

LeSeulArtichaut:
@kinnison I'd need a bit of help here ^^

kinnison:
Your change is behaving as I'd expect.  The goal here is to explain when something which has a target *different* from the toolchain's target. To test this you need a configuration where `rust-std` is available for multiarch in one release, you add the other arch's rust-std, and then in the next release that arch's rust-std is no longer available.

In total I'd expect an error sequence like:

```
info: syncing channel updates for 'nightly-i686-unknown-linux-gnu'
info: latest update on 2019-09-14, rust version 1.37.0 (hash-nightly-3)
info: skipping nightly which is missing installed component 'rust-std-aarch64-unknown-linux-gnu'
info: syncing channel updates for 'nightly-2019-09-13-i686-unknown-linux-gnu'
```

See how the toolchain's target is i686 but the rust-std target is aarch64.

Obviously this will be within the constraints of the test set, but I'm sure you'll get there.  You're very very close.

kinnison:
I can see why this caused @LeSeulArtichaut so much difficulty during testing.  I've reworked a lot of the test mock channel construction code and cleaned some bits up, rebased this change on top, and fixed up the test to be more thorough.  It should be good now I think.

kinnison (approved):
I'm now happy with this (having taken a break and re-read it)

LeSeulArtichaut:
@kinnison Thanks a lot for doing the work! Is there something I need to do for this to be merged?

kinnison:
@LeSeulArtichaut I just need your confirmation that it looks as you expected, given it has your name on some of it.  Then I'll merge.

LeSeulArtichaut:
Well it looks good to me 👍

kinnison:
@LeSeulArtichaut Thanks.  I'll leave it a day or two for @rbtcollins to have a chance to skim it too, just in case :D

PR #2198, opened 2020-01-09
Few improvement for Windows string manipulation

tesuji:
* De-duplicate `string_to_winreg_bytes` function
* Use `encode_wide().chain` when possible

### DIFF
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -506,8 +506,9 @@ pub fn format_path_for_display(path: &str) -> String {
 /// Encodes a utf-8 string as a null-terminated UCS-2 string in bytes
 #[cfg(windows)]
 pub fn string_to_winreg_bytes(s: &str) -> Vec<u8> {
+    use std::ffi::OsStr;
     use std::os::windows::ffi::OsStrExt;
-    let v: Vec<_> = OsString::from(format!("{}\x00", s)).encode_wide().collect();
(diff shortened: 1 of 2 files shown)


kinnison:
As far as I can tell, no pathway this change touches is in any sense performance sensitive.  What made you think this was necessary / worthwhile?  What have I missed?

src/utils/raw.rs
```diff
@@ -352,7 +352,7 @@ pub mod windows {
     pub fn to_u16s<S: AsRef<OsStr>>(s: S) -> io::Result<Vec<u16>> {
         fn inner(s: &OsStr) -> io::Result<Vec<u16>> {
             let mut maybe_result: Vec<u16> = s.encode_wide().collect();
-            if maybe_result.iter().any(|&u| u == 0) {
+            if memchr_u16(0, &maybe_result).is_some() {
```
rbtcollins: I don't think we need this bit - adding in a FIXME  (in the other hunk) in our code base for an unmeasured performance change in a non-inner-loop(*) part of the code  doesn't make sense to me. The rest of the patch looks great.

*: this is only called from symlink_dir, which I don't have exact figures on, but I am reasonably sure its at most O(dists). so maybe 10 times a process run? That 11% improvement is going to be maybe 0.0001 of runtime. If that.

tesuji: That's fair. I removed that commit.

kinnison:
These cleanups now LGTM

PR #2199, opened 2020-01-09
build: Remove rust-docs before updating toolchain

tesuji:


kinnison:
Hmm, if `bash` is available perhaps this is a chance to unify handling of the toolchains across all our platforms.  I like the way it's done in the `linux` template:

```yaml
      - name: Ensure Stable is up to date
        run: |
          if rustc +stable -vV >/dev/null 2>/dev/null; then
            rustup toolchain uninstall stable
          fi
          rustup toolchain install --profile=minimal stable
      - name: Ensure we have our goal target installed
        run: |
          rustup target install "$TARGET"
```

kinnison:
If you agree, please can you update the macos workflow to match too, and retitle the PR accordingly?

tesuji:
The script above also removes the toolchain in case it is up-to-date.

kinnison:
Yes it does - deliberately so - because it ensures that we don't have unexpected components installed.

kinnison:
I've updated macos myself and I'm okay for this change to be merged for now.

Issue #2188, opened 2020-01-06
Can't use rustup-init arguments to set up cross-compiling with mingw


flaviojs:
**Problem**

Can't use rustup-init arguments to set up cross-compiling with mingw in the github runner ubuntu-latest (ubuntu-18.04).

During build I got the unexpected error:
```
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-pc-windows-gnu` target may not be installed
```

I found that rustup-init didn't set up things as expected and forced me to manually add the x86_64-pc-windows-gnu target with:
```
rustup target add x86_64-pc-windows-gnu
```

I was expecting the target/toolchain to be available after init.

**Notes**
First try (bad):
```
$ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain=1.36.0-x86_64-pc-windows-gnu
$ export PATH=$PATH:$HOME/.cargo/bin
$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/runner/.rustup

info: syncing channel updates for '1.36.0-x86_64-unknown-linux-gnu'
info: latest update on 2019-07-04, rust version 1.36.0 (a53f9df32 2019-07-03)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
1.36.0-x86_64-unknown-linux-gnu (overridden by '/home/runner/work/ja2-stracciatella/ja2-stracciatella/rust-toolchain')
rustc 1.36.0 (a53f9df32 2019-07-03)
```

Second try (bad):
```
$curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain=1.36.0 --target=x86_64-pc-windows-gnu
$ export PATH=$PATH:$HOME/.cargo/bin
$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/runner/.rustup

info: syncing channel updates for '1.36.0-x86_64-unknown-linux-gnu'
[...]

PR #2201, opened 2020-01-09
rustup-init: Support updating the installed toolchain

### LINKED ISSUE #2194: Warn when a rustup-init argument is ignored/changed

This is a followup to #2188.
Since `rustup-init` shouldn't be used to set up cross-compiling with mingw, at the very least it should warn when an argument is ignored/changed.

Without a warning I would think that everything is setup as expected.
This leads to wasting a lot of time trying to figure out what is wrong when a problem arises.

I'm not sure what the stated reasons should be but I would like a warning with enough information.

Examples:
 * for --default-toolchain=1.36.0-x86_64-pc-windows-gnu
    ```
    default toolchain changed to 1.36.0-x86_64-unknown-linux-gnu: reason
    ```
 * for --default-toolchain=1.36.0 --target=x86_64-pc-windows-gnu
    ```
    target x86_64-pc-windows-gnu was ignored: reason
    ```

Maybe something else would be appropriate but this is all I can think of without knowledge of the internals.

### LINKED ISSUE #2215: rustup-init skip installing additional target if default toolchain exists
**Problem**

`rustup-init` skips adding additional toolchain via `--target` option when default-toolchain exists.

*Currently*: rustup-init updates itself and stop
*Expected*: rustup should add the `target` if it doesn't exist.
*Context*: I run a script on CI that makes use of `rustup-init` despite of rustup exist or not.

**Steps**

1. Install stable toolchain
```console
rustup install stable
```
2. Try using `rustup-init`
```console
$ rustup-init.exe --default-host=x86_64-pc-windows-msvc --default-toolchain=stable --target=i686-pc-windows-msvc
info: profile set to 'minimal'
info: setting default host triple to x86_64-pc-windows-msvc
info: updating existing rustup installation


Rust is installed now. Great!

To get started you need Cargo's bin directory (%USERPROFILE%\.cargo\bin) in your
PATH
$ cargo build --target i686-pc-windows-msvc
    Updating crates.io index
   Compiling winapi v0.3.8
error[E0463]: can't find crate for `core`
  |
  = note: the `i686-pc-windows-msvc` target may not be installed

For more information about this error, try `rustc --explain E0463`.
error: could not compile `winapi`.

```

**Possible Solution(s)**

**Notes**

Output of `rustup --version`: rustup 1.21.1 (7832b2ebe 2019-12-20)
Output of `rustup show`:

kinnison:
If the user specifies any of --default-toolchain (not none)
or --target or --component when invoking `rustup-init` then
we will update the existing installation toolchains, rather
than the previous behaviour which was to skip toolchain
installation.  This is nominally a breaking change but likely
ends up adhering closer to the principle of least surprise.

At this point, the profile will not be applied, so a change
of profile from minimal to default will not cause additional
components to end up being installed.

In theory this fixes #2194 but will need checking with @flaviojs
Also @lzutao filed another bug, this closes #2215 too.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -758,6 +758,20 @@ fn maybe_install_rust(
     // install, so we leave their setup alone.
     if toolchain == Some("none") {
         info!("skipping toolchain installation");
+        if !components.is_empty() {
+            warn!(
+                "ignoring requested component{}: {}",
+                if components.len() == 1 { "" } else { "s" },
+                components.join(", ")
+            );
+        }
+        if !targets.is_empty() {
+            warn!(
+                "ignoring requested target{}: {}",
+                if targets.len() == 1 { "" } else { "s" },
+                targets.join(", ")
+            );
+        }
         println!();
     } else if user_specified_something || cfg.find_default()?.is_none() {
         let toolchain_str = toolchain.unwrap_or("stable");
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -47,7 +47,7 @@ use std::process::{self, Command};
 
 pub struct InstallOpts<'a> {
     pub default_host_triple: Option<String>,
-    pub default_toolchain: String,
+    pub default_toolchain: Option<String>,
     pub profile: String,
     pub no_modify_path: bool,
     pub components: &'a [&'a str],
@@ -283,9 +283,9 @@ pub fn install(no_prompt: bool, verbose: bool, quiet: bool, mut opts: InstallOpt
         }
         utils::create_rustup_home()?;
         maybe_install_rust(
-            &opts.default_toolchain,
+            opts.default_toolchain.as_deref(),
             &opts.profile,
-            opts.default_host_triple.as_ref(),
+            opts.default_host_triple.as_deref(),
             opts.components,
             opts.targets,
             verbose,
@@ -440,10 +440,10 @@ fn do_pre_install_options_sanity_checks(opts: &InstallOpts) -> Result<()> {
             .as_ref()
             .map(|s| dist::TargetTriple::new(s))
             .unwrap_or_else(TargetTriple::from_host_or_build);
-        let toolchain_to_use = if opts.default_toolchain == "none" {
-            "stable"
-        } else {
-            &opts.default_toolchain
+        let toolchain_to_use = match &opts.default_toolchain {
+            None => "stable",
+            Some(s) if s == "none" => "stable",
+            Some(s) => &s,
         };
         let partial_channel = dist::PartialToolchainDesc::from_str(toolchain_to_use)?;
         let resolved = partial_channel.resolve(&host_triple)?.to_string();
@@ -590,7 +590,9 @@ fn current_install_opts(opts: &InstallOpts) -> String {
             .as_ref()
             .map(|s| TargetTriple::new(s))
             .unwrap_or_else(TargetTriple::from_host_or_build),
-        opts.default_toolchain,
+        opts.default_toolchain
+            .as_deref()
+            .unwrap_or("stable (default)"),
         opts.profile,
         if !opts.no_modify_path { "yes" } else { "no" }
     )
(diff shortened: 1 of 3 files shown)


tesuji:
I wonder whether we could integrate parsing toolchains string to a type.
And use that type instead of processing raw string over and over again
in whole crate.

kinnison:
@lzutao We do have a ToolchainDesc in `dist.rs` and I suppose we could begin to rework APIs to expect one of those but it pushes the handling of bad toolchain names outward which might mean we end up with more spots needing to deal with that specific error rather than more general handling in the core.

rbtcollins:
hmm, so before I even check the code I will note that this will likely result in performance regression bugs on windows until we liaise with Microsoft's Defender team again to get rustup-init included in their scan-mode heuristics as well (I don't believe it currently is - I may be wrong).

rbtcollins:
Hmm, rustup-init already knows how to install toolchains, we must have had it included in the heuristics already. Nevermind - or at least, lets treat that as a separate issue. Looking at the code now.

flaviojs:
Installing/updating according to the arguments is what I originally expected, so this fixes #2188 (my initial issue).

From the changes it seems I'll be able to know if the toolchain was installed or not from the messages, so this fixes #2194.

src/cli/self_update.rs
```diff
@@ -746,21 +748,29 @@ fn maybe_install_rust(
...
-    // possible to select a toolchain then have it not be installed.
-    if toolchain_str == "none" {
+    let user_specified_something =
+        toolchain.is_some() || !targets.is_empty() || !components.is_empty();
+
+    // If the user specified they want no toolchain, we skip this, otherwise
+    // if they specify something directly, or we have no default, then we install
+    // a toolchain (updating if it's already present) and then if neither of
+    // those are true, we have a user who doesn't mind, and already has an
+    // install, so we leave their setup alone.
```
rbtcollins, line 758: I don't think this is quite right. What if they ask for no tool chain and a component? Should that silently do nothing or error?

rbtcollins: Well, it won't be silent, but it won't process the invalid input either.

rbtcollins: other than this, the PR looks good

kinnison: I think that's a good point.  I'd say it's part of `setup_mode.rs` to verify that though.  I've made a note to try and write a test into there.

kinnison: I take that back, clearly it belongs after the user might have made a change.  If they specify targets/components on the CLI and then set the toolchain to none in the interactive prompts, then it still ought to warn.

kinnison: I've pushed a small change, how does that feel to you?

kinnison:
@rbtcollins Are you okay with the current state of things?  I'd like to get this merged if so.

src/cli/self_update.rs
```diff
@@ -746,21 +748,43 @@ fn maybe_install_rust(
...
+                targets.join(", ")
+            );
+        }
         println!();
-    } else if cfg.find_default()?.is_none() {
+    } else if user_specified_something || cfg.find_default()?.is_none() {
+        let toolchain_str = toolchain.unwrap_or("stable");
         let toolchain = cfg.get_toolchain(toolchain_str, false)?;
+        if toolchain.exists() {
+            warn!("Updating existing toolchain, profile choice will be ignored");
```
rbtcollins, line 780: Is this missing logic, or not well defined?

kinnison: At the moment it's not clear what it'd mean - should it only *add* missing components?  Should it make it *be* only the components for the profile (e.g. if you say `minimal` should it remove anything else?) - as such, this is punted just like it is if you do `rustup toolchain install ...` after it's already installed, the profile will be ignored.

flaviojs: `rustup set` says "The default components installed" about profile.
`rustup component` says we can add and remove components

Based on that I would assume that the profile only determines what is added on install. If the toolchain already exists I would expect it to add missing components and leave the rest alone.

Since the rust-init option is called `--profile` (no default) I would also assume it only affects that particular call. No idea if that is the current behavior.

kinnison: So your proposal would be "If the user explicitly invoked `rustup toolchain install` (or `rustup-init` as is considered in this PR) then any components missing from the profile of an already installed toolchain should explicitly be added in.  I *think* that could be consistent but wouldn't be behaviour belonging in *this* PR.  I'd be up for an issue about that being opened though, if either of you agree with my restating of things.

rbtcollins: I'm happy with 'this is not well defined yet, lets open a new issue to discuss'.

That proposed behaviour might be it, but another one might be to coerce back to the profiles exact definition - lets work through it elsewhere.

PR #2204, opened 2020-01-17
Dropped unnecessary curl option

marcobellaccini:
The curl *proto* option is not necessary (since target URL is https), hence it should be dropped.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -25,7 +25,7 @@
 
 <div id="platform-instructions-unix" class="instructions display-none">
   <p>Run the following in your terminal, then follow the onscreen instructions.</p>
-  <pre>curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+  <pre>curl --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
   <p class="other-platforms-help">You appear to be running Unix. If not, <a class="default-platform-button" href="#">display all supported installers</a>.</p>
 </div>
 
@@ -36,7 +36,7 @@
     then follow the onscreen instructions.
   </p>
   <p>If you're a Windows Subsystem for Linux user run the following in your terminal, then follow the onscreen instructions to install Rust.</p>
-  <pre>curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+  <pre>curl --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
   <p class="other-platforms-help">You appear to be running Windows 32-bit. If not, <a class="default-platform-button" href="#">display all supported installers</a>.</p>
 </div>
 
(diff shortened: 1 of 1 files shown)


kinnison:
the `--proto =https` argument means that a MITM cannot force the protocol to downgrade to `http` so it isn't unnecessary as I see it.  Why do you think it's unnecessary?

marcobellaccini:
What kind of HTTPS->HTTP downgrade attacks are you afraid of?
Can you provide a practical example?
curl (and https clients in general) won't accept http-downgrade that easily.

Also, that `proto` option isn't protecting you from TLS/SSL downgrade attacks like POODLE (but that `--tlsv1.2` option will do the job).

Moreover, if your threat model requires a very high degree of security, I think you should consider many other issues first (e.g.: neither [rustup.rs](https://rustup.rs/) nor [rust-lang.org](https://www.rust-lang.org/) seem to be included in [HSTS preload lists](https://hstspreload.org), they also don't support TLS 1.3).

kinnison:
In this instance I'm more interested in belt and braces.  The particular combination of arguments was recommended to us as part of https://github.com/rust-lang/rustup/pull/1716 and I'm reluctant to remove arguments without the author of that change (@sanmai-NL) confirming how you believe cURL works wrt. redirects.

marcobellaccini:
Ok, I understand your point.
However please consider that, in order to perform a redirect, the TLS handshake will have to complete successfully (hence you should be protected).
I read in [https://github.com/rust-lang/rustup/pull/1716](https://github.com/rust-lang/rustup/pull/1716) that @sanmai-NL was concerned about future unsafe redirects to HTTP.
That's right, but it's somewhat fancy (you're trying to partially protect yourselves from some unlikely, future, critical security issue).
And it's quite uncommon too (see what [Node.js](https://github.com/nodesource/distributions/blob/master/README.md), [Docker](https://docs.docker.com/install/linux/docker-ce/ubuntu/), [NGINX](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#installing-a-prebuilt-alpine-linux-package-from-the-official-nginx-repository) are doing for packages and keys when using _curl_).

www/index.html
```diff
@@ -25,7 +25,7 @@
 
 <div id="platform-instructions-unix" class="instructions display-none">
   <p>Run the following in your terminal, then follow the onscreen instructions.</p>
-  <pre>curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+  <pre>curl --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
```
tesuji, line 28: What about removing redundant https scheme: https://curl.haxx.se/docs/manpage.html#--proto
```suggestion
  <pre>curl --tlsv1.2 --proto '=https' -sSf sh.rustup.rs | sh</pre>
```

marcobellaccini: This is not working:
`$ curl --tlsv1.2 --proto '=https' -sSf sh.rustup.rs`
`curl: (1) Protocol "http" not supported or disabled in libcurl`

tesuji: Oh, that's right.

sanmai-NL:
@marcobellaccini: why should we remove security measures when other security measures haven't been implemented yet? Do you have a risk assessment that supports your idea such an attack is highly unlikely?

Please help by implementing the missing security measures.

marcobellaccini:
I have made my comments.
If you disagree, you can just keep that option.
Nothing personal when it comes to software.
Just for the record: I can't add the website to the HSTS preload list, nor I can enable TLS 1.3 - because those actions would require administrative access to the webservers.
I'm closing this PR.

sanmai-NL:
I also don't have administrative access for rustup. You may raise the concerns though. TLS 1.3 is limited by Amazon I'm afraid.

PR #2205, opened 2020-01-18
Update `retry` to 1.0

JohnTitor:
We can reduce dependencies size.

Changelog: https://github.com/jimmycuadra/retry/compare/0.5.1...1.0.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ xz2 = "0.1.3"
 pgp = { version = "0.4", default-features = false}
 
 [dependencies.retry]
-version = "0.5"
+version = "1"
 default-features = false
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -55,7 +55,7 @@ xz2 = "0.1.3"
 pgp = { version = "0.4", default-features = false}
 
 [dependencies.retry]
-version = "0.5"
+version = "1"
 default-features = false
```
JohnTitor, line 59: btw, it there the reason to separate `retry` section?

kinnison: Only because of aesthetics of the author who picked `retry` to use.

kinnison (approved):
LGTM.

PR #2208, opened 2020-01-21
www: show whole shell command at once

### LINKED ISSUE #2207: Website: show whole shell command at once
Since the TLS related options were added, the main https://rustup.rs/ page shows only a fraction of the shell command (see below).

Even copying the command is getting cumbersome with this, and it would be nice to see the whole thing I'm supposed to run...

image

harrysarson:
Makes the width of the shell box slightly larger and the font size slightly smaler.

Fixes: https://github.com/rust-lang/rustup/issues/2207

Before: 
image

After:
image

---
cc:  @birkenfeld

### DIFF
--- a/www/rustup.css
+++ b/www/rustup.css
@@ -63,11 +63,11 @@ body#idx > * {
     margin-left: auto;
     margin-right: auto;
     text-align: center;
-    width: 25em;
+    width: 35em;
 }
 
 body#idx > #pitch {
-    width: 30rem;
+    width: 35rem;
 }
 
 #pitch em {
@@ -88,7 +88,6 @@ body#idx p.other-platforms-help {
     background-color: rgb(250, 250, 250);
     margin-left: auto;
     margin-right: auto;
-    width: 30rem;
     text-align: center;
     border-radius: 3px;
     border: 1px solid rgb(204, 204, 204);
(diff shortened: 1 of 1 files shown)


birkenfeld:
For me, looks like this:

image

So a bit more generous, and we're good :)

www/rustup.css
```diff
@@ -63,11 +63,11 @@ body#idx > * {
     margin-left: auto;
     margin-right: auto;
     text-align: center;
-    width: 25em;
+    width: 30em;
```
harrysarson:
```suggestion
    width: 35em;
```

www/rustup.css
```diff
@@ -96,7 +96,7 @@ body#idx p.other-platforms-help {
 }
 
 .instructions > * {
-    width: 34rem;
+    width: 39rem;
```
harrysarson:
```suggestion
    width: 45rem;
```

www/rustup.css
```diff
@@ -116,11 +116,12 @@ hr {
     margin-left: auto;
     margin-right: auto;
     padding: 1rem;
-    width: 32rem;
+    width: 37rem;
```
harrysarson:
```suggestion
    width: 45rem;
```

harrysarson (commented):
How does it look now?

harrysarson:
@brikenfeld have you had a chance to have a look?

birkenfeld:
LGTM now, thanks!

kinnison (approved):
I see no reason not to accept this.  Website updates happen on releases of rustup normally so you won't see any updates until the release of 1.22.0 but then it should be done.  Thank you all for your efforts cleaning this up.

PR #2209, opened 2020-01-21
tests: Use format macro to generate string

tesuji:


kinnison (approved):
A reasonable little cleanup, thanks.

PR #2210, opened 2020-01-21
Nicer checking for not found dir

tesuji:


kinnison (approved):
Simplification looks good and reduces "OK" pathways, which is nice.

PR #2211, opened 2020-01-21
Remove needless map

tesuji:


kinnison:
The failure case showing here is fascinating -- Do you have any idea why it might be occurring?

tesuji:
The test succeeded on second run: https://github.com/rust-lang/rustup/pull/2211/checks?check_run_id=402222296

kinnison:
Yeah I saw that, but wondered what, if anything, might cause that.  It has exit code zero and yet failed the `.success()` check.  If you have no idea then I guess we put it down to a test bogon?

kinnison (approved):
The change is good.

tesuji:
Sorry I have no ideas.

kinnison:
@lzutao Fair enough, I'll keep an eye out.  Thanks for the patch.

PR #2212, opened 2020-01-22
Added support for aria2c in rustup-init.sh

thedrow:
aria2c uses multiple connections to download a file which should make installation slightly faster on good connections but will improve installation time on bad connections. There's no reason not to use it if it's available.

aria2c already defaults to tlsv1.2 (See https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-min-tls-version) so there's no need to check for it as far as I know.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -372,20 +372,24 @@ ignore() {
     "$@"
 }
 
-# This wraps curl or wget. Try curl first, if not installed,
-# use wget instead.
+# This wraps aria2c, curl or wget. Try aria2c first, if not installed,
+# use curl instead. If curl isn't installed try wget.
 downloader() {
     local _dld
-    if check_cmd curl; then
+    if check_cmd aria2c; then
+        _dld=aria2c
+    elif check_cmd curl; then
         _dld=curl
     elif check_cmd wget; then
         _dld=wget
     else
-        _dld='curl or wget' # to be used in error message of need_cmd
+        _dld='aria2c, curl or wget' # to be used in error message of need_cmd
     fi
 
     if [ "$1" = --check ]; then
         need_cmd "$_dld"
+    elif [ "$_dld" = aria2c ]; then
+        aria2c --allow-overwrite=true --dir / --no-conf=true -o "${2#?}" "$1"
     elif [ "$_dld" = curl ]; then
         if ! check_help_for "$3" curl --proto --tlsv1.2; then
             echo "Warning: Not forcing TLS v1.2, this is potentially less secure"


thedrow:
Funny, it worked locally for me.
I'll find out why exactly.

thedrow:
Fixed :)

rbtcollins:
I'm skeptical of this change for a few reasons.

* static.rust-lang.org is a cloud-front CDN, so there should be a local POP for most people, which means that multiple connections will have diminishing returns (since multiple connections to the same endpoint are mostly useful for dealing with poor BDP issues, and close POPs ameliorate those)
* http/2 is enabled on it, which is somewhat better at opening up a wide TCP window quickly anyway
* Once QUIC aka HTTP/3 is far enough along that AWS enable that on cloudfront, it will be better than multiple connections in and of itself

What performance improvement are you seeing with this change? How many seconds without, how many seconds with? Is your curl built with http2 support? What platform are you on?

emilyalbini:
> static.rust-lang.org is a cloud-front CDN, so there should be a local POP for most people, which means that multiple connections will have diminishing returns (since multiple connections to the same endpoint are mostly useful for dealing with poor BDP issues, and close POPs ameliorate those)

By the way, we only have [Price Class 100](https://aws.amazon.com/cloudfront/pricing/) enabled for static.rust-lang.org, which means we can only use pops in US, Europe, Canada and Israel.

thedrow:
> I'm skeptical of this change for a few reasons.
> 
> * static.rust-lang.org is a cloud-front CDN, so there should be a local POP for most people, which means that multiple connections will have diminishing returns (since multiple connections to the same endpoint are mostly useful for dealing with poor BDP issues, and close POPs ameliorate those)
> * http/2 is enabled on it, which is somewhat better at opening up a wide TCP window quickly anyway
> * Once QUIC aka HTTP/3 is far enough along that AWS enable that on cloudfront, it will be better than multiple connections in and of itself
> 
> What performance improvement are you seeing with this change? How many seconds without, how many seconds with? Is your curl built with http2 support? What platform are you on?

This contribution was born out of boredom (it took me an hour including all fixes) and not out of need but I still think it's useful for others. I'm not seeing any speedup from this since I'm in Israel.
Perhaps someone else can test this.

I agree that HTTP2 & HTTP3 are an improvement. Both of them are planned for aria2c. See https://github.com/aria2/aria2/issues/825 & https://github.com/aria2/aria2/issues/476.

Maybe we should prefer aria2c only if curl wasn't built with HTTP2 support?

kinnison:
While the change is not that complex, I have no experience with aria2c so would be unable to diagnose errors if people had problems with it.  As such, until someone can indicate if this actually does improve matters *for them* I suggest this change is held (not closed).  We can revisit it in a little while if nothing comes of this.

kinnison:
There has been no movement on this for a while, the author admits they aren't too fussed about it, and I don't see the value for the download of `rustup-init` which is hardly the largest thing downloaded during a rust installation.  I'm closing this, and if in the future we decide it's a problem we can revisit the idea.  Thank you for your time.

PR #2214, opened 2020-01-23
Allow installing alongside existing Rust (with warning)

sollyucko:
See https://github.com/rust-lang/rustup/issues/953

FYI, I haven't been able to test this because I run out of disk space, which is why telling users who want to install alongside an existing install to modify the source and recompile doesn't work. (Specificially, I am trying to run nightly Rust on repl.it.)

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -522,3 +522,17 @@ pub fn report_error(e: &Error) {
         }
     }
 }
+
+pub fn ignorable_error(error: crate::errors::Error, no_prompt: bool) -> Result<()> {
+    report_error(&error);
+    if no_prompt {
+        warn!("continuing (because the -y flag is set and the error is ignorable)");
+        Ok(())
+    } else {
+        if confirm("\nContinue? (y/N)", false).unwrap_or(false) {
+            Ok(())
+        } else {
+            Err(error)
+        }
+    }
+}
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -235,9 +235,16 @@ fn canonical_cargo_home() -> Result<String> {
 /// `CARGO_HOME`/bin, hard-linking the various Rust tools to it,
 /// and adding `CARGO_HOME`/bin to PATH.
 pub fn install(no_prompt: bool, verbose: bool, quiet: bool, mut opts: InstallOpts) -> Result<()> {
-    do_pre_install_sanity_checks(no_prompt)?;
+    if !env::var_os("RUSTUP_INIT_SKIP_EXISTENCE_CHECKS").map_or(false, |s| s == "yes") {
+        do_pre_install_sanity_checks(no_prompt)?;
+    }
+
     do_pre_install_options_sanity_checks(&opts)?;
-    check_existence_of_rustc_or_cargo_in_path(no_prompt)?;
+
+    if !env::var_os("RUSTUP_INIT_SKIP_EXISTENCE_CHECKS").map_or(false, |s| s == "yes") {
+        check_existence_of_rustc_or_cargo_in_path(no_prompt)?;
+    }
+
     #[cfg(unix)]
     do_anti_sudo_check(no_prompt)?;
 
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -30,7 +30,7 @@
 //! Deleting the running binary during uninstall is tricky
 //! and racy on Windows.
 
-use crate::common::{self, Confirm};
+use crate::common::{self, ignorable_error, Confirm};
 use crate::errors::*;
 use crate::markdown::md;
 use crate::term2;
@@ -235,7 +235,7 @@ fn canonical_cargo_home() -> Result<String> {
 /// `CARGO_HOME`/bin, hard-linking the various Rust tools to it,
 /// and adding `CARGO_HOME`/bin to PATH.
 pub fn install(no_prompt: bool, verbose: bool, quiet: bool, mut opts: InstallOpts) -> Result<()> {
-    do_pre_install_sanity_checks()?;
+    do_pre_install_sanity_checks(no_prompt)?;
     do_pre_install_options_sanity_checks(&opts)?;
     check_existence_of_rustc_or_cargo_in_path(no_prompt)?;
     #[cfg(unix)]
@@ -378,8 +378,8 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
     // Only the test runner should set this
     let skip_check = env::var_os("RUSTUP_INIT_SKIP_PATH_CHECK");
 
-    // Ignore this check if called with no prompt (-y) or if the environment variable is set
-    if no_prompt || skip_check == Some("yes".into()) {
+    // Skip this if the environment variable is set
+    if skip_check == Some("yes".into()) {
         return Ok(());
     }
 
@@ -390,13 +390,12 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
         err!("Otherwise you may have confusion unless you are careful with your PATH");
         err!("If you are sure that you want both rustup and your already installed Rust");
         err!("then please restart the installation and pass `-y' to bypass this check.");
-        Err("cannot install while Rust is installed".into())
-    } else {
-        Ok(())
+        ignorable_error("cannot install while Rust is installed".into(), no_prompt)?;
     }
+    Ok(())
 }
 
-fn do_pre_install_sanity_checks() -> Result<()> {
+fn do_pre_install_sanity_checks(no_prompt: bool) -> Result<()> {
     let rustc_manifest_path = PathBuf::from("/usr/local/lib/rustlib/manifest-rustc");
     let uninstaller_path = PathBuf::from("/usr/local/lib/rustlib/uninstall.sh");
     let rustup_sh_path = utils::home_dir().unwrap().join(".rustup");
@@ -412,7 +411,7 @@ fn do_pre_install_sanity_checks() -> Result<()> {
             "run `{}` as root to uninstall Rust",
             uninstaller_path.display()
         );
-        // return Err("cannot install while Rust is installed".into());
+        ignorable_error("cannot install while Rust is installed".into(), no_prompt)?;
     }
 
     if rustup_sh_exists {
@@ -422,7 +421,10 @@ fn do_pre_install_sanity_checks() -> Result<()> {
         warn!("or, if you already have rustup installed, you can run");
         warn!("`rustup self update` and `rustup toolchain list` to upgrade");
         warn!("your directory structure");
-        return Err("cannot install while rustup.sh is installed".into());
+        ignorable_error(
+            "cannot install while rustup.sh is installed".into(),
+            no_prompt,
+        )?;
     }
 
     Ok(())
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -391,12 +391,13 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
     }
 
     if let Err(path) = rustc_or_cargo_exists_in_path() {
-        err!("it looks like you have an existing installation of Rust at:");
-        err!("{}", path);
-        err!("rustup should not be installed alongside Rust. Please uninstall your existing Rust first.");
-        err!("Otherwise you may have confusion unless you are careful with your PATH");
-        err!("If you are sure that you want both rustup and your already installed Rust");
-        err!("then please restart the installation and pass `-y' to bypass this check.");
+        warn!("it looks like you have an existing installation of Rust at:");
+        warn!("{}", path);
+        warn!("rustup should not be installed alongside Rust. Please uninstall your existing Rust first.");
+        warn!("Otherwise you may have confusion unless you are careful with your PATH");
+        warn!("If you are sure that you want both rustup and your already installed Rust");
+        warn!("then please reply `y' or `yes' or set RUSTUP_INIT_SKIP_PATH_CHECK to yes");
+        warn!("or pass `-y' to ignore all ignorable checks.");
         ignorable_error("cannot install while Rust is installed".into(), no_prompt)?;
     }
     Ok(())
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -412,7 +412,7 @@ fn do_pre_install_sanity_checks() -> Result<()> {
             "run `{}` as root to uninstall Rust",
             uninstaller_path.display()
         );
-        return Err("cannot install while Rust is installed".into());
+        // return Err("cannot install while Rust is installed".into());
     }
 
     if rustup_sh_exists {
--- a/tests/cli-inst-interactive.rs
+++ b/tests/cli-inst-interactive.rs
@@ -8,6 +8,8 @@ use crate::mock::clitools::{
 };
 use crate::mock::{get_path, restore_path};
 use lazy_static::lazy_static;
+use rustup::utils::raw;
+use std::fs;
 use std::io::Write;
 use std::process::Stdio;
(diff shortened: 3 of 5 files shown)


kinnison:
Hi @sollyucko 

Thanks for your wanting to contribute to Rustup.  I'm sorry that you're having difficulty with `repl.it` and `rustup`.  The change as you propose it is not acceptable because it removes a sanity check which we want to remain present for the most part.  Better would be to make a change similar to the one in #705 - specifically the env var `RUSTUP_INIT_SKIP_PATH_CHECK` and the use of the `no_prompt` flag from the `-y` command like argument.  I think, perhaps, introducing `RUSTUP_INIT_SKIP_INSTALLED_CHECK` might be a good name and you could alter the path check and the manifest sanity check to honour that (and the latter to honour the `no_prompt`).

If you want to work on this, let me know, otherwise it'd be better to open a more descriptive issue detailing how this is necessary for `repl.it` and then close this PR to allow someone else to have a go at fixing it.

I hope you're up for the work, because I think it'd be useful, but I understand if you don't have the space to do it, rustup is quite big to build :D

sollyucko:
What do you think of my latest commit (46cf331)? It currently requires either replying to a confirmation prompt or specifying the -y flag, and produces warnings no matter what.

sollyucko:
Oops, fixed formatting. (c11150c)

sollyucko:
Need to fix test cases...

src/cli/common.rs
```diff
@@ -522,3 +522,18 @@ pub fn report_error(e: &Error) {
         }
     }
 }
+
+pub fn non_fatal_error(error: crate::errors::Error, question: &str, default: bool, no_prompt: bool) -> Result<()> {
+    report_error(&error);
+    if no_prompt {
```
kinnison, line 528: Does this skipping also want an environment variable, or is `-y` sufficient?

src/cli/self_update.rs
```diff
@@ -390,29 +390,28 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
         err!("Otherwise you may have confusion unless you are careful with your PATH");
         err!("If you are sure that you want both rustup and your already installed Rust");
         err!("then please restart the installation and pass `-y' to bypass this check.");
-        Err("cannot install while Rust is installed".into())
-    } else {
-        Ok(())
+        non_fatal_error("cannot install while Rust is installed".into(), "\nContinue? (y/N)", false, no_prompt)?;
```
kinnison: I'm not convinced it's worth passing in this prompt, the sense of the prompt is essential to the logic of your `non_fatal_error()` function so I'd just have that string in there. Ditto for the default really.

src/cli/common.rs
```diff
@@ -522,3 +522,18 @@ pub fn report_error(e: &Error) {
         }
     }
 }
+
+pub fn non_fatal_error(error: crate::errors::Error, question: &str, default: bool, no_prompt: bool) -> Result<()> {
```
kinnison: Perhaps "ignorable_error()" might be a better name?

src/cli/self_update.rs
```diff
@@ -390,29 +390,28 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
...
+fn do_pre_install_sanity_checks(no_prompt: bool) -> Result<()> {
     let rustc_manifest_path = PathBuf::from("/usr/local/lib/rustlib/manifest-rustc");
     let uninstaller_path = PathBuf::from("/usr/local/lib/rustlib/uninstall.sh");
     let rustup_sh_path = utils::home_dir().unwrap().join(".rustup");
     let rustup_sh_version_path = rustup_sh_path.join("rustup-version");
 
     let rustc_exists = rustc_manifest_path.exists() && uninstaller_path.exists();
     let rustup_sh_exists = rustup_sh_version_path.exists();
-
+
```
kinnison: whitespace bogon crept in here.

kinnison (changes requested):
As you say, you'd need to deal with test hiccoughs (I imagine there's plenty of installer tests which will need tweaking, and I'll want to see at least some hint of testing the bypassing of the error if possible), and there're some bits to tweak, but this is making progress.

src/cli/common.rs
sollyucko: Currently, it just uses `-y`, but I could also add an environment variable.

src/cli/self_update.rs
sollyucko: OK, makes sense. I guess i18n would be the most likely reason for changing the prompt, and that it doesn't seem like rustup does that at all right now.

src/cli/common.rs
sollyucko: Good idea.

src/cli/self_update.rs
sollyucko: Hmm, oops.

src/cli/common.rs
sollyucko: I have now added `RUSTUP_INIT_SKIP_CHECKING_EXISTENCE` to completely skip the check.

sollyucko:
Assuming there aren't any more issues, I think all I have left to do is to add some more test cases.

sollyucko:
@kinnison This now has a working implementation and tests for each method of skipping or ignoring the check. Is this ready to merge, or are there any issues I should fix? Thanks for the suggestions, BTW.

kinnison:
Hi @sollyucko thanks for your efforts in improving this work.  I shall review the over-all diff since you have so many commits on this branch.  Assuming I'm happy with the work, we'll need it rebasing into a smaller number of commits which logically express the work you did; rather than the actual process by which you achieved the final result.  That way, the history is more useful if we need to work out why a line of code is the way it is.

I shall review your work tonight.

src/cli/self_update.rs
```diff
@@ -378,8 +385,8 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
     // Only the test runner should set this
     let skip_check = env::var_os("RUSTUP_INIT_SKIP_PATH_CHECK");
 
-    // Ignore this check if called with no prompt (-y) or if the environment variable is set
-    if no_prompt || skip_check == Some("yes".into()) {
+    // Skip this if the environment variable is set
+    if skip_check == Some("yes".into()) {
```
kinnison, line 389: Why hav eyou ceased obeying `no_prompt` here?

tests/cli-inst-interactive.rs
```diff
@@ -266,3 +268,112 @@ fn test_warn_if_complete_profile_is_used() {
...
+
+        cmd.stdin(Stdio::piped());
+        cmd.stdout(Stdio::piped());
+        cmd.stderr(Stdio::piped());
+        let _out = cmd.spawn().unwrap().wait_with_output().unwrap();
+        let out = SanitizedOutput {
+            ok: _out.status.success(),
+            stdout: String::from_utf8(_out.stdout).unwrap(),
+            stderr: String::from_utf8(_out.stderr).unwrap(),
+        };
```
kinnison: I feel like if this wasn't already possible via clitools more directly, then we need to improve that, rather than having this code in a test case.  It makes it harder to see what the test case is up to.

kinnison (changes requested):
Your testing looks excellent.  A few comments are inline, and as previously stated, once things look good we'll need the commit sequence rebasing to clean it up.

I am very excited by this work.

src/cli/self_update.rs
sollyucko: I changed it to show output even when `no_prompt` is used, but still continue afterwards.

tests/cli-inst-interactive.rs
sollyucko: This was copied from `run_input`, not `clitools`.

sollyucko:
Yeah, the commits are definitely a mess...

tests/cli-inst-interactive.rs
kinnison: I see.  It'd be preferable to make `run_input()` more generic (perhaps `run_input_with_env()` with `run_input()` changed to call through that) just so that we don't end up with code duplication for this stuff.

src/cli/self_update.rs
kinnison: Aah yes, I see.  I shall have to give that a go and decide if I like the feel of it, but not until the other issue is resolved and you've rebased things.  It'll be a final check that I'm OK with the feel of the whole feature.

I wonder if we need the two environment variables documented in `README.md`?  Perhaps so, though perhaps as a separate PR afterwards.

tests/cli-inst-interactive.rs
sollyucko: Good idea.

sollyucko:
Does this `git rebase` input look good?

```
p 46cf331 Made some errors non-fatal.
s c11150c Ran rustfmt.
p 5a88830 Fixed test case.
r a2db8ae Renamed `non_fatal_error` to `ignorable_error`.
r 88b7de9 Added environment variable `RUSTUP_INIT_SKIP_CHECKING_EXISTENCE`.
s f55f76e Oops, got environment variable check backwards...
r 0e57beb Added tests for ignorable errors.
s d0292ef Ran rustfmt.
s fa3e3b4 Fixed `#[ignore]` attribute.
s 527be7d Fixed tests (e.g. setting up environment for test).
s 24a0863 Added `use` statements.
s 9ebc560 Parameter types go *after* the name and a colon...
s 30e0273 Fixed error.
s db6db45 rustfmt
s da0a80a Fixed test errors.
r 7445b7a Added test for `RUSTUP_INIT_SKIP_CHECKING_EXISTENCE`.
r d5064eb Created `run_input_with_env`.
s 414cee1 Removed unnecessary `use`s.

# Rebase 703b405..414cee1 onto 703b405 (18 commands)
```

sollyucko:
Oops, I meant `git rebase`.

kinnison:
I'd probably make the `run_input_with_env()` stuff come before the new test cases, but the basic shape of the result looks good.  You'll then want to run again and reword your commits to make the commit messages be nice, rather than the concatenation of messages which squash will result in.

kinnison (commented):
This change is starting to look really good.  Let's get the rebase done, and then I will go over the UX aesthetics by running it on my laptop to see how it feels.  In general this is very promising though.

sollyucko:
Rebase done!

tests/cli-inst-interactive.rs
kinnison: In your rebase I think you lost this bit of the test change.

kinnison (changes requested):
You've managed to lose the use of `run_input_with_env()` in your test case during the rebase.  That needs fixing.

Also, do you need to keep the renaming of `non_fatal()` as a separate commit or can that be folded into the original work?

sollyucko:
> You've managed to lose the use of run_input_with_env() in your test case during the rebase. That needs fixing.

Oh, oops.

> Also, do you need to keep the renaming of non_fatal() as a separate commit or can that be folded into the original work?

I'll squash it in.

sollyucko:
Both of those should now be fixed.

kinnison:
I shall re-review this tonight, Thanks.

kinnison (approved):
The code now looks good, I need to see how I feel when I run this locally and experiment, but I'm pretty sure it's mergeable.

sollyucko:
Sounds good, thanks!

src/cli/self_update.rs
```diff
@@ -390,13 +397,12 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
         err!("Otherwise you may have confusion unless you are careful with your PATH");
         err!("If you are sure that you want both rustup and your already installed Rust");
         err!("then please restart the installation and pass `-y' to bypass this check.");
-        Err("cannot install while Rust is installed".into())
-    } else {
-        Ok(())
+        ignorable_error("cannot install while Rust is installed".into(), no_prompt)?;
```
rbtcollins, line 401: Errata: this is *not* ignorable if the location of the existing rust is in the CARGO_BIN directory that rustup's proxies will be written to. We're making this easier to run into, so perhaps we should check for that at this point.

sollyucko: @rbtcollins Should it be possible to skip it in any way (e.g. environment variable) or not?

rbtcollins: Currently the code will give a warning later about "tool X is already installed", and not touch those binaries. This would mean that rustup isn't actually installed. I don't think it makes sense to allow that to be bypassed, but would like @kinnison 's input. This might be something to do as a separate PR for instance.

kinnison: The cases this PR covers are to do with `rustc` on the `PATH` or about detecting an old-style `rustup.sh` install as I understand the change.  If it happens that it's in `CARGO_BIN` then the user is going to get warnings as you say, and if they ignore those then they get to keep all the pieces.  I don't think we can stop *all* the footguns -- this is simply a way to allow another footgun for someone who really can't install `rustup` otherwise due to immutable underlying layers.

rbtcollins: The concern I have is that this is escalating it from a read-the-docs situation to a y/n prompt, so perhaps a lot easier to perform targeted foot removal.

kinnison: Possibly so.  I shall check the feel of that in my testing which I'm about to undertake.

kinnison: Okay so the issue here is that if the `rustc/cargo` is in `$CARGO_BIN` then the only real pain is that `rustup` will replace them with proxies without prompting.  The only tools you get the warnings about are `rustfmt` and `cargo-fmt` because those used to be provided via `cargo install` rather than always being proxied via `rustup`.  I think that the footgun is permissible in this case because it's *exceedingly* unlikely that the user will have set up this kind of thing without also being aware of what `rustup-init` will end up doing

kinnison:
Hi @sollyucko 

I've now done some user-testing and I'm concerned about some of the wording...

When you have `/usr/bin/rustc` and you try and install you get:

```
error: it looks like you have an existing installation of Rust at:
error: /usr/bin
error: rustup should not be installed alongside Rust. Please uninstall your existing Rust first.
error: Otherwise you may have confusion unless you are careful with your PATH
error: If you are sure that you want both rustup and your already installed Rust
error: then please restart the installation and pass `-y' to bypass this check.

Continue? (y/N) 
```

It seems odd to both tell the user they **have** to restart and pass `-y` *yet* offer them the chance to continue.  We'll need to change that wording a little, perhaps to something like:

```
error: If you are sure that you want both rustup and your already installed Rust
error: then you can continue, though be aware that it may cause confusion depending
error: on your PATH settings.  You can bypass this check by passing `-y` when installing.
```

Otherwise I think I'm okay with this.  So if you can look over and provide a better message, we're close to being able to merge this.

src/cli/self_update.rs
```diff
@@ -390,13 +397,12 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool) -> Result<()> {
         err!("Otherwise you may have confusion unless you are careful with your PATH");
         err!("If you are sure that you want both rustup and your already installed Rust");
         err!("then please restart the installation and pass `-y' to bypass this check.");
```
kinnison: This is the message which needs tweaking now that you can answer `y` rather than **having** to pass `-y`

kinnison (changes requested):

src/cli/self_update.rs
sollyucko: Should I still have the prompt, or force the (possibly overly broad) command-line argument?

kinnison: I still want the prompt, the point is that the message needs tweaking *because* we want the prompt :D

sollyucko: Yep, makes sense, just double-checking.

sollyucko:
Does this look good?

kinnison (approved):
Let's get this merged and see how we feel over time.  Thank you.

PR #2217, opened 2020-01-27
Fix typo

LitoMore:
### Changes

- `Mac OS` -> `macOS`
- `Github` -> `GitHub`


kinnison (approved):
Normally I'd prefer two commits when the workflows are changed, but this is simple enough to verify by eye.  Thank you for these fixes.

PR #2218, opened 2020-01-31
Improve compile time in release profile

tesuji:
By setting proc-macro crates to not use optimization even in
release build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,3 +110,7 @@ path = "src/cli/main.rs"
 [profile.release]
 lto = true
 codegen-units = 1
+
+# Reduce build time by setting proc-macro crates non optimized.
+[profile.release.build-override]
+opt-level = 0


Cargo.toml
```diff
@@ -110,3 +110,7 @@ path = "src/cli/main.rs"
 [profile.release]
 lto = true
 codegen-units = 1
+
+# Reduce build time by setting proc-macro crates non optimized.
+[profile.release.build-override]
```
rbtcollins, line 115: its not clear to me how this only affects proc-macro crates. Can you help me understand ?

tesuji: I read that from https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#overrides

rbtcollins: That says
"""
To override the settings for build scripts, proc macros, and their dependencies, use the build-override table:

# Set the settings for build scripts and proc-macros.
[profile.dev.build-override]
opt-level = 3

    Note: When a dependency is both a normal dependency and a build dependency, Cargo will try to only build it once when --target is not specified. When using build-override, the dependency may need to be built twice, once as a normal dependency and once with the overridden build settings. This may increase initial build times.
"""

So my expectation is that this will cause everything that is in the transitive dependency tree of our build scripts and proc-macros and also in our regular dependencies to be built twice, yes?

I  wonder if that affects us.

Anyhow, a quick unscientific test:

cargo clean; cargo build --release with the patch and without:
Finished release [optimized] target(s) in 2m 25s
Finished release [optimized] target(s) in 3m 02s

So it does seem to make a difference, and those numbers are consistent across multiple runs.  I'm not sure how much diff it makes to an incremental build, and don't have time to test that right now.

kinnison:
Interestingly the macos rules appear to be missing the changes I did to the Linux rules to deal with the existing toolchains on  github actions.  Otherwise this is good.  I'll fix those and then merge this.

PR #2230, opened 2020-02-13
Change README.md to discourage the use of `complete`

### LINKED ISSUE #2228: README about profiles encourages use of `complete` which is a bad idea
The `README.md` profiles section needs adjusting to not mislead people into thinking `complete` is a good idea for the general users because it'll almost never work.

This should be a relatively easy fix to make.

I'm happy to help someone through making this change.

LeSeulArtichaut:
My proposal to resolve #2228.

cc @kinnison

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,12 @@ default (`rustc`, `rust-std`, `cargo`, and `rust-docs`) plus `rustfmt` and
 `clippy`. This profile will be used by `rustup` by default, and it's the one
 recommended for general use.
 * The **complete** profile includes all the components available through
-`rustup`, including `miri` and IDE integration tools (`rls` and `rust-analysis`).
+`rustup`. This should never be used, as it includes *every* component ever included
+in the metadata and thus will almost always fail. If you are looking for a way
+to install devtools such as `miri` or IDE integration tools (`rls`, `rust-analysis`),
+you should use the `default` profile and install the needed additional components
+manually, either by using `rustup component add` or by using `-c` when installing
+the toolchain.
 
 To change the `rustup` profile you can use the `rustup set profile` command. For
 example, to select the minimal profile you can use:


README.md
```diff
@@ -78,7 +78,11 @@ default (`rustc`, `rust-std`, `cargo`, and `rust-docs`) plus `rustfmt` and
 `clippy`. This profile will be used by `rustup` by default, and it's the one
 recommended for general use.
 * The **complete** profile includes all the components available through
-`rustup`, including `miri` and IDE integration tools (`rls` and `rust-analysis`).
+`rustup`. This should never be used, as it includes *every* component ever included
+in the metadata and thus will almost always fail. If you are looking for a way
+to install devtools such as `miri` or IDE integration tools (`rls`, `rust-analysis`),
+you should use the `default` profile and install the needed additional components
+manually by running `rustup component add`.
```
rbtcollins: Another way to add the components is to use the -c option: `rustup toolchain add nightly -c rls -c rust-analysis` for instance. I don't know if its worth mentioning that, but if folk are looking for a one-shot way to make it happen, we have it already.

LeSeulArtichaut: I think it is a bit long for a short description of profiles in a README, but it is definitely worth discussing

kinnison: You could improve the wording to satisfy the `-c` stuff to with something like:
```suggestion
manually, either by using `rustup component add` or by using `-c` when installing
the toolchain.
```

kinnison (commented):
I think it's OK as it is, but if you like my suggestion, please integrate it into your commit.

LeSeulArtichaut:
@kinnison I think this is ready to merge, except if you want me to squash the two commits

kinnison (changes requested):
Please do squash things together (and make sure the commit message is good)

kinnison (approved):

PR #2235, opened 2020-02-21
Let docs pass through with custom toolchains

cuviper:
Following #2116, custom toolchains started to fail `rustup doc`, since
they can't list components to check if `rust-docs` is installed.

    error: toolchain 'system' does not support components
    error: caused by: invalid toolchain name: 'system'

Now custom toolchains just skip that component check, and it's up to the
user to have installed the documentation in their own way.


rbtcollins:
I think this raises a point about our API: custom toolchains shouldn't let us list try to list components - that code should have forced a compile error in the first place to help us understand that this problem would have arisen.

PR #2236, opened 2020-02-22
Fix #2229 detect available memory

### LINKED ISSUE #2229: allocation failure installing rustc
**Problem**

Running `rustup update` to upgrade an install to the 1.41.0 stable release today, I have a repeatable failure:
```info: installing component 'cargo'
  4.3 MiB /   4.3 MiB (100 %)   1.9 MiB/s in  2s ETA:  0s
 15.5 MiB /  15.5 MiB (100 %)   2.6 MiB/s in  5s ETA:  0s
 48.6 MiB /  65.5 MiB ( 74 %)   2.0 MiB/s in 26s ETA:  8smemory allocation of 152166216 bytes failedAborted
```
This is on a mips host with 512 MiB of physical ram. I'd expect this to work on a larger machine, but it also doesn't seem correct that the installer is asking for twice the size of the file it's unpacking.

**Steps**

1. Install rustup and rust 1.40.0.
2. Limit host to half a GB of memory.
3. Invoke `rustup update`.

**Possible Solution(s)**

Since the previous releases all worked, it's possible the rustc package just got big enough to trigger this in the 1.41.0 release.

**Notes**

I ran `rustup self update` to upgrade the tool itself to 1.21.1, but that didn't resolve the issue. I can also reproduce with rustup 1.20.2 re-installed by an older rustup-init binary, so the bug doesn't seem to be particularly new.

Note that this left the host without a working toolchain, since the installer removes the old release before unpacking the new one. I'm also unable to downgrade since `rustup` seems to now require `clippy` which wasn't available for this plafform in previous releases. Is there a way around that?

Command-line `tar xvf` works fine to unpack the archive.

Output of `rustup --version`:
```
```

Output of `rustup show`:
```
Default host: mipsel-unknown-linux-gnu
rustup home:  /home/giles/.rustup

stable-mipsel-unknown-linux-gnu (default)
(rustc does not exist)
```

rbtcollins:
This doesn't implement streaming IO for low memory situations - we still
have the situation that low footprint situations will fail to install,
but while it is the case that rustc's memory footprint is lower than our
unpack footprint this is probably not urgent to fix, though I will get
around to it.

Being less aggressive about unpack buffer size though should reduce the
number of support tickets from folk in these cases, I hope.

We may end up getting tickets from folk with broken ulimit syscalls
though, who knows.


rbtcollins:
One possible critique of the patch is that I haven't wired up the UI to the eventing system. TBH I couldn't be bothered, but I can do that if thats the only issue.

However I'm less and less enthralled by that system as time progresses; as we dont support the library as an API, and it's very cumbersome, I'd really like to bring a better system into being.

src/dist/component/package.rs
```diff
@@ -163,19 +164,37 @@ struct MemoryBudget {
     used: usize,
 }
 
+static mut RAM_NOTICE_SHOWN: bool = false;
```
kinnison: I dislike that this will result in unsafe -- could we use some kind of interior-mutability thing instead, perhaps an `AtomicBool` -- that seems more plausible and shouldn't need unsafe.

src/dist/component/package.rs
```diff
@@ -163,19 +164,37 @@ struct MemoryBudget {
...
-    fn new(max_file_size: usize) -> Self {
-        const DEFAULT_UNPACK_RAM: usize = 400 * 1024 * 1024;
-        let unpack_ram = if let Ok(budget_str) = env::var("RUSTUP_UNPACK_RAM") {
-            if let Ok(budget) = budget_str.parse::<usize>() {
-                budget
-            } else {
-                DEFAULT_UNPACK_RAM
+    fn new(max_file_size: usize, effective_max_ram: usize) -> Self {
+        const DEFAULT_UNPACK_RAM_MAX: usize = 500 * 1024 * 1024;
+        let ram_for_unpacking = effective_max_ram - (100 * 1024 * 1024);
```
kinnison: Perhaps:
```suggestion
        const RAM_ALLOWANCE_FOR_OS: usize = 100 * 1024 * 1024;
        let ram_for_unpacking = effective_max_ram - RAM_ALLOWANCE_FOR_OS;
```

src/dist/component/package.rs
```diff
@@ -163,19 +164,37 @@ struct MemoryBudget {
...
+        let default_max_unpack_ram = std::cmp::min(DEFAULT_UNPACK_RAM_MAX, ram_for_unpacking);
+        let unpack_ram = match env::var("RUSTUP_UNPACK_RAM")
+            .ok()
+            .and_then(|budget_str| budget_str.parse::<usize>().ok())
+        {
+            Some(budget) => budget,
```
kinnison, line 183: Should we do something if this is more than `effective_max_ram` ?

tests/cli-exact.rs
```diff
@@ -38,6 +38,7 @@ info: downloading component 'rust-docs'
 info: downloading component 'rust-std'
 info: downloading component 'rustc'
 info: installing component 'cargo'
+Defaulting to 500.0 MiB unpack ram
```
kinnison: Can we guarantee that all our CI systems will permit this, or should we actually set a limit ourselves to ensure this doesn't occur?

kinnison (changes requested):
I'd prefer to see the use of `AtomicBool` otherwise I think it looks good.  How confident are you about `effective_limits` as a crate for this?  I'll defer to your judgement if you say it looks/feels sound.

src/dist/component/package.rs
rbtcollins: We can. Today we only do one unpack at a time so it is safe, but I will fix this to avoid future foot-guns.

src/dist/component/package.rs
rbtcollins: That overhead is for rustup and the tar buffers, but yes, I'll add a constant for clarity.

src/dist/component/package.rs
rbtcollins: Perhaps. 

One the one hand, they may know that that 100MB overhead is too generous, for instance. On the other hand they know that the autodetection code is broken for some reason. On the gripping hand we may get bug reports that 'when I tell rustup to use X RAM it breaks'.

tests/cli-exact.rs
rbtcollins: It passed CI. Any ulimit / job limit we set will by necessity be lower than what CI has available, not higher.

We can't guarantee developers will have 500MB available, or that IDEs won't put ulimits around test runs etc. But if an IDE puts a hard ulimit around a test run, we can't actually raise it ourselves anyway.

kinnison: OK, I guess we know what to fix if it turns out to be a problem in future.  I've got plenty of RAM for running my tests too :D

src/dist/component/package.rs
```diff
@@ -163,19 +166,40 @@ struct MemoryBudget {
     used: usize,
 }
 
+lazy_static! {
+    static ref RAM_NOTICE_SHOWN: AtomicBool = AtomicBool::new(false);
+}
+
```
kinnison: I'm not *certain* that this needs `lazy_static!` -- I think AtomicBool is meant to be used as a synchronisation primitive.  Certainly a lot of its methods take `&self` rather than `&mut self`

kinnison (changes requested):
I think the use of AtomicBool somehow isn't right because the defaulting message is no longer displayed.  Either way the tests need checking because they're failing.

src/dist/component/package.rs
rbtcollins: ah, I see, right, iterating.

kinnison (approved):
I am really getting to be grumpy about our notification system :D  For now, :+1:

PR #2239, opened 2020-02-25
cli: Allow specifying multiple component or target values in one argument

jplatte:
It has happened a few times to me now that I've tried to install a toolchain with `rustup toolchain install <toolchain> -c a,b,c`, only to get an error that the component "a,b,c" does not exist.

This PR changes the meaning of that command to the (IMO obvious) one of installing the toolchain `<toolchain>` with components `a`, `b` and `c`. Targets can also be specified in one `-t` argument when delimited with commas.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -318,7 +318,8 @@ pub fn cli() -> App<'static, 'static> {
                                 .long("component")
                                 .short("c")
                                 .takes_value(true)
-                                .multiple(true),
+                                .multiple(true)
+                                .use_delimiter(true),
                         )
                         .arg(
                             Arg::with_name("targets")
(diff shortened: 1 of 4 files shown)


kinnison (changes requested):
Nominally this change is good, however if we want to guarantee to support this going into the future then we need a test to demonstrate the functionality.  Could you please sort a test out (probably in cli-v2 or cli-misc) and push it to the PR?

jplatte:
I've added tests for `rustup`, should I create additional ones for `rustup-init`?

payload (approved):

payload (approved):

kinnison (changes requested):
This looks good,  However as you note, there are changes in `setup_mode.rs` which do need verifying too.

Please try and a test to confirm the behaviour in the installer.

jplatte:
@kinnison Done.

kinnison (approved):

PR #2240, opened 2020-02-26
Make list_components safer to work with

rbtcollins:
PR 2116 introduced a regression in linked toolchain support. The
regression was introduced in part because there was no type information
signifying the difference between uninstalled toolchains and toolchains
that don't have components, thus the compiler didn't hint during
development that any discrimination was needed.

This patch introduces a type level discrimination to avoid future issues
of this type. The code is a little crufty due to error_chain glue
issues, but there is a longer term path to making that clean.


kinnison (approved):
I can see why you're wanting us to use typestate control for the toolchains.  For now this does help keep us away from mistakes though. :+1:

PR #2244, opened 2020-02-28
fallback-settings file

rbtcollins:
This provides a way for snaps, distro packages and other such installs
to set the default-toolchain for users that have not gone through the
rustup-init one-time set of questions.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -11,6 +11,7 @@ use pgp::{Deserializable, SignedPublicKey};
 
 use crate::dist::{dist, temp};
 use crate::errors::*;
+use crate::fallback_settings::FallbackSettings;
 use crate::notifications::*;
 use crate::settings::{Settings, SettingsFile, DEFAULT_METADATA_VERSION};
 use crate::toolchain::{Toolchain, UpdateStatus};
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -133,6 +137,13 @@ impl Cfg {
 
         let settings_file = SettingsFile::new(rustup_dir.join("settings.toml"));
 
+        // Centralised file for multi-user systems to provide admin/distributor set initial values.
+        let fallback_settings = if cfg!(not(windows)) {
+            FallbackSettings::new(PathBuf::from(UNIX_FALLBACK_SETTINGS))?
+        } else {
+            None
+        };
+
```
kinnison, line 146: Would it be better to have fallback_settings always be present via `Default::default()`

rbtcollins: I'm torn
it makes a certain bit of code cleaner
on the other hand, it means we erase knowledge of whether a fallback file was supplied, which may be useful later

kinnison: I accept that the knowledge may be useful in the future, so let's not erase it for now.  It's a trivial refactor if we decide later it's worth the cleanup.

PR #2252, opened 2020-03-04
Auto-install toolchains more consistently

rbtcollins:
This combines all auto-installation logic to the root of the codepaths
leading to running a binary, which should lead to auto-installation
happening for all cases, as long as a toolchain has been selected.

Config.find_default no longer verifies the toolchain in order to permit
this auto-installation to take place in one location; the callers of
find_default do not generally need verified toolchains, so this should
be fine.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -309,11 +309,8 @@ impl Cfg {
     }
 
     pub fn which_binary(&self, path: &Path, binary: &str) -> Result<Option<PathBuf>> {
-        if let Some((toolchain, _)) = self.find_override_toolchain_or_default(path)? {
-            Ok(Some(toolchain.binary_file(binary)))
-        } else {
-            Ok(None)
-        }
+        let (toolchain, _) = self.find_or_install_override_toolchain_or_default(path)?;
+        Ok(Some(toolchain.binary_file(binary)))
     }
 
     pub fn upgrade_data(&self) -> Result<()> {
@@ -370,10 +367,7 @@ impl Cfg {
             .with(|s| Ok(s.default_toolchain.clone()))?;
 
         if let Some(name) = opt_name {
-            let toolchain = self
-                .verify_toolchain(&name)
-                .chain_err(|| ErrorKind::ToolchainNotInstalled(name.to_string()))?;
-
+            let toolchain = Toolchain::from(self, &name)?;
             Ok(Some(toolchain))
         } else {
             Ok(None)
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -495,17 +486,30 @@ impl Cfg {
         Ok(None)
     }
 
-    pub fn find_override_toolchain_or_default(
+    pub fn find_or_install_override_toolchain_or_default(
         &self,
         path: &Path,
     ) -> Result<Option<(Toolchain<'_>, Option<OverrideReason>)>> {
-        Ok(
+        if let Some((toolchain, reason)) =
```
kinnison, line 490: This feels a smidge awkward as a double-nested `if let` -- could it be split out a bit?  I know there's a tendency to prefer expressions over statements in Rust code, but this is just a little too nested for my brain to grok easily.

kinnison (commented):
Looks generally good.  Let's see how the CI feels.

tests/cli-v1.rs
```diff
@@ -162,10 +166,10 @@ fn remove_default_toolchain_err_handling() {
     setup(&|config| {
         expect_ok(config, &["rustup", "default", "nightly"]);
         expect_ok(config, &["rustup", "toolchain", "remove", "nightly"]);
-        expect_err(
+        expect_stderr_ok(
             config,
-            &["rustc"],
-            for_host!("toolchain 'nightly-{0}' is not installed"),
+            &["rustc", "--version"],
+            "info: installing component",
```
kinnison, line 172: With this change, I feel the function name for the test isn't brilliant.  If you can't be bothered to fix it I don't mind, otherwise it'd be nice if it were renamed something like `remove_default_toolchain_auto_reinstalls` or somesuch

tests/cli-v2.rs
```diff
@@ -249,10 +253,10 @@ fn remove_default_toolchain_err_handling() {
     setup(&|config| {
         expect_ok(config, &["rustup", "default", "nightly"]);
         expect_ok(config, &["rustup", "toolchain", "remove", "nightly"]);
-        expect_err(
+        expect_stderr_ok(
             config,
-            &["rustc"],
-            for_host!("toolchain 'nightly-{0}' is not installed"),
+            &["rustc", "--version"],
+            "info: installing component",
```
kinnison, line 259: Same comment here.

kinnison (approved):
If you think the names of the test functions are okay as-is, then you're good to merge, otherwise rename them and then merge :D

tests/cli-v1.rs
rbtcollins: `remove_default_toolchain_autoinstalls` is what my fingers put put :P. I noticed after doing it that the very next test, for overridden toolchains, had the same naming bug, but ce la vie.

PR #2259, opened 2020-03-10
Use dependent types to make programming errors harder

rbtcollins:
This extends the minimal fix done the other week into a full blown set of types around the different sorts of distributions in use.


kinnison (approved):
This looks good.  IMO it may may sense to add some `Deref` impls to allow the more constrained types to deref back to the underlying `Toolchain` and it may make sense to add some nice methods to `Toolchain` to get the newtypes (e.g. `foo.as_distributable()` rather than `DistributableToolchain::new(foo)`) but none of that should block merging, this makes things a lot clearer with better separation of concern on the behaviours.

Nice job.

Issue #2233, opened 2020-02-16
Confusing error message when trying to call rust-lldb on windows


skade:
**Problem**

This is a problem I caught on twitter: https://twitter.com/robdaemon/status/1229188704666820608

Calling `rust-lldb` on a windows toolchain msvc seems to result in the following interaction:

```
rust-lldb.exe is not installed...
To install, run `rustup component add lldb-preview`.
```
Running the command results in:

```
component 'lldb-preview' for target x86-64-pc-windows-msvc...
```

This interaction is not useful, especially as rust-lldb is not available for this target.

**Possible Solution(s)**

My suggested solution would be to error out early with a message that MSVC targets are not supported by lldb.

robdaemon:
![rust-lldb](https://user-images.githubusercontent.com/287213/74615406-45f28800-50d5-11ea-9869-f608b4636a43.png)

adding a screenshot here, as my Twitter feed gets purged regularly.

kinnison:
The useful change for this would be to ensure that when raising the `BinaryNotFound` from `Toolchain::create_command()` we should check the toolchain's available components and add a boolean to the raise when the command would not be available in the toolchain.  This would have the added bonus that in nightlies where `rustfmt` or `clippy` is unavailable, we'd explain that up-front rather than risking confusion later.

If someone wishes to tackle this, I'd be pleased to help them find the right places to make such a change.

virome:
Hello kinnison, I would like to take this one, if possible.

rbtcollins:
I think an enum would be more useful.
Result<ToolchainCommand>

enum ToolchainCommand {
UnknownBinary
BinaryNotBuilt
Command(command)
}

Where errors indicate failures like the toolchain isn't actually installed, etc, Command(command) holds the built command, and we can tell the difference between a binary that should exist but is temporarily missing, and permanently - if the manifest has that info?
o

virome:
I was wondering along the same lines. Possibly a new error, maybe something like `CommandNotSupportedOnToolchain` that would be for instances when we've identified it as a command that's commonly available on other platforms but not the current one (how we would do that, I'm not sure yet), whereas `BinaryNotFound` would be for FS level errors.

kinnison:
@virome Hi and I'm excited you'd like to try and fix this.  I'm glad you're agreeing with the approach that @rbtcollins is suggesting, and if you want further discussion on these topics then there is `#wg-rustup` on the rust-lang discord server https://discord.gg/rust-lang

[...]

PR #2264, opened 2020-03-15
Add error message to indicate that a missing command is not available…

virome:
… on the current toolchain

Addresses #2233

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -1,3 +1,4 @@
+use crate::component_for_bin;
 use crate::config::Cfg;
 use crate::dist::dist::TargetTriple;
 use crate::dist::dist::ToolchainDesc;
@@ -274,6 +275,30 @@ impl<'a> InstalledCommonToolchain<'a> {
                 .and_then(|s| s.parse().ok())
                 .unwrap_or(0);
             if recursion_count > env_var::RUST_RECURSION_COUNT_MAX - 1 {
+                let binary_lossy: String = binary.to_string_lossy().into();
+                if let Ok(distributable) = DistributableToolchain::new(self.0) {
+                    if let (Some(component_name), Ok(component_statuses), Ok(Some(manifest))) = (
+                        component_for_bin(&binary_lossy),
+                        distributable.list_components(),
+                        distributable.get_manifest(),
+                    ) {
+                        let component_status = component_statuses
+                            .iter()
+                            .find(|cs| cs.component.short_name(&manifest) == component_name)
+                            .expect(&format!(
+                                "component {} should be in the manifest",
+                                component_name
+                            ));
+                        if !component_status.available {
+                            return Err(ErrorKind::BinaryProvidedByUnavailableComponent(
+                                component_status.component.short_name(&manifest),
+                                binary_lossy,
+                                self.0.name.clone(),
+                            )
+                            .into());
+                        }
+                    }
+                }
                 let defaults = self.0.cfg.get_default()?;
                 return Err(ErrorKind::BinaryNotFound(
                     binary.to_string_lossy().into(),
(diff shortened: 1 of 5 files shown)


src/toolchain.rs
```diff
@@ -325,9 +326,31 @@ impl<'a> Toolchain<'a> {
...
+                        ));
+                    if !component_status.available {
+                        return Err(ErrorKind::BinaryProvidedByUnavailableComponent(
+                            component_status.component.short_name(&manifest),
+                            binary_lossy,
+                            self.name.clone(),
+                        )
+                        .into());
+                    }
+                }
```
virome, line 309: One thing about this approach is it does slow this error path down. Previously, with the BinaryNotFound error, it would return the error instantly. Now, regardless of whether or not the component is available, it takes a moment to return the suggestion. Is there a better way I'm not seeing than looping over the component_statuses?

But since it's behind the path exists check it does not affect the happy path.

virome:
I'll add a test for this, have to learn the test infrastructure first though.

tests/cli-misc.rs
```diff
@@ -445,7 +445,31 @@ fn rls_exists_in_toolchain() {
...
+                EXE_SUFFIX,
+                this_host_triple(),
+            ),
+        );
+
+        set_current_dist_date(config, "2015-01-02");
+        expect_ok(config, &["rustup", "update", "--no-self-update"]);
+        expect_ok(config, &["rustup", "component", "add", "rls"]);
+
+        assert!(config.exedir.join(format!("rls{}", EXE_SUFFIX)).exists());
```
kinnison: This assert checks for the proxy, which you already know exists because of the `expect_err()` call above succeeding in order to get this far.  As such either it should be moved to the top of this test, or just dropped.

src/toolchain.rs
kinnison: I don't think it's a problem to have slowdown in the error reporting path, so I don't object to this.

kinnison (dismissed):
The approach looks okay, and slowing down the error path in order to be more useful is not a problem in my opinion.

This needs rebasing after a recent merge to master, so fix up the test a bit and rebase please.

tests/cli-misc.rs
virome: Ah, I was thinking that the proxies didn't exist until being installed. But that does clarify some things for sure.

virome:
I'm running this on windows, and it works for rust-lldb, but it panics on rust-gdb.
For gdb, component_for_bin is `gdb-preview` but  neither `gdb` nor `gdb-preview` are in the latest stable or nightly manifests. There is a gdb proxy though which is why we can even get to my code.

I can make it not panic anymore, but it would be back to the old 'rust-gdb not installed, please run to install rust-gdb'. 

EDIT: Am I correct that the binaries for rust-gdb and rust-lldb actually come in the rustc package? Also the llvm-preview package is unavailable on unknown-linux-gnu. Is that there as a historical artifact before they were moved to rustc? It's not necessarily a problem for the PR as it stands because the rust-lldb binary is actually found and executed, but I'm wondering if we should be telling windows users that the 'llvm-preview' component is not on windows when in reality it's actually the window's rustc that's missing the rust-lldb executable.

I'm not very knowledgeable on the history of rustup packaging.

kinnison:
I think it's quite possible that those binaries come in the `rustc` component these days.  I'm afraid I'm as clueless as you are with respect to the history of all this.

Regarding telling them of the right component is definitely valuable.  If rust-gdb and rust-lldb aren't actually available on Windows at all; then perhaps we should remove those proxies?

kinnison:
It's actually not right to remove the proxies under windows because they might be used by the msvc toolchain.  I've pushed what I hope will be an appropriate message for when you try and run a command which would normally be available but is not applicable to a given toolchain (e.g. msvc toolchain lacks rust-gdb and rust-lldb, but the windows-gnu toolchain has them)

kinnison:
@virome Do you think this looks right to you now?

virome:
Hmm, let me confirm. I was in the middle of fixing a panic when I stopped to ask that clarifying question. Let me check with these changes.

PR #2266, opened 2020-03-16
Anyhow migration

rbtcollins:
error_chain is giving us more and more headaches; time to let it pass beyond the vale.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -172,15 +172,15 @@ pub fn set_globals(verbose: bool, quiet: bool) -> anyhow::Result<Cfg> {
 pub fn show_channel_update(
     cfg: &Cfg,
     name: &str,
-    updated: rustup::Result<UpdateStatus>,
-) -> Result<()> {
+    updated: anyhow::Result<UpdateStatus>,
+) -> anyhow::Result<()> {
     show_channel_updates(cfg, vec![(name.to_string(), updated)])
 }
 
 fn show_channel_updates(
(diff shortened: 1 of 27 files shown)


src/cli/common.rs
```diff
@@ -3,9 +3,11 @@
 use crate::errors::*;
 use crate::self_update;
 use crate::term2;
+use anyhow;
```
lzutao: We shouldn't need this use statement.

src/errors.rs
```diff
@@ -370,7 +341,168 @@ error_chain! {
...
+#[derive(ThisError, Debug)]
+pub enum RustupError {
+    #[error("invalid toolchain name: '{0}'")]
+    InvalidToolchainName(String),
+    #[error("Unable to proceed. Could not locate working directory.")]
+    LocatingWorkingDir {
+        #[from]
+        source: io::Error,
+    },
+    #[error("failed to set permissions for '{}'", .p.display())]
```
dtolnay, line 342:
```suggestion
    #[error("failed to set permissions for '{p}'")]
```

We'll display the path automatically.

kinnison:
@rbtcollins Are you likely to be able to do any more on this any time soon?  it would be nice to get it merged before we end up pushing changes which make it harder to rebase.

kinnison:
I have a feeling that migrating many of these patches will now be a pain with the rebase needed, however if we can get it to a point that it's working then I'd love to get more of this migration done.

PR #2274, opened 2020-03-29
Escape Windows backslash when outputing to markdown

### LINKED ISSUE #2272: Missing backslash in printed CARGO_HOME path on Windows
**Problem**

I installed Rust using rustup under Windows 10 and noticed a missing backslash in the post install message which tells you the Cargo bin directory:

```
Rust is installed now. Great!

To get started you need Cargo's bin directory (%USERPROFILE%.cargo\bin) in your
PATH

Press the Enter key to continue.
```

It should be `%USERPROFILE%\.cargo\bin` with a backslash after `%USERPROFILE%` to be valid.

Note that it printed the following earlier:

```
It will add the cargo, rustc, rustup and other commands to
Cargo's bin directory, located at:

  C:\Users\Simran\.cargo\bin
```

And it added this to the PATH environment variable correctly.

**Steps**

1. [Download rustup-init.exe (64-bit)](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe)
2. `1) Proceed with installation (default)`
3. Wait until installation completes and check the last block of messages

**Possible Solution(s)**

The message is coming from here:

_src\cli\self_update.rs_

```rust
macro_rules! post_install_msg_win {
    () => {
        r"# Rust is installed now. Great!

To get started you need Cargo's bin directory ({cargo_home}\\bin) in your `PATH`
"
    };
}
```

The backslash before `bin` (which shows correctly) is escaped, but the rest is coming from the variable `cargo_home`. Interestingly, a different variable is used for the other message, `cargo_home_bin`:

```rust
macro_rules! pre_install_msg_template {
…
It will add the `cargo`, `rustc`, `rustup` and other commands to
Cargo's bin directory, located at:

    {cargo_home_bin}
```

It could be used instead of `{cargo_home}\\bin` I guess, but with the env var `%USERPROFILE%` resolved. I'm not sure how else to fix the problem. The following looks alright using a raw string, removing the need for escaping the backslash:

```rust
fn canonical_cargo_home() -> Result<String> {
…
            path_str = String::from(r"%USERPROFILE%\.cargo");
```

Or does it need to be double-escaped to `r"%USERPROFILE%\\.cargo"` (non-raw `"%USERPROFILE%\\\\.cargo"` I suppose)?

**Notes**

Output of `rustup --version`: rustup 1.21.1 (7832b2ebe 2019-12-20)

Output of `rustup show`:

```
rustup home:  C:\Users\Simran\.rustup

rustc 1.42.0 (b8cedc004 2020-03-09)
```

tesuji:
Closes #2272
Duplicate close #2174

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -65,7 +65,7 @@ pub const NEVER_SELF_UPDATE: bool = false;
 // argument of format! needs to be a literal.
 
 macro_rules! pre_install_msg_template {
-    ($platform_msg: expr) => {
+    ($platform_msg:literal) => {
         concat!(
             r"
 # Welcome to Rust!
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -41,6 +41,7 @@ use rustup::utils::Notification;
 use rustup::{Cfg, UpdateStatus};
 use rustup::{DUP_TOOLS, TOOLS};
 use same_file::Handle;
+use std::borrow::Cow;
 use std::env;
 use std::env::consts::EXE_SUFFIX;
 use std::fs;
@@ -215,22 +216,21 @@ static UPDATE_ROOT: &str = "https://static.rust-lang.org/rustup";
 
 /// `CARGO_HOME` suitable for display, possibly with $HOME
 /// substituted for the directory prefix
-fn canonical_cargo_home() -> Result<String> {
+fn canonical_cargo_home() -> Result<Cow<'static, str>> {
     let path = utils::cargo_home()?;
-    let mut path_str = path.to_string_lossy().into_owned();
 
     let default_cargo_home = utils::home_dir()
         .unwrap_or_else(|| PathBuf::from("."))
         .join(".cargo");
-    if default_cargo_home == path {
+    Ok(if default_cargo_home == path {
         if cfg!(unix) {
-            path_str = String::from("$HOME/.cargo");
+            "$HOME/.cargo".into()
         } else {
-            path_str = String::from(r"%USERPROFILE%\.cargo");
+            r"%USERPROFILE%\.cargo".into()
         }
-    }
-
-    Ok(path_str)
+    } else {
+        path.to_string_lossy().into_owned().into()
+    })
 }
 
 /// Installing is a simple matter of copying the running binary to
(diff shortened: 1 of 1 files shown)


tesuji:
Result from windows: https://github.com/rust-lang/rustup/pull/2274/checks?check_run_id=543340236#step:15:21

rbtcollins (approved):
LGTM though I'm not sure if the variation between windows and unix needs to be as large as it is.

ci/actions-templates/windows-builds-template.yaml
```diff
@@ -113,6 +113,12 @@ jobs:
       # A script might be nicer, though these in theory could be made to run
       # independently, assuming the run above succeeded, allowing us to run
       # the full test suite rather than stop on first error, at least in PRs
+      - name: Run the CARGO_HOME in self-installation step
+        run: |
+          set -e
+          rm ~/.cargo/bin/rustup
+          target/${{ matrix.target }}/release/rustup-init.exe -y --default-host ${{ matrix.target }} --default-toolchain=stable --profile=minimal
+        shell: bash
```
kinnison: It's not clear to me what your goal is here.  Why are you installing the built rustup into the test host at this time?
Given this is part of a `do not merge` commit, did you mean to leave it in?

src/cli/self_update.rs
```diff
@@ -65,7 +65,7 @@ pub const NEVER_SELF_UPDATE: bool = false;
 // argument of format! needs to be a literal.
 
 macro_rules! pre_install_msg_template {
-    ($platform_msg: expr) => {
+    ($platform_msg:literal) => {
```
kinnison, line 69: This commit seems unrelated to your other work, why did you do it?

kinnison (changes requested):
There are `[do not merge]`  commits in this sequence.  Please rebase them away.

ci/actions-templates/windows-builds-template.yaml
tesuji: I don't have a Windows machine and I don't wanna run the virtual machine.
So to check if the new installation prompt for Windows displays correctly, 
I add this commit.

src/cli/self_update.rs
tesuji: The macro argument of that macro is always literal to me.
I'd move it to another PR (this change is so small) if you prefer.

PR #2278, opened 2020-04-03
Clarify help message of `rustup toolchain help link`

### LINKED ISSUE #2276: `rustup toolchain help link` has misleading help message for `<toolchain>`
**Problem**

The output from `rustup toolchain help link` contains the following:

> ARGS:
>    <toolchain>    Toolchain name, such as 'stable', 'nightly', or '1.8.0'. For more information see `rustup help
>                   toolchain`

But just after that, the longer explanation also says:

>    'toolchain' is the custom name to be assigned to the new toolchain.
>    Any name is permitted as long as it does not fully match an initial
>    substring of a standard release channel. For example, you can use
>    the names 'latest' or '2017-04-01' but you cannot use 'stable' or
>    'beta-i686' or 'nightly-x86_64-unknown-linux-gnu'.

IOW, the examples given under `ARGS` are not valid to use there (well, maybe 1.8.0 is, it's not all that clear).

**Notes**

Output of `rustup --version`:
```
```

ssomers:
Fixes #2276.
And tries to mimic the style of other options.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -360,10 +360,14 @@ pub fn cli() -> App<'static, 'static> {
                         .after_help(TOOLCHAIN_LINK_HELP)
                         .arg(
                             Arg::with_name("toolchain")
-                                .help(TOOLCHAIN_ARG_HELP)
+                                .help("Custom toolchain name")
                                 .required(true),
                         )
-                        .arg(Arg::with_name("path").required(true)),
+                        .arg(
+                            Arg::with_name("path")
+                                .help("Path to the directory")
+                                .required(true),
+                        ),
                 ),
         )
         .subcommand(


ssomers:
Help message then looks like this:
```
rustup.exe-toolchain-link
Create a custom toolchain by symlinking to a directory

USAGE:
    rustup.exe toolchain link <toolchain> <path>

FLAGS:

ARGS:
    <toolchain>    Custom toolchain name
    <path>         Path to the directory

DISCUSSION:
    'toolchain' is the custom name to be assigned to the new toolchain.
    Any name is permitted as long as it does not fully match an initial
    substring of a standard release channel. For example, you can use
    the names 'latest' or '2017-04-01' but you cannot use 'stable' or
    'beta-i686' or 'nightly-x86_64-unknown-linux-gnu'.

    'path' specifies the directory where the binaries and libraries for
    the custom toolchain can be found. For example, when used for
    development of Rust itself, toolchains can be linked directly out of
    the build directory. After building, you can test out different
    compiler versions as follows:

        $ rustup toolchain link latest-stage1 build/x86_64-unknown-linux-gnu/stage1
        $ rustup override set latest-stage1

    If you now compile a crate in the current directory, the custom
    toolchain 'latest-stage1' will be used.
```

kinnison (approved):
This certainly looks more correct.

Thank you for your contribution.

PR #2299, opened 2020-04-25
ci: $PWD is equivalent to $(pwd)

tesuji:


kinnison (approved):
I guess this simplifies things a little, though I'd not have worried personally.

PR #2300, opened 2020-04-25
Measure impact of LTO to compilation times on CI for non-stable builds

tesuji:


kinnison:
Do you still not have Actions enabled on your own fork of rustup?  This noise is quite irritating, especially when unbidden.

PR #2301, opened 2020-04-25
Clarify help message with the Cargo home directory

futurist:
Before this PR, the help message is:

```
Cargo's bin directory, located at:

        {cargo_home_bin}

This can be modified with the CARGO_HOME environment variable.
```

above **WILL** potentially mislead the user to set `CARGO_HOME` to `{cargo_home_bin}`, which is wrong.

This PR clearify the concept of `the Cargo home directory`, to separate from `Cargo's bin directory`, in fact, it's 2 different concepts and folders.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -74,19 +74,23 @@ macro_rules! pre_install_msg_template {
 This will download and install the official compiler for the Rust
 programming language, and its package manager, Cargo.
 
-It will add the `cargo`, `rustc`, `rustup` and other commands to
-Cargo's bin directory, located at:
-
-    {cargo_home_bin}
-
-This can be modified with the CARGO_HOME environment variable.
-
 Rustup metadata and toolchains will be installed into the Rustup
 home directory, located at:
 
     {rustup_home}
 
 This can be modified with the RUSTUP_HOME environment variable.
+
+The Cargo home directory located at:
+
+    {cargo_home}
+
+This can be modified with the CARGO_HOME environment variable.
+
+The `cargo`, `rustc`, `rustup` and other commands will be added to
+Cargo's bin directory, located at:
+
+    {cargo_home_bin}
 
 ",
             $platform_msg,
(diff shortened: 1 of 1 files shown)


src/cli/self_update.rs
```diff
@@ -75,9 +75,10 @@ This will download and install the official compiler for the Rust
...
 
 It will add the `cargo`, `rustc`, `rustup` and other commands to
-Cargo's bin directory, located at:
+Cargo's bin directory, under the Cargo home directory, which is
+located at:
 
-    {cargo_home_bin}
+    {cargo_home}
 
 This can be modified with the CARGO_HOME environment variable.
```
tesuji:
```suggestion
This can be modified with the CARGO_HOME environment variable
The cargo's bin directory is always under $CARGO_HOME/bin.
```

tesuji (commented):
Rewording to make it clear could be a better choice.

src/cli/self_update.rs
futurist: OK, merged.

tesuji: I don't quite happy with my suggestion. 
The old `cargo_home_bin` is correct. But the paragraph needs to be rephrased.

futurist: _How about below one?_

This will download and install the official compiler for the Rust
programming language, and its package manager, Cargo.

The Cargo's home directory located at:

    {cargo_home}

This can be modified with the CARGO_HOME environment variable.

The `cargo`, `rustc`, `rustup` and other commands will be added to
Cargo's bin directory, located at:

    {cargo_home_bin}

futurist: The last commit with below message shown:

屏幕快照 2020-04-25 下午6 35 32

kinnison (changes requested):
This looks better over-all, but I'd prefer it if you would clean up / squash the commits so that the commit history is nicer.

kinnison (approved):

PR #2304, opened 2020-04-25
ci: Add rustc hash for caching target dir

tesuji:


kinnison (approved):
I had wanted to do this but was unsure how, thank you!

Issue #2285, opened 2020-04-13
Help for proxy programs should somehow grant a peek behind The Matrix


ssomers:
When rustup acts as a proxy (to cargo, rustc, any others?), it does too good a job of hiding itself. `cargo --help` gives no hint to noobs like me that there is somewhere else to look at (like `rustup --help`) to understand examples featuring a `cargo +toolchain` command.

`cargo --help`, or at least `cargo +stable --help`, should either:

- Say up front or down below that they're a wrapper for cargo with an additional option.
- List the +toolchain option among cargo's options (but still make clear that is not part of cargo, not documented there).
- Pass a secret option to the proxy, and/or rely on cargo and others to [document the option itself](https://github.com/rust-lang/cargo/issues/8058)? I sure don't like it.
- Do something else?

Looking at the code, this looks like a non-trivial job for the clap crate to be exploited at src/cli/proxy_mode.rs. Which I'll look into myself, unless someone says it's a bad idea or knows how already.

kinnison:
I guess we could grab the first argument check if it's `--help` and if so, provide a little help about the proxy setup *before* chaining through to the proxied command.  @rbtcollins what do you think about that as a UX?

```shell
$ cargo --help
rustup 1.xx.y (blahblah)
Proxy for `cargo`

USAGE:
    cargo [+toolchain] ...cargo arguments...

---
Rust's package manager

USAGE:
    cargo [OPTIONS] [SUBCOMMAND]

OPTIONS:
    -V, --version           Print version info and exit
        --list              List installed commands
......
```

ssomers:
I tried implementing that, but now I think that if you're new, and you don't care about toolchains and you just installed rustup because that's how you're supposed to start with Rust, this looks complicated and scary.

So I tried a short line after the normal page, and it looks better to me. But `rustup help` doesn't quite explain that it acts like a proxy. Is there or should there be a different help page for proxy use?

kinnison:
Most people won't, if they're new, be using the `+toolchain` stuff unless told to, and at that point they'll probably have it explained to them as to why.  The fact that it's `rustup` and not `cargo` which is handling that won't bother them a lot.  I think it's fine for the proxies to add their ha'pence to the help message but not for adding additional proxy help to the main `rustup`.  I *suppose* you could add a `rustup help proxies` if clap supports that kind of thing (and then `rustup proxies` would probably list all the proxies it has installed and whether the components for them are installed in the current toolchain, hmm, that could be useful)

[...]

PR #2307, opened 2020-04-26
Extend the help message when rustup acts as a proxy

ssomers:
Alternate but not quite right proposal for #2285

### DIFF
--- a/src/cli/proxy_mode.rs
+++ b/src/cli/proxy_mode.rs
@@ -9,10 +9,10 @@ use std::ffi::OsString;
 use std::path::PathBuf;
 use std::process;
 
-pub fn main() -> Result<()> {
+pub fn main(name: &str) -> Result<()> {
     crate::self_update::cleanup_self_updater()?;
 
-    let ExitCode(c) = {
+    let (help, ExitCode(c)) = {
         let _setup = job::setup();
 
         let mut args = env::args_os();
@@ -38,12 +38,22 @@ pub fn main() -> Result<()> {
         } else {
             env::args_os().skip(2).collect()
         };
+        let help = cmd_args.is_empty()
+            || cmd_args == &["help"]
+            || cmd_args == &["--help"]
+            || cmd_args == &["-h"];
 
         let cfg = set_globals(false, true)?;
         cfg.check_metadata_version()?;
-        direct_proxy(&cfg, &arg0, toolchain, &cmd_args)?
+        (help, direct_proxy(&cfg, &arg0, toolchain, &cmd_args)?)
     };
 
+    if help {
+        println!(
+            "\nThis proxy of {} supports an additional +toolchain option, see `rustup help`.",
+            name
+        );
+    }
     process::exit(c)
 }
 
--- a/src/cli/proxy_mode.rs
+++ b/src/cli/proxy_mode.rs
@@ -38,19 +38,16 @@ pub fn main(name: &str) -> Result<()> {
         } else {
             env::args_os().skip(2).collect()
         };
-        let help = cmd_args.is_empty()
-            || cmd_args == &["help"]
-            || cmd_args == &["--help"]
-            || cmd_args == &["-h"];
+        let help = cmd_args.is_empty() || cmd_args[0] == "--help" || cmd_args[0] == "-h";
 
         let cfg = set_globals(false, true)?;
         cfg.check_metadata_version()?;
         (help, direct_proxy(&cfg, &arg0, toolchain, &cmd_args)?)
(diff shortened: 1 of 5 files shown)


ssomers:
I forgot to mention: `cargo` and `rustc` have the same [`ArgRequiredElseHelp`](https://docs.rs/clap/2.33.0/clap/enum.AppSettings.html#variant.ArgRequiredElseHelp), so the empty command line checks out. But `rustc` does not have subcommands, so `rustc help` looks crappy:
```
error: couldn't read help: The system cannot find the file specified. (os error 2)

This proxy of rustc supports an additional +toolchain option, see `rustup help`.
```

In addition, I wonder if the -h and --help option exist like that on all platforms. It might have been /h in Windows (but it isn't) and Mac is uncharted territory for me.

src/cli/proxy_mode.rs
```diff
@@ -38,12 +38,22 @@ pub fn main() -> Result<()> {
         } else {
             env::args_os().skip(2).collect()
         };
+        let help = cmd_args.is_empty()
+            || cmd_args == &["help"]
```
kinnison: Do you want this to happen if someone does `cargo help build` ?  Maybe, maybe not.

src/cli/proxy_mode.rs
```diff
@@ -38,12 +38,22 @@ pub fn main() -> Result<()> {
...
 
         let cfg = set_globals(false, true)?;
         cfg.check_metadata_version()?;
-        direct_proxy(&cfg, &arg0, toolchain, &cmd_args)?
+        (help, direct_proxy(&cfg, &arg0, toolchain, &cmd_args)?)
     };
 
+    if help {
+        println!(
+            "\nThis proxy of {} supports an additional +toolchain option, see `rustup help`.",
```
kinnison: My suggestion of `rustup proxy` (perhaps `rustup proxy list`) would be a nicer spot for this to reference.  Of course, that doesn't exist yet...

kinnison (commented):
I think this is getting somewhere.  It probably needs some tests to make sure it will work.

src/cli/proxy_mode.rs
ssomers: I couldn't have worded it better...

But now I think it shouldn't even hook onto `cargo help`. Once you dive into subcommands, you're on your own. And of course it avoids the awkward output of `rustc help`.

ssomers: And by the way, there's no reason to expect extra flags besides -h or --help. `cargo -V -h` does not give you both version and help, and `cargo --color=always -h` does not colour your day. Help overrules almost everything, including  `cargo -h --baloney`, but not `cargo --baloney -h`.

But that still means it's better to compare the first argument and allow more arguments.

ssomers:
Struggled to test stuff, and not sure exactly what I'm testing. I did't see clear in the naming of modules and test cases, so just added to the "misc" pile.

ssomers:
Oops, never mind. This `direct_proxy` on non-Windows does an exec. I would suggest to put the Windows-only `process::exit` inside `run_command_for_dir` and not pretend that it returns.

We can't do anything after the `direct_proxy` call, and we can't just print some help text before it because the exec may fail (like when the proxy's target doesn't exist). To make this work, we're talking changing process treatment, at least for commands requesting help, and that's still emulating clap. It doesn't seem worth while.

kinnison:
Aah I hadn't thought of that, sorry that you feel your time was wasted.  I still think there's potentially value in working out how to improve matters for users though.

ssomers:
Sure, but I don't see any low hanging fruit any more. Unless you'd want to give up the unix-specific exec and do it with process::status or process::spawn, and possibly face some change in the way file descriptors or what not turn out.

It's not relevant on the road to better help, but I wonder if the current process::status isn't a (minor or philosophical) problem now: you start some cargo build on Windows, inadvertently keeping a lock on the rustup executable, and while that build is brewing, try to update rustup itself? Not that I ever saw `rustup self update` actually do anything. With a spawn instead, rustup could immediately exit in most cases, or wait on the child and complete the help message otherwise.

ssomers:
I realised something else smelled fishy: I had to check the exitcode to avoid producing the extra help text when the proxied help couldn't appear because the proxied binary is missing. That doesn't make sense in hindsight because, on both unix and windows, `run_command_for_dir` returns no exitcode but `Err` if the executable is missing. Also, I'm reminded now that without the check on exitcode, the extra help message is produced 4 times (says the test framework). One hour later I still can't figure out what is actually happening.

PR #2308, opened 2020-04-26
Update mingw version

tesuji:
to match what's currently used in rustc repo.


kinnison (approved):
Thanks for spotting the discrepancy.

PR #2313, opened 2020-04-28
riscv64gc-unknown-linux-gnu support

tblah:
## Current status
`rustup` builds on riscv, tests pass*, and rustup can install toolchains for supported host architectures.

Tested on Debian using Qemu full system emulation.

## To resolve WIP status
- [ ]  rust-lang/rust needs to publish official `riscv64gc-unknown-linux-gnu` host toolchains for `rustup` on riscv to be useful.
- [x] rust-lang/rust needs to publish a docker image to allow CI for `riscv64gc-unknown-linux-gnu` to run
- [x] https://github.com/rust-lang/libc/pull/1745 needs merging so that ca3cec0 can go away

### DIFF
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -46,6 +46,7 @@ jobs:
           - armv7-linux-androideabi       # skip-pr skip-master
           - i686-linux-android            # skip-pr skip-master
           - x86_64-linux-android          # skip-pr skip-master
+          - riscv64gc-unknown-linux-gnu   # skip-pr skip-master skip-stable
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
--- a/ci/actions-templates/README.md
+++ b/ci/actions-templates/README.md
@@ -56,6 +56,7 @@ system.
 | armv7-linux-androideabi       | Yes        | Two   | No     | No         |
 | i686-linux-android            | Yes        | Two   | No     | No         |
 | x86_64-linux-android          | Yes        | Two   | No     | No         |
+| riscv64gc-unknown-linux-gnu   | Yes        | ---   | No     | No         |
 | ----------------------------- | ---------- | ----- | ------ | ---------- |
 | x86_64-apple-darwin           | No         | One   | Yes    | Yes        |
 | ----------------------------- | ---------- | ----- | ------ | ---------- |
--- a/ci/docker/riscv64gc-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/riscv64gc-unknown-linux-gnu/Dockerfile
@@ -1,0 +1,11 @@
+FROM rust-riscv64gc-unknown-linux-gnu
+
+
+# We install a stable toolchain using rustup anyway, this should stop us getting
+# confused and still having this toolchain in $PATH
+RUN /usr/local/lib/rustlib/uninstall.sh
+# Weirdly that doesn't remove these:
+RUN rm /usr/local/bin/cargo /usr/local/bin/rust*
+
+ENV CC_riscv64gc_unknown_linux_gnu=riscv64-unknown-linux-gnu-gcc \
+    CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER=riscv64-unknown-linux-gnu-gcc
--- a/ci/fetch-rust-docker.bash
+++ b/ci/fetch-rust-docker.bash
@@ -32,6 +32,7 @@ case "$TARGET" in
   x86_64-unknown-freebsd)          image=dist-x86_64-freebsd ;;
   x86_64-unknown-linux-gnu)        image=dist-x86_64-linux ;;
   x86_64-unknown-netbsd)           image=dist-x86_64-netbsd ;;
+  riscv64gc-unknown-linux-gnu)     image=dist-various-1 ;;
   *) exit ;;
 esac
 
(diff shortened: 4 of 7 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ chrono = "0.4"
 clap = "2"
 download = { path = "download" }
 error-chain = "0.12"
-effective-limits = "0.4"
+effective-limits = { git = "https://github.com/tblah/effective-limits.rs.git", rev = "5fc9ca2fcf7c698d585325d3aad4a8bbe20b4418" }
```
kinnison: I take it that there needs to be a patch merged and a new release made for this?

kinnison (commented):
Obviously I can't merge while `effective-limits` isn't released, and with that CI change, but the rest could go in immediately if you wanted to either split the PR up, or simplify it in the short term.

mati865:
@tblah regarding your question in infra docker channel.

```
rustc -vV
rustc 1.41.0-dev
binary: rustc
commit-hash: unknown
commit-date: unknown
host: x86_64-unknown-linux-gnu
release: 1.41.0-dev
error: couldn't load codegen backend "/usr/local/lib/rustlib/x86_64-unknown-linux-gnu/codegen-backends/librustc_codegen_llvm-llvm.so": "/lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27\' not found (required by /usr/local/lib/rustlib/x86_64-unknown-linux-gnu/codegen-backends/librustc_codegen_llvm-llvm.so)"
```
^ this is not the Rust that you have installed with `rustup`.
Missing `rustup default stable` somewhere perhaps?

Cargo.toml
```diff
@@ -25,7 +25,7 @@ chrono = "0.4"
 clap = "2"
 download = { path = "download" }
 error-chain = "0.12"
-effective-limits = "0.4"
+effective-limits = { git = "https://github.com/tblah/effective-limits.rs.git", rev = "4873a008d2b441b2d47ce01f62c6f5ec9c157df9" }
```
tesuji: if bump libc, why not `cargo update -p libc` and remove this change.

tblah: The libc changes aren't in a released version yet so I think I have to tell cargo to fetch libc master branch instead of a release.

tblah:
Thanks @mati865 it seems to work now that it is building using the stable tool-chain.

tblah:
@kinnison sorry for the delay while I played with CI.

CI will only work with the effective-limits change, so is the next step to merge only 342642d10207dfa2f3a6667b936bc2a219c23fb5 and make a new WIP PR for CI?

kinnison:
As discussed on IRC, let's wait for the libc fix and then consider merging everything cleanly instead

kinnison (changes requested):
We'll need to undo the effective-limits thing and instead update `libc` and friends when the releases are done.  So let's not consider merging before then.

tblah:
An update on the status of this:

The `libc` version bump landed a little while ago. The last piece of the puzzle is to bump the version of `libc` required by `effective-limits` so that rustup is forced to upgrade. I have a [PR](https://github.com/rbtcollins/effective-limits.rs/pull/4) open for this.

kinnison:
Fantastic, I imagine @rbtcollins will get to that PR when he has a chance.  I think we might be worried about something else with E-L on another issue, so hopefully that could be soon.

kinnison:
Once that's done, if you could also rebase so that we're good in terms of the CI etc, that'd be excellent.

rbtcollins:
effective-limits stuff has merged, so this should be cleanup-able now

tblah:
Thanks @rbtcollins 

@kinnison this should be ready for review

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
           - armv7-linux-androideabi       # skip-pr skip-master
           - i686-linux-android            # skip-pr skip-master
           - x86_64-linux-android          # skip-pr skip-master
+          - riscv64gc-unknown-linux-gnu   # skip-pr skip-master
```
kinnison: This may also need a `skip-stable` because we're not going to release this for now.

kinnison (changes requested):
Fundamentally good stuff.  I'm slightly worried that people might expect a released rustup with the shell script changes, but we can explain it's not there yet, so I'm okay with that change there.  But I think we need a `skip-stable` in the action template, and then we'll need the workflows regenerating (as a separate commit)

ci/actions-templates/linux-builds-template.yaml
tblah: +1 see 67c272166fa3ea1414103f77e4dfd891c58e4580

kinnison (commented):
This is getting very close.  Could you please rebase to fold your two CI changes together; and then as a new commit add the result of running `sh ci/actions-templates/gen-workflows.sh` which will update the contents of the real workflows.

I'm okay if that results in failed CI on the PR for now, but I want to check things are updated properly (and not updated as appropriate for the stable build fr.ex)

tblah:
@kinnison I've done the rebase. Running `ci/actions-templates/gen-workflows.sh` doesn't change anything because `riscv64gc-unknown-linux-gnu` is `skip-pr skip-master skip-stable` (it is skipped in all cases)

kinnison (approved):

kinnison:
Cool, let's hope your efforts to get riscv-linux to tier two complete soon, so we can consider a rustup release with it :D

PR #2314, opened 2020-04-29
Reassert /bin/sh validity

workingjubilee:
Reverts the shebang change in 6b4283f and adds a shellcheck directive
to suppress warnings about POSIX validity and prevent further drift.

The POSIX standard is not, despite its name, a reliable guide to what
features are actually standard, even for POSIX shells. rustup-init.sh
uses `local`, a non-POSIX extension supported by all POSIX shells known
in common use. Thus, it is fine to run it with #!/bin/sh instead.

The Debian Almquist shell is not the only POSIX shell but it is very
common, so is a useful proxy for whether the script actually runs for
non-bash shells.

It's worth noting this was already what was used in CI, via CLI command.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -1,8 +1,12 @@
-#!/bin/bash
+#!/bin/sh
+# shellcheck shell=dash
 
 # This is just a little script that can be downloaded from the internet to
 # install rustup. It just does platform detection, downloads the installer
 # and runs it.
+
+# It runs on Unix shells like {a,ba,da,k,z}sh. It uses the common `local`
+# extension. Note: Most shells limit `local` to 1 var per line, contra bash.
 
 set -u
 


rustup-init.sh
```diff
@@ -1,9 +1,13 @@
-#!/bin/bash
+#!/bin/sh
+# shellcheck shell=dash
 
 # This is just a little script that can be downloaded from the internet to
 # install rustup. It just does platform detection, downloads the installer
 # and runs it.
 
+# It runs on Unix shells like {a,ba,da,k,z}sh. It uses the common `local`
+# extension. Note: Most shells limit `local` to 1 var per line, contra bash.
```
tesuji, line 9:
```suggestion
# extension. Note: Most shells limit `local` to 1 var per line but bash.
```

workingjubilee: The intent of the line is to warn a future dev against testing `local`'s behavior with bash and expecting it to be the same across shells.

Rewording would be fine if the turn of phrase is considered too unusual, but this particular diff would not convey my intent because "but" does not carry quite the same meaning of "in opposition to", and offers a misleading "every shell but bash" interpretation. It's possible but doubtful that bash is the only one that supports a generic extension of multiple locals in 1 line.

kinnison: I agree with @workingjubilee here.  There may be a nicer way to put it, but your suggestion definitely isn't it.  I'm going to merge as is.

PR #2316, opened 2020-05-01
notifications: list all missing components when skipping nightlies

unleashed:
Currently when updating and skipping nightlies due to missing required components only one such missing component is shown. Users might consider removing the requirement for the component and updating to the specific nightly that was missing it, only to find out that there is another, newly discovered missing component.

With this change all required and missing components are shown when a nightly is skipped, so users might be able to quickly identify which nightly version, if any, has an acceptable set of missing components.


kinnison (approved):

kinnison:
Thank you for this

PR #2320, opened 2020-05-05
Fixed the copyright owner & year for the Apache License

rapidclock:
##  Description of Change
The value for the year and owner was blank in the Apache License.
Mirrored the values of the year & owner from the MIT License (LICENSE-MIT).

## Testing
Checked the values in both `LICENSE-APACHE` and `LICENSE-MIT` to have the same values.

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -186,7 +186,7 @@ APPENDIX: How to apply the Apache License to your work.
    same "printed page" as the copyright notice for easier
    identification within third-party archives.
 
-Copyright [yyyy] [name of copyright owner]
+Copyright 2016 The Rust Project Developers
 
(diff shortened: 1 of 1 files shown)


unleashed:
Note that the license text you are modifying is just a _template_ copyright attribution suggestion to add to the project's files, so you could either do that, or add a `NOTICE` file covering the project with the copyright attribution (or do both).

Update: to be clear, changing the license text itself would not be the right thing to do, and it would create issues with automation for license detection.

kinnison:
This is as @unleashed says, not quite the right approach.  The copyright assertion, weak though it is, is in `README.md` and the dates etc. are handled via the source code repository itself.  I'd be willing to entertain an improvement to the `README.md` licence/copyright section instead, but this contribution, appreciated though it is, is not going to be merged.

PR #2322, opened 2020-05-08
Fix the shebang for rustup-init.sh

dimo414:
The [normal installation guidance](https://rustup.rs/) is to pipe the contents of this file into `sh`, which is not `bash` (it may be `dash` or `bash` in POSIX-mode in certain systems, neither of which support most Bash-isms). Given that this script is intended to be run by `sh` the shebang should reflect that.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # This is just a little script that can be downloaded from the internet to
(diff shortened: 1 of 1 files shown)


kinnison:
Hi @dimo414,

While your contribution is appreciated, @workingjubilee was already fixing this in another PR which has since been merged.

Thanks for your time, but you were a smidge late :(

D.

dimo414:
#2314 LGTM, thanks @workingjubilee!

PR #2326, opened 2020-05-10
www: Implement copy button for rustup download command

### LINKED ISSUE #2280: Copy button for Unix installation command on Rustup.rs

On the main [rustup.rs](https://rustup.rs/) installation site, the Unix installation command is very long and copying it requires having to click and drag while scrolling or triple-clicking.

unix command

Having a small button or intractable area that copies the command to the clipboard (and it provides feedback that it has done so) would make the installation process easier.

**Examples**

[Crates.io](https://crates.io/crates/lru) crate pages:
lru

[Npm](https://www.npmjs.com/package/react) package pages:
react

**Notes**

The fact that the crates.io team has already implemented this makes it seem like using/looking at their implementation would be worthwhile.

BeniCheni:
Implement a "_Copy_" button in www/ site page (inspired by [crates.io](https://crates.io/), to fix #2280 

Screen Shot 2020-05-10 at 5 50 53 PM

Screen Shot 2020-05-10 at 5 51 02 PM

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -25,7 +25,17 @@
 
 <div id="platform-instructions-unix" class="instructions display-none">
   <p>Run the following in your terminal, then follow the onscreen instructions.</p>
-  <pre>curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+    <div class="copy-container">
+      <pre>curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+      <button class="copy-button" title="Copy curl command to clipboard to download Rustup" type="button" onclick="handle_copy_button_click()">
+        <div class="copy-icon">
+          <svg width="24" height="25" viewBox="0 0 24 25" xmlns="http://www.w3.org/2000/svg" alt="Copy curl command to clipboard to download Rustup">
+            <path d="M18 20h2v3c0 1-1 2-2 2H2c-.998 0-2-1-2-2V5c0-.911.755-1.667 1.667-1.667h5A3.323 3.323 0 0110 0a3.323 3.323 0 013.333 3.333h5C19.245 3.333 20 4.09 20 5v8.333h-2V9H2v14h16v-3zM3 7h14c0-.911-.793-1.667-1.75-1.667H13.5c-.957 0-1.75-.755-1.75-1.666C11.75 2.755 10.957 2 10 2s-1.75.755-1.75 1.667c0 .911-.793 1.666-1.75 1.666H4.75C3.793 5.333 3 6.09 3 7z"></path><path d="M4 19h6v2H4zM12 11H4v2h8zM4 17h4v-2H4zM15 15v-3l-4.5 4.5L15 21v-3l8.027-.032L23 15z"></path>
+          </svg>
+        </div>
+        <div id="copy-status-message" class="copy-button-text"></div>
+      </button>
+    </div>
   <p class="other-platforms-help">You appear to be running Unix. If not, <a class="default-platform-button" href="#">display all supported installers</a>.</p>
 </div>
 
--- a/www/rustup.js
+++ b/www/rustup.js
@@ -2,6 +2,7 @@
 
 var platforms = ["default", "unknown", "win32", "win64", "unix"];
 var platform_override = null;
+var rustup_install_command = "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh";
 
 function detect_platform() {
     "use strict";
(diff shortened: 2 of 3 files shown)


kinnison (commented):
I think this looks okay, but I'm not an HTML/CSS/JS expert.  I wonder if @pietroalbini might know if this is sane?

emilyalbini:
I'm not familiar with clipboard handling unfortunately :(

kinnison:
Sadly the crates.io impl looks to be Ember based so I have no clue how to compare it.

kinnison:
I've just tested this locally and I get: TypeError: 'clipboard-write' (value of 'name' member of PermissionDescriptor) is not a valid value for enumeration PermissionName.

Was this tested in Firefox at all @BeniCheni ?

workingjubilee:
This does not need to use the Permissions API on Firefox, however Firefox requires a secure context instead. Please consult https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API for further details.

EDIT: Wrong link, whoops.

www/rustup.js
```diff
@@ -164,6 +164,33 @@ function fill_in_bug_report_values() {
...
+            navigator.clipboard.write([clippedItem]).then(function() {
+                document.getElementById('copied-text').innerText = 'Copied!'
+            }, function(error) {
+                console.error('Attempt to write to clipboard failed. Error:');
+                console.dir(error);
+            });
+        } else {
+            console.log('clipboard-permission not granted:', result);
+        }
+    });
```
workingjubilee: Providing a second function to Promise.then() will catch Promise ~~errors~~ rejections but this does not catch logical errors in the JavaScript code on browsers which do not implement this API. This needs to be wrapped in a try / catch block.

www/rustup.js
```diff
@@ -164,6 +164,33 @@ function fill_in_bug_report_values() {
...
+
+    setTimeout(clear_copied_text, 2000)
```
workingjubilee: For a second, I thought we were somehow deleting the copied text itself (i.e. the text we just wrote to the clipboard). It would be nice if this function used a name which indicated clearly its usage of "this function is resetting the message that the user has copied the text." I'd also have it linger slightly longer.

www/index.html
```diff
@@ -25,7 +25,17 @@
...
-  <pre>curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+    <div class="copy-container">
+      <pre>curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+      <button class="copy-button" title="Copy curl command to clipboard to download Rustup" type="button" onclick="handle_copy_button_click()">
+        <div class="copy-icon">
+          <svg width="24" height="25" viewBox="0 0 24 25" xmlns="http://www.w3.org/2000/svg" alt="Copy curl command to clipboard to download Rustup">
+            <path d="M18 20h2v3c0 1-1 2-2 2H2c-.998 0-2-1-2-2V5c0-.911.755-1.667 1.667-1.667h5A3.323 3.323 0 0110 0a3.323 3.323 0 013.333 3.333h5C19.245 3.333 20 4.09 20 5v8.333h-2V9H2v14h16v-3zM3 7h14c0-.911-.793-1.667-1.75-1.667H13.5c-.957 0-1.75-.755-1.75-1.666C11.75 2.755 10.957 2 10 2s-1.75.755-1.75 1.667c0 .911-.793 1.666-1.75 1.666H4.75C3.793 5.333 3 6.09 3 7z"></path><path d="M4 19h6v2H4zM12 11H4v2h8zM4 17h4v-2H4zM15 15v-3l-4.5 4.5L15 21v-3l8.027-.032L23 15z"></path>
+          </svg>
+        </div>
+        <div id="copied-text" class="copy-button-text"></div>
```
workingjubilee: This should have an id more like "copy-status-message" or something, I think, see also the other comment re: the later functions involving it ("text" is ambiguous).

workingjubilee (changes requested):

BeniCheni:
> Was this tested in Firefox at all @BeniCheni ?

>  Please consult https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API for further details.

@kinnison / @workingjubilee, updated to use the `Clipboard.writeText()`, and tested in FireFox / Safari / Chrome. Worked fine in my local.

> This needs to be wrapped in a try / catch block.

> It would be nice if this function used a name which indicated clearly its usage of "this function is resetting the message that the user has copied the text." I'd also have it linger slightly longer.

> This should have an id more like "copy-status-message" or something,

Thanks for the feedback, @workingjubilee. Updated to use `try` / `catch`, renamed to `clear_copy_status_message` function w. 5 second timeout to clear copy status message, and used "copy-status-message" as element ID for the copy button.

PTAL? 🙏

workingjubilee (approved):
Excellent! I have reviewed this on my local and while (due to the vagaries of systems) I recommend this get yet another set of eyes on it before it is merged (hi Kinnison) it looks good and workable. At some point the unstable Web APIs involved will settle and this will need to be redrafted but there is no telling  whether that happens ~3 days, ~3 years, or ~3 decades from now.

kinnison:
It now works for me on Firefox on Linux so let's assume it'll mostly work anywhere.  Thanks @BeniCheni and thanks @workingjubilee for helping and reviewing.

PR #2327, opened 2020-05-11
Fetch image hashes from ci-artifacts.rust-lang.org

emilyalbini:
Direct access to the S3 bucket will eventually be removed.

r? @kinnison


kinnison:
Thanks Pietro

PR #2328, opened 2020-05-12
Return early in check_help_for

Artoria2e5:
check_help_for can fail as soon as the first sign of not meeting the check happens.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -428,11 +428,9 @@ check_help_for() {
     local _arch
     local _cmd
     local _arg
-    local _ok
     _arch="$1"
     shift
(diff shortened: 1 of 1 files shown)


kinnison:
Hi @Artoria2e5 thank you for your change.

I believe that it's OK, but I'd like to understand from you why you feel this is better than the previous form?

If you have good reason to request the early return then I will merge this.

kinnison (approved):

Artoria2e5:
From my point of view, it's simply a good habit to only do what needs to be done.

kinnison:
I'll accept that reasoning.

PR #2331, opened 2020-05-14
Update some help text version numbers

Manishearth:
These don't really matter but might as well have a newer toolchain if folks are trying stuff.


kinnison (approved):
I see no reason this won't be OK, unless a test somewhere was matching on exact help text.

PR #2332, opened 2020-05-16
markdown: Fix indentation of multi-line code blocks

ghost:
Before this commit the "Welcome to Rust!" text was:

```
This path will then be added to your PATH environment variable by
modifying the profile files located at:

  /Users/alexander/.profile
/Users/alexander/.zprofile
```

Notice the second line of the code block is not indented.
This is because the `Text` is passed to us as a single
`&str` (which contains newlines). We need to split the
text and call `LineWrapper.write_indent()` for each line.

### DIFF
--- a/src/cli/markdown.rs
+++ b/src/cli/markdown.rs
@@ -72,6 +72,13 @@ impl<'a, T: Terminal + 'a> LineWrapper<'a, T> {
             self.write_word(word);
         }
     }
+    // Writes code block where each line is indented
+    fn write_code_block(&mut self, text: &str) {
+        for line in text.lines() {
+            self.write_word(line); // Will call write_indent()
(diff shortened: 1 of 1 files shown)


ghost:
This is a very small fix and probably not that important.
But I thought I'd contribute something :)

It's the first thing that macos users see when they install rust.
"Making a Great First Impression" ^^

ghost:
There's another thing: 

```
Current installation options:


   default host triple: x86_64-apple-darwin
     default toolchain: stable (default)
               profile: default
  modify PATH variable: yes
```

Notice the double empty line before the list.
Completely nitpicking here, but do you want me to fix this?

kinnison (approved):
This looks sane, thank you.

PR #2333, opened 2020-05-17
Update Microsoft C++ Build Tools url

ChrisDenton:
Microsoft now have a dedicated page for the C++ Build Tools (which even mentions Rust). It is simpler to direct users to that rather than the full list of Visual Studio 2019 downloads.


rbtcollins:
Thank you

PR #2335, opened 2020-05-17
markdown: Fix double newline before list

ghost:
Before this commit newlines were written by:
- `start_tag()` of `Tag::List`
- `start_tag()` of `Tag::Item`
- `end_tag()` of `Tag::List`

This results in two newlines before the first
list item, because both the `List` and `Item`
call `write_line()`. The simplest solution is
to not write newline in `start_tag` of `List`.

After this commit a newline is written before
each `Item` and a newline is written after the
entire `List`. No more double newlines.


kinnison:
Have you checked over other messages to make sure this doesn't make anything else look odd?

ghost:
I have not, but I'll do that.

Issue #2330, opened 2020-05-14
rustup.rs install should update old toolchains


Manishearth:
**Problem**

(h/t @mcclure for finding this rather annoying bug)

If you are on a system with an older `.rustup` folder, with or without a `.cargo`, rustup will declare success unconditionally, without updating anything. This also happens even if you request a default profile that is not installed, and, curiously, if you attempt to install rustup with a different default toolchain.

**Steps**

1. Get a rustup install on a fresh system with an old `stable`. I did this by installing rust 1.8, and then moving the toolchain and update-hashes folder. Make sure `rustup update` attempts to update, but don't let it finish.
2. Run the rustup string from https://rustup.rs with the default install. It will say `info: updating existing rustup installation`, but nothing will actually be updated.
3. Delete `.cargo`. Try again. It will recreate the `.cargo` folder, but again not update anything.
4. Try installing it with a beta or nightly install, a toolchain that is definitely not installed. It will again not install anything. It will also not update the default!

The last step can be tested on an existing rustup install pretty easily (for testing the others I had to create a docker container), by trying to install rustup with the default set as beta (if you don't have beta installed).

**Possible Solution(s)**

I feel like we should have some additional prompts for the user here.

If the user is attempting to install a default toolchain that _does_ exist in `.rustup` but is out of date, we should tell them: "You have an existing outdated toolchain installed in `~/.rustup`. Would you like to update it? y/n"

If the user is attempting to install a default toolchain that _does not_ exist in `.rustup`, we should just install it, and set it as default.

**Notes**

Output of `rustup --version`: rustup 1.21.1 (7832b2ebe 2019-12-20)
Output of `rustup show`:

<details>

```
manishearth@bb1876555c6d:~$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/manishearth/.rustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu (default)
1.8.0-x86_64-unknown-linux-gnu

active toolchain
----------------

stable-x86_64-unknown-linux-gnu (default)
rustc 1.8.0 (db2939409 2016-04-11)
```

</details>

cc @kinnison

mcclure:
For the record, here is how I encountered this bug "in the wild":

- I fiddled with Rust briefly in 2018 but didn't get beyond building someone else's code. I don't know anything about Rust.
- This week I attempted to build the wgpu-rs examples. There was a cryptic error because my cargo was too old to understand the dependencies.
[...]

PR #2339, opened 2020-05-19
Update on install

kinnison:
This is a semi-prototype, but in theory ready-for-showtime attempt at resolving #2330

Opinions from @Manishearth and @mcclure are invited.

### DIFF
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -82,6 +82,11 @@ pub fn main() -> Result<utils::ExitCode> {
                 .use_delimiter(true),
         )
         .arg(
+            Arg::with_name("no-update-default-toolchain")
+                .long("no-update-default-toolchain")
+                .help("Don't update any existing default toolchain after install"),
+        )
+        .arg(
             Arg::with_name("no-modify-path")
                 .long("no-modify-path")
                 .help("Don't configure the PATH environment variable"),
@@ -106,6 +111,7 @@ pub fn main() -> Result<utils::ExitCode> {
         .value_of("profile")
         .expect("Unreachable: Clap should supply a default");
     let no_modify_path = matches.is_present("no-modify-path");
+    let no_update_toolchain = matches.is_present("no-update-default-toolchain");
 
     let components: Vec<_> = matches
         .values_of("components")
@@ -122,6 +128,7 @@ pub fn main() -> Result<utils::ExitCode> {
         default_toolchain,
         profile: profile.to_owned(),
         no_modify_path,
+        no_update_toolchain,
         components: &components,
         targets: &targets,
     };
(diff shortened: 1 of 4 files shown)


Manishearth:
So I think _always_ prompting about this is a mistake, since all users will have to care about this.

We should either:

 - detect that there is an existing install at the beginning, and _only then_ prompt "There is an existing toolchain installed. Update it? Y/n"
 - run `rustup check` after installation and prompt if necessary

I'd prefer the former UX since then there's only one block of prompts.

rbtcollins:
I've commented on https://github.com/rust-lang/rustup/issues/2330 - I don't think prompting is ever desirable here.

kinnison:
@Manishearth How do you feel about this prototype?

Manishearth (approved):
If i'm reading this right this unconditionally does the equivalent of `rustup update <default>` on install unless otherwise specified?

Seems good to me!

kinnison:
Yep that's the intent.  Okay, @rbtcollins how about you? Are you okay with this?

mcclure:
I'm sorry, I don't know how to test this. I'm sure y'all know what's right for your project tho.

This said, I don't understand why you're trying so hard to avoid a prompt given when last I ran it, rustup.sh already shows a prompt?

Manishearth:
@mcclure the initial prototype prompted for _every_ install, even if there wasn't an existing installation to update, which means a lot of people need to go through an unnecessary prompt.

The current prototype never prompts, which is also fine IMO, it's hard to see a use case for not updating an existing install when you are installing rustup (as opposed to updating it)

src/cli/setup_mode.rs
```diff
@@ -81,6 +81,11 @@ pub fn main() -> Result<utils::ExitCode> {
                 .multiple(true)
                 .use_delimiter(true),
         )
+        .arg(
+            Arg::with_name("no-update-default-toolchain")
```
rbtcollins, line 85: I feel this is a bit verbose. We have --no-self-update; perhaps just --no-update ?

tests/cli-inst-interactive.rs
```diff
@@ -14,6 +14,12 @@ use std::io::Write;
 use std::process::Stdio;
 use std::sync::Mutex;
 
+macro_rules! for_host {
```
rbtcollins, line 17: I don't really feel that a used-once macro that just forms a function once, used in a single test is really pulling its weight - but if you feel it is going to be used again I'm cool with it being added.

rbtcollins (approved):

src/cli/setup_mode.rs
kinnison: I will have a ponder.  `--no-update` feels a little odd but the terseness may be okay.  I'll reconsider in the face of the rest of the args, it is a bit long yes.

tests/cli-inst-interactive.rs
kinnison: I'm intending, once this is merged, to file an e-easy e-mentor help-wanted to move those macros into `mock/clitools.rs`

PR #2342, opened 2020-05-23
Refactoring self_update.rs

workingjubilee:
This PR refactors out self_update.rs' contents into a submodule of two files, one for Unix-specific capability and one for Windows-specific capability, and introduces an import strategy. This was discussed on Discord, but mostly is so that it is less challenging to introduce specific shell-handling logic.


src/cli/self_update/types.rs
```diff
@@ -0,0 +1,8 @@
+#![allow(dead_code)]
+use std::path::PathBuf;
```
rbtcollins: Do you know why this triggers  a dead_code lint?

I'd rather an affirmative name for this file rather than generic 'types'; it seems like just trading one big bucket for a small bucket will grow otherwise.

e.g. 'pathupdate.rs'.

workingjubilee: 1) I really don't know, I will make another go at removing the dead_code, but this was a weirdly "entangled" type so that might be why?
2) Completely reasonable!

I am working on the code that will make that type fully redundant, which is why I was indifferent to this file being sloppy.

workingjubilee: Oh, I realize why now. Splitting out the method which generates a Vec of that type warns "variant never constructed". Still, renaming the file is easy enough.

workingjubilee: @workingjubilee Easy enough if you actually follow up and put the change in all the places that require it, Past!Jubilee.

kinnison:
@rbtcollins Are you satisfied now?

rbtcollins:
The name is fine now; I trust that the content is just moved + minimal changes to support the move? If so LGTM

workingjubilee:
Yep. More "actual" changes will be other PRs.

PR #2344, opened 2020-05-25
Handle effective limits lookups failing

rbtcollins:
QueryInformationForJobObject is failing on some Windows machines,
and we don't have fallback code in place.


kinnison (approved):
If CI is green I'm good.

PR #2351, opened 2020-05-29
Fix README install instruction for shell "sh -s"

joelparkerhenderson:


kinnison:
I am not sure why you feel `-s` is needed here.

joelparkerhenderson:
> I am not sure why you feel -s is needed here.

I'm trying the command using Alpine latest in a Dockerfile:

```
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```

Without `-s` the command fails. It seems to be due to arg issues, as far as I can tell. 

With `-s` the command succeeds. The Alpine docs for `sh` show to use `-s`.

For context, my goal is to run Rust nightly and Rocket, for a demo to teach volunteers about Rust. 

My first attempt at a Dockerfile is here: https://github.com/joelparkerhenderson/demo_rust_rocket/blob/master/Dockerfile

Any advice is very welcome.

rbtcollins:
-s is posixally correct, though most shells do read commands from stdin by default. I guess busybox is just that lean.

kinnison:
Aah interesting.  Is there anywhere else in docs or similar which may updating too?

joelparkerhenderson:
Not that I've found so far. Perhaps this merge can happen, then use it as a way to ask for someone more-skilled that I am to look for any related update areas?

And if you're able to go into wider areas... while learning last week, I found the rust-lang Rust nightly Docker builds, and tried to use them. They look official to me, yet didn't seem to work at all for me; Docker error messages showed permissions problems, and I couldn't figure out how to solve those. When I read the Rust documentation, I didn't find much info about if/how these Rust nightly Docker images are maintained or updated, or who to contact about them.

kinnison:
The dockers which `rustup`'s CI uses are only for use by the `rust-lang` projects.   They're not public.  If you've had trouble with images published on dockerhub then please report them in the proper way via the links I imagine are on dockerhub.

Thank you for your contribution.

joelparkerhenderson:
Thanks! I don't understand what you mean about the dockers...?

What I mean is I found this public page (https://hub.docker.com/r/rustlang/rust) and this public page (https://github.com/rust-lang/docker-rust-nightly) and I was trying to use the nightly tag in my own Docerkfile.

I'm new to Docker for Rust, so this is the first time I've personally seen a public page that lists tags that aren't available to the public.

PR #2354, opened 2020-06-01
In process test support

rbtcollins:
This is incremental work towards in-process tests.

### DIFF
--- a/src/bin/rustup-init.rs
+++ b/src/bin/rustup-init.rs
@@ -14,7 +14,6 @@
 
 #![recursion_limit = "1024"]
 
-use std::env;
 use std::path::PathBuf;
 
 use rs_tracing::*;
@@ -25,41 +24,46 @@ use rustup::cli::proxy_mode;
 use rustup::cli::rustup_mode;
 use rustup::cli::self_update;
 use rustup::cli::setup_mode;
+use rustup::currentprocess::{process, with, OSProcess};
 use rustup::env_var::RUST_RECURSION_COUNT_MAX;
 use rustup::utils::utils;
 
 fn main() {
-    if let Err(ref e) = run_rustup() {
-        common::report_error(e);
-        std::process::exit(1);
-    }
+    let process = OSProcess::default();
+    with(Box::new(process), || match run_rustup() {
+        Err(ref e) => {
+            common::report_error(e);
+            std::process::exit(1);
+        }
+        Ok(utils::ExitCode(c)) => std::process::exit(c),
+    });
 }
 
-fn run_rustup() -> Result<()> {
-    if let Ok(dir) = env::var("RUSTUP_TRACE_DIR") {
+fn run_rustup() -> Result<utils::ExitCode> {
+    if let Ok(dir) = process().var("RUSTUP_TRACE_DIR") {
         open_trace_file!(dir)?;
     }
     let result = run_rustup_inner();
-    if env::var("RUSTUP_TRACE_DIR").is_ok() {
+    if process().var("RUSTUP_TRACE_DIR").is_ok() {
         close_trace_file!();
     }
     result
 }
 
-fn run_rustup_inner() -> Result<()> {
+fn run_rustup_inner() -> Result<utils::ExitCode> {
     // Guard against infinite proxy recursion. This mostly happens due to
     // bugs in rustup.
     do_recursion_guard()?;
 
     // Before we do anything else, ensure we know where we are and who we
     // are because otherwise we cannot proceed usefully.
-    utils::current_dir()?;
+    process().current_dir()?;
     utils::current_exe()?;
 
     // The name of arg0 determines how the program is going to behave
-    let arg0 = match env::var("RUSTUP_FORCE_ARG0") {
+    let arg0 = match process().var("RUSTUP_FORCE_ARG0") {
         Ok(v) => Some(v),
-        Err(_) => env::args().next(),
+        Err(_) => process().args().next(),
     }
     .map(PathBuf::from);
     let name = arg0
@@ -90,7 +94,8 @@ fn run_rustup_inner() -> Result<()> {
 }
 
 fn do_recursion_guard() -> Result<()> {
-    let recursion_count = env::var("RUST_RECURSION_COUNT")
+    let recursion_count = process()
+        .var("RUST_RECURSION_COUNT")
         .ok()
         .and_then(|s| s.parse().ok())
         .unwrap_or(0);
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -14,6 +14,7 @@ use super::errors::*;
 use super::self_update;
 use super::term2;
 use crate::dist::notifications as dist_notifications;
+use crate::process;
 use crate::toolchain::DistributableToolchain;
 use crate::utils::notifications as util_notifications;
 use crate::utils::notify::NotificationLevel;
@@ -23,7 +24,7 @@ use crate::{Cfg, Notification, Toolchain, UpdateStatus};
 pub const WARN_COMPLETE_PROFILE: &str = "downloading with complete profile isn't recommended unless you are a developer of the rust language";
 
 pub fn confirm(question: &str, default: bool) -> Result<bool> {
-    print!("{} ", question);
+    write!(process().stdout(), "{} ", question)?;
     let _ = std::io::stdout().flush();
     let input = read_line()?;
 
@@ -34,7 +35,7 @@ pub fn confirm(question: &str, default: bool) -> Result<bool> {
         _ => false,
     };
 
-    println!();
+    writeln!(process().stdout())?;
 
     Ok(r)
 }
@@ -46,11 +47,11 @@ pub enum Confirm {
 }
 
 pub fn confirm_advanced() -> Result<Confirm> {
-    println!();
-    println!("1) Proceed with installation (default)");
-    println!("2) Customize installation");
-    println!("3) Cancel installation");
-    print!(">");
+    writeln!(process().stdout())?;
+    writeln!(process().stdout(), "1) Proceed with installation (default)")?;
+    writeln!(process().stdout(), "2) Customize installation")?;
+    writeln!(process().stdout(), "3) Cancel installation")?;
+    write!(process().stdout(), ">")?;
 
     let _ = std::io::stdout().flush();
     let input = read_line()?;
@@ -61,17 +62,17 @@ pub fn confirm_advanced() -> Result<Confirm> {
         _ => Confirm::No,
     };
 
-    println!();
+    writeln!(process().stdout())?;
 
     Ok(r)
 }
 
 pub fn question_str(question: &str, default: &str) -> Result<String> {
-    println!("{}", question);
+    writeln!(process().stdout(), "{}", question)?;
     let _ = std::io::stdout().flush();
     let input = read_line()?;
 
-    println!();
+    writeln!(process().stdout())?;
 
     if input.is_empty() {
         Ok(default.to_string())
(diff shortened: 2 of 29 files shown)


rbtcollins:
so, here's the basic idea
read CONTRIBUTING
read the two new traits that are in their own files
the vast bulk is basically just adding <V: VarSource> .. (vars:&V,
to signatures
or to structs
finally changed DownloadCfg  to be Clone based rather than Copy.

I'm not 100% sure that the way I've done this is right.

For instance, we could have just one object that implements multiple traits; and we could have multiple traits combined into one aggregate to make it less noisy in code...

e.g. ProcessInterface: ArgsSource+VarSource + (process:&P...

which we could then break apart for when a narrow interface really is appropriate.

I'm happy enough to do a refactor like that if you like the sounds of it.

download/src/lib.rs
```diff
@@ -7,7 +7,7 @@ use url::Url;
 mod errors;
 pub use crate::errors::*;
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Clone)]
```
kinnison: I'm not sure why you relaxed `Copy` here given it's a trivial enumeration.

src/cli/argsource.rs
```diff
@@ -0,0 +1,54 @@
+/// Abstracts over reading the current process environment variables as a
+/// zero-cost abstraction to support threaded in-process testing.
+use std::env;
+
+pub trait ArgSource {
+    type Args: Iterator<Item = String>;
```
kinnison: Do we really treat environment variables as always being `String` rather than `OsString`?

src/cli/argsource.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+/// Helper for ArgSource over `Vec<String>`
+pub struct VecArgs {
+    v: Vec<String>,
+    i: usize,
+}
+
+impl From<&Vec<String>> for VecArgs {
```
kinnison: Would it be okay to have this as `From<&[String]>` and to have a complementary `From<&[&str]>` perhaps?  I tried to persuade an `impl<S: AsRef<str>> From<&[S]> for Foo` to exist, but it started whining that `&[&str; 2]` wasn't compatible which made me a bit sad.

src/cli/argsource.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn from(source: &Vec<String>) -> Self {
+        let v = source.clone();
+        VecArgs { v, i: 0 }
+    }
+}
+
+impl Iterator for VecArgs {
```
kinnison: I'd have been tempted to just set the iterator type for VecArgs to `Box<dyn Iterator<Item=String>>` and then return `Box::new(self.v.iter().cloned())` fom the `fn args()` but this is okay.

src/cli/argsource.rs
```diff
@@ -0,0 +1,54 @@
...
+    type Item = String;
+    fn next(&mut self) -> Option<String> {
+        if self.i == self.v.len() {
+            return None;
```
kinnison: If you want to retain this implementation then it'd be right to also `impl FusedIterator` for this because you are fused.  Not that we rely on this but I think it does allow theoretically some optimisation.

src/cli/common.rs
```diff
@@ -147,7 +148,10 @@ impl NotifyOnConsole {
     }
 }
 
-pub fn set_globals(verbose: bool, quiet: bool) -> Result<Cfg> {
+pub fn set_globals<V>(vars: &V, verbose: bool, quiet: bool) -> Result<Cfg<V>>
+where
+    V: VarSource + Clone,
```
kinnison: Given how often you say `VarSource + Clone` would it not be easier to just supertrait it?  If we ever need to make a non-Clone VarSource we could revisit it.

src/cli/common.rs
```diff
@@ -157,24 +161,27 @@ pub fn set_globals(verbose: bool, quiet: bool) -> Result<Cfg> {
...
+                return;
+            }
+            console_notifier.borrow_mut().handle(n);
+        }),
+    )?)
 }
 
-pub fn show_channel_update(
-    cfg: &Cfg,
+pub fn show_channel_update<V: VarSource + Clone>(
```
kinnison: We could get rid of a lot of this noise by boxing the VarSource in the `Cfg` instance rather than parameterising it on it.

kinnison: Also, thinking about it, that'd be faster to build too, and not be enough of a performance issue to be worth considering against IMO.

kinnison: Then when you need to pass the var source to other things, make it take a `&dyn VarSource` instead?

src/varsource.rs
```diff
@@ -0,0 +1,38 @@
+/// Abstracts over reading the current process environment variables as a
+/// zero-cost abstraction to support threaded in-process testing.
+use std::collections::hash_map::HashMap;
+use std::env;
+use std::ffi::{OsStr, OsString};
+
+pub trait VarSource {
```
kinnison: I'm half tempted to suggest we do "typesafe environment variables" and gather all environment variable retrieval into one place, by putting a trait `RustupVars` which is implemented for any T which implements `VarSource` and then pass `&dyn RustupVars` around rather than `VarSource` -- perhaps that's a refactor for another time though.

kinnison (commented):
Initial thoughts - I wouldn't **block** on them, but if you think there's merit then it might reduce the size of the diff somewhat.

download/src/lib.rs
rbtcollins: Over-keen I think, I was trying to provoke closer-to-source errors on the trait. Happy to back it out; OTOH we don't need it anywhere either now, if we ever did.

src/cli/argsource.rs
rbtcollins: .collect() on those should do the right thing for us already; this type is *much* easier as an owned type. I mean yes - but in my conversions in run() in the -wip branch, it isn't an issue.

src/cli/argsource.rs
rbtcollins: I tried things in that space. std::slice::Iterator<String> returns &String. Hmm, I can go back and noodle on that, or we can do as another pass to make it tighter. I think the contract is the key thing.

src/cli/argsource.rs
rbtcollins: it allows optimising .fuse(), so sure, we can; not an issue in practice I think

src/cli/common.rs
rbtcollins: By this point I was doing large scale grep-replace and not thinking :) - I only found out about the Clone need when I had the entire patch in place and was substituting in the HashMap layer in the WIP branch and found it was missing. Sure. I think perhaps the refactoring I refer to in the conversation is perhaps a bigger better rearrangement.

src/varsource.rs
rbtcollins: not sure how that looks different

src/cli/argsource.rs
rbtcollins: hmm, I missed an args_os call in proxy-mode

src/cli/common.rs
rbtcollins: thats a possibility too, yes.

src/varsource.rs
kinnison: The idea would be that anywhere using `RustupVars` would call things like `.ram_limit()` and such, but the more I think about it the less I think it's needed right now since we don't tend to read the same variable from more than one place.

src/cli/argsource.rs
rbtcollins: fixed

src/bin/rustup-init.rs
```diff
@@ -25,41 +24,46 @@ use rustup::cli::proxy_mode;
...
+use rustup::currentprocess::{process, with, OSProcess};
 use rustup::env_var::RUST_RECURSION_COUNT_MAX;
 use rustup::utils::utils;
 
 fn main() {
-    if let Err(ref e) = run_rustup() {
-        common::report_error(e);
-        std::process::exit(1);
-    }
+    let process = OSProcess {};
```
kinnison: I'd prefer to see this implement `Default` or else have a trivial `new` method, just because this looks ugly.

src/bin/rustup-init.rs
```diff
@@ -25,41 +24,46 @@ use rustup::cli::proxy_mode;
...
+    with(Box::new(process), || match run_rustup() {
+        Err(ref e) => {
+            common::report_error(e);
+            std::process::exit(1);
+        }
+        Ok(utils::ExitCode(c)) => std::process::exit(c),
```
kinnison, line 38: Should this be `std::process::exit()` or should it really be `process.exit()` with that abstracted through the trait?

src/currentprocess.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Obtain the current instance of CurrentProcess
+pub fn process() -> Box<dyn CurrentProcess> {
+    match PROCESS.with(|p| p.borrow().clone()) {
+        None => panic!("No process instance"),
+        Some(p) => p,
+    }
+}
+
+lazy_static! {
+    static ref HOOK_INSTALLED: Mutex<bool> = Mutex::default();
```
kinnison: `Mutex<bool>` can often be replaced with an `AtomicBool` instance -- does this need the `Mutex` behaviour?

src/currentprocess.rs
```diff
@@ -0,0 +1,228 @@
...
+        let mut installed = HOOK_INSTALLED.lock().unwrap();
+        if *installed == false {
+            let orig_hook = panic::take_hook();
+            panic::set_hook(Box::new(move |info| {
+                clear_process();
+                orig_hook(info);
+            }));
+            *installed = true
+        }
+    }
```
kinnison: This whole thing could be a `Once` perhaps?

kinnison (commented):
Looks fairly good, a few comments to deal with though.

src/bin/rustup-init.rs
rbtcollins: sure, though for something used just once - shrug :). I will add.

src/bin/rustup-init.rs
rbtcollins: Because exit's occur deep in call stacks, process() can't have an exit() method unless we implement that as panic() - which is ugly, or a macro equivalent try! - which isn't a method - or we get into compiler plugins - which is opaque to readers.

We could downcast errors to extract ExitCode's from errors that have them, but errors can still be not propogated, and downcasting with error-chain is going to just make the migration from that more complicated I feel.

We could treat all errors as 'bad thing exit 1' and Ok(explicit error code) as a form of success, even if that means exiting non-zero e.g. for the proxy mode. Which is what I chose

Thus exit() is implemented as structured return: Err(badthing) for error class returns, and Ok(ExitCode(n)) for deliberate non-error returns where that is needed, which is here.

src/currentprocess.rs
rbtcollins: AtomicBool would permit a data race. set_hook is not a CAS operation, and orig_hook is global state.

src/bin/rustup-init.rs
kinnison: Fair enough.  I suppose we just don't do in-process testing of this part of `rustup-init` and we'll be fine. :+1:

kinnison (approved):

src/bin/rustup-init.rs
rbtcollins: Well we can't - its not in the library (it is in bin/ :) ). We could move some of this into the lib, but actually I think this is sufficiently shallow that bugs are remote.

PR #2359, opened 2020-06-02
CI: Check riscv64gc-unknown-linux-gnu for stable releases

tblah:
`riscv64gc-unknown-linux-gnu` is now tier 2
    https://github.com/ehuss/rust-forge/commit/c1ae62fbb3d91af8fc6a55ebb175309d79b091f2

~TODO: b66bc057ac18b0f0e641dc02f604fc742610c249 is only here to verify the CI passes~

### DIFF
--- a/.github/workflows/linux-builds-on-stable.yaml
+++ b/.github/workflows/linux-builds-on-stable.yaml
@@ -38,6 +38,7 @@ jobs:
           - armv7-linux-androideabi       # skip-pr skip-master
           - i686-linux-android            # skip-pr skip-master
           - x86_64-linux-android          # skip-pr skip-master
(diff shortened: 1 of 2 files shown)


tblah:
CI passed so I have dropped b66bc057ac18b0f0e641dc02f604fc742610c249

Log is here https://github.com/rust-lang/rustup/pull/2359/checks?check_run_id=731725929

kinnison:
Unfortunately that link takes me to the 1b54604 check because github actions don't retain from run to run on the same PR afaict.  I'm willing to trust you but will say that if I have issues during a release test I may come back to complain :D

kinnison (approved):
Normally I'd ask for the templates and the workflow updates to be in separate commits, but this is small enough I won't be precious about it.

PR #2367, opened 2020-06-06
In process test

rbtcollins:
Run rustup tests in-process

Using the in-process test support feature, we can now change the test runner to run select CLI operations in-process.

More can be done, but this is the first major step: running rustup-mode in-process much of the time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ error-chain = "0.12"
 effective-limits = "0.5"
 flate2 = "1"
 git-testament = "0.1.4"
-home = "0.5"
+home = { git = "https://github.com/rbtcollins/home", rev="81c8dfcd815765daf384319a04e237feb9d5e72d"  }
 lazy_static = "1"
 libc = "0.2"
 num_cpus = "1.13"
@@ -113,6 +113,3 @@ codegen-units = 1
 # Reduce build time by setting proc-macro crates non optimized.
 [profile.release.build-override]
 opt-level = 0
-
-[patch.crates-io]
-home = { path = "../rustup-home" }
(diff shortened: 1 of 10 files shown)


kinnison:
In its current state, this PR blows my `~/.cargo` and `~/.rustup` away

rbtcollins:
success!

Cargo.toml
```diff
@@ -28,7 +28,7 @@ error-chain = "0.12"
 effective-limits = "0.5"
 flate2 = "1"
 git-testament = "0.1.4"
-home = "0.5"
+home = { git = "https://github.com/rbtcollins/home"  }
```
kinnison: Could we please have a SHA on this, so that if I decide to `cargo update` I don't have to fear messing up?

src/cli/rustup_mode.rs
```diff
@@ -55,7 +55,16 @@ where
...
+    let matches = match cli().get_matches_from_safe(process().args_os()) {
+        Ok(matches) => Ok(matches),
+        Err(e)
+            if e.kind == clap::ErrorKind::HelpDisplayed
+                || e.kind == clap::ErrorKind::VersionDisplayed =>
+        {
+            return Ok(utils::ExitCode(0))
+        }
+        Err(e) => Err(e),
+    }?;
```
kinnison, line 67: It might be nice to have this factored into a function given it's also used in setup_mode.rs and someone fixing something in one might miss the other.

kinnison (changes requested):
Two minor niggles and then I think this is mergeable.  What sort of test time improvement do we see?

Cargo.toml
rbtcollins: You'd know if you did, when your rustup install went away :P.

kinnison (approved):
Okay, I guess you didn't think the refactor worthwhile, fair enough.

rbtcollins:
> 
> 
> Okay, I guess you didn't think the refactor worthwhile, fair enough.

I think its marginal - there's really only one possible spelling here, and it's dictated by clap; and we have full test coverage. I'm happy to swing back and add it in when I get some more time.

PR #2368, opened 2020-06-08
Add a hint to log out/in after installation to the README

### LINKED ISSUE #2317: Add a hint to log out/in after installation to the README

As discussed in #2226, on various flavours of Linux you need to do a log out/in cycle for the updates to `PATH` to take without having to re-source `.profile` in each terminal. However, the README simply states:

```
This directory will be in your $PATH environment variable, which means you can run them from the shell without further configuration. Open a new shell and type the following:
```

This makes it appear that some thing went wrong with your installation if (like me) you didn't pause to read the end of the installation output properly.

A simple reference to needing to log out/in on Linux would be friendly I think. Even as a footnote or a FAQ.

kellda:
Try to fixes #2317 

If you have a better wording feel free to tell me

### DIFF
--- a/README.md
+++ b/README.md
@@ -840,7 +840,7 @@ closing your editor) and try again.
 ### rustup exited sucessfully but I can't run `rustc --version`
 
 To get started you need Cargo's bin directory ({cargo_home}/bin) in your `PATH`
-environment variable. This should be done by rustup. Next time you log in this will be done
+environment variable. This should be done by rustup. Next time you log in this should have been done
 automatically.
 
 ## License
--- a/README.md
+++ b/README.md
@@ -837,6 +837,12 @@ If `rustup` fails to self-update in this way it's usually because RLS is
 running (your editor is open and running RLS). The solution is to stop RLS (by
 closing your editor) and try again.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -837,6 +837,12 @@ If `rustup` fails to self-update in this way it's usually because RLS is
 running (your editor is open and running RLS). The solution is to stop RLS (by
 closing your editor) and try again.
 
+### rustup exited sucessfully but I can't run `rustc --version`
+
+To get started you need Cargo's bin directory ({cargo_home}/bin) in your `PATH`
+environment variable. This should be done by rustup. Next time you log in this will be done
```
kinnison:
```suggestion
environment variable. This should be done by rustup. Next time you log in this should have been done
```

Just in case it doesn't work, let's not assert it *will* but rather than it *should* happen :D

kellda:
Yes, you should not need to log in / out, but if it doesn't works, loging out / in may help...

In fact I just copy-pasted the rustup message given [here](https://github.com/rust-lang/rustup/issues/2317#issuecomment-625741498).

PR #2369, opened 2020-06-08
Document the encoding of rust-toolchain

### LINKED ISSUE #2302: Rustup ignores 'rust-toolchain' when saved in UTF-16 on Windows.
Rustup ignores `rust-toolchain` when saved in UTF-16 on Windows.

To reporoduce:
1. Create `rust-toolchain` with `nigthly` and save as UTF-16 (BE/LE, doesn't matter)
2. Run `rustup show`

Active toolchain will be stable.

Saving the same file in UTF-8 solves the problem

Output of `rustup --version`:
`rustup 1.21.1 (7832b2ebe 2019-12-20)`

Output of `rustup show`:
```
rustup home:  C:\Users\xxx\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc

active toolchain
----------------

nightly-x86_64-pc-windows-msvc (overridden by '\\?\C:\Dev\eezip\rust-toolchain')
rustc 1.44.0-nightly (3360cc3a0 2020-04-24)
```

kellda:
Fixes #2302 

Maybe it's also useful to throw a warning when there is an invalid `rust-toolchain` file

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,7 +406,8 @@ software that pins to a revision from the release archives.
 In these cases the toolchain can be named in the project's directory
 in a file called `rust-toolchain`, the content of which is the name of
 a single `rustup` toolchain, and which is suitable to check in to
-source control.
+source control. This file has to be encoded in US-ASCII (if you are on
+Windows, check the encoding and that it does not starts with a BOM).
 
 The toolchains named in this file have a more restricted form than
 `rustup` toolchains generally, and may only contain the names of the


kinnison:
Since file encoding for `rust-toolchain` is likely only an issue on Windows, where applications have a habit of inserting byte-order-markers for UTF-8 documents (which is madness) it might be sensible to explicitly state that it should be ASCII rather than permitting the mention of UTF-8.

kinnison:
Also regarding validating the content, we are in discussions about improving the handling of `rust-toolchain` so I'll make a note to think about that then too.

kellda:
Is this better worded ? Or should I move the Windows reference to the FAQ / remove it

kinnison (approved):
This wording is acceptable to me.  While there's no real point, I will wait for green CI.

kellda:
> While there's no real point, I will wait for green CI

Did the CI failed just to prove there's a point ?

PR #2371, opened 2020-06-09
Minor wording improvements in 'echo' output

perlun:
It feels to me (as a non-native English speaker) that `enforcing` would sound slightly better in this context than the current `forcing` wording.


kinnison:
The manual page for curl at least says _force_ but I guess perhaps this wording might be better.  I'll have a think, sorry for the delay.

kinnison (approved):

PR #2372, opened 2020-06-10
Disable threading when on single core hardware

### LINKED ISSUE #2365: Unable to install rustc on Raspberry PI W due to low memory
I just tried to update to Rust 1.44.0 on my Raspberry pi zero W (512 MiB of memory). I was previously running 1.43.1 without problems. But during the upgrade I see:
```
 52.1 MiB /  69.1 MiB ( 75 %)   1.2 MiB/s in  1m 14s ETA: 14smemory allocation of 148512752 bytes failedAborted
```

I don't even run X on this thing, so before starting the install/upgrade I only use 36 MiB of my memory. I can't really free up much more than that.

`rustup install 1.43.1` works fine. So it must be 1.44 that passes some magical limit.

**Possible Solution(s)**
Don't store the entire tarball in memory (if that is what actually happens) stream them directly to disk while downloading.

**Notes**

Output of `rustup --version`:
```
```
Output of `rustup show`:
```
Default host: arm-unknown-linux-gnueabihf
rustup home:  /home/pi/.rustup

installed toolchains
--------------------

stable-arm-unknown-linux-gnueabihf (default)
nightly-arm-unknown-linux-gnueabihf
1.43.1-arm-unknown-linux-gnueabihf

active toolchain
----------------

stable-arm-unknown-linux-gnueabihf (default)
(rustc does not exist)
```

faern:
An improvement that will (at least temporarily) fix #2365.

No need to run a thread pool if the number of available CPUs/cores is one anyway. So it can fall back to the special single threaded mode in that case. The single threaded mode uses less memory so it's really useful on low end machines, such as the Raspberry Pi zero and similar.

Since setting the `RUSTUP_IO_THREADS` environment variable to zero on one now produces the same result as `"disabled"` did previously the `"disabled"` value does not need to exist any longer, and the code can be simpler if it just needs to be an integer directly. Since the variable was regarded as unstable in the docs I thought it was fine to change how it behaved.

This new code also prints a warning on stderr if `RUSTUP_IO_THREADS` is set to an invalid value, instead of just silently falling back to automatically detecting the number of cores.

I'm still working on trying to get this tested locally. But the Rpi zero is *incredibly* slow at compiling such a big project 😅 At least it builds. So I figured I could submit the PR and get some feedback.

### DIFF
--- a/src/diskio/mod.rs
+++ b/src/diskio/mod.rs
@@ -59,6 +59,7 @@ use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::time::{Duration, Instant};
 
+use crate::errors::{Result, ResultExt};
 use crate::process;
 use crate::utils::notifications::Notification;
 
@@ -194,20 +195,16 @@ pub fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
 /// Get the executor for disk IO.
 pub fn get_executor<'a>(
     notify_handler: Option<&'a dyn Fn(Notification<'_>)>,
-) -> Box<dyn Executor + 'a> {
+) -> Result<Box<dyn Executor + 'a>> {
     // If this gets lots of use, consider exposing via the config file.
-    if let Ok(thread_str) = process().var("RUSTUP_IO_THREADS") {
-        if thread_str == "disabled" {
-            Box::new(immediate::ImmediateUnpacker::new())
-        } else if let Ok(thread_count) = thread_str.parse::<usize>() {
-            Box::new(threaded::Threaded::new_with_threads(
-                notify_handler,
-                thread_count,
-            ))
-        } else {
-            Box::new(threaded::Threaded::new(notify_handler))
-        }
-    } else {
-        Box::new(threaded::Threaded::new(notify_handler))
-    }
+    let thread_count = match process().var("RUSTUP_IO_THREADS") {
+        Err(_) => num_cpus::get(),
+        Ok(n) => n
+            .parse::<usize>()
+            .chain_err(|| "invalid value in RUSTUP_IO_THREADS. Must be a natural number")?,
+    };
+    Ok(match thread_count {
+        0 | 1 => Box::new(immediate::ImmediateUnpacker::new()),
+        n => Box::new(threaded::Threaded::new(notify_handler, n)),
+    })
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1475,6 +1475,7 @@ dependencies = [
  "home",
  "lazy_static",
  "libc",
+ "num_cpus",
  "opener",
  "openssl",
  "pgp",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ git-testament = "0.1.4"
 home = "0.5"
 lazy_static = "1"
 libc = "0.2"
+num_cpus = "1.13"
 opener = "0.4.0"
 # Used by `curl` or `reqwest` backend although it isn't imported
 # by our rustup.
--- a/README.md
+++ b/README.md
@@ -697,7 +697,7 @@ currently can define only `default_toolchain`.
   Sets the root URL for downloading self-updates.
 
 - `RUSTUP_IO_THREADS` *unstable* (defaults to reported cpu count). Sets the
-  number of threads to perform close IO in. Set to `disabled` to force
+  number of threads to perform close IO in. Set to `1` to force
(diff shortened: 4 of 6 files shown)


faern:
Seems to work as intended on a machine with many cores. So testing successful. But I have not tried on the Raspberry pi yet to see if it actually solves #2365 without having to set custom variables.

src/diskio/mod.rs
```diff
@@ -196,18 +196,19 @@ pub fn get_executor<'a>(
...
-            ))
-        } else {
-            Box::new(threaded::Threaded::new(notify_handler))
+    let thread_count = match process()
+        .var("RUSTUP_IO_THREADS")
+        .map(|x| x.parse::<usize>())
+    {
+        Ok(Ok(n)) => n,
+        Ok(Err(_)) => {
+            eprintln!("Invalid value for RUSTUP_IO_THREADS. Has to be an integer");
```
rbtcollins: Please don't use eprintln! - either use chain_err or a notify callback message. Also it doesn't have to be an integer - it has to be a natural number.

rbtcollins (changes requested):

src/diskio/mod.rs
faern: If I use `chain_err` that means you want me to return an error instead of falling back to `num_cpus::get()`? Sure. I can make the method fallible. Going to be some more work, but if you prefer that.

Yeah I know about the natural numbers. I just thought this sounded better even though it's not technically correct. People must understand that negative numbers are not OK :D But I'll change it.

rbtcollins:
I suggest cross-compiling for the pi, it will be a lot faster, even with the time to get setup :).

faern:
Can you show me a guide for getting that set up on a macOS machine and I'd be super happy.

src/diskio/mod.rs
faern: Done

rbtcollins:
> 
> 
> Can you show me a guide for getting that set up on a macOS machine and I'd be super happy.

https://dev.to/jeikabu/raspberry-pi-zero-raspbian-rust-primer-3aj6

rbtcollins:
@kinnison seeking your thoughts: this is backwards incompatible with the disabled value, but RUSTUP_IO_THREADS is marked*unstable*, so I'm ok taking the heat if folk do find they have scripts they have to update: depending on unstable interfaces should be done by  version locking the dependencies anyway.

rbtcollins (commented):
I'm ok with this - thank you.

There's various bits of code golf that could be done differently but nothing consequential enough to be worth asking for it to be changed. (e.g. chain_err can take a string, for instance).

faern:
Ok. I know `chain_err` can take a string. But you had so insanely many `ErrorKind` variants so I assumed the preferred style was to have explicit error kinds for everything. I did not spend that much time researching all existing `chain_err` instances. I can easily change it if you prefer an "opaque" string error.

faern:
> There's various bits of code golf that could be done differently

Feel free to tell me your opinions. I do care deeply about code cleanliness and I thought this was the nicest way of writing this. Separating getting the value out and actually creating the executor etc.

rbtcollins:
Ok then :) - I try to balance obsession with contributor time :). And Rustup is a quite old codebase with lots of hair on it :). I think taking a string is generally good when an error comes from just one location, making an explicit enum entry for reuse is only worth it when well, we need to reuse it. Or for matching, which is very rare.

The Ok here could be factored to the outside of the expression:

```
 Ok(match thread_count {
        0 | 1 => Box::new(immediate::ImmediateUnpacker::new()),
        n => Box::new(threaded::Threaded::new(notify_handler, n)),
    })
```

Generally with error-chain the pattern is to import @ from errors : 

`use crate::errors::{...}` -> `use crate::errors::*` - its one of the reasons we're trying to move away from it :).

rbtcollins:
Anyhow, this is ready to merge - I've chatted with @kinnison  about the env var change on discord. If you could rebase your changes to remove the exploratory changes, I'll merge this. Thanks!

faern:
> The Ok here could be factored to the outside of the expression

Yeah.. I did not try that since it usually leads to not so nice formatting. But in this case it actually did not add an extra level of indentation so it's fine! I changed it :)

> use crate::errors::{...} -> use crate::errors::* - its one of the reasons we're trying to move away from it :).

I'm not a fan of wildcard imports. Makes it super hard to know what comes from where. And yeah, `error-chain` is pretty ancient at this point :/

I also made the error into just a string, as you wished. Now it should be ready for merge.

rbtcollins:
Thank yoiu

PR #2375, opened 2020-06-12
Add `--print-file` option to rustup-init.sh

tonytheodore:
This is useful when the download and install steps need to be separated.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -27,6 +27,7 @@ FLAGS:
     -q, --quiet             Disable progress output
     -y                      Disable confirmation prompt.
         --no-modify-path    Don't configure the PATH environment variable
+        --print-file        Prints detected arch and file extension
     -h, --help              Prints help information
     -V, --version           Prints version information
 
@@ -83,6 +84,10 @@ main() {
         case "$arg" in
             -h|--help)
                 usage
+                exit 0
+                ;;
+            --print-file)
+                printf '%s\n' "${_arch}/rustup-init${_ext}" 1>&2
                 exit 0
                 ;;
             -y)


tonytheodore:
The [` Run a full build and test` is failing](https://github.com/rust-lang/rustup/runs/764193111?check_suite_focus=true#step:14:312):
```
error: test failed, to rerun pass '--test dist'
Caused by:
  process didn't exit successfully: `/Users/runner/runners/2.263.0/work/rustup/rustup/target/x86_64-apple-darwin/release/deps/dist-e1ce6058a21b3ead --test-threads 1` (signal: 11, SIGSEGV: invalid memory reference)
test add_extension_not_in_manifest ... 
##[error]Process completed with exit code 101.
```

though I'm not sure how it's related.

kinnison:
Unfortunately the help for `rustup-init.sh` is actually the generated help for the `rustup-init` binary so this patch won't be reliable.  What is your goal with this?  It doesn't display the URL to the file it would download.  If you'd prefer that `rustup-init` (and thus `rustup-init.sh`) supported printing the computed "rust" arch triplet, then we should add that both to the binary and to the script, so that the help remains in sync.

tonytheodore:
Indeed, `--print-arch` would be perfect. I initially thought the extension would be useful, but it’s better to return a single value (and we don’t need the extension).

I’m packaging the rust binaries in a [cross-compiler](https://mxe.cc) and we disable downloads during the build phase. We have our own download/checksum logic and I need the same computed arch in both phases. We also have our own mirrors and need to pin versions - just knowing the arch is ideal.

tonytheodore:
> If you'd prefer that rustup-init (and thus rustup-init.sh) supported printing the computed "rust" arch triplet, then we should add that both to the binary and to the script, so that the help remains in sync.

Hmm, a complete option would specify `conflicts_with` https://github.com/rust-lang/rustup/blob/master/src/cli/setup_mode.rs, 

```
        .arg(
            Arg::with_name("quiet")
                .conflicts_with("verbose")
                .short("q")
                .long("quiet")
                .help("Disable progress output"),
```
I’m guessing the `—print-arch` could conflict with anything that does any action?

kinnison:
I think it's okay if it doesn't conflict any, since it's only giving information.  if `rustup-init` had subcommands then it'd be fairly easy, but oh well.

PR #2378, opened 2020-06-16
Fixed name of rust-analyzer IDE tool

trevarj:
Baby contribution :baby: :baby_bottle: that I found while reading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ recommended for general use.
 * The **complete** profile includes all the components available through
 `rustup`. This should never be used, as it includes *every* component ever included
 in the metadata and thus will almost always fail. If you are looking for a way
-to install devtools such as `miri` or IDE integration tools (`rls`, `rust-analysis`),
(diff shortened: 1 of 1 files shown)


kinnison (approved):
Definitely correct.

jonas-schievink:
Wasn't this actually correct? `rust-analysis` is the `-Zsave-analysis` data of the standard library, used by RLS. rust-analyzer is not yet shipped via rustup.

kinnison:
@jonas-schievink Yes in one sense: `rust-analysis` is a real component but it's also not something that people *tend* to manually install since most rls IDE integrations automatically install it because there's no inter-component dependency information in the channels so installing `rls` doesn't auto-install `rust-analysis` -- In the future, we hope to be distributing `rust-analyzer` as a component, so this is sort-of future-proofing the README.  It reads slightly better (IMO) like this.  Ultimately though this is just a README line so it's just about hinting to the user about the `complete` profile not being what you want.

PR #2384, opened 2020-06-22
Describe how to remove components when update fails

### LINKED ISSUE #2355: Describe how to remove components when update fails

When upgrading to a Rust version missing currently installed components, we get the error message including the text

> If you require these components

This has great text for what to do in this case, but we don't discuss the other case!

(Filed due to a Stack Overflow question: [Unable to update stable rustup toolchain: some components unavailable for target i386-apple-ios, armv7s-apple-ios](https://stackoverflow.com/q/62139919/155423))

It would be good to describe how to _remove_ the components if we **don't** need them anymore (ideally copy-pastable).

Something like:

```
rustup component remove --toolchain $TC --target $TG $NAME
```

For each conflicting component.

dmarcoux:
Closes  #2355

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -395,16 +395,26 @@ fn valid_profile_names() -> String {
 }
 
 fn remove_component_msg(cs: &Component, manifest: &Manifest, toolchain: &str) -> String {
-    format!(
-        "rustup component remove --toolchain {}{} {}",
-        toolchain,
-        if let Some(target) = cs.target.as_ref() {
-            format!(" --target {}", target)
-        } else {
-            String::default()
-        },
-        cs.short_name(manifest)
-    )
+    if cs.short_name_in_manifest() == "rust-std" {
+        // We special-case rust-std as it's the stdlib so really you want to do
+        // rustup target remove
+        format!(
+            "    rustup target remove --toolchain {} {}",
+            toolchain,
+            cs.target.as_deref().unwrap_or(toolchain)
+        )
+    } else {
+        format!(
+            "    rustup component remove --toolchain {}{} {}",
+            toolchain,
+            if let Some(target) = cs.target.as_ref() {
+                format!(" --target {}", target)
+            } else {
+                String::default()
+            },
+            cs.short_name(manifest)
+        )
+    }
 }
 
 fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &str) -> String {
@@ -427,7 +437,7 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
 
         let _ = write!(
             buf,
-            "If you don't need the component, you can remove it with:\n{}",
+            "If you don't need the component, you can remove it with:\n\n{}",
             remove_component_msg(&cs[0], manifest, toolchain)
         );
     } else {
@@ -448,7 +458,7 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
             let _ = write!(
                 buf,
                 "some components unavailable for download for channel {}: {}
-                 If you don't need the components, you can remove them with:\n{}\n{}",
+                 If you don't need the components, you can remove them with:\n\n{}\n\n{}",
                 toolchain, cs_str, remove_msg, TOOLSTATE_MSG,
             );
         } else {
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -394,6 +394,19 @@ fn valid_profile_names() -> String {
         .join(", ")
 }
 
+fn remove_component_msg(cs: &Component, manifest: &Manifest, toolchain: &str) -> String {
+    format!(
+        "rustup component remove --toolchain {}{} {}",
+        toolchain,
+        if let Some(target) = cs.target.as_ref() {
+            format!(" --target {}", target)
+        } else {
+            String::default()
+        },
+        cs.short_name(manifest)
+    )
+}
+
 fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &str) -> String {
     assert!(!cs.is_empty());
 
@@ -411,6 +424,12 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
                 ""
             }
         );
+
+        let _ = write!(
+            buf,
+            "If you don't need the component, you can remove it with:\n{}",
+            remove_component_msg(&cs[0], manifest, toolchain)
+        );
     } else {
         let same_target = cs
             .iter()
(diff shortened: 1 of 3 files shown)


tests/dist.rs
```diff
@@ -889,17 +895,194 @@ fn removed_component() {
...
+                    assert!(err.to_string().contains("rustup component remove --toolchain nightly --target x86_64-apple-darwin bonus-2"));
+                }
+                _ => panic!(),
+            }
+        },
+    );
+}
+
+#[test]
+fn removed_components_with_different_target() {
```
dmarcoux: I cannot get this test to pass. Something is wrong in its setup.

tests/dist.rs
```diff
@@ -190,9 +190,23 @@ pub fn create_mock_channel(
...
+    packages.push(bonus_component(
+        "bonus-2",
+        contents.clone(),
+        "x86_64-apple-darwin",
+    ));
+    packages.push(bonus_component(
+        "bonus-3",
+        contents,
+        "x86_64-unknown-linux-gnu",
+    ));
```
kinnison: Is there not an existing scenario which does what you need without adding additional components?

tests/dist.rs
kinnison: Modifying the channel is not the best way to do this.  It's better to have configured a scenario which does this kind of thing.  Ideally for comprehensibility it ought to do it with `rust-std` since that's the commonly used component which might be installed with a target different to that of the toolchain's host.

src/errors.rs
```diff
@@ -386,6 +386,19 @@ fn valid_profile_names() -> String {
...
+        toolchain,
+        if let Some(target) = cs.target.as_ref() {
+            format!(" --target {}", target)
+        } else {
+            String::default()
+        },
+        cs.short_name(manifest)
+    )
+}
+
```
kinnison: Can you please give me some examples of how this would look when the component in question has a target which is the same as the toolchain?  e.g. if nightly-x86_64-unknown-linux-gnu is missing rustfmt-x86_64-unknown-linux-gnu.

kinnison (changes requested):
I apologise for taking so long to get back to you.

The approach is not bad.  I'm worried that you had to add a whole bunch of "bonus" components to our channels rather than using an existing scenario.  I'd prefer a smaller change to one or more scenarios to make your testing possible if you need changes at all.  In addition, as I noted on the test you are finding hard to make work, your chosen approach of modifying channels in-test is awkward and painful, and not an example of good basic test writing (from wherever you found it) instead you should try and use an existing scenario where as the day changes, components come and go.  This more accurately models the `nightly` changes which you're trying to improve the situation for anyway.

kinnison:
@dmarcoux I've selected this issue (and thus PR) for the 1.23.0 milestone.  Are you likely to be able to continue work on it any time soon?  If not, no worries, just let us know here so that someone else can pick up the work and complete it.

dmarcoux:
@kinnison Sorry for the delay in responding to your previous comments. I will have time again at the end of the month after the 25th of July.

kinnison:
Wonderful news, thank you @dmarcoux

tests/dist.rs
dmarcoux: I based the new tests on the existing `removed_component` test. I'm not really sure on how to configure a scenario for tests on removed components if `removed_component` is not a good example.

Edit: I found `unavailable_component` which seems to be a simpler setup. I'll base my tests on it.

dmarcoux:
I might be able to find some time to work on this before the 25th, but I cannot guarantee. Anyway, I already pushed some changes to the tests `unavailable_components_with_different_target` and `unavailable_components_with_same_target`. Could you let me know what you think @kinnison?

kinnison:
I'll take a look, but don't worry about this until you can dedicate a solid chunk of time, just do what you need to do in the meantime and come back when you're ready.  No rush.

src/errors.rs
dmarcoux: With the current changes, it would look like this:

`rustup component remove --toolchain nightly-x86_64-unknown-linux-gnu --target rustfmt-x86_64-unknown-linux-gnu`

Is this wrong?

dmarcoux:
Hey @kinnison, could you have a look please? I just pushed changes and I would like to have your opinion on the changes.

src/errors.rs
kinnison: Yeah, I'd expect `rustup component remove --toolchain nightly-x86_64-unknown-linux-gnu rustfmt` I'm not sure that `--target` makes any sense here, thinking about it some more.

In fact, now I think more, I expect there's really two situations we need to deal with:

1. A "normal" component needs removing (no target is needed)
2. `rust-std` needs removing for a target which doesn't match the toolchain host, in which case we ought to recommend `rustup target remove --toolchain $TOOLCHAIN $TARGET` e.g. `rustup target remove --toolchain nightly-x86_64-unknown-linux-gnu wasm32-unknown-unknown`

kinnison: In addition, as mentioned in #2440 we should check if we need to use the rename tables here.

kinnison:
The tests you add still feel very complicated to me; but perhaps that's needed -- let's resovle the point about how we think the UX should be first though, then I'll reconsider if we can test this more efficiently.

src/errors.rs
dmarcoux: What are the rename tables?

dmarcoux:
Regarding the new tests, I based them on the existing `unavailable_component` test.

src/errors.rs
kinnison: A toolchain channel has a small mapping for renaming components.  e.g. the underlying component for rustfmt is actually rustfmt-preview and we have a rename in place to rename that to rustfmt for users.  If you look at the end of a channel manifest you'll see:

```toml
[renames.clippy]
to = "clippy-preview"

[renames.miri]
to = "miri-preview"

[renames.rls]
to = "rls-preview"

[renames.rustfmt]
to = "rustfmt-preview"
```

We *might* need to look at using https://github.com/rust-lang/rustup/blob/master/src/dist/manifest.rs#L313 I'm not sure.

I guess that to be sure on that part we ought to ensure we're using `rls` in one of the tests since IIRC that is renamed in the test manifests.

dmarcoux:
Sorry for coming to this after so many weeks, but I won't be able to finish this. I hope somebody else can pick this up.

src/errors.rs
kinnison: I've been through the codepath and believe it is using the rename tables so that's fine.

I've also written a patch which special-cases `rust-std` to suggest using `rustup target remove`.

kinnison:
I'm working on tidying the last bits of this since @dmarcoux did a good job of getting as far as they did.

I'll push an updated set of commits as soon as I'm finished tweaking things.

I'm currently trying to tweak things so that we don't emit `--target ...` if it's not needed (i.e. it's the default target)

kinnison:
I've decided that right now the extraneous `--target` is less important.

kinnison (approved):
I think I'm good with this as-is.  Assuming CI goes green.

PR #2385, opened 2020-06-25
Better asserts

rbtcollins:
These assertions were not providing any diagnostic information


kinnison (approved):
Assuming green CI (which I don't doubt ought to occur) +1 to merge this yourself.

PR #2386, opened 2020-06-25
  Fix removing from PATH on windows when last item

rbtcollins:
When the cargo bin is the last item in the PATH on windows, a trailing ;
was left behind, which leave an empty segment in the PATH.

Also refactor into unit tests, for faster testing.

This has arguably the same coverage: we know that windows path setting
works in general due to the presence of at least one end to end test;
and it is implemented by calling this one code path; this pure function
in the code path can therefore be unit tested to examine it's corner
case behaviour rather than testing end to end every possible
interaction.

$ time ...

running 1 test
test cli::self_update::windows::tests::windows_uninstall_removes_semicolon_from_path ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out

real    0m0.041s
user    0m0.000s
sys     0m0.030s

$ time ...

running 1 test
test windows_uninstall_removes_semicolon_from_path ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 56 filtered out

real    0m13.450s
user    0m0.000s
sys     0m0.015s

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -222,6 +222,44 @@ fn get_windows_path_var() -> Result<Option<String>> {
     }
 }
 
+// Returns None if the existing old_path does not need changing
+fn _remove_from_path(old_path: &str, path_str: &str) -> Option<String> {
+    let idx = old_path.find(&path_str)?;
+    // If there's a trailing semicolon (likely, since we probably added one
+    // during install), include that in the substring to remove. We don't search
+    // for that to find the string, because if its the last string in the path,
+    // there may not be.
+    let mut len = path_str.len();
+    if old_path.as_bytes().get(idx + path_str.len()) == Some(&b';') {
+        len += 1;
+    }
+
+    let mut new_path = old_path[..idx].to_string();
+    new_path.push_str(&old_path[idx + len..]);
+    // Don't leave a trailing ; though, we don't want an empty string in the
+    // path.
+    if new_path.ends_with(';') {
+        new_path.pop();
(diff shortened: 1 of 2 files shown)


src/cli/self_update/windows.rs
```diff
@@ -430,3 +439,18 @@ pub fn get_add_path_methods() -> Vec<PathUpdateMethod> {
...
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn windows_uninstall_removes_semicolon_from_path() {
+        assert_eq!(
+            "foo",
+            super::_remove_from_path(
+                r"c:\users\example\.cargo\bin;foo",
```
kinnison, line 457: Could we also have a variant which tests where that semicolon is not present?

kinnison (approved):
One extra unit test wanted, then +1 to merge on green CI.

src/cli/self_update/windows.rs
rbtcollins: No. Or rather this *is* that variant: this tests that when the path shrinks so that the ; is not needed anymore, it is removed. This is the exact test that was replaced.

The reason it *looks* different is that in the other test, the intermediate state where PATH="cargo_home+bin;foo" is never explicitly spelt out - it is created by taking foo; installing cargo; then uninstalling and checking that only foo remains, not ";foo".

Because this is a more focused test, we're not also executing the installation half; thats the only difference.

kinnison (approved):
Merge on green CI

Issue #556, opened 2016-07-01
PATH variable not available in non-interactive shells on Linux


pornel:
I'm trying to deploy code using Ansible, which (very indirectly) runs `cargo` via non-interactive `ssh` session.

Rustup modifies `PATH` in `~/.profile` on Linux, but AFAIK this file is only used for interactive sessions. `~/.bashrc` is used for both interactive and non-interactive sessions, so it may be a better place to modify path.

sbstp:
Could you add a line in your `~/.bashrc` to `source ~/.profile`?

I've done some digging and found [why this is like this](https://github.com/rust-lang-nursery/rustup.rs/issues/211). Basically, for Desktop Environments.

Relevant part:
> For example if a user launches application (IDE or Emacs, for example) via desktop environment and not via command line, then the PATH will be wrong.

Diggsey:
There's basically no good solution to this: `.profile` is about the closest there is to a per-user setting, and excepting some shells which ignore that in favour of a shell-specific file, it *should* be sourced on login.

Unfortunately, for whatever reason, ssh sessions do not run in a login shell, despite logging in the user :angry: 

For the moment at least, if you're a linux user, you're going to have to know where this needs to be sourced to meet your needs. The best we can do is make that as easy as possible.

brson:
Another variant of https://github.com/rust-lang-nursery/rustup.rs/issues/371

PR #2387, opened 2020-06-25
Implement shell PATH fixes via UnixShell trait

### LINKED ISSUE #332: Check if `~/.cargo/bin` already on PATH

### LINKED ISSUE #371: PATH not set post installation on OS X 10.11.4
The installer said it would set the path, after installation, but even after installation there's no sign of rustup.

### LINKED ISSUE #883: PATH to cargo/bin is in wrong file, should be in .bash_profile (CentOS)

### LINKED ISSUE #1881: A more robust post-install environment setup
This is based on my recent frustration reinstalling WSL and rustup, and subsequent discussion here: https://github.com/rust-lang/rustup.rs/issues/975#issuecomment-496667276

The rustup install experience is still frustrating. Ever since modifying rustup to modify only `.profile` instead of the rc files, I don't particularly know how to get cargo to work on my WSL install, and always end up modifying `.bashrc`.

I don't know when or if `.profile` is executed by default on WSL, but some cursory googling suggests it is not. I have only just today figured out how to open a WSL terminal to a login shell.

And I suspect many users don't understand the difference between a login shell and non-login shell. That they have to log out of their GUI and back in (as I understand it - or understand what a login shell is and launch one) is obnoxious. I just want to start working in Rust immediately.

I know that the "correct" thing to do is to modify `.profile`, `.bash_profile`, and maybe `.bash_login` (which is not modified today), but I don't think it provides a good experience, and I'd like to propose a more rubust solution.

- Change `~/.cargo/env` to conditionally modify $PATH if it is not already set.
- At install, source that file from `.profile`, `.bash_profile`, `.bash_login`, `.bashrc`, and `.zshrc`, any of those that exist.
- At uninstall, reverse those changes as is done today.

For purists who don't like this, they can configure the installer to not touch their dotfiles.

It was additionally suggested in the above mentioned thread that the rustup installer itself could immediately spawn a new shell so that the current terminal works immediately. I think this is worth considering. (Edit: though there are challenges here with non-interactive installs, installs that prefer non-bash environments, maybe others. It would also need to be a config option).

It is embarrassing that rust does not "just work" after install. It is the first impression of the rust experience.

Edit: removed some unrelated comments about $CARGO_HOME

### LINKED ISSUE #2106: Installer modifies ~/.profile which is ignored for graphical terminals
**Problem**
When running installer using `curl https://sh.rustup.rs -sSf | sh` in Debian 10 "Buster", installer adds a line to modify `PATH` in `~/.profile`. That file is not used at all in graphical terminals, so this is wrong. I would expect modified `PATH` to be active in both login shells and graphical terminals.

**Steps**
1. run `curl https://sh.rustup.rs -sSf | sh` accepting defaults
2. restart computer
3. start graphical terminal, see that `PATH` has not been modified

**Possible Solution(s)**
The line to modify `PATH` needs to be added to `~/.bashrc` instead of `~/.profile`.

**Notes**

Output of `rustup --version`: none, as rustup is not in PATH
Output of `rustup show`: none, as rustup is not in PATH

workingjubilee:
Resolves #1881, and by doing so also:
Fixes #2106.
Fixes #883.
Fixes #371.
Fixes #332.

This PR significantly changes Rustup's PATH modification strategy:
1) In $CARGO_HOME an `env` shell script is written, and now it is sourced in every single rcfile that Rustup has decided to update.
2) This `env` shell script describes how to update PATH idempotently.
3) A new UnixShell trait is added that describes what to write and where for every shell it is implemented for.
4) It also burns down a lot of outdated code that doesn't rely on this code architecture.

The purpose is to hopefully make it easier to implement new shells in the future and also to fix various incompatibility issues that have historically occurred, especially with regards to GUI terminal emulators.


workingjubilee:
Alright, further cleanup will be incoming as there's still duplicate code in this rough draft (against the previous draft), but this is and will remain mergeable. It took a While to figure out how best to do it because this is going to wind up cutting out a lot of other code (and I found a surprising number of repeated code, even essentially right after itself), so I wound up trashing a few drafts for maintenance concerns. Got to the point where I wanted to just make sure this got up and passing Windows tests, and in front of other eyes.

The main key here is that the UnixShell trait should localize all shell support implementation so that it only happens in one place from now on, so hopefully when a shell support issue gets raised it does not take one aeon.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,94 @@
...
+
+struct Zsh;
+impl UnixShell for Zsh {
+    fn does_exist(&self) -> bool {
+        self.rcfile().map_or(false, |rc| rc.is_file())
+            || matches!(utils::find_cmd(&["zsh"]), Some(_))
+    }
+
+    fn rcfile(&self) -> Option<PathBuf> {
+        let zdotdir = match env::var("ZDOTDIR") {
```
rbtcollins: please use the new process abstraction we've added for supporting unit tests rather than env:: directly see CONTRIBUTING.md

tests/cli-self-upd.rs
```diff
@@ -330,8 +330,8 @@ fn install_adds_path_to_profile() {
 
 #[test]
 #[cfg(unix)]
-fn install_adds_path_to_bash_profile() {
-    install_adds_path_to_rc(".bash_profile");
+fn install_adds_path_to_bashrc() {
```
rbtcollins: Won't this lead to ~/.cargo/bin being in the path as many times as subshells have been entered? Thats why ~/.profile is used for this kind of thing usually - that or more complicated once-and-only-once constructs.

Secondly, adding to bashrc has the negative effect of *not* adding to PATH when someone ssh's into their environment and runs rustup non-interactively.

I think both of these are addressable. But perhaps I'm wrong and your patch does - but there's no documentation update explaining cover message explaining the overall change so I'm a bit unclear on the big picture and just inferring it from the commit contents :).

The google cloud sdk does the sort of idempotent shell shenanigans I'm talking about; adding that into rc files in addition to keeping the profile additions would suffice I think.

tests/cli-self-upd.rs
```diff
@@ -441,8 +441,8 @@ fn uninstall_removes_path_from_profile() {
 
 #[test]
 #[cfg(unix)]
-fn uninstall_removes_path_from_bash_profile() {
-    uninstall_removes_path_from_rc(".bash_profile");
+fn uninstall_removes_path_from_bashrc() {
```
rbtcollins: Once other review changes are taken into consideration, I think you'll probably want to add coverage rather than changing existing tests (unless I'm wrong, which I may be :)). But if so, perhaps consider migrating to focused functional tests rather than more heavyweight tests, because the overall install path has little to do with the changes you're making. Some modest test infra may be needed to support that, if you need a hand bootstrapping that, let me know.

rbtcollins (changes requested):

tests/cli-self-upd.rs
workingjubilee: So, here's the key problem: `.bash_profile` matters not-at-all for the 90% case of a GUI terminal user.
The primary way this is addressed lately is by having .bash_profile `source ~/.bashrc`.
Modifying the exact shell code so it doesn't add it twice is simple, but rustup currently accomplishes this using .bash_profile and thereby makes it so that for most developers, it appears 0 times in PATH.

Pushing it into both .bash_profile and .bashrc could be done but seems like succumbing to madness (bash's madness, specifically, since it results in duplication, and thus it is more opaque to the user how to remove it if they need to manually edit it, but also is an unfortunately plausible answer to the problems of bash).

workingjubilee: The Google Cloud SDK is actually in fact also a wonderful example of "this doesn't actually work for many users precisely because it updates .bash_profile over .bashrc when .bash_profile isn't loaded when I start a terminal."

rbtcollins: google cloud SDK updated rc files when I installed it recently for my new job. Perhaps I did something unusual there.

I appreciate the issues for GUi terminal users, and am sympathetic to them. I use Linux all day every day, but I don't use a Linux GUI (well, I do use i3 from time to time, but thats a digression). 

I'd like to make sure we meet everyones needs.

src/cli/self_update/shell.rs
workingjubilee: Fixed. I just wasn't sure what exactly was going on there.

tests/cli-self-upd.rs
workingjubilee: NB: Discussion elaborated and the conclusion arrived at is that actually, descending into a little madness is fine, but we still have to use an idempotent framework for modifying PATH, and this is unavoidable because all of this might have made sense at one time, but now it's 2020. Further explanation will be documented for future reference in followup commit.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    fn rcfiles(&self) -> Vec<PathBuf> {
+        // FIXME: if zsh exists but is not in the process tree of the shell
+        // on install, $ZDOTDIR may not be loaded and give the wrong result.
+        let zdotdir = match process().var("ZDOTDIR") {
+            Ok(dir) => Some(PathBuf::from(dir)),
+            _ => utils::home_dir(),
+        };
+
+        // Don't bother with .zprofile/.zlogin because .zshrc will always be
```
Nemo157: `.zshrc` won't be sourced for the login shell which sets up the environment for things launched from GUI applications, and I believe having a `.zprofile` to set that up is more common than having a `.zshenv`.

workingjubilee: I know.

GUI terminals can start *without* a login shell, but _do_ source .zshrc, making adding to the login shell irrelevant, and I can point you to the exact lines of code which illustrate that the decision to source .zshrc and .zprofile in zsh are made completely independently of one another:
https://github.com/zsh-users/zsh/blob/19390a1ba8dc983b0a1379058e90cd51ce156815/Src/init.c#1329-1367

.zprofile is excluded from consideration by design.

Nemo157: I'm not talking about terminals. I'm talking about logging directly into a DE (which will run under a login shell) and using an application launcher to open an application that ends up calling a non-interactive script. At no point in that chain will there be an interactive shell which loads `.zshrc`, only the one login shell which loads `.zshenv` and/or `.zprofile` and a non-interactive shell which loads only `.zshenv` if it exists. For users without a `.zshenv` this would result in the script running without its path being updated to include `~/.cargo/bin`.

workingjubilee: I do not believe this hypothetical logic chain actually exists.

workingjubilee: Nonetheless, I believe this is an excellent argument in favor of resolving this permanently in favor of .zshenv (indeed, this kind of conversation was probably precisely why .zshenv was designed!), and will update the logic accordingly.

workingjubilee: I suppose I must concede that this has been an issue before: #556! All the more reason to use zsh's provided solution to this bit of complexity instead of handling it ourselves, in this case.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+//! environments often choose to NOT start login shells in GUI terminals. Thus,
+//! a trend has emerged to place PATH updates in other run-commands (rc) files,
+//! leaving Rustup with few assumptions to build on for fulfilling its promise
+//! to set up PATH appropriately.
+//!
+//! Rustup addresses this by:
+//! 1) using a shell script that updates PATH if the path is not in PATH
+//! 2) sourcing this script in any known and appropriate rc file
+
+use super::canonical_cargo_home;
```
rbtcollins: canonical_cargo_home is a bit of an odd function - it transforms any non-clean unicode characters in a bid to make the path printable. This matters when e.g. folk have paths on disk that are in SHIFT-JIS but their encoding is in CP1252 or other incompatible combinations.

So rather than using this function designed for display, call `cargo_home()?.join("bin").to_str()`. Factoring out a helper for that is *just about* a common enough thing to be worth doing, but I haven't gotten around to it yet.

If `to_str()` fails, then we should fail to write the env file for now (as it would be written with bogus content). A more sophisticated iteration of this work would be to write the env file as a bytestring, not unicode, so that we can write it even if cargo_home isn't valid unicode. I don't know if you want to tackle that right now or not.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+        let cargo_bin = format!("{}/bin", canonical_cargo_home()?);
+        let env_name = utils::cargo_home()?.join(self.name);
+        let env_file = self.content.replace("{cargo_bin}", &cargo_bin);
+        utils::write_file(self.name, &env_name, &env_file)?;
+        Ok(())
+    }
+}
+
+#[allow(dead_code)] // For some reason.
+const POSIX_ENV: &str = include_str!("env");
```
rbtcollins: Given the size of the template, I'd personally prefer to haave it inline here as a raw string. @kinnison what do you think?

workingjubilee: I would write that change without complaint but I must Note Ominously:
That will get messy the moment `env.csh`, `env.fish`, or something similar gets implemented.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+
+macro_rules! support_shells {
```
rbtcollins: This macro doesn't seem worth it: all its saving is Box::new in each line of a constant array initializer for vec! : I worry that having to go see what the macro does impedes clarity more than it helps. This isn't a must-change, just a thought.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+            .take(1)
+            // Pick .bashrc if it exists for GUI terminals
+            .chain(Bash::rc().filter(|rc| rc.is_file()))
+            .collect()
+    }
+}
+
+struct Zsh;
+impl UnixShell for Zsh {
+    fn does_exist(&self) -> bool {
```
rbtcollins, line 174: I'm not completely clear on what this actually tests; I *think* it is:
if zsh in SHELL or ZDOTDIR != "" or ~/.zshenv exists or zsh exists.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+        matches!(process().var("SHELL"), Ok(sh) if sh.contains("zsh"))
+            || matches!(process().var("ZDOTDIR"), Ok(dir) if dir.len() > 0)
+            || self.rcfiles().iter().any(|rc| rc.is_file())
+            || matches!(utils::find_cmd(&["zsh"]), Some(_))
+    }
+
+    fn rcfiles(&self) -> Vec<PathBuf> {
+        // FIXME: if zsh exists but is not in the process tree of the shell
```
rbtcollins: This seems like an intrinsic characteristic unless we get into very convoluted territory where we invoke rustup inside zsh to read out the calculated value of ZDOTDIR.

I think we should either plan on doing that in a future iteration, or not have a FIXME and instead have a FAQ about limitations.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+    fn update_rcs(&self) -> Vec<PathBuf> {
+        // Write to .profile even if it doesn't exist.
+        self.rcfiles()
+    }
+}
+
+struct Bash;
+
+// Bash is the source of many complications because it loads ONLY 1 rc,
+// either a login profile or .bashrc, not both.
```
rbtcollins: It is true that it does that, but every bash_profile I've ever seen has 
```

if [ -f ~/.bashrc ]; then
    source ~/.bashrc
fi

```

in it, so I'm really not sure that this comment is more than us being judgemental, in a way that ill suits us.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+            profiles.push(rc);
+        }
+        profiles
+    }
+
+    fn update_rcs(&self) -> Vec<PathBuf> {
+        Bash::profiles()
+            .into_iter()
+            .filter(|rc| rc.is_file())
+            // bash only reads one "login profile" so pick the one that exists
```
rbtcollins: I don't think the complexity of filtering just one of bash_profile, profile etc makes sense, because we will often have to update both bash_profile and profile when a use has both bash and ash installed.

So because we can't only update just-one of the files, we should instead by much simpler and dumber: update all the files that are present. I think this will make the code substantially smaller as well: no need to separate out the rc vs profiles.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+pub trait UnixShell {
+    // Detects if a shell "exists". Users have multiple shells, so an "eager"
+    // heuristic should be used, assuming shells exist if any traces do.
+    fn does_exist(&self) -> bool;
+
+    // Gives all rcfiles of a given shell that rustup is concerned with.
+    // Used primarily in checking rcfiles for cleanup.
+    fn rcfiles(&self) -> Vec<PathBuf>;
```
rbtcollins, line 87: See below - but I think rcfiles vs update_rcs is an unneeded distinction, because we can be pretty certain we won't be `pure` as far as updating bash's profile files is concerned, because of the intersection with the posix shell definition.

src/cli/self_update/unix.rs
```diff
@@ -57,66 +57,50 @@ pub fn complete_windows_uninstall() -> Result<utils::ExitCode> {
...
 }
 
-pub fn do_remove_from_path(methods: &[PathUpdateMethod]) -> Result<()> {
-    for method in methods {
-        if let PathUpdateMethod::RcFile(ref rcpath) = *method {
-            let file = utils::read_file("rcfile", rcpath)?;
-            let addition = format!("\n{}\n", shell_export_string()?);
+pub fn do_remove_from_path() -> Result<()> {
+    for sh in shell::get_available_shells() {
+        let source_cmd = format!("\n{}\n", sh.source_string()?);
```
rbtcollins, line 62: This is constant in the loop, it could be extracted above it.

src/cli/self_update/unix.rs
```diff
@@ -57,66 +57,50 @@ pub fn complete_windows_uninstall() -> Result<utils::ExitCode> {
...
+        let source_bytes = source_cmd.into_bytes();
```
rbtcollins: this probably needs to be bytes the whole way through - because of the issue I mentioned with potential non-unicode translatable paths.

src/cli/self_update/unix.rs
```diff
@@ -57,66 +57,50 @@ pub fn complete_windows_uninstall() -> Result<utils::ExitCode> {
...
-                .windows(addition_bytes.len())
-                .position(|w| w == &*addition_bytes);
-            if let Some(i) = idx {
-                let mut new_file_bytes = file_bytes[..i].to_vec();
-                new_file_bytes.extend(&file_bytes[i + addition_bytes.len()..]);
-                let new_file = String::from_utf8(new_file_bytes).unwrap();
-                utils::write_file("rcfile", rcpath, &new_file)?;
-            } else {
-                // Weird case. rcfile no longer needs to be modified?
+            if let Some(idx) = file_bytes
```
rbtcollins, line 69: It would perhaps be nice to still work if someone added whitespace by mistake. Thats a separate enhancement obviously, but maybe a FIXME?

src/cli/self_update/shell.rs
workingjubilee: The function should Err in that case, right? What Err should it return? I'm struggling a bit to interpret how to make err_chain work here, looking at it. to_str() gives an Option which I can .ok_or() on. 

This dilemma crops up on either path. Frankly it seems like I'm going to go to the same amount of trouble either way so I might as well write some raw bytes, but we'll see.

src/cli/self_update/unix.rs
```diff
@@ -57,66 +57,50 @@ pub fn complete_windows_uninstall() -> Result<utils::ExitCode> {
...
+        let source_cmd = format!("\n{}", sh.source_string()?);
+        for rc in sh.update_rcs() {
+            if !rc.is_file() || !utils::read_file("rcfile", &rc)?.contains(&source_cmd) {
+                utils::append_file("rcfile", &rc, &source_cmd).chain_err(|| {
                     ErrorKind::WritingShellProfile {
-                        path: rcpath.to_path_buf(),
+                        path: rc.to_path_buf(),
                     }
                 })?;
+                let script = sh.env_script();
```
rbtcollins, line 99: I think this code is suffering over-generalism: we know we've created a single env file on disk in the current code, we don't need to track this in a general fashion. Just write the file once, outside the loop (unconditionally, so that the latest version of the file is written).

tests/cli-self-upd.rs
```diff
@@ -328,12 +328,24 @@ fn install_adds_path_to_profile() {
...
+
 #[test]
 #[cfg(unix)]
 fn install_adds_path_to_bash_profile() {
     install_adds_path_to_rc(".bash_profile");
 }
 
+#[test]
+#[cfg(unix)]
+fn install_adds_path_to_zshenv() {
```
rbtcollins: I'd really like to see these written as unit tests: right now they will run serially due to the mutex in setup(); they are doing about 10k times the work actually under test.

Alternatively, just have one test that has all the files on disk and check they are all modified.

You can see a few recent patches from me, including the unmerged https://github.com/rust-lang/rustup/pull/2412 for some possible sorts of changes to enable this: but basically separating out functional and pure layers and testing logic via pure functions (where possible).

src/cli/self_update/unix.rs
workingjubilee: The reason I wrote it this way was in fact planning ahead for and thinking about other shells, yes. It seemed like an inevitable complication.

src/cli/self_update/shell.rs
workingjubilee: Shell out to zsh and ask it to report back? :laughing:

tests/cli-self-upd.rs
workingjubilee: Cutting down the work sounds good. Pulling them into a single test seems like the current path of least resistance given how much of it is testing on doing IO (therefore impure) in the correct places in the correct manner?

src/cli/self_update/shell.rs
workingjubilee: Noted. Fiiixed?

src/cli/self_update/shell.rs
workingjubilee: Yes. (not anymore)

src/cli/self_update/shell.rs
workingjubilee: It's mostly a vestige of an earlier draft where it did more work. My current thought is that it wound up shrinking down to the point where removing it isn't much of a difference.

src/cli/self_update/shell.rs
rbtcollins: yeah - assuming zsh accepts -c , something like `zsh -c "{ourargv0} report-env ZDOTDIR"`.

src/cli/self_update/shell.rs
rbtcollins: if you choose to defer the work and return an error, you'd say something like

```
let as_str = if let Some(s) = p.to_str() { s } else { bail!("non-unicode homedir") } 
```

src/cli/self_update/shell.rs
rbtcollins: mmm, I would then put those inline in the trait impl for those shells, since we do have that nice structure :).

src/cli/self_update/unix.rs
```diff
@@ -57,66 +57,50 @@ pub fn complete_windows_uninstall() -> Result<utils::ExitCode> {
...
             let file_bytes = file.into_bytes();
-            let addition_bytes = addition.into_bytes();
-
-            let idx = file_bytes
-                .windows(addition_bytes.len())
-                .position(|w| w == &*addition_bytes);
-            if let Some(i) = idx {
-                let mut new_file_bytes = file_bytes[..i].to_vec();
-                new_file_bytes.extend(&file_bytes[i + addition_bytes.len()..]);
-                let new_file = String::from_utf8(new_file_bytes).unwrap();
```
rbtcollins, line 75: one additional thing I missed: this change will no longer uninstall previous installs of rustup properly; I think checking for the prior form of rustup path add would make sense.

rbtcollins:
I think we also need a genuinely new test: that upgrading rustup removes the old added PATH mechanism at the same time it installs the new PATH env mechanism.

voltamachine:
```
Trap Card
```

src/cli/self_update/shell.rs
workingjubilee: Ah, that is simpler than expected then, so I will do that and come back to write the bytestring.

src/cli/self_update/shell.rs
workingjubilee: It does, currently thinking `zsh -c "echo $ZDOTDIR"` or a close variation would be simplest from what I remember of all the APIs involved.

workingjubilee:
Ace of Rebase.
Apologies if this messes up the review flow.

src/cli/self_update/unix.rs
workingjubilee: I agree. Kinnison said I didn't _have_ to do that, but I was going to anyways, and mostly wanted to see an initial review pass on this PR before doing more work (in case I had to throw everything out or something).

It's going to be a distinct function so that it doesn't get tangled in the shell-handling logic, and it will be invoked both coming and going.

src/cli/self_update/shell.rs
workingjubilee: I have a draft of this logic implemented and I've done some basic tinkering around to verify that it should work. Writing tests for it is going to be... interesting.

src/cli/self_update/shell.rs
workingjubilee: > having it as an external script means we could shellcheck it
> So please ensure it's called something like env.sh so we can do that

-Kinnison, Discord @ 2020-07-13 16:56 UTC

I found his reasoning persuasive here so I'm going to fix that up as well.

workingjubilee:
TODO:
- [ ] Add a function that returns an unpretty but defaulting cargo_home
- [ ] Add legacy `export PATH` method removal function
- [ ] Improve tests to verify legacy function works on update and uninstall
- [ ] Reduce test workload
- [ ] Improve testing for zsh shellout (stretch)
- [ ] Factor tests nicely (stretch)
- [ ] Move to bytestrings (stretch)

src/cli/self_update/shell.rs
workingjubilee: Mm, the evolution of the zsh implementation suggests it is going to be a concern anyways.

src/cli/self_update/shell.rs
workingjubilee: A later patch, however.

I should note that Rustup was previously using the [lossy canonical_cargo_home]https://github.com/rust-lang/rustup/blob/7050161b9d89afd1db7d3d374b1acc2bb368383e/src/cli/self_update/unix.rs#L95-L100) in this segment of the code anyways so this is already doing additional fixes.

tests/cli-self-upd.rs
workingjubilee: I'm sorry, I have tried several times to understand it but I simply do not get the unit test framework well enough at all to write these as unit tests. I will not be doing so for this patch.

kinnison:
Goodness there has been a lot of work on this, nice to see @workingjubilee really pushing hard.  The PR is currently in conflict (likely due to some test updates by @rbtcollins) so needs some more tweaking.

How do we feel about the over-all state of this work @rbtcollins ?  I see there's some open discussions around test cases.

workingjubilee:
Should be a simple rebase, I should only have to go back a few steps.

But also I think this will be Rebase MkIII and I would like to be done with this branch in particular if it is mostly a matter of improving this further on top of already acceptable code. I am confident in the integration tests I have written and updated things to, and I am happy to help in reducing the test workload further, but that seems like a second-stage concern that will require some working-out over time because the unit test code is currently in an under-construction state.

rbtcollins:
I think the only thing that is a genuine need is the functionality && test for dealing with older PATH implants - which it looks like @workingjubilee may have done but I haven't yet reviewed? 

I'm not going to hold this up over whether a given test is unit or functional; thats a (very :P) nice to have.

workingjubilee:
Yes, test was added, should be functional!

src/cli/self_update.rs
```diff
@@ -514,23 +513,14 @@ fn pre_install_msg(no_modify_path: bool) -> Result<String> {
...
-                .into_iter()
-                .filter_map(|m| {
-                    if let PathUpdateMethod::RcFile(path) = m {
-                        Some(format!("{}", path.display()))
-                    } else {
-                        None
-                    }
-                })
+        // Brittle code warning: some duplication in unix::do_add_to_path
+        #[cfg(unix)]
```
rbtcollins: I think these blocks should either use https://docs.rs/cfg-if/0.1.10/cfg_if/ or be more clearly disjoint cfg(not(windows)) + cfg(windows)

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,208 @@
...
+        let home = utils::cargo_home()?;
+        let cargo_bin = format!("{}/bin", cargo_home_str()?);
+        let env_name = home.join(self.name);
+        let env_file = self.content.replace("{cargo_bin}", &cargo_bin);
+        utils::write_file(self.name, &env_name, &env_file)?;
+        Ok(())
+    }
+}
+
+// TODO: Update into a bytestring.
```
rbtcollins, line 52: This is for a future iteration yes?

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    fn rcfiles(&self) -> Vec<PathBuf> {
+        match utils::home_dir() {
+            Some(dir) => vec![dir.join(".profile")],
+            _ => vec![],
+        }
+    }
+
+    fn update_rcs(&self) -> Vec<PathBuf> {
+        // Write to .profile even if it doesn't exist.
```
rbtcollins: I think we probably need to say why (e.g. 'because posix shell only loads .profile'). I do agree that configuring a PATH appropriately for posix shell is the right thing to do.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,208 @@
...
+
+struct Bash;
+
+impl UnixShell for Bash {
+    fn does_exist(&self) -> bool {
+        self.update_rcs().len() > 0
+    }
+
+    fn rcfiles(&self) -> Vec<PathBuf> {
+        [".bash_profile", ".bash_login", ".bashrc"]
```
rbtcollins, line 135: I presume .profile is missing because we're confident that Posix shell will apply it? Perhaps a comment here, in case of future refactorings that e.g. permit the user to select just one shell - in that case we'd want to check for .profile for bash I think?

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,208 @@
...
+            }
+        }
+    }
+}
+
+impl UnixShell for Zsh {
+    fn does_exist(&self) -> bool {
+        // zsh has to either be the shell or be callable for zsh setup.
+        matches!(process().var("SHELL"), Ok(sh) if sh.contains("zsh"))
+            || matches!(utils::find_cmd(&["zsh"]), Some(sh) if sh.contains("zsh"))
```
rbtcollins: Does this second branch require the guard? find_cmd shouldn't return anything that fails contains("zsh") right?

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+pub mod mock;
+
+use crate::mock::clitools::{self, expect_ok, Config, Scenario};
+use crate::mock::{get_path, restore_path};
+use lazy_static::lazy_static;
+use std::path::PathBuf;
+use std::sync::Mutex;
+
+pub fn setup(f: &dyn Fn(&Config)) {
+    clitools::setup(Scenario::SimpleV2, &|config| {
```
rbtcollins: You could use Scenario::None here, and add "--default-toolchain", "none" to the init commands, it will make the tests faster and avoid unpacking content that is irrelevant to the codepaths being tested

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
+//! This file contains tests relevant to Rustup's handling of updating PATHs.
+//! It depends on self-update working, so if absolutely everything here breaks,
+//! check those tests as well.
+pub mod mock;
+
+use crate::mock::clitools::{self, expect_ok, Config, Scenario};
+use crate::mock::{get_path, restore_path};
+use lazy_static::lazy_static;
+use std::path::PathBuf;
+use std::sync::Mutex;
```
rbtcollins: please group imports per the CONTRIBUTING.md file

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+        let saved_path = get_path();
+        let _g = scopeguard::guard(saved_path, restore_path);
+
+        f(config);
+    });
+}
+
+#[cfg(unix)]
+mod unix {
+    use super::*;
```
rbtcollins: please group imports per the CONTRIBUTING.md file

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
+//! This file contains tests relevant to Rustup's handling of updating PATHs.
```
rbtcollins, line 1: This is somewhat hard to review because of the combination of cross-file move *and* content change.

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+use std::path::PathBuf;
+use std::sync::Mutex;
+
+pub fn setup(f: &dyn Fn(&Config)) {
+    clitools::setup(Scenario::SimpleV2, &|config| {
+        // Lock protects environment variables
+        lazy_static! {
+            static ref LOCK: Mutex<()> = Mutex::new(());
+        }
+        let _g = LOCK.lock();
```
rbtcollins: This lock is only relevant for PATH modifying tests on windows; I suggest not calling setup() from the unix tests

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+        });
+    }
+}
+
+#[cfg(windows)]
+mod windows {
+    use super::*;
+
+    #[test]
+    fn install_doesnt_modify_path_if_passed_no_modify_path() {
```
rbtcollins: This test should be deleted; I think that has been lost in the rebasing

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+            let new_path = environment.get_raw_value("PATH").unwrap();
+
+            assert_eq!(old_path, new_path);
+        });
+    }
+
+    // HKCU\Environment\PATH may not exist during install, and it may need to be
+    // deleted during uninstall if we remove the last path from it
+    #[test]
+    fn windows_handle_empty_path_registry_key() {
```
rbtcollins: same

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+                .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+                .unwrap();
+            let path = environment.get_raw_value("PATH");
+
+            assert!(path.is_err());
+        });
+    }
+
+    #[test]
+    fn uninstall_removes_path() {
```
rbtcollins: And this one

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+            expect_ok(config, &["rustup-init", "-y"]);
+            expect_ok(config, &["rustup", "self", "uninstall", "-y"]);
+
+            let path = config.cargodir.join("bin").to_string_lossy().to_string();
+            assert!(!get_path().unwrap().contains(&path));
+        });
+    }
+
+    #[test]
+    fn install_adds_path() {
```
rbtcollins: and this

tests/cli-paths.rs
```diff
@@ -0,0 +1,403 @@
...
+            expect_ok(config, &["rustup-init", "-y"]);
+
+            let path = config.cargodir.join("bin").to_string_lossy().to_string();
+            assert!(
+                get_path().unwrap().contains(&path),
+                format!("`{}` not in `{}`", get_path().unwrap(), &path)
+            );
+        });
+    }
+}
```
rbtcollins, line 359: install_uninstall_affect_path may be something you want to move here too for consistency

tests/cli-self-upd.rs
```diff
@@ -900,19 +690,10 @@ fn reinstall_specifying_different_toolchain() {
 }
 
 #[test]
-#[cfg(unix)]
-fn produces_env_file_on_unix() {
+fn install_sets_up_stable() {
```
rbtcollins: this seems to be coming from nowhere - I can't see this test removed elsewhere in the diff.

rbtcollins (changes requested):
I think the code itself is basically sound and we're in fine tuning over some test code rearrangements.

tests/cli-self-upd.rs
workingjubilee: The writing-env-file test moved to cli-paths (because it's part of the path, er, codepath) and this reappearing is git showing up drunk.

src/cli/self_update/shell.rs
workingjubilee: Yes. Noted.

tests/cli-paths.rs
workingjubilee: Apologies.

tests/cli-paths.rs
workingjubilee: I managed to get it exactly backwards, amazing.

tests/cli-paths.rs
workingjubilee: ~~No such variant in the Scenario enum.~~ Ah, Empty!

src/cli/self_update/shell.rs
workingjubilee: Yes, I was intending to follow up in a later PR. It will be a much smaller diff to fix it up (previously rustup was mostly interacting using lossy UTF8, for note).

tests/cli-paths.rs
rbtcollins: :) :) :)

tests/cli-paths.rs
rbtcollins: Not to worry but in future would prefer to review one-or-other. Thanks!

tests/cli-paths.rs
```diff
@@ -306,7 +296,7 @@ export PATH="$HOME/apple/bin"
 
     #[test]
     fn install_doesnt_modify_path_if_passed_no_modify_path() {
-        setup(&|config| {
+        clitools::setup(Scenario::SimpleV2, &|config| {
```
rbtcollins: This could also use Empty if you add the distribution none option.

workingjubilee:
Linux error is spurious (on snapcraft).

src/cli/self_update.rs
```diff
@@ -159,7 +161,7 @@ To get started you need Cargo's bin directory ({cargo_home}/bin) in your `PATH`
 environment variable. Next time you log in this will be done
 automatically.
 
-To configure your current shell run `source {cargo_home}/env`
+To configure your current shell run `source {cargo_home}/env.sh`
```
kinnison: This is changing an API of sorts.  Unless `env` and `env.sh` are valid here.

src/cli/self_update.rs
```diff
@@ -182,7 +184,7 @@ macro_rules! post_install_msg_unix_no_modify_path {
 To get started you need Cargo's bin directory ({cargo_home}/bin) in your `PATH`
 environment variable.
 
-To configure your current shell run `source {cargo_home}/env`
+To configure your current shell run `source {cargo_home}/env.sh`
```
kinnison: Ditto, changing an API.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,213 @@
...
+            None => bail!("Non-Unicode path!"),
+        }
+    })
+}
+
+// TODO: Tcsh (BSD)
+// TODO?: Make a decision on Ion Shell, Power Shell, Nushell
+// Cross-platform non-POSIX shells have not been assessed for integration yet
+fn enumerate_shells() -> Vec<Shell> {
+    vec![Box::new(Posix), Box::new(Bash), Box::new(Zsh)]
```
kinnison, line 73: I wonder if `.xinitrc` or `.xsession` are worth updating if present?

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,213 @@
...
+    // Used primarily in checking rcfiles for cleanup.
+    fn rcfiles(&self) -> Vec<PathBuf>;
+
+    // Gives rcs that should be written to.
+    fn update_rcs(&self) -> Vec<PathBuf>;
+
+    // Writes the relevant env file.
+    fn env_script(&self) -> ShellScript {
+        ShellScript {
+            name: "env.sh",
```
kinnison: This is the API difference I think.  `env` rather than `env.sh` would be expected by users.

tests/cli-self-upd.rs
```diff
@@ -302,219 +302,9 @@ fn uninstall_doesnt_leave_gc_file() {
...
-/// This covers both windows and unix no-path-modification cases due to the code structure.
-fn install_doesnt_modify_path_if_passed_no_modify_path() {
-    setup(&|config| {
-        let profile = config.homedir.join(".profile");
-        expect_ok(config, &["rustup-init", "-y", "--no-modify-path"]);
-        assert!(!profile.exists());
-    });
-}
+#[ignore]
+fn uninstall_stress_test() {}
```
kinnison: Why adding this?

workingjubilee: https://tahoe-lafs.org/~zooko/badmerge/simple.html

src/cli/self_update/shell.rs
workingjubilee: Well yes, you had suggested it be changed for shellcheck convenience. I can change it back.

tests/cli-self-upd.rs
workingjubilee: git rebase is drunk? I didn't intentionally make that edit, but I have had to rebase this several times and some oddities have arisen at this point (git has no concept of changes, it only has diffs between trees).

src/cli/self_update/shell.rs
workingjubilee: I don't think so but I would need to do a more extensive evaluation.

I would add X as a shell if I decided so: it is in fact a user-facing environment allowing interaction with the OS :wink:, even if conventionally we limit "shell" as a term to TUIs and their easily scripted interfaces.

src/cli/self_update/shell.rs
workingjubilee: Or not.

src/cli/self_update/shell.rs
workingjubilee: ( since all the files will likely be named env with different extensions it's probably fine anyways. )

src/cli/self_update/shell.rs
kinnison: Mmm, I was just thinking from the PoV of attempting to ensure the path is in the DE's environment for the purpose of things like keybinding-arranged launch windows.  But it's not important for this release, we can consider it in the future.

kinnison (commented):
I think this is close to ready.  @rbtcollins Could you go through your outstanding comments and resolve/follow up if needed?

Let's see if we can get this through the door.

src/cli/self_update/shell.rs
rbtcollins: Looking through this, I cannot see rcfiles() being called on the trait; only from within trait implementations. All I'm suggesting is that it be removed from the trait unless-and-until something that is calling methods on the trait needs to call it.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,213 @@
...
+        match utils::home_dir() {
+            Some(dir) => vec![dir.join(".profile")],
+            _ => vec![],
+        }
+    }
+
+    fn update_rcs(&self) -> Vec<PathBuf> {
+        // Write to .profile even if it doesn't exist. It's the only rc in the
+        // POSIX spec so it should always be set up.
+        self.rcfiles()
```
rbtcollins, line 121: With the suggestion above, rcfiles here could be inlined into update_rcs for this impl.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,213 @@
...
+            .iter()
+            .filter_map(|rc| utils::home_dir().map(|dir| dir.join(rc)))
+            .collect()
+    }
+
+    fn update_rcs(&self) -> Vec<PathBuf> {
+        self.rcfiles()
+            .into_iter()
+            .filter(|rc| rc.is_file())
+            .collect()
```
rbtcollins, line 145: And this here could use a literal array, or stay with the helper function, though it would need to move to a new impl Bash block.

src/cli/self_update/shell.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    fn update_rcs(&self) -> Vec<PathBuf> {
+        // zsh can change $ZDOTDIR both _before_ AND _during_ reading .zshenv,
+        // so we: write to $ZDOTDIR/.zshenv if-exists ($ZDOTDIR changes before)
+        // OR write to $HOME/.zshenv if it exists (change-during)
+        // if neither exist, we create it ourselves, but using the same logic,
+        // because we must still respond to whether $ZDOTDIR is set or unset.
+        // In any case we only write once.
+        self.rcfiles()
```
rbtcollins, line 194: and here it could be inlined and the other method removed.

rbtcollins (commented):
Sorry about the long delay; busy in life.

I think this is basically busy, the only thing remaining is the question I raised earlier about the unused method on the trait. I'm ok with leaving it as-is, but there really doesn't seem to be any use for it.

src/cli/self_update/shell.rs
workingjubilee: unix.sh, line 65:
```rust
        for rc in sh.rcfiles().iter().filter(|rc| rc.is_file()) {
```

workingjubilee: I know it's basically that one place. ^^;
But it is a method that has a reason for existing.

workingjubilee:
We discussed and there was an understanding gap that has been bridged, in terms of smoothing out the code. The specific requests were in service of a goal I did not understand (and initially did not inquire too deeply about). As I now understand the higher-level goal (difficult-to-itemize improvements in clarity), I can draft some commits in service of that.

That is the only objection standing, however, and I will note that it is the only issue, and should not result in major logical changes, so it could be merged as a later PR. However, as long as this one remains open, I will work on that issue here.

rbtcollins:
Please do continue on and do the improvements we've discussed, but per discord, there's a bias toward getting this in at this stage, it is good enough - so I've merged it.

workingjubilee:
Ah good. That makes patching this work from multiple directions actually easier for me since it's not all one big diff.

PR #2391, opened 2020-06-28
Convert unicode uninstall test to unit test

rbtcollins:
Added a mutex for registry editing unit tests as they will need to
mutually exclude as the functional test suite does. For now I've copied
the support functions; it is small and I've left a pointer.

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -449,6 +449,89 @@ pub fn get_remove_path_methods() -> Result<Vec<PathUpdateMethod>> {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Mutex;
+
+    use lazy_static::lazy_static;
+    use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
+    use winreg::{RegKey, RegValue};
+
+    use crate::currentprocess;
+    use crate::utils::utils;
+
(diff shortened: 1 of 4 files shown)


src/cli/self_update/windows.rs
```diff
@@ -449,6 +449,89 @@ pub fn get_remove_path_methods() -> Result<Vec<PathUpdateMethod>> {
...
+    pub fn get_path() -> Option<String> {
+        let root = RegKey::predef(HKEY_CURRENT_USER);
+        let environment = root
+            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+            .unwrap();
+        // XXX: copied from the mock support crate, but I am suspicous of this
+        // code: This uses ok to allow signalling None for 'delete', but this
+        // can fail e.g. with !(winerror::ERROR_BAD_FILE_TYPE) or other
+        // failures; which will lead to attempting to delete the users path
+        // rather than aborting the test suite.
```
kinnison, line 470: Is it worth noting this in the mock code too?

kinnison (approved):
This looks sane, if you don't think it worth adding the comment then :+1: to merge, otherwise add that comment to mock first.

src/cli/self_update/windows.rs
rbtcollins: good idea

PR #2393, opened 2020-07-01
Fix homedir detection on Windows

rbtcollins:
Correctly pass in the home::OS_ENV instance in non-test code.

### DIFF
--- a/src/currentprocess/homethunk.rs
+++ b/src/currentprocess/homethunk.rs
@@ -5,10 +5,24 @@ use std::ops::Deref;
 use std::path::PathBuf;
 
 use super::CurrentDirSource;
-use super::CurrentProcess;
+use super::HomeProcess;
+use super::OSProcess;
+use super::TestProcess;
 use super::VarSource;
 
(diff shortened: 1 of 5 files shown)


kinnison (approved):

kinnison:
It's a bit crunchy but I'm not sure I see an easier way just yet.  The whole of this will need reviewing more carefully when I have more brain, but this doesn't look wrong so here's hoping.

src/currentprocess/homethunk.rs
```diff
@@ -19,3 +33,15 @@ impl home::Env for Box<dyn CurrentProcess + 'static> {
         VarSource::var_os(self.deref(), key)
     }
 }
+
+impl home::Env for OSProcess {
+    fn home_dir(&self) -> Option<PathBuf> {
+        home::OS_ENV.home_dir()
```
tesuji, line 39: If we use default implementation for `home::Env`, we don't need `home::OS_ENV` here.
cc https://github.com/brson/home/pull/23/files#r449375855

PR #2398, opened 2020-07-07
www: Correct UX of the copy button

### LINKED ISSUE #2397: Style for the snippet displayed on rustup.rs is slightly broken
**Problem**

Copy button won't show or work well, and style for the snippet is broken on some environment.

I'm not sure how often is the website updated but I assume that's related to #2326.

It looks like this to me (Incognito Chrome on Windows):

image

**Steps**

1. Visit https://rustup.rs.
2. See the command and what's happen.

BeniCheni:
Fixes #2397.

Triaged. It seems that the click on the copy button would run into a console error about CSP (Content Security Policy) Unsafe_inline_script:

> Refused to execute inline event handler because it violates the following Content Security Policy directive: "script-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution.

Attempt to fix by a JavaScript solution

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -5,7 +5,6 @@
   <title>rustup.rs - The Rust toolchain installer</title>
   <meta name="keywords" content="Rust, Rust programming language, rustlang, rust-lang, Mozilla Rust, rustup">
   <meta name="description" content="The Rust toolchain installer">
-
   <link rel="stylesheet" href="normalize.css">
   <link rel="stylesheet" href="rustup.css">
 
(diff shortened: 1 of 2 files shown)


emilyalbini:
Our CSP is defined in [`www/website_config.json`](https://github.com/rust-lang/rustup/blob/master/www/website_config.json), so eventual changes should be done there.

I'd prefer not to tweak the CSP: we intentionally made it as strict as possible, and this PR allows way more stuff than necessary (including `unsafe-inline` which kinda defeats the point of the CSP). If the goal is fixing the copy button not working, the better solution would be to move the `onclick` to `rustup.js`:

```js
var buttons = document.querySelectorAll(".copy-button");
for (int i = 0; i < buttons.length; i++) {
    buttons[i].addEventListener("click", handle_copy_button_click);
}
```

BeniCheni:
> the better solution would be to move the onclick to rustup.js

I've tried out @pietroalbini's suggestion, and tested locally using "test" keyword to validate the style & functionality of the copy button.

####  platform-instructions-unix

Untitled-unix

 ####  platform-instructions-win32

Untitled-win32


#### platform-instructions-win64

Untitled-win64

#### platform-instructions-unknown

Untitled-unkown

#### platform-instructions-default

Untitled-default

kinnison:
So, I personally have no reason to believe this won't work, providing @BeniCheni has checked it on both chrome and firefox, and via an HTTP service with a similar CSP to the one on rustup.rs.  @pietroalbini do you think this is OK?

emilyalbini:
Yeah this looks correct.

PR #2403, opened 2020-07-08
sync-dist: invalidate dev.rustup.rs during dev deploy

emilyalbini:
r? @kinnison


kinnison (approved):
I assume this is good, I'll merge it after the release so I don't have a messy history :D

PR #2405, opened 2020-07-08
test: add validate_host_triple in tests/cli-inst-interactive.rs

### LINKED ISSUE #2376: Improve error messages for unsupported platforms.

Hey there! I'm messing around with the rustup package in Void Linux. One of the platforms that we ship binaries for is `armv6l-musl`, which roughly corresponds to the `arm-unknown-linux-musleabihf` target. However, trying to run `rustup-init` on such a device results in the following error message:

```
$ rustup-init
error: Pre-checks for host and toolchain failed: Provided host 'arm-unknown-linux-musleabihf' couldn't be converted to partial triple
If you are unsure of suitable values, the 'stable' toolchain is the default.
Valid host triples look something like: arm-unknown-linux-musleabihf
```

The error here is simply that the Rust team doesn't ship compilers built for this architecture. This should be made clear, because the current error message is a bit cryptic.

**Notes**

#2168 has a similar error message for an unsupported platform.

BeniCheni:
Resolve #2376, per [this comment](https://github.com/rust-lang/rustup/issues/2376#issuecomment-647101399), using a whitelist based on `host_triple` in `fn inner()`.

https://github.com/rust-lang/rustup/blob/bb0fd7b5f56bf5afd030cc2a251bea0c973313ad/src/dist/dist.rs#L234

### DIFF
--- a/tests/cli-inst-interactive.rs
+++ b/tests/cli-inst-interactive.rs
@@ -91,6 +91,42 @@ fn smoke_test() {
     });
 }
 
+// validate host triple against a whitelist to support rustup-init
+#[test]
+fn validate_host_triple() {
+    let host_triple = &this_host_triple();
+    let host_triple_whitelist = [
+        String::from("arm-linux-androideabi"),
+        String::from("armv7-linux-androideabi"),
+        String::from("aarch64-linux-android"),
+        String::from("i686-linux-android"),
+        String::from("x86_64-linux-android"),
+        String::from("x86_64-unknown-linux-gnu"),
+        String::from("x86_64-unknown-linux-musl"),
+        String::from("i686-unknown-linux-gnu"),
+        String::from("mips-unknown-linux-gnu"),
+        String::from("mipsel-unknown-linux-gnu"),
+        String::from("mips64-unknown-linux-gnuabi64"),
+        String::from("mips64el-unknown-linux-gnuabi64"),
+        String::from("arm-unknown-linux-gnueabi"),
+        String::from("armv7-unknown-linux-gnueabihf"),
+        String::from("aarch64-unknown-linux-gnu"),
+        String::from("x86_64-apple-darwin"),
+        String::from("i686-apple-darwin"),
+        String::from("x86_64-unknown-freebsd"),
+        String::from("i686-unknown-freebsd"),
+        String::from("x86_64-unknown-openbsd"),
+        String::from("i686-unknown-openbsd"),
+        String::from("x86_64-unknown-netbsd"),
+        String::from("i686-unknown-netbsd"),
+        String::from("x86_64-unknown-dragonfly"),
+        String::from("x86_64-pc-windows-msvc"),
+        String::from("x86_64-pc-windows-gnu"),
+        String::from("i686-pc-windows-gnu"),
+    ];
+    assert_eq!(true, host_triple_whitelist.contains(host_triple));
+}
+
 #[test]
 fn update() {
     setup(&|config| {


kinnison:
I think it probably makes more sense to simply test that we can parse the triple we're building on.  Since that's the specific problem in the issue - that having built for a given host, we don't recognise that host.

tests/cli-inst-interactive.rs
```diff
@@ -91,6 +91,42 @@ fn smoke_test() {
     });
 }
 
+// validate host triple against a whitelist to support rustup-init
+#[test]
+fn validate_host_triple() {
+    let host_triple = &this_host_triple();
```
rbtcollins, line 97: don't test this_host_triple() - it is a test helper. Test from_build() - that is the function that is confused at source. We can't use from_host_or_build, as that will succeed during cross_builds, whereas from_build will return the bad data during cross_builds.

Validate it using ```PartialTargetTriple::new(&from_build())``` and we should have all we need to make these builds of rustup fail during build time.

Oh, and this should be in a unit test in dist.rs.

rbtcollins:
I don't think this really resolves the related ticket: it will stop such rustups passing self-test, but it won't make the errors more comprehensible

kinnison:
> I don't think this really resolves the related ticket: it will stop such rustups passing self-test, but it won't make the errors more comprehensible

The error which makes no sense is "I can't parse X as a valid triple.  Valid triples look like X" -- if you can't build a `rustup` which would get into this circumstance then at least this error can't happen.

*but* a unit test is not going to resolve this.  Consider the following scenario.

I, a user of platform A, cross-build rustup for platform B which works because B is supported by Rust as a target platform.
I cannot run the test suite because A cannot run binaries intended for B, and such binaries would include unit test binaries as well as the integration tests.
As such, I have no idea that when I run `rustup-init` on platform B, it's going to give the upsetting error above.

That's the scenario we need to be able to protect against if we can.

If we cannot, then it's acceptable to just have the test that Robert suggests, but in conjunction with a rework of the error message as well, and possibly a deliberate check of the `from_build()` parse right at the start of `rustup-init` which has a different an explicit error message about how the binary is unable to run.

rbtcollins:
mm, so we can put something in build.rs as well - basically the same thing I would say - but its a bit chicken and egg, because rustup's crate isn't built, so we need to replicate a chunk of code.

We got onto this particular path from your suggestion about a 'test' as a way to solve it.

But there is another more direct way, which is to change the error reporting, which is what I initially suggested.

kinnison:
I kinda feel like what we want is to factor `PartialTargetTriple` into a separate `.rs` file (currently it's in `src/dist/dist.rs`) which does not import anything from the `rustup` crate.  Then we can directly include that into `build.rs` to validate that the build target is a valid triple as far as rustup is concerned.  From there we can also improve the error messages, but if you can't *build* rustup such that `rustup-init` would give the error in the issue, then that's a good first step IMO.

rbtcollins:
I don't think that will help though, because PartialTargetTriple only checks for structural validity, and the target that started this all off is structurally valid.

kinnison:
Not valid - `musleabihf` is not in the valid environments list -- https://github.com/rust-lang/rustup/blob/master/src/dist/dist.rs#L107

This is my point - it can't be parsed as a partial triple as-is right now.

rbtcollins:
Ah, I had in my head that check being done elsewhere. Hmm, sure.

BeniCheni:
Hi there. Apologize that I've been busy with work and life in the July and August so far, that I couldn't allocate time to follow up with this PR.

I'll close the PR for now, in case there's a contributor interested in helping out #2376.

Will come back to this PR or checkout more rustup when work and life calm down in the future month.

PR #2407, opened 2020-07-09
test: Ensure we can read proxy params from environment

inejge:
Tests for:

1. Correctly retrieving the proxy `(host, port)` tuple from `$https_proxy`  or `$HTTPS_PROXY`

2. Prioritizing `$https_proxy` over `$HTTPS_PROXY`

Using a mutex to serialize environment access in the test function is not strictly necessary as long as there are no other tests running in parallel, but makes it easier to extend testing if the need arises.

### DIFF
--- a/download/tests/read-proxy-env.rs
+++ b/download/tests/read-proxy-env.rs
@@ -1,0 +1,30 @@
+#![cfg(feature = "reqwest-backend")]
+
+use std::env::{remove_var, set_var};
+
+use env_proxy::for_url;
+use url::Url;
+
+fn scrub_env() {
+    remove_var("http_proxy");
+    remove_var("https_proxy");
+    remove_var("HTTPS_PROXY");
+    remove_var("ftp_proxy");
+    remove_var("FTP_PROXY");
+    remove_var("all_proxy");
+    remove_var("ALL_PROXY");
+    remove_var("no_proxy");
+    remove_var("NO_PROXY");
+}
+
+// Tests for correctly retrieving the proxy (host, port) tuple from $https_proxy
+#[test]
+fn read_basic_proxy_params() {
+    scrub_env();
+    set_var("https_proxy", "http://proxy.example.com:8080");
+    let u = Url::parse("https://www.example.org").ok().unwrap();
+    assert_eq!(
+        for_url(&u).host_port(),
+        Some(("proxy.example.com".to_string(), 8080))
+    );
+}


inejge:
OK, slightly modified to deal with variable name case insensitivity on Windows. The reason for the macOS build failure is unclear to me.

kinnison:
That lock is only ever used in that test, so I'm not sure how useful it is.  Better would be to ensure we restore the environment after with a guard like the ones used in the integration tests, or else for `env_proxy` to have a way to be used where it uses provided environment values rather than reading the system env (like we're trying to do for `home`).

The restoration is important otherwise this will probably mess up developer boxes which set environment variables for proxying on Windows.

inejge:
> That lock is only ever used in that test, so I'm not sure how useful it is

No objection there, it can be removed.

> Better would be to ensure we restore the environment after with a guard like the ones used in the integration tests

Those guards restore `PATH` by changing it through registry operations, not by using `env::{set,remove}_var()`. The proxy environment test shouldn't affect either the command shell which invoked it, or the user's session.

kinnison:
Aha, I'm a little sensitive about environment variables in Windows as you might imagine :D  Okay.  (The macos failure was a GHA bogon btw)

download/tests/read-proxy-env.rs
```diff
@@ -0,0 +1,50 @@
...
+#[test]
+fn read_basic_proxy_params() {
+    let _l = LOCK.lock();
+    scrub_env();
+    set_var("https_proxy", "http://proxy.example.com:8080");
+    #[cfg(not(windows))]
+    set_var("HTTPS_PROXY", "http://proxy2.example.com:8080");
+    let u = Url::parse("https://www.example.org").ok().unwrap();
+    assert_eq!(
+        for_url(&u).host_port(),
```
rbtcollins, line 27: This is using the env_proxy function directly: I guess thats what broke, so we probably need to, but its a bit of a shame to be testing our dependencies :(. I haven't looked but if we wrap things up in higher order things maybe those should be checked instead/too.

download/tests/read-proxy-env.rs
```diff
@@ -0,0 +1,50 @@
+#![cfg(feature = "reqwest-backend")]
```
rbtcollins, line 1: Did this bug only affect one backend?

download/tests/read-proxy-env.rs
```diff
@@ -0,0 +1,50 @@
...
+
```
rbtcollins, line 2: separately, this is going to make a separate integration test binary; rather than a unit test. I think this is ok for now, but having it in a larger compilation unit would help prevent proliferation, but we'll need to integrate in with currentprocess() to do that - I haven't looked at env_proxy to see whats involved there.

An alternative approach would be to have a helper that will always be isolated (like this is), but simpler: don't scrub the environment or anything; just print host_port() to stdout; then when we run this helper binary, we can set the environment, allowing us to run different scenarios.

But I don't think we need that approach yet, since what we're doing here is really just a smoke test.

download/tests/read-proxy-env.rs
```diff
@@ -0,0 +1,50 @@
...
+use std::env::{remove_var, set_var};
+use std::sync::Mutex;
+
+use env_proxy::for_url;
+use lazy_static::lazy_static;
+use url::Url;
+
+lazy_static! {
+    static ref LOCK: Mutex<()> = Mutex::new(());
```
rbtcollins: I don't think the lock is of any use right now - we have one and only one test, in one test script, and nothing else is consulting it, so it isn't going to lock anything out: just ditch it.

download/tests/read-proxy-env.rs
```diff
@@ -0,0 +1,50 @@
...
+    remove_var("FTP_PROXY");
+    remove_var("all_proxy");
+    remove_var("ALL_PROXY");
+    remove_var("no_proxy");
+    remove_var("NO_PROXY");
+}
+
+// Tests for:
+// 1. Correctly retrieving the proxy (host, port) tuple from $https_proxy or $HTTPS_PROXY
+// 2. Prioritizing $https_proxy over $HTTPS_PROXY (except on Windows, where env vars
```
rbtcollins: I don't think point 2 is interesting: this is very much in the space of 'checking env_proxy internals' :  the point of this test is to make sure that urls hooks <-> env_proxy are setup properly in case of future dependency skew issues, not to test the definition of env-proxy.

download/tests/read-proxy-env.rs
inejge: Looking at the antecedents of #2399, it should be clear that `env_proxy`'s own logging and tests _did_ detect the breakage, but it wouldn't have helped here; the even higher order problem stemmed from my failure to publish the updated version in a timely manner.

I agree that a higher-level test would be architecturally cleaner, and could imagine spinning up a separate micro-proxy thread (could it be backed by `hyper`? Probably) and attempting to fetch a file through that. In any case, proxies are an important enough use case for `rustup` that I believe that they should be tested for in some manner.

download/tests/read-proxy-env.rs
inejge: Yes, cURL has its own proxy detection logic (which `env_proxy` emulates).

download/tests/read-proxy-env.rs
inejge: Ack, just checking `https_proxy` is probably enough.

inejge:
To summarize, the way forward (in the near term, for this PR) would be to: a) remove locking, b) limit the test to `$http_proxy`. I'd prefer doing it in a force push. Is that on the right track?

kinnison:
> To summarize, the way forward (in the near term, for this PR) would be to: a) remove locking, b) limit the test to `$http_proxy`. I'd prefer doing it in a force push. Is that on the right track?

In general, it's easier to review changes-in-response-to-review if you push new commits to do the change, though if you and @rbtcollins are in agreement on the way forward then jumping immediately to a force-push is fine.

I definitely prefer that a PR be rebased to be clean commits before merging though.

inejge:
Right, here is the updated test as an additional commit, then.

rbtcollins:
LGTM, please squash into a single logical commit to prep for merging

inejge:
Rebased and force-pushed.

PR #2408, opened 2020-07-09
Add rust-analyzer proxy

matklad:


kinnison:
I've added the proxy to the snapcraft bit too, I forgot to mention that earlier on discord.

kinnison (approved):

akurtakov:
Is this change visible for users on nightly channel?

rbtcollins:
We need to cut a release before this will become visible to rustup users.

akurtakov:
Any hope for new release?

kinnison:
I'm afraid with everything going on in the world, I've been too busy to cut a release.  I'll be looking to do so in the near future though.

akurtakov:
Too bad this one didn't make it to 1.23.0 release.

kinnison:
This was explicitly backed out for 1.23.0 because rust-analyzer isn't going to be in `stable` for a while.

akurtakov:
So there is not a chance to have the proxy for nightly only for now?

kinnison:
Proxies are installed into PATH so it's not possible to only provide a proxy for one toolchain, no.  When we have a closer timeline for the component making it to stable we'll put the proxy back in.

kinnison:
You *can* make the link yourself, and then `rust-analyzer +nightly` ought to work; but it's not ideal.

PR #2412, opened 2020-07-12
More unit-testification

rbtcollins:
Please see individual commits


rbtcollins:
@kinnison reviewed offline while github was down.

PR #2414, opened 2020-07-13
Fix path to rustup.exe and rustup-init.exe on self-update

rhysd:
While checking an error on GitHub Actions workflow in my project,

https://github.com/rhysd/wain/runs/865787365?check_suite_focus=true

I noticed the path to 'setup' file on Windows was not correct:

```
C:\Users\runneradmin\.cargo\bin/rustup-init.exe
```

I'm not sure this is the reason of the workflow failure, but this PR fixes the path as follows:

```
C:\Users\runneradmin\.cargo\bin\rustup-init.exe
```

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -34,7 +34,7 @@ use std::borrow::Cow;
 use std::env;
 use std::env::consts::EXE_SUFFIX;
 use std::fs;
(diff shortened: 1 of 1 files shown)


rbtcollins:
Thanks for this; it is not the cause of the error you have suffered: that is a file-in-use issue of some sort.

PR #2416, opened 2020-07-14
More unit testifications

rbtcollins:
See the individual commits please

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,5 +66,6 @@ pub mod fallback_settings;
 mod install;
 mod notifications;
 pub mod settings;
+pub mod test;
 pub mod toolchain;
 pub mod utils;
--- a/tests/cli-self-upd.rs
+++ b/tests/cli-self-upd.rs
@@ -494,62 +494,30 @@ fn when_cargo_home_is_the_default_write_path_specially() {
 
 #[test]
 #[cfg(windows)]
-fn install_adds_path() {
+/// Smoke test for end-to-end code connectivity of the installer path mgmt on windows.
+fn install_uninstall_affect_path() {
     setup(&|config| {
-        expect_ok(config, &["rustup-init", "-y"]);
-
         let path = config.cargodir.join("bin").to_string_lossy().to_string();
+
+        expect_ok(config, &["rustup-init", "-y"]);
         assert!(
             get_path().unwrap().contains(&path),
             format!("`{}` not in `{}`", get_path().unwrap(), &path)
         );
-    });
-}
 
-#[test]
-#[cfg(windows)]
-fn uninstall_removes_path() {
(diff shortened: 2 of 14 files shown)


tests/cli-self-upd.rs
```diff
@@ -494,32 +494,25 @@ fn when_cargo_home_is_the_default_write_path_specially() {
...
-
-        let path = config.cargodir.join("bin").to_string_lossy().to_string();
         assert!(!get_path().unwrap().contains(&path));
     });
 }
 
 #[test]
 #[cfg(unix)]
+/// This covers both windows and unix no-path-modification cases due to the code structure.
 fn install_doesnt_modify_path_if_passed_no_modify_path() {
```
workingjubilee, line 511: This test will not compile on Windows as it is set to be `#[cfg(unix)]`, and so provides no test coverage on Windows. This is observable: It is absent from the build logs as well.

rbtcollins: yes, I know. I think this is fine: we run windows and unix tests together in CI, so the test won't fail to run.

The code structure in the library means that if this test fails, we know that the platform specific do_add_to_path routines are being called inappropriately, and thats from the platform agnostic code - but because the routines that are being called are platform specific the only way we can detect this is by having some platform specific mechanism.

As this is a blackbox test, we can't drop a replacement trait in and mock it without compromising the binary that we build; a future unit test in self_update.rs might provide better coverage in a platform neutral way with a unit-test-specific trait object, but for now I think this is fine.

The trade off is that a windows developer that breaks this test will only find out in CI, which I think is ok.

workingjubilee: Fair enough.

rbtcollins: Oh, I should add, the equivalent platform specific test on the windows side has to modify global state - the windows registry - which is potentially disruptive, and not worth doing for an edge case test in the first place when other coverage means are available.

src/lib.rs
```diff
@@ -66,5 +66,7 @@ pub mod fallback_settings;
 mod install;
 mod notifications;
 pub mod settings;
+// #[cfg(test)]
```
kinnison: ?

kinnison (commented):
It looks consistent, modulo the -gnu failure case we need to resolve, but what's that comment all about?

src/lib.rs
rbtcollins: oh hmm, I think I wanted that to be cfg(test) but it can't be for the integration tests to use the test helpers, they need to be built normally and then dropped during link

kinnison (approved):
This all seems to make sense.

kinnison:
You are right that it took quite a kicking from some heavy boots, but I think this is good.

PR #2419, opened 2020-07-15
Adds Rust favicon

sl4m:
Adds favicons to minimally support Chrome, Firefox, and Safari browsers.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -7,6 +7,9 @@
   <meta name="description" content="The Rust toolchain installer">
   <link rel="stylesheet" href="normalize.css">
   <link rel="stylesheet" href="rustup.css">
+  <link rel="icon" type="image/png" sizes="32x32" href="https://www.rust-lang.org/static/images/favicon-32x32.png">
+  <link rel="icon" type="image/svg+xml" href="https://www.rust-lang.org/static/images/favicon.svg">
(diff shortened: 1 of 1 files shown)


kinnison:
Where did these image files come from?

sl4m:
> Where did these image files come from?

From https://github.com/rust-lang/www.rust-lang.org

kinnison:
Assuming we're allowed to cross-link with the relevant content policies etc, it may make more sense to just reference them from www.rust-lang.org then, rather than copying them over.

sl4m:
I've updated `index.html` to just reference assets from www.rust-lang.org

kinnison:
@pietroalbini What would it take to deploy this to dev- so that we can verify the favicons work like this?

emilyalbini:
@kinnison we need to do the dev promotion of a release.

kinnison:
hmm, not the easiest thing to do with a prospective change then

4ldas (approved):
Seems reasonable. However could add metatags for favicons and stuff too

kinnison:
I'm sorry it has taken so long to come back to this.  At worst if this doesn't work then it simply won't work, right?  Or will it cause browsers to complain to users?

emilyalbini:
If this does not work I don't think users will see an explicit error from browsers. Still, we should try this out before doing the prod promotion.

kinnison:
Okay, I'll merge this and we should remember to check it during the 1.24 release process

Issue #2334, opened 2020-05-17
rustup-init completion message issues


drguildo:
**Problem**

On completion, rustup-init prints:

![image](https://user-images.githubusercontent.com/9282/82148899-f430d980-984c-11ea-9d7d-449a0bd36760.png)

The line wrapping is a bit janky and also there is a backslash missing in the path `%USERPROFILE%.cargo\bin`, which should be `%USERPROFILE%\.cargo\bin`.

Output of `rustup --version`:

`rustup 1.21.1 (7832b2ebe 2019-12-20)`

Output of `rustup show`:

```
Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\Simon\.rustup

stable-x86_64-pc-windows-msvc (default)
rustc 1.43.1 (8d69840ab 2020-05-04)
```

VladKou:
The missing "\" is an issue with the current stable release. It was fixed in https://github.com/rust-lang/rustup/commit/cac1d3a53f554ecacb9bf873ff21cea7c3c08f36

Building rustup from source, and running it shows the expected string:

![image](https://user-images.githubusercontent.com/3132093/82162638-f8fc8a00-985a-11ea-88c3-76d1392e8e34.png)

PR #2336, opened 2020-05-17
Fixing janky line breaks in self_updater.rs

VladKou:
Fixes #2343 #2334   -  janky line breaks in the %USERPROFILE% path help string.

New line breaks:

image

PR #2420, opened 2020-07-16
Fixing janky line breaks in self_updater.rs

rbtcollins:
Fixing janky line breaks in the %USERPROFILE% path help string, as reported in #2334

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -168,9 +168,9 @@ macro_rules! post_install_msg_win {
     () => {
         r"# Rust is installed now. Great!
 
-To get started you need Cargo's bin directory ({cargo_home}\\bin) in your `PATH`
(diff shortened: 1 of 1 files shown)


kinnison:
Did this come from #2336 ?  if so, why not just update that MR.  As a rustup committer you should have had the rights.

rbtcollins:
> 
> 
> Did this come from #2336 ? if so, why not just update that MR. As a rustup committer you should have had the rights.

That capability is able to be turned on and off on a per PR basis... yes it is that PR rebased. I'm not sure how to update that PR.

kinnison:
If you can, it'll say near the bottom of the page (above the checks/merge box) 'Add more commits by pushing to ...'

kinnison:
OOI did this get done somewhere, or should we merge this?

PR #2423, opened 2020-07-16
Consolidate install smoke tests

rbtcollins:
These three smoke tests are not interestingly different: they lose
nothing by being a single test, but we gain by less time in serialised
tests.


kinnison (approved):
LGTM

kinnison:
Free to merge on green.

PR #2427, opened 2020-07-17
cli: exit gracefully when no subcommand

### LINKED ISSUE #2425: Rustup CLI displays "error" message with no text on run
**Problem**
After installing rustup on a fresh system I see an error message when running the rustup command without any arguments. There is no meaningful error message. However, the tool appears to work as expected so the error being triggered does not appear to be critical.

Screen Shot 2020-07-16 at 8 26 12 PM

**Steps**
1. `docker run -it debian:buster`
2. `apt update`
3. `apt install -y curl`
4. `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y`
5. `PATH=$PATH:$HOME/.cargo/bin`
6. `rustup`

**Possible Solution(s)**
I have not found a solution to this yet.

**Notes**

Output of `rustup --version`:
```
root@908b73dbcdbc:/# rustup --version
rustup 1.22.1 (b01adbbc3 2020-07-08)
```
Output of `rustup show`:
```
root@908b73dbcdbc:/# rustup show
rustup home:  /root/.rustup

rustc 1.45.0 (5c1f21c3b 2020-07-13)
```

leighmcculloch:
### What
Exit from rustup CLI gracefully when no subcommand is provided.

### Why
That appears to be the intention because the `SubcommandRequiredElseHelp` app setting is configured on the CLI, however that is not the behavior because the `main` function treats the error indicating no subcommand was provided as an error.

https://github.com/rust-lang/rustup/blob/22dc71adbd7e19de0090f991b3609e5f14910dcb/src/cli/rustup_mode.rs#L152-L159

Instead of exiting gracefully the word `error:` is printed with no error message:
Screen Shot 2020-07-17 at 12 22 44 AM

Fixes #2425 

### Known Limitations
I have not added tests for this change. I'm rather new to rust and this codebase and I could not find tests for this code to modify or change.

### DIFF
--- a/tests/cli-misc.rs
+++ b/tests/cli-misc.rs
@@ -4,6 +4,7 @@
 pub mod mock;
 
 use std::env::consts::EXE_SUFFIX;
+use std::str;
 
 use rustup::for_host;
 use rustup::test::this_host_triple;
@@ -129,8 +130,8 @@ fn show_works_with_dumb_term() {
     });
 }
 
-// Issue #140
-// Don't panic when `target`, `update` etc. are called without subcommands.
+// Issue #2425
+// Exit with error and help output when called without subcommand.
 #[test]
 fn subcommand_required_for_target() {
     setup(&|config| {
@@ -138,12 +139,13 @@ fn subcommand_required_for_target() {
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
         assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert_eq!(out.status.code().unwrap(), 1);
+        assert!(str::from_utf8(&out.stdout).unwrap().contains(&"USAGE"));
     });
 }
 
-// Issue #140
-// Don't panic when `target`, `update` etc. are called without subcommands.
+// Issue #2425
+// Exit with error and help output when called without subcommand.
 #[test]
 fn subcommand_required_for_toolchain() {
     setup(&|config| {
@@ -151,12 +153,13 @@ fn subcommand_required_for_toolchain() {
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
         assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert_eq!(out.status.code().unwrap(), 1);
+        assert!(str::from_utf8(&out.stdout).unwrap().contains(&"USAGE"));
     });
 }
 
-// Issue #140
-// Don't panic when `target`, `update` etc. are called without subcommands.
+// Issue #2425
+// Exit with error and help output when called without subcommand.
 #[test]
 fn subcommand_required_for_override() {
     setup(&|config| {
@@ -164,12 +167,13 @@ fn subcommand_required_for_override() {
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
         assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert_eq!(out.status.code().unwrap(), 1);
+        assert!(str::from_utf8(&out.stdout).unwrap().contains(&"USAGE"));
     });
 }
 
-// Issue #140
-// Don't panic when `target`, `update` etc. are called without subcommands.
+// Issue #2425
+// Exit with error and help output when called without subcommand.
 #[test]
 fn subcommand_required_for_self() {
     setup(&|config| {
@@ -177,7 +181,8 @@ fn subcommand_required_for_self() {
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
         assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert_eq!(out.status.code().unwrap(), 1);
+        assert!(str::from_utf8(&out.stdout).unwrap().contains(&"USAGE"));
     });
 }
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -64,6 +64,10 @@ pub fn main() -> Result<utils::ExitCode> {
             writeln!(process().stdout().lock(), "{}", e.message)?;
             return Ok(utils::ExitCode(0));
         }
+        Err(e) if e.kind == clap::ErrorKind::MissingArgumentOrSubcommand => {
+            writeln!(process().stdout().lock(), "{}", e.message)?;
+            return Ok(utils::ExitCode(1));
+        }
         Err(e) => Err(e),
     }?;
     let verbose = matches.is_present("verbose");


kinnison:
As noted here: https://github.com/rust-lang/rustup/issues/2425#issuecomment-659919859 I think that we might want to investigate why `clap` isn't returning `clap::ErrorKind::HelpDisplayed`

leighmcculloch:
I think `clap` is returning the correct error, it explicitly returns this error when this feature is enabled:

https://github.com/clap-rs/clap/blob/e8d46f403671ebdd14f4f530f82f6ba6fe9c78cc/src/parse/parser.rs#L750-L756

leighmcculloch:
`clap` does the same thing for the similar feature, `ArgRequiredElseHelp`:

https://github.com/clap-rs/clap/blob/e8d46f403671ebdd14f4f530f82f6ba6fe9c78cc/src/parse/validator.rs#L60-L68

So there's consistency in how it responds with these features.

kinnison:
Yes, but it would, by default, report these on stderr and exit with failure: https://github.com/clap-rs/clap/blob/v2-master/src/errors.rs#L385

leighmcculloch:
Interesting, that might be a separate issue though. I don't know.

It's worth noting that `ErrKind::MissingArgumentOrSubcommand` is not the what is returned if the subcommand is missing and `AppSettings::SubcommandRequiredElseHelp` is not used. In that case `ErrorKind::MissingSubcommand` is returned.

It does seem quite confusing.

kinnison:
You've found a difference in behaviour between 1.21 and 1.22 (the presence of `error:`) and possibly a bug in how clap-v2 handles this situation.  I'm still not convinced this is the right "fix" but I'm not going to reject it yet.  Would you care to open a bug on `clap` discussing this?

leighmcculloch:
@kinnison See https://github.com/clap-rs/clap/issues/2021.

kinnison:
Fantastic, let's wait and see what the clap people say about the behaviour.

leighmcculloch:
@kinnison According to https://github.com/clap-rs/clap/issues/2021#issuecomment-660499948 this would be a breaking change if changed and so if it is changed that will be limited to v3. `rustup` is using v2.

leighmcculloch:
Looks like I have some tests to fix, I'll get on that.

kinnison:
Cool, making sure that over all our behaviour is clean is the goal.  You may also want to rebase as I merged another test improvement from Robert earlier today.

kinnison:
I prefer rebases over merges, it makes for a clearer commit sequence to review.  However I can wait for now and we can clean this up later.  Are you sure that the changes in behaviour that your test changes imply are desired?  Are older rustup behaviours now being changed?  I ask because if we've broken how rustup used to behave then we need to be careful if anything might be used in scripting or similar.

leighmcculloch:
>I prefer rebases over merges, it makes for a clearer commit sequence to review.

I force pushed a squash of the commits rebased on master to make the history clean.

>Are you sure that the changes in behaviour that your test changes imply are desired? Are older rustup behaviours now being changed? I ask because if we've broken how rustup used to behave then we need to be careful if anything might be used in scripting or similar.

We could keep the error code return without the `error:` text, but that seems inconsistent. 

What do you recommend?

kinnison:
Sorry it took me so long to get back here.  Thank you for the squash, that makes it easier to read the commit.  As for behaviour, I personally would prefer that whatever the behaviour was in versions before we started this refactor (so 1.21 or earlier) were preserved for anything where scripting might be used to detect errors.  If it's *purely* in interactive behaviours then it's probably fine to change behaviour, though if the help text goes to stderr then it is opportunistic help and should have an exit code which reflects that (i.e. non-zero in some form).

I don't mind losing the `error:` text though.

tests/cli-misc.rs
```diff
@@ -129,55 +129,55 @@ fn show_works_with_dumb_term() {
...
+// Exit gracefully when called without subcommand.
 #[test]
 fn subcommand_required_for_target() {
     setup(&|config| {
         let mut cmd = clitools::cmd(config, "rustup", &["target"]);
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
-        assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert!(out.status.success());
```
rbtcollins: so this is saying that this is no longer an error from a scripting point of view. I'm not sure that that makes sense: it has after all not actually accomplished anything, and so for a scripter, having a non-zero exit status is useful, because this presumably happens due to forgetting to put some calculated $thing together properly.

tests/cli-misc.rs
```diff
@@ -129,55 +129,55 @@ fn show_works_with_dumb_term() {
...
+// Exit gracefully when called without subcommand.
 #[test]
 fn subcommand_required_for_toolchain() {
     setup(&|config| {
         let mut cmd = clitools::cmd(config, "rustup", &["toolchain"]);
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
-        assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert!(out.status.success());
```
rbtcollins: same

tests/cli-misc.rs
```diff
@@ -129,55 +129,55 @@ fn show_works_with_dumb_term() {
...
+// Exit gracefully when called without subcommand.
 #[test]
 fn subcommand_required_for_override() {
     setup(&|config| {
         let mut cmd = clitools::cmd(config, "rustup", &["override"]);
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
-        assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert!(out.status.success());
```
rbtcollins: same

tests/cli-misc.rs
```diff
@@ -129,55 +129,55 @@ fn show_works_with_dumb_term() {
...
+// Exit gracefully when called without subcommand.
 #[test]
 fn subcommand_required_for_self() {
     setup(&|config| {
         let mut cmd = clitools::cmd(config, "rustup", &["self"]);
         clitools::env(config, &mut cmd);
         let out = cmd.output().unwrap();
-        assert!(!out.status.success());
-        assert_ne!(out.status.code().unwrap(), 101);
+        assert!(out.status.success());
```
rbtcollins: same

tests/cli-misc.rs
leighmcculloch: I think this is logical, and some tools definitely return an exit code when no subcommands are provided, but Rust tooling doesn't seem to. For example, running `rustc` or `cargo` without subcommands prints the help output and returns a zero `0` exit code.

But I think for the reasons @kinnison shared [here](https://github.com/rust-lang/rustup/pull/2427#issuecomment-662286661), being more consistent with previous versions, is a strong reason to keep the non-zero exit code.

I'll update the PR to keep the exit code.

leighmcculloch:
@kinnison @rbtcollins The PR is updated to maintain the existing exit code when executed without subcommands, while losing the `error:` prefix on output.

rbtcollins:
Could you please rebase this to reflect the final changes being done, rather than the review-edit-cycle? Thanks!

leighmcculloch:
@rbtcollins The PR is updated with a force-push that squashes the commits into a single commit. The diff and the change are identical.

rbtcollins (approved):

PR #2428, opened 2020-07-17
Better tests for cargo_home creation

rbtcollins:
The blackbox test now tests only what it claims to test; and added a
unit test to test the specific function we expect to create the bins.


kinnison (approved):
Looks good to me. :+1:  merge on green.

PR #2433, opened 2020-07-20
Do not completely remove ~/.rustup/tmp and download

### LINKED ISSUE #2430: Please do not completely remove ~/.rustup/tmp
**Problem**

Hi.

Because rustup's toolchains etc. are very large, I have them in a separate filesystem from my home directory.  (The separate filesystem is not backed up.)  I have made a symlink `~/.rustup/toolchains` into that directory.  Because rustup likes to rename things from its `~/.rustup/tmp` to `toolchains`, I have a symlink for `~/.rustup/tmp` too.

But each time rustup finishes, it not only deletes the contents of tmp - it deletes tmp itself too.  Then the next run makes a new tmp as a directory, but now it is on the wrong filesystem and rustup fails.

**Steps**

1. mv .rustup/toolchains /some/other/fs/
2. ln -s /some/other/fs/toolchains .rustup/
3. mkdir /some/other/fs/tmp
4. ln -s /some/other/fs/tmp .rustup/
5. rustup update
6. Observe the tmp symlink is gone
7. nontrivial rustup update => failure, cannot rename

`error: could not rename component file from '/home/rustcargo/.rustup/tmp/yk0his763iqoxjd2_file' to '/home/rustcargo/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/components'`

**Possible Solution(s)**

I think the best fix would be to simply not delete `~/.rustup/tmp`.  (Regardless of whether it's a symlink or whatever.)

I don't want to move the whole of `~/.rustup` into the non-backed-up filesystem, because that includes the settings, which I would want to keep.

I tried working around the problem with a bind mount for `tmp` and rustup did this:
```
thread 'main' panicked at 'Unable to clean up /home/rustcargo/.rustup/tmp: Os { code: 16, kind: Other, message: "Device or resource busy" }', src/utils/utils.rs:632:13
```

**Notes**

Output of `rustup --version`: `rustup 1.22.1 (b01adbbc3 2020-07-08)`
Output of `rustup show`:
```
rustup home:  /home/rustcargo/.rustup

installed toolchains
--------------------

arm-unknown-linux-gnueabihf
x86_64-unknown-linux-gnu

active toolchain
----------------

rustc 1.46.0-nightly (346aec9b0 2020-07-11)
```

ijackson:
As discussed in #2430, it is better to keep the top-level directory.  It might be a mountpoint, or a symlink which the user wants to keep, or have unusual permissions.

<strike>I wasn't able to find an implementation of "remove contents of this directory but not the directory itself", so I have implemeted one. Maybe this should be in some other crate somewhere but it's not that huge so I just put it here.</strike>  I use the new `remove_dir_contents` in the `remove_dir_all` crate

This change changes the behaviour of `delete_dir_contents` to actually just remove the contents, like the name suggests.  There are two call sites: one to delete tmp and one to delete download.  This is correct behaviour for both.

Fixes: #2430

I have done ad-hoc testing for this, and rerun the main test suite.  If you want a formal fails-before-passes-after test, I could provide one.

### DIFF
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -637,7 +637,7 @@ where
 }
 
 pub fn delete_dir_contents(dir_path: &Path) {
-    match remove_dir_all::remove_dir_all(dir_path) {
+    match remove_dir_all::remove_dir_contents(dir_path) {
         Err(e) if e.kind() != io::ErrorKind::NotFound => {
             panic!("Unable to clean up {}: {:?}", dir_path.display(), e);
         }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1559,10 +1559,11 @@ dependencies = [
 
 [[package]]
 name = "remove_dir_all"
-version = "0.6.1"
+version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7b19f5c2df95a07275e7224924cc62f76f04525f4fda801473f85e325e81977"
+checksum = "882f368737489ea543bc5c340e6f3d34a28c39980bd9a979e47322b26f60ac40"
 dependencies = [
+ "libc",
  "log",
  "num_cpus",
  "rayon",
@@ -1727,7 +1728,7 @@ dependencies = [
  "pulldown-cmark",
  "rand 0.8.3",
  "regex",
- "remove_dir_all 0.6.1",
+ "remove_dir_all 0.7.0",
  "retry",
  "rs_tracing",
  "same-file",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ pgp = {version = "0.7", default-features = false}
 pulldown-cmark = {version = "0.8", default-features = false}
 rand = "0.8"
 regex = "1"
-remove_dir_all = "0.6.0"
+remove_dir_all = "0.7.0"
 same-file = "1"
 scopeguard = "1"
 semver = "0.11"


src/utils/utils.rs
```diff
@@ -627,12 +627,26 @@ where
 }
 
 pub fn delete_dir_contents(dir_path: &Path) {
-    match remove_dir_all::remove_dir_all(dir_path) {
-        Err(e) if e.kind() != io::ErrorKind::NotFound => {
-            panic!("Unable to clean up {}: {:?}", dir_path.display(), e);
+    // Deletes only the contents, and not the directory iteself (in
+    // case it's a mountpoint, or a symlink, or has unusual
+    // permissions which the user wants to preserve).
+    use anyhow::Context;
```
rbtcollins: please put this use at the top of the file

src/utils/utils.rs
```diff
@@ -627,12 +627,26 @@ where
 }
 
 pub fn delete_dir_contents(dir_path: &Path) {
-    match remove_dir_all::remove_dir_all(dir_path) {
-        Err(e) if e.kind() != io::ErrorKind::NotFound => {
-            panic!("Unable to clean up {}: {:?}", dir_path.display(), e);
+    // Deletes only the contents, and not the directory iteself (in
```
rbtcollins: I think this comment would make a lot of sense as a docstring - 
```
/// delete_dir_contents deletes ...
pub fn dele...
```

src/utils/utils.rs
```diff
@@ -627,12 +627,26 @@ where
...
+        let entries = match fs::read_dir(dir_path) {
+            Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(()),
+            r => r,
         }
-        _ => {}
-    }
+        .context("checking directory")?;
+        for entry in entries {
+            let entry_path = entry.context("reading directory")?.path();
+            remove_dir_all::remove_dir_all(&entry_path)
```
rbtcollins: rather than reading the dir here and recursing into remove_dir_all, please patch remove_dir_all directly; inspecting https://github.com/XAMPPRocky/remove_dir_all/blob/master/src/fs.rs#L36 will show you why.  I have commit and publish rights on it, so can ensure a good response time.

I'm slightly  in favour of a new method e.g. remove_dir_contents_all in that crate rather than a parameter to the existing function (but probably refactor the contents to use a common internal helper).

rbtcollins (changes requested):

ijackson:
Robert Collins writes ("Re: [rust-lang/rustup] Do not completely remove ~/.rustup/tmp and download (#2433)"):
> @rbtcollins requested changes on this pull request.

Hi.  Thanks for the review.  Your requested changes look good to me
and I'll go ahead and dothat.

I have a few opinions/questions about this comment:

> In src/utils/utils.rs:
> 
> >          }
> -        _ => {}
> -    }
> +        .context("checking directory")?;
> +        for entry in entries {
> +            let entry_path = entry.context("reading directory")?.path();
> +            remove_dir_all::remove_dir_all(&entry_path)
> 
> rather than reading the dir here and recursing into remove_dir_all, please
> patch remove_dir_all directly; inspecting https://github.com/XAMPPRocky/
> remove_dir_all/blob/master/src/fs.rs#L36 will show you why. I have commit and
> publish rights on it, so can ensure a good response time.

Thanks.  I'm very happy to make this change in the dependency.

Having read all that stuff in remove_dir_all, it seems to me that on
Windows we ought to remove each of the subdirectories separately.

That way when we move the contents to temporary names, the temporary
names are *inside* the toplevel directory (which we are not removing).
That will work better if the toplevel directory is a mount point or
junction or whatever these Windows things are called :-).

Do you agree ?

If so I will implement it that way.  I'll put this rationale in a
comment (or in the commit message if you prefer) for future reference.

> I'm slightly in favour of a new method e.g. remove_dir_contents_all in that
> crate rather than a parameter to the existing function (but probably refactor
> the contents to use a common internal helper).

I agree with your opinion that the public API should be a new
`remove_dir_contents_all` function.

I'll see what the innards come out looking like, but in any case I
will avoid duplication obviously, and do any refactoring needed to
support that.

Regards,
Ian.

-- 
Ian Jackson <ijackson@chiark.greenend.org.uk>   These opinions are my own.  

Pronouns: they/he.  If I emailed you from @fyvzl.net or @evade.org.uk,
that is a private address which bypasses my fierce spamfilter.

rbtcollins:
There's some additional prior art that may give possibly useful context on the fs behaviour https://github.com/cygwin/cygwin/blob/ad101bcb0f55f0eb1a9f60187f949c3decd855e4/winsup/cygwin/syscalls.cc#L675-L1064

>  Having read all that stuff in remove_dir_all, it seems to me that on Windows we ought to remove each of the subdirectories separately.

Do you mean something like 'base_dir should be the dir whose contents we are removing' ? Yes, that seems like an obvious choice since for this new function we don't have the same implied 'has permissions to delete the target' semantics.

ijackson:
Robert Collins writes ("Re: [rust-lang/rustup] Do not completely remove ~/.rustup/tmp and download (#2433)"):
> There's some additional prior art that may give possibly useful context on the
> fs behaviour https://github.com/cygwin/cygwin/blob/
> ad101bcb0f55f0eb1a9f60187f949c3decd855e4/winsup/cygwin/syscalls.cc#L675-L1064
> 
>     Having read all that stuff in remove_dir_all, it seems to me that on
>     Windows we ought to remove each of the subdirectories separately.
> 
> Do you mean something like 'base_dir should be the dir whose contents we are
> removing' ? Yes, that seems like an obvious choice since for this new function
> we don't have the same implied 'has permissions to delete the target'
> semantics.

Yes, precisely.

I'll look at the code again (quite tired now) but I suspect the most
convenient implementation of this going to look very similar to the
modified delete_dir_contents from my initial MR branch - it can call
remove_dir_all as a subroutine and then it will all DTRT.

Ian.

-- 
Ian Jackson <ijackson@chiark.greenend.org.uk>   These opinions are my own.  

Pronouns: they/he.  If I emailed you from @fyvzl.net or @evade.org.uk,
that is a private address which bypasses my fierce spamfilter.

J-Vin (approved):

J-Vin (approved):
👍 🤯

mbrubeck:
@J-Vin, please don't spam PRs with comments or reviews that don't add new information.  It's very distracting. Thanks!

J-Vin (approved):

mbrubeck:
@J-Vin, you need to stop “reviewing” issues that you don’t own. It generates a lot of spam to our project members. We have blocked your account, for now.  Please contact rust-mods@rust-lang.org if you have any questions about this.

rbtcollins:
@ijackson this is now ready to be updated to remove_dir_all 0.7.0 and the new functions you wrote.

ijackson:
@rbtcollins Thanks, now done :-).

kinnison (approved):
Looks good to me, if @rbtcollins is okay with this then we shall merge it after 1.24.2 release is done.

rbtcollins (approved):

PR #2435, opened 2020-07-22
Report error for unsupported platform on 404

pickfire:


rustup-init.sh
```diff
@@ -400,30 +402,50 @@ downloader() {
...
             else
-                curl --proto '=https' --tlsv1.2 --silent --show-error --fail --location "$1" --output "$2"
+                _err=$(curl --proto '=https' --tlsv1.2 --silent --show-error --fail --location "$1" --output "$2" 2>&1)
+                _status=$?
             fi
         fi
+        if [ -n "$_err" ]; then
+            echo "$_err" >&2
+            if echo "$_err" | grep -q 404$; then
+                err "unsupported platform $3"
```
kinnison:
```suggestion
                err "installer for platform '$3' not found, this may be unsupported"
```

rustup-init.sh
```diff
@@ -400,30 +402,50 @@ downloader() {
...
             else
-                wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2"
+                _err=$(wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2" 2>&1)
+                _status=$?
+            fi
+        fi
+        if [ -n "$_err" ]; then
+            echo "$_err" >&2
+            if echo "$_err" | grep -q 'ERROR 404: Not Found\.$'; then
+                err "unsupported platform $3"
```
kinnison: Similar suggestion to the above.

rustup-init.sh
```diff
@@ -400,30 +402,50 @@ downloader() {
...
+                _status=$?
             else
-                wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2"
+                _err=$(wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2" 2>&1)
+                _status=$?
+            fi
+        fi
+        if [ -n "$_err" ]; then
+            echo "$_err" >&2
+            if echo "$_err" | grep -q 'ERROR 404: Not Found\.$'; then
```
kinnison: Will `ERROR 404: Not Found.` me the same in non-english locales?

kinnison (changes requested):
I apologise for how long it took to get to this review; I've been somewhat busy with real-life.

rustup-init.sh
pickfire:
```suggestion
                err "installer for platform '$3' not found, this may be unsupported"
```

rustup-init.sh
pickfire: No, nice findings. I forgot that different locale may have different messages.

```
> wget --https-only --secure-protocol=TLSv1_2 "https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init" -O "/tmp/rustup-init"
--2020-07-27 16:09:57--  https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init
Loaded CA certificate '/etc/ssl/certs/ca-certificates.crt'
Resolving static.rust-lang.org (static.rust-lang.org)... 13.33.198.175, 13.33.198.232, 13.33.198.96, ...
Connecting to static.rust-lang.org (static.rust-lang.org)|13.33.198.175|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2020-07-27 16:09:58 ERROR 404: Not Found.

> env LANG=zh_CN.UTF-8 wget --https-only --secure-protocol=TLSv1_2 "https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init" -O "/tmp/rustup-init"
--2020-07-27 16:10:01--  https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init
已载入 CA 证书“/etc/ssl/certs/ca-certificates.crt”
正在解析主机 static.rust-lang.org (static.rust-lang.org)... 13.33.198.175, 13.33.198.232, 13.33.198.96, ...
正在连接 static.rust-lang.org (static.rust-lang.org)|13.33.198.175|:443... 已连接。
已发出 HTTP 请求，正在等待回应... 404 Not Found
2020-07-27 16:10:02 错误 404：Not Found。

> env LANG=ja_JP.UTF-8 wget --https-only --secure-protocol=TLSv1_2 "https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init" -O "/tmp/rustup-init"
--2020-07-27 16:10:05--  https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init
CA証明書 '/etc/ssl/certs/ca-certificates.crt' をロードしました
static.rust-lang.org (static.rust-lang.org) をDNSに問いあわせています... 13.33.198.175, 13.33.198.232, 13.33.198.96, ...
static.rust-lang.org (static.rust-lang.org)|13.33.198.175|:443 に接続しています... 接続しました。
HTTP による接続要求を送信しました、応答を待っています... 404 Not Found
2020-07-27 16:10:05 エラー 404: Not Found。

> env LANG=zh_TW.UTF-8 wget --https-only --secure-protocol=TLSv1_2 "https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init" -O "/tmp/rustup-init"
--2020-07-27 16:10:10--  https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init
Loaded CA certificate '/etc/ssl/certs/ca-certificates.crt'
Resolving static.rust-lang.org (static.rust-lang.org)... 13.33.198.175, 13.33.198.232, 13.33.198.96, ...
Connecting to static.rust-lang.org (static.rust-lang.org)|13.33.198.175|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2020-07-27 16:10:10 ERROR 404: Not Found.

> env LANG=zh_HK.UTF-8 wget --https-only --secure-protocol=TLSv1_2 "https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init" -O "/tmp/rustup-init"
--2020-07-27 16:10:19--  https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init
Loaded CA certificate '/etc/ssl/certs/ca-certificates.crt'
Resolving static.rust-lang.org (static.rust-lang.org)... 13.33.198.175, 13.33.198.232, 13.33.198.96, ...
Connecting to static.rust-lang.org (static.rust-lang.org)|13.33.198.175|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2020-07-27 16:10:19 ERROR 404: Not Found.

```

```suggestion
            if echo "$_err" | grep -q ' 404: Not Found\.$'; then
```

pickfire: Oh wait, the period is different. I guess we need to check the second last line `404 Not Found`. Luckily I checked against chinese, if it's latin I bet the period is the same. But not sure about this on arabic locale, I never used a right-to-left language except chinese and japanese (which they don't on screen). Maybe someone can test this with middle eastern languages?

rustup-init.sh
```diff
@@ -400,30 +402,50 @@ downloader() {
...
+                _status=$?
             else
-                wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2"
+                _err=$(wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2" 2>&1)
+                _status=$?
+            fi
+        fi
+        if [ -n "$_err" ]; then
+            echo "$_err" >&2
+            if echo "$_err" | grep -q ' 404: Not Found\.$'; then
```
pickfire:
```suggestion
            if echo "$_err" | grep -q ' 404 Not Found$'; then
```

rustup-init.sh
kinnison: I worry that the `Not Found` text happens to be what AWS returns too.  Is there a reason you didn't just grep for `404` like you do in the cURL cases?

rustup-init.sh
```diff
@@ -400,30 +402,50 @@ downloader() {
         get_ciphersuites_for_curl
         _ciphersuites="$RETVAL"
         if [ -n "$_ciphersuites" ]; then
-            curl --proto '=https' --tlsv1.2 --ciphers "$_ciphersuites" --silent --show-error --fail --location "$1" --output "$2"
+            _err=$(curl --proto '=https' --tlsv1.2 --ciphers "$_ciphersuites" --silent --show-error --fail --location "$1" --output "$2" 2>&1)
```
rbtcollins, line 405: this should use the `-w` option to control the output and get the http_code written in machine parseable form. See man curl.

rustup-init.sh
```diff
@@ -400,30 +402,50 @@ downloader() {
...
+                _status=$?
             else
-                wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2"
+                _err=$(wget --https-only --secure-protocol=TLSv1_2 "$1" -O "$2" 2>&1)
+                _status=$?
+            fi
+        fi
+        if [ -n "$_err" ]; then
+            echo "$_err" >&2
+            if echo "$_err" | grep -q ' 404 Not Found$'; then
```
rbtcollins, line 444: I would grep for `ERROR 404`.

kinnison: Would that also suffer from the localisation problem I pointed out?

rustup-init.sh
pickfire: Because I bet AWS won't change that.

rustup-init.sh
pickfire: Yes, but we still want to show the original error in human readable form in case it wasn't 404.

rustup-init.sh
pickfire: Yes, see https://github.com/rust-lang/rustup/pull/2435#discussion_r460714244. `错误 404` for simplified chinese and `エラー 404` for japanese so I bet other localization are there too.

rustup-init.sh
kinnison: Fair enough

rustup-init.sh
rbtcollins: You can include other components in the output with line breaks or whatever other framing you wish.

rustup-init.sh
rbtcollins: Ugh. sorry for the poor suggestion. Given wgets lack of a programmable interface I'm actually inclined to suggest dropping it as a supported downloader. It would make what we support simpler and there are no platforms I know of where curl is not available but wget is.

pickfire: No, I do encounter some platforms where wget is available but curl isn't, I don't think we need to drop support for something we already support.

rustup-init.sh
pickfire: I don't feel like adding some parsing from command messages, I am not sure any way to do this easily. @rbtcollins you could add that in another pull request if you want.

kinnison: Given the way the PR is set up, @rbtcollins ought to be able to push an example/mock-up to your branch for the PR, if he has time and energy.

rustup-init.sh
kinnison: I don't mind if we give slightly less good errors on platforms where we have to fall back to wget.

rbtcollins: Out of interest, what platforms are these? We added this support only 2 years ago in a PR with no attached bug report, that said 'linux systems have this often', but rustup had been around for many years at that point and there are literally no bug reports of it not working at that time. I'm not worried about less-good-errors, I'm worried about less-correct-behaviour.

rustup-init.sh
rbtcollins: I'm worried that what you're doing here is going to break in the future as the output isn't a stable format; curl has a stable format which is the -w format, grepping the non formatted output is just hoping that it happens to work on all curl versions everywhere.

rustup-init.sh
pickfire: Or maybe easier, why not just search for `-w 404` to match exactly 404 for both curl and wget? Will that be incorrect?

rbtcollins: Maybe? What would that look like?

pickfire: `curl $args | grep -w 404`

rustup-init.sh
pickfire: It's just grepping for 404, also I don't think it will change.

rustup-init.sh
kinnison: This still says "Not Found" which is a localisation issue.

pickfire: `404 Not Found` is not localized, did you try `env LANG=zh_CN.UTF-8 wget --https-only --secure-protocol=TLSv1_2 "https://static.rust-lang.org/rustup/dist/armv7-unknown-linux-musleabihf/rustup-init" -O "/tmp/rustup-init"` or something similar?

The line after that however is localized, `错误 404：Not Found。`

kinnison: IIRC we thought that the 'Not Found' was part of the AWS web frontend and thus subject to potential change, however if you think it's part of wget then I'm prepared to let it slide.

pickfire: I believe it's part of wget, I saw the same error message for different sites.

kinnison: Aha, okay

kinnison (approved):

PR #2437, opened 2020-07-22
Removes need for `local`, making script POSIX compliant.

mavenraven:
Removes need for using the `local` keyword that is supported by
a number of shells. Instead, we can create a new sub shell when
calling a given function, which creates a new process without
any shared memory with the rest of the shell script, in the same way
calling any other command does. For example,

```
#!/bin/sh
set -eu
a=1
b=1

mutate_yes() {
	a=2
}

mutate_no() {
	(
		b=2
	)
}

mutate_yes
mutate_no

printf "%s\n" "$a" # prints 2
printf "%s\n" "$b" # prints 1
```
See
https://unix.stackexchange.com/questions/138463/do-parentheses-really-put-the-command-in-a-subshell
for more information.

### DIFF
--- a/ci/actions-templates/centos-fmt-clippy-template.yaml
+++ b/ci/actions-templates/centos-fmt-clippy-template.yaml
@@ -79,7 +79,7 @@ jobs:
             sh ./ci/raw_init.sh
       - name: Run shell checks
         run: |
-          shellcheck -s dash -- rustup-init.sh
+          shellcheck rustup-init.sh
           git ls-files -- '*.sh' | xargs shellcheck -s dash -e SC1090
           git ls-files -- '*.bash' | xargs shellcheck -s bash -e SC1090
       - name: Run formatting checks
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -1,12 +1,8 @@
 #!/bin/sh
-# shellcheck shell=dash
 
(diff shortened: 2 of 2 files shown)


mavenraven:
Closed because I'm not going to care anymore when this get reviewed in 3 months from now.

kinnison:
> Closed because I'm not going to care anymore when this get reviewed in 3 months from now.

I'm a little saddened that this was your reaction to not receiving a review for a few days; but I understand if you don't have time or mind-space to wait.  If you decide to come back to this then hopefully I won't be trying to move my mother in law across the country the exact moment you submit a new PR.

mavenraven:
@kinnison It's not a reaction to not receiving a review. I don't care about rust or rustup at all. Not in a pejorative way, it's just not relevant to what I work on. I only made this PR because I was looking at a project that had a rust library as a dependency, and I noticed the bashisms in your script. So you're more than welcome to take this code [or not]! It's just not something that I care at all about investing any more time in.

kinnison:
I see, fair enough. We actually target POSIX + almquist features as /bin/sh because that's a reasonable assumable subset across our currently targetted platforms.  If we start to get platforms which are less well provided for then I will revisit this.  For now, I think the style using `local` is nicer.  Thank you for your efforts though.

mavenraven:
Yep, no problem!

PR #2438, opened 2020-07-23
Accept TOML format in rust-toolchain

### LINKED ISSUE #2350: specify required components in rust-toolchain file
I wanted to [pin clippy to specific nightly version](https://github.com/rust-lang/rust-clippy/issues/5561) (which will be updated every time the subtree of the rustc repo is synced into the clippy standalone repo) to make sure a `cargo build` will always be able to build clippy.

I wanted to use a `rust-toolchain` file but it turns out, `rustup` only downloads default components for the specified version.
Clippy needs the `rustc-dev` component but there is no way to tell rustup to install that via `rust-toolchain` rendering it pretty useless for that use case. :(

It would be useful to be able to specify required components in the roolchain file.

ebroto:
Closes #2350

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,7 @@ targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]
 ```
 
 If the TOML format is used, the `[toolchain]` section is mandatory,
-but all its entries are optional.
+and at least one property must be specified.
 
 The `rust-toolchain` file is suitable to check in to source control. 
 This file has to be encoded in US-ASCII (if you are on
--- a/README.md
+++ b/README.md
@@ -414,8 +414,8 @@ components = [ "rustfmt", "rustc-dev" ]
 targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]
 ```
 
-If the TOML format is used, the `[toolchain]` section and the `channel` entry
-are mandatory.
+If the TOML format is used, the `[toolchain]` section is mandatory,
+but all its entries are optional.
 
 The `rust-toolchain` file is suitable to check in to source control. 
 This file has to be encoded in US-ASCII (if you are on
--- a/README.md
+++ b/README.md
@@ -404,9 +404,21 @@ source repository. This is most often the case for nightly-only
 software that pins to a revision from the release archives.
 
 In these cases the toolchain can be named in the project's directory
-in a file called `rust-toolchain`, the content of which is the name of
-a single `rustup` toolchain, and which is suitable to check in to
-source control. This file has to be encoded in US-ASCII (if you are on
+in a file called `rust-toolchain`, the content of which is either the name of
+a single `rustup` toolchain, or a TOML file with the following layout:
+
+``` toml
+[toolchain]
+channel = "nightly-2020-07-10"
+components = [ "rustfmt", "rustc-dev" ]
+targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]
+```
+
+If the TOML format is used, the `[toolchain]` section and the `channel` entry
+are mandatory.
+
+The `rust-toolchain` file is suitable to check in to source control. 
+This file has to be encoded in US-ASCII (if you are on
 Windows, check the encoding and that it does not starts with a BOM).
 
 The toolchains named in this file have a more restricted form than
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -480,12 +480,12 @@ pub fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCode> {
             String::new()
         };
         let cwd = utils::current_dir()?;
-        let ovr_toolchain_name = if let Ok(Some((ovr_toolchain, _reason))) = cfg.find_override(&cwd)
-        {
-            ovr_toolchain.name().to_string()
-        } else {
-            String::new()
-        };
+        let ovr_toolchain_name =
+            if let Ok(Some((toolchain, _override_cfg, _reason))) = cfg.find_override(&cwd) {
+                toolchain.name().to_string()
+            } else {
+                String::new()
+            };
         for toolchain in toolchains {
             let if_default = if def_toolchain_name == &*toolchain {
                 " (default)"
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -480,12 +480,11 @@ pub fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCode> {
             String::new()
         };
         let cwd = utils::current_dir()?;
-        let ovr_toolchain_name =
-            if let Ok(Some((toolchain, _override_cfg, _reason))) = cfg.find_override(&cwd) {
-                toolchain.name().to_string()
-            } else {
-                String::new()
-            };
+        let ovr_toolchain_name = if let Ok(Some((toolchain, _reason))) = cfg.find_override(&cwd) {
+            toolchain.name().to_string()
+        } else {
+            String::new()
+        };
         for toolchain in toolchains {
             let if_default = if def_toolchain_name == &*toolchain {
                 " (default)"
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -786,7 +786,7 @@ fn default_(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
         }
 
         let cwd = utils::current_dir()?;
-        if let Some((toolchain, reason)) = cfg.find_override(&cwd)? {
+        if let Some((toolchain, _override_cfg, reason)) = cfg.find_override(&cwd)? {
             info!(
                 "note that the toolchain '{}' is currently in use ({})",
                 toolchain.name(),
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -786,7 +786,7 @@ fn default_(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
         }
 
         let cwd = utils::current_dir()?;
-        if let Some((toolchain, _override_cfg, reason)) = cfg.find_override(&cwd)? {
+        if let Some((toolchain, reason)) = cfg.find_override(&cwd)? {
             info!(
                 "note that the toolchain '{}' is currently in use ({})",
                 toolchain.name(),
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -16,7 +16,7 @@ pub fn main() -> Result<utils::ExitCode> {
         return self_update::self_replace();
     }
 
-    // Internal testament dump used during CI. Not for users.
+    // Internal testament dump used during CI.  Not for users.
     if arg1 == Some("--dump-testament") {
         common::dump_testament()?;
         return Ok(utils::ExitCode(0));
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -16,7 +16,7 @@ pub fn main() -> Result<utils::ExitCode> {
         return self_update::self_replace();
     }
 
-    // Internal testament dump used during CI.  Not for users.
+    // Internal testament dump used during CI. Not for users.
     if arg1 == Some("--dump-testament") {
         common::dump_testament()?;
         return Ok(utils::ExitCode(0));
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,12 +1,14 @@
 use std::borrow::Cow;
 use std::fmt::{self, Display};
 use std::io;
+use std::mem;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::FromStr;
 use std::sync::Arc;
 
 use pgp::{Deserializable, SignedPublicKey};
+use serde::Deserialize;
 
 use crate::dist::download::DownloadCfg;
 use crate::dist::{dist, temp};
@@ -17,6 +19,27 @@ use crate::process;
 use crate::settings::{Settings, SettingsFile, DEFAULT_METADATA_VERSION};
 use crate::toolchain::{DistributableToolchain, Toolchain, UpdateStatus};
 use crate::utils::utils;
+
+#[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+struct OverrideFile {
+    toolchain: ToolchainSection,
+}
+
+#[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+struct ToolchainSection {
+    channel: String,
+    components: Option<Vec<String>>,
+    targets: Option<Vec<String>>,
+}
+
+impl<T: Into<String>> From<T> for ToolchainSection {
+    fn from(channel: T) -> Self {
+        Self {
+            channel: channel.into(),
+            ..Default::default()
+        }
+    }
+}
 
 #[derive(Debug)]
 pub enum OverrideReason {
@@ -33,6 +56,21 @@ impl Display for OverrideReason {
             Self::CommandLine => write!(f, "overridden by +toolchain on the command line"),
             Self::OverrideDB(path) => write!(f, "directory override for '{}'", path.display()),
             Self::ToolchainFile(path) => write!(f, "overridden by '{}'", path.display()),
+        }
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct OverrideCfg {
+    components: Vec<String>,
+    targets: Vec<String>,
+}
+
+impl From<OverrideFile> for OverrideCfg {
+    fn from(file: OverrideFile) -> Self {
+        Self {
+            components: file.toolchain.components.unwrap_or_default(),
+            targets: file.toolchain.targets.unwrap_or_default(),
         }
     }
 }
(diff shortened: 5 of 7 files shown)


ebroto:
Not sure about the macOS CI failure as I can't see the logs :thinking:

rbtcollins:
Builder probably flaked, just run it again

README.md
```diff
@@ -404,9 +404,21 @@ source repository. This is most often the case for nightly-only
...
+a single `rustup` toolchain, or a TOML file with the following layout:
+
+``` toml
+[toolchain]
+channel = "nightly-2020-07-10"
+components = [ "rustfmt", "rustc-dev" ]
+targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]
+```
+
+If the TOML format is used, the `[toolchain]` section and the `channel` entry
```
kinnison: I wonder if we need `channel` to be mandatory -- people may want to use this new more flexible format just to add a target such as `wasm32-unknown-unknown` without stipulating a particular channel.

src/cli/common.rs
```diff
@@ -480,12 +480,12 @@ pub fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCode> {
...
         let cwd = utils::current_dir()?;
-        let ovr_toolchain_name = if let Ok(Some((ovr_toolchain, _reason))) = cfg.find_override(&cwd)
-        {
-            ovr_toolchain.name().to_string()
-        } else {
-            String::new()
-        };
+        let ovr_toolchain_name =
+            if let Ok(Some((toolchain, _override_cfg, _reason))) = cfg.find_override(&cwd) {
+                toolchain.name().to_string()
```
kinnison: This may need tweaking if you agree with my above comment and decide that the channel may be optional in the toml.

src/cli/rustup_mode.rs
```diff
@@ -786,7 +786,7 @@ fn default_(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
         }
 
         let cwd = utils::current_dir()?;
-        if let Some((toolchain, reason)) = cfg.find_override(&cwd)? {
+        if let Some((toolchain, _override_cfg, reason)) = cfg.find_override(&cwd)? {
```
kinnison: Ditto, may need tweaking.

src/cli/setup_mode.rs
```diff
@@ -16,7 +16,7 @@ pub fn main() -> Result<utils::ExitCode> {
         return self_update::self_replace();
     }
 
-    // Internal testament dump used during CI.  Not for users.
+    // Internal testament dump used during CI. Not for users.
```
kinnison: Whitespace bogon crept in, please get rid of this.

src/config.rs
```diff
@@ -18,6 +20,27 @@ use crate::settings::{Settings, SettingsFile, DEFAULT_METADATA_VERSION};
...
 use crate::utils::utils;
 
+#[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+struct OverrideFile {
+    toolchain: ToolchainSection,
+}
+
+#[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+struct ToolchainSection {
+    channel: String,
```
kinnison: My preference would be `channel: Option<String>` as mentioned above.

src/config.rs
```diff
@@ -37,6 +60,21 @@ impl Display for OverrideReason {
     }
 }
 
+#[derive(Debug, Default)]
+pub struct OverrideCfg {
```
kinnison: If we make channel optional as per the above, then this struct becomes less necessary perhaps?  Otherwise if you want to ensure that there remains API separation between the `rust-toolchain` format and the internal structs used, then this should probably contain the `channel: Option<Toolchain<'_>>` too.

src/config.rs
```diff
@@ -510,26 +559,74 @@ impl Cfg {
         Ok(None)
     }
 
+    fn parse_override_file<S: AsRef<str>>(contents: S) -> Result<Option<OverrideFile>> {
+        let contents = contents.as_ref();
+
+        if contents.lines().any(|line| line.starts_with("[toolchain]")) {
+            toml::from_str::<OverrideFile>(contents)
+                .map(|file| Some(file))
+                .map_err(|e| ErrorKind::ParsingOverride(e).into())
```
kinnison: If `channel` is made optional, here is where you'd need to verify that the constructed `OverrideFile` has at least one of `channel`, `components`, or `targets` specified.

tests/cli-rustup.rs
```diff
@@ -1402,6 +1402,174 @@ fn file_override_with_archive() {
...
+            r#"
+[toolchain]
+channel = "nightly-2015-01-01"
+components = [ "rust-src" ]
+"#,
+        )
+        .unwrap();
+
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-nightly-1");
+        expect_stdout_ok(
```
kinnison, line 1461: Ideally if `channel` were optional, this test would **a)** not change channel since it's purpose is to test that the component is added and **b)** ensure that the component you select is *NOT* installed before the toolchain file is written out.  (Note, this is kinda like the below test)

tests/cli-rustup.rs
```diff
@@ -1402,6 +1402,174 @@ fn file_override_with_archive() {
...
+            &["rustup", "component", "list"],
+            "rust-src (installed)",
+        );
+    });
+}
+
+#[test]
+fn file_override_toml_format_add_components_toolchain_misses_components() {
+    setup(&|config| {
+        expect_ok(config, &["rustup", "default", "stable"]);
```
kinnison, line 1472: Why not `rustup default nightly-2015-01-01` rather than installing two toolchains in this test?

tests/cli-rustup.rs
```diff
@@ -1402,6 +1402,174 @@ fn file_override_with_archive() {
...
+"#,
+        )
+        .unwrap();
+
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-nightly-1");
+        expect_stdout_ok(
+            config,
+            &["rustup", "component", "list"],
+            "rust-src (installed)",
+        );
```
kinnison, line 1494: All this does is confirm `rust-src` is available *afterwards* not that it also was not available *before* the toolchain file was written to disk.  Pleaes verify that the component is not installed beforehand.  otherwise changes in default profiles etc, may cause this test to become useless in a way that does not cause a test failure.

tests/cli-rustup.rs
```diff
@@ -1402,6 +1402,174 @@ fn file_override_with_archive() {
...
+            r#"
+[toolchain]
+channel = "nightly-2015-01-01"
+components = [ "rust-bongo" ]
+"#,
+        )
+        .unwrap();
+
+        // The toolchain should be installed and the invalid component skipped
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-nightly-1");
```
kinnison: Is it worth using a variant of this test which looks for the message that `rust-bongo` is skipped?

tests/cli-rustup.rs
```diff
@@ -1402,6 +1402,174 @@ fn file_override_with_archive() {
...
+            "rust-src (installed)",
+        );
+    });
+}
+
+#[test]
+fn file_override_toml_format_add_components_invalid_component() {
+    setup(&|config| {
+        expect_ok(config, &["rustup", "default", "stable"]);
+        expect_ok(
```
kinnison: Again, why installing two toolchains here?

tests/cli-rustup.rs
```diff
@@ -1402,6 +1402,174 @@ fn file_override_with_archive() {
...
+        // The toolchain should be installed and the invalid component skipped
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-nightly-1");
+    });
+}
+
+#[test]
+fn file_override_toml_format_add_components_toolchain_misses_target() {
+    setup(&|config| {
+        expect_ok(config, &["rustup", "default", "stable"]);
+        expect_ok(
```
kinnison: Again, it's not really worth installing both toolchains here.

tests/cli-rustup.rs
```diff
@@ -1402,6 +1402,174 @@ fn file_override_with_archive() {
...
+"#,
+        )
+        .unwrap();
+
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-nightly-1");
+        expect_stdout_ok(
+            config,
+            &["rustup", "component", "list"],
+            "arm-linux-androideabi (installed)",
+        );
```
kinnison: Again, please verify the target is not installed before you write the toolchain file out.

kinnison (commented):
Over-all I'm very impressed.  The single main concern I have is that I think it'd be beneficial to make `channel` optional so that people can use this new format to specify components/targets without constraining the channel of the toolchain.  With that, a bit of refactoring would be appropriate since the result of finding an override configuration might be that there's no toolchain override (though that might come from elsewhere) but there might be components/targets to install.

In addition, the integration tests could be tweaked to be more correct, less prone to risk of silently becoming worthless in the future, and possibly quicker to run, with the tweaks I mention in comments.

This is an excellent first pass.  What do you think about my 'optional `channel`' point?

ebroto:
I think making the channel optional is a great idea, thanks for the suggestion!

I've also tweaked the integration tests with your insights. I'll leave some comments to clarify some decisions I made.

src/config.rs
ebroto: I think decoupling them is a good idea as you mention. Also, I've realized that this struct does not need to be exposed outside the `config` module, so I kept the previous API of `find_override` returning the toolchain and the override reason.

tests/cli-rustup.rs
ebroto: I completely agree with `b)`, I've changed all the related tests to check that :+1:. 

With respect to `a)`, in this particular case I was trying to test that if the toolchain file specifies a toolchain that is not installed it would be installed before adding the component. Let me know if I misunderstood how rustup works with respect to this specific case. 

I've renamed and reordered a bit the tests to try to make my intent more explicit. Also I took advantage of the optional channel to reduce some test cases.

src/config.rs
kinnison: Sounds good.

tests/cli-rustup.rs
kinnison: It's fine that you want to verify the toolchain gets installed; in that context, simply don't install the other toolchain too.  Let's try and keep the test as fast as it can be.

kinnison (commented):
I think this is looking pretty good now.  @rbtcollins Do you have time to skim?

README.md
```diff
@@ -404,9 +404,21 @@ source repository. This is most often the case for nightly-only
...
+are mandatory.
```
rbtcollins: This prose needs updating.

src/config.rs
```diff
@@ -18,6 +19,35 @@ use crate::settings::{Settings, SettingsFile, DEFAULT_METADATA_VERSION};
...
+    targets: Option<Vec<String>>,
+}
+
+impl ToolchainSection {
+    fn is_empty(&self) -> bool {
+        self.channel.is_none() && self.components.is_none() && self.targets.is_none()
+    }
+}
+
+enum OverrideKind {
```
rbtcollins: I'm not sure we want to be passing around an enum to the rest of the code: I think we just want to upgrade the old serialisation format to the new model, and always pass around an OverrideFile.

src/config.rs
```diff
@@ -510,26 +588,90 @@ impl Cfg {
         Ok(None)
     }
 
+    fn parse_override_file<S: AsRef<str>>(contents: S) -> Result<Option<OverrideFile>> {
+        let contents = contents.as_ref();
+
+        if contents.lines().any(|line| line.starts_with("[toolchain]")) {
```
rbtcollins: I highly recommend not taking this approach, which is fragile: toml permits sections in any order, and any future changes we make to the schema will make this code more and more awkward.

What I suggest instead is to attempt to deserialise using serde directly. Because the two branches are serialised differently, you'll need to deserialise it twice rather than using serdes https://serde.rs/enum-representations.html#untagged feature. 

So try once, first, using the TOML struct, and toml format, and the second entry the plain string using https://docs.rs/crate/serde_plain/0.3.0.

That way you're not manually implementing any part of the underlying plumbing, and we can change it further without changing the deserialiser.

Also, this should give us utf8 support rather than US-ASCII.

rbtcollins (commented):
A few thoughts

src/config.rs
ebroto: @rbtcollins @kinnison following our discussion on Discord, I've applied this suggestion by avoiding to manually parse the TOML format, and showing the TOML parse error in addition to the "unknown channel" error if parsing as legacy format yields an invalid toolchain name. I've not used serde-plain because (correct me if I'm wrong) it would not respect the old behavior given that lines following the first one were simply ignored. I believe that same reason prevents trying to parse as legacy format before TOML.

I find this solution is not ideal, but it works. There's still one problem though, we can't control the error message for old rustups, as the contents of the first line could be anything. We can't require the first line to be anything specific if we want to avoid parsing TOML manually (which I think is wise advice, as e.g. `[ toolchain ]` is valid TOML).

All in all, I strongly feel that having a specific `rustup.toml` would get rid of all these problems, if I may propose it again at this stage (one last time, I promise :)). We can prevent having both files by erroring out so diverging contents should not be a concern. The only drawback I can think of is that it could make the integration tests slower, which I believe is currently a concern.

Let me know your thoughts.

src/config.rs
```diff
@@ -37,6 +73,26 @@ impl Display for OverrideReason {
...
+#[derive(Default)]
+struct OverrideCfg<'a> {
+    toolchain: Option<Toolchain<'a>>,
+    components: Vec<String>,
+    targets: Vec<String>,
+}
+
+impl<'a> OverrideCfg<'a> {
+    fn from_file(cfg: &'a Cfg, file: OverrideFile) -> Result<Self> {
+        Ok(Self {
```
rbtcollins, line 85: Does a result (None,None,None) make sense? what will happen when someone has such a config file - will it produce an error? I see the merit in permitting configs where at least one of the fields is not none, but with all none - do we want to treat that as an absent file, or as an error? @kinnison

src/config.rs
```diff
@@ -510,26 +589,97 @@ impl Cfg {
         Ok(None)
     }
 
+    fn parse_override_file<S: AsRef<str>>(
+        contents: S,
+    ) -> (Option<OverrideFile>, Option<ErrorKind>) {
+        let contents = contents.as_ref();
+
+        // Avoid TOML parsing error for backwards compatibility with the legacy format
+        if contents.is_empty() {
```
rbtcollins: This could go in the 'in case of error' block I think, or does lines() not return an empty line for an empty file?

src/config.rs
rbtcollins: the failure mode though is that older rustups will then behave as though there is no rustup file at all, which will be mysterious.

kinnison: I'd rather risk unusual but failing behaviour from an old `rustup` over silent miscompilation with the wrong toolchain, or unusual errors as a result.  Thus no, `rustup.toml` is not going to be a thing at this point.

The way I see it, a reasonable approach would be:

1. An empty `rust-toolchain` file is a specific error, perhaps "Empty `rust-toolchain` file detected.  Please remove it, or else specify the desired toolchain properties in the file".
2. A single-line `rust-toolchain` file is to be treated as the old-form, even if it's `[toolchain]` or similar, the resultant error will probably be obvious
3. Any other file content should be parsed-as-toml and errors reported as appropriate.

src/config.rs
ebroto: I went for treating this case as an empty file in the legacy format, which was not yielding errors.

kinnison: A file consisting of `[toolchain]` only (i.e. the `(None, None, None)` case) is an error IMO.  You must be specifying *something* to warrant a `rust-toolchain` file - similarly, an empty such file is worthless and thus an error as I mentioned in a previous comment.

src/config.rs
ebroto: You are right, in fact, I think I can remove the code completely as long as we ignore the toml error

src/config.rs
kinnison: (While previously an empty file was not an error, I think from now on it *should* be.  and I'm prepared to accept that as a breaking change if people view it as such.)

src/config.rs
ebroto: Alright, I totally understand the concerns about the old rustups, I'll implement the change as described.

Just to explain the way I understood it, is that pre-`rust-toolchain` rustups were ignoring the file too.

rbtcollins: Fair enough, but that was 3 years ago (June 2017), and we can't do anything about those versions: if someone on RHEL or Ubuntu 1604 or whatnot running an LTS version of rustup tries to build, they will get unexpected results, and we may have to support them; but we are at least not creating new silent failure modes: any newer rustup will fail verbosely.

src/config.rs
ebroto: This is not necessary anymore with the new parsing strategy.

src/config.rs
ebroto: With the latest version `OverrideFile` is guaranteed to have at least one `Some`.

rbtcollins (approved):

Issue #2424, opened 2020-07-16
Finish disentangling Windows and Unix logic


workingjubilee:
After #2342 there's still a few instances of the platform-specific code being entangled between the two OS platforms.

In some cases this is because of usage of [`cfg!(expr)`](https://doc.rust-lang.org/std/macro.cfg.html) which works very differently than `#[cfg(attr)]`. This causes useless object code to be compiled on Windows that cannot be properly tested and evaluated and in any case probably fails to even compile due to being dependent on some hidden platform-targeted code. There might also be other increases in compilation times as a result of poor separation of logic. So, ideally, the Windows and Unix logic would be cleanly separated, or at least enough so that neither platform compiles useless, impossible to test, and possibly-erroneous codepaths on the other.

I likely will follow up on this myself but I'm filing this issue for self-reference, because several instances of editing code that was supposed to only affect Unix platforms and then mysteriously breaking Windows in CI (and not because I broke generally-applicable code) was a notable part of my experience with #2387.
```[tasklist]
### Tasks
- [ ] https://github.com/rust-lang/rustup/issues/3936#issuecomment-2222224266
- [ ] https://github.com/rust-lang/rustup/pull/3887
```

djc:
https://github.com/rust-lang/rustup/pull/3887 made some progress on this, too.

PR #2453, opened 2020-08-10
Refactor self_update* to compile less OS-specific code

workingjubilee:
All of this, even removing the complete_windows_uninstall function from self_update/unix.rs is effectively just rearranging things (a function that does nothing but panic was replaced with an inlined cfg_if! panic). So there are no functional changes, it should remove object code from some compilations but not alter the logic of the program. As mentioned in #2424, the main purpose of this is to minimize events where changing OS-dependent code for one OS affects another OS.

This is a followup on but is not the end of #2453 currently because there is a lot of code that is still logically dependent on OS that has no particular reason to be in self_update.rs over self_update/{windows,unix}.rs. In a few cases it makes sense to have such code present, but in a lot of cases it doesn't. A total review of such code is beyond the scope of this PR.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -124,7 +124,7 @@ these changes will be reverted.
     };
 }
 
-#[cfg(unix)]
+#[cfg(not(windows))]
 macro_rules! pre_install_msg_unix {
     () => {
         pre_install_msg_template!(
@@ -155,6 +155,7 @@ but will not be added automatically."
     };
 }
 
+#[cfg(not(windows))]
 macro_rules! post_install_msg_unix {
     () => {
         r"# Rust is installed now. Great!
@@ -168,6 +169,7 @@ To configure your current shell run `source {cargo_home}/env`
     };
 }
 
+#[cfg(windows)]
 macro_rules! post_install_msg_win {
     () => {
         r"# Rust is installed now. Great!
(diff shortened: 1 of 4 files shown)


workingjubilee:
Linux errors are spurious, on Snapcraft net failures. :shrug:

Down to only two cfg! exprs in self_update remaining.

workingjubilee:
One.

workingjubilee:
Zero.

src/cli/self_update.rs
```diff
@@ -189,6 +195,7 @@ To configure your current shell run `source {cargo_home}/env`
     };
 }
 
+#[cfg(windows)]
```
rbtcollins, line 199: would you like this reviewed, or are you planning on moving these now windows only things into the windows self-install module too ?

workingjubilee: This is where it currently stands. Moving things around further is out-of-scope. I only performed what seemed like the most obvious changes.

workingjubilee: A review would be appreciated. In general, I am trying to deliberately block myself from pushing further changes in to a diff (because I know it's a bad habit). I would rather split the work and submit a followup than have the entire thing turned down, and I would rather have less things to rebase.

src/cli/self_update.rs
```diff
@@ -58,14 +58,16 @@ mod shell;
 mod unix;
 #[cfg(windows)]
 mod windows;
+#[cfg(windows)]
+pub use windows::complete_windows_uninstall;
```
rbtcollins: This probably shouldn't be interleaved with the mod definitions; It should just sit in the same-crate imports; I think I tweaked this recently, but if it isn't conflicting, don't worry about it.

rbtcollins (approved):
Looks fine

PR #2455, opened 2020-08-13
Test cleanup of rustup-init functional tests

rbtcollins:
Run all rustup-init tests that can in parallel.

Use common helpers where possible.

Move some tests to more relevant modules.

Fixes test path leakages on Windows I think.


rbtcollins:
Treating the thumbs up on discord as a +1 here :).

PR #2459, opened 2020-08-14
fix feature-hidden clippy warnings (clippy::len_zero)

matthiaskrgr:
found with cargo clippy .. --all-features

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -126,7 +126,7 @@ struct Bash;
 
 impl UnixShell for Bash {
     fn does_exist(&self) -> bool {
-        self.update_rcs().len() > 0
+        !self.update_rcs().is_empty()
     }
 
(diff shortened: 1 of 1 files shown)


rbtcollins:
Thank you. I think there are some in the optional curl downloader as well, but I haven't managed to replicate that locally yet for some reason. I'll merge this when it passes CI.

rbtcollins:
Ah, we need to add --all

matthiaskrgr:
Hm,I checked with `--all-targets --all-features` and this was all I found :D

rbtcollins:
yeah, `cargo +nightly clippy --all --all-targets --all-features -- -D warnings` is the full thing; --all-features will be tricky for Windows developers as installing OS libraries can be fiddly compared to the ease of apt install libcurl-dev etc.

rbtcollins:
https://github.com/rust-lang/rustup/pull/2460 follows this up with additional coverage

PR #2464, opened 2020-08-19
Lower-case unpack ram log message

palango:
All other log messages start with a lower-case word, so adapt the unpac ram message accordingly,


kinnison (approved):
This does improve consistency.  Thank you.

Issue #2447, opened 2020-08-05
Confusion between rustup version and rustc version


kornelski:
Rustup has reached version 1.22, which is high enough that it starts to look like an old rustc version.

To people who are just starting with Rust, it can be confusing. If someone is told they need "Rust 1.31" and they run `rustup --version`, they may think they've got an outdated version of Rust.

I see a couple of ways to improve this situation:

* Bump rustup's version to track rustc version (e.g. Cargo does this). I know rustup isn't tied to rustc version, so these two could be out of sync, but `rustup update` updates itself too, so they should stay more or less in sync in typical usage.

* Make `rustup --version` also display `rustc --version`

kinnison:
Currently `--version` is handled by `clap` I guess we could perhaps just turn it from `--version` being handled by `clap` into treating it more like `rustup show` perhaps?

kinnison:
I've added the help-wanted label because I'd like to see ideas of nice ways to make this less confusing for people without needing to do a rustup release on every rust release just to keep numbers "aligned"

killercup:
Strawman proposal for a simple fix: Overwrite the output for `--verbose` to something like

```
$ rustup --version
rustup 1.23.0 (8932b2fcf 2020-09-20)
NOTE: This is the version for the rustup toolchain manager, not the rustc compiler.
NOTE: The currently active rustc version is `rustc 1.42.0 (b8cedc004 2020-03-09)`
```

(where the last part is calling out to `rustc --version`.)

kinnison:
Yes, that'd be a nice halfway house between the current output and a full on `rustup show` perhaps.

kinnison:
Rustup 1.23.0 is now out.

codemajesty999:
rustup version now is already 1.26.0

simona70:
I have rustup with rustc 1.65 but many of the software require version 1.75... what should I do?

ChrisDenton:
try `rustup install stable`

simona70:
"rustup update" did all the work. but I'm not used to updating outside of Pacman

ChrisDenton:
You would have to talk to your distro about that. Rustup has no control over system packages.

simona70:
yes, thx. good work.

PR #2465, opened 2020-08-19
Quick addition to `--version` to mention rustc

killercup:
As discussed in #2447, rustup versions are getting closer to rustc
versions (because the rustup maintainers are working too much I'm sure).
This leads to confusion when running `rustup --version` while intending
to run `rustc --version`.

The change introduced here will print two additional (stderr) lines
after rustup's default `--version` output. The first mentions this
issue, and the second will print the output of `rustc --version` as a
hint (by trivially calling out to `rustc`, whatever binary that may be).

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -75,12 +75,10 @@ pub fn main() -> Result<utils::ExitCode> {
             info!("This is the version for the rustup toolchain manager, not the rustc compiler.");
 
             fn rustc_version() -> std::result::Result<String, Box<dyn std::error::Error>> {
-                let cmd = Command::new("rustc").arg("--version").output()?;
-                if cmd.status.success() {
-                    Ok(String::from_utf8_lossy(&cmd.stdout).trim().into())
-                } else {
-                    Err(String::from_utf8_lossy(&cmd.stderr).into())
-                }
+                let cfg = &mut common::set_globals(false, true)?;
+                let cwd = std::env::current_dir()?;
+                let toolchain = cfg.find_or_install_override_toolchain_or_default(&cwd)?.0;
+                Ok(toolchain.rustc_version())
             }
 
             match rustc_version() {
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -55,17 +55,46 @@ where
 pub fn main() -> Result<utils::ExitCode> {
     self_update::cleanup_self_updater()?;
 
+    use clap::ErrorKind::*;
     let matches = match cli().get_matches_from_safe(process().args_os()) {
         Ok(matches) => Ok(matches),
-        Err(e)
-            if e.kind == clap::ErrorKind::HelpDisplayed
-                || e.kind == clap::ErrorKind::VersionDisplayed =>
-        {
-            writeln!(process().stdout().lock(), "{}", e.message)?;
+        Err(clap::Error {
+            kind: HelpDisplayed,
+            message,
+            ..
+        }) => {
+            writeln!(process().stdout().lock(), "{}", message)?;
             return Ok(utils::ExitCode(0));
         }
-        Err(e) if e.kind == clap::ErrorKind::MissingArgumentOrSubcommand => {
-            writeln!(process().stdout().lock(), "{}", e.message)?;
+        Err(clap::Error {
+            kind: VersionDisplayed,
+            message,
+            ..
+        }) => {
+            writeln!(process().stdout().lock(), "{}", message)?;
+            info!("This is the version for the rustup toolchain manager, not the rustc compiler.");
+
+            fn rustc_version() -> std::result::Result<String, Box<dyn std::error::Error>> {
+                let cmd = Command::new("rustc").arg("--version").output()?;
+                if cmd.status.success() {
+                    Ok(String::from_utf8_lossy(&cmd.stdout).trim().into())
+                } else {
+                    Err(String::from_utf8_lossy(&cmd.stderr).into())
+                }
+            }
+
+            match rustc_version() {
+                Ok(version) => info!("The currently active `rustc` version is `{}`", version),
+                Err(err) => debug!("Wanted to tell you the current rustc version, too, but ran into this error: {}", err),
+            }
+            return Ok(utils::ExitCode(0));
+        }
+        Err(clap::Error {
+            kind: MissingArgumentOrSubcommand,
+            message,
+            ..
+        }) => {
+            writeln!(process().stdout().lock(), "{}", message)?;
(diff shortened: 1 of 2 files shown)


src/cli/rustup_mode.rs
```diff
@@ -55,17 +55,46 @@ where
...
+            }
             return Ok(utils::ExitCode(0));
         }
-        Err(e) if e.kind == clap::ErrorKind::MissingArgumentOrSubcommand => {
-            writeln!(process().stdout().lock(), "{}", e.message)?;
+        Err(clap::Error {
+            kind: MissingArgumentOrSubcommand,
+            message,
+            ..
+        }) => {
```
killercup, line 101: Changed this to pattern matching for consistency with match arms above it.

src/cli/rustup_mode.rs
```diff
@@ -55,17 +55,46 @@ where
...
+            writeln!(process().stdout().lock(), "{}", message)?;
+            return Ok(utils::ExitCode(0));
+        }
+        Err(clap::Error {
+            kind: VersionDisplayed,
+            message,
+            ..
+        }) => {
+            writeln!(process().stdout().lock(), "{}", message)?;
+            info!("This is the version for the rustup toolchain manager, not the rustc compiler.");
```
killercup, line 75: I hope using the log macros is fine; not sure what the `process().stdout().lock()` dance does here.

killercup (commented):
r? @kinnison

For your entertainment I want to also quote this snippet from my testing:

```console
$ env RUSTUP_FORCE_ARG0=rustup $CARGO_TARGET_DIR/debug/rustup-init --version
   Compiling rustup v1.22.1 (/home/pascal/code/rust-lang/rustup)
    Finished dev [unoptimized + debuginfo] target(s) in 9.35s
rustup 1.22.1+103 (000079b11 2020-08-19) dirty 3 modifications
info: The currently active `rustc` version is `rustup 1.22.1 (1980-01-01)`
```

(Renaming the binary to `rustup` instead of setting the env var fixes this of course.)

src/cli/rustup_mode.rs
rbtcollins: https://github.com/rust-lang/rustup/blob/master/CONTRIBUTING.md#no-direct-use-of-process-state-outside-rustupcurrentprocess

info! does the dance for you, as long as its info! from our macros, not a new import or some such.

rbtcollins:
> (Renaming the binary to rustup instead of setting the env var fixes this of course.)

That suggests to me that you're invoking the proxy rather than grabbing the current active toolchain and invoking rustc via it, which won't have this defect at all, and will be measurably faster on windows, where process creation isn't as cheap as it is on Unix systems.

src/cli/rustup_mode.rs
```diff
@@ -55,17 +55,46 @@ where
...
+
+            fn rustc_version() -> std::result::Result<String, Box<dyn std::error::Error>> {
+                let cmd = Command::new("rustc").arg("--version").output()?;
+                if cmd.status.success() {
+                    Ok(String::from_utf8_lossy(&cmd.stdout).trim().into())
+                } else {
+                    Err(String::from_utf8_lossy(&cmd.stderr).into())
+                }
+            }
```
kinnison: You don't want to use this.  Instead you want to find the default toolchain for the CWD and use `toolchain.rustc_version()`

Look at `Cfg::find_or_install_override_toolchain_or_default` for that.

killercup: I thought about trying to use this function, but was confused how to do it without building up the whole chain. Will look again.

Just curious: Is there any downside to doing this aside from potential performance? (I'm unsure where an additional 0.1s for `--version` would matter.) Arguably, calling `rustc` as I do would also work for broken rustup installs.

src/cli/rustup_mode.rs
killercup: Yep, it's the already imported rustup macro. Thanks for the info! (pun intended)

src/cli/rustup_mode.rs
kinnison: The downside is exactly as you spotted basically, also `rustup +nightly --version` will work if you look up the Cfg properly :D

src/cli/rustup_mode.rs
```diff
@@ -55,17 +55,44 @@ where
...
+            message,
+            ..
+        }) => {
+            writeln!(process().stdout().lock(), "{}", message)?;
+            info!("This is the version for the rustup toolchain manager, not the rustc compiler.");
+
+            fn rustc_version() -> std::result::Result<String, Box<dyn std::error::Error>> {
+                let cfg = &mut common::set_globals(false, true)?;
+                let cwd = std::env::current_dir()?;
+                let toolchain = cfg.find_or_install_override_toolchain_or_default(&cwd)?.0;
```
kinnison, line 86: Unless I'm mistaken, I think this may need to take into account if a `+toolchain` was specified. but I'm not in a position to check right now,

src/cli/rustup_mode.rs
kinnison: :+1:

src/cli/rustup_mode.rs
rbtcollins: `rustup +nightly --version` not working would be surprising to people I think

killercup: True, just added support for that as well. Had to do it in a slightly hack-y way (`process().args().find(|x| x.starts_with('+'))`) since we are in the match arm where clap opted out of parsing the args, but it should be fine for this rather simple edge-case.

src/cli/rustup_mode.rs
```diff
@@ -77,7 +77,14 @@ pub fn main() -> Result<utils::ExitCode> {
             fn rustc_version() -> std::result::Result<String, Box<dyn std::error::Error>> {
                 let cfg = &mut common::set_globals(false, true)?;
                 let cwd = std::env::current_dir()?;
+
+                if let Some(t) = process().args().find(|x| x.starts_with('+')) {
+                    debug!("Fetching rustc version from toolchain `{}`", t);
+                    cfg.set_toolchain_override(&t[1..]);
```
kinnison, line 83: It's really sad that this can't be done with the `matches` because we don't get them back if version is displayed :(

kinnison:
Functionalitywise this looks fine, it'll be worth ensuring in a test that this works though.  Probably just a cli-misc test.

killercup:
Test added. Don't know how granular you test these kinds of things, so I hope the one simple test case is fine. I can always add more tests for edge cases if you want :)

kinnison (approved):
I think this is good enough for now.  If we get unusual bug reports for `--version` we'll revisit more testing.

Issue #2409, opened 2020-07-09
reqwest download backend is lacking socks support


Sherlock-Holo:
**Problem**

**Steps**

1. run `https_proxy=socks5://127.0.0.1:1990 rustup update` and it doesn't use proxy because I watch `iftop -B` and nothing connect to my proxy server
2. use proxychians and see `iftop` it connects to my proxy server

kinnison:
@inejge any ideas?

inejge:
Support for `SOCKS` proxies in `reqwest` is behind the `socks` feature, which (AFAICS) isn't on by default, nor is it enabled in `download`'s `Cargo.toml`; `git blame` points to af3ba46. My testing of `env_proxy` shows that it passes `socks5://...` through.

kinnison:
Okay, sounds like we need to turn that on again :D  Thanks.

gaveen:
@kinnison I'd like to work on this since it's marked both 'easy' and 'mentor'. How can I get help on starting with it?

kinnison:
@gaveen Firstly I suggest that you get a local socks proxy working so that you can confirm if requests are passing through it.  Set up the `https_proxy` variable in the manner suggested in the issue description; and then confirm that `RUSTUP_USE_CURL=1 rustup check` uses the proxy but that `rustup check` does not.  After that you'll need to fork and clone rustup; and look at what it'll take to turn on the socks support in reqwest.  Likely this will involve changing the `download/Cargo.toml` file to enable the requisite feature.

You can come onto the rust-lang discord and find me in `#wg-rustup` if you want more discussion, or else respond here.

PR #2466, opened 2020-08-20
Add socks5 support for reqwest-backend

eidenar:
As discussed in #2409, rustup doesn't use socks5 proxy with reqwest backend. It turned out, that enabling socks feature in Cargo.toml is enough to make it work as intended.

There was no updates from the last person worked on this issue for almost a month, so I decided to pick it up.

### DIFF
--- a/download/tests/read-proxy-env.rs
+++ b/download/tests/read-proxy-env.rs
@@ -1,8 +1,14 @@
 #![cfg(feature = "reqwest-backend")]
 
 use std::env::{remove_var, set_var};
+use std::error::Error;
+use std::net::TcpListener;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::thread;
+use std::time::Duration;
 
 use env_proxy::for_url;
+use reqwest::{blocking::Client, Proxy};
 use url::Url;
 
 fn scrub_env() {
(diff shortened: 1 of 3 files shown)


kinnison:
I wonder if there's any way we can add a test to ensure that we don't accidentally lose this capability?

eidenar:
@kinnison I added a test to check if socks5 feature is enabled. It performs a simple request with non-existent socks5 address from `all_proxy` variable and checks call count. Not sure thought if there is a easier way. But I ran this test without socks feature and it failed as expected.

download/tests/read-proxy-env.rs
```diff
@@ -28,3 +31,37 @@ fn read_basic_proxy_params() {
...
+        .proxy(Proxy::custom(env_proxy))
+        .timeout(Duration::from_secs(1))
+        .build()
+        .unwrap();
+
+    let res = client.get(url.as_str()).send();
+    println!("Called {}", CALL_COUNT.load(Ordering::SeqCst));
+    if let Err(e) = res {
+        let s = e.source().unwrap();
+        // Only 2 calls performed in case of lack socks5 support
```
kinnison: This number feels very fragile to me.  Could we instead create a socket, bind and listen, and then set the proxy to that socket and check that it tries to connect to it?

eidenar: I've modified test to actually check if proxy connects to the socket. But I noticed that sometimes tests can overlap and mess around with env variables. I see 3 options to fix this:
1. Use `serial_test` [crate](https://docs.rs/serial_test/0.4.0/serial_test/) to make those 2 tests simultaneous
2. Use Mutex for the same purpose
3. Or just make tests use different variables

Not sure though which is the best approach in this case

kinnison: Individual integration tests ought to be being run in isolation; though on Windows environment changes are moderately awkward.  We have some other tests for environment in the download crate and they use a mutex to serialise them, perhaps move the test to be like those?

kinnison: I have made a change which I believe will work.

kinnison:
I've rebased this (as `Cargo.lock` was in conflict)

kinnison (changes requested):
At this point I'm satisfied with the shape of this, but @eidenar needs to double-check and then rebase the series so that we have a nice clean commit series of the change and the test, without the corrections.

eidenar:
I've updated commits and messages to be clear, @kinnison.

kinnison (approved):
This looks good to me, thank you for your contribution.

PR #2467, opened 2020-08-25
Update to remove_dir_all 0.6.0

rbtcollins:


kinnison (approved):
Merge on green CI

Cargo.lock
```diff
@@ -1855,7 +1960,7 @@ dependencies = [
  "libc",
  "rand",
  "redox_syscall",
- "remove_dir_all",
+ "remove_dir_all 0.5.3",
```
tesuji, line 1963: So tempfile still use old version ?

rbtcollins: Yes. We will likely get further benefits when it upgrades. However, we use it directly in the removal of components; tempfile is only used in the test suite. Even with the use limited to the direct component usage test suite performance is improved up to under 7m with this patch for me.

Issue #2458, opened 2020-08-14
Allow specifying a custom toolchain in `rust-toolchain`?


anp:
A project at work uses prebuilt toolchain binaries from internal infrastructure. We currently provide [instructions](https://fuchsia.dev/fuchsia-src/development/languages/rust/cargo#cargo-config-gen) on how to link that location with rustup for use with cargo and other tools. Ideally we'd be able to specify that relationship in version control and have rustup 'just work' when run within our directory tree.

I see that [TOML support](https://github.com/rust-lang/rustup/pull/2438) in the `rust-toolchain` file was recently merged (super exciting for other reasons too!) and I'm wondering whether it could be used to specify a "local" custom toolchain for a repository. Maybe something like this?

```toml
[toolchain]
path = "prebuilts/.../bin"
```

kinnison:
Nominally this is semi-supported by the `rustup toolchain link ...` pathways.  However to make this particular use-case function would require the creation of a temporary toolchain whose path is not inside the rustup toolchain paths, not certain how complex that would be.

jonhoo:
This would come in extremely handy for me as well, though for a fairly different use-case. In my case, I have a larger build system that sometimes needs to "fiddle some bits" before `cargo`/`rustc` gets to run. Currently, the way that works is that the user has to invoke the "main" build system, which then internally invokes `cargo`. This works, but means that tools that aren't aware of the larger build system (like `rust-analyzer`) do not invoke the build system, which may lead to new dependencies not being available and such.

One way I've tried to hack around this problem is to tell users to install a custom system-wide toolchain that's really just a thing wrapper around the default toolchain, and which invokes the "fiddle" part of the main build system when necessary before forwarding to cargo. That approach works ok, but it a) requires that users know about and install this system-wide toolchain, and b) it's now a non-local configuration for building projects, which can make debugging build problems harder.

But, with this feature, we could do _much_ better. The first time you run the build system, it could set up a `rust-toolchain` for the project that points at a project-local toolchain that's the aforementioned wrapper. Then, when the user later runs `cargo`, `rustc`, or any other command, they'll automatically be routed through the wrapper script (and thus through the main build system) without having to do any special or system-wide setup!

kinnison:
[...]

PR #2471, opened 2020-08-29
WIP: Thoughts on #2458

kinnison:
This is me starting to think about fixing #2458

I'm mostly putting it up so we can discuss the approach and see if it's sufficient.  Right now it's not tested.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -33,6 +33,7 @@ impl OverrideFile {
 #[derive(Debug, Default, Deserialize, PartialEq, Eq)]
 struct ToolchainSection {
     channel: Option<String>,
+    path: Option<String>,
     components: Option<Vec<String>>,
     targets: Option<Vec<String>>,
 }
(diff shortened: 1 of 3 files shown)


workingjubilee:
So, hm, at least 1 problem:
1. The rust-toolchain file is specified as "[yes, this has to be USASCII.](https://rust-lang.github.io/rustup/overrides.html?highlight=ASCII#the-toolchain-file)" But path names are not necessarily USASCII. USASCII is a subset of UTF8 however, so we're almost OK. But Windows paths are very not UTF8 by default. And I know what a BOM is but many of your users don't.

2. Further, I think a "path" and a "channel" describe what is mostly the same concept ("what point in timespace am I getting my Rust toolchain from?") and this should be clarified somewhat.

For instance, I can see users requesting "how about I use `rust-toolchain` to set my own local interpretation of both `stable` and `nightly` so that I can use $proxy +stable or $proxy +nightly?" That can be denied, of course, but if it is, then I think we should still have a principled approach here, I guess is what I am saying? So if that doesn't work, then we should assign an ident to `rust-toolchain`'s override like `+local` or `+path`.

kinnison:
I think the intent is that the USASCII encoding requirement will be lifted for the `toml` form of the `rust-toolchain` because that assumes utf8 encoding anyway AIUI.  Given that, I suppose we could restrict the path form to the `toml` rather than the single-line variant; however the single-line variant gains (potentially un-advertised) support by dint of supporting `cargo +/path/too/rust build` type constructs which I would want to support.

Paths and channels are subtly different because channel is something which can update over time, whereas path is just some bytes on a filesystem.

Thanks for your input though, it's food for thought, for sure.

kinnison:
Obsoleted by #2678

PR #2475, opened 2020-08-31
Use backticks in CONTRIBUTING.md

pickfire:
Suggested by @lzutao

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,7 +88,7 @@ Similarly, there are many Linux and Unix operating systems images available for
 developers whose usual operating system is Windows. Currently Rustup has no Mac
 OS specific code, so there should be no need to worry about Mac VM images.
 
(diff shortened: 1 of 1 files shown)


kinnison (approved):
Good catch.  I wonder if there's a sensible markdown lint we could run as part of the general checks CI job?

tesuji:
> sensible markdown lint

Clippy [uses remarkjs](https://github.com/rust-lang/rust-clippy/blob/71a5379bda71f19539dabac0232330561e19be1a/.remarkrc), but it will include nodejs to build setup.

kinnison:
I'd rather not add that kind of dependency in unless it's quick to install, but it is something to consider, thanks.

tesuji:
It seems really fast to install on github actions: https://github.com/rust-lang/rust-clippy/runs/1065196865?check_suite_focus=true
But I'm not sure remarkjs could catch issue like this, because I've never used it locally.

pickfire:
Heh, time to rewrite it in rust.

PR #2476, opened 2020-09-01
List unused toolchains

pwoolcoc:
This adds an `--unused` flag to `rustup toolchain list` that will only
list toolchains that aren't part of an override. `stable`, `beta`, and
`nightly` are never considered "unused" regardless of whether they are
in the overrides list

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -1,5 +1,6 @@
 //! Just a dumping ground for cli stuff
 
+use std::collections::HashSet;
 use std::fs;
 use std::io::{BufRead, ErrorKind, Write};
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


pwoolcoc:
example usage:

```
$ rustup toolchain list --unused | xargs rustup toolchain uninstall
info: uninstalling toolchain '1.39.0-x86_64-apple-darwin'
info: toolchain '1.39.0-x86_64-apple-darwin' uninstalled
info: uninstalling toolchain '1.41.0-x86_64-apple-darwin'
info: toolchain '1.41.0-x86_64-apple-darwin' uninstalled
info: uninstalling toolchain '1.42.0-x86_64-apple-darwin'
info: toolchain '1.42.0-x86_64-apple-darwin' uninstalled
info: uninstalling toolchain '1.44.0-x86_64-apple-darwin'
info: toolchain '1.44.0-x86_64-apple-darwin' uninstalled
```

kinnison:
This is an interesting idea, but I wonder how it interacts with users who have `rust-toolchain` files rather than setting overrides through `rustup override` directly.

Would it potentially be interesting instead to maintain a timestamp of when a toolchain was last used, and offer a time-based cleanup?

pwoolcoc:
Ah, yea, I totally forgot about `rust-toolchain` files. I'll go back to the drawing board on this, thanks for the review @kinnison !

kinnison:
I am glad you're not discouraged by the need to think a little more about this.  Feel free to create an issue to gather your ideas and solicit ideas from the community.  Then we can keep track (and link to this closed PR for reference too)

PR #2478, opened 2020-09-03
Deploy docs on stable only.

ehuss:
This changes it so that docs are only updated during a stable release. This ensures users don't see unreleased changes, which can be confusing.

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -3,7 +3,7 @@ name: Deploy Docs
 on:
   push:
     branches:
-      - master
+      - stable
 
 jobs:
   doc:


ehuss:
#2477 should be merged before this, so that the published docs will get updated.  After this is merged, I can follow up with a PR to revert #2477 (already prepared at https://github.com/ehuss/rustup/tree/add-unreleased-docs).

Or, if you don't want to bother with that dance (and you plan to do a new release soon), feel free to just close #2477.

rbtcollins:
I think it would be nice to have a published copy of the master tree
somewhere, like rust docs does of nightly,  but stable should definitely be
the default folk encounter

ehuss:
Sure!  Are you thinking just a different URL, like https://rust-lang.github.io/rustup/master/?  And leave the root for stable?

I can probably put something like that together, though it'll take a little time.

kinnison:
I'd love it if there were a `/devel/` for the current master branch (We may rename the branch, so don't encode `master` in the URL)

ehuss:
Sounds good!  I'll try to put something together.  The initial approach might be a little awkward since the docs don't yet exist on the stable branch.

rbtcollins:
I'd be OK merging a point update to stable of just doc updates if that
would be helpful.

kinnison:
I've merged the revert docs PR, and I'll merge this one after the docs have generated.  Then I'm fine waiting until a release before we try for more complexity.

kinnison:
OK docs have built so I'm merging this.  I look forward to the docs reinstatement PR @ehuss

PR #2486, opened 2020-09-12
Update RUSTUP_UNPACK_RAM docs to reflect code

nashley:
Match the minimum to `MAX_FILE_SIZE`'s value in `src/dist/component/package.rs`
Also attempt to describe the default behavior

### DIFF
--- a/doc/src/environment-variables.md
+++ b/doc/src/environment-variables.md
@@ -28,7 +28,7 @@
   determines the directory that traces will be written too. Traces are of the
   form PID.trace. Traces can be read by the Catapult project [tracing viewer].
 
-- `RUSTUP_UNPACK_RAM` *unstable* (default 400M, min 100M) Caps the amount of
+- `RUSTUP_UNPACK_RAM` *unstable* (default auto or 500M if on a platform such as FreeBSD or NetBSD, min 210M) Caps the amount of
   RAM `rustup` will use for IO tasks while unpacking.
 
 - `RUSTUP_NO_BACKTRACE` Disables backtraces on non-panic errors even when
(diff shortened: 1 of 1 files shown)


nashley:
I'm not sure if the explanation of the default behavior is sufficient or desired, and since it's not a static number, it's difficult to explain it succinctly. 🤔

nashley:
[Minimum value in source](https://github.com/rust-lang/rustup/blob/2d019878da283747476f6a82b0ba6d0c3b68eb67/src/dist/component/package.rs#L300)

Default behavior in source described [here](https://github.com/rust-lang/rustup/blob/2d019878da283747476f6a82b0ba6d0c3b68eb67/src/dist/component/package.rs#L301) and [here](https://github.com/rust-lang/rustup/blob/2d019878da283747476f6a82b0ba6d0c3b68eb67/src/dist/component/package.rs#L175)

doc/src/environment-variables.md
```diff
@@ -28,7 +28,7 @@
   determines the directory that traces will be written too. Traces are of the
   form PID.trace. Traces can be read by the Catapult project [tracing viewer].
 
-- `RUSTUP_UNPACK_RAM` *unstable* (default 400M, min 100M) Caps the amount of
+- `RUSTUP_UNPACK_RAM` *unstable* (default auto or 500M if on a platform such as FreeBSD or NetBSD, min 210M) Caps the amount of
```
nashley: I specified `210M` since `220000000/1024^2=209.80835` and the previous documentation seemed to refer to `M` as a unit of `mebibytes` instead of `megabytes`, but perhaps it'd be better to refer to the more human-friendly `megabytes` instead.

nashley: I think I'll change this to explicitly `MB` instead of using the ambiguous `M` to refer to `MiB`

kinnison (approved):

PR #2487, opened 2020-09-13
Give a better error message when a component isn't installed for a custom toolchain

### LINKED ISSUE #2470: Give a better error message when a component is missing with a linked toolchain
If I try to use a component from a tool that is currently building with `x.py`, rustup currently gives the following error:

```
$ rustdoc +stage1
error: 'rustdoc' is not installed for the toolchain 'stage1'
To install, run `rustup component add rustc --toolchain stage1`
```
The second line is not helpful: when you run it it will always give an error.
```
$ rustup component add rustc --toolchain stage1
error: stage1 is a custom toolchain
```

Instead it should suggest building the tool with x.py:
```
To install, run `x.py build src/tools/rustdoc`.
```

**Notes**

Ideally, rustup would know about compile stages and suggest `--stage 0` or `--stage 2` as necessary. However, if the user is using a local toolchain they should know about stages already, so it's ok to let them figure out for themselves which stage they want.

jyn514:
Closes https://github.com/rust-lang/rustup/issues/2470.

I did the thing you told me not to do which is mention x.py :P let me know if I should remove it.

I'm not sure how to run a custom rustup without installing it on top of my existing install ... `cargo run ` gives me
```
warning: it looks like you have an existing installation of Rust at:
warning: /home/joshua/.local/lib/cargo/bin
warning: rustup should not be installed alongside Rust. Please uninstall your existing Rust first.
warning: Otherwise you may have confusion unless you are careful with your PATH
warning: If you are sure that you want both rustup and your already installed Rust
warning: then please reply `y' or `yes' or set RUSTUP_INIT_SKIP_PATH_CHECK to yes
warning: or pass `-y' to ignore all ignorable checks.
```

Is there a way to test this by hand or do you only have unit tests? I want to see what the output looks like.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,10 +1,10 @@
 #![allow(clippy::large_enum_variant)]
 #![allow(deprecated)] // because of `Error::description` deprecation in `error_chain`
 
-use crate::component_for_bin;
 use crate::dist::dist::Profile;
 use crate::dist::manifest::{Component, Manifest};
 use crate::dist::temp;
+use crate::{component_for_bin, Toolchain};
 use error_chain::error_chain;
 use std::ffi::OsString;
 use std::io::{self, Write};
@@ -444,6 +444,10 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
 }
 
 fn install_msg(bin: &str, toolchain: &str, is_default: bool) -> String {
+    if Toolchain::is_custom_name(toolchain) {
+        return "\nnote: this is a custom toolchain, which cannot use `rustup component add`\n\
+        help: if you built this toolchain from source, and used `rustup component link`, then you may be able to build the component with `x.py`".to_string();
+    }
     match component_for_bin(bin) {
         Some(c) => format!("\nTo install, run `rustup component add {}{}`", c, {
             if is_default {
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -150,8 +150,13 @@ impl<'a> Toolchain<'a> {
 
     // Custom only
     pub fn is_custom(&self) -> bool {
-        ToolchainDesc::from_str(&self.name).is_err()
+        Toolchain::is_custom_name(&self.name)
     }
+
+    pub(crate) fn is_custom_name(name: &str) -> bool {
+        ToolchainDesc::from_str(name).is_err()
+    }
+
     // Distributable only
     pub fn is_tracking(&self) -> bool {
         ToolchainDesc::from_str(&self.name)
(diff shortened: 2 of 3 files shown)


kinnison:
Personally I do `cargo run -- --no-modify-path -y` but it sounds like you have an odd setup there.

I think `CONTRIBUTING.md` carries some instructions on setting up a custom cargo home.

src/errors.rs
```diff
@@ -1,10 +1,10 @@
 #![allow(clippy::large_enum_variant)]
 #![allow(deprecated)] // because of `Error::description` deprecation in `error_chain`
 
-use crate::component_for_bin;
 use crate::dist::dist::Profile;
 use crate::dist::manifest::{Component, Manifest};
 use crate::dist::temp;
+use crate::{component_for_bin, Toolchain};
```
kinnison, line 7: Why was this reordered?

src/errors.rs
```diff
@@ -444,6 +444,10 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
 }
 
 fn install_msg(bin: &str, toolchain: &str, is_default: bool) -> String {
+    if Toolchain::is_custom_name(toolchain) {
+        return format!("\nnote: this is a custom toolchain, which cannot use `rustup component add`\n\
+        help: if you installed this toolchain from source with `rustup component link`, you can build the component with `x.py`");
```
kinnison:
```suggestion
        help: if you built this toolchain from source, and used `rustup component link`, then you may be able to build the component with `x.py`");
```

Basically make it less "yes you can do this"

kinnison (changes requested):
This is pretty close to spot on.  I'll even forgive the `x.py` providing you tweak the wording a bit.

src/errors.rs
jyn514: Oops, this is left over from my failed attempt to make `Toolchain` part of `BinaryNotFound`. I reverted the change.

jyn514: Hmm, it looks like rustfmt really wants me to re-order it:
```
Run cargo fmt --all -- --check
Diff in /home/runner/work/rustup/rustup/src/errors.rs at line 1:
 #![allow(clippy::large_enum_variant)]
 #![allow(deprecated)] // because of `Error::description` deprecation in `error_chain`
 
-use crate::{component_for_bin, Toolchain};
 use crate::dist::dist::Profile;
 use crate::dist::manifest::{Component, Manifest};
 use crate::dist::temp;
Diff in /home/runner/work/rustup/rustup/src/errors.rs at line 8:
+use crate::{component_for_bin, Toolchain};
 use error_chain::error_chain;
 use std::ffi::OsString;
 use std::io::{self, Write};
##[error]Process completed with exit code 1.
```
Seems like a rustfmt bug? Adding a `{` shouldn't change the order.

kinnison: I guess `{` sorts after `d` for rustfmt :(

kinnison (changes requested):
Looks good, please rebase to remove the response commit now I've checked it.

jyn514:
Rebased :)

kinnison (approved):

PR #2491, opened 2020-09-17
Improve 'defaulting to ... unpack ram' message

### LINKED ISSUE #2484: Consider dropping "info: Defaulting to 500.0 MiB unpack ram" for rustup commands.

When I run `rustup component add ...`, I get this line at the end:

    info: Defaulting to 500.0 MiB unpack ram

As discussed at [What does rustup mean by "Defaulting to 500\.0 MiB unpack ram"? \- help \- The Rust Programming Language Forum](https://users.rust-lang.org/t/what-does-rustup-mean-by-defaulting-to-500-0-mib-unpack-ram/48528/4), it appears that the message isn't of much value to the end user, and other values for the amount of RAM available for unpacking crates may not be very relevant.

Either drop the message, saving users' cognitive load, or document why the message is useful to the end user, e.g. under what circumstances the user might want to change the RAM available for unpacking.

**Notes**

```
$ rustup --version
rustup 1.22.1 (b01adbbc3 2020-07-08)
```

camelid:
Fixes #2484.

This makes it more self-explanatory.

### DIFF
--- a/src/utils/notifications.rs
+++ b/src/utils/notifications.rs
@@ -87,7 +87,7 @@ impl<'a> Display for Notification<'a> {
             ),
             SetDefaultBufferSize(size) => write!(
                 f,
-                "defaulting to {} unpack ram",
+                "using up to {} of RAM to unpack components",
                 units::Size::new(*size, units::Unit::B, units::UnitMode::Norm)
             ),
             DownloadingFile(url, _) => write!(f, "downloading file from: '{}'", url),
--- a/src/dist/notifications.rs
+++ b/src/dist/notifications.rs
@@ -194,7 +194,7 @@ impl<'a> Display for Notification<'a> {
                 write!(f, "Force-skipping unavailable component '{}'", component)
             }
             SignatureInvalid(url) => write!(f, "Signature verification failed for '{}'", url),
-            RetryingDownload(url) => write!(f, "Retrying download for '{}'", url),
+            RetryingDownload(url) => write!(f, "retrying download for '{}'", url),
         }
     }
 }
--- a/tests/cli-exact.rs
+++ b/tests/cli-exact.rs
@@ -34,7 +34,7 @@ info: downloading component 'rust-docs'
 info: downloading component 'rust-std'
 info: downloading component 'rustc'
 info: installing component 'cargo'
-info: defaulting to 500.0 MiB unpack ram
+info: using up to 500.0 MiB of RAM to unpack components
 info: installing component 'rust-docs'
 info: installing component 'rust-std'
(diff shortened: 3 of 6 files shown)


src/utils/notifications.rs
```diff
@@ -87,7 +87,7 @@ impl<'a> Display for Notification<'a> {
             ),
             SetDefaultBufferSize(size) => write!(
                 f,
-                "defaulting to {} unpack ram",
+                "using up to {} of RAM for unpacking component",
```
camelid: The settled-upon message was "... for unpacking component**s**" – is that correct, or should it be singular like I have here?

camelid:
**Note to self:** Remember to post an update once this is merged to the [users.rust-lang.org thread](https://users.rust-lang.org/t/what-does-rustup-mean-by-defaulting-to-500-0-mib-unpack-ram/48528)!

src/utils/notifications.rs
```diff
@@ -87,7 +87,7 @@ impl<'a> Display for Notification<'a> {
             ),
             SetDefaultBufferSize(size) => write!(
                 f,
-                "defaulting to {} unpack ram",
+                "using up to {} of RAM to unpack component",
```
camelid: I think it reads better as "to unpack component" over "for unpacking component", so that's why I went with this.

camelid: Also, should it be singular or plural?

kinnison: It should be plural since the message is emitted once for the toolchain, not once per component.

camelid: Okay, I think it confused me because it says it *after* it has already said `info: installing component XYZ`:

https://github.com/rust-lang/rustup/blob/d43d6c181f588b4618e930954c24b9d301d86661/tests/cli-exact.rs#L35-L38

kinnison (changes requested):
The change looks good.  Assuming green CI then the change is good, but I'd prefer it were rebased to be a clean commit rather than still having the two fixup commits

kinnison:
Ci CI failure of aarch64 android is likely a docker snafu from rust-lang/rust (again)

camelid:
I'll squash and force-push which should do a retry.

camelid:
Looks like Docker failed again :(

kinnison (approved):

nealmcb:
Thanks, @camelid!

I've taken the liberty of following up on your "note to self" by noting the new language in the original discussion thread:

https://users.rust-lang.org/t/what-does-rustup-mean-by-defaulting-to-500-0-mib-unpack-ram/48528/6

I hope I got the exact default text right...

Issue #2003, opened 2019-09-18
Add support for aarch64-unknown-linux-musl (alpine armv8)


gidjituser:
Cannot install rust on an arm based alpine linux distribution. In my case armv8 (64-bit). musl is only supported on x86. 

It would be great to support the platform aarch64-unknown-linux-musl

kinnison:
This looks like it ought to be possible, based on the platform support list.  I guess someone needs to work out the right docker bits.  Should be an easy contribution if you want to give it a go.  I'm willing to help if someone shows up on our Discord too.

gidjituser:
Thanks. I am pretty new to rust, however I will try and give it a stab. 

I am currently using latest docker edge for the platform capability. 

```
FROM balenalib/aarch64-alpine
# Install build tools
RUN apk add --no-cache build-base gcc pkgconfig curl file git openssl-dev 
ENV PATH=/root/.cargo/bin:$PATH
# Install Rust
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y 
```

```
docker buildx build --platform linux/arm64 -t my/rust-image .
```

rustup-init.sh fails in get_bitness() as `head -c 5 /proc/self/exe` errors with Bad descriptor. Might need an alternative to getting bitness (32 or 64 bit) without dependencies.

mati865:
There is no possibility to do it right now because `aarch64-unknown-linux-musl` is target only (does not provide `cargo` and `rustc`) because Rust does *not* have enough CI budget.

kinnison:
So it is :(  I shall remove the easy/mentor tags for now, thanks @mati865

gidjituser:
Thanks for the updates. Currently rust arm64 docker images are quite large. Cannot use alpine or build static as that requires musl. Hopefully this will get in the pipeline at some point.

alex:
Is this the sort of thing that would be enabled by the ongoing work to add aarch64 CI for Rust, and making it a T1 platform? (https://github.com/rust-lang/rfcs/pull/2959)

rbtcollins:
Probably; I can't imagine that rustup wouldn't be able to install a Tier 1 triple.

alex:
Actually, I guess I have a slightly more important question: Does this require an aarch64 builder, or could it be done in `src/ci/docker/host-x86_64/dist-aarch64-linux/Dockerfile`, which cross-compiles currently?

[...]

PR #2493, opened 2020-09-17
Add aarch64-unknown-linux-musl support

Gelbpunkt:
See #2003 

I have no clue what to do with the .github/workflows files as they state not to edit them. Same applies to tests/channel-rust-nightly-example2.toml which seems autogenerated.

This is mainly the result of a grep for `aarch64-unknown-linux-gnu` and appropiate changes

### DIFF
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -28,6 +28,7 @@ jobs:
           - armv7-unknown-linux-gnueabihf
           - aarch64-linux-android
           - aarch64-unknown-linux-gnu     # skip-pr
+          - aarch64-unknown-linux-musl
           - powerpc64-unknown-linux-gnu   # skip-pr
           - x86_64-unknown-linux-musl     # skip-pr
           - i686-unknown-linux-gnu        # skip-pr skip-master
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -28,7 +28,7 @@ jobs:
           - armv7-unknown-linux-gnueabihf
           - aarch64-linux-android
           - aarch64-unknown-linux-gnu     # skip-pr
-          - aarch64-unknown-linux-musl
+          - aarch64-unknown-linux-musl    # skip-pr skip-master
           - powerpc64-unknown-linux-gnu   # skip-pr
           - x86_64-unknown-linux-musl     # skip-pr
           - i686-unknown-linux-gnu        # skip-pr skip-master
--- a/ci/docker/aarch64-unknown-linux-musl/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-musl/Dockerfile
@@ -1,0 +1,5 @@
+FROM rust-aarch64-unknown-linux-musl
+
+ENV CC_aarch64_unknown_linux_musl=aarch64-linux-musl-gcc \
+    CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-linux-musl-gcc \
+    RUSTFLAGS="-C target-feature=+crt-static -C link-arg=-lgcc"
--- a/.github/workflows/linux-builds-on-stable.yaml
+++ b/.github/workflows/linux-builds-on-stable.yaml
@@ -20,6 +20,7 @@ jobs:
           - armv7-unknown-linux-gnueabihf
           - aarch64-linux-android
           - aarch64-unknown-linux-gnu     # skip-pr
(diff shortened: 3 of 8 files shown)


ci/actions-templates/linux-builds-template.yaml
```diff
@@ -28,6 +28,7 @@ jobs:
           - armv7-unknown-linux-gnueabihf
           - aarch64-linux-android
           - aarch64-unknown-linux-gnu     # skip-pr
+          - aarch64-unknown-linux-musl    # skip-pr
```
kinnison: It's important that this not be skipped on PR until we know it works

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -55,6 +56,8 @@ jobs:
             snap_arch: i386  # skip-pr skip-master
           - target: aarch64-unknown-linux-gnu # skip-pr
             snap_arch: arm64 # skip-pr
+          - target: aarch64-unknown-linux-musl # skip-pr
+            snap_arch: arm64 # skip-pr
```
kinnison: Is `aarch64-unknown-linux-musl` really a valid snapcraft architecture?  I don't think so.  This hunk should not be present.

kinnison (changes requested):
You need to fix up the stuff I mentioned, and then in a separate commit, please include the results of running `bash ci/actions-templates/gen-workflows.sh`

ci/actions-templates/linux-builds-template.yaml
Gelbpunkt: Done, fixed in 6184ff0

ci/actions-templates/linux-builds-template.yaml
Gelbpunkt: Done, fixed in 6184ff0

Gelbpunkt:
@kinnison all done!

kinnison:
Okay the build has failed - the compiler appears to be missing - https://github.com/rust-lang/rustup/pull/2493/checks?check_run_id=1134461439

This does not bode well.

ci/docker/aarch64-unknown-linux-musl/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM rust-aarch64-unknown-linux-musl
+
+ENV CC_aarch64_unknown_linux_musl=aarch64-unknown-linux-musl-gcc \
+    CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER=aarch64-unknown-linux-musl-gcc
```
kinnison: So the question is, is this the name of the compiler inside the docker?

Gelbpunkt: Seems like it's aarch64-linux-musl, I'll update it

kinnison:
This time the build got a lot further but then failed to link because of the floating point library perhaps?  I am not a musl expert.  Look at https://github.com/rust-lang/rustup/pull/2493/checks?check_run_id=1134498637 and see if you have any ideas.

Gelbpunkt:
> This time the build got a lot further but then failed to link because of the floating point library perhaps? I am not a musl expert. Look at https://github.com/rust-lang/rustup/pull/2493/checks?check_run_id=1134498637 and see if you have any ideas.

Seems like https://github.com/rust-lang/rust/issues/46651 - what would be the steps to set compile arguments in the build here? (Can probably imagine using RUSTFLAGS)

kinnison:
I am not sure as I've never had to do that before.  I think you'll need to see help from outside, but I'd guess setting the ENV in the dockerfile maybe?

Gelbpunkt:
ENV would probably work. Just a question - this seems to run on a stable rust toolchain base, which does not include https://github.com/rust-lang/compiler-builtins/pull/377 (which should fix this). Is it possible to use nightly or are rustflags in ENV fine for now?

kinnison:
We explicitly use stable.  If it makes more sense, we can wait for a release which includes the fix, rather than pushing for things now.

Gelbpunkt:
It seems to work with the rustflags. I don't think it hurts to include them and it is easy to remove once 1.48 is stable and probably better than waiting.

kinnison (commented):
Yep this looks plausible.  This needs rebasing into a commit which adds support for the platform, then another which contains the regenerated workflows.  This platform should probably be `skip-pr skip-master` for now.

Gelbpunkt:
@kinnison I hope this is what you expected :smile: If not, let me know and I can change it

kinnison (approved):

kinnison:
Looks good, thank you.  We'll see how it goes at the next release :D

Issue #2492, opened 2020-09-17
Double `error:` print


maxwase:
**Problem**
Double `error` word print when executing 2nd level unexpected subcommand
Expected 1 `error` word like in all other cases

![image](https://user-images.githubusercontent.com/23321756/93513375-6ebe5280-f92e-11ea-9d62-ccb89e975ec9.png)

**Steps**
1. Execute `rustup link unexpected_option` or `rustup toolchain unexpected_option` or ... (any wrong 2nd level subcommand)

**Notes**

Output of `rustup --version`: rustup 1.22.1 (b01adbbc3 2020-07-08)
Output of `rustup show`: 
```Default host: x86_64-unknown-linux-gnu
rustup home:  /home/max/.rustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
1.41.0-x86_64-unknown-linux-gnu
1.42.0-x86_64-unknown-linux-gnu
1.45.0-x86_64-unknown-linux-gnu

installed targets for active toolchain
--------------------------------------

thumbv6m-none-eabi
thumbv7m-none-eabi
x86_64-unknown-linux-gnu

active toolchain
----------------

stable-x86_64-unknown-linux-gnu (default)
rustc 1.46.0 (04488afe3 2020-08-24)
```

kinnison:
I can confirm this issue still exists on master.  It should be fairly simple to fix.

rbtcollins:
Looks like its been fixed

PR #2496, opened 2020-09-21
Fix double error label for certain error types

eidenar:
Fix for issue #2492 Second error message came from `src/bin/rustup-init.rs:37` in case if error was not matched in `src/cli/rustup_mode.rs:main`. I expanded match conditions for required error types here.


kinnison:
Functionality looks good here, but I'd prefer if you could rebase this into one commit rather than two where one is broken.

eidenar:
@kinnison Thanks for the note. I updated the PR

kinnison:
Thank you :D

PR #2499, opened 2020-09-28
Fix #2490: Don't fail artificially on large files

### LINKED ISSUE #2490: Increase max file size to ~500MB
**Problem**

rustc_driver dll on Windows bundles the whole compiler and LLVM into one file, and unsurprisingly, that's a big binary, currently hovering at ~193 MB. This is quite close to the 200 MB limit (and indeed, depending on how that limit is defined -- powers of 10 or 2, over it).

**Steps**

I can't reproduce myself, but we did get a user report https://github.com/rust-lang/rust/issues/76795 -- maybe the limit used to be smaller?

**Possible Solution(s)**

**Notes**

Output of `rustup --version`: unknown
Output of `rustup show`: unknown

rbtcollins:
We may still fail on them, but allow the failure to actually occur
rather than insisting that we will be unable to handle them.

### DIFF
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -196,6 +196,8 @@ impl MemoryBudget {
             }
         };
 
+        // Future us: this can be removed when IO chunking within a single file is possible: it just helps generate good
+        // errors rather than allocator-failure panics when we hit the large file on a RAM limited system.
         if max_file_size > unpack_ram {
             panic!("RUSTUP_UNPACK_RAM must be larger than {}", max_file_size);
         }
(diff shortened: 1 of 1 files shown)


src/dist/component/package.rs
```diff
@@ -356,7 +358,16 @@ fn unpack_without_first_dir<'a, R: Read>(
 
         let size = entry.header().size()?;
         if size > MAX_FILE_SIZE {
-            return Err(format!("File too big {} {}", relpath.display(), size).into());
+            // If we cannot tell the user we will either succeed (great), or fail (and we may get a bug report), either
+            // way, we will most likely get reports from users about this, so the possible set of custom builds etc that
+            // don't report are not a great concern.
+            if let Some(notify_handler) = notify_handler {
+                notify_handler(Notification::Error(format!(
```
kinnison, line 365: Should this be a warning rather than an error?  In theory we may succeed anyway if swap helps out, and if so, an error seems risky.  Perhaps "Warning file {} is too large ({}) and may fail to unpack" ?

rbtcollins: I want people to tell us, so we can bump the number quickly :). Also its a smaller patch.

kinnison: Reasoning accepted.

kinnison (approved):

PR #2508, opened 2020-10-03
Chunked disk IO

rbtcollins:
Just some cleanups at this stage.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ cfg-if = "1.0"
 chrono = "0.4"
 clap = "2"
 download = {path = "download"}
+effective-limits = "0.5.2"
 error-chain = "0.12"
 flate2 = "1"
 git-testament = "0.1.4"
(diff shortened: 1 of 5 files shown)


kinnison:
These cleanups look good - are you intending to take this further or is it blocked on effective-limits updates or somesuch?

rbtcollins:
I intend to take it further but we can of course land it incrementally if
you like

kinnison:
Have you been able to verify the freebsd and netbsd builds?  If so, then I suggest relabelling the PR and landing the cleanups since they add support for effective-limits in those platforms.

rbtcollins:
I have not - who would be a good person to ask to do that?

kinnison:
Having built them, I'm not sure other than using a VM or somesuch.  Perhaps if you can build them you could ask on the internals.rust-lang.org forum for a tester?  Or reddit?

rbtcollins:
To be clear; this doesn't add effective-limits supports on those platforms - it just moves the handling of the lack of support into effective-limits - and goes *some way* to adding support via the related PRs on libc etc.

I would like to just merge this I think and ask for testers after its in master - since passing CI on the relevant platforms in effective-limits itself.

kinnison:
You have a Cargo.lock conflict but I remain positive on merging this otherwise.

kinnison (approved):

kinnison:
Good to merge on green

PR #2513, opened 2020-10-08
Update deprecated GitHub Actions commands.

ehuss:
The old method of using `::set-env` and `::add-path` have been deprecated, see https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/.

This includes a second commit to work around a GitHub Actions cache bug on macOS.

### DIFF
--- a/.github/workflows/centos-fmt-clippy-on-all.yaml
+++ b/.github/workflows/centos-fmt-clippy-on-all.yaml
@@ -37,7 +37,7 @@ jobs:
           mkdir -p ~/.cargo/{registry,git}
       - name: Set environment variables appropriately for the build
         run: |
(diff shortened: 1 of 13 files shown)


ehuss:
Note that CI caching is not really working for any job that doesn't set `SKIP_TESTS`.  I filed https://github.com/rust-lang/cargo/issues/8762 with a description of the problem.  This will be somewhat mitigated by https://github.com/rust-lang/cargo/pull/8755 which should be on nightly (that will fix CI caching, but all the dependencies are still being built twice).  Some workarounds would be:
* Add `lto` to `[profile.bench]`.
* Don't use `--release` when running tests.

kinnison (approved):
Thanks for doing this.  It's always nice when things change under us :+1:

Issue #2515, opened 2020-10-09
Migration to new way to set env var and path in GH Actions


tesuji:
Deprecation announcement blog: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

New way: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

An affected file: https://github.com/rust-lang/rustup/blob/243b453934929ea19f4172f626018531bab45ddd/ci/actions-templates/linux-builds-template.yaml#L82-L83

tesuji:
Maybe fixed by #2513.

kinnison:
This has been merged now.

PR #2521, opened 2020-10-15
Build for aarch64-apple-darwin

shepmaster:


kinnison:
The macos changes need to go in the actions template, otherwise they'll be lost next time we regenerate the actions files.  Also I think the set-env stuff needs changing now :D

shepmaster:
Ah, I was waiting for https://github.com/rust-lang/rustup/issues/2515 and it looks like it’s been addressed.

shepmaster:
> The macos changes need to go in the actions template

Done.

> the set-env stuff needs changing now

Done.

kinnison:
As a final point before I merge this, have you confirmed that `rustup` believes this is a valid *host* triple?

kinnison:
I think it should do, and I seem to recall you running your cross-built binary on an aarch64 mac, but I just want to be sure

shepmaster:
> have you confirmed that `rustup` believes this is a valid _host_ triple?

If I understand you, I think so. `rustup` actually [works “as-is” today](https://github.com/shepmaster/rust/blob/silicon/silicon/README.md#compiling-on-native-aarch64-hardware):

> ```
> arch --x86_64 \
> sh <(curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs) \
> --default-host aarch64-apple-darwin \
> --default-toolchain nightly
> ```

I expect that this PR removes the need to bring in the x86-64 emulator. A future change to the shell script should reduce it to the usual one-liner.

PR #2529, opened 2020-10-18
do_msvc_check: set default choice to no

### LINKED ISSUE #2514: Default choice when running rustup-init without build tools installed is to continue

Running `rustup-init.exe` on a computer without C++ build tools installed display an useful message explaining to the
user how to install the requirement, while indicating that this requirement may be ignored if the user want to target only GNU ABI, or if he know what he's doing.

The default action prompted is to continue the installation process anyway, and I think it may not be the best idea. A lot of users will just want to get started with rust and will likely need the build tools installed before installing the toolchain. The other scenario will happen less often.

I think the default choice should be to not continue the installation process to prevent mistakes, and letting people who *know what they are doing* choosing to continue the installation process anyway.

Capture

Changing the default action to **No** to prevent any mistakes.

**Notes**

I think this is an opinionated suggestion and may not be the best one, therefore it's rather an open question than an issue.

creekorful:
Before that the behavior on Windows when running rustup-init
without having build tools installed was to continue.

Now the user need to explicitly choose to continue installation,
and thus we are making sure they are carefully reading the warning
message.

Closes: #2514

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -307,7 +307,7 @@ pub fn install(
             warn!("installing msvc toolchain without its prerequisites");
         } else {
             md(&mut term, MSVC_MESSAGE);
-            if !common::confirm("\nContinue? (Y/n)", true)? {
+            if !common::confirm("\nContinue? (y/N)", false)? {
                 info!("aborting installation");
                 return Ok(utils::ExitCode(0));
             }


kinnison:
Before I can accept this (and I really do want to) we should take a moment to reflect on whether this might break anyone's CI.  Is it plausible anyone would be programmatically responding to this prompt because they know they're using mingw, but without passing `-y` to the init process?

creekorful:
I guess that's plausible. But I think that users should be passing `-y` flag to ensure skipping behavior rather than relying on interactive prompt (especially in CI).

So this _could_ be a breaking change for existing CI.

What should we do?

kinnison:
We make a call, my vote is with you - that if you want to deal with prompts like this you should be passing `-y` but I'd like @rbtcollins to weigh in.  Robert, if you agree with us, you can merge this rather than metoo'ing :D

rbtcollins:
I think this needs a release note and perhaps even a doc update - I haven't reviewed the doc instructions but I have a suspicion that we have something in there that overlaps this.

J-Vin:
Why does this skip test and find no cache?

rbtcollins:
> 
> 
> Why does this skip test and find no cache?

You'll need to be more specific about your question; the builds look fine to me.

J-Vin:
1s

1Run actions/cache@v2 

2 with: 

3 path: target 

4 key: master-fixes/2514-Linux-cargo-clippy-target-dir-18bf6b4f01a6feaf7259ba7cdae58031af1b7b39-93e7e6dc29aba5e98d8c3fcffd40e248c5356201d78692e0e5584960a6779d37 

5 restore-keys: master-Linux-cargo-clippy-target-dir-18bf6b4f01a6feaf7259ba7cdae58031af1b7b39-93e7e6dc29aba5e98d8c3fcffd40e248c5356201d78692e0e5584960a6779d37 

6Cache not found for input keys: master-fixes/2514-Linux-cargo-clippy-target-dir-18bf6b4f01a6feaf7259ba7cdae58031af1b7b39-93e7e6dc29aba5e98d8c3fcffd40e248c5356201d78692e0e5584960a6779d37, master-Linux-cargo-clippy-target-dir-18bf6b4f01a6feaf7259ba7cdae58031af1b7b39-93e7e6dc29aba5e98d8c3fcffd40e248c5356201d78692e0e5584960a6779d37

J-Vin:

22Warning: Not enforcing TLS v1.2, this is potentially less secure 

23info: profile set to 'default' 

24info: default host triple is x86_64-unknown-linux-gnu 

25info: skipping toolchain installation

J-Vin:
Im not great at explaining what I mean, I apologize. Y'all are the experts... I just see errors allll the time. They jump out at me and do not let go. **for o for days** also... Extra header error means what?

kinnison:
The caches aren't always executed cleanly because of limits on github actions' side.  Don't worry about them.

kinnison:
I think that @rbtcollins is right that we need to double check the documentation.  Can you go through the installation instructions for Windows and see if we need to tweak that a bit to explain this.  If we decide to merge then I'm happy to write the release note as part of the announce / changelog when we do the release, I'll just have to make an issue to track that.

creekorful:
Hello, I have reviewed the Windows install instructions [here](https://rust-lang.github.io/rustup/installation/windows.html) and haven't found something to update. The documentation looks valid and up-to-date AFAIC.
Can someone else check this to make sure I haven't forgot anything?

kinnison:
I agree that the docs don't mention anything to the effect of the questions, nor their defaults.

All I'm wondering is if there's value in mentioning something since those who use `rustup` but expect to be using the GNU toolchain may find the change in default surprising.  Then again people who use muscle-memory to install `rustup` are probably few and far between.

@rbtcollins If you agree, I suggest we merge this and ensure that a release note mentions the change.

creekorful:
Hello there!

Any news on this changes? :)

Issue #2520, opened 2020-10-11
Warn when downloading a toolchain for a different platform


mk12:
# Problem

I recently tried to cross compile a Windows binary from my MacBook.

After some research, I figured I needed to install a new toolchain:

```
$ rustup toolchain install stable-x86_64-pc-windows-msvc
...
  stable-x86_64-pc-windows-msvc installed - (rustc does not exist)
```

It apparently succeeded (exit status 0), but then when I try to build it can't find cargo:

```
$ cargo +stable-x86_64-pc-windows-msvc build
error: the 'cargo' binary, normally provided by the 'cargo' component, is not applicable to the 'stable-x86_64-pc-windows-msvc' toolchain
```

After more searching I realized this is giving me a toolchain for building _on_ Windows, i.e. cargo.exe. What I really want is to just have one toolchain for my Mac, but add a target _for_ Windows:

```
$ rustup target add x86_64-pc-windows-msvc
$ cargo build --target x86_64-pc-windows-msvc
```

(It still doesn't work because there are [more steps](https://v4.chriskrycho.com/2016/using-rust-for-scripting.html#fnref1) to get linking to work, but at least I'm on the right track now.)

# Request

It would have saved me an hour if rustup had printed something like:

```
warning: toolchain stable-x86_64-pc-windows-msvc cannot run on this platform; use `rustup target add` if you want to compile to a non-host target
```

kinnison:
Unfortunately it's not entirely obvious which host toolchains will run sucessfully on any given system.  For example, I can run "natively" any of x86_64 linux, amd64 linux, i686 linux, x86_64 windows, and i686 windows binaries on my Linux laptop.

I suppose a warning when the toolchain build does not match the system host would be plausible, but it'd annoy people particularly who install i686 versions of the toolchain on their x86_64 systems.

Perhaps softening the wording to:

```
warning: toolchain stable-x86_64-pc-windows-msvc may not be able to run on this system.
warning: If you meant to build software to target that platform, perhaps try `rustup target add x86_64-pc-windows-msvc` instead?
```

What do you think?

mk12:
(By running windows binaries natively on linux, do you mean using Wine?)

For sure, I just quickly typed out that warning message, assuming it would be given some more thought if this is implemented :) Your softer wording looks good to me, and would have solved my problem.

kinnison:
Yes, with a bunch of bits and bobs, wine, qemu, etc. you can run a lot of non-native binaries as though they were native.

[...]

PR #2534, opened 2020-10-22
warn toolchain on different platform

frbimo:
**Fixes** 
#2520 

**Testing**
`cargo test`

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -15,7 +15,9 @@ use super::self_update;
 use super::term2;
 use super::term2::Terminal;
 use super::topical_doc;
-use crate::dist::dist::{PartialTargetTriple, PartialToolchainDesc, Profile, TargetTriple};
+use crate::dist::dist::{
+    PartialTargetTriple, PartialToolchainDesc, Profile, TargetTriple, ToolchainDesc,
+};
 use crate::dist::manifest::Component;
 use crate::process;
 use crate::toolchain::{CustomToolchain, DistributableToolchain};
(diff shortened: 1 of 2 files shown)


frbimo:
@kinnison kindly reminder*

src/cli/rustup_mode.rs
```diff
@@ -904,6 +906,32 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
...
+                Ok(x) => x.has_triple(),
+                _ => false,
+            };
+
+            if toolchain_has_triple {
+                let host_arch = TargetTriple::from_host_or_build();
+                match ToolchainDesc::from_str(name) {
+                    Ok(toolchain_desc) => {
+                        let target_triple = toolchain_desc.target.clone();
+                        if host_arch.ne(&target_triple) {
```
kinnison: Why did you feel the need to clone that here?

kinnison (commented):
There's only one tiny niggle around what I feel might be a wasted `clone()` otherwise this looks excellent.  I'm especially grateful you added a test for this.

src/cli/rustup_mode.rs
frbimo: fixed

frbimo:
@kinnison ptal

kinnison (approved):
This looks good to me.  One final chance for @rbtcollins to have a look in case the approach or wording isn't to his taste, otherwise I'll look to merge it fairly soon.

rbtcollins:
Should this ask interactively perhaps?

kinnison:
I'm loathe to add an interactive prompt to a part of `rustup` which has never been interactive before now.  If we think interactivity makes sense then we have to detect if there's a tty on stdin and stuff like that, otherwise it'd be a pain for anyone who has automated this in CI or similar.  Also we simply don't know for sure so warning non-interactively is probably enough.

kinnison:
I rebased to clean up the conflict and am minded to merge this as I remain of the opinion that interactive prompting doesn't seem ideal to me.

kinnison:
The CI badness is snapcraft being odd.

PR #2536, opened 2020-10-23
Resolving proxy environment variables for curl backend

### LINKED ISSUE #1343: Supports HTTPS proxy authorization
Are there a way to support HTTPS proxy authorization? Idealy it should be 

    export $proxy-authorization="Basic YWxhZGRpbjpvcGVuc2VzYW1l"

Otherwise, for organizations that require proxy authorization, we could not connect.

I also quickly looked at the code under `src/download/lib.rs`, couldn't find "authorization" or "authentication" so I assume this function is not currently existing.

whizsid:
Fixes #1343

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -129,11 +129,62 @@ pub fn download_to_path_with_backend(
 pub mod curl {
     use super::Event;
     use crate::errors::*;
-    use curl::easy::Easy;
+    use curl::easy::{Auth, Easy};
     use std::cell::RefCell;
+    use std::env::var_os;
     use std::str;
     use std::time::Duration;
     use url::Url;
+
+    struct ProxyInfo {
+        url: String,
+        port: Option<u16>,
+        username: Option<String>,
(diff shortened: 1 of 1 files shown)


kinnison:
I don't think this is the right approach for this.  Not least, we already use env_proxy for picking proxies for URLs so we should be using that here, rather than replicating the logic yet again.  I acknowledge that you've had a look at Reqwest and cURL but I'm not sure that you've provided anything more functional than what is already in those libraries.  I'd much prefer to see the reqwest and/or cURL libraries updated if truly they do not accept credentials in the proxy URLs yet.

Also anything which touches the proxying code needs very careful testing these days because it appears to be quite delicate in some cases our users have.

Could you start by explaining where the cURL code processes the proxy environment variables and how you determined that it doesn't support authentication information stored in them?

whizsid:
Sorry! I think I misunderstood the actual issue. So I am closing it. Thank you @kinnison for the review.

PR #2539, opened 2020-10-27
Fix formatting of "To configure your current shell" message

camelid:


kinnison (approved):
Looks fairly benign and does look nicer.

PR #2540, opened 2020-10-27
Allow selecting toolchains to uninstall with glob pattern

### LINKED ISSUE #2530: Glob support for rustup toolchain remove
I have 63 toolchains, culminating in 40GB of hard drive use. I'd love to remove all nightlies, but I have to add the nightlies one by one. It would be great if one could specify regexes like `rustup toolchain remove nightly-*`.

camelid:
Closes #2530.

### Correct usage

```console
$ rustup toolchain list
nightly-2020-01-01-x86_64-apple-darwin
nightly-2020-06-01-x86_64-apple-darwin
nightly-x86_64-apple-darwin

$ rustup toolchain uninstall 'nightly-20*'
info: uninstalling toolchain 'nightly-2020-01-01-x86_64-apple-darwin'
info: toolchain 'nightly-2020-01-01-x86_64-apple-darwin' uninstalled
info: uninstalling toolchain 'nightly-2020-06-01-x86_64-apple-darwin'
info: toolchain 'nightly-2020-06-01-x86_64-apple-darwin' uninstalled

$ rustup toolchain list
nightly-x86_64-apple-darwin
```

### Invalid pattern

```console
$ rustup toolchain uninstall '***'
error: Pattern syntax error near position 2: wildcards are either regular `*` or recursive `**`
```

**Work-in-progress:**

- [x] Errors need improvement (instead of just panicking)
- [x] Needs test
- [ ] Needs CLI docs

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -406,13 +406,6 @@ pub fn cli() -> App<'static, 'static> {
                         .about("Uninstall a toolchain")
                         .alias("remove")
                         .arg(
-                            Arg::with_name("pattern")
-                                .help("Treat arguments as glob patterns")
-                                .short("p")
-                                .long("pattern")
-                                .takes_value(false),
-                        )
-                        .arg(
                             Arg::with_name("toolchain")
                                 .help(TOOLCHAIN_ARG_HELP)
                                 .required(true)
@@ -1320,23 +1313,17 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
 }
 
 fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    if m.is_present("pattern") {
-        for pattern_str in m.values_of("toolchain").unwrap() {
-            let pattern = Pattern::new(&pattern_str)?;
+    for pattern_str in m.values_of("toolchain").unwrap() {
+        let pattern = Pattern::new(&pattern_str)?;
 
-            let mut toolchains = cfg.get_toolchains_from_glob(pattern)?.peekable();
-            if toolchains.peek().is_none() {
-                info!("no toolchains matched pattern '{}'", pattern_str);
-                return Ok(utils::ExitCode(0));
-            }
+        let mut toolchains = cfg.get_toolchains_from_glob(pattern)?.peekable();
 
+        if toolchains.peek().is_some() {
             for toolchain in toolchains {
                 toolchain.remove()?;
             }
-        }
-    } else {
-        for toolchain in m.values_of("toolchain").unwrap() {
-            let toolchain = cfg.get_toolchain(toolchain, false)?;
+        } else {
+            let toolchain = cfg.get_toolchain(pattern_str, false)?;
             toolchain.remove()?;
         }
     }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1332,9 +1332,6 @@ fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode
         }
     } else {
         for toolchain in m.values_of("toolchain").unwrap() {
-            if m.is_present("regex") {
-            } else {
-            }
             let toolchain = cfg.get_toolchain(toolchain, false)?;
             toolchain.remove()?;
         }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1319,10 +1319,12 @@ fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode
         let mut toolchains = cfg.get_toolchains_from_glob(pattern)?.peekable();
 
         if toolchains.peek().is_some() {
+            // This pattern matched some toolchains, so remove each of the ones it matched.
             for toolchain in toolchains {
                 toolchain.remove()?;
             }
         } else {
+            // It didn't match any toolchains, so treat it as a partial toolchain specifier.
             let toolchain = cfg.get_toolchain(pattern_str, false)?;
             toolchain.remove()?;
         }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -7,6 +7,7 @@ use std::process::Command;
 use std::str::FromStr;
 
 use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, Shell, SubCommand};
+use regex::Regex;
 
 use super::common;
 use super::errors::*;
@@ -1324,7 +1325,7 @@ fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode
             "exactly one regex filter must be supplied"
         );
 
-        let regex = regex::Regex::from_str(m.values_of("toolchain").unwrap().next().unwrap())
+        let regex = Regex::from_str(m.values_of("toolchain").unwrap().next().unwrap())
             .expect("invalid regex");
 
         for toolchain in cfg.get_toolchains_from_regex(regex)? {
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -7,7 +7,7 @@ use std::process::Command;
 use std::str::FromStr;
 
 use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, Shell, SubCommand};
-use regex::Regex;
+use glob::Pattern;
 
 use super::common;
 use super::errors::*;
@@ -406,9 +406,10 @@ pub fn cli() -> App<'static, 'static> {
                         .about("Uninstall a toolchain")
                         .alias("remove")
                         .arg(
-                            Arg::with_name("regex")
-                                .help("Use a regular expression to select the toolchains to uninstall")
-                                .long("regex")
+                            Arg::with_name("pattern")
+                                .help("Treat arguments as glob patterns")
+                                .short("p")
+                                .long("pattern")
                                 .takes_value(false),
                         )
                         .arg(
@@ -1319,17 +1320,12 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
 }
 
 fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    if m.is_present("regex") {
-        assert!(
-            m.values_of("toolchain").unwrap().len() == 1,
-            "exactly one regex filter must be supplied"
-        );
-
-        let regex = Regex::from_str(m.values_of("toolchain").unwrap().next().unwrap())
-            .expect("invalid regex");
-
-        for toolchain in cfg.get_toolchains_from_regex(regex)? {
-            toolchain.remove()?;
+    if m.is_present("pattern") {
+        for pattern_str in m.values_of("toolchain").unwrap() {
+            let pattern = Pattern::new(pattern_str)?;
+            for toolchain in cfg.get_toolchains_from_glob(pattern)? {
+                toolchain.remove()?;
+            }
         }
     } else {
         for toolchain in m.values_of("toolchain").unwrap() {
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -405,6 +405,12 @@ pub fn cli() -> App<'static, 'static> {
                         .about("Uninstall a toolchain")
                         .alias("remove")
                         .arg(
+                            Arg::with_name("regex")
+                                .help("Use a regular expression to select the toolchains to uninstall")
+                                .long("regex")
+                                .takes_value(false),
+                        )
+                        .arg(
                             Arg::with_name("toolchain")
                                 .help(TOOLCHAIN_ARG_HELP)
                                 .required(true)
@@ -1312,10 +1318,28 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
 }
 
 fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    for toolchain in m.values_of("toolchain").unwrap() {
-        let toolchain = cfg.get_toolchain(toolchain, false)?;
-        toolchain.remove()?;
+    if m.is_present("regex") {
+        assert!(
+            m.values_of("toolchain").unwrap().len() == 1,
+            "exactly one regex filter must be supplied"
+        );
+
+        let regex = regex::Regex::from_str(m.values_of("toolchain").unwrap().next().unwrap())
+            .expect("invalid regex");
+
+        for toolchain in cfg.get_toolchains_from_regex(regex)? {
+            toolchain.remove()?;
+        }
+    } else {
+        for toolchain in m.values_of("toolchain").unwrap() {
+            if m.is_present("regex") {
+            } else {
+            }
+            let toolchain = cfg.get_toolchain(toolchain, false)?;
+            toolchain.remove()?;
+        }
     }
+
     Ok(utils::ExitCode(0))
 }
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1322,8 +1322,15 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
 fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     if m.is_present("pattern") {
         for pattern_str in m.values_of("toolchain").unwrap() {
-            let pattern = Pattern::new(pattern_str)?;
-            for toolchain in cfg.get_toolchains_from_glob(pattern)? {
+            let pattern = Pattern::new(&pattern_str)?;
+
+            let mut toolchains = cfg.get_toolchains_from_glob(pattern)?.peekable();
+            if toolchains.peek().is_none() {
+                info!("no toolchains matched pattern '{}'", pattern_str);
+                return Ok(utils::ExitCode(0));
+            }
+
+            for toolchain in toolchains {
                 toolchain.remove()?;
             }
         }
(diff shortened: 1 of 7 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1312,10 +1318,28 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
...
 
 fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    for toolchain in m.values_of("toolchain").unwrap() {
-        let toolchain = cfg.get_toolchain(toolchain, false)?;
-        toolchain.remove()?;
+    if m.is_present("regex") {
+        assert!(
+            m.values_of("toolchain").unwrap().len() == 1,
+            "exactly one regex filter must be supplied"
+        );
```
camelid: This should be a nice error message. How should I do that?

src/cli/rustup_mode.rs
```diff
@@ -1312,10 +1318,28 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
...
+
+        let regex = regex::Regex::from_str(m.values_of("toolchain").unwrap().next().unwrap())
+            .expect("invalid regex");
```
camelid: Same here.

src/cli/rustup_mode.rs
```diff
@@ -404,6 +404,12 @@ pub fn cli() -> App<'static, 'static> {
                     SubCommand::with_name("uninstall")
                         .about("Uninstall a toolchain")
                         .alias("remove")
+                        .arg(
+                            Arg::with_name("regex")
+                                .help("Use a regular expression to select the toolchains to uninstall")
+                                .long("regex")
+                                .takes_value(false),
```
camelid: I did not add the option of `-r` since that's often used to mean "recursive" (e.g. `grep -r`) and I don't want to confuse people.

src/cli/rustup_mode.rs
```diff
@@ -404,6 +405,12 @@ pub fn cli() -> App<'static, 'static> {
...
+                        )
```
camelid: Side note: I noticed that rustfmt didn't format this part at all; I had to do it manually. Is that expected? Maybe the expression building up the CLI is too big?

src/cli/rustup_mode.rs
kinnison: Ideally you should put it as a clap validation function then clap will enforce it for us.

kinnison:
One comment, before I even look at the code too deeply, I'm not sure regular expressions are ideal for this, instead perhaps it's best to advertise this as shell-glob style patterns.  i.e. `stable-*` rather than `stable-.*` being needed?

src/cli/rustup_mode.rs
kinnison: I'm wondering if we need this at all (flag that is) -- instead we could just treat the name as a pattern if there's a `*` in it.

src/cli/rustup_mode.rs
kinnison: This chunk of code is too big for rustfmt.  We need to break it up, but that's something for another day.

src/cli/rustup_mode.rs
kinnison: But as I mention above, it may be better to just never use this flag, and simply accept that clap ought to require at least one toolchain for `remove` and we treat the name as a pattern if it has a `*` in it?

src/cli/rustup_mode.rs
kinnison: If it were a simple pattern rather than a regex this wouldn't be necessary.  Otherwise it ought to be a clap validator function checking this.

src/config.rs
```diff
@@ -362,6 +364,14 @@ impl Cfg {
         Toolchain::from(self, name)
     }
 
+    pub fn get_toolchains_from_regex(&self, regex: Regex) -> Result<Vec<Toolchain<'_>>> {
```
kinnison: Would it be possible to return a `Result<impl Iterator<Item=Toolchain<'_>>>` instead?  No need to `.collect()` in theory.

src/config.rs
```diff
@@ -694,18 +704,21 @@ impl Cfg {
...
-                .collect();
-
-            utils::toolchain_sort(&mut toolchains);
+        let mut toolchains: Vec<_> = self.list_toolchains_iter()?.collect();
+        utils::toolchain_sort(&mut toolchains);
+        Ok(toolchains)
+    }
 
-            Ok(toolchains)
+    fn list_toolchains_iter(&self) -> Result<Box<dyn Iterator<Item = String>>> {
```
kinnison, line 714: These days, could that be a `Result<impl Iterator<Item=String>>` ?

camelid:
I'm not fully sure what you mean by "advertise as shell-glob style patterns": It takes a regex, not a glob pattern. I realized that I confused myself with the example I tried because I used the regex `nightly-*` which actually matches the substring `nightly`, `nightly-`, `nightly--`, etc.

kinnison:
Most "mortals" are not regex savvy, but they understand shell glob patterns (i.e. `*` not `.*`)

camelid:
So are you suggesting that we only allow the metacharacter `*`, escape all other characters, replace the `*` with `.*`, and then pass it to regex? I think I agree that we should have it be a shell glob pattern, but I'm not sure what the best way to implement it is.

src/config.rs
camelid: I don't think so. I tried that, but it didn't work because the `else` branch returns `iter::empty()`, which is a different concrete type.

src/config.rs
camelid: Sure, then we could collect only when we need to.

camelid:
Oh, I guess we could use the `glob` crate to do this. Is that what you were thinking?

kinnison:
If I were working in C I'd just use the `fnmatch()` function so if there's a similar crate then that might be best.  Otherwise as you suggest, the transformation of escaping metacharacters other than * and then changing * for .* would work.

src/cli/rustup_mode.rs
camelid: Actually you can have invalid patterns. I added an error case for it.

src/cli/rustup_mode.rs
camelid: It supports multiple patterns now.

src/cli/rustup_mode.rs
```diff
@@ -404,6 +405,13 @@ pub fn cli() -> App<'static, 'static> {
                     SubCommand::with_name("uninstall")
                         .about("Uninstall a toolchain")
                         .alias("remove")
+                        .arg(
+                            Arg::with_name("pattern")
+                                .help("Treat arguments as glob patterns")
+                                .short("p")
+                                .long("pattern")
+                                .takes_value(false),
+                        )
```
camelid: We have a couple options here:

1. **Treat all toolchains as patterns.** Upside: simpler to use. Downside: could be confusing if that's not what people expected.

2. **Only treat toolchains with metacharacters (`*` or `?` I think).** Upside: also simpler to use. Downside: we'd have to figure out a way to make sure we catch edge-cases, etc.

3. **Keep it as an optional behavior enabled with a flag.** We may want to bikeshed whether this should be `--pattern` or `--glob`. Perhaps look at prior art?

src/cli/rustup_mode.rs
```diff
@@ -1312,10 +1320,20 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
...
 fn toolchain_remove(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    for toolchain in m.values_of("toolchain").unwrap() {
-        let toolchain = cfg.get_toolchain(toolchain, false)?;
-        toolchain.remove()?;
+    if m.is_present("pattern") {
+        for pattern_str in m.values_of("toolchain").unwrap() {
+            let pattern = Pattern::new(pattern_str)?;
+            for toolchain in cfg.get_toolchains_from_glob(pattern)? {
+                toolchain.remove()?;
+            }
```
camelid: Should we print a message if the pattern didn't match any toolchains so the user notices the silent "failure"?

src/config.rs
```diff
@@ -362,6 +364,16 @@ impl Cfg {
         Toolchain::from(self, name)
     }
 
+    pub fn get_toolchains_from_glob(
+        &self,
+        pattern: Pattern,
+    ) -> Result<impl Iterator<Item = Toolchain<'_>>> {
```
camelid, line 370: I'm calling the function `...glob` since "pattern" could mean `std::pattern::Pattern`, which is not the meaning here.

src/cli/rustup_mode.rs
kinnison: Yes, if nothing was removed it should be an error, same as if you try to remove a toolchain which is not installed now.  However I remain interested in if there's a way to not need the `--pattern` flag at all.  Could we simply always treat the incoming `toolchain` arguments as patterns?

camelid: See my comment: https://github.com/rust-lang/rustup/pull/2540#discussion_r513770900

src/cli/rustup_mode.rs
```diff
@@ -1312,10 +1320,27 @@ fn toolchain_link(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
...
-        let toolchain = cfg.get_toolchain(toolchain, false)?;
-        toolchain.remove()?;
+    if m.is_present("pattern") {
+        for pattern_str in m.values_of("toolchain").unwrap() {
+            let pattern = Pattern::new(&pattern_str)?;
+
+            let mut toolchains = cfg.get_toolchains_from_glob(pattern)?.peekable();
+            if toolchains.peek().is_none() {
+                info!("no toolchains matched pattern '{}'", pattern_str);
+                return Ok(utils::ExitCode(0));
```
camelid: I think this should return a non-zero exit code, but I modelled this after `rustup toolchain uninstall some-toolchain`, which just shows a message with a successful exit code. Should I change both of them to return an error?

src/cli/rustup_mode.rs
kinnison: If people currently put shell glob characters into their toolchain names then they're probably in a world of pain without us treating all incoming names as patterns, so my vote is for 1.  @rbtcollins what do you think?

src/cli/rustup_mode.rs
kinnison: If it has always been "safe" in the shell scripting sense to "remove" a toolchain which isn't present, then we wouldn't want to break peoples' scripting by making them check first before trying.  Better to be informational but consistent with past behaviour here I think.

src/cli/rustup_mode.rs
kinnison: Let's collate conversation on this point over there.

src/cli/rustup_mode.rs
rbtcollins: depends on the patterns we're talking about . '.' for instance from regex patterns is present in version numbers. If we're just talking '*' and '?' as pattern characters, then I think we're fine.

src/cli/rustup_mode.rs
rbtcollins: I'm not entirely sure that this makes sense. Patterns are new; failure to match a pattern is a new situation. We can detect a toolchain string that has no pattern characters and do the prior behaviour, if we want to.

That means, we can take a first principles design here if appropriate.

src/cli/rustup_mode.rs
camelid: The docs for the glob crate say:

- `?` matches any single character.

- `*` matches any (possibly empty) sequence of characters.

- `**` matches the current directory and arbitrary subdirectories. This
  sequence **must** form a single path component, so both `**a` and `b**`
  are invalid and will result in an error.  A sequence of more than two
  consecutive `*` characters is also invalid.

- `[...]` matches any character inside the brackets.  Character sequences
  can also specify ranges of characters, as ordered by Unicode, so e.g.
  `[0-9]` specifies any character between 0 and 9 inclusive. An unclosed
  bracket is invalid.

- `[!...]` is the negation of `[...]`, i.e. it matches any characters
  **not** in the brackets.

- The metacharacters `?`, `*`, `[`, `]` can be matched by using brackets
  (e.g. `[?]`).  When a `]` occurs immediately following `[` or `[!` then it
  is interpreted as being part of, rather then ending, the character set, so
  `]` and NOT `]` can be matched by `[]]` and `[!]]` respectively.  The `-`
  character can be specified inside a character sequence pattern by placing
  it at the start or the end, e.g. `[abc-]`.

So I think it's fine to go with option 1.

tests/cli-rustup.rs
```diff
@@ -1224,6 +1224,21 @@ fn toolchain_uninstall_is_like_uninstall() {
...
+        expect_ok(config, &["rustup", "uninstall", "nightly-*"]);
+        let mut cmd = clitools::cmd(config, "rustup", &["show"]);
+        clitools::env(config, &mut cmd);
+        let out = cmd.output().unwrap();
+        assert!(out.status.success());
+        let stdout = String::from_utf8(out.stdout).unwrap();
+        assert!(!stdout.contains(for_host!("'stable-{}'")));
+        assert!(!stdout.contains(for_host!("'nightly-2015-01-01-{}'")));
+    });
+}
```
camelid, line 1240: I just copied this test code from other tests and modified it. Is the test correct?

camelid (commented):
I think all this needs is some documentation, and then it's ready to go! Where should I put the docs?

camelid:
@kinnison Could you review this again? Thanks!

src/config.rs
```diff
@@ -362,6 +364,16 @@ impl Cfg {
         Toolchain::from(self, name)
     }
 
+    pub fn get_toolchains_from_glob(
+        &self,
+        pattern: Pattern,
```
kinnison, line 369: Is it right for this to take `Pattern` rather than `&Pattern` ?  I realise in this case it's not massively relevant but as a general API design I feel like `&Pattern` might be right.

src/config.rs
```diff
@@ -362,6 +364,16 @@ impl Cfg {
...
+    ) -> Result<impl Iterator<Item = Toolchain<'_>>> {
+        Ok(self
+            .list_toolchains_iter()?
+            .filter(move |toolchain| pattern.matches(toolchain))
```
kinnison, line 373: If we took `&Pattern` I think this `move` might go away.

src/config.rs
```diff
@@ -362,6 +364,16 @@ impl Cfg {
...
+            .map(move |toolchain| Toolchain::from(self, &toolchain).unwrap()))
```
kinnison, line 374: Why does this closure need `move` ?

tests/cli-rustup.rs
kinnison: You probably want to start the test by installing and asserting that stable and nightly *are* there.  Otherwise I think this is fine.

tests/mock/clitools.rs
```diff
@@ -99,9 +99,9 @@ pub fn setup(s: Scenario, f: &dyn Fn(&mut Config)) {
     env::remove_var("RUSTUP_TOOLCHAIN");
     env::remove_var("SHELL");
     env::remove_var("ZDOTDIR");
-    // clap does it's own terminal colour probing, and that isn't
+    // clap does its own terminal colour probing, and that isn't
     // trait-controllable, but it does honour the terminal. To avoid testing
-    // claps code, lie about whatever terminal this process was started under.
+    // clap's code, lie about whatever terminal this process was started under.
```
kinnison, line 104: While these fixes are correct, they seem a little odd in this PR :D

src/config.rs
camelid: We could theoretically take `&Pattern`, but then we run into lifetime issues, so I didn't think it was worth it given that it will make the code more complex. I am still open to changing it though.

src/config.rs
camelid: I think because otherwise it reborrows `self`? I don't know enough about how reborrows work, but this is the error:

```
error[E0373]: closure may outlive the current function, but it borrows `self`, which is owned by the current function
   --> src/config.rs:374:18
    |
374 |             .map(|toolchain| Toolchain::from(self, &toolchain).unwrap()))
    |                  ^^^^^^^^^^^                 ---- `self` is borrowed here
    |                  |
    |                  may outlive borrowed value `self`
    |
note: closure is returned here
   --> src/config.rs:370:10
    |
370 |     ) -> Result<impl Iterator<Item = Toolchain<'_>>> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: to force the closure to take ownership of `self` (and any other referenced variables), use the `move` keyword
    |
374 |             .map(move |toolchain| Toolchain::from(self, &toolchain).unwrap()))
    |                  ^^^^^^^^^^^^^^^^
```

tests/mock/clitools.rs
camelid: Yeah, I noticed them while I was working on some code, and it's sometimes annoying to open a whole new PR just to fix two typos in a comment :)

src/config.rs
kinnison: Aah because the iterator gets returned, fair enough.

src/config.rs
kinnison: I see, okay, no problem

rami3l:
Closing in favor of https://github.com/rust-lang/rustup/pull/3862, as discussed in https://github.com/rust-lang/rustup/issues/3519#issuecomment-1784723949. Thanks a lot for helping out nonetheless!

djc:
@rami3l can you elaborate on your rationale for linking this to #3862? Seems like this is still a reasonable request?

rami3l:
> @rami3l can you elaborate on your rationale for linking this to #3862? Seems like this is still a reasonable request?

@djc The idea is that one would be able to do `rustup toolchain list --quiet | grep nightly-2023- | xargs rustup uninstall` after #3862. Not sure if this addition is still worth it 🤔

If the demand is still there we can make another PR with @camelid as the co-author, I have no problem with that.

est31:
It might still make sense to have glob support. At the least one should print the command you suggested when someone tries to use globs with rustup: it's not trivial to come up with it, nor is it guaranteed to always work in the future.

rami3l:
> It might still make sense to have glob support. At the least one should print the command you suggested when someone tries to use globs with rustup: it's not trivial to come up with it, nor is it guaranteed to always work in the future.

@est31 Indeed, I've added #2530 back to our roadmap.

PR #2543, opened 2020-10-31
fix typo:  rust-toolchain, not rust*up*-toolchain

MaulingMonkey:
Rest of the file correctly refers to `rust-toolchain` already.


kinnison:
Oopsie, well spotted!

PR #2544, opened 2020-10-31
Add docs for proxies

### LINKED ISSUE #2538: List proxies in the rustup book
As mentioned in #2532 it would be good to list all the proxies and what they are and which component contains them - this should go into the book somewhere fairly obvious.

This should be easy to do, and I'd like it to be in 1.23.0

chansuke:
Fixes #2538.

### DIFF
--- a/doc/src/concepts/proxies.md
+++ b/doc/src/concepts/proxies.md
@@ -1,0 +1,23 @@
+# Proxies
+
+`rustup` provides a number of wrappers for common Rust tools.
+These are called _proxies_ and represent commands which are
+provided by the various [components][].
+
+The list of proxies is currently static in `rustup` and is as follows:
+
+[components]: concepts/components.html
+
+- `rustc` is the compiler for the Rust programming language, provided by the project itself and comes from the `rustc` component.
+
+- `rustdoc` is a tool distributed in the `rustc` component which helps you to generate documentation for Rust projects.
+
+- `cargo` is the Rust package manager which downloads your Rust package’s dependencies, compiles your packages, makes distributable packages, and uploads them to crates.io (the Rust community’s package registry). It comes from the `cargo` component.
+
+- `rust-lldb` and `rust-gdb` are simple wrappers around the `lldb` and `gdb` debuggers respectively. The wrappers enable some pretty-printing of Rust values and add some convenience features to the debuggers by means of their scripting interfaces.
+
+- `rls` and `rust-analyzer` are two different generations of IDE integration tooling. They implement the language-server protocol to permit IDEs and editors such as Visual Studio Code, ViM, or Emacs, access to the semantics of the Rust code you are editing. They come from their respectively named components.
+
+- `cargo-clippy` and `clippy-driver` are related to the `clippy` linting tool which provides extra checks for common mistakes and stylistic choices and it comes from the `clippy` component.
+
+- `cargo-miri` is an experimental interpreter for Rust's mid-level intermediate representation (MIR) and it comes from the `miri` component.


doc/src/proxies.md
```diff
@@ -0,0 +1,23 @@
+# Proxies
+
+`rustup` provides proxy wrapper through [components](https://github.com/rust-lang/rustup/blob/master/doc/src/concepts/components.md) and you can see the list of the proxies [here](https://github.com/rust-lang/rustup/blob/22dc71adbd7e19de0090f991b3609e5f14910dcb/src/lib.rs#L17-L28)
```
kinnison: This might read slightly better as something like:

```markdown
`rustup` provides a number of wrappers for common Rust tools.
These are called _proxies_ and represent commands which are
provided by the various [components][].

The list of proxies is currently static in `rustup` and is as follows:

[components]: concepts/components.html
```

doc/src/proxies.md
```diff
@@ -0,0 +1,23 @@
...
+
+- `rustc`: `rustc` is a proxy for the compiler for the Rust programming language, provided by the project itself and it matches `rustc` component.
```
kinnison: I think, given this is explicitly already a list of proxies, we can simplify the language here a little:

```suggestion
- `rustc` is the compiler for the Rust programming language, provided by the project itself and comes from the `rustc` component.
```

Note how I don't say "proxy" and I don't bother repeating the name of the proxy either, there's simply no need.

kinnison: You can make similar changes to all the lines.  No need to keep saying "proxy"

doc/src/proxies.md
```diff
@@ -0,0 +1,23 @@
...
+
+- `rustdoc`: `rustdoc` is a proxy for the standard Rust distribution ships with a tool called rustdoc that generates documentation for Rust projects and it matches `rustc` component.
```
kinnison: Again the language here is very complex, a simpler example could be:

```suggestion
- `rustdoc` is a tool distributed in the `rustc` component which helps you to generate documentation for Rust projects.
```

kinnison: Similar simplifications can be made to other entries.

In general, try not to be over-complex in your language or description, simpler is better at this point.

kinnison (changes requested):
Thank you for starting the work on this.  It's a good start, but the language could do with some tweaks to make it slightly easier for others to consume.

In addition, perhaps it should be in the `concepts/` subtree of the docs?  I'm not sure, but it feels like it should be, just like the `components` file is.

doc/src/proxies.md
chansuke: @kinnison I appreciate your feedback. I agree that moving the doc under the `concepts/` subtree is more descriptive and clear.

doc/src/concepts/proxies.md
```diff
@@ -0,0 +1,29 @@
...
+
+`rustup` provides a number of wrappers for common Rust tools.
+These are called _proxies_ and represent commands which are
+provided by the various [components][].
+
+The list of proxies is currently static in `rustup` and is as follows:
+
+[components]: concepts/components.html
+
+- `rustc`: `rustc` is the compiler for the Rust programming language, provided by the project itself and come the from the `rustc` component.
```
kinnison: This is still repeating the `rustc` bit.  Simpler would be:

```suggestion
- `rustc` is the compiler for the Rust programming language, provided by the project itself and come the from the `rustc` component.
```

kinnison: Similar simplifications can be done throughout and also will resolve some of the consistency issues below where you sometimes use backticks on command names and sometimes fail to.

doc/src/concepts/proxies.md
```diff
@@ -0,0 +1,29 @@
...
+
+- `rustdoc`: `rustdoc` is a tool distributed in the `rustc` component which helps you to generate documentation for Rust projects.
+
+- `cargo`: cargo is the rust package manager which downloads your Rust package’s dependencies, compiles your packages, makes distributable packages, and uploads them to crates.io and the Rust community’s package registry and it comes from the `cargo` component.
```
kinnison: Here you refer to 'rust' and to 'Rust' in the same sentence.  It'd be better to capitalise Rust consistently.

doc/src/concepts/proxies.md
```diff
@@ -0,0 +1,29 @@
...
+
+- `rust-lldb`: rust-lldb is a debugger component of the LLVM project. It is built as a set of reusable components that extensively use existing libraries from LLVM, such as the Clang expression parser and LLVM disassembler and it comes from the `rustc` component.
+
+- `rust-gdb`: rust-gdb helps you to pretty print the program with GDB python. You can see the detail in [here](https://github.com/rust-lang/rust/blob/master/src/etc/rust-gdb).It comes from the `rustc` component.
```
kinnison: Are these descriptions lifted from somewhere else?  They feel a little disjointed and confusing.  `rust-lldb` is as much a wrapper as `rust-gdb` and they're both there simply to make it easier to debug Rust code using the `lldb` and `gdb` tools respectively.  I'd be tempted to wrap both of these up with a single bullet point of the form:

```markdown
- `rust-lldb` and `rust-gdb` are simple wrappers around the `lldb` and `gdb` debuggers respectively.  The wrappers enable some pretty-printing of Rust values and add some convenience features to the debuggers by means of their scripting interfaces.
````

File: doc/src/concepts/proxies.md [hunk 7]
```diff
@@ -0,0 +1,29 @@
...
+
+- `rls`: rls is a server that runs in the background, provides IDEs, editors, and other tools with information about Rust programs. It supports functionality such as 'goto definition', symbol search, reformatting, and code completion, and enables renaming and refactorings and it comes from the `rls` component.
+
+- `rust-analyzer`: rust-analyzer is an IDE integration which is an experimental modular compiler frontend for the Rust language.
```
kinnison: Again I'd be tempted to combine these two together:

```markdown
- `rls` and `rust-analyzer` are two different generations of IDE integration tooling.  They implement the language-server protocol to permit IDEs and editors such as Visual Studio Code, ViM, or Emacs, access to the semantics of the Rust code you are editing.  They come from their respectively named components.
```

kinnison (changes requested):
I apologise for taking so long to get around to this.  When you first updated, I assumed you were going to rebase your changes after making them so I didn't review the content immediately.

I think you're getting closer to what I feel would be good for this documentation.  Thank you for continuing to work on this.

doc/src/concepts/proxies.md
chansuke: Thank you for your review and help. Yeah, I read the documents of LLDB & GDB and source code and tried to explain it as much as possible but that's too disjoiinted and unhelpful.

doc/src/concepts/proxies.md
```diff
@@ -0,0 +1,25 @@
...
+
+`rustup` provides a number of wrappers for common Rust tools.
+These are called _proxies_ and represent commands which are
+provided by the various [components][].
+
+The list of proxies is currently static in `rustup` and is as follows:
+
+[components]: concepts/components.html
+
+- `rustc` is the compiler for the Rust programming language, provided by the project itself and come from the `rustc` component.
```
kinnison:
```suggestion
- `rustc` is the compiler for the Rust programming language, provided by the project itself and comes from the `rustc` component.
```

doc/src/concepts/proxies.md
```diff
@@ -0,0 +1,25 @@
...
+
+- `rustdoc` is a tool distributed in the `rustc` component which helps you to generate documentation for Rust projects.
+
+- `cargo` is the Rust package manager which downloads your Rust package’s dependencies, compiles your packages, makes distributable packages, and uploads them to crates.io and the Rust community’s package registry and it comes from the `cargo` component.
```
kinnison:
```suggestion
- `cargo` is the Rust package manager which downloads your Rust package’s dependencies, compiles your packages, makes distributable packages, and uploads them to crates.io (the Rust community’s package registry). It comes from the `cargo` component.
```

doc/src/concepts/proxies.md
```diff
@@ -0,0 +1,25 @@
...
+
+- `rust-lldb` and `rust-gdb` are simple wrappers around the `lldb` and `gdb` debuggers respectively. The wrappers enable some pretty-printing of Rust values and add some convenience features to the debuggers by means of their scripting interfaces.
+
+- `rls` and `rust-analyzer` are two different generations of IDE integration tooling. They implement the language-server protocol to permit IDEs and editors such as Visual Studio Code, ViM, or Emacs, access to the semantics of the Rust code you are editing. They come from their respectively named components.
+
+- `cargo-clippy` is a lint tool which provides extra checks for common mistakes and stylistic choices and it comes from the `clippy` component.
+
+- [clippy-driver](https://github.com/rust-lang/rust-clippy/blob/master/src/driver.rs) is a driver that gives description of the provided lint with an option and also checks a package and it comes from the `clippy` component.
```
kinnison: I think `clippy-driver` is actually semi-deprecated so perhaps combine it into the `cargo-clippy` line with:

```markdown
- `cargo-clippy` and `clippy-driver` are related to the `clippy` linting tool which ...
```

kinnison (changes requested):
Getting closer.  Perhaps after your next tweaks you could rebase your commits too, so that if the text is good, we might be able to merge.

kinnison (approved):
This looks great, thank you for your effort.

PR #2545, opened 2020-10-31
Add env variable to override the fallback settings file path

### LINKED ISSUE #2456: Insufficient test isolation for `/etc/rustup/settings.toml` files
**Problem**
When we added the fallback settings capability, we failed to make it properly isolated for testing :(. As such, an `/etc/rustup/settings.toml` file that specifies `stable` as the toolchain will cause `update` (soon to be `update_once` in cli-exact.rs to fail, because the default toolchain *will* be set before the initial update.

Since we need this to be per-test, we probably want to allow an env variable to override the path to it, then set that up in the mock test support glue.

frazar:
This PR introduces the support for overriding the path of the fallback settings file (`/etc/rustup/settings.toml`) by specifying the `RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS` environment variable.

The fix can be verified by running the following commands
```
$ # Create an `/etc/rustup/settings.toml` file specifying a default toolchain
$ echo "default_toolchain = 'stable'" | sudo tee /etc/rustup/settings.toml
default_toolchain = 'stable'
$ cargo test update_once # Fails!
$ export RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS=/dev/null 
$ cargo test update_once # Does not fail
```

Fixes #2456 

Issue #2456  also mentions that the env variable should be "set up in the mock test support glue", but I'm not sure what it means..

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -196,7 +196,14 @@ impl Cfg {
 
         // Centralised file for multi-user systems to provide admin/distributor set initial values.
         let fallback_settings = if cfg!(not(windows)) {
-            FallbackSettings::new(PathBuf::from(UNIX_FALLBACK_SETTINGS))?
+            // If present, use the RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS environment
+            // variable as settings path, or UNIX_FALLBACK_SETTINGS otherwise
+            FallbackSettings::new(
+                match process().var("RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS") {
+                    Ok(s) => PathBuf::from(s),
+                    Err(_) => PathBuf::from(UNIX_FALLBACK_SETTINGS),
+                },
+            )?
         } else {
             None
         };
--- a/tests/cli-rustup.rs
+++ b/tests/cli-rustup.rs
@@ -1940,3 +1940,41 @@ fn check_host_goes_away() {
         );
     })
 }
+
+#[cfg(unix)]
+#[test]
+fn check_unix_settings_fallback() {
+    setup(&|config| {
+        // No default toolchain specified yet
+        expect_err(
+            config,
+            &["rustup", "default"],
+            r"no default toolchain configured",
+        );
+
+        // Default toolchain specified in fallback settings file
+        let mock_settings_file = config.current_dir().join("mock_fallback_settings.toml");
+        raw::write_file(
+            &mock_settings_file,
+            for_host!(r"default_toolchain = 'nightly-{0}'"),
+        )
+        .unwrap();
+
+        let mut cmd = clitools::cmd(config, "rustup", &["default"]);
+        clitools::env(config, &mut cmd);
+
+        // Override the path to the fallback settings file to be the mock file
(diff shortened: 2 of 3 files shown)


src/config.rs
```diff
@@ -196,7 +196,12 @@ impl Cfg {
 
         // Centralised file for multi-user systems to provide admin/distributor set initial values.
         let fallback_settings = if cfg!(not(windows)) {
-            FallbackSettings::new(PathBuf::from(UNIX_FALLBACK_SETTINGS))?
+            // If present, use the RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS environment
+            // variable as settings path, or UNIX_FALLBACK_SETTINGS otherwise
+            FallbackSettings::new(PathBuf::from(
+                option_env!("RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS")
+                    .unwrap_or(UNIX_FALLBACK_SETTINGS),
```
kinnison: `option_env!()` is a compile-time expansion of the environment variable.

You probably meant to retrieve the runtime environment which we do via our process code as:

```rust
process().var("RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS").unwrap_or(UNIX_FALLBACK_SETTINGS)
```

kinnison (commented):
Thank you for trying to fix this issue.

There's a minor snafu in your code which will be easy for you to fix, however the important reason for wanting this feature is so that our test code doesn't use `/etc/rustup/settings.toml` when testing `rustup`.  In here: https://github.com/rust-lang/rustup/blob/master/tests/mock/clitools.rs#L410 you will see how we set up the environment to run test code, at minimum setting the environment variable to something we know won't exist might be a good starting point.  However this could also be an opportunity to introduce the ability to test the fallback code.

kinnison:
While it pains me to say, you didn't format the code or test it before submitting that.  There's a little more work you need to do, likely involving a bit of shifting around of the `PathBuf::from` and friends before it'll compile.

Also, have you given any thought to a way we might test this?

frazar:
Sorry, I didn't mean to actually push this! It's still a work in progress, I'll update the PR title accordingly. 

Thanks for your advices!

frazar:
I think this PR is ready for a second review!

The unit test setup was updated to set  the `RUSTUP_OVERRIDE_UNIX_FALLBACK_SETTINGS` variable with a bogus value, as suggested. This should make sure that the actual fallback settings file does not interefere with testing.

Also, a unit test was added for Unix plarforms where the fallback settings mechanism is tested.

Let me know if further changes are required e.g. squashing the commits.

kinnison (approved):
I think this looks good.  @rbtcollins since this touches the process/env stuff I'd like you to just run your eye over it; but I think it's right.

kinnison:
Content-wise it looks good.  If @rbtcollins is okay with it, then I'd like it rebased into a neater commit if you don't mind.

frazar:
Commits squashed! Let me know if anything else needs fixing

kinnison:
I'm going to assume Robert is too busy right now, and so since I think it's good I'll merge it, thank you.

PR #2551, opened 2020-11-12
Allow installing `major.minor` channels

### LINKED ISSUE #794: Allow omitting patch version in toolchain specifiers
Example use cases:

## Installation / update
```
$ rustup install 1.8
error: toolchain '1.8' is not installed
```

(I've filed #793 for the confusing error message.)

I think this should install the highest version matching [semver requirement](https://github.com/steveklabnik/semver#requirements) `~1.8`. For the recent case of `1.12.1`, installing `1.12` should therefore install `1.12.1`. `1.12.1` should be installed even if the user has `1.12.0` installed.

## Run

```
$ rustup run 1.8 rustc test.rs
error: toolchain '1.8' is not installed
```

This should be equivalent to specifying the highest *installed* version matching `~1.8`. In this case, I expect it to be equivalent to `rustup run 1.8.0 cargo test`.

For `1.12`, if the user has `1.12.0` installed, then

```
$ rustup run 1.12 rustc test.rs
```

should run with `1.12.0`, but if they have `1.12.1` installed, it should run with `1.12.1`.

Together with #792, this leads to really confusing behavior: `rustup run 1.8 cargo test` will silently use nightly, since even if you have `1.8.0` installed, that's *not* `1.8`.

carols10cents:
This fixes #794!!! 🎉 🌮 

I would recommend not merging+releasing this until after 1.48 is released, because that's the first stable release that will [write out the new manifests](https://github.com/rust-lang/rust/pull/76107). We could even wait for a patch release to happen, to verify that the patch release overwrites the major.minor manifests correctly, but that's not predictable, so it's up to you.

I tested this out locally installing into a `home` subdir and [since the manifests have been backfilled for old versions](https://github.com/rust-lang/rust-central-station/pull/954), I was able to do the following:

```
$ home/bin/rustup install 1.47 # install a channel that hasn't had point releases
info: syncing channel updates for '1.47-x86_64-apple-darwin'
info: latest update on 2020-10-08, rust version 1.47.0 (18bf6b4f0 2020-10-07)
...snip...
  1.47-x86_64-apple-darwin installed - rustc 1.47.0 (18bf6b4f0 2020-10-07)

$ home/bin/rustup toolchain list # list all toolchains
nightly-2020-04-22-x86_64-apple-darwin
1.47-x86_64-apple-darwin

$ home/bin/rustup toolchain uninstall 1.47 # uninstall a toolchain of this style
info: uninstalling toolchain '1.47-x86_64-apple-darwin'
info: toolchain '1.47-x86_64-apple-darwin' uninstalled

$ home/bin/rustup install 1.45 # install a channel that has had point releases, I got the latest one
info: syncing channel updates for '1.45-x86_64-apple-darwin'
info: latest update on 2020-08-03, rust version 1.45.2 (d3fb005a3 2020-07-31)
...snip...
  1.45-x86_64-apple-darwin installed - rustc 1.45.2 (d3fb005a3 2020-07-31)

$ home/bin/rustup run 1.45 rustc --version
rustc 1.45.2 (d3fb005a3 2020-07-31)

$ home/bin/rustc +1.45 --version
rustc 1.45.2 (d3fb005a3 2020-07-31)
```

Please let me know if there's anything else I should test out manually!

### DIFF
--- a/doc/src/concepts/toolchains.md
+++ b/doc/src/concepts/toolchains.md
@@ -15,15 +15,15 @@ Standard release channel toolchain names have the following form:
 ```
 <channel>[-<date>][-<host>]
 
-<channel>       = stable|beta|nightly|<version>
+<channel>       = stable|beta|nightly|<major.minor>|<major.minor.patch>
 <date>          = YYYY-MM-DD
 <host>          = <target-triple>
 ```
 
-'channel' is either a named release channel or an explicit version number,
-such as `1.42.0`. Channel names can be optionally appended with an archive
-date, as in `nightly-2014-12-18`, in which case the toolchain is downloaded
-from the archive for that date.
+'channel' is a named release channel, a major and minor version number such as
+`1.42`, or a fully specified version number, such as `1.42.0`. Channel names
+can be optionally appended with an archive date, as in `nightly-2014-12-18`, in
+which case the toolchain is downloaded from the archive for that date.
 
 Finally, the host may be specified as a target triple. This is most useful for
 installing a 32-bit compiler on a 64-bit platform, or for installing the
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -55,14 +55,15 @@ pub static TOOLCHAIN_HELP: &str = r"DISCUSSION:
 
         <channel>[-<date>][-<host>]
 
-        <channel>       = stable|beta|nightly|<version>
+        <channel>       = stable|beta|nightly|<major.minor>|<major.minor.patch>
         <date>          = YYYY-MM-DD
         <host>          = <target-triple>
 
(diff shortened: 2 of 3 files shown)


kinnison (approved):
Thanks for working on this @carols10cents :heart: 

I would like to see some integration testing for this, to ensure things like `rustup update` will install updated channels etc. however since there's nothing obvious to me to suggest this may not work, I won't necessarily block on doing that since the current integration testing framework can be a bit of a pain to update if you're not up for the faff :grin: 

I certainly appreciate the unit tests.

@rbtcollins I'm minded to merge this, even if @carols10cents isn't up for doing an integration test.  How do you feel?

kamalmarhubi:
Thanks @carols10cents <3

PR #2555, opened 2020-11-17
Validate target triple in build script

### LINKED ISSUE #2376: Improve error messages for unsupported platforms.

kinnison:
I have factored `PartialTargetTriple` out into a separate file and then used that in `build.rs`.

Also by emitting `rerun-if...` we might make the build process a tad faster since we won't keep rerunning the build script if we don't need to.

This fixes #2376

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,30 @@
 use std::env;
 
+include!("src/dist/triple.rs");
+
+pub fn from_build() -> Result<PartialTargetTriple, String> {
+    let triple = if let Ok(triple) = env::var("RUSTUP_OVERRIDE_BUILD_TRIPLE") {
+        triple
+    } else {
+        env::var("TARGET").unwrap()
+    };
+    PartialTargetTriple::new(&triple).ok_or(triple)
+}
+
 fn main() {
+    println!("cargo:rerun-if-env-changed=RUSTUP_OVERRIDE_BUILD_TRIPLE");
+    println!("cargo:rerun-if-env-changed=TARGET");
+    match from_build() {
+        Ok(triple) => eprintln!("Computed build based partial target triple: {:#?}", triple),
+        Err(s) => {
+            eprintln!("Unable to parse target '{}' as a PartialTargetTriple", s);
+            eprintln!(
+                "If you are attempting to bootstrap a new target you may need to adjust the\n\
+               permitted values found in src/dist/triple.rs"
+            );
+            std::process::abort();
+        }
+    }
     let target = env::var("TARGET").unwrap();
     println!("cargo:rustc-env=TARGET={}", target);
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,10 @@ version = "0.3"
 [dev-dependencies]
 walkdir = "2"
 
+[build-dependencies]
+lazy_static = "1"
+regex = "1"
+
 [workspace]
 members = ["download"]
 
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -14,6 +14,7 @@ use crate::dist::manifestation::{Changes, Manifestation, UpdateStatus};
 use crate::dist::notifications::*;
 use crate::dist::prefix::InstallPrefix;
 use crate::dist::temp;
+pub use crate::dist::triple::*;
 use crate::errors::*;
 use crate::process;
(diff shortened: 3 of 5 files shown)


build.rs
```diff
@@ -1,6 +1,30 @@
...
 
+include!("src/dist/triple.rs");
+
+pub fn from_build() -> Result<PartialTargetTriple, String> {
+    if let Some(triple) = env::var("RUSTUP_OVERRIDE_BUILD_TRIPLE").ok() {
+        PartialTargetTriple::new(&triple).ok_or(triple)
+    } else {
+        let triple = env::var("TARGET").unwrap();
+        PartialTargetTriple::new(&triple).ok_or(triple)
+    }
```
jyn514:
```suggestion
    let triple = env::var("RUSTUP_OVERRIDE_BUILD_TRIPLE").unwrap_or_else(|| env::var("TARGET").unwrap());
    PartialTargetTriple::new(&triple).ok_or(triple)
```

build.rs
```diff
@@ -1,6 +1,30 @@
...
+    println!("cargo:rerun-if-env-changed=TARGET");
+    match from_build() {
+        Ok(triple) => eprintln!("Computed build based partial target triple: {:#?}", triple),
+        Err(s) => {
+            eprintln!("Unable to parse target '{}' as a PartialTargetTriple", s);
+            eprintln!(
+                "If you are attempting to bootstrap a new target you may need to adjust the\n\
+               permitted values found in src/dist/triple.rs"
+            );
+            std::process::abort();
```
jyn514, line 25: Hmm, is there any reason you chose to abort instead of exiting?
```suggestion
            std::process::exit(1);
```

build.rs
```diff
@@ -1,6 +1,30 @@
...
+        PartialTargetTriple::new(&triple).ok_or(triple)
+    } else {
+        let triple = env::var("TARGET").unwrap();
+        PartialTargetTriple::new(&triple).ok_or(triple)
+    }
+}
+
 fn main() {
+    println!("cargo:rerun-if-env-changed=RUSTUP_OVERRIDE_BUILD_TRIPLE");
+    println!("cargo:rerun-if-env-changed=TARGET");
```
jyn514, line 16: I think you might also need to tell cargo if the build script changes? Not sure if it infers that automatically.
```suggestion
    println!("cargo:rerun-if-env-changed=TARGET");
    println!("cargo:rerun-if-changed=build.rs");
```

jyn514 (commented):
You mentioned in the issue you were going to make this a unit test - any reason to put it in build.rs instead?

(also, in case anyone else wants to review the moved files, https://github.com/rust-lang/rust/issues/76268#mentoring-instructions-for-reviewers should be helpful)

build.rs
kinnison: Cargo always tracks the *source* inputs to the build script automatically.

build.rs
kinnison: I chose to abort because I wasn't *certain* that non-zero was correct on all platforms.

I can change it if you really want, but this seemed easiest/clearest.  I also considered just panic!()ing

build.rs
kinnison: While that is fewer lines of code, I'm not sure it's necessarily clearer.  I'll go for deduplicating the call to `::new` but I will use the `if let` rather than the `unwrap_or_else` approach.

build.rs
jyn514: Hmm, ok. I guess rust doesn't have an EXIT_FAILURE like C does, I'm not used to thinking about portability :smile:

PR #2560, opened 2020-11-21
chore: Temporarily back out rust-analyzer proxy

kinnison:
For the 1.23.0 release we are temporarily backing out the rust-analyzer
proxy as r-a will only be on nightly at this point and so it's better if
we don't confuse people into expecting it on stable.  We'll reintroduce
it by reverting this commit when we're ready.

Signed-off-by: Daniel Silverstone (WSL2) <dsilvers@digital-scurf.org>

### DIFF
--- a/ci/snapcraft.bash
+++ b/ci/snapcraft.bash
@@ -43,7 +43,7 @@ fi
 # as rustup.$PROXY.  If we have an alias which is not a supported proxy name
 # then rustup might get sad.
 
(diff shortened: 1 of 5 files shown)


kinnison:
@pietroalbini Assuming this is okay, I think it's the last code change we *need* for 1.23 - I'll then sort out the changelog (which needs the r-a mention backing out) and then look at blog posts.

jyn514 (approved):

PR #2561, opened 2020-11-21
dist: Ensure we treat two-part versions as tracking

kinnison:
Without this, `rustup update` will not update two-part version channels such as `1.48`.

This is a necessary extension of #2551

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -464,9 +464,16 @@ impl ToolchainDesc {
         }
     }
 
+    /// Toolchain channels are considered 'tracking' if it is one of the named channels
+    /// such as `stable`, or is an incomplete version such as `1.48`, and the
+    /// date field is empty.
     pub fn is_tracking(&self) -> bool {
         let channels = ["nightly", "beta", "stable"];
-        channels.iter().any(|x| *x == self.channel) && self.date.is_none()
+        lazy_static! {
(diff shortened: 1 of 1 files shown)


kinnison:
I cannot release 1.23.0 of rustup claiming to support two-part versions unless we merge this I think.

src/dist/dist.rs
```diff
@@ -464,9 +464,16 @@ impl ToolchainDesc {
...
     }
 
+    /// Toolchain channels are considered 'tracking' if it is one of the named channels
+    /// such as `stable`, or is an incomplete version such as `1.48`, and the
+    /// date field is empty.
     pub fn is_tracking(&self) -> bool {
         let channels = ["nightly", "beta", "stable"];
-        channels.iter().any(|x| *x == self.channel) && self.date.is_none()
+        lazy_static! {
+            static ref TRACKING_VERSION: Regex = Regex::new(r"^\d{1}\.\d{1,3}$").unwrap();
```
jyn514, line 473: Hmm, I guess the minor version needs 3 digits for when Rust gets a `1.100` release. Can you add that as a comment?

I think 3 digits is fine - back of the envelope math says we'd need 100 years of rust releases to run out of numbers, which seems like more than far enough away.

carols10cents (approved):
LGTM! Thank you, I didn't realize this needed to be changed too ❤️

src/dist/dist.rs
kinnison: We've been 3 digits in minor for a while now.  If we have to change it, it's not hard :D

PR #2562, opened 2020-11-21
UX: Report non-installable toolchains

### LINKED ISSUE #2547: Trying to install 1.8.0 on freebsd succeeds without actually installing the toolchain
**Steps**

1. Beg, borrow, or steal a FreeBSD machine
2. Install rustup.
3. `rustup install 1.8.0`

```
info: sync channel updates ...
info: latest update on 2016-04-12, rust version 1.8.0 ...

  1.8.0-x86_64-unknown-freedbsd unchanged - (toolchain not installed)
```

**Possible Solution(s)**

Rustup should either install the toolchain or give an error that it's not available.

**Notes**

Output of `rustup --version`: 1.22.1
Output of `rustup show`: omitted because I can't get copy/paste to work in my VM.

kinnison:
If you pick a default toolchain which is not installable, or if you
try to `rustup toolchain install` something non-installable then we
should report it as such and exit non-zero so scripting can fail.

This fixes #2547 

Starting this as draft because this needs tests but I'm feeling too tired to write some for now.

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -5,7 +5,7 @@ use crate::dist::dist;
 use crate::dist::download::DownloadCfg;
 use crate::dist::prefix::InstallPrefix;
 use crate::dist::Notification;
-use crate::errors::Result;
+use crate::errors::{ErrorKind, Result};
 use crate::notifications::Notification as RootNotification;
 use crate::toolchain::{CustomToolchain, DistributableToolchain, Toolchain, UpdateStatus};
 use crate::utils::utils;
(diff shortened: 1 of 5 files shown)


kinnison:
Hmm, I really need to write the tests for this so that I can finish it out.

kinnison (commented):
I am now satisfied with this patch, assuming green CI I'll un-draft the PR

kinnison:
@jyn514 Are you able to check if this would solve your particular problem?  I can only try on Linux by default.

jyn514:
Worked like a charm! `error: toolchain '1.8.0-freebsd' is not installable`

Thanks for the fix!

J-Vin (approved):

J-Vin:
👍

src/install.rs
```diff
@@ -76,7 +76,12 @@ impl<'a> InstallMethod<'a> {
             (false, _) => UpdateStatus::Unchanged,
         };
 
-        Ok(status)
+        // Final check, to ensure we're installed
+        if !toolchain.exists() {
```
jyn514, line 80: When could this happen? I know for freebsd 1.8 it happened, but do you know why?

kinnison: Yes, it's when all the components are *present* in the channel, but they're marked unavailable for installation.

kinnison: Other approaches I tried were much more complex than this simple cross-check at the end.

PR #2566, opened 2020-11-23
chore: Correct potential panic should the installer time travel into …

### LINKED ISSUE #1984: Panic because `supplied instant is later than self` during download.
While testing the release of 1.19.0 we had an unreproducible instance of the download of a component panicing because `supplied instant is later than self`

Since we don't use `Instant` anywhere in our code, this must be in a dependency, but until we get a traceback we can't do much about it.  This is therefore here to track.

If you can reproduce it, please comment below with a traceback.

kinnison:
…the past

This should fix #1984 or at least ameliorate it.

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -118,7 +118,7 @@ impl DownloadTracker {
         match self.last_sec {
             None => self.last_sec = Some(current_time),
             Some(prev) => {
-                let elapsed = current_time - prev;
+                let elapsed = current_time.saturating_duration_since(prev);
                 if elapsed >= Duration::from_secs(1) {
                     if self.display_progress {
                         self.display();
@@ -166,7 +166,7 @@ impl DownloadTracker {
                 let len = self.downloaded_last_few_secs.len();
                 let speed = if len > 0 { sum / len } else { 0 };
                 let speed_h = Size::new(speed, unit, UnitMode::Rate);
-                let elapsed_h = start_sec.elapsed();
+                let elapsed_h = Instant::now().saturating_duration_since(start_sec);
 
                 // First, move to the start of the current line and clear it.
                 let _ = self.term.carriage_return();
--- a/src/diskio/mod.rs
+++ b/src/diskio/mod.rs
@@ -151,7 +151,9 @@ pub fn perform(item: &mut Item) {
         Kind::Directory => create_dir(&item.full_path),
         Kind::File(ref contents) => write_file(&item.full_path, &contents, item.mode),
     };
-    item.finish = item.start.map(|s| s.elapsed());
+    item.finish = item
+        .start
+        .map(|s| Instant::now().saturating_duration_since(s));
 }
 
 #[allow(unused_variables)]


kinnison:
If either of @sloanp486 or @tbarusseau were able to fetch this branch and try it I'd appreciate it, but if it doesn't actually fix #1984 then they can comment after the release and we can reopen it for further investigation.

tbarusseau:
This might be a dumb question, but the only thing that I can run from this repo (without rust installed) is seemingly `rustup-init.sh`, which seems to download a pre-compiled binary from the internet. How do I test without rust installed?

kinnison:
You need to `cargo run -- --no-modify-path -y` which will build and install a developer version of rustup, but if that's too much risk/faff for you then no worries, I'll just listen out for if things break in the future :D

tbarusseau:
I managed to compile and run it, and it seems to work just fine now. Thanks 😄

cyril23:
> You need to `cargo run -- --no-modify-path -y` which will build and install a developer version of rustup, [...]

doesn't really work for me:

```
apt-get install cargo pkg-config libssl-dev
git clone https://github.com/rust-lang/rustup.git && cd rustup/ && cargo run -- --no-modify-path -y
....
  Compiling hyper v0.13.8
   Compiling serde_urlencoded v0.6.1
   Compiling ed25519-dalek v1.0.1
   Compiling toml v0.5.7
   Compiling rs_tracing v1.0.1
   Compiling rsa v0.3.0
   Compiling pgp v0.7.1
   Compiling hyper-tls v0.4.3
   Compiling reqwest v0.10.8
   Compiling download v0.6.7 (/root/rustup/download)
error: attributes are not yet allowed on `if` expressions
   --> src/cli/self_update.rs:304:5
    |
304 |     #[cfg(windows)]
    |     ^^^^^^^^^^^^^^^

error: could not compile `rustup`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

- Debian 10, fresh installed

kinnison:
@cyril23 Rustup is not meant to be built with older compilers.  We tend to follow current stable compiler.

cyril23:
> @cyril23 Rustup is not meant to be built with older compilers. We tend to follow current stable compiler.

so my Debian packages are too old, I understand.

By the way I had similar errors like https://github.com/rust-lang/rustup/issues/1984#issuecomment-731605699 when trying to install rust the recommended way via `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh` after a fresh Debian 10 installation (`uname -r` shows `4.19.0-12-amd64`) in Windows Virtual Box: [rust_error.log](https://github.com/rust-lang/rustup/files/5598858/rust_error.log)

kinnison:
@cyril23 I think that virtualbox must have a broken sense of time.  I hope that the new release will mitigate that when we can make a release.

cyril23:
> @cyril23 I think that virtualbox must have a broken sense of time. 

Yes, you're right. It's actually the fault of Virtual Box. I've just tried it out again in VMware Player with a fresh Debian 10, and it installed without any problems. [rust_vmware.log](https://github.com/rust-lang/rustup/files/5599215/rust_vmware.log)

PR #2567, opened 2020-11-24
Fix proxies docs.

ehuss:
There were a few minor mistakes in #2544, most notably that the file wasn't added to SUMMARY.md which means it doesn't render at all.

### DIFF
--- a/doc/src/SUMMARY.md
+++ b/doc/src/SUMMARY.md
@@ -10,6 +10,7 @@
     - [Toolchains](concepts/toolchains.md)
     - [Components](concepts/components.md)
     - [Profiles](concepts/profiles.md)
+    - [Proxies](concepts/proxies.md)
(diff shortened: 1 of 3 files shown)


kinnison (approved):
Thanks for this.

I wonder if it'd be possible to render the docs in a PR and somehow view them?  Gitlab lets you view artifacts for a CI pipelie, is there an equivalent in github actions?

ehuss:
There looks like there is an [upload-artifact action](https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts), but it requires a REST API to fetch the artifacts.  I have never used it, so I am not too familiar with it.

BTW, I'm still planning to implement the ability to view the "master" docs (after PRs merge to master), but I think I might wait till after the release, since that will make it easier once the `stable` branch has been updated.

PR #2573, opened 2020-11-25
Toolchain reinstalling fix 2571

### LINKED ISSUE #2571: New toolchain file format causes unneeded channel syncs
Reported by @Nemo157 here: https://internals.rust-lang.org/t/new-rustup-release-staged-and-needing-testers/13446/3

The unnecessary syncs are repeatable *and* break if no internet connection is available.

This needs fixing before 1.23.0 can go out.

kinnison:
This fixes #2571 by rewriting the check for "is this already installed"

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -625,17 +625,37 @@ impl Cfg {
             targets: &[&str],
         ) -> Result<bool> {
             let components_requested = !components.is_empty() || !targets.is_empty();
-
+            // If we're here, the toolchain exists on disk and is a dist toolchain
+            // so we should attempt to load its manifest
+            let manifest = if let Some(manifest) = distributable.get_manifest()? {
+                manifest
+            } else {
+                // If we can't read the manifest we'd best try and install
+                return Ok(true);
+            };
             match (distributable.list_components(), components_requested) {
                 // If the toolchain does not support components but there were components requested, bubble up the error
                 (Err(e), true) => Err(e),
-                (Ok(installed_components), _) => {
-                    Ok(components.iter().chain(targets.iter()).all(|name| {
-                        installed_components.iter().any(|status| {
-                            status.component.short_name_in_manifest() == name && status.installed
+                // Otherwise check if all the components we want are installed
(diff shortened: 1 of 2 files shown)


rbtcollins (commented):
This is a little hard to infer from the diff alone. What caused the repeated syncs?

kinnison:
The issue was twofold.

1. We weren't running the component names through the rename tables and so "clippy" being "clippy-preview" in the manifest was missed
2. We weren't treating targets as `rust-std` properly

src/config.rs
```diff
@@ -625,17 +625,37 @@ impl Cfg {
...
         ) -> Result<bool> {
             let components_requested = !components.is_empty() || !targets.is_empty();
-
+            // If we're here, the toolchain exists on disk and is a dist toolchain
+            // so we should attempt to load its manifest
+            let manifest = if let Some(manifest) = distributable.get_manifest()? {
+                manifest
+            } else {
+                // If we can't read the manifest we'd best try and install
+                return Ok(true);
```
rbtcollins, line 634: will this remove the corrupt dist?

kinnison: No it'll bubble the situation up a level which will then try to ensure the component set in question and report something along the lines of:

```
error: toolchain 'nightly-2020-11-24-x86_64-unknown-linux-gnu' does not support components
```

I think this is fair enough.

rbtcollins: Except that that should support components, and the reason it doesn't is that its corrupt, and we know its corrupt because we expected it to support components, and we returned Ok(true) rather than signalling the impossible situation...

I think this is good enough to do a release with; I don't think its good enough to consider done.

kinnison: Actually no, I got that entirely wrong - argh.  Let me faff.

kinnison: I think when I tested this, I must have accidentally reinstalled the test release rather than the version I was editing.  I need to push a change to make that `Ok(false)` and then instead of being crap about it, it'll heal the installation.

PR #2585, opened 2020-11-29
MacOS: do not depend on Homebrew

jeroen:
See https://github.com/rust-lang/rustup/issues/2583
This should prevent inadvertent runtime dependencies on homebrew libs.


kinnison:
This won't work because the moment the CI workflows are regenerated the change will be lost.  Also it seems a little heavyhanded.  Was #2584 not good enough?

kinnison:
I'm hoping the change I merged which will refuse anything linked to `/usr/local` will be sufficient, we can reopen this if it's not.

PR #2586, opened 2020-11-29
Profile in rust toolchain

### LINKED ISSUE #2579: rust-toolchain should allow setting the equivalent of `--profile minimal`

For things like CI that can make do with the minimal profile but also need additional components, I usually do something like this to install the desired toolchain:

```sh
rust_toolchain="$(cat rust-toolchain)"
rustup toolchain install "$rust_toolchain" --profile minimal --component clippy,rustfmt
rustup default "$rust_toolchain"
```

The new TOML format allows this to be skipped since the first invocation of `rustc` / `cargo` will auto-install the toolchain with the required components, *except* that it does so using the default profile and thus also installs `rust-docs`.

It would be nice if, when rustup installs a new toolchain as a result of a discovered TOML rust-toolchain file, it either defaults to using the minimal profile if `components` is specified (too late now perhaps due to backward compatibility), or the TOML has a `profile` field I can set to `"minimal"`, or some other way.

kinnison:
If I'm right this should fix #2579

@jyn514 Is this what you were expecting / desiring?  If so, you can look at the built artifacts and fetch a test rustup to try.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -35,6 +35,7 @@ struct ToolchainSection {
     channel: Option<String>,
     components: Option<Vec<String>>,
     targets: Option<Vec<String>>,
+    profile: Option<String>,
 }
 
 impl ToolchainSection {
@@ -78,6 +79,7 @@ struct OverrideCfg<'a> {
     toolchain: Option<Toolchain<'a>>,
     components: Vec<String>,
     targets: Vec<String>,
+    profile: Option<dist::Profile>,
 }
 
 impl<'a> OverrideCfg<'a> {
@@ -89,6 +91,10 @@ impl<'a> OverrideCfg<'a> {
             },
             components: file.toolchain.components.unwrap_or_default(),
             targets: file.toolchain.targets.unwrap_or_default(),
+            profile: match file.toolchain.profile {
+                Some(name) => Some(dist::Profile::from_str(&name)?),
+                None => None,
+            },
         })
     }
 }
@@ -667,6 +673,7 @@ impl Cfg {
                         toolchain,
                         components,
                         targets,
+                        profile,
                     },
                     reason,
                 )) => {
@@ -911,6 +918,7 @@ mod tests {
                     channel: Some(contents.into()),
                     components: None,
                     targets: None,
+                    profile: None,
                 }
             }
         );
@@ -922,6 +930,7 @@ mod tests {
 channel = "nightly-2020-07-10"
 components = [ "rustfmt", "rustc-dev" ]
 targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]
+profile = "default"
 "#;
 
         let result = Cfg::parse_override_file(contents);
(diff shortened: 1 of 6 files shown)


src/toolchain.rs
```diff
@@ -685,12 +686,15 @@ impl<'a> DistributableToolchain<'a> {
...
+        profile: Option<Profile>,
     ) -> Result<UpdateStatus> {
         let update_hash = self.update_hash()?;
         let old_date = self.get_manifest().ok().and_then(|m| m.map(|m| m.date));
         InstallMethod::Dist {
             desc: &self.desc()?,
-            profile: self.0.cfg.get_profile()?,
+            profile: profile
+                .map(|p| Ok(p))
+                .unwrap_or_else(|| self.0.cfg.get_profile())?,
```
jyn514:
```suggestion
            profile: profile.map_or_else(Ok, || self.0.cfg.get_profile())?,
```
or if you don't like `map_or_else` :P just changing the closure to `Ok` directly.

tests/cli-rustup.rs
```diff
@@ -1548,6 +1548,35 @@ components = [ "rust-bongo" ]
...
+[toolchain]
+profile = "minimal"
+channel = "nightly"
+"#,
+        )
+        .unwrap();
+        expect_not_stdout_ok(
+            config,
+            &["rustup", "component", "list"],
+            for_host!("rust-docs-{} (installed)"),
```
jyn514, line 1575: This says 'expect the component not to be present'? Looks good if so :)

jyn514 (approved):
Just a tiny nit. Thanks so much for adding this!

jyn514:
> If so, you can look at the built artifacts and fetch a test rustup to try.

~~How can I look at the built artifacts?~~ Found it, it was in the GitHub actions screen. I had to be logged in to download the artifact for some reason.

jyn514:
Tested it out and it works great, just like I'd expect :)

src/toolchain.rs
kinnison: I keep forgetting that variant constructors are functions, I also was unaware of `.map_or_else()` which I now need to add to my brain.

tests/cli-rustup.rs
kinnison: Correct.  it's a tad unintuitive, but the idea is to demonstrate that the profile overrode the user setting.

src/config.rs
```diff
@@ -89,6 +91,10 @@ impl<'a> OverrideCfg<'a> {
             },
             components: file.toolchain.components.unwrap_or_default(),
             targets: file.toolchain.targets.unwrap_or_default(),
+            profile: match file.toolchain.profile {
+                Some(name) => Some(dist::Profile::from_str(&name)?),
+                None => None,
+            },
```
toothbrush7777777: You could use `Option::map` here, right?

kinnison: Mmm, a combination of `.map()` and `.transpose()` might make that work.

toothbrush7777777:
Yeah, I was thinking of `and_then`, not `map`.

src/config.rs
toothbrush7777777: Yeah, I was thinking of `and_then`, not `map`.

kinnison:
I've worked through the code suggestions and if this goes green I'll merge.

PR #2587, opened 2020-11-30
Add missing space in error message

camelid:
Before the error looked like:

> Sometimes not all components are available in any given nightly.If you
> don't need the component, you can remove it with:

Note how it's missing a space after the period after the first sentence.


kinnison (approved):
This looks good, thank you.

PR #2589, opened 2020-11-30
Ensure env file exists

### LINKED ISSUE #2578: rustup is not creating $HOME/.cargo/env
I don't want `$HOME/.cargo/env` being sourced in my `$HOME/.profile` or my `$HOME/.bashrc`.
Still, it is convenient to have `$HOME/.cargo/env` generated since I will source this file anyway under certain circumstances.

However, `$HOME/.cargo/env` is not created when I perform the installation as shown below

```bash
$ curl https://sh.rustup.rs -sSf | sh -s -- --no-modify-path -y
```

Bug or feature?

Thanks a lot,

kinnison:
This should fix #2578

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -144,6 +144,11 @@ pub fn do_add_to_path() -> Result<()> {
     _apply_new_path(new_path)
 }
 
+// Nothing to do for Windows for now
+pub fn do_write_env_files() -> Result<()> {
+    Ok(())
+}
+
 fn _apply_new_path(new_path: Option<String>) -> Result<()> {
     use std::ptr;
     use winapi::shared::minwindef::*;
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -338,6 +338,7 @@ pub fn install(
 
     let install_res: Result<utils::ExitCode> = (|| {
         install_bins()?;
+        do_write_env_files()?;
         if !opts.no_modify_path {
             do_add_to_path()?;
         }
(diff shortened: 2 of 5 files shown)


kinnison:
@workingjubilee opinions?

kinnison:
This appears to be sane to me, and I need to re-prepare 1.23.1 so I'm merging it.

src/cli/self_update/windows.rs
```diff
@@ -144,6 +144,11 @@ pub fn do_add_to_path() -> Result<()> {
     _apply_new_path(new_path)
 }
 
+// Nothing to do for Windows for now
+pub fn do_write_env_files() -> Result<()> {
+    Ok(())
+}
+
```
workingjubilee, line 151: This is a literally useless function i.e. it does nothing except further enmesh the two OS' APIs.

I submitted 3 PRs with the express intention in each one of reducing the frequency of these in the codebase so that Linux and Windows devs would not have to consult enmeshed OS APIs, and I was hardly given even 24 hours to review this, whereas at least one of my PRs was held up for weeks in review for not being pretty enough or not using APIs that had no explanation for their usage.

kinnison: This is a fair assessment.  It does increase coupling in an API you worked very hard to decouple.  I can only apologise for the timing and the presentation of this change, I was working under (perhaps imagined) pressure to get a fix out for users because 1.23.0 broke a behaviour people were reasonably relying on and it had broken a number of CIs (not just this, but also the macos 10.13 problem).

I will work on a PR to decouple this again, I apologise for rushing through something less than ideal.

PR #2595, opened 2020-12-03
chore: Use new platform-support URL

### LINKED ISSUE #2588: Update the platform support URL
[Rustup currently links to the platform support page when it can't install a target](https://github.com/rust-lang/rustup/blob/cc9ce760ec38880e495237d2d96b3fe3dde146d8/src/dist/manifest.rs#L382), but the URL moved.

The platform support page is now included in the "rustc book", which is shipped on doc.rust-lang.org riding the trains. It'd be nice to detect which channel the user is trying to install, and show the link to the proper URL:

```
https://doc.rust-lang.org/{channel}/rustc/platform-support.html
```

For >= 1.47.0 we can also directly use the version number instead of "stable" as the channel.

kinnison:
This fixes #2588

### DIFF
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -13,7 +13,7 @@ case "$(uname -s)" in
   * ) FEATURES=('--features' 'vendored-openssl') ;;
 esac
 
-# rustc only supports armv7: https://forge.rust-lang.org/release/platform-support.html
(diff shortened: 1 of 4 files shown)


kinnison:
@pietroalbini Will this be enough?

emilyalbini:
r=me

PR #2596, opened 2020-12-03
cli: Support prefixes of toolchains in + argument

### LINKED ISSUE #1491: Support abbreviating toolchain names (`+n`)
If rustup allowed abbreviating toolchain names via unambiguous prefixes, then `cargo +n build` would work instead of `cargo +nightly build`, which would save quite a bit of typing.

kinnison:
Where +n is unique, expand it to +nightly (or whatever).
This means if you have nightly, beta, and stable, then
`cargo +n run` will use nightly.  If the prefix you provide
is not unique then we won't expand it and you'll get the
same error you would have before.

This will fix #1491

Currently a draft because it needs tests.  Seeking opinions from @rbtcollins and @joshtriplett who originally requested the feature.

```console
dsilvers@listless:~$ rustup toolchain list
beta-x86_64-unknown-linux-gnu
nightly-2020-11-24-x86_64-unknown-linux-gnu
dsilvers@listless:~$ rustup show active-toolchain
dsilvers@listless:~$ rustup +beta show active-toolchain
beta-x86_64-unknown-linux-gnu (overridden by +toolchain on the command line)
dsilvers@listless:~$ rustup +nightly show active-toolchain
nightly-x86_64-unknown-linux-gnu (overridden by +toolchain on the command line)
dsilvers@listless:~$ rustup +n show active-toolchain
error: override toolchain 'n' is not installed
error: caused by: the +toolchain on the command line specifies an uninstalled toolchain
dsilvers@listless:~$ rustup +b show active-toolchain
beta-x86_64-unknown-linux-gnu (overridden by +toolchain on the command line)
dsilvers@listless:~$ 
```

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -82,7 +82,7 @@ pub fn main() -> Result<utils::ExitCode> {
 
                 if let Some(t) = process().args().find(|x| x.starts_with('+')) {
                     debug!("Fetching rustc version from toolchain `{}`", t);
-                    cfg.set_toolchain_override(&t[1..]);
+                    cfg.set_toolchain_override(&t[1..])?;
                 }
 
                 let toolchain = cfg.find_or_install_override_toolchain_or_default(&cwd)?.0;
@@ -118,7 +118,7 @@ pub fn main() -> Result<utils::ExitCode> {
     let cfg = &mut common::set_globals(verbose, quiet)?;
 
     if let Some(t) = matches.value_of("+toolchain") {
-        cfg.set_toolchain_override(&t[1..]);
+        cfg.set_toolchain_override(&t[1..])?;
     }
 
     if maybe_upgrade_data(cfg, &matches)? {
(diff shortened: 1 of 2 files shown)


joshtriplett:
This looks awesome! Thank you so much.

kinnison:
@joshtriplett Excellent, thank you.  I'll look at ensuring this is clean/tidy but it looks good to be in 1.24

rbtcollins:
Hmm, in not sure this should apply to version numbers as strings.

+1.3 invoking +1.33.0 seems wrong but perhaps it is ok. The key issue would be if it can be fooled into invoking a lower version than expected...

kinnison:
I'd be happy to exclude any toolchain which matches the channel regexps already from prefix matching.  Would that be right?  It'd prevent `nightly` matching `nightly-localbuild` or somesuch, but perhaps that's not a bad thing.

rbtcollins:
that could be good yeah. I'm sorry I still don't have the bandwidth to fully consider things, but it would address my concern

kinnison:
Okay thanks Robert.  I'll sort out a draft of that too when I come back to writing tests for this.

PR #2599, opened 2020-12-07
Add support for aarch64-apple-darwin

berkus:
Also, when rustup is installed, somewhere the default host is set to x86_64-apple-darwin even on aarch64-apple-darwin machine. I couldn't find that place, but it lead to some crazy behavior re: toolchains - see https://github.com/rust-lang/cargo/issues/8955
It probably needs to be fixed too, but I don't know where exactly.


kinnison:
While this patch isn't bad it's also not necessary since this function actually ends up only being used in places where either it's windows-only, or else via the from_host_or_build which will use the build triplet on arm64 macs.

The presence of x86_64-apple-darwin is likely down to you previously having had rustup installed and perhaps a dodgy settings.toml left over?

berkus:
@kinnison when it didnt work the first time, i definitely removed entire rustup installation including the config files.

I probably should re-test doing complete wipe and install, but I'd wait for another absolutely clean mac without rust and try it there, will open a new issue if it reproduces. Closing this one.

kinnison:
Okay thanks @berkus

Issue #2572, opened 2020-11-25
Indicate default option with all steps


89z:
Using this file:

https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-gnu/rustup-init.exe

I get this question, which presents with two choices, and marks the default
choice:

~~~
If you will be targeting the GNU ABI or otherwise know what you are
doing then it is fine to continue installation without the build
tools, but otherwise, install the C++ build tools before proceeding.

Continue? (Y/n)
~~~

Then I get this, which again marks the default:

~~~
   default host triple: x86_64-pc-windows-msvc
     default toolchain: stable (default)
               profile: default
  modify PATH variable: yes

1) Proceed with installation (default)
2) Customize installation
3) Cancel installation
~~~

However then I get this, which has a default but is not indicated:

~~~
I'm going to ask you the value of each of these installation options.
You may simply press the Enter key to leave unchanged.

Default host triple?
~~~

Then I get this, which has a default but is not indicated:

~~~
Default toolchain? (stable/beta/nightly/none)
~~~

Then I get this, which has a default but is not indicated:

~~~
Profile (which tools and data to install)? (minimal/default/complete)
~~~

Then I get this, which has a default but is not indicated:

~~~
Modify PATH variable? (y/n)
~~~

kinnison:
Improving the prompting here ought to be fairly easy for someone looking to get into hacking on Rustup.

If you are interested in helping to fix this, please indicate so on the issue such that others don't duplicate your work.

If you require mentoring/assistance, ask on this issue, or want to discuss things before starting, you can find me or others on the `#wg-rustup` channel of the Rust discord server (https://discord.gg/rust-lang)

a8nJ:
Currently I am downloading this:

https://static.rust-lang.org/dist/rust-1.48.0-x86_64-pc-windows-gnu.msi

and extracting with a program I wrote that extracts MSI files. Ideally what I
would like is just a Zip file of the above, but I dont think its available.

I think RustUp has the potential to be a good alternative, but the idea of
RustUp just throwing away 100 MB every time its run doesnt sit well with me.
So since that issue is a WONTFIX, I am unlikely to use RustUp going forward.

If you need to close the issue because I wont be trying to fix it myself, so be
it.

kinnison:
[...]

PR #2605, opened 2020-12-11
#2572 - Add defaults to common::question_*

londospark:
Hopefully this is cleaner than specifying the default as a string in each use of the `common::question_{str|bool}` functions.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -68,7 +68,7 @@ pub fn confirm_advanced() -> Result<Confirm> {
 }
 
 pub fn question_str(question: &str, default: &str) -> Result<String> {
-    writeln!(process().stdout(), "{}", question)?;
+    writeln!(process().stdout(), "{} [{}]", question, default)?;
     let _ = std::io::stdout().flush();
     let input = read_line()?;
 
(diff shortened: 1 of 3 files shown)


londospark:
Fixes #2572

kinnison:
This covers the last of four questions - is there other work you have planned for the other questions hilighted in the issue you're fixing?

kinnison:
Also I wonder if we might get a cli-exact test to cover this perhaps?

londospark:
I'll have a look at how to write those, this covers all 4 of the questions by adding the default to the end.

kinnison:
So it does, I missed the effect of the first hunk.  If we don't have any cli-exact tests doing prompted configuration then it ought to be fairly easy to add.  If you want any help with that, ping me on Discord and if I'm around I'll do my best.

londospark:
Would it make more sense to do those tests in `cli-inst-interactive.rs` as it seems to be about testing the installation process? That, and it has the work done already for taking inputs.

kinnison (changes requested):
The tests look good.  Thank you for that.

This does indeed seem to do exactly what it says on the tin.  It should indeed resolve the issue you set out to fix.

Before I can merge it I would like the commits cleanly rebased.  A commit for the tests and a commit for the functionality would be best.  Your choice which order to put them in.

If you need help doing that, ping me.  Also if you're having difficulty with vscode formatting stuff, check that 'Editor: Format On Save' is set and that 'Editor: Format On Save Mode' is set to 'file'.

kinnison (approved):
:+1: thank you

PR #2607, opened 2020-12-12
rustup-init.sh: Support Big Sur major version

### LINKED ISSUE #2606: Rustup doesn't check the platform OS version correctly.
I previously mis-reported this at https://github.com/rust-lang/rust/issues/79924
**Problem**

**Steps**

1. Try to install rust using Rustup on an Apple Silicon laptop.

`% curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`

Expected this to install rust, however there are two problems, lets focus on the first one in this ticket: (The other issue was I was expecting it to install the tier-2 platform, but it seems I need the nightly for that, which will be fine).
```
Warning: Detected OS X platform older than 10.13
Warning: Not enforcing TLS v1.2, this is potentially less secure
```

Big Sur is definitely not older than 10.13!

**Possible Solution(s)**

Correctly parse the output for the OS version, which is as follows:

`% sw_vers -productVersion`
`11.0.1`

**Notes**

Output of `rustup --version`:
Output of `rustup show`:

kinnison:
In theory this fixes #2606 by detecting major os x version 11 in rustup-init.sh

I would appreciate it if @arafangion would check this, and also @shepmaster

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -493,15 +493,27 @@ check_help_for() {
 
     case "$_arch" in
 
-        # If we're running on OS-X, older than 10.13, then we always
-        # fail to find these options to force fallback
         *darwin*)
         if check_cmd sw_vers; then
-            if [ "$(sw_vers -productVersion | cut -d. -f2)" -lt 13 ]; then
-                # Older than 10.13
-                echo "Warning: Detected OS X platform older than 10.13"
-                return 1
-            fi
+            case $(sw_vers -productVersion) in
+                10.*)
+                    # If we're running on macOS, older than 10.13, then we always
+                    # fail to find these options to force fallback
+                    if [ "$(sw_vers -productVersion | cut -d. -f2)" -lt 13 ]; then
+                        # Older than 10.13
+                        echo "Warning: Detected macOS platform older than 10.13"
+                        return 1
+                    fi
+                    ;;
+                11.*)
+                    # We assume Big Sur will be OK for now
+                    ;;
+                *)
+                    # Unknown product version, warn and continue
+                    echo "Warning: Detected unknown macOS major version: $(sw_vers -productVersion)"
+                    echo "Warning TLS capabilities detection may fail"
+                    ;;
+            esac
         fi
         ;;
 


rustup-init.sh
```diff
@@ -493,15 +493,27 @@ check_help_for() {
...
         *darwin*)
         if check_cmd sw_vers; then
-            if [ "$(sw_vers -productVersion | cut -d. -f2)" -lt 13 ]; then
-                # Older than 10.13
-                echo "Warning: Detected OS X platform older than 10.13"
-                return 1
-            fi
+            case $(sw_vers -productVersion) in
+                10.*)
+                    # If we're running on OS-X, older than 10.13, then we always
```
shepmaster: Being pedantic...

- "Mac OS X" 10.0 
- "OS X" 10.8
- "macOS" 10.12

"OS-X" was never a proper name. Since rustup stops at 10.13, I'd suggest changing to "macOS" consistently in user-facing text as well as comments.

rustup-init.sh
```diff
@@ -493,15 +493,27 @@ check_help_for() {
...
+                    # fail to find these options to force fallback
+                    if [ "$(sw_vers -productVersion | cut -d. -f2)" -lt 13 ]; then
```
shepmaster, line 502: The internal pain I have that this shell command is run multiple times is silly, but I'll state it regardless.

rustup-init.sh
```diff
@@ -493,15 +493,27 @@ check_help_for() {
...
+                        # Older than 10.13
+                        echo "Warning: Detected OS X platform older than 10.13"
+                        return 1
+                    fi
+                    ;;
+                11.*)
```
shepmaster, line 508: It *shouldn't* matter, but I'll point out that Big Sur [is _sometimes_ reported as 10.16](https://eclecticlight.co/2020/07/21/big-sur-is-both-10-16-and-11-0-its-official/). I doubt that will come into play for Rustup.

rustup-init.sh
kinnison: That's fair.  I shall make that tweak.

rustup-init.sh
kinnison: Given the cut is different each time, I figured I'd just rerun it.  If this were a real programming language I'd have split it into parts once.  But sw_vers should be pretty fast to run, and consistent through a rustup-init or the user has worse problems :D

rustup-init.sh
kinnison: Is this perhaps why you don't see the issue?

rustup-init.sh
shepmaster: Oh, it’s definitely not a real problem, especially considering that this should only be run like twice ever.

rustup-init.sh
shepmaster: It’s possible somehow, but running sw_vers *did* show `11...`

arafangion: I wonder if it's possibly related to running BigSur on M1 as opposed to on Intel?  I am running this on M1.
The 10.16 may have also been the version reported in beta releases, and might now have been cleaned up? (Just guessing here, really).

arafangion: This difference appears to be due to changes introduced in 01fd3a19774e97839f6b3c52075a491ea9479abd

The code changes otherwise seem to work fine.

arafangion: Regarding 10.16 vs 11.0, this seems to explain it: https://eclecticlight.co/2020/07/21/big-sur-is-both-10-16-and-11-0-its-official/

(Summary: It depends on whether the calling environment has `SYSTEM_VERSION_COMPAT=1` set, it seems)

PR #2609, opened 2020-12-12
Don't build rustls on platforms unsupported by ring

### LINKED ISSUE #2608: CI fails on powerpc64-unknown-linux-gnu
Since merging of #2517, the master branch CI fails on powerpc64-unknown-linux-gnu. [Link to a failing build](https://github.com/rust-lang/rustup/runs/1542098778). Relevant excerpt from the log:

```
error: failed to run custom build command for `ring v0.16.17`

Caused by:
  process didn't exit successfully: `/checkout/target/release/build/ring-69b64e93d224eaca/build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.17/build.rs:375:10
```

The failing line is [here](https://github.com/briansmith/ring/blob/f247250ee0b96770ac34e2c1fbf23e3f3efb0d84/build.rs#L375). It turns out that ring doesn't support the powerpc platform: https://github.com/briansmith/ring/issues/389

I think the best resolution would be to disable the builtin rustls support on the powerpc platform.

cc @kinnison

est31:
Should fix #2608 but haven't tested it. Let's see what CI has to say about it :)

### DIFF
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -7,10 +7,21 @@ export RUST_BACKTRACE=1
 rustc -vV
 cargo -vV
 
-FEATURES=()
+
+FEATURES=('--no-default-features' '--features' 'curl-backend,reqwest-backend,reqwest-default-tls')
 case "$(uname -s)" in
   *NT* ) ;; # Windows NT
-  * ) FEATURES=('--features' 'vendored-openssl') ;;
+  * ) FEATURES+=('--features' 'vendored-openssl') ;;
+esac
+
+case "$TARGET" in
+  # these platforms aren't supported by ring:
+  powerpc* ) ;;
+  mips* ) ;;
+  riscv* ) ;;
+  s390x* ) ;;
+  # default case, build with rustls enabled
+  * ) FEATURES+=('--features' 'reqwest-rustls-tls') ;;
 esac
 
 # rustc only supports armv7: https://doc.rust-lang.org/nightly/rustc/platform-support.html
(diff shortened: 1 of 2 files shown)


kinnison:
@est31 That change looks reasonable.  Have you tried enabling one or more of those platforms to ensure that it correctly excludes the rustls support?  If you're not sure how to, you'll need to edit the template for the CI workflow and then run the janky generation script which will regenerate the .github/workflows files - I don't have the right to push to your branch I think, so I can't do it for you.

est31:
@kinnison I've changed it as requested. Once we get the CI logs i'll change it back. Yeah I usually disable the "allow edits by maintainers" thing. It's a bit weird that it's enabled by default IMO. Nothing about you :).

est31:
Wow nice actually found a mistake in the script... had to be `case "$TARGET" in` instead of `case "$($TARGET)" in`... classic copy & paste mistake...

est31:
Works: https://github.com/rust-lang/rustup/actions/runs/418946505

Screenshot_20201213_161346

I'll remove the last commit that changed the CI configs. r? @kinnison

kinnison (approved):
This makes sense.  Good catch on that match arm too.

kinnison:
Assuming this goes green I'll merge it in.

PR #2610, opened 2020-12-15
use indicatif progressbar for downloads

milahu:
start solving #1818 #1835 etc

@kinnison said

> It'll be quite a bit of work to rearrange **all the output** to go properly via that, but it would potentially allow some cleanup.

> I think that switching to indicatif or another similar library would be an interesting approach. It'd be a significant departure for the current UI if we also funnelled **all messaging** through such a beast, so we'd need to think carefully.

why refactor all the prints? i just want a download progressbar

code is not tested. github CI, what do you say?

edit: CI says no. also my cargo still says no:

```
$ cargo update
error: failed to parse manifest at `Cargo.toml`

Caused by:
  feature `profile-overrides` is required

this Cargo does not support nightly features, but if you
switch to nightly channel you can add
`cargo-features = ["profile-overrides"]` to enable this feature
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ error-chain = "0.12"
 flate2 = "1"
 git-testament = "0.1.4"
 home = {git = "https://github.com/rbtcollins/home", rev = "a243ee2fbee6022c57d56f5aa79aefe194eabe53"}
+indicatif = "*"
 lazy_static = "1"
 libc = "0.2"
 num_cpus = "1.13"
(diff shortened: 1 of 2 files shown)


kinnison:
Are you intending to persue this?  if so it'd be sensible to get to the point you can run `cargo test --all` properly locally before pushing to the CI.  Also since progress bars aren't part of our CI since they only appear on terninals, you will have to extensively test locally to be certain nothing has broken.

You can install a locally built rustup with `cargo run -- --no-modify-path -y` once you're happy with the build.

rami3l:
I'm closing this stale PR in favor of #4426, which hopefully will land in the upcoming release of rustup.

Many thanks for the previous work on this topic nonetheless, @milahu!

PR #2615, opened 2020-12-19
feature: include self-update status in check

### LINKED ISSUE #2577: Include rustup self-update status in `rustup check`
When checking whether toolchain updates are available it would be useful to see that there is also a rustup update available.

Include rustup's own version status in the output of `rustup check`

**Notes**

0xPoe:
close https://github.com/rust-lang/rustup/issues/2577

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -8,13 +8,13 @@ use std::str::FromStr;
 
 use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, Shell, SubCommand};
 
-use super::common;
 use super::errors::*;
 use super::help::*;
 use super::self_update;
 use super::term2;
 use super::term2::Terminal;
 use super::topical_doc;
+use super::{common, self_update::get_available_rustup_version};
 use crate::dist::dist::{
     PartialTargetTriple, PartialToolchainDesc, Profile, TargetTriple, ToolchainDesc,
 };
@@ -316,7 +316,7 @@ pub fn cli() -> App<'static, 'static> {
                         .takes_value(false),
                 ),
         )
-        .subcommand(SubCommand::with_name("check").about("Check for updates to Rust toolchains"))
+        .subcommand(SubCommand::with_name("check").about("Check for updates to Rust toolchains and rustup"))
         .subcommand(
             SubCommand::with_name("default")
                 .about("Set the default toolchain")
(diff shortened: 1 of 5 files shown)


hi-rustin:
@Nemo157 @kinnison  I think maybe we should support "rustup self check" instead of just including the status in the "rustup check" result.

0xPoe:
@kinnison Could you please take look?  Thanks!

kinnison:
This looks pretty good.  I'm not in a position to test this out today, but will endeavour to do so in the next few days.

kinnison (changes requested):
In general I like this, though I think it'd be slightly better if `rustup check` contained a line for rustup in the case of there being no update.  e.g.

`rustup - Up to date : 1.23.1`

Also, could you clean up your commit message when you're done?  Currently it feels like you made some number of commits, then squashed them together without cleaning it up.

This is very close to being mergeable.

0xPoe:
> In general I like this, though I think it'd be slightly better if `rustup check` contained a line for rustup in the case of there being no update. e.g.

Added.

0xPoe:
Thanks for your review! ❤️  @kinnison

kinnison (changes requested):
Please would you add a test for the new message?

0xPoe:
> Please would you add a test for the new message?

Added.

kinnison (changes requested):
Code content now looks good.  Could you please fix up your commit message now.  It is a history of everything you did and as such is not useful.

I'd suggest something along the lines of:

```
feature: include self-update status in check
```

Once the commit message is clean, I can merge this.

0xPoe:
> Code content now looks good. Could you please fix up your commit message now. It is a history of everything you did and as such is not useful.
> 
> I'd suggest something along the lines of:
> 
> ```
> feature: include self-update status in check
> ```
> 
> Once the commit message is clean, I can merge this.

Done, thanks for your review!

0xPoe:
@kinnison ping~ 😄

kinnison:
Hi, there's still a large body in the commit message which reads like a history of the commit which isn't ideal.  I've pushed a rebase which cleans that up for you, and assuming I didn't break anything and the CI goes green, I'll merge this.  Thank you for your work on this feature.

0xPoe:
> Hi, there's still a large body in the commit message which reads like a history of the commit which isn't ideal. I've pushed a rebase which cleans that up for you, and assuming I didn't break anything and the CI goes green, I'll merge this. Thank you for your work on this feature.

Thanks for your review!

kinnison (approved):

PR #2616, opened 2020-12-20
Use POSIX-compliant `.` instead of `source` in .profile

Xyene:
The invocation to `source` introduced in 850adef is not POSIX-compliant,
and not available in some shells. Use `.` instead, which will work in
all shells. Concretely, this was failing in Docker builds, where
`/bin/sh` is not Bash.

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -98,7 +98,7 @@ pub trait UnixShell {
     }
 
     fn source_string(&self) -> Result<String> {
-        Ok(format!(r#"source "{}/env""#, cargo_home_str()?))
+        Ok(format!(r#". "{}/env""#, cargo_home_str()?))
     }
 }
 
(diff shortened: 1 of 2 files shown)


kinnison (approved):
Thank you for this.

Xyene:
Thanks! Any idea when this will roll to rustup.sh? No big rush, just want to know whether it'll happen ~soon, or if I should work around it locally.

kinnison:
Continue to work around locally for now I'm afraid.  I will be trying to do a release early in Jan but can't guarantee when.

DamianFekete:
@kinnison , I think there is a small issue with this change, but I'm also not sure if this is acceptable for this project or not.
* Uninstall/update won't work as expected with previously installed versions (`unix.rs#do_remove_from_path` now searches for `. "$HOME/.cargo/env"` although the resource file was modified with `source "$HOME/.cargo/env"`)
* Rust will source the `env` file twice (not a practical problem, but not so nice):

``` shell
# ~/.bashrc
source "$HOME/.cargo/env"
. "$HOME/.cargo/env"
```

kinnison:
Hmm, that's a good point, we should probably detect both kinds in order that people don't end up confused on uninstallation. I've filed #2623 to this effect.

PR #2619, opened 2020-12-25
Fix missing space in error message

dralley:
This PR is 2 commits. The first simply fixes and issue that wasn't totally resolved by 5414c42 (the other branch of the "if" statement wasn't covered). The second goes further and does some cleanup that I thought would be worthwhile (removing duplicated code, making the channel quoted as in ```unavailable for download for channel 'beta'```, making the use of newline more consistent, etc.).

new behavior: left
Screenshot from 2020-12-25 00-07-45

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -431,7 +431,7 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
             if toolchain.starts_with("nightly") {
                 "\nSometimes not all components are available in any given nightly. "
             } else {
-                ""
+                " "
             }
         );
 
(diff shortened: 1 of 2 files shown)


dralley:
Looks like I broke the expectations of the tests.

```
(actual)
    error: component 'rust-std' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel 'nightly'
 (expected)
    error: component 'rust-std' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
```

I can either drop that part of the patch or update the tests -- which would you prefer?

kinnison:
Sorry for the delay, winter break is getting in the way of my rust work 😁 

I think that as you're improving error messages, it makes most sense to adjust the tests to match your improved messages, however it'll be worth checking each time you update the tests to ensure that the resulting message actually does look better once in context.

Most rustup tests don't look for the exact message so you likely won't have to adjust many.

Thank you for this.

dralley:
@kinnison No worries, enjoy your vacation :)

I updated the test expectations

kinnison (approved):
This looks good, thank you :+1:

Issue #2575, opened 2020-11-26
Additional tests for `rust-toolchain` desirable


kinnison:
We could do with some additional testing of the `rust-toolchain` functionality.  Particularly around healing damaged toolchain installations, ensuring that the right components/targets are installed into it, etc.

This is related to #2574

chansuke:
I'd like to work on this.

kinnison:
Wonderful.  Perhaps a good start would be for you to look at all the tests which already work on rust-toolchain, and then come up with some proposed tests as comments on this issue.  Once we have a good idea of what tests you should write, it'll be worth putting hand-to-keyboard to produce some test code.

chansuke:
@kinnison I appreciate your comment and suggestion. I will start working at this weekend.

chansuke:
@kinnison I read through the test codes... My idea is that creating the function that generates a damaged toolchain in `tests/mock/clitools.rs`. After that, I will start adding some test codes. Is that the right approach...? Thanks.

kinnison:
I'm not sure you need to create broken toolchains in `clitools.rs`.

Let's first consider the 'checking components and targets work' part of the above:

A possible testing scenario might be:

1. a project has a `rust-toolchain` stating that it wants `nightly` with the `rust-src` component installed.
2. the user already has `nightly` but without `rust-src`
3. we should cause `rust-src` to be installed when the user interacts with rustup (e.g. `rustup show active-toolchain`) in that project

You could extend the above to include, for example, the user deliberately uninstalling a needed component or target, and verifying that on interacting with `rustup show active-toolchain` in that directory, the uninstalled bits are reinstalled.

Now for the 'healing' aspect, specifically the linked PR deals with the fact that if the already installed toolchain looked like a dist toolchain but somehow was missing its manifest file then we would have failed to "heal" it.  To understand what's going on there we need to think through the code which leads up to that point:

1. The override file talks about a toolchain which ought to be a dist toolchain (e.g. `nightly`)
2. The toolchain is already installed
3. The installed toolchain is damaged (missing its manifest file)
4. The user interacts with rustup in a way which causes it to need to use that override (e.g. a `rust-toolchain` file is present with `nightly` in it and the user runs `rustup show active-toolchain`)

To achieve 3 there, all we need do in a test is:

1. cause `nightly` to be installed (e.g. `rustup default nightly` - there's lots of examples of this in the test suite)
[...]

PR #2620, opened 2020-12-26
Add test for damaged toolchain

chansuke:
#2575 
Add test for damaged toolchain installations.

### DIFF
--- a/tests/cli-rustup.rs
+++ b/tests/cli-rustup.rs
@@ -11,8 +11,8 @@ use rustup::test::this_host_triple;
 use rustup::utils::raw;
 
 use crate::mock::clitools::{
-    self, expect_err, expect_not_stdout_ok, expect_ok, expect_ok_ex, expect_stderr_ok,
-    expect_stdout_ok, run, set_current_dist_date, Config, Scenario,
+    self, expect_err, expect_not_stderr_ok, expect_not_stdout_ok, expect_ok, expect_ok_ex,
+    expect_stderr_ok, expect_stdout_ok, run, set_current_dist_date, Config, Scenario,
 };
 
 macro_rules! for_host_and_home {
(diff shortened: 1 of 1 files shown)


kinnison:
This is a good start, though we should probably detect some of the text indicating it has just healed the toolchain in the second invocation.

kinnison (changes requested):
This is getting nearer to validating what we want, however in theory the test is just asserting that every time we 'show active-toolchain' we sync the channel.

We should run `rustup show active-toolchain` before removing the channel toml file and assert that it *doesn't* sync the toolchain; then remove the file, run it again, assert that it *does* sync it, and then run it a third time and assert that it doesn't sync it once more.

That way we show "it doesn't heal unless it needs to; and then having done so it is healed so doesn't heal unnecessarily again"

kinnison:
Also your test is failing so you should be sure to check it before pushing again.

`cargo test --test cli-rustup -- heal_damaged_toolchain` will run just your test.

syncing0x:
> #2575 
> Add test for damaged toolchain installations.

chansuke:
@kinnison I appreciate your comment. I will fix the error.

tests/cli-rustup.rs
```diff
@@ -803,6 +803,33 @@ fn list_default_and_override_toolchain() {
     });
 }
 
+#[test]
+fn heal_damaged_toolchain() {
+    setup(&|config| {
+        expect_ok(config, &["rustup", "default", "nightly"]);
+        expect_ok(config, &["rustup", "show", "active-toolchain"]);
```
kinnison: When I suggested adding this check, it was because we need to ensure that when the toolchain doesn't need healing we don't try syncing the channel once more.  As such, something like the following would make sense:

```suggestion
        expect_not_stderr_ok(config, &["rustup", "show", "active-toolchain"], "syncing channel updates");
```

kinnison (changes requested):
You're *so close* to where we need to be.

What I was trying to get at with my suggestion was that we should *ENSURE* that the first `rustup show active-toolchain` does **NOT** cause a re-sync of the channel since it shouldn't need to.  Could you try my suggestion (I've not tested it myself, but in theory you can check that it's doing the right thing by using `expect_stderr_ok` first, which should fail, and then change that to `expect_not_stderr_ok` and it should then pass through.

I'm sorry to keep coming back to you on this, I *could* just make the change myself and merge your work; but I got the feeling you would appreciate working through the process to understand why I'm looking for the changes.

If on the other hand you just want this finished then I'm happy to make the update myself and merge your work.

tests/cli-rustup.rs
chansuke: @kinnison I appreciate the patient support and yes, I would like to finish this.

kinnison (approved):
Fantastic, thank you so much

PR #2622, opened 2020-12-27
Detect and warn about x32

nabijaczleweli:
Error message wrapped to eighty columns

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -159,6 +159,17 @@ get_bitness() {
     fi
 }
 
+is_host_amd64_elf() {
+    need_cmd head
+    need_cmd tail
+    # ELF e_machine detection without dependencies beyond coreutils.
+    # Two-byte field at offset 0x12 indicates the CPU,
+    # but we're interested in it being 0x3E to indicate amd64, or not that.
+    local _current_exe_machine
+    _current_exe_machine=$(head -c 19 /proc/self/exe | tail -c 1)
+    [ "$_current_exe_machine" = "$(printf '\076')" ]
+}
+
 get_endianness() {
     local cputype=$1
     local suffix_eb=$2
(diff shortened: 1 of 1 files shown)


kinnison:
This change looks interesting, thank you for putting the effort in.

Can you suggest an easy way for me to install an x32 system in a VM so that I can test this?  Is there a distro you suggest for this?

nabijaczleweli:
As far as a VM goes, I suggest following [this post](https://nabijaczleweli.xyz/content/blogn_t/001-x32-in-2020.html) for doing a x32 installation of Debian, which AFAIK is the only distribution that supports the ABI for the whole userland.

However, a chroot with a procfs mount should work just as well (but you'll need your host kernel to understand x32; if you're on Debian you'll need to boot with `syscall.x32=y`, if on Gentoo, it should Just Work with default config).

Something like
```
debootstrap --arch=x32 --variant=minbase --include=debian-ports-archive-keyring --no-check-gpg sid sid-x32/ https://deb.debian.org/debian-ports
mount --bind /proc/ sid-x32/proc/
chroot sid-x32/
```
should give you a bare-bones x32 userland.

(Technically all you need is an x32 `head(1)`, but that's more of a main in the arse than is worth.)

kinnison:
I read through those instructions but don't feel comfortable messing around with debootstrap etc.  I'm fairly confident that this will work, having reread the shell so I'm going to merge this providing you can confirm you've tried the updated shell script on all of:

1. x86_64 with 64bit userland (should be entirely unaffected)
2. x86_64 with real 32bit userland (should be fine by your `else` clause)
3. x86_64 with x32 userland (I assume you've tested this already since otherwise you wouldn't have submitted the patch)

Are you able to confirm all of those are good with your change?  Sorry I can't get local testing set up but I'm in and out of Windows right now and don't have the time/space to do it.

nabijaczleweli:
As far as I can tell, yes, all of those end up with the right cputype.

kinnison:
Fantastic, thank you!

Issue #2626, opened 2021-01-01
improve rustc wrapper startup time?


matthiaskrgr:
`time  ~/.rustup/toolchains/master/bin/rustc  >& /dev/null`
`0,040 total`

`time  ~/.cargo/bin/rustc  >& /dev/null`
`0,272 total`

That's almost 7x slower.
I wonder if there is some something we can do to get these numbers a bit closer?

kinnison:
I can confirm that the difference in performance is around 6x:

```
listless% hyperfine ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc rustc 
Benchmark #1: /home/dsilvers/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc
  Time (mean ± σ):      11.3 ms ±   0.9 ms    [User: 6.9 ms, System: 4.3 ms]
  Range (min … max):    10.1 ms …  15.4 ms    252 runs
 
Benchmark #2: rustc
  Time (mean ± σ):      71.2 ms ±   2.3 ms    [User: 60.0 ms, System: 11.2 ms]
  Range (min … max):    67.3 ms …  76.2 ms    41 runs
 
Summary
  '/home/dsilvers/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc' ran
    6.31 ± 0.54 times faster than 'rustc'
listless%                                                                                

```

kinnison:
Looking locally, ca. 15ms are before main() is entered (presumably due to dynamic link time for all the libraries rustup uses by dint of being linked to curl).

It looks like about 25ms is taken parsing manifest files, sadly this is also done twice, and thus accounts for the vast majority of the extra time.  The toml parser must be quite inefficient.

It might be possible to knock one of those parse steps on the head since in theory we only need to parse once, but it'll involve quite a bit of refactoring.

If someone wants to begin to look at this, then the place to start is the config.rs file, around line 620 which is `find_or_install_override_toolchain_or_default()`.  In the sub-function `components_exist()` you can see it acquire the manifest (since it needs it) and then it lists the components in order to check that everything needed is installed.  The `list_components()` function though ends up loading the manifest itself.  Perhaps the distributable toolchain could cache the manifest having loaded it?

rbtcollins:
To help triage this, how much does this contribute to typical build times? like, - sure we can put a chunk of time into making this snappy, and that would be good, but if a typical build is 1500ms ... ?

[...]

PR #2627, opened 2021-01-02
dist: Support a bincoded manifest file for performance reasons

kinnison:
This goes some of the way to mitigating #2626 but isn't a "fix" per-se.

Not least, we need to be sure of whether this is valid.

### DIFF
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -52,18 +52,22 @@ pub fn if_not_empty<S: PartialEq<str>>(s: S) -> Option<S> {
     }
 }
 
-pub fn write_file(path: &Path, contents: &str) -> io::Result<()> {
+pub fn write_file_bytes(path: &Path, contents: &[u8]) -> io::Result<()> {
     let mut file = fs::OpenOptions::new()
         .write(true)
         .truncate(true)
         .create(true)
         .open(path)?;
 
-    io::Write::write_all(&mut file, contents.as_bytes())?;
+    io::Write::write_all(&mut file, contents)?;
 
     file.sync_data()?;
 
     Ok(())
+}
+
+pub fn write_file(path: &Path, contents: &str) -> io::Result<()> {
+    write_file_bytes(path, contents.as_bytes())
 }
 
 pub fn filter_file<F: FnMut(&str) -> bool>(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,6 +134,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
+name = "bincode"
+version = "1.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f30d3a39baa26f9651f17b375061f3233dde33424a8b72b0dbe93a68a0bc896d"
+dependencies = [
+ "byteorder",
+ "serde",
+]
+
+[[package]]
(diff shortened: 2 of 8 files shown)


kinnison:
I've rewritten this as a serialisation of the parsed manifest as a bincoded file.  This is basically the same performance as toml parsing the trimmed manifest, but doesn't involve trimming which was debateable as to its correctness.

We need to introduce a version indicator for this so that we can detect if we should fall back to reading the toml and rewriting the bincode in case of changing our manifest structures.

rbtcollins:
Lets use flatbuf not bincode. https://www.reddit.com/r/rust/comments/cmq6k9/bincode_for_other_languages/ vs https://google.github.io/flatbuffers/flatbuffers_support.html

src/utils/raw.rs
```diff
@@ -52,20 +52,24 @@ pub fn if_not_empty<S: PartialEq<str>>(s: S) -> Option<S> {
...
 
-    io::Write::write_all(&mut file, contents.as_bytes())?;
+    io::Write::write_all(&mut file, contents)?;
 
     file.sync_data()?;
 
     Ok(())
 }
 
+pub fn write_file(path: &Path, contents: &str) -> io::Result<()> {
```
rbtcollins, line 69: I don't think this pays for itself vs write_file(path, contents.as_bytes())?;

kinnison:
Is there a flatbuffers crate with serde support?

src/utils/raw.rs
```diff
@@ -52,20 +52,24 @@ pub fn if_not_empty<S: PartialEq<str>>(s: S) -> Option<S> {
     }
 }
 
-pub fn write_file(path: &Path, contents: &str) -> io::Result<()> {
+pub fn write_file_bytes(path: &Path, contents: &[u8]) -> io::Result<()> {
```
rbtcollins, line 55: I note this is doing a sync_data - this is an important part of the contract of the function; if we're renaming it perhaps consider exposing that at the same time - e.g. namespacing it or adding _synced or something.

rbtcollins:
There is for flexbuffers - https://github.com/google/flatbuffers/tree/master/rust/flexbuffers - but I'm not sure of the story for flatfbuffers.

kinnison:
Okay so flexbuffers look plausible vs. bincode, though as it's an internal cache implementation detail why are you adamant we shouldn't use bincode?

src/utils/raw.rs
kinnison: I agree with this idea and will sort it out

src/utils/raw.rs
kinnison: Fair enough, I'll sort out a refactor commit alongside this which pushes that up to the call sites.

rbtcollins:
If we need to debug it or introspect it, flatbuffers has more tooling available as it isn't rust-only with relatively few users. ditto flexbuffers; flatbuffers is the schemad version, I don't think the lack of serde support should be an issue though I haven't looked into it closely - an alternative would be protobuf, the tower protobuf glue is pretty nice

kinnison:
I'm concerned about minimising the impact of the effort if we're do this soon.  I was thinking of treating the binary as a cache and if it failed to load falling back to the toml.  The serde capability just means it's much less effort for us in terms of implementation.

Debuggability is a good argument against bincode though.  Flexbuffers look plausible if a  bit more awkward to implement than bincode, yaml, json, etc.

bjorn3:
Cargo also uses bincode for certain caches like fingerprints. Flatbuffers having a schema would make the caches a bit bigger I think and will likely encourage others to inspect this implementation detail.

PR #2628, opened 2021-01-04
Update mdbook

ehuss:
Changelog to 0.4.5: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-045

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Install mdbook
         run: |
           mkdir mdbook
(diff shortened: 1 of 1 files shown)


kinnison (approved):
Thanks, do we need to rerun this on stable too?

ehuss:
I'll need to spend some time to figure out how to build it on stable without a release.  I'm also still planning on having it redeploy on each merge to master in a separate directory.

kinnison:
I'd like to be preparing 1.24.0 fairly soon, so unless you think the risk is very high, we can wait for that to update our deployed docs?

PR #2631, opened 2021-01-06
Add gnux32 to `LIST_ENVS`

nabijaczleweli:
This got rustup to accept set default-host and the installer to not choke on it, though not much further

### DIFF
--- a/src/dist/triple.rs
+++ b/src/dist/triple.rs
@@ -36,6 +36,7 @@ static LIST_OSES: &[&str] = &[
 ];
 static LIST_ENVS: &[&str] = &[
     "gnu",
+    "gnux32",
     "msvc",
     "gnueabi",
     "gnueabihf",


kinnison:
gnux32 is a std-only platform according to the platform-support information here: https://doc.rust-lang.org/nightly/rustc/platform-support.html

As a first step to ensuring rustup can support the platform once it becomes std+host this is useful, but as you say it won't get you much further on the route to actually getting rustup usefully on that host platform yet.

I'll merge this, but as noted, it's only a small step on the route.

Are you part of any team trying to get gnux32 to the point that it'll be std+host supported at tier2 ?

nabijaczleweli:
Am not; should I be?

kinnison:
Well I have no idea if there even is a team attempting that - I guess it'd be a good idea for you to join the rust zulip and talk with the release team about what would be needed to get from std to std+host.

PR #2634, opened 2021-01-09
paths: Clean up use of 'source' rather than '.'

### LINKED ISSUE #2623: Need to detect both `source` and `.` when installing and uninstalling rustup.
In  #2616 we changed from `source` to `.` in order to be more compatible with non-bash shells.  We should probably use `source` for bash and `.` for POSIX sh if we can separate the two, or else just use `.` but detect both kinds so that we don't end up with double invocations and/or failing to remove an invocation on uninstallation as per https://github.com/rust-lang/rustup/pull/2616#issuecomment-750950080

kinnison:
This should fix #2623

@DamianFekete what do you think?

### DIFF
--- a/src/cli/self_update/unix.rs
+++ b/src/cli/self_update/unix.rs
@@ -145,22 +145,38 @@ pub fn self_replace() -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
-fn remove_legacy_paths() -> Result<()> {
-    let export = format!("export PATH=\"{}/bin:$PATH\"\n", shell::cargo_home_str()?).into_bytes();
+fn remove_legacy_source_command(source_cmd: String) -> Result<()> {
+    let cmd_bytes = source_cmd.into_bytes();
     for rc in shell::legacy_paths().filter(|rc| rc.is_file()) {
         let file = utils::read_file("rcfile", &rc)?;
         let file_bytes = file.into_bytes();
         // FIXME: This is whitespace sensitive where it should not be.
         if let Some(idx) = file_bytes
-            .windows(export.len())
-            .position(|w| w == export.as_slice())
+            .windows(cmd_bytes.len())
+            .position(|w| w == cmd_bytes.as_slice())
         {
             // Here we rewrite the file without the offending line.
             let mut new_bytes = file_bytes[..idx].to_vec();
-            new_bytes.extend(&file_bytes[idx + export.len()..]);
+            new_bytes.extend(&file_bytes[idx + cmd_bytes.len()..]);
             let new_file = String::from_utf8(new_bytes).unwrap();
             utils::write_file("rcfile", &rc, &new_file)?;
         }
     }
+    Ok(())
+}
+
+fn remove_legacy_paths() -> Result<()> {
+    // Before the work to support more kinds of shells, which was released in
+    // version 1.23.0 of rustup, we always inserted this line instead, which is
+    // now considered legacy
+    remove_legacy_source_command(format!(
+        "export PATH=\"{}/bin:$PATH\"\n",
+        shell::cargo_home_str()?
+    ))?;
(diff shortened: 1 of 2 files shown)


src/cli/self_update/unix.rs
```diff
@@ -164,3 +164,17 @@ fn remove_legacy_paths() -> Result<()> {
     }
     Ok(())
 }
+
+fn remove_legacy_paths() -> Result<()> {
+    // Before the work to support more kinds of shells, we always inserted this
+    // which is now considered legacy
```
DamianFekete: I personally prefer more precise comments, because coming from outside some months later I won't know what "before the work to support" is or when it was done. If possible I would write "Before/Until version XYZ ...".

src/cli/self_update/unix.rs
```diff
@@ -164,3 +164,17 @@ fn remove_legacy_paths() -> Result<()> {
...
+    remove_legacy_path(format!(
+        "export PATH=\"{}/bin:$PATH\"\n",
+        shell::cargo_home_str()?
+    ))?;
+    // We had a release which used `source` rather than `.` which while widely
+    // supported, isn't actually POSIX, so we clean that up.
```
DamianFekete: I would prepend the comment with the Github issue number (it helps when the code is moved several times and git blame doesn't quickly provide something useful). It may also help to provide the rustup version, so that someone can decide in a year (or N years) that the remove_legacy_path is not needed anymore and delete the code.

src/cli/self_update/unix.rs
```diff
@@ -145,8 +145,8 @@ pub fn self_replace() -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
-fn remove_legacy_paths() -> Result<()> {
-    let export = format!("export PATH=\"{}/bin:$PATH\"\n", shell::cargo_home_str()?).into_bytes();
+fn remove_legacy_path(path: String) -> Result<()> {
+    let export = path.into_bytes();
```
DamianFekete: `path` is not actually a path but a line, so I would rename both the function (`remove_line`) and the parameter (`line`?).
The line is also not always an "export" line.

tests/cli-paths.rs
```diff
@@ -32,6 +32,10 @@ export PATH="$HOME/apple/bin"
         format!(". \"{dir}/{sh}\"\n", dir = dir, sh = sh)
     }
 
+    fn bad_source(dir: impl Display, sh: impl Display) -> String {
```
DamianFekete: `s/bad_source/source_pre_version_XYZ` ? Or `non_posix_source` ?
It's not easily clear why the function is called "bad".

Maybe extracting `&bad_source("$HOME/.cargo", POSIX_SH)` as a constant and adding a comment to make it clear what it means would also help down the line?

DamianFekete (approved):
@kinnison, LGTM.

I've added some comments, but keep in mind that I don't know Rust, the code base or the code style, so feel free to ignore them. I've noticed the issue as I was looking into also supporting the [fish shell](https://fishshell.com/).

Thanks for acting on the issue so promptly!

src/cli/self_update/unix.rs
kinnison: I agree, mentioned 1.23.0

src/cli/self_update/unix.rs
kinnison: Mentioned both 1.23 and #2623

tests/cli-paths.rs
kinnison: Relabelled the function `non_posix_source` which I think is probably good enough.

I also added an expositionary comment as to why this is present.

src/cli/self_update/unix.rs
kinnison: Renamed the function to `remove_legacy_source_command` and the argument to `source_cmd`.  Finally renamed `export` as `cmd_bytes` - All very good points, thank you for taking the time to review this for me.

PR #2637, opened 2021-01-12
download: Use an IP address when testing socks support

kinnison:
In order that we can test socks, we have to attempt an HTTP request
in toto.  Unfortunately if the test environment has broken (or
intermittently failing) DNS then this test can fail incorrectly.

By changing the request to an IP address we can check that the
SOCKS proxy is used without needing to rely on the DNS.


rbtcollins:
Lgtm

PR #2641, opened 2021-01-20
Fix incorrect documentation

jyn514:
Rustup explicitly targets building on stable.


kinnison (approved):
Out of date docs make me sad 😢 So thank you for this fix 😁

PR #2649, opened 2021-02-05
Handle PATHs with non-unicodes values on Windows

### LINKED ISSUE #265: Non-unicode values of HKCU\Environment\PATH are not handled
This registry key is modified on install and uninstall. If it contains a non-unicode value rustup will panic.

kellda:
Fixes #265

### DIFF
--- a/src/cli/self_update/test.rs
+++ b/src/cli/self_update/test.rs
@@ -5,39 +5,34 @@ use std::sync::Mutex;
 use lazy_static::lazy_static;
 #[cfg(not(unix))]
 use winreg::{
-    enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE},
+    enums::{HKEY_CURRENT_USER, KEY_READ, KEY_WRITE},
     RegKey, RegValue,
 };
 
 #[cfg(not(unix))]
-use crate::utils::utils;
-
-#[cfg(not(unix))]
-pub fn get_path() -> Option<String> {
+pub fn get_path() -> Option<RegValue> {
     let root = RegKey::predef(HKEY_CURRENT_USER);
     let environment = root
         .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
         .unwrap();
-    // XXX: copied from the mock support crate, but I am suspicous of this
-    // code: This uses ok to allow signalling None for 'delete', but this
-    // can fail e.g. with !(winerror::ERROR_BAD_FILE_TYPE) or other
-    // failures; which will lead to attempting to delete the users path
-    // rather than aborting the test suite.
-    environment.get_value("PATH").ok()
+    match environment.get_raw_value("PATH") {
+        Ok(val) => Some(val),
+        Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => None,
+        Err(e) => panic!(
+            "Error getting PATH: {}\nBetter abort to avoid trahsing it.",
+            e
+        ),
+    }
 }
 
 #[cfg(not(unix))]
-fn restore_path(p: Option<String>) {
+fn restore_path(p: Option<RegValue>) {
     let root = RegKey::predef(HKEY_CURRENT_USER);
     let environment = root
         .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
         .unwrap();
     if let Some(p) = p.as_ref() {
-        let reg_value = RegValue {
-            bytes: utils::string_to_winreg_bytes(&p),
-            vtype: RegType::REG_EXPAND_SZ,
-        };
-        environment.set_raw_value("PATH", &reg_value).unwrap();
+        environment.set_raw_value("PATH", &p).unwrap();
     } else {
         let _ = environment.delete_value("PATH");
     }
@@ -60,9 +55,9 @@ pub fn with_saved_path(f: &dyn Fn()) {
 }
 
 #[cfg(unix)]
-pub fn get_path() -> Option<String> {
+pub fn get_path() -> Option<()> {
     None
 }
 
 #[cfg(unix)]
-fn restore_path(_: Option<String>) {}
+fn restore_path(_: Option<()>) {}
--- a/src/cli/self_update/test.rs
+++ b/src/cli/self_update/test.rs
@@ -10,18 +10,15 @@ use winreg::{
 };
 
 #[cfg(not(unix))]
-pub fn get_path() -> Option<RegValue> {
+pub fn get_path() -> std::io::Result<Option<RegValue>> {
     let root = RegKey::predef(HKEY_CURRENT_USER);
     let environment = root
         .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
         .unwrap();
     match environment.get_raw_value("PATH") {
-        Ok(val) => Some(val),
-        Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => None,
-        Err(e) => panic!(
-            "Error getting PATH: {}\nBetter abort to avoid trashing it.",
-            e
-        ),
+        Ok(val) => Ok(Some(val)),
+        Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => Ok(None),
+        Err(e) => Err(e),
     }
 }
 
@@ -48,15 +45,15 @@ pub fn with_saved_path(f: &dyn Fn()) {
 
     // On windows these tests mess with the user's PATH. Save
     // and restore them here to keep from trashing things.
-    let saved_path = get_path();
+    let saved_path = get_path().expect("Error getting PATH: Better abort to avoid trashing it.");
     let _g = scopeguard::guard(saved_path, restore_path);
 
     f();
 }
 
 #[cfg(unix)]
-pub fn get_path() -> Option<()> {
-    None
+pub fn get_path() -> std::io::Result<Option<()>> {
+    Ok(None)
 }
 
 #[cfg(unix)]
--- a/src/cli/self_update/test.rs
+++ b/src/cli/self_update/test.rs
@@ -19,7 +19,7 @@ pub fn get_path() -> Option<RegValue> {
         Ok(val) => Some(val),
         Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => None,
         Err(e) => panic!(
-            "Error getting PATH: {}\nBetter abort to avoid trahsing it.",
+            "Error getting PATH: {}\nBetter abort to avoid trashing it.",
             e
         ),
     }
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -144,7 +144,7 @@ pub fn do_add_to_path() -> Result<()> {
     _apply_new_path(new_path)
 }
 
-fn _apply_new_path(new_path: Option<String>) -> Result<()> {
+fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
     use std::ptr;
     use winapi::shared::minwindef::*;
     use winapi::um::winuser::{
@@ -169,7 +169,7 @@ fn _apply_new_path(new_path: Option<String>) -> Result<()> {
             .chain_err(|| ErrorKind::PermissionDenied)?;
     } else {
         let reg_value = RegValue {
-            bytes: utils::string_to_winreg_bytes(&new_path),
+            bytes: utils::string_to_winreg_bytes(new_path),
             vtype: RegType::REG_EXPAND_SZ,
         };
         environment
@@ -194,9 +194,9 @@ fn _apply_new_path(new_path: Option<String>) -> Result<()> {
 }
 
 // Get the windows PATH variable out of the registry as a String. If
-// this returns None then the PATH variable is not unicode and we
+// this returns None then the PATH variable is not a string and we
 // should not mess with it.
-fn get_windows_path_var() -> Result<Option<String>> {
+fn get_windows_path_var() -> Result<Option<Vec<u16>>> {
     use std::io;
     use winreg::enums::{HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
     use winreg::RegKey;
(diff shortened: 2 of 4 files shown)


kinnison:
Fundamentally I think this looks okay, but I am very much not experienced in Windows development, particularly around this.  Could, we, from a cleanliness perspective, use OsString in the APIs rather than Vec<u16> or would that be even more complex?

@rbtcollins Could you please spare some time to review this, as I feel quite lost when it comes to faffing with Windows PATHs etc.

src/cli/self_update/test.rs
```diff
@@ -5,39 +5,34 @@ use std::sync::Mutex;
...
-    // code: This uses ok to allow signalling None for 'delete', but this
-    // can fail e.g. with !(winerror::ERROR_BAD_FILE_TYPE) or other
-    // failures; which will lead to attempting to delete the users path
-    // rather than aborting the test suite.
-    environment.get_value("PATH").ok()
+    match environment.get_raw_value("PATH") {
+        Ok(val) => Some(val),
+        Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => None,
+        Err(e) => panic!(
+            "Error getting PATH: {}\nBetter abort to avoid trahsing it.",
```
rbtcollins: typo 'trashing'

src/cli/self_update/test.rs
```diff
@@ -5,39 +5,34 @@ use std::sync::Mutex;
...
+    enums::{HKEY_CURRENT_USER, KEY_READ, KEY_WRITE},
     RegKey, RegValue,
 };
 
 #[cfg(not(unix))]
-use crate::utils::utils;
-
-#[cfg(not(unix))]
-pub fn get_path() -> Option<String> {
+pub fn get_path() -> Option<RegValue> {
```
rbtcollins: This would be a larger change, but how do you feel about changing the signature to Result<Option<RegValue>> rather than panicing ? We have other test stuff to unwind, and generally we've been heading towards Result...

src/cli/self_update/windows.rs
```diff
@@ -144,7 +144,7 @@ pub fn do_add_to_path() -> Result<()> {
     _apply_new_path(new_path)
 }
 
-fn _apply_new_path(new_path: Option<String>) -> Result<()> {
+fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
```
rbtcollins, line 147: This should be an [OsString](https://doc.rust-lang.org/std/ffi/struct.OsString.html) as thats what we can accurately get from the Path type, and we'll need to fix up the home crate, and thats the type I'll ask for there as well to support this there.

src/cli/self_update/windows.rs
```diff
@@ -194,9 +194,9 @@ fn _apply_new_path(new_path: Option<String>) -> Result<()> {
 }
 
 // Get the windows PATH variable out of the registry as a String. If
-// this returns None then the PATH variable is not unicode and we
+// this returns None then the PATH variable is not a string and we
 // should not mess with it.
-fn get_windows_path_var() -> Result<Option<String>> {
+fn get_windows_path_var() -> Result<Option<Vec<u16>>> {
```
rbtcollins, line 199: OsString again: OsString on windows can represent any u16 string I believe, though if I'm wrong, I'd like to know. See this trait to help with this particular code path: https://doc.rust-lang.org/stable/std/os/windows/ffi/trait.OsStringExt.html

src/cli/self_update/windows.rs
```diff
@@ -212,63 +212,72 @@ fn get_windows_path_var() -> Result<Option<String>> {
...
-        Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(Some(String::new())),
+        Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(Some(Vec::new())),
         Err(e) => Err(e).chain_err(|| ErrorKind::WindowsUninstallMadness),
     }
 }
 
 // Returns None if the existing old_path does not need changing, otherwise
 // prepends the path_str to old_path, handling empty old_path appropriately.
-fn _add_to_path(old_path: &str, path_str: String) -> Option<String> {
+fn _add_to_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>> {
```
rbtcollins, line 229: OsString etc...

src/cli/self_update/windows.rs
```diff
@@ -212,63 +212,72 @@ fn get_windows_path_var() -> Result<Option<String>> {
...
+    use std::ffi::OsString;
+    use std::os::windows::ffi::OsStrExt;
+
     let windows_path = get_windows_path_var()?;
-    let path_str = utils::cargo_home()?
-        .join("bin")
-        .to_string_lossy()
-        .into_owned();
-    Ok(windows_path.and_then(|old_path| f(&old_path, path_str)))
+    let mut path_str = utils::cargo_home()?;
```
rbtcollins, line 277: ok nice, but lets fix the types

src/cli/self_update/windows.rs
```diff
@@ -402,26 +411,33 @@ pub fn delete_rustup_and_cargo_home() -> Result<()> {
...
+
     use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
     use winreg::{RegKey, RegValue};
 
     use crate::currentprocess;
     use crate::test::with_saved_path;
     use crate::utils::utils;
 
+    fn wide(str: &str) -> Vec<u16> {
+        OsString::from(str).encode_wide().collect()
```
rbtcollins, line 454: I think it can be simpler no?

`OsString::from_wide(&str[..]);`

I wouldn't even have the function

src/cli/self_update/windows.rs
```diff
@@ -437,23 +453,19 @@ mod tests {
                     .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
                     .unwrap();
                 let reg_value = RegValue {
-                    bytes: vec![
-                        0x00, 0xD8, // leading surrogate
-                        0x01, 0x01, // bogus trailing surrogate
-                        0x00, 0x00,
-                    ], // null
-                    vtype: RegType::REG_EXPAND_SZ,
+                    bytes: vec![0x00, 0xD8, 0x01, 0x01, 0x00, 0x00],
```
rbtcollins: please don't discard the explanatory comments about this byte sequence

src/cli/self_update/windows.rs
```diff
@@ -402,26 +411,33 @@ pub fn delete_rustup_and_cargo_home() -> Result<()> {
...
-                r"c:\users\example\.cargo\bin".into()
+                wide(r"c:\users\example\.cargo\bin;foo"),
+                wide(r"c:\users\example\.cargo\bin")
             )
         );
     }
 
     #[test]
-    fn windows_doesnt_mess_with_a_non_unicode_path() {
+    fn windows_doesnt_mess_with_a_non_string_path() {
```
rbtcollins: this is still a non-unicode path, and we still want it not-messed with, so I don't think the title needs altering, does it? Perhaps you want to change it to "non_unicode_paths_are_handled_safely" or something

src/cli/self_update/windows.rs
```diff
@@ -437,23 +453,19 @@ mod tests {
...
-                assert_eq!(None, super::_with_path_cargo_home_bin(panic).unwrap());
+                assert_eq!(
+                    None,
+                    super::_with_path_cargo_home_bin(|_, _| panic!("called")).unwrap()
+                );
             })
         });
         assert_eq!(
-            r"warning: the registry key HKEY_CURRENT_USER\Environment\PATH does not contain valid Unicode. Not modifying the PATH variable
+            r"warning: the registry key HKEY_CURRENT_USER\Environment\PATH is not a string. Not modifying the PATH variable
```
rbtcollins: I don't understand how we'll still see this error if your change to handle non-unicode values is complete.

Perhaps you're adding a new corner case, or anticipating some other sort of error?

If that non-unicode bytestring is able to trigger the error, then we don't actually handle non-unicode strings yet.
If it isn't able to trigger the error, but something else is generating the error, the test isn't clear.

src/cli/self_update/windows.rs
```diff
@@ -556,10 +568,10 @@ mod tests {
     #[test]
     fn windows_uninstall_removes_semicolon_from_path_suffix() {
         assert_eq!(
-            "foo",
+            wide("foo"),
```
rbtcollins, line 620: does .into() work ? OsString::from(ruststring) should work, and ruststring.into() OsString should reflexively too.

src/utils/utils.rs
```diff
@@ -538,10 +539,8 @@ pub fn format_path_for_display(path: &str) -> String {
 
 /// Encodes a utf-8 string as a null-terminated UCS-2 string in bytes
 #[cfg(windows)]
-pub fn string_to_winreg_bytes(s: &str) -> Vec<u8> {
-    use std::ffi::OsStr;
-    use std::os::windows::ffi::OsStrExt;
-    let v: Vec<u16> = OsStr::new(s).encode_wide().chain(Some(0)).collect();
+pub fn string_to_winreg_bytes(mut v: Vec<u16>) -> Vec<u8> {
```
rbtcollins: uhm, nope, lets not do this. I think we can just delete this helper if we switch to OsString.

rbtcollins (changes requested):
Thank you very much for closing this tech debt of ours. I think we have more than one copy of that PATH set-restore code that will need updating to be fully safe, but doing the core like you're doing is a very good start and sufficient to be merged once we work through my feedback here.

kellda:
@kinnison @rbtcollins thanks, I'll take time tomorrow to make the reqpired changes.

**About `Vec<u16>` and `OsString`:** I understand `OsString` would be cleaner, and I initialy tried to use it, but it has currently really few methods available (i.e. just `push` and `clear` really modify it). So most of the time using `OsString` would mean `encode_wide`, edit, `from_wide`, so I don't really see the point here.

src/cli/self_update/test.rs
kellda: Where and how should we handle this result then ?
- In `with_saved_path` or in each test ?
- By ignoring the test or by panicking ?

src/cli/self_update/windows.rs
```diff
@@ -212,63 +212,72 @@ fn get_windows_path_var() -> Result<Option<String>> {
...
     }
 }
 
 // Returns None if the existing old_path does not need changing
-fn _remove_from_path(old_path: &str, path_str: String) -> Option<String> {
-    let idx = old_path.find(&path_str)?;
+fn _remove_from_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>> {
+    let idx = old_path
+        .windows(path_str.len())
+        .position(|path| path == path_str)?;
```
kellda, line 249: @rbtcollins

This function can't be implemented with only `OsString`s. This is why I used `Vec<u16>` everywhere.

Should I rewrite it to take and return `OsString`s and convert to `Vec<u16>` and back ? Or can I just leave `Vec<u16>` everywhere because that's what can be edited ?

rbtcollins: So _remove_from_path is the only complicated case right? Everywhere else is simple comparison or push or clear cases. I think constraining the semantics to signal that OsString is the intent, and containing the complexity to this one place is quite tolerable - it won't be a performance problem after all. 

OTOH the Vec<u16> is already contained in this file, so it won't spread. Ultimately I'm going to be ok either way, and I'm mainly happy to have this contribution.

But my preference, if you're up for it, is to use OsString most places.

kellda: ~Yes, it is the only one, but it is also half of them (the other one being `_add_to_path`).~ In fact, it is also needed in `_add_to_path`, i.e. it is needed everywhere we want to do more than just passing it further.

Note that it will also require converting between `OsString` and `Vec` in `string_from_winreg_value` and `string_to_winreg_bytes`. Since this file is windows-specific (and to avoid four type conversions) I'd rather keep `Vec`s.

~That said now I've explained why I did things the way I did I'll be fine with whatever your decision is.~

kellda:
Done:
- Fixed the typo
- Enhanced tests for non Unicode strings
- There is still a test to check we don't mess with non string values (e.g. binary, dword, qword)

Questions:
- Should I return a result from `get_path` and [where to handle it](https://github.com/rust-lang/rustup/pull/2649#discussion_r571830383)?
- ~Do you still think I should encode the path in a `OsString` to pass it around ?~
- Is there anything else I missed ?

src/cli/self_update/windows.rs
kellda: In fact, we also need `Vec`s in `_add_to_path` to check whether cargo_home is already in the path, which makes me quite strongly against converting `OsString`s back and forth for no real benefit. (Though I could add `type OsString = Vec<u16>` and then write `OsString` instead of `Vec<u16>` is you really want)

rbtcollins:
Vec<u16> is fine everywhere then. I had thought add_to_path could  avoid it, but if it can't there really is a clear win with Vec... Is there a ticket on OsString's limits?

kellda:
> I had thought add_to_path could avoid it

I also through so, but when I tried to see what it would look like to use `OsString` I realized it checks whether cargo_home is already in the path and that can't be done with `OsString`.

I did not find an open issue in the rust repo, but I think this is a known limitation.

rbtcollins:
I think its worth opening one; this is a decent example of it being desired but impossible.

kellda:
I just found that these issues are in the rfcs repo, not the rust one (so there is rust-lang/rfcs#900).

Btw can you re-review my PR ?

src/cli/self_update/windows.rs
```diff
@@ -524,6 +521,41 @@ mod tests {
         });
     }
 
+    #[test]
+    fn windows_doesnt_mess_with_a_non_string_path() {
```
rbtcollins, line 554: much better - thank you!

src/utils/utils.rs
rbtcollins: so AFAICT this is only used from windows.rs now, lets just include it in that file and remove it from utils.rs - it will be cleaner and less of an attractive nuisance I think.

tests/cli-paths.rs
```diff
@@ -309,21 +309,69 @@ mod windows {
...
                 expect_ok(config, &["rustup", "self", "uninstall", "-y"]);
-                assert!(!get_path().unwrap().contains(&path));
+                assert!(!get_path().unwrap().unwrap().to_string().contains(&path));
+            })
+        });
+    }
+
+    #[test]
+    /// Smoke test for end-to-end code connectivity of the installer path mgmt on windows.
+    fn install_uninstall_affect_path_with_non_unicode() {
```
rbtcollins, line 336: thank you. I would have actually been ok with no functional test given the excellent local coverage we have now for this part of the system, but :shrug:

rbtcollins (changes requested):
One last change: move the utils function you change into windows.rs (and probably change the name to something that reflects what it is actually doing for your new code) and we're good.

kellda:
Done !

src/cli/self_update/windows.rs
```diff
@@ -285,6 +285,36 @@ pub fn do_remove_from_path() -> Result<()> {
     _apply_new_path(new_path)
 }
 
+/// Encodes a utf-8 string as a null-terminated UCS-2 string in bytes
```
rbtcollins: I think this comment is wrong though yes? the input isn't a utf8 string ;)

kellda: Hope it is good now, not really sure of how to describe the operation

rbtcollins:
Thank you very much for adding this improvement!

Issue #183, opened 2016-03-29
rustup-setup.sh uses incompatible mktemp invocation


brson:
```
$ curl https://sh.rustup.rs -sSf | sh
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
rustup: command failed: mktemp -d
info: downloading installer
mkdir: : No such file or directory
rustup: command failed: mkdir -p
```
on OSX 10.10 after uninstalling multirust with

```
$ curl -sf https://raw.githubusercontent.com/brson/multirust/master/blastoff.sh | sh -s -- --uninstall.
```

[Source](https://internals.rust-lang.org/t/beta-testing-rustup-rs/3316/14?u=brson).

PR #2650, opened 2021-02-05
Fix template name for mkdir

### LINKED ISSUE #2553: mktemp: too few X's in template ‘rustup’
**Problem**

Installation scripts don't work on ubuntu 20

```bash
$ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
mktemp: too few X's in template ‘rustup’
rustup: command failed: mktemp -d -t rustup
info: downloading installer
mkdir: cannot create directory ‘’: No such file or directory
rustup: command failed: mkdir -p

$lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04 LTS
Release:	20.04
Codename:	focal
```

**Steps**

```bash
$mktemp -d -t rustup
mktemp: too few X's in template ‘rustup’

$type -a mktemp
mktemp is /usr/bin/mktemp
mktemp is /bin/mktemp
```

**Possible Solution(s)**

Create the temporary directory based on template `rustup-XXX`

kellda:
Fix #2553 

Is it really all what is needed ?

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -72,7 +72,7 @@ main() {
     local _url="${RUSTUP_UPDATE_ROOT}/dist/${_arch}/rustup-init${_ext}"
 
     local _dir
-    _dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup)"
+    _dir="$(ensure mktemp -d)"
     local _file="${_dir}/rustup-init${_ext}"
 
     local _ansi_escapes_are_valid=false


kinnison:
I agree that this looks like the right fix, but I wonder if there's any way we can test this somehow.  If you can think of a way to add a test to our CI which validates this then that'd be cool, if you don't think you can, given the intermittent behaviour the bug reporters have had, then we can merge this since I can't see at all how it could be a bad thing regardless.

kellda:
I really don't see how to test this in CI, since this is only executed when `mktemp -d` fails. (Btw I don't see why/how `mktemp -d` can fails, but it seems it happened once)

kellda:
@kinnison Do you have any idea on how to test this in CI ? Is there anything else I have to do before this can be merged ?

rustup-init.sh
```diff
@@ -72,7 +72,7 @@ main() {
     local _url="${RUSTUP_UPDATE_ROOT}/dist/${_arch}/rustup-init${_ext}"
 
     local _dir
-    _dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup)"
+    _dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup-XXX)"
```
rbtcollins: lets just remove the || path here - even OpenBSD mktemp supports -d as-is. https://man.openbsd.org/mktemp.1 : there's no possible case I can see where we're making things better with this.

#183 is the bug that this was meant to fix initially. mktemp on OS X - https://github.com/rust-lang/rustup/commit/7719d9143bfeb9dd9e719a07a0c3e2b3d3ad8732

Now, here's the man page for OS X mktemp: https://ss64.com/osx/mktemp.html

again, -d is supported.

So, I don't know why it was failing for rustup-setup as reported in #183, but the fix was not correctly diagnosed: the original command was valid, and the fix was invalid: note the language from the man page "The template may be any file name with some number of `Xs' appended to it, for example /tmp/temp.XXXX. ".

So - lets:
* go back to a single mktemp -d
* remove 2>/dev/null so that we can actually find out what goes wrong in the long tail here and fix the real bug in future

kinnison: This makes a lot of sense.  I wasn't in the mood to go digging in other OS manpages, so thanks for that Robert.

rbtcollins: You're welcome :)

kellda: Done

rbtcollins: We want
```
local _dir="$(ensure mktemp -d)"
```

Your commit looks like 
```
_dir="$(mktemp -d 2>/dev/null || ensure mktemp -d -t rustup-XXX)"
```

to me - this might be github glitching though, since its been acting odd overnight...

kellda: Right, I forgot to stage my changes before committing... should be good now

rbtcollins:
Thank you.

PR #2653, opened 2021-02-07
Allow `.toml` extension for `rust-toolchain` files

### LINKED ISSUE #2652: Optional `.toml` extension for `rust-toolchain` file?

The `rust-toolchain` file accepts TOML syntax since https://github.com/rust-lang/rustup/pull/2438, but it still can't have a `toml` file extension (i.e. `rust-toolchain.toml`). Thus, the used markup language is not clear to both users and IDEs.

Allow an optional `.toml` extension for the file so that both `rust-toolchain` and `rust-toolchain.toml` are valid. 

**Notes**

A similar approach was taken for Cargo's `.cargo/config` files, which now can be named `.cargo/config.toml` too: https://github.com/rust-lang/cargo/pull/7295

phil-opp:
This pull request adds support for an optional `.toml` extension for `rust-toolchain` files, i.e. the file can now be also named `rust-toolchain.toml`. Only the [TOML syntax](https://github.com/rust-lang/rustup/pull/2438) is supported in `rust-toolchain.toml` files. The advantage of the `.toml` extension is that both users and IDEs directly recognize that the file is written in TOML.

If both `rust-toolchain` and `rust-toolchain.toml` files are present in a directory, rustup outputs a warning and chooses the `rust-toolchain` file. This way, this change is fully backwards compatible.

Resolves #2652 

cc @kinnison

### DIFF
--- a/doc/src/overrides.md
+++ b/doc/src/overrides.md
@@ -8,14 +8,14 @@ and override which toolchain is used:
    +beta`.
 2. The `RUSTUP_TOOLCHAIN` environment variable.
 3. A [directory override], set with the `rustup override` command.
-4. The [`rust-toolchain`] file.
+4. The [`rust-toolchain.toml`] file.
 5. The [default toolchain].
 
 The toolchain is chosen in the order listed above, using the first one that is
 specified. There is one exception though: directory overrides and the
-`rust-toolchain` file are also preferred by their proximity to the current
+`rust-toolchain.toml` file are also preferred by their proximity to the current
 directory. That is, these two override methods are discovered by walking up
-the directory tree toward the filesystem root, and a `rust-toolchain` file
+the directory tree toward the filesystem root, and a `rust-toolchain.toml` file
 that is closer to the current directory will be preferred over a directory
 override that is further away.
 
@@ -24,7 +24,7 @@ To verify which toolchain is active use `rustup show`.
 [toolchain]: concepts/toolchains.md
 [toolchain override shorthand]: #toolchain-override-shorthand
 [directory override]: #directory-overrides
-[`rust-toolchain`]: #the-toolchain-file
+[`rust-toolchain.toml`]: #the-toolchain-file
 [default toolchain]: #default-toolchain
 
 ## Toolchain override shorthand
@@ -74,8 +74,11 @@ case for nightly-only software that pins to a revision from the release
 archives.
 
 In these cases the toolchain can be named in the project's directory in a file
-called `rust-toolchain`, the content of which is either the name of a single
-`rustup` toolchain, or a TOML file with the following layout:
+called `rust-toolchain.toml` or `rust-toolchain`. If both files are present in
+a directory, the latter is used for backwards compatibility. The files use the
+[TOML] format and have the following layout:
+
+[TOML]: https://toml.io/
 
 ``` toml
 [toolchain]
@@ -85,14 +88,19 @@ targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]
 profile = "minimal"
 ```
 
-If the TOML format is used, the `[toolchain]` section is mandatory, and at
-least one property must be specified.
+The `[toolchain]` section is mandatory, and at least one property must be
+specified.
 
-The `rust-toolchain` file is suitable to check in to source control. This file
-has to be encoded in US-ASCII (if you are on Windows, check the encoding and
-that it does not starts with a BOM).
+For backwards compatibility, `rust-toolchain` files also support a legacy
+format that only contains a toolchain name without any TOML encoding, e.g.
+just `nightly-2021-01-21`. The file has to be encoded in US-ASCII this case
+(if you are on Windows, check the encoding and that it does not starts with a
+BOM). The legacy format is not available in `rust-toolchain.toml` files.
 
-The toolchains named in this file have a more restricted form than `rustup`
+The `rust-toolchain.toml`/`rust-toolchain` files are suitable to check in to
+source control.
+
+The toolchains named in these files have a more restricted form than `rustup`
 toolchains generally, and may only contain the names of the three release
 channels, 'stable', 'beta', 'nightly', Rust version numbers, like '1.0.0', and
 optionally an archive date, like 'nightly-2017-01-01'. They may not name
--- a/src/config.rs
+++ b/src/config.rs
@@ -567,10 +567,36 @@ impl Cfg {
                 return Ok(Some((name.into(), reason)));
             }
 
-            // Then look for 'rust-toolchain'
-            let toolchain_file = d.join("rust-toolchain");
-            if let Ok(contents) = utils::read_file("toolchain file", &toolchain_file) {
-                let override_file = Cfg::parse_override_file(contents)?;
+            // Then look for 'rust-toolchain' or 'rust-toolchain.toml'
+            let path_rust_toolchain = d.join("rust-toolchain");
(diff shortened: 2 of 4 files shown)


phil-opp:
I'm opening this PR as a draft since there is still a small TODO:

https://github.com/rust-lang/rustup/blob/e729cf5709d11827b64f3ffc484082931c89efcb/src/config.rs#L589-L594

I'm not sure if that's the best way to print the warning. I saw that there is a `warn!` macro in the `cli` module, but it isn't available in `config.rs`.

src/config.rs
```diff
@@ -567,10 +567,37 @@ impl Cfg {
...
+                }
+                (Ok(contents), Ok(_)) => {
+                    // both `rust-toolchain` and `rust-toolchain.toml` exist
+
+                    // TODO: is there a better way to print a warning?
+                    eprintln!(
+                        "warning: both `{0}` and `{1}` exist. Using `{0}`",
+                        path_rust_toolchain.display(),
+                        path_rust_toolchain_toml.display(),
+                    );
```
kinnison: This message should be a notification sent via the notification handler which is in the `notify` local variable set at the top of this function.

src/config.rs
```diff
@@ -590,12 +617,12 @@ impl Cfg {
         Ok(None)
     }
 
-    fn parse_override_file<S: AsRef<str>>(contents: S) -> Result<OverrideFile> {
+    fn parse_override_file<S: AsRef<str>>(contents: S, toml_only: bool) -> Result<OverrideFile> {
```
kinnison: While booleans are easy, I find myself wishing this was a enum such as:

```rust
pub enum ParseMode {
    OnlyToml,
    Both,
}
```

That way we're calling `Cfg::parse_override_file(contents, ParseMode::Both)` or similar, making it a little clearer.  The compiler will end up reducing it to something as efficient as the boolean, but we're clear at callsites, and type-safe in the face of the API changing in the future.

src/config.rs
```diff
@@ -590,12 +617,12 @@ impl Cfg {
...
         let contents = contents.as_ref();
 
         match contents.lines().count() {
             0 => Err(ErrorKind::EmptyOverrideFile.into()),
-            1 => {
+            1 if !toml_only => {
```
kinnison: With the enum rather than the boolean, this would likely be nicely done as:

```rust
match (contents.lines.count(), parse_mode) {
    (0, _) => ...
    (1, ParseMode::Both) => {...}
    _ => {...}
}
```

tests/cli-rustup.rs
```diff
@@ -2065,3 +2065,53 @@ warning: If you meant to build software to target that platform, perhaps try `ru
...
+
+/// Checks that `rust-toolchain.toml` files are considered
+#[test]
+fn rust_toolchain_toml() {
+    setup(&|config| {
+        let cwd = config.current_dir();
+        let toolchain_file = cwd.join("rust-toolchain.toml");
+        raw::write_file(&toolchain_file, "[toolchain]\nchannel = \"nightly\"").unwrap();
+
+        expect_stdout_ok(config, &["rustc", "--version"], "hash-nightly-2");
```
kinnison, line 2083: I'm not certain but I think `nightly` is the default in this kind of test.  Probably worth ensuring that the output of `rustc --version` is not `hash-nightly-2` before you write the toolchain file, and is that afterwards.  To demonstrate that it's the toolchain file which has caused it.

kinnison (changes requested):
Firstly I'd like to apologise for how long it has taken me to look at this.  I've been somewhat busy at work but my silence wasn't very polite and so I'm sorry.

In general this patch is excellent.  You hilighted the issue with the warning yourself -- you'll need to add a notification kind, and thread that through so that we report it as a warning.  There's plenty of examples in the code but if you want some help let me know.

I'm starting to become averse to boolean values being passed to functions like this though, and if you're up to it, I'd love to see that boolean replaced with a logical enumeration which makes it clear what the values mean.

phil-opp:
Thanks a lot for the review, I fully agree with all your suggestions! No worries about the slight delay. 

I just pushed a new version that should implement your suggestions. Let me know if I did everything right with the new notification kind.

src/config.rs
J-Vin: ✓

J-Vin (approved):

kinnison (changes requested):
I think the only thing left to do is to confirm that the test is *proving* that the toolchain file was in use.  If you can't come up with a way to do that, then ping me and I'll see what I can come up with.

kinnison:
Also, if you're up for it, we should document this cleanly in the relevant section of the book.

phil-opp:
@kinnison Could you clarify what you mean with "proving that the toolchain file was in use"? I already pushed [a commit](https://github.com/rust-lang/rustup/pull/2653/commits/acc69e8594d298c4d334939ce98d3e2f81c99a0c) that checks that `rustc --version` errors before the toolchain file is created. So we can be quite sure that the `rust-toolchain.toml` file is considered by `rustup` because it's the only change between the two `rustc --version` calls.

I'll look into the book changes.

kinnison:
Aha yes, that makes sense, sorry I missed that assertion because I was so focussed on thinking that there was a 'rustup default nightly' as there often is in these tests.  No need to change the test again.  Once you've made the doc tweaks (which I expect to be pretty easy for you) you can mark this ready for final review (i.e. stop it being draft) and then we can look to merge it.  Excellent work.

phil-opp:
Ok great, thanks!

phil-opp:
I updated the _Overrides_ chapter in https://github.com/rust-lang/rustup/pull/2653/commits/21f72289eeddcdfea3378769648807ca819141b9. Let me know if you have any suggestions for improvements.

doc/src/overrides.md
```diff
@@ -85,14 +88,19 @@ targets = [ "wasm32-unknown-unknown", "thumbv2-none-eabi" ]
...
+format that only contains a toolchain name without any TOML encoding, e.g.
+just `nightly-2021-01-21`. This format is not available in
+`rust-toolchain.toml` files.
 
-The `rust-toolchain` file is suitable to check in to source control. This file
-has to be encoded in US-ASCII (if you are on Windows, check the encoding and
-that it does not starts with a BOM).
+The `rust-toolchain.toml`/`rust-toolchain` files are suitable to check in to
+source control. The files have to be encoded in US-ASCII (if you are on
+Windows, check the encoding and that it does not starts with a BOM).
```
kinnison: Only `rust-toolchain` needs to be US-ASCII in theory `rust-toolchain.toml` should always be `utf-8`

kinnison (changes requested):
Almost there.  Could you fix the one doc hiccough, and then rebase your commits so that we don't have 'do foo, fix foo' type commits, but rather a small number of clean commits.  i.e. what you'd have done if you'd know what this would be right from the start?

If you can't rebase then don't worry, but I'd much prefer a clean commit series if possible.

We're nearly there 👍🏻

doc/src/overrides.md
phil-opp: Do you mean only the legacy format of the `rust-toolchain` file or also when the TOML format is used? I assume it only affects the legacy format (see the latest commit), but please let me know if I'm wrong.

I'll rebase as soon as this is ready.

kinnison: Yep, that's correct.  Rebase and I think we're done.

phil-opp:
Rebased.

kinnison (approved):
:+1: I'll merge assuming a green CI (which we can expect)

PR #2657, opened 2021-02-11
Better error message when no toolchain is installed

### LINKED ISSUE #395: 'no updatable toolchains installed'
After my first runing `rustup-init.exe`, I got `rustup.exe`, `cargo.exe` and `rustc.exe` installed at `.cargo/bin`, that's OK. But no toolchain was installed normally for network reasons.

Then I run: `rustup update`
... and expect it will install some toolchains for me, but it just says:

`info: no updatable toolchains installed`

Yes I can do:

```
rustup update stable
rustup update beta
rustup update nightly
rustup update any_other_if_exists
```

But I think at least a `rustup update ALL` (does the similar work as rustup-init) will make my life easier.

---

Just another thing:

```
C:\Users\liigo>cargo
```

Why not suggest user HOW to config or install a toolchain here?

kellda:
Fix #395

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -297,7 +297,7 @@ error_chain! {
         }
         ToolchainNotSelected {
             description("toolchain is not selected")
-            display("no override and no default toolchain set")
+            display("no override and no default toolchain set; run 'rustup default stable' to set the stable toolchain as default")
         }
         OverrideToolchainNotInstalled(t: String) {
             description("override toolchain is not installed")


src/errors.rs
```diff
@@ -297,7 +297,7 @@ error_chain! {
         }
         ToolchainNotSelected {
             description("toolchain is not selected")
-            display("no override and no default toolchain set")
+            display("no override and no default toolchain set; run 'rustup default [toolchain]' to select one")
```
rbtcollins: I don't like the word 'select' here, I think it implies showing a menu - which it won't (but we do on windows during first install kindof).

I can suggest two alternatives to move this forward.

1. s/select/set/
2. extend the places this is used to determine the default toolchain for the platform, and then pass that into the error, so that it can become a copy-and-pastable valid command.

kellda: 1. Sure
2. If there is no default toolchain set, can we really get a default toolchain ? Maybe we should just suggest `rustup default stable`.

rbtcollins: Sure, though stable is just shorthand for stable-{host-triple}

kellda: I changed it. I don't think it is worth to suggest typing the whole host triple.

src/errors.rs
```diff
@@ -297,7 +297,7 @@ error_chain! {
         }
         ToolchainNotSelected {
             description("toolchain is not selected")
-            display("no override and no default toolchain set")
+            display("no override and no default toolchain set; run 'rustup default stable' to download and set the stable toolchain as default")
```
rbtcollins: lets not mention downloading here: it may already be present, or not, and its not the essence of the thing.

kinnison:
I believe the test failure is a spurious one, and I've filed an issue to look into it.  In the meantime, are you now happy with the wording Robert?

daveloyall:
I'm attempting to contribute to this conversation even though I don't know what I'm doing.  (Maybe that is helpful.  Maybe people who know what they are doing never see this error message anyway.)

I thought it would be useful to build rust programs on windows under the GNU tools provided by Msys2... So, my toolchain is `nightly-x86_64-pc-windows-gnu`.  I followed a guide..

`rustup toolchain list` returns that toolchain, the full name.  And then `rustup default nightly-x86_64-pc-windows-gnu` seems to have stopped `cargo` from printing out the error message described here in #2657.

If I understand correctly (and that's a big if!), if I had done `rustup default stable` instead.. then I would have ended up using the Microsoft compiler, instead of the GNU one...

I hope this helps.

rbtcollins:
@daveloyall we haven't released this change yet.

PR #2665, opened 2021-02-16
Deploy docs for latest master.

ehuss:
This changes it so that every merge to master and stable will trigger the docs to be rebuilt.  The stable docs appear at https://rust-lang.github.io/rustup/ and the latest master appears at https://rust-lang.github.io/rustup/devel/.

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - stable
+      - master
 
 jobs:
   doc:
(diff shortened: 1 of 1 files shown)


kinnison:
As I'm reading this it looks like changes to master will also rebuild stable's with the new mdbook.  I think that's probably fine.

Is there any way mdbook can fail, and if so, is it worth having a doc workflow which generates an artifact someone could download from a PR to check that their docs built okay?

I'll merge this though because it looks good. ❤️

ehuss:
> Is there any way mdbook can fail, and if so, is it worth having a doc workflow which generates an artifact someone could download from a PR to check that their docs built okay?

Generally, there are very few ways mdbook can fail in normal operation.  Usually mistakes in `book.toml` or a malformed `SUMMARY.md`.  

I'm not sure I understand what you mean about generating an artifact.  You can run `mdbook build` locally and generally get the same results.  AFAIK, the only way to store artifacts with just GitHub's setup is via GitHub Releases, which we wouldn't want to use, or committing them to a git repository.  Otherwise, you have to setup your own server to host it.

kinnison:
We can serve artifacts from pipelines (we actually provide rustup-init from the pipelines) but I think it's not that useful if mdbook is easy to run locally.

PR #2667, opened 2021-02-16
Handle modifying rc files that don't have a trailing newline

### LINKED ISSUE #2636: rcfiles: We need to cope correctly when the rcfile does not have a newline at the end and we need to append the source command.
I am totally new to Rust and I wanted to install it. I followed the installation guide on the [website](https://www.rust-lang.org/learn/get-started). I use Linux Mint, thus I ran the following in a terminal:

```bash
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```
After this, when I open a new terminal, the ``bash: export: `/home/user/.cargo/env': not a valid identifier`` message appears on top of the terminal.

I figured out it is due to the following line which has been added during the installation in the `.bashrc` file:

```bash
export PATH=$PATH:$HOME/.local/binsource "$HOME/.cargo/env"
```

In the end, I solved the identifier error changing the previous line in the `.bashrc` file with this:

```bash
export PATH=$PATH:$HOME/.local/binsource:$HOME/.cargo/env
```

I open this issue because I think fixing this in the installation procedure could help some new inexperienced user (like me) facing it at the end of the installation procedure

mythmon:
This changes the flow so that instead of checking if `&rc.is_file`, it just tries to read the file and handles the error. I think this an ok way to handle it, though it might produce a little more IO. It simplifies the logic a lot.

I couldn't find any tests that cover this function, and I'm not comfortable adding tests right now, but I could try and tackle that with some guidance, or if there is a similar test I could use an example.

Fixes #2636.

### DIFF
--- a/src/cli/self_update/unix.rs
+++ b/src/cli/self_update/unix.rs
@@ -83,14 +83,20 @@ pub fn do_remove_from_path() -> Result<()> {
 pub fn do_add_to_path() -> Result<()> {
     for sh in shell::get_available_shells() {
         let source_cmd = sh.source_string()?;
+        let source_cmd_with_newline = format!("\n{}", &source_cmd);
+
         for rc in sh.update_rcs() {
-            if !rc.is_file() || !utils::read_file("rcfile", &rc)?.contains(&source_cmd) {
-                utils::append_file("rcfile", &rc, &source_cmd).chain_err(|| {
-                    ErrorKind::WritingShellProfile {
-                        path: rc.to_path_buf(),
-                    }
-                })?;
-            }
+            let cmd_to_write = match utils::read_file("rcfile", &rc) {
+                Ok(contents) if contents.contains(&source_cmd) => continue,
+                Ok(contents) if !contents.ends_with('\n') => &source_cmd_with_newline,
+                _ => &source_cmd,
+            };
+
+            utils::append_file("rcfile", &rc, &cmd_to_write).chain_err(|| {
+                ErrorKind::WritingShellProfile {
+                    path: rc.to_path_buf(),
+                }
+            })?;
         }
     }
 
--- a/tests/cli-paths.rs
+++ b/tests/cli-paths.rs
@@ -158,6 +158,47 @@ export PATH="$HOME/apple/bin"
     }
 
     #[test]
+    fn install_adds_path_to_rc_handling_no_newline() {
+        clitools::setup(Scenario::Empty, &|config| {
+            let profile = config.homedir.join(".profile");
+            let fake_rc_modified = FAKE_RC.strip_suffix('\n').expect("Should end in a newline");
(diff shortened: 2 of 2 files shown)


mythmon:
I found the tests in the CI build, and added one for this. I ran `cargo test` before pushing this PR. Any idea why it didn't run that test?

kinnison:
I'm not sure why the test didn't run for you locally, you can clearly see it running in CI here - https://github.com/rust-lang/rustup/pull/2667/checks?check_run_id=1914524215#step:17:1345

src/cli/self_update/unix.rs
```diff
@@ -83,14 +83,20 @@ pub fn do_remove_from_path() -> Result<()> {
...
         for rc in sh.update_rcs() {
-            if !rc.is_file() || !utils::read_file("rcfile", &rc)?.contains(&source_cmd) {
-                utils::append_file("rcfile", &rc, &source_cmd).chain_err(|| {
-                    ErrorKind::WritingShellProfile {
-                        path: rc.to_path_buf(),
-                    }
-                })?;
-            }
+            let cmd_to_write = match utils::read_file("rcfile", &rc) {
+                Ok(contents) if contents.contains(&source_cmd) => break,
```
kinnison: I fear this `break` ought to be `continue` so that it considers the other rc files.

tests/cli-paths.rs
```diff
@@ -150,6 +150,21 @@ export PATH="$HOME/apple/bin"
...
     }
 
+    #[test]
+    fn install_adds_path_to_rc_handling_no_newline() {
+        clitools::setup(Scenario::Empty, &|config| {
+            let profile = config.homedir.join(".profile");
+            let fake_rc_modified = FAKE_RC.strip_suffix('\n').expect("Should end in a newline");
+            raw::write_file(&profile, fake_rc_modified).unwrap();
+            expect_ok(config, &INIT_NONE);
+            expect_ok(config, &INIT_NONE);
```
kinnison: Why are you running rustup-init twice here?  As in, it's worth putting a coment in to explain what's going on, i.e. what we're trying to ensure does not rear its head once more.

kinnison (changes requested):
One potential tweak to be done to coments in the tests, and a `break` which I *think* should be a `continue`

Otherwise this looks good, thank you for making the effort to contribute.

kinnison (changes requested):
This content is looking excellent, thank you for the extra test too, that is very useful.

Could you rebase the commits so that it's a clean change, and then we can look to merge this.

Thank you for your work.

mythmon:
Alright, rebased and squashed down to one commit. Thanks for the review and guidance!

kinnison (approved):
:+1:

I'll merge on green-CI

PR #2669, opened 2021-02-17
Add on_log_quit.

### LINKED ISSUE #2590: rustup-init leaves user with colored terminal
*(If this doesn't belong here, please tell me which repo to move it to, thanks!)*

**Problem**

`rustup`'s setup script, `rustup-init`, leaves the user with a colored terminal when exiting interrupted.

**Steps**

1. Execute `rustup-init`
2. At the first prompt, interrupt by Ctrl-C
3. User's shell is now red.

**Possible Solution(s)**
Check the interrupted-exit code maybe? Make sure that the color is reset before exiting (Since `^C` is displayed in **red**)

**Notes**

OS: Win 10 (1909)
Output of `rustup --version`: `rustup 1.23.0 (00924c9ba 2020-11-27)`
Output of `rustup-init --version`: `rustup-init 1.22.1 (b01adbbc3 2020-07-08)`
Output of `rustup show`:
```
rustup home:  C:\Users\kangj\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc

active toolchain
----------------

rustc 1.48.0-nightly (6af1bdda5 2020-09-15)
```

guilt:
Request for comments;

Closes https://github.com/rust-lang/rustup/issues/2590
 
 I was unable to find a working at_exit that is publicly available in stable rust today. So I decided to post this and ask what may be the right way to attempt this?

### DIFF
--- a/src/bin/rustup-init.rs
+++ b/src/bin/rustup-init.rs
@@ -13,14 +13,19 @@
 //! different name.
 
 #![recursion_limit = "1024"]
+#![feature(rt)]
 
 use std::path::PathBuf;
+#[cfg(feature = "rt")]
+use std::rt;
 
 use cfg_if::cfg_if;
 use rs_tracing::*;
 
 use rustup::cli::common;
 use rustup::cli::errors::*;
+#[cfg(feature = "rt")]
+use rustup::cli::log;
 use rustup::cli::proxy_mode;
 use rustup::cli::rustup_mode;
 #[cfg(windows)]
@@ -42,6 +47,10 @@ fn main() {
 }
 
 fn run_rustup() -> Result<utils::ExitCode> {
+    #[cfg(feature = "rt")]
+    {
+        let _ = rt::at_exit(log::on_quit_log);
+    }
     if let Ok(dir) = process().var("RUSTUP_TRACE_DIR") {
         open_trace_file!(dir)?;
     }
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -23,6 +23,11 @@ macro_rules! debug {
     ( $ ( $ arg : tt ) * ) => ( $crate::cli::log::debug_fmt ( format_args ! ( $ ( $ arg ) * ) ) )
 }
 
+pub fn on_quit_log() {
+    let mut t = term2::stderr();
+    let _ = t.reset();
+}
+
 pub fn warn_fmt(args: fmt::Arguments<'_>) {
     let mut t = term2::stderr();
     let _ = t.fg(term2::color::YELLOW);


kinnison:
Does this demonstrably fix the issue in question?  It's relying on a feature flag which means we won't accept it into Rustup because we build with `stable` for releases, but it could point the way to warranting investigation into a way to resolve this in a stable release if you are certain it functions.

In the meantime, since this is not mergeable as-is regardless of correctness, I have marked the PR as a draft

guilt:
Thanks! 

I will generally wait for the at_exit to get merged to stable/beta at this point. This isn't a priority for 1.24.x, but the reset handler would have to be called somewhere.

To reproduce, I would basically introduce some artificial delays in the log code, press Ctrl-C/SIGKILL and see if it resets correctly. It will need some time from me to report what I find, on macOS/Windows/Linux. I am still unsure how to write a testcase for this, this is a 'noob' issue for me to get acquainted with the code.

I am going to try testing this with some signal handling code as well and report what I found. But this isn't an uncommon issue. A lot of commands fail and I usually run `stty sane` at that point before typing stuff on the Terminal again.

rbtcollins:
I don't think at_exit is the right way to tackle this. It's kindof sad Rust is getting at_exit hooks really. But I digress.

A simple signal handler on unix family OSes, and [SetConsoleCtrlHandler](https://www.reddit.com/r/rust/comments/57lcm6/if_rust_doesnt_have_exceptions_what_happens_on/d8syo1a/?utm_source=reddit&utm_medium=web2x&context=3) on Windows, can call the term reset function we want - but ideally we'd be integrated into stack unwinding to allow cleaning up of held locks etc - while we should be crash safe, we can be more efficient sometimes if we have cleanup code :)

rbtcollins:
Also see https://rust-cli.github.io/book/in-depth/signals.html and perhaps https://doc.rust-lang.org/nomicon/unwinding.html as well

workingjubilee:
> I don't think at_exit is the right way to tackle this. It's kindof sad Rust is getting at_exit hooks really. But I digress.

That's because it is not. This PR relies on the unstable-internal `rt` feature which is not meant to be used by ordinary consumers of the Rust standard library which are not directly implementing libraries and drivers for the Rust compiler or instrumenting the compiler and its runtime.

> I will generally wait for the at_exit to get merged to stable/beta at this point. This isn't a priority for 1.24.x, but the reset handler would have to be called somewhere.

For the above-mentioned reasons, this is extremely unlikely to happen.
See https://doc.rust-lang.org/unstable-book/library-features/rt.html

PR #2670, opened 2021-02-17
Add an entry to the installed programs on windows

### LINKED ISSUE #1618: Put an item into the Add / Remove Programs directory
It doesn't seem right that people should be stuck not knowing how to uninstall rustup. If you really want to make someone feel like they don't have control of their computer, make it so that when they fail to get started with Rust, they don't even know how to put things back the way they were.

It should be possible to integrate rustup with the Windows software manager.

kellda:
Fixes #1618

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -344,6 +344,7 @@ pub fn install(
         do_write_env_files()?;
 
         if !opts.no_modify_path {
+            do_add_to_programs()?;
             do_add_to_path()?;
         }
         utils::create_rustup_home()?;
@@ -855,6 +856,7 @@ pub fn uninstall(no_prompt: bool) -> Result<utils::ExitCode> {
 
     // Remove CARGO_HOME/bin from PATH
     do_remove_from_path()?;
+    do_remove_from_programs()?;
 
     // Delete everything in CARGO_HOME *except* the rustup bin
 
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -1,4 +1,6 @@
 use std::env::consts::EXE_SUFFIX;
+use std::ffi::{OsStr, OsString};
+use std::os::windows::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 use std::process::Command;
 
@@ -9,6 +11,9 @@ use crate::dist::dist::TargetTriple;
 use crate::process;
 use crate::utils::utils;
 use crate::utils::Notification;
+
+use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
+use winreg::{RegKey, RegValue};
 
 pub fn ensure_prompt() -> Result<()> {
     writeln!(process().stdout(),)?;
@@ -42,7 +47,6 @@ pub fn do_msvc_check(opts: &InstallOpts<'_>) -> Result<bool> {
 
 /// Run by rustup-gc-$num.exe to delete CARGO_HOME
 pub fn complete_windows_uninstall() -> Result<utils::ExitCode> {
-    use std::ffi::OsStr;
     use std::process::Stdio;
 
     wait_for_parent()?;
@@ -150,8 +154,6 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
     use winapi::um::winuser::{
         SendMessageTimeoutA, HWND_BROADCAST, SMTO_ABORTIFHUNG, WM_SETTINGCHANGE,
     };
-    use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
-    use winreg::{RegKey, RegValue};
 
     let new_path = match new_path {
         Some(new_path) => new_path,
@@ -198,8 +200,6 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
 // should not mess with it.
 fn get_windows_path_var() -> Result<Option<Vec<u16>>> {
     use std::io;
-    use winreg::enums::{HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
-    use winreg::RegKey;
 
     let root = RegKey::predef(HKEY_CURRENT_USER);
     let environment = root
@@ -270,9 +270,6 @@ fn _with_path_cargo_home_bin<F>(f: F) -> Result<Option<Vec<u16>>>
 where
     F: FnOnce(Vec<u16>, Vec<u16>) -> Option<Vec<u16>>,
 {
-    use std::ffi::OsString;
-    use std::os::windows::ffi::OsStrExt;
-
     let windows_path = get_windows_path_var()?;
     let mut path_str = utils::cargo_home()?;
     path_str.push("bin");
@@ -283,6 +280,55 @@ where
 pub fn do_remove_from_path() -> Result<()> {
     let new_path = _with_path_cargo_home_bin(_remove_from_path)?;
     _apply_new_path(new_path)
+}
+
+const RUSTUP_UNINSTALL_ENTRY: &str = r"Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup";
+
+pub fn do_add_to_programs() -> Result<()> {
+    use std::path::PathBuf;
+
+    let key = RegKey::predef(HKEY_CURRENT_USER)
+        .create_subkey(RUSTUP_UNINSTALL_ENTRY)
+        .chain_err(|| ErrorKind::PermissionDenied)?
+        .0;
+
+    // Don't overwrite registry if Rustup is already installed
+    let prev = key
+        .get_raw_value("UninstallString")
+        .map(|val| from_winreg_value(&val));
+    if let Ok(Some(s)) = prev {
+        let mut path = PathBuf::from(OsString::from_wide(&s));
+        path.pop();
+        if path.exists() {
+            return Ok(());
+        }
+    }
+
+    let mut path = utils::cargo_home()?;
+    path.push("bin\rustup.exe");
+    let mut uninstall_cmd = OsString::from("\"");
+    uninstall_cmd.push(path);
+    uninstall_cmd.push("\" self uninstall");
+
+    let reg_value = RegValue {
+        bytes: to_winreg_bytes(uninstall_cmd.encode_wide().collect()),
+        vtype: RegType::REG_SZ,
+    };
+
+    key.set_raw_value("UninstallString", &reg_value)
+        .chain_err(|| ErrorKind::PermissionDenied)?;
+    key.set_value("DisplayName", &"Rustup: the Rust toolchain installer")
+        .chain_err(|| ErrorKind::PermissionDenied)?;
+
+    Ok(())
+}
+
+pub fn do_remove_from_programs() -> Result<()> {
+    match RegKey::predef(HKEY_CURRENT_USER).delete_subkey_all(RUSTUP_UNINSTALL_ENTRY) {
+        Ok(()) => Ok(()),
+        Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => Ok(()),
+        Err(e) => Err(e).chain_err(|| ErrorKind::PermissionDenied),
+    }
 }
 
 /// Convert a vector UCS-2 chars to a null-terminated UCS-2 string in bytes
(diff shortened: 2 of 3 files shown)


kellda:
I have a few questions about how this should be implemented:

- Where should the functions I added be called ?
- Should they be gated on some option (like for modifying the path)

- What should be displayed in the list of installed programs ? (e.g. "rustup" or "The Rust Programming Language")
- Should the registry key be a fixed string (easier to find in the registry) or a randomly generated GUID (easier to have multiple installations of rustup) ?

kinnison:
Taking your questions in order:

1. These functions likely will end up called from `cli::self_update::install` and `cli::self_update::windows::complete_windows_uninstall` though these are just guesses right now
2. I feel like adding to the windows registry uninstall set should probably just be done, but one big problem is that if people installed without modifying the path then they might just choose to delete the folders.  Perhaps only adding to the registry if PATH was modified makes sense.
3. I would suggest `rustup - Rust toolchain manager` or similar
4. My gut feeling is that in general there'll only ever be one installation of `rustup` on a Windows system.  *HOWEVER* we do support more than one, and nominally people *could* have several.  I'd suggest, if possible, that we work on the basis of a GUID derived from the combo of `CARGO_HOME` and `RUSTUP_HOME` though I don't know how that'd go, and if we did that, then the text for `3` would need to be augmented with those values so that the user knew which they were uninstalling.

Ideally we'd get someone who actually uses Windows seriously, rather than treats it like a toy OS (me), to weigh in with some understanding of how people *expect* this kind of thing to present itself.  Otherwise we're just going to be speculating as to what might be best, without being informed.

Thank you for your work thus-far.

kellda:
About 1. and 2., I think it makes sense to call these functions in `cli::self_update::[un]install`, just after `do_[add_to/remove_from]_path`.

3. sounds good.

About 4., I was mainly thinking about testing rustup while developing it and in the test suite. On Windows most programs just uninstall any previous version before installing themselves. I'm for adding uninstall information only if it doesn't already contain valid ones and only remove it if it points to the installation we're removing.

I think rustup uninstall RUSTUP_HOME regardless of where is the called installation. Is this correct ?

kinnison:
Sorry it has taken so long to get back to you on this one.  I'm struggling to cope with the fact that Windows is making us conflate global and local configuration to some extent.  `rustup` can be installed multiple times, and I think each uninstall entry will have to handle the specific `RUSTUP_HOME` and `CARGO_HOME` it's related to - however the installations could feasibly *overlap* (it'd be broken to do so, but it's at least slightly possible to some extent) - I'm beginning to wonder if I was right to ask about coping with uninstalling multiple variants.  Perhaps we should go back to your much simpler original idea of a single entry?

kellda:
I also think that it is not really needed nor useful to have multiple entries in the registry when there is multiple installations. IMO there should be one "main" installation in the registry, the one that was installed first and that's called when `RUSTUP_HOME` is not set/overridden.

kinnison:
So only set the uninstall in the registry if `RUSTUP_HOME` and `CARGO_HOME` are not overridden in the environment?  That sounds plausible.

kellda:
I didn't implemented it that way, but I can do it so.

What is currently implemented is:
- Check if there is already an entry in the registry
- If there isn't, add uninstall information
- If the entry points to a valid location, don't modify the registry
- If the entry is invalid, override it with info of this installation

And on uninstall, only remove the registry's entry if it points to the currently running executable.

Is it better to check the state of env variables or that of the registry ?

kinnison:
I'm afraid I'm not very good at reasoning out what a Windows user would expect here.  @rbtcollins Do you happen to have any insight which might help?

src/cli/self_update.rs
```diff
@@ -344,6 +344,8 @@ pub fn install(
 
         if !opts.no_modify_path {
             do_add_to_path()?;
+            #[cfg(windows)]
+            do_add_to_programs()?;
```
rbtcollins, line 347: There are desktop environments on Unix OSes too that have similar app registration mechanisms that we should perhaps be using. Rather than cfg windows, perhaps just a register_program() call, and a stub for unix platforms, and something real for windows?

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
     _apply_new_path(new_path)
 }
 
+pub fn do_add_to_programs() -> Result<()> {
+    use std::path::PathBuf;
+
+    let key = RegKey::predef(HKEY_CURRENT_USER)
+        .create_subkey(r"Software\Microsoft\Windows\CurrentVersion\Uninstall\rustup")
```
rbtcollins: This should be `Rustup`, not `rustup`, as the project name is capitalised.

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
...
+        .chain_err(|| ErrorKind::PermissionDenied)?
```
rbtcollins, line 292: Is that the only possible error that can occur?

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
...
+        path.pop();
+        if path.exists() {
+            return Ok(());
+        }
+    }
+
+    let mut path = utils::cargo_home()?;
+    path.push("bin\rustup.exe");
+    let mut uninstall_cmd = path.into_os_string();
+    uninstall_cmd.push(" self uninstall");
```
rbtcollins: This doesn't seem to correctly quote the path if the home directory contains spaces. I'm not sure if we can use a REG_MULTI_SZ here? That would be better.

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
...
+
+    let reg_value = RegValue {
+        bytes: to_winreg_bytes(uninstall_cmd.encode_wide().collect()),
+        vtype: RegType::REG_SZ,
+    };
+
+    key.set_raw_value("UninstallString", &reg_value)
+        .chain_err(|| ErrorKind::PermissionDenied)?;
```
rbtcollins, line 319: similarly here, is this narrowing the reported errors too much ?

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
...
+    key.set_value("DisplayName", &"rustup - Rust toolchain manager")
+        .chain_err(|| ErrorKind::PermissionDenied)?;
```
rbtcollins, line 321: and so on

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
...
+    Ok(())
+}
+
+pub fn do_remove_from_programs() -> Result<()> {
+    use std::io;
+    use std::path::PathBuf;
+
+    let root = RegKey::predef(HKEY_CURRENT_USER);
+
+    let key = match root.open_subkey(r"Software\Microsoft\Windows\CurrentVersion\Uninstall\rustup")
```
rbtcollins: lets factor this out into a const or static

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
...
+    {
+        Ok(key) => key,
+        Err(ref e) if e.kind() == io::ErrorKind::NotFound => return Ok(()),
+        Err(e) => return Err(e).chain_err(|| ErrorKind::PermissionDenied),
```
rbtcollins: error folding...

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,75 @@ pub fn do_remove_from_path() -> Result<()> {
...
+    };
+
+    // Don't remove uninstall information from another installation
```
rbtcollins: I think we should remove this: if someone has moved their homedir around for whatever reason, this logic will make it impossible to remove the entry from their programs list - and thats super annoying.

OTOH removing an entry thats not for a parallel install done using CARGO_HOME + RUSTUP_HOME variables, is somewhat harmless, since the user can still use rustup, and still uninstall directly, as long as we document that.

rbtcollins (changes requested):

rbtcollins:
@kinnison review done. rustup *cannot* be installed multiple times today. Well, not meaningfully. You just think it can :P. 

The PATH behaviour really means that rustup is only installed once: the first entry on the path wins, and anything else loses. Some Windows programs both add themselves to the registry PATH and offer a batch script to add themselves to the process environment block PATH by spawning a command shell, setting the variables and dropping the user into it, thus allowing the user to work around the PATH first-wins issue.

So someone setting custom CARGO_HOME and RUSTUP_HOME - both on Windows and elsewhere - really is taking responsibility for managing the definition of installed, and I think our best bet is just to ignore that.

src/cli/self_update/windows.rs
kellda: I didn't really think about it as I took it from the PATH handling functions, which also deals with the registry.

Looking through `src/cli/errors.rs` only `PermissionDenied` seems appropriate. Functions in winreg return an `io::Result`, which will also be displayed when an error occur.

src/cli/self_update/windows.rs
kellda: REG_MULTI_SZ only use the first string as if it was a REG_SZ, so it doesn't help here. Quoting the path should work.

src/cli/self_update/windows.rs
kellda: You're talking about the registry path, right ? What about the text displayed in the list of installed programs (`rustup - Rust toolchain manager` or `Rustup - Rust toolchain manager`)

kellda: By the way, I can't see any place where the project name is capitalised.

rustup.rs:

> rustup is an installer for...

>rustup is an official Rust project.

README.md:

>rustup: the Rust toolchain installer

rbtcollins: `Rustup` please. We discussed this on discord prior to my saying this in the review; I realise the prior art in the repo may be inconsistent.

src/cli/self_update/windows.rs
rbtcollins: https://docs.rs/winreg/0.8.0/winreg/struct.RegKey.html#method.create_subkey is the docs, looking at the source its backed by https://docs.rs/winreg/0.8.0/src/winreg/lib.rs.html#387-389 which is backed by winapi_reg::RegCreateKeyExW which means we should look at https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regcreatekeyexw which helpfully is unclear.

Anyhow, I guess this will have to do for now, though I'm certain other errors can occur. shrug.

kellda: An alternative would be to add a variant in cli::errors::ErrorKind for errors that occur when using the registry

src/cli/self_update.rs
```diff
@@ -344,6 +344,7 @@ pub fn install(
 
         if !opts.no_modify_path {
             do_add_to_path()?;
```
rbtcollins: I think the add to programs should be done first, before adding to path.

src/cli/self_update/windows.rs
```diff
@@ -285,6 +282,55 @@ pub fn do_remove_from_path() -> Result<()> {
...
+
+pub fn do_add_to_programs() -> Result<()> {
+    use std::path::PathBuf;
+
+    let key = RegKey::predef(HKEY_CURRENT_USER)
+        .create_subkey(RUSTUP_UNINSTALL_ENTRY)
+        .chain_err(|| ErrorKind::PermissionDenied)?
+        .0;
+
+    // Don't overwrite registry if Rustup is already installed
```
rbtcollins, line 295: I'm not certain this is right, but for now lets leave it. I think it might be more that we want to determine default RUSTUP_HOME being overridden etc, and otherwise we should be writing the current correct value into the registry each time.

rbtcollins (changes requested):
Once you've done this change please rebase - we don't want the incremental changes that lead the PR, just the clear set of changes.

src/cli/self_update/windows.rs
glandium: What would be the "current correct value"? As someone who overrides RUSTUP_HOME and use temporary installs of  rustup in parallel of the standard one, I wouldn't want those temporary installs to replace the standard one in the registry.

kellda: A later install wont override a previous one, as long as the previous install is still installed.
However when uninstalling, the registry key will be unconditionally removed, and then recreated on the next install. See https://github.com/rust-lang/rustup/pull/2670#discussion_r592748920

By the way, any `rustup self uninstall` will unistall rustup in the current `RUSTUP_HOME`, regardless of where the executable you're running is.

PR #2674, opened 2021-02-20
Set `upgrade` to the alias of `update`

### LINKED ISSUE #2660: Consider `upgrade` as alias for `update`
Some people are used to saying `upgrade` when wanting to update toolchains.  As such it might be nice to offer this as an alias of `rustup update` which could either be visible or hidden.

I am happy to mentor someone through this work.

chansuke:
Fixes #2660

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -297,6 +297,7 @@ pub fn cli() -> App<'static, 'static> {
         .subcommand(
             SubCommand::with_name("update")
                 .about("Update Rust toolchains and rustup")
+                .aliases(&["upgrade"])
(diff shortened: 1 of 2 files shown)


kinnison (approved):
I think this looks good.  The use of `aliases` rather than `visible_aliases` means I am happy we don't need to update the docs to this effect.

I do not intend to advertise this in the changelog, beyond thanking you for your contribution, because I don't want to *encourage* the use of this alias, merely have it as a fortuitous benefit for those who accidentally need it.

I shall merge assuming I get a green CI

chansuke:
@kinnison Thank you for your comments and feedback!!

PR #2675, opened 2021-02-21
compression: Refactor handling of compression formats

kinnison:
In preparation for adding zstd support to rustup, this rework
means that we treat compression formats we have available (and know)
generically without breaking if unexpected formats are present,
or if no format we know is present.

In theory this means that we won't error out if a component is labelled "available" but then has no compression format we understand.  Still things ought to behave well enough.

### DIFF
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -47,16 +47,34 @@ pub enum PackageTargets {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct TargetedPackage {
-    pub bins: Option<PackageBins>,
+    pub bins: Vec<(CompressionKind, HashedBinary)>,
     pub components: Vec<Component>,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum CompressionKind {
+    GZip,
+    XZ,
+}
+
+/// Each compression kind, in order of preference for use, from most desirable
+/// to least desirable.
+static COMPRESSION_KIND_PREFERENCE_ORDER: &[CompressionKind] =
+    &[CompressionKind::XZ, CompressionKind::GZip];
+
+impl CompressionKind {
+    const fn key_prefix(self) -> &'static str {
+        match self {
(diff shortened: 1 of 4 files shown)


src/dist/manifest.rs
```diff
@@ -47,16 +47,34 @@ pub enum PackageTargets {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct TargetedPackage {
-    pub bins: Option<PackageBins>,
+    pub bins: Vec<(CompressionKind, HashedBinary)>,
```
rbtcollins, line 50: Rather than a tuple, what about a data carrying enum ?

```
GZip(HashedBinary),
XZ(HashedBinary),
ZStd(HashedBinary)
```

Or alternatively, a field in HashedBinary:

`pub kind CompressionKind,`
I'm not sure it would turn out better, but I think if we always want to know we shouldn't rely on storing them adjacently in a vector

rbtcollins (approved):
I think this is a nice little improvement, but I think you're leaving some more niceness on the table. That shouldn't block this in any way - we can always do it later.

src/dist/manifest.rs
kinnison: I would have stored this as a `HashMap<CompressionKind, HashedBinary>` but for the fact that (a) HashMap is not PartialEq, and (b) I wanted to retain a preference order in there.  As such, given you don't feel the need to block on this, I'm going to merge.

kinnison: As for an enum which carries data, that makes storing a preference order in a const a bit harder :D

rbtcollins: So there are two separate concerns here which I think you've tangled up.

One is the knowledge that there are serialised prefixes in the external table.
The other is the tangled threads through the callers of the kind, which is shown by the places this touched and the (kind, url, path) tuple later on.

We don't necessarily need one kind for both of these things, and the preference order is only relevant, if at all, for the former.

Issue #1858, opened 2019-05-18
Prefer faster compression methods under good network


ishitatsuyuki:
**Describe the problem you are trying to solve**

Decompression can be an bottleneck if the network is fast. XZ (LZMA) is surprising slow and can quickly dominate the time spent on installation.

For network connections faster than broadband, we should consider not using LZMA compression. This includes home fiber and datacenters, notably CI environments.

**Describe the solution you'd like**

- Add a configuration so the user can select preferred compression method.
- (If possible) Add an `auto` option that tries to detect network speed.
- Consider distributing archives with new algorithms such as zstandard.

**Notes**

[Ubuntu proposal case study](https://lists.ubuntu.com/archives/ubuntu-devel/2018-March/040211.html).

kinnison:
For the most part, the compression scheme selected is up to the tool which generates the compiler releases.  As far as I know, however, the manifest format only allows for a single download file per component, so we wouldn't be able to choose compression methods for the user.  I imagine that we use xz for compression ratio reasons.

rbtcollins:
@ishitatsuyuki do you have some profiling data showing that decompression is a bottleneck here? One thing we could very easily do if it is a bottleneck is move decompression to a dedicated thread.

ishitatsuyuki:
Reference file: `183M    rust-std-1.34.2-x86_64-pc-windows-gnu.tar`  
CPU: `Intel(R) Core(TM) i7-6500U CPU`

| Compression | Compressed size | Download time | Decompression time |
|-------------|-----------------|---------------|--------------------|
| gz          | 71M             | 4.4s          | 1.25s              |
| xz          | 56M             | 3.4s          | 3.50s              |
| zstd (-19)  | 59M             | N/A           | 0.34s              |

You can clearly see that when the internet is fast, decompression time is a significant part of installation time (on Linux where I/O doesn't get interfered by antimalware). Moving it to a dedicated thread **doesn't make anything faster** since the CPU is the bottleneck.

I run a NVMe drive so files can be written almost instantly, but for rotating disks it may take much more time.

My Wi-Fi which is used for download test had an average throughput of 16.2MB/s. Using a wired connection is even faster, yielding around 80MB/s. This resembles typical datacenter networking speed.

As a bonus, zstd decompression is faster than any other methods listed here, while retaining a similar compression ratio to xz. Time to adopt.

rbtcollins:
Thank you for the profiling data.

So, lets break this into several aspects.

[...]

PR #2676, opened 2021-02-21
Add zstd support

### LINKED ISSUE #2488: Add support for `.zst` for component tarballs in channels

As per #1858 and other places, it would be good to support `zstd` compression for component packages in channels.  This would allow for smaller, faster-to-decompress components which is a benefit all round.

Support for `.zst` as a file extension and `zstd` decompression as part of Rustup.

**Notes**

This ought mostly to be constrained to the `src/dist/` tree, particularly the `manifestation.rs` and `component/package.rs` files.  Some test support will also be needed.

Since `zstd` generally compresses better and decompresses more quickly than `xz`, we should prefer `.zst` over `.xz` where it is present in the manifest.

Once implemented, the compiler team could start to produce channels with `zstd` compressed content

kinnison:
This builds atop #2675 and adds support for ZStd compression, both to rustup and to the test suite.

This fixes #2488 and in theory goes some way toward #1858 though given this is a linear preference of zst over xz over gz it's not entirely a fix.

This patch assumes it's okay to insert zstd into manifest v2 files without bumping the manifest version.  It follows the `xz_tar` pattern with `zst_tar` et al.  Also since it builds atop the refactor in #2675 it doesn't *require* that gz still be present, paving the way for seamless conversion from gz+xz, via some combination of gz+(xz and/or zst) to a pure zst manifest if that is of interest to the infra team.

I'd appreciate opinions from @rust-lang/infra on this - should we bump the manifest version since this implicitly makes the gz files optional?

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -7,7 +7,9 @@ use retry::delay::NoDelay;
 use retry::{retry, OperationResult};
 
 use crate::config::PgpPublicKey;
-use crate::dist::component::{Components, Package, TarGzPackage, TarXzPackage, Transaction};
+use crate::dist::component::{
+    Components, Package, TarGzPackage, TarXzPackage, TarZStdPackage, Transaction,
+};
 use crate::dist::config::Config;
 use crate::dist::dist::{Profile, TargetTriple, DEFAULT_DIST_SERVER};
 use crate::dist::download::{DownloadCfg, File};
@@ -238,6 +240,7 @@ impl Manifestation {
             };
             let gz;
             let xz;
+            let zst;
             let reader =
                 utils::FileReaderWithProgress::new_file(&installer_file, &notification_converter)?;
             let package: &dyn Package = match format {
@@ -248,6 +251,10 @@ impl Manifestation {
                 CompressionKind::XZ => {
                     xz = TarXzPackage::new(reader, temp_cfg, Some(&notification_converter))?;
                     &xz
+                }
+                CompressionKind::ZStd => {
+                    zst = TarZStdPackage::new(reader, temp_cfg, Some(&notification_converter))?;
+                    &zst
                 }
             };
 
--- a/tests/dist.rs
+++ b/tests/dist.rs
@@ -573,33 +573,8 @@ fn setup_from_dist_server(
     );
 }
 
-#[test]
-fn initial_install() {
-    setup(None, GZOnly, &|url,
-                          toolchain,
-                          prefix,
-                          download_cfg,
-                          temp_cfg| {
-        update_from_dist(
-            url,
-            toolchain,
-            prefix,
-            &[],
-            &[],
-            download_cfg,
-            temp_cfg,
-            false,
-        )
-        .unwrap();
-
-        assert!(utils::path_exists(&prefix.path().join("bin/rustc")));
-        assert!(utils::path_exists(&prefix.path().join("lib/libstd.rlib")));
-    });
-}
-
-#[test]
-fn initial_install_xz() {
-    setup(None, AddXZ, &|url,
+fn initial_install(comps: Compressions) {
+    setup(None, comps, &|url,
                          toolchain,
                          prefix,
                          download_cfg,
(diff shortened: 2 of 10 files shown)


src/dist/manifestation.rs
```diff
@@ -11,22 +11,19 @@ use crate::dist::component::{Components, Package, TarGzPackage, TarXzPackage, Tr
...
-use crate::dist::manifest::{Component, Manifest, TargetedPackage};
+use crate::dist::manifest::{Component, CompressionKind, Manifest, TargetedPackage};
 use crate::dist::notifications::*;
 use crate::dist::prefix::InstallPrefix;
 use crate::dist::temp;
 use crate::errors::*;
 use crate::process;
 use crate::utils::utils;
 
+use super::component::TarZStdPackage;
```
rbtcollins: not clear why this and only this package type appears here - shouldn't it be on line 10 with the others?

tests/dist.rs
```diff
@@ -496,24 +498,50 @@ fn uninstall(
     Ok(())
 }
 
+#[derive(Copy, Clone, Debug)]
+enum Compressions {
+    GZOnly,
+    AddXZ,
+    AddZStd,
+    AddBoth,
```
rbtcollins: AddBoth isn't very meaningful without the context of Compressions::

tests/dist.rs
```diff
@@ -496,24 +498,50 @@ fn uninstall(
     Ok(())
 }
 
+#[derive(Copy, Clone, Debug)]
+enum Compressions {
```
rbtcollins, line 500: just listing the compressions in an enum in vec![] syntax wouldn't be much longer and would allow more combinations in future. This is fine, just food for thought.

tests/dist.rs
```diff
@@ -496,24 +498,50 @@ fn uninstall(
...
+        comps.enable_xz(),
+        comps.enable_zst(),
+        f,
+    );
 }
 
 fn setup_from_dist_server(
     server: MockDistServer,
     url: &Url,
     enable_xz: bool,
```
rbtcollins: You've done this lovely enum above, and then not here.

rbtcollins (changes requested):
I think the 'prefer X under Y circumstance thing can reasonably be added later. I have some review bits that should be addressed.

gz is already implicitly optional in the current code.

src/dist/manifestation.rs
kinnison: Looks like when rust-analyzer auto-added the import it didn't group it for some reason.  Fixed.

tests/dist.rs
kinnison: Fair, changed to `AddXZAndZStd` since ultimately it's not actually used (yet)

tests/dist.rs
kinnison: Yeah, I think ultimately we'll want either a vec, or some kind of bitfield or somesuch, for now as you say, this will do.

tests/dist.rs
kinnison: That'd be because I'm a plonker.  Moved the logic in one level.  I'm not going to push it into `server.write()` for now though.  I think that refactor can wait for the vec/otherwise refactor later.

tests/dist.rs
```diff
@@ -552,7 +575,31 @@ fn setup_from_dist_server(
 
 #[test]
 fn initial_install() {
```
rbtcollins: these initial install functions are all super similar. Lets make a helper.

tests/mock/dist.rs
```diff
@@ -153,6 +155,11 @@ impl MockDistServer {
             } else {
                 None
             };
+            let zst_hash = if enable_zst {
```
rbtcollins, line 158: This to me looks like a loop trying to break free. Something for the future I guess.

tests/mock/dist.rs
```diff
@@ -356,6 +364,13 @@ impl MockDistServer {
                     );
                     toml_target.insert(String::from("xz_hash"), toml::Value::String(xz_hash));
                 }
+                if let Some(zst_hash) = hash.zst {
```
rbtcollins, line 367: Can't the dist abstraction write this out ?

tests/mock/dist.rs
```diff
@@ -466,6 +482,10 @@ fn create_tarball(relpath: &Path, src: &Path, dst: &Path) -> io::Result<()> {
             xzwriter = xz2::write::XzEncoder::new(outfile, 0);
             &mut xzwriter
         }
+        s if s.ends_with(".tar.zst") => {
```
rbtcollins, line 485: These blocks here look like an enum trying to break free

rbtcollins (approved):
More suggestions but nothing that prevents merging.

tests/dist.rs
kinnison: Agreed.  Refactored.

tests/mock/dist.rs
kinnison: Yeah, it'll be nice to make this more generic and loopy in the future.

tests/mock/dist.rs
kinnison: I'll admit I lazily extended what existed already.  We can certainly log to refactor this part of the tests.

tests/mock/dist.rs
kinnison: Yeah, I wasn't terribly happy about duplicating this code, but again I think the refactor can come later.

kinnison:
I've rebased this and added the refactor of the `initial_install*` functions.

rbtcollins (approved):

Issue #2681, opened 2021-02-25
shell restarting message can be confusing


Gurkenglas:
**Describe the problem you are trying to solve**
rustup-init added `%USERPROFILE%\.cargo\bin` to `PATH` but said
```
To get started you need Cargo's bin directory (%USERPROFILE%\.cargo\bin) in
your PATH environment variable. Future applications will automatically
have the correct environment, but you may need to restart your current shell.
```
at the end, so I unnecessarily went to add it.

**Describe the solution you'd like**
That message should not appear if `%USERPROFILE%\.cargo\bin` is in `PATH` by the end.

rbtcollins:
I'm not sure how to communicate the situation better than we have.

PATH isn't just the registry setting. It is also an ephemeral value in the process environment block of each process. And the cmd prompt that that message is shown in does not have the cargo bin directory in the PATH that is in the process environment block.

Restarting that shell will pickup the value from the registry, as the message says you need to do.

Gurkenglas:
Ah, I see. I thought the sentence described the situation as it would be after I finish adding `%USERPROFILE%\.cargo\bin` to the registry setting.

`To get started you may need to restart your current shell. This would reload its PATH environment variable to include Cargo's bin directory (%USERPROFILE%\.cargo\bin).`

rbtcollins:
I think thats a good improvement. Would you like to put a patch forward?

0xPoe:
@rbtcollins This issue seems to have been fixed and we should be able to close it now?

kinnison:
Indeed, thank you @hi-rustin

This ought to be in the next release.

PR #2684, opened 2021-02-28
Clarify shell restarting message

Gurkenglas:
Addresses https://github.com/rust-lang/rustup/issues/2681.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -162,9 +162,9 @@ macro_rules! post_install_msg_unix {
     () => {
         r"# Rust is installed now. Great!
 
-To get started you need Cargo's bin directory ({cargo_home}/bin) in your `PATH`
-environment variable. Next time you log in this will be done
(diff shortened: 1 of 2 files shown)


src/cli/self_update.rs
```diff
@@ -162,9 +162,9 @@ macro_rules! post_install_msg_unix {
     () => {
         r"# Rust is installed now. Great!
 
-To get started you need Cargo's bin directory ({cargo_home}/bin) in your `PATH`
-environment variable. Next time you log in this will be done
-automatically.
+To get started you may need to log out and back in.
```
rbtcollins: Perhaps.  ...out and back in, or start a new shell. (We've changed what we do on unix since these message was written, and any new shell will now affect PATH.

rbtcollins (changes requested):
One small tweak requested :)

rbtcollins (approved):

PR #2695, opened 2021-03-15
bug-2672

rbtcollins:
We need a two-step process for this. Step 1 is to announce the
deprecation of no-force options on update/install of toolchains and
accept --force. Step 2, in the next release, is to make --force actually
required.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -316,7 +316,12 @@ pub fn cli() -> App<'static, 'static> {
                         .help("Force an update, even if some components are missing")
                         .long("force")
                         .takes_value(false),
-                ),
+                )
+                .arg(
+                    Arg::with_name("force-non-host")
+                        .help("Install toolchains that require an emulator. See https://github.com/rust-lang/rustup/wiki/Non-host-toolchains")
+                        .long("force-non-host")
+                        .takes_value(false)),
         )
         .subcommand(SubCommand::with_name("check").about("Check for updates to Rust toolchains and rustup"))
         .subcommand(
@@ -918,6 +923,7 @@ fn check_updates(cfg: &Cfg) -> Result<utils::ExitCode> {
 
 fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     let self_update = !m.is_present("no-self-update") && !self_update::NEVER_SELF_UPDATE;
+    let forced = m.is_present("force-non-host");
     if let Some(p) = m.value_of("profile") {
         let p = Profile::from_str(p)?;
         cfg.set_profile_override(p);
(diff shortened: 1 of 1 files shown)


kinnison:
Is your hope that the presence of the `error:` will trigger some CIs to fail, causing people to update sooner?

I think we need to ponder the wording carefully, as well as the timing of requiring `--force` but the concept is good.  I don't personally have a problem with the wording you chose, but I'm aware that in the past my judgement on these kinds of things isn't great. I wonder if we could get some help from the compiler's error messages team?

rbtcollins:
The error won't cause a exit code of 1, but it will pollute stderr. I hope folk will notice it and update. I intend to file a bug on cross and get them to update immediately, so that they are ready for our next release where we conceptually do the actual enforcement.

kinnison:
A lot of people shouted at me that I'd break CI systems for emitting a deprecation error (but not failing to run) because of systems scanning stderr for 'error:' so I guess they'd notice pretty quickly if that's the case.  Honestly this is such a corner case of use, providing `cross` works properly that I don't mind.

I wonder if we should have a specific flag for this though, since `--force` implies more than just this particular override.  Again I don't mind if you think we should keep things simple.

rbtcollins:
Talking to people about the error message sounds like a fine idea; perhaps we can ask at the wg-sync? I can add a -not-host suffix to this flag easily, and its nice future proofing, so sure lets do that.

src/cli/rustup_mode.rs
```diff
@@ -939,7 +940,8 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
                 let host_arch = TargetTriple::from_host_or_build();
                 if let Ok(toolchain_desc) = ToolchainDesc::from_str(name) {
                     let target_triple = toolchain_desc.target;
-                    if host_arch.ne(&target_triple) {
+                    if !forced && host_arch.ne(&target_triple) {
+                        err!("DEPRECATED: please use --force to permit this in future releases of rustup.");
```
jyn514: Maybe make it clear that the behavior is allowed currently? But the current text seems fine too.
```suggestion
                        err!("DEPRECATED: future versions of rustup will require --force to install a non-host toolchain as the default");
```

src/cli/rustup_mode.rs
```diff
@@ -918,6 +918,7 @@ fn check_updates(cfg: &Cfg) -> Result<utils::ExitCode> {
 
 fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     let self_update = !m.is_present("no-self-update") && !self_update::NEVER_SELF_UPDATE;
+    let forced = m.is_present("force-non-host");
```
kinnison, line 926: This will need a commensurate change in the Clap definitions earlier in the file, and in the error mesage below.

kinnison (approved):

PR #2698, opened 2021-03-21
Docs: indicate the default settings.toml location

marcospb19:
After having trouble finding the `settings.toml` file in my system, I thought this could be useful as there is no obvious indication that the `${RUSTUP_HOME}` is by default `~/.rustup`, maybe that should've been pointed out too?

I'm a non-native english speaker, so any corrections would be appreciated.

### DIFF
--- a/doc/src/configuration.md
+++ b/doc/src/configuration.md
@@ -1,9 +1,9 @@
 # Configuration
 
-Rustup has a settings file in [TOML](https://github.com/toml-lang/toml) format
-at `${RUSTUP_HOME}/settings.toml`. The schema for this file is not part of the
-public interface for rustup - the rustup CLI should be used to query and set
-settings.
+Rustup has a [TOML](https://github.com/toml-lang/toml) settings file at
+`${RUSTUP_HOME}/settings.toml` (which defaults to `~/.rustup` or
+`%USERPROFILE%/.rustup`). The schema for this file is not part of the public
+interface for rustup - the rustup CLI should be used to query and set settings.
 
 On Unix operating systems a fallback settings file is consulted for some
 settings. This fallback file is located at `/etc/rustup/settings.toml` and


doc/src/configuration.md
```diff
@@ -1,10 +1,10 @@
 # Configuration
 
-Rustup has a settings file in [TOML](https://github.com/toml-lang/toml) format
-at `${RUSTUP_HOME}/settings.toml`. The schema for this file is not part of the
-public interface for rustup - the rustup CLI should be used to query and set
-settings.
+Rustup has a [TOML](https://github.com/toml-lang/toml) settings file at
+`${RUSTUP_HOME}/settings.toml`. The schema for this file is not part of the
+public interface for rustup - the rustup CLI should be preferred to query and
+set settings.
```
marcospb19: These changes weren't the PR target, but I took the chance to try to improve it.

rbtcollins: I like the rewording of the start of the paragraph. However the second part of the change, from 'should be used' to 'preferred' weakens the meaning. We mean 'do not edit this file directly' and 'do not query this file or parse it directly, because we may change its contents, format or location at any time.'

doc/src/configuration.md
```diff
@@ -1,10 +1,10 @@
...
 
-On Unix operating systems a fallback settings file is consulted for some
-settings. This fallback file is located at `/etc/rustup/settings.toml` and
-currently can define only `default_toolchain`.
+On Unix systems, by default, it is located at `${HOME}/.rustup/settings.toml`,
+and a fallback settings file at `/etc/rustup/settings.toml` is available, but
+can currently only define `default_toolchain`.
```
rbtcollins: This change makes the meaning less clear I feel. There are two meanings included in this paragraph now:

- That there is a fallback settings file at a certain location.
- That the location for the non-public settings file is is at a particular location.

I am not keen on documenting the location of the non-public settings file with any more clarity. We don't want people accessing it except when debugging a problem the CLI cannot cope with. If we really want to add this exposition, please keep the explanation in the paragraph about the non-public settings file, rather than in the paragraph about the public machine-wide settings file.

doc/src/configuration.md
marcospb19: I do agree, I will revert the changes at the second part :+1: .

doc/src/configuration.md
marcospb19: I'm not aware of that "non-public" means in your message, I don't know much about this file because this `configuration` section isn't very informative at the moment.

I do think that omitting the location is bad for users, and would prefer to be told the location, plus a "don't change it" advice.

I also like to know where `rustup` is keeping it's stuff so I don't accidentally mess it up. And I did not found the `.rustup` information anywhere, so that worried me, maybe if I did a better search, but it makes sense to be present at the configurations section.

marcospb19: That said, you recommend I revert all the changes in the second paragraph and add the location piece of information to the first one?

rbtcollins: yes

marcospb19:
@rbtcollins thanks for your reviews, I have re-pushed with the corrections.

doc/src/configuration.md
```diff
@@ -1,10 +1,10 @@
 # Configuration
 
-Rustup has a settings file in [TOML](https://github.com/toml-lang/toml) format
-at `${RUSTUP_HOME}/settings.toml`. The schema for this file is not part of the
-public interface for rustup - the rustup CLI should be used to query and set
-settings.
+Rustup has a [TOML](https://github.com/toml-lang/toml) settings file at
+`${RUSTUP_HOME}/settings.toml` (which defaults to `~/.rustup` or
```
0xPoe, line 4: Do we need to specify which OS it is for?

marcospb19: I copied from [the Environment Variables section](https://rust-lang.github.io/rustup/environment-variables.html):
image

The OS isn't specified, so I assumed it was ok.

doc/src/configuration.md
```diff
@@ -1,10 +1,10 @@
...
+`%USERPROFILE%/.rustup`). The schema for this file is not part of the public
+interface for rustup - the rustup CLI should be used to query and set settings.
 
 On Unix operating systems a fallback settings file is consulted for some
 settings. This fallback file is located at `/etc/rustup/settings.toml` and
-currently can define only `default_toolchain`.
+currently can only define `default_toolchain`.
```
rbtcollins: This is an incorrect change: https://dictionary.cambridge.org/grammar/british-grammar/only 
define is a verb, the section on focus and `only` as an adverb applies. The focus here is not the subject (`default_toolchain`), but the ability to define, so putting only in the mid position removes the focus from the subject.

marcospb19: Thanks for the correction! I'll fix it.
EDIT: fixed, marking this conversation as resolved.

PR #2710, opened 2021-04-05
show active-toolchain with rustc info

### LINKED ISSUE #2685: rustup show active-toolchain could be more detailed
Right now `rustup show active-toolchain` shows this on my system:
````
````

It would be nice to not only know the active channel but perhaps also show the actual compiler version and the corresponding hash? (perhaps through a `--verbose` flag?)
````
nightly-x86_64-unknown-linux-gnu (default) - 1.52.0-nightly (94736c434 2021-02-27)
````

### PR DESCRIPTION
**Pull request by 0xPoe**
close https://github.com/rust-lang/rustup/issues/2685

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1197,9 +1197,9 @@ fn show_active_toolchain(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCod
         Err(e) => return Err(e.into()),
         Ok((toolchain, reason)) => {
             if let Some(reason) = reason {
-                writeln!(process().stdout(), "{} ({})", toolchain.name(), reason,)?;
+                writeln!(process().stdout(), "{} ({})", toolchain.name(), reason)?;
             } else {
-                writeln!(process().stdout(), "{} (default)", toolchain.name(),)?;
+                writeln!(process().stdout(), "{} (default)", toolchain.name())?;
             }
             if verbose {
                 writeln!(process().stdout(), "{}", toolchain.rustc_version())?;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by 0xPoe** [comment]
cc:  @kinnison 

Could you please take a look?

**Comment by 0xPoe** [comment]
@kinnison Ping~

If I disturb you, please let me know. I will stop pinging you and wait for your review.

**Comment by kinnison** [comment]
Could you please add a comment here showing what a **full** output of `rustup show` displays (before/after your PR if it has changed) and what `rustup show --verbose` shows?

I'm not entirely sure it's worth adding `--verbose` rather than just always showing the information so I'd like to see.

**Comment by 0xPoe** [comment]
Without my change:
```sh
```

with my change but without verbose:
```sh
➜  rustup git:(rustin-patch-show) ./home/bin/rustup  show active-toolchain                                     
➜  rustup git:(rustin-patch-show) 
```

with my change and verbose:
```sh
➜  rustup git:(rustin-patch-show) ./home/bin/rustup  show active-toolchain --verbose
stable-x86_64-apple-darwin (default) - rustc 1.51.0 (2fd73fabe 2021-03-23)
➜  rustup git:(rustin-patch-show) 
```

0xPoe:
@kinnison ping~

rbtcollins:
I think showing it in the same format that `rustup show` does would be better: one per line. Concretely this is better for folk using sed to query etc.

0xPoe:
> I think showing it in the same format that `rustup show` does would be better: one per line. Concretely this is better for folk using sed to query etc.

Fixed.

kinnison (approved):

PR #2715, opened 2021-04-12
docs, comments: Editing and style cleanup

### LINKED ISSUE #2693: Clarify GNU vs MSVC for Windows

[The book says](https://github.com/rust-lang/rustup/blob/master/doc/src/installation/other.md):

> On Windows, download and run the [rustup-init.exe built for `i686-pc-windows-gnu` target][setup]. In general, this is the build of `rustup` one should install on Windows. Despite being built against the GNU toolchain, _the Windows build of `rustup` will install Rust for the MSVC toolchain if it detects that MSVC is installed_.

However, the [official Rust download page](https://www.rust-lang.org/tools/install) links to the MSVC version of `rustup-init`, going against the recommendation from the book.

Additionally, I ran the GNU version of `rustup-init` in a clean `mcr.microsoft.com/windows:1809` Docker container, which does not have MSVC, as confirmed by the fact that `cargo build` tells me I am missing the linker, and yet I still get an MSVC toolchain by default.

So it seems that the book is wrong about the MSVC detection, or it does not work reliably.

To me, it seems the correct solution would be to recommend the MSVC version of `rustup-init`, but in any case the recommendation from the book and official download link should probably be aligned.

**Notes**

See above.

workingjubilee:
Closes #2693.
Stylistic choices for the "Rustup" name based on remarks in #2670. I have followed, where possible, the intuition that the binary itself is named `rustup` (or `rustup.exe`) and that the _notional_ software is named Rustup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# rustup: the Rust toolchain installer
+# Rustup: the Rust toolchain installer
 
 | Master CI    | Build Status                                             |
 |--------------|----------------------------------------------------------|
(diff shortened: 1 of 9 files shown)


kinnison (approved):

kinnison:
Thank you.  I am awful at remaining consistent, so this kind of cleanup is very much appreciated.

PR #2716, opened 2021-04-12
report error when rustup is invoked with unsupported arg0

### LINKED ISSUE #2286: An error should be reported when rustup is invoked with unsupported arg0
**Problem**
In a windows docker image I had `C:/rustup-init.exe` with a Unicode hyphen which [caused the executable name match](https://github.com/rust-lang/rustup/blob/master/src/cli/main.rs#L80) to always default to `Some(_)` whereas `n.starts_with("rustup-init")` should be true with the right encoding.

**Steps**
1. replace the hyphen in `rustup-init.exe` with a [unicode hyphen](https://www.fileformat.info/info/unicode/char/2010/index.htm)
2. `rustup-init.exe --help`
3. an error is shown.
    In my docker image, I had "error: no default toolchain configured"

**Possible Solution(s)**
Encode the `OsStr` with UTF-8 encoding. 

**Notes**
Ended up compiling and testing this bug with the `1.21.1` tag

0xPoe:
close https://github.com/rust-lang/rustup/issues/2286

### DIFF
--- a/src/bin/rustup-init.rs
+++ b/src/bin/rustup-init.rs
@@ -29,6 +29,7 @@ use rustup::cli::setup_mode;
 use rustup::currentprocess::{process, with, OSProcess};
 use rustup::env_var::RUST_RECURSION_COUNT_MAX;
 use rustup::utils::utils;
+use rustup::TOOLS;
 
 fn main() {
     let process = OSProcess::default();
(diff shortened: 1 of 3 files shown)


kinnison:
This patch will prevent people from experimenting with other proxy names in custom toolchains.  I'm not sure if that's a bad thing or not since such custom proxies wouldn't be updated by a `rustup self update`.  @rbtcollins are you okay with this (minor) change in API ?  If so, then I'm good to have this merged.

rbtcollins:
custom proxies aren't a use case we have really supported; I think we could start supporting that, but lack of support shouldn't stop us solving concrete issues like this.

PR #2721, opened 2021-04-14
anyhow instead of error_chain

rbtcollins:
Drop error_chain which is no longer supported, evolving, causes code deprecation warnings, forces everything into one surfaced massive enum and generally is breaking our brains. Use anyhow! for most errors, and ThisError for errors used multiple times or for which we do need to match on.


kinnison:
I have started to review this but I think there's something squiffy about the commit series because the delta github shows me includes things definitely already merged to master.

Could you please rebase this and clean up the commit series to just be your changes?

tests/cli-v2.rs
```diff
@@ -804,7 +804,7 @@ fn add_target_v1_toolchain() {
                 clitools::CROSS_ARCH1,
                 "--toolchain=nightly",
             ],
-            for_host!("toolchain 'nightly-{0}' does not support components"),
+            for_host!("Missing manifest in toolchain 'nightly-{0}'"),
```
kinnison, line 807: This error message change seems bogus - There is a manifest, it's just a V1 manifest which doesn't support the component concept, no?

src/utils/utils.rs
```diff
@@ -522,11 +510,12 @@ fn dot_dir(name: &str) -> Option<PathBuf> {
 }
 
 pub fn rustup_home_in_user_dir() -> Result<PathBuf> {
-    dot_dir(".rustup").ok_or_else(|| ErrorKind::RustupHome.into())
+    // XXX: This error message seems wrong/bogus.
+    dot_dir(".rustup").ok_or_else(|| anyhow::anyhow!("couldn't find value of RUSTUP_HOME"))
```
kinnison, line 514: I suppose the question here is how could `home_dir()` fail since that's the problem here.  Perhaps:
```suggestion
    dot_dir(".rustup").ok_or_else(|| anyhow::anyhow!("couldn't find home directory to construct RUSTUP_HOME"))
```
?

kinnison (changes requested):
There are some inconsistencies in use of `downcast()` vs `downcast_ref()` in a few places, and a few extra `.clone()`s which might be avoidable (though they're in test paths)

I've highlighted one or two things, but otherwise I'm up for merging this modulo you having a final pass for happiness and tweaking the few items I mention.

rbtcollins:
downcast vs downcast_ref - depends on whether the error is being rethrown or consumed generally: I wouldn't call it inconsistent so much as having different contexts. Of course, spread over a year some inconsistency can always occur

src/utils/utils.rs
rbtcollins: RUSTUP_HOME might be unset
HOME might be unset
the users SID might not have a home dir set in the registry properly
....

rbtcollins: the patch just inlines the message for now, and the XXX is me whinging about our qualitay.

kinnison: Okay, this can sit out for now.

tests/cli-v2.rs
rbtcollins: hmm, lets see https://github.com/rust-lang/rustup/pull/2721/files/8add44771eccee37e53964dcd9fa8d8f2a3d812c#diff-1e51556c6aa8a6243798d60243781b6acabd184d6d26c6c5f484eefd7b9e49c2R923 is the key bit: so this is (load it up fully) - 
```
remove_component()
 if let Some(manifest) = manifestation.load_manifest()? {
   ...} else {
     -> change here
     Err MissingManifest
   -> .change finishes
 ... }

I don't recall what led to this delta. But its clearly a manifest that doesn't load from manifestation.load_manifest(), which doesn't know how to load V1 manifests.

I don't think this code is wrong with this change, but I agree its not related to anyhow. Want it backed out?

kinnison: I'm not terribly worried since V1 manifests are only for the very oldest toolchains anyway.  let's not back it out for now.

kinnison (approved):

kinnison:
If the light is green, the trap is clean!

PR #2727, opened 2021-04-19
Replace & add tags for product names

dingyx99:
This pull request only contains changes about `rustup website`.
Using `<em></em>` for product names is not capable, according to [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em#usage_notes). This may cause some translator (like Microsoft Translator) to translate product names.
BTW, Before I commit, I found that even if I change tags, MS Translator still translate names lol, so this pr is only for make this webpage much more correct.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -20,9 +20,9 @@
 </a>
 
 <p id="pitch">
-  <em>rustup</em> is an installer for<br/>
+  <i>rustup</i> is an installer for<br/>
   the systems programming language
-  <a href="https://www.rust-lang.org">Rust</a>
+  <a href="https://www.rust-lang.org"><i>Rust</i></a>
 </p>
 
 <div id="platform-instructions-unix" class="instructions display-none">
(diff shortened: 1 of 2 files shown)


kinnison:
I don't like that this changes from "emphasis" to "italics" and then uses CSS to remove the slant.  Are you sure this is necessary to improve translation behaviour?

rbtcollins:
We should probably use `<cite>` per those docs.

dingyx99:
> I don't like that this changes from "emphasis" to "italics" and then uses CSS to remove the slant. Are you sure this is necessary to improve translation behaviour?

According to MDN docs, it **should** have improvements. However, after I opened this PR, I have a try at local, still with Microsoft Translator, seemed to be no effect. But Google Translate worked fine, even without this PR.
Maybe there's something wrong with MS Translate. And I'll close this PR asap.

PR #1071, opened 2017-04-16
Enforce timeouts for reading rustc version.

lifthrasiir:
Some rustc versions (1.3.0 through 1.10.0 if my analysis is correct) are no longer working in recent macOS Sierra, in that they do not produce a correct output nor do not properly terminate in a timely
manner (e.g. loops indefinitely).

Having them in the current toolchain makes `rustup update` or (sometimes) `rustup show` unresponsive, so a timeout of one second has been added to correctly show that they are no longer working.

Fixes #766.

PR #2731, opened 2021-04-23
Increase timeout for toolchain commands for Big Surf

vors:
I noticed yesterday that running `rustup show` still sometimes times out on MacOS Big Surf. Bumping timeout one more time.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -245,11 +245,15 @@ impl<'a> Toolchain<'a> {
 
                 // some toolchains are faulty with some combinations of platforms and
                 // may fail to launch but also to timely terminate.
-                // (known cases include Rust 1.3.0 through 1.10.0 in recent macOS Sierra.)
+                // (
+                //   known cases include Rust versions
+                //     - 1.3.0 through 1.10.0 in recent macOS Sierra.
+                //     - 1.10.0 through 1.51.0 in macOS Big Surf.
+                // )
                 // we guard against such cases by enforcing a reasonable timeout to read.
                 let mut line1 = None;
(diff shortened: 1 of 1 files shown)


vors:
This is continuation of work from #1071

vors:
Perhaps it's time to root-cause the unresponsiveness, instead of bumping the timeout one more time. Since restart seems to fix the issue. But we may need help from Apple friends for that.

src/toolchain.rs
```diff
@@ -245,11 +245,15 @@ impl<'a> Toolchain<'a> {
...
+                // (
+                //   known cases include Rust versions
+                //     - 1.3.0 through 1.10.0 in recent macOS Sierra.
+                //     - 1.10.0 through 1.51.0 in macOS Big Surf.
+                // )
                 // we guard against such cases by enforcing a reasonable timeout to read.
                 let mut line1 = None;
                 if let Ok(mut child) = cmd.spawn() {
-                    let timeout = Duration::new(10, 0);
+                    let timeout = Duration::new(24, 0);
```
workingjubilee, line 256: > "a reasonable timeout to read"

> 24 seconds

I believe it would be appropriate to:
- actually solve the problem
XOR
- remove the description "reasonable", which was added when this was **1** second.

workingjubilee: If you go with the latter, I suggest using `Duration::from_secs` to make it explicit what the current state of affairs is at a brief glance.

workingjubilee:
What actual hardware is this problem being observed on? x86_64 Macs? aarch64? PowerPCs?

vors:
> actually solve the problem

Could not agree more

PR #2732, opened 2021-04-25
Give a better error when std is missing for a target

jyn514:
- Note that not all platforms have std pre-compiled
- Suggest `cargo build -Zbuild-std` on nightly

After https://github.com/rust-lang/rust/pull/84450, you can end up with a situation like this:
```
$ rustc +stage1 src/main.rs --target x86_64-unknown-uefi
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-uefi` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-uefi`

For more information about this error, try `rustc --explain E0463`.

$ rustup target add x86_64-unknown-uefi

error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'x86_64-unknown-uefi'
```

That doesn't give you much info on how to proceed. After this PR, it will be more clear:

```
$ rustup target add x86_64-unknown-uefi

error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'x86_64-unknown-uefi'
```

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -351,11 +351,16 @@ error_chain! {
         }
         UnknownComponent(t: String, c: String, s: Option<String>) {
             description("toolchain does not contain component")
-            display("toolchain '{}' does not contain component {}{}", t, c, if let Some(suggestion) = s {
-                format!("; did you mean '{}'?", suggestion)
-            } else {
-                "".to_string()
-            })
+            display("toolchain '{}' does not contain component {}{}{}", t, c, if let Some(suggestion) = s {
+                    format!("; did you mean '{}'?", suggestion)
+                } else {
+                    "".to_string()
+                }, if c.contains("rust-std") {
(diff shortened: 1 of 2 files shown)


jyn514:
Unlike https://github.com/rust-lang/rust/pull/84450, this recommends `-Z build-std` whenever you're using a nightly toolchain, because you can always install cargo with rustup and it seems strange to be using rustup but not cargo.

rbtcollins:
rustup and no cargo will occur when someone builds a rust toolchain from source and links it in. However I agree that this is the common case suggestion to make, and I think anyone using such a non-downloadable toolchain will be sophisticated enough to figure it out with the pointer this gives them. @kinnison thoughts?

kinnison (approved):

Issue #2730, opened 2021-04-23
Detect dead code better


rbtcollins:
We should stop using 'pub' and instead use 'crate' within rustup. Rustup has no exported API, but having pub functions results in dead types and functions not being detected. https://github.com/rust-lang/rust/pull/79061 is an example of this technique applied to rustdoc

0xPoe:
I am working on this.

0xPoe:
- [ ] After most of the cleanup is done, we should globally introduce #! [deny(unreachable_pub)].

0xPoe:
I think we can close it now. It's done.

PR #2735, opened 2021-04-26
Make all tests functions and structs crate-private

0xPoe:
part of https://github.com/rust-lang/rustup/issues/2730

### DIFF
--- a/tests/mock/clitools.rs
+++ b/tests/mock/clitools.rs
@@ -28,31 +28,31 @@ use crate::mock::topical_doc_data;
 use crate::mock::{MockComponentBuilder, MockFile, MockInstallerBuilder};
 
 /// The configuration used by the tests in this module
-pub struct Config {
+pub(crate) struct Config {
     /// Where we put the rustup / rustc / cargo bins
-    pub exedir: PathBuf,
+    pub(crate) exedir: PathBuf,
     /// The distribution server
-    pub distdir: PathBuf,
+    pub(crate) distdir: PathBuf,
     /// RUSTUP_HOME
-    pub rustupdir: rustup_test::RustupHome,
+    pub(crate) rustupdir: rustup_test::RustupHome,
     /// Custom toolchains
-    pub customdir: PathBuf,
+    pub(crate) customdir: PathBuf,
     /// CARGO_HOME
-    pub cargodir: PathBuf,
+    pub(crate) cargodir: PathBuf,
     /// ~
-    pub homedir: PathBuf,
+    pub(crate) homedir: PathBuf,
     /// An empty directory. Tests should not write to this.
-    pub emptydir: PathBuf,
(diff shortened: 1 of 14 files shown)


0xPoe:
I found a lot of methods and variables that are not being used, should I remove them all?

cc: @kinnison @rbtcollins

0xPoe:
/hold

Waiting for anyhow wroks laned.

0xPoe:
/unhold

I think it's ready for review.

kinnison:
I'm not entirely sure if this is valuable here.  The way our tests are structured makes it quite hard to know for sure about which bits are used in which test files.  Also `pub` of any kind inside `test-$foo.rs` seems a little odd to me anyway.

I'm going to defer to @rbtcollins for this particular PR.

0xPoe:
Because our test is a standalone crate, it warns on cargo check, but not on cargo clippy.

I don't have too good a solution for it.

kinnison:
I wonder if, instead, what we need is to restructure our tests/ dir into tests/dist.rs and tests/cli.rs with clitools/ and clitests/ as subdirs - i.e. build the dist tests (which import the rustup module) and the cli tests as two integration binaries rather than the nearly 10 we have now.  Then the clitools stuff becomes a module used by only one binary (the cli tests) and as such controlling unused functions etc. becomes much more plausible.  If we did that though, we'd need to be sure that we'd bottomed the potentially dodgy parallelism in some of the tests.

tests/mock/clitools.rs
```diff
@@ -28,31 +28,31 @@ use crate::mock::topical_doc_data;
 use crate::mock::{MockComponentBuilder, MockFile, MockInstallerBuilder};
 
 /// The configuration used by the tests in this module
-pub struct Config {
+pub(crate) struct Config {
     /// Where we put the rustup / rustc / cargo bins
-    pub exedir: PathBuf,
+    pub(crate) exedir: PathBuf,
```
rbtcollins, line 33: we shouldn't do this - just the pub(crate) on the struct, per our discussions and that bug.

0xPoe: No, if we remove the modifier, it will be private. The bug fixed is the use of pub in pub(crate).

0xPoe: Could this be another bug?

rbtcollins: I'm not suggesting private, I'm suggesting simple `pub`

0xPoe: Oh, you mean in this case we don't need to change it to pub(crate) but keep it pub?

rbtcollins:
Exactly

0xPoe:
Close it for now, as it causes cargo check to misrepresent a lot of unused warnings. It can be very disruptive until we can sort out and re-architect the tests.

PR #2744, opened 2021-05-01
Make clippy happy

0xPoe:


src/cli/rustup_mode.rs
```diff
@@ -863,7 +863,7 @@ fn default_(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     } else {
         let default_toolchain: Result<String> = cfg
             .get_default()?
-            .ok_or(anyhow!("no default toolchain configured"));
```
rbtcollins, line 866: Why doesn't it like this one ?

0xPoe: https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call

```rust
    /// Arguments passed to `ok_or` are eagerly evaluated; if you are passing the
    /// result of a function call, it is recommended to use [`ok_or_else`], which is
    /// lazily evaluated.
```

rbtcollins: hmm, I had figured it was cheap enough since its a static string etc, but nothing we're doing should be a hotspot here, so shrug.

Issue #2717, opened 2021-04-12
Installer gets killed during installation


kinnison:
Please I am unable to install rust 1.51.0 on Ubuntu 20.10 (groovy).
What could be wrong?

```
eshikafe@eshikafe-vb:~$ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
info: downloading installer

Welcome to Rust!

This will download and install the official compiler for the Rust
programming language, and its package manager, Cargo.

Rustup metadata and toolchains will be installed into the Rustup
home directory, located at:

  /home/eshikafe/.rustup

This can be modified with the RUSTUP_HOME environment variable.

The Cargo home directory located at:

  /home/eshikafe/.cargo

This can be modified with the CARGO_HOME environment variable.

The cargo, rustc, rustup and other commands will be added to
Cargo's bin directory, located at:

  /home/eshikafe/.cargo/bin

This path will then be added to your PATH environment variable by
modifying the profile files located at:

  /home/eshikafe/.profile
  /home/eshikafe/.bashrc

You can uninstall at any time with rustup self uninstall and
these changes will be reverted.

Current installation options:


   default host triple: x86_64-unknown-linux-gnu
     default toolchain: stable (default)
               profile: default
  modify PATH variable: yes

1) Proceed with installation (default)
2) Customize installation
3) Cancel installation
>1

info: profile set to 'default'
info: default host triple is x86_64-unknown-linux-gnu
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: latest update on 2021-03-25, rust version 1.51.0 (2fd73fabe 2021-03-23)
info: downloading component 'cargo'
  6.0 MiB /   6.0 MiB (100 %)   3.4 MiB/s in  2s ETA:  0s
info: downloading component 'clippy'
info: downloading component 'rust-docs'
 14.9 MiB /  14.9 MiB (100 %)   1.4 MiB/s in 17s ETA:  0s
info: downloading component 'rust-std'
 24.9 MiB /  24.9 MiB (100 %)   1.3 MiB/s in 23s ETA:  0s 
info: downloading component 'rustc'
 50.4 MiB /  50.4 MiB (100 %)   3.2 MiB/s in 19s ETA:  0s
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: using up to 500.0 MiB of RAM to unpack components
info: installing component 'clippy'
info: installing component 'rust-docs'
[...]

PR #2746, opened 2021-05-03
Add link reference definitions to CHANGELOG.md

foresterre:
# Description

The changelog of rustup already contained link labels, but did not contain matching [link reference definitions](https://github.github.com/gfm/#link-reference-definition). This commit adds matching link reference definitions for all previous releases. The link reference definitions point to the diff of two releases, except for the first release in the changelog, which points to the release tag. This allows markdown viewers, to show clickable links to the diff for each release.

# Preview of changes

Previously, the link labels looked like this:

## [1.24.1] - 2021-04-27

with the markdown definition:

```md
## [1.24.1] - 2021-04-27
```

Using this commit, the link labels look like this:

## [1.24.0] - 2021-04-27

[1.24.0]: https://github.com/rust-lang/rustup/compare/1.23.1...1.24.0

with the markdown definition:

```md
## [1.24.0] - 2021-04-27

...

[1.24.0]: https://github.com/rust-lang/rustup/compare/1.23.1...1.24.0

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ part of a Rust component.
 We're sorry for pain this may have caused.
 
 [pr#2738]: https://github.com/rust-lang/rustup/pull/2738
+[1.24.1]: https://github.com/rust-lang/rustup/releases/tag/1.24.1
 
 ## [1.24.0] - 2021-04-27
 
@@ -152,6 +153,7 @@ Thanks go to:
 [pr#2604]: https://github.com/rust-lang/rustup/pull/2604
 [pr#2517]: https://github.com/rust-lang/rustup/pull/2517
 [pr#2592]: https://github.com/rust-lang/rustup/pull/2592
+[1.24.0]: https://github.com/rust-lang/rustup/releases/tag/1.24.0
 
 ## [1.23.1] - 2020-12-01
 
@@ -172,6 +174,8 @@ Thanks go to:
 - Joshua M. Clulow
 - Neil Mitchell
 - Richard Gomes
+
+[1.23.1]: https://github.com/rust-lang/rustup/releases/tag/1.23.1
 
 ## [1.23.0] - 2020-11-27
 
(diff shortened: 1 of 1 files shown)


kinnison:
I think the square brackets were stylistic rather than meant to be link points, however I'm not entirely averse to this idea.  I will note that almost always that delta will be near impossible for someone to read usefully though.  @rbtcollins If you think this is "not insane" to do then please merge.  There's no need to fire the actions since none of them consume the changelog.

rbtcollins:
Interesting idea. Uh, I think its probably unlikely to be useful for people to read such a diff - but perhaps it is for you - what prompted this patch?

Thanks

foresterre:
Sometimes I need to search for a specific change I read about in the changelog and use these (links to the) diffs to search for the specific commit in which the described change was made.

Yesterday, I went through the rustup changelog for something similar (issues #2717 and #2748) and thought, hey, the link definitions are missing, let me add those while waiting for a build 😜. I thought I recognized the format as similar to [keep a changelog](https://keepachangelog.com/en/1.0.0/).

rbtcollins:
I'm torn, I support giving folk information, but I'm not sure how useful it will be: we have lots of bugs that look similar but are distinct, and we strongly prefer that folk file new bugs unless they are hitting one of the few bugs we've escalated to a general-problem status (and pinned / gardened to make super clear).

kinnison:
> Sometimes I need to search for a specific change I read about in the changelog 

We link to the PRs whenever we talk about a change normally.  If we're failing to do that then we should fix that.

My gut feeling is that these links simply won't be useful for more than a very small number of people and will likely confuse others more.  I'd be fine with linking to the tag page instead - https://github.com/rust-lang/rustup/releases/tag/1.24.1 for example - which has a compare button if someone wishes to use it.

foresterre:
> > Sometimes I need to search for a specific change I read about in the changelog
> 
> We link to the PRs whenever we talk about a change normally. If we're failing to do that then we should fix that.

My apologies, they are there indeed, I don't understand how I missed these links. 

> I'd be fine with linking to the tag page instead - https://github.com/rust-lang/rustup/releases/tag/1.24.1 for example - which has a compare button if someone wishes to use it.

I updated the links to point to the tag release pages instead of the diffs.

kinnison:
OK cool, this looks good.  Thank you.  I shall endeavour to do similarly on future releases :D

PR #2750, opened 2021-05-04
Fixes #2748 by filling the streaming buffers fully

### LINKED ISSUE #2748: rustup 1.24.1 fails to install Rust 1.38.0 on Windows in Travis CI due to failure to allocate memory
**Problem**

`rustup` 1.24.1 fails to install Rust 1.38.0 on Windows in Travis CI due to failure to allocate memory. Interestingly installing Rust 1.51.1 with this version of `rustup` does succeed. Here's an example log:

```
$ curl -LSso /tmp/rustup.sh https://sh.rustup.rs && sh /tmp/rustup.sh -y --default-toolchain "$RUST_TOOLCHAIN" --profile minimal
info: downloading installer
info: profile set to 'minimal'
info: default host triple is x86_64-pc-windows-msvc
info: syncing channel updates for '1.38.0-x86_64-pc-windows-gnu'
info: latest update on 2019-09-26, rust version 1.38.0 (625451e37 2019-09-23)
info: downloading component 'rust-mingw'
info: using up to 500.0 MiB of RAM to unpack components
info: installing component 'rust-mingw'
memory allocation of 16777216 bytes failed
The command "curl -LSso /tmp/rustup.sh https://sh.rustup.rs && sh /tmp/rustup.sh -y --default-toolchain "$RUST_TOOLCHAIN" --profile minimal" exited with 1.
```

Installing Rust 1.38.0 in this environment works with `rustup` 1.23.1.

**Steps**

1. Attempt to install Rust 1.38.0 in a Windows Travis CI build, see https://github.com/wezm/windows-travis-rustup/blob/master/.travis.yml for a minimal reproduction.

**Possible Solution(s)**

Pinning `rustup` to 1.23.1 successfully installs Rust 1.38.0.

**Notes**

* See https://travis-ci.com/github/wezm/windows-travis-rustup/builds/224818911 for example build.
* According to [the Travis docs](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) the Windows builders have 8Gb of RAM.

Output of `rustup --version`: 
Output of `rustup show`:

rbtcollins:
Diagnosis by Daniel. Code by Robert.

### DIFF
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -364,16 +364,18 @@ fn unpack_without_first_dir<'a, R: Read>(
             continue;
         }
 
+        struct SenderEntry<'a, 'b, R: std::io::Read> {
+            sender: Box<dyn FnMut(Vec<u8>) -> bool + 'a>,
+            entry: tar::Entry<'b, R>,
+        }
+
         /// true if either no sender_entry was provided, or the incremental file
         /// has been fully dispatched.
         fn flush_ios<'a, R: std::io::Read, P: AsRef<Path>>(
             mut budget: &mut MemoryBudget,
             io_executor: &dyn Executor,
             mut directories: &mut HashMap<PathBuf, DirStatus>,
-            mut sender_entry: Option<&mut (
-                Box<dyn FnMut(Vec<u8>) -> bool + 'a>,
-                &mut tar::Entry<'_, R>,
-            )>,
+            mut sender_entry: Option<&mut SenderEntry<'a, '_, R>>,
             full_path: P,
         ) -> Result<bool> {
             let mut result = sender_entry.is_none();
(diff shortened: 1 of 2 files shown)


src/dist/component/package.rs
```diff
@@ -384,16 +386,20 @@ fn unpack_without_first_dir<'a, R: Read>(
                 trigger_children(&*io_executor, &mut directories, &mut budget, op)?;
             }
             // Maybe stream a file incrementally
-            if let Some((sender, entry)) = sender_entry.as_mut() {
+            if let Some(sender) = sender_entry.as_mut() {
                 if budget.available() as u64 >= IO_CHUNK_SIZE {
                     let mut v = vec![0; IO_CHUNK_SIZE as usize];
```
kinnison, line 391: If we're using `read_to_end` then there's little point in initialising the vec here. The `.take()` will hint the size to `read_to_end()` and we won't have to waste zero-initialising the buffer.  This may as well be `Vec::new()`
```suggestion
                    let mut v = Vec::new();
```

kinnison:
If we're making this change can we also add `drop(contents);` before the `chunk_complete_callback(len);` in `diskio::write_file_incremental()` (`mod.rs`)?

src/dist/component/package.rs
rbtcollins: It would be great if it did. But nopity nope nope nope nope. https://doc.rust-lang.org/src/std/io/mod.rs.html#355

rbtcollins:
> 
> 
> If we're making this change can we also add `drop(contents);` before the `chunk_complete_callback(len);` in `diskio::write_file_incremental()` (`mod.rs`)?

Sure. Though as discussed, don't think its material.

src/dist/component/package.rs
rbtcollins: https://doc.rust-lang.org/src/std/io/mod.rs.html#2416 sorry ... but still, note it takes the min. Not the max.

kinnison:
> Sure. Though as discussed, don't think its material.

It dropped the max-rss by not-quite-a-chunk in my basic testing

Issue #2686, opened 2021-03-01
Command for "install stuff from `rust-toolchain.toml`"


jmaargh:
Apologies if this already exists, but I couldn't find it in the docs or shell help. Or if there is an alternate workflow for this.

I love `rust-toolchain.toml`, it's a great idea. It would be even more useful if there were a command such as `rustup override install`, which would look at the current `rust-toolchain.toml` and install the toolchain and components specified there. It would be super useful both when exploring projects for the first time, but also when using disposable container-based development environments.

I'm unopinionated on the exact syntax of this command. Also happy to help implement this if the maintainers want it.

```[tasklist]
### Tasks
- [ ] https://github.com/rust-lang/rustup/issues/3972
```

rbtcollins:
See #1397

jmaargh:
I feel like there's value in keeping this issue different from #1397 for a couple of reasons:

1. The main point of #1397 is to disable a certain behaviour, this is to add that behaviour somewhere else (something that was discussed in #1397, but isn't the main point)

2. It's not clear to me from #1397 whether what's being discussed is installing _just_ the toolchain per-se, or also the components that can be set in a `rust-toolchain.toml` file. This issue is suggesting ensuring that *everything* specified in `rust-toolchain.toml` is installed and available.

kinnison:
I think that ensuring `rust-toolchain{,.toml}` is properly honoured is an all-or-nothing thing - it couldn't "just" be the toolchain itself, no matter what.

You're right that #1397 is about ensuring that we don't over-indulge in autoinstallation, and I think this is a good home for how we then permit manual triggering of what was previously automatic.

I continue to think something like `rustup toolchain ensure [PATH]` seems like a good approach, but further straw men for discussion would be good.

jmaargh:
I'm easy about the syntax. `rustup toolchain ensure [path]` works for me.

I was thinking re-using `rustup toolchain install`, so:
- `rustup toolchain install [toolchain]` works as currently
- `rustup toolchain install [path to rust-toolchain.toml]` installs according to the specified file
- `rustup toolchian install [path to directory]` finds the `rust-toolchain.toml` that would apply to that directory (if any) and asks for confirmation before installing, with the confirmation prompt specifying exactly which file and that you could have specified a toolchain instead.
- `rustup toolchain install` acts as `rustup toolchian install .`

kinnison:
Given all the other arguments `rustup toolchain install` can be given, I'd favour a separate verb.

[...]

PR #2751, opened 2021-05-04
Allow rustup-init to install the default toolchain from a toolchain file

### LINKED ISSUE #2749: rustup-init with rust-toolchain
With the introduction of the TOML version of `rust-toolchain`, there isn't an easy way to install the exact version and components without also first installing the default toolchain. Where as with the old one line version you could write `rustup-init --default-toolchain $(cat rust-toolchain)`.

`rustup-init --from-file ./rust-toolchain` or something similar.

foresterre:
This PR allows rustup-init to install a default toolchain from a specified toolchain file.

In the current draft PR:
1) The toolchain file is only checked for the [channel](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) field, which to the best of my knowledge corresponds to the `toolchain` field. I could also add overrides for the other fields which can be specified in the toolchain file: the `target` (in rust-toolchain.toml)/`host`(in rustup), the `components` and the `profile`.

2) I re-used the name suggested by @XAMPPRocky, `--from-file`.

3) The toolchain specified by the toolchain file is only used if the `--default-toolchain` option is not given.

3) The parse mode is determined based on the file extension, if it has a `toml` file extension, the `TomlOnly` mode is used, else, the legacy mode is available. When the toolchain file is used to override current toolchain, the toolchain file name is fixed. In the suggested option this is not necessarily the case, so I opted to take this middle ground for now.

4) I will add extra test cases in a bit (I'm a bit figuring out how the testing setup works :))

closes #2749

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -43,6 +43,7 @@ OPTIONS:
         --default-host <default-host>              Choose a default host triple
         --default-toolchain <default-toolchain>    Choose a default toolchain to install
         --default-toolchain none                   Do not install any toolchains
+        --from-file <rust-toolchain-file>          Use the default toolchain specified by the rust-toolchain file
         --profile [minimal|default|complete]       Choose a profile
     -c, --component <components>...                Component name to also install
     -t, --target <targets>...                      Target name to also install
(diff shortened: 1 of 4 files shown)


rbtcollins:
Please see #2686 : I'm glad you're putting effort into this, but the immediate effort needed is probably some design / hallway testing work.

T-256:
I think #3983 could also close this PR.

rami3l:
@T-256 Thanks for the triaging! Indeed, this has been closed by #3983.

@foresterre Many thanks nonetheless for the PR :)

PR #2752, opened 2021-05-05
Use utils::remove_file in utils::ensure_file_removed

### LINKED ISSUE #2734: latent file removal bug
I just noticed that `ensure_file_removed` doesn't have the retry logic that `remove_file` does (both in `utils.rs`: the difference between them is solely that a file-not-found error is ignored by ensure-file-removed; so we should apply the retry logic - in fact we should just layer `ensure_file_removed` on `remove_file`.

joshrotenberg:
This change calls `utils::remove_file` in `utils::ensure_file_removed` instead of calling `fs::remove_file` in order to take advantage of the retry. The logic is essentially the same: if the result is an error but the error type is `NotFound`, return `Ok(())`, otherwise return whatever the actual result is. With this change, the error type is buried a little bit deeper in the result.

In addition, some basic tests have been added to verify the functionality of both `remove_file` and `ensure_file_removed` with regards to this change.

Fixes #2734.

### DIFF
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -392,10 +392,14 @@ pub fn remove_file(name: &'static str, path: &Path) -> Result<()> {
 }
 
 pub fn ensure_file_removed(name: &'static str, path: &Path) -> Result<()> {
-    let result = fs::remove_file(path);
+    let result = remove_file(name, path);
     if let Err(err) = &result {
-        if err.kind() == io::ErrorKind::NotFound {
-            return Ok(());
+        if let Some(retry::Error::Operation { error: e, .. }) =
+            err.downcast_ref::<retry::Error<io::Error>>()
+        {
+            if e.kind() == io::ErrorKind::NotFound {
+                return Ok(());
+            }
         }
     }
     result.with_context(|| RustupError::RemovingFile {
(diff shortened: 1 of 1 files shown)


kinnison:
Is there a good reason that you didn't add those unit-like tests as unit tests in utils.rs, rather than creating an integration test file?

joshrotenberg:
> Is there a good reason that you didn't add those unit-like tests as unit tests in utils.rs, rather than creating an integration test file?

No, actually. I'll move them.

src/utils/utils.rs
```diff
@@ -756,4 +760,41 @@ mod tests {
...
+        let tempdir = tempfile::Builder::new().prefix("rustup").tempdir().unwrap();
+        let f_path = tempdir.path().join("f");
+        File::create(&f_path).unwrap();
+
+        assert!(f_path.exists());
+
+        assert!(remove_file("f", &f_path).is_ok());
+        assert!(!f_path.exists());
+
+        let result = remove_file("f", &f_path);
```
0xPoe:
```suggestion
        assert!(f_path.exists());
        assert!(remove_file("f", &f_path).is_ok());
        
        assert!(!f_path.exists());
        let result = remove_file("f", &f_path);
```

src/utils/utils.rs
```diff
@@ -756,4 +760,41 @@ mod tests {
...
+        let err = result.unwrap_err();
+
+        match err.downcast_ref::<RustupError>() {
+            Some(RustupError::RemovingFile { name, path }) => {
+                assert_eq!(*name, "f");
+                assert_eq!(path.clone(), f_path);
+            }
+            _ => panic!(),
```
0xPoe: Would it be better to add a msg?

src/utils/utils.rs
```diff
@@ -756,4 +760,41 @@ mod tests {
...
+
+    #[test]
+    fn test_ensure_file_removed() {
+        let tempdir = tempfile::Builder::new().prefix("rustup").tempdir().unwrap();
+        let f_path = tempdir.path().join("f");
+        File::create(&f_path).unwrap();
+
+        assert!(f_path.exists());
+
+        assert!(ensure_file_removed("f", &f_path).is_ok());
```
0xPoe:
```suggestion
        assert!(f_path.exists());
        assert!(ensure_file_removed("f", &f_path).is_ok());
```

0xPoe (commented):
Thanks for your contribution.
rest LGTM. It's better to squash your commits.

src/utils/utils.rs
joshrotenberg: Added a message here.

joshrotenberg:
> Thanks for your contribution.
> rest LGTM. It's better to squash your commits.

Thanks. I'll remember to squash next time.

kinnison:
I've approved the workflow to run.  If everything goes green then I'd like to see a cleaned up commit series before I do a final pass

joshrotenberg:
> I've approved the workflow to run. If everything goes green then I'd like to see a cleaned up commit series before I do a final pass

`fmt` fixed and commits squashed, but ...

>> memory allocation of 16777216 bytes failed
>> Error: Process completed with exit code 1.

I'm guessing this is a transient error, but I don't know for sure.

kinnison:
> > I've approved the workflow to run. If everything goes green then I'd like to see a cleaned up commit series before I do a final pass
> 
> `fmt` fixed and commits squashed, but ...
> 
> > > memory allocation of 16777216 bytes failed
> > > Error: Process completed with exit code 1.
> 
> I'm guessing this is a transient error, but I don't know for sure.

This is part of a problem with Rustup we're trying to bottom out on another PR, so don't worry that you might have caused it.

kinnison (approved):
This looks good to me, assuming a green CI (modulo any 1.24 related iffies) I'll merge.

rbtcollins:
@joshrotenberg Thank you for picking this bug up and fixing it ! <3

PR #2755, opened 2021-05-05
Fix #2754 by testing file content

### LINKED ISSUE #2754: Test needed - Verify streaming file installation
While preparing 1.24.2 we encountered a problem where streamed files were all NUL bytes.

This was easy to fix, but highlights the fact that our test set does not include any files large enough to trigger the streaming installation mode.

We should ensure that at least one (and at most a small number) of our tests validate that (a) the streaming mode is invoked (i.e. files larger than `IO_CHUNK_SIZE`) and (b) that the resultant installed file has not been corrupted.

rbtcollins:


kinnison:
This test validates the incremental file *writer* but not the process of constructing the chunks in `package.rs` which was actually where the bug existed.

rbtcollins:
Argh. Thanks. I don't think that's accessible enough to test short of constructing a test archive containing a streamed file.

rbtcollins:
I'm closing this - the patch can come in via the branch and the bug can stay open

PR #2756, opened 2021-05-05
Slabs

rbtcollins:
be smarter about memory management to reduce hypothetical memory fragmentation


kinnison:
Hiding the 'using 500.0MiB' message has broken a number of the 'exact' tests I think.

kinnison:
I think this is looking good.  I'd just like to know what the memory consumption difference is between 1.24.1 and this for Windows runs, because my gut feeling is that it's using more RAM under Linux, though that may simply be that it's being more efficient with the RAM I have.

kinnison:
I have asked someone on #2759 to try an artifact from this PR.

rbtcollins:
I have a number of non-rustup things to do: it will be a week or so before I can sit down and dig up a massif or similar for windows.

kinnison:
Based on the experience in #2759 I'm going to merge this so we can push out a 1.24.2 for beta again.

PR #2763, opened 2021-05-10
Support set auto-self-update mode

### LINKED ISSUE #2576: Setting to only check for a self-update during `rustup update`
When testing the staging rustup release I had it "upgrade" back to the previous version when I ran `rustup update nightly`. This made me reconsider somehow aliasing in `rustup update` to include `--no-self-update` so that I could control when rustup itself updates.

Something like `rustup set update-self-update check-only` that will check whether a self-update is available and report it, rather than installing it straight away.

**Notes**

0xPoe:
close https://github.com/rust-lang/rustup/issues/2576

1. The PR supports three different modes of auto self update: enable/disable/check-only.
 - When enabled, commands such as install or update will update rustup itself. 
 - When disable, commands such as install or update will **not** update rustup itself. Only self udpate will update rustup itself. 
 - When check-only, commands such as install or update will **not** update rustup itself but will check for updates. Only self udpate will update rustup itself. 

2. This PR removes a large number of "--no-self-update" args in tests(setup will turn off auto self update by default.) 
3. Add tests for check-only and enable. 
4. add docs

### DIFF
--- a/doc/src/basics.md
+++ b/doc/src/basics.md
@@ -30,9 +30,22 @@ info: downloading self-updates
 
 ## Keeping `rustup` up to date
 
-Running `rustup update` also checks for updates to `rustup` itself and automatically
-installs the latest version. To manually update `rustup` only,
-without updating installed toolchains, type `rustup self update`:
+If your `rustup` was built with the `no-self-update` feature, it can not update
+itself. This is not the default, and only versions of `rustup` built with
+`--no-default-features`, or obtained from a third-party distributor who has
+disabled it (such as the Ubuntu snap store).
+
+Otherwise Rustup can update itself. It is possible to control Rustup's automatic
+self update mechanism with the `auto-self-update` configuration variable. This
+setting supports three values: `enable` and `disable` and `check-only`.
+
+* `disable` will ensure that no automatic self updating actions are taken.
+* `enable` will mean that `rustup update` and similar commands will also check for, and install, any update to Rustup.
+* `check-only` will cause any automatic self update to check and report on any updates, but not to automatically install them.
+
+Whether `auto-self-update` is `enable` or not, you can request that Rustup
+update itself to the latest version of `rustup` by running `rustup self update`.
+This will not download new toolchains:
 
 ```console
 $ rustup self update
@@ -40,10 +53,10 @@ info: checking for self-updates
 info: downloading self-updates
 ```
 
-> #### Disable automatic self-updates
-> `rustup` will automatically update itself at the end of any toolchain installation.
-> You can prevent this automatic behaviour by passing the `--no-self-update` argument
-> when running `rustup update` or `rustup toolchain install`.
+### Disabling self updates on a per-invocation basis
+> Self updates can also be suppressed on individual invocations of `rustup` by
+> passing the argurment `--no-self-update`  when running `rustup update` or
+> `rustup toolchain install`.
 
 ## Help system
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -13,7 +13,11 @@ use super::self_update;
 use super::term2;
 use super::term2::Terminal;
 use super::topical_doc;
-use super::{common, self_update::get_available_rustup_version};
+use super::{
+    common,
+    self_update::{check_rustup_update, SelfUpdateMode},
+};
+use crate::cli::errors::CLIError;
 use crate::dist::dist::{
     PartialTargetTriple, PartialToolchainDesc, Profile, TargetTriple, ToolchainDesc,
 };
@@ -185,6 +189,7 @@ pub fn main() -> Result<utils::ExitCode> {
         ("set", Some(c)) => match c.subcommand() {
             ("default-host", Some(m)) => set_default_host_triple(cfg, m)?,
             ("profile", Some(m)) => set_profile(cfg, m)?,
+            ("auto-self-update", Some(m)) => set_auto_self_update(cfg, m)?,
             (_, _) => unreachable!(),
         },
         ("completions", Some(c)) => {
@@ -707,6 +712,16 @@ pub fn cli() -> App<'static, 'static> {
                                 .possible_values(Profile::names())
                                 .default_value(Profile::default_name()),
                         ),
+                )
+                .subcommand(
+                    SubCommand::with_name("auto-self-update")
+                        .about("The rustup auto self update mode")
+                        .arg(
+                            Arg::with_name("auto-self-update-mode")
+                                .required(true)
+                                .possible_values(SelfUpdateMode::modes())
+                                .default_value(SelfUpdateMode::default_mode()),
+                        ),
                 ),
         );
 
@@ -911,31 +926,20 @@ fn check_updates(cfg: &Cfg) -> Result<utils::ExitCode> {
         }
     }
 
-    // Get current rustup version
-    let current_version = env!("CARGO_PKG_VERSION");
+    check_rustup_update()?;
 
-    // Get available rustup version
-    let available_version = get_available_rustup_version()?;
-
-    let _ = t.attr(term2::Attr::Bold);
-    write!(t, "rustup - ")?;
-
-    if current_version != available_version {
-        let _ = t.fg(term2::color::YELLOW);
-        write!(t, "Update available")?;
-        let _ = t.reset();
-        writeln!(t, " : {} -> {}", current_version, available_version)?;
-    } else {
-        let _ = t.fg(term2::color::GREEN);
-        write!(t, "Up to date")?;
-        let _ = t.reset();
-        writeln!(t, " : {}", current_version)?;
-    }
     Ok(utils::ExitCode(0))
 }
 
 fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    let self_update = !m.is_present("no-self-update") && !self_update::NEVER_SELF_UPDATE;
+    let self_update_mode = cfg.get_self_update_mode()?;
+    // Priority: no-self-update feature > self_update_mode > no-self-update args.
+    // Update only if rustup does **not** have the no-self-update feature,
+    // and auto-self-update is configured to **enable**
+    // and has **no** no-self-update parameter.
+    let self_update = !self_update::NEVER_SELF_UPDATE
+        && self_update_mode == SelfUpdateMode::Enable
+        && !m.is_present("no-self-update");
     let forced = m.is_present("force-non-host");
     if let Some(p) = m.value_of("profile") {
         let p = Profile::from_str(p)?;
@@ -1019,6 +1023,10 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
         info!("cleaning up downloads & tmp directories");
         utils::delete_dir_contents(&cfg.download_dir);
         cfg.temp_cfg.clean();
+    }
+
+    if !self_update::NEVER_SELF_UPDATE && self_update_mode == SelfUpdateMode::CheckOnly {
+        check_rustup_update()?;
     }
 
     Ok(utils::ExitCode(0))
@@ -1575,6 +1583,20 @@ fn set_default_host_triple(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitC
 
 fn set_profile(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     cfg.set_profile(&m.value_of("profile-name").unwrap())?;
+    Ok(utils::ExitCode(0))
+}
+
+fn set_auto_self_update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
+    if self_update::NEVER_SELF_UPDATE {
+        let mut args = crate::process().args_os();
+        let arg0 = args.next().map(PathBuf::from);
+        let arg0 = arg0
+            .as_ref()
+            .and_then(|a| a.to_str())
+            .ok_or(CLIError::NoExeName)?;
+        warn!("{} is built with the no-self-update feature: setting auto-self-update will not have any effect.",arg0);
+    }
+    cfg.set_auto_self_update(&m.value_of("auto-self-update-mode").unwrap())?;
     Ok(utils::ExitCode(0))
 }
 
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -48,8 +48,10 @@ use std::borrow::Cow;
 use std::env;
 use std::env::consts::EXE_SUFFIX;
 use std::fs;
+use std::io::Write;
 use std::path::{Component, Path, PathBuf, MAIN_SEPARATOR};
 use std::process::Command;
+use std::str::FromStr;
 
 use anyhow::{anyhow, Context, Result};
 use cfg_if::cfg_if;
@@ -59,6 +61,7 @@ use super::common::{self, ignorable_error, Confirm};
 use super::errors::*;
 use super::markdown::md;
 use super::term2;
+use crate::cli::term2::Terminal;
 use crate::dist::dist::{self, Profile, TargetTriple};
 use crate::process;
 use crate::toolchain::{DistributableToolchain, Toolchain};
@@ -85,6 +88,51 @@ pub struct InstallOpts<'a> {
 pub const NEVER_SELF_UPDATE: bool = true;
 #[cfg(not(feature = "no-self-update"))]
 pub const NEVER_SELF_UPDATE: bool = false;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum SelfUpdateMode {
+    Enable,
+    Disable,
+    CheckOnly,
+}
+
+impl SelfUpdateMode {
+    pub fn modes() -> &'static [&'static str] {
+        &["enable", "disable", "check-only"]
+    }
+
+    pub fn default_mode() -> &'static str {
+        "enable"
+    }
+}
+
+impl FromStr for SelfUpdateMode {
+    type Err = anyhow::Error;
+
+    fn from_str(mode: &str) -> Result<Self> {
+        match mode {
+            "enable" => Ok(Self::Enable),
+            "disable" => Ok(Self::Disable),
+            "check-only" => Ok(Self::CheckOnly),
+            _ => Err(anyhow!(format!(
+                "unknown self update mode: '{}'; valid modes are {}",
+                mode,
+                valid_self_update_modes(),
+            ))),
+        }
+    }
+}
+
+impl ToString for SelfUpdateMode {
+    fn to_string(&self) -> String {
+        match self {
+            SelfUpdateMode::Enable => "enable",
+            SelfUpdateMode::Disable => "disable",
+            SelfUpdateMode::CheckOnly => "check-only",
+        }
+        .into()
+    }
+}
 
 // The big installation messages. These are macros because the first
 // argument of format! needs to be a literal.
@@ -491,7 +539,6 @@ fn do_pre_install_sanity_checks(no_prompt: bool) -> Result<()> {
 }
 
 fn do_pre_install_options_sanity_checks(opts: &InstallOpts<'_>) -> Result<()> {
-    use std::str::FromStr;
     // Verify that the installation options are vaguely sane
     (|| {
         let host_triple = opts
@@ -1103,6 +1150,32 @@ pub fn get_available_rustup_version() -> Result<String> {
     Ok(String::from(available_version))
 }
 
+pub fn check_rustup_update() -> Result<()> {
+    let mut t = term2::stdout();
+    // Get current rustup version
+    let current_version = env!("CARGO_PKG_VERSION");
+
+    // Get available rustup version
+    let available_version = get_available_rustup_version()?;
+
+    let _ = t.attr(term2::Attr::Bold);
+    write!(t, "rustup - ")?;
+
+    if current_version != available_version {
+        let _ = t.fg(term2::color::YELLOW);
+        write!(t, "Update available")?;
+        let _ = t.reset();
+        writeln!(t, " : {} -> {}", current_version, available_version)?;
+    } else {
+        let _ = t.fg(term2::color::GREEN);
+        write!(t, "Up to date")?;
+        let _ = t.reset();
+        writeln!(t, " : {}", current_version)?;
+    }
+
+    Ok(())
+}
+
 pub fn cleanup_self_updater() -> Result<()> {
     let cargo_home = utils::cargo_home()?;
     let setup = cargo_home.join(&format!("bin/rustup-init{}", EXE_SUFFIX));
@@ -1112,6 +1185,14 @@ pub fn cleanup_self_updater() -> Result<()> {
     }
 
     Ok(())
+}
+
+pub fn valid_self_update_modes() -> String {
+    SelfUpdateMode::modes()
+        .iter()
+        .map(|s| format!("'{}'", s))
+        .collect::<Vec<_>>()
+        .join(", ")
 }
 
 #[cfg(test)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -11,6 +11,7 @@ use pgp::{Deserializable, SignedPublicKey};
 use serde::Deserialize;
 use thiserror::Error as ThisError;
 
+use crate::cli::self_update::SelfUpdateMode;
 use crate::dist::download::DownloadCfg;
 use crate::dist::{
     dist::{self, valid_profile_names},
@@ -405,6 +406,20 @@ impl Cfg {
         Ok(())
     }
 
+    pub fn set_auto_self_update(&mut self, mode: &str) -> Result<()> {
+        match SelfUpdateMode::from_str(mode) {
+            Ok(update_mode) => {
+                self.settings_file.with_mut(|s| {
+                    s.auto_self_update = Some(update_mode);
+                    Ok(())
+                })?;
+                (self.notify_handler)(Notification::SetSelfUpdate(mode));
+                Ok(())
+            }
+            Err(err) => Err(err),
+        }
+    }
+
     pub fn set_toolchain_override(&mut self, toolchain_override: &str) {
         self.toolchain_override = Some(toolchain_override.to_owned());
     }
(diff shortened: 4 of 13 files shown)


0xPoe:
/hold

The PR cannot be review yet.

- [x] Better code 
- [x] Add tests

src/cli/rustup_mode.rs
```diff
@@ -8,12 +8,13 @@ use std::str::FromStr;
...
 use clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, Shell, SubCommand};
 
+use super::common;
 use super::help::*;
 use super::self_update;
 use super::term2;
 use super::term2::Terminal;
 use super::topical_doc;
-use super::{common, self_update::get_available_rustup_version};
+use crate::cli::self_update::{check_rustup_update, SELF_UPDATE_MODES};
```
rbtcollins: Why the change from super::... to crate::cli::... ?

I'm not against it, just curious?

```
use super::{common, self_update::{check_rustup_update, SELF_UPDATE_MODES}}
```

would have worked fine

src/cli/rustup_mode.rs
```diff
@@ -911,31 +923,16 @@ fn check_updates(cfg: &Cfg) -> Result<utils::ExitCode> {
...
-    }
     Ok(utils::ExitCode(0))
 }
 
 fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    let self_update = !m.is_present("no-self-update") && !self_update::NEVER_SELF_UPDATE;
+    let self_update_mode = cfg.get_self_update_mode()?;
+    let self_update = !m.is_present("no-self-update")
+        && !self_update::NEVER_SELF_UPDATE
+        && self_update_mode == "enable";
```
rbtcollins: I am not sure I am interpreting this boolean expression correctly - and even if I am, I'm not certain of the intent or user story here; this new feature needs documenting, and that would be a good place to describe the interactions of all the various bits:
- it might be compiled out
- it might be controlled via the command line
- it might be controlled via state

Do places of settings have precedence, or values of settings, or some mixture of both? Lets write it down clearly and see if it makes sense.

src/cli/rustup_mode.rs
0xPoe: Oops! My IDE imported it for me. Fixed.

src/cli/rustup_mode.rs
0xPoe: Added. Update only if rustup does **not** have the no-self-update feature, and auto-self-update is configured to **enable** and has **no** no-self-update parameter.

All other cases will not be updated.

Priority: no-self-update feature > self_update_mode > no-self-update args.

doc/src/basics.md
```diff
@@ -40,8 +40,14 @@ This is the essence of `rustup`.
 
 ## Keeping rustup up to date
 
-Running `rustup update` also checks for updates to `rustup` and automatically
-installs the latest version. To manually check for updates and install the
+We control the automatic update of rustup by setting the auto-self-update configuration. 
+The auto-self-update setting supports three values: `enable` and `disable` and `check-only`.
```
kinnison:
```suggestion
It is possible to control Rustup's automatic self update mechanism with the `auto-self-update` configuration
variable.  This setting supports three values: `enable` and `disable` and `check-only`.
```

doc/src/basics.md
```diff
@@ -40,8 +40,14 @@ This is the essence of `rustup`.
...
+
+When it is `enable`, running `rustup update` will also check for updates to `rustup` and automatically install the latest version.
+If it is `disable`, the available updates is not checked and installed. 
+In addition, you can set auto-self-update to `check-only` if you want to check rustup for available updates during `rustup update` as well.
```
kinnison:
```suggestion
* `disable` will ensure that no automatic self updating actions are taken.
* `enable` will mean that `rustup update` and similar commands will also check-for, and install, any update to Rustup.
* `check-only` will cause any automatic self update to check and report on any updates, but not to automatically install them.
```

doc/src/basics.md
```diff
@@ -40,8 +40,14 @@ This is the essence of `rustup`.
...
+
+To manually force check for updates and install the
```
kinnison:
```suggestion
Whether automatic self-update is enabled or not, you can request that Rustup check for, and update it self to the
```

doc/src/basics.md
```diff
@@ -40,8 +40,14 @@ This is the essence of `rustup`.
...
 latest version of `rustup` without updating installed toolchains type `rustup
```
kinnison:
```suggestion
latest version of `rustup` without updating any installed toolchains by running `rustup
```

src/cli/rustup_mode.rs
```diff
@@ -1021,6 +1023,10 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
         cfg.temp_cfg.clean();
     }
 
+    if !self_update && self_update_mode == "check-only" {
```
kinnison: Should we check even if `--no-self-update` is specified?  I don't mind if the thought is "yes we should" but I wanted to raise it as a potential inconsistency here.

src/config.rs
```diff
@@ -405,6 +406,22 @@ impl Cfg {
         Ok(())
     }
 
+    pub fn set_auto_self_update(&mut self, mode: &str) -> Result<()> {
```
kinnison, line 409: I'm not entirely sure I like this use of strings here, I think I'd prefer an enumeration for the mode, but I also won't block this PR on that if you don't want to rework our internal APIs to use an enum.

src/notifications.rs
```diff
@@ -102,6 +104,7 @@ impl<'a> Display for Notification<'a> {
                 name
             ),
             SetProfile(name) => write!(f, "profile set to '{}'", name),
+            SetSelfUpdate(mode) => write!(f, "self update mode to '{}'", mode),
```
kinnison:
```suggestion
            SetSelfUpdate(mode) => write!(f, "auto-self-update mode set to '{}'", mode),
```

kinnison (changes requested):
There are a number of whitespace changes in this PR.  I don't mind them too much, but in future it's best if this kind of reformatting is only done in independent commits/PRs.

I have made a number of wording change suggestions.  You do not have to accept them all, but in general I think they're likely to improve matters.

src/cli/rustup_mode.rs
0xPoe: I think if it's check-only then it's better to always check for updates. I simplified the logic. In fact the first condition is useless.

src/config.rs
0xPoe: Yes, I totally agree that using enumeration is better. I added a fixme and I will change our code to support enums in the next PR.

src/cli/rustup_mode.rs
0xPoe: Also if the user has set check-only I think they may no longer use --no-self-update.

0xPoe:
> There are a number of whitespace changes in this PR. I don't mind them too much, but in future it's best if this kind of reformatting is only done in independent commits/PRs.

Got it.

src/cli/rustup_mode.rs
0xPoe: Simply put, --no-self-update only useful when auto_self_update is enable and #[cfg(not(feature = "no-self-update"))].

0xPoe: Priority: no-self-update feature > self_update_mode > no-self-update args.

0xPoe:
@kinnison Could you please take another look? All comments addressed. Thanks!

0xPoe:
Conflicts have been resolved.

doc/src/basics.md
```diff
@@ -30,20 +30,29 @@ info: downloading self-updates
...
+## Keeping rustup up to date
+
+It is possible to control Rustup's automatic self update mechanism with the `auto-self-update` configuration
+variable.  This setting supports three values: `enable` and `disable` and `check-only`.
+
+* `disable` will ensure that no automatic self updating actions are taken.
+* `enable` will mean that `rustup update` and similar commands will also check-for, and install, any update to Rustup.
+* `check-only` will cause any automatic self update to check and report on any updates, but not to automatically install them.
+
+Whether automatic self-update is enabled or not, you can request that Rustup check for, and update it self to the
```
kinnison:
```suggestion
Whether automatic self-update is enabled or not, you can request that Rustup check for, and update itself to the
```

doc/src/basics.md
```diff
@@ -30,20 +30,29 @@ info: downloading self-updates
...
 ```
 
-> #### Disable automatic self-updates
-> `rustup` will automatically update itself at the end of any toolchain installation.
-> You can prevent this automatic behaviour by passing the `--no-self-update` argument
-> when running `rustup update` or `rustup toolchain install`.
+**Note**: If auto-self-update is `enable`, `rustup` will automatically update itself at the end of any
+toolchain installation as well.  You can prevent this automatic behaviour by
+passing the `--no-self-update` argument when running `rustup update` or
+`rustup toolchain install`.
```
kinnison: A previous change deliberately called this out in a block quote with its own low-tier heading.  I suggest you don't undo that since it's a good readability improvement.

src/cli/rustup_mode.rs
```diff
@@ -1578,6 +1585,11 @@ fn set_profile(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
+fn set_auto_self_update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
+    cfg.set_auto_self_update(&m.value_of("auto-self-update-mode").unwrap())?;
+    Ok(utils::ExitCode(0))
+}
+
```
kinnison: I wonder if this function should warn the user if rustup was built with the `no-self-update` feature and they set it to `enable` ?

kinnison (changes requested):
A couple of doc tweaks and a question which might be "no need" or might lead to a slight code change - otherwise I think this is nearly ready to merge.

doc/src/basics.md
0xPoe: Fixed.

doc/src/basics.md
0xPoe: Added.

src/cli/rustup_mode.rs
0xPoe: Added.

0xPoe:
Commits squashed.

doc/src/basics.md
```diff
@@ -30,20 +30,30 @@ info: downloading self-updates
...
 
-> #### Disable automatic self-updates
-> `rustup` will automatically update itself at the end of any toolchain installation.
-> You can prevent this automatic behaviour by passing the `--no-self-update` argument
-> when running `rustup update` or `rustup toolchain install`.
+### Disable automatic self-updates with the parameter of the command
+If auto-self-update is `enable`, `rustup` will automatically update itself at the end of any
+toolchain installation as well.  You can prevent this automatic behaviour by
+passing the `--no-self-update` argument when running `rustup update` or
+`rustup toolchain install`.
```
kinnison: You still seem to have removed the blockquote (`>`) here which was added to provide callout clarity.

```suggestion
> ### Disable automatic self-updates with the parameter of the command
> If auto-self-update is `enable`, `rustup` will automatically update itself at the end of any
> toolchain installation as well.  You can prevent this automatic behaviour by
> passing the `--no-self-update` argument when running `rustup update` or
> `rustup toolchain install`.
```

kinnison (changes requested):

doc/src/basics.md
0xPoe: Sorry I missed it. 
Added. Thanks for your review!

kinnison (approved):
Assuming CI goes green, I think this is okay.  @rbtcollins ?

doc/src/basics.md
```diff
@@ -30,20 +30,30 @@ info: downloading self-updates
...
 info: checking for self-updates
 info: downloading self-updates
 ```
 
-> #### Disable automatic self-updates
-> `rustup` will automatically update itself at the end of any toolchain installation.
-> You can prevent this automatic behaviour by passing the `--no-self-update` argument
-> when running `rustup update` or `rustup toolchain install`.
+> ### Disable automatic self-updates with the parameter of the command
+> If auto-self-update is `enable`, `rustup` will automatically update itself at the end of any
```
rbtcollins: This should probably mention that the default is `enable`?

doc/src/basics.md
```diff
@@ -30,20 +30,30 @@ info: downloading self-updates
...
+It is possible to control Rustup's automatic self update mechanism with the `auto-self-update` configuration
+variable.  This setting supports three values: `enable` and `disable` and `check-only`.
+
+* `disable` will ensure that no automatic self updating actions are taken.
+* `enable` will mean that `rustup update` and similar commands will also check-for, and install, any update to Rustup.
+* `check-only` will cause any automatic self update to check and report on any updates, but not to automatically install them.
+
+Whether automatic self-update is enabled or not, you can request that Rustup check for, and update itself to the
+latest version of `rustup` without updating any installed toolchains by running `rustup
+self update`:
```
rbtcollins: I think this section also needs to mention the feature that can be disable that disables this capability entirely.

src/cli/rustup_mode.rs
```diff
@@ -911,31 +925,20 @@ fn check_updates(cfg: &Cfg) -> Result<utils::ExitCode> {
...
-        let _ = t.reset();
-        writeln!(t, " : {}", current_version)?;
-    }
     Ok(utils::ExitCode(0))
 }
 
 fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
-    let self_update = !m.is_present("no-self-update") && !self_update::NEVER_SELF_UPDATE;
+    let self_update_mode = cfg.get_self_update_mode()?;
+    // Priority: no-self-update feature > self_update_mode > no-self-update args.
```
rbtcollins, line 936: This should be documented somewhere users can see it; since users need to reason about it.

src/settings.rs
```diff
@@ -84,6 +89,7 @@ impl Default for Settings {
             profile: Some("default".to_owned()),
             overrides: BTreeMap::new(),
             pgp_keys: None,
+            auto_self_update: Some("enable".to_owned()),
```
rbtcollins: For a typed valued in the core here, I'd -really- rather have validated data not strings: I think it will allow fail fast rather than undefined behaviour.

src/cli/self_update.rs
```diff
@@ -1103,6 +1106,32 @@ pub fn get_available_rustup_version() -> Result<String> {
     Ok(String::from(available_version))
 }
 
+pub fn check_rustup_update() -> Result<()> {
```
rbtcollins, line 1153: Just an idea, what about putting this in self-update:

pseudocode:

```
fn self_update(mode: UpdateMode) -> Result<()> {
   if #[cfg(no_self_update)] {
       return Ok(());
   }  else {
      match mode {
         UpdateMode::Disable => { return Ok(()) },
         UpdateMode::CheckOnly => { ... check logic ... }
         UpdateMode::Enable => { ... update logic ... {
     }
  }
}
```

Then you don't need a separate call for checking self updates and can have all the logic about whether self update are possible or not in just one place.

doc/src/basics.md
0xPoe: Mentioned.

doc/src/basics.md
0xPoe: It was mentioned at the very beginning.

src/cli/rustup_mode.rs
0xPoe: The documentation now mentions them in order of priority.

src/settings.rs
0xPoe: A FIXME has been added and I will change our code to support enum in the next PR, now it is difficult to use enum as a configuration item. https://github.com/rust-lang/rustup/blob/3baa99556fd2c06faaf03527a567bc026d5d0899/src/config.rs#L409

In addition this fn now validates the value entered by the user.

src/cli/self_update.rs
0xPoe: This fn is also used in `rustup check`, so I think we can keep it.

rbtcollins: I deliberately suggested an interface taking a mode - that would allow it to be used from `rustup check` as well

src/cli/rustup_mode.rs
```diff
@@ -1578,6 +1585,14 @@ fn set_profile(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
+fn set_auto_self_update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
+    if self_update::NEVER_SELF_UPDATE {
+        warn!("Your rustup is built with the no-self-update feature, so setting auto-self-update will not have any effect.");
```
rbtcollins: Rather than this conversational style, lets be specific about what rustup is involved:

```/home/foo/.cargo/bin/rustup is built with no-self-update feature: setting auto-self-update will not have any effect.```

File: src/settings.rs (hunk 17 above)
Comment (rbtcollins) [reply src/settings.rs]: I understand. I'm expressing a different opinion to @kinnison : I think converting the core to an enum is very straight forward, and there's no need to wait for another PR to do that - this PR itself sat for a week with no movement because the reviewers were busy (very sorry about that!) - but in that time the enum stage could have happened, since we all seem to be agreed that its needed.

If there is some substantial problem doing the enum work that will delay things, then I'm ok merging-and-improving, but I've looked, and it seems very shallow: have a TryInto implementation from str to the Enum type; use it in the toml parser, use it in the config setter, and you're done.

**Review by rbtcollins — CHANGES_REQUESTED** [review]

File: src/cli/self_update.rs (hunk 18 above)
Comment (0xPoe) [reply src/cli/self_update.rs:1153]: The check in the rustup check does not actually check UpdateMode. So I don't think it's necessary to couple it?

File: src/settings.rs (hunk 17 above)
Comment (0xPoe) [reply src/settings.rs]: I agree that these tasks are straightforward, but I actually want to modify the profile-related code as well. So that part of the change is not related to this pr. It only extends the pr merge cycle(It's been fifteen days.). So I would prefer to change the configuration item in the next PR.

File: src/cli/rustup_mode.rs (hunk 19 above)
Comment (0xPoe) [reply src/cli/rustup_mode.rs]: Good idea! Added.

File: src/cli/self_update.rs (hunk 18 above)
Comment (0xPoe) [reply src/cli/self_update.rs:1153]: The logic you describe is actually only used in the update method. The value of UpdateMode is not considered elsewhere.
https://github.com/rust-lang/rustup/blob/330a5b2a2bb60389747e58e82c4646bc0a5bd13d/src/cli/rustup_mode.rs#L933

File: doc/src/basics.md [hunk 20]
```diff
@@ -30,20 +30,31 @@ info: downloading self-updates
 
 ## Keeping `rustup` up to date
 
-Running `rustup update` also checks for updates to `rustup` itself and automatically
-installs the latest version. To manually update `rustup` only,
-without updating installed toolchains, type `rustup self update`:
+If your `rustup` build with the `no-self-update` feature, it will not be able to update itself.
```
rbtcollins: Grammar fix: "was built with"

doc/src/basics.md
```diff
@@ -30,20 +30,31 @@ info: downloading self-updates
...
+
+If your `rustup` build without the `no-self-update` feature,
```
rbtcollins: same grammar fix needed here

doc/src/basics.md
rbtcollins: Suggest adding that this is the default and that the rustups available from https://rustup.rs are built this way.

doc/src/basics.md
```diff
@@ -30,20 +30,31 @@ info: downloading self-updates
...
-installs the latest version. To manually update `rustup` only,
-without updating installed toolchains, type `rustup self update`:
+If your `rustup` build with the `no-self-update` feature, it will not be able to update itself.
+
+If your `rustup` build without the `no-self-update` feature, 
+it is possible to control Rustup's automatic self update mechanism with the `auto-self-update` configuration variable. 
+This setting supports three values: `enable` and `disable` and `check-only`.
+
+* `disable` will ensure that no automatic self updating actions are taken.
+* `enable` will mean that `rustup update` and similar commands will also check-for, and install, any update to Rustup.
```
rbtcollins: check for should not be hypenated here. Just say "commands will also check for, and install, any new version of Rustup.

doc/src/basics.md
```diff
@@ -30,20 +30,31 @@ info: downloading self-updates
...
 $ rustup self update
 info: checking for self-updates
 info: downloading self-updates
 ```
 
-> #### Disable automatic self-updates
-> `rustup` will automatically update itself at the end of any toolchain installation.
-> You can prevent this automatic behaviour by passing the `--no-self-update` argument
-> when running `rustup update` or `rustup toolchain install`.
+> ### Disable automatic self-updates with the parameter of the command
```
rbtcollins: The conditional clause makes this hard to process. Perhaps we start over now things are different. How about
```
### Disabling self updates on a per-invocation basis
> Self updates can also be suppressed on individual invocations of `rustup` by passing the argurment `--no-self-update`  when running `rustup update` or `rustup toolchain install`.
```

src/cli/rustup_mode.rs
```diff
@@ -1578,6 +1585,14 @@ fn set_profile(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
+fn set_auto_self_update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
+    if self_update::NEVER_SELF_UPDATE {
+        warn!("{} is built with the no-self-update feature: setting auto-self-update will not have any effect.",cfg.rustup_dir.display());
```
rbtcollins: I think you need arg0 + os.path.realpath, like from proxy_mode.rs
```
        let mut args = crate::process().args_os();
        let arg0 = args.next().map(PathBuf::from);
```
     rustup_dir is just the directory that the common case install path for the binary is: we want the actual path, which for a snap, is some random directory.

src/cli/self_update.rs
```diff
@@ -86,9 +89,53 @@ pub const NEVER_SELF_UPDATE: bool = true;
...
+    }
+}
+
+impl ToString for SelfUpdateMode {
+    fn to_string(&self) -> String {
+        let modes = Self::modes();
+        match self {
+            SelfUpdateMode::Enable => modes[0].to_string(),
+            SelfUpdateMode::Disable => modes[1].to_string(),
+            SelfUpdateMode::CheckOnly => modes[2].to_string(),
```
rbtcollins: I don't think the modes array here adds anything: you have connaisence of value between the ordering of the array and the values in the lookup, vs connaisence of value between an inline string here and a string in FromStr; string interning should optimise the binary size the same either way, and so it becomes what is easier to read: 

```
match self {
   SelfUpdateMode::Enable => "enable",
   SelfUpdateMode::Disable => "disable"
   SelfUpdateMode::CheckOnly => "check-only"
}.into()
```

would be my preference - but its ok if you don't choose to change it; just making a note in passing.

src/settings.rs
```diff
@@ -84,6 +91,7 @@ impl Default for Settings {
             profile: Some("default".to_owned()),
             overrides: BTreeMap::new(),
             pgp_keys: None,
+            auto_self_update: Some(SelfUpdateMode::Enable),
```
rbtcollins: Either this should default to None, or the type should not have Option<> in it.

If we want to support writing the file back with an unset key unset (which I think is probably good), then we need to keep None as the default value and only cast up to Enable on queries. Otherwise new files will be written with a value in them.

And get_self_update_mode already returns Enable when the value is None....

rbtcollins (commented):
looking quite good

0xPoe:
@rbtcollins All comments addressed. Thanks for your review!

0xPoe:
@rbtcollins @kinnison  Can we merge it? Are there any other issues?

src/cli/rustup_mode.rs
```diff
@@ -1578,6 +1586,17 @@ fn set_profile(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
+fn set_auto_self_update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
+    if self_update::NEVER_SELF_UPDATE {
+        let mut args = crate::process().args_os();
+        let arg0 = args.next().map(PathBuf::from);
+        let arg0 = arg0.as_ref().and_then(|a| a.to_str());
+        warn!("{} is built with the no-self-update feature: setting auto-self-update will not have any effect.",arg0.ok_or(CLIError::NoExeFilePath)?);
```
kinnison:
```suggestion
        let arg0 = arg0.as_ref().and_then(|a| a.to_str());
        let arg0 = arg0.ok_or(CLIError::NoExeFilePath)?;
        warn!("{} is built with the no-self-update feature: setting auto-self-update will not have any effect.",arg0);
```

(I don't mind at all if you don't take this, but I think it'd look neater, the `warn` line is already very long)

kinnison (approved):
One tiny suggestion, otherwise I'm good with this.  If @rbtcollins is happy then we can merge IMO.

src/cli/rustup_mode.rs
0xPoe: Fixed. Thanks for your review.

0xPoe:
It seems the docker image is broken.

doc/src/basics.md
```diff
@@ -30,20 +30,32 @@ info: downloading self-updates
 
 ## Keeping `rustup` up to date
 
-Running `rustup update` also checks for updates to `rustup` itself and automatically
-installs the latest version. To manually update `rustup` only,
-without updating installed toolchains, type `rustup self update`:
+If your `rustup` was build with the `no-self-update` feature, it will not be able to update itself.
```
rbtcollins: should be past tense 

'was built with'

src/cli/rustup_mode.rs
```diff
@@ -1578,6 +1586,20 @@ fn set_profile(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
...
 }
 
+fn set_auto_self_update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
+    if self_update::NEVER_SELF_UPDATE {
+        let mut args = crate::process().args_os();
+        let arg0 = args.next().map(PathBuf::from);
+        let arg0 = arg0
+            .as_ref()
+            .and_then(|a| a.to_str())
+            .ok_or(CLIError::NoExeFilePath)?;
```
rbtcollins: We already have NoExeName : I don't think is different enough to bother with a new error (and as the error is never caught and never repeated, we can just use anyhow!(...)

rbtcollins (commented):
Minor nits - I'll fix these I think

0xPoe:
@rbtcollins @kinnison Thanks again for your review!

PR #2768, opened 2021-05-13
docs: Edit "Basic usage" page

pierwill:
Makes edits throughout for clarity, concision, and formatting.

Moves the help section to the bottom of the page.

Uses the block quote syntax for the note on disabling automatic self-updates.


kinnison (approved):
I like this rework.  Thank you for your effort.

I shall merge this after 1.24.2 is released.

PR #2772, opened 2021-05-16
Update windows.rs

asv7c2:
Escape fix


kinnison (approved):
I am embarassed that we missed this, thank you for the fix.

Assuming green CI, I'll merge this after the 1.24.2 effort.

0xPoe:
It looks like the GitHub actions service is out of service! Maybe the test will be triggered after the fix.

PR #2779, opened 2021-05-23
Fix low-memory issue and lower tier platforms with no sysinfo

kinnison:
Lower memory (where 32M might be the bare minimum we can offer the unpacker) or lower-tier platforms which lack sysinfo support (where we default to 32M of unpack RAM) were **hanging** during install.  Turns out this is because we were failing to account for the **release** of the last chunk in an incremental file unpack.  This wasn't a memory leak per-se, but it did mean that the accounting system refused to let the component proceed because it didn't think there was a free chunk to use.

This pair of commits first introduces a test change which demonstrates the issue and then introduces a fix.

If this is merged, we should do a release in order to deal with #2774 asap.

### DIFF
--- a/src/diskio/mod.rs
+++ b/src/diskio/mod.rs
@@ -328,6 +328,10 @@ pub(crate) trait Executor {
 
     /// Query the memory budget to see if a particular size buffer is available
     fn buffer_available(&self, len: usize) -> bool;
+
+    #[cfg(test)]
+    /// Query the memory budget to see how much of the buffer pool is in use
+    fn buffer_used(&self) -> usize;
 }
 
 /// Trivial single threaded IO to be used from executors.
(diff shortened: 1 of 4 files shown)


src/diskio/mod.rs
```diff
@@ -418,9 +422,10 @@ pub(crate) fn write_file_incremental<P: AsRef<Path>, F: Fn(usize)>(
             let len = contents.len();
             // Length 0 vector is used for clean EOF signalling.
             if len == 0 {
+                drop(contents);
```
rbtcollins, line 426: Probably should add a 
`trace_scope!("EOF_chunk", "name": path_display, "len": len);` at the top of this scope.

rbtcollins (commented):
.

kinnison:
Scope trace added, hope that's good now.

rbtcollins:
heh check the timeline... I had already done so :)

chrisduerr:
@kinnison mentioned the potential for a release in order to deal with this. Is a 1.24.3 release planned? If so, I'd be curious if there are any other things blocking it.

kinnison:
@chrisduerr Once #2782 is done I think a new release is plausible.  Assuming Robert has time to review and merge it soon, I'll aim to stage a beta release early in the week.

PR #2782, opened 2021-05-23
install: improve heuristic for deprecation warning

### LINKED ISSUE #2765: Requiring --force-non-host for *-pc-windows-gnu
When I try to install a *-pc-windows-gnu toolchain on Windows, I always get something like
```
error: DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain as the default.
warning: toolchain '1.37.0-x86_64-pc-windows-gnu' may not be able to run on this system.
warning: If you meant to build software to target that platform, perhaps try `rustup target add x86_64-pc-windows-gnu` instead?
```

Does it really make sense to require `--force-non-host` for a GNU toolchain install under Windows? I do not want to cross compile, so I don't think it makes sense to use `rustup target add` instead.

### LINKED ISSUE #2773: --force-non-host warning is too broad
**Problem**
`rustup update stable-msvc` (or other combinations like `nightly-gnu`) on Windows generates a warning:

```
error: DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain as the default.
warning: toolchain 'stable-msvc' may not be able to run on this system.
warning: If you meant to build software to target that platform, perhaps try `rustup target add msvc` instead?
```

**Possible Solution(s)**
Don't warn about installing msvc or gnu on windows.

**Notes**

Output of `rustup --version`: rustup 1.24.2 (755e2b07e 2021-05-12)
Output of `rustup show`:
rustup home:  C:\Users\User\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc

active toolchain
----------------

rustc 1.52.0-nightly (fe1bf8e05 2021-02-23)

kinnison:
Nominally this will fix #2773 and fix #2765

Basically adjust the heuristic from just "host != target" to account for Windows a little more.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -962,7 +962,7 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
                 let host_arch = TargetTriple::from_host_or_build();
                 if let Ok(toolchain_desc) = ToolchainDesc::from_str(name) {
                     let target_triple = toolchain_desc.target;
-                    if !forced && host_arch.ne(&target_triple) {
+                    if !forced && !host_arch.can_run(&target_triple)? {
                         err!("DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain as the default.");
                         warn!(
                             "toolchain '{}' may not be able to run on this system.",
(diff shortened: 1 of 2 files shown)


rbtcollins (commented):
Does can_run get called with self == other ? 

perhaps a couple of tests?

kinnison:
Good catch re `self==other` - that could happen and would have failed.  I've written a unit test for this and assuming it passes I'll push soon.

kinnison:
@rbtcollins If you're okay with this, I'd like to get it merged and then look at doing another staged beta release so that we can get the fixes out soon.

PR #2783, opened 2021-05-25
Add freebsd CI test suite

GuillaumeGomez:


ci/actions-templates/linux-builds-template.yaml
```diff
@@ -34,7 +34,6 @@ jobs:
           - i686-unknown-linux-gnu        # skip-pr skip-master
           - arm-unknown-linux-gnueabi     # skip-pr skip-master
           - arm-unknown-linux-gnueabihf   # skip-pr skip-master
-          - x86_64-unknown-freebsd        # skip-pr skip-master
```
kinnison: This change is going to cause releases of freebsd to break.

ci/actions-templates/freebsd-builds-template.yaml
```diff
@@ -0,0 +1,83 @@
...
+            #
+            echo "========="
+            echo "Run formatting checks"
+            cargo fmt --all -- --check
+            #
+            echo "========="
+            echo "Run cargo check and clippy"
+            cargo check --all --all-targets
+            git ls-files -- '*.rs' | xargs touch
+            cargo clippy --all --all-targets
```
kinnison: These aren't needed on FreeBSD, they're part of the general checks

kinnison (changes requested):
The way you've done this, we wouldn't get the FreeBSD artifact released properly.  Either the FreeBSD build needs to support the s3 upload or else we need to not use it on the Stable branch and instead just do the cross-build from Linux for that.

GuillaumeGomez:
Updated!

ci/actions-templates/freebsd-builds-template.yaml
```diff
@@ -0,0 +1,73 @@
...
+            rustup default beta
+            #
+            echo "========="
+            echo "Ensure we have the components we need"
+            rustup component add rustfmt
+            rustup component add clippy
+            #
+            echo "========="
+            echo "Run the freebsd check"
+            sh ci/raw_init.sh
```
kinnison: This isn't the ideal script to be running here, you probably meant to use:
```suggestion
            bash ci/run.bash
```

But you'll also need to have set up some environment variables first,

You'll want `TARGET` to be set, `SKIP_TESTS` to be unset/empty, `LIBZ_SYS_STATIC=1` at minimum.

kinnison (changes requested):

ci/actions-templates/freebsd-builds-template.yaml
kinnison: You know, I said *bash* for a reason :D

GuillaumeGomez: Yes but bash wasn't found. Need to install it I guess...

kinnison:
Based on the discussion on Discord, I'd guess that https://github.com/rust-lang/rustup/runs/2787265203?check_suite_focus=true#step:3:2747 is caused by the `unimplemented!()` in https://github.com/rust-lang/rustup/blob/7a051f7f02075b54da58bf59d107b635518b1ba5/src/test.rs#L102-L109 that was identified by @workingjubilee in the channel.

kinnison:
I am going to try and fix things without disabling a bunch of tests, sorry for the noise...

kinnison:
The remaining test failures are because we need a newer `effective-limits` for a newer `sys-info` before we can merge.

kinnison:
I've added a temporary change to a git version of effective-limits to see if that fixes the tests

kinnison:
The test which is failing is not one which fails on my freebsd VM at home.

asomers:
Using a nested VM is awkward, inefficient and expensive (even if it's free for you, it costs money for somebody).  Might I suggest using Cirrus-CI instead?  It natively supports FreeBSD.  You can keep the other OSes on Github Actions if you want.

kinnison:
While this is a clever approach and we may readdress this in the future, I don't feel like we have the energy to pursue it for now.  I'm going to close this PR, but would happily accept someone coming up with a less onerous approach in the future.  Ideally if Github Actions ever add proper BSD support.

PR #2784, opened 2021-05-28
Use profile enum in settings

0xPoe:
Maybe we all use enumerations? cc: @rbtcollins  @kinnison

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -6,6 +6,7 @@ use std::str::FromStr;
 use anyhow::{Context, Result};
 
 use crate::cli::self_update::SelfUpdateMode;
+use crate::dist::dist::Profile;
 use crate::errors::*;
 use crate::notifications::*;
 use crate::toml_utils::*;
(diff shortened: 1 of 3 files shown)


kinnison (approved):
I think this is a good incremental cleanup.  @rbtcollins , are you in agreement that this is a step in the right direction?

src/settings.rs
```diff
@@ -157,11 +158,18 @@ impl Settings {
             },
             None => None,
         };
+        let profile = match get_opt_string(&mut table, "profile", path)? {
+            Some(profile) => match Profile::from_str(profile.as_str()) {
```
rbtcollins: this could use Result::ok() I think, no? 

``` let profile = get_opt_string(&mut table, "profile", path)?.map(|o| Profile::from_str(o.as_str()).ok());```

0xPoe: It returns option`option<option<profile>>`. But we want `option<profile>`.

0xPoe: We should use `and_then`.

0xPoe: Thanks for your review!

0xPoe:
@kinnison @rbtcollins Can we merge it now? 😆

kinnison (approved):

PR #2787, opened 2021-05-31
Make all utils functions and structs crate-private and cleanup unused functions

0xPoe:
part of  https://github.com/rust-lang/rustup/issues/2730
Because we use some methods in our tests, we need to set it to pub.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -246,7 +246,7 @@ pub struct Cfg {
 }
 
 impl Cfg {
-    pub fn from_env(notify_handler: Arc<dyn Fn(Notification<'_>)>) -> Result<Self> {
+    pub(crate) fn from_env(notify_handler: Arc<dyn Fn(Notification<'_>)>) -> Result<Self> {
         // Set up the rustup home directory
         let rustup_dir = utils::rustup_home()?;
 
@@ -359,7 +359,7 @@ impl Cfg {
     }
 
     /// construct a download configuration
-    pub fn download_cfg<'a>(
+    pub(crate) fn download_cfg<'a>(
         &'a self,
         notify_handler: &'a dyn Fn(crate::dist::Notification<'_>),
     ) -> DownloadCfg<'a> {
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -1,17 +1,13 @@
 use std::env;
-use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::io;
 use std::io::Write;
 use std::path::Path;
-use std::process::{Command, ExitStatus};
 use std::str;
-
-use thiserror::Error as ThisError;
 
 use crate::process;
 
-pub fn ensure_dir_exists<P: AsRef<Path>, F: FnOnce(&Path)>(
+pub(crate) fn ensure_dir_exists<P: AsRef<Path>, F: FnOnce(&Path)>(
     path: P,
     callback: F,
 ) -> io::Result<bool> {
@@ -23,7 +19,7 @@ pub fn ensure_dir_exists<P: AsRef<Path>, F: FnOnce(&Path)>(
     }
 }
 
-pub fn is_directory<P: AsRef<Path>>(path: P) -> bool {
+pub(crate) fn is_directory<P: AsRef<Path>>(path: P) -> bool {
     fs::metadata(path).ok().as_ref().map(fs::Metadata::is_dir) == Some(true)
 }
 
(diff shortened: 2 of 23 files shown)


0xPoe:
@kinnison @rbtcollins  Could you please take a look? Thanks!

src/config.rs
```diff
@@ -242,11 +242,11 @@ pub struct Cfg {
     pub env_override: Option<String>,
     pub dist_root_url: String,
     pub dist_root_server: String,
-    pub notify_handler: Arc<dyn Fn(Notification<'_>)>,
+    pub(crate) notify_handler: Arc<dyn Fn(Notification<'_>)>,
```
kinnison: I don't recall if we decided this works right now?

Long term, I'd prefer to get rid of the pub fields entirely and switch to accessor methods, but that's a much larger piece of work.

kinnison (commented):
For the most part I think this is fine - one query, and a quick check from @rbtcollins before this cleanup can go in.

I particularly like that you found what looks like some dead code.

src/config.rs
0xPoe: @rbtcollins mentioned here https://github.com/rust-lang/rustup/pull/2735#discussion_r642432379 that we need to keep pub instead of using pub(crate) and I modified it.

kinnison (approved):
Looks plausible to me. @rbtcollins are you okay for this to merge?

0xPoe:
Conflict resolved, @rbtcollins Could you please take a look?

kinnison:
Please don't merge into your branch, instead rebase onto master.  If you need help with that, I can provide it.

0xPoe:
> Please don't merge into your branch, instead rebase onto master. If you need help with that, I can provide it.

Rebased!

src/utils/raw.rs
```diff
@@ -214,31 +210,11 @@ fn symlink_junction_inner(target: &Path, junction: &Path) -> io::Result<()> {
...
 }
 
-pub fn hardlink(src: &Path, dest: &Path) -> io::Result<()> {
+pub(crate) fn hardlink(src: &Path, dest: &Path) -> io::Result<()> {
     let _ = fs::remove_file(dest);
     fs::hard_link(src, dest)
 }
 
-#[derive(Debug, ThisError)]
-pub enum CommandError {
```
rbtcollins, line 223: So this is just ununused?

rbtcollins (changes requested):
I think we need to introduce 
```
#![deny(unreachable_pub)]
```
To all of our global lints - demo here, try turning it off and on

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dd28683bfd53acd9b2d3bdc3c92db32e

src/utils/raw.rs
0xPoe: Yes. It doesn't look like it's still being used anywhere else.

0xPoe:
> I think we need to introduce
> 
> ```
> #![deny(unreachable_pub)]
> ```
> 
> To all of our global lints - demo here, try turning it off and on
> 
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dd28683bfd53acd9b2d3bdc3c92db32e

@rbtcollins I agree that we should introduce that lint, but should I open it after my current cleanup is done? Right now we have a few more modules of code that need to be modified. If we open it now, there will be a lot of code that needs to be changed that will be warned. Do you think it is acceptable for us to open it at the end? Because the cli module has a lot of code that needs to be changed and I will change it in my next PR.

0xPoe:
Added a [to-do item](https://github.com/rust-lang/rustup/issues/2730#issuecomment-864515156) to the issue.

rbtcollins:
I had another question on CommandError :)..

Re the lint: I think the lint is part of this bug; it doesn't have to be part of this PR - so it can certainly come in as another patch if it has a lot of dead code it picks up.

Oh - but heres an idea: you could be building the modules you touch with it, even if you don't commit the lint, you commit the results of the cleanup of the lint to the modules so each module is only touched once.

0xPoe:
> I had another question on CommandError :)..

Is this it? https://github.com/rust-lang/rustup/pull/2787#discussion_r654885112 I should have replied to you?



> Oh - but heres an idea: you could be building the modules you touch with it, even if you don't commit the lint, you commit the results of the cleanup of the lint to the modules so each module is only touched once.

Good idea, just tried it in the utils mod and no warnings were reported.

rbtcollins:
please squash the two commits together, then I think we merge this

0xPoe:
> please squash the two commits together, then I think we merge this

squashed! Thanks again for your review!

Issue #1826, opened 2019-05-02
Progress indicator broken on Alacritty with ConPTY


LPGhatguy:
## Problem
When updating a toolchain, Rustup tries to draw over its previous progress indicators, but fails, and instead produces beautiful art:

![unknown (6)](https://user-images.githubusercontent.com/654598/57061223-f68bba00-6c70-11e9-91b5-feb0474ca55f.png)

## Steps
1. `rustup update stable`, or another command that causes Rustup to output progress

## Possible Solution(s)
Bash in Alacritty in Windows is a pretty weird config. 😄 

## Notes
Windows 10, Build 1809.

Alacritty has `enable_experimental_conpty_backend` set, which uses the Windows ConPTY API (1809+ only) instead of WinPTY, which is pretty buggy.

```
$ alacritty --version
alacritty 0.3.2
$ rustup --version
rustup 1.18.1 :: 1.18.0+20 (462683b3a 2019-04-25) dirty 1 modification
```

rbtcollins:
So this is interesting; it looks like we're either not sending a \r correctly, or the number of spaces being sent to overwrite content is wrong, leading to wrapping to the next line, and so on. Probably some diagnostic testing will be needed to determine what terminfo we're ending up with with term today.

rbtcollins:
Capturing from discord:
`TERM= rustup toolchain add nightly` displays correctly but without colour. TERM=xterm-256color otherwise.

The `term` package fallsback to probing the Windows Console via console APIs when TERM is unset, and if that fails, rustup then creates a Terminal object that tries to work with dumb terminals (e.g. just depends on `\r` and discards attempts to do bold/colour etc).

So there are two possible code paths we're encountering here when you set `TERM=`
 - Windows console detected by `term`.
 - Rustup no-terminfo-fallback.

So this raises several aspects:
 -> either we have some weird bug, or alacritty isn't actually xterm-256color compatible. 
 -> we may have a conpty detection bug
 ->or we may have a bug doing colour on a detected windows console with conpty

For the latter two need to know which of the code paths we're following; for the former, that seems most likely to be an alacritty issue.
I'm not sure how to tell which we've encountered from a binary; I suggest building a rustup with some debug code (insert into  `    fn make_terminal<T>(terminfo: Option<TermInfo>) -> Option<Box<Terminal<Output = T> + Send>>
` in term2.rs). - that should let you determine which path is being followed.

[...]

PR #2790, opened 2021-06-09
Replace term with termcolor

### LINKED ISSUE #1818: Shift away from `term` crate

### LINKED ISSUE #2292: Colors are broken on MSYS2 shell.
**Problem**

Colors are broken in VSCode terminal with MSYS2 shell. For example, `rustup show` outputs this:

image

But, Cargo's colors work fine:

image

So this isn't an issue on my end afaik.

**Steps**

1. Use Windows, MSYS2, and VSCode terminal
2. Use the MSYS2 shell with
```json
    "terminal.integrated.shell.windows": "C:\\msys64\\msys2_shell.cmd",
    "terminal.integrated.shellArgs.windows": [
        "-defterm",
        "-no-start",
        "-use-full-path",
        "-mingw",
        "-here"
    ],
```
3. Run a command which outputs color like `rustup show`

**Possible Solution(s)**

Copy how Cargo is doing colors.

**Notes**

- Output of `rustup --version`: rustup 1.21.1 (7832b2ebe 2019-12-20)
- Output of `rustup show`: See above.
- Output of `echo $TERM`: xterm-256color

aloucks:
Fixes #1818
Fixes #2292
CC #1826

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -11,5 +11,5 @@ pub mod proxy_mode;
 pub mod rustup_mode;
 pub mod self_update;
 pub mod setup_mode;
-pub mod term2;
+pub(crate) mod term2;
 mod topical_doc;
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -11,5 +11,5 @@ pub mod proxy_mode;
 pub mod rustup_mode;
 pub mod self_update;
 pub mod setup_mode;
-mod term2;
+pub mod term2;
 mod topical_doc;
--- a/src/cli/term2.rs
+++ b/src/cli/term2.rs
@@ -7,21 +7,32 @@ use std::io::Write;
 
 use termcolor::{ColorChoice, ColorSpec, StandardStream, WriteColor};
 
-pub(crate) trait Terminal: io::Write {
-    fn fg(&mut self, color: Color) -> io::Result<()>;
-    fn bg(&mut self, color: Color) -> io::Result<()>;
-    fn attr(&mut self, attr: Attr) -> io::Result<()>;
-    fn reset(&mut self) -> io::Result<()>;
-    fn carriage_return(&mut self) -> io::Result<()>;
-}
+use crate::process;
 
+/// Public via Terminal
 #[derive(Copy, Clone, Debug)]
-pub(crate) enum Color {
+pub enum Color {
     Red,
     Green,
     Yellow,
     Blue,
     Magenta,
+}
+
+/// Public via Terminal
+#[derive(Copy, Clone, Debug)]
+pub enum Attr {
+    Bold,
+    ForegroundColor(Color),
+}
+
+/// Public via currentprocess::filesource
+pub trait Terminal: io::Write {
+    fn fg(&mut self, color: Color) -> io::Result<()>;
+    fn bg(&mut self, color: Color) -> io::Result<()>;
+    fn attr(&mut self, attr: Attr) -> io::Result<()>;
+    fn reset(&mut self) -> io::Result<()>;
+    fn carriage_return(&mut self) -> io::Result<()>;
 }
 
 impl From<Color> for termcolor::Color {
@@ -36,25 +47,13 @@ impl From<Color> for termcolor::Color {
     }
 }
 
-#[derive(Copy, Clone, Debug)]
-pub(crate) enum Attr {
-    Bold,
-    ForegroundColor(Color),
-}
-
 use crate::currentprocess::filesource::Isatty;
 
-// Decorator to:
-// - Disable all terminal controls on non-tty's
-// - Swallow errors when we try to use features a terminal doesn't have
-//   such as setting colours when no TermInfo DB is present
+/// Disable all terminal controls on non-tty's
 pub(crate) struct AutomationFriendlyTerminal {
     stream: StandardStream,
     color: ColorSpec,
 }
-
-pub(crate) type StdoutTerminal = AutomationFriendlyTerminal;
-pub(crate) type StderrTerminal = AutomationFriendlyTerminal;
 
 impl Isatty for AutomationFriendlyTerminal {
     fn isatty(&self) -> bool {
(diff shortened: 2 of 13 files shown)


aloucks:
The build is failing due to the output being sent to the real `stderr` and `stdout`, which causes some test failures. The `currentprocess` testing framework sets global state and makes things a little difficult.

I haven't yet determined the least invasive way to fix the tests. Feedback is welcome!

rbtcollins:
> The build is failing due to the output being sent to the real `stderr` and `stdout`, which causes some test failures

What sends the output there? If termcolor does it, then termcolor needs to be used in a way that it doesn't do that. `stderr` and `stdout` are global state and thus prevent concurrent testing, though there is some horrid capture glue to pretend they aren't in core.

`currentprocess` lets us fully isolate each test and run an entire CLI style program or components of one, in-process without forking: a massive boost to test efficiency and reliability particularly on Windows.

All it requires is that there is no implicit global state: that things are concurrent instance safe.

src/cli.rs
```diff
@@ -11,5 +11,5 @@ pub mod proxy_mode;
 pub mod rustup_mode;
 pub mod self_update;
 pub mod setup_mode;
-mod term2;
+pub mod term2;
```
rbtcollins: its not documented yet but we prefer pub(crate) at this point

src/cli/term2.rs
```diff
@@ -3,238 +3,150 @@
...
-            .unwrap()
-    }
-    /// Return a Terminal wrapping stdout, or None if a terminal couldn't be
-    /// opened.
-    #[allow(unused)]
-    pub fn stdout(terminfo: Option<TermInfo>) -> Option<Box<StdoutTerminal>> {
-        make_terminal(terminfo, io::stdout)
-    }
+#[derive(Copy, Clone, Debug)]
+pub enum Color {
```
rbtcollins, line 17: Is there any reason not to just re-export termcolor::Color ? This enum doesn't seem to be interestingly different to termcolor::Color

src/cli/term2.rs
```diff
@@ -3,238 +3,150 @@
...
-    static ref TERMINFO: Mutex<Option<term::terminfo::TermInfo>> =
-        Mutex::new(term::terminfo::TermInfo::from_env().ok());
-}
-
-pub(crate) fn stdout() -> StdoutTerminal {
-    let info_result = TERMINFO.lock().unwrap().clone();
-    AutomationFriendlyTerminal(termhack::make_terminal_with_fallback(info_result, || {
-        process().stdout()
-    }))
+pub(crate) fn stdout() -> Box<dyn Terminal> {
```
rbtcollins, line 181: I'm very concerned that this is undoing the performance work we did w.r.t. terminfo probing on linux and on Windows - can you step me through the details of termcolors behaviour in this regard : does it make assumptions? Does it probe? 

If it doesn't probe and just uses an OS compile switch, what sort of compat are we dealing with with things like illumOS and so on?

src/utils/tty.rs
```diff
@@ -1,42 +1,7 @@
...
-#[inline]
-#[cfg(unix)]
-fn isatty(fd: libc::c_int) -> bool {
-    unsafe { libc::isatty(fd) == 1 }
-}
-
-#[inline]
-#[cfg(windows)]
-fn isatty(fd: winapi::shared::minwindef::DWORD) -> bool {
-    if std::env::var("MSYSTEM").is_ok() {
```
rbtcollins, line 31: have you tested with MSYS shells?

rbtcollins: (Or I guess MSYS2 w/in VSCode ?)

src/cli/term2.rs
rbtcollins: Ah yes - https://lessis.me/atty/src/atty/lib.rs.html#66

so we do indeed still need to cache information, though perhaps not the way we were: we may be better off lifting this up to the process() abstraction, allowing it to maintain coherency of the tty metadata with the process stream abstraction, probe once there, and then constructing things around it as you do should just be an in-process overhead, not kernel transitions and other noise.

src/cli/term2.rs
aloucks: The intent was to encapsulate all terminal related things in `term2` so that future refactor or swapping of `termcolor` for something else has minimal impact.

src/utils/tty.rs
aloucks: Yes

src/cli/term2.rs
aloucks: > undoing the performance work

Could you elaborate? Is there something to measure? I guess I don't consider any of this to be performance sensitive code and I didn't notice any sluggishness with the output (in Windows 10).

src/cli/term2.rs
rbtcollins: Currently rustup doesn't offer a crate; so we can simply hit all code paths that are affected in a single commit like you are if a future change is made. I'd like @kinnison's opinion too, but my view today is that we value less code over future possibilities.

src/cli/term2.rs
rbtcollins: It's IO, on Windows in some cases its up to 4 syscalls per entry into the term codepath if that atty call is being made. Thats significant when we're unpacking up to 60k files per run (20k for stable, beta and unstable each), and also deleting as many for each.

Best way to evaluate regressions will be to measure the total syscalls before and after in both normal and verbose mode.

aloucks: > on Windows in some cases its up to 4 syscalls per entry into the term codepath if that atty call is being made. Thats significant when we're unpacking up to 60k files per run (20k for stable, beta and unstable each), and also deleting as many for each.

What rustup operation are you executing that's producing that much output? I've never seen anywhere near 60k lines. I was usually uninstalling and reinstalling rust nightly for testing. Can you provide a command line example that I can use as a baseline for comparison?

aloucks:
@rbtcollins Is there any interest in this or should I close it?

rbtcollins:
Sorry, been very busy with life recently; yes I think there is interest. In terms of output - rustup isn't very chatty for users - deliberately, but if you use RUSTUP_DEBUG=1 rustup -v it will output the most it can - which is a couple hundred lines right now. I'm pretty sure in the past we had per-file-debug!() calls, and may again in future too.

aloucks:
@rbtcollins I've pushed a change to cache the terminal. I can resolve the merge conflicts if we're good to move forward.

rbtcollins:
I'll try to get to this over the xmas break

kinnison:
> I'll try to get to this over the xmas break

@rbtcollins That sounds fantastic, thank you.

rbtcollins:
Ok finally got there. 

I've fixed the merge conflicts so you don't have to.

There are two issues I see here.

Firstly, this is failing on windows:
```

test cli::self_update::windows::tests::windows_doesnt_mess_with_a_non_string_path ... FAILED
```

I'm not sure why. 

Secondly, looking at how the caching has been done I need to think about it a bit.

Should we cache as globals, or as state on OSProcess. Probably what you've done is right, but its a thing I need to page in.

aloucks:
I think the terminal locking needs to be refactored to work closer to how it did before. I think there needs to be something like `SharedTerminal::lock() -> SharedTerminalGuard` and then have a `Write` impl for the guard instead (and move all attribute functions to the guard). I'm not sure if this is related to the test failures.

rbtcollins:
I don't think the test failure is due to the lack of locking: if you look at TestProcess you can see there is a Vec<u8> that all writes get stored to, and there is another test `default_toolchain_is_stable` which succeeds. (Thus my saying I'm not sure why the failure is happening :P) 

The locking at the CurrentProcess level was there originally solely to match the stdlibs semantics around stdout/err, so we could pass instances into trait-bound calls that expect that.

0xPoe:
I will take care of this recently because we [ran into a bug](https://internals.rust-lang.org/t/help-verify-rustup-1-26-0-release/18577/6?u=hi-rustin) after bumping the term version.

I will open a new PR to move it forward.

Thanks for your contribution!

PR #2796, opened 2021-06-17
Better documentation for toolchain file backward compatibility

0xPoe:
There is only one toolchain file, should it be better to represent the file as a code block for readability?

### DIFF
--- a/doc/src/overrides.md
+++ b/doc/src/overrides.md
@@ -91,9 +91,14 @@ profile = "minimal"
 The `[toolchain]` section is mandatory, and at least one property must be
 specified. `channel` and `path` are mutually exclusive.
 
-For backwards compatibility, `rust-toolchain` files also support a legacy
+For backwards compatibility, the `rust-toolchain` file also supports a legacy
 format that only contains a toolchain name without any TOML encoding, e.g.
-just `nightly-2021-01-21`. The file has to be encoded in US-ASCII this case
+
+```
+nightly-2021-01-21
+```
+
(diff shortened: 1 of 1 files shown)


doc/src/overrides.md
```diff
@@ -91,9 +91,14 @@ profile = "minimal"
 The `[toolchain]` section is mandatory, and at least one property must be
 specified. `channel` and `path` are mutually exclusive.
 
-For backwards compatibility, `rust-toolchain` files also support a legacy
+For backwards compatibility, the `rust-toolchain` file also supports a legacy
```
rbtcollins, line 94: this change isn't correct as I understand english grammar. `files` is correct because we are not talking about a single instance but about the class, like `people`. `people support a government` would be an similar construction.

doc/src/overrides.md
```diff
@@ -91,9 +91,14 @@ profile = "minimal"
...
 format that only contains a toolchain name without any TOML encoding, e.g.
-just `nightly-2021-01-21`. The file has to be encoded in US-ASCII this case
+
+```
+nightly-2021-01-21
+```
+
+The file has to be encoded in US-ASCII this case
```
rbtcollins, line 101: I could go either way here, I'm not sure what the block format gains us. If we are switching to that, lets either put all the information about it before the block, rather than splitting it around the block. It is so simple I would be inclined to keep it inline, unless we have evidence of it being systematically confusing.

0xPoe: Got it. Let me close it.

Issue #1397, opened 2018-04-17
`rustup show` should not force-install the default toolchain if it is not installed.


johnthagen:
This behavior confused me:

```
$ rustup uninstall nightly-2018-04-01-x86_64-apple-darwin
info: uninstalling toolchain 'nightly-2018-04-01-x86_64-apple-darwin'
info: toolchain 'nightly-2018-04-01-x86_64-apple-darwin' uninstalled
$ rustup show
Default host: x86_64-apple-darwin

info: syncing channel updates for 'nightly-2018-04-01-x86_64-apple-darwin'
info: latest update on 2018-04-01, rust version 1.26.0-nightly (517f24025 2018-03-31)
info: downloading component 'rustc'
 57.3 MiB /  57.3 MiB (100 %)   7.8 MiB/s ETA:   0 s                
info: downloading component 'rust-std'
 46.2 MiB /  46.2 MiB (100 %)   7.4 MiB/s ETA:   0 s                
info: downloading component 'cargo'
  3.0 MiB /   3.0 MiB (100 %)   1.5 MiB/s ETA:   0 s                
info: downloading component 'rust-docs'
  6.8 MiB /   6.8 MiB (100 %)   3.1 MiB/s ETA:   0 s                
info: installing component 'rustc'
info: installing component 'rust-std'
info: installing component 'cargo'
info: installing component 'rust-docs'
installed toolchains
--------------------

stable-x86_64-apple-darwin
nightly-i686-apple-darwin
nightly-x86_64-apple-darwin (default)

active toolchain
----------------

nightly-2018-04-01-x86_64-apple-darwin (directory override for '/Users/user/GitHub/unrust')
rustc 1.26.0-nightly (517f24025 2018-03-31)
```

Is `rustup show` supposed to actually start installing components? I was under the impression that it was just for listing out what you have installed.

I'm on a macOS 10.12.6 host.

kinnison:
The behaviour here comes from the fact that the default toolchain was uninstalled.  Rustup forces the default toolchain to exist in many places semi-automatically.  This is done to reduce surprise in a number of places, but has the side-effect of being surprising here.  Perhaps we need an `info: Default toolchain not installed, attempting to install...` which would make it clear why it's happening?

johnthagen:
@kinnison Personally I think a command like `show` should never actually install something and change the user's environment. That just seems very odd to me and I can't think of another example of a tool that does something like that.

kinnison:
[...]

PR #2797, opened 2021-06-17
Document automatic installation behavior via `rustup show`

### LINKED ISSUE #2639: Document autoinstallation behaviour on overrides page of docs
In some cases it can confuse a small but not insignificant number of people that the override files can cause autoinstallation of toolchains, and in other cases people are not sure *what* command to use to trigger this activity best.

We should document the autoinstallation behaviour as part of https://rust-lang.github.io/rustup/overrides.html and we should make recommendations about the "best" way to trigger it, for example so that CI providers can implement it.

0xPoe:
close https://github.com/rust-lang/rustup/issues/2639

### DIFF
--- a/doc/src/overrides.md
+++ b/doc/src/overrides.md
@@ -19,7 +19,9 @@ the directory tree toward the filesystem root, and a `rust-toolchain.toml` file
 that is closer to the current directory will be preferred over a directory
 override that is further away.
 
-To verify which toolchain is active use `rustup show`.
+To verify which toolchain is active, you can use `rustup show`, 
+which will also try to install the corresponding
+toolchain if the current one has not been installed according to the above rules.
 
 [toolchain]: concepts/toolchains.md
 [toolchain override shorthand]: #toolchain-override-shorthand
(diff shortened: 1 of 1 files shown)


rbtcollins:
I'm torn on this - I haven't even read the patch yet, but see https://github.com/rust-lang/rustup/issues/2686 and https://github.com/rust-lang/rustup/issues/1397

This patch is documenting things we consider bugs. It is good to have the current behaviour documented, OTOH then folk may remember what they read and not-re-read the docs after we fix those bugs. perhaps a note to the effect that this is under review or something, and a link to the bugs?

0xPoe:
> perhaps a note to the effect that this is under review or something, and a link to the bugs?

Add a note for it. Thanks for the review, I didn't notice that it was considered a bug. 🤣

0xPoe:
Also, I'm not sure that if it's considered a bug, then we should still document it. Because @kinnison his original idea was to let people know about the behavior and use it in CI. But if it is a bug, will it have a big impact on users when it is fixed?

rbtcollins:
yeah, CI users will use 'rustup ensure' or whatever verb we end up on.'

0xPoe:
@rbtcollins @kinnison Should I close the PR or is this clarification acceptable for now? (I don't currently recommend it in the description for use in CI)

PR #2805, opened 2021-06-24
Mention the rustdoc in components.md

### LINKED ISSUE #2801: Clarify in which component and/or profile `rustdoc` is
It is not clear from https://github.com/rust-lang/rustup/blob/master/doc/src/concepts/components.md from which component `rustdoc` comes from, and therefore it is not clear that it is included in the minimal profile.

Clarify in https://github.com/rust-lang/rustup/blob/master/doc/src/concepts/components.md from which component `rustdoc` comes from.

Perhaps also clarify in https://github.com/rust-lang/rustup/blob/master/doc/src/concepts/profiles.md that the components in the minimal profile also include other tools.

0xPoe:
close https://github.com/rust-lang/rustup/issues/2801

### DIFF
--- a/doc/src/concepts/components.md
+++ b/doc/src/concepts/components.md
@@ -30,7 +30,7 @@ Most components have a target-triple suffix, such as
 The set of available components may vary with different releases and
 toolchains. The following is an overview of the different components:
 
-* `rustc` — The Rust compiler.
+* `rustc` — The Rust compiler and [Rustdoc].
 * `cargo` — [Cargo] is a package manager and build tool.
 * `rustfmt` — [Rustfmt] is a tool for automatically formatting code.
 * `rust-std` — This is the Rust [standard library]. There is a separate
@@ -76,6 +76,7 @@ details.
 [build-std]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#build-std
 [miri]: https://github.com/rust-lang/miri/
 [RLS]: https://github.com/rust-lang/rls
+[rustdoc]: https://doc.rust-lang.org/rustdoc/
 [cargo]: https://doc.rust-lang.org/cargo/
 [clippy]: https://github.com/rust-lang/rust-clippy
 [LLVM]: https://llvm.org/


doc/src/concepts/components.md
```diff
@@ -76,6 +76,7 @@ details.
 [build-std]: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#build-std
 [miri]: https://github.com/rust-lang/miri/
 [RLS]: https://github.com/rust-lang/rls
+[rustdoc]: https://github.com/rust-lang/rust/tree/master/src/tools/rustdoc
```
ehuss: I think this should link to the rustdoc documentation at https://doc.rust-lang.org/rustdoc/

0xPoe: I'm not sure exactly what links need to be used here, if the emphasis is on internal tools then the current links might be better? If the emphasis is on rusctdoc itself maybe a link to the documentation would be better?

ehuss: The intent here is to provide information for a user of rustup to learn what each component is.  Some of these (like miri or rustfmt) link to github because they don't have any other documentation.

0xPoe: Changed. Thanks for your review!

doc/src/concepts/components.md
```diff
@@ -30,7 +30,7 @@ Most components have a target-triple suffix, such as
 The set of available components may vary with different releases and
 toolchains. The following is an overview of the different components:
 
-* `rustc` — The Rust compiler.
+* `rustc` — The Rust compiler, including its internal tool [rustdoc].
```
jyn514: I do not think this is a good description. Rustdoc is a user-facing tool.
```suggestion
* `rustc` — The Rust compiler and [Rustdoc].
```

jyn514 (changes requested):

0xPoe:
Commits squashed.

cc: @kinnison @rbtcollins  Cloud you please take a look? Thanks!

PR #2808, opened 2021-06-26
Make all dist functions and structs crate-private and cleanup unused functions

0xPoe:
part of #2730

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,7 +2,7 @@ use std::env;
 
 include!("src/dist/triple.rs");
 
-pub fn from_build() -> Result<PartialTargetTriple, String> {
+fn from_build() -> Result<PartialTargetTriple, String> {
     let triple = if let Ok(triple) = env::var("RUSTUP_OVERRIDE_BUILD_TRIPLE") {
         triple
     } else {
--- a/src/dist/config.rs
+++ b/src/dist/config.rs
@@ -4,8 +4,8 @@ use super::manifest::Component;
 use crate::errors::*;
 use crate::utils::toml_utils::*;
 
-pub const SUPPORTED_CONFIG_VERSIONS: [&str; 1] = ["1"];
-pub const DEFAULT_CONFIG_VERSION: &str = "1";
+pub(crate) const SUPPORTED_CONFIG_VERSIONS: [&str; 1] = ["1"];
+pub(crate) const DEFAULT_CONFIG_VERSION: &str = "1";
 
 #[derive(Clone, Debug)]
 pub struct Config {
@@ -14,7 +14,7 @@ pub struct Config {
 }
 
 impl Config {
-    pub fn from_toml(mut table: toml::value::Table, path: &str) -> Result<Self> {
+    pub(crate) fn from_toml(mut table: toml::value::Table, path: &str) -> Result<Self> {
         let config_version = get_string(&mut table, "config_version", path)?;
         if !SUPPORTED_CONFIG_VERSIONS.contains(&&*config_version) {
             bail!(RustupError::UnsupportedVersion(config_version));
@@ -29,7 +29,7 @@ impl Config {
             components,
         })
     }
-    pub fn into_toml(self) -> toml::value::Table {
+    pub(crate) fn into_toml(self) -> toml::value::Table {
         let mut result = toml::value::Table::new();
         result.insert(
             "config_version".to_owned(),
@@ -42,12 +42,12 @@ impl Config {
         result
     }
 
-    pub fn parse(data: &str) -> Result<Self> {
+    pub(crate) fn parse(data: &str) -> Result<Self> {
         let value = toml::from_str(data).context("error parsing manifest")?;
         Self::from_toml(value, "")
     }
 
-    pub fn stringify(self) -> String {
+    pub(crate) fn stringify(self) -> String {
         toml::Value::Table(self.into_toml()).to_string()
     }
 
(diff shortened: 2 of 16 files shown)


build.rs
```diff
@@ -2,7 +2,7 @@ use std::env;
 
 include!("src/dist/triple.rs");
 
-pub fn from_build() -> Result<PartialTargetTriple, String> {
+pub(crate) fn from_build() -> Result<PartialTargetTriple, String> {
```
kinnison: This is `build.rs` in reality this doesn't need to be `pub` at all.

src/dist/dist.rs
```diff
@@ -106,11 +106,11 @@ struct ParsedToolchainDesc {
...
 #[derive(Debug, Clone)]
-pub struct PartialToolchainDesc {
+pub(crate) struct PartialToolchainDesc {
     // Either "nightly", "stable", "beta", or an explicit version number
-    pub channel: String,
-    pub date: Option<String>,
-    pub target: PartialTargetTriple,
+    pub(crate) channel: String,
+    pub(crate) date: Option<String>,
+    pub(crate) target: PartialTargetTriple,
```
kinnison: I'm confused, I thought we'd agreed this kind of change was unnecessary.

src/dist/config.rs
```diff
@@ -4,17 +4,17 @@ use super::manifest::Component;
...
+pub(crate) const SUPPORTED_CONFIG_VERSIONS: [&str; 1] = ["1"];
+pub(crate) const DEFAULT_CONFIG_VERSION: &str = "1";
 
 #[derive(Clone, Debug)]
-pub struct Config {
-    pub config_version: String,
-    pub components: Vec<Component>,
+pub(crate) struct Config {
+    pub(crate) config_version: String,
+    pub(crate) components: Vec<Component>,
```
kinnison: Did we not decide that `pub(crate)` on fields means basically nothing in a `pub(crate)` struct?

src/dist/temp.rs
```diff
@@ -4,21 +4,21 @@ use std::io;
...
     #[error("could not create temp root {}" ,.0.display())]
-    CreatingRoot(PathBuf),
+    Root(PathBuf),
     #[error("could not create temp file {}",.0.display())]
-    CreatingFile(PathBuf),
+    File(PathBuf),
     #[error("could not create temp directory {}",.0.display())]
-    CreatingDirectory(PathBuf),
+    Directory(PathBuf),
 }
```
kinnison, line 22: I do not like having this name change conflated with a `pub(crate)` refactor.  Also I'm not convinced by it anyway.  The variant names now do not read intuitively.  `CreatingError::Root` has the concepts of 'create' and 'error' reversed in an english grammatical sense.  `Error::CreatingRoot` makes much more sense.

src/dist/triple.rs
```diff
@@ -47,14 +47,14 @@ static LIST_ENVS: &[&str] = &[
...
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct PartialTargetTriple {
-    pub arch: Option<String>,
-    pub os: Option<String>,
-    pub env: Option<String>,
+pub(crate) struct PartialTargetTriple {
+    pub(crate) arch: Option<String>,
+    pub(crate) os: Option<String>,
+    pub(crate) env: Option<String>,
```
kinnison: Again, confused by `pub(crate) $field` inside a `pub(crate) struct`.

src/dist/temp.rs
0xPoe: I changed it because I saw clippy's warning when I modified the code. Please see: https://rust-lang.github.io/rust-clippy/master/#enum_variant_names

0xPoe:
@kinnison Thanks for your review!

Regarding the use of pub(crate) in the pub(crate) structure I understand that it is useful? Because the visibility of the fields of a pub(crate) structure is private by default and cannot be accessed by other mods in the same crate. So we can use pub(crate)? If we use pub, then other crates will be able to use it. This would result in the accessibility being amplified?

An example of a compilation error that does not use pub(crate): https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ece1d5cb3da661daf651d6dbd575d21b

An example of a successful compilation using pub(crate): https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5dc0ff67c55081d1ffec5fd0666bc6f3

I'm not sure if I'm understanding this correctly?

0xPoe:
@kinnison ping~ 

Could you please take a look? Thanks!

kinnison:
Hi, the thing is that with the struct being `pub(crate)` there's no difference in meaning between `pub` and `pub(crate)` on the fields.  Except the latter implies that the crate has privileged and special knowledge of the interior of the structure which frankly it doesn't.  I'd prefer `pub(crate)` on the struct, and then just plain `pub` on the fields.  There can be no access amplification because nothing outside of the crate can address the struct type at all.

0xPoe:
> Hi, the thing is that with the struct being `pub(crate)` there's no difference in meaning between `pub` and `pub(crate)` on the fields. Except the latter implies that the crate has privileged and special knowledge of the interior of the structure which frankly it doesn't. I'd prefer `pub(crate)` on the struct, and then just plain `pub` on the fields. There can be no access amplification because nothing outside of the crate can address the struct type at all.

I modified them all to pub. Thanks for you review!

PR #2811, opened 2021-06-30
Add rust-gdbgui to the list of proxied tools

willbush:
Attempts to complete https://github.com/rust-lang/rustup/pull/1950 which was
closed due to inactivity.

It appears that everywhere the rust-gdb wrapper script is available so is the rust-gdbgui.
https://github.com/rust-lang/rust/blob/a435b49e86d16e98dcc6595dd471f95e823f41aa/src/bootstrap/dist.rs#L513


kinnison (approved):
I apologise for how long this has sat here.  It looks good to me.

PR #2814, opened 2021-07-01
ci: verify checksum of sccache

pierwill:
Adds a checksum verification to this CI script.


kinnison:
I think this has broken our CI somewhat -- https://github.com/rust-lang/rustup/runs/2978245107?check_suite_focus=true#step:16:496

PR #2815, opened 2021-07-02
Detect '-y' short option in rustup-init.sh more robustly

### LINKED ISSUE #2813: When `-y` is combined with other short options, rustup-init cannot be used without a terminal.
**Problem**
For example, it cannot be invoked with `-qy`, but rather requires `-q -y`.

This is because before the options are passed to the main binary, they are scanned with a simplistic technique to detect whether an interactive tty is required:

```
    for arg in "$@"; do
        case "$arg" in
            -h|--help)
                usage
                exit 0
                ;;
            -y)
                # user wants to skip the prompt -- we don't need /dev/tty
                need_tty=no
                ;;
            *)
                ;;
        esac
    done
```

If `need_tty` is not cleared, the script will refuse to continue running unless stdout is a terminal.

**Steps**

1. `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -qy > /dev/null` and note the failure.
2. `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -qy` and note that confirmation was _not_ asked for; i.e., the underlying binary does indeed understand that the `y` option has been specified, so the failure in step 1 was overly restrictive.
3.

**Possible Solution(s)**
Use `getopts` in the script; e.g. https://stackoverflow.com/a/28466267/242814

**Notes**

Output of `rustup --version`: irrelevant.
Output of `rustup show`: irrelevant.

umanwizard:
Fixes #2813

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -90,15 +90,32 @@ main() {
     local need_tty=yes
     for arg in "$@"; do
         case "$arg" in
-            -h|--help)
+            --help)
                 usage
                 exit 0
                 ;;
-            -y)
-                # user wants to skip the prompt -- we don't need /dev/tty
-                need_tty=no
-                ;;
             *)
+                OPTIND=1
+                if [ "${arg%%--*}" = "" ]; then
+                    # Long option (other than --help);
+                    # don't attempt to interpret it.
+                    continue
+                fi
+                while getopts :hy sub_arg "$arg"; do
+                    case "$sub_arg" in
+                        h)
+                            usage
+                            exit 0
+                            ;;
+                        y)
+                            # user wants to skip the prompt --
+                            # we don't need /dev/tty
+                            need_tty=no
+                            ;;
+                        *)
+                            ;;
+                        esac
+                done
                 ;;
         esac
     done


kinnison:
Thank you for making this change, before I review it fully, I notice it uses the `getopts` shell builtin -- are you confident this will work with `/bin/sh` on BSDs and Illumos (i.e. ksh and friends)?  If so then good, but if you can't be sure then we need to find some people to test it because we don't have anything to that effect in our CI currently really.

umanwizard:
This uses pure POSIX shell features, so it should work on any Unix-like OS.

I have tested on FreeBSD and it works fine there.

I was not able to test the full script on Illumos because the version I tried does not support rustup at all, but I have tested `getopts` and shell substitution syntax there and verified that they work the way POSIX specifies, so I am confident that the script will work there too.

kinnison:
Okay thanks for doing that check.

rustup-init.sh
```diff
@@ -90,15 +90,32 @@ main() {
...
-                need_tty=no
-                ;;
             *)
+                OPTIND=1
+                if [ "${arg%%--*}" = "" ]; then
+                    # Long option (other than --help);
+                    # don't attempt to interpret it.
+                    continue
+                fi
+                while getopts :hy sub_arg "$arg"; do
```
kinnison, line 104: Can you confirm if this will work in combination with long options?  i.e. if we had `--profile=blah -y --default-toolchain=whatever` would that work?

umanwizard: Yes. The above logic will skip those.

```
if [ "${arg%%--*}" = "" ]; then
    # Long option (other than --help);
    # don't attempt to interpret it.
    continue
fi
```

The test `if [ "${arg%%--*}" = "" ]` simply checks whether the first two characters of the argument are `--`, and we skip checking for -y and -h if so.

umanwizard:
Ping, @kinnison , can this be merged?

kinnison (approved):

kinnison:
Yep, thank you for your contribution.

PR #2816, opened 2021-07-03
Correct mistake in musl sccache fetcher

kinnison:
This PR primarily exists to fix a bug exposed by #2814 but also it adds a `.dockerignore` file because without it we send all of `.git` and in the case of a normal working tree also `target` to the docker daemon which is slow and unnecessary.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+target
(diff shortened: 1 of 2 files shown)


kinnison:
@pierwill can you please check I've not broken what you were trying to achieve?

rbtcollins (approved):

pierwill:
> @pierwill can you please check I've not broken what you were trying to achieve?

Gave it a try locally. Looks good to me!

Thanks for the fix 🙏

Issue #1772, opened 2019-04-17
Visual Studio 2019


TjeuKayim:
**Describe the problem you are trying to solve**

Rustup prompts for installing the Microsoft C++ build tools on Windows. It refers to Visual Studio 2017, and links to https://aka.ms/buildtools. I noticed that this information is outdated.

On 2 April 2019, Visual Studio 2019 was released. The link https://aka.ms/buildtools is pointing to https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2017, but that anchor does not exist anymore.

https://github.com/rust-lang/rustup.rs/blob/8f1d93edbd037fbc849e72c7aef66a61146ee458/src/cli/self_update.rs#L175-L190

**Describe the solution you'd like**

The instructions printed by the CLI need an update. The help text should refer to Visual Studio 2019. Instructions to install 'Visual C++ Build Tools 2017' should be removed. It seems like the official downloads are no longer available.

Maybe the package manager chocolatey can be mentioned as an alternative to acquire the build-tools.

**Notes**
Steps to install build tools 2019:
1. Visit https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019
1. Download "Build Tools for Visual Studio 2019"
1. Start the installer and enable the checkbox "C++ build tools"
1. Run rustup-init.exe

I verified this process on a clean Windows 10 machine. Afterwards Cargo runs successfully.

Screenshot of Visual Studio 2019 installer:

![afbeelding](https://user-images.githubusercontent.com/15987676/56306661-58153a00-6143-11e9-8b3a-82051451c593.png)

TjeuKayim:
I just noticed that the content at 'https://www.rust-lang.org/tools/install' is already updated for Visual Studio 2019.
https://github.com/rust-lang/www.rust-lang.org/pull/736

Could you please have a look at this @yoshuawuyts?

I searched the Github oranisation rust-lang for "Visual Studio 2017", to find that the Rust Book also needs an update.
https://github.com/rust-lang/book/pull/1923

yoshuawuyts:
@TjeuKayim thanks for the ping and raising this! -- I don't maintain rustup, but I'm sure someone from the tools team will be able to help with this soon!

kinnison:
@TjeuKayim Thanks for the issue, and the PR.  I've asked on the PR for you to also update our `README.md` appropriately so that the two parts are not contradicting one another.

TjeuKayim:
[...]

PR #2822, opened 2021-07-11
Update help text to align with new VS Build Tools installer

thadguidry:


TjeuKayim:
I also ran into this, the visual studio installer changed again, this pull request would indeed clear up the confusion.

As pointed out in https://github.com/rust-lang/rust/issues/43039#issuecomment-589321158, It is also possible to use the tab Single Components and selecting these items: “MSVC v142 - VS 2019 C++ x64/x86 build tools” and “Windows 10 SDK (1.0.18362.0)”. Only selecting these two saves 4.51 GB disk space compared to selecting the "Desktop development with C++" workload. So, maybe it's good to mention this alternative in the additional instructions you wrote.

Related https://github.com/rust-lang/rustup/issues/1772#issuecomment-869030067

src/cli/self_update.rs
```diff
@@ -282,8 +282,9 @@ tools:
 
     https://visualstudio.microsoft.com/visual-cpp-build-tools/
 
-Please ensure the Windows 10 SDK and the English language pack components
-are included when installing the Visual C++ Build Tools.
+Check its box for "Desktop development with C++" which will ensure that
+the Windows 10 SDK is selected.  If your locale language is not English,
+then additionally check the box for English under Language packs.
```
thadguidry:
```suggestion
then additionally check the box for English under Language packs.  Or minimally, use the tab Single Components and select these items: “MSVC - VS 2019 C++ x64/x86 build tools” and “Windows 10 SDK”
```

thadguidry: @TjeuKayim So like above?

TjeuKayim: Nice, that's what I meant.

src/cli/self_update.rs
```diff
@@ -282,8 +282,9 @@ tools:
 
     https://visualstudio.microsoft.com/visual-cpp-build-tools/
 
-Please ensure the Windows 10 SDK and the English language pack components
-are included when installing the Visual C++ Build Tools.
+Check its box for "Desktop development with C++" which will ensure that
+the Windows 10 SDK is selected.  If your locale language is not English,
+then additionally check the box for English under Language packs.  Or minimally, use the tab Single Components and select these items: “MSVC - VS 2019 C++ x64/x86 build tools” and “Windows 10 SDK”
```
TjeuKayim: I just noticed that you used unicode characters `“..”` for the left and right quote. This is inconsistent with the rest of the text that uses the ASCII `".."`

PR #2829, opened 2021-07-19
Make all cli functions and structs crate-private

0xPoe:
part of #2730

### DIFF
--- a/src/cli/markdown.rs
+++ b/src/cli/markdown.rs
@@ -11,7 +11,7 @@ struct LineWrapper<'a, T: Terminal> {
     indent: u32,
     margin: u32,
     pos: u32,
-    pub w: &'a mut T,
+    w: &'a mut T,
 }
 
 impl<'a, T: Terminal + 'a> LineWrapper<'a, T> {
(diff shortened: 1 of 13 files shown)


0xPoe:
@kinnison Could you please take a look? Thanks!

rbtcollins:
@hi-rustin have you checked that there is no lint on Windows builds?

0xPoe:
> @hi-rustin have you checked that there is no lint on Windows builds?

I checked the CI relevant build log and found no warnings, do I still need to check it manually?

0xPoe:
@rbtcollins @kinnison Do you have time to help look at it? Thanks!

src/cli/markdown.rs
```diff
@@ -11,7 +11,7 @@ struct LineWrapper<'a, T: Terminal> {
     indent: u32,
     margin: u32,
     pos: u32,
-    pub w: &'a mut T,
+    pub(crate) w: &'a mut T,
```
rbtcollins: this is a public attribute in a private struct. What happens if you just make the attribute non-public?

0xPoe: Addressed. Good catch! Thank you.

PR #2832, opened 2021-07-22
Provide a better error when CWD is not available.

ehuss:
This adds some context when rustup is unable to determine the current directory. Previously it would give a cryptic error:

```
error: No such file or directory (os error 2)
```

This should update it to provide an explanation:

```
error: Unable to proceed. Could not locate working directory.: No such file or directory (os error 2)
```

### DIFF
--- a/src/bin/rustup-init.rs
+++ b/src/bin/rustup-init.rs
@@ -59,7 +59,7 @@ fn run_rustup_inner() -> Result<utils::ExitCode> {
 
     // Before we do anything else, ensure we know where we are and who we
     // are because otherwise we cannot proceed usefully.
-    process().current_dir()?;
+    utils::current_dir()?;
     utils::current_exe()?;
 
     // The name of arg0 determines how the program is going to behave
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -475,7 +475,7 @@ pub(crate) fn make_executable(path: &Path) -> Result<()> {
     inner(path)
 }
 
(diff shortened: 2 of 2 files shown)


ehuss:
I'm a little confused, as there seems to be three different ways to get the current directory (`env::current_dir`, `utils::current_dir`, and `process().current_dir()`), and they seem to be used a little inconsistently, so I'm not sure how this should work.

kinnison:
`process().current_dir()` should be used in application code.  `utils::current_dir()` does that, but also provides the context around looking for the CWD.

https://github.com/rust-lang/rustup/blob/master/src/utils/utils.rs#L478-L482

the only use of `env::current_dir` I can see which might be faulty is https://github.com/rust-lang/rustup/blob/master/src/cli/rustup_mode.rs#L90 which should probably be a `utils::current_dir()` in reality.

src/bin/rustup-init.rs
```diff
@@ -59,7 +60,9 @@ fn run_rustup_inner() -> Result<utils::ExitCode> {
 
     // Before we do anything else, ensure we know where we are and who we
     // are because otherwise we cannot proceed usefully.
-    process().current_dir()?;
+    process()
+        .current_dir()
+        .context(RustupError::LocatingWorkingDir)?;
```
kinnison: This is essentially the content of `utils::current_dir()`

ehuss: Ah, indeed, updated.

0xPoe (approved):
LGTM.

PR #2834, opened 2021-08-03
Update cross-compilation.md

zoodirector:
Added information how to configure the linker for a specific target.

### DIFF
--- a/doc/src/cross-compilation.md
+++ b/doc/src/cross-compilation.md
@@ -30,6 +30,13 @@ installing the NDK components as well.
 
 [Android NDK]: https://developer.android.com/tools/sdk/ndk/index.html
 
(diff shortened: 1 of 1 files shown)


kinnison:
I'm sorry that it has taken so long for me to come back to this - I kinda feel like this doesn't entirely belong in the rustup docs, since that's Cargo configuration really.  Is there some useful bit of the cargo docs we could link to instead?  (If not, I'm not 100% against merging this hint, but if it's possible to refer the user somewhere else instead then that'd be better since they'll learn more than just this hint).

zoodirector:
Thanks @kinnison, you are right! I have added the link to the cargo book instead.

kinnison (approved):

Issue #2612, opened 2020-12-17
aarch64 windows


zhmushan:
Could we build an installer for `aarch64-pc-windows-msvc` ? (I'm working on Surface Pro X)

kinnison:
In theory we could.  std and host are built for this target.  We need to know by what means (cross from Windows x86_64? cross from Linux?) and for that I shall request a hint from @pietroalbini on how we build that target's rustc/cargo.

emilyalbini:
For rustc/cargo we're cross-compiling from Windows x86_64.

kinnison:
Cool, would you expect GH Actions to just work if I `rustup target install aarch64-pc-windows-msvc` and then `cargo build --target=aarch64-pc-windows-msvc` ?  If so I should be able to add the action pretty easily.

emilyalbini:
I'm not sure exactly, I was not involved in setting that CI up. I guess trying won't hurt?

kinnison:
Probably wouldn't hurt, no.  I'll see if I can knock up a PR to check

kinnison:
A brief test suggests this will be a significantly bigger job than thought.

Hyper doesn't support tokio 0.3 yet.  tokio 0.2 relies on mio 0.6 - mio 0.6 relies on winapi 0.2.

I believe the fixes for aarch64 didn't make it into winapi 0.2 so we'd need to fix hyper to work on tokio 0.3 (and redo our test suite with that, which is less hassle) in order to eliminate incompatible winapi deps.

kinnison:
```
tokio v0.2.24
├── h2 v0.2.7
│   └── hyper v0.13.9
│       ├── hyper-rustls v0.21.0
│       │   └── reqwest v0.10.10
│       │       └── download v0.6.8 (/home/dsilvers/dev-git/rustup/download)
│       │           └── rustup v1.23.1 (/home/dsilvers/dev-git/rustup)
│       ├── hyper-tls v0.4.3
│       │   └── reqwest v0.10.10 (*)
│       └── reqwest v0.10.10 (*)
├── hyper v0.13.9 (*)
├── hyper-rustls v0.21.0 (*)
├── hyper-tls v0.4.3 (*)
├── reqwest v0.10.10 (*)
├── tokio-rustls v0.14.1
│   ├── hyper-rustls v0.21.0 (*)
│   └── reqwest v0.10.10 (*)
├── tokio-socks v0.3.0
│   └── reqwest v0.10.10 (*)
[...]

PR #2835, opened 2021-08-04
Support building rustup for Windows arm64

Alovchin91:
This PR adds support for building rustup for Windows arm64 using aarch64-pc-windows-msvc toolchain.

Things to note:

* CI templates are present but do not run builds on Windows arm64 yet
* Ring still does not support Windows arm64. See https://github.com/briansmith/ring/issues/1167

See #2612 for discussion

### DIFF
--- a/.github/workflows/windows-builds-on-pr.yaml
+++ b/.github/workflows/windows-builds-on-pr.yaml
@@ -26,8 +26,6 @@ jobs:
           - target: i686-pc-windows-gnu
             mingwdir: mingw32
             mingw: https://ci-mirrors.rust-lang.org/rustc/i686-6.3.0-release-posix-dwarf-rt_v5-rev2.7z
-          - target: aarch64-pc-windows-msvc
-            skip_tests: yes
     steps:
       - uses: actions/checkout@v2
         # v2 defaults to a shallow checkout, but we need at least to the previous tag
--- a/.github/workflows/windows-builds-on-pr.yaml
+++ b/.github/workflows/windows-builds-on-pr.yaml
@@ -26,6 +26,8 @@ jobs:
           - target: i686-pc-windows-gnu
             mingwdir: mingw32
             mingw: https://ci-mirrors.rust-lang.org/rustc/i686-6.3.0-release-posix-dwarf-rt_v5-rev2.7z
+          - target: aarch64-pc-windows-msvc
+            skip_tests: yes
     steps:
       - uses: actions/checkout@v2
         # v2 defaults to a shallow checkout, but we need at least to the previous tag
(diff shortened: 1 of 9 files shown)


Alovchin91:
Test results: (see attachments)

All run under rustc 1.55.0-beta.3 (53fd98ca7 2021-07-31) installed by that same arm64 rustup 🙂

Note: for download tests I had to manually remove rustls-tls (requires ring) from Cargo.toml default features because cargo test doesn't seem to respect --no-default-features/--features flags when passed a -p parameter.

[dump_testament.txt](https://github.com/rust-lang/rustup/files/6934801/dump_testament.txt)
[cargo_test.txt](https://github.com/rust-lang/rustup/files/6934802/cargo_test.txt)
[download_test.txt](https://github.com/rust-lang/rustup/files/6934804/download_test.txt)

.github/workflows/windows-builds-on-pr.yaml
```diff
@@ -26,6 +26,8 @@ jobs:
           - target: i686-pc-windows-gnu
             mingwdir: mingw32
             mingw: https://ci-mirrors.rust-lang.org/rustc/i686-6.3.0-release-posix-dwarf-rt_v5-rev2.7z
+          - target: aarch64-pc-windows-msvc
```
Alovchin91: Oh, oops, it actually did build after all 😅 Nice verification though! Should we then keep it enabled at least for Stable?

kinnison:
When you updated the CI actions, did you update the template and then run the script to regenerate the workflow files?  If so then I need to poke the scripts.  Otherwise this PR looks good.  I'll check that over and then consider merging.

Alovchin91:
> When you updated the CI actions, did you update the template and then run the script to regenerate the workflow files?

Yep, that's what I did. Apparently I also need to mark those parameters in the "include" section with "skip-pr skip-master" otherwise they seem to be picked up by the CI.

kinnison:
I think you must be right - it is over-executing on PRs for sure.  Try adding the markers to the include section too please, yes.

Alovchin91:
Sure! Do you think it makes sense to disable win-arm64 builds for now (since we seem to be making some good progress on \*ring\*: https://github.com/briansmith/ring/pull/1339), or should we still enable them for stable or master, for example?

kinnison:
I'm not sure if we're OK to release until ring is sorted, as such let's build on master, but not pr or stable.

Alovchin91:
👍 Done. Also updated the README.

Alovchin91:
Hi @kinnison Do you think this could be merged now? 🙂

kinnison (approved):

Issue #2189, opened 2020-01-06
rustup.sh fails on /bin/sh on openindiana with local: not found


exorcist365:
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
sh[378]: local: not found [No such file or directory]
sh[170]: local: not found [No such file or directory]
rustup: unrecognized OS type: SunOS

emilyalbini:
Transferred the issue.

kinnison:
Hi @JoanM1lev thanks for contacting us.

At this time, Illumos is tier 2 and only `std` is built.  As such, `rustup` can't be used on that platform.  You can use Linux or another OS and cross-build for that target, but not build natively as far as I understand.

kinnison:
Oh in addition, if `sh` on Illumos isn't POSIX then that'd explain why it failed on `local` and friends.

rbtcollins:
We'd be happy with patches to detect a POSIX sh on openindiana and invoke it appropriately, though that won't help with the lack of built components etc. I've tagged this help wanted to indicate that the rustup team themselves are not going to get to this (while Ilumos is tier 2 anyhow).

jclulow:
NB: As of the latest release of `rustup`, I believe the script now correctly detects the `ksh93` shell that doesn't support `local` and suggests to the user they should use `bash` instead:

```
$ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
rustup does not work with this ksh93 version; please try bash!
```

PR #2839, opened 2021-08-12
Add illumos/OpenIndiana platform check in rustup.js

mdaverde:
On OpenIndiana (illumos), [rustup.rs](https://rustup.rs/) currently shows that it doesn't recognize the platform:

rusup-illumos-not-supported-resized

This change clears the confusion and adds the missing platform check to rustup.js:

rustup-illumos-support-resized

I'm tagging @jclulow to verify if the platform checks added are thorough enough but also to verify that the [curl command itself works on OpenIndiana](https://github.com/rust-lang/rustup/issues/2189#issuecomment-735007324). On OpenIndiana 5.11, I get a wall of `local: not found` errors but piping it into bash does make it work. I can try to look into the script later

Cheers!


kinnison:
I shall merge this, though there remain potential problems with the OI release which need looking into.

PR #2841, opened 2021-08-19
Make all src functions and structs crate-private

0xPoe:
part of https://github.com/rust-lang/rustup/pull/2829

### DIFF
--- a/src/currentprocess.rs
+++ b/src/currentprocess.rs
@@ -191,13 +191,13 @@ impl ProcessSource for OSProcess {
 
 #[derive(Clone, Debug, Default)]
 pub struct TestProcess {
-    pub(crate) cwd: PathBuf,
-    pub(crate) args: Vec<String>,
-    pub(crate) vars: HashMap<String, String>,
-    pub(crate) id: u64,
-    pub(crate) stdin: TestStdinInner,
-    pub(crate) stdout: TestWriterInner,
-    pub(crate) stderr: TestWriterInner,
+    pub cwd: PathBuf,
+    pub args: Vec<String>,
+    pub vars: HashMap<String, String>,
+    pub id: u64,
+    pub stdin: TestStdinInner,
+    pub stdout: TestWriterInner,
+    pub stderr: TestWriterInner,
 }
 
 impl TestProcess {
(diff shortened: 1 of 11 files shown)


0xPoe:
@rbtcollins @kinnison Could you please take a look when you have time? Thanks! ❤️

src/currentprocess.rs
```diff
@@ -191,13 +191,13 @@ impl ProcessSource for OSProcess {
...
 #[derive(Clone, Debug, Default)]
 pub struct TestProcess {
-    pub(crate) cwd: PathBuf,
-    pub(crate) args: Vec<String>,
-    pub(crate) vars: HashMap<String, String>,
-    pub(crate) id: u64,
-    pub(crate) stdin: TestStdinInner,
-    pub(crate) stdout: TestWriterInner,
-    pub(crate) stderr: TestWriterInner,
+    pub cwd: PathBuf,
```
rbtcollins, line 194: Whats going on with this hunk?

0xPoe: Should we not be able to use pub(crate) in the pub field? In the first few PRs we didn't seem to want to use pub(crate) on the field.

rbtcollins: I'm not judging, I'm trying to understand. `TestProcess` is used in both unit tests and integration tests; thus `TestProcess` must be `pub`. Fields of `TestProcess` are read from what contexts? Our goal with this arc of work is to make it so that we get lint errors if anything is accidentally unused, whether it be a mod, struct/enum, function or field of a struct.

We didn't need a visibility path of `pub struct` -> `pub(crate) field` on internal parts of the system, because we could instead do `pub(crate) struct` -> `pub field`. But the struct here is pub. So a `pub` field could be unused and not warn, right? Try it - add a new unused field with the same visibility and lets see what happens?

0xPoe: Thanks for the comment! ❤️ I will try it out over the weekend.

0xPoe: > Fields of TestProcess are read from what contexts?

These fields are used in unit tests.

> But the struct here is pub. So a pub field could be unused and not warn, right?

Yes. I tested it locally as well.

0xPoe: @rbtcollins Can we merge it now? Do you have any other questions?

0xPoe:
@kinnison Could you please take a look? Thanks!

kinnison (approved):
I think this looks like a good improvement to our ability to see what's going on, thank you!

PR #2845, opened 2021-09-07
Fixes #2807 adds info message if self-update is disabled

### LINKED ISSUE #2807: `rustup update` does NOT report when self-update is disabled
**Problem**

As noted in #2757, `rustup` can't update itself when `rustup` is installed using a system package manager like homebrew. `rustup self update` reports an error in this case:

```
rustup self update
error: self-update is disabled for this build of rustup
error: you should probably use your system package manager to update rustup
```

`rustup update --help` (that is, without `self`) says:

```
rustup-update
Update Rust toolchains and rustup
...
DISCUSSION:
    With no toolchain specified, the `update` command updates each of
    the installed toolchains from the official release channels, then
    updates rustup itself.
```

I expected `rustup update` to report the same `self-update is disabled` error that `rustup self update` reported when `rustup update` tries to update `self`, but instead it silently fails to update `rustup`:

```
rustup update
info: syncing channel updates for 'stable-aarch64-apple-darwin'
info: syncing channel updates for 'beta-aarch64-apple-darwin'
  stable-aarch64-apple-darwin unchanged - rustc 1.53.0 (53cb7b09b 2021-06-17)
    beta-aarch64-apple-darwin unchanged - rustc 1.54.0-beta.1 (bf62f4de3 2021-06-23)
```

Note the lack of `info: checking for self-updates`.

Someone who installed `rustup` through homebrew and has never run `rustup self update` would think that `rustup update` *was* updating `rustup` (they also wouldn't know that usually the output contains `info: checking for self-updates`), and would have trouble figuring out why `rustup` wasn't updating.

**Steps**

1. Install `rustup` with homebrew
2. Run `rustup update`
3. Observe no error message where there should be one and no `info: checking for self-updates` in the output

**Possible Solution(s)**

Show the same error for `rustup update` that `rustup self update` shows when self-update is disabled.

**Notes**

Output of `rustup --version`: rustup 1.24.3 (ce5817a94 2021-05-31)

kushaldas:
If self-update is disabled via features during build time, then
we should better info message during `update` that the `rustup`
itself is not updated.

## How did I test?

- `cargo build --features no-self-update`
- `mkdir home`
- `RUSTUP_HOME=home CARGO_HOME=home target/debug/rustup-init --no-modify-path -y`
- `RUSTUP_HOME=home CARGO_HOME=home ./home/bin/rustup update`

Output from the last command:

```

  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.54.0 (a178d0322 2021-07-26)

info: self-update is disabled for this build of rustup
info: any updates to rustup will need to be fetched with your system package manager
```

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1028,6 +1028,11 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
         check_rustup_update()?;
     }
 
+    if self_update::NEVER_SELF_UPDATE {
+        info!("self-update is disabled for this build of rustup");
+        info!("any updates to rustup will need to be fetched with your system package manager")
+    }
+
     Ok(utils::ExitCode(0))
 }
 


kinnison:
@carols10cents How do you feel about this - do you think it fixes #2807 in a way you think will work for others?

kinnison (approved):

PR #2847, opened 2021-09-08
Port rustup to Sequoia PGP.

teythoon:
(Edited from original)

In order to port Rustup to Sequoia-PGP there are a number of steps remaining:

* [x] Make the CI able to build it at all
* [x] Verify on Windows, MacOS, and other tier one platforms
* [ ] Verify (by adjusting the CI temporarily) that we can build all the other targets too
* [x] Examine licence considerations involved in using Sequoia, likely either Rustup or Sequoia would have to re-licence. (Gathering all contributors to relicence rustup might be significantly difficult, or at least very very lengthy).

If further steps are identified, the above list can be amended.


kinnison:
Thank you for this - naturally there are many things we need to sort out before this could be merged.  I shall edit the top posting to include a checklist.

teythoon:
Ftr, we're [porting to RustCrypto](https://gitlab.com/sequoia-pgp/sequoia/-/commits/justus/rustcrypto).  Due to their own admission, RustCrypto is not ready for production use, but as you used rpgp before, and they also use RustCrypto, you are probably fine with Sequoia backed by RustCrypto.

kinnison:
That is interesting indeed, it'd reduce the complexity of the CI work for sure.  Do you have an ETA on an alpha version with that support?

teythoon:
ETA on the alpha is now.  There are some lxd-related errors that I don't feel responsible for.

kinnison:
Sadly snap isn't always reliable, I've requeued them.

The 32-bit Windows build test is worrying though:

```
   Compiling sha1collisiondetection v0.2.3
LLVM ERROR: Invalid LLVMRustVisibility value!
error: could not compile `sha1collisiondetection`
```

teythoon:
Yeah, I don't know what that is about.  I can report that we do successfully compile for 32-bit Windows: https://gitlab.com/sequoia-pgp/sequoia/-/jobs/1590997603

teythoon:
(Also, sha1collisiondetiction is a crate that we use for all backends, not just for the new RustCrypto one.)

kinnison:
The only difference I can see is 1.55.0 vs. 1.54.0 -- can you try your build with 1.55.0 ?  Maybe there's a compiler bug.

teythoon:
Seems to build the sha1cd crate fine, but fails for different reasons (new warnings being errors in our ci): https://gitlab.com/sequoia-pgp/sequoia/-/jobs/1592512848

kinnison (commented):
I *think* this looks good, but we do want the CI to run.  If necessary I'll see if I can trigger things.

teythoon:
As of sequoia-openpgp 1.5, we're LGPL2.0+.

kinnison:
At minimum, https://github.com/rust-lang/rust/issues/87813 is blocking this because of how sha1collisiondetection breaks the compile.

teythoon:
Now with the workaround developed by Josh Stone in https://bugzilla.redhat.com/show_bug.cgi?id=1990657.

kinnison (commented):
That the full CI is green heartens me greatly.  I need to spend a bit of time with the particular code bits, but I'm on the verge of approval

PR #2849, opened 2021-09-18
Fix duplicated PATH entries

### LINKED ISSUE #2848: Deduplicate PATH entries added during build (e.g. `~/.cargo/bin`)
**Problem**
When running `cargo build` (or `cargo check`, `cargo clippy`), the `$PATH` env var contains `~/.cargo/bin` twice. My `$PATH` contains it once (by sourcing `~/.cargo/env` as per standard rustup install)

This addition to `$PATH` is mostly harmless, except when:

1. A dependent package has a build.rs containing `cargo:rerun-if-env-changed=PATH`
2. User switches between running `cargo build` and using something like `rust-analyzer`

The problem being:

1. As user is editing, `rust-analyzer` is building and writes the "non-duplicated" $PATH into the fingerprinting file
2. User switches to terminal and runs `cargo build` or similar, cargo adds the duplicated entry and the fingerprinter detects that `$PATH` changes and triggers a rebuild
3. All downstream crates of the dependent package trbuild
4. Switching back to rust-analyzer, it also rebuilds for the same reason

In short, the different $PATH causes lots of additional rebuilds (in some cases)

One example of this being pyo3:
https://github.com/PyO3/pyo3/issues/1708
..which does have to re-run if $PATH changes, because a different Python interpreter might be added via a virtualenv

**Steps (simpler)**
Smallest reproduction case I can think of is:

```
dbr:~/pathexample$ cat Cargo.toml 
[package]
name = "pathexample"
version = "0.1.0"
edition = "2018"

build = "build.rs"

[dependencies]
dbr:~/pathexample$ cat build.rs 
fn main(){
    dbg!(env!("PATH"));
    panic!();
}
dbr:~/pathexample$ cargo build
Dupe path found /home/dbr/.cargo/bin
   Compiling pathexample v0.1.0 (/home/dbr/pathexample)
error: failed to run custom build command for `pathexample v0.1.0 (/home/dbr/pathexample)`

Caused by:
  process didn't exit successfully: `/home/dbr/pathexample/target/debug/build/pathexample-2395f9cd1cf04944/build-script-build` (exit code: 101)
  --- stderr
  [build.rs:2] env!("PATH") = "/home/dbr/.cargo/bin:/home/dbr/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games"
  thread 'main' panicked at 'explicit panic', build.rs:3:5
dbr:~/pathexample$ echo $PATH
/home/dbr/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
```

(Noting the `[build.rs:2] env!("PATH")` contains `/home/dbr/.cargo/bin` twice, whereas `echo $PATH` is "correct")

If I bypass rustup (..I think..) by setting $PATH to only include a specific toolchain, the duplication doesn't occur:

```
dbr:~/pathexample$ env PATH=/home/dbr/.rustup/toolchains/1.54-x86_64-unknown-linux-gnu/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games cargo build
   Compiling pathexample v0.1.0 (/home/dbr/pathexample)
error: failed to run custom build command for `pathexample v0.1.0 (/home/dbr/pathexample)`

Caused by:
  process didn't exit successfully: `/home/dbr/pathexample/target/debug/build/pathexample-875cafc10592a6c9/build-script-build` (exit status: 101)
  --- stderr
  [build.rs:2] env!("PATH") = "/home/dbr/.rustup/toolchains/1.54-x86_64-unknown-linux-gnu/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games"
  thread 'main' panicked at 'explicit panic', build.rs:3:5
```

**Steps (real world)**
Larger repro is to:

1. Make a simple crate with `pyo3 = "*"` as a dependency
2. run an editor with `rust-analyzer` and make some trivial changes
3. switch to a terminal and run the following:

```
CARGO_LOG=cargo::core::compiler::fingerprint=info cargo build
```
..and note the lines similar to:
```
INFO  cargo::core::compiler::fingerprint]     err: env var `PATH` changed: previously Some("/home/dbr/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games"), now Some("/home/dbr/.cargo/bin:/home/dbr/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games")
```

**Possible Solution(s)**
:shrug: 

**Notes**

Output of `rustup --version`:

```
$ rustup --version
info: The currently active `rustc` version is `rustc 1.51.0 (2fd73fabe 2021-03-23)`
```

Output of `rustup show`:
```
$ rustup show
rustup home:  /home/dbr/.rustup

installed toolchains
--------------------

nightly-2021-04-15-x86_64-unknown-linux-gnu
nightly-2021-07-05-x86_64-unknown-linux-gnu
1.46-x86_64-unknown-linux-gnu
1.48-x86_64-unknown-linux-gnu
1.51-x86_64-unknown-linux-gnu (default)
1.53-x86_64-unknown-linux-gnu
1.54-x86_64-unknown-linux-gnu

wasm32-unknown-unknown
x86_64-unknown-linux-gnu

active toolchain
----------------

1.51-x86_64-unknown-linux-gnu (default)
rustc 1.51.0 (2fd73fabe 2021-03-23)
```

chansuke:
Fixes #2848

### DIFF
--- a/src/env_var.rs
+++ b/src/env_var.rs
@@ -8,7 +8,7 @@ use crate::process;
 
 pub const RUST_RECURSION_COUNT_MAX: u32 = 20;
 
-/// This can be removed when https://github.com/rust-lang/rust/issues/ 44434 is
+/// This can be removed when https://github.com/rust-lang/rust/issues/44434 is
 /// stablised.
 pub(crate) trait ProcessEnvs {
     fn env<K, V>(&mut self, key: K, val: V) -> &mut Self
--- a/src/env_var.rs
+++ b/src/env_var.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
 use std::env;
 use std::ffi::OsStr;
 use std::path::PathBuf;
@@ -44,7 +44,7 @@ fn append_path<E: ProcessEnvs>(name: &str, value: Vec<PathBuf>, cmd: &mut E) {
     }
 }
 
-pub(crate) fn prepend_path(name: &str, prepend: Vec<PathBuf>, cmd: &mut Command) {
+pub(crate) fn prepend_path<E: ProcessEnvs>(name: &str, prepend: Vec<PathBuf>, cmd: &mut E) {
     let old_value = process().var_os(name);
     let parts = if let Some(ref v) = old_value {
         let mut tail = env::split_paths(v).collect::<VecDeque<_>>();
@@ -222,8 +222,8 @@ mod tests {
                         &Some(
                             env::join_paths(
                                 vec![
-                                    "/home/a/.cargo/bin",
                                     "/home/z/.cargo/bin",
+                                    "/home/a/.cargo/bin",
                                     "/home/b/.cargo/bin"
                                 ]
                                 .iter()
--- a/src/env_var.rs
+++ b/src/env_var.rs
@@ -1,5 +1,6 @@
 use std::collections::VecDeque;
 use std::env;
+use std::ffi::OsStr;
 use std::path::PathBuf;
 use std::process::Command;
 
@@ -7,13 +8,34 @@ use crate::process;
 
 pub const RUST_RECURSION_COUNT_MAX: u32 = 20;
 
+/// This can be removed when https://github.com/rust-lang/rust/issues/ 44434 is
+/// stablised.
+pub(crate) trait ProcessEnvs {
+    fn env<K, V>(&mut self, key: K, val: V) -> &mut Self
+    where
+        Self: Sized,
+        K: AsRef<OsStr>,
+        V: AsRef<OsStr>;
+}
+
+impl ProcessEnvs for Command {
+    fn env<K, V>(&mut self, key: K, val: V) -> &mut Self
+    where
+        Self: Sized,
+        K: AsRef<OsStr>,
+        V: AsRef<OsStr>,
+    {
+        self.env(key, val)
+    }
+}
+
 #[allow(unused)]
-fn append_path(name: &str, value: Vec<PathBuf>, cmd: &mut Command) {
+fn append_path<E: ProcessEnvs>(name: &str, value: Vec<PathBuf>, cmd: &mut E) {
     let old_value = process().var_os(name);
     let mut parts: Vec<PathBuf>;
     if let Some(ref v) = old_value {
-        parts = env::split_paths(v).collect();
-        parts.extend(value);
+        let old_paths: Vec<PathBuf> = env::split_paths(v).collect::<Vec<_>>();
+        parts = concat_uniq_paths(old_paths, value);
     } else {
         parts = value;
     }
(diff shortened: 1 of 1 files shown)


rbtcollins:
This is going to need a test case, I suggest a unit test of some sort close to the code, since the external contract is only changing in the manner of the duplication, and getting at this will be a little tricky.

chansuke:
@rbtcollins Thanks for your comment. I will add some unit test.

chansuke:
I am trying to add a unit test in `tests/cli-misc.rs` but it gives `panicked at 'No process instance'` error. How can I get the result of [cmd.env(name, new_value);](https://github.com/rust-lang/rustup/blob/master/src/env_var.rs#L35) in the test case?
```
    let a = OsString::from("/home/a/.cargo/bin");
    let path_a = PathBuf::from(a);
    let b = OsString::from("/home/b/.cargo/bin");
    let path_b = PathBuf::from(b);
    let c = OsString::from("/home/c/.cargo/bin");
    let path_c = PathBuf::from(c);

    let path_entries = vec![path_a, path_b, path_c];
    let mut cmd = Command::new("test");
    // add duplicated path
    env_var::prepend_path("/home/c/.cargo/bin", path_entries, &mut cmd);
    // trying to get the result of `prepend_path()` with `get_envs()` because `prepend_path` itself doesn't return any value.
    let envs: Vec<(&OsStr, Option<&OsStr>)> = cmd.get_envs().collect();
    // gives `panicked at 'No process instance'` error
    assert_eq!(envs, &[(OsStr::new("wip"), Some(OsStr::new("wip")))]);
```

kinnison:
Unit tests belong in the crate, not in `tests/` which is where we have our integration tests.
There is the `TestProcess` struct in the `currentprocess` module which is provided for this purpose.  You can see its use in a number of unit tests throughout the `src/` tree.  It should be fairly easy to deduce how to construct an appropriate test, but if you can't, then I'm sure Robert (or I) can help further.

chansuke:
I had a wrong understanding. I will try to fix it. Thanks so much for your comment.

chansuke:
I'm using `cmd.get_envs()`, an unstable function to check the result of `prepend_path`, but it fails clippy check on CI. Is there any workaround for this?

src/env_var.rs
```diff
@@ -27,6 +27,8 @@ pub fn prepend_path(name: &str, value: Vec<PathBuf>, cmd: &mut Command) {
     if let Some(ref v) = old_value {
         parts = value;
         parts.extend(env::split_paths(v).collect::<Vec<_>>());
+        parts.sort();
```
rbtcollins: Sorted PATH entries changes the semantic meaning of them: they are not a set, they are a sorted list, because 'bash' in a lower-index entry wins over 'bash' in a higher index entry.

src/env_var.rs
```diff
@@ -45,3 +47,52 @@ pub fn inc(name: &str, cmd: &mut Command) {
...
+                let path_a = PathBuf::from(a);
+                path_entries.push(path_a);
+                // unique PATH.
+                let z = OsString::from("/home/z/.cargo/bin");
+                let path_z = PathBuf::from(z);
+                path_entries.push(path_z);
+
+                prepend_path("PATH", path_entries, &mut cmd);
+
+                let envs: Vec<(&OsStr, Option<&OsStr>)> = cmd.get_envs().collect();
```
rbtcollins: one way you could avoid using this unstable method is to have the code you're testing depend on a trait defined in this module.

Implement it once for Command, and once for a trivial test struct (or use mockall), and then you can assert on what the set env is, rather than interacting with Command at all.

rbtcollins (changes requested):

rbtcollins:
I think this will work but I have a couple of suggestions:
- append_path should be changed likewise to prevent a regression in future.
- value is itself a vec but not deduped
- perhaps a helper function that takes two vecs and returns a unified vec taken in-order from vec 1 and vec 2 and with all duplicates not added to the output.

chansuke:
@rbtcollins Thanks so much for your suggestion. I will fix this.

src/env_var.rs
chansuke: I'm trying to reimplement the unit test but I'm stuck on this part. Which trait should I depend on? 
> a trait defined in this module.

rbtcollins: Define one:
```rust
trait ProcessEnvs {
    type Item;
    type Iter: Iterator;
    fn get_envs(&self) -> Self::Iter<Item=Self::Item>;
    fn env<K, V>(&mut self, key: K, val: V) -> &mut Self
where
    K: AsRef<OsStr>,
    V: AsRef<OsStr>;
}
```

And then in your test you can just use a mock (e.g. with mockall) or a manually constructed test double. Change prepend/append _path to be generic over that trait, and provide an impl for Process.

chansuke:
```
pub fn append_path(
    name: &str,
    value: Vec<PathBuf>,
    cmd: &mut dyn ProcessEnvs<Item = Command, Iter = vec::IntoIter<Command>>,
) {
```
I am stack on `the trait `ProcessEnvs` cannot be made into an object` error. I tried to separate the trait but it didn't work. This is the latest [commit](https://github.com/rust-lang/rustup/pull/2849/commits/afe83e0c0fc194c3be4e492309dd38b5a1b95c99).

chansuke:
Thanks for the support. I am stuck for about a month and would like to close it.

rbtcollins:
Sorry I didn't realise you're stuck. https://internals.rust-lang.org/t/generic-methods-over-object-safe-traits/6774 is the thing to read to understand this - and the reason you needed trait objects is because you didn't make prepend_path generic over an implementation of ProcessEnvs. I've got a fix - should be pushed up in a minute.

However the patch fails tests now :)

rbtcollins:
I've also rebased all the incremental commits together - there isn't enough here for a large commit history. 
@chansuke do you want to finish this off, or should someone else do that?

dmartin:
@chansuke @rbtcollins Hi, I don't have much experience with rustup's internals nor the time to dig too deeply into them right now, but I have been affected by https://github.com/PyO3/pyo3/issues/1708 and would like to help push this across the finish line if possible.

If the only thing blocking this PR from being merged is the failing test on Windows, I submitted a PR with a fix [here](https://github.com/chansuke/rustup/pull/1) (though I'm not sure whether it's considered okay to call `.unwrap()` as I did within tests). Let me know if I can do anything else to help get this merged in!

chansuke:
@rbtcollins @dmartin 
sorry for the late reply I'd like to finish this.

gussmith23:
Wondering what the status of this is -- seems to be blocking https://github.com/PyO3/pyo3/issues/1708 (among other things)

chansuke:
I have requested the review to @rbtcollins

klimburg:
I too have come here from https://github.com/PyO3/pyo3/issues/1708

@rbtcollins / @chansuke what is needed to push this to completion? Is there someone else who can review?

rbtcollins (approved):
wow I had forgotten this. thank you. I'm sure we could code golf it more but the essentials are fine:
- it fixes the bug
- it has a test to prevent accidental recurrence

rbtcollins:
I'm quite confident the android failure is unrelated; will merge this.

Issue #2860, opened 2021-10-06
Feature request: add User-Agent header to http requests


jiegec:
**Describe the problem you are trying to solve**

Hi rustup developers, currently the requests sent by rustup do not include `User-Agent` header. It should include something like `rustup/v1.24.3` or something.

The background is, we are mirroring rustup at https://mirrors.tuna.tsinghua.edu.cn/rustup and we are blocking requests without user-agent headers.

**Describe the solution you'd like**

Set `User-Agent` header in curl and reqwest backend.

**Notes**

takkuumi:
same. rust-static is very slowly in china, i use tuna or ustc mirror, it's broken in last week.

PR #2861, opened 2021-10-06
Add user agent to http requests

jiegec:
Fixing #2860

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -132,6 +132,8 @@ pub fn download_to_path_with_backend(
     })
 }
 
+static USER_AGENT: &str = concat!("rustup/", env!("CARGO_PKG_VERSION"));
+
 /// Download via libcurl; encrypt with the native (or OpenSSl) TLS
 /// stack via libcurl
 #[cfg(feature = "curl-backend")]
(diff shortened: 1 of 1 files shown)


download/src/lib.rs
```diff
@@ -132,6 +132,8 @@ pub fn download_to_path_with_backend(
     })
 }
 
+static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
```
ehuss: Is there a particular reason to use PKG_NAME here?  I think in this case, the name is called "download", which is not very descriptive.

jiegec: Okay, let's use `rustup` instead.

download/src/lib.rs
```diff
@@ -132,6 +132,8 @@ pub fn download_to_path_with_backend(
     })
 }
 
+static USER_AGENT: &str = concat!("rustup/", env!("CARGO_PKG_VERSION"));
```
kinnison, line 135: This is still going to be the `download` package's version, so `0.6.9` currently.

jiegec: How to get the version from outer package?

skyzh: Maybe a custom build.rs 😂😂

kinnison: Or we change the rules so that the download sub-crate gets the same version during the release process.  That'd just need an update to the release process described in `CONTRIBUTING.md`

rbtcollins: Or we merge the crates. Or we pass the value in from the location of use.

jiegec:
Closed in favor of #2953

PR #2862, opened 2021-10-08
Use correct format to load TTF files

sondr3:
I got a bunch of warnings when visiting the website on my Firefox saying `downloadable font: no supported format found`. The root cause is that using `ttf` is an invalid format for `@font-face`, you can [see](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face) the valid specifiers here. However this does change the install box font quite drastically when the font is actually loaded:

### Before

image

### After

image

In my case it was using my preferred monospace font, DejaVu Sans Mono since I do not have the rest of the fonts for the `pre`.

### DIFF
--- a/www/rustup.css
+++ b/www/rustup.css
@@ -21,14 +21,14 @@
     font-family: 'Work Sans';
     font-style: normal;
     font-weight: 500;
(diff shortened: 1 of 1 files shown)


kinnison (approved):
Thank you for this fix

PR #2863, opened 2021-10-08
Beauty changes for install command

sondr3:
This is a slight follow-up from #2862, while changing the fonts I also ended up doing some beauty changes to the command container. Mostly because I didn't like how the scrollbars looked.

### Before

image

### After

image

### After with #2862 

image

Inspiration from the new GHCup website.


kinnison (approved):
That looks much nicer, thank you.

PR #2865, opened 2021-10-11
Symlink rustup-init when in no self update mode

Firstyear:
When we are in no self update configuration, it's likely that we have an externally managed rustup binary. During rustup setup it copies the rustup binary into ~/.cargo/bin/, which then means that rustup will *never* be updated in the future, even if it can update toolchains.

When we have compiled the no self update mode, we can assume external management of rustup and in that situation we should symlink back to rustup instead of copying.

An example use case is that for OpenSUSE we plan to package and provide rustup for developers as an alternate method to get rustup. Because our packages manage the updates of rustup, we want no self update, and we want users who use rustup to link back to the rustup in /usr/bin/rustup which ensures they get updates from their base system. 

Alternately, we could use a difference feature flag for this change rather than no-self-update.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -680,7 +680,11 @@ fn install_bins() -> Result<()> {
     if rustup_path.exists() {
         utils::remove_file("rustup-bin", &rustup_path)?;
     }
-    utils::copy_file(&this_exe_path, &rustup_path)?;
+    if cfg!(feature = "no-self-update") {
+        utils::symlink_file(&this_exe_path, &rustup_path)?;
+    } else {
+        utils::copy_file(&this_exe_path, &rustup_path)?;
+    }
     utils::make_executable(&rustup_path)?;
     install_proxies()
 }
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -165,6 +165,7 @@ Cargo's bin directory, located at:
     {cargo_home_bin}
 
 ",
+            pre_install_msg_linking!(),
             $platform_msg,
             r#"
 
@@ -173,6 +174,26 @@ these changes will be reverted.
 "#
         )
     };
+}
+
+#[cfg(feature = "no-self-update")]
+macro_rules! pre_install_msg_linking {
+    () => {
+        r"
+These commands will be symlinked to `rustup-init`.
+
+`WARNING` - if the path to `rustup-init` changes, your Rust install
+will cease to function.
+
+"
+    }
+}
+
+#[cfg(not(feature = "no-self-update"))]
+macro_rules! pre_install_msg_linking {
+    () => {
+        ""
+    }
 }
 
 #[cfg(not(windows))]
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -306,7 +306,7 @@ pub fn hardlink_file(src: &Path, dest: &Path) -> Result<()> {
 }
 
 #[cfg(unix)]
-fn symlink_file(src: &Path, dest: &Path) -> Result<()> {
+pub(crate) fn symlink_file(src: &Path, dest: &Path) -> Result<()> {
     std::os::unix::fs::symlink(src, dest).with_context(|| RustupError::LinkingFile {
         src: PathBuf::from(src),
         dest: PathBuf::from(dest),
@@ -314,7 +314,7 @@ fn symlink_file(src: &Path, dest: &Path) -> Result<()> {
 }
 
 #[cfg(windows)]
-fn symlink_file(src: &Path, dest: &Path) -> Result<()> {
+pub(crate) fn symlink_file(src: &Path, dest: &Path) -> Result<()> {
     // we are supposed to not use symlink on windows
     Err(anyhow!(RustupError::LinkingFile {
         src: PathBuf::from(src),


vadorovsky:
For more context - we have the rpm package for rustup defined here:

https://build.opensuse.org/package/show/devel:languages:rust/rustup
https://build.opensuse.org/package/view_file/devel:languages:rust/rustup/rustup.spec?expand=1

So we install the binary into `/usr/bin/rustup` (then add `/usr/bin/rustup-init` symlink) which is supposed to be updated only with rpm.

I think the symlink change could be helpful also for the package in AUR.

kinnison:
If you're packaging rustup with no-self-update then you should realistically be providing all the proxies and **not** rustup-init.

Consider how the prototype `snap` works, or how `rustup` is packaged on NixOS.

Firstyear:
> If you're packaging rustup with no-self-update then you should realistically be providing all the proxies and **not** rustup-init.
> 
> Consider how the prototype `snap` works, or how `rustup` is packaged on NixOS.

Sure, but rustup still wants to install itself into the users home directory. Does rustup have a way to account for this so that the "installed" proxy can then be updated? That's more the issue I'm trying to solve here than anything ....

kinnison:
There's a critical bit of functional separation you're not quite grasping here, so perhaps I'm not being clear enough.  I shall attempt to explain it more thoroughly since it can be a tad confusing due to the chimeric nature of the rustup binary.

1. `rustup-init` wants to install proxies into a user's `CARGO_HOME` (typically in their homedir)
2. It does this by copying _itself_ as `rustup` in there, and then hard-linking itself
3. Then those proxies identify what mode to work in based on how they are invoked.  i.e. the same binary handles `rustup` as well as `rustc`, `cargo`, etc.
4. In `no-self-update` mode, `rustup self update` will never run which means it'll never be able to update those proxies itself.
5. As such, **packaged** rustup should work in one of two ways:
6. (a) by packaging `rustup-init` and accepting that after running `rustup-init` the user will no longer be beholden to the packaging system for updates (to fix this case for homebrew, we should ensure that the *content* of `rustup-init` is copied, rather than the symlink that homebrew provides and that the homebrew `rustup-init` is not built with no-self-update)
7. (b) by packaging `rustup` and providing all the proxy binaries as links to that binary.  This way the user never runs `rustup-init` but instead just runs `rustup default stable` or somesuch to get started.  This differs slightly from published getting-started but it wouldn't be hard for packaging systems to explain how to deal with that, and that way the user uses the packaged `rustup` always, it can be built with no-self-update (and indeed should).

Examples of the second mechanism of packaging is the prototype `snap` whose structure can be seen in the `rustup` codebase; and also NixOS's packaging of rustup which you can see here: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/rust/rustup/default.nix

Is that any clearer?

Firstyear:
Hi there, thanks for explaining further. I already understood this process, excepting 7(b). 

I think what I'm proposing here is helpers for 6(a). When you call rustup-init, symlinks are created for the proxies to allow the rustup binary to update, and then the proxies need no further interaction for them to be automatically updated. This PR actually would assist with 6(a) by automating that process so the proxies are softlinks instead. 

This way there is one rustup binary, and we just softlink to it from the required proxy locations allowing a single point of update when in no-self-update mode so an external tool can do the updates.

src/cli/self_update.rs
```diff
@@ -159,8 +159,8 @@ The Cargo home directory located at:
 
 This can be modified with the CARGO_HOME environment variable.
 
-The `cargo`, `rustc`, `rustup` and other commands will be added to
-Cargo's bin directory, located at:
+The `cargo`, `rustc` and other commands will be added to Cargo's
+bin directory, located at:
```
kinnison: This change claims `rustup` will no longer be put into `$CARGO_HOME/bin` but your change still puts it there, via symlink.

src/cli/self_update.rs
```diff
@@ -680,7 +680,11 @@ fn install_bins() -> Result<()> {
     if rustup_path.exists() {
         utils::remove_file("rustup-bin", &rustup_path)?;
     }
-    utils::copy_file(&this_exe_path, &rustup_path)?;
+    if cfg!(feature = "no-self-update") {
+        utils::symlink_file(&this_exe_path, &rustup_path)?;
```
kinnison, line 705: I'm tempted to request that you emit a warning here that we are making a symbolic link not a copy, and as such, if the location of `rustup-init` changes in future, the installation may cease to work.

kinnison (changes requested):
I'm prepared to entertain merging this if it truly makes things easier.  I've commented on the related issue around hardlinks to symlinks breaking a backup program on big-sur.  I still don't understand why you wouldn't just want the packaging to be done properly though.

src/cli/self_update.rs
Firstyear: I have undone this.

src/cli/self_update.rs
Firstyear: I'm okay with a warning. I've added a second commit that put's a warning in the pre-install message. Let me know what you think of it.

Firstyear:
>  I still don't understand why you wouldn't just want the packaging to be done properly though.

Well, in the example you provided of 6(a), then there is no point to packaging rustup-init because the user won't need it after the first run as rustup-init then becomes it's own self contained install. 

The issue with 7(b) is that it's a "different" workflow. Many rust users are used to "rustup-init" as the first step, and even rustup in how it communicates to a user is predominately through rustup-init. For example, "rustup default stable" doesn't provide any of the same messaging about the install and process like rustup-init does. The second risk is that we need to always ship the correct links to each proxy binary which could be added/removed in a way that we miss or are not aware of. 

By leaving rustup responsible for "what components exist" and managing them in the path, it can do the correct thing with the symlinks as required, and then as a platform we just need to update the rustup-init binary that those elements point to. As well leaving rustup-init as the first contact point, we retain the communication and workflow that developers expect when they engage with rust and setup their environment. And finally, it means that rustup-init is upgraded and updated correctly overtime with platform updates which removes one more thing that a user "may" forget to do.

rbtcollins:
I don't understand why 7b being different is bad.

From my perspective the reason for an intermediary shipping rustup at all
is precisely because of a desire to offer a different workflow.

If rustup obtained through e.g. ports is the same as rustup obtained from
upstream, why bother?

Internally, rustup only creates proxies during self update. A no self
update rustup is one that won't create proxies, because it won't reach

 ​        ​
https://github.com/rust-lang/rustup/blob/master/src/cli/self_update.rs#L391

So the packaging system that is shipping rustup is also responsible for
shipping the proxies, and the proxies should not live in the user homedir.

kinnison:
I am inclined to agree with Robert that realistically homebrew should be shipping rustup properly (i.e. as rustup and with all the relevant proxies set up already), or not at all.  Shipping a `rustup-init` just isn't the right thing to do, unless it's a shell script which goes and fetches our installer rather than expecting the user to run `curl` themselves.

Firstyear:
In order to do this then, we need to know what are the definitive list of proxies so that we can programatically create them.

kinnison:
There's no programmatic way to extract that proxy list short of trying `rustup-init` into a temp dir which you could do during packaging if you package on native platforms.  If you need to determine it without executing `rustup-init` then you'd need to accept that the proxy list is fairly static and that we'd inform of changes in the proxy list in the changelog of a release.

The currently proxy list is: `rustc`, `rustdoc`, `cargo`, `rust-lldb`, `rust-gdb`, `rust-gdbgui`, `rls`, `cargo-clippy`, `clippy-driver`, `cargo-miri`, `rustfmt`, `cargo-fmt`.

Firstyear:
Okay, I have changed the package as a result.

kinnison:
@Firstyear If you need any more help with the packaging, let me know and I'll see what I can do to help.

Firstyear:
No problem, thanks. I think a way to list what proxies existed would be the most helpful thing for future, but for now this works.

PR #2869, opened 2021-10-19
Added retry flag support for curl invocations

### LINKED ISSUE #2853: Modify `rustup-init.sh` to support retrying the download of `rustup-init`

I am a newbie trying to install rustup. it is having difficulties on downloading required packages from static.rust-lang.org because the connection constantly drops randomly, it tries to reconnect 3 times all starting from zero (does the host not have resume capability!?, I dont know) then just fails the install progress. each reconnection fails at different percentages, one was at 15% while another at 95%.

Instead of halting the progress, inform them and prompt users if they want to try for more.

Also, check beforehand that if the host has resume capability and again inform users about that. currently I am not sure who to blame for starting over the download progress.

**Notes**

as of now, I could not complete the installation and unfortunately lost appetite towards rust for now. I hope problems would be solved till next time.

facklambda:
Should fix https://github.com/rust-lang/rustup/issues/2853

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -473,6 +473,7 @@ downloader() {
     local _ciphersuites
     local _err
     local _status
+    local _retry
     if check_cmd curl; then
         _dld=curl
     elif check_cmd wget; then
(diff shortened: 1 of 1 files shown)


facklambda:
My fix involves utilizing the pre-existing `check_help_for` function to check if `--retry` flag exists in the curl command's help.
If it does contain --retry, it proceeds to set `--retry 3` as the curl option.
This results in 3 retries with the default exponential back-off.

kinnison (approved):
this looks reasonable, thank you for the work.

PR #2877, opened 2021-10-21
Remove macOS aarch-64 xcode-select directory.

### LINKED ISSUE #2875: PR Builds for macOS aarch64 failing due to xcode-select directory
**Problem**
Builds for macOS aarch64 are failing during the `aarch64-specific items` step, reporting the error `xcode-select: error: invalid developer directory '/Applications/Xcode_12.2.app/Contents/Developer/'`.

**Steps**

1. Open a PR with cosmetic changes (e.g. https://github.com/rust-lang/rustup/pull/2874)
2. It fails in the way described.

**Possible Solution(s)**
My guess is that the latest version of xcode may have changed, and the specified dir doesn't exist anymore. Googling the error brings up stackoverflows of other people experiencing similar issues and suggestions to set the directory to `/Library/Developer/CommandLineTools`.

**Notes**

rustup version and show not relevant for this issue

wcarss:
12.2 is likely out of date, and it is breaking PR builds. It likely
isn't required to use the beta compiler anymore[1], so we'll ~~just remove
the line altogether.~~ specify xcode 13.0 and SDK 11.3.

1 - see this [version table](https://en.wikipedia.org/wiki/Xcode#Xcode_11.x_-_13.x_(since_SwiftUI_framework)) on wikipedia

If successful, fixes https://github.com/rust-lang/rustup/issues/2875.

### DIFF
--- a/.github/workflows/macos-builds-on-all.yaml
+++ b/.github/workflows/macos-builds-on-all.yaml
@@ -78,13 +78,6 @@ jobs:
 
           # Can't run tests: cross-compiling
           echo "SKIP_TESTS=yes" >> $GITHUB_ENV
-
-          # Use the beta compiler
-          sudo xcode-select -s /Applications/Xcode_12.2.app/Contents/Developer/
-
-          # Set SDK environment variables
-          echo "SDKROOT=$(xcrun -sdk macosx11.0 --show-sdk-path)" >> $GITHUB_ENV
-          echo "MACOSX_DEPLOYMENT_TARGET=$(xcrun -sdk macosx11.0 --show-sdk-platform-version)" >> $GITHUB_ENV
         if: matrix.target == 'aarch64-apple-darwin'
       - name: Ensure we have our goal target installed
         run: |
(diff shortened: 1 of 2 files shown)


wcarss:
I don't think removing it is correct. The build (despite getting past this stage) now reports:
```
xcodebuild: error: SDK "macosx11.0" cannot be located.
xcodebuild: error: SDK "macosx11.0" cannot be located.
xcrun: error: unable to lookup item 'Path' in SDK 'macosx11.0'
xcodebuild: error: SDK "macosx11.0" cannot be located.
xcodebuild: error: SDK "macosx11.0" cannot be located.
xcrun: error: unable to lookup item 'PlatformVersion' in SDK 'macosx11.0'
```
so I'll try adding it back in and referencing Xcode 13.0, the current version.

wcarss:
Looks like specifying xcode 13.0 and sdk 11.3 did it, but I suspect that not specifying either would also work from now on and be less fragile across future xcode/SDK releases.

Someone who actually has an M1 might want to sanity-check this before it merges! I'm just a pattern-matching dog who tries numbers out until lights go green. 🐕

ehuss:
fwiw, in the rust-lang/rust repository, it was chosen to just remove the explicit xcode select in https://github.com/rust-lang/rust/pull/89849.

wcarss:
ah, thanks for the context @ehuss! I'll replace these with one where the sdk and xcode specifics are all dropped.

Not sure what the issue might be with that windows-build permissions problem /shrug

kinnison (approved):
Thanks for this

Issue #2793, opened 2021-06-14
Environment variable to disable `rust-toolchain.toml`


jonas-schievink:
**Describe the problem you are trying to solve**

In order to support VS Code's "workspace trust" feature in rust-analyzer (https://github.com/rust-analyzer/rust-analyzer/issues/9224), we'd like to disable `rust-toolchain.toml` support, since that allows arbitrary code execution.

**Describe the solution you'd like**

The easiest solution would be some environment variable that simply disables this toolchain file, but I'm open to alternatives.

matklad:
I'd phrase this more general terms. We need an env variable that indicates that the contents of the file-system is "untrusted". So that, if we set `UNTRUSTED=1`, running `rustc -v`  and such can't steal the secrets and such. I think this is more useful than just the narrow sense of "disabling toolchain file" because:

* it guarantees forward security:  if rustup grows another way to opt-in into insecure (for untrusted code) behaviors, tools that set the env var will work correctly.
* it covers other attack vectors. For example, it probably should forbid overriding `RUSTUP_DIST_SERVER` as well
* it gives better usability: I think allowing toolchain.toml should be fine for "standard" toolchains, it's only path overrides which are problematic.

kinnison:
I agree that a more general solution is better.  There's  a number of possible approaches here.  The approaches outlined so far are good, an alternative is a more explicit approach.  Here's a starter-for-ten on that kind of approach:

Any `rust-toolchain` or `rust-toolchain.toml` must be explicitly enabled by the user in order to work a'la `direnv`.

UX would be approximately:

1. User clones a repository containing a toolchain file
2. By some mechanism they are informed that the directory is not yet trusted (e.g. they run `cargo build`)
3. They choose to trust that directory with `rustup set trusted` or they choose to override the toolchain with `rustup override`
4. They proceed to do their work

The trusting/overrides would not be stored in the dir, but rather in `$RUSTUP_HOME` and as such should be considered trustable.

What do people feel about that as an approach?

bjorn3:
That will break CI as there is no user to approve the `rust-toolchain` and nobody has such approval as part of the CI config yet.

matklad:
Yeah, that's much better, provided that we don't actually break real-world CIs. I guess that's ok, if the logic triggers only for path-based overrides. I think it's OK, and probably better, if rustup downloads official toolchains without questions. 

[...]

PR #2884, opened 2021-10-28
Add support for [env] in rust-toolchain.toml

### LINKED ISSUE #2883: Set environment variables in rust-toolchain.toml

I'm managing a sort of meta-build system that manages packages built in many different languages. It aims to delegate to the idiomatic build tools for each language as much as possible rather than re-implement them, which means it relies heavily on being able to express all necessary configuration through idiomatic configuration files. Unfortunately, there are some configuration options that cannot (currently) be expressed easily in configuration options, and instead have to take the form of environment variables. These are unfortunate because while the meta-build system knows to set those environment variables before it delegates to Cargo, it breaks the ability for human users to "just run `cargo`", which also tends to break IDE integration provided by rust-analyzer and the like, as they also just execute cargo directly and expect that to work.

Some of these can be dealt with by using the new [`[env]` section in `.cargo/config.toml`](https://doc.rust-lang.org/stable/cargo/reference/config.html#env), but since those are only set _after_ Cargo is invoked, it doesn't allow setting environment variables that might affect Cargo itself or Rustup toolchains. While it would be possible to set these using something like [direnv](https://direnv.net/), that is an additional tool that users would have to have set up and working, and if they don't things would fail in surprising and hard-to-diagnose ways.

Concretely, I want to change `PATH` and `CARGO_HOME`:

 - I want to change `PATH` so that tools provided by the meta-build system (and idiomatic dependencies declared therein) are made available to custom toolchains. The meta-build system injects custom toolchains that ensure that the local build uses the rustc/cargo vended through the meta-build system rather than whatever the user happens to have installed locally, and those toolchains in turn rely on plenty of commands being available on `PATH` to do their job. They coincidentally _also_ rely on other custom environment variables, which I would _also_ set in a hypothetical rust-toolchain.toml `[env]` section.
 - I want to change `CARGO_HOME` both to control where Cargo places its various cache files so that they end up in a place that the meta-build system can track (and clean) and where they don't clutter the users' normal Cargo setup, and to avoid accidentally picking up user-specific configuration from `~/.cargo/config.toml` which might make local build artifacts differ in unpredictable ways leading to hard-to-diagnose issues. `CARGO_HOME` needs to be set before `cargo` is invoked, as it affects where Cargo looks for configuration files in the first place, and thus can't be set in Cargo's `[env]`.

I propose we add an `[env]` section to `rust-toolchain.toml` similar to the one [supported by Cargo](https://doc.rust-lang.org/stable/cargo/reference/config.html#env) which allows setting environment variables inside Rustup itself _before_ dispatching to the binary in the toolchain.

**Open questions**

 - The ability to set environment variables is fairly powerful, and could be abused by setting `HOME` or `PATH` in such a way that program execution does "bad things". In all likelihood, this means `[env]` should be subject to the same "trust decision" as is being proposed in https://github.com/rust-lang/rustup/issues/2793.
 - Cargo's `[env]` syntax allows setting config-relative paths rather than absolute ones with a `relative = true` option. Should we support the same? I don't know of such paths existing in Rustup currently.
 - Cargo's `[env]` does not (currently) provide a mechanism for _augmenting_ an environment variable, such as by allowing `PATH` to be set to some value that includes `$PATH`. Should it? My instinct is to follow the design of Cargo's `[env]` section closely, and only add this if Cargo adds support for the same.

jonhoo:
Fixes #2883.

Should probably not be merged until the open questions from #2883 are resolved, and notably the implication of the discussion in https://github.com/rust-lang/rustup/issues/2793 is taken into account.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1165,11 +1165,11 @@ fn show(cfg: &Cfg) -> Result<utils::ExitCode> {
 
         match active_toolchain {
             Ok(atc) => match atc {
-                (ref toolchain, Some(ref reason)) => {
+                (ref toolchain, _, Some(ref reason)) => {
                     writeln!(t, "{} ({})", toolchain.name(), reason)?;
                     writeln!(t, "{}", toolchain.rustc_version())?;
                 }
-                (ref toolchain, None) => {
+                (ref toolchain, _, None) => {
                     writeln!(t, "{} (default)", toolchain.name())?;
                     writeln!(t, "{}", toolchain.rustc_version())?;
                 }
@@ -1221,7 +1221,7 @@ fn show_active_toolchain(cfg: &Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCod
                 return Err(e);
             }
         }
-        Ok((toolchain, reason)) => {
+        Ok((toolchain, _, reason)) => {
             if let Some(reason) = reason {
                 writeln!(process().stdout(), "{} ({})", toolchain.name(), reason)?;
             } else {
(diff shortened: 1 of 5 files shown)


kinnison:
Thanks Jon, I've marked this as draft for now because until we're confident the decisions are right, I don't want to risk merging this.  I'll try and get to code review at some point soonish though, I promise.

PR #2888, opened 2021-10-31
Migration of GitHub issue templates to issue forms

noritada:
This PR provides GitHub issue forms converted from issue templates.

GitHub issue forms (beta) [has been available since June](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) and used in some projects such as cargo.  Although it is still in beta, it provides more pretty and better structured forms than templates, so I have tried converting existing templates into forms.  You can preview forms by looking into [each files in the branch](https://github.com/noritada/rustup/tree/features/issue-forms/.github/ISSUE_TEMPLATE).

I would be grateful if you could point out whether it is reasonable to migrate to forms, and if so, whether the changes are reasonable.

Documentation:
- [Syntax for issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)
- [Syntax for form schema](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema)


kinnison (approved):
These look reasonable

kinnison:
I think this looks nice, let's give it a go.  We can always revert if it proves problematic.

PR #2896, opened 2021-11-10
Change monospace font from Inconsolata to Fira Mono

sondr3:
This is (yet) another follow-up to my previous PRs (#2863 and #2862), primarily because after discussing the font with a couple of friends we didn't really like how Inconsolata looked, both on its own and with the general "feel" of the site. I also slightly tweaked the `font-family` clause to grab a wider selection of user fonts.

## Before

image

## After

image


kinnison (approved):
While I personally like (and actively use in many places) inconsolata, I don't mind if someone with more of an eye to design thinks that fira code is a better fit for the website.  The cleanup work in the CSS is also worthwhile.

PR #2898, opened 2021-11-10
Optimization: parse manifest only once

mkeeter:
I was building a project which invoked `cargo` multiple times, and noticed that a no-op build (i.e. a build with no local changes) was surprisingly slow.

Generating a flamegraph of one `cargo` call, I noticed the following:

Screen Shot 2021-11-10 at 1 40 11 PM

About half the work is done in `Manifestation::load_manifest`, which is called twice to parse *the same* 700K TOML file ([here](https://github.com/rust-lang/rustup/blob/master/src/config.rs#L742) and [here](https://github.com/rust-lang/rustup/blob/master/src/config.rs#L750)).

This PR adds a new `struct` in `toolchain.rs` which encapsulates this parsed manifest plus other relevant data.  This `struct` is then used in `config.rs` to eliminate the double-parsing.

I also use this `struct` to clean up duplicate code in `toolchain.rs`, e.g. [this comment](https://github.com/rust-lang/rustup/blob/master/src/toolchain.rs#L874):
```
// Overlapping code with get_manifest :/.
```

I see a 15-25% speedup in a no-op build, depending on the project!

- [X] Tests all pass (at least on my machine!)
- [X] No new `clippy` lints (there are a few pre-existing lints in areas I haven't touched)
- [X] `rustfmt` run

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -529,30 +529,21 @@ impl<'a> DistributableToolchain<'a> {
 
     // Installed only.
     pub(crate) fn add_component(&self, mut component: Component) -> Result<()> {
-        if !self.0.exists() {
-            return Err(RustupError::ToolchainNotInstalled(self.0.name.to_owned()).into());
-        }
-
-        let toolchain = &self.0.name;
-        let toolchain = ToolchainDesc::from_str(toolchain).expect("must be valid");
-
-        let prefix = InstallPrefix::from(self.0.path.to_owned());
-        let manifestation = Manifestation::open(prefix, toolchain.target.clone())?;
-
-        if let Some(manifest) = manifestation.load_manifest()? {
+        if let Some(desc) = self.get_toolchain_desc_with_manifest()? {
             // Rename the component if necessary.
-            if let Some(c) = manifest.rename_component(&component) {
+            if let Some(c) = desc.manifest.rename_component(&component) {
                 component = c;
             }
 
             // Validate the component name
-            let rust_pkg = manifest
+            let rust_pkg = desc
+                .manifest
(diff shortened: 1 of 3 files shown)


kinnison:
You have reached a similar conclusion to that which led me in #2626 to think about writing #2627.

My gut feeling is that it'd be *best* to rewrite the toml into something which can be parsed more quickly, rather than complicating the internal pathways in rustup regarding parsed toolchain manifests.

I will have a read through your change, but I'm sceptical that as-is it'll be the right answer.

I am grateful for the drive-by cleanup work though, that idea is excellent.

src/toolchain.rs
```diff
@@ -972,6 +908,68 @@ impl<'a> DistributableToolchain<'a> {
     }
 }
 
+/// Helper type to avoid parsing a manifest more than once
+pub(crate) struct ToolchainDescWithManifest {
+    toolchain: ToolchainDesc,
+    manifestation: Manifestation,
+    pub(crate) manifest: Manifest,
```
kinnison: With the struct being `pub(crate)` there's no need for the `(crate)` modifier on this field.

kinnison (changes requested):
I take back my comment, I think the extra complication of that struct is simple enough that I do not object to it. This looks very worthwhile to include, thank you!

Could you please deal with the one comment I made on the code; and then when you're done, rebase the changes together, since I prefer to not merge "run rustfmt" type patches into the code if I can avoid it.

Thank you again.

mkeeter:
Cool – I made that one change, then squashed into a single commit and force-pushed it, so we should be good to go!

PR #2902, opened 2021-11-14
globally introduce #![deny(unreachable_pub)]

### LINKED ISSUE #2730: Detect dead code better

0xPoe:
close https://github.com/rust-lang/rustup/issues/2730

I know we don't want to use pub(crate) in pub(crate) struct, but it looks like this lint suggests we do. Do you have any comments on how to handle it?

### DIFF
--- a/src/dist/config.rs
+++ b/src/dist/config.rs
@@ -8,9 +8,9 @@ pub(crate) const SUPPORTED_CONFIG_VERSIONS: [&str; 1] = ["1"];
 pub(crate) const DEFAULT_CONFIG_VERSION: &str = "1";
 
 #[derive(Clone, Debug)]
-pub struct Config {
-    pub config_version: String,
-    pub components: Vec<Component>,
+pub(crate) struct Config {
+    pub(crate) config_version: String,
+    pub(crate) components: Vec<Component>,
 }
 
 impl Config {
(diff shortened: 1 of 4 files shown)


kinnison:
I'm not sure what the right answer is here - the extra `(crate)` is entirely redundant as far as I understand it.  Perhaps we need to check in with that lint author to find out?

0xPoe:
@Nemo157 I remember you helped to modify this lint last year, do you have any suggestions for using create field in the crate structure? Thanks!

Nemo157:
🤔 I did? I don't remember that. It does make sense to me that fields could be excluded from this lint, they are not _items_ which are what it's meant to lint against, but it also makes some sense to lint against them too and have a fully consistent "`pub` means external users can access it" so that locally you can tell whether things are part of the public API (though this matters less for this usecase where it looks like the `unreachable_pub` lint is being used as a manual implementation of <https://github.com/rust-lang/rust/issues/74970>).

0xPoe:
Maybe this doesn't make much sense for field checking. So I close it.

PR #2906, opened 2021-11-23
Remove dead code warning for test-only `size` field for `diskio::Item`

jyn514:


0xPoe:
We use it in our tests. Maybe we can use cfg(test).

I seem to have missed it in https://github.com/rust-lang/rustup/pull/2901.

jyn514:
Ah, ok. I can add `cfg(test)` to it instead.

0xPoe (approved):
LGTM. Thanks!

rbtcollins:
Thank you for fixing this. I"m a bit nervous about our test coverage here, but we can always iterate with other patches in future.

PR #2910, opened 2021-11-27
Warn on unknown key in toolchain file

### LINKED ISSUE #2876: Unknown keys in rust-toolchain.toml are silently ignored

Aelerinya:
Closes #2876

I took the conservative approach of only displaying a warning when an unknown key is found, so that an older version of rustup can still parse future toolchain files with unknown keys.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1755,6 +1755,7 @@ dependencies = [
  "scopeguard",
  "semver",
  "serde",
+ "serde_ignored",
  "sha2",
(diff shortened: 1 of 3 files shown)


rbtcollins:
Thank you for working on this. I've just updated the bug report with an updated specification from @kinnison and I - the work you've done is probably still absolutely usable, but theres a little more to do - we need to document the versions toolchain keys are added in in the docs, and the situation should be a hard error.

PR #2933, opened 2022-01-13
use new cargo fmt option

calebcartwright:
As of v1.58, cargo fmt now supports the --check flag directly. Updating it here (and in a few other r-l repos) both because it's more succinct and so more people will see/become aware

### DIFF
--- a/.github/workflows/centos-fmt-clippy-on-all.yaml
+++ b/.github/workflows/centos-fmt-clippy-on-all.yaml
@@ -87,7 +87,7 @@ jobs:
           git ls-files -- '*.bash' | xargs shellcheck -s bash -e SC1090
       - name: Run formatting checks
         run: |
-          cargo fmt --all -- --check
+          cargo fmt --all --check
       - name: Run cargo check and clippy
         run: |
           cargo check --all --all-targets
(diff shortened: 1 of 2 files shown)


0xPoe (commented):
Thanks for your contribution!
It looks like we should modify the https://github.com/rust-lang/rustup/blob/master/ci/actions-templates/centos-fmt-clippy-template.yaml file instead of editing this file directly?

calebcartwright:
> Thanks for your contribution! It looks like we should modify the https://github.com/rust-lang/rustup/blob/master/ci/actions-templates/centos-fmt-clippy-template.yaml file instead of editing this file directly?

D'oh. Yes indeed, thank you. Not sure how I missed the very obvious comment at the top of the file :man_facepalming: Will update this later

0xPoe (commented):
I assume you need to generate new files with the latest template?

calebcartwright:
> I assume you need to generate new files with the latest template?

I hope that's not a question for me as I really have no idea how you all are managing your CI :smile:

0xPoe:
@calebcartwright Checkout https://github.com/rust-lang/rustup/tree/master/ci/actions-templates#the-workflows

I think you can use `gen-workflows.sh` to gen the new actions files.

PR #2936, opened 2022-01-24
Make clippy happy

0xPoe:
Please check each commit.

### DIFF
--- a/src/diskio/immediate.rs
+++ b/src/diskio/immediate.rs
@@ -204,7 +204,7 @@ impl IncrementalFileWriter {
     fn write(&mut self, chunk: Vec<u8>) -> std::result::Result<bool, io::Error> {
         let mut state = self.state.lock().unwrap();
         if let Some(ref mut state) = *state {
-            if let Some(ref mut file) = (&mut self.file).as_mut() {
+            if let Some(ref mut file) = self.file.as_mut() {
                 // Length 0 vector is used for clean EOF signalling.
                 if chunk.is_empty() {
                     trace_scoped!("close", "name:": self.path_display);
--- a/src/dist/signatures.rs
+++ b/src/dist/signatures.rs
@@ -56,20 +56,13 @@ impl VerificationHelper for Helper<'_> {
         for layer in structure.into_iter() {
             match layer {
                 MessageLayer::SignatureGroup { results } => {
-                    for result in results {
-                        match result {
-                            Ok(GoodChecksum { ka, .. }) => {
-                                // A good signature!  Find the index
-                                // of the singer key and return
-                                // success.
-                                self.index = self.certs.iter().position(|c| c.cert() == ka.cert());
-                                assert!(self.index.is_some());
-                                return Ok(());
-                            }
-                            _ => {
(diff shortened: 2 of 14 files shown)


src/currentprocess/filesource.rs
```diff
@@ -24,8 +24,8 @@ impl Stdin for io::Stdin {
     fn lock(&self) -> Box<dyn StdinLock + '_> {
         Box::new(io::Stdin::lock(self))
     }
-    fn read_line(&self, buf: &mut String) -> Result<usize> {
-        io::Stdin::read_line(self, buf)
+    fn read_line(&self, buf: &mut str) -> Result<usize> {
+        io::Stdin::read_line(self, &mut buf.to_owned())
```
kinnison: This is confusing, surely that won't work?

src/currentprocess/filesource.rs
```diff
@@ -68,8 +68,8 @@ impl Stdin for TestStdin {
             inner: self.0.lock().unwrap_or_else(|e| e.into_inner()),
         })
     }
-    fn read_line(&self, buf: &mut String) -> Result<usize> {
-        self.lock().read_line(buf)
+    fn read_line(&self, buf: &mut str) -> Result<usize> {
+        self.lock().read_line(&mut buf.to_owned())
```
kinnison: Ditto, won't this end up losing the read-in content?

src/diskio/immediate.rs
```diff
@@ -204,7 +204,7 @@ impl IncrementalFileWriter {
     fn write(&mut self, chunk: Vec<u8>) -> std::result::Result<bool, io::Error> {
         let mut state = self.state.lock().unwrap();
         if let Some(ref mut state) = *state {
-            if let Some(ref mut file) = (&mut self.file).as_mut() {
+            if let Some(ref mut file) = (self.file).as_mut() {
```
kinnison:
```suggestion
            if let Some(ref mut file) = self.file.as_mut() {
```
?

src/dist/signatures.rs
```diff
@@ -56,20 +56,12 @@ impl VerificationHelper for Helper<'_> {
...
-                            _ => {
-                                // We ignore any errors.
-                            }
-                        }
+                    for GoodChecksum { ka, .. } in results.into_iter().flatten() {
+                        // A good signature!  Find the index
+                        // of the singer key and return
+                        // success.
+                        self.index = self.certs.iter().position(|c| c.cert() == ka.cert());
+                        assert!(self.index.is_some());
```
kinnison: This loses the nice explicit indicator that we are ignoring errors during signature verification, which is an open question for us.  We shouldn't hide that.  If you're going to do this refactor then please ensure there's a comment to the effect that the `.flatten()` causes us to ignore errors for now.  Also fix the comment `singer` -> `signer` please :D

kinnison (changes requested):

src/currentprocess/filesource.rs
0xPoe: Yes, this should be a false positive.

kinnison (approved):

PR #2939, opened 2022-01-30
Fix custom toolchain name in the example

yerke:
In the rest of the page we use `my-toolchain` rather than `myrust`.


0xPoe (approved):
Looks reasonable to me.

PR #2945, opened 2022-02-05
Upgrade rustup to the 2021 edition

### LINKED ISSUE #2944: Update Rust toolchain to edition 2021
Edition 2021 has been released for a long time, and we can upgrade to the latest edition.

0xPoe:
close https://github.com/rust-lang/rustup/issues/2944

We should follow up with the latest edition.


kinnison (commented):
The change to `ci/run.bash` is confusing me - why is this needed for the edition change?

0xPoe:
> The change to `ci/run.bash` is confusing me - why is this needed for the edition change?

Because the download package does not have this `vendored-openssl` feature, it will report an error under the new edition.

kinnison (approved):

PR #2951, opened 2022-02-19
Support busybox-w32 in rustup-init.sh

SpaceManiac:
[busybox-w32](https://frippery.org/busybox/) differs from currently-supported shells in some ways:
- `uname -s` is `Windows_NT` instead of anything currently recognized
- `/dev/tty` does not exist
  - The fix currently checks whether stdin is already a terminal, but could instead be checking that `/dev/tty` exists

On versions of Windows 10 where `curl` comes included, it's already preferred over busybox-w32's `wget`.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -135,7 +135,7 @@ main() {
         exit 1
     fi
 
-    if [ "$need_tty" = "yes" ]; then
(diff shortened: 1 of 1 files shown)


kinnison:
Sorry it took me so long to spot this, thank you.

Issue #2739, opened 2021-04-27
Testing for proxying, including both TOOLS and DUP_TOOLS


workingjubilee:
(Note: this has been edited by the rustup team)

1.24.0 shipped with the ability to run `cargo fmt`, untested.

1.24.1 will not have that defect.
We need to test to make sure that:
- rustup correctly proxies to all of its core tools rather than the subset currently checked.
- rustup correctly proxies to its `DUP_TOOLS` list.

rbtcollins:
Rustup doesn't do the third Item on your list.

0xPoe:
Test all proxies on macOS Catalina 10.15.7 19H2 x86_64

All proxies working well.

The package cargo-cache for cargo was also tested.

See: https://discord.com/channels/442252698964721669/463480252723888159/836798461134897152

0xPoe:
If anyone wants to help test, you can use the following command to upgrade to 1.24.1(linux/macos/git-bash):

```sh
$ export RUSTUP_UPDATE_ROOT=https://dev-static.rust-lang.org/rustup
$ rustup self update
info: checking for self-updates
info: downloading self-update
  rustup updated - 1.24.1 (from 1.23.1)
```

0xPoe:
@rustbot claim

I will try to test it.

0xPoe:
[Added unit tests to it](https://github.com/rust-lang/rustup/pull/2952), maybe we need better integration tests. My intention is to write a script and install and run these proxy tools in the CI, do you have a better way to test them?

PR #2952, opened 2022-02-27
Add tests for proxyable tools

0xPoe:
par tof https://github.com/rust-lang/rustup/issues/2739

Add is_proxyable_tools fn and test it.


kinnison:
I guess it's one fewer place to make the mistake so I'm ok with this, but I think an integration test of some kind will be useful to have in the future too.

PR #2953, opened 2022-02-28
Add the user agent for http requests

### LINKED ISSUE #2860: Feature request: add User-Agent header to http requests

0xPoe:
close https://github.com/rust-lang/rustup/issues/2860
close https://github.com/rust-lang/rustup/pull/2861

Unify rustup and download package versions. (See: https://github.com/rust-lang/rustup/pull/2861#discussion_r744646528)
Add the user agent for http requests and update the release process docs.


0xPoe:
cc: @kinnison @jiegec 
Could you please take a look? Thanks! 💚 💙 💜 💛 ❤️

kinnison:
I suppose that tweak of just making sure we use the same crate version for both parts makes sense. :+1:

Issue #2947, opened 2022-02-10
Windows MSVC: automate installing Visual Studio components


ChrisDenton:
### Problem you are trying to solve

Using Rust for `msvc` toolchains  requires some components from Visual Studio to be installed: However, this requires the user to download the vs installer and select the correct components. This can be a source of friction for new users.

A good, regularly updated, install guide would help but it'd be great if rustup itself could ease the install process.

### Solution you'd like

I would suggest this workflow:

* If Visual Studio is not installed at all then offer to download and install the Community Edition with the required components.
* If Visual Studio is installed but is missing the required components, offer to install what is needed.

Alternatively rustup could just do the second bullet point. The user would have to manually install Visual Studio but rustup would be able to fix anything missed by the user. In any case, this might be a good first step if this feature request is accepted.

### Notes

Rustup already has logic to prompt the user to install Visual Studio if necessary.

The Visual Studio installer can be downloaded by using winget (if installed) or via `https://aka.ms/vs/17/release/vs_community.exe`.

The install can be [automated](https://docs.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2022) using command line arguments. It can be run either with a GUI (where rustup can set the components to be installed by default) or in "quiet" mode without a GUI or user interaction. In the latter case, the [license terms](https://visualstudio.microsoft.com/license-terms/) would need to be linked by rustup itself.

If Visual Studio is installed but components are missing then [`vswhere`](https://github.com/Microsoft/vswhere#visual-studio-locator) can be used to find the existing installer, which in turn can be used to install missing components. Fortunately Visual Studio installs vswhere to a stable location so this should be easy to find. Alternatively there is a COM interface that can be used to query Visual Studio directly. This is what [cc-rs](https://github.com/alexcrichton/cc-rs) uses internally so perhaps features could be added to that crate to support this use case.

kinnison:
This sounds like a very clever way to deal with this.  If you think you can drive this functionality forward then it would be nice to have; though I don't know how easily we'd be able to CI this.

ChrisDenton:
A simple improvement to start with might be to update the instructions to tell the user how to run the Visual Studio Installer from the command line with the C++ workload pre-selected. For example:

Using winget:

```cmd
winget install --id Microsoft.VisualStudio.2022.Community --override "--add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended --focusedUi --wait"
```

[...]

PR #2954, opened 2022-03-03
Offer to auto install VS 2022

ChrisDenton:
This PR aims to improve installation on MSVC hosts when the user does not have Visual Studio already installed. The default workflow will download and install VS 2022 unless the user declines.

For all other cases (e.g. Visual Studio is installed but lacks the required components) it falls back to the current manual install message. This could be further improved in the future but on reflection I think I'd like to work out the new VS install case first because I feel this will have the biggest impact. Users who are already familiar with Visual Studio will have an easier time understanding the install instructions.

See #2947

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -1,5 +1,6 @@
-use std::env::consts::EXE_SUFFIX;
+use std::env::{consts::EXE_SUFFIX, split_paths};
 use std::ffi::{OsStr, OsString};
+use std::fmt;
 use std::os::windows::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 use std::process::Command;
@@ -24,12 +25,18 @@ pub(crate) fn ensure_prompt() -> Result<()> {
     Ok(())
 }
 
+#[derive(PartialEq, Eq)]
+pub(crate) enum VsInstallPlan {
+    Automatic,
+    Manual,
+}
+
 // Provide guidance about setting up MSVC if it doesn't appear to be
 // installed
-pub(crate) fn do_msvc_check(opts: &InstallOpts<'_>) -> bool {
+pub(crate) fn do_msvc_check(opts: &InstallOpts<'_>) -> Option<VsInstallPlan> {
     // Test suite skips this since it's env dependent
     if process().var("RUSTUP_INIT_SKIP_MSVC_CHECK").is_ok() {
-        return true;
+        return None;
     }
 
     use cc::windows_registry;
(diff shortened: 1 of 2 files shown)


ChrisDenton:
I think two issues with this PR are going to be testing and improving the messages displayed to users. Also this does spawn an interactive GUI installer which is not something rustup has done before.

kyle-rader:
I just went through the manual process of trying to find the right things to be installed, I haven't looked closely yet, but VS 2022 Enterprise installed with several packs (that we use internally at Microsoft) still doesn't cover the requirements. I am happy to help test this.

src/cli/self_update/windows.rs
```diff
@@ -41,10 +48,102 @@ pub(crate) fn do_msvc_check(opts: &InstallOpts<'_>) -> bool {
...
+            path.push("kernel32.lib");
+            if path.exists() {
+                has_libs = true;
+            }
+        }
+    };
+    if !has_libs {
+        cmd.args([
+            "--add",
+            "Microsoft.VisualStudio.Component.Windows11SDK.22000",
```
kinnison, line 146: Is this the lowest common denominator?  I don't have Windows 11, would this cause an SDK which would build stuff which wouldn't run on my Windows PC?

src/cli/self_update/windows.rs
```diff
@@ -41,10 +48,102 @@ pub(crate) fn do_msvc_check(opts: &InstallOpts<'_>) -> bool {
...
+    // download the installer
+    let visual_studio_url = utils::parse_url("https://aka.ms/vs/17/release/vs_community.exe")?;
+
+    let tempdir = tempfile::Builder::new()
+        .prefix("rustup-visualstudio")
+        .tempdir()
+        .context("error creating temp directory")?;
+
+    let visual_studio = tempdir.path().join("vs_setup.exe");
+    utils::download_file(&visual_studio_url, &visual_studio, None, &|_| ())?;
```
kinnison: It's a pity you don't offer a download progress bar here. I accept it might be a bit of a faff, but if you can sort a notification handler out to get a progress bar, that'd be nice for people.

src/cli/self_update/windows.rs
```diff
@@ -41,10 +48,102 @@ pub(crate) fn do_msvc_check(opts: &InstallOpts<'_>) -> bool {
...
+
+    // Run the installer. Arguments are documented at:
+    // https://docs.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio
+    let mut cmd = Command::new(visual_studio);
+    cmd.arg("--wait")
+        // Display an interactive GUI focused on installing just the selected components.
+        .arg("--focusedUi")
+        // Add the linker and C runtime libraries.
+        .args(["--add", "Microsoft.VisualStudio.Component.VC.Tools.x86.x64"]);
```
kinnison, line 130: Is this component the correct name regardless of if we're i686 or x86_64 ?

kinnison (commented):
I'm pretty OK with this - I have no useful way to check (My Windows system isn't for development) so if you can answer my questions and / or make the changes I suggest then I feel like we're close to being able to merge this if someone else can check it.

src/cli/self_update/windows.rs
ChrisDenton: The latest SDK can be used to build for older versions. From [their website](https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/) (my bold): 
> The Windows SDK (10.0.22000) for Windows 11 provides the latest headers, libraries, metadata, and tools for building Windows applications. Use this SDK to build Universal Windows Platform (UWP) and Win32 applications for Windows 11 and **previous Windows releases**.

src/cli/self_update/windows.rs
ChrisDenton: That would indeed be good. I just have to learn how 🙂. It should only be a small download though (about 1.6 MB) as it bootstraps the actual install.

src/cli/self_update/windows.rs
ChrisDenton: Yes. The name means x86/x64 (i.e. "and") rather than meaning `x86_64`, if that makes sense. So basically it means both 32 and 64 bit.

src/cli/self_update/windows.rs
kinnison: OK, thanks for the clarification.

src/cli/self_update/windows.rs
kinnison: It's not a blocker, just if you want to have a go at it you're welcome to :D

src/cli/self_update/windows.rs
kinnison: Aha OK thanks you for the clarification.

src/cli/self_update/windows.rs
ChrisDenton: Yep, I can definitely try!

ChrisDenton:
> I just went through the manual process of trying to find the right things to be installed, I haven't looked closely yet, but VS 2022 Enterprise installed with several packs (that we use internally at Microsoft) still doesn't cover the requirements. I am happy to help test this.

Thanks! Currently this PR is aimed at getting the OSS or home developer up and running with only what's needed for a pure Rust project using the Community Edition. The enterprise use case can be more complicated because we need to select the right edition and make sure to test for existing versions of any tools or libraries rather than installing the latest. That is something I plan to handle too but I'm worried about doing too much in one PR.

Also enterprise projects are more likely to involve a mix of languages, frameworks, etc and other requirements which Rustup can't know about. This PR uses a "focused" GUI mode to concentrate on what Rustup will install but for other use cases it may be more important to display the full GUI.

Sorry if I'm rambling. I've just been thinking a lot about various different use cases.

kyle-rader:
No worries, not rambling :). I was just saying that I agree this can be a confusing process and that even with my enterprise installation with 20GB of components installed, I still was having trouble getting it working. I've tested the steps here manually so far and they did fix my Rust installation. This weekend I will try to build this and run some tests in a VM. I think there is a way we can also have the UI that is launched be completely non-interactive.

ChrisDenton:
> I think there is a way we can also have the UI that is launched be completely non-interactive.

Yeah there is. The problem is mostly licensing. Showing an interactive UI "solves" that by requiring the user to click through the prompts that link to the licenses. We can maybe solve that by linking to https://visualstudio.microsoft.com/license-terms/ and asking if the user agrees. Assuming that link is stable.

src/cli/self_update/windows.rs
ChrisDenton: Ok, I've added a progress bar.

ChrisDenton:
Hm, there seems to be an issue with the arm linux CI?

> Error response from daemon: No such image

src/cli/self_update/windows.rs
kinnison: That looks quite promising, I've restarted the CI to clear that ARM problem.

kinnison:
On the basis that as it stands I can't easily test this, I'm going to merge and when we come to do the release I'll specifically seek feedback on this component during the public test phase.

PR #2955, opened 2022-03-03
Add CI check for FreeBSD

GuillaumeGomez:
To do so, I used cirrus-ci (it'll need to be setup to allow to run the checks on this repository too). You can see my original PR and the checks passing there: https://github.com/GuillaumeGomez/rustup/pull/4

The only thing I didn't test was for uploading artifacts to S3.

For the CI scripts, I tried to keep the same logic with the `gen-workflow.sh`. Basically it's simply a copy-paste.


ci/cirrus-templates/gen-workflows.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+INPATH=$(dirname "$0")
+OUTPATH="${INPATH}/../.."
+
+cp freebsd.yaml "$OUTPATH/.cirrus.yml"
```
kinnison, line 6: This whole file feels a bit odd to have, though I suppose we might want to template it in future.

ci/cirrus-templates/freebsd.yaml
```diff
@@ -0,0 +1,24 @@
...
+    chown -R tester .
+    sudo -u tester bash ci/cirrus-templates/script.bash
+  upload_to_s3:
+    only_if: $CIRRUS_BRANCH == "stable"
+    script:
+      - pkg install -y py36-pip bash
+      - pip-3.6 install -U setuptools
+      - pip-3.6 install awscli
+      - bash ci/prepare-deploy.bash
+      - aws s3 cp --recursive deploy/ s3://dev-static-rust-lang-org/rustup/
```
kinnison: Do we need to set AWS_DEFAULT_REGION here?

tests/cli-paths.rs
```diff
@@ -109,6 +109,7 @@ export PATH="$HOME/apple/bin"
         });
     }
 
+    // This test should NOT be run as root!
```
kinnison, line 112: Where did this comment come from?

ci/cirrus-templates/gen-workflows.sh
GuillaumeGomez: I kept the same logic as github actions. If it's useless, I can remove it.

ci/cirrus-templates/freebsd.yaml
GuillaumeGomez: No idea. I couldn't test this part so if something needs to be added, please tell me. :)

tests/cli-paths.rs
GuillaumeGomez: By default, on cirrus CI we run scripts as root. So this test was failing and I couldn't understand why. The reason is simple: when root, the file rights are overlooked, making this test fail. Considering how much time I spent on this issue, I felt like the problem could be solved faster if we add support for another target and encounter the same issue.

ci/cirrus-templates/freebsd.yaml
kinnison: I'm tempted to say "remove this" and we'll stick with stable releases from the cross-built binary for now.  Otherwise we need to be sure about this *and* disable the freebsd cross builds.

tests/cli-paths.rs
kinnison: Ouch. OK

ci/cirrus-templates/freebsd.yaml
GuillaumeGomez: Done! I have put the removal in its own commit in case we want to have the S3 upload later on so it's easy to revert.

kinnison (approved):

PR #2958, opened 2022-03-10
Set RUSTC and RUSTDOC env for child processes run through the proxy

jyn514:
This serves two purposes:
1. It avoids the overhead of having to re-parse `settings.toml` on each proxy.
   In projects with many crates, this can have non-trivial overhead.
2. For build scripts, it provides an absolute path for RUSTC rather than forcing them to look it up in PATH.
   This allows them to find the sysroot path for the toolchain, rather than ending up somewhere in CARGO_HOME.

The original motivation for this was to allow `bootstrap` in rust-lang/rust
to allow building stage0 tools without python, only the host rust toolchain. See https://github.com/rust-lang/rust/issues/94797.

cc @Mark-Simulacrum - do you think this is a reasonable approach to fixing the original issue? It does have the downside that we can't fix it until ~the next rustup release, but I think it's the least hacky way to do it. Other things I can think of involve running `rustup which rustc` and such, which seems pretty not great if we don't know whether `rustc` is managed by rustup or not.

---

Reverted in #3034.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -401,6 +401,19 @@ impl<'a> InstalledCommonToolchain<'a> {
 
         cmd.env("RUSTUP_TOOLCHAIN", &self.0.name);
         cmd.env("RUSTUP_HOME", &self.0.cfg.rustup_dir);
+        // It's valid (although slightly cursed) to tell a rustup cargo proxy to use a custom rustc
+        // from a different toolchain.
+        if env::var("RUSTC").is_err() {
+            // Set RUSTC for the spawned binary. This serves two purposes:
+            // 1. It avoids the overhead of having to re-parse `settings.toml` on each proxy.
+            //    In projects with many crates, this can have non-trivial overhead.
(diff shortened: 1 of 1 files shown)


jyn514:
Ok, I found a better way to do this for bootstrap specifically: https://github.com/rust-lang/rust/pull/94806. I do think it could be useful to set these variables though, the decreased build time especially seems tempting.

kinnison:
Are you confident this is right when using rustup to patch a rustc-only dev build with stable's cargo?

jyn514:
Yes.
```
$ which cargo
/home/jnelson/rust-lang/rustup/home/bin/cargo
$ cargo +stable rustc -q -- --version
rustc 1.59.0 (9d1b2106e 2022-02-23)
$ cargo +stage2 rustc -q -- --version
rustc 1.61.0-nightly (85ce7fdfa 2022-03-10)
$ rustup toolchain list -v | grep stage2
stage2	/home/jnelson/rust-lang/rust/build/x86_64-unknown-linux-gnu/stage2/
```

Issue #2838, opened 2021-08-11
rust-gdb and rust-lldb not working using stable-x86_64-pc-windows-gnu toolchain


PietroValerio:
**Problem**

expected debugget output but this message pops up
```the 'rust-gdb.exe' binary, normally provided by the 'rustc' component, is not applicable to the 'stable-x86_64-pc-windows-gnu' toolchain```

**Steps**

1.type rustup-gdb or rustup-lldb

**Possible Solution(s)**

**Notes**
Operating system:Windows 10 pro version 10.0.19042 Build 19042
Using mingw-64
Output of `rustup --version`: rustup 1.24.3 (ce5817a94 2021-05-31)
Output of `rustup show`:
  Default host: x86_64-pc-windows-gnu
  stable-x86_64-pc-windows-gnu (default)
  rustc 1.54.0 (a178d0322 2021-07-26)

**Comment by ehuss** [comment]
I think (maybe) the issue is that rustup is assuming all executables end with `.exe` [here](https://github.com/rust-lang/rustup/blob/a441c41679386cf62ebd9e49af6d8b37cd792af6/src/toolchain.rs#L326-L335), but scripts like rust-gdb are shell scripts which do not have the extension.

**Comment by kinnison** [comment]
Certainly on Windows we blindly append the extension `.exe` -- perhaps

1. we should use `PATHEXT` rather than always assuming `.exe`
2. we should consider *how* to have `rust-gdb` and `rust-lldb` scripts which will work on native Windows and then give the new scripts appropriate extensions.

Failing that, I suppose we could detect the missing executable, but the fact that the component is already installed and present a better error message.

**Comment by prdas31** [comment]
Same issue ...

```
D:\RustTest01>rust-lldb
error: the 'rust-lldb.exe' binary, normally provided by the 'rustc' component, is not applicable to the 'stable-x86_64-pc-windows-msvc' toolchain
````

```
D:\RustTest01>which rust-lldb
/d/Programs/Rust/.cargo/bin/rust-lldb
```

**Comment by selvavm** [comment]
Based on the [link](https://github.com/eclipse/corrosion/blob/master/documentation/Troubleshooting.md#debugging-on-windows), I am able to debug rust in windows with rust-gdb in Visual Studio Code using the below configuration,
```
{
            "type": "gdb",
            "request": "launch",
            "name": "Debug Launch",
            "target": "${workspaceRoot}/target/debug/${workspaceRootFolderName}.exe",
[...]

PR #2962, opened 2022-03-11
Add smoke test for proxies

### LINKED ISSUE #2739: Testing for proxying, including both TOOLS and DUP_TOOLS

0xPoe:
close https://github.com/rust-lang/rustup/issues/2739

- Add smoke test for proxies
- Gen the new workflows


ci/smoke-test.bash
```diff
@@ -0,0 +1,56 @@
...
+rustc --version
+rustdoc --version
+cargo --version
+rustfmt --version
+cargo-fmt --version
+cargo-clippy --version
+clippy-driver --version
+# rust-lldb --version
+# rust-gdb -h
+# rust-gdbgui -h
```
kinnison: Why not running these ?

0xPoe: Because they have other dependencies and there are related issues that have not been resolved. So we can skip it for now.
cc: https://github.com/rust-lang/rustup/issues/2838 and https://github.com/rust-lang/rust/issues/61282

rbtcollins:
I'm quite curious why this smoke test is a bash script, only running in CI, and not a rust integration test running in our test suite.

I can see the value in something that explicitly inspects the expected set of proxies, makes sure they can all be invoked, and are all correct hardlinks to each other. I don't *think* we have that today.

But I see lots of technical debt in this particular testing approach and am not a fan.

0xPoe:
> I can see the value in something that explicitly inspects the expected set of proxies, makes sure they can all be invoked, and are all correct hardlinks to each other. I don't _think_ we have that today.

@rbtcollins  Sounds good, can you provide some guidance? I'm trying to invoke the bin out of mock and it looks like it goes into recursion and then fails.

```rust
 let path = &config.cargodir.join(&format!("bin/{}{}", tool, EXE_SUFFIX));
 expect_component_executable(config, path.to_str().unwrap());
```

```log
out.stderr (1 lines):
    error: the 'rustdoc' binary, normally provided by the 'rustc' component, is not applicable to the 'stable-x86_64-apple-darwin' toolchain
```

PR #2963, opened 2022-03-12
curl https://sh.rustup.rs -sSf | sh

Shutch748:
My info Shutch748:master


Shutch748 (commented):
Mine

hi-rustin (changes requested):
Hello. 👋 Why do you want to delete it? What does your PR title mean?

kinnison:
The user looks very much like a bot to me, I've reported them so that GitHub can review their activity and decide.

PR #2964, opened 2022-03-12
simplify code a bit

matthiaskrgr:


kinnison (changes requested):
Hi Mathias, this is actually a clippy lint that we disagree with because it's making the shape of adding arguments different in the two places.

We have an issue open in clippy around this here - https://github.com/rust-lang/rust-clippy/issues/7071

I suppose in reality we should add an allow for that lint, just to keep things quieter, but we won't apply the refactor you suggest here.

matthiaskrgr:
Well I got rid of the extra push entirely :upside_down_face:  but feel free to close if that's too convoluted. :sweat_smile:

tests/mock/clitools.rs
```diff
@@ -637,17 +637,16 @@ where
     let mut vars: HashMap<String, String> = HashMap::default();
     self::env(config, &mut vars);
     vars.extend(env.iter().map(|(k, v)| (k.to_string(), v.to_string())));
-    let mut arg_strings: Vec<Box<str>> = Vec::new();
-    arg_strings.push(name.to_owned().into_boxed_str());
-    for arg in args {
-        arg_strings.push(
+    let arg_strings = Some(name.to_owned().into_boxed_str())
```
rbtcollins, line 640: This is definitely not cleaner: iterating over a Some() to avoid a clippy lint... in test code.

rbtcollins (changes requested):
Clarity is paramount in this test harness code - hypothetical performance advantages are irrelevant (at least for now), because test performance is dominated by the IO and layering we have, not in memory operations like this.

matthiaskrgr:
Alright, closing ^^

PR #2968, opened 2022-03-18
add `force-non-host` flag to install subcommand

Emilgardis:
This touches on standing issue in https://github.com/cross-rs/cross/issues/536

ping https://github.com/cross-rs/cross/pull/554

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -296,6 +296,11 @@ pub(crate) fn cli() -> App<'static, 'static> {
                         .help("Force an update, even if some components are missing")
                         .long("force")
                         .takes_value(false),
+                ).arg(
+                    Arg::with_name("force-non-host")
+                        .help("Install toolchains that require an emulator. See https://github.com/rust-lang/rustup/wiki/Non-host-toolchains")
+                        .long("force-non-host")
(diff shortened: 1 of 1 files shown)


Emilgardis:
Has anyone been able to review this? The CI failure is unrelated.

Emilgardis:
what's the status for this?

it's unexpected that `install` tells the user to use `--force-non-host`, but doesn't allow it in this context.

kinnison:
So `rustup install` is a hidden semi-alias for `rustup toolchain install` which is the official installation command for toolchains.

It's unfortunate that so many people expected `rustup install` and so I was not permitted to remove it (or even cause it to emit information about how it's deprecated) because the way it's implemented causes this kind of problem for people with legacy use of the hidden alias.

I'm OK with adding this patch, but honestly I'd rather that `cross` also switched to the correct command form.

Could you please rebase this patch atop current master so that the CI hopefully shows clean?  I don't appear to be able to restart it.

Emilgardis:
I'm fine with changing the command used in cross, but I think it makes sense to make this work as well.

rebased

Emilgardis:
We actually already use `toolchain install` in the suggested change to use `--force-non-host`, but this pr doesn't touch that. Fixed that now :D

PR #2978, opened 2022-04-14
Don't prepend CARGO_HOME/bin unnecessarily

### LINKED ISSUE #7431: Cargo should not prepend `$HOME/.cargo/bin` to `PATH` for subcommands when it's already in `PATH`, to avoid unnecessary recompilations
`cargo` is [prepending `$HOME/.cargo/bin`/`%HOME%\.cargo\bin` to `PATH`](https://github.com/passcod/cargo-watch/issues/106#issuecomment-445569998), even if `PATH` already contains this dir.
Cargo should only prepend it if it's not already in `PATH`!
Currently this behavior causes unnecessarily recompilations when any dependency uses `rerun-if-env-changed=PATH`, such as [`libsqlite3-sys`](https://github.com/jgallagher/rusqlite/issues/435).
(I have to wait several minutes every time when switching between `cargo check` in my editor and having `cargo watch -x run` running in cmd.exe, because cargo passes an unnecessarily modified PATH to cargo-watch, even though my PATH already contains `%HOME%\.cargo\bin`.)

jonhoo:
The current logic forces nested invocations to execute `cargo` and
friends from `$CARGO_HOME/bin`. This makes rustup break in environments
where the appropriate rustup proxies to use happen to be installed
elsewhere (earlier on `$PATH`), and using the binaries in
`$CARGO_HOME/bin` would not work correctly.

It also ensures that Rustup won't change `$PATH` "just for the heck of
it", which _should_ help reduce unnecessary re-compilations when
downstream build logic notices that `$PATH` changes (since it will no
longer).

Helps with #2848.

Fixes rust-lang/cargo#7431.

### DIFF
--- a/src/env_var.rs
+++ b/src/env_var.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::env;
 use std::path::PathBuf;
 use std::process::Command;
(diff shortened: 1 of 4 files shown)


jonhoo:
The armv7 build failure appears to be a Docker problem. Daniel's take (on Discord) was:
> That can happen in the rust-lang/rust repo doesn't quite do the right things
> It'll resolve after the next nightly build I think

The FreeBSD failure appears to be a time-out unrelated to this change.

GuillaumeGomez:
For the timeout issue on cirrus-ci, you can increase the build time limit like this:

```
task: 
  timeout_in: 120m
```

Issue #2581, opened 2020-11-28
Investigate and implement support for TLS1.3 in `rustup-init.sh`


kinnison:
Whatever fronts `https://rustup.rs` now supports TLS1.3 - we should look into what it'd take to detect support in the system `curl` or `wget` binaries in order to adjust `rustup-init.sh` to support TLS1.3

Also we should see if there's any hope we can detect the TLS version used to visit https://rustup.rs/ and offer a TLS1.3 install command to copy/paste if that's possible.

sanmai-NL:
I think we're good to go already. We use [`--tlsv1.2`](https://github.com/rust-lang/rustup/blob/e80ad246b33cf49d93ef74c87c4d5f5611a02ba7/rustup-init.sh#L436) which means `curl` [uses TLS v1.3 by preference](https://curl.se/mail/archive-2018-12/0013.html
), if it knows about it. As for Wget, its behavior [isn't documented as such](https://www.gnu.org/software/wget/manual/wget.html#HTTPS-_0028SSL_002fTLS_0029-Options) but it [seems to do the same](https://stackoverflow.com/questions/59788637/how-to-force-wget-to-use-tls1-2).

kinnison:
That's still opportunistic.  If we want to *force* 1.3 we'd need to say `--tlsv1.3` surely?

sanmai-NL:
Only if you decide that you needn't be compatible with with older clients. But that is contrary to the logic you describe in your opening post. Instead, you may want to turn this into a feature to warn users about imperfect security during their usage.

BryanQuigley:
Pretty sure all modern browsers will use TLS1.3, so detecting that their browser supports it won't be very predictive - although I see in [OS detection is](https://github.com/kinnison/rustup/commit/615ed4e265c702cdc2ad025e944a92d8068abde2) already done for some. 

Curl's [changelog](https://curl.se/changes.html) indicates tls1.3 was added in 7.52.0 - December 21 2016.

A quick [pkgs search](https://pkgs.org/search/?q=curl) indicates that a lot would be fine. Highlighting the major ones (IMHO) that would not support tls1.3:
CentOS 6/7 (without backports),  Ubuntu 14.04 ESM/16.04. Slackware 14.1/14.2 (really old?)

Both Cent/RHEL 6 and Ubuntu 14.04 are both in some sort of enterprisey life support mode. Ubuntu 16.04 goes enterprisey life support mode in April 2021. CentOS 7 in 2024.

[...]

PR #2996, opened 2022-05-26
Use TLSv1.3 with curl if specified at all

pothos:
The curl option specified to use TLSv1.2 explicity while nowadays 1.3
is availalble and recommended.
Switch to specifying 1.3 instead of 1.2 for the command that downloads 
the install script. The rustup-init.sh script itself it left with the 
ciphersuite selection plus 1.2 and fallbacks as is.

See https://github.com/rust-lang/book/pull/3130

### DIFF
--- a/doc/src/installation/index.md
+++ b/doc/src/installation/index.md
@@ -49,7 +49,7 @@ installation of `rustup` and then install `nightly` along with `clippy` or
 `miri`, first install `rustup` without a toolchain:
 
 ```console
-curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y
+curl --proto '=https' --tlsv1.3 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y
 ```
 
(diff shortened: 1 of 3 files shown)


pothos:
Related, can someone point me at the repo for the website source code? The 1.2 option is used there, too: https://rustup.rs/

Edit: Found more places: https://github.com/search?q=org:rust-lang+rustup+tlsv1.2&type=code

pothos:
I didn't touch the logic for ciphersuite selection with 1.2 and the fallbacks. Should we create an issue for that? Edit: No, it already exists: https://github.com/rust-lang/rustup/issues/2581

rbtcollins:
What are the compatibility implications of this? My understanding is that the reason we specify the tls parameter is defence-in-depth to prevent down grade attacks, but we do presumably want some somewhat older OSes to be able to install still?

pothos:
This got discussed in https://github.com/rust-lang/www.rust-lang.org/pull/1670 - I'm fine with closing this if it shouldn't stay open for the future.

kinnison:
Closed - we can revisit if/when tls1.3 is sufficiently widespread that we're not risking hurting people

PR #3005, opened 2022-06-12
Add safe-directories.

ehuss:
This is an implementation of [RFC 3279](https://github.com/rust-lang/rfcs/pull/3279).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2033,6 +2033,7 @@ dependencies = [
  "serde",
  "sha2 0.10.2",
  "sharded-slab",
+ "shell-escape",
  "strsim 0.10.0",
  "tar",
  "tempfile",
@@ -2274,6 +2275,12 @@ checksum = "900fba806f70c630b0a382d0d825e17a0f19fcd059a2ade1ff237bcddf446b31"
 dependencies = [
  "lazy_static",
 ]
+
+[[package]]
+name = "shell-escape"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "45bb67a18fa91266cc7807181f62f9178a6873bfad7dc788c42e6430db40184f"
 
 [[package]]
 name = "signature"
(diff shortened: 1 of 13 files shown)


ehuss:
Questions/Notes:
* I did not add any sort of "unstable" flag to enable this feature. Which approach do you think would be good?
    * Merge this on a separate branch, and create a testing release from that?
    * Merge on master, and revert it if a release needs to be made before we commit to this feature?
    * Add some unstable gate, such as a CLI flag or environment variable like RUSTUP_UNSTABLE_SAFE_DIRECTORIES=true to mirror the Cargo one.
      This might be the most desirable, but requires changing a bunch of things to support it.

* Should I bump DEFAULT_METADATA_VERSION? I think the only consequence of not doing that is that an older version of rustup may delete the safe_directories settings. If it is bumped, then older rustups will just fail, which seems worse.

* I can't tell why some code uses macros like info! and others use a `Notification`.

* I can't tell why some code is in `rustup_mode.rs` and some is in `common.rs`.

* The ownership checking code is copied from Cargo until a new `cargo_util` is published.

kinnison:
How did the `cargo` part of this go Eric?  I'm reluctant to start validating this until we know Cargo makes sense.

ehuss:
I need to follow up on some things, and I have been procrastinating. 🐌

rbtcollins:
Perhaps rustup and cargo should share their safe directories flag? Would we ever want them to disagree?

ehuss:
I'm not sure I understand the question fully. Can you say more about what you mean?

If it is about why there is a separate `CARGO_SAFE_DIRECTORIES` and `RUSTUP_SAFE_DIRECTORIES`, the reason is because cargo can be used without rustup (and it would be a little strange to set a `RUSTUP_` setting without rustup). However, if you are using rustup, then the intent is that the user should only configure `RUSTUP_SAFE_DIRECTORIES` (because cargo will read that) so in essence there is only one setting for most users.

I suppose it may be possible to drop `RUSTUP_SAFE_DIRECTORIES` and have rustup read `CARGO_SAFE_DIRECTORIES`, but that feels potentially confusing.

rbtcollins:
I'm hoping to entirely remove relative-path toolchain directories: I've asked around a bit and so far found no use case for them.

That then leaves absolute path toolchain directories - this seems safe to me; would welcome thoughts on that.

PR #3009, opened 2022-06-13
Fix unused imports warning

ChrisDenton:


ehuss:
I think a possibly simpler solution is the following:

```diff
diff --git a/src/cli/self_update.rs b/src/cli/self_update.rs
index a4c00d57..f6278dad 100644
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -462,7 +462,7 @@ pub(crate) fn install(
     })();

     if let Err(e) = install_res {
-        common::report_error(&e);
+        report_error(&e);

         // On windows, where installation happens in a console
         // that may have opened just for this purpose, give
```

This line is used on all platforms, and will make the unused import error go away.

ChrisDenton:
Good point!

ehuss (approved):

rbtcollins (approved):
I think the code change is fine, but the title - and thus the PR merge commit - is confusing.

ChrisDenton:
Oh sorry, I should have thought more about that. Is this better?

PR #3010, opened 2022-06-14
Fix detection of aarch64 host on Windows

arlosi:
`rustup` does not currently auto-detect `aarch64-pc-windows-msvc` hosts. Instead `x86_64-pc-windows-msvc` is used because `GetNativeSystemInfo` does not support detection of aarch64 hosts.

From the docs:
> If the function is called from an x86 or x64 application running on a 64-bit system that does not have an Intel64 or x64 processor (such as ARM64), it will return information as if the system is x86 only if x86 emulation is supported (or x64 if x64 emulation is also supported).

The alternative function used in this change [IsWow64Process2](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process2) supports detecting aarch64 hosts.

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -229,27 +229,80 @@ impl TargetTriple {
         #[cfg(windows)]
         fn inner() -> Option<TargetTriple> {
             use std::mem;
-            use winapi::um::sysinfoapi::GetNativeSystemInfo;
 
-            // First detect architecture
-            const PROCESSOR_ARCHITECTURE_AMD64: u16 = 9;
-            const PROCESSOR_ARCHITECTURE_INTEL: u16 = 0;
-            const PROCESSOR_ARCHITECTURE_ARM64: u16 = 12;
+            /// Get the host architecture using `IsWow64Process2`. This function
+            /// produces the most accurate results (supports detecting aarch64), but
+            /// it is only available on Windows 10 1511+, so we use `GetProcAddress`
+            /// to maintain backward compatibility with older Windows versions.
+            fn arch_primary() -> Option<&'static str> {
+                use winapi::shared::minwindef::BOOL;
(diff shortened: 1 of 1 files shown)


arlosi:
I should note that `IsWow64Process2` is only available on Windows 10 or newer. Is that acceptable? I couldn't find the minimum requirements for rustup.

If we need to go back further, I can do runtime detection of whether the function is available.

ChrisDenton:
Note that Rust itself currently supports Windows 7+. Presumably rustup should attempt to do the same unless the maintenance burden gets too high.

src/dist/dist.rs
```diff
@@ -328,12 +334,12 @@ impl TargetTriple {
         let ret = if partial_self.os != partial_other.os {
             false
         } else if partial_self.os.as_deref() == Some("pc-windows") {
-            // Windows is a special case here, we know we can run 32bit on 64bit
-            // and we know we can run gnu and msvc on the same system
-            // We don't immediately assume we can cross between x86 and aarch64 though
+            // Windows is a special case here: we can run gnu and msvc on the same system,
+            // i686 can run on x86_64, and both x86_64 and i686 can run on aarch64 through emulation
```
dpaoliello: x86_64 can only run emulated on AArch64 on Win11 - you can use `GetMachineTypeAttributes` to check if the device can run x86_64 code (although that API was added in Win11, so maybe moot point?).

arlosi: Thanks. I'll change it to "aarch64 can only run i686" to be conservative here.

arlosi:
I've updated the PR to use runtime detection of `IsWow64Process2` with a fallback to the original method using `GetNativeSystemInfo`

dpaoliello (approved):

PR #3012, opened 2022-06-19
Correctly propagate subshell failures in rustup-init

SabrinaJewson:
If `mktemp -d` were to fail in the current script, `rustup-init.sh` would report the error but keep running anyway. This fixes that by checking for a nonzero exit status and if so exiting the outer shell.


kinnison (approved):
This looks good, thank you for catching this odd corner case

PR #3032, opened 2022-07-12
Unify the msg of the prompt when downloading self-update

0xPoe:
We prompt for `self-updates` on `rustup update` but `self-update` on `rustup self update`. We should unify it.


kinnison:
While I agree with the premise, I think I'd prefer to normalise to `self-update` since we only download one thing (the new rustup)

0xPoe:
> We prompt for `self-updates` on `rustup update` but `self-update` on `rustup self update`. We should unify it.

Make sense to me.

kinnison (approved):

PR #3044, opened 2022-07-19
Add `up` as an alias of the `update` command

jbalintbiro:


jbalintbiro:
The failing Linux x86_64 build seems to be unrelated to the PR, it failed to download the docker image `sha256:9c23ba7777b9bca87c073017ab67e29c9ac1eb2e152bbd534f50d10ed40a637e` before the build.

kinnison (approved):
This looks as we agreed

EwoutH:
Is there documentation, a reference issue or other information on this change available? Since the PR  and commit body don’t contain any motivation.

kinnison:
@EwoutH based on the timing and the way it was merged, I'd say this was discussed on the Rustup discord channel.  Without digging through discord history I'd say it's likely that the submitter asked if we'd mind having the short alias and I wouldn't have objected to it and so they submitted the PR.  What are you hoping for in terms of documented history?

EwoutH:
Mainly motivation / proportionality for adding the `up` alias. It seems the first abbreviation alias as far as I know.

danielsilverstone-ct:
Motivation - Someone wanted it enough to talk to us about it, justified it enough at the time for me to say "file a PR" and then they cared enough to actually file the PR including test code.

Proportionality - as abbreviated aliases go, it's far far less of a problem than `rustup install` being an "alias" for `rustup toolchain install`.  I know a lot of people like the cargo short-aliases (eg `cargo b` `cargo t` etc) and most people only use the `update` feature of rustup on a regular basis.  So `rustup up` kinda makes sense to have.  These days I'd probably ask for it to be a hidden alias rather than a visible one, but I don't see why I'd deny its existence.

PR #3047, opened 2022-07-25
Add DisplayVersion for rustup to registry on Windows.

### LINKED ISSUE #3055: Register Rustup version in Windows registry for add/remove programs

In the windows add/remove programs (or apps and features) screen, the version of Rustup isn't registered. This can be pointed to the following block of code:

https://github.com/rust-lang/rustup/blob/6bc5d2c340e1dd9880b68564a19f0dea384c849c/src/cli/self_update/windows.rs#L526-L531

This makes it impossible for the [Windows Package Manager CLI (aka winget)](https://github.com/microsoft/winget-cli) to detect updates:

```cmd
winget upgrade --include-unknown                                                          
Name                                 Id              Version Available Source                                           
-----------------------------------------------------------------------------                                           
Rustup: the Rust toolchain installer Rustlang.Rustup Unknown 1.25.5    winget                                           
1 upgrades available.                                                                                                   
```

### Solution you'd like

Add the `DisplayVersion` string to the registry containing the rustup version (e.g. 1.25.1

### Notes

Note that the commit hash should be left out from the DisplayVersion as it is a hash and doesn't have a particular order.

jedieaston:
Hello! I'm a contributor to the Windows Package Manager Community Repository ([microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs)), and we recently had an issue posted to that repo which revealed that Rustup was not publishing a `DisplayVersion` to Add and Remove Programs during its install (although it was posting its `DisplayName`). 

The `DisplayVersion` is how winget and other tools determine when updates to software are needed, so it is highly recommended to set it whenever possible. This PR does just that, by making it so when Rustup is installed, it adds its version number to the registry alongside its name. 

This is the most Rust I've ever written, so let me know if I did something wrong and I can fix it :)

Tested: manually by compiling a new version of Rustup and installing it in a VM. If this needs more strenuous tests, I'm happy to oblige.

Resolves: https://github.com/rust-lang/rustup/issues/3055

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -523,10 +523,14 @@ pub(crate) fn do_add_to_programs() -> Result<()> {
         vtype: RegType::REG_SZ,
     };
 
+    let current_version: &str = env!("CARGO_PKG_VERSION");
+
     key.set_raw_value("UninstallString", &reg_value)
         .context("Failed to set uninstall string")?;
     key.set_value("DisplayName", &"Rustup: the Rust toolchain installer")
         .context("Failed to set display name")?;
+    key.set_value("DisplayVersion", &current_version)
+        .context("Failed to set display version")?;
 
     Ok(())
 }


LeonarddeR:
Thanks @jedieaston

I had exactly the same issue. My approach is slightly different:

```rust
    key.set_value("DisplayVersion", &common::version())
        .context("Failed to set display version")?;
```

Note that this sets the version in the form as outputted by `rustup --version` (i.e. it includes the commit hash and release date). As long as every new release changes either one of the major.minor.patch numbers that shouldn't be a problem.

jedieaston:
We prefer for there not to be extraneous information in the DisplayVersion outside of the number on the winget side, as it makes it easier to do comparisons for upgrades. If it is common to do multiple releases of Rustup under the same exact version number I can change it, but I think any stable releases would increment the `CARGO_PKG_VERSION` which is always just a number (at least, according to [this](https://github.com/rust-lang/rustup/blob/master/CONTRIBUTING.md#version-numbers)).

PR #3049, opened 2022-07-27
Add Fig as an autocomplete method

brendanfalk:
Hey there! [Fig](https://fig.io) provides autocomplete for 300+ CLI tools including rustup. It looks like this:

image

We think this will make users more efficient with rustup CLI, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!

### DIFF
--- a/doc/src/installation/index.md
+++ b/doc/src/installation/index.md
@@ -100,3 +100,12 @@ For `zsh`, you must then add the following line in your `~/.zshrc` before
 ```zsh
 fpath+=~/.zfunc
 ```
+
+### Fig
+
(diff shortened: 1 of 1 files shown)


kinnison:
The way you amended the docs here suggests that we (Rustup) can affect Fig's autocompletions because we're saying the support is there - is that true? is fig using *our* completions data or is it somehow built into Fig without reference to the Rustup itself?

rbtcollins:
I don't want `rustup` to get into the business of advertising complementary products: we don't have the bandwidth to pick winners, and if we don't pick then we will be creating a canvas for adverts for an unbounded collection. 

Documenting what `rustup` has specific code for, or common questions from users to reduce our support bandwidth are qualitatively different choices.

PR #3051, opened 2022-07-29
Make auto-self-update setting work when no-self-update feature is enabled

### LINKED ISSUE #3040: "self-update is disabled for this build of rustup" message on every `rustup update` is annoying

I run `rustup update` and other commands to update the software on my PC every day. I know my installation of rustup is managed by my package manager and don't need to be reminded that it can't update itself every time I want to update my Rust installation.

### Solution you'd like

A config option to suppress the message would work for me.

### Notes

As someone who prefers to use software packaged by my distro, a big reason this is grating to me is because it reminds me of software that intentionally degrades the user experience when packaged by people other than its author.

autumnontape:
Fixes #3040.

This patch series makes rustup work in the way that makes the most sense to me personally when built with self updates disabled. I've split four distinct, individually conservative changes I'd make into four patches. If you'd like me to drop or modify some of them, that's fine, of course. The first patch ("don't log on update if no-self-update is enabled but auto-self-update=disable") is the one that matters to me.

If you'd like docs or tests changes, I can do that, too, but no-self-update already doesn't seem to be covered by either. Enabling the feature causes the tests to fail even without these patches.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1032,11 +1032,11 @@ fn update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
         cfg.temp_cfg.clean();
     }
 
-    if !self_update::NEVER_SELF_UPDATE && self_update_mode == SelfUpdateMode::CheckOnly {
+    if self_update_mode == SelfUpdateMode::CheckOnly {
         check_rustup_update()?;
     }
 
-    if self_update::NEVER_SELF_UPDATE && self_update_mode != SelfUpdateMode::Disable {
+    if self_update::NEVER_SELF_UPDATE && self_update_mode == SelfUpdateMode::Enable {
         info!("self-update is disabled for this build of rustup");
         info!("any updates to rustup will need to be fetched with your system package manager")
     }
@@ -1610,8 +1610,7 @@ fn set_profile(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
 fn set_auto_self_update(cfg: &mut Cfg, m: &ArgMatches<'_>) -> Result<utils::ExitCode> {
     cfg.set_auto_self_update(m.value_of("auto-self-update-mode").unwrap())?;
 
-    let new_self_update_mode = cfg.get_self_update_mode()?;
-    if self_update::NEVER_SELF_UPDATE && new_self_update_mode != SelfUpdateMode::Disable {
+    if self_update::NEVER_SELF_UPDATE && cfg.get_self_update_mode()? == SelfUpdateMode::Enable {
         let mut args = crate::process().args_os();
         let arg0 = args.next().map(PathBuf::from);
         let arg0 = arg0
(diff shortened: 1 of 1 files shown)


rbtcollins:
I'd quite like to remove the no self update feature entirely.

That said, can you explain the user interaction model you are aiming for here?

autumnontape:
@rbtcollins Remove it in favor of what? Do you not want distros to package rustup?

The idea behind these patches is this:

- rustup built with no-self-update should work the same as rustup built without it as much as possible, hence why I made it able to check for self updates but not install them.
- rustup built with no-self-update should always warn the user when it can't achieve the same behavior as rustup built without it, and only then, so it shouldn't warn if auto-self-update is disable or check-only.

autumnontape:
If you're wondering more about how I expect the messages to be helpful to users, this is what I'm envisioning:

1. A user installs rustup through a package manager.
2. They run `rustup update` and see the message.
3. They now understand that they can update rustup through their package manager if they didn't before, and they can follow the message's directions so they won't have to see it in the future.
4. If they're ever led to turn auto-self-update back on, a warning reminds them that it won't work with their build of rustup.

I do think a docs change to mention that rustup won't update itself when installed through a package manager would also be nice, if people were confused enough by the docs for a message to be added to the program itself.

kinnison:
I think I'd like to review this when I have more brain-space.  I *am* aware that this can be annoying (I actually use rustup via nixos on some of my systems so I'm used to seeing the message, but that doesn't mean it's entirely pleasant to keep encountering).

autumnontape:
It still bothers me seeing this message every day. Are you interested in a change like this?

rami3l:
@autumnontape Thanks for asking! I'm personally using Homebrew's source-distributed build, with `no-self-update` enabled, and I'm hitting the same notifications every day. That said, I'm also even more bothered by Codex CLI's "there's a new version available but we can't do the update for you here", so personally I don't think we should move towards that direction, sorry.

OTOH, actually as you might have seen this is not the previous project team anymore so the opinions might have shifted a bit... so I'd like to listen to the current team's opinion: cc @djc @ChrisDenton @FranciscoTGouveia

autumnontape:
I could drop the third commit if you like. My reasoning for it was that `enable` does strictly more than `check-only`, which does strictly more than `disable`, so it seemed intuitive for `enable` to degrade to the next-highest-power behavior when it couldn't perform all of its functions. But I won't have `auto-self-update=enable` on any of my systems, so its behavior doesn't concern me.

ChrisDenton:
> so I'd like to listen to the current team's opinion

Hm, I don't think the message as it currently stands is particularly actionable by the user. Ok, so they're not on the very latest version but there's not really anything they can do about that (short of uninstalling the packaged rustup and manually installing the latest). Telling them repeatedly won't change that. If it's handled by their package manager then presumably it will get the update soon enough.

I do get the argument that instructions on the website are for the latest version so it may be surprising if some things don't work as documented. But nowadays I suspect that's a rare enough occurrence that it's not worth bothering everyone because of it. Maybe I can see an argument if rustup is way out of date but even then it feels excessive to keep constantly telling the user that.

FranciscoTGouveia:
I believe the message is important when the user explicitly runs `rustup self update`; in that case, we should clearly explain why it doesn't work.

However, showing it on every `rustup update` feels unnecessary.
If the user didn't ask to self-update, the warning just becomes noise.

So I would favor keeping the message for `rustup self update` only and avoiding emitting it during regular `rustup update` executions.

djc:
> So I would favor keeping the message for `rustup self update` only and avoiding emitting it during regular `rustup update` executions.

This sounds reasonable to me.

PR #3067, opened 2022-09-06
Add details of each setting in the toolchain file

nicholasbishop:
The format of the new section is modeled after the cargo docs, for example:
https://doc.rust-lang.org/cargo/reference/profiles.html#profile-settings

This is mostly a reorganization of the existing content to make it
easier to scan for relevant details, but a few new details have also
been added:

1. For `channel`, the format is now explained in the same format as in
   concepts/toolchains.md, but with the more limited form that
   `rust-toolchain.toml` accepts. This is based on the text that was
   here previously; I haven't verified if the code is actually this
   strict.
2. The `profile` section mentions that the default profile is not
   necessarily the profile named `default`, which is subtle detail that
   can be confusing.
3. The `components` section notes that the list is additive with the
   profile's list of components.
4. The `targets` field is now documented, previously it wasn't mentioned
   other than to say it's ignored with `path`.

### DIFF
--- a/doc/src/overrides.md
+++ b/doc/src/overrides.md
@@ -113,23 +113,70 @@ that uses the new TOML encoding in the `rust-toolchain` file. You need to upgrad
 The `rust-toolchain.toml`/`rust-toolchain` files are suitable to check in to
 source control.
 
-The toolchains named in these files have a more restricted form than `rustup`
-toolchains generally, and may only contain the names of the three release
-channels, 'stable', 'beta', 'nightly', Rust version numbers, like '1.0.0', and
-optionally an archive date, like 'nightly-2017-01-01'. They may not name
-custom toolchains, nor host-specific toolchains. To use a custom local
-toolchain, you can instead use a `path` toolchain:
+### Toolchain file settings
 
-``` toml
-[toolchain]
-path = "/path/to/local/toolchain"
+#### channel
+
+The `channel` setting specifies which [toolchain] to use. The value is a
+string in the following form:
+
+```
+<channel>[-<date>]
+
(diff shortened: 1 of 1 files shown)


doc/src/overrides.md
```diff
@@ -113,23 +113,71 @@ that uses the new TOML encoding in the `rust-toolchain` file. You need to upgrad
...
+
+#### profile
+
+The `profile` setting names a group of components to be installed. The
+value is a string. The valid options are: `minimal`, `default`, and
+`complete`. See [profiles] for details of each.
+
+Note that if not specified, the `default` profile is not necessarily
+used, as a different default profile might have been set with `rustup
+set profile`. For example, Github's CI runners use the `minimal` profile
```
0xPoe: If you use other actions like https://github.com/actions-rs/toolchain#inputs, it will be `default` instead of minimal. So I prefer not to mention this here.

nicholasbishop: OK, I have removed that sentence.

0xPoe (approved):
Looks good to me. Thanks for your patch!
But I am not a native English speaker, so I'd like to wait for approval from @kinnison or @rbtcollins before we merge it.

0xPoe:
You can check out this [rendered page](https://github.com/nicholasbishop/rustup/blob/bishop-specify-additive/doc/src/overrides.md).

nicholasbishop:
Friendly ping, would love to get this merged :)

0xPoe:
I am going to merge this PR. If we find some grammar issues we can fix them later. Thanks! Sorry for the long wait.

PR #3073, opened 2022-09-21
Update home usage to new env fn

LucioFranco:
This is the corresponding PR for https://github.com/brson/home/pull/29 which updates how rustup uses the `home` crates `Env` features for testing. 

This is still a draft PR pending a new release of the home crate with the new features.

cc @kinnison

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -974,8 +974,9 @@ dependencies = [
 
 [[package]]
 name = "home"
-version = "0.5.3"
-source = "git+https://github.com/rbtcollins/home?rev=a243ee2fbee6022c57d56f5aa79aefe194eabe53#a243ee2fbee6022c57d56f5aa79aefe194eabe53"
+version = "0.5.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


rbtcollins:
Sure, looks fine I guess - aesthetically it has stuttering. I realise this isn't a rust style guide issue but it is one that I find very helpful in making nice APIs. (And not the extreme version presented in some golang talks!).

But not having to maintain home is much more beneficial than a few extra env:s here and there.

That said, I think this would be a smaller patch if instead of the insertion of `::env` you added `use home::env as home` at the top of each affected source file.

LucioFranco:
> That said, I think this would be a smaller patch if instead of the insertion of ::env you added use home::env as home at the top of each affected source file.

Oh that is a great idea, I didn't think of that, will make that change.

kinnison (approved):
As it stands, I think that this is clean and clear.  Once a new `home` release is done, I look forward to this being updated for the new `home` version.

LucioFranco:
Hm I feel like the ring failure has nothing to do with my changes?

0xPoe:
> Hm I feel like the ring failure has nothing to do with my changes?

Yes. See: https://github.com/rust-lang/rust/pull/102332#issuecomment-1272730482

0xPoe (approved):
Thanks!

Issue #3078, opened 2022-10-04
Reconsider recommending `--profile minimal` for CI (or add CI tools to the minimal profile)


Victor-N-Suadicani:
### Problem you are trying to solve

The official Rust Docker images do not contain clippy and rustfmt. This has been the [source](https://github.com/rust-lang/docker-rust/issues/16) of [several](https://github.com/rust-lang/docker-rust/issues/37) [issues](https://github.com/rust-lang/docker-rust/issues/97).

As far as I can tell, they haven't been added because the Rust Docker images should [only install what is provided by default by rustup](https://github.com/rust-lang/docker-rust/issues/37#issuecomment-445292336) (which seems fair). And since [rustup recommends the minimal profile for CI](https://rust-lang.github.io/rustup/concepts/profiles.html), it seems fair that this is the one used by the Rust Docker images (since such Docker images are often used in CI).

However I think this is a bad recommendation. Linting (clippy) and code format verification (rustfmt) are very, very common tasks in CI. Recommending a profile for CI that doesn't have the capacity to lint code or verify its format is bound to eventually lead to problems when users try to do those things (which just happened to me, and probably why there are all those previously mentioned issues in the docker-rust repository).

### Solution you'd like

Rustup ought to recommend a profile for CI that includes the capacity to lint and format code. This can be achieved in any number of ways:

1. Add rustfmt and clippy to the minimal profile (probably not desirable)
2. Recommend the default profile for use in CI.
3. Create a new profile specifically tailored to CI that includes rustfmt and clippy.

I don't really care which solution is chosen - the point is that given such a recommendation, I think the clear choice for the Rust Docker images would be to use the image that includes rustfmt and clippy, which would solve the problem as stated above and solve the issues in the docker-rust repository.

rbtcollins:
Its still not clear what you're trying to solve.

Is it build time in CI?

Is it the issue with having rustup working across multiple layers if you modify components post install?

Certainly default is not a good profile for CI in my view : default includes documentation which is 10's of thousands of files that noone will ever read.

Victor-N-Suadicani:
The problem I'm trying to solve is what is mentioned in the linked issues, such as https://github.com/rust-lang/docker-rust/issues/97

[...]

PR #3080, opened 2022-10-07
Recommend the default profile for CI

Victor-N-Suadicani:
Solution number 2 for #3078. See that issue for details.


rbtcollins:
I don't think we have consensus on the recommendation question at this point. I can say that minimal was created to be suitable for CI, so if its not, I think a better approach is to revisit the definition of minimal, rather than changing the recommendation

Victor-N-Suadicani:
That's fair :)

PR #3089, opened 2022-10-12
Add automatic resume flag when retrying download with curl

darkkeh:
Should fix the resume issue that caused #2853 to be reopened.

Simply adds the resume flag `-C` and tells curl to automatically figure out where to resume the download with `-`.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -607,14 +607,17 @@ check_help_for() {
 
 # Check if curl supports the --retry flag, then pass it to the curl invocation.
 check_curl_for_retry_support() {
-  local _retry_supported=""
-  # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
-  if check_help_for "notspecified" "curl" "--retry"; then
-    _retry_supported="--retry 3"
-  fi
+    local _retry_supported=""
+    # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
+    if check_help_for "notspecified" "curl" "--retry"; then
+        _retry_supported="--retry 3"
+        if check_help_for "notspecified" "curl" "--continue-at"; then
+            # "-C -" tells curl to automatically find where to resume the download when retrying.
+            _retry_supported="--retry 3 -C -"
+        fi
+    fi
 
-  RETVAL="$_retry_supported"
-
+    RETVAL="$_retry_supported"
 }
 
 # Return cipher suite string specified by user, otherwise return strong TLS 1.2-1.3 cipher suites


0xPoe (commented):
Thanks for working on this! 
How can I test it? Did you do some tests for this patch?

darkkeh:
Sorry for being a bit slow to respond yesterday ended up being busy!

I did some manual testing by kill switching my network with my VPN (adding a video below to show the behavior). 
[unknown.webm](https://user-images.githubusercontent.com/16773900/195946353-a2e3d2f9-1923-4048-8eac-668d5ac77bbb.webm)

I could look into adding an actual test case for this but I would probably need some guidance as I'm fairly new to rust and I wouldn't exactly be sure how to go about testing this properly

rustup-init.sh
```diff
@@ -610,7 +610,8 @@ check_curl_for_retry_support() {
   local _retry_supported=""
   # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
   if check_help_for "notspecified" "curl" "--retry"; then
-    _retry_supported="--retry 3"
+    # "-C -" tells curl to automatically find where to resume the download when retrying.
+    _retry_supported="--retry 3 -C -"
```
0xPoe: Do we need to check the `--continue-at` flag support?

0xPoe:
> I did some manual testing by kill switching my network with my VPN (adding a video below to show the behavior).
> [unknown.webm](https://user-images.githubusercontent.com/16773900/195946353-a2e3d2f9-1923-4048-8eac-668d5ac77bbb.webm)

Looks good. Thank you!



> I could look into adding an actual test case for this but I would probably need some guidance as I'm fairly new to rust and I wouldn't exactly be sure how to go about testing this properly

I don't think there's a way to test it in code, so manual testing is good enough.

rustup-init.sh
darkkeh: I can take a look at adding that tomorrow just to make sure, I did check when it was added when I was adding it and it seems to have been supported since 1998 I think it was? But probably better safe than sorry so I'll try to add a check for the support tomorrow

darkkeh:
Sorry that it took a while again ended up being more busy than expected.

Added the check to verify that `--continue-at` is available and cleaned up some indentation

0xPoe (approved):
Looks good to me. Could you please squash your PR and make it just one commit?

Thanks!

darkkeh:
There we go, commits have been squashed!

0xPoe:
Thanks again! 💚 💙 💜 💛 ❤️

PR #3092, opened 2022-10-14
Avoid removing partial paths when uninstalling

### LINKED ISSUE #2967: installer produced an invalid .bash_profile
### Problem

Installing rustup via the bash command messed up my .bash_profile.

The installer truncated the line `[ -f "$HOME/.cargo/env" ] && source "$HOME/.cargo/env"` to `[ -f "$HOME/.cargo/env" ] &&` (which is now applying the conditional to some unrelated line) and then appended `. "$HOME/.cargo/env"` to the end of the file.

Here it is as a patch, my original file in red(-) and the rustup installer's modifications in green(+):

```diff
--- a/.bash_profile
+++ b/.bash_profile
@@ -1,7 +1,6 @@
 [[ $- == *i* ]] && [ -f ~/.bashrc ] && source ~/.bashrc
 
-[ -f "$HOME/.cargo/env" ] && source "$HOME/.cargo/env"
-
+[ -f "$HOME/.cargo/env" ] && 
 # added by Anaconda3 5.0.0 installer
 export PATH="/Users/bronson/anaconda3/bin:$PATH"
 
@@ -9,3 +8,4 @@ export PATH="/Users/bronson/anaconda3/bin:$PATH"
 if [ -e "$HOME/.nix-profile/etc/profile.d/nix.sh" ]; then
   . "$HOME/.nix-profile/etc/profile.d/nix.sh"
 fi
+. "$HOME/.cargo/env"
```

### Steps

1. Add this line to the start of your .bash_profile: `[ -f "$HOME/.cargo/env" ] && source "$HOME/.cargo/env"`
2. Install rustup via bash: `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
3. That line has been modified ... poorly. now the conditional is mistakenly applied to whatever the next line is, and the `source` line has been moved the end of the file.

### Possible Solution(s)

I wish the installer wouldn't try to modify lines it doesn't understand.

### Notes

_No response_

### Rustup version

```console
info: The currently active `rustc` version is `rustc 1.59.0 (9d1b2106e 2022-02-23)`
```

### Installed toolchains

```console
rustup home:  /Users/bronson/.rustup

rustc 1.59.0 (9d1b2106e 2022-02-23)
```

x3ro:
Fixes #2967

* Adds a regression test
* Changes the behaviour of `do_remove_from_paths` so that it only removes full lines
* Added normalization of whitespaces (this could also be moved to a separate PR, or removed entirely)

One thing I'm not 100% sure about is whether it's okay to call `.lines()` and then later re-build the file with `.join("\n")`. For example in the case of CRLF line breaks, this would certainly rewrite the file entirely. Is this a concern for the Unix implementation?

I'm fairly sure this still needs some work, but the overall idea is correct, I believe. Feedback welcome :)

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -949,7 +949,7 @@ pub(crate) fn uninstall(no_prompt: bool) -> Result<utils::ExitCode> {
     info!("removing cargo home");
 
     // Remove CARGO_HOME/bin from PATH
-    do_remove_from_path()?;
+    do_remove_from_paths()?;
     do_remove_from_programs()?;
 
     // Delete everything in CARGO_HOME *except* the rustup bin
--- a/src/cli/self_update/unix.rs
+++ b/src/cli/self_update/unix.rs
@@ -3,11 +3,13 @@ use std::process::Command;
 
 use anyhow::{bail, Context, Result};
 
-use super::install_bins;
-use super::shell;
+use crate::cli::self_update::shell::Shell;
 use crate::process;
 use crate::utils::utils;
 use crate::utils::Notification;
+
+use super::install_bins;
+use super::shell;
 
 // If the user is trying to install with sudo, on some systems this will
 // result in writing root-owned files to the user's home directory, because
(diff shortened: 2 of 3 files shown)


rami3l:
Oops, commenting on the wrong thread, my apologies 🙏

PR #3064, opened 2022-08-21
Bump `clap` to 3.0

zohnannor:
Bump `clap` to v3 which fixes incorrect executable name printing in subcommand's help texts on Windows (it comes as `rustup.exe-subcommand` in v2). I don't think this issue was reported to `rustup` but https://github.com/clap-rs/clap/pull/3693
Add `clap_complete` as some of `clap`'s functionality was moved into this crate.
Fix related tests.

close https://github.com/rust-lang/rustup/issues/2920

PR #3094, opened 2022-10-15
Bump `clap` to 4.0

### LINKED ISSUE #3128: PowerShell completions installation does not work as documented
### Problem

Depending on the particular setup, the installation either doesn't work as documented (incorrect profile file, so it has no effect) or causes an error when a new PowerShell instance is opened, if the profile file is correct but not empty (in my case, it contained the single line  `Import-Module posh-git`):

```text
ParserError: D:\Users\<xxxxxxxx>\Documents\PowerShell\Microsoft.PowerShell_profile.ps1:4
Line |
   4 |  using namespace System.Management.Automation
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | A 'using' statement must appear before any other statements in a script.
```

For completeness sake, the output of `pwsh --version` is `PowerShell 7.3.1`

### Steps

1. Use the new PowerShell `pwsh` instead of the bundled PowerShell OR have your Documents library set to a different path
2. Install rustup PowerShell completions by following the instructions given by `rustup completions --help`

   ```powershell
   Test-Path $profile # shows, true, so I skip New-Item -path $profile -type file -force
   rustup completions powershell >> ${env:USERPROFILE}\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1
   ```
3. Open a new PowerShell instance and try to use the completions

### Possible Solution(s)

On my system, `${env:USERPROFILE}\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1` does not match `$profile` (or `$profile.CurrentUserAllHosts`). So in general, the command has no effect except creating an unused file. At the very, least, the documentation should be corrected to suggest running `rustup completions powershell >> $profile` and remove the text about `New-Item -path $profile -type file -force` creating the above file (it doesn't).

Since the generated completions are incompatible with a non-empty profile, I would go further to remove the suggestion of appending to the profile entirely, and suggest something like:

```powershell
rustup completions powershell > "$(Split-Path $profile)\rustup_completions.ps1"
Write-Output ". `"`$PSScriptRoot\rustup_completions.ps1`"" >> $profile
```

Addendum: Instead of `$profile` which evaluates to `$profile.CurrentUserCurrentHost`, it may make sense to also suggest  `$profile.CurrentUserAllHosts` as an alternative, although even that will not necessarily affect all PowerShell versions, since `pwsh` has its profile files under `D:\Users\<xxxxxxxx>\Documents\PowerShell\` for me, while the bundled `powershell` has its profiles under `D:\Users\<xxxxxxxx>\Documents\WindowsPowerShell\`, *neither* of which matches the documented `${env:USERPROFILE}\Documents\WindowsPowerShell` because I have my Documents library on a different drive:

Screenshot of the properties dialog for the Documents library showing the Location tab with a non-default location on drive D (with part of the path redacted)

### Notes

~~Regarding the second part of the bug, I have attempted to move the output into a separate file `rustup_completions.ps1` in the same directory as my profile script and sourcing it via `. "$PSScriptRoot\rustup_completions.ps1"`. This still has no effect. The code is definitely run as checked by adding a `Write-Output` to the script, suggesting a third part to this bug.~~ Update: The completions work when written to a separate file and sourced, but partial completions are unsupported since rustup uses clap 2 and partial completions are [only supported since 3.0.6](https://github.com/clap-rs/clap/pull/3409). This annoyance would be solved with https://github.com/rust-lang/rustup/pull/3094

### Rustup version

```console
```

### Installed toolchains

```console
rustup home:  C:\Users\<xxxxxxxx>\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc
1.34.0-x86_64-pc-windows-msvc

armv7-unknown-linux-gnueabihf
nvptx64-nvidia-cuda
thumbv6m-none-eabi
wasm32-wasi
x86_64-pc-windows-gnu
x86_64-pc-windows-msvc

active toolchain
----------------

```

zohnannor:
Depends on #3064 (see https://github.com/rust-lang/rustup/pull/3064#issuecomment-1276929460)

fixes #3128 apparently

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,17 +146,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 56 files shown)


0xPoe:
@zohnannor Friendly ping~

Could you please solve the conflicts and move forward with this PR? Thanks!

rami3l:
Superseded by #3444. Thanks for your efforts, @zohnannor!

PR #3101, opened 2022-10-29
Fix the android build

palfrey:
This was broken due to https://github.com/rust-lang/rust/pull/102332 but is now blocked on https://github.com/rust-lang/rust/pull/103744 for correct gcc vs unwind support on later NDKs


palfrey:
All sorted now the rust 1.66.0 release is out.

PR #3108, opened 2022-11-11
Add support for fish shell

### LINKED ISSUE #478: PATH support doesn't work for fish shell

WaffleLapkin:
This is an attempt to fix https://github.com/rust-lang/rustup/issues/478.

I'm not sure how to test if it works though, do I have to install local build of rustup in a vm somehow? 🤔

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -233,16 +233,7 @@ impl UnixShell for Fish {
     }
 
     fn update_rcs(&self) -> Vec<PathBuf> {
-        if let Ok(home) = process().var("XDG_CONFIG_HOME") {
-            let mut path = PathBuf::from(home);
-            path.push("fish/config.d/rustup.fish");
-            vec![path]
-        } else if let Some(mut path) = utils::home_dir() {
-            path.push(".config/fish/config.d/rustup.fish");
-            vec![path]
-        } else {
-            Vec::new()
-        }
+        self.rcfiles()
     }
 }
 
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -235,6 +235,17 @@ impl UnixShell for Fish {
     fn update_rcs(&self) -> Vec<PathBuf> {
         self.rcfiles()
     }
+
+    fn env_script(&self) -> ShellScript {
+        ShellScript {
+            name: "env.fish",
+            content: include_str!("env.fish"),
+        }
+    }
+
+    fn source_string(&self) -> Result<String> {
+        Ok(format!(r#". "{}/env.fish""#, cargo_home_str()?))
+    }
 }
 
 pub(crate) fn legacy_paths() -> impl Iterator<Item = PathBuf> {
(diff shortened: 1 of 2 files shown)


src/cli/self_update/shell.rs
```diff
@@ -201,6 +206,46 @@ impl UnixShell for Zsh {
...
+
+        p0.into_iter().chain(p1).collect()
+    }
+
+    fn update_rcs(&self) -> Vec<PathBuf> {
+        if let Ok(home) = process().var("XDG_CONFIG_HOME") {
+            let mut path = PathBuf::from(home);
+            path.push("fish/config.d/rustup.fish");
+            vec![path]
+        } else if let Some(mut path) = utils::home_dir() {
```
rbtcollins: Sorry for not getting to this earlier.

We generally try to write to all the places that a shell might read. The helper we wrote is idempotent to support this. What would happen if that was done here?

WaffleLapkin: Hm, to be honest I've lost all the context since I've opened this PR... Do you mean I should just call `rcfiles` here?

src/cli/self_update/shell.rs
```diff
@@ -201,6 +206,46 @@ impl UnixShell for Zsh {
...
+        matches!(process().var("SHELL"), Ok(sh) if sh.contains("fish"))
+            || matches!(utils::find_cmd(&["fish"]), Some(_))
+    }
+
+    // > "$XDG_CONFIG_HOME/fish/conf.d" (or "~/.config/fish/conf.d" if that variable is unset) for the user
+    // from <https://github.com/fish-shell/fish-shell/issues/3170#issuecomment-228311857>
+    fn rcfiles(&self) -> Vec<PathBuf> {
+        let p0 = process().var("XDG_CONFIG_HOME").ok().map(|p| {
+            let mut path = PathBuf::from(p);
+            path.push("fish/config.d/rustup.fish");
```
WaffleLapkin: This is a typo, as the comment suggests, this should be `conf.d`, not `config.d`

WaffleLapkin:
Hm, I've also noticed that the default `env` script doesn't work with fish, so I've written a custom one. `. ~/path/to/repo/rustup/src/cli/self_update/env.fish`  does add the desired path to the `$PATH` (after the `{cargo_bin}` replacement of course), so I think this should work.

src/cli/self_update/shell.rs
rbtcollins: approximately - look at zsh for the typical pattern

WaffleLapkin: @rbtcollins the current version calls `rcfiles` here.

Is there anything more that I can do here?

rami3l (approved):
The new version looks good!

@rbtcollins may I ask why this has been blocked from merging for so long?

srcrip:
Yeah same, is there any reason this can't be merged?

djc:
@sevensidedmarble there's no point in piling on -- please bear with us while we try to increase review capacity.

alex (approved):

YoutacRandS-VA (commented):
Loving the changes/updates

PR #3121, opened 2022-12-07
Fix shell check warning

0xPoe:
See: https://github.com/rust-lang/rustup/actions/runs/3622731921/jobs/6133180965
It may be that our shell checker is updated.

[SC1091](https://www.shellcheck.net/wiki/SC1091) is pretty annoying and it prints some warning in our test scripts. So I excluded it.

### DIFF
--- a/ci/cirrus-templates/script.bash
+++ b/ci/cirrus-templates/script.bash
@@ -1,4 +1,4 @@
-#!/bin/bash                                                                                            
+#!/bin/bash
 
 set -ex
 
@@ -26,6 +26,7 @@ echo "Install Rustup using ./rustup-init.sh"
 
 sh rustup-init.sh --default-toolchain=stable --profile=minimal -y
 # It's the equivalent of `source`
+# shellcheck source=src/cli/self_update/env.sh
 source "$HOME"/.cargo/env
 
 echo "========="
--- a/ci/fetch-rust-docker.bash
+++ b/ci/fetch-rust-docker.bash
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 script_dir=$(cd "$(dirname "$0")" && pwd)
+# shellcheck source=ci/shared.bash
 . "$script_dir/shared.bash"
 
 set -e
--- a/ci/raw_init.sh
+++ b/ci/raw_init.sh
@@ -3,5 +3,6 @@
 set -ex
 
 sh ./rustup-init.sh --default-toolchain none -y
+# shellcheck source=/dev/null
 . "$HOME"/.cargo/env
 rustup -Vv
--- a/ci/raw_init.sh
+++ b/ci/raw_init.sh
@@ -3,6 +3,6 @@
 set -ex
 
 sh ./rustup-init.sh --default-toolchain none -y
-# shellcheck source=/dev/null
+# shellcheck source=src/cli/self_update/env.sh
 . "$HOME"/.cargo/env
 rustup -Vv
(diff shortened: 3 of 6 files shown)


rbtcollins:
Hi! so SC1091 means shellcheck couldn't read the file. I can't see those action logs they are too far back; perhaps we can see them somewhere else? I think it would be good to fix the problem rather than disabling the checks.

ci/cirrus-templates/script.bash
```diff
@@ -26,6 +26,7 @@ echo "Install Rustup using ./rustup-init.sh"
 
 sh rustup-init.sh --default-toolchain=stable --profile=minimal -y
 # It's the equivalent of `source`
+# shellcheck source=/dev/null
```
rbtcollins: ok so this one is failing because the script wants to import from a file that won't exist until rustup is installed. But this is linted *after* rustup is installed, so I would expect it to find the source.

ci/fetch-rust-docker.bash
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 script_dir=$(cd "$(dirname "$0")" && pwd)
+# shellcheck source=/dev/null
```
rbtcollins: this should just work, but if its not, 
`# shellcheck source=ci/shared.bash` should

ci/raw_init.sh
```diff
@@ -3,5 +3,6 @@
 set -ex
 
 sh ./rustup-init.sh --default-toolchain none -y
+# shellcheck source=/dev/null
```
rbtcollins: similar to [ci/cirrus-templates/script.bash](https://github.com/rust-lang/rustup/pull/3121/files#diff-53a66607b98fd4e58616b06387fd72e3fc80f3874086f3bd3376878792e1833d) this should be working.

rustup-init.sh
```diff
@@ -597,7 +597,7 @@ check_help_for() {
     esac
 
     for _arg in "$@"; do
-        if ! "$_cmd" --help $_category | grep -q -- "$_arg"; then
+        if ! "$_cmd" --help "$_category" | grep -q -- "$_arg"; then
```
rbtcollins, line 600: We definitely need this fix

ci/fetch-rust-docker.bash
0xPoe: Yes, it works. Thanks!

ci/cirrus-templates/script.bash
0xPoe: I think it only checks the source file from this repo. So we need to find the `env` file in the source code, but it is generated from `src/cli/self_update/env.sh`. And I tried set it in https://github.com/rust-lang/rustup/pull/3121/commits/d74db478af4085aff37bf9d1310d84db2b7c2b75.

It failed: https://github.com/rust-lang/rustup/actions/runs/3888788652/jobs/6636457603

It seems shellcheck can not use it as `env`.

ci/cirrus-templates/script.bash
```diff
@@ -26,6 +26,7 @@ echo "Install Rustup using ./rustup-init.sh"
 
 sh rustup-init.sh --default-toolchain=stable --profile=minimal -y
 # It's the equivalent of `source`
+# shellcheck source=src/cli/self_update/env.sh
```
rbtcollins, line 29: Adding '-x' to the shellcheck invocation should fix this. See https://github.com/koalaman/shellcheck/issues/1012#issuecomment-359294974

0xPoe: Thank you a lot! It works.

rbtcollins:
Thank you for all the work on this

0xPoe:

PR #3127, opened 2022-12-28
Improve a help message for `rustup doc --rustdoc`

noritada:
This PR improves a help message for `rustup doc --rustdoc`.

Any suggestions for a better message or other changes we should make in conjunction with this would be appreciated.
Many thanks.

### Reason

It's not a fatal problem, but the current help message "Generate documentation for Rust projects" reads for a moment as if the command itself has an action to generate documentation.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1515,7 +1515,7 @@ const DOCS_DATA: &[(&str, &str, &str,)] = &[
     ("reference", "The Rust Reference", "reference/index.html"),
     ("rust-by-example", "A collection of runnable examples that illustrate various Rust concepts and standard libraries", "rust-by-example/index.html"),
     ("rustc", "The compiler for the Rust programming language", "rustc/index.html"),
-    ("rustdoc", "Generate documentation for Rust projects", "rustdoc/index.html"),
+    ("rustdoc", "Documentation generator for Rust projects", "rustdoc/index.html"),
     ("std", "Standard library API documentation", "std/index.html"),
     ("test", "Support code for rustc's built in unit-test and micro-benchmarking framework", "test/index.html"),
     ("unstable-book", "The Unstable Book", "unstable-book/index.html"),


hi-rustin:
Thanks for the patch!

> It's not a fatal problem, but the current help message "Generate documentation for Rust projects" reads for a moment as if the command itself has an action to generate documentation.

I think it is correct. It actually does the generate job.

You can check out https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html. It says `The standard Rust distribution ships with a tool called rustdoc. Its job is to generate documentation for Rust projects.`.

noritada:
@hi-rustin Thanks for your comments!

Sorry my explanation is not clear. Yes, the `rustdoc` command is a command that generates documentation. But my point is that `rustup doc --rustdoc` is only to open documentation about `rustdoc`, not to do generation.

In the description of the flags, `--help` and `--path` are flags that change the behavior, so their _behavior_ is described. But since the others are flags that display the documentation, their descriptions are basically displaying the _title and contents of the document_. `--rustdoc` is the only one where the _behavior_ is described.

```
FLAGS:
        --alloc              The Rust core allocation and collections library
        --book               The Rust Programming Language book
        --cargo              The Cargo Book
        --core               The Rust Core Library
        --edition-guide      The Rust Edition Guide
        --embedded-book      The Embedded Rust Book
    -h, --help               Prints help information
        --nomicon            The Dark Arts of Advanced and Unsafe Rust Programming
        --path               Only print the path to the documentation
        --proc_macro         A support library for macro authors when defining new macros
        --reference          The Rust Reference
        --rust-by-example    A collection of runnable examples that illustrate various Rust concepts and standard
                             libraries
        --rustc              The compiler for the Rust programming language
        --rustdoc            Generate documentation for Rust projects
        --std                Standard library API documentation
        --test               Support code for rustc's built in unit-test and micro-benchmarking framework
        --unstable-book      The Unstable Book
```

0xPoe (approved):
Looks good to me. Thanks a lot!

Happy new year!

PR #3096, opened 2022-10-19
Make clippy happy

0xPoe:
Try `cargo +nightly clippy --all --all-targets`.
[Remove needless borrow](https://github.com/rust-lang/rustup/commit/9d35fe07b131a3421eff984371337e0721de6394)
[Use auto deref](https://github.com/rust-lang/rustup/commit/3b18348225cf9937e22653e970313f2f2db20e97)

PR #3133, opened 2023-01-04
Fix clippy error

chansuke:
Fix clippy error.

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -47,12 +47,13 @@ fn download_with_backend(
     }
 }
 
+type EventResult<'a> = &'a dyn Fn(Event<'_>) -> Result<()>;
 pub fn download_to_path_with_backend(
     backend: Backend,
     url: &Url,
     path: &Path,
     resume_from_partial: bool,
-    callback: Option<&dyn Fn(Event<'_>) -> Result<()>>,
+    callback: Option<EventResult<'_>>,
 ) -> Result<()> {
     use std::cell::RefCell;
     use std::fs::remove_file;
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -47,13 +47,13 @@ fn download_with_backend(
     }
 }
 
-type EventResult<'a> = &'a dyn Fn(Event<'_>) -> Result<()>;
+type DownloadCallback<'a> = &'a dyn Fn(Event<'_>) -> Result<()>;
 pub fn download_to_path_with_backend(
     backend: Backend,
     url: &Url,
     path: &Path,
     resume_from_partial: bool,
-    callback: Option<EventResult<'_>>,
+    callback: Option<DownloadCallback<'_>>,
 ) -> Result<()> {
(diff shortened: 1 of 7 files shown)


src/cli/self_update.rs
```diff
@@ -89,7 +89,7 @@ pub(crate) const NEVER_SELF_UPDATE: bool = true;
 #[cfg(not(feature = "no-self-update"))]
 pub(crate) const NEVER_SELF_UPDATE: bool = false;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
0xPoe, line 92: Which lint requires this change?

0xPoe (commented):
Thanks a lot!

If you can separate your changes with the lint name by commit, it would be great.

src/cli/self_update.rs
chansuke: @hi-rustin Thank you so much for your review. I fixed [derive_partial_eq_without_eq](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq) lint error.

chansuke:
@hi-rustin I have updated the commit log with clippy name. Some of the messages are duplicated because `cargo clippy` doesn't show the error at once and I procedurally fixed the log:pray:

download/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 //! Easy file downloading
 #![deny(rust_2018_idioms)]
+#![allow(clippy::type_complexity)]
```
0xPoe: I prefer to warn it. We may find a better way to fix it in the future.

0xPoe (commented):
Could you please squash your first and third commits? I think they are from the same lint.
rest LGTM. Thanks again!

0xPoe:
Also, could you please rebase your base branch? It seems you still using the 2022 master branch.

download/src/lib.rs
0xPoe: @chansuke Could you please remove this? Then we can merge this PR.

chansuke:
Sorry for many time. I will fix this tonight(JST)

download/src/lib.rs
```diff
@@ -47,12 +47,13 @@ fn download_with_backend(
     }
 }
 
+type EventResult<'a> = &'a dyn Fn(Event<'_>) -> Result<()>;
```
0xPoe: We had some discussions on this change before: https://github.com/rust-lang/rustup/pull/3096#discussion_r1005434248
```suggestion
type DownloadCallback<'a> = &'a dyn Fn(Event<'_>) -> Result<()>;
```

0xPoe (approved):
Thanks again.  I'm sorry for this back-and-forth discussion. I should have said it all at once.
Because this is a nit change. I'll push it directly.

chansuke:
No problem. Thank you so much for your help.

PR #3152, opened 2023-01-11
Renovate tweaks

rbtcollins:
- enable lockfile maintenance to update Cargo.lock without Cargo
- don't create PRs until checks have completed for the change
- wait 3 days for woops! on new releases
- prefer lockfile updates over changes to Cargo.toml

The last one I think is not working correctly for Cargo - not harmful, just still more noise than we'd prefer - I'm working with renovate to track this down.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -5,5 +5,11 @@
     ],
     "labels": [
         "dependencies"
-    ]
+    ],
+    "lockFileMaintenance": {
+        "enabled": true
+    },
+    "prCreation": "not-pending",
+    "rangeStrategy": "replace",
+    "stabilityDays": 3
 }


.github/renovate.json
```diff
@@ -5,5 +5,11 @@
     ],
     "labels": [
         "dependencies"
-    ]
+    ],
+    "lockFileMaintenance": {
+        "enabled": true
+    },
+    "prCreation": "not-pending",
```
0xPoe, line 12: I'm just curious how the Renovate runs the tests before it creates the pull request. Do we need to enable GitHub Actions for its branches? Because we only run tests for master and pr.

rbtcollins: > > I'm not too sure, but it seems to work elsewhere with just pr actions

0xPoe (approved):
Thanks! Let's move on to see what's going on.

PR #3160, opened 2023-01-15
Test Renovate branches directly

rbtcollins:
- Permit tests to run on branches matching renovate/*, configured the
  same as PR tests. This permits Renovate to pre-test branches, and when
  we're comfortable will permit automatic merging if desired.
- Tell renovate where our actions templates are stored.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -11,5 +11,10 @@
     },
     "prCreation": "not-pending",
     "rangeStrategy": "replace",
-    "stabilityDays": 3
+    "stabilityDays": 3,
+    "github-actions": {
+        "fileMatch": [
+            "^ci\\/.*/[^/]+\\.ya?ml$"
+        ]
+    }
 }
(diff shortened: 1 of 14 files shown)


.github/renovate.json
```diff
@@ -11,5 +11,10 @@
     },
     "prCreation": "not-pending",
     "rangeStrategy": "replace",
-    "stabilityDays": 3
+    "stabilityDays": 3,
+    "github-actions": {
+        "fileMatch": [
+            "^ci\\/.*/[^/]+\\.ya?ml$"
```
0xPoe, line 17: Does this regex also match this file https://github.com/rust-lang/rustup/blob/master/ci/cirrus-templates/freebsd.yaml?

0xPoe (commented):
Thanks! Rest LGTM.

.github/renovate.json
rbtcollins: yes. Is that a problem?

0xPoe: This file is not for GitHub Actions. I don't know whether it makes sense to match it.

rbtcollins: If it has versions that look the same, best case is they are applicable, worst case is we ignore it later. e.g. I don't know if it makes sense but I see little harm in waiting and seeing

PR #3167, opened 2023-01-17
Improve msvc install instructions

ChrisDenton:
This is an attempt to address concerns with the msvc install instructions by:

* Highlighting the relevant area of images (note: I'm not an artist).
* Expanding on why Visual Studio Community edition is not right for everyone.


rbtcollins (approved):
thank you!

PR #3174, opened 2023-01-18
Do not allow `none` as custom toolchain name

### LINKED ISSUE #3130: rustup allows `rustup toolchain link none`
### Problem

This is not *necessarily* a bug, but it seems very confusing.

```
; rustup default none
info: default toolchain unset
; rustup toolchain link none build/aarch64-unknown-linux-gnu/stage0
; cargo
error: rustup could not choose a version of cargo to run, because one wasn't specified explicitly, and no default is configured.
; cargo +none --version
cargo 1.67.0-beta.3 (f6e737b1e 2022-12-02)
```

### Steps

1. Clone rust-lang/rust and run `./x.py --help
2. Run `rustup toolchain link none build/aarch64-unknown-linux-gnu/stage0` (as appropriate for your host triple)
3. Run `cargo +none --version`

### Possible Solution(s)

I think disallowing `none` for a custom toolchain name would be easier to understand, something like the existing error for reserved channel names:
```
; rustup toolchain link stable build/aarch64-unknown-linux-gnu/stage0
error: invalid custom toolchain name: 'stable-aarch64-unknown-linux-gnu'
```

### Notes

_No response_

### Rustup version

```console
```

### Installed toolchains

```console
> rustup show -v
rustup home:  /home/gh-jyn514/.local/lib/rustup

installed toolchains
--------------------

```
(this seems like a second bug: I definitely have at least nightly installed.)

ashutoshvarma:
Fix #3130 

Throws when `none` is used for custom toolchain name.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -465,7 +465,7 @@ pub struct CustomToolchain<'a>(&'a Toolchain<'a>);
 
 impl<'a> CustomToolchain<'a> {
     pub fn new(toolchain: &'a Toolchain<'a>) -> Result<CustomToolchain<'a>> {
-        if toolchain.is_custom() {
+        if toolchain.is_custom() && toolchain.name() != "none" {
             Ok(CustomToolchain(toolchain))
         } else {
             Err(anyhow!(format!(
--- a/tests/cli-rustup.rs
+++ b/tests/cli-rustup.rs
@@ -513,6 +513,12 @@ fn link() {
         expect_ok(config, &["rustup", "update", "nightly"]);
         expect_ok(config, &["rustup", "default", "nightly"]);
         expect_stdout_ok(config, &["rustup", "show"], "custom");
+        // should not allow 'none' for custom toolchain name
+        expect_err(
+            config,
+            &["rustup", "toolchain", "link", "none", &path],
+            "error: invalid custom toolchain name: 'none'",
+        );
     });
 }
 


src/toolchain.rs
```diff
@@ -465,7 +465,7 @@ pub struct CustomToolchain<'a>(&'a Toolchain<'a>);
 
 impl<'a> CustomToolchain<'a> {
     pub fn new(toolchain: &'a Toolchain<'a>) -> Result<CustomToolchain<'a>> {
-        if toolchain.is_custom() {
+        if toolchain.is_custom() && toolchain.name() != "none" {
```
0xPoe, line 468: @rbtcollins Do you think we need to move this special name into `is_custom_name` fn? Or we can just determine it here?

0xPoe (commented):
Thanks for working on this!

src/toolchain.rs
jyn514: I think it's *not* a custom name, right? the whole point is that we're not allowing it as a name for custom toolchains.

0xPoe: > I think it's _not_ a custom name, right?

Actually, in the current `is_custom_name` fn we treat it as a custom name. https://github.com/rust-lang/rustup/blob/060b8f9e3f5a8a810b49108aa57cee2e784ee104/src/toolchain.rs#L194

So it's more like a special case and probably we can put it into `is_custom_name`.

jyn514 (approved):
This seems fine to me; I think if you prevent creating a toolchain with this name that's enough and you don't need to also change other parts of rustup. In fact I would prefer not to for compatibility with people who happened to already have this name in an old version of rustup and then upgraded their install.

tests/cli-rustup.rs
```diff
@@ -513,6 +513,12 @@ fn link() {
         expect_ok(config, &["rustup", "update", "nightly"]);
         expect_ok(config, &["rustup", "default", "nightly"]);
         expect_stdout_ok(config, &["rustup", "show"], "custom");
+        // should not allow 'none' for custom toolchain name
```
rbtcollins, line 516: Could you please make this a separate test? Thanks!

src/toolchain.rs
rbtcollins: I agree with @hi-rustin that this is perhaps the wrong place.

A quick look at the code shows that we haven't integrated the "none" value of a toolchain systematically into the code. We shouldn't need to patch it here and there and everywhere.

We have things like `resolve_toolchain(&name)` that might be a decent place to flat out prohibit 'none' as a toolchain name, irrespective of custom-or-not

rbtcollins (changes requested):

rbtcollins:
I have a large refactoring of our toolchain name handling that will obsolete this.

rbtcollins:
See #3340

PR #3175, opened 2023-01-18
Lint the codebase with beta clippy

0xPoe:
Maybe we don't wanna land the lint suggestions before it is stable?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,10 @@
 1. Fork it!
 2. Create your feature branch: `git checkout -b my-new-feature`
 3. Test it: `cargo test`
-4. Lint it: `cargo +nightly clippy --all --all-targets -- -D warnings`
+4. Lint it: `cargo +beta clippy --all --all-targets -- -D warnings`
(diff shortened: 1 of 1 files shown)


rbtcollins:
Perhaps. It is nice when folk are active to see whats coming, but I certainly don't think having nightly lints prevent merges is  helpful. I think that that suggestion is there because we have an action that does clippy on nightly. Perhaps we removed that already.

rbtcollins:
https://github.com/rust-lang/rustup/blob/master/.github/workflows/centos-fmt-clippy-on-all.yaml#L68 still lints with beta. I'm not sure what the right path is here.

Perhaps add more explanation - e.g. we don't want to be surprised by incoming lints, but unstable lints don't always make it to stable?

Or perhaps instruct the use of beta?

0xPoe:
> Perhaps add more explanation - e.g. we don't want to be surprised by incoming lints, but unstable lints don't always make it to stable?
> 
> Or perhaps instruct the use of beta?

Use the beta version and also added some explanations. Could you please take another look?

rbtcollins (approved):

PR #3177, opened 2023-01-22
Bash completion install doc improvements

scop:
See individual commits for details.

### DIFF
--- a/doc/user-guide/src/installation/index.md
+++ b/doc/user-guide/src/installation/index.md
@@ -73,10 +73,10 @@ simple as using one of the following:
 
 ```console
 # Bash
-$ rustup completions bash > ~/.local/share/bash-completion/completions/rustup
+$ printf 'eval -- "$("$1" completions bash)"\n' > ~/.local/share/bash-completion/completions/rustup
 
 # Bash (macOS/Homebrew)
-$ rustup completions bash > $(brew --prefix)/etc/bash_completion.d/rustup.bash-completion
+$ printf 'eval -- "$(rustup completions bash)"\n' > $(brew --prefix)/etc/bash_completion.d/rustup.bash-completion
 
 # Fish
 $ mkdir -p ~/.config/fish/completions
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -169,10 +169,9 @@ pub(crate) static COMPLETIONS_HELP: &str = r"Discussion:
 
     Bash:
 
-    Completion files are commonly stored in
-    `/usr/share/bash-completion/completions` for
-    system-wide commands, but can be stored in
-    `~/.local/share/bash-completion/completions` for user-specific commands.
+    System wide completion files are commonly stored in
+    `/usr/share/bash-completion/completions`, and user-specific ones
+    can be stored in `~/.local/share/bash-completion/completions`.
     Run the command:
 
         $ mkdir -p ~/.local/share/bash-completion/completions
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -175,7 +175,7 @@ pub(crate) static COMPLETIONS_HELP: &str = r"Discussion:
     Run the command:
 
         $ mkdir -p ~/.local/share/bash-completion/completions
-        $ rustup completions bash >> ~/.local/share/bash-completion/completions/rustup
+        $ rustup completions bash > ~/.local/share/bash-completion/completions/rustup
 
     This installs the completion script. You may have to log out and
     log back in to your shell session for the changes to take effect.
@@ -270,7 +270,7 @@ pub(crate) static COMPLETIONS_HELP: &str = r"Discussion:
 
     Bash:
 
-        $ rustup completions bash cargo >> ~/.local/share/bash-completion/completions/cargo
+        $ rustup completions bash cargo > ~/.local/share/bash-completion/completions/cargo
 
     Zsh:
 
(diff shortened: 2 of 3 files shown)


rbtcollins:
I've never looked particularly closely at completions. What does bash do to avoid running that command on every completion request?

scop:
> What does bash do to avoid running that command on every completion request?

https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html has the background details. There's no anchor to link to on that page, but search for "There is some support for dynamically modifying completions" towards the end of it.

More specifically for this particular case, the https://github.com/scop/bash-completion project that I co-maintain is "in charge" of loading completions dynamically from the mentioned `.../bash-completion/completions` dirs (system and user wide), contains a completion loader that implements the above: https://github.com/scop/bash-completion/blob/fd7eadc541ef15fdb897911613d6939bcc1bb742/bash_completion#L2713-L2724

scop:
Rebased/resolved conflicts.

rami3l:
@scop In order to make the lights green again, you still have to update the UI test to match the new output. Please have a look at files in `tests/suite/cli-ui/` and change them accordingly (like what was done in e.g. https://github.com/rust-lang/rustup/pull/3542).

scop:
Updated, but no success yet: something seems to be changing backslash escapes to slashes in the outputs, don't know what should be done about that.

rami3l:
> Updated, but no success yet: something seems to be changing backslash escapes to slashes in the outputs, don't know what should be done about that.

@scop Sorry for the late reply!

I'll paste the original error below for the sake of convenience:

```diff
---- expected: stdout
++++ actual:   stdout
   1    1 | ...
   2    2 | Generate tab-completion scripts for your shell
   3    3 | 
   4    4 | Usage: rustup[EXE] completions [shell] [command]
   5    5 | 
          ⋮
  27   27 |     `/usr/share/bash-completion/completions`, and user-specific ones
  28   28 |     can be stored in `~/.local/share/bash-completion/completions`.
  29   29 |     Run the command:
  30   30 | 
  31   31 |         $ mkdir -p ~/.local/share/bash-completion/completions
  32      -         $ printf 'eval -- "$("$1" completions bash)"\n' \
       32 +         $ printf 'eval -- "$("$1" completions bash)"/n' /
  33   33 |               > ~/.local/share/bash-completion/completions/rustup
  34   34 | 
  35   35 |     This installs the completion script. You may have to log out and
  36   36 |     log back in to your shell session for the changes to take effect.
  37   37 | 
          ⋮
  39   39 | 
  40   40 |     Homebrew stores bash completion files within the Homebrew directory.
  41   41 |     With the `bash-completion` brew formula installed, run the command:
  42   42 | 
  43   43 |         $ mkdir -p $(brew --prefix)/etc/bash_completion.d
  44      -         $ printf 'eval -- "$(rustup completions bash)"\n' \
       44 +         $ printf 'eval -- "$(rustup completions bash)"/n' /
  45   45 |               > $(brew --prefix)/etc/bash_completion.d/rustup.bash-completion
  46   46 | 
  47   47 |     Fish:
  48   48 | 
  49   49 |     Fish completion files are commonly stored in
          ⋮
 125  125 |     toolchain. Not all shells are currently supported. Here are examples for
 126  126 |     the currently supported shells.
 127  127 | 
 128  128 |     Bash:
 129  129 | 
 130      -         $ printf 'eval -- "$(rustup completions bash cargo)"\n' \
      130 +         $ printf 'eval -- "$(rustup completions bash cargo)"/n' /
 131  131 |               > ~/.local/share/bash-completion/completions/cargo
 132  132 | 
 133  133 |     Zsh:
 134  134 | 
 135  135 |         $ rustup completions zsh cargo > ~/.zfunc/_cargo
```



~~IIRC this has something to do with [`after_help`](https://docs.rs/clap/latest/clap/struct.Command.html#method.after_help)'s markup support.~~
~~I'll look into it and hopefully find a solution real soon.~~

(cc @epage)

Update: As it turns out, the replacement of `\`s is made by `trycmd`, as explained in https://github.com/rust-lang/rustup/pull/3177#issuecomment-1896093126.

epage:
`trycmd` changes backslashes to forward slashes so that paths can be represented in a more neutral way across platforms.  This was modeled off of `cargo-test-support`.

rami3l:
@epage Thanks for your quick response!

This seems to be https://github.com/assert-rs/trycmd/issues/91 then.
@djc Before resolving that issue, maybe we should just accept the hack and use `/` in the snapshots?

djc:
> This seems to be [assert-rs/trycmd#91](https://github.com/assert-rs/trycmd/issues/91) then. @djc Before resolving that issue, maybe we should just accept the hack and use `/` in the snapshots?

Seems fine?

doc/user-guide/src/installation/index.md
```diff
@@ -73,10 +73,10 @@ simple as using one of the following:
 
 ```console
 # Bash
-$ rustup completions bash > ~/.local/share/bash-completion/completions/rustup
+$ printf 'eval -- "$("$1" completions bash)"\n' > ~/.local/share/bash-completion/completions/rustup
```
djc, line 76: What is the value of using `$1` here? It seems rather obfuscated to me.

scop: It ensures we're invoking the same `rustup` to generate completions as the one we are generating completions for. For example with a full path `/some/where/rustup <TAB>` that is not the `rustup` in `$PATH` -- we want to generate using `/some/where/rustup completions bash`, not `rustup completions bash`.

PR #3178, opened 2023-01-23
Don't add toolchain bin to PATH on Windows

### LINKED ISSUE #3036: rustup 1.25: On Windows, nested cargo invocation with a toolchain specified fails
### Problem

This is a regression in rustup 1.25 on Windows. If a program run via `cargo run` tries to run a command like `cargo +nightly ...`, it fails even if `RUSTC` is unset in the environment.

Splitting this off from https://github.com/rust-lang/rustup/issues/3031#issuecomment-1181867695 because https://github.com/rust-lang/rustup/issues/3031#issuecomment-1182253988 indicates it's a separate problem.

### Steps

https://github.com/nicholasbishop/rustup-nested-bug-repro#rustup-nested-bug-repro contains a repro of this bug.

Here's the `main.rs` of that repo:
```rust
use std::process::Command;
use std::env;

fn main() {
    // Modify the path if any argument is passed in.
    let modify_path = env::args().len() == 2;

    let mut cmd = Command::new("cargo");
    cmd.env_remove("RUSTC");
    cmd.env_remove("RUSTDOC");
    cmd.args(&["+nightly", "version"]);

    if modify_path {
        let orig_path = env::var_os("PATH").unwrap_or_default();
        let modified_split_paths = env::split_paths(&orig_path).filter(|path| {
            !path
                .components()
                .any(|component| component.as_os_str() == ".rustup")
        });
        let modified_path = env::join_paths(modified_split_paths).expect("invalid PATH");
        cmd.env("PATH", modified_path);
    }

    println!("running command: {:?}", cmd);
    let s = cmd.status().unwrap();
    assert!(s.success());
}
```

On Linux this command succeeds:

```
$ cargo run
   Compiling rustup-nested-bug-repro v0.1.0 (/var/home/nbishop/src/rustup-nested-bug-repro)
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/rustup-nested-bug-repro`
running command: "cargo" "+nightly" "version"
cargo 1.64.0-nightly (b1dd22e66 2022-07-09)
```

On Windows it fails:

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target\debug\rustup-nested-bug-repro.exe`
running command: "cargo" "+nightly" "version"
error: no such subcommand: `+nightly`
thread 'main' panicked at 'assertion failed: s.success()', src\main.rs:26:5
error: process didn't exit successfully: `target\debug\rustup-nested-bug-repro.exe` (exit code: 101)
```

But it succeeds if the `PATH` is modified to remove toolchain entries
added by rustup:

```
$ cargo run modify
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target\debug\rustup-nested-bug-repro.exe modify`
running command: "cargo" "+nightly" "version"
cargo 1.64.0-nightly (b1dd22e66 2022-07-09)
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
$ rustup --version
rustup 1.25.0 (90365aa81 2022-06-15)
info: The currently active `rustc` version is `rustc 1.62.0 (a8314ef7d 2022-06-27)`
```

### Installed toolchains

```console
$ rustup show
rustup home:  C:\Users\nicholasbishop\.rustup

installed toolchains
--------------------

active toolchain
----------------

rustc 1.62.0 (a8314ef7d 2022-06-27)
```

### LINKED ISSUE #3394: `cargo test` incompatible with latest rustup
The 0.7 release of rustup changed how `PATH` is managed, and now `cargo test` where `cargo` is actually a rustup proxy will no longer work for this project itself.

I believe this has to do with something related to binaries eventually being compiled against the standard library dynamically and then the standard library dll can't be found at runtime. This'll need to be fixed with `PATH` munging *somewhere* although I'm not quite sure where.

ehuss:
This removes the addition of the `<toolchain>/bin` directory to PATH on Windows. This fixes it so that recursive invocations of tools like `cargo` will use the rustup proxies instead of executing the original `<toolchain>/bin` executable, allowing those recursive invocations to do things like use `+toolchain` shorthands.

There is a bit of a long history for this behavior:
1. Originally, rustup would put `.cargo/bin` and `<toolchain>/bin` prepended to PATH.
2. #812 changed it to remove the `<toolchain>/bin` 
    * This was changed because as described in #809, when using a toolchain link, rustup will fall back to "nightly" if `cargo` is not in the toolchain. However, that would result in `cargo +stage1 ...` always using nightly.
3.  The `<toolchain>/bin` was added again in #1093, trying to fix rust-lang/cargo#3394. There, `cargo test` of the cargo project itself was not working. The reason isn't listed in the bug. I don't think it is an issue anymore, as cargo's testsuite now explicitly circumvents the rustup wrappers.
    * There were some other issues such as `cargo install clippy`. This would put a `cargo-clippy` binary in `~/.cargo/bin`. That binary needed to launch and load clippy-driver, which needed to load rustc-driver. Without PATH munging, this would fail. I don't think this is an issue anymore, since clippy is no longer distributed like that.
4. #2978 changed it so that the order of PATH was preserved. However, that introduced a problem now that `~/.cargo/bin` ended up behind the `<toolchain>/bin`, re-introducing the problem where recursive calls don't see the proxies. 

Fixes #3036

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -453,6 +453,25 @@ impl<'a> InstalledCommonToolchain<'a> {
             path_entries.push(cargo_home.join("bin"));
         }
 
+        if cfg!(target_os = "windows") {
+            // Historically rustup has included the bin directory in PATH to
+            // work around some bugs (see
+            // https://github.com/rust-lang/rustup/pull/3178 for more
+            // information). This shouldn't be needed anymore, and it causes
+            // problems because calling tools recursively (like `cargo
+            // +nightly metadata` from within a cargo subcommand). The
+            // recursive call won't work because it is not executing the
+            // proxy, so the `+` toolchain override doesn't work.
+            //
+            // This is opt-in to allow us to get more real-world testing.
+            if process()
+                .var_os("RUSTUP_WINDOWS_PATH_ADD_BIN")
+                .map_or(true, |s| s == "1")
+            {
+                path_entries.push(self.0.path.join("bin"));
+            }
+        }
+
         env_var::prepend_path("PATH", path_entries, cmd);
     }
 }
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -453,10 +453,6 @@ impl<'a> InstalledCommonToolchain<'a> {
             path_entries.push(cargo_home.join("bin"));
         }
 
-        if cfg!(target_os = "windows") {
-            path_entries.push(self.0.path.join("bin"));
-        }
-
         env_var::prepend_path("PATH", path_entries, cmd);
     }
 }
--- a/tests/mock/mock_bin_src.rs
+++ b/tests/mock/mock_bin_src.rs
@@ -61,6 +61,18 @@ fn main() {
             let rustc = &format!("rustc{}", EXE_SUFFIX);
             Command::new(rustc).arg("--version").status().unwrap();
         }
+        Some("--recursive-cargo-subcommand") => {
+            Command::new("cargo-foo")
+                .arg("--recursive-cargo")
+                .status()
+                .unwrap();
+        }
+        Some("--recursive-cargo") => {
+            Command::new("cargo")
+                .args(&["+nightly", "--version"])
+                .status()
+                .unwrap();
+        }
         Some("--echo-args") => {
             let mut out = io::stderr();
             for arg in args {
@@ -71,7 +83,7 @@ fn main() {
             let mut out = io::stderr();
             writeln!(out, "{}", std::env::var("PATH").unwrap()).unwrap();
         }
-        _ => panic!("bad mock proxy commandline"),
+        arg => panic!("bad mock proxy commandline: {:?}", arg),
     }
 }
 
--- a/tests/mock/mock_bin_src.rs
+++ b/tests/mock/mock_bin_src.rs
@@ -62,16 +62,18 @@ fn main() {
             Command::new(rustc).arg("--version").status().unwrap();
         }
         Some("--recursive-cargo-subcommand") => {
-            Command::new("cargo-foo")
+            let status = Command::new("cargo-foo")
                 .arg("--recursive-cargo")
                 .status()
                 .unwrap();
+            assert!(status.success());
         }
         Some("--recursive-cargo") => {
-            Command::new("cargo")
+            let status = Command::new("cargo")
                 .args(&["+nightly", "--version"])
                 .status()
                 .unwrap();
+            assert!(status.success());
         }
         Some("--echo-args") => {
             let mut out = io::stderr();
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -585,7 +585,21 @@ fn recursive_cargo() {
             fs::create_dir_all(&cargo_bin_path).unwrap();
             fs::copy(&real_mock_cargo, &cargo_subcommand).unwrap();
 
-            config.expect_stdout_ok(&["cargo", "--recursive-cargo-subcommand"], "hash-nightly-2");
+            // Verify the default behavior, which is currently broken on Windows.
+            let args = &["cargo", "--recursive-cargo-subcommand"];
+            if cfg!(windows) {
+                config.expect_err(
+                    &["cargo", "--recursive-cargo-subcommand"],
+                    "bad mock proxy commandline",
+                );
+            }
+
+            // Try the opt-in, which should fix it.
+            let out = config.run(args[0], &args[1..], &[("RUSTUP_WINDOWS_PATH_ADD_BIN", "0")]);
+            if !out.ok || !out.stdout.contains("hash-nightly-2") {
+                clitools::print_command(args, &out);
+                panic!("expected hash-nightly-2 in output");
+            }
         });
     });
 }
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -547,6 +547,49 @@ fn fallback_cargo_calls_correct_rustc() {
     });
 }
 
+// Checks that cargo can recursively invoke itself with rustup shorthand (via
+// the proxy).
+//
+// This involves a series of chained commands:
+//
+// 1. Calls `cargo --recursive-cargo-subcommand`
+// 2. The rustup `cargo` proxy launches, and launches the "mock" nightly cargo exe.
+// 3. The nightly "mock" cargo sees --recursive-cargo-subcommand, and launches
+//    `cargo-foo --recursive-cargo`
+// 4. `cargo-foo` sees `--recursive-cargo` and launches `cargo +nightly --version`
+// 5. The rustup `cargo` proxy launches, and launches the "mock" nightly cargo exe.
+// 6. The nightly "mock" cargo sees `--version` and prints the version.
+//
+// Previously, rustup would place the toolchain's `bin` directory in PATH for
+// Windows due to some DLL issues. However, those aren't necessary anymore.
+// If the toolchain `bin` directory is in PATH, then this test would fail in
+// step 5 because the `cargo` executable would be the "mock" nightly cargo,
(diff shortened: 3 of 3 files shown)


0xPoe (approved):
Looks good to me! Thanks!

0xPoe:
@rbtcollins Could you please take a look? Thanks!

rbtcollins:
[edited, I found the bits I was confused on]

maybe_do_cargo_fallback does the selection of toolchain. But it doesn't add +nightly etc - it directly picks the executable to use. The recursive call with `+nightly` in the command is not being done by rustup as far as I can tell. And https://github.com/rust-lang/rustup/issues/3031#issuecomment-1182253988 agrees with me on that.

Now some of the history also has to do with [Windows binary search order](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa). I think. 

we do this thing where we make a [temp directory for custom toolchains that are missing binaries](https://github.com/rust-lang/rustup/blob/d4e1f1708ac4044e5c1a7cd7c032741370221e17/src/config.rs#L952), and in [toolchain.rs](https://github.com/rust-lang/rustup/blob/d4e1f1708ac4044e5c1a7cd7c032741370221e17/src/toolchain.rs#L609)

In particular, the presence of the toolchain bin dir in PATH shouldn't affect whether cargo sees rustc from the same toolchain, or a rustc proxy, unless an explicitly overridden RUSTC_WRAPPER is in use, or cargo is calculating the specific binary to run itself. (is it?).

I'm really hesitant to permute this again without a functional test case that exercises the end to end behaviour we want (e.g. cargo -> build.rs -> cargo-but-must-be-rustup-proxy, rather than just the particular implementation we think is right today.

But then that raises a second question - is such an end to end behaviour reasonable? @jonhoo was working (from memory - could be the wrong person) on ensuring that once rustup calls into a command, nothing bounces out through the proxies again. This would make process invocation faster by removing unneeded CreateProcess, which is slow on Windows, but also avoid the manifest processing code in rustup for all platforms. That seems a bit incompatible with the end to end behaviour being discussed - can we get some convergence on our desires here? Then the code to support it should be easier to reason about.

For instance, if we could say: "Rustup figures out the right installed binary to run, and then sets variables {RUSTC, ...} pointing specifically at that binary, improving compile speed. Code that wants to recursively use *rustup*[1] should unset [list of variables] to disable this feature." then that might help both that performance work and this reentrant use case to work. That would also suggest that removing the toolchain dir would be sensible, because we'd be saying that our interface for finding binaries is 
1. PATH to find rustup proxies, then the proxy calculates the exact binary
2. overrides via environment variables like RUSTC, to let toolchain components avoid the overhead of (1) inside inner loops

What are your thoughts?

rbtcollins (changes requested):
The new test is failing on FreeBSD, I haven't looked into why, but  it doesn't look like a FreeBSD build flake.

ehuss:
Thanks for taking a look!

I understand this is a risky change. I have done a fair amount of local testing to try to shake out any possible concerns, but it's possible there are some cases that this might cause issues.

> The new test is failing on FreeBSD, I haven't looked into why, but it doesn't look like a FreeBSD build flake.

Oh, sorry, didn't catch that. I fixed it by changing a hard-link of the mock cargo to a copy. `current_exe()` was for some reason reporting the other hard-link file. I couldn't reproduce in a VM, but it seems to be fixed now.

> I'm really hesitant to permute this again without a functional test case that exercises the end to end behaviour we want (e.g. cargo -> build.rs -> cargo-but-must-be-rustup-proxy, rather than just the particular implementation we think is right today.

Maybe I am uncertain what this means, but I believe the `recursive_cargo` test is intended to exercise this exact scenario. It does cargo (proxy) -> cargo (mock) -> cargo-foo -> cargo (proxy) -> cargo (mock), where "cargo-foo" is simulating a build-script or third-party cargo command.

>  is such an end to end behaviour reasonable?

I can understand the hesitation here. For the scenario of improving the performance, that is related to #2958 (which was reverted in #3034), tracked in #3035. The solution I think for that is for rustup to set an environment variable to points to the toolchain bin directory (like `~/.rustup/toolchains/nightly-aarch64-apple-darwin/bin`). Cargo can then take that environment variable and use it to find rustc and rustdoc and execute them directly (instead of via the proxies in PATH). That should avoid the problems that were encountered with #2958.  I don't think setting `RUSTC` directly will work well, and I think this is an alternative that should be relatively easy.  I can perhaps look at implementing that if you are interested.

rbtcollins:
I will need to think more on this, but I very much appreciate the depth
you've gone to.

One possibility that occurs to me - could we make this feature flagged, so
that we have a release with it off, but set an option / env variable to
turn it on, and get more feedback?

ehuss:
> could we make this feature flagged, so that we have a release with it off, but set an option / env variable to turn it on, and get more feedback?

Sure, I went ahead and pushed an update the includes an opt-in. I'm a little concerned that it is very difficult to get people to test things like this, but it seems safer to at least have some kind of escape hatch.

ehuss:
@rbtcollins Did you have any more questions about this?

rbtcollins:
@ehuss How do you see the several related things tying together? e.g. if cargo assumes the toolchain path is predictable as we discussed, is this PR still relevant?

ehuss:
Yea, the optimization for cargo directly invoking the executables will help avoid the performance loss on Windows as a consequence of this change. This change is still necessary because that optimization will only apply to invoking `rustc` or `rustdoc` directly from within cargo. That optimization won't fix the issue here where there is:

1. User runs `cargo subcommand` which launches the `cargo` proxy. 
2. Proxy sets PATH to include `~/.rustup/toolchains/stable-x86_64-pc-windows-msvc/bin`. 
3. Proxy executes the real `cargo.exe`
4. `cargo` invokes `cargo-subcommand` from `~/.cargo/bin`
5. `cargo-subcommand` executes `rustc +nightly foo.rs`. This fails because the PATH modification in step 2 circumvented the proxies.

This PR's opt-in changes it to:

1. User runs `cargo subcommand` which launches the `cargo` proxy. 
2. Proxy executes the real `cargo.exe`
3. `cargo` invokes `cargo-subcommand` from `~/.cargo/bin`
4. `cargo-subcommand` executes `rustc +nightly foo.rs`, which launches the `rustc` proxy from `~/.cargo/bin`.
5. The proxy finds the correct `rustc` to run (nightly).

The direct execution optimization (which I will hopefully post soon) will help with the following (which is a different scenario):

Today's behavior (on Windows only):
1. User runs `cargo build` which launches the `cargo` proxy.
2. Proxy sets PATH to include `~/.rustup/toolchains/stable-x86_64-pc-windows-msvc/bin`. 
3. Proxy executes the real `cargo.exe`
4. Cargo executes `rustc` from PATH which is directly in `~/.rustup/toolchains/stable-x86_64-pc-windows-msvc/bin`.

With the opt-in from this PR, that would become (matching the behavior of non-Windows platforms):

1. User runs `cargo build` which launches the `cargo` proxy.
2. Proxy executes the real `cargo.exe`
3. Cargo executes `rustc` from PATH which is the `~/.cargo/bin` proxy.
4. The proxy launches the real `rustc` from `~/.rustup/toolchains/stable-x86_64-pc-windows-msvc/bin`

The optimization will bring it to the following on all platforms:

1. User runs `cargo build` which launches the `cargo` proxy.
2. Proxy executes the real `cargo.exe`
3. Cargo detects it is running under rustup and executes `rustc` directly from `~/.rustup/toolchains/stable-x86_64-pc-windows-msvc/bin`.

rbtcollins:
ok got it. So yes, lets merge this - we're mid release right now, but straight after the release I'll bring this in.

rbtcollins (approved):

rbtcollins:
Merging now since we have to re-stage after the GPG fixups.

PR #3186, opened 2023-02-01
Accept SHA1 keys and release 1.25.2

emilyalbini:
This PR prepares the 1.25.2 rustup release, adding SHA1 to the list of hashing algorithms explicitly allowed by Sequoia. This is needed since the Rust key uses SHA1 for its own signatures, and Sequoia has a 2023-02-01 cutoff for the use of SHA1. That is resulting in signature verification failed warnings for rustup users.


src/dist/signatures.rs
```diff
@@ -8,11 +8,25 @@ use anyhow::Result;
...
 use crate::config::PgpPublicKey;
 
+pub(crate) fn sequoia_policy() -> StandardPolicy<'static> {
+    let mut policy = StandardPolicy::new();
+
+    // The builtin Rust signature key uses SHA1 for its own signature, even though the individual
+    // signatures generated by it nowadays use SHA512. Sequoia added a cutoff date for SHA1,
+    // 2023-02-01, which caused warnings to be displayed to rustup users from that day onwards.
+    //
+    // To keep supporting the Rust signature key, we allow the SHA1 algorithm in rustup withou a
```
bjorn3: nit:
```suggestion
    // To keep supporting the Rust signature key, we allow the SHA1 algorithm in rustup without a
```

src/config.rs
```diff
@@ -198,7 +199,9 @@ impl PgpPublicKey {
         let keyid = format_hex(cert.keyid().as_bytes(), "-", 4)?;
         let algo = cert.primary_key().pk_algo();
         let fpr = format_hex(cert.fingerprint().as_bytes(), " ", 2)?;
-        let p = policy::StandardPolicy::new();
+        let mut p = sequoia_policy();
+        p.accept_hash(sequoia_openpgp::types::HashAlgorithm::SHA1);
```
nwalfield: `sequoia_policy` already calls `accept_hash(SHA1)`

emilyalbini: Good catch! That was a leftover from a previous attempt, will fix it soon.

src/dist/signatures.rs
```diff
@@ -8,11 +8,25 @@ use anyhow::Result;
...
+pub(crate) fn sequoia_policy() -> StandardPolicy<'static> {
+    let mut policy = StandardPolicy::new();
+
+    // The builtin Rust signature key uses SHA1 for its own signature, even though the individual
+    // signatures generated by it nowadays use SHA512. Sequoia added a cutoff date for SHA1,
+    // 2023-02-01, which caused warnings to be displayed to rustup users from that day onwards.
+    //
+    // To keep supporting the Rust signature key, we allow the SHA1 algorithm in rustup without a
+    // cutoff date.
+    policy.accept_hash(HashAlgorithm::SHA1);
```
nwalfield: Since this is only about the binding signature, you should probably only allow SHA-1 in contexts where collision resistance is not required instead of also for data signatures.  See [this](https://docs.sequoia-pgp.org/sequoia_openpgp/policy/struct.StandardPolicy.html#method.reject_hash_property_at) for more details.  Try something like: `policy.reject_hash_policy_at(HashAlgorithm::SHA1, HashAlgoSecurity::SecondPreImageResistance, std::time::UNIX_EPOCH)`

nwalfield (changes requested):

0xPoe:
@pietroalbini The CI failed because you missing some changes from https://github.com/rust-lang/rustup/pull/3121. You need to add the `-x`  arg to shellcheck.

emilyalbini:
Thanks @hi-rustin, cherry-picked that commit too.

src/dist/signatures.rs
emilyalbini: For now I'd prefer to use the current approach, as we do not know whether old release signatures still use SHA-1 or all release signatures always used SHA-512.

nwalfield: SHA-1 on data has been disabled since 2013.  So, this is creating attack surface relative to yesterday.

0xPoe (approved):
Looks good to me! But I'm not familiar with the release process.
Should we merge this change back to the master? I'm not sure. It would be nice if Daniel had time to look at it.

0xPoe:
The CI failed because of the code format.

kinnison (commented):
Modulo Neil's concern I'm okay with this.  @hi-rustin I trust you to confirm the CI tweaks are OK, as for release process, Pietro knows what to do, and once things are settled on the release, merging the changes back into master can happen.

Then we can consider how to resolve this longer term in a better fashion.

g2p:
Updating the embedded key might be a better option:

It should be possible (for someone with the secret key) to regenerate the internal signatures without updating the key id, using this sequoia tool: https://gitlab.com/sequoia-pgp/keyring-linter (https://lib.rs/crates/sequoia-keyring-linter)

src/dist/signatures.rs
emilyalbini: Hmm ok, sounds good, I updated to only allow `HashAlgoSecurity::SecondPreImageResistance` and confirmed the warning is not there with it.

emilyalbini:
Also disabled Clippy since 1.67 included a *loooooooot* of warnings.

kinnison:
> Updating the embedded key might be a better option:
> 
> It should be possible (for someone with the secret key) to regenerate the internal signatures without updating the key id, using this sequoia tool: https://gitlab.com/sequoia-pgp/keyring-linter

I think that this is a good idea, but probably might take longer to be confident about given the lack of experience with OpenPGP in the team right now.

As a plan I'd say "do this patch now, update key later, do new release with new key, and then make rollover plans properly" would make the most sense - we need to do the minimal change which gets our user-base to not need to worry about the warnings until we can protect them more fully - as it stands, we're still only warning, not erroring out, so the relative attack surface vs. protection aspects are semi-moot points anyway IMO.

0xPoe:
> Also disabled Clippy since 1.67 included a _loooooooot_ of warnings.

I suppose the warnings would not break the rustup's CI.

emilyalbini:
> I suppose the warnings would not break the rustup's CI.

Oh yeah, didn't notice rustup's CI doesn't deny warnings. Also pushed a super hacky way to get the Android CI hopefully working.

emilyalbini:
CI seems to be going, will merge this to let the build on `stable` finish sooner.

PR #3193, opened 2023-02-04
Bump the centos image version

0xPoe:
Bump the centos image and try to use the higher version GLIBC.
Otherwise, the CI will fail. See: https://github.com/rust-lang/rustup/actions/runs/4089637947/jobs/7052485676


rbtcollins:
looks like 7 failures. Have you looked into them?

Issue #3182, opened 2023-01-26
Support `.rust-toolchain.toml`


ojeda:
### Problem you are trying to solve

In the Linux kernel, we would have liked to take advantage of `rustup`'s `rust-toolchain.toml` feature to simplify our Quick Start instructions (see https://github.com/Rust-for-Linux/linux/issues/953) back when we needed to pin our compiler version.

However, it appears that there is no support for a hidden version of that configuration file, i.e. `.rust-toolchain.toml`, which would mean it is very visible when listing the root directory of the kernel sources.

However, most kernel developers did not use Rust, and even for those that did, the file is not really relevant: it would have very rarely changed and if they needed to change the toolchain version, they should have used a directory override in any case, rather than modifying the file.

Thus it is really not needed to show it all the time in listings, and so a hidden `.rust-toolchain.toml` file would be better. I guess for other projects that may be the case as well.

This would follow other files like `.clang-format`, `.gitignore`, `.gitattributes`, `.mailmap`, `.rustfmt.toml`, etc. where configuration is not directly related to the project (say, a `Kconfig` file) but to tooling.

### Solution you'd like

Support for the `.rust-toolchain.toml` filename.

rbtcollins:
Its not clear to me that rust-toolchain does what you want in that ticket.

You're not using a custom built toolchain, and you have developers using newer toolchains (and thus the expectation) that newer toolchains should work. Have you considered setting `rust-version` in your Cargo.toml to require a known minimum version and then lean on Rust's compatability guarantees, as an alternate approach?

`rust-toolchain{,.toml}` files can [cause arbitrary code execution](https://github.com/rust-lang/rustup/issues/2879), and IDEs are looking to [*disable* toolchain file support](https://github.com/rust-lang/rustup/issues/2793). In that context I don't think a hidden file is a good idea.

TheAlgorythm:
> Have you considered setting rust-version in your Cargo.toml to require a known minimum version and then lean on Rust's compatability guarantees, as an alternate approach?

I don't think, that this will work for the kernel, as it requires a specific version of rustc if I'm not mistaken.
Further `rust-toolchain.toml` can also specify `components` which isn't possible in `Cargo.toml`.

ojeda:
> You're not using a custom built toolchain, and you have developers using newer toolchains (and thus the expectation) that newer toolchains should work

No, sorry. The kernel uses a bunch of unstable Rust features, and thus we are pinned to a particular version so far.

[...]

PR #3195, opened 2023-02-05
add support for `.rust-toolchain.toml`

TheAlgorythm:
This PR adds support for `.rust-toolchain.toml` files like described in #3182.
I'm not sure if this is the normal procedure or is a RFC needed?

### DIFF
--- a/doc/src/overrides.md
+++ b/doc/src/overrides.md
@@ -77,10 +77,15 @@ case for nightly-only software that pins to a revision from the release
 archives.
 
 In these cases the toolchain can be named in the project's directory in a file
-called `.rust-toolchain.toml`, `rust-toolchain.toml` or `rust-toolchain`.
-If multiple files are present in a directory, the latter present one is used
-for backwards compatibility. The files use the [TOML] format and have the
-following layout:
+called `.rust-toolchain.toml`, `rust-toolchain.toml` or `rust-toolchain`.  If
+multiple files are present in a directory, the following precedence will be
+applied for backwards compatibility:
+
+1. `rust-toolchain`
+2. `rust-toolchain.toml`
+3. `.rust-toolchain.toml`
+
+The files use the [TOML] format and have the following layout:
 
 [TOML]: https://toml.io/
 
--- a/doc/src/overrides.md
+++ b/doc/src/overrides.md
@@ -77,9 +77,10 @@ case for nightly-only software that pins to a revision from the release
 archives.
 
 In these cases the toolchain can be named in the project's directory in a file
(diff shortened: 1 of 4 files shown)


bjorn3:
cargo-crev will need to be updated too if this PR ends up getting accepted: https://github.com/crev-dev/cargo-crev/blob/5d9574d43634013ef73684f2af35f37db8ee8bfe/crev-lib/src/util/mod.rs#L118-L122

ojeda:
`doc/src/overrides.md` here should be updated too.

doc/src/overrides.md
```diff
@@ -77,9 +77,10 @@ case for nightly-only software that pins to a revision from the release
 archives.
 
 In these cases the toolchain can be named in the project's directory in a file
-called `rust-toolchain.toml` or `rust-toolchain`. If both files are present in
-a directory, the latter is used for backwards compatibility. The files use the
-[TOML] format and have the following layout:
+called `.rust-toolchain.toml`, `rust-toolchain.toml` or `rust-toolchain`.
+If multiple files are present in a directory, the latter present one is used
```
ojeda: "the latter in the list" could clarify since there are more than 2 now.

ojeda (commented):
Also: "The legacy format is not available in `rust-toolchain.toml` files." (in between the other two doc changes) could be updated to mention both files explicitly.

doc/src/overrides.md
TheAlgorythm: Maybe it would be better to use an enumeration for the precedence of the files.

TheAlgorythm: Did the enumeration. Sorry for the bad phrasing in the first documentation attempt.

ojeda: No worries at all! Thanks for working on this. :)

rbtcollins:
So at this point both @kinnison and I dislike hiding the file. I'm not sure we can provide a sensible rational that can be argued through though. For now, since this would potentially affect all users and would not be a reversible decision, I think we should not merge a patch adding it.

ojeda:
> So at this point both @kinnison and I dislike hiding the file. I'm not sure we can provide a sensible rational that can be argued through though.

Well, we provided a rational argument in the original issue, so perhaps you can find something there in particular that you do not like? Otherwise, it is hard to reply.

> For now, since this would potentially affect all users and would not be a reversible decision, I think we should not merge a patch adding it.

I agree a fair amount of people would probably start using it (which is one reason it would be nice to have! :), but it is backwards compatible, so I am not sure I understand how it would affect all users -- it should not affect any existing one, no?

PR #3205, opened 2023-02-11
Add test for cli ui

0xPoe:
ref https://github.com/rust-lang/rustup/pull/3064#issuecomment-1414789331

Add ui tests for cli. Then we can try to bump clap's version.

Some notes:
1. Because of [this bug](https://github.com/assert-rs/trycmd/issues/197), so we need to use `...` to ignore the first line of the subcommand bin name.
2. Because we don't have the `man` command on windows, we need to skip the man's test on windows.

### DIFF
--- a/tests/cli-ui.rs
+++ b/tests/cli-ui.rs
@@ -1,0 +1,13 @@
+use std::fs;
+
+#[test]
+fn ui_tests() {
+    let t = trycmd::TestCases::new();
+    let rustup_init = trycmd::cargo::cargo_bin("rustup-init");
+    let rustup = trycmd::cargo::cargo_bin("rustup");
+    t.register_bin("rustup-init", &rustup_init);
+    // Copy rustup-init to rustup so that the tests can run it.
+    fs::copy(&rustup_init, &rustup).unwrap();
+    t.register_bin("rustup", &rustup);
+    t.case("tests/cli-ui/*.toml");
+}
--- a/tests/cli-ui.rs
+++ b/tests/cli-ui.rs
@@ -10,4 +10,9 @@ fn ui_tests() {
     fs::copy(&rustup_init, &rustup).unwrap();
     t.register_bin("rustup", &rustup);
     t.case("tests/cli-ui/*.toml");
+    #[cfg(target_os = "windows")]
+    {
+        // On windows, we don't have man command, so skip the tests.
+        t.skip("tests/cli-ui/rustup_man_cmd_help_flag_stdout.toml");
+    }
 }
--- a/tests/cli-ui/rustup-init_help_flag_stdout.toml
+++ b/tests/cli-ui/rustup-init_help_flag_stdout.toml
@@ -1,0 +1,27 @@
+bin.name = "rustup-init"
+args = ["--help"]
+status.code = 0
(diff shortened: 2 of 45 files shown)


src/utils/utils.rs
```diff
@@ -341,7 +341,7 @@ where
     })
 }
 
-pub(crate) fn copy_file(src: &Path, dest: &Path) -> Result<()> {
+pub fn copy_file(src: &Path, dest: &Path) -> Result<()> {
```
rbtcollins: This is presumably being used from some test code; I'd like to keep the truely public functions as minimal as we can - just `fs::copy` is probably more than sufficient for the test suite case

tests/cli-ui.rs
```diff
@@ -0,0 +1,18 @@
+use rustup::utils::utils;
+
+#[test]
+fn ui_tests() {
+    let t = trycmd::TestCases::new();
+    let rustup_init = trycmd::cargo::cargo_bin("rustup-init");
+    let rustup = trycmd::cargo::cargo_bin("rustup");
+    t.register_bin("rustup-init", &rustup_init);
+    // Copy rustup-init to rustup so that the tests can run it.
+    utils::copy_file(&rustup_init, &rustup).unwrap();
```
rbtcollins: ah yes here; this can just use `fs::copy`, it doesn't need all the heuristics and UI glue that we have in utils.

tests/cli-ui/rustup-init_help_flag_stdout.toml
```diff
@@ -0,0 +1,27 @@
...
+status.code = 0
+stdout = """
+rustup-init [..]
+The installer for rustup
+
+USAGE:
+    rustup-init[EXE] [FLAGS] [OPTIONS]
+
+FLAGS:
+    -v, --verbose                        Enable verbose output
```
rbtcollins, line 12: I'm not sure if this will be stable as clap evolves - we may want to do things like

`--verbose[...] Enable verbose output`  : but we can do that if we need to, not right now ;)

src/utils/utils.rs
0xPoe: Revered! Thanks!

tests/cli-ui/rustup-init_help_flag_stdout.toml
0xPoe: Yes. we can update it if we need to.

tests/cli-ui.rs
```diff
@@ -0,0 +1,18 @@
+use std::fs;
+
+#[test]
+fn ui_tests() {
+    let t = trycmd::TestCases::new();
+    let rustup_init = trycmd::cargo::cargo_bin("rustup-init");
+    let rustup = trycmd::cargo::cargo_bin("rustup");
+    t.register_bin("rustup-init", &rustup_init);
+    // Copy rustup-init to rustup so that the tests can run it.
+    fs::copy(&rustup_init, &rustup).unwrap();
```
rbtcollins, line 10: You could also export RUSTUP_FORCE_ARG0='rust', but the copy is fine.

rbtcollins (approved):

Issue #3031, opened 2022-07-12
rustup 1.25 sets RUSTC, which overrides custom toolchain specified in sub-Cargo invocation


mykmelez:
### Problem

I have a Rust build wrapper (xtask) that invokes Cargo with a custom toolchain `+esp` and a toolchain-specific target `--target xtensa-esp32s3-espidf`, so I can use a simple command like `cargo xtask build` to cross-compile my project with a more complex command like `cargo +esp --target xtensa-esp32s3-espidf`.

This works with rustup 1.24, but rustup 1.25 defines RUSTC in the environment of the build wrapper, setting it to the default compiler, which causes the sub-Cargo invocation in the build wrapper to ignore the custom toolchain specified in the `+esp` argument and instead invoke the rustc compiler specified in RUSTC.

And since that compiler doesn't support the toolchain-specific target, the invocation fails:

>error: failed to run `rustc` to learn about target-specific information
>
>Caused by:
>process didn't exit successfully: `/Users/myk/.rustup/toolchains/1.59.0-aarch64-apple-darwin/bin/rustc - --crate-name ___ --print=file-names --target xtensa-esp32s3-espidf --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 1)
>--- stderr
>error: Error loading target specification: Could not find specification for target "xtensa-esp32s3-espidf". Run `rustc --print target-list` for a list of built-in targets

### Steps

I don't yet have a set of steps to reproduce, but I'll work on a minimal test case.

### Possible Solution(s)

It isn't clear why RUSTC now gets defined in the environment, but if there isn't a good reason for it, then reverting that change would obviously fix the bug. (I'm working around the issue at the moment by undefining RUSTC in the build wrapper before it invokes Cargo.) If there is a good reason for it, then I'm not sure what the solution would be.

### Notes

I wonder if #3025 and/or #3029 are related.

### Rustup version

```console
Version that works:

>\> rustup --version
>rustup 1.24.0 (2f894d923 2021-04-24)

Version that is broken:

>\> rustup --version
>rustup 1.25.0 (90365aa81 2022-06-15)

In both cases:

>info: This is the version for the rustup toolchain manager, not the rustc compiler.
>info: The currently active `rustc` version is `rustc 1.59.0 (9d1b2106e 2022-02-23)`
```

### Installed toolchains

```console
> rustup show
[...]

PR #3207, opened 2023-02-12
Add RUSTUP_TOOLCHAIN_DIR

### LINKED ISSUE #3035: un-revert "Set RUSTC and RUSTDOC env for child processes run through the proxy"

We had some nice performance enhancements but hadn't caught an edge case well enough (which has led to a reversion). Lets support that and then re-enable the code

### Solution you'd like

Robust automatic detection of different toolchain selection in recursive rustup proxy use. Nothing special needed by callers, except passing through variables

### Notes

- #2958 
- #3034

ehuss:
This adds an environment variable `RUSTUP_TOOLCHAIN_DIR` that is set when running a toolchain command. This is intended to be used by Cargo to execute rustc and rustdoc directly without going through the rustup wrappers. This should help improve performance on non-Windows platforms, and to avoid the performance loss on Windows in #3178.

Modifying PATH isn't an option because it prevents recursive commands from being able to run via the proxies (#812).
Setting RUSTC isn't an option because it breaks recursive invocations (#3029 and #3031).
This option *should* be safe, since cargo will only invoke the toolchain binary if no other setting is specified. Recursive invocations should continue to be able to run the rustup wrappers.

https://github.com/ehuss/cargo/commit/f6d949b80fb385d331c73ff82a43a1eb2e04a764 is a prototype of what the cargo side looks like.

I have done performance testing on a variety of machines running macOS and Linux across 10 different projects. Typical improvement for a clean cargo check is about 1.07 to 1.16 times faster (with a peak of 1.25).

Closes #3035

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -401,6 +401,7 @@ impl<'a> InstalledCommonToolchain<'a> {
 
         cmd.env("RUSTUP_TOOLCHAIN", &self.0.name);
         cmd.env("RUSTUP_HOME", &self.0.cfg.rustup_dir);
+        cmd.env("RUSTUP_TOOLCHAIN_DIR", &self.0.path);
     }
 
     fn set_ldpath(&self, cmd: &mut Command) {
--- a/tests/cli-rustup.rs
+++ b/tests/cli-rustup.rs
@@ -4,7 +4,7 @@ pub mod mock;
 
 use std::env::consts::EXE_SUFFIX;
 use std::fs;
-use std::path::{PathBuf, MAIN_SEPARATOR};
+use std::path::{Path, PathBuf, MAIN_SEPARATOR};
 
 use rustup::for_host;
 use rustup::test::this_host_triple;
@@ -2249,6 +2249,22 @@ fn warn_on_duplicate_rust_toolchain_file() {
                 toolchain_file_1.canonicalize().unwrap().display(),
                 toolchain_file_2.canonicalize().unwrap().display(),
             ),
+        );
+    });
+}
+
+/// Checks that the RUSTUP_TOOLCHAIN_DIR is set.
+#[test]
+fn toolchain_dir_env() {
+    setup(&|config| {
+        expect_ok(config, &["rustup", "default", "nightly"]);
+        let output = clitools::run(config, "rustup", &["which", "rustc"], &[]);
+        let real_mock_rustc = Path::new(output.stdout.trim());
+        let toolchain_dir = real_mock_rustc.parent().unwrap().parent().unwrap();
+        expect_stderr_ok(
+            config,
+            &["rustc", "--echo-env", "RUSTUP_TOOLCHAIN_DIR"],
+            toolchain_dir.to_str().unwrap(),
         );
     });
 }
--- a/tests/mock/mock_bin_src.rs
+++ b/tests/mock/mock_bin_src.rs
@@ -71,6 +71,11 @@ fn main() {
             let mut out = io::stderr();
             writeln!(out, "{}", std::env::var("PATH").unwrap()).unwrap();
         }
+        Some("--echo-env") => {
+            let mut out = io::stderr();
+            let var = args.next().unwrap();
+            writeln!(out, "{}", std::env::var(var).unwrap_or_default()).unwrap();
+        }
         _ => panic!("bad mock proxy commandline"),
     }
 }


rbtcollins:
This looks appealing to me. Thank you.

So to be sure I understand:
- we keep the proxies on the path so code such as RUSTC_WRAPPERS, and build.rs scripts can use +toolchain syntax without fear
- but we want the cargo->rust interface to be fast, so we're going to set a variable for cargo to use to find rust directly 

I think cargo will need to know to reset that variable : consider this stack

cargo(proxy) (sets var)
cargo
some-build-rs
explicit-call-to-a-different-toolchain-cargo (uses var)
wrong-rustc-invoked

I think we need to keep this var hermetic: vars are great for tunnelling but know no boundaries :/.

What do you think?

ehuss:
>  explicit-call-to-a-different-toolchain-cargo (uses var)
> wrong-rustc-invoked

Can you say more about what this step looks like? If it is something like executing `cargo +nightly build`, then that should execute the proxy which should reset all of the environment variables (RUSTUP_TOOLCHAIN, RUSTUP_TOOLCHAIN_DIR, etc.).  Or if the script overrides the toolchain with the `RUSTUP_TOOLCHAIN` environment variable, then that should also work (and reset `RUSTUP_TOOLCHAIN_DIR` to point to the new toolchain).

rbtcollins:
> > explicit-call-to-a-different-toolchain-cargo (uses var)
> > wrong-rustc-invoked
> 
> Can you say more about what this step looks like? If it is something like executing `cargo +nightly build`, then that should execute the proxy which should reset all of the environment variables (RUSTUP_TOOLCHAIN, RUSTUP_TOOLCHAIN_DIR, etc.).

Lets say someone has a custom toolchain they aren't managing with rustup. They do `/path/to/that/toolchains/cargo do-something-here`.

ehuss:
I'm still not clear exactly what would happen differently in that circumstance.

Before:
cargo (proxy) -> Sets RUSTUP_TOOLCHAIN -> cargo (real) -> build-script -> calls `/other/cargo` -> other cargo calls `rustc` (proxy) -> rustc proxy sees `RUSTUP_TOOLCHAIN` and executes the toolchain from the original cargo

New:
cargo (proxy) -> Sets RUSTUP_TOOLCHAIN and RUSTUP_TOOLCHAIN_DIR -> cargo (real) -> build-script -> calls /other/cargo` -> other cargo calls `rustc` directly from RUSTUP_TOOLCHAIN_DIR

In both cases, the `rustc` should be the one from the original toolchain determined from the first cargo.

A build script or extension could intentionally modify `PATH` to have a different toolchain earlier in PATH. I could see that causing it to use a different `rustc`. FWIW, I think the scenario of a build script or extension trying to intentionally circumvent rustup in that fashion seems fairly unlikely (and lurks in the evil "you shouldn't do that" category).

rbtcollins:
> I'm still not clear exactly what would happen differently in that circumstance.
> 
> Before: cargo (proxy) -> Sets RUSTUP_TOOLCHAIN -> cargo (real) -> build-script -> calls `/other/cargo` -> other cargo calls `rustc` (proxy) -> rustc proxy sees `RUSTUP_TOOLCHAIN` and executes the toolchain from the original cargo

Ah, I'd forgotten we set that too; ok so thats clear.

> New: cargo (proxy) -> Sets RUSTUP_TOOLCHAIN and RUSTUP_TOOLCHAIN_DIR -> cargo (real) -> build-script -> calls /other/cargo`-> other cargo calls`rustc` directly from RUSTUP_TOOLCHAIN_DIR

yep. What if someone sets RUSTUP_TOOLCHAIN in there somewhere?

> In both cases, the `rustc` should be the one from the original toolchain determined from the first cargo.
> 
> A build script or extension could intentionally modify `PATH` to have a different toolchain earlier in PATH. I could see that causing it to use a different `rustc`. FWIW, I think the scenario of a build script or extension trying to intentionally circumvent rustup in that fashion seems fairly unlikely (and lurks in the evil "you shouldn't do that" category).

I have largely given up assuming good intentions from users :). I don't mean that in a negative way, but rather lets be conservative rather than surprised.

rbtcollins:
I have a proposal for a variation on this. Two in fact.

1) Rust looks at the version of the toolchain and if new enough sets RUSTUP_TOOLCHAIN to `{"toolchain":toolchain, "toolchain_dir": thedir}`

2) we set RUSTUP_TOOLCHAIN_DATA=`{"toolchain":toolchain, "toolchain_dir": thedir}` and cargo checks to see if the toolchain value is different to RUSTUP_TOOLCHAIN before using any other fields.

JSON is an arbitrary choice; we could use protobuf or whatever you like.

ehuss:
> yep. What if someone sets RUSTUP_TOOLCHAIN in there somewhere?

Yea, I think in that case it would also be different.

---

I'd be concerned about changing the format of `RUSTUP_TOOLCHAIN`, since there are likely things inspecting it. [Code search](https://cs.github.com/?scopeName=All+repos&scope=&q=language%3Arust+%5C%22RUSTUP_TOOLCHAIN%5C%22) shows a fair number of hits, which would likely be a breaking change for them.

Another option I was reluctant to suggest is that Cargo could *unset* `RUSTUP_TOOLCHAIN_DIR` when calling other processes. That would ensure it isn't "sticky", and only `RUSTUP_TOOLCHAIN` would drive things. That should address the concern about an inner-non-rustup cargo calling something with `RUSTUP_TOOLCHAIN` changed or modifying `PATH`. That should make sure the behavior doesn't change (at least for things driven with cargo).

rbtcollins:
So given these three options:
- add ..._DIR
- add ..._DIR and unset it in cargo (which is what I originally asked for in https://github.com/rust-lang/rustup/pull/3207#issuecomment-1427151207)
- add ..._DATA with a built-in check code [the value of RUST_TOOLCHAIN when it was set]

What do you prefer and why?

ehuss:
Of the latter two, I don't have a strong preference. Using an encoded form is more flexible, but a little more complicated. I'm happy to do either, do you have a preference?

rbtcollins:
I am terrible at predicting the future ;).

If we're going to want to pass more info to cargo over time, having an extensible format is nice, because we can treat this as a bit of an API.

If we really just need this one thing to fix things, then _DIR + resetting it is clearly less work.

My hunch is that we'll find other things in future, but the question is whether they will be coupled like this is, or distinct and thus safer to simply add a new variable for

rbtcollins:
And I found a problem with the idea of Cargo programmatically converting RUSTUP_TOOLCHAIN -> the dir. We now support toolchains in arbitrary locations on disk in rustup-toolchain.toml.

Whether that was a mistake or not (I think it is) its out there now, and we'd have to be graceful in pulling it back. 

OTOH an arbitrary absolute file path is now a valid toolchain name in RUSTUP_TOOLCHAIN. So cargo could look for that - but that will run into some trouble with respect to trusted_dirs. You should trust any arbitrary file path IFF rustup trusts it, but we have no method to signal that today, since most any process can invoke cargo with an env variable set :/.

ehuss:
Thanks for the heads up. I think if someone does that, then the optimization just won't work, which I'm fine with for now.

rbtcollins:
I think we can probably close this at this point?

ehuss:
Yea, it's fine to close. We can revisit if needed in the future. Thanks for the review and discussion!

PR #3209, opened 2023-02-12
Add ui test for the rustup-init.sh

0xPoe:
Add UI test for the `rustup-init.sh`. Also compare two help texts to make sure they are equal.


0xPoe:
This PR is still working in the process. I hope I can fix it tomorrow.

src/cli/setup_mode.rs
```diff
@@ -22,7 +22,7 @@ pub fn main() -> Result<utils::ExitCode> {
         return Ok(utils::ExitCode(0));
     }
 
-    // XXX: If you change anything here, please make the same changes in rustup-init.sh
```
rbtcollins, line 25: The XXX's were to grab developers attentions, not a 'thing to fix' or 'TODO'; I'm not sure we need them, but I'm not sure we don't.

0xPoe: Using `NOTICE` now.

0xPoe: Thanks for your review!  💚 💙 💜 💛 ❤️

rustup-init.sh
```diff
@@ -22,28 +22,28 @@ set -u
...
+    -y                                   Disable confirmation prompt.
+        --no-update-default-toolchain    Don't update any existing default toolchain after install
+        --no-modify-path                 Don't configure the PATH environment variable
+    -h, --help                           Prints help information
+    -V, --version                        Prints version information
 
 OPTIONS:
         --default-host <default-host>              Choose a default host triple
         --default-toolchain <default-toolchain>    Choose a default toolchain to install
-        --default-toolchain none                   Do not install any toolchains
```
rbtcollins, line 45: So this is interesting. I think we lose something by not explicitly saying 'use none to get no toolchain installed'

rbtcollins: Is there any reason - other than having the same help as the executable - for this part of the change?

0xPoe: > Is there any reason - other than having the same help as the executable - for this part of the change?

Nope. The only reason is to want to make sure they are the same.

rbtcollins: I'd rather we either:
- keep them different
- teach the binary rustup-init how to offer that tip

we added that because folk were not realising that this was how it worked.

0xPoe: Addressed. I also think it's rarely used by users.

rbtcollins (approved):

0xPoe:
Thanks for your review again!  💚 💙 💜 💛 ❤️

PR #3210, opened 2023-02-12
doc: add links for the aarch64-apple-darwin files

vrischmann:
The links on the [Other installation methods](https://rust-lang.github.io/rustup/installation/other.html) page don't mention the `aarch64-apple-darwin` builds, this PR adds them.


rbtcollins (approved):
I wonder if we shouldn't autogenerate that table

PR #3214, opened 2023-02-14
Work improving check_updates_none test performance

rbtcollins:
- Move tests of aliases out of the test - they are unrelated
- rename the trycmd test for better precision
- clarify test command output
- measure the time test commands take to run


0xPoe (approved):
Looks good to me. Thanks!

0xPoe:
It seems you need to format your code.

PR #3216, opened 2023-02-15
clippy: Apply clippy suggestions

illia-bobyr:
`CONTRIBUTING.md` suggests one runs this, before publishing a change:
```bash
cargo clippy --all --all-targets --all-features -- -D warnings
```
These changes make `cargo clippy` output on `master` warning free.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -88,7 +88,7 @@ pub fn this_host_triple() -> String {
         // For windows, this host may be different to the target: we may be
         // building with i686 toolchain, but on an x86_64 host, so run the
         // actual detection logic and trust it.
-        let tp = Box::new(currentprocess::TestProcess::default());
+        let tp = Box::<currentprocess::TestProcess>::default();
         return currentprocess::with(tp, || TargetTriple::from_host().unwrap().to_string());
     }
     let arch = if cfg!(target_arch = "x86") {
--- a/tests/dist.rs
+++ b/tests/dist.rs
@@ -32,11 +32,10 @@ use crate::mock::{MockComponentBuilder, MockFile, MockInstallerBuilder};
 
 const SHA256_HASH_LEN: usize = 64;
 
+pub type EditChannel = dyn Fn(/* date */ &str, &mut MockChannel);
+
 // Creates a mock dist server populated with some test data
-pub fn create_mock_dist_server(
-    path: &Path,
-    edit: Option<&dyn Fn(&str, &mut MockChannel)>,
-) -> MockDistServer {
+pub fn create_mock_dist_server(path: &Path, edit: Option<&EditChannel>) -> MockDistServer {
     MockDistServer {
         path: path.to_owned(),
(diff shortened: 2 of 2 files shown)


src/test.rs
```diff
@@ -88,7 +88,7 @@ pub fn this_host_triple() -> String {
         // For windows, this host may be different to the target: we may be
         // building with i686 toolchain, but on an x86_64 host, so run the
         // actual detection logic and trust it.
-        let tp = Box::new(currentprocess::TestProcess::default());
+        let tp = Box::<currentprocess::TestProcess>::default();
```
0xPoe, line 91: Please see: https://github.com/rust-lang/rustup/pull/3096#discussion_r1005435968

0xPoe (commented):
Thanks for your PR.

We have some discussion about the `Box::default()` change before. Maybe we can revert it.

src/test.rs
illia-bobyr: Definitely can, but it is what the stable clippy is suggesting today.
During that discussion, it was still a nightly only check.
Maybe, instead of just removing the patch, I should change it to

```rust
#[allow(clippy::box_default)]
```

For this particular location.  With a link to that discussion?
Otherwise, it may come again, as clippy will keep complaining :)

illia-bobyr: I totally agree with the consistency argument.
But, if this is what clippy is suggesting, I wonder if most new code will be written like this now.

rbtcollins: I would like to see a blanket allow on this for rustup.

tests/dist.rs
```diff
@@ -32,11 +32,10 @@ use crate::mock::{MockComponentBuilder, MockFile, MockInstallerBuilder};
...
 const SHA256_HASH_LEN: usize = 64;
 
+pub type EditChannel = dyn Fn(/* date */ &str, &mut MockChannel);
+
 // Creates a mock dist server populated with some test data
-pub fn create_mock_dist_server(
-    path: &Path,
-    edit: Option<&dyn Fn(&str, &mut MockChannel)>,
-) -> MockDistServer {
+pub fn create_mock_dist_server(path: &Path, edit: Option<&EditChannel>) -> MockDistServer {
```
rbtcollins, line 38: So EditChannel is just defining the function signature that these two functions use as a closure.

We're separating out two things that are tightly coupled. I don't think that increases readability: it increases cognitive load.

Please allow() it at the module level. (I have a PR in flight that refactors this test code towards immutability, so the whole edit concept is going to fade anyhow).

0xPoe:
Already done at https://github.com/rust-lang/rustup/pull/3276/files. So I am going to close this PR.

Thanks for your contribution!

0xPoe:
@gautamprikshit1 I would like to suggest asking the original author before picking up the work from their PRs next time. This way people don't race on the same thing and the original author may feel better. ❤️ Thank you all.

PR #3219, opened 2023-02-16
constant dist servers in tests

rbtcollins:
Introduces a concept of constant dist servers in the test suite, and uses it to improve performance of cli-exact and cli-rustup tests.


0xPoe (approved):
Thanks a lot!

PR #3222, opened 2023-02-18
Move test calls from fn(&config) to self.fn 

rbtcollins:
This will permit some future refactorings, but also is somewhat more idiomatic given that the functions aren't generic


rbtcollins:
will fix win tests tomorrow

PR #3224, opened 2023-02-19
Do all builds together in CI

rbtcollins:
This may help CI performance by reducing thrashing.

### DIFF
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -30,13 +30,15 @@ if [ "$TARGET" = arm-linux-androideabi ]; then
   export CFLAGS='-march=armv7'
 fi
 
-cargo build --locked --release --target "$TARGET" "${FEATURES[@]}"
-
-runtest () {
-  cargo test --locked --release --target "$TARGET" "${FEATURES[@]}" "$@"
+target_cargo() {
+    cmd="$1"
+    shift
+    cargo "${cmd}" --locked --release --target "$TARGET" "${FEATURES[@]}" "$@"
 }
 
-run_download_pkg_test() {
+target_cargo build
+
+download_pkg_test() {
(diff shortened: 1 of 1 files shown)


rbtcollins:
Comparing times to https://github.com/rust-lang/rustup/pull/3223
(this vs base)
linux x86_64 38m v 34m
windows 42m v 46m
mac x64 51m v 49m
windows x86-gnu 52m v 60m
windows i686 52m v 55m

I suspect cross tenancy noise means these are variable anyway, but the key thing is that the slowest builds have gotten faster, so this should faster overall if that holds up; but if its just noise, we're no worse off.

We could also look at building-without-testing the unit tests on other platforms.

rbtcollins:
Comparing times to https://github.com/rust-lang/rustup/pull/3223
(this vs base)
linux x86_64 38m v 34m
windows 42m v 46m
mac x64 51m v 49m
windows x86-gnu 52m v 60m
windows i686 52m v 55m

I suspect cross tenancy noise means these are variable anyway, but the key thing is that the slowest builds have gotten faster, so this should faster overall if that holds up; but if its just noise, we're no worse off.

We could also look at building-without-testing the unit tests on other platforms.

rbtcollins:
Comparing times to https://github.com/rust-lang/rustup/pull/3223
(this vs base)
linux x86_64 38m v 34m
windows 42m v 46m
mac x64 51m v 49m
windows x86-gnu 52m v 60m
windows i686 52m v 55m

I suspect cross tenancy noise means these are variable anyway, but the key thing is that the slowest builds have gotten faster, so this should faster overall if that holds up; but if its just noise, we're no worse off.

We could also look at building-without-testing the unit tests on other platforms.

0xPoe (approved):

0xPoe:
Thanks!

PR #3225, opened 2023-02-19
Make `rustup show` output info in a more logical order

### LINKED ISSUE #1397: `rustup show` should not force-install the default toolchain if it is not installed.

majaha:
I ran `rustup show` and noticed that its output is in the order:
installed toolchains -> installed targets for active toolchain -> active toolchain
I thought it would make sense to put "active toolchain" before " installed targets for active toolchain" so that they in order of decreasing scope:
installed toolchains -> active toolchain -> installed targets for active toolchain
 so that is what this patch does 🙂
Thoughts?

*Before:*
```
Matt@Matts-PC /c/d/p/r/rustup (show-order)> rustup show
rustup home:  C:\Users\Matt\.rustup

installed toolchains
--------------------

wasm32-unknown-unknown
x86_64-pc-windows-msvc

active toolchain
----------------

rustc 1.67.1 (d5a82bbd2 2023-02-07)

Matt@Matts-PC /c/d/p/r/rustup (show-order)>
```
*After:*
```
Matt@Matts-PC /c/d/p/r/rustup (show-order)> RUSTUP_HOME=home CARGO_HOME=home home/bin/rustup show
rustup home:  D:\programming\repos\rustup\home

installed toolchains
--------------------

active toolchain
----------------

rustc 1.67.1 (d5a82bbd2 2023-02-07)

wasm32-unknown-unknown
x86_64-pc-windows-msvc

Matt@Matts-PC /c/d/p/r/rustup (show-order)>
```

Fixes https://github.com/rust-lang/rustup/issues/1397

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -14,6 +14,7 @@ use once_cell::sync::Lazy;
 
 use super::self_update;
 use crate::cli::download_tracker::DownloadTracker;
+use crate::config::ActiveReason;
 use crate::currentprocess::{
     argsource::ArgSource,
     filesource::{StdinSource, StdoutSource},
@@ -466,11 +467,15 @@ pub(crate) fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCod
     } else {
         let def_toolchain_name = cfg.get_default()?.map(|t| (&t).into());
         let cwd = utils::current_dir()?;
-        let ovr_toolchain_name = if let Ok(Some((toolchain, _reason))) = cfg.find_override(&cwd) {
-            Some(toolchain)
-        } else {
-            None
-        };
+        let ovr_toolchain_name =
+            if let Ok(Some((toolchain, reason))) = cfg.find_active_toolchain(&cwd) {
+                match reason {
+                    ActiveReason::Default => None,
+                    _ => Some(toolchain),
+                }
+            } else {
+                None
+            };
         for toolchain in toolchains {
             let if_default = if def_toolchain_name.as_ref() == Some(&toolchain) {
                 " (default)"
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -14,7 +14,6 @@ use once_cell::sync::Lazy;
 
 use super::self_update;
 use crate::cli::download_tracker::DownloadTracker;
-use crate::config::ActiveReason;
 use crate::currentprocess::{
     argsource::ArgSource,
     filesource::{StdinSource, StdoutSource},
@@ -24,6 +23,7 @@ use crate::currentprocess::{
 use crate::dist::dist::{TargetTriple, ToolchainDesc};
 use crate::dist::manifest::ComponentStatus;
 use crate::install::UpdateStatus;
+use crate::toolchain::names::{LocalToolchainName, ToolchainName};
 use crate::utils::notifications as util_notifications;
 use crate::utils::notify::NotificationLevel;
 use crate::utils::utils;
@@ -426,78 +426,72 @@ fn list_items(
     Ok(utils::ExitCode(0))
 }
 
-fn print_toolchain_path(
-    cfg: &Cfg,
-    toolchain: &str,
-    if_default: &str,
-    if_override: &str,
-    verbose: bool,
-) -> Result<()> {
-    let toolchain_path = cfg.toolchains_dir.join(toolchain);
-    let toolchain_meta = fs::symlink_metadata(&toolchain_path)?;
-    let toolchain_path = if verbose {
-        if toolchain_meta.is_dir() {
-            format!("\t{}", toolchain_path.display())
-        } else {
-            format!("\t{}", fs::read_link(toolchain_path)?.display())
-        }
-    } else {
-        String::new()
-    };
-    writeln!(
-        process().stdout().lock(),
-        "{}{}{}{}",
-        &toolchain,
-        if_default,
-        if_override,
-        toolchain_path
-    )?;
-    Ok(())
-}
-
 pub(crate) fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCode> {
-    // Work with LocalToolchainName to accommodate path based overrides
-    let toolchains = cfg
-        .list_toolchains()?
-        .iter()
-        .map(Into::into)
-        .collect::<Vec<_>>();
+    let toolchains = cfg.list_toolchains()?;
     if toolchains.is_empty() {
         writeln!(process().stdout().lock(), "no installed toolchains")?;
     } else {
-        let def_toolchain_name = cfg.get_default()?.map(|t| (&t).into());
+        let default_toolchain_name = cfg.get_default()?;
         let cwd = utils::current_dir()?;
-        let ovr_toolchain_name =
-            if let Ok(Some((toolchain, reason))) = cfg.find_active_toolchain(&cwd) {
-                match reason {
-                    ActiveReason::Default => None,
-                    _ => Some(toolchain),
-                }
+        let active_toolchain_name: Option<ToolchainName> =
+            if let Ok(Some((LocalToolchainName::Named(toolchain), _reason))) =
+                cfg.find_active_toolchain(&cwd)
+            {
+                Some(toolchain)
             } else {
                 None
             };
-        for toolchain in toolchains {
-            let if_default = if def_toolchain_name.as_ref() == Some(&toolchain) {
-                " (default)"
-            } else {
-                ""
-            };
-            let if_override = if ovr_toolchain_name.as_ref() == Some(&toolchain) {
-                " (override)"
-            } else {
-                ""
-            };
 
-            print_toolchain_path(
+        for toolchain in toolchains {
+            let is_default_toolchain = default_toolchain_name.as_ref() == Some(&toolchain);
+            let is_active_toolchain = active_toolchain_name.as_ref() == Some(&toolchain);
+
+            print_toolchain(
                 cfg,
                 &toolchain.to_string(),
-                if_default,
-                if_override,
+                is_default_toolchain,
+                is_active_toolchain,
                 verbose,
             )
             .context("Failed to list toolchains' directories")?;
         }
     }
+
+    fn print_toolchain(
+        cfg: &Cfg,
+        toolchain: &str,
+        is_default: bool,
+        is_active: bool,
+        verbose: bool,
+    ) -> Result<()> {
+        let toolchain_path = cfg.toolchains_dir.join(toolchain);
+        let toolchain_meta = fs::symlink_metadata(&toolchain_path)?;
+        let toolchain_path = if verbose {
+            if toolchain_meta.is_dir() {
+                format!(" {}", toolchain_path.display())
+            } else {
+                format!(" {}", fs::read_link(toolchain_path)?.display())
+            }
+        } else {
+            String::new()
+        };
+        let status_str = match (is_default, is_active) {
+            (true, true) => " (active, default)",
+            (true, false) => " (default)",
+            (false, true) => " (active)",
+            (false, false) => "",
+        };
+
+        writeln!(
+            process().stdout().lock(),
+            "{}{}{}",
+            &toolchain,
+            status_str,
+            toolchain_path
+        )?;
+        Ok(())
+    }
+
     Ok(utils::ExitCode(0))
 }
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -20,6 +20,7 @@ use crate::{
         topical_doc,
     },
     command,
+    config::ActiveReason,
     currentprocess::{
         argsource::ArgSource,
         filesource::{StderrSource, StdoutSource},
@@ -864,8 +865,10 @@ fn default_(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
         };
 
         let cwd = utils::current_dir()?;
-        if let Some((toolchain, reason)) = cfg.find_override(&cwd)? {
-            info!("note that the toolchain '{toolchain}' is currently in use ({reason})");
+        if let Some((toolchain, reason)) = cfg.find_active_toolchain(&cwd)? {
+            if !matches!(reason, ActiveReason::Default) {
+                info!("note that the toolchain '{toolchain}' is currently in use ({reason})");
+            }
         }
     } else {
         let default_toolchain = cfg
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -872,10 +872,15 @@ fn default_(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
             }
         }
     } else {
-        let default_toolchain = cfg
-            .get_default()?
-            .ok_or_else(|| anyhow!("no default toolchain configured"))?;
-        writeln!(process().stdout().lock(), "{default_toolchain} (default)")?;
+        match cfg.get_default()? {
+            Some(default_toolchain) => {
+                writeln!(process().stdout().lock(), "{default_toolchain} (default)")?;
+            }
+            None => writeln!(
+                process().stdout().lock(),
+                "no default toolchain is configured"
+            )?,
+        }
     }
 
     Ok(utils::ExitCode(0))
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -110,7 +110,7 @@ pub fn main() -> Result<utils::ExitCode> {
                     cfg.set_toolchain_override(&ResolvableToolchainName::try_from(&t[1..])?);
                 }
 
-                let toolchain = cfg.find_or_install_override_toolchain_or_default(&cwd)?.0;
+                let toolchain = cfg.find_or_install_active_toolchain(&cwd)?.0;
 
                 Ok(toolchain.rustc_version())
             }
@@ -1082,7 +1082,7 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let cwd = utils::current_dir()?;
     let installed_toolchains = cfg.list_toolchains()?;
     // XXX: we may want a find_without_install capability for show.
-    let active_toolchain = cfg.find_or_install_override_toolchain_or_default(&cwd);
+    let active_toolchain = cfg.find_or_install_active_toolchain(&cwd);
 
     // active_toolchain will carry the reason we don't have one in its detail.
     let active_targets = if let Ok(ref at) = active_toolchain {
@@ -1175,16 +1175,10 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
         }
 
         match active_toolchain {
-            Ok(atc) => match atc {
-                (ref toolchain, Some(ref reason)) => {
-                    writeln!(t.lock(), "{} ({})", toolchain.name(), reason)?;
-                    writeln!(t.lock(), "{}", toolchain.rustc_version())?;
-                }
-                (ref toolchain, None) => {
-                    writeln!(t.lock(), "{} (default)", toolchain.name())?;
-                    writeln!(t.lock(), "{}", toolchain.rustc_version())?;
-                }
-            },
+            Ok((ref toolchain, ref reason)) => {
+                writeln!(t.lock(), "{} ({})", toolchain.name(), reason)?;
+                writeln!(t.lock(), "{}", toolchain.rustc_version())?;
+            }
             Err(err) => {
                 let root_cause = err.root_cause();
                 if let Some(RustupError::ToolchainNotSelected) =
@@ -1223,7 +1217,7 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
 fn show_active_toolchain(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let verbose = m.get_flag("verbose");
     let cwd = utils::current_dir()?;
-    match cfg.find_or_install_override_toolchain_or_default(&cwd) {
+    match cfg.find_or_install_active_toolchain(&cwd) {
         Err(e) => {
             let root_cause = e.root_cause();
             if let Some(RustupError::ToolchainNotSelected) =
@@ -1234,16 +1228,12 @@ fn show_active_toolchain(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
             }
         }
         Ok((toolchain, reason)) => {
-            if let Some(reason) = reason {
-                writeln!(
-                    process().stdout().lock(),
-                    "{} ({})",
-                    toolchain.name(),
-                    reason
-                )?;
-            } else {
-                writeln!(process().stdout().lock(), "{} (default)", toolchain.name())?;
-            }
+            writeln!(
+                process().stdout().lock(),
+                "{} ({})",
+                toolchain.name(),
+                reason
+            )?;
             if verbose {
                 writeln!(process().stdout().lock(), "{}", toolchain.rustc_version())?;
             }
@@ -1408,7 +1398,7 @@ fn explicit_or_dir_toolchain2(
         }
         None => {
             let cwd = utils::current_dir()?;
-            let (toolchain, _) = cfg.find_or_install_override_toolchain_or_default(&cwd)?;
+            let (toolchain, _) = cfg.find_or_install_active_toolchain(&cwd)?;
 
             Ok(toolchain)
         }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -20,7 +20,7 @@ use crate::{
         topical_doc,
     },
     command,
-    config::ActiveReason,
+    config::{new_toolchain_with_reason, ActiveReason},
     currentprocess::{
         argsource::ArgSource,
         filesource::{StderrSource, StdoutSource},
@@ -38,8 +38,9 @@ use crate::{
         names::{
             custom_toolchain_name_parser, maybe_resolvable_toolchainame_parser,
             partial_toolchain_desc_parser, resolvable_local_toolchainame_parser,
-            resolvable_toolchainame_parser, CustomToolchainName, MaybeResolvableToolchainName,
-            ResolvableLocalToolchainName, ResolvableToolchainName, ToolchainName,
+            resolvable_toolchainame_parser, CustomToolchainName, LocalToolchainName,
+            MaybeResolvableToolchainName, ResolvableLocalToolchainName, ResolvableToolchainName,
+            ToolchainName,
         },
         toolchain::Toolchain,
     },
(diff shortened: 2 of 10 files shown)


rbtcollins:
My gut reaction is that folk usually want the active toolchain information, so having it first or last is best; and last means they don't need to scan far for it or use head.

Neither first nor last will match your concept though.

Thanks for demonstrating the idea clearly though!

majaha:
Thanks for considering it!

I've fixed the tests I broke, amended the commit and commit message and force pushed. Apologies to anyone who pulled the original commit.

majaha:
There's always `rustup show active-toolchain [-v]`, although it is admittedly longer
```fish
Matt@Matts-PC /c/d/p/r/rustup (show-order)> rustup show active-toolchain
Matt@Matts-PC /c/d/p/r/rustup (show-order)> rustup show active-toolchain -v
rustc 1.67.1 (d5a82bbd2 2023-02-07)
```

majaha:
As an aside, would it be possible to reformat the reason the active toolchain was chosen? At the moment the floating `(default)` is a little misleading and mysterious if you don't already know what it means. And it visually matches how it looks in "installed toolchains" even though it means something different.  (My initial reaction to it was "Why is there a default active toolchain if there can only be one active toolchain? 🤔" but of course, that's not what it means 🙂).
```
active toolchain
----------------
rustc 1.67.1 (d5a82bbd2 2023-02-07)
```
->
```
active toolchain
----------------
stable-x86_64-pc-windows-msvc
rustc 1.67.1 (d5a82bbd2 2023-02-07)
Reason: is the default toolchain
/or/
Reason: overridden by +toolchain on the command line
```

Also, there's some scope for code deduplication between `show_active_toolchain()` and `show()`

workingjubilee:
I feel like "\nReason:" is nondescript (reason for what? it's dissociated from the phrase "active toolchain" by multiple lines by that point) and the " (inline comment)" style is better, but I don't think " (default)" is a good explanation either.

majaha:
It's a little terse, I'll admit. I think because it's right next to the other parts of the active toolchain, you can work out what it's the reason for using context from the reason message. But I would be open to:
```
active toolchain
----------------
stable-x86_64-pc-windows-msvc
rustc 1.67.1 (d5a82bbd2 2023-02-07)
Reason it's active: it's the default toolchain
```

Either way, it's certainly more descriptive than the other lines! Perhaps it should really be:
```
active toolchain
----------------
Toolchain name: stable-x86_64-pc-windows-msvc
Compiler version: rustc 1.67.1 (d5a82bbd2 2023-02-07)
Reason it's active: it's the default toolchain
```

workingjubilee:
Hm. Why not reframe it and instead
```
active toolchain: default
----------------
stable-x86_64-pc-windows-msvc
rustc 1.67.1 (d5a82bbd2 2023-02-07)
```

rbtcollins:
What would happen if we put the reason on the same line as Active toolchain.

`Active toolchain (the default)`
`Active toolchain (selected by RUSTUP_TOOLCHAIN)`
...

(or something similar).

Having thought more about this I do like the idea of making targets subordinate to the toolchain. Perhaps like this:

```
RUSTUP_HOME=home CARGO_HOME=home home/bin/rustup show
rustup home:  D:\programming\repos\rustup\home

installed toolchains
===================

active toolchain
================

channel: stable-x86_64-pc-windows-msvc
version: rustc 1.67.1 (d5a82bbd2 2023-02-07)
reason: default
targets: 
--------

wasm32-unknown-unknown
x86_64-pc-windows-msvc
```

majaha:
Yes, that's definitely an improvement!

Although I think we could do away with the blank lines after the `=====` and `-----` lines.

djc:
@majaha would you be able to rebase this and follow up on the discussion so far?

majaha:
Okay, I've rebased the old patch.

We're probably not ready to merge yet though, as we were still discussing and iterating on what we wanted to do. I'm going to have another think about what we want here.

majaha:
Okay, I've spent some more time thinking about the output format, incorporating some other's ideas, and this is what I've come up with:

```
$ rustup show
rustup home:  C:\Users\Matt\.rustup

installed toolchains
====================

active toolchain
================
compiler: rustc 1.73.0 (cc66ad468 2023-10-03)
active because: overridden by +toolchain on the command line
targets:
--------
wasm32-unknown-unknown
x86_64-pc-windows-msvc
```

I thought about putting the active reason in the header like this:
```
active toolchain [by default]
================
```
but decided against it as the `rust-toolchain` file override reason can be arbitrarily long, and will probably wrap for some people, causing some headache with what to prevent the wrapped `====` line being ugly:
```
active toolchain [overridden by 'D:\foo\bar\baz\...\rust-toolchain']
================
```

We should also change the format of `rustup show active-toolchain` and split it into two lines:
```
$ RUSTUP_TOOLCHAIN=nightly rustup show active-toolchain
nightly-x86_64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)
```
⬇
```
$ RUSTUP_TOOLCHAIN=nightly rustup show active-toolchain
active because: environment override by RUSTUP_TOOLCHAIN
```

rami3l:
> ```
> $ RUSTUP_TOOLCHAIN=nightly rustup show active-toolchain
> nightly-x86_64-pc-windows-msvc
> active because: environment override by RUSTUP_TOOLCHAIN
> ```

Indeed, breaking that into two lines might be a good idea. @majaha, it looks like we are moving in the right direction!

> ```
> active toolchain [overridden by 'D:\foo\bar\baz\...\rust-toolchain']
> ================
> ```

I still prefer using `active because:` here.

However, I'd like to add an `(active)` in the first section, which is a deliberate duplication but will hopefully make things clearer:

```
$ rustup show
rustup home:  C:\Users\Matt\.rustup

installed toolchains
====================
stable-x86_64-pc-windows-msvc (active)

active toolchain
================
compiler: rustc 1.73.0 (cc66ad468 2023-10-03)
active because: overridden by +toolchain on the command line
targets:
--------
wasm32-unknown-unknown
x86_64-pc-windows-msvc
```

... and we might want to use `(active, default)` for the non-overridden state.

rami3l:
@rbtcollins @djc Instead of changing to `=` and `-` combined as proposed above, what if we keep the `-`, and add indentation to the mix (IIRC it should be 2 spaces to be coherent)?

```sh
<snip>

active toolchain
----------------
compiler: rustc 1.73.0 (cc66ad468 2023-10-03)
active because: overridden by +toolchain on the command line
targets:
  wasm32-unknown-unknown
  x86_64-pc-windows-msvc
```

Other than this I have no outstanding concerns at the moment.

majaha:
> I still prefer using `active because:` here.

Sorry, do you mean that you want it to be on the same line as the header next to `active toolchain`, or that you want it to be on it's own line underneath?

rami3l:
> > I still prefer using `active because:` here.
> 
> Sorry, do you mean that you want it to be on the same line as the header next to `active toolchain`, or that you want it to be on it's own line underneath?

A separate line underneath, just like in my snippets above.

majaha:
Ah, gotcha. Yeah, I think you're version looks quite clean!

djc:
This looks good to me. So can we get the code updated to match the proposed style?

majaha:
Okay, I'm working on this.

Here's a thought I had: Should we always have the `installed toolchains` and `active toolchain -> installed targets` lists, even if they have only a single entry? I think that would be clearer and more intuitive. If you haven't seen them before, it's not obvious that their non-presence is giving you information. It's too implicit, you have to go and carefully read the help page to pick up on it.

rami3l:
@majaha As a quick reminder, with the currently proposed style, the output should be like:

```console
$ RUSTUP_TOOLCHAIN=nightly rustup show active-toolchain
active because: environment override by RUSTUP_TOOLCHAIN
```

_(From https://github.com/rust-lang/rustup/pull/3225#issuecomment-1782058807)_

---

... with `(active)` and `(default)` in `installed toolchains`:

```console
$ rustup show
rustup home:  C:\Users\Matt\.rustup

installed toolchains
--------------------
stable-x86_64-pc-windows-msvc (active)

active toolchain
----------------
compiler: rustc 1.73.0 (cc66ad468 2023-10-03)
active because: overridden by +toolchain on the command line
targets:
  wasm32-unknown-unknown
  x86_64-pc-windows-msvc
```

_(From https://github.com/rust-lang/rustup/pull/3225#issuecomment-1793316750)_

---

Also, `(active, default)` should be used if the default toolchain is active.

_(From https://github.com/rust-lang/rustup/pull/3225#issuecomment-1793314607)_

majaha:
I've uploaded the new patch, which includes the changes and a bit of a refactoring:
1. Change the output format of `rustup show`, to be in a more logical
   order and have more pleasing formatting
2. Update the formatting of `rustup show active-toolchain` and
   `rustup toolchain list` to match the new `rustup show` format.
3. `rustup +nightly show` will no longer install the nightly toolchain
   if it isn't already installed. Ditto for
   `rustup show active-toolchain`. Fixes https://github.com/rust-lang/rustup/issues/1397
4. Fix a bug where the RUSTUP_TOOLCHAIN environment variable took
   priority over the +toolchain command line option, and add a test for
   override priority.
5. `rustup default` no longer errors when there is no default toolchain
   configured, it prints a message instead.
6. Update the docs, and fix an issue where they incorrectly stated
   that `rust-toolchain.toml` config files couldn't specify custom
   names. (Text last touched here, see commit message: https://github.com/rust-lang/rustup/commit/7bfbd3b96c29898de001cbda7730fa440d35bd3b)

djc:
@majaha this sounds great, except I'd like to request basically one commit per bullet point, which should make it a lot easier to review this. Do you think that would be feasible?

majaha:
Ok, I've split out as best as I can. The first one is still quite large, but I can't really split it much further without meticulously reimplementing the bugs that were fixed. The commit message should help though; start by reading through the changes to `OverrideCfg`, and noting that `OverrideReason` has turned into `ActiveReason`, and the rest mostly follows from there.

src/config.rs
```diff
@@ -115,59 +118,157 @@ impl Display for OverrideReason {
...
-                    }
-                    // We -do not- support relative paths, they permit trivial
-                    // completely arbitrary code execution in a directory.
-                    // Longer term we'll not support path based toolchains at
-                    // all, because they also permit arbitrary code execution,
-                    // though with more challenges to exploit.
-                    Some((&PathBasedToolchainName::try_from(&path as &Path)?).into())
-                }
-                (Some(channel), Some(path)) => {
+        let toolchain_name: ToolchainName = match (file.toolchain.channel, file.toolchain.path) {
```
djc: Here, I'd like a separate commit for the refactoring that splits the `match` for `toolchain_name` out of the `Self` initializer.

src/config.rs
```diff
@@ -96,17 +96,20 @@ impl<T: Into<String>> From<T> for OverrideFile {
     }
 }
 
+// Represents the reason why the active toolchain is active.
 #[derive(Debug)]
-pub(crate) enum OverrideReason {
+pub(crate) enum ActiveReason {
```
djc, line 100: This rename should be in a separate commit.

src/cli/common.rs
```diff
@@ -491,22 +492,24 @@ fn print_toolchain_path(
...
-        let ovr_toolchain_name = if let Ok(Some((toolchain, _reason))) = cfg.find_override(&cwd) {
-            Some(toolchain)
-        } else {
-            None
-        };
+        let ovr_toolchain_name: Option<ToolchainName> =
+            if let Ok(Some((LocalToolchainName::Named(toolchain), reason))) =
+                cfg.find_active_toolchain(&cwd)
+            {
+                if !matches!(reason, crate::config::ActiveReason::Default) {
```
djc: Style nit: let's import `ActiveReason` here, and use a `match` expression` instead of `matches!()`. Also please avoid type annotations (here and elsewhere) -- I think this one is probably unnecessary.

src/cli/common.rs
```diff
@@ -491,22 +492,24 @@ fn print_toolchain_path(
 }
 
 pub(crate) fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCode> {
-    // Work with LocalToolchainName to accommodate path based overrides
-    let toolchains = cfg
-        .list_toolchains()?
-        .iter()
-        .map(Into::into)
-        .collect::<Vec<_>>();
+    let toolchains: Vec<ToolchainName> = cfg.list_toolchains()?;
```
djc: This doesn't need a type annotation, please remove it.

src/config.rs
```diff
@@ -115,59 +118,157 @@ impl Display for OverrideReason {
...
-    targets: Vec<String>,
-    profile: Option<dist::Profile>,
+/// Calls Toolchain::new(), but augments the error message with more context
+/// from the ActiveReason if the toolchain isn't installed.
+pub(crate) fn new_toolchain_with_reason<'a>(
+    cfg: &'a Cfg,
+    name: LocalToolchainName,
+    reason: &ActiveReason,
+) -> Result<Toolchain<'a>> {
+    match Toolchain::new(cfg, name.clone()) {
```
djc, line 127: Please rewrite this to reduce rightward drift:

```rust
match Toolchain::new(cfg, name.clone()) {
     Err(RustupError::ToolchainNotInstalled(_)) => {},
     result => return Ok(result?),
}

..
```

src/config.rs
```diff
@@ -443,105 +544,62 @@ impl Cfg {
             .transpose()?)
     }
 
-    pub(crate) fn find_override(
+    pub(crate) fn find_active_toolchain(
```
djc, line 547: I think this rename should probably in a commit by itself?

djc (commented):
I'm sorry, but I do think the first commit should be split up -- as is, this is very hard to review. I reviewed that commit only so far, suggesting some ways to make it smaller. Overall the direction looks great!

src/config.rs
rami3l: @djc I'm not so sure about that... Although it could possibly make the PR easier to review, the intermediate state that you're looking for might have inconsistent semantics (that is, a "new" type/fn name with an "old" implementation which cannot match that name). IMHO it's like intentionally introducing bugs into the codebase 🤔

src/config.rs
rami3l: @djc I'm not so sure about that... Although it could possibly make the PR easier to review, the intermediate state that you're looking for might have inconsistent semantics (that is, a "new" type/fn name with an "old" implementation which cannot match that name). IMHO it's like intentionally introducing bugs into the codebase 🤔

src/config.rs
rami3l: I second this.

majaha:
Looks like a random test failure. Can we run it again?

src/config.rs
djc: I generally don't think this is a big issue? Like, it's only within the context of a few commits (that are part of the same PR), after which the branch is in a consistent state again.

majaha: My instinct is that one semantic change should be in one commit, the code should compile and pass tests at each commit, and that artificially small commits aren't helpful. That's what I'm used to from my experience working on different projects. I don't really see any value in having the branch in an inconsistent state to begin with.

src/config.rs
```diff
@@ -96,78 +96,179 @@ impl<T: Into<String>> From<T> for OverrideFile {
...
+                        "toolchain options are ignored for a custom toolchain ({})",
+                        name
+                    )
+                }
+                Self::Custom(name)
+            }
         })
     }
+
+    fn get_local_toolchain_name(self) -> LocalToolchainName {
```
rami3l: According to our [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html), `get_*` should probably be replaced with `into_`.

src/config.rs
```diff
@@ -443,106 +544,72 @@ impl Cfg {
...
-        // Then walk up the directory tree from 'path' looking for either the
-        // directory in override database, or a `rust-toolchain` file.
-        if override_.is_none() {
-            self.settings_file.with(|s| {
-                override_ = self.find_override_from_dir_walk(path, s)?;
-
-                Ok(())
-            })?;
+    ) -> Result<Option<(LocalToolchainName, ActiveReason)>> {
+        if let Some((override_config, reason)) = self.find_override_config(path)? {
```
rami3l: How about using `Ok(if let ...)` here instead of `if let ... { Ok(...) } else { Ok(...) }`?

src/cli/rustup_mode.rs
```diff
@@ -11,6 +11,7 @@ use clap::{
 };
 use clap_complete::Shell;
 
+use crate::config::new_toolchain_with_reason;
```
rami3l: This import line should be merged to the group right below.

src/cli/rustup_mode.rs
```diff
@@ -1078,133 +1088,112 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
+        };
+
+    let (active_toolchain_name, _active_reason) = active_toolchain_and_reason
+        .as_ref()
+        .map(|atar| (&atar.0, &atar.1))
+        .unzip();
+
+    let active_targets: Vec<ComponentStatus> = active_toolchain_name
+        .and_then(|atn| match atn {
+            ToolchainName::Official(desc) => DistributableToolchain::new(cfg, desc.clone()).ok(),
```
rami3l, line 1106: Is there a particular reason for deviating from the original `DistributableToolchain::try_from(&at.0)` here?

src/cli/common.rs
```diff
@@ -461,74 +462,72 @@ pub(crate) fn list_installed_components(distributable: DistributableToolchain<'_
...
+            if toolchain_meta.is_dir() {
+                format!(" {}", toolchain_path.display())
+            } else {
+                format!(" {}", fs::read_link(toolchain_path)?.display())
+            }
+        } else {
+            String::new()
+        };
+        let status_str = match (is_default, is_active) {
+            (true, true) => " (default, active)",
```
rami3l: As stated in https://github.com/rust-lang/rustup/pull/3225#issuecomment-1793314607 I personally think it's better to use `(active, default)` since `active` will be more important info that I'll look for in this output.

src/cli/rustup_mode.rs
```diff
@@ -1227,31 +1218,27 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
-            } else {
-                writeln!(process().stdout().lock(), "{} (default)", toolchain.name())?;
-            }
-            if verbose {
-                writeln!(process().stdout().lock(), "{}", toolchain.rustc_version())?;
             }
         }
+        None => writeln!(
+            process().stdout().lock(),
+            "no default toolchain is configured"
```
rami3l: You might want to say `active` instead of `default` here?

rami3l (approved):

rami3l:
@majaha Sorry for the late review! I'm finally gaining bandwidth since a few days ago 💦

src/cli/rustup_mode.rs
majaha: Yes: It's because `DistributableToolchain::try_from()` takes a `Toolchain`, which represents an installed toolchain on disk. This is what is returned by `find_or_install_active_toolchain()`.
The new code works with `ToolchainName`, which is what's returned by `find_active_toolchain()` above. This represents the name of a toolchain, which doesn't necessarily have to be installed.
Using `find_active_toolchain()` rather than `find_or_install_active_toolchain()` is what makes `rustup show` no longer install the active toolchain, if it isn't already installed.

majaha:
No worries, hope you enjoyed your break! I've addressed your comments and uploaded the new version.

src/cli/rustup_mode.rs
rami3l: @majaha I see. Thanks for your detailed explanation!

Should `CustomToolchainName` be considered as well though? In this case I think it's okay to assume the toolchain is installed.

rami3l:
@majaha Looks like the `cargo fmt` check has failed, would you mind updating your code again?

majaha:
Ah sorry, there's always something. Should be fixed now.

src/cli/rustup_mode.rs
majaha: This code is for determining the list of targets that are installed for the active toolchain. Rustup only has the ability to query this for Official toolchains, not custom ones, as far as I know. That's what the previous code does, and that's what the `DistributableToolchain` is used for.

To be honest, the variable names could be a bit clearer. And I might put `CustomToolchainName` in the match statement directly. Explicit is better than implicit and all that. I'll change this quickly.

rami3l:
~~@djc I've reviewed this PR and it looks ready in its current state.~~
~~I'd like to merge this PR soon and return to #3501, if that sounds okay to you.~~

Update: After my discussion with @rbtcollins last night we decided to postpone this change to the next milestone to better integrate this PR with the massive changes to be made in #3367 and #3596.

rbtcollins:
Hi, I know I'm coming to the code aspects of this late.

This PR is not focused on changing the output; it is instead:
- changing how we model toolchain selection internally
- changing the semantics for config setting file operations
- refactoring the show function to decompose it somewhat
- and changing the ordering

And perhaps more I've missed.

I'd like to take each of these separately to allow decent discussion about the merits of each, please.

djc:
> Update: After my discussion with @rbtcollins last night we decided to postpone this change to the next milestone to better integrate this PR with the massive changes to be made in #3367 and #3596.

I usually prefer not to consider Discord load-bearing in terms of project decisions. What is the rationale for postponing this change? #3367 and #3596 are large changes but are also mostly mechanical in nature.

rbtcollins:
> What is the rationale for postponing this change? #3367 and #3596 are large changes but are also mostly mechanical in nature.

https://github.com/rust-lang/rustup/pull/3225#issuecomment-1913593183 highlights my concerns.

majaha:
I'm open to discussing the changes. Do you have any specific issues with the code? I also agree that there's no need to postpone this beyond https://github.com/rust-lang/rustup/pull/3367 and https://github.com/rust-lang/rustup/pull/3596, they seem to be a largely separate issue - not to mention that this PR has been finished for some time now, it should really go in first.

djc:
> I'd like to take each of these separately to allow decent discussion about the merits of each, please.

Aren't these already factored into separate commits within this PR?

rami3l:
I still want to merge this, probably before v1.28.0 now that we definitely have a v1.27.1 to launch. I'll try to find some time to give it look once again. My apologies to @majaha for your long wait!

src/cli/rustup_mode.rs
```diff
@@ -1078,133 +1087,106 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
+
+    let (active_toolchain_name, _active_reason) = active_toolchain_and_reason
+        .as_ref()
+        .map(|atar| (&atar.0, &atar.1))
+        .unzip();
+
+    let active_toolchain_targets: Vec<TargetTriple> = active_toolchain_name
+        .and_then(|atn| match atn {
+            ToolchainName::Official(desc) => DistributableToolchain::new(cfg, desc.clone()).ok(),
+            ToolchainName::Custom(_) => None,
```
rami3l, line 1108:
```suggestion
            // So far, it is not possible to list targets for a custom toolchain.
            ToolchainName::Custom(_) => None,
```

rami3l (approved):
v1.27.1 took us longer than expected, but I didn't forget about this PR. I just had another look, and I'm again confirmed that we can include this into v1.28.0.

majaha:
Sounds good. Let me know if I can help with anything.

rami3l:
> Sounds good. Let me know if I can help with anything.

@majaha Looks like your code will generate some clippy warnings with newer Rust. Would you mind fixing them? ❤️

src/cli/rustup_mode.rs
```diff
@@ -1078,133 +1087,107 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
+        match active_toolchain_and_reason {
+            Some((active_toolchain_name, active_reason)) => {
+                let active_toolchain = new_toolchain_with_reason(
+                    cfg,
+                    active_toolchain_name.clone().into(),
+                    &active_reason,
+                )?;
+                writeln!(t.lock(), "name: {}", active_toolchain.name())?;
+                writeln!(t.lock(), "compiler: {}", active_toolchain.rustc_version())?;
+                writeln!(t.lock(), "active because: {}", active_reason.to_string())?;
```
```suggestion
                writeln!(t.lock(), "active because: {}", active_reason)?;
```

src/cli/rustup_mode.rs
```diff
@@ -1227,31 +1212,24 @@ fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
-            {
-            } else {
-                return Err(e);
-            }
-        }
-        Ok((toolchain, reason)) => {
-            if let Some(reason) = reason {
+    match cfg.find_active_toolchain(&cwd)? {
+        Some((toolchain_name, reason)) => {
+            let toolchain = new_toolchain_with_reason(cfg, toolchain_name.clone().into(), &reason)?;
```
```suggestion
            let toolchain = new_toolchain_with_reason(cfg, toolchain_name.clone(), &reason)?;
```

src/cli/download_tracker.rs
```diff
@@ -181,7 +181,7 @@ impl DownloadTracker {
                         let percent = (self.total_downloaded as f64 / content_len as f64) * 100.;
                         let remaining = content_len - self.total_downloaded;
                         let eta_h = Duration::from_secs(if speed == 0 {
-                            std::u64::MAX
+                            u64::MAX
```
djc: Should really inline the clippy fixes into the commit that originated them (maybe try https://github.com/tummychow/git-absorb).

rami3l: @djc TIL about git-absorb! I will give it a try on this one. However the std::u64::MAX wasn't introduced in this patch at all, so it cannot be squashed.

rami3l: Never mind, looks like it has been done elsewhere, so I can safely drop it.

rami3l: @djc Thanks a lot for the recommendation! `git-absorb` literally saved my day :)

majaha:
Nice, well done on merging this guys!

PR #3226, opened 2023-02-19
Fix some more clippy-related issues.

rbtcollins:
Format strings with embedded variables, and passing in structs by value
when they implement the requested trait directly. Neither seem to affect
readability so shouldn't be contentious.


0xPoe (approved):

0xPoe:
Thanks!

PR #3227, opened 2023-02-19
Config mutable

rbtcollins:
Change enough of the Config methods to be mutable to permit the following style of code:
```
test(&|config| {
        config.with_scenario(Scenario::ArchivesV2, &|config| {
            config.expect_ok_ex(....
```

This requires that expect_ok_ex takes a mut reference to self, and everything it calls into.

Without this such a `with_scenario` function would require internal mutability, which would be effectively preventing the borrow checker doing its job.

Alternatively it could make a new Config object with only the distdir field changed, but then we have a sort of anti-aliasing occuring where folk could be confused about what config instance is being referred to.

Another alternative is to not wrap the code under test, but then its possible for the state to be changed and not popped back... possibly that would be better; so this is genuinely asking for thoughts around it

### DIFF
--- a/src/cli/self_update/test.rs
+++ b/src/cli/self_update/test.rs
@@ -36,7 +36,7 @@ fn restore_path(p: Option<RegValue>) {
 }
 
 /// Support testing of code that mutates global path state
-pub fn with_saved_path(f: &dyn Fn()) {
+pub fn with_saved_path(f: &mut dyn FnMut()) {
     // Lock protects concurrent mutation of registry
     lazy_static! {
         static ref LOCK: Mutex<()> = Mutex::new(());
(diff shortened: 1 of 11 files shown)


0xPoe:
> Another alternative is to not wrap the code under test, but then its possible for the state to be changed and not popped back... possibly that would be better; so this is genuinely asking for thoughts around it

The code changes look good to me. But I am wondering if can we make it more obvious instead of using nested closures.
Like: 
```rust
let config = Config::newxxx();

config.set_update_server(xxx).set_scenario(xxx).expect_std_out(r#"xxx"#).run("arg1","arg2");
```
Then we own the config in the whole test and we can do whatever we want.

This looks more clear. What do you think? If you think this change is too big. We can move forward with your PR. I can help improve it later.

rbtcollins:
I have the closure approach working for the performance improvement, so would like to finish that arc, and explore alternatives separately

0xPoe (approved):
Thanks!

PR #3233, opened 2023-02-22
Use workspace inheritance

rbtcollins:
This should reduce rebuilds of packages during testing as they will be
identical.


0xPoe (approved):

0xPoe:
Thanks!

PR #3238, opened 2023-02-23
Use a single integration test binary

rbtcollins:
Much of our test time is linking each test; a single bonanza binary lets
us maintain tests individually, but amortize build/link costs
effectively.


tests/test_bonanza.rs
```diff
@@ -0,0 +1,2 @@
+mod mock;
+mod tests;
```
0xPoe: Can we call it `testsuite`? Then we do not have this kind of `tests/tests` path.

rbtcollins: done

PR #3239, opened 2023-02-23
Test-uses-hardlinks

rbtcollins:
Much of our tests IO overhead is copying a 14MB binary into every test
dir. Unix IO is cheap, but not free. 450 tests ~= 6.6GB of IO. Not all
at the same time, not to the same file, and spread out over time, so
peak rates don't apply. Hardlink rustup-init everywhere in the test
suite, and use link-breaking techniques for the small number of cases
that genuinely need a new file.

### DIFF
--- a/tests/mock/clitools.rs
+++ b/tests/mock/clitools.rs
@@ -9,7 +9,7 @@ use std::fs;
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::process::Command;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock, RwLockWriteGuard};
 use std::time::Instant;
 
 use enum_map::{enum_map, Enum, EnumMap};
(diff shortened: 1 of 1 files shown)


0xPoe:
Wow! It is much faster! 😍

rbtcollins:
There was a spurious failure of one windows build; that seems resolved by retrying - we may have some nondeterminism to chase down, but there's no obvious fault that I can see; and for this large improvement I'm ok with occasional re-runs

rbtcollins:
@hi-rustin this is ready for review

tests/mock/clitools.rs
```diff
@@ -221,19 +232,50 @@ pub fn setup_test_state(test_dist_dir: tempfile::TempDir) -> (tempfile::TempDir,
...
+                // break the *original* so new tests form a new distinct set of
+                // links. Do this by copying to link, breaking the source,
+                // linking back.
+                **lock = 0;
+                fs::copy(original, link)?;
+                fs::remove_file(original)?;
+                hard_link(link, original)
+            }
+        }
+        let mut _lock = CMD_LOCK.write().unwrap();
```
0xPoe:
```suggestion
        let mut lock = CMD_LOCK.write().unwrap();
```
We use it below. So probably we should not have a `_` prefix.

0xPoe (approved):
Thanks!

tests/mock/clitools.rs
rbtcollins: good point, I'll tweak this.

PR #3244, opened 2023-02-25
Bump remove_dir_all

rbtcollins:
While this fixes a TOCTOU bug in the dependency, rustup is run either as an unprivileged process, or as a privileged process operating on a file tree that the privileged user owns; it is not a setuid binary and that mitigates much of the vulnerability: a deliberately misconfigured sudoers or similar would be required. Thus this hardens rustup against misconfiguration, but does not close an active vulnerability in how rustup is designed to be used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,6 +58,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aligned"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "80a21b9440a626c7fc8573a9e3d3a06b75c7c97754c2949bc7857b90353ca655"
+dependencies = [
+ "as-slice",
+]
+
+[[package]]
 name = "android_system_properties"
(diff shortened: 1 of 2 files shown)


0xPoe:
> While this fixes a TOCTOU bug in the dependency,

Sorry, I am not sure I get the point.

Do you mean rustup will delete some files that users do not expect to delete? How will the `parallel` feature help this?

rbtcollins:
Current master uses a version of remove_dir_all that has parallel enabled by default. This PR just keeps the current feature. The parallel deletion feature helps with the performance of removing toolchains - which have 10's of thousands of files.

The reason for the upgrade is a new release - as usual - but also there is a CVE in older versions that means we should upgrade now rather than later.

0xPoe (approved):
I found the [description in the changelog](https://github.com/XAMPPRocky/remove_dir_all/blob/master/CHANGELOG.md#security-changes) for this CVE. Thanks!

PR #3247, opened 2023-02-27
Separate release jobs

rbtcollins:
This is layered on the hardlink PR; it hopefully makes things even faster by only building whats needed in release mode.

PR #3252, opened 2023-03-03
Run the tests on x86_64-pc-windows-msvc

0xPoe:
I think we turned off the tests on windows by accident in this PR. https://github.com/rust-lang/rustup/pull/3247


rbtcollins:
No, this is not needed. You can see 
image
 that the windows tests were run.

0xPoe:
> No, this is not needed.

We just compiled it and do not run any tests.

rbtcollins:
I'd like to make the matrix easier to understand, but this is something we can iterate on. Thanks!

PR #3298, opened 2023-03-31
Add support for listing components with target omitted

cdown:
Configuration management like Chef or Ansible commonly checks this list to know whether a component is present, with a view to installing it if it is not.

This is slightly unwieldy at present because while installing can use an implicit target (e.g., `rustup component add foo`), target-specific components always show their target when listing (e.g., `foo-x86_64-unknown-linux-gnu`). The naive way to tackle this is to just strip after the first dash, but of course this doesn't work well with things like rust-analyzer.

There are other ways to achieve this (e.g., split from the end, since the target is always a triple, although that can also get tricky when the component is target agnostic), but it would be nice to just have a --no-target flag which permits just showing the component short name for tools which just want to simply check if the component is installed or not, and don't really care about anything other than the implied target.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -384,12 +384,17 @@ pub(crate) fn list_installed_targets(toolchain: &Toolchain<'_>) -> Result<utils:
     Ok(utils::ExitCode(0))
 }
 
-pub(crate) fn list_components(toolchain: &Toolchain<'_>) -> Result<utils::ExitCode> {
+pub(crate) fn list_components(toolchain: &Toolchain<'_>, short: bool) -> Result<utils::ExitCode> {
     let mut t = term2::stdout();
     let distributable = DistributableToolchain::new_for_components(toolchain)?;
     let components = distributable.list_components()?;
     for component in components {
-        let name = component.name;
+        let name = if short {
+            component.component.short_name_in_manifest()
+        } else {
+            &component.name
+        };
+
         if component.installed {
             t.attr(term2::Attr::Bold)?;
             writeln!(t, "{name} (installed)")?;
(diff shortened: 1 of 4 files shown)


rbtcollins:
Hi, the problem is that the target for the component may not be the target that you want or expect. 

I suggest either parsing them properly(*), or use rustup as an idempotent command:

For instance on Windows (so slow to invoke commands).

```
$ time `rustup component add clippy`
info: component 'clippy' for target 'x86_64-pc-windows-msvc' is up to date

real    0m0.235s
user    0m0.000s
sys     0m0.000s
$ echo $?
0
```

If its really going to make things better I'm totally fine with adding this moderate complexity, but given the distribution model we use I'm quite certain that this patch will just hide genuine bugs for those configuration management systems.

A compromise would be to only show non-default-target targets on components, which might actually be a nice UI improvement that we could do without an option. I'd like to get some wider input on that before doing it, because we haven't changed this part of rustup in quite a long time.

For 'parsing them properly' the triples we use are quite standard gnu style host triples, so a parser for those should exist in Ruby, Python etc already - take all the substrings from the right side of the component until one validates correctly.

Alternatively, or in conjunction you can hard code the list of components. While it is strictly speaking data driven, I think the reality is that for your use case, you have a fixed set of components you're querying for : this is not data driven, and you  can be completely sure of the expected outputs.

cdown:
Thanks for the detailed reply! I think that clarifies things nicely.

> A compromise would be to only show non-default-target targets on components, which might actually be a nice UI improvement that we could do without an option. I'd like to get some wider input on that before doing it, because we haven't changed this part of rustup in quite a long time.

That would also work. Who would need to be part of that discussion? :-)

rbtcollins:
@cdown probably a working patch (to surface any non obvious complexities) and then perhaps a discussion on rust-internals ? We have a new area lead for devtools I think (if the nominations stuff has completed), so they might have an idea of concrete suspects to round up.

cdown:
I will drop this for now. Thanks for looking!

PR #3310, opened 2023-04-04
Better error when a ValueValidation error occurs

0xPoe:
Yet another solution for https://github.com/clap-rs/clap/discussions/4799

See https://github.com/clap-rs/clap/discussions/4799#discussioncomment-5460445

### DIFF
--- a/tests/suite/cli-ui/rustup/rustup_unknown_arg_stdout.toml
+++ b/tests/suite/cli-ui/rustup/rustup_unknown_arg_stdout.toml
@@ -1,51 +1,9 @@
 bin.name = "rustup"
 args = ["random"]
 status.code = 1
-stdout = """
-rustup [..]
-The Rust toolchain installer
-
-USAGE:
-    rustup[EXE] [OPTIONS] [+toolchain] <SUBCOMMAND>
-
-ARGS:
-    <+toolchain>    release channel (e.g. +stable) or custom toolchain to set override
-
-OPTIONS:
-    -v, --verbose    Enable verbose output
-    -q, --quiet      Disable progress output
-    -h, --help       Print help information
-    -V, --version    Print version information
-
-SUBCOMMANDS:
-    show           Show the active and installed toolchains or profiles
-    update         Update Rust toolchains and rustup
-    check          Check for updates to Rust toolchains and rustup
-    default        Set the default toolchain
-    toolchain      Modify or query the installed toolchains
-    target         Modify a toolchain's supported targets
-    component      Modify a toolchain's installed components
-    override       Modify directory toolchain overrides
-    run            Run a command with an environment configured for a given toolchain
-    which          Display which binary will be run for a given command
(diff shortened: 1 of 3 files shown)


tests/suite/cli-ui/rustup/rustup_unknown_arg_stdout.toml
```diff
@@ -0,0 +1,51 @@
+bin.name = "rustup"
+args = ["random"]
+status.code = 1
+stdout = """
+rustup 1.25.2+252 (3438037c7 2023-04-02) dirty 2 modifications
```
0xPoe: TODO: UPDATE IT

tests/suite/cli-ui/rustup/rustup_unknown_arg_stdout.toml
```diff
@@ -0,0 +1,51 @@
...
+    check          Check for updates to Rust toolchains and rustup
+    default        Set the default toolchain
+    toolchain      Modify or query the installed toolchains
+    target         Modify a toolchain's supported targets
+    component      Modify a toolchain's installed components
+    override       Modify directory toolchain overrides
+    run            Run a command with an environment configured for a given toolchain
+    which          Display which binary will be run for a given command
+    doc            Open the documentation for the current toolchain
+    man            View the man page for a given command
```
0xPoe: TODO: UPDATE IT

rbtcollins:
The test shows nice behaviour. It seems like clap has some way to go to make this have a really nice interface.

0xPoe:
> The test shows nice behaviour. It seems like clap has some way to go to make this have a really nice interface.

I changed it again. Because the help text in [this commit](https://github.com/rust-lang/rustup/pull/3310/commits/b579e23a7503f7943fd80c7d5a1da67a8aebfa22) is still not very user-friendly. Because it does not show the value which the user passed to the rustup.

Also, I think print help text for users is not consistent with other errors. So I prefer to just tell the users why we print the `+toolchain` error. Instead of printing the whole help text(it still does not explain why we treat it as  a toolchain name.)

tests/suite/cli_misc.rs
```diff
@@ -946,7 +946,10 @@ fn override_by_toolchain_on_the_command_line() {
         );
         config.expect_err(
             &["rustup", "@stable", "which", "rustc"],
-            r#"Invalid value "@stable" for '<+toolchain>': Toolchain overrides must begin with '+'"#,
+            "Invalid value \"@stable\" for '<+toolchain>': \"@stable\" is not a valid subcommand, \
+            so it was interpreted as a toolchain name, but it is also invalid. \
+            To override the toolchain using the 'rustup +toolchain' syntax, \
+            make sure to prefix the toolchain override with a '+'",
```
rbtcollins: We can probably delete this bit of the integration test; your new test confirms the same behaviour

PR #3333, opened 2023-04-25
Compile static Mutex where possible

workingjubilee:
A few of Rustup's `lazy_static!` instances are for simple locks. These are easy to evaluate at compile time, and doing so is stable, so reduce usage of the more heavyweight dependency.

I intend to follow up with more ambitious refactorings to slim down this kind of initializer code (more of it actually in the CLI, thus more of it likely to affect startup time), but these two were so trivial it seemed obvious enough to submit this first.


0xPoe (approved):
Looks good to me. Thanks!

PR #3338, opened 2023-04-27
Suggest right toolchain when running clippy

0xPoe:
We should use beat as we discussed before.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,12 +4,14 @@
 2. Create your feature branch: `git checkout -b my-new-feature`
 3. Test it: `cargo test`
 4. Lint it: `cargo +beta clippy --all --all-targets -- -D warnings`
+
 > We use `cargo clippy` to ensure high-quality code and to enforce a set of best practices for Rust programming. However, not all lints provided by `cargo clippy` are relevant or applicable to our project.
 > We may choose to ignore some lints if they are unstable, experimental, or specific to our project.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -4,12 +4,14 @@
...
 > We use `cargo clippy` to ensure high-quality code and to enforce a set of best practices for Rust programming. However, not all lints provided by `cargo clippy` are relevant or applicable to our project.
 > We may choose to ignore some lints if they are unstable, experimental, or specific to our project.
 > If you are unsure about a lint, please ask us in the [rustup Discord channel](https://discord.com/channels/442252698964721669/463480252723888159).
-5. Commit your changes: `git commit -am 'Add some feature'`
-6. Push to the branch: `git push origin my-new-feature`
-7. Submit a pull request :D
+
+1. Commit your changes: `git commit -am 'Add some feature'`
+2. Push to the branch: `git push origin my-new-feature`
+3. Submit a pull request :D
```
rbtcollins: Is the renumbering deliberate, or a side effect of some editor plugin? These steps are still part of the sequence started on line 3.

rbtcollins (changes requested):
.

CONTRIBUTING.md
0xPoe: Oops. Updated by my markdown linter. Reverted!

rbtcollins (approved):

PR #3340, opened 2023-04-28
Rework Toolchain model and drop relative file path overrides

### LINKED ISSUE #3130: rustup allows `rustup toolchain link none`

rbtcollins:
Rework Toolchain model and drop relative file path overrides
    
Relative path overrides permit a freshly downloaded source tree to
execute arbitrary code on any rustup command that executes a binary from
the configured toolchain, and its a reasonable tradeoff for us to remove
this feature. Absolute path overrides are kept intact - these were added
to support users of large monorepo tool systems, and can be kept with
reasonable safety.
    
Introduce an interior model for toolchain names which allows the
elimination of conditional code for custom vs named vs path based
toolchains. Fixes #3130. Close https://github.com/rust-lang/rustup/pull/3174
    
Finishes the separation of Toolchain vs DistributableToolchain vs
CustomToolchain started some time ago.
    
Moves some test support into the src tree in order to keep symbols
private and adds a 'test' feature required for self-testing, ensuring
that that code doesn't affect production builds.

Adds in a new build that tests the combination of features together (on just one platform), to ensure all combinations are valid.


src/cli/common.rs
```diff
@@ -178,51 +181,71 @@ pub(crate) fn set_globals(verbose: bool, quiet: bool) -> Result<Cfg> {
 
 pub(crate) fn show_channel_update(
     cfg: &Cfg,
-    name: &str,
+    name: PackageUpdate,
```
0xPoe, line 184: Perhaps we should consider using a different term than "name"?

src/toolchain/names.rs
```diff
@@ -0,0 +1,712 @@
+//! Overview of toolchain modeling.
+//!
+//! From the user (including config files, toolchain files and manifests) we get
+//! a String. Strings are convertable into `MaybeOfficialToolchainName`,
```
0xPoe:
```suggestion
//! a String. Strings are convertible into `MaybeOfficialToolchainName`,
```

tests/suite/cli-ui/rustup/rustup_unknown_arg_stdout.toml
```diff
@@ -3,7 +3,7 @@ args = ["random"]
 status.code = 1
 stdout = ""
 stderr = """
-error: Invalid value \"random\" for '<+toolchain>': \"random\" is not a valid subcommand, so it was interpreted as a toolchain name, but it is also invalid. To override the toolchain using the 'rustup +toolchain' syntax, make sure to prefix the toolchain override with a '+'
+error: Invalid value \"random\" for '<+toolchain>': error: \"random\" is not a valid subcommand, so it was interpreted as a toolchain name, but it is also invalid. To override the toolchain using the 'rustup +toolchain' syntax, make sure to prefix the toolchain override with a '+'
```
0xPoe, line 6: There are two `error:`. It is a little bit confusing.

src/toolchain/names.rs
```diff
@@ -0,0 +1,712 @@
...
+    //         let candidate = validate(candidate)?;
+    //         if candidate.parse::<PartialToolchainDesc>().is_ok()
+    //             || candidate == "none"
+    //             || candidate.contains('/')
+    //             || candidate.contains('\\')
+    //         {
+    //             Err(InvalidName::CustomName(candidate.into()))
+    //         } else {
+    //             Ok(CustomToolchainName(candidate.into()))
+    //         }
```
0xPoe: Can we remove this?

0xPoe (approved):
I have noticed a few minor issues. 

Looks great and much clearer. Thank you!

src/cli/common.rs
rbtcollins: I think a future cleanup can just remove this intermediate function entirely

src/toolchain/names.rs
rbtcollins: oh, yes, I had it here for reference.

matklad:
>Relative path overrides permit a freshly downloaded source tree to
execute arbitrary code on any rustup command that executes a binary from
the configured toolchain, and its a reasonable tradeoff for us to remove
this feature. Absolute path overrides are kept intact - these were added
to support users of large monorepo tool systems, and can be kept with
reasonable safety.

I think relative paths are a subset of absolute paths, so, security-wise, this is a no-op.

Namely, `./foo/bar` relative path is equivalent to `/proc/self/cwd/foo/bar` absolute path.

matklad:
Verified with

```
[toolchain]
# path = ".cargo"
path = "/proc/self/cwd/.cargo"
```

on

https://github.com/jonas-schievink/mallory

jonhoo:
That's definitely unfortunate. I think removing path-based toolchains altogether is going to break things, especially for more corporate users with additional build orchestration around Cargo (e.g., Amazon). How feasible is it to disallow specifically `/proc` (and maybe also `/sys`)? Do we know of other absolute-to-relative proxies like this?

matklad:
@rbtcollins is the issue that absolute paths are also unsafe tracked anywhere?

rbtcollins:
Not yet, please do open a bug from this, and we'll have to discuss whether to give up, filter some absolute paths and hope that's sufficient, or some other thing.

PR #2828, opened 2021-07-16
Switch to indicatif for the progress bar

kirawi:
Resolves #1826 #1835

Output on Alacritty:
```
info: profile set to 'default'
info: default host triple is x86_64-pc-windows-msvc
info: syncing channel updates for 'stable-x86_64-pc-windows-msvc'
info: latest update on 2021-06-17, rust version 1.53.0 (53cb7b09b 2021-06-17)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
 16.13MiB / 16.13MiB (100%) 5.85MiB/s in 2s ETA: 0s
info: downloading component 'rust-std'
 22.78MiB / 22.78MiB (100%) 3.75MiB/s in 6s ETA: 0s
info: downloading component 'rustc'
 59.68MiB / 59.68MiB (100%) 4.08MiB/s in 14s ETA: 0s
info: downloading component 'rustfmt'
info: installing component 'cargo'
 3.59MiB / 3.59MiB (100%) 2.29MiB/s in 1s ETA: 0s
info: installing component 'clippy'
info: installing component 'rust-docs'
 16.13MiB / 16.13MiB (100%) 632.90KiB/s in 26s ETA: 0s
info: installing component 'rust-std'
 22.78MiB / 22.78MiB (100%) 1.79MiB/s in 12s ETA: 0s
info: installing component 'rustc'
 59.68MiB / 59.68MiB (100%) 3.20MiB/s in 18s ETA: 0s
info: installing component 'rustfmt'
info: default toolchain set to 'stable-x86_64-pc-windows-msvc'

  stable-x86_64-pc-windows-msvc installed - rustc 1.53.0 (53cb7b09b 2021-06-17)


Rust is installed now. Great!
```
Alternatively, I can also look into integrating `crossterm` instead.

26/8/2021: I'll get back to this when I can 😅, very busy right now.

PR #3349, opened 2023-05-06
Use indicatif as the progress bar

### LINKED ISSUE #1826: Progress indicator broken on Alacritty with ConPTY

### LINKED ISSUE #1835: Investigate using a community maintained progress bar

0xPoe:
Close https://github.com/rust-lang/rustup/pull/2828

Close https://github.com/rust-lang/rustup/issues/1826 

Close https://github.com/rust-lang/rustup/issues/1835

Big thanks for https://github.com/rust-lang/rustup/pull/2828

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -348,6 +348,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "acbf1af155f9b9ef647e42cdc158db4b64a1b61f743629225fde6f3e0be2a7c7"
 
 [[package]]
+name = "console"
+version = "0.15.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3d79fbe8970a77e3e34151cc13d3b3e248aa0faaecb9f6091fa07ebefe5ad60"
+dependencies = [
+ "encode_unicode",
+ "lazy_static",
+ "libc",
+ "unicode-width",
+ "windows-sys 0.42.0",
+]
+
+[[package]]
 name = "content_inspector"
 version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


ryanavella:
I can confirm this resolves the issue for me on FreeBSD + xterm[-256color]. My symptoms were identical to #1826 even though that was reported for a different environment (Alacritty + ConPTY)

0xPoe:
This PR will cause some display issues when changing the window size of the terminal. So I am going to close it for now.

0xPoe:
I will revisit it and try to fix the resize issue.

cc: @rust-lang/rustup

PR #3351, opened 2023-05-08
Replace term with termcolor

### LINKED ISSUE #1818: Shift away from `term` crate

### LINKED ISSUE #2292: Colors are broken on MSYS2 shell.

0xPoe:
close https://github.com/rust-lang/rustup/issues/1818

Based on https://github.com/rust-lang/rustup/pull/2790/files. Big thanks for [aloucks](https://github.com/aloucks).

Probably close https://github.com/rust-lang/rustup/issues/2292.


src/cli/term.rs
```diff
@@ -0,0 +1,160 @@
...
+        self.inner.lock().unwrap().reset()
+    }
+
+    fn carriage_return(&mut self) -> io::Result<()> {
+        self.inner.lock().unwrap().carriage_return()
+    }
+}
+
+pub(crate) fn stdout() -> Box<dyn Terminal> {
+    Box::new(SharedTerminal {
```
0xPoe: I've discovered that caching the stdout is quite difficult in this situation. As there are some tests running in the same process, using a once cell for caching is not a straightforward solution.

A possible solution could be to utilize `process().id()` to decide whether a new terminal needs to be created. However, this approach may involve certain lock options and mutability concerns. I am uncertain if this is a good solution since we aim for improved performance.

0xPoe: Only cache the `isatty` to solve this problem. See https://github.com/rust-lang/rustup/pull/3351/commits/f76f961b407f731da9499446d8e7458fcbfb9e03

Cargo.toml
```diff
@@ -16,12 +16,7 @@ version = "1.26.0"
 
 [features]
 curl-backend = ["download/curl-backend"]
-default = [
-    "curl-backend",
-    "reqwest-backend",
-    "reqwest-default-tls",
-    "reqwest-rustls-tls",
-]
+default = ["curl-backend", "reqwest-backend", "reqwest-default-tls", "reqwest-rustls-tls"]
```
rbtcollins: I'd like us to figure out a canonical representation for this file and then stick with it. I've been using the vscode `Better toml` extension which has an option to sort table content, something very nice for avoiding spurious conflicts.

Cargo.toml
```diff
@@ -93,6 +81,7 @@ url.workspace = true
 wait-timeout = "0.2"
 xz2 = "0.1.3"
 zstd = "0.12"
+atty = "0.2.14"
```
rbtcollins: Please sort this (see my suggestion about IDE support up above - though finding a third party CLI tool to do it would be nice too).

src/cli/markdown.rs
```diff
@@ -1,20 +1,17 @@
...
-use std::io;
-
 use pulldown_cmark::{Event, Tag};
 
-use super::term2::{color, Attr, Terminal};
+use super::term::{Attr, Color, Terminal};
 
 // Handles the wrapping of text written to the console
-struct LineWrapper<'a, T: Terminal> {
+struct LineWrapper<'a> {
```
rbtcollins, line 9: What drives the change to `dyn Terminal` rather than a type parameter?

src/cli/rustup_mode.rs
```diff
@@ -76,11 +75,11 @@ pub fn main() -> Result<utils::ExitCode> {
     let matches = match cli().try_get_matches_from(process().args_os()) {
         Ok(matches) => Ok(matches),
         Err(err) if err.kind() == DisplayHelp => {
-            write!(process().stdout().lock(), "{err}")?;
```
rbtcollins: If I remember correctly the use of `.lock()` was needed to avoid race conditions, because stdout returns a thread-usable, but not locked, output; and write! is not atomic.

src/utils/tty.rs
```diff
@@ -1,42 +1,17 @@
...
-#[cfg(windows)]
 pub(crate) fn stderr_isatty() -> bool {
-    isatty(winapi::um::winbase::STD_ERROR_HANDLE)
-}
-
-#[cfg(unix)]
-pub(crate) fn stdout_isatty() -> bool {
-    isatty(libc::STDOUT_FILENO)
+    // Cache the terminal detection for performance reasons.
+    // atty::is() is a relatively expensive IO operation.
```
rbtcollins: This should be cached in the Process abstraction, not here: because its process specific but we run multiple processes in parallel during testing.

It looks like the old code was likely overlooked / buggy.

src/currentprocess/filesource.rs
```diff
@@ -193,14 +185,36 @@ impl Isatty for TestWriter {
...
+    fn attr(&mut self, _: crate::cli::term::Attr) -> io::Result<()> {
+        Ok(())
+    }
+
+    fn reset(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+
+    fn carriage_return(&mut self) -> io::Result<()> {
+        self.0.lock().unwrap().write(b"\r").map(|_| ())
```
rbtcollins: `Ok(self.0.lock()?.write(b"\r")?;())`

might be a little better. Or not - whatever you like.

src/currentprocess/filesource.rs
```diff
@@ -166,20 +167,11 @@ impl Write for TestWriterLock<'_> {
...
-        Box::new(TestWriterLock {
-            inner: self.0.lock().unwrap_or_else(|e| e.into_inner()),
-        })
-    }
-}
-
 impl Write for TestWriter {
     fn write(&mut self, buf: &[u8]) -> Result<usize> {
-        self.lock().write(buf)
+        self.0.lock().unwrap().write(buf)
```
rbtcollins: I'd -really- like us to avoid `unwrap()` where its not needed. Can't `?` be used?

src/cli/term.rs
rbtcollins: so, process FileSource could answer isatty, to push that into the abstraction over env + fd's that is process().

Then caching that is local - it would have to drop the cache under some circumstances for correctness, but I don't think we ever play those games with fd 0/1/2 that would cause that to be needed.

That said, there are two things that raise concerns for me.

One is the removal of Writer from FileSource - that would then require Terminal to implement Writer itself - but it doesn't:

https://docs.rs/termcolor/latest/src/termcolor/lib.rs.html#443-445 shows that LockedStandardTerminal is the equivalent. 

For thread safety, and this is something Rust won't detect today (or perhaps it was and that is what you found hard last week) we *must* not use impl Write, but a locked Write (analogous to https://doc.rust-lang.org/std/io/struct.StdoutLock.html).

Failure to do that will lead to write striping from different threads.

rbtcollins (changes requested):
My suggestion:

Don't push Terminal into FileSource, instead keep the abstraction we had before - Writer and Write - and use StandardStream and StandardStreamLock as appropriate on top of that. Push isatty into FileSource (it probably should be anyway for testing support), but we shouldn't be constructing these a lot anyway - even with cross-thread concerns. In term.rs it could make sense to have a long lived reference to a StandardStream.

Note that the Inner is simple : https://docs.rs/termcolor/latest/src/termcolor/lib.rs.html#465 - and should be threadsafe itself if W is, so its not clear what was going on. 

Do you have the git ref that shows the problem you encountered?

0xPoe:
> Do you have the git ref that shows the problem you encountered?

You can check this one https://github.com/hi-rustin/rustup/commit/0cd54157915e554a6845a12cd5a6fb87d77bb22f

The test failure is caused by we cache the terminal globally. But we execute multiple tests in the same process when we do the integration test.
For example:
```rust
#[test]
fn remove_override() {
    for keyword in &["remove", "unset"] {
        test(&|config| {
            config.with_scenario(Scenario::SimpleV2, &|config| {
                let cwd = config.current_dir();
                config.expect_ok(&["rustup", "override", "add", "nightly"]);
                config.expect_ok_ex(
                    &["rustup", "override", keyword],
                    r"",
                    &format!("info: override toolchain for '{}' removed\n", cwd.display()),
                );
            })
        });
    }
}
```

You can notice there are two operations in the same closure. Although they are using different test processes, the program process is the same.

Cargo.toml
0xPoe: It is weird. I am also using `Better toml`.

0xPoe: Solved by using `Even Better TOML`.

src/currentprocess/filesource.rs
0xPoe: Updated!

0xPoe: Updated!

src/currentprocess/filesource.rs
0xPoe: There is no from implementation for this error. So I use expect here.

src/cli/rustup_mode.rs
0xPoe: Some parts of our code are locked and some parts are not. What is the basis for it?

rbtcollins: This is what happens with unlocked terminals and threads:
```
ColorSpec { fg_color: Some(Cyan), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
ColorSpec { fg_color: Some(Cyan), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
ColorSpec { fg_color: Some(Cyan), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
ColorSpec { fg_color: Some(Cyan), bg_color: None, bold: false, intense: ColorSpec { fg_color: Some(Green), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
ColorSpec { fg_color: Some(Green), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
ColorSpec { fg_color: Some(Green), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
ColorSpec { fg_color: Some(Green), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
ColorSpec { fg_color: Some(Green), bg_color: None, bold: false, intense: false, underline: false, dimmed: false, italic: false, reset: true, strikethrough: false } text!
```

Generated from https://github.com/rbtcollins/console-race

rbtcollins: If you `cargo run -- 1` instead of simple cargo run, you'll get a version using locked streams, and that will not mix up colors or lines at all. Example above showing the problem was generated on Linux.

Basically *everything* that is writing a non-atomic unit of IO to the console must use a locked writer.

So for instance a single refresh of a multiline progress bar must run against a locked stream.
The outputting of a status message that makes exactly one IO call - e.g. 'foo bar\n' - is implicitly locked in some cases - the write() syscall is generally defined as race-free - it either writes some bytes or fails, but doesn't write some bytes interleaved with another write() syscall. The problem is that write! and friends don't make a single syscall. They make many. And higher level abstractions like termcolor make even more syscalls. One for setting the color, one for the content being written.

So at the file descriptor level we get:
write(1, b"<color escape code");
write(1, b"Color Spec {");
write(1, b"fg_color: ");
write(1, b"Some(");
write(1, b"Cyan");
write(1, b"), ");

And when two threads are doing that at the same time, any one thread is sequential, but the shared file descriptor can see any of those from any thread with any arbitrary interleaving.

rbtcollins:
> > Do you have the git ref that shows the problem you encountered?
> 
> You can check this one [hi-rustin@0cd5415](https://github.com/hi-rustin/rustup/commit/0cd54157915e554a6845a12cd5a6fb87d77bb22f)
...
> You can notice there are two operations in the same closure. Although they are using different test processes, the program process is the same.

Ah I think I meant further back than that - you were asking about a thread compile error - missing Sync or some such thing somewhere.

src/cli/markdown.rs
0xPoe: Because we return a Box<dyn Terminal> from the `term::stdout`.

src/utils/tty.rs
0xPoe: Fixed in https://github.com/rust-lang/rustup/pull/3351/commits/4b3d9e790ffe9770c2b03db0a561294e71cd29e2

src/cli/rustup_mode.rs
0xPoe: Fixed in https://github.com/rust-lang/rustup/pull/3351/commits/8af7e6a4d0d85ab9c4c8af43db4b2af5d8868a2d

0xPoe:
> Ah I think I meant further back than that - you were asking about a thread compile error - missing Sync or some such thing somewhere.

Oh, I want to pass a `TermLike` into indicatif. It requires the Sync trait. You can see https://docs.rs/indicatif/latest/indicatif/trait.TermLike.html.

src/cli/rustup_mode.rs
0xPoe: Maybe we can improve it in the future. We could pass a context to each function that contains the terminal and provides the `info`, `error`, and `warn` functions. This way we don't need to go around accessing process().stdout().

Cargo.toml
```diff
@@ -62,7 +63,7 @@ home = "0.5.4"
 lazy_static.workspace = true
 libc = "0.2"
 num_cpus = "1.15"
-once_cell = { workspace = true, optional = true }
```
0xPoe: I will revert this change. It is not necessary now.

0xPoe (commented):
I will squash the commits and do some manual tests after your review.

src/cli/rustup_mode.rs
rbtcollins: Maybe! The underlying challenge is this: rust std defines stdout / stderr / stdin as thread-agnostic globals (modelled via a function wrapping a OnceLock) rather than a File which would have the normal borrow checker constraints over it.

So there is an accessible-via-any thread struct with interior mutability as the *contract* for interacting with the terminal.

Then CurrentProcess comes in, which seeks to replace this global state with isolation, to permit concurrent tests without the overhead of forking/execing :- it has to meet the existing contract, because other libraries and code often want to work that way.

src/cli/term.rs
```diff
@@ -0,0 +1,120 @@
...
+pub trait Terminal: Writer {
+    fn fg(&mut self, color: Color) -> io::Result<()>;
+    fn bg(&mut self, color: Color) -> io::Result<()>;
+    fn attr(&mut self, attr: Attr) -> io::Result<()>;
+    fn reset(&mut self) -> io::Result<()>;
+    fn carriage_return(&mut self) -> io::Result<()>;
+}
+
+/// Disable all terminal controls on non-tty's.
+pub(crate) struct StandardStreamTerminal {
```
rbtcollins: I think this should be pub(super), not pub(crate) - the terminal abstraction is really related to the edge of the code, and shouldn't leak to the core code base; I note that it has leaked to CurrentProcess, which pub(super) would have prevented.

src/currentprocess/filesource.rs
```diff
@@ -79,71 +79,38 @@ impl StdinSource for super::TestProcess {
...
-/// Stand-in for std::io::Stdout
-pub trait Writer: Write + Isatty + Send {
+/// Holds a terminal or other output stream.
+pub trait Writer: Send + Sync {
+    /// Locks the terminal so that you can write to it.
     fn lock(&self) -> Box<dyn WriterLock + '_>;
 }
 
-/// Stand-in for std::io::stdout
+// -------------- stdout -------------------------------
```
rbtcollins, line 110: Why the change from a docstring to a mere comment?

src/currentprocess/filesource.rs
```diff
@@ -79,71 +79,38 @@ impl StdinSource for super::TestProcess {
...
 }
 
-// -------------- stdout -------------------------------
-
-pub trait Isatty {
-    fn isatty(&self) -> bool;
-}
-
-/// Stand-in for std::io::StdoutLock
+/// Allows you to write to a terminal or other output stream.
```
rbtcollins: I think this loses key information - std::io::StdoutLock is a crucial part of the behaviour this is emulating. I'm happy for more information to be added to the docstring, but I don't think replacing what is there with what you've got is sufficient.

src/currentprocess/filesource.rs
```diff
@@ -79,71 +79,38 @@ impl StdinSource for super::TestProcess {
...
+    /// Locks the terminal so that you can write to it.
     fn lock(&self) -> Box<dyn WriterLock + '_>;
 }
 
-/// Stand-in for std::io::stdout
+// -------------- stdout -------------------------------
+
 pub trait StdoutSource {
-    fn stdout(&self) -> Box<dyn Writer>;
+    fn stdout(&self) -> Box<dyn Terminal>;
```
rbtcollins: This should still return Writer in my opinion. Can you help me understand why it needs to return Terminal?

src/currentprocess/filesource.rs
```diff
@@ -79,71 +79,38 @@ impl StdinSource for super::TestProcess {
...
 }
 
 // -------------- stderr -------------------------------
 
-/// Stand-in for std::io::stderr
```
rbtcollins, line 108: I'm not clear why this docstring is removed either?

src/cli/markdown.rs
rbtcollins: Ok, but why do we do that?

rbtcollins:
ok, so coming back to the dyn question; because we implement an analog to std::io::stdout(), most code doesn't know the concrete type that will be returned. So we return a Box<dyn Stdout> rather than being specialised by some T=std::io::StdOut. Doing that might be better, but it would require threading that type T throughout the code base - I did attempt that, and it was pretty awful - thousands and thousands of lines of churn, and a type parameter on nearly every struct in the codebase. Console IO probably isn't such a common call-path that removing the one dyn is critical (but perhaps in future).

However, I see a more subtle problem with your patch.

```
pub(super) struct StandardStreamTerminal {
    
    stream: StandardStream,
    color: ColorSpec,
}
```

This depends on term_color::StandardStream, which itself is defined as:
```
pub struct StandardStream {
    wtr: LossyStandardStream<WriterInner<IoStandardStream>>,
}
```

Thats problematic because our TestProcess has no room to hook into that that all. I *presume* thats why CurrentProcess was changed to return Terminal types, not Stdout / Stderr look-alikes.

Thats a shame, but I guess natural.

I'd like to suggest we put these crate updates aside and don't do them yet.

Instead, lets refactor all our console IO to the following shape without changing dependencies, and then after that it should be really easy to drop new things in.

The new shape is:
- always run rustup under a tokio runtime [so we can write this once rather than twice]
- (for non-proxy mode only) we have a dedicated console IO thread, that both handles read/prompts, and stdout/stderr.
- both an async and a sync proxy object that lets some arbitrary thread make requests from, or submit IO to, the dedicated console IO thread. The requests should be things that should be safe to interleave (so e.g. the proxy object shouldn't implement `Write` but `Writer`, and only submit writes on unlock()).
  - the sync version can block when IO backpressure occurs
  - the async version needs to poll NotReady when IO backpressure occurs
  - probably backed by [tokio](https://docs.rs/tokio/1.28.1/tokio/index.html)::[sync](https://docs.rs/tokio/1.28.1/tokio/sync/index.html)::[mpsc](https://docs.rs/tokio/1.28.1/tokio/sync/mpsc/index.html#)
 - the existing Std{out/err}Source family would not be modified.
 - a new couple of traits now return the proxy to the IO thread ... and we go through all our macros and everything else to migrate to the new interface
 - test code should still use the same cross-thread interface, permitting the return type of StdoutSource etc to become a concrete type
 - the implementation running in the IO worker thread is now where the type specialisation occurs - it is implementation specific - real IO or test IO
 - we have a Terminal abstraction parallel to the Stdout/Stderr etc abstractions in the currentprocess module. I would lean towards a new file 'terminalsource.rs', but perhaps putting it in 'filesource.rs' would be ok.
   - this too would offer both sync and async forms and be communicating over mpsc to the IO thread
   - the IO thread implementation is where we would be able to switch to termcolor without any challenge or problems, and it will naturally have a single instance for the lifetime of rustup processes.
  
Theres a little hair in there related to getting into an async mode, but I think I've got that cracked - the otel code seems pretty reliable at this point, so expanding our async coverage bit by bit could be really nice.

The main bit thats missing is adding in a shim to the tokio runtime for every new thread we create - which I think at this point is just the gnarly IO worker pools, which I think don't even actually do console IO themselves ... so we might be able to defer that for a while yet.

The key benefit of this approach is that we can do a series of small patches and not have to end up with a big-change at any point. Even though I did that just recently with ToolchainNames its something to avoid wherever possible!

Concrete steps would be:
1) make the tokio runtime always enabled
2) write a new console IO thread (that does nothing) and start it in both `main` (so around `maybe_trace_rustup`) and from the `run_inprocess` test helper (or nearby).
3) start with the console IO thread using the current StdoutSource etc functions.
5) add in the new interface to get proxies that talk to this console IO thread
6) migrate the rest of the codebase to use the new proxy based interface
7) make the original StdoutSource etc private, or move them, or even delete; at the same time make the console IO thread be specialised, so we have a different IO thread implementation for real IO vs test IO. `main()` would then know to use real IO, and everything test would not.
8) add interface for getting a terminal that just uses the current terminal abstraction (which will be stacked on the mpsc channel ultimately), and migrate the terminal-getting code calls to use that.
9) migrate the terminal implementation logic into the console IO thread implementations (sibling struct or whatever, nicely abstracted, but specific to real IO, and for the test IO thread an implementation that tosses colour away entirely). This changes the terminal implementation the rest of the uses from terminal(mpsc-proxy-for-stdout) to terminal-that-is-mpsc-proxy-to-io-thread, and the terminal itself becomes terminal(stdout) or terminal(stderr) with no polymorphism needed.
10) we then switch terminal for term_color, and only the console IO thread cares.

src/cli/term.rs
0xPoe: There isn't a good way not to expose it. Because we need to get the terminal implementation in the current process. So we at least need to expose it for `CurrentProcess`.

src/currentprocess/filesource.rs
hi-rustin: Because the original comment said it is `Stand-in` instead of `Stand-out`. I guess it is a typo.

src/currentprocess/filesource.rs
0xPoe: As you said in https://github.com/rust-lang/rustup/pull/3351#issuecomment-1556025976, we cannot abstract it with the writer.

src/currentprocess/filesource.rs
0xPoe: Reverted! Thanks!

src/currentprocess/filesource.rs
0xPoe: Added more info.

src/currentprocess/filesource.rs
0xPoe: Revered!

0xPoe (commented):
Self check

rbtcollins:
Putting organisational style comments here, as discord isn't googleable :P.

I think the approach I'm proposing only has one real question, which is "is tokio's platform support wide enough for rustup" ?

[Rust Tiers](https://doc.rust-lang.org/nightly/rustc/platform-support.html) are basically trailing definitions - a platform with everything working and tested is Tier 1, a platform where folk can developer is Tier2+host tools, and one everything builds is Tier 2, and Tier 3 is best effort.

So for Rustup to adopt something that we can't fix easily ourselves as a dependency, we want it to support Tier 1 and Tier2+Host Tools, otherwise we're reducing our coverage accidentally. Supporting more - like Android and iOS - is nice but not a must-have as far as I can tell.

Tokio [platform support](https://docs.rs/tokio/latest/tokio/#supported-platforms) doesn't vary by architecture, only OS, so it seems that they consider an issue on some arch that they cover the OS for to be a bug. Doesn't mean there won't be any, but at least it would be something they consider important? 

[MIO platform](https://crates.io/crates/mio#platforms) support is the thing that really dominates tokio support. 

So comparing:

| platform | Tier | Tokio | Mio |
+----------+----+-----+-----+
| Linux | 1 | Yes | Yes |
| Windows | 1 | Yes | Yes |
| macOS | 1 | Yes | Yes |
| FreeBSD | 2+ | Yes | Yes |
| Illumos | 2+ | Mio | [Yes](https://github.com/tokio-rs/mio/blob/0757ec88387035eaa1bc723993fc3405f218fb7b/Makefile#L2) | 
| NetBSD | 2+ | Mio | [Yes](https://github.com/tokio-rs/mio/blob/0757ec88387035eaa1bc723993fc3405f218fb7b/Makefile#L2) |
| iOS | 2 | Mio | [Yes](https://github.com/tokio-rs/mio/blob/0757ec88387035eaa1bc723993fc3405f218fb7b/Makefile#L2) |
| fuchsia | 2 | Mio | No |
| android API 21 |  2 | Yes | Yes |
| arm-none-* | 2 | Mio | [No](https://github.com/tokio-rs/mio/blob/0757ec88387035eaa1bc723993fc3405f218fb7b/Makefile#L2) |
| WASM | 2 | Mio | [No](https://github.com/tokio-rs/mio/blob/0757ec88387035eaa1bc723993fc3405f218fb7b/Makefile#L2) |
| Solaris | 2 | Mio | No |
...

I think this is pretty safe, but we could do an advance test where we only introduce the runtime don't use it, and see who screams.

src/cli/common.rs
```diff
@@ -382,11 +384,11 @@ pub(crate) fn list_targets(distributable: DistributableToolchain<'_>) -> Result<
                 .as_ref()
                 .expect("rust-std should have a target");
             if component.installed {
-                let _ = t.attr(term2::Attr::Bold);
-                let _ = writeln!(t, "{target} (installed)");
+                let _ = t.attr(terminalsource::Attr::Bold);
+                let _ = writeln!(t.lock(), "{target} (installed)");
```
rbtcollins, line 393: A note for future investigation .... is the `let _ =` needed still? I presume its discarding the fallible from the call, but we're in a function that returns Result, so `?` should be usable.

src/cli/common.rs
```diff
@@ -397,7 +399,9 @@ pub(crate) fn list_targets(distributable: DistributableToolchain<'_>) -> Result<
 pub(crate) fn list_installed_targets(
     distributable: DistributableToolchain<'_>,
 ) -> Result<utils::ExitCode> {
-    let mut t = term2::stdout();
+    let t = process().stdout();
+    let mut t_lock = t.lock();
```
rbtcollins: We should probably not lock around non-console IO; instead locking just before the for loop.

src/cli/download_tracker.rs
```diff
@@ -81,7 +74,7 @@ impl DownloadTracker {
                 true
             }
             Notification::Install(In::Utils(Un::DownloadDataReceived(data))) => {
-                if tty::stdout_isatty() {
+                if process().is_stdout_a_tty() {
```
rbtcollins: We could also spell this `process().stdout().is_a_tty()`. I don't know that it makes much difference. What do you think?

src/currentprocess/terminalsource.rs
```diff
@@ -0,0 +1,231 @@
...
+    fn bg(&mut self, color: Color) -> io::Result<()>;
+    fn attr(&mut self, attr: Attr) -> io::Result<()>;
+    fn reset(&mut self) -> io::Result<()>;
+    fn carriage_return(&mut self) -> io::Result<()>;
+}
+
+// -------------- stdout -------------------------------
+
+/// Stand-in for std::io::stdout.
+pub trait StdoutTerminalSource {
```
rbtcollins: I don't understand why this moves across from filesource.

We have outstanding feature requests for an API (e.g. JSON output on stdout, maybe even an RPC style thing), where we'd want to be operating entirely separately from the terminal layer.

Is it possible to do this:
filesource - unchanged vs master
terminalsource - adds in a terminal() method ?

code that previously used stdout doesn't change
code that previously used term2, uses process().terminal()

?

src/currentprocess/terminalsource.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    fn reset(&mut self) -> io::Result<()> {
+        if !self.stream.supports_color() {
+            return Ok(());
+        }
+        self.stream.reset()
+    }
+
+    fn carriage_return(&mut self) -> io::Result<()> {
```
rbtcollins: I think we should drop this method. It existed because the other terminal abstraction we used provided it, but if termcolor just wants a \r in the stream, we can just make our code do that directly. 

This can be done in a later PR to keep this one focused on the like-for-like exchange.

src/currentprocess/terminalsource.rs
```diff
@@ -0,0 +1,231 @@
...
+    };
+    StandardStreamTerminal {
+        stream: termcolor::StandardStream::stdout(choice),
+        color: ColorSpec::new(),
+    }
+}
+
+/// Returns a `StandardStreamTerminal` for stderr. If stderr is not a tty, then
+/// the returned `StandardStreamTerminal` will not support coloring.
+fn stderr() -> StandardStreamTerminal {
```
rbtcollins: Genuine question - do we want a terminal abstraction on stderr? If so then for my suggestion above about `.terminal()` we could add `.terminal()` to the Stdout / Stderr traits. 

So it would look like

`process.stderr().terminal()`

src/currentprocess.rs
```diff
@@ -195,17 +199,45 @@ pub trait ProcessSource {
...
+    fn is_stdout_a_tty(&self) -> bool;
 }
 
 // ----------- real process -----------------
 
-#[derive(Clone, Debug, Default)]
-pub struct OSProcess {}
+#[derive(Clone, Debug)]
+pub struct OSProcess {
+    is_stderr_a_tty: bool,
```
rbtcollins: These clearly have to be here to be cached, but I think we can make the accessing functions live on Writer.

Then e.g.

```
impl Writer for io::Stdout {
    ...
    fn is_a_tty(&self) -> bool {
        process().is_stdout_a_tty()
    }
```

... except that that won't work...
... except that it *will* if we merge PR 3368 first.

This can be done as an evolution on top of your PR of course, no need to hold it up.

Cargo.toml
```diff
@@ -79,7 +79,7 @@ sharded-slab = "0.1.1"
 strsim = "0.10"
 tar = "0.4.26"
 tempfile.workspace = true
-term = "=0.5.1" # FIXME(issue #1818, #1826, and friends)
+termcolor = "1.2"
```
rbtcollins: please add this to the 'workspace dependencies' table below, and this line here should be
```
termcolor.workspace = true
```

This will help us keep a single version of the dependency in our dependencies even if we refactor rustup in future.

rbtcollins (commented):
This is looking pretty good.

I have left a number of comments.

There is one place in particular that I would like to see different which is the move of stdout/stderr into the terminal responsibility. At least in my head the streams and the terminal abstraction over them are really quite different. Having the terminal in currentprocess is required because of the termcolor design, so this is looking much nicer from that perspective, but I'd still like to see it layered, so that we can have terminal-free stdout when needed.

src/cli/common.rs
0xPoe: I will take a look later.

src/currentprocess/terminalsource.rs
0xPoe: I changed it to:
1. `process().stdout()`
2. `process().stdout_terminal()`

src/currentprocess/terminalsource.rs
0xPoe: I will change it in my next PR.

src/currentprocess/terminalsource.rs
0xPoe: I guess there is no good way to do that. Because the stderr is a writer and the terminal is also a writer, it does not make sense to implement a terminal for a writer.

rbtcollins: I think there are two separate discussions here that perhaps we're not connecting on.

One is what termcolor requires: we agree that it has its own definition of stdout/stderr etc and can't layer on a provided struct/trait.

The other is what we offer *within* rustup. 
A) Do we supply a termcolor terminal (even wrapped by a struct) to code that is doing low-level operations or
B) do we supply a low level struct with less features to such code, and a termcolor struct (wrapped sure) to higher level code

I think B is better. The lock() feature will prevent cross talk between these layers (and we should also do more like making progress bars run on stderr, but e.g. JSON output on stdout).

Cargo.toml
```diff
@@ -182,6 +182,7 @@ opentelemetry-otlp = { version = "0.11.0" }
 proptest = "1.1.0"
 rustup-macros = { path = "rustup-macros" }
 tempfile = "3.5"
+termcolor.workspace = true
```
rbtcollins: this and the other change in this file are in the reverse spot from each other.

`workspace.dependencies` needs to specify a version
`dependencies` needs to specify that it uses the workspace version.

0xPoe: Damn! 😢

0xPoe:
Test report:

1. On MacOS vscode zsh:
image

2. On MacOS  Warp:
image

3. On Windows vscode powershell:

image

4. On Windows vscode MINGW64:
image

5. On Linux vscode bash:
image

0xPoe (commented):
Self check

0xPoe:
@rbtcollins Thanks for your review and help!  💚 💙 💜 💛 ❤️

PR #3354, opened 2023-05-10
fix(rustup): use warning if noprompt

arazabishov:
Downgrade the log level from error to warning on home dir mismatch when opting in for noprompt (-y).

### DIFF
--- a/src/cli/self_update/unix.rs
+++ b/src/cli/self_update/unix.rs
@@ -34,9 +34,21 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool) -> Result<utils::ExitCode> {
     match home_mismatch() {
         (false, _, _) => {}
         (true, env_home, euid_home) => {
-            err!("$HOME differs from euid-obtained home directory: you may be using sudo");
-            err!("$HOME directory: {}", env_home.display());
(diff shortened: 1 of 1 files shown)


rbtcollins:
Can I ask what prompted the PR? What problem are you solving by making this change?

arazabishov:
Sure, apologies for the lack of context. When installing dependencies in Vercel's environment, this error is picked up by tooling and bubbled up in UI despite opting in for silent installation. This does not fail the build, but there is also not much we (users of Vercel) can do about Vercel's environment/container configuration. 

Generally, if there is an error I would expect the whole thing to fail, which it does if no -y flag is passed. However, if -y is passed and the installation process is not failing, it becomes somewhat misleading. 

If you believe this PR is not useful, please feel free to close it :)

rbtcollins:
So I see the concern. That said, two things come to mind.

1) If you're root, you have the ability to make a new user and sudo to that user. So you could fix this from your side with little performance impact.

2) have you opened a ticket on Vercel? They might be happy to fix it if they are aware of it?

I'm hesitant to downgrade the warning, because we really do want this one to be seen. Its true that noprompt is focused on automation, but the emergent problem can still bite folk, and making sure they are aware to facilitate debugging is important.

PR #3360, opened 2023-05-21
`rustup show` shouldn't install things

### LINKED ISSUE #1397: `rustup show` should not force-install the default toolchain if it is not installed.

john-h-k:
Fixes #1397

Currently `rustup show` will install missing components or missing toolchains when ran. This has a couple of issues:
* It's unintuitive - I don't think anyone expects a `show` command to install things
* It causes lots of programs to accidentally cause installations
  * Lots of shell prompts use `rustup show` to show current toolchain and similar. This means `cd`'ing into a rust project with an uninstalled toolchain can cause the shell to freeze while it installs the entire toolchain

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -660,6 +660,22 @@ impl Cfg {
         &self,
         path: &Path,
     ) -> Result<(Toolchain<'_>, Option<OverrideReason>)> {
+        self.find_toolchain(path, true)
+    }
+
+    #[cfg_attr(feature = "otel", tracing::instrument(skip_all))]
+    pub(crate) fn find_override_toolchain_or_default(
+        &self,
+        path: &Path,
+    ) -> Result<(Toolchain<'_>, Option<OverrideReason>)> {
+        self.find_toolchain(path, false)
+    }
+
+    fn find_toolchain(
+        &self,
+        path: &Path,
+        install_if_missing: bool,
+    ) -> Result<(Toolchain<'_>, Option<OverrideReason>)> {
         let (toolchain, components, targets, reason, profile) =
             match self.find_override_config(path)? {
                 Some((
@@ -688,7 +704,7 @@ impl Cfg {
                 let components: Vec<_> = components.iter().map(AsRef::as_ref).collect();
                 let targets: Vec<_> = targets.iter().map(AsRef::as_ref).collect();
                 let toolchain = match DistributableToolchain::new(self, desc.clone()) {
-                    Err(RustupError::ToolchainNotInstalled(_)) => {
+                    Err(RustupError::ToolchainNotInstalled(_)) if install_if_missing => {
                         DistributableToolchain::install(
                             self,
                             &desc,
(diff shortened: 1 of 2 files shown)


rbtcollins:
Please see the history in that bug, from around https://github.com/rust-lang/rustup/issues/1397#issuecomment-782710497

In particular, this is a breaking change, we rather suspect we need an RFC, though rustup hasn't been doing that process often so I'd expect some friction there.

Regardless of the RFC thing though, we have to have the replacement UX present before we contract the functionality.

We have rough consensus in that bug that:
 - we should add an ensure verb that autoinstall the current toolchain name. E.g. `rustup +stable ensure` will install stable, `rustup ensure` will install the default or otherwise overridden toolchain, etc. Possibly that should be `rustup toolchain ensure` or some other spelling. This is a new command that doesn't exist yet.
 - we then then stop autoinstalling on anything outside of proxy-mode

To make this change will require a decent time investment. Adding the new command is quite easy; then you have to follow that up by migrating the test suite to the new command. And then finally we can remote autoinstalling for the non-proxy-mode commands.

I think that that last step should be staggered over at least 2 release cycles and coordinated with a rust-lang blog post.

src/config.rs
```diff
@@ -659,6 +659,22 @@ impl Cfg {
...
+    ) -> Result<(Toolchain<'_>, Option<OverrideReason>)> {
+        self.find_toolchain(path, true)
+    }
+
+    #[cfg_attr(feature = "otel", tracing::instrument(skip_all))]
+    pub(crate) fn find_override_toolchain_or_default(
+        &self,
+        path: &Path,
+    ) -> Result<(Toolchain<'_>, Option<OverrideReason>)> {
+        self.find_toolchain(path, false)
```
rbtcollins, line 671: this is just find_toolchain with a create option of false. I don't think the new function is worth it.

rbtcollins:
I'm going to close this PR as it doesn't seem to be getting updated; please do reopen it if you start work on the suggestions in https://github.com/rust-lang/rustup/pull/3360#issuecomment-1556239862

PR #3363, opened 2023-05-25
CI support for loongarch64-unknown-linux-gnu

heiher:
This PR adds CI support for `loongarch64-unknown-linux-gnu` target and skip stable until rust 1.71 is released.


rbtcollins (changes requested):
I don't think we want to take a dependency on nightly. Once loongarch64 builds reliably on stable we can just add it as another platform we validate.

PR #3364, opened 2023-05-25
Fix compile on rust nightly

heiher:
This PR fixes compile on rust nightly (1.71.0-nightly) with non-default features:

```
--no-default-features --features curl-backend,reqwest-backend,reqwest-default-tls --features vendored-openssl
```

### Compile errors

```
error: cannot find macro `lazy_static` in this scope
   --> download/src/lib.rs:345:5
    |
345 |     lazy_static! {
    |     ^^^^^^^^^^^
    |
    = help: consider importing this macro:
            lazy_static::lazy_static
    = note: `lazy_static` is in scope, but it is a crate, not a macro

error[E0425]: cannot find value `CLIENT_DEFAULT_TLS` in this scope
   --> download/src/lib.rs:379:37
    |
379 |             TlsBackend::Default => &CLIENT_DEFAULT_TLS,
    |                                     ^^^^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
```

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -269,7 +269,7 @@ pub mod reqwest_be {
     use std::time::Duration;
 
     use anyhow::{anyhow, Context, Result};
-    #[cfg(feature = "reqwest-rustls-tls")]
+    #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-default-tls"))]
     use lazy_static::lazy_static;
     use reqwest::blocking::{Client, ClientBuilder, Response};
     use reqwest::{header, Proxy};


xen0n (commented):
In my own experiments this only happens with the `reqwest-rustls-tls` feature disabled but `reqwest-backend` and `reqwest-default-tls` enabled. Given the context of your work that's `loongarch64` enablement, maybe mentioning the intended use case (no `ring` yet, so no `reqwest-rustls-tls`) would further justify the change.

download/src/lib.rs
```diff
@@ -269,8 +269,6 @@ pub mod reqwest_be {
     use std::time::Duration;
 
     use anyhow::{anyhow, Context, Result};
-    #[cfg(feature = "reqwest-rustls-tls")]
```
rbtcollins, line 272: Please just put a `any` in here with the additional feature that requires rustls.

rbtcollins (changes requested):

PR #3366, opened 2023-05-26
Improve error message for adding/removing invalid target

### LINKED ISSUE #3318: Improve error message for `rustup target add invalid-target`

Consider this error:
```
; rustup target add wasm
error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not contain component 'rust-std' for target 'wasm'
```
The pointer to platform-support.html is good and helpful, but the rest of these messages are a bit misleading. The problem is that `wasm` isn't a target, I need to pick a specific version of wasm:
```
; rustc --print=target-list | grep wasm
wasm32-unknown-emscripten
wasm32-unknown-unknown
wasm32-wasi
wasm64-unknown-unknown
```

### Solution you'd like

Have `rustup` check the output of `rustc --print=target-list` for the specified toolchain and give a more helpful error if it doesn't have the target. Maybe something like this:
```
; rustup target add wasm
error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not support target 'wasm'
note: you can see a list of supported targets with `rustc --print=target-list`
note: if you are adding support for a new target to rustc itself, see https://rustc-dev-guide.rust-lang.org/building/new-target.html
```

### Notes

_No response_

0xPoe:
close https://github.com/rust-lang/rustup/issues/3318

Check the target first.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -130,6 +130,17 @@ pub(crate) enum RustupError {
         target: TargetTriple,
         suggestion: Option<String>,
     },
+    #[error("toolchain '{}' does not have target '{}' installed{}\n", .desc, .target,
+    if let Some(suggestion) = .suggestion {
+        format!("; did you mean '{suggestion}'?")
+    } else {
+        "".to_string()
+    })]
+    TargetNotInstalled {
+        desc: ToolchainDesc,
+        target: TargetTriple,
+        suggestion: Option<String>,
+    },
     #[error("unknown metadata version: '{0}'")]
     UnknownMetadataVersion(String),
     #[error("manifest version '{0}' is not supported")]
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -10,7 +10,10 @@ use std::path::PathBuf;
 use thiserror::Error as ThisError;
 use url::Url;
 
-use crate::{currentprocess::process, dist::dist::ToolchainDesc};
+use crate::{
+    currentprocess::process,
+    dist::dist::{TargetTriple, ToolchainDesc},
+};
 use crate::{
     dist::manifest::{Component, Manifest},
     toolchain::names::{PathBasedToolchainName, ToolchainName},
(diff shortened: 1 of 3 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1377,7 +1377,7 @@ fn get_target(m: &ArgMatches, distributable: &DistributableToolchain<'_>) -> Opt
 fn component_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let toolchain = explicit_or_dir_toolchain(cfg, m)?;
     let distributable = DistributableToolchain::try_from(&toolchain)?;
-    let target = get_target(m, &distributable);
+    let target: Option<TargetTriple> = get_target(m, &distributable);
```
rbtcollins: What requires the type annotation?

src/errors.rs
```diff
@@ -114,6 +117,14 @@ pub(crate) enum RustupError {
         component: String,
         suggestion: Option<String>,
     },
+    #[error("toolchain '{}' does not support target '{}'{}", .desc, .target,
+    "\nnote: you can see a list of supported targets with `rustc --print=target-list`\n\
+    note: if you are adding support for a new target to rustc itself, see https://rustc-dev-guide.rust-lang.org/building/new-target.html
+    ".to_string())]
```
rbtcollins: Using {} to add in a constant string feels weird. Any reason we can't just put it in the main format string? (perhaps using a multi-line string for easier reading)?

rbtcollins (commented):
this looks broadly fine

src/cli/rustup_mode.rs
0xPoe: I guess my RA got some issues. Reverted.

src/errors.rs
0xPoe: Fixed. Thanks! Make sense.

src/errors.rs
```diff
@@ -114,6 +117,30 @@ pub(crate) enum RustupError {
...
+    } else {
+        "".to_string()
+    })]
+    UnknownTarget {
+        desc: ToolchainDesc,
+        target: TargetTriple,
+        suggestion: Option<String>,
+    },
+    #[error("toolchain '{}' does not have target '{}' installed{}\n", .desc, .target,
+    if let Some(suggestion) = .suggestion {
```
rbtcollins, line 134: This is duplicated. What about a helper? `suggest_message(.suggestion)` ?

rbtcollins (approved):

rbtcollins:
I'm merging this, but have left a suggestion to reduce code duplication

jyn514:
this is awesome, thank you both ❤️❤️

PR #3367, opened 2023-05-26
Switch from sync to a tokio runtime

rbtcollins:
Being able to use async where relevant across the codebase makes it possible to use async constructs where relevant. rustup has many places where that can be useful: downloading channel metadata, distribution content, unpacking that content to disk, are work that could usefully proceed in parallel - and async provides a good abstraction for that.

We already have a sophisticated disk IO layer that accomodates various OS latency-inducing behaviours, and adapting that to async without any regressions could be very interesting too - but for now, it co-exists nicely with an async core.

One not necessarily obvious caveat - we can't use tokio::main because it doesn't provide the per-worker-thread injection that we use to facilitate isolated tests. (tl;dr std::env::var/args being global reflect reality well, but not writing tests that interact with those layers).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,6 @@ otel = [
     "dep:tracing-subscriber",
     "dep:opentelemetry",
     "dep:opentelemetry_sdk",
-    "dep:tokio",
-    "dep:tracing",
 ]
 
 # Exports code dependent on private interfaces for the integration test suite
@@ -85,11 +83,11 @@ tempfile.workspace = true
 termcolor.workspace = true
 thiserror.workspace = true
 threadpool = "1"
-tokio = { workspace = true, optional = true }
+tokio.workspace = true
 toml = "0.8"
 tracing-opentelemetry = { workspace = true, optional = true }
 tracing-subscriber = { workspace = true, optional = true, features = ["env-filter"] }
-tracing = { workspace = true, optional = true }
+tracing.workspace = true
 url.workspace = true
 wait-timeout = "0.2"
 walkdir = { workspace = true, optional = true }
@@ -145,7 +143,8 @@ rustup-macros = { path = "rustup-macros" }
 tempfile = "3.8"
 termcolor = "1.2"
 thiserror = "1.0"
-tokio = { version = "1.26.0", default-features = false, features = ["rt-multi-thread"] }
+tokio = { version = "1.26.0", default-features = false, features = ["macros", "rt-multi-thread"] }
+tokio-stream = { version = "0.1.14" }
 tracing = "0.1"
 tracing-opentelemetry = "0.24"
 tracing-subscriber = "0.3.16"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,6 +83,7 @@ tempfile.workspace = true
 termcolor.workspace = true
 thiserror.workspace = true
 threadpool = "1"
+tokio-retry.workspace = true
 tokio.workspace = true
 toml = "0.8"
 tracing-opentelemetry = { workspace = true, optional = true }
@@ -144,6 +145,7 @@ tempfile = "3.8"
 termcolor = "1.2"
 thiserror = "1.0"
 tokio = { version = "1.26.0", default-features = false, features = ["macros", "rt-multi-thread"] }
+tokio-retry = { version = "0.3.0" }
 tokio-stream = { version = "0.1.14" }
 tracing = "0.1"
 tracing-opentelemetry = "0.24"
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -50,7 +50,7 @@ pub enum Event<'a> {
 
 type DownloadCallback<'a> = &'a dyn Fn(Event<'_>) -> Result<()>;
 
-fn download_with_backend(
+async fn download_with_backend(
     backend: Backend,
     url: &Url,
     resume_from: u64,
@@ -58,26 +58,7 @@ fn download_with_backend(
 ) -> Result<()> {
     match backend {
         Backend::Curl => curl::download(url, resume_from, callback),
-        Backend::Reqwest(tls) => {
-            // reqwest is async; this function is sync.
-            match Handle::try_current() {
-                Ok(current) => {
-                    // hide the asyncness for now.
-                    task::block_in_place(|| {
-                        current.block_on(reqwest_be::download(url, resume_from, callback, tls))
-                    })
-                }
-                Err(_) => {
-                    // Make a runtime to hide the asyncness.
-                    tokio::runtime::Runtime::new()?.block_on(reqwest_be::download(
-                        url,
-                        resume_from,
-                        callback,
-                        tls,
-                    ))
-                }
-            }
-        }
+        Backend::Reqwest(tls) => reqwest_be::download(url, resume_from, callback, tls).await,
     }
 }
 
@@ -145,17 +126,43 @@ pub fn download_to_path_with_backend(
 
         let file = RefCell::new(file);
 
-        download_with_backend(backend, url, resume_from, &|event| {
-            if let Event::DownloadDataReceived(data) = event {
-                file.borrow_mut()
-                    .write_all(data)
-                    .context("unable to write download to disk")?;
+        match Handle::try_current() {
+            Ok(current) => {
+                // hide the asyncness for now.
+                task::block_in_place(|| {
+                    current.block_on(download_with_backend(backend, url, resume_from, &|event| {
+                        if let Event::DownloadDataReceived(data) = event {
+                            file.borrow_mut()
+                                .write_all(data)
+                                .context("unable to write download to disk")?;
+                        }
+                        match callback {
+                            Some(cb) => cb(event),
+                            None => Ok(()),
+                        }
+                    }))
+                })
             }
-            match callback {
-                Some(cb) => cb(event),
-                None => Ok(()),
+            Err(_) => {
+                // Make a runtime to hide the asyncness.
+                tokio::runtime::Runtime::new()?.block_on(download_with_backend(
+                    backend,
+                    url,
+                    resume_from,
+                    &|event| {
+                        if let Event::DownloadDataReceived(data) = event {
+                            file.borrow_mut()
+                                .write_all(data)
+                                .context("unable to write download to disk")?;
+                        }
+                        match callback {
+                            Some(cb) => cb(event),
+                            None => Ok(()),
+                        }
+                    },
+                ))
             }
-        })?;
+        }?;
 
         file.borrow_mut()
             .sync_data()
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 
 use anyhow::Context;
 pub use anyhow::Result;
+use tokio::{runtime::Handle, task};
 use url::Url;
 
 mod errors;
@@ -57,7 +58,26 @@ fn download_with_backend(
 ) -> Result<()> {
     match backend {
         Backend::Curl => curl::download(url, resume_from, callback),
-        Backend::Reqwest(tls) => reqwest_be::download(url, resume_from, callback, tls),
+        Backend::Reqwest(tls) => {
+            // reqwest is async; this function is sync.
+            match Handle::try_current() {
+                Ok(current) => {
+                    // hide the asyncness for now.
+                    task::block_in_place(|| {
+                        current.block_on(reqwest_be::download(url, resume_from, callback, tls))
+                    })
+                }
+                Err(_) => {
+                    // Make a runtime to hide the asyncness.
+                    tokio::runtime::Runtime::new()?.block_on(reqwest_be::download(
+                        url,
+                        resume_from,
+                        callback,
+                        tls,
+                    ))
+                }
+            }
+        }
     }
 }
 
@@ -285,15 +305,15 @@ pub mod reqwest_be {
     use anyhow::{anyhow, Context, Result};
     #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-default-tls"))]
     use once_cell::sync::Lazy;
-    use reqwest::blocking::{Client, ClientBuilder, Response};
-    use reqwest::{header, Proxy};
+    use reqwest::{header, Client, ClientBuilder, Proxy, Response};
+    use tokio_stream::StreamExt;
     use url::Url;
 
     use super::Event;
     use super::TlsBackend;
     use crate::errors::*;
 
-    pub fn download(
+    pub async fn download(
         url: &Url,
         resume_from: u64,
         callback: &dyn Fn(Event<'_>) -> Result<()>,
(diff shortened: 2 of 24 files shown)


djc:
Having briefly skimmed over some parts of this: what limits us from just using `#[tokio::main]` and making things natively async, rather than all the callback stuff and `block_on()` type calls?

rbtcollins:
> Having briefly skimmed over some parts of this: what limits us from just using `#[tokio::main]` and making things natively async, rather than all the callback stuff and `block_on()` type calls?

This is in `draft` mode at the moment because it is incomplete.

The block-on stuff is scaffolding during the transition.

The starting point was:
sync main()
...
reqwest blocking http client
tokio runtime invoked here

With an async main, that will blow up at runtime, because you get
tokio runtime
async main()
...
reqwest blocking http client
tokio runtime invoked here - panic

If that is changed naively by just making main() async, and reqwest async, then we get rust's async function tainting everything and will change all the relevant calls to async, **but** there are a number of places where types need to change to be compatible with async requirements; so the commit becomes huge and unwieldy very quickly.

So the approach I'm taking is some scaffolding that lets us work this incrementally, and once complete it will just be nice async code.

djc:
Makes sense, thanks for the explanation. Can you give an example of where/how types need to change? I worry that with this approach, you'll need another big set of commits to unwind all the `block_on()` stuff again after switching `main`. Instead, maybe it would be possible to use the top-down approach to isolate the required type changes in their own commits, and then jump directly to full async mode from there by adding `async` keywords all over the place?

rbtcollins:
Once done the aggregate diff should be precisely the changes needed, and I can rebase squash it all away. We did this with the removal of `failure`. Many many small steps, nice outcome :)

djc:
Oh, I guess that makes sense. Other projects I help maintain prefer to rebase-merge so that the small commits make it into main, which can be helpful if you need to chase down a regression later.

rbtcollins:
Thats also valid to do; lets assess when this is at the next inflection point (which will be async native with no thunk, but no refactorings for better behaviour)

rbtcollins:
This is nearly complete, I think its mostly mechanical now, except for the closures in the test suite - there's some lifetime challenges to address there

rbtcollins:
This now complete modulo any refactoring of the commit series we might want to do.

The key notes are:
- we can't use tokio::main because it doesn't provide the per-worker-thread injection that we use to facilitate isolated tests. (tl;dr std::env::var/args being global reflect reality well, but not writing tests that interact with those layers).

Cargo.toml
```diff
@@ -1,14 +1,14 @@
 [package]
-name = "rustup"
-version = "1.26.0"
-edition = "2021"
+build = "build.rs"
```
djc: I'd prefer not to have reorganization of `Cargo.toml` in this PR (for both crates), since this makes it much harder to see actual changes to the contents that are related to the main topic of this PR.

(Unless we move all of the actual `async` changes into one commit that can be reviewed independently, but actually it looks like these `Cargo.toml` changes are a step in the wrong direction IMO.)

Cargo.toml
```diff
@@ -159,16 +162,24 @@ anyhow = "1.0.69"
 derivative = "2.2.0"
 enum_dispatch = "0.3.11"
 fs_at = "0.1.6"
+futures = "0.3"
+lazy_static = "1"
```
djc: AFAIK lazy_static has been deprecated in favor of once_cell (which we already have as a dependency).

Also it looks like lazy_static doesn't actually get used? I guess the `Cargo.toml` changes are mostly some rebase artifact?

download/src/lib.rs
```diff
@@ -61,74 +61,81 @@ pub fn download_to_path_with_backend(
...
+                    0
+                };
 
-            possible_partial.seek(SeekFrom::End(0))?;
-
-            (possible_partial, downloaded_so_far)
-        } else {
-            (
-                OpenOptions::new()
+                // TODO: blocking call
```
djc: Should this be addressed before merging?

download/src/lib.rs
```diff
@@ -61,74 +61,81 @@ pub fn download_to_path_with_backend(
...
+            };
 
-        file.borrow_mut()
-            .sync_data()
-            .context("unable to sync download to disk")?;
+            let file = RefCell::new(file);
 
-        Ok(())
-    }()
+            // TODO: the sync callback will stall the async runtime if IO calls block, which is OS dependent. Rearrange.
```
djc: Should this be addressed before merging?

src/bin/rustup-init.rs
```diff
@@ -26,29 +26,36 @@ use rustup::cli::rustup_mode;
...
 
-    let process = OSProcess::default();
-    with(process.into(), || match maybe_trace_rustup() {
-        Err(e) => {
-            common::report_error(&e);
-            std::process::exit(1);
+    let process = OSProcess::new();
+    let mut builder = Builder::new_multi_thread();
+    builder.enable_all();
+    with_runtime(process.into(), builder, {
```
djc, line 42: Do we need the outer block here?

src/bin/rustup-init.rs
```diff
@@ -26,29 +26,36 @@ use rustup::cli::rustup_mode;
...
 fn main() {
     #[cfg(windows)]
     pre_rustup_main_init();
 
-    let process = OSProcess::default();
-    with(process.into(), || match maybe_trace_rustup() {
-        Err(e) => {
-            common::report_error(&e);
-            std::process::exit(1);
+    let process = OSProcess::new();
```
djc: Why the change from `OSProcess::default()` to `OSProcess::new()` here? That seems unrelated.

src/cli/proxy_mode.rs
```diff
@@ -12,7 +12,7 @@ use crate::{
 };
 
 #[cfg_attr(feature = "otel", tracing::instrument)]
-pub fn main(arg0: &str) -> Result<ExitCode> {
+pub async fn main(arg0: &str) -> Result<ExitCode> {
```
djc: Have you checked the performance impact of these changes on proxy invocations?

src/cli/rustup_mode.rs
```diff
@@ -84,8 +85,33 @@ where
     callee(cfg, matches)
 }
 
+async fn deprecated_async<'a, FF, F, B, R>(
+    instead: &str,
+    cfg: &'a mut Cfg,
+    matches: B,
+    callee: FF,
```
djc: I would suggest writing this as `impl FnOnce(&'a mut Cfg, B) -> impl Future<Output = R> + 'a`, which seems easier to understand to me.

src/cli/rustup_mode.rs
```diff
@@ -84,8 +85,33 @@ where
...
+    ));
+    (cfg.notify_handler)(Notification::PlainVerboseMessage(
+        "The exact API of this command may change without warning",
+    ));
+    (cfg.notify_handler)(Notification::PlainVerboseMessage(
+        "Eventually this command will be a true alias.  Until then:",
+    ));
+    (cfg.notify_handler)(Notification::PlainVerboseMessage(&format!(
+        "  Please use `rustup {instead}` instead"
+    )));
```
djc: Given that this is exactly the same as in sync `deprecated()`, let's deduplicate? Alternatively, is there a good reason not to use the async variant for the `uninstall` command so we don't need both?

src/cli/self_update.rs
```diff
@@ -356,9 +356,10 @@ fn canonical_cargo_home() -> Result<Cow<'static, str>> {
 }
 
 /// Installing is a simple matter of copying the running binary to
-/// `CARGO_HOME`/bin, hard-linking the various Rust tools to it,
-/// and adding `CARGO_HOME`/bin to PATH.
-pub(crate) fn install(
+/// `CARGO_HOME`/bin, hard-linking the various Rust tools to it, and adding
+/// `CARGO_HOME`/bin to PATH by dropping shell scripts around the place, except
+/// on windows, and also we need a linker on Windows...
```
djc: Seems unrelated, can we isolate this change/defer it for now?

src/cli/self_update.rs
```diff
@@ -1074,10 +1080,10 @@ pub(crate) fn uninstall(no_prompt: bool) -> Result<utils::ExitCode> {
 /// (and on windows this process will not be running to do it),
 /// rustup-init is stored in `CARGO_HOME`/bin, and then deleted next
 /// time rustup runs.
-pub(crate) fn update(cfg: &Cfg) -> Result<utils::ExitCode> {
-    common::warn_if_host_is_emulated();
-
+pub(crate) async fn update(cfg: &Cfg) -> Result<utils::ExitCode> {
     use common::SelfUpdatePermission::*;
+
+    common::warn_if_host_is_emulated();
```
djc, line 1085: For a large change like this, I'd prefer not to make unrelated changes.

src/cli/setup_mode.rs
```diff
@@ -1,5 +1,7 @@
 use anyhow::Result;
-use clap::{builder::PossibleValuesParser, value_parser, Arg, ArgAction, Command};
+use clap::{
+    builder::PossibleValuesParser, error::ErrorKind, value_parser, Arg, ArgAction, Command,
```
djc: This too seems unrelated.

src/currentprocess.rs
```diff
@@ -139,20 +140,89 @@ pub fn with<F, R>(process: Process, f: F) -> R
...
-    PROCESS.with(|p| {
+/// Run a function in the context of a process definition and a tokio runtime.
+///
+/// The process state is injected into a thread-local in every work thread of
+/// the runtime, but this requires access to the runtime builder, so this
+/// function must be the one to create the runtime.
+pub fn with_runtime<'a, FR, R>(
+    process: Process,
+    mut runtime_builder: tokio::runtime::Builder,
+    fut: FR,
```
djc: Suggest writing this as `impl Future<Output = R> + 'a`.

src/currentprocess.rs
```diff
@@ -139,20 +140,89 @@ pub fn with<F, R>(process: Process, f: F) -> R
...
+) -> R
+where
+    FR: Future<Output = R> + 'a,
+{
+    ensure_hook();
+
+    let start_process = process.clone();
+    let unpark_process = process.clone();
+    let runtime = runtime_builder
+        // propogate to blocking threads
```
djc: Typo: "propogate" -> "propagate" (also below).

djc (commented):
This looks mostly okay to me.

src/bin/rustup-init.rs
rami3l: @djc I believe this is just a `rustfmt` specificity.

rami3l (approved):

Cargo.toml
rbtcollins: I can back that out, its a side effect of the vscode even better toml plugin. The actual dep changes of course are needed.

Cargo.toml
rbtcollins: https://github.com/rust-lang/rustup/pull/3367#discussion_r1560042952

download/src/lib.rs
rbtcollins: oh, I missed that I had these TODOs left. I will review.

src/cli/proxy_mode.rs
rbtcollins: Not rigorously, but I know that most of our proxy overhead is in manifest parsing today, so I doubt that we'd get a measurable change even if the runtime initialisation were a concern.

download/src/lib.rs
rbtcollins: I think this is fine a as blocking call for now : we download a small number of files during install, and it is very unlikely that this is going to cause a performance problem. Eventually of course we should do it, but the goal here was to get a tokio runtime in place at parity, not to make perf/concurrency improvements, since those will require reasoning about resource limits and so on in a very careful fashion.

download/src/lib.rs
rbtcollins: No. Same as above, we should do so eventually.

The cb passed in does two things:
 - write a block of the file, which can/will block
 - calls the next layer out callback, which will be our download rate tracker

As we only download one file at the moment, there is no performance concern about the IO blocking - a rearranged async callback would want to prevent more data being downloaded to avoid OOM if disk IO were to stall (e.g. on NFS); thus this is appropriate for now, and when we make it async, it will be in order of supporting concurrent download of files, which is a different feature that we can all bring together

src/bin/rustup-init.rs
rbtcollins: It was the elevation of stderr_is_a_ttry into the OSProcess struct, that meant that either a manual Default implementation was needed, or new(), and I have no idea why I decided to do both :).

I made stderr_is_a_tty state to resurrect a previous efficiency optimisation we had for Windows, so that tty probing only happens one per process, not once per terminal construction. It really should be a separate patch; and will need redoing with your other refactoring.

src/bin/rustup-init.rs
rbtcollins: rustfmt insisted on it.

src/cli/self_update.rs
rbtcollins: sorry, what unrelated change?

src/cli/setup_mode.rs
rbtcollins: error::ErrorKind is moved from within main() to a file level import. I don't recall if I had to do that do deal with a transient state during the refactoring, or whether I could move it back now. That said, I usually prefer top of module imports, so we can break if out if you like, but I'm not super excited by it :)

src/currentprocess.rs
rbtcollins: fixed

Cargo.toml
rbtcollins: lazy_static is because this branch started before 

commit 867dfc7d46e46bc718bb4b947291990bdf4c7e61
Author: Dirkjan Ochtman <dirkjan@ochtman.nl>
Date:   Mon Nov 6 13:41:22 2023 +0100

    Replace usage of lazy_static with once_cell


fixed.

src/currentprocess.rs
rbtcollins: done

src/bin/rustup-init.rs
djc: I'm not sure I understand what you mean here. Seems that `stderr_is_a_tty` is already state in `OSProcess` on `master`, and `default()` and `new()` are interchangeable, so I'd prefer to remove this change from this PR.

src/cli/self_update.rs
djc: Changing the order of calling `warn_if_host_is_emulated()` vs importing from `SelfUpdatePermission`.

src/cli/setup_mode.rs
djc: I prefer top of module imports too, but when reviewing a large change I prefer for these not to be intertwined with the other changes.

src/cli/self_update.rs
rbtcollins: ah yes. Do you want me to back it out?

djc: Yes, I think that is preferrable.

djc (approved):

djc:
Have rebased this on current master and addressed my own comments (pretty much all about unrelated changes). I think this is good to go. @rami3l do you want to take another look? I just went through and compared each commit to the original PR in its state before I touched it and the differences are pretty minimal.

src/cli/self_update.rs
rami3l: @djc Can we mark this as resolved now?

rami3l:
Thanks a lot for your hard work, @rbtcollins and @djc!

PR #3368, opened 2023-05-27
Improve CurrentProcess

rbtcollins:
To bring in tokio runtime pervasively we need to evolve CurrentProcess
to be a bit more friendly with Send/Sync. Getting at the concrete types
is actually clearer than what we had I think, and should ease that
process.


0xPoe (approved):

0xPoe:
Thanks!

PR #3370, opened 2023-05-28
Group updates to opentelemetry together

rbtcollins:
This might not be the exact pattern we need - we can iterate on it
easily.

The Rust opentelemetry ecosystem has tight coupling across versions
between releases, so it is unlikely for single-dependency upgrades to
work.


djc (approved):
Drive-by review: yup, this makes sense to me (as an opentelemetry-rust co-maintainer).

PR #3371, opened 2023-05-30
Make download_tracker thread safe.

rbtcollins:
Not however concurrency safe. This is sufficient to permit introducing
an async runtime throughout rustup, but not sufficient for us to start
tracking many concurrent downloads.


0xPoe (approved):
Thanks!

PR #3379, opened 2023-06-08
Make clippy happy

0xPoe:
See the commits.

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -61,7 +61,7 @@ impl DownloadTracker {
             term: process().stdout().terminal(),
             displayed_charcount: None,
             units: vec![Unit::B],
-            display_progress: display_progress,
+            display_progress,
         }))
     }
 
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -195,7 +195,7 @@ impl UnixShell for Zsh {
         self.rcfiles()
             .into_iter()
             .filter(|env| env.is_file())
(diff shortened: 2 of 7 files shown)


src/cli/common.rs
```diff
@@ -5,7 +5,7 @@ use std::fmt::Display;
 use std::fs;
 use std::io::{BufRead, ErrorKind, Write};
 use std::path::Path;
-use std::sync::Arc;
+use std::rc::Rc;
```
rbtcollins: I think this one will get resolved as we move further into async. I think we should just permit it or ignore it if its not stabilised.

src/config.rs
```diff
@@ -3,8 +3,8 @@ use std::fmt::{self, Display};
 use std::io;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::rc::Rc;
 use std::str::FromStr;
-use std::sync::Arc;
```
rbtcollins: ditto

src/dist/manifestation/tests.rs
```diff
@@ -2079,7 +2080,7 @@ fn reuse_downloaded_file() {
                           temp_cfg| {
         prevent_installation(prefix);
 
-        let reuse_notification_fired = Arc::new(Cell::new(false));
+        let reuse_notification_fired = Rc::new(Cell::new(false));
```
rbtcollins: ditto

src/cli/common.rs
0xPoe: Updated. Make sense.

PR #3380, opened 2023-06-08
Allow rustup doc books open subsections and paragraphs

### LINKED ISSUE #3345: `rustup doc --book` to directly link to subsections and paragraphs of the book

i came up recently with [this case in Rustlings](https://github.com/rust-lang/rustlings/issues/1501), where i wanted the hints in the exercises to point out sections of the book to be cross-referenced when completely offline, but all the hints in those exercises links to online resources instead.

main problem is, someone pointed out there's way too many variables to keep track of to be worth maintained in Rustlings alone, ranging from browser installation checks, to the file URLs which Rustup already handles. so the current options in Rustlings' side as of now may become either make the user setup a Hosts file redirect into the offline book, or omit links altogether and just hint the user to the chapter section and the existence of the `rustup doc` commands, as opposed to having both the links and the commands on all relevant hints.

so my thought with this was, if it was maintained on Rustup's side instead, that'll probably be worth it better since it's gonna become easily accessible outside anyway.

### Solution you'd like

the access would be made for commands like

```bash
$ rustup doc --book chapter-part[-section[-subsection]]
```

for instance,
```bash
$ rustup doc --book 03-05-02-02
#not sure if this would require changing how all the HTML page links and headers are written
```

would ammount to `file:///…/book/ch03-05-control-flow.html#returning-values-from-loops`  
but maybe an easier approach could be to instead index all headers in the book together and make them accessible by name directly like so:

```bash 
#this does the same except it's easier to see where you're going to from the start.
$ rustup doc --book returning-values-from-loops 
```

i think this would enable for easier access to notes and not just the use for the initial learning of the language. what do you think?

### Notes

_No response_

YegorStolyarov:
**Reason:** additional param to doc command, topic, is only supported for rust reference. I tried to implement minimalistic changes to make it work with books as well. 
based on  this issue #3345 

**What does it do:** exact links to offline docs in books.
for example, rustup doc --rustc platform-support/armv6k-nintendo-3ds.html#building-the-target now opens exact chapter and paragraph in rustc book

**Code changes:**
- added logic case for condition (book flag) AND (topic param provided) . now topic param is ignored in case of opening book
- have to add "file:///" to browser url because opener crate which used in rustup have problems to recognize url as file when it contains "#" symbol.

my first PR, thank you for the feedback.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1589,18 +1589,21 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let cached_path: String;
     let mut has_docs_data_link = false;
 
-    let doc_url = if let Some((short, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
-        if let Some(topic) = m.get_one::<String>("topic") {
-            has_docs_data_link = true;
-            cached_path = format!("{}/{}", short, topic);
-            cached_path.as_str()
-        } else { path }
-    } else if let Some(topic) = m.get_one::<String>("topic") {
-        topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
-        topical_path.to_str().unwrap()
-    } else {
-        "index.html"
-    };
+    let doc_url =
+        if let Some((short, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+            if let Some(topic) = m.get_one::<String>("topic") {
+                has_docs_data_link = true;
+                cached_path = format!("{}/{}", short, topic);
+                cached_path.as_str()
+            } else {
+                path
+            }
+        } else if let Some(topic) = m.get_one::<String>("topic") {
+            topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
+            topical_path.to_str().unwrap()
+        } else {
+            "index.html"
+        };
 
     if m.get_flag("path") {
         let doc_path = toolchain.doc_path(doc_url)?;
@@ -1612,7 +1615,6 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
             let url_path = format!("file:///{}", url_path_buf.to_str().unwrap());
             let doc_path = Path::new(&url_path);
             utils::open_browser(doc_path)?
-
         } else {
             toolchain.open_docs(doc_url)?;
         }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1586,12 +1586,18 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     };
 
     let topical_path: PathBuf;
+    let cached_path: String;
+    let mut has_docs_data_link = false;
 
-    let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
+    let doc_url = if let Some((short, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+        if let Some(topic) = m.get_one::<String>("topic") {
+            has_docs_data_link = true;
+            cached_path = format!("{}/{}", short, topic);
+            cached_path.as_str()
+        } else { path }
+    } else if let Some(topic) = m.get_one::<String>("topic") {
         topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
         topical_path.to_str().unwrap()
-    } else if let Some((_, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
-        path
     } else {
         "index.html"
     };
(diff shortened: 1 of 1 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1586,22 +1586,38 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
-
-    let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
-        topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
-        topical_path.to_str().unwrap()
-    } else if let Some((_, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
-        path
-    } else {
-        "index.html"
-    };
+    let cached_path: String;
```
0xPoe, line 1589: Why do we need to define this one here? Maybe we can put it into the `if` statement.

src/cli/rustup_mode.rs
```diff
@@ -1586,22 +1586,38 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
+        };
 
     if m.get_flag("path") {
         let doc_path = toolchain.doc_path(doc_url)?;
         writeln!(process().stdout().lock(), "{}", doc_path.display())?;
         Ok(utils::ExitCode(0))
     } else {
-        toolchain.open_docs(doc_url)?;
+        if has_docs_data_link {
+            let url_path_buf = toolchain.doc_path(doc_url)?;
```
0xPoe, line 1614: I think we can put it into `open_docs`. We don't need to leak it here. And the overall logic is very similar with the `open_docs`.

0xPoe (commented):
Thanks for your contribution.

@rbtcollins Do you have any thoughts about adding this feature?

rbtcollins:
I like the concept of the feature.

I think we should see how it interacts:
- using fragments with rustup doc ... on
  - mac
  - windows
  - linux
  - wsl

If the fragment doesn't get consistently forwarded to the browser then it may just cause more confusion.

I'd also love to see more testing around doc TBH. We have a very good test coverage in most of rustup, but some bits are not as strong ... I think here an appropriate testing strategy would be to do a unit test in the same file, checking for the emitted url for a few combinations of parameters. It might help to break out a helper function that takes the Option<>'s from clap and returns the url.

src/cli/rustup_mode.rs
YegorStolyarov: This is the way to resolve &str borrowing issue. If we put this definition into if statement we will get error:
error[E0597]: `cached_path` does not live long enough. because source String must live as long as let doc_url lives.
TBH I just used the same workaround that is already used  with `let topical_path: PathBuf;` one line above.
maybe there is a nicer way to solve this.

src/cli/rustup_mode.rs
YegorStolyarov: I have tried to use `open_docs()`, but there was a problem with differentiating web links from file links when we use urls with anchors. Problem is caused by "#" symbol in url and located in the opener crate

So I had to use `utils::open_browser()` to prepend "file:///" to the url arbitrary, after generating url with `toolchain.doc_path()`

Or maybe we should go further and prepend "file:///" everywhere?

YegorStolyarov:
Currently I'm stuck with writing tests for my feature.
It's difficult for me to reproduce needed context using existing test helpers, or I am doing it wrong.

I tried to move url generating logic into separate function:
```
fn get_doc_url(m: &ArgMatches, t: &Toolchain<'_>) -> Result<(String, bool)> {
    if let Some((short, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
        if let Some(topic) = m.get_one::<String>("topic") {
            let cached_path = format!("{}/{}", short, topic);
            Ok((cached_path, true))
        } else {
            Ok((path.to_string(), false))
        }
    } else if let Some(topic) = m.get_one::<String>("topic") {
        let topical_path = topical_doc::local_path(&t.doc_path("").unwrap(), topic)?;
        Ok((topical_path.to_str().unwrap().to_string(), false))
    } else {
        Ok(("index.html".to_string(), false))
    }
}
```
So it needs a mock toolchain for test.

Need some clues or an advice.

0xPoe:
> I tried to move url generating logic into separate function:

If you moved it into a separate function, then you can write some unit tests to cover it.

rbtcollins:
Toolchain takes a Cfg, a LocalToolchainName and a path (to the toolchain root dir). 

Cfg has a pretty shallow needs for testing. See [this example](https://github.com/rust-lang/rustup/blob/132f9e762f13019040ce5694c6c8408f90d49bbd/src/dist/component/tests.rs#L23). We could of course make it better.

LocalToolchainName is a very narrow object with no exterior state, you can just create the one you want. E.g. `LocalToolchainName::from(ToolchainDesc::from_str("stable").unwrap()).unwrap();`

The path - since this is a unit test and meant to be in a read-only code path, you could just supply any arbitrary string you want quite sensibly I think.

`test::mock::topical_doc_data` might be useful, or at least relevant.

rami3l:
I'm closing this PR since it has been superseded by #4070. Thanks again @YegorStolyarov for your previous efforts!

PR #3396, opened 2023-07-01
Use available_parallelism to replace the `num_cpus`crate

0xPoe:
close https://github.com/rust-lang/rustup/pull/3393

Use available_parallelism to replace the `num_cpus`crate. It has already been stable for a while. So I think we can replace it safely.

### DIFF
--- a/src/diskio/mod.rs
+++ b/src/diskio/mod.rs
@@ -60,6 +60,7 @@ use std::io::{self, Write};
 use std::ops::{Deref, DerefMut};
 use std::path::{Path, PathBuf};
 use std::sync::mpsc::Receiver;
+use std::thread::available_parallelism;
 use std::time::{Duration, Instant};
 use std::{fmt::Debug, fs::OpenOptions};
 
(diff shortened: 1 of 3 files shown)


weihanglo (approved):
Looks good!

src/diskio/mod.rs
```diff
@@ -452,7 +453,7 @@ pub(crate) fn get_executor<'a>(
 ) -> Result<Box<dyn Executor + 'a>> {
     // If this gets lots of use, consider exposing via the config file.
     let thread_count = match process().var("RUSTUP_IO_THREADS") {
-        Err(_) => num_cpus::get(),
+        Err(_) => available_parallelism()?.get(),
```
rbtcollins:
```suggestion
        Err(_) => available_parallelism().map(|p|p.get()).unwrap_or(1),
```

or something similar - the docs for available parallelism document failure cases where it is better for rustup to run slower than to fail.

0xPoe: Sounds good!

PR #3402, opened 2023-07-08
buf writes to components to prevent small writes

klensy:
Buf writes to components.

Without this fix, 100kb file `manifest-rust-src` written line by line, consisting of about 3k writes; with this pr ~ 20 writes.

### DIFF
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -64,7 +64,7 @@ pub(crate) fn append_file(name: &'static str, path: &Path, line: &str) -> Result
 
 pub(crate) fn write_line(
     name: &'static str,
-    file: &mut File,
+    mut file: impl Write,
     path: &Path,
     line: &str,
 ) -> Result<()> {
--- a/src/dist/component/components.rs
+++ b/src/dist/component/components.rs
@@ -2,6 +2,7 @@
 //! `Components` and `DirectoryPackage` are the two sides of the
 //! installation / uninstallation process.
 
+use std::io::BufWriter;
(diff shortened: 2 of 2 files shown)


0xPoe:
@klensy Thanks for your patch! The CI failed.

rbtcollins:
Thank you for spotting this and putting a patch together. I think it will be easiest to merely put the buf writer where add_file is called, as it has only one call site outside of a) tests, b) single-IO write_file, and c) other uninteresting layering. https://github.com/rust-lang/rustup.rs/blob/fix-renovate/src/dist/component/components.rs#L125

rbtcollins (changes requested):
See my earlier comment for a suggestion

src/utils/utils.rs
```diff
@@ -64,7 +64,7 @@ pub(crate) fn append_file(name: &'static str, path: &Path, line: &str) -> Result
 
 pub(crate) fn write_line(
     name: &'static str,
-    file: &mut File,
+    file: &mut dyn Write,
```
djc: I've noticed in the past that the vtable indirection can cost some performance for I/O, and given that there's only one call site it probably couldn't hurt to use `impl Write` instead of `dyn Write` here.

(IMO we might also instead just inline `utils::write_line()` given that it only has one caller, which would avoid needing to specify the type -- but that should probably be in a separate commit.)

klensy: Ok.

djc (approved):

rbtcollins:
Thanks for the patch. merged now.

PR #3404, opened 2023-07-12
rustup-init.sh: Check for kernel UAPI compatibility on LoongArch

xen0n:
In preparation for the next release which will feature LoongArch support, for a smoother UX in case the user is running an incompatible system (see the commit message & comment for some details).

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -240,6 +240,67 @@ get_endianness() {
     fi
 }
 
+# Detect the Linux/LoongArch UAPI flavor, with all errors being non-fatal.
+# Returns 0 or 234 in case of successful detection, 1 otherwise (/tmp being
+# noexec, or other causes).
+check_loongarch_uapi() {
+    need_cmd base64
+
+    local _tmp
+    if ! _tmp="$(ensure mktemp)"; then
+        return 1
+    fi
+
+    # Minimal Linux/LoongArch UAPI detection, exiting with 0 in case of
+    # upstream ("new world") UAPI, and 234 (-EINVAL truncated) in case of
+    # old-world (as deployed on several early commercial Linux distributions
+    # for LoongArch).
+    #
+    # See https://gist.github.com/xen0n/5ee04aaa6cecc5c7794b9a0c3b65fc7f for
+    # source to this helper binary.
+    ignore base64 -d > "$_tmp" <<EOF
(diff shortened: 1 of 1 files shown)


xry111:
Would it be easier to just check `uname -r` like the following?

```
printf "5.19.0\n$(uname -r)\n" | sort -C -V
```

xen0n:
> Would it be easier to just check `uname -r` like the following?
> 
> ```
> printf "5.19.0\n$(uname -r)\n" | sort -C -V
> ```

I don't have guarantees that a 5.19+ but old-world kernel would not get created and/or deployed to end users. If such guarantee would be given I'd agree this is the simpler way to go.

xry111:
> > Would it be easier to just check `uname -r` like the following?
> > ```
> > printf "5.19.0\n$(uname -r)\n" | sort -C -V
> > ```
> 
> I don't have guarantees that a 5.19+ but old-world kernel would not get created and/or deployed to end users. If such guarantee would be given I'd agree this is the simpler way to go.

Perhaps we can ask Huacai...  To me those organizations must keep to use the old-world will only update kernels to a new patch version anyway.  They are quite conservative on any upgrading.

xen0n:
Self note: If we were to consider some of the early and experimental new-world distros (that have incompatible glibc symbol versions), the statically-linked helper included won't properly detect them as incompatible. We would need to link to at least one versioned symbol in `libc` in order to check this, but I don't think this case is worth supporting due to the explicit incompatibility notice and small user base.

xen0n:
> > > Would it be easier to just check `uname -r` like the following?
> > > ```
> > > printf "5.19.0\n$(uname -r)\n" | sort -C -V
> > > ```
> > 
> > 
> > I don't have guarantees that a 5.19+ but old-world kernel would not get created and/or deployed to end users. If such guarantee would be given I'd agree this is the simpler way to go.
> 
> Perhaps we can ask Huacai... To me those organizations must keep to use the old-world will only update kernels to a new patch version anyway. They are quite conservative on any upgrading.

I asked Huacai and the answer is "theoretically this can be done, but ultimately it depends on who's writing code".

Obviously, I don't trust the corporate devs making the commercial distros to always do the right thing, so I'd say checking kernel ABI would be more appropriate. (After a couple of years it may finally become evident whether the assumption actually holds, but at that time I'd expect the number of old-world-based developers to shrink enough that this is no longer a concern.)

Issue #3316, opened 2023-04-11
remove rls support


rbtcollins:
With rust-analyzer taking over, should we remove the rls specific tests, and/or the rls proxy?

mattheww:
The `rls` proxy is something of an active nuisance.

In its default configuration in emacs 29, eglot knows that both rls and rust-analyzer exist, sees that both are available on PATH, and prompts the user rather than just choosing to run rust-analyzer.

lnicola:
I'd blame that on eglot, it shouldn't be trying to run rls. Consider filing an issue there.

(That's not to say the RA proxy itself isn't annoying if you compile your own RA, but it's the best way we have to make rustup-managed RA work. But that's a completely different topic.)

mattheww:
Eglot upstream removed the reference to `rls` from its default configuration some months ago, but of course it takes time for these things to filter through to users, and there may well be other clients out there in a similar position.

rami3l:
I believe all work required in this repo has concluded in #3920. As such, I'm closing this issue as completed now. For further work, please refer to https://github.com/rust-lang/rust/pull/126856.

PR #3410, opened 2023-07-17
Remove rls and rust-analysis from components list.

burakemir:
This is a documentation-only change.

Since rls and rust-analysis do not seem to be useful, removing those from the component list.

RLS was deprecated over a year ago. https://blog.rust-lang.org/2022/07/01/RLS-deprecation.html
The rust-analysis component seems to only contain a placeholder file with the contents

```
{"warning": "The `rust-analysis` component has been removed." }"
```

I checked a few occurrences in rustup toolchain manifest files, and they only seem to contain that "removed.json".

I believe no one will miss these things.

### DIFF
--- a/.github/workflows/linux-builds-on-master.yaml
+++ b/.github/workflows/linux-builds-on-master.yaml
@@ -129,7 +129,7 @@ jobs:
             "${DOCKER}" \
             -c 'PATH="${PATH}":/rustc-sysroot/bin bash ci/run.bash'
       - name: Upload the built artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: matrix.mode == 'release'
         with:
           name: rustup-init-${{ matrix.target }}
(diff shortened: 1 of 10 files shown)


0xPoe:
cc: https://github.com/rust-lang/rustup/issues/3316

0xPoe (approved):
Looks good  to me.

0xPoe:
@burakemir Please rebase your branch with the latest master branch to fix the test.

rbtcollins:
I think we should rearrange rather than remove these entries. Rustup installs both current and old versions of toolchains, and those components remain valid in older releases: the removal only affects newer channel metadata, not existing.

Something like 
```
### Previous components

These components have been deprecated and are not published in new Rust releases.

<table here>
```

rbtcollins (changes requested):
Changes requested in my other comment

burakemir:
Done.

djc (approved):

PR #3421, opened 2023-07-28
Capturing IO error in download_file_with_resume

### LINKED ISSUE #2937: Need better error when trying to install with low disk space
### Problem

I tried to install Rust when disk space was low and got the following error:

```error: component download failed for cargo-x86_64-unknown-linux-gnu: could not download file from 'https://static.rust-lang.org/dist/2022-01-20/cargo-1.58.1-x86_64-unknown-linux-gnu.tar.xz' to '/root/.rustup/downloads/1d3f01b973ea3de5174d0a48b50828cf913bafc223ecbf768a0293602cff0cdc.partial'```

which led me look for networking problems.

I would expect the error to indicate that the download could not be done due to limited disk space.

### Steps

1. Use an x86-64 Linux
2. Fill up your disk
3. Run the Rust installer with default options

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
Rustup was not installed when the error happened, but now that I made some disk space and re-ran the installer, the version is reported as:

info: The currently active `rustc` version is `rustc 1.58.1 (db9d1b20b 2022-01-20)`
```

### Installed toolchains

```console
rustup home:  /home/user/.rustup

rustc 1.58.1 (db9d1b20b 2022-01-20)
```

ongchi:
This patch matches the I/O error in `utils::utils::download_file_with_resume` and makes the `RustupError::IOError` transparent, which improves the error message when an I/O error occurs.

Fixes: #2937

### DIFF
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -151,6 +151,9 @@ pub(crate) fn download_file_with_resume(
     match download_file_(url, path, hasher, resume_from_partial, notify_handler) {
         Ok(_) => Ok(()),
         Err(e) => {
+            if let Some(_) = e.downcast_ref::<std::io::Error>() {
+                return Err(e);
+            }
             let is_client_error = match e.downcast_ref::<DEK>() {
                 // Specifically treat the bad partial range error as not our
                 // fault in case it was something odd which happened.


ongchi:
I found that most errors are of the type `anyhow::Error` with context that is a bit messy, and the error chain only displays the first message and sometimes irrelevant.
Would it be easier to change most errors with type `anyhow::Error` to a concrete error type? I'm not sure, but I think most errors in rustup should cause the program to exit with an error immediately, so error chaining might not be necessary.

src/utils/utils.rs
```diff
@@ -151,23 +151,35 @@ pub(crate) fn download_file_with_resume(
     match download_file_(url, path, hasher, resume_from_partial, notify_handler) {
         Ok(_) => Ok(()),
         Err(e) => {
+            let mut maybe_io_error = None;
```
rbtcollins: Do we need the temp variable at all? This looks like it would be amenable to an early exit approach instead.

rbtcollins:
Thank you for this patch.

Rustup's codebase is quite old, I think we're on the 3rd or fourth iteration of 'find a good error handling pattern'.

I have no in-principle objection to concrete error types everywhere, but I suspect it will be rather messier for at least a while, and that I do worry about since the code is already a strain to comprehend quite often, with many subtly different concepts packed densely. (e.g. the difference between what users give rustup as toolchain names, and what it uses itself, and those things and what an installed toolchain look like ... is an area we only just improved, and its still a lot of cognitive load).

tl;dr by all means do some experiments and we can discuss, but its not clear that it will be a net improvement,.

ongchi:
I think it would be better to wrap up `std::io::Error` by `DownloadError:IoError`, then we can just downcast to `DownloadError` with less surprise.

src/utils/utils.rs
ongchi: This makes sense, we should just early return in the match arm.

src/utils/utils.rs
```diff
@@ -156,6 +156,9 @@ pub(crate) fn download_file_with_resume(
                 // fault in case it was something odd which happened.
                 Some(DEK::HttpStatus(416)) => false,
                 Some(DEK::HttpStatus(400..=499)) | Some(DEK::FileNotFound) => true,
+                Some(DEK::IoError(io_err)) => {
+                    return Err(RustupError::IOError(std::io::Error::from(io_err.kind())).into());
```
rbtcollins: isn't io_err an IO error itself? Why do we need to construct a new one here?

ongchi: The `io_err` is an `&Error` and cannot be moved out, but I just realized we could return `Err(e)` in this match arm.

I'm also thinking about moving the logic here into `download::download_to_path_with_backend` and also moving these error variants into `DownloadError`, and not having to downcast `anyhow::Error` to a concrete error type. Would this be a good approach?

rbtcollins: If you want to experiment with it, and it turns out nicely, that does sound good.

ongchi: After investigation, I found some rustup logic relays on these error types. It would not be very beneficial to move these errors into download crate.

ongchi:
https://github.com/rust-lang/rustup/blob/871d9287010e774b4016ee492a1bac69bdd17d4c/download/src/lib.rs#L100-L116

Shall we clear up the cached files? Would resuming from partial ever happen since we clear it up when an error occurred?

rbtcollins:
> https://github.com/rust-lang/rustup/blob/871d9287010e774b4016ee492a1bac69bdd17d4c/download/src/lib.rs#L100-L116
> 
> Shall we clear up the cached files? Would resuming from partial ever happen since we clear it up when an error occurred?

I think the comment above that is saying that the directory downloads are cached in should exist before reaching this point. And that on failure we try to remove the cache entry for this particular file.

Then the other comment is saying that it is probably ok to keep the file for some errors (e.g. connection lost mid-way), but we don't do that yet.

In terms of restructuring here, I think that should be a different PR at a minimum. And we need to consider these edge cases:
- partial files may exist even though we remove files on failure because of:
  - rustup might be terminated uncleanly (SIGILL, SIGQUIT, power failure, older version of rustup
  - the download cache directory might not exist because a user deletes it in a race with rustup (not currently handled, but I'm not sure we fail particularly gracefully either)

download/src/lib.rs
```diff
@@ -35,108 +35,87 @@ pub enum Event<'a> {
...
+                };
+            }
+        }
 
-            let mut possible_partial = OpenOptions::new()
-                .write(true)
-                .create(true)
-                .open(path)
-                .context("error opening file for download")?;
+        let mut possible_partial = OpenOptions::new().write(true).create(true).open(path)?;
```
rbtcollins: This drops the .context - will the resulting error be shown to users? The problem with file open errors is that they can come from anywhere...

download/src/lib.rs
```diff
@@ -35,108 +35,87 @@ pub enum Event<'a> {
...
         })?;
 
-        file.borrow_mut()
-            .sync_data()
-            .context("unable to sync download to disk")?;
-
-        Ok(())
-    }()
-    .map_err(|e| {
+        Ok(file.borrow_mut().sync_data()?)
```
rbtcollins: same here.

download/src/lib.rs
```diff
@@ -35,108 +35,87 @@ pub enum Event<'a> {
...
+    }() {
         // TODO: We currently clear up the cached download on any error, should we restrict it to a subset?
-        if let Err(file_err) = remove_file(path).context("cleaning up cached downloads") {
-            file_err.context(e)
-        } else {
-            e
-        }
-    })
+        remove_file(path)?;
```
rbtcollins: same here - this will lose the error information won't it?

download/src/lib.rs
```diff
@@ -277,40 +250,56 @@ pub mod reqwest_be {
...
         resume_from: u64,
-        callback: &dyn Fn(Event<'_>) -> Result<()>,
+        callback: Callback<'_>,
         tls: TlsBackend,
     ) -> Result<()> {
-        // Short-circuit reqwest for the "file:" URL scheme
-        if download_from_file_url(url, resume_from, callback)? {
-            return Ok(());
-        }
+        // The file scheme is mostly for use by tests to mock the dist server.
```
rbtcollins, line 262: It is also used by some organisations with mirrors on NFS - that kind of thing. I think it is a mistake to over-emphasis the test usage here - if we truely wanted to make it test only, we'd make that support a compile time feature.

rbtcollins: I'm curious why the inlining of the separate function?

rbtcollins:
I've had another look at this, and it seems the size of the change has grown quite a bit? It would help me review if we keep this PR focused on the functional change - improving the error message for users. 

Refactorings are their own intent, and need to be reviewed and explained from a different angle

ongchi:
Sorry, I should create a new issue or PR.
Thank you for your time.

djc:
Can you maybe show example outputs of how your change affected `rustup` output?

ongchi:
The rustup will always return an error of RustupError::DownloadNotExists or RustupError::DownloadingFile, and the original message will be like this:

> error: component download failed for rust-std-aarch64-apple-darwin: could not download file from 'https://static.rust-lang.org/dist/2023-08-21/rust-std-nightly-aarch64-apple-darwin.tar.xz' to '/Volumes/rustup/home/downloads/79eea81780aa687bff712e32cc9149786e4f094bfa5bf13e3879cfcc7fb9254a.partial'

The error comes from https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/download/src/lib.rs#L52-L140
which will check if partial is accessible before downloading files from the network. If an error happens, it will clear up the partial file before returning an error wrapped by `anyhow::Error`. It's confusing when out of disk space.

This PR makes the error message contain error context from `download::download_to_path_with_backend`, making the message much more relevant than before.

> error: component download failed for rust-std-aarch64-apple-darwin: unable to write download to disk

djc (approved):
Thanks, that looks like a big improvement!

rbtcollins (approved):

PR #3435, opened 2023-08-13
Support `RUSTUP_TERM_COLOR` as an override environment variable

### LINKED ISSUE #3433: Launching `rustup` from `Command` does not return a colored output

When I run this command:

```rust
Command::new("rustup")
        .args(["target", "add", target])
        .output()
        .expect("Failed to run rustup");
```

I cannot get a colored text as output. It does not happen when I run the same command directly from my terminal.

### Solution you'd like

Would it be possible to add a `--color` option even for the `rustup` installer? Thanks in advance! 

### Notes

_No response_

rami3l:
Closes #3433, getting my feet wet at the same time.

The variable name and the description are directly copied from the [Cargo Book](https://doc.rust-lang.org/cargo/reference/config.html#termcolor).

### Concerns

- [x] Is the naming of the variable ideal?
- [ ] Are there any test cases to be added for this change?
- [ ] ~~Should we add a CLI flag (e.g. `--color`) for this as well?~~
  - [ ] ~~If so, should we use Clap's `env` feature?~~

### DIFF
--- a/src/currentprocess/terminalsource.rs
+++ b/src/currentprocess/terminalsource.rs
@@ -19,6 +19,8 @@ pub(super) enum StreamSelector {
     Stderr,
     #[cfg(feature = "test")]
     TestWriter(TestWriter),
+    #[cfg(feature = "test")]
+    TestTtyWriter(TestWriter),
 }
 
 impl StreamSelector {
@@ -36,6 +38,8 @@ impl StreamSelector {
             },
             #[cfg(feature = "test")]
             StreamSelector::TestWriter(_) => false,
+            #[cfg(feature = "test")]
+            StreamSelector::TestTtyWriter(_) => true,
         }
     }
 }
@@ -55,7 +59,7 @@ pub struct ColorableTerminal {
 enum TerminalInner {
     StandardStream(StandardStream, ColorSpec),
     #[cfg(feature = "test")]
-    TestWriter(TestWriter),
+    TestWriter(TestWriter, ColorChoice),
 }
 
 pub struct ColorableTerminalLocked {
@@ -94,7 +98,9 @@ impl ColorableTerminal {
                 TerminalInner::StandardStream(StandardStream::stderr(choice), ColorSpec::new())
             }
             #[cfg(feature = "test")]
-            StreamSelector::TestWriter(w) => TerminalInner::TestWriter(w),
+            StreamSelector::TestWriter(w) | StreamSelector::TestTtyWriter(w) => {
+                TerminalInner::TestWriter(w, choice)
+            }
         };
         ColorableTerminal {
             inner: Arc::new(Mutex::new(inner)),
@@ -122,7 +128,7 @@ impl ColorableTerminal {
                     TerminalInnerLocked::StandardStream(locked)
                 }
                 #[cfg(feature = "test")]
-                TerminalInner::TestWriter(w) => TerminalInnerLocked::TestWriter(w.lock()),
+                TerminalInner::TestWriter(w, _) => TerminalInnerLocked::TestWriter(w.lock()),
             });
             // ColorableTerminalLocked { inner, guard, locked }
             uninit.assume_init()
@@ -136,7 +142,7 @@ impl ColorableTerminal {
                 s.set_color(spec)
             }
             #[cfg(feature = "test")]
-            TerminalInner::TestWriter(_) => Ok(()),
+            TerminalInner::TestWriter(_, _) => Ok(()),
         }
     }
 
@@ -147,7 +153,7 @@ impl ColorableTerminal {
                 s.set_color(spec)
             }
             #[cfg(feature = "test")]
-            TerminalInner::TestWriter(_) => Ok(()),
+            TerminalInner::TestWriter(_, _) => Ok(()),
         }
     }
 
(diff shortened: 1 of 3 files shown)


doc/src/environment-variables.md
```diff
@@ -29,6 +29,12 @@
   determines the directory that traces will be written too. Traces are of the
   form PID.trace. Traces can be read by the Catapult project [tracing viewer].
 
+- `RUSTUP_TERM_COLOR` (default: `auto` in TTY, `never` otherwise) Controls whether colored output is used in the terminal. Possible values:
```
rbtcollins: I don't know how to parse the content in the brackets. Are you describing how `auto` works? If so perhaps just state the default, and describe the `auto` behaviour below.

src/currentprocess/terminalsource.rs
```diff
@@ -9,9 +9,10 @@ use std::{
 pub(crate) use termcolor::Color;
 use termcolor::{ColorChoice, ColorSpec, StandardStream, StandardStreamLock, WriteColor};
 
+use super::{process, varsource::VarSource};
```
rbtcollins: Please keep this import grouped with the other super:: import

src/currentprocess/terminalsource.rs
```diff
@@ -73,14 +74,17 @@ enum TerminalInnerLocked {
 }
 
 impl ColorableTerminal {
-    /// Construct a terminal for the selected stream. Colors written to the
-    /// terminal will be discarded if the stream is not a tty.
+    /// Construct a terminal for the selected stream.
+    /// Colors written to the terminal will be discarded if the stream is not a tty,
+    /// unless the `RUSTUP_TERM_COLOR` override environment variable is set.
```
rbtcollins:
```suggestion
    /// A terminal that supports colorisation of a stream.
    /// If `RUSTUP_TERM_COLOR` is set to `always`, or if the stream is a tty and
    /// `RUSTUP_TERM_COLOR` either unset or set to `auto`, then 
    /// color commands will be sent to the stream.
    /// Otherwise color commands are discarded.
```

src/currentprocess/terminalsource.rs
```diff
@@ -73,14 +74,17 @@ enum TerminalInnerLocked {
...
     pub(super) fn new(stream: StreamSelector) -> Self {
-        let is_a_tty = stream.is_a_tty();
-        let choice = if is_a_tty {
-            ColorChoice::Auto
-        } else {
-            ColorChoice::Never
+        let env_override = process().var("RUSTUP_TERM_COLOR").ok();
+        let choice = match env_override.as_deref() {
+            Some("always") => ColorChoice::Always,
+            Some("auto") => ColorChoice::Auto,
```
rbtcollins: this choice is not needed: it is the same as `_ if stream.is_a_tty` by intent. I suggest just eliding it, so there will be:
Some(always)
Some(never)
_ if stream.is_a_tty
_ => ColorChoice::Never

rbtcollins:
Thank you for the patch.

> Is the naming of the variable ideal?

I think its fine.

> Is there any test cases to be added for this change?

yes, unit tests in terminalsource would be best I think.

> Should we add a CLI flag (e.g. --color) for this as well?

We generally don't expose variables as CLI parameters.

doc/src/environment-variables.md
rami3l: Done.

src/currentprocess/terminalsource.rs
rami3l: Done.

src/currentprocess/terminalsource.rs
rami3l: Done.

src/currentprocess/terminalsource.rs
rami3l: I agree. I also removed the `.ok()` part to match on `Ok` instead.

src/currentprocess/terminalsource.rs
```diff
@@ -73,14 +77,18 @@ enum TerminalInnerLocked {
...
     pub(super) fn new(stream: StreamSelector) -> Self {
-        let is_a_tty = stream.is_a_tty();
-        let choice = if is_a_tty {
-            ColorChoice::Auto
-        } else {
-            ColorChoice::Never
+        let env_override = process().var("RUSTUP_TERM_COLOR");
+        let choice = match env_override.as_deref() {
+            Ok("always") => ColorChoice::Always,
+            Ok("never") => ColorChoice::Never,
```
rbtcollins, line 91: I think the value of the environment variable should be lowercased (or upper) - either way, a case error from a user shouldn't prevent this code detecting their intent.

rami3l:
Hmmm. IMO that CI error has nothing to do with the changes I've made... (cc https://github.com/rust-lang/rustup/issues/3437)

weihanglo:
(drive-by comment)

Some related issue in rust-lang/cargo, if we want to make sure the behavior is in sync.

* https://github.com/rust-lang/cargo/issues/11211#issuecomment-1274499277 (switched to `std::io::IsTerminal` though)
* https://github.com/rust-lang/cargo/issues/9835

weihanglo:
(drive-by comment)

Some related issue in rust-lang/cargo, if we want to make sure the behavior is in sync.

* https://github.com/rust-lang/cargo/issues/11211#issuecomment-1274499277 (switched to `std::io::IsTerminal` though)
* https://github.com/rust-lang/cargo/issues/9835

PR #3436, opened 2023-08-13
Dev-docs-toc

rbtcollins:
We had a suggestion that the dev docs would be easier to read if there
was a table of contents. Rather than making one by hand I thought using
mdbook would be better.


djc (approved):

djc:
Drive-by review, this looks good to me!

PR #3438, opened 2023-08-14
macOS `uname -m` can lie due to Rosetta shenanigans

### LINKED ISSUE #3419: Upstream a bug fix for platform detection on MacOS
The Julia programming language has a tool Juliaup which was not only heavily inspired by rustup, but I also copied the [rustup-init.sh](https://github.com/rust-lang/rustup/blob/master/rustup-init.sh) script as a starting point. First of all: thanks so much for an awesome project, the inspiration and the code, it made building Juliaup sooo much easier!

We recently fixed a pretty bizzar corner case bug with the shell script re platform detection, and I just thought I'd let you know in case you want to copy that over into your `rustup-init.sh` :) The PR is https://github.com/JuliaLang/juliaup/pull/701.

rami3l:
Closes #3419.

Originally provided in https://github.com/JuliaLang/juliaup/pull/701, this shell script fix uses `sysctl` to prevent Rosetta 2 from lying about the system's actual CPU architecture.

---

Possibly related: #3068

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -261,11 +261,22 @@ get_architecture() {
         # get the real answer, but that's hard to ensure, so instead we use
         # `sysctl` (which doesn't lie) to check for the actual architecture.
         if [ "$_cputype" = i386 ]; then
-            if sysctl hw.optional.x86_64 | grep -q ': 1'; then
+            # Handling i386 compatibility mode in older macOS versions (<10.15)
+            # running on x86_64-based Macs.
+            # Starting from 10.15, macOS explicitly bans all i386 binaries from running.
+            # See: <https://support.apple.com/en-us/HT208436>
+
+            # Avoid `sysctl: unknown oid` stderr output and/or non-zero exit code.
+            if sysctl hw.optional.x86_64 2> /dev/null || true | grep -q ': 1'; then
                 _cputype=x86_64
             fi
         elif [ "$_cputype" = x86_64 ]; then
-            if sysctl hw.optional.arm64 | grep -q ': 1'; then
+            # Handling x86-64 compatibility mode (a.k.a. Rosetta 2)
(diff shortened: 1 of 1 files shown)


djc (approved):
Drive-by review: this makes sense to me.

Are the CI runs expected to be flaky?

rami3l:
@djc Thanks for the review!

> Are the CI runs expected to be flaky?

I don't think so. The causes of 3 unsuccessful runs are completely different. What happened? 🤔️

rami3l:
There was indeed a tiny detail regarding `sysctl`'s stderr output and/or exit code that might break the CI.
Now there should be no more CI-related problems related to the changes I've made :)

rbtcollins (commented):
So it's not possible to get an i386 binary running on Apple silicon?

Perhaps querying hw.optional and grepping for the sub keys we understand would be better?

rami3l:
> So it's not possible to get an i386 binary running on Apple silicon?

@rbtcollins No.

AFAIK Apple's transition to ARM64 happened after having explicitly banned all x86_32 binaries in a previous macOS major upgrade, so an x86_32 binary can never run on ARM64 even with Rosetta 2 enabled.

The x86_32 compatibility part is for the old macOS versions before that ban.

rbtcollins:
> @rbtcollins No.
> 
> AFAIK Apple's transition to ARM64 happened after having explicitly banned all x86_32 binaries in a previous macOS major upgrade, so an x86_32 binary can never run on ARM64 even with Rosetta 2 enabled.
> 
> The x86_32 compatibility part is for the old macOS versions before that ban.

Great. Could you please document that in a comment - I think its sufficiently specialised knowledge that we'll forget in a few weeks :)

rami3l:
> Great. Could you please document that in a comment - I think its sufficiently specialised knowledge that we'll forget in a few weeks :)

@rbtcollins Done.

Issue #3290, opened 2023-03-27
Rustup will panic if you using `rustup -v` or `rustup -q`


0xPoe:
### Problem

Rustup will panic if you using `rustup -v` or `rustup -q`.

```console
thread 'main' panicked at 'internal error: entered unreachable code', src/cli/rustup_mode.rs:219:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Steps

1. `rustup -q` or `rustup -v`

### Possible Solution(s)

None

### Notes

If you run it in rustup 1.25.2, you will get the help text instead of panic.

### Rustup version

```console
➜  ~ rustup --version
rustup 1.26.0 (9ac66b5ee 2023-03-22)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.70.0-nightly (1459b3128 2023-03-23)`
```

### Installed toolchains

```console
➜  ~ rustup show
Default host: aarch64-apple-darwin
rustup home:  /Users/hi-rustin/.rustup

installed toolchains
--------------------

stable-aarch64-apple-darwin
nightly-2022-11-15-aarch64-apple-darwin
nightly-aarch64-apple-darwin (default)

active toolchain
----------------

nightly-aarch64-apple-darwin (default)
rustc 1.70.0-nightly (1459b3128 2023-03-23)
```

Issue #3156, opened 2023-01-14
Duplicative auto completions


aurexav:
### Problem

Duplicative auto completions

### Steps

```
rustup completions zsh rustup > ~/.cache/oh-my-zsh/completions/_rustup
```

```
rustup c <TAB>

check        -- Check for updates to Rust toolchains and rustup
completions  -- Generate tab-completion scripts for your shell
component    -- Modify a toolchain's installed components
# It also starts the path completion, weird
code/
```

```
rustup check <TAB>

# All these are duplicative
check           -- Check for updates to Rust toolchains and rustup
completions     -- Generate tab-completion scripts for your shell
component       -- Modify a toolchain's installed components
default         -- Set the default toolchain
doc             -- Open the documentation for the current toolchain
dump-testament  -- Dump information about the build
help            -- Prints this message or the help of the given subcommand(s)
install         -- Update Rust toolchains
man             -- View the man page for a given command
override        -- Modify directory toolchain overrides
run             -- Run a command with an environment configured for a given toolchain
self            -- Modify the rustup installation
set             -- Alter rustup settings
show            -- Show the active and installed toolchains or profiles
target          -- Modify a toolchain's supported targets
toolchain       -- Modify or query the installed toolchains
uninstall       -- Uninstall Rust toolchains
update          -- Update Rust toolchains and rustup
which           -- Display which binary will be run for a given command
```

```
rustup check check --v

--help     -h  -- Prints help information
--version  -V  -- Prints version information
```

```
rustup check check --version

error: Found argument 'check' which wasn't expected, or isn't valid in this context

USAGE:
    rustup check

For more information try --help
```

### Rustup version

```console
rustup 1.25.1 (bb60b1e89 2022-07-12)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.68.0-nightly (0b90256ad 2023-01-13)`
```

### Installed toolchains

```console
Default host: aarch64-apple-darwin
rustup home:  /Users/xavier/.rustup

installed toolchains
--------------------

stable-aarch64-apple-darwin
nightly-aarch64-apple-darwin (default)

installed targets for active toolchain
--------------------------------------

aarch64-apple-darwin
wasm32-unknown-unknown

active toolchain
----------------

nightly-aarch64-apple-darwin (default)
rustc 1.68.0-nightly (0b90256ad 2023-01-13)
```

rbtcollins:
[...]

PR #3444, opened 2023-08-15
Upgrade to clap 4

### LINKED ISSUE #3128: PowerShell completions installation does not work as documented

### LINKED ISSUE #3443: atty potential unaligned read
atty potential unaligned read

(This bug is created from a dependabot alert - but for rustup it is not a security vuln today. There is a latent bug risk if we were to install a custom allocator, so this should get fixed).

`atty` is present in Cargo.lock via `clap`. We're currently on version 3 of clap, possibly v4 drops the dependency, or possibly the work needs to be done in `clap` first.

Open Opened June 30, 2023 22:43 on atty (Rust) · Cargo.lock
| Package | Affected versions | Patched version
|-|-|-
| atty  (Rust) | <= 0.2.14 | None

On windows, atty dereferences a potentially unaligned pointer.

In practice however, the pointer won't be unaligned unless a custom global allocator is used.

In particular, the System allocator on windows uses HeapAlloc, which guarantees a large enough alignment.
atty is Unmaintained

A Pull Request with a fix has been provided over a year ago but the maintainer seems to be unreachable.

Last release of atty was almost 3 years ago.
Possible Alternative(s)

The below list has not been vetted in any way and may or may not contain alternatives;

    [std::io::IsTerminal](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html) - Stable since Rust 1.70.0\n
    [is-terminal](https://crates.io/crates/is-terminal) - Standalone crate supporting Rust older than 1.70.0"

djc:
Review the changelog here: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#400---2022-09-28.

Previously: #3094, which seems stalled.

Fixes #3443 and, apparently, fixes #3128.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,17 +130,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -311,51 +300,40 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.2.25"
+version = "4.3.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ea181bf566f71cb9a5d17a59e1871af638180a18fb0035c92ae62b705207123"
+checksum = "c27cdf28c0f604ba3f512b0c9a409f8de8513e4816705deb0498b627e7c3a3fd"
 dependencies = [
- "atty",
- "bitflags 1.3.2",
- "clap_derive",
+ "clap_builder",
+]
+
+[[package]]
+name = "clap_builder"
+version = "4.3.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08a9f1ab5e9f01a9b81f202e8562eb9a10de70abf9eaeac1be465c28b75aa4aa"
+dependencies = [
+ "anstream",
+ "anstyle",
  "clap_lex",
- "indexmap 1.9.3",
  "strsim",
- "termcolor",
  "terminal_size",
- "textwrap",
 ]
 
 [[package]]
 name = "clap_complete"
-version = "3.2.5"
+version = "4.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f7a2e0a962c45ce25afce14220bc24f9dade0a1787f185cecf96bfba7847cd8"
+checksum = "5fc443334c81a804575546c5a8a79b4913b50e28d69232903604cada1de817ce"
 dependencies = [
  "clap",
 ]
 
 [[package]]
-name = "clap_derive"
-version = "3.2.25"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ae6371b8bdc8b7d3959e9cf7b22d4435ef3e79e138688421ec654acf8c81b008"
-dependencies = [
- "heck",
- "proc-macro-error",
- "proc-macro2",
- "quote",
- "syn 1.0.109",
-]
-
-[[package]]
 name = "clap_lex"
-version = "0.2.4"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2850f2f5a82cbf437dd5af4d49848fbdfc27c157c3d010345776f952765261c5"
-dependencies = [
- "os_str_bytes",
-]
+checksum = "2da6da31387c7e4ef160ffab6d5e7f00c42626fe39aea70a7b0f1773f7dd6c1b"
 
 [[package]]
 name = "colorchoice"
@@ -876,21 +854,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c6201b9ff9fd90a5a3bac2e56a830d0caa509576f0e503818ee82c181b3437a"
 
 [[package]]
-name = "heck"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
-
-[[package]]
-name = "hermit-abi"
-version = "0.1.19"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62b467343b94ba476dcb2500d242dadbb39557df889310ac77c5d99100aaac33"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "hermit-abi"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1054,7 +1017,7 @@ version = "1.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "eae7b9aee968036d54dce06cebaefd919e4472e753296daccd6d344e3e2df0c2"
 dependencies = [
- "hermit-abi 0.3.2",
+ "hermit-abi",
  "libc",
  "windows-sys 0.48.0",
 ]
@@ -1071,7 +1034,7 @@ version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cb0889898416213fab133e1d33a0e5858a48177452750691bde3666d0fdbaf8b"
 dependencies = [
- "hermit-abi 0.3.2",
+ "hermit-abi",
  "rustix 0.38.8",
  "windows-sys 0.48.0",
 ]
@@ -1301,7 +1264,7 @@ version = "1.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4161fcb6d602d4d2081af7c3a45852d875a03dd337a6bfdd6e06407b61342a43"
 dependencies = [
- "hermit-abi 0.3.2",
+ "hermit-abi",
  "libc",
 ]
 
@@ -1494,12 +1457,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "os_str_bytes"
-version = "6.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d5d9eb14b174ee9aa2ef96dc2b94637a2d4b6e7cb873c7e171f0c20c6cf3eac"
-
-[[package]]
 name = "overload"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1554,30 +1511,6 @@ name = "ppv-lite86"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
-
-[[package]]
-name = "proc-macro-error"
-version = "1.0.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "da25490ff9892aab3fcf7c36f08cfb902dd3e71ca0f9f9517bea02a73a5ce38c"
-dependencies = [
- "proc-macro-error-attr",
- "proc-macro2",
- "quote",
- "syn 1.0.109",
- "version_check",
-]
-
-[[package]]
-name = "proc-macro-error-attr"
-version = "1.0.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1be40180e52ecc98ad80b184934baf3d0d29f979574e439af5a55274b35f869"
-dependencies = [
- "proc-macro2",
- "quote",
- "version_check",
-]
 
 [[package]]
 name = "proc-macro2"
@@ -2356,15 +2289,6 @@ checksum = "8e6bf6f19e9f8ed8d4048dc22981458ebcf406d67e94cd422e5ecd73d63b3237"
 dependencies = [
  "rustix 0.37.23",
  "windows-sys 0.48.0",
-]
-
-[[package]]
-name = "textwrap"
-version = "0.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "222a222a5bfe1bba4a77b45ec488a741b3cb8872e5e499451fd7d0129c9c7c3d"
-dependencies = [
- "terminal_size",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,6 +317,7 @@ checksum = "4ea181bf566f71cb9a5d17a59e1871af638180a18fb0035c92ae62b705207123"
 dependencies = [
  "atty",
  "bitflags 1.3.2",
+ "clap_derive",
  "clap_lex",
  "indexmap 1.9.3",
  "strsim",
@@ -332,6 +333,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f7a2e0a962c45ce25afce14220bc24f9dade0a1787f185cecf96bfba7847cd8"
 dependencies = [
  "clap",
+]
+
+[[package]]
+name = "clap_derive"
+version = "3.2.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ae6371b8bdc8b7d3959e9cf7b22d4435ef3e79e138688421ec654acf8c81b008"
+dependencies = [
+ "heck",
+ "proc-macro-error",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -860,6 +874,12 @@ name = "hashbrown"
 version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c6201b9ff9fd90a5a3bac2e56a830d0caa509576f0e503818ee82c181b3437a"
+
+[[package]]
+name = "heck"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
 
 [[package]]
 name = "hermit-abi"
@@ -1534,6 +1554,30 @@ name = "ppv-lite86"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
+
+[[package]]
+name = "proc-macro-error"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da25490ff9892aab3fcf7c36f08cfb902dd3e71ca0f9f9517bea02a73a5ce38c"
+dependencies = [
+ "proc-macro-error-attr",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+ "version_check",
+]
+
+[[package]]
+name = "proc-macro-error-attr"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a1be40180e52ecc98ad80b184934baf3d0d29f979574e439af5a55274b35f869"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "version_check",
+]
 
 [[package]]
 name = "proc-macro2"
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -26,47 +26,36 @@ RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
 usage() {
     cat <<EOF
 rustup-init 1.26.0 (577bf51ae 2023-04-05)
+
 The installer for rustup
 
-USAGE:
-    rustup-init [OPTIONS]
+Usage: rustup-init[EXE] [OPTIONS]
 
-OPTIONS:
-    -v, --verbose
-            Enable verbose output
-
-    -q, --quiet
-            Disable progress output
-
-    -y
-            Disable confirmation prompt.
-
-        --default-host <default-host>
-            Choose a default host triple
-
-        --default-toolchain <default-toolchain>
-            Choose a default toolchain to install. Use 'none' to not install any toolchains at all
-
-        --profile <profile>
-            [default: default] [possible values: minimal, default, complete]
-
-    -c, --component <components>...
-            Component name to also install
-
-    -t, --target <targets>...
-            Target name to also install
-
-        --no-update-default-toolchain
-            Don't update any existing default toolchain after install
-
-        --no-modify-path
-            Don't configure the PATH environment variable
-
-    -h, --help
-            Print help information
-
-    -V, --version
-            Print version information
+Options:
+  -v, --verbose
+          Enable verbose output
+  -q, --quiet
+          Disable progress output
+  -y
+          Disable confirmation prompt.
+      --default-host <default-host>
+          Choose a default host triple
+      --default-toolchain <default-toolchain>
+          Choose a default toolchain to install. Use 'none' to not install any toolchains at all
+      --profile <profile>
+          [default: default] [possible values: minimal, default, complete]
+  -c, --component <components>...
+          Component name to also install
+  -t, --target <targets>...
+          Target name to also install
+      --no-update-default-toolchain
+          Don't update any existing default toolchain after install
+      --no-modify-path
+          Don't configure the PATH environment variable
+  -h, --help
+          Print help
+  -V, --version
+          Print version
 EOF
 }
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -253,7 +253,8 @@ pub(crate) fn cli() -> Command<'static> {
         .about("The Rust toolchain installer")
         .after_help(RUSTUP_HELP)
         .global_setting(AppSettings::DeriveDisplayOrder)
-        .setting(AppSettings::SubcommandRequiredElseHelp)
+        .subcommand_required(true)
+        .arg_required_else_help(true)
         .arg(
             verbose_arg("Enable verbose output"),
         )
@@ -400,7 +401,8 @@ pub(crate) fn cli() -> Command<'static> {
             Command::new("toolchain")
                 .about("Modify or query the installed toolchains")
                 .after_help(TOOLCHAIN_HELP)
-                .setting(AppSettings::SubcommandRequiredElseHelp)
+                .subcommand_required(true)
+                .arg_required_else_help(true)
                 .subcommand(
                     Command::new("list")
                         .about("List installed toolchains")
@@ -508,7 +510,8 @@ pub(crate) fn cli() -> Command<'static> {
         .subcommand(
             Command::new("target")
                 .about("Modify a toolchain's supported targets")
-                .setting(AppSettings::SubcommandRequiredElseHelp)
+                .subcommand_required(true)
+                .arg_required_else_help(true)
                 .subcommand(
                     Command::new("list")
                         .about("List installed and available targets")
@@ -571,7 +574,8 @@ pub(crate) fn cli() -> Command<'static> {
         .subcommand(
             Command::new("component")
                 .about("Modify a toolchain's installed components")
-                .setting(AppSettings::SubcommandRequiredElseHelp)
+                .subcommand_required(true)
+                .arg_required_else_help(true)
                 .subcommand(
                     Command::new("list")
                         .about("List installed and available components")
@@ -630,7 +634,8 @@ pub(crate) fn cli() -> Command<'static> {
             Command::new("override")
                 .about("Modify directory toolchain overrides")
                 .after_help(OVERRIDE_HELP)
-                .setting(AppSettings::SubcommandRequiredElseHelp)
+                .subcommand_required(true)
+                .arg_required_else_help(true)
                 .subcommand(
                     Command::new("list").about("List directory toolchain overrides"),
                 )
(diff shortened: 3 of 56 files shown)


src/cli/rustup_mode.rs
```diff
@@ -247,12 +248,11 @@ pub fn main() -> Result<utils::ExitCode> {
     })
 }
 
-pub(crate) fn cli() -> Command<'static> {
+pub(crate) fn cli() -> Command {
     let mut app = Command::new("rustup")
         .version(common::version())
         .about("The Rust toolchain installer")
         .after_help(RUSTUP_HELP)
-        .global_setting(AppSettings::DeriveDisplayOrder)
```
djc, line 255: Per the changelog, this has become the default.

djc:
The `rustup_init_ui_doc_text_tests` failure is confusing to me, I tried to get it lined up but the way the test fails doesn't quite make sense to me.

rbtcollins:
Thank you for taking this on.

The first line is an expected line - `rustup-init ...` 
The second line `The installer for rustup` is line 1 of the clap4 output and line 2 of the expected output.

Line `3 -Usage ...` is new in the clap4 output
Line `   4 + USAGE:` is expected but not in the clap4 output

etc.

We're ok with changes to the help layout as long as:
- its internally consistent between subcommands and the like
- it is clean and readable

There was discussion about this during the clap3 upgrade - I forget whether on discord or in the clap3 PR (perhaps in both in fact).

```diff
---- expected: stdout
++++ actual:   stdout
        1 + rustup-init 1.26.0 (577bf51ae 2023-04-05)
   1    2 | The installer for rustup
   2    3 | 
   3      - Usage: rustup-init [OPTIONS]
        4 + USAGE:
        5 +     rustup-init [OPTIONS]
        6 + 
        7 + OPTIONS:
        8 +     -v, --verbose
        9 +             Enable verbose output
       10 + 
       11 +     -q, --quiet
       12 +             Disable progress output
       13 + 
       14 +     -y
       15 +             Disable confirmation prompt.
       16 + 
       17 +         --default-host <default-host>
       18 +             Choose a default host triple
       19 + 
       20 +         --default-toolchain <default-toolchain>
       21 +             Choose a default toolchain to install. Use 'none' to not install any toolchains at all
       22 + 
       23 +         --profile <profile>
       24 +             [default: default] [possible values: minimal, default, complete]
       25 + 
       26 +     -c, --component <components>...
       27 +             Component name to also install
       28 + 
       29 +     -t, --target <targets>...
       30 +             Target name to also install
       31 + 
       32 +         --no-update-default-toolchain
       33 +             Don't update any existing default toolchain after install
       34 + 
       35 +         --no-modify-path
       36 +             Don't configure the PATH environment variable
       37 + 
       38 +     -h, --help
       39 +             Print help information
   4   40 | 
   5      - Options:
   6      -   -v, --verbose
   7      -           Enable verbose output
   8      -   -q, --quiet
   9      -           Disable progress output
  10      -   -y
  11      -           Disable confirmation prompt.
  12      -       --default-host <default-host>
  13      -           Choose a default host triple
  14      -       --default-toolchain <default-toolchain>
  15      -           Choose a default toolchain to install. Use 'none' to not install any toolchains at all
  16      -       --profile <profile>
  17      -           [default: default] [possible values: minimal, default, complete]
  18      -   -c, --component <components>...
  19      -           Component name to also install
  20      -   -t, --target <targets>...
  21      -           Target name to also install
  22      -       --no-update-default-toolchain
  23      -           Don't update any existing default toolchain after install
  24      -       --no-modify-path
  25      -           Don't configure the PATH environment variable
  26      -   -h, --help
  27      -           Print help
  28      -   -V, --version
  29      -           Print version
       41 +     -V, --version
       42 +             Print version information
```

Cargo.lock
```diff
@@ -120,17 +120,6 @@ dependencies = [
  "syn 2.0.15",
 ]
 
-[[package]]
```
rbtcollins, line 123: Also fixes #3443

Cargo.lock
```diff
@@ -120,17 +120,6 @@ dependencies = [
...
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
```
rbtcollins, line 141: Also fixes https://github.com/rust-lang/rustup/issues/3443

rbtcollins: oh heh already linked.

rbtcollins:
One other thing, once the tests are passing and we've figured out everything so the output won't churn on you, please also update the bootstrap shell scripts' help to match - we keep them in sync for visual consistency.

src/cli/rustup_mode.rs
```diff
@@ -5,9 +5,10 @@ use std::process;
 use std::str::FromStr;
 
 use anyhow::{anyhow, Error, Result};
+use clap::ValueEnum;
```
rbtcollins: (nit) please include this in the `{..}` of other clap imports.

src/cli/rustup_mode.rs
```diff
@@ -89,7 +90,7 @@ where
 pub fn main() -> Result<utils::ExitCode> {
     self_update::cleanup_self_updater()?;
 
-    use clap::ErrorKind::*;
+    use clap::error::ErrorKind::*;
```
rbtcollins, line 93: `clap::error::ErrorKind` is getting a bit verbose/deep. Perhaps 

```suggestion
    use clap::error::{ErrorKind::*, ErrorKind as ClapErrorKind};
```

src/cli/rustup_mode.rs
```diff
@@ -270,9 +271,9 @@ pub(crate) fn cli() -> Command<'static> {
                 .help("release channel (e.g. +stable) or custom toolchain to set override")
                 .value_parser(|s: &str| {
                     if let Some(stripped) = s.strip_prefix('+') {
-                        ResolvableToolchainName::try_from(stripped).map_err(|e| clap::Error::raw(clap::ErrorKind::InvalidValue, e))
+                        ResolvableToolchainName::try_from(stripped).map_err(|e| clap::Error::raw(clap::error::ErrorKind::InvalidValue, e))
```
rbtcollins: perhaps import the clap error types locally to this function too, then shorten these to

```suggestion
                        ResolvableToolchainName::try_from(stripped).map_err(|e| raw(InvalidValue, e))
```

rbtcollins:
The changes look sane, though I'm no clap expert, so I defer to you on all that!

The only caveat that isn't looking ok is the removal of the rustup version details from the top of the help. Its not necessarily a show stopper, but it does make me nervous, sometimes we have behavioural expectations in the oddest places.

If its easy to re-instate that, that would be great. Otherwise if we can check that the version shows up *somewhere* in the help, that would be sufficient.

src/cli/rustup_mode.rs
djc: Done.

djc:
Addressed your suggestions, including adding back the version to the `rustup-init` help.

samchouse:
Hey @djc, I've tried this PR out as you suggested on #3156 and completions are completely broken now on ZSH.
```
rustup on  clap-4 is 📦 v1.26.0 via 🦀 v1.71.1 
❯ rustup
_rustup:850: parse error near `\n'
_rustup:850: parse error near `\n'
_rustup:850: parse error near `\n'
```

djc:
@Xenfo did you reinstall the completions? I personally don't have any experience with shell completion setup stuff, but all I did here was change a version number for the `clap_complete` crate.

samchouse:
Yes I did, `home/bin/rustup completions zsh >> ~/.oh-my-zsh/completions/_rustup`.

samchouse:
Ignore what I said, looks like it was appending to the completion file. That issue is fixed but the issue from #3156 is still present.

samchouse:
It looks like it works after providing 2 main commands.
image

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
```diff
@@ -1,48 +1,13 @@
 bin.name = "rustup"
 args = ["-v"]
 status.code = 1
-stdout = """
-rustup [..]
-The Rust toolchain installer
+stdout = ""
+stderr = """
+error: error: 'rustup' requires a subcommand but one was not provided
```
djc: @rbtcollins any clue why the `[..]` is not doing the right thing here?

samchouse:
I've tracked the issue down to https://github.com/rust-lang/rustup/blob/master/src/cli/rustup_mode.rs#L268-L278. Removing that and https://github.com/rust-lang/rustup/blob/master/src/cli/rustup_mode.rs#L147-L149 causes completions to work as expected.

Edit: it's because of the argument before the command, it causes completions not to know if the first command was specified for the toolchain or as an actual command.

tests/suite/cli_ui.rs
```diff
@@ -55,6 +55,9 @@ fn rustup_init_ui_doc_text_tests() {
         .unwrap();
 
         // Make sure that the help output of rustup-init and rustup-init.sh are the same.
-        assert_eq!(rustup_init_help_std_out, rustup_init_sh_help_std_out)
+        assert_eq!(
+            rustup_init_help_std_out.stdout.unwrap(),
+            rustup_init_sh_help_std_out.stdout.unwrap()
+        )
```
rbtcollins, line 65: I have no objection here, but I'm curious why the change

djc: Just trying to reduce the noise from this. We should probably use something like the similar-asserts crate to make the errors from this less inscrutable.

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
rbtcollins: This behaviour is surprising. `rustup -v` should print the version and basic help, not an error about subcommand selection.

tests/suite/cli_ui.rs
rbtcollins: Thats a great suggestion. Feel free to add it in incrementally where you need it.

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
djc: Good catch, https://github.com/clap-rs/clap/discussions/5076.

rami3l: I agree with @epage's point in https://github.com/clap-rs/clap/discussions/5076#discussioncomment-6752420 and I don't even know what `rustup -v` is supposed to do personally... 👀

epage: If this test had instead used `-y` or `--no-modify-path`, would the answer be any different of what this should do?

rbtcollins: This is a new test added during the clap-3 migration (specifically if I remember correctly during the upgrade aftermath when we got folk to test and report on bugs).

This test was added in https://github.com/rust-lang/rustup/pull/3291 referencing https://github.com/rust-lang/rustup/issues/3290.

I think it is reasonable to expect that `rustup` on its own with only globally scoped, debug-affecting flags like `-v` or `-q` should behave the same as only `rustup` - or perhaps output something more or less such as startup debug messages.

`-y` and `--no-modify-path` are install-mode specific flags, not global.

rami3l: @rbtcollins The point here is, as the code is currently written, `rustup` in isolation has no clear meaning but a fallback behavior.

It is possibly better to make a default (but hidden) subcommand to just print the help string... But this doesn't seem supported by clap right now.

rbtcollins: @rami3l I don't entirely follow?

`rustup` as shipped today - not this PR, but the in use binary, has the following behaviour:
`$ rustup` -> outputs help
`$ rustup -v` -> outputs help ('verbosely' - no actual change)
`$ rustup -q` -> outputs help ('quietly' - no actual change)

This PR changes that to an error message that is somewhat inscrutable and certainly doesn't guide users to realise that they need to remove the `-v`/`-q`. 

So what options do we have? We could:

- stay on clap 3.
  -  Pros: UI stays the same.
  -  Cons: its dead upstream as 4 is released and we still have that attr CVE.
- accept the current behaviour in this PR. 
  - Pros: we get future bug fixes from the clap developers
  - Cons: we have to deal with bug reports from confused users
- find a way to retain the existing behaviour with clap 4.
  - Pros: we get both the pros of the first two options
  - Cons: more development is needed and its not clear how to do it

rami3l: @rbtcollins I meant we should implement that help message-printing behavior **explicitly as `rustup`'s default behavior** (e.g. by calling [`print_long_help`](https://docs.rs/clap/4.3.23/clap/struct.Command.html#method.print_long_help)), rather than relying on `arg_required_else_help` whose semantics has already slightly shifted. I'm not sure if this can be done properly though, but I'll find some time to see if I can get a demo working.

rami3l: @djc Luckily, this patch does seem to work on your 0988a514f5c84cfd6d1e4c5cc0b1003b23ea5eb4 (you might need to change the return code if necessary):

```diff
diff --git a/src/cli/rustup_mode.rs b/src/cli/rustup_mode.rs
index 001f86b7..644ecd6b 100644
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -243,7 +243,10 @@ pub fn main() -> Result<utils::ExitCode> {
             }
             _ => unreachable!(),
         },
-        None => unreachable!(),
+        None => {
+            cli().print_long_help()?;
+            utils::ExitCode(0)
+        }
     })
 }
 
@@ -253,8 +256,7 @@ pub(crate) fn cli() -> Command {
         .about("The Rust toolchain installer")
         .before_help(format!("rustup {}", common::version()))
         .after_help(RUSTUP_HELP)
-        .subcommand_required(true)
-        .arg_required_else_help(true)
+        .subcommand_required(false)
         .arg(
             verbose_arg("Enable verbose output"),
         )
```

<details>
<summary>Example output</summary>

```sh
> ./target/debug/rustup -v
verbose: read metadata version: '12'
rustup 1.26.0+100 (0988a514f 2023-08-16) dirty 1 modification

The Rust toolchain installer

Usage: rustup [OPTIONS] [+toolchain] [COMMAND]

Commands:
  show         Show the active and installed toolchains or profiles
  update       Update Rust toolchains and rustup
  check        Check for updates to Rust toolchains and rustup
  default      Set the default toolchain
  toolchain    Modify or query the installed toolchains
  target       Modify a toolchain's supported targets
  component    Modify a toolchain's installed components
  override     Modify directory toolchain overrides
  run          Run a command with an environment configured for a given toolchain
  which        Display which binary will be run for a given command
  doc          Open the documentation for the current toolchain
  man          View the man page for a given command
  self         Modify the rustup installation
  set          Alter rustup settings
  completions  Generate tab-completion scripts for your shell
  help         Print this message or the help of the given subcommand(s)

Arguments:
  [+toolchain]
          release channel (e.g. +stable) or custom toolchain to set override

Options:
  -v, --verbose
          Enable verbose output

  -q, --quiet
          Disable progress output

  -h, --help
          Print help

  -V, --version
          Print version

Discussion:
    Rustup installs The Rust Programming Language from the official
    release channels, enabling you to easily switch between stable,
    beta, and nightly compilers and keep them updated. It makes
    cross-compiling simpler with binary builds of the standard library
    for common platforms.

    learn Rust.
```
</details>

djc: @rami3l thanks, looks great! I changed the `ExitCode` to 1 since I think we should still technically signal this as an error, and pushed the branch again to see if it can pass CI this time.

djc: > If this test had instead used `-y` or `--no-modify-path`, would the answer be any different of what this should do?

IMO yes, in that case it should also act according to the "missing subcommand" logic.

rami3l: @djc
Looking at your changes, I think you probably have forgotten to update [tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml](https://github.com/rust-lang/rustup/pull/3444/files#diff-7b4864dbf5072fa19eca9a80e540c15591f0c3dc41ac47823f662c23a1554af2)...

Also, `.subcommand_required(false)` at the top level is required to make this work (it could be `false` by default, I don't remember exactly,  but you cannot set it to `true` anyway).

djc: Right, I fixed those issues but the test still fails, because apparently a `verbose: read metadata version '12'` is coming through in the output in some way that trycmd doesn't deal with well (as in, the location of the notification seems to be shifting around in the output).

rami3l: @djc That's interesting. I can reproduce this (see the output above) but I have no idea what this is about, sorry :/

djc: Just generic output from `-v` at initialization, I think. CI fails in yet another way:

https://github.com/rust-lang/rustup/actions/runs/5927865297/job/16072261180?pr=3444#step:17:632

epage: > `rustup` as shipped today - not this PR, but the in use binary, has the following behaviour: `$ rustup` -> outputs help `$ rustup -v` -> outputs help ('verbosely' - no actual change) `$ rustup -q` -> outputs help ('quietly' - no actual change)
> 
> This PR changes that to an error message that is somewhat inscrutable and certainly doesn't guide users to realise that they need to remove the `-v`/`-q`.

imo this analysis is backwards.  I do not see why a user would pass in `rustup -v` and expect help output.  To me, this feels like we are overall making a mountain out of a molehill and putting in more thought and code design work to something that 99% won't even notice or care about.

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
```diff
@@ -1,41 +1,33 @@
 bin.name = "rustup"
 args = ["-v"]
```
rbtcollins:
```suggestion
args = ["-q"]
```

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
rbtcollins: The metadata output is the result of -v having effect; I suggest changing to -q for the test, which is effectively guaranteed to never be adding more output.

@epage I can appreciate the sentiment about engineering effort / reward. But it seems to me that you take issue with the value judgement not the analysis - you haven't disputed the historical behaviour, argued that the error message is easily understood or proposed alternative ways forward. 

So if the question is about the relative merits of option 2 vs option 3, what I can say right now is that `rustup` is under-resourced in terms of human capacity. And triaging and improving non-functional bugs takes quite a bit of time. We could of course say 'if it isn't a functional problem, we will just ignore/close such bugs'.... but I don't feel that that would be a good way of upholding the excellent work that preceeded my joining the project a few years back.

Perceptions of quality in software often hinge on non-happy-path, non-functional-requirements. We know that at least some people went and tried this path, or we wouldn't have the additional test added to cover it. With it covered, we won't get another bug report about it, which is good. 

And @djc seems to have quite rapidly figured out and solved the problem, so thats nice.

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
djc: Ah, good idea.

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
djc: Well, CI is still failing, which is not so nice.

epage: I take issue with the analysis itself but since it started at the root of the analysis, I didn't go further into it.  I didn't just remove `SubcommandRequiredElseHelp` for code size and binary size.

epage: And for a project that is under resourced, flowing against the stream seems like it'd be the opposite of what you'd want to be doing.  I'm not saying this in terms of accepting bugs but of practices

rustup-init.sh
```diff
@@ -26,47 +26,36 @@ RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
 usage() {
     cat <<EOF
 rustup-init 1.26.0 (577bf51ae 2023-04-05)
```
epage, line 28: > The only caveat that isn't looking ok is the removal of the rustup version details from the top of the help. Its not necessarily a show stopper, but it does make me nervous, sometimes we have behavioural expectations in the oddest places.
>
> If its easy to re-instate that, that would be great. Otherwise if we can check that the version shows up somewhere in the help, that would be sufficient.

If people are relying on parsing `--help` output, we will break them; there are no guarentees.

We intentionally dropped this to recover vertical space.  If you want precedence / consistency with other rust commands, `cargo` does not include the version.

src/cli/rustup_mode.rs
```diff
@@ -269,10 +272,11 @@ pub(crate) fn cli() -> Command<'static> {
...
                 .help("release channel (e.g. +stable) or custom toolchain to set override")
                 .value_parser(|s: &str| {
+                    use clap::{Error, error::ErrorKind};
                     if let Some(stripped) = s.strip_prefix('+') {
-                        ResolvableToolchainName::try_from(stripped).map_err(|e| clap::Error::raw(clap::ErrorKind::InvalidValue, e))
+                        ResolvableToolchainName::try_from(stripped).map_err(|e| Error::raw(ErrorKind::InvalidValue, e))
                     } else {
-                        Err(clap::Error::raw(clap::ErrorKind::InvalidSubcommand, format!("\"{s}\" is not a valid subcommand, so it was interpreted as a toolchain name, but it is also invalid. {TOOLCHAIN_OVERRIDE_ERROR}")))
+                        Err(Error::raw(ErrorKind::InvalidSubcommand, format!("\"{s}\" is not a valid subcommand, so it was interpreted as a toolchain name, but it is also invalid. {TOOLCHAIN_OVERRIDE_ERROR}")))
                     }
```
epage, line 280: Honestly, I'm surprised this works.  The intention was to support any user errors.  I'm surprised that clap errors render well when nested.  I wouldn't be surprised if we accidentally break this

epage:
> I've tracked the issue down to https://github.com/rust-lang/rustup/blob/master/src/cli/rustup_mode.rs#L268-L278. Removing that and https://github.com/rust-lang/rustup/blob/master/src/cli/rustup_mode.rs#L147-L149 causes completions to work as expected.
> 
> Edit: it's because of the argument before the command, it causes completions not to know if the first command was specified for the toolchain or as an actual command.

FYI [completest](docs.rs/completest/) now exists for testing of completions.  clap has adopted it to test the completion generators.  I'd like to adopt it in cargo to test our hand written completions but haven't yet.

Also, clap is (slowly) working towards completions being implemented mostly in Rust which should offer more correctness and flexibility

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
rbtcollins: I think it depends on where we think our time is going no? We're getting quite meta at this point though.

rbtcollins:
failure windows dev:
```
---- expected: stderr
++++ actual:   stderr
   1    1 | rustup [..]
   2    2 | 
   3    3 | The Rust toolchain installer
   4    4 | 
  5      - Usage: rustup[EXE] [OPTIONS] [+toolchain] [COMMAND]
   5 + Usage: rustup [OPTIONS] [+toolchain] [COMMAND]
   6    6 | 
   7    7 | Commands:
   8    8 |   show         Show the active and installed toolchains or profiles
   9    9 |   update       Update Rust toolchains and rustup
  10   10 |   check        Check for updates to Rust toolchains and rustup
          ⋮
  47   47 |     for common platforms.
  48   48 | 
  49   49 |     If you are new to Rust consider running `rustup doc --book` to
  50   50 |     learn Rust.
  51   51 | 
```

mac is a flake.

Linux is 
```
---- expected: stderr
++++ actual:   stderr
   1      - rustup [..]
   2      - 
   ...
  51      - 
        1 + error: could not create home directory: '/.rustup': Permission denied (os error 13)
```

tests/suite/cli-ui/rustup/rustup_only_options_stdout.toml
```diff
@@ -1,41 +1,50 @@
...
-    check          Check for updates to Rust toolchains and rustup
-    default        Set the default toolchain
-    toolchain      Modify or query the installed toolchains
-    target         Modify a toolchain's supported targets
-    component      Modify a toolchain's installed components
-    override       Modify directory toolchain overrides
-    run            Run a command with an environment configured for a given toolchain
-    which          Display which binary will be run for a given command
-    doc            Open the documentation for the current toolchain
+Usage: rustup[EXE] [OPTIONS] [+toolchain] [COMMAND]
```
rbtcollins: I'm not sure why, but the help for this particular case seems to be skipping .exe on Windows. Try this suggestion?
```suggestion
Usage: rustup [OPTIONS] [+toolchain] [COMMAND]
```

src/cli/rustup_mode.rs
rbtcollins: What do you suggest we do instead?

rbtcollins:
At a guess I think the linux error is coming from https://github.com/rust-lang/rustup/blob/2fa788a74978bb4ec0054708db55d1fbb73a6053/src/config.rs#L198-L199 - that is `from_env()` is being called for some reason.

The test suite that runs the trycmd stuff isn't setting up isolated environments (HOME temp dir, environment variables) so this looks like an interaction with the global environment, and could well be super racy.

That said the reason this is triggering that code path isn't too obvious to me. I'll see if get time this weekend to checkout the branch and poke around.

edit: noticed it is `'/.rustup'` so this is running with HOME=`/` I think. Perhaps trycmd does that automatically.

src/cli/rustup_mode.rs
epage: Use any normal error type that implements `Error`

djc:
All in all I feel like these trycmd tests are superflaky and hard to work with.

rbtcollins:
> All in all I feel like these trycmd tests are superflaky and hard to work with.

They are certainly sensitive to changes, possibly because we've only just added them this year and not had much time to figure out where we want our sensitivity and where not.

I'm not sure that they are flaky: I don't recall a single CI run that failed nondeterministically because of them.

src/cli/rustup_mode.rs
rbtcollins: Ok, and - just checking for complete clarity - so I don't need to use clap error types in those helpers?

epage: Correct.  If curious, [this](https://docs.rs/clap/latest/clap/builder/trait.TypedValueParser.html#impl-TypedValueParser-for-F) is the relevant trait implementation you need to match

djc:
Sure, I guess "overly sensitive" is more accurate than flaky.

rbtcollins:
I've fixed the failure on windows.

On linux I've validated that it is a hermiticity issue leading to the test error. Prepping a separate patch for that.

rbtcollins:
https://github.com/rust-lang/rustup/pull/3470 should fix the isolation failure. I haven't figured out yet what this patch is doing that introduces the assertion of settings.toml, though probably we should look into that - writing to the settings file when nothing is being explicitly changed is a bit odd.

rbtcollins (approved):

PR #3446, opened 2023-08-16
Flatten Cargo workspace

djc:
I generally prefer flat workspaces, that is, no packages nested within other Cargo packages. I find this clearer, since the packages end up not really having a hierarchy in how Cargo actually treats them, and it separates the root of repository files cleanly from the rustup crate's files.


rami3l:
A possibly silly question: will this force most current PRs to be rebased?

djc:
> A possibly silly question: will this force most current PRs to be rebased?

Not silly at all. I'm not sure exactly what you're asking. Is there a bors workflow here that will ping every PR to rebase if conflicts happen? I think this should not cause many conflicts because Git can usually merge across renames. I would mainly expect issues for changes to `Cargo.toml`, since it got split in this PR.

rami3l:
> Is there a bors workflow here that will ping every PR to rebase if conflicts happen?

Looking at the history of this repo, there doesn't seem to be one, unfortunately.

> I think this should not cause many conflicts because Git can usually merge across renames. I would mainly expect issues for changes to `Cargo.toml`, since it got split in this PR.

Oh I see, thanks for the clarification :)

djc:
FWIW, happy to dig into the CI failures here but would like to first hear a pronouncement on whether this would be accepted if it were to pass CI.

rbtcollins:
We don't use bors in rustup; there's been talk of setting it up but no-one has had the time.

That said, I'd rather not do this change at the moment, for two reasons.

Firstly, rustup is an application with exactly one main output; the other crates are internal concerns, not published to crates.io, and really are subordinate from a human developer perspective.

Secondly, we're overdue to take a pause and figure out what crates we want. `rustup-macros` is only a separate crate because that is a requirement from rust itself for that sort of functionality; download is a crate with an awkward interface, and a while back we had a cli crate that we've removed - and doing so has enabled order of magnitude improvements to the test suite. I can sketch out the broad concerns without too much cognitive load if digging into this is important for you, but if its not, this would probably be better done by a larger team once we've grown them.

djc:
Feel free to close this if you don't want to spend more time on this, but...

Seems to me that flattening is useful even if we only have the two rustup and rustup-macros crate, and it doesn't seem likely that the requirement to have a separate crate for procedural macros is going away any time soon. Also, this PR took me like 15 minutes so far -- it's not a big investment in my mind to make further changes later. So even if we inline the download crate again (which sounds like it might make sense going forward), I still think this would be a nice change.

rbtcollins:
I'll close this then, but thank you for raising the discussion!

PR #3447, opened 2023-08-16
Avoid installing protoc for most CI workflows

djc:
It seems like it should only be needed for the `all-features` workflow, which enables `otel` via `--all-features`.

Because we get rate-limited by something inside the setup-protoc Action, this is a fairly frequent source of spurious CI failures.

This was introduced in #3340 according to git blame.

### DIFF
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -131,10 +131,6 @@ jobs:
           echo "DOCKER=$DOCKER" >> $GITHUB_ENV
       - name: Fetch the docker
         run: bash ci/fetch-rust-docker.bash "${TARGET}"
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v1
-        with:
-          version: '3.x'
       - name: Maybe build a docker from there
         run: |
           if [ -f "ci/docker/$DOCKER/Dockerfile" ]; then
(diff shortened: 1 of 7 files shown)


djc:
Seems to pass CI (macOS still going right now) except for the previously failing Linux armv7-unknown-linux-gnueabihf workflows.

rbtcollins:
I hadn't noticed rate limiting being at the heart of the failures - do you have an example we can refer to?

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -131,10 +131,6 @@ jobs:
           echo "DOCKER=$DOCKER" >> $GITHUB_ENV
       - name: Fetch the docker
         run: bash ci/fetch-rust-docker.bash "${TARGET}"
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v1
-        with:
-          version: '3.x'
```
rbtcollins, line 137:
```
  with:
    repo-token: ${{ secrets.GITHUB_TOKEN }}
```

Seems like the canonical approach to avoid rate limiting, from the setup-protoc repo.

rami3l:
> I hadn't noticed rate limiting being at the heart of the failures - do you have an example we can refer to?

@rbtcollins At least I can confirm that this is not uncommon:
https://github.com/rust-lang/rustup/actions/runs/5851656996/job/15862744488

ci/actions-templates/linux-builds-template.yaml
djc: Added this in another commit to `all-features.yaml`.

PR #3449, opened 2023-08-17
Refactor: Use download_cfg.notify_handler in update()

rbtcollins:
All calls to update() were just passing it in already.


djc (approved):
Nice!

PR #3451, opened 2023-08-17
Fix handling of async tests

rbtcollins:
The ident parsing was broken.

### DIFF
--- a/rustup-macros/src/lib.rs
+++ b/rustup-macros/src/lib.rs
@@ -2,7 +2,6 @@
 
 use ::quote::quote;
 use proc_macro2::TokenStream;
-use quote::format_ident;
 use syn::{parse_macro_input, parse_quote, Block, Expr, ItemFn, LitStr};
 
 /// Custom wrapper macro around `#[test]` and `#[tokio::test]`.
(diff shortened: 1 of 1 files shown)


rustup-macros/src/lib.rs
```diff
@@ -80,8 +79,10 @@ pub fn unit_test(
 
 fn test_inner(mod_path: String, mut input: ItemFn) -> syn::Result<TokenStream> {
     if input.sig.asyncness.is_some() {
-        let before_ident = format_ident!("{}::before_test_async", mod_path);
-        let after_ident = format_ident!("{}::after_test_async", mod_path);
+        let before_ident = format!("{}::before_test_async", mod_path);
+        let before_ident = syn::parse_str::<Expr>(&before_ident)?;
```
djc, line 83: Shouldn't that be a `Path` rather than an `Expr`? If not, maybe give an example?

Maybe use inline format args?

rbtcollins: Expr is what works below - this part of rust isn't my forte, so if you wanted to muck around and make it more beautiful I suspect you'd get better code than I've written :).

Yes, it could use inline formatting, the format_ident macro didn't from memory, and I didn't think to change when changing.

That said there is more in the file that this is consistent with, so I'm inclined to come back later, and update the whole file.

djc: Fair enough.

djc (approved):

PR #3453, opened 2023-08-18
Refine suggestions about missing components

### LINKED ISSUE #3418: Confusing `component add` error message
### Problem

I tried to add component `miri` and received the following message:
```
$ rustup component add miri
error: component 'miri' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel '1.69.0'
If you don't need the component, you can remove it with:

    rustup component remove --toolchain 1.69.0 --target x86_64-unknown-linux-gnu miri
```
I.e., I tried to add the component (unsuccessfully), and the instructions tell me how to remove it.

### Steps

Run `rustup component add miri` in a directory where `x86_64-unknown-linux-gnu` is the default toolchain.

### Possible Solution(s)

Adjust or eliminate the irrelevant error message.

### Notes

_No response_

### Rustup version

```console
info: The currently active `rustc` version is `rustc 1.69.0 (84c898d65 2023-04-16)`
```
```


### Installed toolchains

```console
rustup home:  /home/smoelius/.rustup

installed toolchains
--------------------

stable-2020-06-18-x86_64-unknown-linux-gnu
nightly-2023-06-29-x86_64-unknown-linux-gnu
nightly-2023-07-14-x86_64-unknown-linux-gnu
bisector-nightly-2023-05-28-x86_64-unknown-linux-gnu
bpf
solana
1.59.0-x86_64-unknown-linux-gnu
1.60.0-x86_64-unknown-linux-gnu
1.66.1-x86_64-unknown-linux-gnu
1.67.1-x86_64-unknown-linux-gnu
1.68.2-x86_64-unknown-linux-gnu
1.69.0-x86_64-unknown-linux-gnu

active toolchain
----------------

1.69.0-x86_64-unknown-linux-gnu (overridden by '/home/smoelius/gh/trailofbits/audit-solana/rust-toolchain.toml')
rustc 1.69.0 (84c898d65 2023-04-16)
```
```

### PR DESCRIPTION
**Pull request by rami3l**
Closes #3418. I also cleaned up the current code a bit along the way.

cc #2355

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -43,6 +43,13 @@ static TOOLCHAIN_CHANNELS: &[&str] = &[
     r"\d{1}\.\d{1,3}(?:\.\d{1,2})?",
 ];
 
+/// Returns a error message indicating that certain [`Component`]s are missing in a toolchain distribution.
+///
+/// This message is currently used exclusively in toolchain-wide operations,
+/// otherwise [`component_unavailable_msg`](../../errors/fn.component_unavailable_msg.html) will be used.
+///
+/// # Panics
+/// This function will panic when the collection of unavailable components `cs` is empty.
 fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &str) -> String {
     let mut buf = vec![];
     let suggestion = format!("    rustup toolchain add {toolchain} --profile minimal");
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -44,45 +44,49 @@ static TOOLCHAIN_CHANNELS: &[&str] = &[
 ];
 
 fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &str) -> String {
-    assert!(!cs.is_empty());
     let mut buf = vec![];
     let suggestion = format!("    rustup toolchain add {toolchain} --profile minimal");
     let nightly_tips = "Sometimes not all components are available in any given nightly. ";
 
-    if cs.len() == 1 {
-        let _ = writeln!(
-            buf,
-            "component {} is unavailable for download for channel '{}'",
-            &cs[0].description(manifest),
-            toolchain,
-        );
+    match cs {
+        [] => panic!("`components_missing_msg` should not be called with an empty collection of unavailable components"),
+        [c] => {
+            let _ = writeln!(
+                buf,
+                "component {} is unavailable for download for channel '{}'",
+                c.description(manifest),
+                toolchain,
+            );
 
-        if toolchain.starts_with("nightly") {
-            let _ = write!(buf, "{nightly_tips}");
+            if toolchain.starts_with("nightly") {
+                let _ = write!(buf, "{nightly_tips}");
+            }
+
+            let _ = write!(
+                buf,
+                "If you don't need the component, you could try a minimal installation with:\n\n{suggestion}"
+            );
         }
+        cs => {
+            let cs_str = cs
+                .iter()
+                .map(|c| c.description(manifest))
+                .collect::<Vec<_>>()
+                .join(", ");
+            let _ = write!(
+                buf,
+                "some components unavailable for download for channel '{toolchain}': {cs_str}"
+            );
 
-        let _ = write!(
-            buf,
-            "If you don't need the component, you could try a minimal installation with:\n\n{suggestion}"
-        );
-    } else {
-        let cs_str = cs
-            .iter()
-            .map(|c| c.description(manifest))
-            .collect::<Vec<_>>()
-            .join(", ");
-        let _ = write!(
-            buf,
-            "some components unavailable for download for channel '{toolchain}': {cs_str}"
-        );
+            if toolchain.starts_with("nightly") {
+                let _ = write!(buf, "{nightly_tips}");
+            }
 
-        if toolchain.starts_with("nightly") {
-            let _ = write!(buf, "{nightly_tips}");
+            let _ = write!(
+                buf,
+                "If you don't need the components, you could try a minimal installation with:\n\n{suggestion}"
+            );
         }
-        let _ = write!(
-            buf,
-            "If you don't need the components, you could try a minimal installation with:\n\n{suggestion}"
-        );
     }
 
     String::from_utf8(buf).unwrap()
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -43,6 +43,14 @@ static TOOLCHAIN_CHANNELS: &[&str] = &[
     r"\d{1}\.\d{1,3}(?:\.\d{1,2})?",
 ];
 
+const TOOLSTATE_MSG: &str =
+    "If you require these components, please install and use the latest successful build version,\n\
+     which you can find at <https://rust-lang.github.io/rustup-components-history>.\n\nAfter determining \
+     the correct date, install it with a command such as:\n\n    \
+     rustup toolchain install nightly-2018-12-27\n\n\
+     Then you can use the toolchain with commands such as:\n\n    \
+     cargo +nightly-2018-12-27 build";
+
 /// Returns a error message indicating that certain [`Component`]s are missing in a toolchain distribution.
 ///
 /// This message is currently used exclusively in toolchain-wide operations,
@@ -71,7 +79,7 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
 
             let _ = write!(
                 buf,
-                "If you don't need the component, you could try a minimal installation with:\n\n{suggestion}"
+                "If you don't need the component, you could try a minimal installation with:\n\n{suggestion}\n\n{TOOLSTATE_MSG}"
             );
         }
         cs => {
@@ -91,7 +99,7 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
 
             let _ = write!(
                 buf,
-                "If you don't need the components, you could try a minimal installation with:\n\n{suggestion}"
+                "If you don't need the components, you could try a minimal installation with:\n\n{suggestion}\n\n{TOOLSTATE_MSG}"
             );
         }
     }
--- a/src/dist/manifestation/tests.rs
+++ b/src/dist/manifestation/tests.rs
@@ -772,8 +772,17 @@ fn unavailable_component() {
             )
             .unwrap_err();
             match err.downcast::<RustupError>() {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by djc — APPROVED** [review]

File: src/errors.rs [hunk 1]
```diff
@@ -171,57 +171,69 @@ fn remove_component_msg(cs: &Component, manifest: &Manifest, toolchain: &str) ->
...
-        let _ = writeln!(
-            buf,
-            "component {} is unavailable for download for channel '{}'",
-            &cs[0].description(manifest),
-            toolchain,
-        );
-        if toolchain.starts_with("nightly") {
-            let _ = write!(
+    match cs {
+        [] => panic!("`component_unavailable_msg` should not be called with an empty collection of unavailable components"),
```
rbtcollins, line 152: This has added a new panic to code that couldn't panic before. For `rustup` panics are quite a concern, though in an error path is -perhaps- acceptable.

In this case I can see there is only one construction point for the error value that formats with this helper. Please add a doc comment to the variant that the helper will panic; we can look at using https://docs.rs/nonempty/latest/nonempty/ or similar later.

rami3l: @rbtcollins IMHO it didn't. This line corresponds to the original `assert!(!cs.is_empty());` line which is also a panic point if `cs` matches `[]`.
... but I'd be happy to change this for the better by, for example, adding a "panic" section in the docs.

rami3l: ... but I'd be happy to change this for the better by, for example, adding a "panic" section in the docs.

rbtcollins: ah thats fair! I think we should still add a doccomment to the variant, because this is a timebomb otherwise.

src/errors.rs
```diff
@@ -171,57 +171,69 @@ fn remove_component_msg(cs: &Component, manifest: &Manifest, toolchain: &str) ->
...
+                    buf,
+                    "Sometimes not all components are available in any given nightly. "
+                );
+            }
+            _ = write!(
+                    buf,
+                    "If you are trying to update an existing toolchain and no longer need these components,\n\
+                    you might want to remove them first with:\n\n{remove_msg}\n\n{TOOLSTATE_MSG}",
+            );
+        }
```
rbtcollins, line 197: its quite hard to determine what you've actually changed here - the change combines a refactoring (fine), and new message  logic (also fine), but even though I've looked in:
- the PR message
- the commit messages
- test changes (none)
I can't tell what the actual intent is without manually stepping through in my head every possible case.

It would be a lot easier to reason about if the refactoring and the functional changes were separate (e.g. two commits in one PR - like you did for the other module that you refactored).

I realise that that isn't always possible, and if it wasn't here then I guess its as good as we can get :).

The other norm we have is to ensure test coverage for functional behaviours. We're in the process of moving tests closer to the domain for the test, so looking at the existing test coverage isn't necessarily the best way to figure out where tests should go - but its ok to put them in the blackbox test suite under `tests/` and we can move them later.

rami3l: > It would be a lot easier to reason about if the refactoring and the functional changes were separate (e.g. two commits in one PR - like you did for the other module that you refactored).

Yes, I am aware of that. I will leave the current commit as it is, but I will definitely improve this from now on.

> The other norm we have is to ensure test coverage for functional behaviours. We're in the process of moving tests closer to the domain for the test, so looking at the existing test coverage isn't necessarily the best way to figure out where tests should go - but its ok to put them in the blackbox test suite under `tests/` and we can move them later.

I'd be glad to add blackbox tests for this one (which will be done in another commit), please bear with me 🙏

rami3l:
@rbtcollins Looking at the blackbox test suite, it seems to me that this fix is more than what it seems to be.

In toolchain **updates**, `component_unavailable_msg`  

https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/errors.rs#L173

does get called, but the generated error then gets **caught and re-thrown**

https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/dist/dist.rs#L938-L946

to `components_missing_msg`

https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/dist/dist.rs#L45

, which instead suggests using a `minimal` profile, so the original suggestion in the former (uninstall the component in question) might not be applicable at all anymore, as the error message from the former can only be seen in some edge cases (like adding `miri` to `stable`, but not upgrading to a new `nightly` which lacks `rls`).

To move this fix forward, I could simply remove the suggestion in `component_unavailable_msg`, however, I want to know more about the semantic differences between these two functions in order to make the final decision.

For example, since this help message is very well-written, I wish to be able to see it in toolchain updates as well:

https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/errors.rs#L22-L28

rami3l:
Note: `dist::manifestation::tests::unavailable_component` and friends test `rustup::dist::manifestation::Manifestation::update` directly (by checking the very `rustup component remove` suggestion), which happens before the catch-rethrow dance in `rustup::dist::dist::try_update_from_dist_`, so these tests are expected to fail at the moment.

... and they are not accurate anyway since the string form of those errors (before the catch-rethrow dance) is NOT ALWAYS visible to end users, which might cause further confusions for maintainers later on.

rami3l:
I've yet to see any use case or test case that disobeys the observation stated in https://github.com/rust-lang/rustup/pull/3453/commits/6b5cb9f904ce9acdacc416848f5ba050a0c6849f 's commit message.

Therefore, I think by now it is reasonable to do the following:

- [x] Add a new blackbox test case to ensure that #3418 does get fixed.
- [x] Move `TOOLSTATE_MSG` to `dist` to serve toolchain-wide operations (since it also talks about toolchain changes) by appending it to the end of `component_missing_msg`'s output.

src/dist/manifestation/tests.rs
```diff
@@ -992,13 +1019,24 @@ fn unavailable_components_is_target() {
...
+                    toolchain,
+                }) => {
+                    assert_eq!(toolchain, "nightly");
+                    let descriptions = components
+                        .iter()
+                        .map(|c| c.description(&manifest))
+                        .collect::<Vec<_>>();
+                    assert_eq!(
+                        descriptions,
+                        [
```
rami3l, line 1034: This actually exposes an error in the original test case, which only tested the INCLUSION rather than the whole collection.

src/dist/manifestation/tests.rs
```diff
@@ -992,13 +1019,24 @@ fn unavailable_components_is_target() {
...
+                            "'rust-std' for target 'x86_64-apple-darwin'",
+                            "'rust-std' for target 'i686-apple-darwin'",
+                            "'rust-std' for target 'i686-unknown-linux-gnu'"
+                        ]
+                    );
```
rami3l, line 1039: Here, I am changing the tests so that they no longer rely on the textual representation of the error.

This has actually exposed a mistake in the original test case, which only tested the INCLUSION rather than the whole collection.

src/errors.rs
rami3l: @rbtcollins The functional test changes now reflect this PR's original intention.

rbtcollins:
> I've yet to see any use case or test case that disobeys the observation stated in [6b5cb9f](https://github.com/rust-lang/rustup/commit/6b5cb9f904ce9acdacc416848f5ba050a0c6849f) 's commit message.
> 
> Therefore, I think by now it is reasonable to do the following:
> 
>     * [x]  Add a new blackbox test case to ensure that [Confusing `component add` error message #3418](https://github.com/rust-lang/rustup/issues/3418) does get fixed.
> 
>     * [x]  Move `TOOLSTATE_MSG` to `dist` to serve toolchain-wide operations (since it also talks about toolchain changes) by appending it to the end of `component_missing_msg`'s output.

seems reasonable to me

rami3l:
> > I've yet to see any use case or test case that disobeys the observation stated in [6b5cb9f](https://github.com/rust-lang/rustup/commit/6b5cb9f904ce9acdacc416848f5ba050a0c6849f) 's commit message.
> > Therefore, I think by now it is reasonable to do the following:
> > ```
> > * [x]  Add a new blackbox test case to ensure that [Confusing `component add` error message #3418](https://github.com/rust-lang/rustup/issues/3418) does get fixed.
> > 
> > * [x]  Move `TOOLSTATE_MSG` to `dist` to serve toolchain-wide operations (since it also talks about toolchain changes) by appending it to the end of `component_missing_msg`'s output.
> > ```
> 
> seems reasonable to me

@rbtcollins  ... and those two tasks have been done, together with a new black box test case for the latter.
Would you mind reviewing my changes?

src/dist/dist.rs
```diff
@@ -43,6 +43,14 @@ static TOOLCHAIN_CHANNELS: &[&str] = &[
     r"\d{1}\.\d{1,3}(?:\.\d{1,2})?",
 ];
 
+const TOOLSTATE_MSG: &str =
```
djc, line 46: The commit message mentions to `Move` this, but I don't see it being removed anywhere. Is that an oversight?

src/dist/dist.rs
```diff
@@ -43,46 +43,57 @@ static TOOLCHAIN_CHANNELS: &[&str] = &[
...
+        [c] => {
+            _ = writeln!(
+                buf,
+                "component {} is unavailable for download for channel '{}'",
+                c.description(manifest),
+                toolchain,
+            );
+
+            if toolchain.starts_with("nightly") {
+                _ = write!(buf, "{nightly_tips}");
```
djc: Nit: not a fan of this change to remove the `let` keywords. IMO the `let _ =` is more conventional.

djc (commented):
Please keep unrelated changes in separate commits, makes it much easier to review.

src/dist/dist.rs
rami3l: Will do.

src/dist/dist.rs
rami3l: Sorry for previously being unclear about this. This is an old PR and I've learned a lot on how to separate a big patch into relevant small commits since then.

The message was previously deleted in https://github.com/rust-lang/rustup/pull/3453/commits/9107c0235ceeb1b227e333495903b78acb135499. Maybe it's better to say that I "recycled" it here since this error message has been used in a wrong module for another kind of error (https://github.com/rust-lang/rustup/pull/3453#issuecomment-1685269047).

rami3l: @djc I rewrote the history to make the move actually appear in the last commit. Sorry about the inconvenience as I'm still improving my collaboration skills!

djc (approved):

PR #3462, opened 2023-08-22
[WIP] Debugging silent download failures on AArch64 Windows

djc:


rami3l:
@djc Is this PR still relevant? Because now ARM64 runners can run Windows and I'd like to enable the full test suite for it just like the rest of tier 2 :)

djc:
No clue...

rami3l:
> No clue...

No worries! It looks like #4290 is doing great though... We can always come back if the problem resurfaces.

PR #3463, opened 2023-08-23
Fix update Rust crate winreg to 0.51

rami3l:
Supercedes #3456.

> * [ ]  If you want to rebase/retry this PR, check this box

I cannot tick that box so I had to make another PR instead.
cc @rbtcollins

### DIFF
--- a/tests/suite/cli_paths.rs
+++ b/tests/suite/cli_paths.rs
@@ -362,7 +362,7 @@ mod windows {
     fn install_uninstall_affect_path() {
         clitools::test(Scenario::Empty, &|config| {
             with_saved_path(&mut || {
-                let path = format!("{:?}", config.cargodir.join("bin").to_string_lossy());
+                let path = config.cargodir.join("bin").display().to_string();
 
                 config.expect_ok(&INIT_NONE);
                 assert!(
--- a/tests/suite/cli_paths.rs
+++ b/tests/suite/cli_paths.rs
@@ -362,22 +362,19 @@ mod windows {
     fn install_uninstall_affect_path() {
         clitools::test(Scenario::Empty, &|config| {
             with_saved_path(&mut || {
-                let path = config.cargodir.join("bin").display().to_string();
+                let cfg_path = config.cargodir.join("bin").display().to_string();
+                let get_path_ = || get_path().unwrap().unwrap().to_string();
 
                 config.expect_ok(&INIT_NONE);
                 assert!(
-                    get_path()
-                        .unwrap()
-                        .unwrap()
-                        .to_string()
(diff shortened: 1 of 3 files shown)


tests/suite/cli_paths.rs
```diff
@@ -362,7 +362,8 @@ mod windows {
     fn install_uninstall_affect_path() {
         clitools::test(Scenario::Empty, &|config| {
             with_saved_path(&mut || {
-                let path = format!("{:?}", config.cargodir.join("bin").to_string_lossy());
+                let path = format!("{:?}", config.cargodir.join("bin").to_string_lossy())
```
djc: Can you give a bit or context? How did the winreg bump change the `config.cargodir` value and why should we fix it in this test instead of at some other level?

rami3l: @djc This test depends on https://togithub.com/gentoo90/winreg-rs/issues/52 which is a breaking change in formatting. The value itself should remain the same (cc https://github.com/rust-lang/rustup/pull/3456#issuecomment-1685202827).

djc: So should we switch to formatting with `{}` instead of `{:?}` like upstream, instead? Even after looking at the upstream issue, this fix looks more like a workaround to me.

rami3l: @djc Indeed, I have no idea why the previous code was printing a string with `Debug`.

rami3l: @djc I changed that to use `config.cargodir.join("bin").display()` instead. The point here is that a path cannot be `Display`ed directly, and if I use `Debug` it'll have double backslashes.

tests/suite/cli_paths.rs
```diff
@@ -362,22 +362,19 @@ mod windows {
     fn install_uninstall_affect_path() {
         clitools::test(Scenario::Empty, &|config| {
             with_saved_path(&mut || {
-                let path = format!("{:?}", config.cargodir.join("bin").to_string_lossy());
+                let cfg_path = config.cargodir.join("bin").display().to_string();
+                let get_path_ = || get_path().unwrap().unwrap().to_string();
```
djc, line 366: This fix looks good to me, thanks for iterating!

This would have been easier to review if you had left the refactoring in a separate commit from the fix related to the winreg update (I would also put the fix for the winreg update in the same commit as the winreg update).

djc (approved):

tests/suite/cli_paths.rs
rami3l: The commit history has been rewritten, as promised.

rami3l:
The CI failure, as usual, is not related to this change.

PR #3464, opened 2023-08-23
Remove redundant message if an error occurs during package extraction

ongchi:
This is the message when there is an error on extracting, it does contain the source of the cause. For example, when out of space on extraction:
> error: failed to extract package (perhaps you ran out of disk space?): No space left on device (os error 28)

This PR remove the message in parentheses like this:
> error: failed to extract package: No space left on device (os error 28)

Note: the error message already contains the source of cause by the selector "{:#}" in `cli::common::report_error`.
See also: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations

### DIFF
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -151,7 +151,7 @@ impl<'a> TarPackage<'a> {
         // $pkgname-$version-$target. Skip that directory when
         // unpacking.
         unpack_without_first_dir(&mut archive, &temp_dir, notify_handler)
-            .context("failed to extract package (perhaps you ran out of disk space?)")?;
+            .context("failed to extract package")?;
 
         Ok(TarPackage(
             DirectoryPackage::new(temp_dir.to_owned(), false)?,


djc (commented):
So yes, for that specific `ENOSPC` error it would still make sense. My question is if there are other potential errors for `unpack_without_first_dir()` that might be indirectly caused by a lack of disk space, which could still benefit from the existing hint?

djc:
The hint was added in https://github.com/rust-lang/rustup/pull/2721/files#diff-47e56d6d43f3702f3f6de0b34ad1988bcccfdba0d493fdfa247c8fd544195327R153, which is sort of generic (conversion to anyhow), so it doesn't necessarily look like there was a specific reason to add this there?

ongchi:
Possible errors are return immediately, I think this should be obvious, except for these lines:
https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/dist/component/package.rs#L314
https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/dist/component/package.rs#L358
https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/dist/component/package.rs#L493
https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/dist/component/package.rs#L517

The `item` or `op` above is an instance of `CompletedIo`.
https://github.com/rust-lang/rustup/blob/a4dd7d03501def7ee3e4e2961bca7f6ec3b2e1aa/src/diskio/mod.rs#L187-L210

The `io_executor` may produce multiple results, but only the the first error would be returned. This is the only reason I can imagine why the error source returned is not an `ENOSPC` but might be the root cause.
However, I believe user still has to resolve any error from `unpack_without_first_dir`, as an `ENOSPC` message will eventually present if it is one of the reasons for the failure.

djc (approved):
It does seem fair to conclude that this is not the right place to narrow the context to suggest `ENOSPC` if that is necessary at all. Given that was added in a commit that was arguably unrelated, I think it makes sense to remove it all, though maybe @rbtcollins remembers why he thought adding it here was useful.

PR #3466, opened 2023-08-25
Mention `brew install rustup-init` in the user guide

### LINKED ISSUE #3460: Document how to install (and use) `rustup` with `brew`

AFAIK, the topic of installing and using `rustup` along with homebrew (`brew`) is not covered anywhere in the public docs, but it is also a common source of confusion among macOS users wanting to learn about Rust (https://github.com/rust-lang/rustup/issues/3458, [^1]).

[^1]: https://users.rust-lang.org/t/compile-failed-with-target-aarch64-linux-android/92439/14

### Solution you'd like

I am looking forward to some doc improvements in this respect.

### Notes

_No response_

rami3l:
Closes #3460.


doc/user-guide/src/installation/other.md
```diff
@@ -29,6 +29,32 @@ $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default
 $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --profile minimal --default-toolchain nightly
 ```
 
+
+## Using a package manager
+
+> Please note that the rustup project is not maintaining any package mentioned in this section.
+> If you have encountered any problems installing `rustup` with a package manager,
+> please contact the package maintainer(s) for further information.
```
rami3l, line 37: I added this section to clarify that this is not the "blessed" way of installing `rustup` (so it is up to the user to decide which method they would use), hoping to reduce responsibility confusions.

But I haven't written anything like that before... Probably the wording could use some improvement here.

doc/user-guide/src/SUMMARY.md
```diff
@@ -3,9 +3,9 @@
 [Introduction](index.md)
 - [Installation](installation/index.md)
     - [Windows](installation/windows.md)
-        - [MSVC Prerequistes](installation/windows-msvc.md)
+        - [MSVC prerequistes](installation/windows-msvc.md)
     - [Other installation methods](installation/other.md)
-    - [Package managers](installation/package-managers.md)
+    - [Adopting external toolchains](installation/adopting-external-toolchains.md)
```
rami3l: I found `package-managers` ambiguous since it could mean installing `rustup` using a package manager, or installing `rustup` while already using externally-managed `rust` toolchains.

Hopefully this change will make that ambiguity disappear.

ehuss: Can you please add a redirect when renaming a file? 

In `book.toml`, you can add something like:

```toml
[output.html.redirect]
"/installation/package-managers.html" = "adopting-external-toolchains.html"
```

(I think that syntax is correct, but please test it.)

rami3l: Will do.

rami3l: @ehuss
I tested it locally and it works without any problem. Thanks a lot!

doc/user-guide/src/installation/other.md
djc: How does this display in the rendered documentation?

doc/user-guide/book.toml
```diff
@@ -6,3 +6,6 @@ title = "The rustup book"
 edit-url-template = "https://github.com/rust-lang/rustup/edit/master/doc/user-guide/{path}"
 git-repository-url = "https://github.com/rust-lang/rustup/tree/master/doc/user-guide"
 site-url = "https://rust-lang.github.io/rustup/"
+
+[output.html.redirect]
+"/installation/package-managers.html" = "adopting-external-toolchains.html"
```
djc: I would squash this into the renaming commit.

doc/user-guide/src/SUMMARY.md
djc: Hmm, not completely sure this feels like an improvement. "Package manager" is a clear existing term of art, whereas "external toolchain" seems like a term you invented. Maybe there's a better way to disambiguate?

djc (commented):
Nice improvements overall!

doc/user-guide/book.toml
rami3l: @djc No worries, I'll rewrite all history once there are no more outstanding concerns.

doc/user-guide/src/SUMMARY.md
rami3l: @djc I originally used something like "Adopting toolchain installations unmanaged by `rustup`" but it was a bit too long. I didn't mean to coin a new term here. Any suggestions?

doc/user-guide/src/installation/other.md
rami3l: @djc If you mean the quote-as-callout part, have a look at the bottom of https://doc.rust-lang.org/stable/book. I got my inspiration from there.

It also uses bold text together with the quote though.

Anyway, this is what it looks like on my machine:

image

doc/user-guide/src/SUMMARY.md
djc: That seems pretty good. Maybe just drop the "installations" part?

rami3l: @djc It looks like this on my machine...
image

But once I use something like "Adopting unmanaged toolchains" you might ask again: is "unmanaged toolchains" a new term?

doc/user-guide/src/installation/other.md
djc: Okay, looks nice! I was wondering if mdBook natively has something like "warning" blocks which I think I've seen in similar context, but quick skimming of the mdBook docs didn't turn up anything like that.

doc/user-guide/src/SUMMARY.md
djc: I slightly prefer "unmanaged" over "external" since it seems little more self-explanatory, but not sure how we decide on brevity/layout vs being explicit.

doc/user-guide/src/installation/other.md
ChrisDenton: Github flavoured markdown has ["warning" blocks](https://github.com/orgs/community/discussions/16925) but I don't *think* mdbook have implemented that.

doc/user-guide/src/SUMMARY.md
rami3l: cc @rbtcollins

rbtcollins: I think that that doc really deals with two things.

One is about toolchains, but the other, bigger one, is that rustup refuses to install if someone already has rust installed via any other method.

I can understand that 'package managers' is too abbreviated, but it is honestly 90+% of the bug reports we get from that space - Windows and Mac users seem to find rustup themselves; folk on Linux tend to apt-get or whatever and then when the version isn't satisfactory, or something else specifically wants rustup, then try to install rustup.

Perhaps a title of 'Installing when rust is already installed' ?

rami3l: @rbtcollins That is a nice point being made. However, I still want to make it shorter, something like `Already installed Rust?` (question mark included) ...

At least in the case of https://github.com/rust-lang/rustup/issues/953 this title is concise enough while still being easily discoverable:

image

rami3l: If you still think this is not appropriate, please feel free to comment below and I'll un-resolve this conversation :)

doc/user-guide/src/SUMMARY.md
```diff
@@ -3,9 +3,9 @@
 [Introduction](index.md)
 - [Installation](installation/index.md)
     - [Windows](installation/windows.md)
-        - [MSVC Prerequistes](installation/windows-msvc.md)
+        - [MSVC prerequistes](installation/windows-msvc.md)
```
djc: Note, there's a pre-existing typo here, it should be "prerequisites". Can you fix that here?

djc (approved):

doc/user-guide/src/SUMMARY.md
rami3l: Done!

rami3l:
I had a bad connection at the time so I demanded multiple reruns for OpenBSD by accident. [One](https://cirrus-ci.com/task/5010144048709632) did succeed but is not displayed on GitHub.
Merging as if the CI were green.

PR #3467, opened 2023-08-25
Fix conversion from PartialTargetTriple to TargetTriple

### LINKED ISSUE #3166: rustup component remove component-NOT-TOOLCHAIN-TARGET errors

I ran `rustup component list --installed` and saw that I had the component `rust-std-wasm32-unknown-unknown` installed, which I wanted to delete. I thought that I could simply copy-paste the component name into the `remove` command like this:
```
rustup component remove rust-std-wasm32-unknown-unknown
```
However, this doesn't work with `remove`. I got the following unhelpful error message:
```
error: toolchain 'nightly-x86_64-apple-darwin' does not contain component 'rust-std-wasm32-unknown-unknown' for target 'x86_64-apple-darwin'
```
(Also the last line doesn't make sense because I'm trying to uninstall something here, not install it.)

### Solution you'd like

The command that actually works is:
```
rustup component remove --target wasm32-unknown-unknown rust-std
```
I would like it if rustup told me that. You could assume that the component names which are output by `list` can also be passed back into rustup as a component name.

### Notes

Version 1.25.1

ongchi:
Most of valid target triple have this structure:
> arch-os-env

For instance, a windows system with gnu environment would be:
> <ins>x86_64</ins>-<ins>pc-windows</ins>-<ins>gnu</ins>

But a triple without `env` could be a valid triple:
> <ins>x86_64</ins>-<ins>apple-darwin</ins>

This PR makes convertion of `PartialTargetTriple` without an `env` field a valid `TargetTriple` and updates the list of possible targets of `PartialTargetTriple`.
This also makes `Component::new_with_target` able to create a valid target from a triple without `env`.

Fix: #3166


src/dist/triple.rs
```diff
@@ -5,46 +5,138 @@ use regex::Regex;
 // the PartialTargetTriple.
 
 static LIST_ARCHS: &[&str] = &[
-    "i386",
-    "i586",
-    "i686",
-    "x86_64",
+    "aarch64",
```
djc, line 8: Just to check, where did you source these lists of current arches, OSes and envs?

src/dist/dist.rs
```diff
@@ -406,16 +406,13 @@ impl TargetTriple {
...
-                value.arch.unwrap(),
-                value.os.unwrap(),
-                value.env.unwrap()
-            )))
-        } else {
-            Err("Incomplete / bad target triple")
-        }
+        let arch = value.arch.ok_or("Incomplete / bad target triple")?;
+        let os = value.os.ok_or("Incomplete / bad target triple")?;
+        let triple = match value.env {
```
djc: Tiny style nit: I would spell this `Ok(Self(match value.env { .. }))` directly, which avoids a low-value single use binding.

djc (approved):
This looks good, thanks!

djc:
Maybe it would be good to add some tests for the new behavior? Especially the high-level CLI behavior as discussed in #3166.

src/dist/triple.rs
ongchi: I copied all the targets from the [Platform Support section of the rustc book](https://doc.rust-lang.org/rustc/platform-support.html) and created the lists manually. I'm not sure if these lists already exist somewhere, so I created them manually.

djc: Sure, maybe add a docstring linking to a section in the book for each of these constants?

src/dist/dist.rs
ongchi: Sure, willing to do so.

rbtcollins:
Thanks for the patch. 

I have two questions. 

1) Its not clear to me that this fixes the linked bug, which is about removing a fully qualified target component `rustup component remove rust-std-wasm32-unknown-unknown` from a `nightly-x86_64-apple-darwin` toolchain. A unit test showing that `rust-std-wasm32-unknown-unknown` was failing in this codepath would be one way to show that more clearly.

2) I have a suspicion that this is the wrong direction to take the fix.

One possibility is to hook in a call to resolve() with the context of the toolchain.

Another possibility is to look at what would happen if instead we defined `x86_64-apple-darwin` as arch=`x86_64`, os=`apple`, env=`darwin` ?

The problem I am worried about is that `i686-pc-windows` is not a valid `TargetTriple` - it has to be resolved [resolve(](https://github.com/rust-lang/rustup/blob/b9ef6537858c671fed7e5ed170bdccdaff727e93/src/dist/dist.rs#L440) - which confusingly actually goes from a string -> ToolchainDesc with a fallback context of the particular host that we're configured for (e.g. `cfg.get_default_host_triple()`).

So a missing env gets filled in - and I'm still used to the older machine-vendor-os triples from gcc - https://wiki.osdev.org/Target_Triplet  / https://github.com/gcc-mirror/gcc/blob/master/config.sub#L1877C6-L1877C37 - where apple is a vendor and darwin is the os.

ongchi:
Trying to clarify this issue as best I can.

1. I think add/remove components by name with target string is intended to be supported, the behavior of `component_add` and `component_remove` will trying to parse target from user input before refer to current toolchain. But I agree the failing message is not clear when input target is not valid with this patch, it would better to hint user that the target is nonexisted.


2. Since the list of supported targets could be obtained from `common::list_targets()`, maybe we could just match from this list for target validation?

The rustup will try to extract target triple from the substring from the user input. If the substring of user input does not match the combination of archs, oses and envs in `PartialTargetTriple`, then the string will be treated as a component name without target suffix.

For the case in #3166:

> rustup component remove rust-std-wasm32-unknown-unknown

Because of rustup cannot parse `wasm32-unknown-unknown` as a valid target, the `rust-std-wasm32-unknown-unknown` here is treated as a component name, and the target inferred to be `x86_64-apple-darwin` for this case. The error message as following:

> error: toolchain 'nightly-x86_64-apple-darwin' does not contain component 'rust-std-wasm32-unknown-unknown' for target 'x86_64-apple-darwin'
> note: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html

The actual component and target should be `rust-std` and `wasm32-unknown-unknown` in this error message respectively.
This is subtle because component name may also contains a '-' character (`rust-std` in this case), the component name is determined by user input subtract from possible target suffix currently (`TryFrom<PartialTargetTriple>`).

The valid component list could be obtained from unique values from `common::list_components`, this would be better than relying on pattern match and decrease maintaining efforts.

Most error messages should remain, except the input component is totally gibberish and does not match any valid component and target, then display a different message to address this error.

src/dist/dist.rs
```diff
@@ -406,16 +406,12 @@ impl TargetTriple {
...
-                "{}-{}-{}",
-                value.arch.unwrap(),
-                value.os.unwrap(),
-                value.env.unwrap()
-            )))
-        } else {
-            Err("Incomplete / bad target triple")
-        }
+        let arch = value.arch.ok_or("Incomplete / bad target triple")?;
+        let os = value.os.ok_or("Incomplete / bad target triple")?;
```
rami3l, line 410: Sorry for the late review 🙇, but wouldn't it be problematic to pass `arch` and `os` as strings as if they were valid? Is there a particular reason not to fail early here?

ongchi: I don't understand. The `ok_or` transforms the `Option<T>` into a `Result<T, E>`, so, these two lines would be fail early if any of `arch` or `os` is None.

rami3l: @ongchi Thanks for your explanation! Must have been a misread from my side, sorry 🙇

PR #3468, opened 2023-08-29
Support `stable-DELTA` as a version descriptor

### LINKED ISSUE #2291: Support `stable-X` as a version descriptor

In Rust-Embedded land, we are struggling with the management of our *Minimum Supported Rust Version*. That is, the promise we make to our users that compilers all the way back to our MSRV will compile this code, and we won't use any features stabilised after that compiler was released.

Currently we set an MSRV like a stake in the sand, and then time moves on and the stake gets further and further away. One of our crates has an MSRV of 1.30 - thats 72 weeks old at the time of writing. What I'd like to do is bring that MSRV along with us - trailing it behind us on a piece of string.

Basically in my CI files I want to "rustup install stable-2", and if stable is 1.40, it would install 1.38. If stable was 1.51 it would install 1.49. Then I can set my Travis CI matrix to build on:

* stable
* stable-1
* stable-2

And I never have to touch the CI file again.

**Notes**

At the moment, to achieve a rolling MSRV we would have to touch the CI file every six weeks when there's a new stable release. That or `rustup install stable`, grok the current version with some grep/sed foo, subtract two minor versions (not easy in Bash) and then install that version.

rami3l:
Closes #2291.

The current implementation proposal is posted in https://github.com/rust-lang/rustup/issues/2291#issuecomment-1689220651.

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -159,6 +159,24 @@ static TRIPLE_MIPS64_UNKNOWN_LINUX_GNUABI64: &str = "mips64-unknown-linux-gnuabi
 #[cfg(all(not(windows), target_endian = "little"))]
 static TRIPLE_MIPS64_UNKNOWN_LINUX_GNUABI64: &str = "mips64el-unknown-linux-gnuabi64";
 
+/// Some old Rust versions don't have v2 manifests, but they don't have point releases either,
+/// so to make the two-part version numbers work for these versions, specially turn
+/// them into their corresponding ".0" version.
+fn convert_old_two_part_version(version: &str) -> &str {
+    match version {
+        "1.0" => "1.0.0",
+        "1.1" => "1.1.0",
+        "1.2" => "1.2.0",
+        "1.3" => "1.3.0",
+        "1.4" => "1.4.0",
+        "1.5" => "1.5.0",
+        "1.6" => "1.6.0",
+        "1.7" => "1.7.0",
+        "1.8" => "1.8.0",
+        other => other,
+    }
+}
+
 impl FromStr for ParsedToolchainDesc {
     type Err = anyhow::Error;
     fn from_str(desc: &str) -> Result<Self> {
@@ -181,21 +199,7 @@ impl FromStr for ParsedToolchainDesc {
                 }
             }
 
-            // These versions don't have v2 manifests, but they don't have point releases either,
-            // so to make the two-part version numbers work for these versions, specially turn
-            // them into their corresponding ".0" version.
-            let channel = match c.get(1).unwrap().as_str() {
-                "1.0" => "1.0.0",
-                "1.1" => "1.1.0",
-                "1.2" => "1.2.0",
-                "1.3" => "1.3.0",
-                "1.4" => "1.4.0",
-                "1.5" => "1.5.0",
-                "1.6" => "1.6.0",
-                "1.7" => "1.7.0",
-                "1.8" => "1.8.0",
-                other => other,
-            };
+            let channel = convert_old_two_part_version(c.get(1).unwrap().as_str());
 
             Self {
                 channel: channel.to_owned(),
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -1,5 +1,6 @@
 use std::collections::HashSet;
 use std::env;
+use std::ffi::OsString;
 use std::fmt;
 use std::io::Write;
 use std::ops::Deref;
@@ -26,7 +27,10 @@ use crate::{
     },
     errors::RustupError,
     process,
-    toolchain::names::{ResolvableToolchainName, ToolchainName},
+    toolchain::{
+        names::{ResolvableLocalToolchainName, ToolchainName},
+        toolchain::Toolchain,
+    },
     utils::utils,
 };
 pub static DEFAULT_DIST_SERVER: &str = "https://static.rust-lang.org";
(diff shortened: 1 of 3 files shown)


src/dist/dist.rs
```diff
@@ -159,6 +159,24 @@ static TRIPLE_MIPS64_UNKNOWN_LINUX_GNUABI64: &str = "mips64-unknown-linux-gnuabi
 #[cfg(all(not(windows), target_endian = "little"))]
 static TRIPLE_MIPS64_UNKNOWN_LINUX_GNUABI64: &str = "mips64el-unknown-linux-gnuabi64";
 
+/// Some old Rust versions don't have v2 manifests, but they don't have point releases either,
```
djc, line 170: I would suggest splitting this commit into smaller constituent parts that make exactly one conceptual change (like "extract `convert_old_two_part_version()` into a separate function"). It should then also be obvious from that commit why it is necessary (in this case, it's not clear why you extracted this).

(As part of this, your commits should have a more precise name, unlike "Refactor `dist`" which is very general.)

src/dist/dist.rs
```diff
@@ -172,43 +190,25 @@ impl FromStr for ParsedToolchainDesc {
...
 
-        let d = TOOLCHAIN_CHANNEL_RE.captures(desc).map(|c| {
-            fn fn_map(s: &str) -> Option<String> {
-                if s.is_empty() {
-                    None
-                } else {
-                    Some(s.to_owned())
-                }
-            }
+        TOOLCHAIN_CHANNEL_RE
```
djc, line 248: I think this way of writing this would be clearer:

```rust
let channel = match TOOLCHAIN_CHANNEL_RE.captures(desc) {
    Some(c) => c,
    None => return RustupError::InvalidToolchainName(desc.to_owned().into()),
};

..
```

src/dist/dist.rs
```diff
@@ -543,6 +568,35 @@ impl ToolchainDesc {
         (channels.iter().any(|x| *x == self.channel) || TRACKING_VERSION.is_match(&self.channel))
             && self.date.is_none()
     }
+
+    pub(crate) fn resolve_delta(channel: &str, delta: u64) -> Result<String> {
+        if channel != "stable" {
```
djc: If this is the only use of `channel`, I'd probably handle this condition outside of the function.

djc (commented):
Early review, hope this helps.

src/dist/dist.rs
rami3l: The original intent was to leave room for other deltas (I don't know what `beta-2` means, but it could possibly mean something later on)... But indeed it could be a premature optimization.

src/dist/dist.rs
rami3l: > (in this case, it's not clear why you extracted this)

I used this function a second time in a later commit. I have been rewriting the history very early on in this PR, and yes this might look like I can perfectly predict things, but of course this is not the case :)

If you have any ideas about how to split the commits better given that context, please don't hesitate to tell me :D
(and commit renaming is always easier than a full rebase)

djc: > I used this function a second time in a later commit.

Yup, that's completely fine and a good reason for extracting the function. Might be the kind of thing that you can put in a commit message.

I think I see 3 changes here that are mostly separate from each other:

1. Extract `convert_old_two_part_version()`
2. Early return for error case (as suggested in my other comment)
3. Extract `filter_non_empty()` function (though I would suggest having this take an index directly, naming it maybe more like `non_empty_component()`?)

src/dist/dist.rs
djc: Yes, I'd simplify this for now -- can always be changed later.

rami3l:
> Early review, hope this helps.

@djc Thanks a lot for your kindness!

However, at the moment this PR does not even work as intended in many cases, nor does it contain the required black and/or white box tests, so it's very much a WIP as indicated in the commit message.

Of course I'll still have to rewrite the history many times, and when I'm sure it does work, I'll be happy to take the WIP hat off this PR, and maybe then we can do more of this refinement thing more efficiently :)

Thanks a lot for your current suggestions anyway! I'll try to extract the essence of it and make it really nice for the reviewers like you to follow along, with all the interesting corners carefully explained!

djc:
> However, at the moment this PR does not even work as intended in many cases, nor does it contain the required black and/or white box tests, so it's very much a WIP as indicated in the commit message.

I'm happy to leave the process of how you want to develop this PR to you. I thought it would be useful to give you feedback early since in my experience creating small commits from the start can be a lot easier than splitting up larger commits later -- and many contributors don't like the additional work involved in splitting commits.

I'll wait until you mark this as ready for review before I next review this. 👍

rami3l:
> in my experience creating small commits from the start can be a lot easier than splitting up larger commits later -- and many contributors don't like the additional work involved in splitting commits.

@djc Thanks! After a few previous PRs I'm starting to make sense of that point now :)

rami3l:
Issue: Weird infinite recursion caused by `RUSTUP_TOOLCHAIN`

```sh
> RUSTUP_HOME=home CARGO_HOME=home RUSTUP_TOOLCHAIN=stable-2 ./home/bin/rustc --version

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
fish: Job 1, 'RUSTUP_HOME=home CARGO_HOME=hom…' terminated by signal SIGABRT (Abort)
```

---

As it turns out, to resolve `stable-2` we must first resolve `stable`, and the latter cannot rely on `crate::cli::set_globals` and friends when `RUSTUP_TOOLCHAIN` is set, since that will again lead to the resolution of `stable-2`.

rami3l:
Closing as the relevant issue has been marked as "not planned".

PR #3470, opened 2023-08-29
Isolate trycmd tests from environment

rbtcollins:


djc (approved):
Thanks, that seems like it might help!

PR #3475, opened 2023-09-08
rustup-init.sh ksh and zsh compatibility fixes

### LINKED ISSUE #3388: rustup-init.sh's solaris/illumos bail doesn't work properly
### Problem

I tried running rustup-init.sh with `sh` on a recent install of illumos. Looking at the source code, I expected it to error with a nice error message telling me to use bash instead: https://github.com/rust-lang/rustup/blob/ab81ee3c5ec3a41e9eaa35ae3108070adf751775/rustup-init.sh#L11-L16

Instead, I got random errors about `local` not working.

My `KSH_VERSION` is `Version jM 93+ 2012-08-01`.
`uname -a` is `SunOS illumos 5.11 illumos-676abcb77c i86pc i386 i86pc`

### Steps

1. install a recent install of illumos (I used openindiana `OI-hipster-gui-20230502`)
2. run rustup-init.sh

### Possible Solution(s)

It seems a like a more robust detection would be to check for `SunOS` in `uname`, but I'm not sure how well that works on Solaris and haven't tested it (but I would assume that Solaris has `SunOS` too since "SunOS" predates "Solaris").

Alternatively, also checking for my version may work too, that version is pretty old too so it seems unlikely to change frequently.

### Notes

_No response_

### Rustup version

```console
https://github.com/rust-lang/rustup/blob/ab81ee3c5ec3a41e9eaa35ae3108070adf751775/rustup-init.sh
```

### Installed toolchains

```console
none
```

### LINKED ISSUE #3474: rustup-init.sh fails in zsh due to no word-splitting of curl "--retry 3"
### Problem

zsh does not enable word-splitting by default and this causes the rustup-init.sh script to fail with:
```
info: downloading installer
curl: option --retry 3 -C -: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
```
This is because zsh is interpreting `--retry 3 -C -` as a single argument, instead of splitting to `--retry`, `3`, `-C`, `-`.

Tested with Linux zsh 5.8.1.

### Steps

1. Download rustup-init.sh to a machine with a curl supporting `curl --retry`
2. Execute `zsh rustup-init.sh`

### Possible Solution(s)

Use `setopt shwordsplit` when zsh is detected.

### Notes

_No response_

### Rustup version

```console
```

### Installed toolchains

```console
N/A
```

alexhudspith:
Fixes #3388
Fixes #3474

Tested with:
|Shell|Versions|
|-|-|
|bash|(ubuntu) 5.1.16(1)-release|
|dash|(ubuntu) 0.5.11+git20210903+057cd650a4ed-3build1<br>(debian) 0.5.12-2|
|ksh93|(illumos) Version jM 93u+ 2012-08-01<br>(ubuntu) Version AJM 93u+m/1.0.0-beta.2 2021-12-17|
|mksh|(ubuntu) @(#)MIRBSD KSH R59 2022/01/31 +Debian|
|zsh|(ubuntu) 5.8.1|

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -19,6 +19,10 @@ has_local() {
 
 has_local 2>/dev/null || alias local=typeset
 
+# zsh does not split words by default, Required for curl retry arguments below.
+if [ -n "$ZSH_VERSION" ]; then
+    setopt shwordsplit
+fi
 
 set -u
 
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 # shellcheck shell=dash
+# shellcheck disable=SC2039  # local is non-POSIX
 
 # This is just a little script that can be downloaded from the internet to
 # install rustup. It just does platform detection, downloads the installer
@@ -8,13 +9,15 @@
 # It runs on Unix shells like {a,ba,da,k,z}sh. It uses the common `local`
 # extension. Note: Most shells limit `local` to 1 var per line, contra bash.
 
-if [ "$KSH_VERSION" = 'Version JM 93t+ 2010-03-05' ]; then
-    # The version of ksh93 that ships with many illumos systems does not
-    # support the "local" extension.  Print a message rather than fail in
-    # subtle ways later on:
-    echo 'rustup does not work with this ksh93 version; please try bash!' >&2
-    exit 1
-fi
+# Some versions of ksh have no `local` keyword. Alias it to `typeset`, but
+# beware this makes variables global with f()-style function syntax in ksh93.
+# mksh has this alias by default.
+has_local() {
+    # shellcheck disable=SC2034  # deliberately unused
+    local _has_local
+}
+
+has_local 2>/dev/null || alias local=typeset
 
 
 set -u


djc (approved):
I'm not very experienced with shell, but these changes seem to make sense to me.

rami3l (dismissed):

rustup-init.sh
```diff
@@ -8,14 +9,20 @@
...
+has_local() {
+    # shellcheck disable=SC2034  # deliberately unused
+    local _has_local
+}
 
+has_local 2>/dev/null || alias local=typeset
+
+# zsh does not split words by default, Required for curl retry arguments below.
+if [ -n "$ZSH_VERSION" ]; then
+    setopt shwordsplit
```
rami3l, line 24: Concern: I don't use `zsh` personally, but does this `setopt` make persistent changes somehow? (I have seen similar concerns elsewhere such as https://github.com/sdkman/sdkman-cli/issues/210.)

rami3l: It seems to me that one solution might be calling `setopt local_options shwordsplit` in a function (I'm looking at [`nvm.sh`](https://github.com/nvm-sh/nvm/blob/8a83b36688e270e460803ab842e3333393b523c7/nvm.sh#L1497), for example). When the function exits, `shwordsplit` will be restored to the original value.

alexhudspith: Good point. If the script is directly `source`d, then yes it would persist for as long as that shell session is alive. I'm not sure if that's something that's recommended though, rather then executing via `zsh rustup-init.sh` or as on [rustup.rs](https://rustup.rs/): `curl ... rustup-init.sh | zsh`.

alexhudspith: > It seems to me that one solution might be calling `setopt local_options shwordsplit` in a function (I'm looking at [`nvm.sh`](https://github.com/nvm-sh/nvm/blob/8a83b36688e270e460803ab842e3333393b523c7/nvm.sh#L1497), for example).

Yeah, that sounds better. I'll take a look.

rami3l: I see. Thanks for the clarification!

rami3l (approved):

rustup-init.sh
rami3l: @alexhudspith Oops I have a bad network connection here, sorry for the delay 😅

rami3l:
@alexhudspith Sorry, I somehow missed your reply when approving those changes. The `setopt local_options shwordsplit` fix is still nice to have though: you can make a new PR for that if you like 🙇‍♀️

alexhudspith:
@rami3l No problem, thanks!

PR #3488, opened 2023-09-22
Do not fallback to "arm" in rustup-init.sh on aarch64 with 32-bit userland

alex:
On aarch64, neon will not be present in /proc/cpuinfo, instead asimd is there. Previously, we'd fall back to arm when running with a 32-bit userland, even though armv7 is perfectly appropriate in this case.

This would then run into the issue described in https://github.com/rust-lang/rust/issues/58414 with "CP15 barrier emulation".

This PR fixes that.


rami3l (approved):
Looks good to me!

djc (approved):

PR #3493, opened 2023-09-28
Windows library loading

ChrisDenton:
When loading DLLs, the loader uses a [standard search order](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-unpackaged-apps) that may include the application directory. To ensure the right DLLs get loaded, this PR makes sure to force DLLs to be loaded from system32. This is done using [`/DEPENDENTLOADFLAG`](https://learn.microsoft.com/en-us/cpp/build/reference/dependentloadflag?view=msvc-170) msvc linker option. However this only works for Windows 10 RS1 and higher (which admittedly is almost all Windows 10 users) and will have no effect on earlier OSes. So as a fallback we use the [delay load](https://learn.microsoft.com/en-us/cpp/build/reference/delayload-delay-load-import?view=msvc-170) mechanism to get the DLLs loaded at runtime so that we can use [SetDefaultDllDirectories](https://learn.microsoft.com/en-us/windows/win32/api/LibLoaderAPI/nf-libloaderapi-setdefaultdlldirectories) to affect DLL loading.  Note that DLLs on the system's list of "Known DLLs" are always loaded from system32 so don't need to be delay loaded this way.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -27,4 +27,44 @@ fn main() {
     }
     let target = env::var("TARGET").unwrap();
     println!("cargo:rustc-env=TARGET={target}");
+
+    // Set linker options specific to Windows MSVC.
+    let target_os = env::var("CARGO_CFG_TARGET_OS");
+    let target_env = env::var("CARGO_CFG_TARGET_ENV");
+    if !(target_os.as_deref() == Ok("windows") && target_env.as_deref() == Ok("msvc")) {
+        return;
+    }
+
+    // # Only search system32 for DLLs
+    //
+    // This applies to DLLs loaded at load time. However, this setting is ignored
+    // before Windows 10 RS1 (aka 1601).
+    // https://learn.microsoft.com/en-us/cpp/build/reference/dependentloadflag?view=msvc-170
+    println!("cargo:cargo:rustc-link-arg-bin=rustup-init=/DEPENDENTLOADFLAG:0x800");
+
+    // # Delay load
+    //
+    // Delay load dlls that are not "known DLLs"[1].
+    // Known DLLs are always loaded from the system directory whereas other DLLs
+    // are loaded from the application directory. By delay loading the latter
+    // we can ensure they are instead loaded from the system directory.
+    // [1]: https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#factors-that-affect-searching
+    //
+    // This will work on all supported Windows versions but it relies on
+    // us using `SetDefaultDllDirectories` before any libraries are loaded.
+    // See also: src/bin/rustup-init.rs
+    let delay_load_dlls = ["bcrypt", "powrprof", "secur32"];
+    for dll in delay_load_dlls {
+        println!("cargo:rustc-link-arg-bin=rustup-init=/delayload:{dll}.dll");
+    }
+    // When using delayload, it's necessary to also link delayimp.lib
+    // https://learn.microsoft.com/en-us/cpp/build/reference/dependentloadflag?view=msvc-170
+    println!("cargo:rustc-link-arg-bin=rustup-init=delayimp.lib");
+
+    // # Turn linker warnings into errors
+    //
+    // Rust hides linker warnings meaning mistakes may go unnoticed.
+    // Turning them into errors forces them to be displayed (and the build to fail).
+    // If we do want to ignore specific warnings then `/IGNORE:` should be used.
+    println!("cargo:cargo:rustc-link-arg-bin=rustup-init=/WX");
 }
--- a/src/bin/rustup-init.rs
+++ b/src/bin/rustup-init.rs
@@ -29,6 +29,9 @@ use rustup::is_proxyable_tools;
 use rustup::utils::utils;
 
 fn main() {
+    #[cfg(windows)]
+    pre_rustup_main_init();
+
     let process = OSProcess::default();
     with(process.into(), || match maybe_trace_rustup() {
         Err(e) => {
@@ -162,4 +165,22 @@ fn do_recursion_guard() -> Result<()> {
     }
 
     Ok(())
+}
+
+/// Windows pre-main security mitigations.
+///
+/// This attempts to defend against malicious DLLs that may sit alongside
+/// rustup-init in the user's download folder.
+#[cfg(windows)]
+pub fn pre_rustup_main_init() {
+    use winapi::um::libloaderapi::{SetDefaultDllDirectories, LOAD_LIBRARY_SEARCH_SYSTEM32};
+    // Default to loading delay loaded DLLs from the system directory.
+    // For DLLs loaded at load time, this relies on the `delayload` linker flag.
+    // This is only necessary prior to Windows 10 RS1. See build.rs for details.
+    unsafe {
+        let result = SetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_SYSTEM32);
+        // SetDefaultDllDirectories should never fail if given valid arguments.
+        // But just to be safe and to catch mistakes, assert that it succeeded.
+        assert_ne!(result, 0);
+    }
 }


src/bin/rustup-init.rs
```diff
@@ -163,3 +166,19 @@ fn do_recursion_guard() -> Result<()> {
...
+/// Windows pre-main security mitigations.
+///
+/// This is attempts to defend against malicious DLLs that may sit alongside
+/// rustup-init in the user's download folder.
+#[cfg(windows)]
+pub fn pre_rustup_main_init() {
+    use winapi::um::libloaderapi::{SetDefaultDllDirectories, LOAD_LIBRARY_SEARCH_SYSTEM32};
+    // Default to loading delay loaded DLLs from the system directory.
+    unsafe {
+        let result = SetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_SYSTEM32);
```
cuviper, line 181: The [docs](https://learn.microsoft.com/en-us/windows/win32/api/LibLoaderAPI/nf-libloaderapi-setdefaultdlldirectories) for this function seem to say that it's not directly available on Win7:
> Windows 7, Windows Server 2008 R2, Windows Vista and Windows Server 2008:  To call this function in an application, use the [GetProcAddress](https://learn.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-getprocaddress) function to retrieve its address from Kernel32.dll. [KB2533623](https://support.microsoft.com/kb/2533623) must be installed on the target platform.

So... how did this call work? Are we not setting up the build in a Win7 mode? (until we actually drop that support)

ChrisDenton: Hm, this works in my Win7 VM. Presumably the advice to use `GetProcAddress` is because the old SDK libraries didn't include them and they require an OS update to work. You get an error message if the OS update is not installed but it's quite easy to find [answers on the web](https://superuser.com/questions/1492060/procedure-entry-point-setdefaultdlldirectories-could-not-be-located-kernel32-dll) pointing to the necessary update. I think it's ok to require users to have installed security updates, especially for such an old OS.

But yes, Windows 7 hasn't been tested in CI in a long time. The same goes for all rustlang repos I'm aware of.

build.rs
```diff
@@ -27,4 +27,36 @@ fn main() {
     }
     let target = env::var("TARGET").unwrap();
     println!("cargo:rustc-env=TARGET={target}");
+
+    let target_os = env::var("CARGO_CFG_TARGET_OS");
+    let target_env = env::var("CARGO_CFG_TARGET_ENV");
+    if Ok("windows") == target_os.as_deref() && Ok("msvc") == target_env.as_deref() {
```
djc: Nit: please switch the constants with the run-time values, which seems more idiomatic. I'd also like to the invert the condition here to do an early return for non-Windows-MSVC targets.

Next, please add some comments here similar to the PR description to describe what problem we're trying to solve here.

build.rs
```diff
@@ -27,4 +27,36 @@ fn main() {
...
+        // using `SetDefaultDllDirectories` before any libraries are loaded.
+        let delay_load_dlls = ["bcrypt", "powrprof", "secur32"];
+        for dll in delay_load_dlls {
+            println!("cargo:rustc-link-arg-bin=rustup-init=/delayload:{dll}.dll");
+        }
+        println!("cargo:rustc-link-arg-bin=rustup-init=delayimp.lib");
+
+        // # Turn linker warnings into errors
+        //
+        // Rust hides linker warnings meaning mistakes may go unnoticed.
```
djc: Are there generally good reasons to hide these warnings "upstream" (I guess in rustc)? Or is there a good reason that rustup's behavior, in particular, should be different from the default?

build.rs
```diff
@@ -27,4 +27,36 @@ fn main() {
...
+        // # Only search system32 for DLLs
+        //
+        // This applies to DLLs loaded at load time. However, this setting is ignored
+        // before Windows 10 RS1.
+        println!("cargo:cargo:rustc-link-arg-bin=rustup-init=/DEPENDENTLOADFLAG:0x800");
+
+        // # Delay load
+        //
+        // Delay load dlls that are not "known DLLs".
+        // Known DLLs are always loaded from the system directory whereas other DLLs
```
djc: Do you have some reference documentation links for this?

src/bin/rustup-init.rs
```diff
@@ -163,3 +166,19 @@ fn do_recursion_guard() -> Result<()> {
...
     Ok(())
 }
+
+/// Windows pre-main security mitigations.
+///
+/// This is attempts to defend against malicious DLLs that may sit alongside
+/// rustup-init in the user's download folder.
+#[cfg(windows)]
+pub fn pre_rustup_main_init() {
+    use winapi::um::libloaderapi::{SetDefaultDllDirectories, LOAD_LIBRARY_SEARCH_SYSTEM32};
```
djc, line 176: Maybe orthogonal, but should we start using windows-sys here?

src/bin/rustup-init.rs
djc: So the Rust 1.72 [blog post](https://blog.rust-lang.org/2023/08/24/Rust-1.72.0.html#future-windows-compatibility) had this text:

> In a future release we're planning to increase the minimum supported Windows version to 10. The accepted proposal in compiler [MCP 651](https://github.com/rust-lang/compiler-team/issues/651) is that Rust 1.75 will be the last to officially support Windows 7, 8, and 8.1. When Rust 1.76 is released in February 2024, only Windows 10 and later will be supported as tier-1 targets. This change will apply both as a host compiler and as a compilation target.

How do you think those changes (5 months from now) should interact with this code?

djc (commented):
I'm curious, is there a particular motivation for doing this now?

(The rustup project isn't doing great on institutional memory right now, so sorry if I'm asking dumb questions, just trying to make sure we have -- and document -- sufficient state for this going forward.)

src/bin/rustup-init.rs
ChrisDenton: It won't change anything until rust's minimum supported version is Windows 10 RS1 (aka 1607). I can add a similar comment to the one in the build script and note how they're related.

build.rs
ChrisDenton: Do you mean something like:

```rust
if target_os.as_deref() != Ok("windows") && target_env.as_deref() != Ok("msvc") {
    return;
}
```

build.rs
ChrisDenton: Personally I think rustc should not be hiding linker warnings but I'm not trying to litigate that here. However, the reason I think it's especially important in this case is that we're directly and explicitly setting some linker arguments and it would not be good if problems with them (e.g. typos) were being silently ignored.

build.rs
ChrisDenton: Best is https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#factors-that-affect-searching which mentions how it affects search order and where to find the list of known DLLs. Unfortunately the list itself is not documented.

src/bin/rustup-init.rs
ChrisDenton: I'm surprised there's not an issue for that already but I do think that's beyond the scope of this PR.

build.rs
djc: Yup!

ChrisDenton:
Ok, I've updated the comments based on the review. To avoid repeating myself, I only added a short comment to src/bin/rustup-init.rs but referenced build.rs for further info.

build.rs
```diff
@@ -27,4 +27,42 @@ fn main() {
...
+    // [1]: https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#factors-that-affect-searching
+    //
+    // This will work on all supported Windows versions but it relies on
+    // us using `SetDefaultDllDirectories` before any libraries are loaded.
+    // See also: src/bin/rustup-init.rs
+    let delay_load_dlls = ["bcrypt", "powrprof", "secur32"];
+    for dll in delay_load_dlls {
+        println!("cargo:rustc-link-arg-bin=rustup-init=/delayload:{dll}.dll");
+    }
+    println!("cargo:rustc-link-arg-bin=rustup-init=delayimp.lib");
```
djc, line 62: AFAICT `delayimp.lib` is not mentioned on the linked page. Do you have a reference for this/why this is necessary?

ChrisDenton: Ah, see the documentation for the [delayload linker option](https://learn.microsoft.com/en-us/cpp/build/reference/delayload-delay-load-import?view=msvc-170). I've added a comment that links to it.

ChrisDenton:
I'm not sure why some checks are failing after adding a comment. The failures look entirely unrelated in any case.

djc (approved):
LGTM, thanks!

djc:
It looks like the failures happen because we run clippy from beta? It's a little hard to tell because the CI setup is pretty complicated...

djc:
The CI issues should be fixed after merging #3497, would you mind rebasing so we can get a clean CI run?

build.rs
```diff
@@ -27,4 +27,44 @@ fn main() {
     }
     let target = env::var("TARGET").unwrap();
     println!("cargo:rustc-env=TARGET={target}");
+
+    // Set linker options specific to Windows MSVC.
+    let target_os = env::var("CARGO_CFG_TARGET_OS");
+    let target_env = env::var("CARGO_CFG_TARGET_ENV");
+    if target_os.as_deref() != Ok("windows") && target_env.as_deref() != Ok("msvc") {
```
rami3l: The De Morgan's Law is being dangerous here: looking at L31, I think you meant:

```suggestion
    if !(target_os.as_deref() == Ok("windows") && target_env.as_deref() == Ok("msvc")) {
```

I could've used `||` but it's less clear.

ChrisDenton: Oh yes. Refactoring is dangerous 🙂.

djc: Nice catch!

rami3l (approved):

ChrisDenton:
Btw, sorry for the noisy force pushes, I committed the change on github then accidentally overwrote the github state with my local state so had to manually reapply 😳

build.rs
rami3l: "Given enough eyeballs, all bugs are shallow" :)

PR #3500, opened 2023-10-05
Avoid warning for unused variant

djc:
The `StreamSelector::TestTtyWriter` variant was previously guarded with `feature = test` but not `test`. Because it's only used in unit tests in the same module that means rustc would yield a warning for an unused variable if the crate is compiled with `--features test` but without actually running the tests.


rami3l (approved):
Sorry, it's on me. Nice catch!

PR #3445, opened 2023-08-16
Refine suggestions of sourcing `$HOME/.cargo/env`

rami3l:
Closes #3415.

PR #3506, opened 2023-10-10
Add suggestions to mention sourcing `env.fish`

### LINKED ISSUE #3504: Add `source` instructions after installation for `fish` users

After the merge of #3108, the `source` instructions after installation for `fish` users are still missing in the help messages of `rustup`.

### Solution you'd like

Make a new PR à la #3445 appending extra info for `fish` users.

### Notes

_No response_

rami3l:
Continuation of #3445 and #3108.
Closes #3504.

I decided to use a new table-like layout inspired by [`venv`](https://docs.python.org/3/library/venv.html#how-venvs-work).

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -225,13 +225,12 @@ This would reload your `PATH` environment variable to include
 Cargo's bin directory ({cargo_home}/bin).
 
 To configure your current shell, you need to source
-the `env` file ({cargo_home}/env).
+the corresponding `env` file under {cargo_home}.
 
-This is usually done by running:
-    source "{cargo_home}/env"
-
-The `source` command above might need to be replaced with `.`
-on shells like `ash`, `dash` and `pdksh`.
+This is usually done by running one of the following:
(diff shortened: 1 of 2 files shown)


src/cli/self_update.rs
```diff
@@ -225,13 +225,12 @@ This would reload your `PATH` environment variable to include
 Cargo's bin directory ({cargo_home}/bin).
 
 To configure your current shell, you need to source
-the `env` file ({cargo_home}/env).
+the corresponding `env` file under {cargo_home}.
```
djc, line 222: Maybe it's worth deduplicating this text before updating it? What do you think?

djc (approved):

src/cli/self_update.rs
rami3l: @djc You're right. I'll have a look at this a bit later.

rami3l: I've extracted that part as a new `macro`, though I'm a little afraid that the readability of this part has reduced a bit.

PR #3518, opened 2023-10-28
Rewrite `toolchain_sort` to prevent panics

### LINKED ISSUE #3517: `rustup toolchain link` with special characters in the toolchain name causes panics
I found that using special characters in the `rustup toolchain link` command can cause panics:
```bash
matt@Matts-PC:~/rust$ rustup toolchain link "foo(bar)" ~/rust/build/host/stage0/
matt@Matts-PC:~/rust$ RUST_BACKTRACE=1 rustup toolchain list
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("unexpected character in pre-release identifier")', src/utils/utils.rs:549:83
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:64:14
   2: core::result::unwrap_failed
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/result.rs:1790:5
   3: rustup::utils::utils::toolchain_sort::special_version
   4: rustup::utils::utils::toolchain_sort::toolchain_sort_key
   5: alloc::slice::<impl [T]>::sort_by::{{closure}}
   6: rustup::config::Cfg::list_toolchains
   7: rustup::cli::rustup_mode::toolchain_list
   8: rustup::cli::rustup_mode::main
   9: rustup_init::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
matt@Matts-PC:~/rust$
```

I also got similar results with `"foo#bar"`

I'd assume the problem is that `rustup toolchain link` needs to sanitise it's input more thoroughly.

This was tested on `rustup 1.26.0 (5af9b9484 2023-04-05)`

rami3l:
Fixes #3517 by completely rewriting the `toolchain_sort` function.

The original code uses the hack of putting everything in `semver::Version::pre` which can easily fail.

### DIFF
--- a/src/toolchain/names.rs
+++ b/src/toolchain/names.rs
@@ -313,8 +313,6 @@ impl Display for ToolchainName {
 /// 2. `X.Y.Z-suffix` names, sorted by semver rules on `X.Y.Z`, then by `suffix`.
 /// 3. Other names, sorted alphanumerically.
 pub(crate) fn toolchain_sort(v: &mut [ToolchainName]) {
-    use semver::Version;
-
     v.sort_by_key(|name| {
         let s = name.to_string();
         if s.starts_with("stable") {
@@ -327,7 +325,7 @@ pub(crate) fn toolchain_sort(v: &mut [ToolchainName]) {
             return (2, None, s);
         }
         if let Some((ver_str, suffix)) = s.split_once('-') {
-            if let Ok(ver) = Version::parse(ver_str) {
+            if let Ok(ver) = semver::Version::parse(ver_str) {
                 return (3, Some(ver), suffix.to_owned());
             }
         }
--- a/src/toolchain/names.rs
+++ b/src/toolchain/names.rs
@@ -321,7 +321,8 @@ pub(crate) fn toolchain_sort(v: &mut [ToolchainName]) {
         }
     }
 
-    fn toolchain_sort_key(s: &str) -> Version {
+    v.sort_by_key(|name| {
+        let s: &str = &format!("{name}");
         if s.starts_with("stable") {
             special_version(0, s)
         } else if s.starts_with("beta") {
@@ -331,9 +332,7 @@ pub(crate) fn toolchain_sort(v: &mut [ToolchainName]) {
         } else {
             Version::parse(&s.replace('_', "-")).unwrap_or_else(|_| special_version(3, s))
         }
-    }
-
-    v.sort_by_key(|name| toolchain_sort_key(&format!("{name}")));
+    });
 }
 
 /// ResolvableLocalToolchainName is used to process values set in
--- a/src/toolchain/names.rs
+++ b/src/toolchain/names.rs
@@ -675,6 +675,9 @@ mod tests {
             "1.2.0-x86_64-unknown-linux-gnu",
             "1.8.0-x86_64-unknown-linux-gnu",
             "1.10.0-x86_64-unknown-linux-gnu",
+            "bar(baz)",
+            "foo#bar",
+            "this.is.not-a+semver",
         ]
         .into_iter()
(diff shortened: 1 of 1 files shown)


src/toolchain/names.rs
```diff
@@ -308,38 +308,31 @@ impl Display for ToolchainName {
...
-            major: 0,
-            minor: 0,
-            patch: 0,
-            pre: Prerelease::new(&format!("pre.{}.{}", ord, s.replace('_', "-"))).unwrap(),
-            build: BuildMetadata::EMPTY,
-        }
-    }
-
-    fn toolchain_sort_key(s: &str) -> Version {
+    v.sort_by_key(|it| {
```
djc: Maybe rename `it` to `name`?

src/toolchain/names.rs
```diff
@@ -308,38 +308,31 @@ impl Display for ToolchainName {
...
-        let b_key = toolchain_sort_key(b_str);
-        a_key.cmp(&b_key)
-    });
+        if s.starts_with("beta") {
+            return (1, None, s);
+        }
+        if s.starts_with("nightly") {
+            return (2, None, s);
+        }
+        if let Some(dash_idx) = s.find('-') {
```
djc: I think we could use something like `split_once()` here?

src/toolchain/names.rs
```diff
@@ -308,38 +308,31 @@ impl Display for ToolchainName {
     }
 }
 
+/// Sorts [`ToolchainName`]s in the following order:
```
djc, line 311: Sorry to be a PITA again, but would you mind splitting this into multiple commits to make it clearer how you got here? I think I see four here:

* Switch from `Version` to a tuple type for the key
* Switch from `sort_by()` to `sort_by_key()`
* Inlining `special_version()`
* Inlining `toolchain_sort_key()`

rami3l: @djc Please mind that the current semantics is not exactly the same as the previous panic-prone one (I worked backwards from the test case to figure out what I should do), so it has been a rewrite rather than a gradual refactoring, and in this case I think "faking" a refactor is not that effective. Of course, I'll be glad to do that if you wish so.

djc: I think that is exactly why it is important to split this, because I want to be able to have a smaller commit that only has the semantic changes so it's easier to review it in isolation. That doesn't seem like "faking" a refactor, but rather having the refactoring as a precursor to making the semantic changes more obvious.

src/toolchain/names.rs
rami3l: Done.

src/toolchain/names.rs
rami3l: Done. That's one more allocation but I think it's negligible in this case anyway.

src/toolchain/names.rs
rami3l: @djc I've split the history into 7 commits, and only [c27f007](https://github.com/rust-lang/rustup/pull/3518/commits/c27f007695d7776b1e88fd58c0056790682eab86) is an actual semantic change.

The change was made because:
1. More toolchain names are available now because they will no longer cause panics in `.unwrap()`.
1. I didn't find anything in the current docs explaining what the semantics should be.

... so I improvised a bit according to my guesses (making sure that the adjusted test case is passing, of course).

djc (approved):

PR #3521, opened 2023-10-30
Suggest installing MSYS2 for  `windows-gnu`

### LINKED ISSUE #3520: Suggest installing mingw when using windows-gnu abi

Currently [the Windows docs say](https://rust-lang.github.io/rustup/installation/windows.html):

> No additional software installation is necessary for basic use of the GNU build.

While this is strictly true, the gnu libraries and tools shipped with Rust are intended to be minimal. Importantly, exactly what is shipped may change from version to version depending on internal implementation details. And even if something is bundled with Rust, it may potentially be stripped down for a single use case or else be an older version (or both).

### Solution you'd like

Make clear that the user may want to install [msys2](https://msys2.github.io/) with mingw tools, even if it's not strictly necessary.

### Notes

_No response_

rami3l:
Fixes #3520.

### DIFF
--- a/doc/user-guide/src/installation/windows.md
+++ b/doc/user-guide/src/installation/windows.md
@@ -7,11 +7,13 @@ the native (MSVC) ABI used by [Visual Studio], and the GNU ABI used by the
 [GCC toolchain]. Which version of Rust you need depends largely on what C/C++
 libraries you want to interoperate with: for interop with software produced by
 Visual Studio use the MSVC build of Rust; for interop with GNU software built
-using the [MinGW/MSYS2 toolchain] use the GNU build.
+using the [MinGW/MSYS2 toolchain][MSYS2] use the GNU build.
 
 When targeting the MSVC ABI, Rust additionally requires an [installation of
 Visual Studio][msvc install] so `rustc` can use its linker and libraries.
-No additional software installation is necessary for basic use of the GNU build.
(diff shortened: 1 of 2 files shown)


doc/user-guide/src/installation/windows.md
```diff
@@ -7,11 +7,13 @@ the native (MSVC) ABI used by [Visual Studio], and the GNU ABI used by the
...
 Visual Studio use the MSVC build of Rust; for interop with GNU software built
-using the [MinGW/MSYS2 toolchain] use the GNU build.
+using the [MinGW/MSYS2 toolchain][MSYS2] use the GNU build.
 
 When targeting the MSVC ABI, Rust additionally requires an [installation of
 Visual Studio][msvc install] so `rustc` can use its linker and libraries.
-No additional software installation is necessary for basic use of the GNU build.
+
+When targeting the GNU ABI, no additional software is strictly required for basic use.
+However, you may still want to install [MSYS2] with MinGW tools for the best experience.
```
rbtcollins: I think a little extra detail would help.
```suggestion
Many library crates will not be able to compile until the full  [MSYS2] with MinGW has been installed..
```

rami3l: Done.

djc (approved):

PR #3522, opened 2023-11-03
Fix the test toolchain_broken_symlink on Windows

### LINKED ISSUE #3516: The test `toolchain_broken_symlink` is failing on Windows
The test `suite::cli_misc::toolchain_broken_symlink` is failing on my Windows 10 machine.

I tested this on commit b5d74ac69307de29cee586c779bdb6b3aee7f7af

I ran `cargo test --features=test toolchain_broken_symlink` and got this output:

```
...
---- suite::cli_misc::toolchain_broken_symlink stdout ----
ran: rustup ["set", "default-host", "x86_64-pc-windows-msvc"]
inprocess: true
status: Some(0)
duration: 0.170s
stdout:
====

====

stderr:
====

====

ran: rustup ["set", "auto-self-update", "disable"]
inprocess: true
status: Some(0)
duration: 0.056s
stdout:
====

====

stderr:
====
info: auto-self-update mode set to 'disable'

====

thread 'suite::cli_misc::toolchain_broken_symlink' panicked at tests\suite\cli_misc.rs:680:35:
called `Result::unwrap()` on an `Err` value: Os { code: 1314, kind: Uncategorized, message: "A required privilege is not held by the client." }


failures:
    suite::cli_misc::toolchain_broken_symlink

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 373 filtered out; finished in 1.71s

error: test failed, to rerun pass `--test test_bonanza`
```

majaha:
This fixes https://github.com/rust-lang/rustup/issues/3516

`rustup` on Windows uses "directory junctions" instead of symbolic links for `rustup toolchain link`. These are somewhere between a symbolic and a hard link, and require fewer permissions. However, the test in question was trying to use symbolic links. It now uses the same codepath to create the link as `toolchain link` itself.

### DIFF
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -131,7 +131,7 @@ pub fn append_file(dest: &Path, line: &str) -> io::Result<()> {
     Ok(())
 }
 
-pub(crate) fn symlink_dir(src: &Path, dest: &Path) -> io::Result<()> {
+pub fn symlink_dir(src: &Path, dest: &Path) -> io::Result<()> {
     #[cfg(windows)]
     fn symlink_dir_inner(src: &Path, dest: &Path) -> io::Result<()> {
         // std's symlink uses Windows's symlink function, which requires
--- a/tests/suite/cli_misc.rs
+++ b/tests/suite/cli_misc.rs
@@ -665,29 +665,22 @@ fn rename_rls_remove() {
 #[test]
 #[cfg(any(unix, windows))]
 fn toolchain_broken_symlink() {
+    use rustup::utils::raw::symlink_dir;
     use std::fs;
     use std::path::Path;
-
-    #[cfg(unix)]
-    fn create_symlink_dir<P: AsRef<Path>, Q: AsRef<Path>>(src: P, dst: Q) {
-        use std::os::unix::fs;
-        fs::symlink(src, dst).unwrap();
-    }
-
-    #[cfg(windows)]
-    fn create_symlink_dir<P: AsRef<Path>, Q: AsRef<Path>>(src: P, dst: Q) {
-        use std::os::windows::fs;
-        fs::symlink_dir(src, dst).unwrap();
-    }
 
     clitools::test(Scenario::None, &|config| {
         // We artificially create a broken symlink toolchain -- but this can also happen "legitimately"
         // by having a proper toolchain there, using "toolchain link", and later removing the directory.
         fs::create_dir(config.rustupdir.join("toolchains")).unwrap();
-        create_symlink_dir(
-            config.rustupdir.join("this-directory-does-not-exist"),
-            config.rustupdir.join("toolchains").join("test"),
-        );
+        fs::create_dir(config.rustupdir.join("this-directory-does-not-exist")).unwrap();
+        symlink_dir(
+            &config.rustupdir.join("this-directory-does-not-exist"),
+            &config.rustupdir.join("toolchains").join("test"),
+        )
+        .unwrap();
+        fs::remove_dir(config.rustupdir.join("this-directory-does-not-exist")).unwrap();
+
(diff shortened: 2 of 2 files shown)


src/utils/raw.rs
```diff
@@ -131,7 +131,7 @@ pub fn append_file(dest: &Path, line: &str) -> io::Result<()> {
     Ok(())
 }
 
-pub(crate) fn symlink_dir(src: &Path, dest: &Path) -> io::Result<()> {
+pub fn symlink_dir(src: &Path, dest: &Path) -> io::Result<()> {
```
majaha, line 134: I had to change this from `pub(crate)` to `pub` to call it, I hope that's not a problem.

tests/suite/cli_misc.rs
```diff
@@ -667,27 +667,19 @@ fn rename_rls_remove() {
...
-        create_symlink_dir(
-            config.rustupdir.join("this-directory-does-not-exist"),
-            config.rustupdir.join("toolchains").join("test"),
-        );
+        fs::create_dir(config.rustupdir.join("this-directory-does-not-exist")).unwrap();
+        symlink_dir(
+            &config.rustupdir.join("this-directory-does-not-exist"),
+            &config.rustupdir.join("toolchains").join("test"),
+        ).unwrap();
+        fs::remove_dir(config.rustupdir.join("this-directory-does-not-exist")).unwrap();
```
majaha, line 682: `symlink_dir()` calls `fs::canonicalize()`, which requires the path to actually exist, so I had to create the directory and delete it afterwards.

src/utils/raw.rs
rami3l: I think it's okay, the `rustup` lib APIs are currently not intended for third-party usages anyway.

rami3l (dismissed):
Thanks a lot for the contribution, and this patch looks good to me!

djc (approved):

djc:
Thanks!

rami3l:
@djc Oops, I guess you missed my dismissed review?

rami3l:
@djc Sorry I had a bad network so the following message didn't get out:

> I just asked @rbtcollins and his suggestion is that we should not add `pub` API here unless it's covered by `#[cfg(feature = "test")]`.
> 
> In this case, I suggest keeping the original `pub(crate) fn` and adding a `pub fn` guarded by the `"test"` feature that wraps over it. Something similar to what's done in https://stackoverflow.com/a/47698997:
> 
> ```rs
> #[cfg(feature = "test")]
> pub fn test_create_symlink_dir(src: &Path, dest: &Path) -> io::Result<()> {
>     symlink_dir(src, dest)
> }
> ```

This requires a minor follow-up PR I guess.

majaha:
I had a grep through the codebase and found that there are several cases of the same sort of pattern, so I think it might be fine? @rbtcollins what do you think?

https://github.com/rust-lang/rustup/blob/13e53706aefc59527c56d03a6118f0240811bf7d/tests/suite/cli_exact.rs#L451
https://github.com/rust-lang/rustup/blob/13e53706aefc59527c56d03a6118f0240811bf7d/tests/suite/cli_v1.rs#L170
https://github.com/rust-lang/rustup/blob/13e53706aefc59527c56d03a6118f0240811bf7d/tests/suite/cli_v2.rs#L265
https://github.com/rust-lang/rustup/blob/13e53706aefc59527c56d03a6118f0240811bf7d/tests/suite/cli_self_upd.rs#L786

rbtcollins:
@majaha we're in a (long) transition to get an actual public API that could in principle be used to interact with rustup. That really just means that only things we intend to support in the public crate api are in the api.

And the steps towards that are:
- removing things that were just exported for testing convenience
- adding things we decide to export 

Those existing under-the-hood examples are all places we consider bugs to fix.

PR #3524, opened 2023-11-06
Assorted dependency updates

djc:


Cargo.toml
```diff
@@ -60,16 +56,21 @@ flate2 = "1"
 fs_at.workspace = true
 git-testament = "0.2"
 home = "0.5.4"
-lazy_static.workspace = true
 libc = "0.2"
-once_cell = { workspace = true, optional = true }
+once_cell.workspace = true
 opener = "0.6.0"
+# Used by `curl` or `reqwest` backend although it isn't imported by rustup:
```
rami3l: I see what you mean here, but I'm afraid for someone without any context of this change, these comments might seem to refer to `openssl` and some other stuff below.

The following should be able to limit the scope of this comment to `openssl`:

```suggestion
# `openssl` is used by `curl` or `reqwest` backend although it isn't imported by rustup:
```

src/dist/triple.rs
```diff
@@ -68,15 +68,17 @@ impl PartialTargetTriple {
...
                 LIST_ARCHS.join("|"),
                 LIST_OSES.join("|"),
                 LIST_ENVS.join("|")
-            );
-            static ref RE: Regex = Regex::new(&PATTERN).unwrap();
-        }
+            )
+        });
+
+        static RE: Lazy<Regex> = Lazy::new(|| Regex::new(&PATTERN).unwrap());
```
rami3l: Maybe we can inline `PATTERN` as long as it's only used for `RE`'s initialization? The current form gives me the impression that the former is used elsewhere...

Cargo.toml
djc: Fair, fixed as suggested!

src/dist/triple.rs
djc: Yeah, good idea -- took care of that for both instances of this pattern.

rami3l (approved):
Oops, just got sidetracked by another issue for a while. LGTM!

PR #3543, opened 2023-11-25
Upgrade hyper to 1.0

dzvon:
hyper 1.0 is a breaking change, so we need to update the code to use the
new API. This commit also introduces two new dependencies, hyper-util
and http-body-util, which are used for functionality that was previously
provided by hyper 0.14.

### DIFF
--- a/download/Cargo.toml
+++ b/download/Cargo.toml
@@ -23,6 +23,8 @@ thiserror.workspace = true
 url.workspace = true
 
 [dev-dependencies]
-hyper = { version = "0.14", default-features = false, features = ["tcp", "server"] }
+http-body-util = "0.1.0"
+hyper = { version = "1.0", default-features = false, features = ["server", "http1"] }
+hyper-util = { version = "0.1.1", features = ["tokio"] }
 tempfile.workspace = true
 tokio = { workspace = true, default-features = false, features = ["sync"] }
--- a/download/tests/support/mod.rs
+++ b/download/tests/support/mod.rs
@@ -6,9 +6,11 @@ use std::path::Path;
 use std::sync::mpsc::{channel, Sender};
 use std::thread;
 
-use hyper::server::conn::AddrStream;
-use hyper::service::{make_service_fn, service_fn};
-use hyper::{Body, Request};
+use http_body_util::Full;
+use hyper::body::Bytes;
+use hyper::server::conn::http1;
+use hyper::service::service_fn;
+use hyper::Request;
 use tempfile::TempDir;
 
 pub fn tmp_dir() -> TempDir {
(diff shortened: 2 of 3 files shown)


download/Cargo.toml
```diff
@@ -23,6 +23,8 @@ thiserror.workspace = true
 url.workspace = true
 
 [dev-dependencies]
-hyper = { version = "0.14", default-features = false, features = ["tcp", "server"] }
+hyper = { version = "1.0", default-features = false, features = ["server", "http1"] }
+hyper-util = { version = "0.1.1", features = ["tokio"] }
+http-body-util = "0.1.0"
```
rami3l: We might want to sort the deps alphabetically here.

download/tests/support/mod.rs
```diff
@@ -32,25 +34,34 @@ pub fn write_file(path: &Path, contents: &str) {
 }
 
 async fn run_server(addr_tx: Sender<SocketAddr>, addr: SocketAddr, contents: Vec<u8>) {
```
rami3l, line 39: IMHO it would be useful if you could add a comment explaining this function is adapted from https://hyper.rs/guides/1/server/hello-world. I found it out myself but maybe I wouldn't have to :)

rami3l:
Supersedes https://github.com/rust-lang/rustup/pull/3534.

@dzvon Thanks a lot for the PR! This is really helpful given the currently limited bandwidth on my side 🙇

download/Cargo.toml
dzvon: Done. Nice habit :D

download/tests/support/mod.rs
dzvon: Agree. The guide link will be helpful.

download/tests/support/mod.rs
```diff
@@ -31,26 +33,39 @@ pub fn write_file(path: &Path, contents: &str) {
     file.sync_data().expect("writing test data");
 }
 
+// This is a dead simple hyper server implementation, if we want to know more about
+// how to create a hyper server, we can check the following link:
+//
+// https://hyper.rs/guides/1/server/hello-world/
```
rami3l: Nit: This could be made even simpler :)

```suggestion
// A dead simple hyper server implementation.
// For more info, see:
// https://hyper.rs/guides/1/server/hello-world/
```

download/Cargo.toml
```diff
@@ -23,6 +23,8 @@ thiserror.workspace = true
 url.workspace = true
 
 [dev-dependencies]
-hyper = { version = "0.14", default-features = false, features = ["tcp", "server"] }
+http-body-util = "0.1.0"
```
rami3l, line 26: This change should be squashed into the first commit instead of the second one.

dzvon: Sure, I thought I could squash all the commits once everything is done, or you could do it through the _Merge pull request_ option. Either way is fine with me.

rami3l: Okay, I'll squash it when merging then.

dzvon: @rami3l Never mind, bro. I squashed it already. Hahaha :D

rami3l (approved):

Issue #1728, opened 2019-03-27
Miri should be presented as miri-preview until it's stable


cuviper:
**Describe the problem you are trying to solve**

We would like the miri component to be presented as "miri-preview" until it has been deemed stable.

**Describe the solution you'd like**

Any advice on how to accomplish this would be fine. :smile:

**Notes**

The component was renamed to `miri-preview` in rust-lang/rust#59236, so the current `channel-rust-nightly.toml` describes miri like this:

```toml
[pkg.miri-preview]
version = "0.1.0-nightly (72b4ee038 2019-03-11)"
[pkg.miri-preview.target.aarch64-unknown-linux-gnu]
available = true
url = "https://static.rust-lang.org/dist/2019-03-27/miri-nightly-aarch64-unknown-linux-gnu.tar.gz"
hash = "17c2b1ad92b2621460c6ae2bc3025b32d088c7ba15ef65c5da003645fd348d5d"
xz_url = "https://static.rust-lang.org/dist/2019-03-27/miri-nightly-aarch64-unknown-linux-gnu.tar.xz"
xz_hash = "e3683df7c976f029eec95f693cd302ddb003d01444db917613d20386595d0744"

# ... and other targets ...

[renames.miri]
to = "miri-preview"
```

As a result, you can use either `rustup component add miri` or `miri-preview`. This is fine, but the displayed "info" always calls it just `miri`, as does `rustup component list`. We ought to still present this as `miri-preview` until the tool is actually deemed stable.

I found that the other tools like `rustfmt`, which *are* stable, are defined the same way, with `[pkg.rustfmt-preview]` and `[renames.rustfmt] to = "rustfmt-preview"`. We should continue to use the non-preview name for presenting these stable tools, which is why I think we may need some enhancement for miri's case, without affecting existing tool manifests.

Maybe we could just "flip" miri for now, defining `[pkg.miri]` and then rename the other way `[renames.miri-preview] to = "miri"`. That feels like a backwards kludge, but I guess we wouldn't need rustup changes then.

cc @rust-lang/release @RalfJung @oli-obk @mati865

mati865:
I wanted to go `miri-preview` but I didn't found way to test it locally. In some places it was called `miri` and in the other it was `miri-preview` so I just made sure it works either way.
You can see it's history here: https://github.com/rust-lang/miri/issues/520

cuviper:
[...]

PR #3578, opened 2023-12-11
'llvm-tools-preview' component is now named 'llvm-tools'

mbrossard:
The `llvm-tools-preview` component is now name `llvm-tools`. It seems this change happened in version 1.67 since `rustup +1.67 component add llvm-tools` succeeds but `rustup +1.66 component add llvm-tools` fails.

### DIFF
--- a/doc/user-guide/src/concepts/components.md
+++ b/doc/user-guide/src/concepts/components.md
@@ -57,8 +57,7 @@ toolchains. The following is an overview of the different components:
   [RLS].
 * `rust-mingw` — This contains a linker and platform libraries for building on
   the `x86_64-pc-windows-gnu` platform.
-* `llvm-tools-preview` — This is an experimental component which contains a
-  collection of [LLVM] tools.
+* `llvm-tools` — This component contains a collection of [LLVM] tools.
 * `rustc-dev` — This component contains the compiler as a library. Most users
   will not need this; it is only needed for development *of* tools that link
   to the compiler, such as making modifications to [Clippy].


ehuss:
I'm a little uncertain if this is the right change to make. The llvm-tools component is not considered stable, and should keep the preview tag. I think it is actually a bug that it works without the `-preview` part. I can't remember or find the history of why `-preview` gets ignored (I think it has something to do with how renames were implemented). It looks like #1728 tried to at least fix it so that miri appeared as `-preview`, but even that seemed to not work.

mbrossard:
It seems https://static.rust-lang.org/dist/2023-01-26/channel-rust-1.67.toml contains:
```
[renames.llvm-tools]
to = "llvm-tools-preview"
```
Whereas https://static.rust-lang.org/dist/2022-12-15/channel-rust-1.66.toml did not have it.

It looks related to https://github.com/rust-lang/rust/commit/a3dd94e702f2d640f5f84f9f43f25fd79dfeae95
(or related changes in https://github.com/rust-lang/rust/commits/master/src/tools/build-manifest/src/main.rs around Oct 27th 2022).

rbtcollins:
This is the correct change as of  1.67. You can see in the manifest - https://static.rust-lang.org/dist/channel-rust-stable.toml - that llvm-tools is renamed to llvm-tools-preview: so the package real name is unaltered, but users should now enter llvm-tools.

However, rustup is not connected to that versioning scheme, but its a problem that becomes rarer over time.

rbtcollins (approved):

smoelius:
> This is the correct change as of 1.67. You can see in the manifest - https://static.rust-lang.org/dist/channel-rust-stable.toml - that llvm-tools is renamed to llvm-tools-preview: so the package real name is unaltered, but users should now enter llvm-tools.

@ehuss Is this your conclusion as well? I had the impression from https://github.com/rust-lang/rust/issues/85658#issuecomment-1889988970 that you had the opposite conslusion.

ehuss:
Correct, there hasn't been a decision on this, yet, so I think it may be premature to merge this.

Although, the cat is out of the bag, so I'm not sure if we can put it back. I'll start a topic with the release team and see if they have any thoughts.

smoelius:
@ehuss Please forgive me for bothering you, but were you able to determine whether `llvm-tools` is now the official term, as opposed to `llvm-tools-preview`?

ehuss:
The discussion was at https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/llvm-tools.20stabilization.

I think the consensus is that we shouldn't remove `llvm-tools` and break people using it. However, it probably should not be considered stabilized. There are no guarantees about which tools are included, no documentation about which tools are included, where they are located (and some thinking that they are currently in the wrong directory), or how to find them.

I'll copy one of the last comments here:

> I agree that simply saying "llvm-tools may break at any time" is uncomfortable, but I think the fairest statement is that "llvm-tools are provided as-is from upstream LLVM". In most cases, they aren't going to break all the time or even occasionally. At the same time, we're not going to take on maintaining a tool LLVM has dropped just to fulfil a stability promise. I think the vast majority of users would totally understand this position and not have an issue with that.
> 
> This feels very similar to the Target Tier Policy where we do not promise perpetual support for any particular target, just a level of support in a given release. Breakage in upstream LLVM has even necessitated demoting targets from Tier 2 to Tier 3 because they're simply too broken to be at that level.

I think further tracking of stabilization would be at https://github.com/rust-lang/rust/issues/85658. I posted https://github.com/rust-lang/rustup/pull/3707 to add a note indicating the current status.

smoelius:
Thank you very much for the detailed response, which was far more than I'd hoped for!

PR #3579, opened 2023-12-12
CI: Enable openssl and rustls on loongarch64

heiher:
This PR updates `openssl-src` to `v300.2.1+3.2.0`, reverts the disabling of `openssl` and enabling `rustls` on loongarch64.


djc (approved):
Thanks for following up!

rami3l:
@djc The Android build failed but maybe using https://github.com/Stebalien/xattr/releases/tag/v1.1.3 will help. https://github.com/rust-lang/rustup/pull/3582 also hit this today. I suggest rebasing #3582 after a manual upgrade, whereas I think it'd be okay if I merge this directly.

Issue #3567, opened 2023-12-07
Meta: Consider introducing GitHub Merge Queue to the repo


rami3l:
> Maybe it's time to introduce a "conclusion" CI action and a merge queue. We already have `automerge=true` configured in renovate, but since there's no "conclusion" configured in our branch protection settings, auto-merging will not work. OTOH once that's been added, renovate will automatically put semver-compatible updates into the merge queue (cf. https://github.com/renovatebot/renovate/issues/20569).

_Originally posted by @rami3l in https://github.com/rust-lang/rustup/issues/3565#issuecomment-1845309099_

--- 

## Tasks

- [x] Merge https://github.com/rust-lang/rustup/pull/3582.
- [x] Merge #3633.
- [x] Tell the mods to enable GHMQ and to point the branch protection rules to the newly-added `conclusion` job.

djc:
Sounds good to me!

rami3l:
@djc For the record, an example of the `conclusion` step can be found in [PyO3](https://github.com/PyO3/pyo3/blob/42601f3af94242b017402b763a495798a92da8f8/.github/workflows/ci.yml#L452-L472), however this will probably require moving all CI workflows into a single file (which should be done with a refactoring of `ci/actions-templates/gen-workflows.sh`), so this might take a while.

rami3l:
~~New concern raised in https://github.com/rust-lang/rustup/pull/3582#issuecomment-1877287794.~~

Update: Addressed in #3627.

rami3l:
GHMQ is now online with "rebase and merge" as the default action.

Thanks a lot for your help and support, @jdno and @djc!

PR #3582, opened 2023-12-14
ci: streamline CI actions to prepare for merge queue introduction

rami3l:
This is one of the prerequisites for closing https://github.com/rust-lang/rustup/issues/3567.

## Changes

- [x] Merge all `*-on-*.yaml` into `ci.yaml`.
- [x] Merge `all-features` into `ci.yaml`.
- [x] Merge `test-docs` into `ci.yaml`.
- [x] Add `freebsd-builds` to `ci.yaml`.
  - [x] Disassemble and reorganize [ci/cirrus-templates](https://github.com/rust-lang/rustup/tree/6115956bafb0ae1e27385f0acad928aa15167f39/ci/cirrus-templates).
- [x] Add `conclusion` to `ci.yaml`.
- [x] Make the CI fail if the generated `ci.yaml` is not up to date.


ci/cirrus-templates/script.bash
```diff
@@ -14,7 +14,7 @@ echo "========="
 echo "Display the current git status"
 
 git status
-git describe
+git describe --always
```
rami3l: Wow, moving the FreeBSD CI to GitHub Actions is surprisingly doable (and we're not losing too much in terms of speed either)! The only problem is that `git describe` will fail.

```console
  Display the current git status
  + echo =========
  + echo 'Display the current git status'
  + git status
  HEAD detached at pull/3582/merge
  nothing to commit, working tree clean
  + git describe
  fatal: No tags can describe '8e5f1d4501d79fbd321b83d57f4cdb6dd4ae121b'.
  Try --always, or create some tags.
```

... whereas ideally we should see something like:

```console
Display the current git status
+ git status
HEAD detached at pull/3582/merge
nothing to commit, working tree clean
+ git describe
1.26.0-1143-gda432be5
```

Changing to `git describe --always` will work here, but later on one of the tests will fail instead, possibly with the version string.

```console
  Display the current git status
  + echo =========
  + echo 'Display the current git status'
  + git status
  HEAD detached at pull/3582/merge
  nothing to commit, working tree clean
  + git describe --always
  674a79e

...

  ran: rustup ["self", "update"]
  inprocess: false
  status: Some(0)
  duration: 0.077s
  stdout:
  ====
    rustup updated - 674a79e2c (from 1.26.0)
  
  
  ====
  
  stderr:
  ====
  info: downloading self-update
  
  ====
  
  
  > rustup self update
  out.ok: true
  out.stdout (2 lines):
        rustup updated - 674a79e2c (from 1.26.0)
      
      
  out.stderr (2 lines):
      info: downloading self-update
      
  expected.ok: true
  expected.stdout (2 lines):
        rustup updated - 1.26.0 (from 1.26.0)
      
      
  expected.stderr (2 lines):
      info: downloading self-update
      
  [src/test/mock/clitools.rs:603] out.stdout == stdout = false
  [src/test/mock/clitools.rs:604] out.stderr == stderr = true
  thread 'suite::cli_self_upd::update_exact' panicked at tests/suite/cli_self_upd.rs:300:16:
  explicit panic
```

rami3l: Solved by using `git fetch --prune --unshallow --tags` (cc https://github.com/actions/checkout/issues/206).

rami3l: Weirdly enough, what worked yesterday shows the following today:

```console
+ git fetch --prune --unshallow --tags
  fatal: shallow file has changed since we read it
```

rami3l: I have finally solved this issue by fetching the tags outside of FreeBSD.

.github/workflows/ci.yaml
rami3l, line 1: This file is generated by the script, please ignore it while reviewing.

PS: Actually, it's probably better to ignore `.github/workflows/**` changes entirely.

djc (approved):
Yeah, I think that makes sense. Thanks for working through this!

rami3l:
@djc Concern: If I'm not mistaken, the conditional `if: ${{ (github.event.push && github.ref_name == 'master') || github.event.schedule }}` doesn't seem to work (returns false on the latest `master` push), and it unfortunately hasn't been tested on PR.

The conditional is necessary since we want the workflow to always run to qualify for being used in branch protection.

What could've happened?

PS: This should not influence the future merge queue integration though, since it depends on a separate `merge_group` case.

rami3l:
... I checked the docs again and I don't see why, but probably we should use `github.event_name == 'push'` instead of `github.event.push`?

PR #3584, opened 2023-12-14
Fix rustup-init failure to read ZDOTDIR from zsh when SHELL is not zsh

ghost:
If the SHELL environment variable does not include "zsh", rustup-init attempts to call zsh to see if ZDOTDIR is set.

It intends to call
```
    zsh -c 'echo $ZDOTDIR'
```
but actually called
```
    zsh -c "'echo $ZDOTDIR'"
```

The second form fails with this output to stderr:
```
    zsh:1: command not found: echo $ZDOTDIR
```

Additionally, the default "echo" output includes a newline, and this was being added to the PathBuf.  This would cause later is_file() checks to fail, ignoring the rcfile.

Added a test, and tested manually.

```
mkdir ~/my-zdot-dir
touch ~/my-zdot-dir/.zshenv
cargo build && ZDOTDIR=~/my-zdot-dir SHELL=/bin/sh target/debug/rustup-init
```
    
On master, notice ~/.zshenv is modified
On this branch, notice ~/my-zdot-dir/.zshenv is modified

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1344,9 +1344,9 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "openssl-src"
-version = "300.2.0+3.2.0"
+version = "300.2.1+3.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


djc (approved):
Thanks, this looks good to me!

PR #3591, opened 2023-12-18
docs: move "rls" and "rust-analysis" to separate section "previous…"

burakemir:
I had created a PR before #3410 but the source file moved, I messed up and decided to start clean.

As requested, just moves the component description into a new section instead of deleting.
This PR addresses part of #3316.

Description from #3410 copied over for convenience:
RLS was deprecated over a year ago. https://blog.rust-lang.org/2022/07/01/RLS-deprecation.html
The rust-analysis component seems to only contain a placeholder file with the contents
{"warning": "The `rust-analysis` component has been removed." }"
I checked a few occurrences in rustup toolchain manifest files, and they only seem to contain that "removed.json".
I believe no one will miss these things.


rbtcollins (approved):
thank you for your patch!

Issue #3501, opened 2023-10-05
1.27.0 release planning


djc:
@alex asked (privately) for a release after #3488 and #3490 so let's use this issue for tracking remaining work that we'd like to include in the next release and any other process surrounding the release process?

@rbtcollins can you share any context/thoughts around how releases have been managed in the past?

(I'm calling it 1.26.1 since I don't think there are any large changes in the public interface since the previous release, but I haven't checked that deeply so maybe 1.27.0 might be warranted?)

rami3l:
@djc As a regular user I'd like to see `1.27.0`. You got us clap4 after all (#3444).

PS: I'm still using `rustup 1.26.0`, so you probably meant `1.26.1` in the title?

djc:
Ahh yes -- I somehow seemed to remember that 1.26.2 was the latest release.

So why do you think it should be 1.27.0? Just because there are "new features"?

rami3l:
> So why do you think it should be 1.27.0? Just because there are "new features"?

Do you remember the large amount of diffs that you had to change in order to pass the black box tests (for clap4)? I'm afraid someone is depending on that, though (s)he probably shouldn't.

rami3l:
I think we can probably add a new milestone here? There are a few PRs that I want to address before cutting a new release 👀

djc:
That works too. I like having a release planning issue because it's something interested users can subscribe to to keep up to date with the progress and comments can be added to discuss scope changes -- so in other projects I usually like to use task lists in a planning issue like this one.

rami3l:
@djc Good. Let's keep it this way then.

I'll keep my list of issues "of interest" here:
* https://github.com/rust-lang/rustup/pull/3453
* https://github.com/rust-lang/rustup/pull/3108
  * https://github.com/rust-lang/rustup/pull/3506
* https://github.com/rust-lang/rustup/pull/3475
* https://github.com/rust-lang/rustup/pull/3464
* https://github.com/rust-lang/rustup/pull/3068
* https://github.com/rust-lang/rustup/pull/3225

The idea is that these are mostly PRs that have been reviewed but require checking again.

rbtcollins:
The release process is documented in the rustup contributors guide. https://rust-lang.github.io/rustup/dev-guide/release-process.html

[tribal] We often ask for testers on reddit / rust-users
[...]

PR #3596, opened 2023-12-21
refactor(cli): rewrite `rustup` with `clap_derive`

### LINKED ISSUE #3549: Consider using `clap_derive` for argument parsing
Migrating to [`clap_derive`](https://docs.rs/clap/latest/clap/_derive/) while utilizing the [ValueParser selection mechanism](https://docs.rs/clap/latest/clap/macro.value_parser.html) should help with avoiding strange panics like the ones described in #3548.

cc https://github.com/rust-lang/rustup/issues/3548#issuecomment-1836043599

rami3l:
Closes #3549.

## Rationale

Currently, to add a new subcommand/argument to `rustup` or `rustup-init`, one will have to manually maintain consistencies across 3 places (`Command`, `ArgMatches` and dispatching). Failing to do so will unavoidably lead to panics.

Hopefully, with `clap_derive`, it'll be much easier to dive in the codebase and make changes, as:
- All the subcommands/arguments will be explicit (not hidden in the specific dispatching functions, like `run()` for `rustup run`);
- To add a new subcommand/argument there are no more 3, but only 2 places to change (`Parser` and dispatching), and this time one will be covered by strongly-typed APIs.

Initially, I decided to experiment with `rustup-init` since it's relatively isolated, and I was surprised to find out that even with such a simple command, migrating to `clap_derive` already allows us to address some inconsistencies.

Given that successful experience, I think a complete migration will be not only feasible but fruitful, which can hopefully reduce accidental complexities in maintaining this project.

~~PS: Please don't merge this before closing https://github.com/rust-lang/rustup/issues/3501.~~

## Progress

- [x] ~~rustup-init~~ (became #3814)
- [x] rustup
  - [x] rustup install [^2][^3]
  - [x] rustup uninstall [^2][^3]
  - [x] rustup dump-testament [^2]
  - [x] rustup show [^1]
  - [x] rustup update
  - [x] rustup check
  - [x] rustup default
  - [x] rustup toolchain [^1]
  - [x] rustup target [^1]
  - [x] rustup component [^1]
  - [x] rustup override [^1]
  - [x] rustup run
  - [x] rustup which
  - [x] rustup doc
  - [x] rustup man
  - [x] rustup self [^1]
  - [x] rustup set [^1]
  - [x] rustup completions

[^1]: This command has subcommands.
[^2]: This command is hidden.
[^3]: This command is implemented as an alias of the corresponding subcommand under `rustup toolchain`.


djc:
Thanks for working on this!

> Please don't merge this before closing #3501.

Why do you want to keep this out of 1.27.0? It maybe doesn't need to go in, but I don't see any harm from it being merged?

(Is that the only reason this is in Draft state, or are there other reasons?)

rami3l:
> Why do you want to keep this out of 1.27.0? It maybe doesn't need to go in, but I don't see any harm from it being merged?
> 
> (Is that the only reason this is in Draft state, or are there other reasons?)

@djc Oh, it's the only reason actually. IMO the overall migration is a huge amount of work that consists of multiple patches and I think it's quite weird to ship a new version half-migrated.

I plan to do this on a per-subcommand basis when it comes to `rustup`, but since `rustup-init` is quite self-contained I think it's also possible to merge it if the rest of the team thinks it's okay?

src/cli/setup_mode.rs
```diff
@@ -1,112 +1,92 @@
...
+    /// Choose a default toolchain to install. Use 'none' to not install any toolchains at all
+    #[arg(long, value_parser = maybe_official_toolchainame_parser)]
+    default_toolchain: Option<MaybeOfficialToolchainName>,
+
+    #[arg(long, value_parser = PossibleValuesParser::new(Profile::names()), default_value = Profile::default_name())]
+    profile: String,
+
+    /// Component name to also install
+    #[arg(short, long, value_delimiter = ',', num_args = 1..)]
+    components: Option<Vec<String>>,
```
djc: I think that for `components` and `targets` we can just use `Vec<String>` rather than `Option<Vec<String>>`? Seems simpler.

src/cli/setup_mode.rs
```diff
@@ -116,39 +96,28 @@ pub fn main() -> Result<utils::ExitCode> {
...
-        no_update_toolchain,
-        components: &components,
-        targets: &targets,
+        no_update_toolchain: no_update_default_toolchain,
+        components: &components
+            .iter()
+            .flatten()
+            .map(|s| &**s)
+            .collect::<Vec<_>>(),
+        targets: &targets.iter().flatten().map(|s| &**s).collect::<Vec<_>>(),
```
djc: This feels a little silly? It's probably straightforward to instead change `InstallOpts` to hold a `Vec<String>` for both `components` and `targets`.

src/toolchain/names.rs
```diff
@@ -95,6 +95,14 @@ macro_rules! try_from_str {
                 $to::validate(&value)
             }
         }
+
+        impl FromStr for $to {
```
djc: Exactly what I had in mind, nice!

djc (approved):

djc:
I can see how it might be weird to ship half-migrated so I guess I agree maybe we shouldn't merge this yet until we have the other CLIs migrated.

src/cli/setup_mode.rs
rami3l: You've probably guessed it: this happens for a reason.

Here I'm sticking to the type in the original code because this combined with `num_args = 1..` allows us to generate `-t, --targets <TARGETS>...` in the help string.

The initial version of this PR used `Vec<String>` and IIRC it generated something like `-t, --targets <TARGETS>`  instead.

src/cli/setup_mode.rs
rami3l: @djc This is indeed a bit awkward.

I think this change it's a bit OT but I guess it's okay to add a tiny refactoring to this PR though, if the size of this change remains relatively small. Let me have a look...

src/cli/setup_mode.rs
djc: So this is just to keep the `...` in the CLI docs around? IMO it would be better to just lose the `...`.

rami3l: @djc I believe there's also a corner case where someone could run `rustup-init targets=`. This is forbidden by the original but would be allowed by `Vec<String>`. Do we want that though?

rami3l: @djc Oh, I think I've found a way to do that properly: `Vec<String>` can be used with `num_args = 1..` and we can get the best of both worlds. I'll do that instead. (Please note that `num_args = 0..` will allow `rustup-init --targets` which is quite weird.)

src/cli/setup_mode.rs
rami3l: @djc I've changed my mind...

It seems that this `InstallOpts` is just passing this `&[&str]` down to other functions. I think it's quite standard to have `&[&str]` as function arguments, although this would mean that the `|s| &**s` dance will certainly happen in one way or another.. (I do this in my project as well, although I prefer `AsRef::as_ref`.)

cc https://users.rust-lang.org/t/vec-string-to-str/12619

src/cli/setup_mode.rs
```diff
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use clap::{builder::PossibleValuesParser, Parser};
+use itertools::Itertools;
```
djc: IMO this is a few different changes that warrant being in a separate commit:

1. Use itertools for `collect_vec()`
2. Import `ErrorKind`
3. Re-order the warning logging
4. Avoid unnecessary mapping

I'm also not a big fan of (1). IMO the use of `Itertools` is a little magic, and I don't feel like `collect_vec()` is all that much easier to understand than `collect::<Vec<_>>()`.

Given that this PR might take a while to be merged if we want to change everything to derive style, maybe these smaller changes should be in a separate PR that we can merge directly? (The same might go for the change to `maybe_official_toolchain_parser()`.)

rami3l: @djc Thanks for your timely reply! I'll split this into separate PRs soon.

rami3l: @djc I've made #3644. My plan for this commit in particular is that I'll
- Abandon `.collect_vec()` now that you don't like it.
- Leave the `if` branch reorder here since it's meaningless over there.
- Put the rest over to #3644.

rami3l:
A reminder for myself: don't rebase this PR until #3644 gets merged.

src/cli/setup_mode.rs
rami3l: PR rebased in accordance with the plan above.

rami3l:
@djc How you would like to see my `rustup` porting history?

I guess porting `rustup-init` is quite independent and deserves its own commit.

However, porting `rustup` is a long and gradual process, requiring hacks and scaffolds here and there which are added to glue the code back together only to be removed later etc. which IMHO is not very interesting for reviewing purposes.
Looking at your #3444, maybe it's reasonable to just squash all those changes into one big commit in the end?

djc:
As long as there is not a lot of back and forth, I think your current commit history of doing one subcommand at a time looks pretty good actually -- I think this will make it fairly straightforward to review. Looks like you're making good progress!

rami3l:
@djc Just found out that our clap builder broke `rustfmt`, so before my simplifying that that part of the code, our source file wasn't really being formatted.

I feel a bit guilty right now about using auto-formatting on save to cause weird diffs...

PS: Also, to my surprise, will all the scaffolds removed in the end, the whole `rustup` refactor is not that hard to read in one piece.

rami3l:
A reminder for myself: don't rebase this PR until #3650 gets merged.

src/cli/rustup_mode.rs
```diff
@@ -1052,11 +1084,9 @@ fn which(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
 }
 
 #[cfg_attr(feature = "otel", tracing::instrument(skip_all))]
-fn show(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
+fn show(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCode> {
```
djc, line 924: IMO it would be nice to take care of these changes in a separate commit, since it's pretty much independent of the change to derive-based argument parsing.

src/cli/rustup_mode.rs
```diff
@@ -84,6 +85,63 @@ where
...
+    /// Display the computed value of RUSTUP_HOME
+    Home,
+
+    /// Show the default profile used for the `rustup install` command
+    Profile,
+}
+
+impl Rustup {
+    fn dispatch(self, cfg: &mut Cfg) -> Result<utils::ExitCode> {
+        match self.subcmd {
```
djc: Pre-existing issue, but maybe hoist the `handle_epipe()` call out of the `match` in a separate commit?

src/cli/rustup_mode.rs
```diff
@@ -66,25 +66,6 @@ fn handle_epipe(res: Result<utils::ExitCode>) -> Result<utils::ExitCode> {
     }
 }
 
-fn deprecated<F, B, R>(instead: &str, cfg: &mut Cfg, matches: B, callee: F) -> R
```
djc, line 70: I think the removal of this (plus any other changes that are directly related to it) should be in a separate commit.

src/cli/rustup_mode.rs
```diff
@@ -950,7 +877,7 @@ fn check_updates(cfg: &Cfg) -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
-fn update(cfg: &mut Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
+fn update(cfg: &mut Cfg, opts: UpdateOpts) -> Result<utils::ExitCode> {
```
djc, line 795: Here, too, I think it would be nice to move this change to a separate commit -- maybe even make a separate PR for changes like these? Would help digest the reviews.

src/cli/rustup_mode.rs
```diff
@@ -36,10 +36,10 @@ use crate::{
     toolchain::{
         distributable::DistributableToolchain,
         names::{
-            custom_toolchain_name_parser, maybe_resolvable_toolchainame_parser,
-            partial_toolchain_desc_parser, resolvable_local_toolchainame_parser,
-            resolvable_toolchainame_parser, CustomToolchainName, MaybeResolvableToolchainName,
-            ResolvableLocalToolchainName, ResolvableToolchainName, ToolchainName,
+            maybe_resolvable_toolchainame_parser, partial_toolchain_desc_parser,
```
djc: IMO "remove dead code" commits are an anti-pattern -- these should be removed in the commit(s) that made the code dead. To the extent that these are replaced with more idiomatic conversions, recommend putting that in a separate commit.

tests/suite/cli_misc.rs
```diff
@@ -1028,29 +1028,3 @@ fn toolchain_link_then_list_verbose() {
         config.expect_stdout_ok(&["rustup", "toolchain", "list", "-v"], "/custom-1");
     });
 }
-
-#[test]
-fn deprecated_interfaces() {
```
djc, line 1026: This change belongs in the commit that actually removed these notices.

src/cli/rustup_mode.rs
```diff
@@ -761,20 +757,21 @@ pub(crate) fn cli() -> Command {
                                 .default_value(SelfUpdateMode::default_mode()),
                         ),
                 ),
+        )
+        .subcommand(
```
djc: Can we fold these changes into the commit that originally introduced `augment_subcommands()`?

src/cli/rustup_mode.rs
```diff
@@ -795,10 +793,10 @@ fn maybe_upgrade_data(cfg: &Cfg, m: &ArgMatches) -> Result<bool> {
     }
 }
 
-fn default_(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
+fn default_(cfg: &Cfg, toolchain: Option<MaybeResolvableToolchainName>) -> Result<utils::ExitCode> {
```
djc, line 709: As before.

src/cli/rustup_mode.rs
```diff
@@ -1353,13 +1347,21 @@ fn component_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
-fn explicit_desc_or_dir_toolchain<'a>(cfg: &'a Cfg, m: &ArgMatches) -> Result<Toolchain<'a>> {
+// FIXME: Delete this.
+fn explicit_desc_or_dir_toolchain_old<'a>(cfg: &'a Cfg, m: &ArgMatches) -> Result<Toolchain<'a>> {
     let toolchain = m
         .get_one::<PartialToolchainDesc>("toolchain")
         .map(Into::into);
     explicit_or_dir_toolchain2(cfg, toolchain)
 }
 
+fn explicit_desc_or_dir_toolchain<'a>(
```
djc: IMO this one-line function isn't worth it (with 5 lines of overhead!). Just inline into the two callers for now?

src/cli/rustup_mode.rs
```diff
@@ -1355,10 +1355,10 @@ fn explicit_desc_or_dir_toolchain_old<'a>(cfg: &'a Cfg, m: &ArgMatches) -> Resul
     explicit_or_dir_toolchain2(cfg, toolchain)
 }
 
-fn explicit_desc_or_dir_toolchain<'a>(
-    cfg: &'a Cfg,
+fn explicit_desc_or_dir_toolchain(
```
djc, line 1245: Should go into the commit introducing/changing this function.

src/cli/rustup_mode.rs
```diff
@@ -1297,25 +1299,34 @@ fn target_remove(
     Ok(utils::ExitCode(0))
 }
 
-fn component_list(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
-    let toolchain = explicit_desc_or_dir_toolchain_old(cfg, m)?;
+fn component_list(
+    cfg: &Cfg,
+    toolchain: Option<PartialToolchainDesc>,
```
djc, line 1190: As before.

src/cli/rustup_mode.rs
```diff
@@ -1418,11 +1414,14 @@ fn toolchain_remove(cfg: &mut Cfg, opts: UninstallOpts) -> Result<utils::ExitCod
     Ok(utils::ExitCode(0))
 }
 
-fn override_add(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
-    let toolchain_name = m.get_one::<ResolvableToolchainName>("toolchain").unwrap();
-    let toolchain_name = toolchain_name.resolve(&cfg.get_default_host_triple()?)?;
+fn override_add(
+    cfg: &Cfg,
+    toolchain: ResolvableToolchainName,
```
djc, line 1294: As before.

djc (commented):
This is looking great! I think it would be nice to prepare for this PR by doing a separate PR that removes all the `&ArgMatches` arguments, which seems like a good idea separate from this PR because it reduces coupling between clap and the rest of the crate's components.

src/cli/rustup_mode.rs
rami3l: I'm not sure. This is a `rustfmt` bug so the code cannot be formatted with its original length, only after some truncations... I think it's not worth it to figure out how to format the original code **manually** only to see it completely removed in the end, right?

rami3l: Oops the original context is gone after a rebase... I don't know what to do now.

rami3l: Oh I have to click into every commit I've made to see the relevant contexts (looks like a GitHub issue)... But there are 17 of them already!

src/cli/rustup_mode.rs
rami3l: @djc I see at least 8 callers in the resulting code... As this is a gradual migration, the `_old()` function will be removed at the end, as clarified by the comment above it.

src/cli/rustup_mode.rs
rami3l: > IMO it would be nice to take care of these changes in a separate commit, since it's pretty much independent of the change to derive-based argument parsing.

@djc I'm not quite sure. Changes like this are a direct consequence of no longer having access to `&ArgMatches` in the `clap_derive` mode. Also, these changes are quite mechanical, in that these parameters were actually _there already_ in the original code, or more precisely, in the first lines of the function body.

For example, in this particular function, the change to use `verbose: bool` corresponds to:

https://github.com/rust-lang/rustup/blob/66b7a442c074c5cd597ea1436e1bafb0807b46e0/src/cli/rustup_mode.rs#L1064-L1067

The refactoring process is very `rustc`-directed in practice: `rustc` reports that `m: &ArgMatches` is unavailable -> move the `m`-derived variable in question to the parameter list.

Theoretically I could've done a first commit changing this parameter/moving the `m`-derived variables to the call site, and then delete those `.get_flag()` lines immediately in the next commit. However, I don't see too much of a difference between the two regarding the readability, since in the 2-commit case, the areas of those two adjacent changes won't really overlap. Also, doubling the number of commits required in this PR, I believe, will make both reviewing and rebasing more difficult.

Thus, I don't see too much additional value being created by separating this kind of changes into new commits.

I agree with the other comments you've made, thus I've rebased the PR again to address those and to ensure the atomicity of commits.

djc: IMO the reason this makes sense is that the `ArgMatches` approach creates an unnecessary coupling between the CLI frontend and the core functionality of the tools. One reason clap's derive macros are good is that they decrease the amount of coupling, but IMO it's still independently a good idea to decrease coupling in order to make the codebase more modular.

For example, if we ever wanted to offer a library interface we wouldn't want to have these `ArgMatches` arguments in there. As such, IMO it's really bad design to have these functions take `ArgMatches` directly, and correcting that makes sense even without the migration to derive macros.

(That said, if you really don't want to do the work to split these changes off I can live with that.)

rami3l: @djc Thanks a lot for your clarification, I can see what you mean now. I'm really sorry but I don't have enough bandwidth recently to spend on this PR (like I did back in December '23), so I guess I'll probably leave it as it is.

I'll make sure to ask for your opinions earlier next time 🙇

PR #3601, opened 2023-12-24
Fix component remove error by component name with explicit target triple

### LINKED ISSUE #3166: rustup component remove component-NOT-TOOLCHAIN-TARGET errors

ongchi:
Fix #3166

Follow by #3467, another approach to fix the error.

## Rationale

 The basic point of this PR is to improve the error message when Rustup somehow cannot recognize a component's full name by extracting the component name first from the manifest when parsing `rust-std-x-y-z`, so that at least `x-y-z` is intended to be a target triple. This prevents Rustup from falling back on the default target triple, producing an error message that is not very helpful (see the quote below).

The direct cause of #3166 is that `wasm32-unknown-unknown` hasn't been hardcoded into our codebase:

https://github.com/rust-lang/rustup/blob/a6c9fae915f41d611ab88ddd2228e19ea978ecaa/src/dist/triple.rs#L7-L37

... but that's another issue (later split into #3783), i.e. we really should not hardcode target triples in our artifact, as the complete list changes quite regularly.

The old `component_(add|remove)()` functions rely on the `new_with_target()` function; on the other hand, the new implementation of those two functions in this PR is based on `distributable` and `fallback_target`, which relies on detecting the component names (fetched from the distributable manifest) first, so will be more robust implementation-wise.

> For the case in #3166:
> 
> > rustup component remove rust-std-wasm32-unknown-unknown
> 
> Because of rustup cannot parse `wasm32-unknown-unknown` as a valid target, the `rust-std-wasm32-unknown-unknown` here is treated as a component name, and the target inferred to be `x86_64-apple-darwin` for this case. The error message as following:
> 
> > error: toolchain 'nightly-x86_64-apple-darwin' does not contain component 'rust-std-wasm32-unknown-unknown' for target 'x86_64-apple-darwin'
> > note: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html
> 
> The actual component and target should be `rust-std` and `wasm32-unknown-unknown` in this error message respectively.
> This is subtle because component name may also contains a '-' character (`rust-std` in this case), the component name is determined by user input subtract from possible target suffix currently (`TryFrom<PartialTargetTriple>`).

_https://github.com/rust-lang/rustup/pull/3467#issuecomment-1714078977_

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1364,8 +1364,7 @@ fn component_add(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let target = get_target(m, &distributable);
 
     for component in m.get_many::<String>("component").unwrap() {
-        let new_component = Component::new_with_target(component, false)
-            .unwrap_or_else(|| Component::new(component.to_string(), target.clone(), true));
+        let new_component = Component::try_new(component, &distributable, target.as_ref())?;
         distributable.add_component(new_component)?;
     }
 
@@ -1385,8 +1384,7 @@ fn component_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let target = get_target(m, &distributable);
 
     for component in m.get_many::<String>("component").unwrap() {
-        let new_component = Component::new_with_target(component, false)
-            .unwrap_or_else(|| Component::new(component.to_string(), target.clone(), true));
+        let new_component = Component::try_new(component, &distributable, target.as_ref())?;
         distributable.remove_component(new_component)?;
     }
 
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -18,8 +18,9 @@ use std::str::FromStr;
 
 use anyhow::{anyhow, bail, Context, Result};
 
-use crate::dist::dist::{PartialTargetTriple, Profile, TargetTriple};
+use crate::dist::dist::{Profile, TargetTriple};
 use crate::errors::*;
+use crate::toolchain::distributable::DistributableToolchain;
 use crate::utils::toml_utils::*;
 
 use super::{config::Config, dist::ToolchainDesc};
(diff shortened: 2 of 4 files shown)


rami3l (approved):
LGTM, and really sorry for the long wait! I personally would like to get this into Rustup v1.28.0.

src/cli/rustup_mode.rs
```diff
@@ -1379,14 +1378,46 @@ fn get_target(m: &ArgMatches, distributable: &DistributableToolchain<'_>) -> Opt
         .or_else(|| Some(distributable.desc().target.clone()))
 }
 
+fn get_component(
```
rami3l: I saw you replaced `Component::new_with_target` with this function. Could we make this a new `try_constructor` of the `Component` type? By `try_constructor` I'm referring to something like [`Box::try_new()`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.try_new), this way maybe it could be more IDE-friendly?

tests/suite/cli_rustup.rs
```diff
@@ -1366,6 +1366,27 @@ fn add_component() {
     });
 }
 
+#[test]
```
rami3l, line 1369: Do we have a test case testing the "sad path" of the story?

src/cli/rustup_mode.rs
ongchi: I think it's fair. This function has been moved to `Component::try_new` by [5c6801d](https://github.com/rust-lang/rustup/pull/3601/commits/5c6801dca2d1e268afdf33168370550e15d61089).

tests/suite/cli_rustup.rs
ongchi: More tests added by [d6ef648](https://github.com/rust-lang/rustup/pull/3601/commits/d6ef648f01549dd47dc11d2a93446cccd2b003c6).

rami3l (approved):
Nice work :)

src/cli/rustup_mode.rs
rami3l: > I think it's fair. This function has been moved to `Component::try_new` by [5c6801d](https://github.com/rust-lang/rustup/pull/3601/commits/5c6801dca2d1e268afdf33168370550e15d61089).

@ongchi Thanks! IIRC you added that function (?) and then moved it elsewhere in the same PR, so maybe it's reasonable to squash this commit into the first one?

ongchi: With no further modifications, I just squash commits and update commit messages.

src/dist/manifest.rs
```diff
@@ -588,21 +589,37 @@ impl Component {
...
-            let target = &pkg_with_target[pos + 1..];
-            if let Some(partial) = PartialTargetTriple::new(target) {
-                if let Ok(triple) = TargetTriple::try_from(partial) {
-                    return Some(Self {
-                        pkg: pkg.to_string(),
-                        target: Some(triple),
-                        is_extension,
-                    });
-                }
+    pub(crate) fn try_new(
```
djc, line 592: IMO neither the PR description nor the commit message make it clear what this change is trying to achieve. I could go dig through the issue thread, but that's pretty inefficient. Can you make sure it's a little easier to address what these changes are designed to do? I see a substantial increase in complexity here and I think such a change should be well-motivated.

rami3l (approved):
Thanks a lot! I think this PR is good to go.

rami3l:
@djc @rbtcollins I guess we can ship this with v1.27.1 as well? As I see it, only `rustup component (add|remove)` is affected here and it's rather unlikely for this PR to introduce a regression somewhere given all our existing tests combined with the new ones...

src/dist/manifest.rs
rami3l: @djc Are you suggesting that at the top of this PR (https://github.com/rust-lang/rustup/pull/3601#issue-2055114196) there should be a short briefing/rationale part? If so I agree with this point, but the code changes per se looks fine to me.

djc: > @djc Are you suggesting that at the top of this PR ([#3601 (comment)](https://github.com/rust-lang/rustup/pull/3601#issue-2055114196)) there should be a short briefing/rationale part? If so I agree with this point, but the code changes per se looks fine to me.

Yes, the PR description should make it more clear what issue/problem it is addressing.

rami3l: ~~@ongchi Could you please add a rationale section in https://github.com/rust-lang/rustup/pull/3601#issue-2055114196? Many thanks in advance!~~

@djc Nevermind, I found out that I can add the rationale myself.

djc (approved):

PR #3633, opened 2024-01-05
feat(ci): enable the `merge_group` trigger

rami3l:
Adding a second patch as requested in https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Enable.20merge.20queue.20for.20Rustup.

@jdno Does this look okay?


djc (approved):

jdno (approved):
Yes, I think we can give this a try! 🙂

rami3l:
> Yes, I think we can give this a try! 🙂

@jdno Please let us know when the merge queue is online, many thanks 🙏

Issue #3319, opened 2023-04-12
Give a warning before uninstalling the active/default toolchain, or the last/host target for a toolchain


jyn514:
### Problem you are trying to solve

Rustup currently lets you do some silly things:
```
; rustup target list --installed
x86_64-unknown-linux-gnu
; rustup target remove x86_64-unknown-linux-gnu
info: removing component 'rust-std'
; cargo build
   Compiling inner v0.1.0 (/home/jyn/src/example/inner)
error[E0463]: can't find crate for `std`
```
This isn't a bug, but it seems ... confusing.

### Solution you'd like

 I think rustup should warn in either of the following cases:
1. If you're trying to remove the host target, which means you can no longer build proc-macros or build scripts
2. If you're trying to remove the last remaining target installed
3. If you're uninstalling the active/default toolchain

```[tasklist]
### Tasks
- [ ] https://github.com/rust-lang/rustup/pull/3637
- [ ] https://github.com/rust-lang/rustup/pull/3924
```

rami3l:
I think I can try to tackle this after https://github.com/rust-lang/rustup/pull/3225, where the logic of detecting the active/default toolchain is refined.

PR #3637, opened 2024-01-15
feat(cli): warn when removing the last/host target for a toolchain

rami3l:
Addresses the concerns 1) and 2) in https://github.com/rust-lang/rustup/issues/3319.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ flate2 = "1"
 fs_at.workspace = true
 git-testament = "0.2"
 home = "0.5.4"
+itertools = "0.12"
 libc = "0.2"
 once_cell.workspace = true
 opener = "0.6.0"
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -10,6 +10,7 @@ use clap::{
     Arg, ArgAction, ArgGroup, ArgMatches, Command, ValueEnum,
 };
 use clap_complete::Shell;
+use itertools::Itertools;
 
 use crate::{
     cli::{
@@ -1334,11 +1335,23 @@ fn target_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let distributable = DistributableToolchain::try_from(&toolchain)?;
 
     for target in m.get_many::<String>("target").unwrap() {
-        let new_component = Component::new(
-            "rust-std".to_string(),
-            Some(TargetTriple::new(target)),
-            false,
-        );
+        let target = TargetTriple::new(target);
+        let default_target = cfg.get_default_host_triple()?;
+        if target == default_target {
+            warn!("after removing the default host target, proc-macros and build scripts might no longer build");
+        }
+        let has_at_most_one_target = {
+            let components = distributable.components()?;
+            // Every component target that is not `None` (wildcard).
+            let targets = components
+                .iter()
+                .filter_map(|c| c.installed.then(|| c.component.target.clone()).flatten());
+            targets.unique().at_most_one().is_ok()
+        };
+        if has_at_most_one_target {
+            warn!("after removing the last target, no build targets will be available");
+        }
+        let new_component = Component::new("rust-std".to_string(), Some(target), false);
         distributable.remove_component(new_component)?;
     }
 
--- a/src/toolchain/distributable.rs
+++ b/src/toolchain/distributable.rs
@@ -115,6 +115,13 @@ impl<'a> DistributableToolchain<'a> {
         Ok(())
     }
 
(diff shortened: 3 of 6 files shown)


src/toolchain/distributable.rs
```diff
@@ -115,6 +115,14 @@ impl<'a> DistributableToolchain<'a> {
         Ok(())
     }
 
+    pub(crate) fn components(&self) -> anyhow::Result<Vec<crate::dist::manifest::ComponentStatus>> {
+        let manifestation = self.get_manifestation()?;
+        let config = manifestation.read_config()?.unwrap_or_default();
+        let manifest = self.get_manifest()?;
+        let components = manifest.query_components(self.desc(), &config)?;
+        Ok(components)
```
djc: Style nit: I would avoid binding `components` here, in favor of just yielding `manifest.query_components(self.desc(), &config)`.

Cargo.toml
```diff
@@ -94,6 +94,7 @@ wait-timeout = "0.2"
 walkdir = { workspace = true, optional = true }
 xz2 = "0.1.3"
 zstd = "0.13"
+itertools = "0.12.0"
```
djc: Nit: please keep this in alphabetic order. In keeping with the style for other dependencies, prefer eliding trailing zero version components.

src/cli/rustup_mode.rs
```diff
@@ -1334,11 +1335,22 @@ fn target_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let distributable = DistributableToolchain::try_from(&toolchain)?;
 
     for target in m.get_many::<String>("target").unwrap() {
-        let new_component = Component::new(
-            "rust-std".to_string(),
-            Some(TargetTriple::new(target)),
-            false,
-        );
+        let target = TargetTriple::new(target);
+        let default_target = cfg.get_default_host_triple().ok();
```
djc: Why not return any error from `get_default_host_triple()` up the stack? Same for `distributable.components()`.

src/toolchain/distributable.rs
rami3l: Done.

Cargo.toml
rami3l: Done.

src/cli/rustup_mode.rs
rami3l: @djc Nice question! I'm a bit afraid of introducing unnecessary points of failure here since I'm just enhancing the error message, not performing some critical logic. However I can change that if you'd like.

rami3l: @djc Changed in 3810964e1483e75e35dd3d7731bfb22887709b43.

rami3l:
@djc I've changed the behavior to be early return on error according to your suggestion, and the tests currently show no issue with that. However, I'm still wondering whether just printing a warning would be a better option...

src/toolchain/distributable.rs
```diff
@@ -115,6 +115,13 @@ impl<'a> DistributableToolchain<'a> {
         Ok(())
     }
 
+    pub(crate) fn components(&self) -> anyhow::Result<Vec<crate::dist::manifest::ComponentStatus>> {
```
djc, line 118: Style nit: let's import `ComponentStatus` instead of using a fully-qualified path for it?

src/cli/rustup_mode.rs
```diff
@@ -1334,11 +1335,23 @@ fn target_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
+            warn!("after removing the default host target, proc-macros and build scripts might no longer build");
+        }
+        let has_at_most_one_target = {
+            let components = distributable.components()?;
+            // Every component target that is not `None` (wildcard).
+            let targets = components
+                .iter()
+                .filter_map(|c| c.installed.then(|| c.component.target.clone()).flatten());
+            targets.unique().at_most_one().is_ok()
+        };
```
djc, line 1350: What do you think about this alternative?

```rust
let has_at_most_one_target = distributable
            .components()?
            .into_iter()
            .filter_map(|c| match (c.installed, c.component.target) {
                (true, Some(t)) => Some(t),
                _ => None,
            })
            .unique()
            .at_most_one()
            .is_ok();
```

This is a bit longer (in formatted line count), but it avoids the nested combinators and the need for cloning the target triples.

djc (approved):

rami3l:
@djc Thanks for your suggestions! Sorry for having enabled merge on approval... I guess I'll do that in a separate PR then.

src/cli/rustup_mode.rs
rami3l: cf. #3643.

src/toolchain/distributable.rs
rami3l: cf. #3643.

PR #3643, opened 2024-01-18
feat(cli): warn when removing the last/host target for a toolchain, pt. 2

rami3l:
Addresses the concerns in https://github.com/rust-lang/rustup/pull/3637#discussion_r1457125300 and https://github.com/rust-lang/rustup/pull/3637#discussion_r1457147688.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1340,14 +1340,17 @@ fn target_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
         if target == default_target {
             warn!("after removing the default host target, proc-macros and build scripts might no longer build");
         }
-        let has_at_most_one_target = {
-            let components = distributable.components()?;
-            // Every component target that is not `None` (wildcard).
-            let targets = components
-                .iter()
-                .filter_map(|c| c.installed.then(|| c.component.target.clone()).flatten());
-            targets.unique().at_most_one().is_ok()
-        };
+        // Whether we have at most 1 component target that is not `None` (wildcard).
+        let has_at_most_one_target = distributable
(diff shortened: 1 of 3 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1340,14 +1340,17 @@ fn target_remove(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
-            targets.unique().at_most_one().is_ok()
-        };
+        // Whether we have at most 1 component target that is not `None` (wildcard).
+        let has_at_most_one_target = distributable
+            .components()?
+            .into_iter()
+            .filter_map(|c| match (c.installed, c.component.target) {
+                (true, Some(t)) => Some(t),
+                _ => None,
+            })
```
rami3l, line 1350: Maybe this would be even better?

```suggestion
            .filter(|c| c.installed)
            .filter_map(|c| c.component.target)
```

djc: Personally I prefer my variant: it seems confusing to me (and potentially less efficient) to run two closures in subsequent combinators, when these could just be combined into a single function.

djc (approved):

src/cli/rustup_mode.rs
rami3l: @djc Thanks for your opinion! I'll stick with your version then.

rami3l:
@djc Looks like we're hitting https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/rustup.20CI.20fails.20to.20install.20Docker.20images.20for.20armv7-linux again. Did we do anything in particular to workaround/solve this issue when it last happened?

PR #3664, opened 2024-02-08
Fix minor typo

ghost:


djc (commented):
"won't be doing" seems more currect to me than "aren't be doing"?

xen0n (commented):
Here "[you] won't be doing ..." means "[you] are not going to do ... in the short term / in the foreseeable future", while "[you] aren't doing ..." sounds more like "[you] will not do ... right now / at all". So I personally feel the original text sounds more natural.

Mandatory disclaimer though: I'm not a native speaker of English.

ghost:
> "won't be doing" seems more currect to me than "aren't be doing"?

Yes, that’s what I meant. I guess that’s what I get for making a PR at 2AM xD

ghost:
Okay this is strange. I created a PR to fix this a couple months ago and it was merged. It seems like the docs themselves haven’t updated with that PR yet, so ignore this one.

PR #3666, opened 2024-02-10
Update mdbook and update some source issues.

ehuss:
This updates `mdbook` for the docs deploy, and updates the source markdown to not include unicode that may not be obvious to someone editing it.

* Updates mdbook to 0.4.37. There have been quite a few small changes since the last update: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-0437
* Removes the invisible U+00A0 no-break-space spaces. Those weren't intentional, it was just an artifact of how I used my keyboard.
* Changes the em-dash to `---` so that it is clearer to people editing the doc.
* Adds `curly-quotes=true` which enables [Smart Punctuation](https://rust-lang.github.io/mdBook/format/markdown.html#smart-punctuation) which is needed for the em-dash translation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1190,7 +1190,7 @@ jobs:
       - name: Install mdbook
         run: |
           mkdir mdbook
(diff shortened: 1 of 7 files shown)


rami3l (approved):
Many thanks @ehuss!

PR #3653, opened 2024-01-29
dist: bump `rustup` version to `1.27.0`

rami3l:
cc #3501.

## Concerns

- [x] Should we make a beta-driving phase for this?

---

# Changelog

## [1.27.0] - 2024-03-08

This long-awaited Rustup release has gathered all the new features and fixes since April 2023.
These changes include improvements in Rustup's maintainability, user experience, compatibility and documentation quality.

The headlines of this release are:
- Basic support for `fish` shell has been added.
- Support for the `loongarch64-unknown-linux-gnu` host platform has been added.

Also, it's worth mentioning that Dirkjan Ochtman and rami3l have joined the team and are coordinating this new release.

Finally, the project seems to have attracted a total of 23 new contributors within this release cycle. Looking forward to seeing you again in the future!

### Added

- Add basic support for `fish` shell [pr#3108] 
- Add the `RUSTUP_TERM_COLOR` environment variable to force the use of colored output [pr#3435]
- Improve `rustup-init.sh`'s compatibility with `ksh` and `zsh` [pr#3475]
- Add a warning when running under Rosetta 2 [pr#3068]
- Add browser detection for RISC-V 64 platform [pr#3642]
- Add a warning when removing the last/host target for a toolchain [pr#3637]

### Changed

- Upgrade `clap` to v4 [pr#3444]
- Fix incorrect platform detection on macOS aarch64 due to Rosetta 2 [pr#3438]
- Fix incorrect platform detection on 32-bit Linux userland with a 64-bit kernel [pr#3488] [pr#3490]
- Improve Windows system32 DLL loading mechanism [pr#3493]
- Improve suggestions about missing components [pr#3453]
- Fix handling of toolchain names with special characters [pr#3518]
- Fix panic in `component list --toolchain stable` [pr#3548]
- Rename `llvm-tools-preview` component to `llvm-tools` [pr#3578]
- Try using symlinks if possible on Windows [pr#3687]
- Bump a lot of dependencies to their latest versions [pr#renovate-bot]

Thanks go to:

- Anthony Perkins (acperkins)
- Tianqi (airstone42)
- Alex Gaynor (alex)
- Alex Hudspith (alexhudspith)
- Alan Somers (asomers)
- Brett (brettearle)
- Burak Emir (burakemir)
- Chris Denton (ChrisDenton)
- cui fliter (cuishuang)
- Dirkjan Ochtman (djc)
- Dezhi Wu (dzvon)
- Eric Swanson (ericswanson-dfinity)
[...]

PR #3668, opened 2024-02-11
fix(ci): fix CI failures on the `stable` branch

rami3l:
Addresses https://github.com/rust-lang/rustup/pull/3653#issuecomment-1936981569.

## Progress 

- [x] `i686-unknown-linux-gnu`
- [x] `mips*-unknown-linux-gnu*` *4
- [x] ~~`i686-linux-android`: Upstream issue (https://github.com/sfackler/rust-openssl/issues/2163, https://github.com/openssl/openssl/issues/18059).~~ Worked around by downgrading `openssl-sys` to 0.9.92.
  - Probably also related to https://github.com/openssl/openssl/issues/14083#issuecomment-812980399, https://github.com/microsoft/vcpkg/pull/24801 and https://github.com/llvm/llvm-project/issues/73361.
- [x] `riscv64gc-unknown-linux-gnu`: Half-unsolved upstream issue. ~~Probably requires an additional fix as described in https://github.com/openssl/openssl/issues/22871#issuecomment-1866621698.~~ Seemingly worked around in https://github.com/alexcrichton/openssl-src-rs/pull/230.
- [x] Autosquash-and-rebase the PR, dropping the CI target changes.


Cargo.lock
```diff
@@ -1337,18 +1337,18 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "openssl-src"
-version = "300.2.3+3.2.1"
+version = "111.28.1+1.1.1w"
```
djc, line 1340: If we're going to do this, should get rid of the perl IPC-CMD thing for now?

djc (approved):
Great work cleaning this up!

Cargo.lock
rami3l: I didn't notice the downgrade `openssl-src` until now... I'd say we are at a very weird state currently. This fix is required for OpenSSL v3 which we could update to in the future, but it's just half-done: it's fixed with x64, but not fixed with x86 until now. I'd say within this downgrade commit I'll remove the perl IPC-CMD thing for both, but at the same time let's keep the previous commit that fixes the build for v3 on x86.

ci/docker/i686-unknown-linux-gnu/Dockerfile
```diff
@@ -1,6 +1 @@
 FROM rust-i686-unknown-linux-gnu
-
-# Install `perl-IPC-Cmd` to make OpenSSL v3 happy.
```
djc: Instead of making this change in this commit, I think it would be better to remove the earlier commit from this PR.

rami3l: Done!

djc (approved):

PR #3673, opened 2024-02-17
Fix ambiguous "Proceed with installation" prompt

### LINKED ISSUE #3429: 'Proceed with installation (default)' prompt is ambiguous
### Problem

After customizing the installation, rustup shows "Proceed with installation (default)" instead of "Proceed with installation (custom)" suggesting it is going to ignore what the user has done and run a default installation instead.

### Steps

1. run rustup
2. when presented with the three options, pick `2) Customize installation`
3. customize the installation
4. observe the label for `1)` after customizing and being back in the menu

### Possible Solution(s)

Change the phrasing to `1) Proceed with installation (custom)` after the user has customized the installation.

### Notes

This is a text bug, not a code bug.

### Rustup version

```console
Windows 10 x64:


>rustup-init --version
```
```


### Installed toolchains

```console
none, rustup hasn't installed anything yet.
```

brettearle:
Changed test and prompt of default prompt in stdout.
Fix #3429

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -56,12 +56,13 @@ pub(crate) enum Confirm {
     Advanced,
 }
 
-pub(crate) fn confirm_advanced() -> Result<Confirm> {
+pub(crate) fn confirm_advanced(customized_install: bool) -> Result<Confirm> {
     writeln!(process().stdout().lock())?;
-    writeln!(
-        process().stdout().lock(),
-        "1) Proceed with installation (default)"
-    )?;
+    let first_option = match customized_install {
+        true => "1) Proceed with selected options (default - just press enter)",
+        false => "1) Proceed with standard installation (default - just press enter)",
+    };
+    writeln!(process().stdout().lock(), "{first_option}")?;
     writeln!(process().stdout().lock(), "2) Customize installation")?;
     writeln!(process().stdout().lock(), "3) Cancel installation")?;
     write!(process().stdout().lock(), ">")?;
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -433,10 +433,10 @@ pub(crate) fn install(
         let msg = pre_install_msg(opts.no_modify_path)?;
 
         md(&mut term, msg);
-
+        let mut customized_install = false;
         loop {
             md(&mut term, current_install_opts(&opts));
-            match common::confirm_advanced()? {
+            match common::confirm_advanced(customized_install)? {
                 Confirm::No => {
                     info!("aborting installation");
                     return Ok(utils::ExitCode(0));
@@ -445,6 +445,7 @@ pub(crate) fn install(
                     break;
                 }
                 Confirm::Advanced => {
+                    customized_install = true;
                     opts = customize_install(opts)?;
                 }
             }
--- a/tests/suite/cli_inst_interactive.rs
+++ b/tests/suite/cli_inst_interactive.rs
@@ -90,7 +90,7 @@ Current installation options:
                profile: default
   modify PATH variable: no
 
-1) Proceed with installation (default)
+1) Proceed with standard installation (default - just press enter)
 2) Customize installation
 3) Cancel installation
 >


brettearle:
Still a WIP, need to refactor a little

rami3l:
@brettearle I'm converting this to a draft since you've declared it to be a WIP. Please feel free to convert it back when you're ready :)

brettearle:
Alright all done @rami3l, unless you need me to change something :)

src/cli/self_update.rs
```diff
@@ -433,10 +433,10 @@ pub(crate) fn install(
         let msg = pre_install_msg(opts.no_modify_path)?;
 
         md(&mut term, msg);
-
+        let mut customised_install: bool = false;
```
djc: Please omit the type annotation, which is unnecessary, and rename `customised_install` to `customized_install` to match convention for US English spelling.

src/cli/common.rs
```diff
@@ -56,12 +56,18 @@ pub(crate) enum Confirm {
...
 }
 
-pub(crate) fn confirm_advanced() -> Result<Confirm> {
+pub(crate) fn confirm_advanced(customised_install: &bool) -> Result<Confirm> {
     writeln!(process().stdout().lock())?;
-    writeln!(
-        process().stdout().lock(),
-        "1) Proceed with installation (default - just press enter)"
-    )?;
+    let one_prompt = match customised_install {
```
djc: I'd rather avoid the extra line here.

src/cli/common.rs
```diff
@@ -56,12 +56,18 @@ pub(crate) enum Confirm {
...
+        true => "Proceed with custom install. Press Enter to run custom install",
+        false => "Proceed with standard install. Press Enter to run (1)",
+    };
+    writeln!(process().stdout().lock(), "{one_prompt}")?;
+    let first_option = match customised_install {
+        true => "1) Proceed with customised install",
```
djc: I think we should keep the `(default - just press enter)` and just change the wording before that in the first option. I'd suggest "Proceed with selected options (default - just press enter)" in the customized case.

src/cli/self_update.rs
brettearle: this has become a enum now. this seems to be outdated

brettearle: will change the spelling though

src/cli/common.rs
brettearle: as in 60?

src/cli/common.rs
brettearle: too easy

brettearle:
Ready for review again, happy to change if required :)

src/cli/self_update.rs
```diff
@@ -143,6 +143,11 @@ impl ToString for SelfUpdateMode {
     }
 }
 
+pub enum CustomizedInstall {
```
djc: IMO `Yes`/`No` enums like this are an anti-pattern. Instead, it should be `enum Install { Custom, Default }`. But really not sure we need an enum for this case over a `bool`.

brettearle: good point. Changed back to bool

brettearle: was doing mainly for readability but can see how its bad

src/cli/common.rs
```diff
@@ -12,7 +12,7 @@ use anyhow::{anyhow, Context, Result};
 use git_testament::{git_testament, render_testament};
 use once_cell::sync::Lazy;
 
-use super::self_update::{self, CustomizedInstall};
+use super::self_update::{self};
```
djc: Should remove `::{self}`, this is unnecessary.

src/cli/common.rs
```diff
@@ -56,13 +56,11 @@ pub(crate) enum Confirm {
     Advanced,
 }
 
-pub(crate) fn confirm_advanced(customized_install: &CustomizedInstall) -> Result<Confirm> {
+pub(crate) fn confirm_advanced(customized_install: &bool) -> Result<Confirm> {
```
djc: There's no point in passing `bool` by reference, just pass it directly here.

src/cli/common.rs
brettearle: done :)

src/cli/common.rs
brettearle: Is that best practice in rust in general? pass reference only when necessary I mean

also done :)

djc (approved):

src/cli/common.rs
djc: Yes -- in particular, `Copy` types (including `bool`, integer and floating point types) are typically passed by-value.

djc:
Please squash these changes into a single commit (I usually do this with `git rebase -i`).

src/cli/common.rs
brettearle: Sweet as, going in the anki

brettearle:
Sweet as, will do.

brettearle:
Gday, Is that rebase done the right way? Just unsure about using force to over write it.

rami3l (approved):
LGTM, thanks a lot!

djc:
Yes, the rebase is great -- thanks!

PR #3674, opened 2024-02-17
added feedback to doc flag and doc in terminal when opening browser

### LINKED ISSUE #3663: Say what ought to happen when running rustup doc.

On macOS.

I ran \"rustup doc --book\" on my command line. It returned, and nothing appeared to happen. Something did happen, the site is open on my browser, but I use desktops, and it doesn't raise things quite correctly (not rust's problem).

```
    learn Rust.
✔ ~/src/CatHerd [master|✚ 1…2]
14:01 $ rustup doc --book
✔ ~/src/CatHerd [master|✚ 1…2]
```

### Solution you'd like

if the thing had printed "Opening the Rust Book in your browser", I would have slid over there, and all would be good.

### Notes

_No response_

brettearle:
Fix #3663

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1583,9 +1583,14 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
         topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
         topical_path.to_str().unwrap()
-    } else if let Some((_, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+    } else if let Some((name, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+        writeln!(
+            process().stderr().lock(),
+            "Opening docs named `{name}` in your browser"
+        )?;
         path
     } else {
+        writeln!(process().stderr().lock(), "Opening docs in your browser")?;
         "index.html"
     };
 


MrTheSaw (approved):
That's the sort of thing I was thinking. Looks good.

rami3l:
@brettearle I think with Rustup you can forget about merging the master occasionally: GitHub Merge Queue will do this for you eventually, and doing so now is almost certainly a no-op.

brettearle:
Too easy, I will not merge in future unless git yells on the push 🫸

src/cli/rustup_mode.rs
```diff
@@ -1598,9 +1598,15 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
         topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
         topical_path.to_str().unwrap()
-    } else if let Some((_, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+    } else if let Some((name, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+        writeln!(
+            process().stdout().lock(),
+            "Opening docs -{} in your browser",
```
djc: The `-` prefix looks strange to me, what's your motivation for adding this?

brettearle: Used it to represent name of flag

djc: Sure, but why use a `-` prefix for it?

brettearle: Not much thought went into it. Happy to change

brettearle: Was trying to stay with what the user put in command line

brettearle:
Removed the prefix

djc (approved):
Now that I think about it, this should probably go to stderr instead of stdout. Otherwise this looks okay to me.

brettearle:
@djc why output it as an error? happy to change just would like to know reasoning as I must be missing some knowledge

djc:
stderr is generally what is used for status updates/log-like output to the user; stdout should be used for showing the requested output.

src/cli/rustup_mode.rs
```diff
@@ -1598,9 +1598,15 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
         topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
         topical_path.to_str().unwrap()
-    } else if let Some((_, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+    } else if let Some((name, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
+        writeln!(
+            process().stderr().lock(),
+            "Opening docs {} in your browser",
+            name
```
rami3l: I suggest refining the wording a bit more:

```suggestion
            "Opening docs named `{name}` in your browser",
```

... since the current one might still seem a bit weird when the `name` is e.g. "rustdoc" or "unstable-book".

djc:
(Please also squash the commits here into a single one.)

rami3l (changes requested):
https://github.com/rust-lang/rustup/pull/3674#issuecomment-1978520247 still needs to be addressed. The current code still prints to stdout.

brettearle:
Yeah not sure what happened there. Will sort.

djc (approved):

rami3l (approved):

PR #3681, opened 2024-02-23
ci: update runners for macOS-related workflows

rami3l:
The current `macos-latest` workflow points to `macos-12`, but it will [point to `macos-14` (which is arm64-based) very soon](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available):

> The macOS 12 runner image will remain latest until migration of the latest YAML workflow label to macOS 14 in Q2 FY24 (April – June 2024).

So it's necessary for us to migrate from `macos-12` to `macos-14` and/or `macos-13` (which is x64-based), or we might face a failure sometime in 24Q2 due to inconsistent target triples.

In this PR, `macos-13` is used for x64 builds, and `macos-14` for arm64 builds (in a future update, this should be replaced with `macos-latest`). Also, the full test suite for the latter (`aarch64-apple-darwin`) has been enabled to get it ready for tier 1 promotion.

cc https://github.com/rust-lang/cargo/pull/13377

PS: As usual, please ignore `.github/workflows/ci.yaml` when reviewing this PR.


djc (approved):
Nice!

PR #3687, opened 2024-02-27
Windows: Try using symlinks if they're allowed

### LINKED ISSUE #3677: Prefer using symlinks on Windows

On Windows rustup currently uses junction points exclusively. https://github.com/rust-lang/rustup/blob/d55d41c8f23a6aac4db8b9d5f9ba257c341e5d08/src/utils/raw.rs#L136-L141

(note: that comment is wrong, user symlinks can be enabled by granting a single permission or more recently by enabling developer mode)

Junction points are always absolute which makes moving rustup (e.g. when setting up a [dev drive](https://learn.microsoft.com/en-us/windows/dev-drive/)) a more fraught process than it needs to be.

### Solution you'd like

Use [symlink_dir](https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_dir.html) and only fallback to juncture points if that fails.

### Notes

_No response_

ChrisDenton:
On Windows, creating symlinks is not allowed unless developer mode is enabled or a specific permission is granted. Therefore this optimistically attempts to use symlinks first, before falling back to juncture points in the event that fails.

Resolves #3677

### DIFF
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -136,6 +136,8 @@ pub fn symlink_dir(src: &Path, dest: &Path) -> io::Result<()> {
     fn symlink_dir_inner(src: &Path, dest: &Path) -> io::Result<()> {
         // On Windows creating symlinks isn't allowed by default so if it fails
         // we fallback to creating a directory junction.
+        // We prefer to use symlinks here because junction point paths, unlike symlinks,
+        // must always be absolute. This makes moving the rustup directory difficult.
         std::os::windows::fs::symlink_dir(src, dest).or_else(|_| symlink_junction_inner(src, dest))
     }
     #[cfg(not(windows))]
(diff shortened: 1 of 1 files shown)


djc (approved):
Seems nice! Maybe add a comment to explain why we think symlinks are the better option here?

ChrisDenton:
Something like that?

djc (approved):

djc:
This seems okay, going to merge this. Thanks!

rami3l:
@ChrisDenton I don't use Windows personally but I have a question: will this create a possibly inconsistent state where one part of the installation is organized with symlinks and another part with juncture points?

ChrisDenton:
That could happen but I can't think of a case where the type of link would matter to us. Except, as previously mentioned, when moving the rustup directory. In that case you'll end up with some links being broken while others aren't but I'm not sure that's any worse than all links being broken.

PR #3696, opened 2024-03-07
www: Fix copy icon position in Safari

### LINKED ISSUE #3512: Clipboard copy icon misaligned on Safari
### Problem

On Safari (macOS and iOS), the clipboard copy icon is misaligned. It clips halfway through the top of the button instead of being in the centre. I'm using macOS 14.0 and iOS 17.0.3.

image

### Steps

1. View rustup.rs using Safari

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
N/A
```

### Installed toolchains

```console
N/A
```

noirgif:
This patch fixes the copy-icon position in Safari Version 17.3.1 (19617.2.4.11.12). It appears out of the box because "top: 50%" has no effect on Safari.

The 27px is from 50% of `height(58px) - border-top-width(2px) - border-bottom-width(2px)`.

Before:
image

After:
image

It does not change the position of the icon on my Chrome 122.0.6261.94.

Fixes #3512.

### DIFF
--- a/www/rustup.css
+++ b/www/rustup.css
@@ -165,7 +165,7 @@ hr {
     position: relative;
     width: fit-content;
     height: fit-content;
-    top: 50%;
+    top: 27px;
     left: 50%;
     transform: translate(-50%, -50%);
 }


djc (approved):
Thanks! This also looks good to me in Firefox (on macOS).

www/rustup.css
```diff
@@ -165,7 +165,7 @@ hr {
     position: relative;
     width: fit-content;
     height: fit-content;
-    top: 50%;
+    top: 27px;
```
rami3l, line 168: I'm not very familiar with CSS, so please correct me if I'm wrong, but I believe `top: 50%` with `-webkit-top: 27px` would be a better option here?

noirgif: image
Good idea, ~~but it seems that `top` takes precedence~~.

rami3l: @noirgif Just did some extra research and it turns out the ordering is important (`top` should come after `-webkit-top`): https://stackoverflow.com/a/7080674

noirgif: There doesn't seem to be a CSS extension called `-webkit-top`: https://developer.mozilla.org/en-US/docs/Web/CSS/Webkit_Extensions

rami3l: That's fair, sorry for the noise 🙇

PR #3697, opened 2024-03-08
Fix #3112 serve docs on localhost:{random port}

### LINKED ISSUE #3112: file:// urls for doc incompatible with snap/flatpak/container-based-development

There are various issues around confinement and working with containers on Linux where the path of the html docs either ends up wrong or the browser does not have permission to open the html docs.

- a flatpak or snap of firefox cannot open files in /home (it would need to be granted permission)
- if i'm developing inside a container (eg a vscode remote container) the paths rustup produces are incorrect outside the container

### Solution you'd like

Provide an option to serve the docs as a port like `mdbook serve` ( https://rust-lang.github.io/mdBook/cli/serve.html ), maybe like `rustup doc --serve`

This allows for both confined applications and developing inside a container to then work as they don't need to access the html files via a path, but instead just open `localhost:8080` etc.

### Notes

This would resolve issues like #3110 #2997 #2988 as there is no issue with confinement on ports.

If this is something this is possible, it would be good for `cargo doc --serve` to also exist as the same problem occurs with `cargo doc --open`. And this would then make both rustup and cargo consistent with `mdbook serve` which has this ability.

brettearle:
A starting implementation looking for feedback and whether I am on the right path


Cargo.toml
```diff
@@ -12,13 +12,13 @@ build = "build.rs"
 
 [features]
 curl-backend = ["download/curl-backend"]
+
```
djc, line 15: Please avoid making unrelated changes.

Cargo.toml
```diff
@@ -86,6 +86,7 @@ termcolor.workspace = true
 thiserror.workspace = true
 threadpool = "1"
 tokio = { workspace = true, optional = true }
+tiny_http = "0.11"
```
djc, line 93: Given that we're already depending on hyper for the reqwest download backend, I think we should be using hyper (0.14) for the server, too.

src/cli/rustup_mode.rs
```diff
@@ -725,6 +726,9 @@ pub(crate) fn cli() -> Command {
                         .value_parser(partial_toolchain_desc_parser),
                 )
                 .arg(Arg::new("topic").help(TOPIC_ARG_HELP))
+                .subcommand(
+                    Command::new("servedoc")
```
djc, line 730: I think this should probably be `doc --serve` rather than a new command.

src/cli/rustup_mode.rs
```diff
@@ -1608,6 +1612,44 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
+                        let file = std::fs::File::open(path).unwrap();
+                        request.respond(Response::from_file(file)).unwrap();
+                    }
+                    None => {
+                        let file = std::fs::File::open(&doc_path_index).unwrap();
+                        request.respond(Response::from_file(file)).unwrap();
+                    }
+                }
+            }
+        }
```
djc, line 1657: This should be outlined into a separate function.

Cargo.toml
brettearle: Sorry. Will avoid

djc:
(Also please make sure to run `cargo fmt` and committing the result before pushing.)

Cargo.toml
brettearle: Too easy, will do

src/cli/rustup_mode.rs
brettearle: Alright, so this would be instead of a specific doc flag and just do base docs or in addition to specific flag "doc --nomicon --serve"?

src/cli/rustup_mode.rs
brettearle: Too easy, will do

src/cli/rustup_mode.rs
djc: It should be a new `--serve` flag on the existing `doc` command, that works orthogonally to other flags. Maybe it should take an optional port number or `SocketAddr` (like `192.168.0.1:3112`) as argument?

PR #3702, opened 2024-03-11
Fix dead_code and unused_imports warnings

ehuss:
This fixes some warnings for dead_code and unused_imports, where I believe rustc has recently started emitting in more recent versions.


djc (approved):
Thanks!

Issue #3256, opened 2023-03-08
No way to prevent the Rustup installer from creating unwanted folders & files in the home dir


lgarron:
### Problem

I prefer to use Homebrew to handle all my toolchains, but I've heard a lot of good things about Rustup and gave it chance last week. However, I was unpleasantly surprised that the Rustup install process creates *a whole bunch* of files and folders directly in my home directory. There is an issue about following XDG conventions for config (https://github.com/rust-lang/rustup/issues/247), but this seems to be a broader issue. So far I've seen the following created, that I'm 100% certain all did not exist before:

- `~/.cargo`, which requires `$CARGO_HOME` to avoid.
- `~/.rustup`, which requires `$RUSTUP_HOME` to avoid.
- `~/.profile`, which I do not use and do not want on my system.
- `~/.zshenv`, which I *also* do not use and do not want on my system.

And just now, I found that using those env vars and deleting those dirs and files (and *then* reinstalling from scratch) wasn't enough — there's still a `~/.rustup/settings.toml` file. Is it safe to delete? I specifically opted out of that folder, and now I have to research if it's safe to delete again.

To me, this is a horrible first impression. It suggests that **Rustup is extremely callous about externalities and resource usage**, which makes me wary of anything else it might do in the future. I've seen Python and Ruby installers break things often enough, and I'm really hoping Rustup won't be like them.

It's like I invited a contractor into my house, only to find they've installed something in the garage instead of the dedicated shed — and then installed a bunch of extra stuff around the house "just in case I need it" even though I very much *don't* need or want it. I would really appreciate if Rustup didn't do so, and I'd be happy to contribute fixes if I know they'd be accepted.

### Steps

Run the installation command from https://rustup.rs/ on macOS, optionally with `$CARGO_HOME` and `$RUSTUP_HOME` set.

### Possible Solution(s)

In order of respectfulness:

1. Ideally, do not create **any** home directory folders as part of the install process, even by/for other tools. If you want to create them, make it clear that you will and give a prompt to opt out / abort the install process.
2. Alternatively, follow XDG/platform conventions per https://github.com/rust-lang/rustup/issues/247 by default, without requiring any env vars.
3. Use `$XDG_DATA_HOME` if it's set.
4. Don't create shell profile files if they don't already exist, particularly if I'm not using the shell in question (also see https://github.com/rust-lang/rustup/issues/3251, https://github.com/rust-lang/rustup/issues/478).

### Rustup version

```console
rustup 1.25.2
[...]

PR #3712, opened 2024-03-13
fix(shell): create parent dir before appending to rcfiles

### LINKED ISSUE #3706: Rustup-init fails when `fish` is installed but without config files generated
### Problem

I launched a GitHub Codespace, ran `curl https://sh.rustup.rs -sSf | sh`, got `error: could not amend shell profile: '/home/codespace/.config/fish/conf.d/rustup.fish': could not write rcfile file: '/home/codespace/.config/fish/conf.d/rustup.fish': No such file or directory (os error 2)`

Even though `fish` is installed in the codespace, its config is not pre-generated and the default shell is `bash`.
As I understood after 5 minutes of scrolling the code, `rustup` [only checks if the shell's binary exists](https://github.com/rust-lang/rustup/blob/4d3281f3bdf8eac76c5998e664ff5db9d7fbf6e0/src/cli/self_update/shell.rs#L83C18-L83C18), not if the config files exist
This doesn't happen with other shells since their config files are typically located in the home folder

### Steps

1. Ensure `fish` is on PATH
2. Ensure fish's config directory doesn't exist
3. Run `rustup` with default settings (with `modify PATH variable`)

### Possible Solution(s)

_No response_

### Notes

Looks like conf.d directory is not created automatically even after running `fish`

### Rustup version

Rustup did not install, but
```console
@4310V343k ➜ /tmp/tmp.MNjnDjzYbe $ ./rustup-init -V
rustup-init 1.27.0 (bbb9276d2 2024-03-08)
```

### Installed toolchains

Rustup did not install

rami3l:
Fixes #3706, ~~addressing https://rust-lang.github.io/rust-clippy/master/#/ineffective_open_options at the same time~~ (superseded by https://github.com/rust-lang/rustup/pull/3713).

## Concerns

- [x] ~~It's better to test the CI build in a container or something before actually merging this PR.~~ (Tested on Ubuntu 22.04.4 LTS x86_64.)

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -233,7 +233,11 @@ impl UnixShell for Fish {
     }
 
     fn update_rcs(&self) -> Vec<PathBuf> {
-        self.rcfiles()
+        // The first rcfile takes precedence.
+        match self.rcfiles().into_iter().next() {
+            Some(path) => vec![path],
+            None => vec![],
+        }
     }
 
     fn env_script(&self) -> ShellScript {
--- a/src/cli/self_update/unix.rs
+++ b/src/cli/self_update/unix.rs
@@ -92,6 +92,13 @@ pub(crate) fn do_add_to_path() -> Result<()> {
                 _ => &source_cmd,
             };
 
+            let rc_dir = rc.parent().with_context(|| {
+                format!(
+                    "parent directory doesn't exist for rcfile path: `{}`",
+                    rc.display()
+                )
+            })?;
+            utils::ensure_dir_exists("rcfile dir", rc_dir, &|_: Notification<'_>| ())?;
             utils::append_file("rcfile", &rc, cmd_to_write)
                 .with_context(|| format!("could not amend shell profile: '{}'", rc.display()))?;
         }


src/cli/self_update/shell.rs
```diff
@@ -233,7 +233,8 @@ impl UnixShell for Fish {
     }
 
     fn update_rcs(&self) -> Vec<PathBuf> {
-        self.rcfiles()
+        // The first rcfile takes precedence.
```
djc, line 236: This doesn't obviously make sense to me. Why is this necessary? At the very least deserves some more comments on why we're doing this.

src/cli/self_update/unix.rs
```diff
@@ -92,6 +92,13 @@ pub(crate) fn do_add_to_path() -> Result<()> {
                 _ => &source_cmd,
             };
 
+            let rc_dir = rc.parent().with_context(|| {
+                format!(
+                    "could not find parent directory for rcfile path: `{}`",
```
djc: Nit: maybe "could not find parent directory" -> "no parent directory"? This isn't actually about "finding" the directory (in the filesystem), but about an rcfile path having no directories at all.

src/cli/self_update/shell.rs
rami3l: @djc Sure! The snippet below explains the differences between `.rcfiles()` and `.update_rcs()`.

https://github.com/rust-lang/rustup/blob/3fcd56283311ab1cc24ab275bc7e44b9d7f62a8d/src/cli/self_update/shell.rs#L91-L96

The problem with the current code is that, we have two rc paths in fish, one (`p1`) being the **_fallback_** of the other (`p0`).

https://github.com/rust-lang/rustup/blob/3fcd56283311ab1cc24ab275bc7e44b9d7f62a8d/src/cli/self_update/shell.rs#L218-L230

The issue here is, according to the current implementation, if both paths are valid, then we'll create and append to **_both_** of them. (It's okay to check both paths for cleanup, but that's another story.)

https://github.com/rust-lang/rustup/blob/f6dbe2e4cc186c7624e76691ece2ff37c87cf565/src/cli/self_update/unix.rs#L88-L93

This is the expected behavior in some cases, but clearly not for fish due to this fallback relationship. I've found some similar code in the Zsh support (note the `.take(1)`):

https://github.com/rust-lang/rustup/blob/f6dbe2e4cc186c7624e76691ece2ff37c87cf565/src/cli/self_update/shell.rs#L186-L191

https://github.com/rust-lang/rustup/blob/f6dbe2e4cc186c7624e76691ece2ff37c87cf565/src/cli/self_update/shell.rs#L193-L205

src/cli/self_update/unix.rs
rami3l: @djc Done!

src/cli/self_update/shell.rs
djc: So should we change `update_rcs()` to yield `PathBuf` instead of `Vec<PathBuf>`? It seems conceptually confusing to do this only for some shells -- do we update all rcs for `Bash`?

Anyway, I guess this change is fine, but let's add some comments to explain the logic.

This way of writing it avoids the `cloned()`, and thus seems cleaner to me:

```rust
match self.rcfiles().into_iter().next() {
    Some(path) => vec![path],
    None => vec![],
}
```

What do you think?

rami3l: > So should we change `update_rcs()` to yield `PathBuf` instead of `Vec<PathBuf>`? It seems conceptually confusing to do this only for some shells -- do we update all rcs for `Bash`?

@djc I'm not a bash expert, but at least the implementation is a bit more complex with Bash indeed, which seems to consider what exists on the current file system as a file and what doesn't.

https://github.com/rust-lang/rustup/blob/f6dbe2e4cc186c7624e76691ece2ff37c87cf565/src/cli/self_update/shell.rs#L138-L152

... a quick blame points to https://github.com/rust-lang/rustup/pull/2387. @workingjubilee maybe you can provide more context on this one?

rami3l: > This way of writing it avoids the `cloned()`, and thus seems cleaner to me:
> 
> ```rust
> match self.rcfiles().into_iter().next() {
>     Some(path) => vec![path],
>     None => vec![],
> }
> ```
> 
> What do you think?

@djc As you wish.

rami3l: > This way of writing it avoids the `cloned()`, and thus seems cleaner to me:
> 
> ```rust
> match self.rcfiles().into_iter().next() {
>     Some(path) => vec![path],
>     None => vec![],
> }
> ```
> 
> What do you think?

@djc Done.

djc: Would still like some more comments about why we're doing this here.

djc (approved):

src/cli/self_update/shell.rs
workingjubilee: > https://github.com/rust-lang/rustup/blob/f6dbe2e4cc186c7624e76691ece2ff37c87cf565/src/cli/self_update/shell.rs#L138-L152
> 
> ... a quick blame points to #2387. @workingjubilee maybe you can provide more context on this one?

My original implementation was very cautious to avoid writing new files to the system, so as to leave the system maximally undisturbed. The problem was that people would then use shells that had not generated even stub rcfiles for themselves. This was most prominent with zsh, as I recall, where macOS, last I checked, generates neither a zshrc nor a zprofile. So what you are seeing is that original implementation, before it was paved over with our current reality.

workingjubilee: We also originally did not handle fish at all.

rami3l: @workingjubilee Thanks for your quick response! Apart from what you said above, we are particularly curious about why `update_rcs` is plural 👀

rami3l: > We also originally did not handle fish at all.

... and that's exactly why we need this context to evaluate our fish support implementation :]

rami3l:
@djc I guess we shouldn't be blocking the inclusion on a reply. I'm merging this now.

src/cli/self_update/shell.rs
workingjubilee: Why would it be singular?

djc: It was just my intuition that it wouldn't make sense to "update" multiple RC files per shell? It seems like if we want to make the rustup proxies work in fish, it would be enough if we update one file? So in that sense `update_rcs() -> Vec<PathBuf>` should be more like `update_rc() -> PathBuf`?

(I'm not big on shell customization so my understanding might be wrong.)

workingjubilee: @djc While imprecise on technical terms, my remarks in https://github.com/rust-lang/rustup/issues/3256#issuecomment-1462306973 remain an accurate summary of the actual situation. In short: No. Not at all. That requires knowledge of not only the OS, not only the shell (and we don't actually know that), but the specific version and distribution of the software.

PR #3713, opened 2024-03-13
Apply clippy suggestions and deny warnings in CI

djc:
Your commit in #3712 fixes one instance, but there are actually multiple clippy warnings.

Fix them all, and update the CI config to fail CI jobs that introduce new clippy warnings.


rami3l (approved):
LGTM. Thanks a lot!

PR #3715, opened 2024-03-14
Replaced `.` with `source` in fish shell's `source_string`

### LINKED ISSUE #3714: Replace `.` with `source` in `conf.d/rustup.fish`

While fish shell technically supports `.` as an alias for [`source`](https://fishshell.com/docs/current/cmds/source.html), this is deprecated. However, after running `rustup-init`, the contents of `~/.config/fish/conf.d/rustup.fish` are `. "$HOME/.cargo/env.fish"`, which will (probably) eventually fail.

### Solution you'd like

During installation of rustup, it should use `source` instead of `.` when generating code for `fish` shell. The contents of `rustup.fish` should be `source "$HOME/.cargo/env.fish"`.

### Notes

Version: rustup 1.27.0 (2024-03-08)

Elsewhere, rustup seems aware of the impending change:

```
Rust is installed now. Great!

To get started you may need to restart your current shell.
This would reload your PATH environment variable to include
Cargo's bin directory ($HOME/.cargo/bin).

```

rben01:
Closes https://github.com/rust-lang/rustup/issues/3714

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -244,7 +244,7 @@ impl UnixShell for Fish {
     }
 
     fn source_string(&self) -> Result<String> {
-        Ok(format!(r#". "{}/env.fish""#, cargo_home_str()?))
+        Ok(format!(r#"source "{}/env.fish""#, cargo_home_str()?))
     }
 }
 


djc (approved):
Thanks!

PR #3719, opened 2024-03-15
docs(dev-guide): update `release-process.md` to match the new workflow based on GitHub Merge Queue

### LINKED ISSUE #3672: Meta: Update `release-process.md` to match the new workflow based on GitHub Merge Queue

In the past we have used a release process based on `git merge`. However, with the inclusion of GHMQ, this is no longer possible. Thus, the instructions regarding cutting a new release should be updated accordingly.

### Solution you'd like

Add/change a few points in the release process to match the current situation, notably:
- Use `git checkout stable && git merge --ff-only master` to update the `stable` branch, which should always contain a subset of `master`'s commit history.
- Update the commit shasum in `rustup-init.sh` not in the release PR, but in a dedicated consequent PR right before tagging.
  - This commit shasum is the one right before the release. When we were using `git merge` it used to point to the exact commit bumping the version in `Cargo.lock`, but since the `master` history is linear now (meaning a new version will appear earlier on `master` than on `stable`, which is indeed the case for many other projects), it should be clearer if we point it to the commit finalizing the `CHANGELOG` instead (cf. https://github.com/rust-lang/rustup/issues/3501#issuecomment-1936968961).
- Tag the release onto the latest `stable` commit. 
- How beta-testing works in Rustup via the `dev` environment, and when exactly to inform `t-release`.

### Notes

Closes #3501.

rami3l:
Closes #3672.

### DIFF
--- a/doc/dev-guide/src/release-process.md
+++ b/doc/dev-guide/src/release-process.md
@@ -3,33 +3,65 @@
 Before making a release, ensure that `rustup-init.sh` is behaving correctly,
 and that you're satisfied that nothing in the ecosystem is breaking because
 of the update. A useful set of things to check includes verifying that
-real-world toolchains install okay, and that `rls-vscode` isn't broken by
+real-world toolchains install okay, and that `rust-analyzer` isn't broken by
 the release. While it's not our responsibility if they depend on non-stable
 APIs, we should behave well if we can.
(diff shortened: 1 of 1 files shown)


djc (approved):
Looks great!

PR #3720, opened 2024-03-16
fix(ci): disable `cargo clippy` on `*-windows-gnu`

### LINKED ISSUE #3718: CI: "This perl implementation doesn't produce Windows like paths"
### Problem

Since a few days ago, our `master` CI (more specifically, `build-windows-master (dev, x86_64-pc-windows-gnu)`) has been reproducibly failing with:

```console
error: failed to run custom build command for `openssl-sys v0.9.92`

Caused by:
  process didn't exit successfully: `D:\a\rustup\rustup\target\debug\build\openssl-sys-5fcaeaa88f6b1b9a\build-script-main` (exit code: 101)
  --- stdout
  cargo:rerun-if-env-changed=X86_64_PC_WINDOWS_MSVC_OPENSSL_NO_VENDOR
  X86_64_PC_WINDOWS_MSVC_OPENSSL_NO_VENDOR unset
  cargo:rerun-if-env-changed=OPENSSL_NO_VENDOR
  OPENSSL_NO_VENDOR unset
  running "perl" "./Configure" "--prefix=D:\\a\\rustup\\rustup\\target\\debug\\build\\openssl-sys-3e664c57f5dedb7e\\out\\openssl-build\\install" "--openssldir=SYS$MANAGER:[OPENSSL]" "no-dso" "no-shared" "no-ssl3" "no-unit-test" "no-comp" "no-zlib" "no-zlib-dynamic" "no-md2" "no-rc5" "no-weak-ssl-ciphers" "no-camellia" "no-idea" "no-seed" "no-engine" "no-asm" "VC-WIN64A"
  Configuring OpenSSL version 1.1.1w (0x1010117fL) for VC-WIN64A
  Using os-specific seed configuration

  --- stderr

  ******************************************************************************
  This perl implementation doesn't produce Windows like paths (with backward
  slash directory separators).  Please use an implementation that matches your
  building platform.

  This Perl version: 5.38.2 for x86_64-msys-thread-multi
  ******************************************************************************
  thread 'main' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\openssl-src-111.28.1+1.1.1w\src\lib.rs:506:13:



  Error configuring OpenSSL build:
      Command: "perl" "./Configure" "--prefix=D:\\a\\rustup\\rustup\\target\\debug\\build\\openssl-sys-3e664c57f5dedb7e\\out\\openssl-build\\install" "--openssldir=SYS$MANAGER:[OPENSSL]" "no-dso" "no-shared" "no-ssl3" "no-unit-test" "no-comp" "no-zlib" "no-zlib-dynamic" "no-md2" "no-rc5" "no-weak-ssl-ciphers" "no-camellia" "no-idea" "no-seed" "no-engine" "no-asm" "VC-WIN64A"
      Exit status: exit code: 127
```

### Steps

N/A

### Possible Solution(s)

~~It seems to me that it's not really an OpenSSL issue (since the release build is green). Rather, this might have been caused by https://github.com/rust-lang/rustup/pull/3713 passing a different (and breaking) set of features to `cargo check` and/or `cargo clippy`, which breaks stuff on some targets not included in our CI triggered on PR.~~

This looks like https://stackoverflow.com/a/43974521.

cc @djc

### Notes

_No response_

### Rustup version

```console
N/A
```

### Installed toolchains

```console
N/A
```

rami3l:
The solution at https://stackoverflow.com/a/43974521 doesn't seem to be working for us, maybe that's because we're not controlling what distribution of Perl we'll be using for our CI builds that easily.

As a workaround, I think it's reasonable to disable `cargo clippy` on `*-windows-gnu` altogether, and keep it this way at least until the removal of OpenSSL. Otherwise, we won't be able to even ship v1.27.1.

Fixes #3718.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -110,7 +110,7 @@ jobs:
           TARGET: ${{ matrix.target }}
           BUILD_PROFILE: ${{ matrix.mode }}
         run: bash ci/run.bash
-      - name: Run cargo check and clippy
+      - name: Run cargo check
         if: matrix.mode != 'release'
         env:
           TARGET: ${{ matrix.target }}
@@ -118,7 +118,12 @@ jobs:
         run: |
           cargo check --all --all-targets --features test
           git ls-files -- '*.rs' | xargs touch
-          cargo clippy --all-targets --all-features -- --deny warnings
+      - name: Run cargo clippy
+        if: matrix.mode != 'release' && matrix.mingw == ''
+        env:
+          TARGET: ${{ matrix.target }}
+        run: |
+          cargo clippy --all-targets --all-features -- -D warnings
       - name: Upload the built artifact
         if: matrix.mode == 'release'
         uses: actions/upload-artifact@v4
@@ -247,7 +252,7 @@ jobs:
           TARGET: ${{ matrix.target }}
           BUILD_PROFILE: ${{ matrix.mode }}
         run: bash ci/run.bash
-      - name: Run cargo check and clippy
+      - name: Run cargo check
         if: matrix.mode != 'release'
         env:
           TARGET: ${{ matrix.target }}
@@ -255,7 +260,12 @@ jobs:
         run: |
           cargo check --all --all-targets --features test
           git ls-files -- '*.rs' | xargs touch
-          cargo clippy --all-targets --all-features -- --deny warnings
+      - name: Run cargo clippy
+        if: matrix.mode != 'release' && matrix.mingw == ''
+        env:
+          TARGET: ${{ matrix.target }}
+        run: |
+          cargo clippy --all-targets --all-features -- -D warnings
       - name: Upload the built artifact
         if: matrix.mode == 'release'
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


djc:
Thanks for taking this on -- I'm not sure I'm fully understanding why a problem arises when we're running clippy for GNU Windows builds?

rami3l:
@djc AFAIK it's somehow still about either OpenSSL or openssl-src, similar to 90% of our previous CI build problems.

The error log complains that we're using a Perl version that doesn't like backslashes. I looked at https://github.com/openssl/openssl/issues/9048 and it seems to me that for building with MingW we do need a MingW Perl... Except according to the error message we ***are*** using a MingW Perl. Maybe it's about the backslashed path being passed in? Maybe the build script doesn't know we're building for `windows-gnu`? 😵‍💫

After quite a few retries I've given up on solving this temporary build problem (given that the builds we're supposed to ship work perfectly fine, it's just the clippy part that's broken).

PS: Enabling all features for clippy implies `vendored-openssl` which is known to cause problems on Windows...

https://github.com/rust-lang/rustup/blob/6e8769c768990c72678de0f6fe092030474ccbcf/ci/run.bash#L11-L15

This conditional dates back to 2019 where Windows CI builds was first added (https://github.com/rust-lang/rustup/commit/9d9a83748d008619f5869985cde1b15380d424a7)!

djc (approved):

djc:
Thanks, sounds like it definitely makes sense to just skip clippy for this target!

PR #3722, opened 2024-03-18
feat: log `RUSTUP_UPDATE_ROOT`, `RUSTUP_DIST_SERVER` or `RUSTUP_DIST_ROOT` on `RUSTUP_DEBUG`

### LINKED ISSUE #3680: Log the `registry` used when run `rustup update/install` (and other commands use registry).

I want to know the registry current used.

### Solution you'd like

Log to screen directly.

### Notes

_No response_

rami3l:
Closes #3680.

This PR makes log the value(s) of `RUSTUP_UPDATE_ROOT`, `RUSTUP_DIST_SERVER` or `RUSTUP_DIST_ROOT` if:
- The environment variable is itself set, and
- `RUSTUP_DEBUG` is set.

This is a tiny yet important feature addition for our users who need to manually designate a Rustup update/dist server, either because they have to [use a mirror server for acceleration](https://mirrors.tuna.tsinghua.edu.cn/help/rustup), or because they need to [switch to the `dev` environment for beta testing](https://rust-lang.github.io/rustup/dev-guide/release-process.html).

The `RUSTUP_DEBUG` predicate minimizes both the extra noise and the required changes to the current test suite.

### DIFF
--- a/doc/user-guide/src/environment-variables.md
+++ b/doc/user-guide/src/environment-variables.md
@@ -1,23 +1,23 @@
 # Environment variables
 
-- `RUSTUP_HOME` (default: `~/.rustup` or `%USERPROFILE%/.rustup`) Sets the
+- `RUSTUP_HOME` (default: `~/.rustup` or `%USERPROFILE%/.rustup`). Sets the
   root `rustup` folder, used for storing installed toolchains and
   configuration options.
 
-- `RUSTUP_TOOLCHAIN` (default: none) If set, will [override] the toolchain used
+- `RUSTUP_TOOLCHAIN` (default: none). If set, will [override] the toolchain used
   for all rust tool invocations. A toolchain with this name should be installed,
   or invocations will fail. This can specify custom toolchains, installable
   toolchains, or the absolute path to a toolchain.
 
-- `RUSTUP_DIST_SERVER` (default: `https://static.rust-lang.org`) Sets the root
+- `RUSTUP_DIST_SERVER` (default: `https://static.rust-lang.org`). Sets the root
   URL for downloading static resources related to Rust. You can change this to
   instead use a local mirror, or to test the binaries from the staging
   directory.
 
-- `RUSTUP_DIST_ROOT` (default: `https://static.rust-lang.org/dist`)
-  Deprecated. Use `RUSTUP_DIST_SERVER` instead.
+- ~~`RUSTUP_DIST_ROOT`~~ *deprecated* (default: `https://static.rust-lang.org/dist`).
+  Use `RUSTUP_DIST_SERVER` instead.
 
-- `RUSTUP_UPDATE_ROOT` (default `https://static.rust-lang.org/rustup`) Sets
+- `RUSTUP_UPDATE_ROOT` (default `https://static.rust-lang.org/rustup`). Sets
   the root URL for downloading self-update.
 
 - `RUSTUP_IO_THREADS` *unstable* (defaults to reported cpu count). Sets the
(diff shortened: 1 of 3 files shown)


rami3l:
Note: The tests are failing as expected. I'll first use these tests to see whether the feature is working, and then add omission indicators to the UI tests if everything goes well.

---

Oops, that's a whole lot of tests! This might need to wait for a bit longer since I'm a bit occupied with other stuff recently...

src/cli/self_update.rs
```diff
@@ -332,7 +332,14 @@ These components can be acquired through a Visual Studio installer.
 
 "#;
 
-static UPDATE_ROOT: &str = "https://static.rust-lang.org/rustup";
+static DEFAULT_UPDATE_ROOT: &str = "https://static.rust-lang.org/rustup";
+
+fn update_root() -> String {
+    process()
+        .var("RUSTUP_UPDATE_ROOT")
+        .inspect(|url| info!("`RUSTUP_UPDATE_ROOT` has been set to `{url}`"))
```
rami3l: Concern: It might not be necessary to print it each time this `update_root` is fetched. Ideally we should print all three altogether at launch time.

src/config.rs
```diff
@@ -230,13 +230,17 @@ impl Cfg {
             .transpose()?;
 
         let dist_root_server = match process().var("RUSTUP_DIST_SERVER") {
-            Ok(ref s) if !s.is_empty() => s.clone(),
+            Ok(s) if !s.is_empty() => {
+                info!("`RUSTUP_DIST_SERVER` has been set to `{s}`");
```
rami3l: Using `info!()` will interfere with `exact` tests. We need to introduce `...` to those tests and make it skip the first line(s), like we did in the UI tests.

rami3l: Maybe it's better to verbose-print those things...

rbtcollins: What if you log at debug! not info!.

rami3l: @rbtcollins I just got a bit sidetracked due to personal reasons. Will try.

rami3l: @rbtcollins Done, although the visibility of this feature has thus gone down a bit. I think a docs change should compensate.

src/cli/self_update.rs
rami3l: I think https://github.com/rust-lang/rustup/pull/3722#discussion_r1560041699 has made it less annoying so I'll leave the logic as-is.

src/config.rs
rami3l: @rbtcollins Done.

doc/user-guide/src/environment-variables.md
```diff
@@ -14,13 +14,13 @@
...
-- `RUSTUP_DIST_ROOT` (default: `https://static.rust-lang.org/dist`)
-  Deprecated. Use `RUSTUP_DIST_SERVER` instead.
+- ~~`RUSTUP_DIST_ROOT`~~ *deprecated* (default: `https://static.rust-lang.org/dist`)
+  Use `RUSTUP_DIST_SERVER` instead.
 
 - `RUSTUP_UPDATE_ROOT` (default `https://static.rust-lang.org/rustup`) Sets
   the root URL for downloading self-update.
 
-- `RUSTUP_IO_THREADS` *unstable* (defaults to reported cpu count). Sets the
+- `RUSTUP_IO_THREADS` *unstable* (defaults to reported cpu count) Sets the
```
rbtcollins: Dropping the `.` leads to poor prose.

doc/user-guide/src/environment-variables.md
```diff
@@ -14,13 +14,13 @@
   instead use a local mirror, or to test the binaries from the staging
   directory.
 
-- `RUSTUP_DIST_ROOT` (default: `https://static.rust-lang.org/dist`)
-  Deprecated. Use `RUSTUP_DIST_SERVER` instead.
+- ~~`RUSTUP_DIST_ROOT`~~ *deprecated* (default: `https://static.rust-lang.org/dist`)
```
rbtcollins: missing . after first sentence

doc/user-guide/src/environment-variables.md
```diff
@@ -29,6 +29,8 @@
   determines the directory that traces will be written too. Traces are of the
   form PID.trace. Traces can be read by the Catapult project [tracing viewer].
 
+- `RUSTUP_DEBUG` *unstable* When set, enables rustup's debug logging.
```
rbtcollins: Missing . after first sentence

doc/user-guide/src/environment-variables.md
rami3l: @rbtcollins That's the style followed by the rest of the file. If you wish to change it, I'm happy to change all of them.

Here is a simple comparison between the two:

image

image

Personally I don't see any readability issues with either of them. What do you think?

djc (approved):
LGTM, thanks!

doc/user-guide/src/environment-variables.md
rbtcollins: Ugh, I hate it when we're already inconsistent!

I think the . present helps scanning the docs - please do use that style.

rami3l: Done.

PR #3724, opened 2024-03-19
Fix "component add" error message

### LINKED ISSUE #3721: Bugs in "component add" message
I think there may be a couple of bugs here:
https://github.com/rust-lang/rustup/blob/6e8769c768990c72678de0f6fe092030474ccbcf/src/toolchain/distributable.rs#L441-L445
- `selector` is not part of the format string, so `{toolchain}` won't get filled in.
- There appears to be a space missing between `{selector}` and `{component}` name.

smoelius:
Fixes #3721 (cc: @rami3l)

### DIFF
--- a/tests/suite/cli_misc.rs
+++ b/tests/suite/cli_misc.rs
@@ -542,6 +542,22 @@ fn run_rls_when_not_installed() {
 }
 
 #[test]
+fn run_rls_when_not_installed_for_nightly() {
+    setup(&|config| {
+        config.expect_ok(&["rustup", "default", "stable"]);
+        config.expect_ok(&["rustup", "toolchain", "install", "nightly"]);
+        config.expect_err(
+            &["rls", "+nightly", "--version"],
+            &format!(
+                "'rls{}' is not installed for the toolchain 'nightly-{}'.\nTo install, run `rustup component add --toolchain nightly-{1} rls`",
+                EXE_SUFFIX,
+                this_host_triple(),
+            ),
+        );
+    });
+}
+
+#[test]
 fn run_rust_lldb_when_not_in_toolchain() {
     clitools::test(Scenario::UnavailableRls, &|config| {
         set_current_dist_date(config, "2015-01-01");
--- a/src/toolchain/distributable.rs
+++ b/src/toolchain/distributable.rs
@@ -439,8 +439,8 @@ impl<'a> DistributableToolchain<'a> {
             } else {
                 // available, not installed, recommend installation
                 let selector = match self.cfg.get_default()? {
-                    Some(ToolchainName::Official(n)) if n == self.desc => "",
-                    _ => " --toolchain {toolchain}",
+                    Some(ToolchainName::Official(n)) if n == self.desc => String::new(),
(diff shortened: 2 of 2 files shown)


rami3l:
@smoelius Thanks a lot, and don't worry about the `cargo check` error!

It's caused by a new `beta` release earlier this week. I'll send a patch to fix it, on which you'd be able to safely rebase this patch.

tests/suite/cli_misc.rs
```diff
@@ -541,6 +541,22 @@ fn run_rls_when_not_installed() {
     });
 }
 
+#[test]
+fn run_rls_when_not_installed_for_nightly() {
+    setup(&|config| {
+        config.expect_ok(&["rustup", "default", "stable"]);
+        config.expect_ok(&["rustup", "install", "nightly"]);
```
rami3l: `rustup install`... I know you all know and use it, but it's a hidden command, so it's behavior can theoretically be broken someday :)

```suggestion
        config.expect_ok(&["rustup", "toolchain", "install", "nightly"]);
```

rami3l (approved):
Nice work!

tests/suite/cli_misc.rs
djc: Are we already emitting a warning about this when the command is executed?

rami3l: > Are we already emitting a warning about this when the command is executed?

@djc I believe we do:

https://github.com/rust-lang/rustup/blob/7790fae8cd49a1c55b6251fab1d5bdde3c5b05ab/src/cli/rustup_mode.rs#L68-L85

https://github.com/rust-lang/rustup/blob/7790fae8cd49a1c55b6251fab1d5bdde3c5b05ab/src/cli/rustup_mode.rs#L168

Actually I guess we've already discussed this and I do wish to make this a true alias in #3596, but that's another story :)

For now I want fewer tests to rely on `rustup install`.

djc: Curious how/why @smoelius decided to include it in this test.

rami3l: @djc Last time I checked, this interface has remain unchanged for years and `rustup toolchain install` has evolved away from it (so they do slightly different things now in v1.27), but the simplest form stays equivalent to `rustup toolchain install` and still has a bunch of users. That's why I decided to making it an alias instead of removing it altogether.

smoelius: > Curious how/why @smoelius decided to include it in this test.

Not sure what you meant, so here are two answers.

Why that syntax? Just because that is what I am accustomed to using.

Why install the nightly toolchain? Because without it, one gets a "toolchain ... is not installed" error.

rami3l:
@smoelius Sorry for all the trouble! I've made https://github.com/rust-lang/rustup/pull/3725 for you which has been merged just now.

As a final step before merging, could you please
- Rebase this PR on the latest `master`, 
- Use `fixup` to squash the 2nd commit into the 1st one?

This should fix the CI errors for you.

Thanks in advance! 🙇

smoelius:
@rami3l Sure. No problem.

PR #3725, opened 2024-03-19
ci: use stable Rust for all clippy lints

rami3l:
Addresses the CI failure in #3724.

## Concerns

- [x] Should we stop using `beta` for clippy lints, now that we're forcing clippy to pass? I don't believe many are actually using `beta` so `cargo clippy --fix` can be difficult as well.

### DIFF
--- a/ci/actions-templates/centos-fmt-clippy-template.yaml
+++ b/ci/actions-templates/centos-fmt-clippy-template.yaml
@@ -49,18 +49,7 @@ jobs: # skip-all
           restore-keys: ${{ github.base_ref }}-${{ runner.os }}-cargo-clippy-target-dir-${{ steps.cargo-target-cache.outputs.rust_hash }}-${{ hashFiles('**/Cargo.lock') }}
       - name: Install Rustup using ./rustup-init.sh
         run: |
-          sh ./rustup-init.sh --default-toolchain=none --profile=minimal -y
-      - name: Ensure Beta is up to date
-        run: |
-          if rustc +beta -vV >/dev/null 2>/dev/null; then
-            rustup toolchain uninstall beta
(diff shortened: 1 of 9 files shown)


djc (approved):

djc:
> * Should we stop using `beta` for clippy lints, now that we're forcing clippy to pass? I don't believe many are actually using `beta` so `cargo clippy --fix` can be difficult as well.

Yeah, I think it would be good to just run clippy from stable.

ci/actions-templates/centos-fmt-clippy-template.yaml
```diff
@@ -49,14 +49,7 @@ jobs: # skip-all
...
         run: |
-          sh ./rustup-init.sh --default-toolchain=none --profile=minimal -y
-      - name: Ensure Beta is up to date
-        run: |
-          if rustc +beta -vV >/dev/null 2>/dev/null; then
-            rustup toolchain uninstall beta
-          fi
-          rustup toolchain install --profile=minimal beta
-          rustup default beta
+          sh ./rustup-init.sh --default-toolchain=stable --profile=minimal -y
```
djc: Kinda unrelated, but I assume this would be more efficient if we just install the full toolchain directly rather than adding rustfmt/clippy at a later stage?

rami3l: @djc Thanks!

djc (approved):

ci/actions-templates/centos-fmt-clippy-template.yaml
djc: FWIW, I would have made this a separate commit.

PR #3742, opened 2024-03-27
fix(utils): don't use `O_NOFOLLOW` in `open_dir()`

### LINKED ISSUE #3737: rustup fails with symlink RUSTUP_HOME/toolchains
### Problem

```
% rustup update
error: I/O Error: Too many levels of symbolic links (os error 40)
```

BTW: This message isn't very helpful.

I'm having a symlink at ~/.rustup/toolchains pointing to ~/.cache/rustup/toolchains

### Steps

1. `mkdir ~/rustup-test && cd ~/.rustup && cp -r downloads tmp update-hashes settings.toml ~/rustup-test`
2. `ln -s ~/.rustup/toolchains ~/rustup-test/`
3. `RUSTUP_HOME=$HOME/rustup-test rustup update`

### Possible Solution(s)

_No response_

### Notes

This happens since the update from

```
   nightly-x86_64-unknown-linux-gnu updated - rustc 1.78.0-nightly (3b1717c05 2024-03-10) (from rustc 1.78.0-nightly (2d24fe591 2024-03-09))
   stable-x86_64-unknown-linux-gnu unchanged - rustc 1.76.0 (07dca489a 2024-02-04)
```

to

```
    nightly-x86_64-unknown-linux-gnu updated - rustc 1.78.0-nightly (3cbb93223 2024-03-13) (from rustc 1.78.0-nig
htly (3b1717c05 2024-03-10))
   stable-x86_64-unknown-linux-gnu unchanged - rustc 1.76.0 (07dca489a 2024-02-04)
```

```
% RUSTUP_HOME=$HOME/rustup-test strace -Zk -e t=openat rustup update
10:27:49.009219 openat(AT_FDCWD</home/joerg/git/rustup>, "/home/joerg/kein_Backup/rustup-test/toolchains", O_RDONLY|O_NOFOLLOW|O_CLOEXEC) = -1 ELOOP (Too many levels of symbolic links) <0.000013>
 > /usr/lib/x86_64-linux-gnu/libc.so.6(__open64+0x51) [0xf7861]
 > /home/joerg/.cache/cargo/bin/rustup(std::sys::unix::fs::File::open_c+0xe8) [0x511148]
 > /home/joerg/.cache/cargo/bin/rustup(std::fs::OpenOptions::_open+0x12a) [0x5391fa]
 > /home/joerg/.cache/cargo/bin/rustup(rustup::toolchain::toolchain::Toolchain::exists+0x172) [0x3fb192]
 > /home/joerg/.cache/cargo/bin/rustup(rustup::toolchain::toolchain::Toolchain::new+0x34) [0x3fac74]
 > /home/joerg/.cache/cargo/bin/rustup(rustup::toolchain::distributable::DistributableToolchain::new+0x3f) [0x3d92af]
 > /home/joerg/.cache/cargo/bin/rustup(<core::iter::adapters::GenericShunt<I,R> as core::iter::traits::iterator::Iterator>::next+0xf80) [0x4e8550]
 > /home/joerg/.cache/cargo/bin/rustup(rustup::config::Cfg::list_channels+0x9a) [0x4e692a]
 > /home/joerg/.cache/cargo/bin/rustup(rustup::cli::rustup_mode::update+0x3b6) [0x4a38c6]
 > /home/joerg/.cache/cargo/bin/rustup(rustup::cli::rustup_mode::main+0x1b2d) [0x47e09d]
 > /home/joerg/.cache/cargo/bin/rustup(rustup_init::main+0x1262) [0xeea32]
 > /home/joerg/.cache/cargo/bin/rustup(std::sys_common::backtrace::__rust_begin_short_backtrace+0x3) [0xec8c3]
 > /home/joerg/.cache/cargo/bin/rustup(main+0x414) [0xef3c4]
 > /usr/lib/x86_64-linux-gnu/libc.so.6(__libc_init_first+0x8a) [0x276ca]
 > /usr/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x85) [0x27785]
 > /home/joerg/.cache/cargo/bin/rustup(_start+0x29) [0xec7f7]
error: I/O Error: Too many levels of symbolic links (os error 40)
```

### Rustup version

```console
rustup 1.27.0 (bbb9276d2 2024-03-08)
info: The currently active `rustc` version is `rustc 1.79.0-nightly (2f090c30d 2024-03-23)`
```

### Installed toolchains

```console
rustup home:  /home/joerg/.rustup

installed toolchains
--------------------

active toolchain
----------------

rustc 1.79.0-nightly (2f090c30d 2024-03-23)
```

rami3l:
Fixes #3737.

For more context on this particular approach, see https://github.com/rust-lang/rustup/issues/3737#issuecomment-2021842441.

## Concerns
- [x] Testing: can we bake a smoke test to prevent regression?

### DIFF
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -45,11 +45,9 @@ pub fn open_dir(p: &Path) -> std::io::Result<File> {
 #[cfg(not(windows))]
 pub fn open_dir(p: &Path) -> std::io::Result<File> {
     use std::fs::OpenOptions;
-    use std::os::unix::fs::OpenOptionsExt;
 
     let mut options = OpenOptions::new();
     options.read(true);
-    options.custom_flags(libc::O_NOFOLLOW);
     options.open(p)
 }
 
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -31,7 +31,7 @@ pub fn is_file<P: AsRef<Path>>(path: P) -> bool {
 }
 
 #[cfg(windows)]
-pub fn open_dir(p: &Path) -> std::io::Result<File> {
+pub fn open_dir_following_links(p: &Path) -> std::io::Result<File> {
     use std::fs::OpenOptions;
     use std::os::windows::fs::OpenOptionsExt;
 
@@ -42,8 +42,9 @@ pub fn open_dir(p: &Path) -> std::io::Result<File> {
     options.custom_flags(FILE_FLAG_BACKUP_SEMANTICS);
     options.open(p)
 }
+
 #[cfg(not(windows))]
-pub fn open_dir(p: &Path) -> std::io::Result<File> {
+pub fn open_dir_following_links(p: &Path) -> std::io::Result<File> {
     use std::fs::OpenOptions;
 
     let mut options = OpenOptions::new();
--- a/tests/suite/cli_misc.rs
+++ b/tests/suite/cli_misc.rs
@@ -677,7 +677,33 @@ fn rename_rls_remove() {
     });
 }
 
+// issue #3737
+/// `~/.rustup/toolchains` is permitted to be a symlink.
+#[test]
+#[cfg(any(unix, windows))]
+fn toolchains_symlink() {
+    use rustup::utils::raw::symlink_dir;
+    use std::fs;
+
+    clitools::test(Scenario::SimpleV2, &|config| {
+        let cwd = config.current_dir();
+        let test_toolchains = cwd.join("toolchains-test");
+        fs::create_dir(&test_toolchains).unwrap();
+        symlink_dir(&test_toolchains, &config.rustupdir.join("toolchains")).unwrap();
+
+        config.expect_ok(&["rustup", "default", "nightly"]);
+        config.expect_ok_contains(&["rustup", "toolchain", "list"], "nightly", "");
+        config.expect_ok_contains(&["rustc", "--version"], "hash-nightly-2", "");
+        config.expect_ok(&["rustup", "toolchain", "uninstall", "nightly"]);
+        config.expect_stdout_ok(
+            &["rustup", "toolchain", "list"],
+            "no installed toolchains\n",
+        );
+    });
+}
+
 // issue #1169
+/// A toolchain that is a stale symlink should be correctly uninstalled.
 #[test]
 #[cfg(any(unix, windows))]
 fn toolchain_broken_symlink() {
--- a/src/toolchain/toolchain.rs
+++ b/src/toolchain/toolchain.rs
@@ -19,7 +19,7 @@ use crate::{
     currentprocess::{process, varsource::VarSource},
     env_var, install,
     notifications::Notification,
-    utils::{raw::open_dir, utils},
+    utils::{raw::open_dir_following_links, utils},
     RustupError,
 };
 
@@ -62,7 +62,7 @@ impl<'a> Toolchain<'a> {
         let base_name = path
             .file_name()
             .ok_or_else(|| RustupError::InvalidToolchainName(name.to_string()))?;
-        let parent_dir = match open_dir(parent) {
+        let parent_dir = match open_dir_following_links(parent) {
             Ok(d) => d,
             Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(false),
             e => e?,


src/utils/raw.rs
```diff
@@ -45,11 +45,9 @@ pub fn open_dir(p: &Path) -> std::io::Result<File> {
 #[cfg(not(windows))]
 pub fn open_dir(p: &Path) -> std::io::Result<File> {
     use std::fs::OpenOptions;
-    use std::os::unix::fs::OpenOptionsExt;
 
     let mut options = OpenOptions::new();
     options.read(true);
-    options.custom_flags(libc::O_NOFOLLOW);
```
djc, line 52: Do we understand why this code was there in the first place? Seems like this might have security implications?

rbtcollins: I'm pretty sure it does, as this is the low level helper used to open directories; and rustup deals with many paths that are required to be directories... as opposed to something like ~/.rustup/toolchains which could be either a dir or a symlink.

tests/suite/cli_misc.rs
```diff
@@ -713,6 +713,45 @@ info: toolchain 'test' uninstalled
     });
 }
 
+// issue #3737
+#[test]
+#[cfg(any(unix, windows))]
+fn toolchain_symlink() {
```
rbtcollins:
```suggestion
/// the ~/.rustup/toolchains path is permitted to be a symlink
fn toolchains_symlink() {
```

tests/suite/cli_misc.rs
```diff
@@ -713,6 +713,45 @@ info: toolchain 'test' uninstalled
...
+    use rustup::utils::raw::symlink_dir;
+    use std::fs;
+
+    clitools::test(Scenario::SimpleV2, &|config| {
+        config.expect_ok(&["rustup", "toolchain", "install", "nightly"]);
```
rbtcollins: this installs a toolchain before running test code; I don't think you want/need that.

tests/suite/cli_misc.rs
```diff
@@ -713,6 +713,45 @@ info: toolchain 'test' uninstalled
...
+
+        let cwd = config.current_dir();
+        let test_rustupdir = cwd.join("rustup-test");
+        let test_toolchains = test_rustupdir.join("toolchains");
+        fs::create_dir_all(&test_toolchains).unwrap();
+        symlink_dir(&config.rustupdir.join("toolchains"), &test_toolchains).unwrap();
+
+        let home_override = &[("RUSTUP_HOME", test_rustupdir.to_str().unwrap())];
```
rbtcollins: no need to override RUSTUP_HOME - its already isolated before any closure code runs

tests/suite/cli_misc.rs
```diff
@@ -713,6 +713,45 @@ info: toolchain 'test' uninstalled
...
+        let out = config.run(
+            "rustup",
+            ["toolchain", "uninstall", "nightly"],
+            home_override,
+        );
+        assert!(out.ok);
+        config.expect_stdout_ok(
+            &["rustup", "toolchain", "list"],
+            "no installed toolchains\n",
+        );
```
rbtcollins: I think this test is ok; it might be better to have a test closer to the code involved - but I'm sure there are still some big obstacles to that.

src/utils/raw.rs
rbtcollins: I think changing the code that opens toolchains itself - whereever that is - to resolve the link, sanity check (/ would be invalid), and then open the resolved path, or alternatively open the directory itself, would be a good approach.

tests/suite/cli_misc.rs
rami3l: I'm just repeating what's been reported in https://github.com/rust-lang/rustup/issues/3737#issue-2204264492: with a `RUSTUP_HOME` override and a symlink pointing to the default `toolchains` in it, Rustup should function correctly.

Of course I can do it another way around, which is creating a `toolchains` elsewhere and make a symlink under the default `RUSTUP_HOME`, install the nightly and verify it's there, uninstall it and verify it's gone.

tests/suite/cli_misc.rs
rami3l: See https://github.com/rust-lang/rustup/pull/3742#discussion_r1542202816.

tests/suite/cli_misc.rs
rami3l: I want to do this as well, however I put this function right below

https://github.com/rust-lang/rustup/blob/29ce9ee0ae3e5f12e217d939f6e82c05fbf0d4dd/tests/suite/cli_misc.rs#L680-L683

for a reason. Maybe I should change the name of that test accordingly.

src/utils/raw.rs
rami3l: > I'm pretty sure it does, as this is the low level helper used to open directories; and rustup deals with many paths that are required to be directories... as opposed to something like ~/.rustup/toolchains which could be either a dir or a symlink.

@rbtcollins I guess so, but I don't currently see this "helper function" being used elsewhere...

> OTOH the only caller of this function is `Toolchain::exists`, and I don't see a particular reason why we can't follow a symlink for this purpose.
_https://github.com/rust-lang/rustup/issues/3737#issuecomment-2016776591_

image

... so a name change or even a simple docstring would help in this case. However, by "changing the code that opens toolchains itself" we will actually lose this function due to `dead_code`.

---

If I understand what you wrote correctly, the Windows version of `open_dir()` currently follows symlinks, but the Unix version doesn't; however the two versions should both follow symlinks (or not) for the sake of consistency.

> the open with FILE_FLAG_BACKUP_SEMANTICS is required to read the link out of a link, compared to linux where a different syscall is used to read links.
_https://github.com/rust-lang/rustup/issues/3737#issuecomment-2023475496_

https://github.com/rust-lang/rustup/blob/0c501d55c54031a992395078d535b6d2574dc693/src/utils/raw.rs#L33-L44

rami3l: Renamed `open_dir()` as suggested by @rbtcollins.

rbtcollins (approved):

PR #3748, opened 2024-03-29
fix(doc): don't show the opening message when --path is used

0x61nas:
This might seem like a travel or unnecessary thing, especially since the message is written to `stderr`  so it won't affect piping or similar use cases.
But since `stdout` and `stderr` are ambiguous in most terminals (i.e., they're printed in the output) this might cause some confusion and/or annoyance to some users.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1594,18 +1594,15 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     };
 
     let topical_path: PathBuf;
+    let mut doc_name = m.get_one::<String>("topic").map(|s| s.as_str());
 
-    let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
+    let doc_url = if let Some(topic) = doc_name {
         topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
         topical_path.to_str().unwrap()
     } else if let Some((name, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
-        writeln!(
-            process().stderr().lock(),
-            "Opening docs named `{name}` in your browser"
-        )?;
+        doc_name = Some(name);
         path
     } else {
-        writeln!(process().stderr().lock(), "Opening docs in your browser")?;
         "index.html"
     };
 
@@ -1614,6 +1611,14 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
         writeln!(process().stdout().lock(), "{}", doc_path.display())?;
         Ok(utils::ExitCode(0))
     } else {
+        if let Some(name) = doc_name {
+            writeln!(
+                process().stderr().lock(),
+                "Opening docs named `{name}` in your browser"
+            )?;
+        } else {
+            writeln!(process().stderr().lock(), "Opening docs in your browser")?;
+        }
         toolchain.open_docs(doc_url)?;
         Ok(utils::ExitCode(0))
     }


src/cli/rustup_mode.rs
```diff
@@ -1594,18 +1594,15 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     };
 
     let topical_path: PathBuf;
+    let mut topic_name = None;
```
```suggestion
    let mut topic_name = m.get_one::<String>("topic");
```

src/cli/rustup_mode.rs
```diff
@@ -1594,18 +1594,15 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
 
     let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
         topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
         topical_path.to_str().unwrap()
     } else if let Some((name, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
-        writeln!(
-            process().stderr().lock(),
-            "Opening docs named `{name}` in your browser"
-        )?;
+        topic_name = Some(name);
```
rami3l, line 1603:
```suggestion
        doc_name = Some(name);
```

src/cli/rustup_mode.rs
```diff
@@ -1594,18 +1594,15 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     };
 
     let topical_path: PathBuf;
+    let mut topic_name = None;
 
     let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
```
```suggestion
    let doc_url = if let Some(topic) = topic_name {
```

rami3l (approved):
LGTM!

Besides, your `topic_name` does remind me of an edge case that I forgot to handle in #3674: `rustup doc std::fs`.
Reusing this variable gives ```Opening docs named `std::fs` in your browser```, I guess that looks just right?

0x61nas:
> LGTM!
> 
> Besides, your `topic_name` does remind me of an edge case that I forgot to handle in #3674: `rustup doc std::fs`. Reusing this variable gives `` Opening docs named `std::fs` in your browser ``, I guess that looks just right?

~this edge case are still exists, when i try `rustup doc std::fs` it rising this message `Opening docs in your browser`, as if i didn't specify a topic, but still opens the correct page~

sorry, i forget to run the init command, yup every thing seams to work fine now

src/cli/rustup_mode.rs
```diff
@@ -1594,18 +1594,14 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     };
 
     let topical_path: PathBuf;
+    let doc_name = m.get_one::<String>("topic");
```
```suggestion
    let mut doc_name = m.get_one::<String>("topic");
```

src/cli/rustup_mode.rs
```diff
@@ -1594,18 +1594,14 @@ fn doc(cfg: &Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
...
-    let doc_url = if let Some(topic) = m.get_one::<String>("topic") {
+    let doc_url = if let Some(topic) = doc_name {
         topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
         topical_path.to_str().unwrap()
-    } else if let Some((name, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
-        writeln!(
-            process().stderr().lock(),
-            "Opening docs named `{name}` in your browser"
-        )?;
+    } else if let Some((_, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
```
rami3l: `rustup doc --cargo` should also give something... I remember you had this one in your original commit, right? 
```suggestion
    } else if let Some((name, _, path)) = DOCS_DATA.iter().find(|(name, _, _)| m.get_flag(name)) {
        doc_name = Some(name);
```

0x61nas: is this `name` can be different from the provided one?

rami3l: > is this `name` can be different from the provided one?

@0x61nas Sorry, I don't quite see what you mean.

Anyway, in the case of `rustup doc --cargo`, `name` should be assigned "cargo".

And given the `else if` at the beginning of this line, `doc_name` is either something like `Some("std::fs")` (via `rustup doc std::fs`) or `Some("cargo")` (via `rustup doc --cargo`) or `None`.

rami3l: @0x61nas If you were asking if `None` is ever possible, then the answer would be yes, but for fallback purposes mostly. In that case, I think it's okay not to provide a "pretty" name.

0x61nas: oh, i didnt see that, appreciate your response, thanks

rami3l:
@0x61nas Thanks again!

PR #3750, opened 2024-03-30
Add hr to Windows instructions

pie-flavor:
This makes it more visually distinct that there are two different sets of instructions for Windows, in case you are ignoring all of the text and skipping straight to the command. A user on Discord did this and was very confused.


rami3l:
@pie-flavor Thanks! Since this is a visual design change, maybe you could provide a screenshot for this change for the record?

pie-flavor:
Sure. It currently looks like this:

And I'm proposing it look like this:

rami3l (approved):
Nice work, and thanks again for the help @pie-flavor!

PR #3752, opened 2024-04-01
fix(utils): unlink input path in `delete_dir_contents()` if it's a link

rami3l:
Closes #3344.

PR #3753, opened 2024-04-01
Revert "fix(utils): unlink input path in `delete_dir_contents()` if it's a link"

rbtcollins:
Reverts rust-lang/rustup#3752

This commit  looks like a regression in multiple directiions: it uses fs::remove_dir_all which has had issues with Windows reliable deletiing in the past - and is single threads vs remove_dir_all; and it removes the link which will regress https://github.com/rust-lang/rustup/issues/2430.

A better fix is to resolve the link in the calling code and then use remove_dir_contents as before.


djc (approved):
Thanks for being on top of this!

rami3l:
@rbtcollins Sorry about that! Will try again later 🙇

PR #3754, opened 2024-04-02
fix(utils): resolve input path in `delete_dir_contents()` if it's a link

### LINKED ISSUE #3344: Regression when ~/.rustup/downloads and/or tmp are symlinks
### Problem

```
rustcargo@zealot:/volatile/rustcargo/Rustup/Arti/arti$ ls -ald ~/.rustup/{tmp,downloads}{/,}
lrwxrwxrwx 1 rustcargo rustcargo      37 Jun  8  2021 /home/rustcargo/.rustup/downloads -> /volatile/rustcargo/rustup/downloads/
drwxrwxr-x 2 rustcargo rustcargo    4096 May  3 12:43 /home/rustcargo/.rustup/downloads/
lrwxrwxrwx 1 rustcargo rustcargo      31 Jun  8  2021 /home/rustcargo/.rustup/tmp -> /volatile/rustcargo/rustup/tmp/
drwxrwxr-x 2 rustcargo rustcargo 1249280 May  3 12:45 /home/rustcargo/.rustup/tmp/
rustcargo@zealot:/volatile/rustcargo/Rustup/Arti/arti$ rustup update
info: syncing channel updates for 'beta-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.31-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.37-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.40-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.51-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.53-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.54-x86_64-unknown-linux-gnu'
info: syncing channel updates for '1.63-x86_64-unknown-linux-gnu'

   stable-x86_64-unknown-linux-gnu unchanged - rustc 1.69.0 (84c898d65 2023-04-16)
     beta-x86_64-unknown-linux-gnu unchanged - rustc 1.70.0-beta.2 (071f14baa 2023-04-30)
  nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.71.0-nightly (a368898de 2023-05-02)
     1.31-x86_64-unknown-linux-gnu unchanged - rustc 1.31.1 (b6c32da9b 2018-12-18)
     1.37-x86_64-unknown-linux-gnu unchanged - rustc 1.37.0 (eae3437df 2019-08-13)
     1.40-x86_64-unknown-linux-gnu unchanged - rustc 1.40.0 (73528e339 2019-12-16)
     1.51-x86_64-unknown-linux-gnu unchanged - rustc 1.51.0 (2fd73fabe 2021-03-23)
     1.53-x86_64-unknown-linux-gnu unchanged - rustc 1.53.0 (53cb7b09b 2021-06-17)
     1.54-x86_64-unknown-linux-gnu unchanged - rustc 1.54.0 (a178d0322 2021-07-26)
     1.63-x86_64-unknown-linux-gnu unchanged - rustc 1.63.0 (4b91a6ea7 2022-08-08)

thread 'main' panicked at 'Unable to clean up /home/rustcargo/.rustup/downloads: Os { code: 40, kind: FilesystemLoop, message: "Too many levels of symbolic links" }', src/utils/utils.rs:650:13
```

### Steps

1. Make `~/.rustup/tmp` and `~/.rustup/downloads` symlinks to like-named directories elsewhere
2. Run rustup

### Possible Solution(s)

_No response_

### Notes

I assume something has changed in rustup's handling of these directories.  This used to work.  I could probably bisect it if it's not obvious.

### Rustup version

```console
rustcargo@zealot:/volatile/rustcargo/Rustup/Arti/arti$ rustup --version
info: The currently active `rustc` version is `rustc 1.70.0-beta.2 (071f14baa 2023-04-30)`
rustcargo@zealot:/volatile/rustcargo/Rustup/Arti/arti$ 
```

### Installed toolchains

```console
rustcargo@zealot:/volatile/rustcargo/Rustup/Arti/arti$ rustup show
rustup home:  /home/rustcargo/.rustup

installed toolchains
--------------------

beta-x86_64-unknown-linux-gnu (default)
nightly-2018-12-19-x86_64-unknown-linux-gnu
nightly-2022-06-24-x86_64-unknown-linux-gnu
nightly-2022-11-12-x86_64-unknown-linux-gnu
1.31-x86_64-unknown-linux-gnu
1.37-x86_64-unknown-linux-gnu
1.40-x86_64-unknown-linux-gnu
1.51-x86_64-unknown-linux-gnu
1.53-x86_64-unknown-linux-gnu
1.54-x86_64-unknown-linux-gnu
1.63-x86_64-unknown-linux-gnu
local
stage1
1.63.0-x86_64-unknown-linux-gnu

wasm32-unknown-unknown
x86_64-unknown-linux-gnu
x86_64-unknown-linux-musl

active toolchain
----------------

beta-x86_64-unknown-linux-gnu (default)
rustc 1.70.0-beta.2 (071f14baa 2023-04-30)

rustcargo@zealot:/volatile/rustcargo/Rustup/Arti/arti$ 
```

rami3l:
Closes #3344, superseding #3452:

> A better fix is to resolve the link in the calling code and then use remove_dir_contents as before.
_https://github.com/rust-lang/rustup/pull/3753#issue-2218949990_

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1007,7 +1007,7 @@ fn update(cfg: &mut Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     } else {
         common::update_all_channels(cfg, self_update, m.get_flag("force"))?;
         info!("cleaning up downloads & tmp directories");
-        utils::delete_dir_contents(&cfg.download_dir);
+        utils::delete_dir_contents_following_links(&cfg.download_dir);
         cfg.temp_cfg.clean();
     }
 
--- a/src/dist/temp.rs
+++ b/src/dist/temp.rs
@@ -158,7 +158,7 @@ impl Cfg {
     }
 
     pub(crate) fn clean(&self) {
-        utils::delete_dir_contents(&self.root_directory);
+        utils::delete_dir_contents_following_links(&self.root_directory);
     }
 }
 
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -610,7 +610,7 @@ where
     })
 }
 
-pub(crate) fn delete_dir_contents(dir_path: &Path) {
+pub(crate) fn delete_dir_contents_following_links(dir_path: &Path) {
     use remove_dir_all::RemoveDir;
 
     match raw::open_dir_following_links(dir_path).and_then(|mut p| p.remove_dir_contents(None)) {
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -611,7 +611,9 @@ where
 }
 
 pub(crate) fn delete_dir_contents(dir_path: &Path) {
-    match remove_dir_all::remove_dir_contents(dir_path) {
+    use remove_dir_all::RemoveDir;
+
+    match raw::open_dir_following_links(dir_path).and_then(|mut p| p.remove_dir_contents(None)) {
         Err(e) if e.kind() != io::ErrorKind::NotFound => {
             panic!("Unable to clean up {}: {:?}", dir_path.display(), e);
         }
--- a/tests/suite/cli_misc.rs
+++ b/tests/suite/cli_misc.rs
@@ -702,6 +702,36 @@ fn toolchains_symlink() {
     });
 }
 
+// issue #3344
+/// `~/.rustup/tmp` and `~/.rustup/downloads` are permitted to be symlinks.
+#[test]
+#[cfg(any(unix, windows))]
+fn tmp_downloads_symlink() {
+    use rustup::utils::raw::symlink_dir;
+    use std::fs;
+
+    clitools::test(Scenario::ArchivesV2, &|config| {
+        let cwd = config.current_dir();
+
+        let test_tmp = cwd.join("tmp-test");
+        fs::create_dir(&test_tmp).unwrap();
+        symlink_dir(&test_tmp, &config.rustupdir.join("tmp")).unwrap();
+
+        let test_downloads = cwd.join("tmp-downloads");
+        fs::create_dir(&test_downloads).unwrap();
+        symlink_dir(&test_downloads, &config.rustupdir.join("downloads")).unwrap();
+
+        set_current_dist_date(config, "2015-01-01");
+        config.expect_ok(&["rustup", "default", "nightly"]);
+
+        set_current_dist_date(config, "2015-01-02");
+        config.expect_ok(&["rustup", "update"]);
+
+        assert!(config.rustupdir.join("tmp").exists());
+        assert!(config.rustupdir.join("downloads").exists());
+    });
+}
+
 // issue #1169
 /// A toolchain that is a stale symlink should be correctly uninstalled.
 #[test]


tests/suite/cli_misc.rs
```diff
@@ -702,6 +702,36 @@ fn toolchains_symlink() {
     });
 }
 
+// issue #3344
```
rbtcollins, line 705: was this deliberately a comment line rather than doc line?

tests/suite/cli_misc.rs
```diff
@@ -702,6 +702,36 @@ fn toolchains_symlink() {
...
+/// `~/.rustup/tmp` and `~/.rustup/downloads` are permitted to be symlinks.
+#[test]
+#[cfg(any(unix, windows))]
```
rbtcollins, line 708: isn't this equivalent to all platforms?

src/utils/utils.rs
```diff
@@ -611,7 +611,9 @@ where
 }
 
 pub(crate) fn delete_dir_contents(dir_path: &Path) {
```
rbtcollins: this looks pretty good. I think perhaps renaming this function as well (or having a non-following-version and a following-version for callers where we don't want to escape out of our trees) would be a good idea, but I haven't checked in the code where delete_dir_contents is invoked today.

tests/suite/cli_misc.rs
rami3l: Yes, this is to be coherent with other tests above and below.

tests/suite/cli_misc.rs
rami3l: There are indeed other platforms such as wasm, although I don't know if Rustup is meaningful there or not. Similar to other tests above/below it, this one relies on the fact that we compile for Unix or Windows. I just copied the code over as I mentioned above.

djc: Given that platforms like WASM exist, this seems okay.

src/utils/utils.rs
djc: The only caller is `rustup_mode::update()`, which calls this directly with `cfg.download_dir` and via `temp::Cfg`, which calls this with `temp::Cfg::root_directory`.

djc (approved):

src/utils/utils.rs
rami3l: @rbtcollins I've changed the function name accordingly.

src/cli/rustup_mode.rs
```diff
@@ -1007,7 +1007,7 @@ fn update(cfg: &mut Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
     } else {
         common::update_all_channels(cfg, self_update, m.get_flag("force"))?;
         info!("cleaning up downloads & tmp directories");
-        utils::delete_dir_contents(&cfg.download_dir);
+        utils::delete_dir_contents_following_links(&cfg.download_dir);
```
rbtcollins, line 1010: ok so this fine to be a symlink; its a dir only the user controls.

src/dist/temp.rs
```diff
@@ -158,7 +158,7 @@ impl Cfg {
     }
 
     pub(crate) fn clean(&self) {
-        utils::delete_dir_contents(&self.root_directory);
+        utils::delete_dir_contents_following_links(&self.root_directory);
```
rbtcollins, line 161: this should never be a symlink, it is created on line 105. Its probably ok to permit following links in order to get this regression fixed, but longer term we should only be deleting dirs here, since getting a symlink is impossible without someone altering state

tests/suite/cli_misc.rs
rbtcollins: I think this is a side effect of code evolution, it isn't the pattern in general, and rustup does depend on hardlinks to operate at all. Are there platforms that have hardlinks, but not symlinks?

src/dist/temp.rs
rami3l: @rbtcollins Hmmm https://github.com/rust-lang/rustup/issues/3344 explicitly mentioned the case where `tmp` is a link, so I'm not quite sure about that. BTW L105 calls `raw::ensure_dir_exists` and is actually compatible with a pre-existing link (cf. the test case).

tests/suite/cli_misc.rs
rami3l: @rbtcollins Looks like symlinks are also available on WASM/WASI: https://doc.rust-lang.org/std/os/wasi/fs/fn.symlink.html, so I'm also okay with removing this line. Anyway Rustup in its current form will fail to build on WASM and not all dependencies of this test work on WASM anyway.

src/dist/temp.rs
rbtcollins: You added a regression test for that, so empirical validatation of whether its the one or not should be quite easy. I don't think it is as this is the tmp root for one dist operation, not the dir that temp operations are rooted in. But if I'm wrong thats good to know too :)  - and perhaps capture somehow

rbtcollins: On the other other hand, this is clean() which suggests preservation of the dir - so ok, lets leave this as a future puzzle.

rbtcollins (approved):

src/dist/temp.rs
rami3l: > You added a regression test for that, so empirical validatation of whether its the one or not should be quite easy.

@rbtcollins Replacing this line with the old implementation gives:

```console
out.stderr (18 lines):
    info: syncing channel updates for 'nightly-aarch64-apple-darwin'
    info: latest update on 2015-01-02, rust version 1.3.0 (hash-nightly-2)
    info: removing previous version of component 'cargo'
    info: removing previous version of component 'rust-docs'
    info: removing previous version of component 'rust-std'
    info: removing previous version of component 'rustc'
    thread 'main' panicked at src/utils/utils.rs:616:13:
    Unable to clean up <snip>/rustupRd3iIp/tmp: Os { code: 62, kind: FilesystemLoop, message: "Too many levels of symbolic links" }
```

So I guess both pieces are indeed required for the fix to work.

PR #3761, opened 2024-04-09
Rename temp::Cfg to temp::Context

djc:
While reviewing #3754 I was a little surprised to note that the `temp::Cfg` type was different from `config::Config` (and then apparently there's also `dist::config::Config`). As far as I can see `temp::Cfg` also really doesn't represent "configuration" in the sense that it's input provided by the user or an external process. Renamed it to `TemporaryContext` instead.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1008,7 +1008,7 @@ fn update(cfg: &mut Cfg, m: &ArgMatches) -> Result<utils::ExitCode> {
         common::update_all_channels(cfg, self_update, m.get_flag("force"))?;
         info!("cleaning up downloads & tmp directories");
         utils::delete_dir_contents_following_links(&cfg.download_dir);
-        cfg.temp_cfg.clean();
+        cfg.tmp_cx.clean();
     }
 
     if !self_update::NEVER_SELF_UPDATE && self_update_mode == SelfUpdateMode::CheckOnly {
--- a/src/config.rs
+++ b/src/config.rs
@@ -179,7 +179,7 @@ pub(crate) struct Cfg {
     pub toolchains_dir: PathBuf,
     pub update_hash_dir: PathBuf,
     pub download_dir: PathBuf,
-    pub temp_cfg: temp::Cfg,
+    pub tmp_cx: temp::Context,
     pub toolchain_override: Option<ResolvableToolchainName>,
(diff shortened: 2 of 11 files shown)


rami3l (approved):
The intention of the original naming is that it [should be referenced with `temp::Cfg`](https://rust-lang.github.io/rust-clippy/master/#/module_name_repetitions), although I think `Cfg` is definitely confusing...

Frankly speaking I would personally prefer `temp::Context`, but since we already have other stuff like `download::DownloadCfg` I think this is acceptable as well.

rami3l:
~~BTW pedantically speaking we should have a `BREAKING` marker on the commit message that changes the API name :)~~

djc:
> Frankly speaking I would personally prefer `temp::Context`, but since we already have other stuff like `download::DownloadCfg` I think this is acceptable as well.

That's fair -- changed it to `temp::Context` instead.

> BTW pedantically speaking we should have a `BREAKING` marker on the commit message that changes the API name :)

Why? Is there anything tracking those? IIRC we don't even pretend to have a stable library API. (FWIW in the many projects that I maintain none use `BREAKING` markers in commit messages.)

rami3l:
> > BTW pedantically speaking we should have a `BREAKING` marker on the commit message that changes the API name :)
> 
> Why? Is there anything tracking those? IIRC we don't even pretend to have a stable library API. (FWIW in the many projects that I maintain none use `BREAKING` markers in commit messages.)

@djc I agree with you that we don't have a stable set of external APIs. OTOH I do have the impression somewhere down the issue threads there's a post saying in the long run we're moving this way... but I failed to find the exact location. Anyway, I think it'd be better to just forget about what I said on this one for now 🤦‍♀️

rami3l:
@djc There seems to be a conflict with #3763 that went undetected by GitHub but has been caught by the CI. (Thanks, GHMQ!)

rbtcollins:
>      @djc I agree with you that we don't have a stable set of external APIs. OTOH I do have the impression somewhere down the issue threads there's a post saying in the long run we're moving this way... but I failed to find the exact location. Anyway, I think it'd be better to just forget about what I said on this one for now 🤦‍♀️

We don't have an external API at all.

I think we should have a *process* interface that is machine readable - e.g. protobuf - but not a use-as-library interface, since we have little to no evidence that there is a need for a library interface.

So there is no need to pay a compatibilty tax on APIs

PR #3763, opened 2024-04-09
Remove derivative dependency in favor of manual implementation

djc:
As an added benefit, derivative was the only dependency still relying on syn 1, so this should help compile times.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,7 +130,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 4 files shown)


rami3l (approved):
It's true that `#[derivative(Debug = "ignore")]` looks like a very rare case in our codebase and it's not worth including `derivative` in our workspace for just that.

Let's wait for https://github.com/rust-lang/rust/issues/37009 to land on Rust std one day, if it's possble at all XD

rbtcollins:
Its nice to have the compile times improved (potentially - not measured), but this increases cognitive load when changes to the struct are made :(, so I'd really rather we solved that by choosing a non-syn backed crate, rather than removing the proc-macro entirely.

PR #3765, opened 2024-04-09
Use std IsTerminal interface

djc:
Was added to std in 1.70.


rami3l (approved):
LGTM!

... do we happen to have an MSRV requirement somewhere though?
(I don't see any, so probably everything is fine.)

djc:
From what I understand we're just relying on stable.

rbtcollins:
> From what I understand we're just relying on stable.
Rustup has no library interface, and we don't support targets that aren't reasonably stable themselves yet, so as long as we can build the binary in CI, we're good to pick whatever we want.

In the past I think we even shipped from beta at one point, but for many years now its been stable, with lints checked on beta so we can provide early feedback to candidate lints (which  I note @djc / @rami3l have change to be stable too now)

PR #3767, opened 2024-04-09
Use pattern matching to make Debug impl for Cfg more robust

djc:
Follow-up from #3763:

> > Its nice to have the compile times improved (potentially - not measured), but this increases cognitive load when changes to the struct are made :(, so I'd really rather we solved that by choosing a non-syn backed crate, rather than removing the proc-macro entirely.

This change should help reduce cognitive overload when changes to the `struct` are made.

For me, crates like derivative also have cognitive load because their exact workings are kind of magic -- compared to a few lines of code doing a manual impl, that doesn't seem like a good deal to me.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -953,18 +953,33 @@ impl Cfg {
 
 impl Debug for Cfg {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self {
+            profile_override,
+            rustup_dir,
+            settings_file,
+            fallback_settings,
+            toolchains_dir,
+            update_hash_dir,
+            download_dir,
+            tmp_cx,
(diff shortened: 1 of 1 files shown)


rbtcollins (approved):

rbtcollins:
Thank you, this does address much of my concern. It will still require manual effort but at least it won't require thinking ;).

I wish Rust had a better story for compile time code execution, because I do totally get the its-magic aspect particularly for troubleshooting issues that arise when the generators go wrong.

OTOH, it is very convenient, and I've learned to embrace the magic much of the time :)

djc:
Don't get me wrong, I love my procedural macros. I just think that all of the little derive utility crates generally add very limited value over manual implementations, which are pretty straightforward and entirely "in view".

PR #3769, opened 2024-04-11
fix: add missing links for loongarch64

heiher:


rami3l (approved):
@heiher LGTM! As long as you're sure these are indeed valid links, I guess we're good to go.

PR #3770, opened 2024-04-12
fix(windows): update `DisplayVersion` on `rustup self update`

### LINKED ISSUE #3739: rustup self update doesn't update DisplayVersion
### Problem

When updating rustup from a prior release using `rustup update`, the DisplayVersion in the registry isn't updated from the old release version. This will result in tools like winget thinking rustup is outdated, and might confuse users.

### Steps

1. Install an older release of rustup that sets DisplayVersion such as 1.26.0
2. Run `rustup update`, and check that rustup is updated to the latest version
3. Open the "Uninstall a program" menu in control panel or check `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Rustup` in the registry to see the outdated version number

### Possible Solution(s)

Maybe `do_add_to_programs` isn't called for updates? I haven't checked the code yet.

### Notes

Related to #3047 and #3055

### Rustup version

```console
rustup 1.27.0 (bbb9276d2 2024-03-08)
info: The currently active `rustc` version is `rustc 1.77.0 (aedd173a2 2024-03-17)`
```

### Installed toolchains

```console
rustup home:  C:\Users\User\.rustup

rustc 1.77.0 (aedd173a2 2024-03-17)
```

rami3l:
Closes #3739.

My daily driver is a Mac (and OpenSSL refuses to cross-build here) so this is pretty much a best-effort blind fix.

cc @ChrisDenton 

## Concerns

- [x] ~~Is it necessary to add a test case for this one? If so, how? (Considering the sandboxing requirements, doing this properly might be hard.)~~ A sandboxed test case has been added.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1106,12 +1106,9 @@ pub(crate) fn update(cfg: &Cfg) -> Result<utils::ExitCode> {
 
     match prepare_update()? {
         Some(setup_path) => {
-            let version = match get_new_rustup_version(&setup_path) {
-                Some(new_version) => parse_new_rustup_version(new_version),
-                None => {
-                    err!("failed to get rustup version");
-                    return Ok(utils::ExitCode(1));
-                }
+            let Some(version) = get_and_parse_new_rustup_version(&setup_path) else {
+                err!("failed to get rustup version");
+                return Ok(utils::ExitCode(1));
             };
 
             let _ = common::show_channel_update(
@@ -1133,6 +1130,10 @@ pub(crate) fn update(cfg: &Cfg) -> Result<utils::ExitCode> {
     }
 
     Ok(utils::ExitCode(0))
+}
+
+fn get_and_parse_new_rustup_version(path: &Path) -> Option<String> {
+    get_new_rustup_version(path).map(parse_new_rustup_version)
 }
 
 fn get_new_rustup_version(path: &Path) -> Option<String> {
--- a/src/cli/self_update/test.rs
+++ b/src/cli/self_update/test.rs
@@ -2,13 +2,13 @@
 
 use std::sync::Mutex;
 
-#[cfg(not(unix))]
+#[cfg(windows)]
 use winreg::{
     enums::{HKEY_CURRENT_USER, KEY_READ, KEY_WRITE},
     RegKey, RegValue,
 };
 
-#[cfg(not(unix))]
+#[cfg(windows)]
 pub fn get_path() -> std::io::Result<Option<RegValue>> {
     let root = RegKey::predef(HKEY_CURRENT_USER);
     let environment = root
@@ -21,7 +21,7 @@ pub fn get_path() -> std::io::Result<Option<RegValue>> {
     }
 }
 
-#[cfg(not(unix))]
+#[cfg(windows)]
 fn restore_path(p: Option<RegValue>) {
     let root = RegKey::predef(HKEY_CURRENT_USER);
     let environment = root
--- a/src/cli/self_update/test.rs
+++ b/src/cli/self_update/test.rs
@@ -8,6 +8,28 @@ use winreg::{
     RegKey, RegValue,
 };
 
+/// Support testing of code that mutates global state
+pub fn with_saved_global_state<S>(
+    getter: impl Fn() -> io::Result<S>,
+    setter: impl Fn(S),
+    f: &mut dyn FnMut(),
+) {
+    // Lock protects concurrent mutation of registry
+    static LOCK: Mutex<()> = Mutex::new(());
+    let _g = LOCK.lock();
+
+    // Save and restore the global state here to keep from trashing things.
+    let saved_state =
+        getter().expect("Error getting global state: Better abort to avoid trashing it");
+    let _g = scopeguard::guard(saved_state, setter);
+
+    f();
+}
+
+pub fn with_saved_path(f: &mut dyn FnMut()) {
+    with_saved_global_state(get_path, restore_path, f)
+}
+
 #[cfg(windows)]
 pub fn get_path() -> io::Result<Option<RegValue>> {
     let root = RegKey::predef(HKEY_CURRENT_USER);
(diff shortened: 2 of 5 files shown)


ChrisDenton:
This looks fine to me. I do wonder if failure should be a warning instead of an error? It's not technically critical to the update process, though it is important to inform the user of it.

tests/suite/cli_self_upd.rs
```diff
@@ -304,7 +304,24 @@ info: downloading self-update
             &["rustup", "self", "update"],
             &format!("  rustup updated - {version} (from {version})\n\n",),
```
rami3l, line 308: This is a bit weird since in this test `version` appears 2 times (so they are the same thing?), and technically my addition can produce a false negative if the `DisplayVersion` is somehow not updated. However I cannot think of a better place to add it... We could possibly change the `DisplayVersion` before `rustup self update` to be sure, but that breaks the sandboxing.

rami3l: I chose to remove the `DisplayVersion` and see if `rustup self update` re-adds it. This also breaks the sandbox but in a more harmless way, I expect.

src/cli/self_update.rs
```diff
@@ -1128,6 +1125,10 @@ pub(crate) fn update(cfg: &Cfg) -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
+fn get_and_parse_new_rustup_version(path: &Path) -> Option<String> {
```
rbtcollins, line 1135: I'm not sure this one-off function carries its own weight.

```
  let Some(version) = get_new_rustup_version(&setup_path).map(parse_new_rustup_version) else {...}
```

doesn't have enough complexity, reusability, or cyclomatic complexity to worry about

src/cli/self_update/windows.rs
```diff
@@ -500,13 +500,23 @@ pub(crate) fn do_remove_from_path() -> Result<()> {
...
+    Ok(RegKey::predef(HKEY_CURRENT_USER)
+        .create_subkey(RUSTUP_UNINSTALL_ENTRY)
+        .context("Failed creating uninstall key")?
+        .0)
+}
+
+pub(crate) fn do_update_programs_display_version(version: &str) -> Result<()> {
+    rustup_uninstall_reg_key()?
+        .set_value("DisplayVersion", &version)
+        .context("Failed to set display version")
```
rbtcollins: perhaps either DisplayVersion (as it is a concrete thing) or

`Failed to record version for uninstall` (as this is the conceptual layer)

src/cli/self_update.rs
rbtcollins: nvm I see its reused

tests/suite/cli_self_upd.rs
```diff
@@ -304,7 +304,38 @@ info: downloading self-update
...
+
+    update_setup(&|config, _| {
+        config.expect_ok(&["rustup-init", "-y", "--no-modify-path"]);
+
+        winreg::RegKey::predef(winreg::enums::HKEY_CURRENT_USER)
+            .create_subkey(RUSTUP_UNINSTALL_ENTRY)
+            .unwrap()
+            .0
+            .delete_value("DisplayVersion")
+            .unwrap();
```
rbtcollins: That deletes, not overwrites, thus it tests introducing, not replacing.

I suggest setting the value to a an old release of rustup rather than deleting

rami3l: @rbtcollins You are right and I intended to do so, it's just that this test modifies a global state (the winreg). This hasn't been done before, so I'm not sure what is considered proper use of winreg in the tests... Leaving a (realistic enough) placeholder in an intermediate state seems weird to me.

Ideally, if this test fails midway, the user should be able to immediately identify that `DisplayVersion` is a placeholder.

src/cli/self_update/windows.rs
rami3l: @rbtcollins That's the original error message. But I agree we can change it, ideally in another commit than the refactoring.

rami3l: Done.

tests/suite/cli_self_upd.rs
rami3l: I've chosen `9.999.99`. Hopefully it's clear enough that this is not a real version number.

djc (approved):

rami3l (commented):
I think it'll still be better to let @rbtcollins decide whether my response is satisfying and whether this PR is ready to be merged :)

tests/suite/cli_self_upd.rs
rbtcollins: @rami3l oh indeed; we have isolation for such tests already.

Probably this test needs to be reworked - sorry, brain not fully switched on. see https://github.com/rust-lang/rustup/blob/master/src/cli/self_update/test.rs#L38 - thats the core helper, and

https://github.com/search?q=repo%3Arust-lang%2Frustup%20with_saved_path&type=code

which shows it being used in both unit tests and integration tests.

I have no strong view about whether your choice of an integration test or unit is better - though a considerable preference for unit testing rather than integration tests.

Anyhow, what is needed is to extend that helper, or write a similar one, to manage the application uninstall data, and then use that in the test

rami3l: > I have no strong view about whether your choice of an integration test or unit is better - though a considerable preference for unit testing rather than integration tests.

@rbtcollins Regarding this particular case I do think integration test is better since we're ensuring the resolution of that issue (i.e. the correct behavior is exposed to the end user). The current `self update` implementation is a bit convoluted and it took me quite a while to figure out where exactly to add the code modifying `DisplayVersion`... 

> Anyhow, what is needed is to extend that helper, or write a similar one, to manage the application uninstall data, and then use that in the test

Will do.

rami3l: @rbtcollins Done.

src/cli/self_update/test.rs
```diff
@@ -1,57 +1,85 @@
...
+#[cfg(windows)]
+fn restore_path(p: Option<RegValue>) {
+    restore_reg_value(&RegKey::predef(HKEY_CURRENT_USER), "Environment", "PATH", p)
+}
+
 #[cfg(unix)]
 fn restore_path(_: Option<()>) {}
+
+#[cfg(windows)]
+pub fn with_saved_programs_display_version(f: &mut dyn FnMut()) {
```
rami3l: Looking back, it could be that this is overly specific for `rustup`'s public API.

We could possibly expose `with_saved_reg_value(root, subkey, name, callback)` instead, and `with_saved_programs_display_version(callback)` can then be easily defined on the tests side by pinning these arguments.

src/cli/self_update/test.rs
```diff
@@ -1,57 +1,85 @@
...
+}
+
+#[cfg(windows)]
+pub fn get_path() -> io::Result<Option<RegValue>> {
+    get_reg_value(&RegKey::predef(HKEY_CURRENT_USER), "Environment", "PATH")
+}
+
 #[cfg(unix)]
-pub fn get_path() -> std::io::Result<Option<()>> {
+pub fn get_path() -> io::Result<Option<()>> {
```
rami3l: Also, this does nothing on Unix according to `rust-analyzer`: it's not even referenced once. It's probably better to make it Windows-exclusive.

src/cli/self_update/test.rs
rbtcollins: This is in the test API anyway, so I don't see any issue.

PR #3774, opened 2024-04-14
mark few deps as optional, which is used only from some features

klensy:


djc (approved):

djc:
Makes sense, thanks! Have you looked at whether there is CI coverage of the various possible feature sets?

klensy:
No, i didn't checked CI, only manually grepped sources for usage and tried to build with curl/reqwest backends.

rami3l (approved):
@djc `cargo-all-features` in our CI seems to be working.

Thanks a lot @klensy!

djc:
Yeah, I was wondering about the `--no-default-features` case.

rami3l:
> Yeah, I was wondering about the `--no-default-features` case.

@djc It would surprise me if `cargo-all-features` considers default features at all.

Turns out that's indeed not the case:
https://github.com/frewsxcv/cargo-all-features/blob/fe252f463d6cf2969bf7ce5f911cc6222bc80c9d/src/test_runner.rs#L25

djc:
I meant I was wondering if we have CI coverage of compiling with no default features.

djc:
And the answer is no, because:

```rust
#[cfg(all(not(feature = "reqwest-backend"), not(feature = "curl-backend")))]
compile_error!("Must enable at least one backend");
```

rami3l:
> And the answer is no, because:
> 
> ```rust
> #[cfg(all(not(feature = "reqwest-backend"), not(feature = "curl-backend")))]
> compile_error!("Must enable at least one backend");
> ```

@djc That case has been explicitly excluded to prevent CI failure:

https://github.com/rust-lang/rustup/blob/f26d16b3b940bfb7b5e5aad7ee8f7121569f1d60/Cargo.toml#L193-L195

Apart from that, our `cargo-all-features` CI does test all possible feature combinations since it always adds `--no-default-features` to the call, as mentioned in https://github.com/rust-lang/rustup/pull/3774#issuecomment-2054024219.

PS: You have probably misunderstood me: when I say `cargo-all-features` I mean the test plugin to ensure all feature combinations compile, rather than the `cargo` flag named `--all-features`.

Issue #3651, opened 2024-01-27
`rustup set default-host` does not change the default toolchain


jyn514:
### Problem you are trying to solve

i have a `stable-x86_64-pc-windows-gnu` toolchain as my default:
```
PS C:\Users\jyn\src\example> rustup default
stable-x86_64-pc-windows-gnu (default)
```
i want it to be `-msvc` instead so i can use windbg (i don't have gdb installed). i ran `rustup set default-host x86_64-pc-windows-msvc`, which succeeded, but did not install any toolchains or change the default:
```
PS C:\Users\jyn\src\example> rustup show
Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\jyn\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-gnu (default)
nightly-x86_64-pc-windows-gnu
stage1

active toolchain
----------------

stable-x86_64-pc-windows-gnu (default)
rustc 1.75.0 (82e1608df 2023-12-21)
```
then i tried `rustup update stable`, which *did* install the -msvc toolchain, but did not change the default:
```
info: syncing channel updates for 'stable-x86_64-pc-windows-msvc'
...
PS C:\Users\jyn\src\example> rustup default
stable-x86_64-pc-windows-gnu (default)
```

### Solution you'd like

`rustup set default-host foo` should behave the same as `rustup set default-host foo && rustup default $channel-foo`, where `$channel` is the channel of the current default toolchain.

ideally, rustup would give a warning if the default toolchain did not match the default host (e.g. `rustup default nightly-x86_64-pc-windows-gnu` with a default-host of `msvc`).

### Notes

rustup 1.26.0 (5af9b9484 2023-04-05)

rami3l:
Following the long-term goal of eliminating implicit installation (https://github.com/rust-lang/rustup/issues/3635), I think the best thing to do here is to add a warning.

jyn514:
@rami3l i still think this suggestion from the description is the best approach:

> `rustup set default-host foo` should behave the same as `rustup set default-host foo && rustup default $channel-foo`, where $channel is the channel of the current default toolchain.

if `rustup default nightly` for a non-installed toolchain warns instead of installing, that's fine. but i do not think rustup should leave the default toolchain so that it doesn't match the default host.

rami3l:
Hello @jyn514, sorry for the late reply but do you see #4945 as a potential solution to this issue?

[...]

PR #3778, opened 2024-04-19
Update windows.md

Womoemy:
Modified the command to choose the 64-bit GNU toolchain as the linker. The previous one didn't work as intended and returned an error.

### DIFF
--- a/doc/user-guide/src/installation/windows.md
+++ b/doc/user-guide/src/installation/windows.md
@@ -34,7 +34,7 @@ $ rustup set default-host i686-pc-windows-msvc
 Or to choose the 64 bit GNU toolchain:
 
 ```console
-$ rustup set default-host x86_64-pc-windows-gnu
+$ rustup default stable-x86_64-pc-windows-gnu
 ```
 
 Since the MSVC ABI provides the best interoperation with other Windows


ehuss:
The original command looks correct to me. Can you show what error you got?

Womoemy:
Alright. Let me describe the error I faced in detail.
Following instructions in the Rust Book tutorial, I ran `rustc main.rs` to compile my program and it exited with this error ` rustc main.rs error: linking with link.exe failed: exit code: 1`.
A Google search of the error and I got results that linked back to the Rust Book docs suggesting I change the default toolchain,
then as stated in the docs I ran `rustup set default-host x86_64-pc-windows-gnu` and nothing happened then I tried to re-compile my program, I got this error message 
firsterror

So, I saw another result from Stack Overflow to try this other variation of the command `rustup default stable-x86_64-pc-windows-gnu` and viola 
firstsuccess
then I retried compiling my program and it compiled.
Hence I just thought to leave a PR.

P.S.: I appreciate your prompt feedback, it speaks much about the health of the project and the dedication of the team. Kudos!

rami3l:
@Womoemy Thanks for sending this PR! Looks like you're setting up Rust on Windows with MSVC.

On the one hand, the Rust Book is aware that its instructions for Windows/MSVC users are not clear enough, so recently a change has been made (https://github.com/rust-lang/book/pull/3616) to make them point to the relevant section of the [Rustup User Guide](https://rust-lang.github.io/rustup/installation/windows-msvc.html).

For the moment being, you can use the latter link for the MSVC setup instructions.

On the other hand, if you want to use Windows/GNU instead, `rustup set default-host x86_64-pc-windows-gnu`  _is_ the correct command to change the default host triple, there's nothing to be changed about that.

However, we do recognize that the step above is not enough to change the default toolchain being used (i.e. you need to run `rustup default stable` again), and that might cause confusion. https://github.com/rust-lang/rustup/issues/3651 has been tracking this very issue, and we expect to address it soon.

Given the two points above, I'm closing this PR now. Thanks again for offering your help to improve the Rustup project, and have a nice day!

Issue #3689, opened 2024-03-01
Rustup cannot download hash file for toolchains channel information under Wine


llde:
### Problem

Hi, When trying to install a MSVC toolchain using Rustup under wine (x64 and x86 rustup versions for windows affected) it fails with this error:
```console
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to 'C:\users\lorenzo\.rustup\tmp\scoucda6ta418hod_file': failed to make network request: error sending request for url (https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256): operation timed out: operation timed out
```

I did see this error seems to happen also on different platforms, but it's consistent on Wine. It's possible that it's not a specific wine deficiency, but worse performance under wine (socket performances are not on par with windows, expecially asyncronous and non blocking ones) are making the problem manifest.
I can download toolchains from the native linux rustup, however it's easier to be able to manage the msvc rust toolchain inside the Wine Prefix where Visual Studio 2019 and the C++ Build tools are installed

RUSTUP_USE_CURL=1 is a workaround, and allow rustup to properly download 

Thanks for the attention

### Steps

1) export WINEPREFIX="(PAth for VS wine prefix)"
2) wine rustup-init.exe
3) Press enter and wait

### Possible Solution(s)

RUSTUP_USE_CURL=1 is a workaround

### Notes

Wine, Wine-Staging and Wine-TKG affected. Problem existed in wine 7.x

### Rustup version

```console
1.26.0  (but this problem existed in all previous version for some years)
```

### Installed toolchains

```console
None
```

djc:
Reported this as a reqwest issue: https://github.com/seanmonstar/reqwest/issues/2158.

llde:
RUSTUP_USE_RUSTLS=1 seems to be another workaround with last rustup version. Without any envvars setted problems still persist
@djc

rami3l:
> RUSTUP_USE_RUSTLS=1 seems to be another workaround with last rustup version. Without any envvars setted problems still persist

@llde That kinda makes sense since the current default is to use Windows' own TLS stack, and I have no idea if that works well under Wine or not...

rami3l:
@djc I think since we've been using rustls by default for a while, plus the positive report regaring its use in Wine such as https://github.com/rust-lang/rustup/issues/4748 in rustup v1.28, we can safely close this issue as stale.

Issue #3391, opened 2023-06-24
Can't install, fails with: "internal error inside Hyper and/or its dependencies, please report"


mgalgs:
### Problem

I'm trying to install Rust using the recommended `rustup` command:

    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

but it's failing like so (with `--verbose`):

```
info: profile set to 'default'
info: default host triple is x86_64-unknown-linux-gnu
verbose: installing toolchain 'stable-x86_64-unknown-linux-gnu'
verbose: toolchain directory: '/home/mgalgs/.rustup/toolchains/stable-x86_64-unknown-linux-gnu'
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
verbose: creating temp file: /home/mgalgs/.rustup/tmp/z0knt4q4ahiof37m_file
verbose: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
verbose: downloading with reqwest
verbose: removing toolchain directory: '/home/mgalgs/.rustup/toolchains/stable-x86_64-unknown-linux-gnu'
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/home/mgalgs/.rustup/tmp/z0knt4q4ahiof37m_file'

Caused by:
    0: failed to make network request
    1: error sending request for url (https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256): internal error inside Hyper and/or its dependencies, please report
    2: internal error inside Hyper and/or its dependencies, please report
```

I've also tried forcing the `curl` fallback download helper as recommended in #1328 (by setting `RUSTUP_USE_CURL=1`), but it fails with a different error:

```
info: profile set to 'default'
info: default host triple is x86_64-unknown-linux-gnu
verbose: installing toolchain 'stable-x86_64-unknown-linux-gnu'
verbose: toolchain directory: '/home/mgalgs/.rustup/toolchains/stable-x86_64-unknown-linux-gnu'
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
verbose: creating temp file: /home/mgalgs/.rustup/tmp/0hk5cenyeo3kf8o5_file
verbose: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
verbose: downloading with curl
verbose: manifest not found. trying legacy manifest
[...]

Issue #3334, opened 2023-04-25
reqwest http handler causes silent failure on AAarch64 Windows


naixx:
### Problem

When I run `rustup-init.exe` it doesn't download any toolchains.

### Steps

1. `rustup default stable`
```
C:\Users\a\.cargo\bin>rustup default stable
info: syncing channel updates for 'stable-x86_64-pc-windows-msvc'

C:\Users\a\.cargo\bin>
```
2. No output on explicit install
```
C:\Users\a\.cargo\bin>rustup toolchain install stable
info: syncing channel updates for 'stable-x86_64-pc-windows-msvc'

C:\Users\a\.cargo\bin>
```
3. No downloads in logging
```
C:\Users\a\.cargo\bin>rustup --verbose update stable
verbose: read metadata version: '12'
verbose: installing toolchain 'stable-x86_64-pc-windows-msvc'
verbose: toolchain directory: 'C:\Users\a\.rustup\toolchains\stable-x86_64-pc-windows-msvc'
info: syncing channel updates for 'stable-x86_64-pc-windows-msvc'
verbose: creating temp file: C:\Users\astra\.rustup\tmp\cizkbcsni3xp7r1c_file
verbose: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
verbose: downloading with reqwest

C:\Users\a\.cargo\bin>
```

### Possible Solution(s)

Kinnison from discord suggested to use `RUSTUP_USE_CURL=1`
And that helped
His comments are:
```
OK that's interesting, it implies something is wrong with how reqwest is looking for stuff on your system and thus it fails
(a) it shouldn't silently fail
and (b) eventually RUSTUP_USE_CURL is going away
```

### Rustup version

```console
1.26
```

### Installed toolchains

```console
no toolchains
```

Alovchin91:
Thanks for the workaround @naixx @kinnison! I've just faced it on Windows aarch64. I'm gonna take a look if I can nail the issue down.

djc:
@Alovchin91 @naixx would either you be game to try to chase this down? The way I would do it is just by injecting a bunch of `dbg!()` spew in the download code (`printf`-style debugging) (starting from [download_to_path_with_backend()](https://github.com/rust-lang/rustup/blob/master/download/src/lib.rs#L52C22-L52C22), but I don't have any AArch64 Windows machines available to me. If it helps, I can make a branch with a bunch of spew on it and then you can compile + run it to see what happens?

[...]

Issue #3122, opened 2022-12-09
rustup-init hangs in armv7 docker container running on an arm64 Linux with `reqwest` backend


messense:
### Problem

Running `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh` in armv7 docker container running on an arm64 Linux hangs.

### Steps

Run a docker container with `--platform linux/arm/v7` on an arm64 Linux, install curl and run the sh.rustup.sh script

```bash
$ docker run --rm -it --platform linux/arm/v7 ubuntu:22.04 bash
root@668fa2822782:/# apt update && apt install curl -y
root@668fa2822782:/# curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --verbose
info: downloading installer
info: profile set to 'default'
info: default host triple is aarch64-unknown-linux-gnu
verbose: installing toolchain 'stable-aarch64-unknown-linux-gnu'
verbose: toolchain directory: '/root/.rustup/toolchains/stable-aarch64-unknown-linux-gnu'
info: syncing channel updates for 'stable-aarch64-unknown-linux-gnu'
verbose: creating temp file: /root/.rustup/tmp/dqlfqns1_reba5ms_file
verbose: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
verbose: downloading with reqwest
```

hangs forever at the `verbose: downloading with reqwest` step, use `RUSTUP_USE_CURL=1` works fine

```bash
root@668fa2822782:/# RUSTUP_USE_CURL=1 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | RUSTUP_USE_CURL=1 sh -s -- -y --verbose
info: downloading installer
info: profile set to 'default'
info: default host triple is aarch64-unknown-linux-gnu
verbose: installing toolchain 'stable-aarch64-unknown-linux-gnu'
verbose: toolchain directory: '/root/.rustup/toolchains/stable-aarch64-unknown-linux-gnu'
info: syncing channel updates for 'stable-aarch64-unknown-linux-gnu'
verbose: creating temp file: /root/.rustup/tmp/0lfud7winasgeiob_file
verbose: downloading file from: 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
verbose: downloading with curl
verbose: deleted temp file: /root/.rustup/tmp/0lfud7winasgeiob_file
[...]

PR #3788, opened 2024-04-25
refactor(download)!: remove curl backend

rami3l:
Part of https://github.com/rust-lang/rustup/issues/3790.

Revives #2034, superseding #1657.

(Thanks @lnicola!)

### Switching to the curl backend via `RUSTUP_USE_CURL` will no longer be a workaround

- https://github.com/rust-lang/rustup/issues/3122 (incomplete)
- https://github.com/rust-lang/rustup/issues/3334
- https://github.com/rust-lang/rustup/issues/3689 (works with `reqwest/rustls`)
- https://github.com/rust-lang/rustup/issues/3391 (incomplete, upstream)
- https://github.com/rust-lang/rustup/issues/1856 (certificate)


djc:
Have you taken stock of open issues referencing being solved by reverting to the curl backend?

Have you considered the alternative of using curl with the rustls TLS backend?

(Note: I am very much in favor of ditch curl and OpenSSL in favor of maximally-Rust solutions, but want to make sure we minimize the risk of getting users stuck.)

rami3l:
> Have you taken stock of open issues referencing being solved by reverting to the curl backend?

@djc I've added a new section to track this.

> Have you considered the alternative of using curl with the rustls TLS backend?

Hmmm, with this option we are basically introducing `curl/rustls` to replace `curl/openssl`, and eliminating the notion of TLS backend (so no more `reqwest/openssl`) in our repo, right?

First, I'm not sure if this would cause more problems or not, since we're introducing something new. In other words, can these existing issues be worked around with `curl/rustls`?
~~Also, I'm not quite familiar with the building process, will we be shipping `rustls` twice with `curl/rustls` + `reqwest/rustls` with this approach?~~ Looks like we won't.

But indeed, this would allow the removal of `openssl` to happen sooner. I guess I can try both...

PS: The 1.28.0 milestone is indicative, it only means I think this should be considered in this release cycle.

djc:
So as I understand it:

* The default currently is to use reqwest with native-tls (OpenSSL on Linux, native stacks on macOS and Windows)
* If `RUSTUP_USE_CURL` is set, we use the curl backend, which defaults to openssl
* If `RUSTUP_USE_RUSTLS` is set, we use reqwest with rustls

There is no way of knowing whether the curl + rustls option would solve some of the issues that people have seen with using reqwest.

Perhaps we should compile or release builds with tracing support built in so that people can give us more detailed logging without compiling their own rustup.

At least for some of the issues, there's the suggestion that native-tls might actually be at fault, such that reqwest + rustls might actually do better than the default reqwest + native-tls option.

I would advocate making the reqwest + rustls option the default first and shipping that -- I have high trust in the rustls maintainers (there are two full-time maintainers and myself) so I think any rustls issuers could be resolved pretty quickly. The one thing I would like is to ship support for rustls-platform-verifier in reqwest.

rami3l:
> Perhaps we should compile or release builds with tracing support built in so that people can give us more detailed logging without compiling their own rustup.

@djc I guess that's a plus, yeah.

> At least for some of the issues, there's the suggestion that native-tls might actually be at fault, such that reqwest + rustls might actually do better than the default reqwest + native-tls option.
> 
> I would advocate making the reqwest + rustls option the default first and shipping that -- I have high trust in the rustls maintainers (there are two full-time maintainers and myself) so I think any rustls issuers could be resolved pretty quickly. The one thing I would like is to ship support for rustls-platform-verifier in reqwest.

Do you think now is the right time to try out `rustls` even without `rustls-platform-verifier`?

djc:
> > At least for some of the issues, there's the suggestion that native-tls might actually be at fault, such that reqwest + rustls might actually do better than the default reqwest + native-tls option.
> > I would advocate making the reqwest + rustls option the default first and shipping that -- I have high trust in the rustls maintainers (there are two full-time maintainers and myself) so I think any rustls issuers could be resolved pretty quickly. The one thing I would like is to ship support for rustls-platform-verifier in reqwest.
> 
> Do you think now is the right time to try out `rustls` even without `rustls-platform-verifier`?

Yes, I think it would make sense to flip the default for the reqwest backend to rustls, but in that case I feel like we should keep the curl backend around for now as a backup option. And should definitely include a good tracing setup so that we can ask folks for more detailed information if issues occur.

rami3l:
> Yes, I think it would make sense to flip the default for the reqwest backend to rustls, but in that case I feel like we should keep the curl backend around for now as a backup option. And should definitely include a good tracing setup so that we can ask folks for more detailed information if issues occur.

@djc In that case I guess we can still keep this PR as a potential second step. Just to be sure, should I remove `reqwest/openssl` as well while changing `curl/openssl` to `curl/rustls` in that first step?

djc:
> > Yes, I think it would make sense to flip the default for the reqwest backend to rustls, but in that case I feel like we should keep the curl backend around for now as a backup option. And should definitely include a good tracing setup so that we can ask folks for more detailed information if issues occur.
> 
> @djc In that case I guess we can still keep this PR as a potential second step. Just to be sure, should I remove `reqwest/openssl` as well while changing `curl/openssl` to `curl/rustls` in that first step?

Yes, I think this could be a good second step. To be clear, I don't think we should make the curl + openssl -> curl + rustls change at the same time as promoting the reqwest + rustls backend to be the primary, since curl + openssl is currently a reliable fallback for scenarios where the default doesn't work, so I think we should keep it for a while longer. So I think the steps are:

1. Include tracing support with tracing-subscriber (I don't think we need opentelemetry stuff?) by default, make sure we don't constrain the `trace!()` level statically (using tracing Cargo features)
2. Switch the default from reqwest + native-tls to reqwest + rustls, keeping the curl + openssl fallback for now
.. put out a release and wait a few months for feedback
3. Try again if we can ditch curl + openssl (and reqwest + native-tls)

djc:
So does this mean we're branching off 1.30? Or do we think we can do this even in 1.29.1?

djc:
(I forget, do we have a warning now when curl is enabled via an environment variable?)

ChrisDenton:
I definitely think this should constitute a big change. People had been using it as a fallback if they were having issues. Hopefully that's not a problem any longer but still.

ChrisDenton:
@rami3l you may need to remove bcrypt from here if it's no longer used: https://github.com/rust-lang/rustup/blob/bd69b96d9f6cc112d386b58f4d784fb60d1469f5/build.rs#L69

rami3l:
> So does this mean we're branching off 1.30? Or do we think we can do this even in 1.29.1?

@djc As I said I will pursue https://github.com/rust-lang/rustup/issues/4738 so this will be in v1.30, and I think the best way to land the workflow change is to actually make a breaking change :) I don't like repairing a vessel on the sea but I guess we have to do exactly that sooner or later regarding this...

That said I think it's a good idea to first make the backport branch, then bump the version, and finally merge this PR. This is also why I'm keeping it as a draft for now.

> do we have a warning now when curl is enabled via an environment variable?

It's been enabled since a long time ago. https://github.com/rust-lang/rustup/commit/3cd156c463414a77f3b19f767ac671684a4a97cb was merged and shipped in May 2025 via v1.28.2. v1.30 will happen in 2027 if we go on like that XD

ChrisDenton:
Is the intent to go ahead with this for the next release or hold off awhile longer?

rami3l:
> Is the intent to go ahead with this for the next release or hold off awhile longer?

@chrisdenton I want to merge it soon and, thanks to our new release flow, ship it next year. So the actual intent is to do both really XD

djc:
So it feels like you've spun up an entire backporting workflow when there really wasn't a need to, and now we'll need to backport every fix, instead of holding off on merging this PR some more.

rami3l:
> So it feels like you've spun up an entire backporting workflow when there really wasn't a need to, and now we'll need to backport every fix, instead of holding off on merging this PR some more.

@djc Thanks for the question! The original intention of my introducing the backport flow is not to backport everything we have in the trunk branch; rather it is to enable releasing of patch releases when large refactoring is underway, such as during GSoC. I was reminded of the Windows MSVC breakage that @chrisdenton had to reiterate in multiple issues among other things.

For this patch in particular, as you said, if we were to introduce any major changes in the download logic, I can always wait and rebase this PR later, but in the meantime we have to tell everyone who contributed "don't need to add new features to the curl backend" which is not trivial. This has also happened to @FranciscoTGouveia when he was adding the download retry feature (#4695); I'd argue that it's a non-trivial task either way.

rami3l:
I think this PR is now ready to be reviewed.

I'll handle the backport-related grind caused by it which hopefully won't be a lot... The release/1.29 branch still retains the old CI config so will continue to serve us for the v1.29.x releases to come.

djc (approved):

rami3l:
It'd be better if I add `!` marks here to prevent this from being backported by mistake. I'll re-enqueue this when that is done...

PR #3795, opened 2024-04-27
refactor(download): use `DownloadCallBack` in `download_with_backend()`

rami3l:
A trivial patch cherry-picked from #3788.

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -35,19 +35,19 @@ pub enum Event<'a> {
     DownloadDataReceived(&'a [u8]),
 }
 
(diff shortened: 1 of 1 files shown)


djc (approved):
Seems okay, honestly I don't think type aliases like this are useful, so I'd prefer to just get rid of it.

rami3l:
> Seems okay, honestly I don't think type aliases like this are useful, so I'd prefer to just get rid of it.

@djc This type was introduced to address [clippy::type_complexity](https://rust-lang.github.io/rust-clippy/master/#/type_complexity) in #3133. I don't think it's worth it to override the lint level for this type 🤔

Issue #3350, opened 2023-05-06
shouldn't install script be quiet?


vic1707:
### Problem you are trying to solve

Hello,
I'm currently working on some installation scripts and I'd like to install rust somewhat silently (or at least restrict the output to only errors).
I found the `-q|--quiet` option in the script but it doesn't affect the script itself, only `rustup` as far as I understand.
```bash
# curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -q -y       
info: downloading installer
info: profile set to 'default'
info: default host triple is aarch64-unknown-linux-gnu
info: syncing channel updates for 'stable-aarch64-unknown-linux-gnu'
info: latest update on 2023-04-20, rust version 1.69.0 (84c898d65 2023-04-16)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
info: default toolchain set to 'stable-aarch64-unknown-linux-gnu'

  stable-aarch64-unknown-linux-gnu installed - rustc 1.69.0 (84c898d65 2023-04-16)


Rust is installed now. Great!

To get started you may need to restart your current shell.
This would reload your PATH environment variable to include
Cargo's bin directory ($HOME/.cargo/bin).

To configure your current shell, run:
source "$HOME/.cargo/env"
```
All `printf` and `echo` lines in the script are postfixed with `1>&2` which redirect `stdout` to `stderr` even though the script has dedicated `say` and `err` function. 

By doing so we are completely unable to only allow errors to be printed out and the `-q|--quiet` doesn't do what its name suggests.
We are also unable to bypass this restriction by setting `1>/dev/null` ourselves.

### Solution you'd like

I think the best solution would be apply the `-q|--quiet` flag to the install script.
Another solution would be to let `stdout` be what it is instead of redirecting it to `stderr` (removing most of the `1>&2` inside the script).
At least I'd like to see some sort of warning/message in the `--help` section of the script.

### Notes

English isn't my primary language, I'm sorry if I made mistakes or if anything sounds bad.

```[tasklist]
### Tasks
- [ ] #3910
- [ ] https://github.com/rust-lang/rustup/issues/3900
```

[...]

PR #3800, opened 2024-05-01
fix(dist/linux): handle the possible unavailability of `/proc` in `rustup-init.sh`

### LINKED ISSUE #2700: `rustup-init.sh` fails to detect platform correctly under `docker buildx` which lacks `/proc`
rustup-init.sh installs the incorrect rustc and other binaries because of a failure to detect arch.
Problems: `/proc/self/exe` does not exist during docker build, so i686/386 etc is detected incorrectly as x86_64 due to failure on line 153. mips64 likely suffers the same issue because it also uses get_bitness

`grep '^Features' /proc/cpuinfo | grep -q -v neon` fails and ARMv6 is incorrectly detected as arm7 on line 367
Running the downloaded binary fails with `/lib/ld-linux-armhf.so.3: No such file or directory` (because it isnt armhf, it is armel at `/lib/ld-linux.so.3` )

Logs and example code to produce the container and error:
https://gist.github.com/miigotu/2a0b80677420d806c96d8e792ae6652e

Note: gcc inside the container reports the correct info, kernel reports x86_64
```bash
root@386c88edfbc5:/# uname -m
x86_64
root@386c88edfbc5:/# gcc -dumpmachine | sed "s/-/-$(uname -p)-/"
i686-unknown-linux-gnu
```
and
```bash
root@dac44b74e4c0:/# uname -m
armv7l
root@dac44b74e4c0:/# gcc -dumpmachine | sed "s/-/-$(uname -p)-/"
arm-unknown-linux-gnueabi
```

rami3l:
Closes https://github.com/rust-lang/rustup/issues/2700. I'm no shell expert, just trying to do my best...

This fix is based on @miigotu and @kerberjg's find in https://github.com/rust-lang/rustup/issues/2700#issuecomment-1266454666 and https://github.com/rust-lang/rustup/issues/2700#issuecomment-1367488985 respectively. It first refers to `/proc/self/exe` if it's accessible, then `$SHELL` if it's set, and finally `/bin/sh`. This executable is then explicitly passed to functions that previously relied on `/proc/self/exe`.

> 有備無患 yǒu bèi wú huàn
> - to be prepared against adverse events

## Concerns

- [x] ~~What about the `grep '^Features' /proc/cpuinfo` line which also relies on `/proc`?~~ Fixed.
    - Currently, if `grep` fails (because e.g. `/proc/cpuinfo` is unavailable or malformed), we still assume `armv7`. This doesn't work in the case described in #2700, where we're using ARMv6. OTOH, falling back to `arm` is much safer in this case and will resolve the aforementioned issue.
- [x] How to test this patch?
    - ~~The `podman` instance I use seems to have `/proc` access.~~ I've tested the patch with a modified `/proc` path (`/fakeproc`) and the logic seems to work.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -539,8 +539,9 @@ get_architecture() {
     # and fall back to arm.
     # See https://github.com/rust-lang/rustup.rs/issues/587.
     if [ "$_ostype" = "unknown-linux-gnueabihf" ] && [ "$_cputype" = armv7 ]; then
-        if ensure grep '^Features' /proc/cpuinfo | grep -E -q -v 'neon|simd'; then
-            # At least one processor does not have NEON (which is asimd on armv8+).
+        if ! (ensure grep '^Features' /proc/cpuinfo | grep -E -q 'neon|simd') ; then
+            # Either `/proc/cpuinfo` is malformed or unavailable, or
+            # at least one processor does not have NEON (which is asimd on armv8+).
             _cputype=arm
         fi
     fi
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -177,12 +177,23 @@ main() {
     return "$_retval"
 }
 
-check_proc() {
-    # Check for /proc by looking for the /proc/self/exe link
+get_current_exe() {
+    # Returns the executable used for system architecture detection
     # This is only run on Linux
-    if ! test -L /proc/self/exe ; then
-        err "fatal: Unable to find /proc/self/exe.  Is /proc mounted?  Installation cannot proceed without /proc."
+    local _current_exe
+    if test -L /proc/self/exe ; then
+        _current_exe=/proc/self/exe
+    else
+        printf '%s\n' "Warning: Unable to find /proc/self/exe. System architecture detection might be inaccurate." 1>&2
+        if test -n "$SHELL" ; then
+            _current_exe=$SHELL
+        else
+            need_cmd /bin/sh
+            _current_exe=/bin/sh
+        fi
+        printf '%s\n' "Warning: Falling back to $_current_exe." 1>&2
     fi
+    echo "$_current_exe"
 }
 
 get_bitness() {
@@ -193,8 +204,9 @@ get_bitness() {
     #   0x02 for 64-bit.
     # The printf builtin on some shells like dash only supports octal
     # escape sequences, so we use those.
+    local _current_exe=$1
     local _current_exe_head
-    _current_exe_head=$(head -c 5 /proc/self/exe )
+    _current_exe_head=$(head -c 5 "$_current_exe")
     if [ "$_current_exe_head" = "$(printf '\177ELF\001')" ]; then
         echo 32
     elif [ "$_current_exe_head" = "$(printf '\177ELF\002')" ]; then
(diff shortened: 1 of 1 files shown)


rustup-init.sh
```diff
@@ -540,7 +540,7 @@ get_architecture() {
     # and fall back to arm.
     # See https://github.com/rust-lang/rustup.rs/issues/587.
     if [ "$_ostype" = "unknown-linux-gnueabihf" ] && [ "$_cputype" = armv7 ]; then
-        if ensure grep '^Features' /proc/cpuinfo | grep -E -q -v 'neon|simd'; then
```
rami3l, line 526: Currently, if `grep` fails (because e.g. /proc/cpuinfo is unavailable or malformed), we still assume `armv7`. This doesn't work with the case described in https://github.com/rust-lang/rustup/issues/2700#issue-841531263 when we're using ARMv6.

rustup-init.sh
```diff
@@ -356,16 +371,18 @@ get_architecture() {
...
     case "$_ostype" in
 
         Android)
             _ostype=linux-android
             ;;
 
         Linux)
-            check_proc
+            get_current_exe
+            _current_exe=$RETVAL
```
rbtcollins: Any reason not to use

`_current_exe=get_current_exe`

rbtcollins:
I think this looks reasonable.

There are other probes we could use, such as uname and lsb_release and gcc but they all have edge cases and caveats. One question for you then GTG IMO.

rustup-init.sh
rami3l: @rbtcollins I assume you meant `_current_exe=$(get_current_exe)`?

The `echo "Warning: ..."` calls in the function body will interfere with the function's output when it's being used like so (unless the `1>&2` trick is used, as I understand it). Looking at the surrounding code, it seems like the `$RETVAL` trick is used exactly in this case to specify the actual value to be returned.

I referred to `downloader()` as an example when writing this. This is how it calls other functions via `$RETVAL`:

https://github.com/rust-lang/rustup/blob/2a5a69e0914ff419554d684ca71eb1d72c72bcb3/rustup-init.sh#L596-L599

This is how it issues warnings:

https://github.com/rust-lang/rustup/blob/2a5a69e0914ff419554d684ca71eb1d72c72bcb3/rustup-init.sh#L604

Maybe we should use `1>&2` for `echo` commands instead? But in that case we might hit #3350, right?

rustup-init.sh
```diff
@@ -177,12 +177,23 @@ main() {
...
+get_current_exe() {
+    # Returns the executable used for system architecture detection
     # This is only run on Linux
-    if ! test -L /proc/self/exe ; then
-        err "fatal: Unable to find /proc/self/exe.  Is /proc mounted?  Installation cannot proceed without /proc."
+    local _current_exe
+    if test -L /proc/self/exe ; then
+        _current_exe=/proc/self/exe
+    else
+        echo "Warning: Unable to find /proc/self/exe.  System architecture detection might be inaccurate."
```
rami3l: Oops, a tiny typo here...

```suggestion
        echo "Warning: Unable to find /proc/self/exe. System architecture detection might be inaccurate."
```

rustup-init.sh
rami3l: @rbtcollins I just had another look of #3350, and I believe I should switch to `1>&2` for this function.

See https://github.com/rust-lang/rustup/issues/3350#issuecomment-2158213483 for my proposal on the direction Rustup should be moving towards in general WRT stdout/stderr handling.

rami3l: @rbtcollins Done.

rustup-init.sh
rami3l: Done.

rbtcollins (approved):

Issue #3790, opened 2024-04-26
Tracking: Simplify download and/or TLS backends


rami3l:
Part of https://github.com/rust-lang/rustup/issues/1611.

## Motivation

The direct motivation of this change is to simplify Rustup's building process in order to make Rustup more cross-build friendly and prevent surprise `stable`-only CI build errors like the ones we've encountered while bumping OpenSSL to v3, i.e. https://github.com/rust-lang/rustup/pull/3478, https://github.com/rust-lang/rustup/pull/3668, https://github.com/rust-lang/rustup/issues/4707, etc.

## Status quo

> * The default currently is to use reqwest with native-tls (OpenSSL on Linux, native stacks on macOS and Windows)
> * If `RUSTUP_USE_CURL` is set, we use the curl backend, which defaults to openssl
> * If `RUSTUP_USE_RUSTLS` is set, we use reqwest with rustls

_https://github.com/rust-lang/rustup/pull/3788#issuecomment-2076819462_

## Goal

The eventual goal would be removing `RUSTUP_USE_CURL` and/or `RUSTUP_USE_RUSTLS`, leaving a single stack "in favor of maximally-Rust solutions", while "making sure we minimize the risk of getting users stuck".

To minimize frictions during this change, some necessary refactoring patches will be developed in the meantime, with a primary focus on improving Rustup's observability by interfacing with the `tracing` ecosystem.

_https://github.com/rust-lang/rustup/pull/3788#issuecomment-2076679922_

### Tasks

- https://github.com/rust-lang/rustup/pull/3793
- https://github.com/rust-lang/rustup/pull/3868
- https://github.com/rust-lang/rustup/pull/3803
- https://github.com/rust-lang/rustup/pull/3879
- https://github.com/rust-lang/rustup/pull/3873
- https://github.com/rust-lang/rustup/pull/3871
- https://github.com/rust-lang/rustup/pull/3877
- https://github.com/rust-lang/rustup/pull/3891
- https://github.com/rust-lang/rustup/pull/3798
- https://github.com/rust-lang/rustup/pull/3903
- https://github.com/rust-lang/rustup/pull/3815
- https://github.com/rust-lang/rustup/pull/3788
- #4277

djc:
Also, many benefits from rustls just being more secure by default than OpenSSL.

PR #3803, opened 2024-05-02
chore(dist/features): ship `tracing` and friends by default

rami3l:
> [!WARNING]  
> Since https://github.com/rust-lang/rustup/pull/3876, `RUST_LOG` has been replaced with **`RUSTUP_LOG`** for configuring Rustup's internal logging system. The rest of the thread has used `RUST_LOG` due to historic reasons.

Part of #3790.

## Rationale

Currently, helping out the Rustup team by enabling local tracing is quite a tedious process (esp. for community contributors), requiring rebuilding Rustup from the exact commit with an extra feature, `otel`:

https://github.com/rust-lang/rustup/blob/54dd3d00fd20e64dc522517a0d7be4285570a2f1/doc/dev-guide/src/tracing.md?plain=1#L13-L36

After some experiment, it turned out that we actually can ship the tracing features by default without forcing the user to face OTEL connection errors on a daily basis.

To clarify, this *does not* mean Rustup is setting up a central (a.k.a. phone-home-style) telemetry mechanism, and we will keep the tracing disabled by default unless `RUST_LOG` has been explicitly set.

## Concerns

- [x] Should we eliminate `RUSTUP_DEBUG` in favor of `RUST_LOG=trace`? (Yes.)
- [x] ~~Should we remove `opentelemetry` while keeping `tracing` (https://github.com/rust-lang/rustup/pull/3788#issuecomment-2079165721)? If we should, the `otel` feature should be renamed.~~ (Not yet, see https://github.com/rust-lang/rustup/pull/3803#issuecomment-2092349333.)

### DIFF
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -1,94 +1,148 @@
 use std::{fmt, io::Write};
 
+use termcolor::{Color, ColorSpec, WriteColor};
+use tracing::{level_filters::LevelFilter, Event, Subscriber};
+use tracing_subscriber::{
+    fmt::{
+        format::{self, FormatEvent, FormatFields},
+        FmtContext,
+    },
+    registry::LookupSpan,
+    EnvFilter, Layer,
+};
+
 #[cfg(feature = "otel")]
 use once_cell::sync::Lazy;
 #[cfg(feature = "otel")]
 use opentelemetry_sdk::trace::Tracer;
-#[cfg(feature = "otel")]
-use tracing::Subscriber;
-#[cfg(feature = "otel")]
-use tracing_subscriber::{registry::LookupSpan, EnvFilter, Layer};
 
-use crate::currentprocess::{process, terminalsource};
+use crate::{currentprocess::Process, utils::notify::NotificationLevel};
 
 macro_rules! warn {
-    ( $ ( $ arg : tt ) * ) => ( $crate::cli::log::warn_fmt ( format_args ! ( $ ( $ arg ) * ) ) )
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::warn ! ( $ ( $ arg ) * )  )
 }
+
 macro_rules! err {
-    ( $ ( $ arg : tt ) * ) => ( $crate::cli::log::err_fmt ( format_args ! ( $ ( $ arg ) * ) ) )
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::error ! ( $ ( $ arg ) * )  )
 }
+
 macro_rules! info {
-    ( $ ( $ arg : tt ) * ) => ( $crate::cli::log::info_fmt ( format_args ! ( $ ( $ arg ) * ) ) )
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::info ! ( $ ( $ arg ) * )  )
 }
 
 macro_rules! verbose {
-    ( $ ( $ arg : tt ) * ) => ( $crate::cli::log::verbose_fmt ( format_args ! ( $ ( $ arg ) * ) ) )
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::debug ! ( $ ( $ arg ) * )  )
 }
 
 macro_rules! debug {
-    ( $ ( $ arg : tt ) * ) => ( $crate::cli::log::debug_fmt ( format_args ! ( $ ( $ arg ) * ) ) )
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::trace ! ( $ ( $ arg ) * )  )
 }
 
-pub(crate) fn warn_fmt(args: fmt::Arguments<'_>) {
-    let mut t = process().stderr().terminal();
-    let _ = t.fg(terminalsource::Color::Yellow);
-    let _ = t.attr(terminalsource::Attr::Bold);
-    let _ = write!(t.lock(), "warning: ");
-    let _ = t.reset();
-    let _ = t.lock().write_fmt(args);
-    let _ = writeln!(t.lock());
+pub fn tracing_subscriber(process: Process) -> impl tracing::Subscriber {
+    use tracing_subscriber::{layer::SubscriberExt, Registry};
+
+    #[cfg(feature = "otel")]
+    let telemetry = telemetry(&process);
+    let console_logger = console_logger(process);
+    #[cfg(feature = "otel")]
+    {
+        Registry::default().with(console_logger).with(telemetry)
+    }
+    #[cfg(not(feature = "otel"))]
+    {
+        Registry::default().with(console_logger)
+    }
 }
 
-pub(crate) fn err_fmt(args: fmt::Arguments<'_>) {
-    let mut t = process().stderr().terminal();
-    let _ = t.fg(terminalsource::Color::Red);
-    let _ = t.attr(terminalsource::Attr::Bold);
-    let _ = write!(t.lock(), "error: ");
-    let _ = t.reset();
-    let _ = t.lock().write_fmt(args);
-    let _ = writeln!(t.lock());
+/// A [`tracing::Subscriber`] [`Layer`][`tracing_subscriber::Layer`] that prints out the log
+/// lines to the current [`Process`]' `stderr`.
+///
+/// When the `RUST_LOG` environment variable is present, a standard [`tracing_subscriber`]
+/// formatter will be used according to the filtering directives set in its value.
+/// Otherwise, this logger will use [`EventFormatter`] to mimic "classic" Rustup `stderr` output.
+fn console_logger<S>(process: Process) -> impl Layer<S>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    let has_ansi = process.stderr().is_a_tty();
+    let maybe_rust_log_directives = process.var("RUST_LOG");
+    let logger = tracing_subscriber::fmt::layer()
+        .with_writer(move || process.stderr())
+        .with_ansi(has_ansi);
+    if let Ok(directives) = maybe_rust_log_directives {
+        let env_filter = EnvFilter::builder()
+            .with_default_directive(LevelFilter::INFO.into())
+            .parse_lossy(directives);
+        logger.compact().with_filter(env_filter).boxed()
+    } else {
+        // Receive log lines from Rustup only.
+        let env_filter = EnvFilter::new("rustup=DEBUG");
+        logger
+            .event_format(EventFormatter)
+            .with_filter(env_filter)
+            .boxed()
+    }
 }
 
-pub(crate) fn info_fmt(args: fmt::Arguments<'_>) {
-    let mut t = process().stderr().terminal();
-    let _ = t.attr(terminalsource::Attr::Bold);
-    let _ = write!(t.lock(), "info: ");
-    let _ = t.reset();
-    let _ = t.lock().write_fmt(args);
-    let _ = writeln!(t.lock());
+// Adapted from
+// https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.FormatEvent.html#examples
+struct EventFormatter;
+
+impl<S, N> FormatEvent<S, N> for EventFormatter
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+    N: for<'a> FormatFields<'a> + 'static,
+{
+    fn format_event(
+        &self,
+        ctx: &FmtContext<'_, S, N>,
+        mut writer: format::Writer<'_>,
+        event: &Event<'_>,
+    ) -> fmt::Result {
+        let has_ansi = writer.has_ansi_escapes();
+        let level = NotificationLevel::from(*event.metadata().level());
+        {
+            let mut buf = termcolor::Buffer::ansi();
+            if has_ansi {
+                _ = buf.set_color(ColorSpec::new().set_bold(true).set_fg(level.fg_color()));
+            }
+            _ = write!(buf, "{level}: ");
+            if has_ansi {
+                _ = buf.reset();
+            }
+            writer.write_str(std::str::from_utf8(buf.as_slice()).unwrap())?;
+        }
+        ctx.field_format().format_fields(writer.by_ref(), event)?;
+        writeln!(writer)
+    }
 }
 
-pub(crate) fn verbose_fmt(args: fmt::Arguments<'_>) {
-    let mut t = process().stderr().terminal();
-    let _ = t.fg(terminalsource::Color::Magenta);
-    let _ = t.attr(terminalsource::Attr::Bold);
-    let _ = write!(t.lock(), "verbose: ");
-    let _ = t.reset();
-    let _ = t.lock().write_fmt(args);
-    let _ = writeln!(t.lock());
-}
-
-pub(crate) fn debug_fmt(args: fmt::Arguments<'_>) {
-    if process().var("RUSTUP_DEBUG").is_ok() {
-        let mut t = process().stderr().terminal();
-        let _ = t.fg(terminalsource::Color::Blue);
-        let _ = t.attr(terminalsource::Attr::Bold);
-        let _ = write!(t.lock(), "debug: ");
-        let _ = t.reset();
-        let _ = t.lock().write_fmt(args);
-        let _ = writeln!(t.lock());
+impl NotificationLevel {
+    fn fg_color(&self) -> Option<Color> {
+        match self {
+            NotificationLevel::Debug => Some(Color::Blue),
+            NotificationLevel::Verbose => Some(Color::Magenta),
+            NotificationLevel::Info => None,
+            NotificationLevel::Warn => Some(Color::Yellow),
+            NotificationLevel::Error => Some(Color::Red),
+        }
     }
 }
 
 /// A [`tracing::Subscriber`] [`Layer`][`tracing_subscriber::Layer`] that corresponds to Rustup's
 /// optional `opentelemetry` (a.k.a. `otel`) feature.
 #[cfg(feature = "otel")]
-pub fn telemetry<S>() -> impl Layer<S>
+fn telemetry<S>(process: &Process) -> impl Layer<S>
 where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
-    // NOTE: This reads from the real environment variables instead of `process().var_os()`.
-    let env_filter = EnvFilter::try_from_default_env().unwrap_or(EnvFilter::new("INFO"));
+    let env_filter = if let Ok(directives) = process.var("RUST_LOG") {
+        EnvFilter::builder()
+            .with_default_directive(LevelFilter::TRACE.into())
+            .parse_lossy(directives)
+    } else {
+        EnvFilter::new("rustup=TRACE")
+    };
     tracing_opentelemetry::layer()
         .with_tracer(TELEMETRY_DEFAULT_TRACER.clone())
         .with_filter(env_filter)
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -18,24 +18,24 @@ use opentelemetry_sdk::trace::Tracer;
 
 use crate::{currentprocess::Process, utils::notify::NotificationLevel};
 
-macro_rules! warn {
-    ( $ ( $ arg : tt ) * ) => ( ::tracing::warn ! ( $ ( $ arg ) * )  )
-}
-
-macro_rules! err {
-    ( $ ( $ arg : tt ) * ) => ( ::tracing::error ! ( $ ( $ arg ) * )  )
-}
-
-macro_rules! info {
-    ( $ ( $ arg : tt ) * ) => ( ::tracing::info ! ( $ ( $ arg ) * )  )
+macro_rules! debug {
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::trace ! ( $ ( $ arg ) * )  )
 }
 
 macro_rules! verbose {
     ( $ ( $ arg : tt ) * ) => ( ::tracing::debug ! ( $ ( $ arg ) * )  )
 }
 
-macro_rules! debug {
-    ( $ ( $ arg : tt ) * ) => ( ::tracing::trace ! ( $ ( $ arg ) * )  )
+macro_rules! info {
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::info ! ( $ ( $ arg ) * )  )
+}
+
+macro_rules! warn {
+    ( $ ( $ arg : tt ) * ) => ( ::tracing::warn ! ( $ ( $ arg ) * )  )
+}
+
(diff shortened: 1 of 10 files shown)


rami3l:
I think the artifact size changes are acceptable even if we keep `opentelemetry`:

### BEFORE

image

### AFTER

image

djc:
I think we should use a basic console-based tracing-subscriber setup:

```rust
pub(super) fn subscribe() -> tracing::subscriber::DefaultGuard {
    let sub = tracing_subscriber::FmtSubscriber::builder()
        .with_max_level(tracing::Level::TRACE)
        .with_writer(|| TestWriter)
        .finish();
    tracing::subscriber::set_default(sub)
}

struct TestWriter;

impl Write for TestWriter {
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        print!(
            "{}",
            str::from_utf8(buf).expect("tried to log invalid UTF-8")
        );
        Ok(buf.len())
    }
    fn flush(&mut self) -> io::Result<()> {
        io::stdout().flush()
    }
}
```

This is what I've been using in Quinn for many years. Every test just starts by calling `let _guard = subscribe();`, which has been a highly effective tool for test observability.

If we do this we can massively simplify the scaffolding for traces/logging:

- Get rid of all the opentelemetry dependencies, which I don't think we need
- Remove the use of the `otel` feature
- Remove the custom `test` macro

rbtcollins:
> Currently, helping out the Rustup team by enabling local tracing is quite a tedious process (esp. for community contributors), requiring rebuilding Rustup from the exact commit with an extra feature, otel:

I don't think this is true. We've not asked people to build with otel enabled that I'm remembering. The OS level traces we use to debug fundamental problems are from strace / truss and other similar tools. Otel / tracing! is not deployed widely enough within rustup to be a replacement for such things.

rbtcollins:
>  Get rid of all the opentelemetry dependencies, which I don't think we need
>  Remove the use of the otel feature
>  Remove the custom test macro

Please don't - while the OS level debugging is vital, for doing investigations on performance, having a nice report with spans and the detailed call tree is very useful, and since we configure it off by default it has very little overhead to maintain or build with. Really only the all-features-build test matters.

djc:
> > Get rid of all the opentelemetry dependencies, which I don't think we need
> > Remove the use of the otel feature
> > Remove the custom test macro
> 
> Please don't - while the OS level debugging is vital, for doing investigations on performance, having a nice report with spans and the detailed call tree is very useful, and since we configure it off by default it has very little overhead to maintain or build with. Really only the all-features-build test matters.

How many times have you used it in the past year? IMO while custom test macro + opentelemetry dependencies may not impose run-time overhead for downstream users, it does impose significant maintenance overhead that may not be warranted for the additional insight compared to just tracing-subscriber built-in (and maybe tokio-console level output).

djc:
> > Currently, helping out the Rustup team by enabling local tracing is quite a tedious process (esp. for community contributors), requiring rebuilding Rustup from the exact commit with an extra feature, otel:
> 
> I don't think this is true. We've not asked people to build with otel enabled that I'm remembering. The OS level traces we use to debug fundamental problems are from strace / truss and other similar tools. Otel / tracing! is not deployed widely enough within rustup to be a replacement for such things.

IMO the important point is that we should have a user-facing solution like "enable RUST_LOG=trace and give us the output of that", which seems like a decent method of getting better insight into problems that happen only in specific environments, which seems to be an important source of issues for rustup.

rami3l:
I just checked and it looks like tokio-console doesn’t currently have a timeline view (https://github.com/tokio-rs/console/issues/129), so I imagine `opentelemetry` and `jaeger` are here to stay for longer...

For now I plan to:
- Ship `tracing` by default with a console-based subscriber.
- If possible, reimplement our current logging system using that subscriber (with a `fmt::layer()` that mimics the original output style).
- Keep the `opentelemetry` related stuff behind the `otel` feature.

More specifically, I imagine having multiple subscribers (https://github.com/tokio-rs/tracing/issues/971) based on env vars and features:

- A "classic" subscriber that targets `process().stderr()`, has a classic output format, will only print `rustup` log lines up to a certain level (`info`), and will be disabled if `RUST_LOG` is set.
- A "tracing" subscriber that also targets `process().stderr()` and is not limited to `rustup` (so we could have `tonic` log lines as well, for example). Its activation is mutually exclusive with the "classic" subscriber, and the precise logging level will be controlled by `RUST_LOG`.
- An OpenTelemetry subscriber (could be replaced by `tokio-console` in the future, but not just yet) available behind the `otel` feature.


Finally:

- [x] We need to make sure this subscriber's use of CLI colors is coherent with the current system (incl. env variable controls via `RUSTUP_TERM_COLOR`, etc).
- [x] With the consistent use of `RUST_LOG`, `RUSTUP_DEBUG` should be retired accordingly.
- [x] Make `tracing::instrument` always run, i.e. no longer behind `#[cfg_attr(feature = "otel", ..)]` (split into #3873).

~~Waiting for #3367 might be worthwhile, since this will change the startup process.~~ Merged.

djc:
#3367 has been merged, would be good to rebase this!

rami3l:
The CI now works, but I have no idea how to properly inject the `otel` subscriber into tests, because it's a global state, but the new `stderr` subscriber is local (depends on `process()`), and there's seemingly no way for them to live together properly according to my knowledge of the `tracing` APIs.

@rbtcollins @djc Am I missing something, or does the current architecture need a greater change (e.g. even more asynchronization) to allow this to happen? I'm a bit confused...

PS: It should be noted that everything works perfectly fine if we only need the `stderr` logger but not the `otel` logger for tests.

## Details

The status quo is that the only subscriber config we have (i.e. the `otel` one) remains the same across all tests, so we're using a `static` item, and we register it with `tracing::subscriber::set_global_default()`:

https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/src/test.rs#L235-L237
https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/src/test.rs#L275-L276
https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/src/test.rs#L280-L308

However now the new `stderr` logger (and thus the entire subscriber layer) relies on `process()` in order to attach to `stderr` properly, so the old approach won't just work.

To make things worse, in `before_test()` and `before_test_async()` the `process()` context is not yet available, so one possibility would be re-injecting subscriber(s) locally (instead of globally like before) within functions like `with()` and `with_runtime()`. This unfortunately will make the previously-decoupled `currentprocess` and `test` modules couple:

https://github.com/rust-lang/rustup/blob/6d54875d2c8acc7c52727c18feea16f396a1f1b7/src/currentprocess.rs#L147-L150
https://github.com/rust-lang/rustup/blob/6d54875d2c8acc7c52727c18feea16f396a1f1b7/src/currentprocess.rs#L252

This is probably not ideal for two reasons:
- `with*()` look like universal functions that are used not only for tests (are they?), meaning I cannot depend on `static TRACER` in this new implementation. However, we must still be able to flush the `TRACER` at the end of the function when we are actually in the tests.
    https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/src/test.rs#L294-L308
    However, if we do intend to use `with*` for tests only, then exposing and using `TRACER` within them don't look like that much of an issue. 
- 3ba08da98221b2941aeb858e54cebc059859ffb7 has experimented with the `stderr` logger and everything works. However once the local `otel` logger comes into play I start to get errors within `with()` that indicates the lack of a Tokio context:
    ```console
    thread 'suite::cli_exact::cross_install_indicates_target' panicked at /Users/rami3l/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tonic-0.11.0/src/transport/service/executor.rs:15:9:                                                                                                             
    there is no reactor running, must be called from the context of a Tokio 1.x runtime                                                                            
    ```
    
I wish I could have a way of appending to rather than [overriding](https://docs.rs/tracing/latest/tracing/instrument/trait.WithSubscriber.html#method.with_subscriber) the current subscriber.

rbtcollins:
> How many times have you used it in the past year? IMO while custom test macro + opentelemetry dependencies may not impose run-time overhead for downstream users, it does impose significant maintenance overhead that may not be warranted for the additional insight compared to just tracing-subscriber built-in (and maybe tokio-console level output).

I have been largely absent for the last year, so perhaps a better thing to ask is 'of the times you've been investigating performance, how often have you used detailed call graph with opentelemetry' - and the answer there is every single time. I depend on it.

> IMO the important point is that we should have a user-facing solution like "enable RUST_LOG=trace and give us the output of that", which seems like a decent method of getting better insight into problems that happen only in specific environments, which seems to be an important source of issues for rustup.

I think that that is important too, for sure.

rami3l:
Regarding https://github.com/rust-lang/rustup/pull/3803#issuecomment-2143703392, I think a reasonable solution would be making all unit tests run under `#[tokio::test]`, so that we can stick to a new local subscriber all the time.

This prevents https://docs.rs/opentelemetry-otlp/latest/opentelemetry_otlp/struct.OtlpTracePipeline.html#method.install_batch from complaining about not being executed in a `tokio` context, and allows the removal of static `TRACER` and `TRACE_RUNTIME` altogether. The perf penalty here should be minimal.

I'll split that change into a preparatory PR (#3868).

rami3l:
@djc The new approach proposed in https://github.com/rust-lang/rustup/pull/3871#pullrequestreview-2115593544 has worked! This PR has been thus unblocked from test refactoring and I'll just need to do some rather simple tweaks (see https://github.com/rust-lang/rustup/pull/3803#issuecomment-2092349333 for the list of remaining tasks) to get it ready 🎉 

PS: I've also changed the signature of `tracing_subscriber()` to accept a `Process` in order to align with #3871 better.

rami3l:
@rbtcollins @djc I've just finished all 3 cleanup tasks, unfortunately however the 3rd (the final) commit (https://github.com/rust-lang/rustup/pull/3803/commits/0cf0c6bdd35f9bf8076d78627eab0ee6f01e3a88) has made our CI red. To make sure that only this commit was bad though, I've reverted it temporarily in https://github.com/rust-lang/rustup/pull/3803/commits/1b7662e9890524ab3a795696aee226eea4a499d2, and indeed the CI is green again.

I don't see how a mechanical change like https://github.com/rust-lang/rustup/pull/3803/commits/0cf0c6bdd35f9bf8076d78627eab0ee6f01e3a88 can cause Rustup to stack overflow on Windows (https://github.com/rust-lang/rustup/actions/runs/9515433133/job/26229657328)... Am I missing something?

djc:
Maybe try a branch where you apply https://github.com/rust-lang/rustup/commit/0cf0c6bdd35f9bf8076d78627eab0ee6f01e3a88 to 1.27.1 directly (or some other commit maybe from before #3868) to see if that works? It might be some "interesting" interaction that causes recursion in an unobvious way?

rami3l:
> Maybe try a branch where you apply [0cf0c6b](https://github.com/rust-lang/rustup/commit/0cf0c6bdd35f9bf8076d78627eab0ee6f01e3a88) to 1.27.1 directly (or some other commit maybe from before #3868) to see if that works? It might be some "interesting" interaction that causes recursion in an unobvious way?

@djc That's a good idea! I'll give it a try tomorrow.

That being said, the overall `tracing` thing is working pretty well, so it might make some sense to leave https://github.com/rust-lang/rustup/commit/0cf0c6bdd35f9bf8076d78627eab0ee6f01e3a88 for another PR. We'll have to test it against another branch via a PR anyways.

djc:
> That being said, the overall `tracing` thing is working pretty well, so it might make some sense to leave [0cf0c6b](https://github.com/rust-lang/rustup/commit/0cf0c6bdd35f9bf8076d78627eab0ee6f01e3a88) for another PR.

Oh yes, I'm in favor of merging without it!

src/cli/log.rs
```diff
@@ -1,94 +1,149 @@
...
-use tracing::Subscriber;
-#[cfg(feature = "otel")]
-use tracing_subscriber::{registry::LookupSpan, EnvFilter, Layer};
 
-use crate::currentprocess::{process, terminalsource};
+use crate::{currentprocess::Process, utils::notify::NotificationLevel};
 
-macro_rules! warn {
-    ( $ ( $ arg : tt ) * ) => ( $crate::cli::log::warn_fmt ( format_args ! ( $ ( $ arg ) * ) ) )
+macro_rules! debug {
```
djc, line 21: I'd like to keep these in their original order so that the diff more clearly shows what changed exactly.

src/cli/log.rs
```diff
@@ -1,94 +1,149 @@
...
+/// lines to the current [`Process`]' `stderr`.
+///
+/// When the `RUST_LOG` environment variable is present, a standard [`tracing_subscriber`]
+/// formatter will be used according to the filtering directives set in its value.
+/// Otherwise, this logger will use [`EventFormatter`] to mimic "classic" Rustup `stderr` output.
+fn console_logger<S>(process: Process) -> impl Layer<S>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    let maybe_rust_log_directives = process.var_os("RUST_LOG").clone();
```
djc: Don't think we need the `clone()` here?

Also this feels like it's doign a bunch of manual reimplementation of things that tracing-subscriber offer for us already -- not sure what the goal is here?

src/cli/log.rs
```diff
@@ -1,94 +1,149 @@
...
-    let mut t = process().stderr().terminal();
-    let _ = t.fg(terminalsource::Color::Magenta);
-    let _ = t.attr(terminalsource::Attr::Bold);
-    let _ = write!(t.lock(), "verbose: ");
-    let _ = t.reset();
-    let _ = t.lock().write_fmt(args);
-    let _ = writeln!(t.lock());
+// Adapted from
+// https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.FormatEvent.html#examples
+struct EventFormatter;
```
djc, line 95: Can we just use tracing-subscriber's default event formatter? Why not?

src/cli/log.rs
```diff
@@ -1,94 +1,149 @@
...
 /// optional `opentelemetry` (a.k.a. `otel`) feature.
 #[cfg(feature = "otel")]
-pub fn telemetry<S>() -> impl Layer<S>
+fn telemetry<S>(process: Process) -> impl Layer<S>
 where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
-    // NOTE: This reads from the real environment variables instead of `process().var_os()`.
-    let env_filter = EnvFilter::try_from_default_env().unwrap_or(EnvFilter::new("INFO"));
+    let maybe_rust_log_directives = process.var_os("RUST_LOG").clone();
```
djc: No need to `clone()`, right? Here, too, I wonder why we can't stick with what tracing-subscriber can provide us.

src/utils/notify.rs
```diff
@@ -1,8 +1,34 @@
+use std::fmt;
+
 #[derive(Debug)]
 pub(crate) enum NotificationLevel {
+    Debug,
```
djc, line 7: Let's not reorder these here?

src/utils/notify.rs
```diff
@@ -1,8 +1,34 @@
...
+            NotificationLevel::Warn => "warning",
+            NotificationLevel::Error => "error",
+        })
+    }
+}
+
+impl From<tracing::Level> for NotificationLevel {
+    fn from(level: tracing::Level) -> Self {
+        match level {
+            tracing::Level::TRACE => Self::Debug,
```
djc: I'd probably import `tracing::Level` in this module.

src/cli/log.rs
```diff
@@ -1,94 +1,149 @@
...
+            NotificationLevel::Error => Some(Color::Red),
+        }
     }
 }
 
 /// A [`tracing::Subscriber`] [`Layer`][`tracing_subscriber::Layer`] that corresponds to Rustup's
 /// optional `opentelemetry` (a.k.a. `otel`) feature.
 #[cfg(feature = "otel")]
-pub fn telemetry<S>() -> impl Layer<S>
+fn telemetry<S>(process: Process) -> impl Layer<S>
```
djc: I think this can take a `&Process`?

src/cli/log.rs
```diff
@@ -65,7 +65,16 @@ where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
     let maybe_rust_log_directives = process.var_os("RUST_LOG").clone();
-    let has_ansi = process.stderr().is_a_tty();
+    let has_ansi = {
```
djc: Can we avoid this nested scope in favor of just assigning `env_override` the value of the `match`?

src/cli/log.rs
```diff
@@ -65,7 +65,16 @@ where
...
+        let env_override = process.var("RUSTUP_TERM_COLOR").map(|it| it.to_lowercase());
+        match env_override.as_deref() {
```
djc: Instead of doing `to_lowercase()`, consider doing `Ok(s) if s.eq_ignore_ascii_case("always")`.

src/cli/log.rs
rami3l: @djc The goal here is to mirror the original Rustup output when `RUST_LOG` is unset, see https://github.com/rust-lang/rustup/pull/3803#issuecomment-2092349333. On the other hand, the default formatter from `tracing_subsriber` is used once `RUST_LOG` is set.

src/cli/log.rs
rami3l: @djc In fact, we are already using it. See https://github.com/rust-lang/rustup/pull/3803#discussion_r1640490409.

Apart from this particular point, I think your other points (like the one regarding unnecessary `.clone()`s) are pretty valid, and I'll address them shortly.

src/utils/notify.rs
rami3l: Hmmm, I guess the rationale here is that `Level` is ambiguous since there are essentially two level types and if I directly `use tracing::Level` it seems less clear... I've considered `use tracing::Level as TracingLevel` but this is not doing any good.

src/cli/log.rs
rami3l: > No need to clone(), right?

@djc That's probably what's left of a unsuccessful previous attempt. Fixed.

> I wonder why we can't stick with what tracing-subscriber can provide us.

This is because the default from `tracing_subscriber` actually doesn't consider our injected `env` and `stderr`. I've still used `tracing_subscriber`'s functionalities to a maximum though except for those two things.

Also, that's why I said we'd better remove injections altogether (`clap` also suffers from the same issue as it directly reads from `env` if we want its `env` integrations), but maybe not today...

PS: It was me who added the `NOTE` above to remind myself of that fact.

src/cli/log.rs
rami3l: @djc This is copied from somewhere else in the codebase (also written by myself, as I've probably missed this `eq_ignore_ascii_case` thing), so I guess it'd be reasonable to change that part of the code accordingly in a separate commit.

src/cli/log.rs
rami3l: @djc I've split the reordering in another commit. This helps parsing the different levels when I was writing the patch, since `tracing::Level` and `NotificationLevel` have different nomenclatures.

src/cli/log.rs
rami3l: @djc FWIW, the current implementation gives:

```console
> RUSTUP_FORCE_ARG0=rustup rustup run stable cargo run -- default stable
     Running `target/debug/rustup-init default stable`
info: using existing install for 'stable-aarch64-apple-darwin'
info: default toolchain set to 'stable-aarch64-apple-darwin'

  stable-aarch64-apple-darwin unchanged - rustc 1.79.0 (129f3b996 2024-06-10)

info: note that the toolchain 'stable-aarch64-apple-darwin' is currently in use (overriden by environment variable RUSTUP_TOOLCHAIN)
```

However, once we set `RUST_LOG`:

```console
> RUST_LOG=trace RUSTUP_FORCE_ARG0=rustup rustup run stable cargo run -- default stable
     Running `target/debug/rustup-init default stable`
2024-06-15T02:22:11.801871Z  INFO rustup::cli::common: using existing install for 'stable-aarch64-apple-darwin'
2024-06-15T02:22:11.814343Z  INFO rustup::cli::common: default toolchain set to 'stable-aarch64-apple-darwin'

  stable-aarch64-apple-darwin unchanged - rustc 1.79.0 (129f3b996 2024-06-10)

2024-06-15T02:22:11.822315Z  INFO rustup::cli::rustup_mode: note that the toolchain 'stable-aarch64-apple-darwin' is currently in use (overriden by environment variable RUSTUP_TOOLCHAIN)
```

rami3l: > Don't think we need the `clone()` here?

@djc Fixed.

> Also this feels like it's doign a bunch of manual reimplementation of things that tracing-subscriber offer for us already -- not sure what the goal is here?

The goal here is to mirror the original Rustup output when `RUST_LOG` is not set (see https://github.com/rust-lang/rustup/pull/3803#issuecomment-2092349333 for the original plan):

```console
> RUSTUP_FORCE_ARG0=rustup rustup run stable cargo run -- default stable
     Running `target/debug/rustup-init default stable`
info: using existing install for 'stable-aarch64-apple-darwin'
info: default toolchain set to 'stable-aarch64-apple-darwin'

  stable-aarch64-apple-darwin unchanged - rustc 1.79.0 (129f3b996 2024-06-10)

info: note that the toolchain 'stable-aarch64-apple-darwin' is currently in use (overriden by environment variable RUSTUP_TOOLCHAIN)
```

This means that I won't have to change our integration tests too much. Please note that there's no "refresh" mode for a whole bunch of handwritten tests, so keeping the old output is a very important aspect in the migration.

That said, I'm totally open to changing the output format if we get to migrated to another test framework such as `snapbox` that supports refreshing in the future (like in https://github.com/rust-lang/cargo/issues/14039).

On the other hand, once `RUST_LOG` is set, **the default formatter from `tracing_subsriber` is used** instead:

```console
> RUST_LOG=trace RUSTUP_FORCE_ARG0=rustup rustup run stable cargo run -- default stable
     Running `target/debug/rustup-init default stable`
2024-06-15T02:22:11.801871Z  INFO rustup::cli::common: using existing install for 'stable-aarch64-apple-darwin'
2024-06-15T02:22:11.814343Z  INFO rustup::cli::common: default toolchain set to 'stable-aarch64-apple-darwin'

  stable-aarch64-apple-darwin unchanged - rustc 1.79.0 (129f3b996 2024-06-10)

2024-06-15T02:22:11.822315Z  INFO rustup::cli::rustup_mode: note that the toolchain 'stable-aarch64-apple-darwin' is currently in use (overriden by environment variable RUSTUP_TOOLCHAIN)
```

I'll leave the relevant doc changes to another PR, so that we can work on it in parallel together with #3873 and #3871.

djc (approved):
Looks great!

PR #3810, opened 2024-05-07
fix(ci/freebsd): fix `invalid peer certificate: UnknownIssuer`

rami3l:
Caused by #3798, this is essentially https://github.com/astral-sh/uv/issues/3369. However, we didn't choose to switch to bundled CA certs, otherwise suddenly Rustup binaries will come with an expiration date :/

https://github.com/rust-lang/rustup/actions/runs/8978258543/job/24658381207 has passed after adding `pkg install -y ca_root_nss` to the setup.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1158,7 +1158,7 @@ jobs:
         usesh: true
         copyback: false
         prepare: |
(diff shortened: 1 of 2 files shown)


djc (approved):
Nice!

PR #3811, opened 2024-05-07
fix(filesource): make some constructs only available via the `test` feature

rami3l:
Cherry-picked from #3803.

This PR has fixed the following warnings in https://github.com/rust-lang/rustup/actions/runs/8978258543/job/24658381207:

```rs
 warning: struct `TestStdinLock` is never constructed
    --> src/currentprocess/filesource.rs:47:8
     |
  47 | struct TestStdinLock<'a> {
     |        ^^^^^^^^^^^^^
     |
     = note: `#[warn(dead_code)]` on by default
  
  warning: type alias `TestStdinInner` is never used
    --> src/currentprocess/filesource.rs:68:17
     |
  68 | pub(crate) type TestStdinInner = Arc<Mutex<Cursor<String>>>;
     |                 ^^^^^^^^^^^^^^
  
  warning: struct `TestStdin` is never constructed
    --> src/currentprocess/filesource.rs:70:8
     |
  70 | struct TestStdin(TestStdinInner);
     |        ^^^^^^^^^
  
  warning: struct `TestWriterLock` is never constructed
     --> src/currentprocess/filesource.rs:182:19
      |
  182 | pub(super) struct TestWriterLock<'a> {
      |                   ^^^^^^^^^^^^^^
  
  warning: `rustup` (lib) generated 4 warnings
```

### DIFF
--- a/src/currentprocess/filesource.rs
+++ b/src/currentprocess/filesource.rs
@@ -1,11 +1,14 @@
-use std::io::{self, BufRead, Cursor, Read, Result, Write};
-use std::sync::{Arc, Mutex, MutexGuard};
+use std::io::{self, BufRead, Read, Result, Write};
+#[cfg(feature = "test")]
+use std::{
+    io::Cursor,
+    sync::{Arc, Mutex, MutexGuard},
+};
 
 use enum_dispatch::enum_dispatch;
 
-use crate::currentprocess::process;
-
 use super::terminalsource::{ColorableTerminal, StreamSelector};
+use crate::currentprocess::process;
 
 /// Stand-in for std::io::Stdin
 pub trait Stdin {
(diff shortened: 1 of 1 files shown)


src/currentprocess/filesource.rs
```diff
@@ -1,11 +1,15 @@
...
+use std::io::{self, BufRead, Read, Result, Write};
 
 use enum_dispatch::enum_dispatch;
 
+use super::terminalsource::{ColorableTerminal, StreamSelector};
 use crate::currentprocess::process;
 
-use super::terminalsource::{ColorableTerminal, StreamSelector};
+#[cfg(feature = "test")]
+use std::{
```
djc, line 3: Prefer to keep this next to the other std imports (in the same block).

djc (approved):

src/currentprocess/filesource.rs
rami3l: Done. (It's not possible to put a `#[cfg()]` between commas, so that's the best I could do.)

PR #3813, opened 2024-05-07
Reduce code duplication for show commands

djc:
This stuck out at me during my review of #3596.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -21,6 +21,7 @@ use crate::currentprocess::{
     varsource::VarSource,
 };
 use crate::dist::dist::{TargetTriple, ToolchainDesc};
+use crate::dist::manifest::ComponentStatus;
 use crate::install::UpdateStatus;
 use crate::utils::notifications as util_notifications;
 use crate::utils::notify::NotificationLevel;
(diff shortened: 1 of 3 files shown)


rami3l:
@djc Thanks! I remember I wanted to do this after #3637 (where I introduced `.components()`), but I somehow forgot it 🤣

src/cli/common.rs
```diff
@@ -375,88 +376,55 @@ where
...
-            } else if component.available {
-                let _ = writeln!(t.lock(), "{target}");
-            }
-        }
-    }
-
-    Ok(utils::ExitCode(0))
+pub(crate) fn list_targets(
+    distributable: DistributableToolchain<'_>,
+    installed_only: bool,
```
rami3l, line 381: I don't know what you think in terms of API design, but I remember having read somewhere that Rust "hates boolean parameters", since you basically don't know what boolean literals mean in a function call.

Your particular uses of this function look perfectly fine, but considering possible future uses (esp. from external contributors), I would've refactored this into two `pub(crate) fn`s such as `list_all_targets()` and `list_installed_targets()` calling one private implementation that has that boolean parameter...

I'll let you decide on this one.

src/cli/common.rs
```diff
@@ -375,88 +376,55 @@ where
...
-                let _ = writeln!(t.lock(), "{target} (installed)");
-                let _ = t.reset();
-            } else if component.available {
-                let _ = writeln!(t.lock(), "{target}");
-            }
-        }
-    }
-
-    Ok(utils::ExitCode(0))
+pub(crate) fn list_targets(
```
rami3l, line 379: I've read somewhere that Rust "hates boolean parameters", as the readability of function calls would be problematic esp. when you pass in boolean literals.

Your particular use of this parameter (passing in named variables) here seems fine, but considering future contributions (esp. external ones) I would've refactored this into two functions calling one private impl passing `true` and `false` respectively.

That being said, I'll let you decide on this one. (I had the same concern regarding the APIs in #3596, but my justification was that a clap-dispatched function e.g. `fn doc()` is really only used for `rustup doc` and nothing else.)

rami3l (approved):
Thanks a lot!

src/cli/common.rs
djc: IMO the notion that Rust "hates boolean parameters" is pretty dogmatic/overblown. I agree that `bool` is not a great choice for widely used functions (especially in the public API), or if there are multiple `bool` arguments in the same function call, or if the role of the `bool` value is particularly sensitive (removing some code). In this case, none of those are relevant.

> Your particular use of this parameter (passing in named variables) here seems fine, but considering future contributions (esp. external ones) I would've refactored this into two functions calling one private impl passing `true` and `false` respectively.

To my eyes, this solution is worse than the problem. Function wrappers that add limited functionality to an inner function they call just make names verbose and obscure shared purpose.

As such, I think this usage is fine. In other cases, the right solution IME might be to have a custom enum -- for this case maybe something like `enum LocalState { Installed, NotInstalled }`, variants like `::Yes` or `True` are another anti-pattern, but that has a decent amount of "overhead" in terms of extra code needed so I don't think doing so here is worth it.

PR #3815, opened 2024-05-09
feat(download): reflect the download/TLS backends in the user agent

rami3l:
Part of #3790.

https://www.reddit.com/r/rust/comments/1cmzxeu/call_for_testing_rustups_reqwest_backend_with/l357qz1 suggests that we could make UAs more detailed to perform some sort of telemetry-based analysis on the download/TLS backend segmentation.

I just got the confirmation from @Mark-Simulacrum that they can obtain a certain amount of user agent strings for the analysis.

We should've done this sooner (ideally before v1.27.1), but since we've missed the window, the current plan would be looking at how many have switched to `RUSTUP_USE_RUSTLS=0` when it gets shipped.

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -12,7 +12,19 @@ pub use crate::errors::*;
 
 /// User agent header value for HTTP request.
 /// See: https://github.com/rust-lang/rustup/issues/2860.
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Nice!

PR #3817, opened 2024-05-09
add command suggestion

### LINKED ISSUE #3812: `rustup update self` should suggest `rustup self update`

A new-ish user got confused about the command for updating rustup itself and tried `rustup update self` which gave this error:

```
error: error: invalid value 'self' for '[toolchain]...': invalid toolchain name: 'self'

For more information, try '--help'.
: invalid toolchain name: 'self'
```

### Solution you'd like

Maybe something like:

```
note: use `rustup self update` to update rustup itself
```

### Notes

I'm also not sure why the ``: invalid toolchain name: 'self'`` is repeated at the end. That seems redundant.
```[tasklist]
### Tasks
- [ ] https://github.com/rust-lang/rustup/issues/3905
```

fbrv:
Closes #3812.

### DIFF
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -141,11 +141,11 @@ fn handle_suggestion() -> Result<()> {
         "note: use `rustup self update` to update rustup itself",
     );
 
-    let args: Vec<String> = std::env::args().into_iter().skip(1).collect();
+    let args: Vec<String> = std::env::args().skip(1).collect();
     for i in 0..args.len() {
         for j in i + 1..=args.len() {
             let slice = &args[i..j];
-            let slice = slice.iter().cloned().collect::<Vec<String>>();
+            let slice = slice.to_vec();
             if let Some(&message) = suggestions.get(&slice) {
                 writeln!(process().stdout().lock(), "\n{}", message)?;
                 break;
--- a/src/cli/setup_mode.rs
+++ b/src/cli/setup_mode.rs
@@ -142,15 +142,9 @@ fn handle_suggestion() -> Result<()> {
     );
 
     let args: Vec<String> = std::env::args().skip(1).collect();
-    for i in 0..args.len() {
-        for j in i + 1..=args.len() {
-            let slice = &args[i..j];
-            let slice = slice.to_vec();
-            if let Some(&message) = suggestions.get(&slice) {
-                writeln!(process().stdout().lock(), "\n{}", message)?;
-                break;
-            }
-        }
+    if let Some(&message) = suggestions.get(&args) {
+        writeln!(process().stdout().lock(), "\n{}", message)?;
+        return Ok(());
     }
 
     Ok(())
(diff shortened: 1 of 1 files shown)


fbrv:
@rami3l that's the overall idea, still not super sure about the suggestion logic. At first I was thinking to pattern matching only for a specific use case but then changed to possibly extend to how many case are necessary by inserting a new entry inside the hashmap, what do you think?

src/cli/setup_mode.rs
```diff
@@ -97,7 +100,11 @@ pub fn main() -> Result<utils::ExitCode> {
             write!(process().stdout().lock(), "{e}")?;
             return Ok(utils::ExitCode(0));
         }
-        Err(e) => return Err(e.into()),
+        Err(e) => {
```
rami3l, line 103: I believe that's not the right place to add the handling logic for `rustup update`. Have you actually tested the code that you added?

Please have a look at the `rustup update` handling logic [here](https://github.com/rust-lang/rustup/blob/66b7a442c074c5cd597ea1436e1bafb0807b46e0/src/cli/rustup_mode.rs#L930), and find out where that error was thrown, catch it, print another message using the `info!()` macro if that is the expected type of error, and finally re-throw it (you may adjust the ordering of the error message and the new one if that makes the implementation simpler).

After that, please add a regression test in an appropriate spot (like [here](https://github.com/rust-lang/rustup/blob/66b7a442c074c5cd597ea1436e1bafb0807b46e0/tests/suite/cli_rustup.rs#L103-L124)) to ensure the resolution of the issue.

src/cli/setup_mode.rs
```diff
@@ -125,3 +132,20 @@ pub fn main() -> Result<utils::ExitCode> {
 
     self_update::install(no_prompt, verbose, quiet, opts)
 }
+
+fn handle_suggestion() -> Result<()> {
+    let mut suggestions = HashMap::new();
```
rami3l, line 137: IMHO using a `HashMap` has added no value to the current implementation. There's no need to "plan ahead" in this particular situation. If a HashMap turns out to be required, it can be added in another patch in the future, but not right now.

src/cli/setup_mode.rs
```diff
@@ -125,3 +132,20 @@ pub fn main() -> Result<utils::ExitCode> {
...
+
+    suggestions.insert(
+        vec![String::from("update"), String::from("self")],
+        "note: use `rustup self update` to update rustup itself",
+    );
+
+    let args: Vec<String> = std::env::args().skip(1).collect();
```
rami3l, line 144: Using `std::env::args()` is not appropriate here, see https://github.com/rust-lang/rustup/pull/3817#discussion_r1596154685. Instead, you should extract all necessary info from the `m: &ArgMatches` parameter.

src/cli/setup_mode.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
```
rami3l, line 1: Finally, at Rustup it is recommended to use atomic commits with informative commit messages, please have a look at the examples on the `master` branch. This means that you'll have to do `git rebase -i` for quite a bunch of times to get the right commits (and the right messages).

PR #3826, opened 2024-05-14
Omit "ETA 0s" output in download tracker

barafael:
When running rustup, the ETA is shown and continuously updated. After a download has finished, ETA is 0, and it is printed for each download (marked **bold**)

  8.0 MiB /   8.0 MiB (100 %)   5.0 MiB/s in  1s **ETA:  0s**
 15.1 MiB /  15.1 MiB (100 %)   5.3 MiB/s in  2s **ETA:  0s**
 24.3 MiB /  24.3 MiB (100 %)   5.3 MiB/s in  4s **ETA:  0s**

This change omits printing the ETA when it is 0s:

  8.0 MiB /   8.0 MiB (100 %)   5.5 MiB/s in  1s         
 15.1 MiB /  15.1 MiB (100 %)   6.0 MiB/s in  2s         
 24.3 MiB /  24.3 MiB (100 %)   5.6 MiB/s in  4s

Not a big change :) lmk what you think :)

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -186,13 +186,16 @@ impl DownloadTracker {
                             (remaining / speed) as u64
                         });
                         format!(
-                            "{} / {} ({:3.0} %) {} in {} ETA: {}",
+                            "{} / {} ({:3.0} %) {} in {}{}",
                             total_h,
                             content_len_h,
                             percent,
                             speed_h,
                             elapsed_h.display(),
-                            eta_h.display(),
+                            match eta_h {
+                                Duration::ZERO => format_args!(""),
+                                _ => format_args!(" ETA: {}", eta_h.display()),
+                            }
                         )
                     }
                     None => format!(


src/cli/download_tracker.rs
```diff
@@ -186,13 +186,15 @@ impl DownloadTracker {
...
+                            "{} / {} ({:3.0} %) {} in {}{}",
                             total_h,
                             content_len_h,
                             percent,
                             speed_h,
                             elapsed_h.display(),
-                            eta_h.display(),
+                            (eta_h != Duration::ZERO)
+                                .then(|| format!(" ETA: {}", eta_h.display()))
+                                .unwrap_or_default()
```
djc: I think I'd prefer to formulate it like this:

```rust
match eta_h {
     Duration::ZERO => format_args!(""),
     _ => format_args!(" ETA: {}", eta_h.display()),
}
```

djc (commented):
Good idea, suggested a slightly different way of spelling it.

src/cli/download_tracker.rs
barafael: hmm, that doesn't work because the `format_args!` creates a temporary which doesn't survive the match scope

djc:
That didn't exactly work, maybe bind `let eta_format = eta_h.display();` before the outer `format!()` call?

barafael:
I think format_args only works as direct argument to `format!`. I can't get it to work in any other way at least. But yes, the original double `format!` is also not ideal

djc:
Fixed in #3827.

PR #3832, opened 2024-05-16
refactor(filesource): replace repetitive `#[cfg()]` usages with `cfg_if!{}`

rami3l:
It occurred to me right after #3811 that we actually have `cfg_if` in our imports, so using it more should provide better ergonomics and prevent issues like #3811 happen again (i.e. "I forgot to add `#[cfg()]` to some constructs but not others").

I plan to adjust #3803 to use `cfg_if!{}` in `rustup-init.rs` as well, where we also tend to use quite a lot of conditionally compiled constructs.

### DIFF
--- a/src/currentprocess/filesource.rs
+++ b/src/currentprocess/filesource.rs
@@ -5,6 +5,7 @@ use std::{
     sync::{Arc, Mutex, MutexGuard},
 };
 
+use cfg_if::cfg_if;
 use enum_dispatch::enum_dispatch;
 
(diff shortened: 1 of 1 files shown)


djc:
A simpler way of doing this would be to just move all the `test`-gated implementations into an inline module which is guarded with `#[cfg(feature = "test")]` and a single `use` statement that exports from the inline module. I think I'd prefer that over the `cfg_if!{}` approach, which is less standard and noisier.

PR #3833, opened 2024-05-16
refactor(filesource): replace repetitive `#[cfg()]` usages with a new `mod`

rami3l:
Another solution to #3832's issue, as suggested in https://github.com/rust-lang/rustup/pull/3832#issuecomment-2114247663.

Supersedes #3832.

### DIFF
--- a/src/currentprocess/filesource.rs
+++ b/src/currentprocess/filesource.rs
@@ -1,9 +1,4 @@
 use std::io::{self, BufRead, Read, Result, Write};
-#[cfg(feature = "test")]
-use std::{
-    io::Cursor,
-    sync::{Arc, Mutex, MutexGuard},
-};
 
 use enum_dispatch::enum_dispatch;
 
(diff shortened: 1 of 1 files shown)


src/currentprocess/filesource.rs
```diff
@@ -1,12 +1,9 @@
...
-#[cfg(feature = "test")]
-use std::{
-    io::Cursor,
-    sync::{Arc, Mutex, MutexGuard},
-};
 
 use enum_dispatch::enum_dispatch;
 
+#[cfg(feature = "test")]
+pub(crate) use self::tests::*;
```
djc: I would put this near the `tests` module definition (probably underneath).

src/currentprocess/filesource.rs
```diff
@@ -187,81 +132,126 @@ impl StderrSource for super::OSProcess {
     }
 }
 
-// ----------------------- test support for writers ------------------
-
 #[cfg(feature = "test")]
-pub(super) struct TestWriterLock<'a> {
-    inner: MutexGuard<'a, Vec<u8>>,
-}
+mod tests {
```
djc: I'd prefer another name, since `tests` is the convention that's used for actual tests. Maybe `test_support` or `test_sources`?

djc (approved):

src/currentprocess/filesource.rs
rami3l: Done.

src/currentprocess/filesource.rs
rami3l: Done.

djc (approved):

PR #3834, opened 2024-05-16
Tweaks from derived clap parser review

djc:
Follow-up from #3596. I think none of this is really things that originated in that PR, just trying to improve the codebase as I notice stuff.


src/cli/rustup_mode.rs
```diff
@@ -1515,12 +1518,6 @@ fn set_auto_self_update(cfg: &mut Cfg, auto_self_update_mode: &str) -> Result<ut
     Ok(utils::ExitCode(0))
 }
 
-#[cfg_attr(feature = "otel", tracing::instrument(skip_all))]
```
rami3l, line 1540: A question for @rbtcollins: when do we decide that we need to instrument a function?

rami3l (approved):
Thank you so much for the work! I do think this has addressed many concerns I have had in terms of code readability :)

PS: There seems to be some comptime errors to be addressed, but this doesn't influence my approval.

Issue #3514, opened 2023-10-22
Make Rustup install a non-emulated toolchain in an emulated environment


rami3l:
As a continuation of #3068:

> Is it possible to teach 'rustup self update' to install the correct non-emulated binary (even if there isn't a new rustup version to update to)?
> 
> I think we'd also need to change the default target in the rustup config, since that will be intel rather than aarch64 too.
> 
> I'm not sure UX wise whether doing it silently is sensible. I worry about IDE integrations.
> 
> What about this as a proposal:
> 
> 0) we add some docs about this
> 1) add a config option to disable emulation detection (default is enabled)
> 2) if emulation detection is enabled:
>   a) all proxies stop working and output an error directing folk to our docs
>   b) rustup self update with no options is unchanged
>   c) `rustup self update --to-native`, a new option to self update, will:
>       1) if the default host triple matches the emulation, rewrite it to be the native triple
>       2) if the default toolchain, and any override toolchains, specify the emulated triple, rewrite them to match the native triple
>       3) download the native rustup-init and install it into place
> 
> I'm not suggesting deleting toolchains because that gets into rapidly mounting complexity: cross.rs for instance wants lots of non-native toolchains present, linked toolchains can be wrong too and we don't have their arch etc.
> 
> Changing the defaults and overrides will cause implicit installation of the rewritten toolchain specifications automatically at next use.

_Originally posted by @rbtcollins in https://github.com/rust-lang/rustup/issues/3068#issuecomment-1474436068_

PR #3840, opened 2024-05-23
Finalise Rustup Windows Arm64 support

Alovchin91:
This PR finalises Windows Arm64 support by doing the following:

* ~~Removes the workaround to exclude Ring on Windows Arm64. Ring 0.17.x fully supports it.~~ Split-off into #3845
* ~Adds `aarch64-pc-windows-msvc` download link to cloud-invalidation.txt so that [this script](https://github.com/rust-lang/rust-forge/blob/master/blacksmith/src/lib.rs) could pick it up.~
  * ~Does the same for `aarch64-apple-darwin` because why not?~
* ~Updates the [Other Installation Methods page](https://forge.rust-lang.org/infra/other-installation-methods.html) to include Windows Arm64 version of Rustup.~ Split-off into #3854 
* Updates the [rustup.rs](https://rustup.rs) website to offer the arm64 version of Rustup on Windows Arm64.

The last part is the most "fragile" in my view because I don't know how the artefacts get published on rustup.rs, and detecting the arm64 CPU also required some async code that won't work in Firefox (but VS Code website’s arm64 detection doesn't work in Firefox either).

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -87,6 +87,29 @@
   <p class="other-platforms-help">You appear to be running Windows 64-bit. If not, <a class="default-platform-button" href="#">display all supported installers</a>.</p>
 </div>
 
+<div id="platform-instructions-win-arm64" class="instructions display-none">
+  <p>
+    To install Rust, download and run
+    <a class="windows-download" href="https://win.rustup.rs/aarch64">rustup&#x2011;init.exe</a>
+    then follow the onscreen instructions.
+  </p>
+  <p>You may also need the <a href="https://rust-lang.github.io/rustup/installation/windows-msvc.html">Visual Studio prerequisites</a>.</p>
+  <hr/>
+  <p>If you're a Windows Subsystem for Linux user run the following in your terminal, then follow the onscreen instructions to install Rust.</p>
+  <div class="copy-container">
+    <pre class="rustup-command">curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh</pre>
+    <button id="copy-button-win-arm64" class="copy-button" title="Copy curl command to clipboard to download Rustup" type="button">
+      <div class="copy-icon">
+        <svg width="24" height="25" viewBox="0 0 24 25" xmlns="http://www.w3.org/2000/svg" alt="Copy curl command to clipboard to download Rustup">
+          <path d="M18 20h2v3c0 1-1 2-2 2H2c-.998 0-2-1-2-2V5c0-.911.755-1.667 1.667-1.667h5A3.323 3.323 0 0110 0a3.323 3.323 0 013.333 3.333h5C19.245 3.333 20 4.09 20 5v8.333h-2V9H2v14h16v-3zM3 7h14c0-.911-.793-1.667-1.75-1.667H13.5c-.957 0-1.75-.755-1.75-1.666C11.75 2.755 10.957 2 10 2s-1.75.755-1.75 1.667c0 .911-.793 1.666-1.75 1.666H4.75C3.793 5.333 3 6.09 3 7z"></path><path d="M4 19h6v2H4zM12 11H4v2h8zM4 17h4v-2H4zM15 15v-3l-4.5 4.5L15 21v-3l8.027-.032L23 15z"></path>
+        </svg>
+      </div>
+      <div id="copy-status-message-win-arm64" class="copy-button-text">Copied!</div>
+    </button>
+  </div>
+  <p class="other-platforms-help">You appear to be running Windows on Arm. If not, <a class="default-platform-button" href="#">display all supported installers</a>.</p>
+</div>
+
 <div id="platform-instructions-unknown" class="instructions display-none">
   <!-- unrecognized platform: ask for help -->
   <p>I don't recognize your platform.</p>
@@ -134,6 +157,16 @@
 
   <div>
     <p>
+      If you are running Windows on Arm,<br/>download and run
+      <a class="windows-download" href="https://win.rustup.rs/aarch64">rustup&#x2011;init.exe</a>
+      then follow the onscreen instructions.
+    </p>
+  </div>
+
+  <hr/>
+
+  <div>
+    <p>
       If you are running Windows 32-bit,<br/>download and run
       <a class="windows-download" href="https://win.rustup.rs/i686">rustup&#x2011;init.exe</a>
       then follow the onscreen instructions.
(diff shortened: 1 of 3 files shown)


www/index.html
```diff
@@ -171,6 +204,16 @@
 
   <hr/>
 
+  <div>
+    <p>
+      If you are running Windows on Arm,<br/>download and run
+      <a class="windows-download" href="https://win.rustup.rs/aarch64">rustup&#x2011;init.exe</a>
```
rami3l, line 210: Before we merge this, I think we should let @rust-lang/infra have a look at the artifact deployment: it looks like the URL is invalid now, but I guess the URL itself looks perfectly fine.

doc/user-guide/src/installation/other.md
```diff
@@ -7,7 +7,7 @@ by platform:
   `x86_64-pc-windows-msvc` target][setup]. In general, this is the build of
   `rustup` one should install on Windows. This will require the Visual C++
   Build Tools 2019 or equivalent (Visual Studio 2019, etc.) to already be
-  installed. If you would prefer to install GNU toolchains or the i686
+  installed. If you would prefer to install GNU toolchains or the i686 or aarch64
```
rami3l: I think this is not the intended way to do this. If Windows on ARM64 lacks proper host triple detection, we should fix it instead of documenting the wrong behavior.

Alovchin91: No, it detects the host triple correctly. This was useful for x64 emulated Rustup, but is likely not needed anymore if the right artefact is published.

Alovchin91: I’ll rephrase this section; right now it recommends to use x64 Rustup in all cases, but I think since Ring workaround is not needed anymore there’s no reason not to recommend the aarch64 Rustup to Windows Arm users.

www/index.html
Alovchin91: I think I’ll split off the removal of the Ring workaround into a separate PR then, so that it could be merged sooner. Then this PR will be exclusively about the delivery.

rami3l: @Alovchin91 Do you think it's okay to remove the trailing ` # skip-pr` hint in the following line:
 
https://github.com/rust-lang/rustup/blob/2a5a69e0914ff419554d684ca71eb1d72c72bcb3/ci/actions-templates/windows-builds-template.yaml#L23

... so that we might gain more confidence about the quality of the build? Currently no tests will be run for this artifact, but the fact that it compiles means something too at least...

doc/user-guide/src/installation/other.md
rami3l: > No, it detects the host triple correctly. This was useful for x64 emulated Rustup, but is likely not needed anymore if the right artefact is published.

@Alovchin91 Let's track this in https://github.com/rust-lang/rustup/issues/3514.

www/index.html
Alovchin91: Yup, sounds good to me. Done!

doc/user-guide/src/installation/other.md
Alovchin91: Rephrased. I'm actually not sure if #3514 is relevant here, since I don't understand the scenario where one would choose to use emulated Rustup with a native toolchain 🤔 It might be useful though to let Rustup upgrade itself from an emulated to a native version.

www/index.html
Alovchin91: Also split the Ring part off into #3845

djc:
(This largely looks good to me, modulo one nit: would be nice to rebase away the merge commit.)

Alovchin91:
> (would be nice to rebase away the merge commit.)

Done! I've been rebasing for years now so I didn't realise "Update fork" button would actually create a merge commit 😅

rami3l:
> > (would be nice to rebase away the merge commit.)
> 
> Done! I've been rebasing for years now so I didn't realise "Update fork" button would actually create a merge commit 😅

@Alovchin91 You should've clicked "update with rebase". GitHub's default behavior is merging, unfortunately we cannot change it for now.

doc/user-guide/src/installation/other.md
```diff
@@ -3,9 +3,10 @@
 The primary installation method, as described at <https://rustup.rs>, differs
 by platform:
 
-* On Windows, download and run the [`rustup-init.exe` built for the
-  `x86_64-pc-windows-msvc` target][setup]. In general, this is the build of
-  `rustup` one should install on Windows. This will require the Visual C++
+* On Windows, download and run the `rustup-init.exe` built for the
+  [`x86_64-pc-windows-msvc`] or [`aarch64-pc-windows-msvc`] target.
+  In general, this is the build of `rustup` one should install on Windows,
+  depending on the OS architecture. This will require the Visual C++
```
rami3l: One tiny nit:

```suggestion
* On Windows, download and run the `rustup-init.exe` built for the
  [`x86_64-pc-windows-msvc`] or [`aarch64-pc-windows-msvc`] target,
  depending on your OS architecture. In general, this is the build of
  `rustup` one should install on Windows. This will require the Visual C++
```

doc/user-guide/src/installation/other.md
rami3l: @Alovchin91 Imagine one uses another app that installs Rustup. I don't know for Windows, but on macOS the guarantee is that once x64 is in the call chain, the rest of it is under emulation, so this will clearly cause problems. However this is out of the scope of this PR anyway, so don't worry 🙇

rami3l (approved):
LGTM! I'll ping t-infra again and let's see what we can do about https://win.rustup.rs/aarch64:
https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/.60win.2Erustup.2Ers.60.20should.20provide.20an.20.60aarch64.60.20installer

doc/user-guide/src/installation/other.md
Alovchin91: Done!

Alovchin91:
@rami3l Do you think it would make sense to split everything except rustup.rs website part off to a new PR that could be merged without waiting for infra? 🤔 It might already be helpful to have proper links in Other Installation Methods + the PR pipeline.

rami3l:
> @rami3l Do you think it would make sense to split everything except rustup.rs website part off to a new PR that could be merged without waiting for infra? 🤔 It might already be helpful to have proper links in Other Installation Methods + the PR pipeline.

@Alovchin91 Yes, definitely. You can also use https://static.rust-lang.org/rustup/dist/aarch64-pc-windows-msvc/rustup-init.exe temporarily instead of https://win.rustup.rs/aarch64 first, then make a small PR to change that into the final URL. It's your call.

Alovchin91:
@rami3l I think I'd prefer to keep it tidy and wait for the infra ☺️ Split off into #3854

rami3l:
~~Status: Waiting for infra's deployment of https://win.rustup.rs/aarch64.~~ I just tested and the link seems to work without issues! Thanks a lot, @jdno!

A note to @Alovchin91: for the actual changes being made here to be in effect, we might have to wait until the next official release of Rustup. The latest version of the docs, however, can be found [here](https://rust-lang.github.io/rustup/devel/installation/other.html). Thank you for your patience :)

Issue #3784, opened 2024-04-23
Decide on a better strategy for release notes/changelogs


djc:
Currently, we're generating changelogs with a Python script. This is trying to mimic the exact format that rustup posts on the Rust blog have been in the past, but it would be nice to simplify this approach.

See more discussion in https://github.com/rust-lang/rustup/pull/3773#discussion_r1576087894.

rami3l:
> > I don't think we should necessarily make the blog posts on the Rust blog actually like they have been before. In particular, IMO listing all contributors in the blog post is overkill (the Rust blog doesn't do that for Rust releases) and could be deferred to the changelog here.
> 
> Yeah, it's kinda tedious really. I wish I wouldn't have to spend like an hour (maybe two or more for the 0.1s?) per release night just to figure this out. Maybe the release blog posts are written in this way because... people tend to do what the Romans do to be safe, but things don't have to be this way.

_https://github.com/rust-lang/rustup/pull/3773#discussion_r1576087894_

@djc I have an idea! What about making [rust-lang/thanks](https://github.com/rust-lang/thanks) recognize us as well? The fact that we're currently not recognized on https://thanks.rust-lang.org is something I been always wanting to point out since a while ago. This way, we can just use a link towards that page to replace the whole section, just like Rustc does right now; also, all the Rustup contributors can then get the credit they deserve!

rami3l:
On the other hand, https://github.com/rust-lang/rfcs/pull/3503 has been merged and it seems that the `cargo` feature is already partially implemented. I think we can adopt `cargo-script` early with either this script or [`gen-workflows.sh`](https://github.com/rust-lang/rustup/blob/4cf9e1f90b6ab46ca43fbac50e2501c856e92169/ci/actions-templates/gen-workflows.sh).

djc:
Adding rustup to thanks seems nice, but I worry that it might be quite a bit of work, since from the description it looks like it might currently be tied into rust-lang/rust stuff.

I prefer boring/familiar stuff because it will be easier to maintain so I'd prefer to avoid `cargo script` until it actually lands in the stable toolchain.

PR #3841, opened 2024-05-24
fix(dist/triple): ensure `dist::triple::known` is up to date with `platforms`

### LINKED ISSUE #3783: `rustup component (add|remove)` should not rely on hardcoded target triples
### Verification

### Problem

> The direct cause of #3166 is that `wasm32-unknown-unknown` hasn't been hardcoded into our codebase:
> 
> https://github.com/rust-lang/rustup/blob/a6c9fae915f41d611ab88ddd2228e19ea978ecaa/src/dist/triple.rs#L7-L37
>
> [...] we really should not hardcode target triples in our artifact, as the complete list changes quite regularly.

_https://github.com/rust-lang/rustup/pull/3601#issue-2055114196_

### Rustup version

```console
rustup 1.27.0+74 (a6c9fae91 2024-04-21) dirty 1 modification
info: The currently active `rustc` version is `rustc 1.77.2 (25ef9e3d8 2024-04-09)`
```

rami3l:
Closes #3783.

By introducing [`platforms`](https://github.com/rustsec/rustsec/tree/main/platforms) as a new dependency, this PR tries to address both the problems that come with generated source files within `src`, as well as the issue that the list of known target triples is a moving target:
- Since `build.rs` already depends on full target triples to work, it no longer has to `include!()` some file inside `src`;
    > Grepping the whole repo showed that `RUSTUP_OVERRIDE_BUILD_TRIPLE` has been used only once in the actual business logic, and exactly as a full triple, so this partial triple thing in `build.rs` is probably not very useful after all?
    > 
    > https://github.com/rust-lang/rustup/blob/2a5a69e0914ff419554d684ca71eb1d72c72bcb3/src/dist/dist.rs#L261
    > https://github.com/rust-lang/rustup/blob/2a5a69e0914ff419554d684ca71eb1d72c72bcb3/src/dist/dist.rs#L266-L268
    _https://github.com/rust-lang/rustup/issues/3783#issuecomment-2128970109_
- The source file containing lists of known target triple segments, albeit generated, is committed directly into the codebase as a regular Rust module to enhance visibility;
- The code generation coherence test is based on a specific version of `platforms`, so the output will not change as long as `platforms`' version is locked.

## Concerns
- [x] ~~Are custom triples involved in this change?~~ No, they are handled by another mechanism (`CustomToolchainName`).
- [ ] Will the original approach of building a regex of all list members be not scalable?
    https://github.com/rust-lang/rustup/blob/2a5a69e0914ff419554d684ca71eb1d72c72bcb3/src/dist/triple.rs#L70-L79


build.rs
```diff
@@ -12,6 +17,96 @@ fn from_build() -> Result<String, String> {
...
+        .filter_map(|p| parse_triple(p.target_triple))
+    {
+        archs.insert(arch);
+        oses.insert(os);
+        if !env.is_empty() {
+            envs.insert(env);
+        }
+    }
+
+    let dst = Path::new(&env::var("OUT_DIR").unwrap()).join("known_triples.rs");
```
djc: I strongly dislike code generation/inclusion via `OUT_DIR` because it ends up making things less visible. Would still prefer to generate this in an integration test instead. If you don't want to make CI fail, we can stick `#[ignore]` on the test so that it isn't run by default.

rami3l: @djc Yes, and that's why this PR is marked as a draft.

> The current Implementation is very dumb, but at least it gets things done.

It's... PoC quality, I'd say. However it does seem to make things work. At the moment that's all what counts, I have no intention to merge this as-is.

rami3l: > I strongly dislike code generation/inclusion via `OUT_DIR` because it ends up making things less visible. Would still prefer to generate this in an integration test instead. If you don't want to make CI fail, we can stick `#[ignore]` on the test so that it isn't run by default.

@djc Regarding this point I'll see how Homebrew performs a similar task. On each formula change it'll optionally update the file via a bot. Like what we'll expect when updating dependencies.

You're essentially asking for something like https://github.com/pinnacles/validate-go-generate...

It would not be too hard to replicate that in Rust using the "examples" trick like you have suggested in #3784. I'll use `platforms` as usual.

The source file in question will come with a special marker either in the path name or in the file itself (I myself prefer the former). On each CI we delete the generated file, regenerate it, and let `git` handle the rest.

djc: https://github.com/chronotope/chrono/blob/main/tests/win_bindings.rs, https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-stackdriver/tests/generate.rs, there are a bunch more examples out there.

rami3l: @djc I've changed the implementation accordingly. How do you think?

build.rs
```diff
@@ -3,12 +3,9 @@ use std::env;
 include!("src/dist/triple.rs");
 
 fn from_build() -> Result<PartialTargetTriple, String> {
-    let triple = if let Ok(triple) = env::var("RUSTUP_OVERRIDE_BUILD_TRIPLE") {
-        triple
-    } else {
-        env::var("TARGET").unwrap()
-    };
     PartialTargetTriple::new(&triple).ok_or(triple)
+    let triple =
```
djc, line 6: This change looks wrong, defining `triple` after using it.

build.rs
```diff
@@ -1,23 +1,26 @@
...
     match from_build() {
-        Ok(triple) => eprintln!("Computed build based partial target triple: {triple:#?}"),
+        Ok(triple) => eprintln!("Computed build based on target triple: {triple:#?}"),
         Err(s) => {
-            eprintln!("Unable to parse target '{s}' as a PartialTargetTriple");
+            eprintln!("Unable to parse target '{s}' as an existing target triple");
             eprintln!(
-                "If you are attempting to bootstrap a new target you may need to adjust the\n\
-               permitted values found in src/dist/triple.rs"
+                "If you are attempting to bootstrap a new target, you might need to update `platforms` to a newer version in `Cargo.toml`"
```
djc: Leave off "to a newer version in `Cargo.toml`"? A `Cargo.lock` update might be sufficient and I don't think the added detail is necessary here.

Do you know/understand why we need to emit the `rustc-env` value?

src/dist/triple.rs
```diff
@@ -1,51 +1,9 @@
...
-    "msvc",
-    "gnueabi",
-    "gnueabihf",
-    "gnuabi64",
-    "androideabi",
-    "android",
-    "musl",
-];
+use once_cell::sync::Lazy;
+use regex::Regex;
```
djc: Let's keep these imports at the top of the file?

src/dist/triple.rs
```diff
@@ -1,51 +1,9 @@
-use once_cell::sync::Lazy;
-use regex::Regex;
+mod known;
 
-// These lists contain the targets known to rustup, and used to build
-// the PartialTargetTriple.
+pub use self::known::{LIST_ARCHS, LIST_ENVS, LIST_OSES};
```
djc: Would be nice to split this commit in two parts: one that moves the constants to a separate file and another that starts using them in `toolchain::names`.

We don't need to redirect via `self::` here, we can import from `known::` directly.

tests/suite/known_triples.rs
```diff
@@ -0,0 +1,92 @@
...
+
+#[test]
+fn gen_known_triples() {
+    let output = "src/dist/triple/known.rs";
+    let existing = std::fs::read_to_string(output).unwrap();
+    assert_eq!(known_triples(), existing);
+}
+
+/// Generates the lists of known target architectures, OSes and environments.
+fn known_triples() -> String {
```
djc: I would prefer to just inline the contents of this function, since it only has a single caller anyway.

tests/suite/known_triples.rs
```diff
@@ -0,0 +1,92 @@
+use std::collections::BTreeSet;
+
+use platforms::Platform;
+
+#[test]
+fn gen_known_triples() {
+    let output = "src/dist/triple/known.rs";
+    let existing = std::fs::read_to_string(output).unwrap();
+    assert_eq!(known_triples(), existing);
```
djc: The developer ergonomics are nicer if you can move the write the new output into the file in the failure case here, so that a human user can easily commit the updates and move on. See the examples I linked before.

tests/suite/known_triples.rs
```diff
@@ -0,0 +1,92 @@
...
+        match triple.split('-').collect::<Vec<_>>()[..] {
+            [arch, os] => Some((arch, os, "")),
+            [arch, os @ ("none" | "linux"), env] => Some((arch, os, env)),
+            [arch, _, _] => Some((arch, &triple[(arch.len() + 1)..], "")),
+            [arch, _, _, env] => Some((
+                arch,
+                &triple[(arch.len() + 1)..(triple.len() - env.len() - 1)],
+                env,
+            )),
+            _ => None,
```
djc: Do these cases happen? Why? (Can you show an example?)

djc (commented):
Nice! I have a bunch of small suggestions but this is looking pretty good!

(Sorry for taking a while to review it.)

build.rs
rami3l: @djc I don't understand how this commit has been created. Maybe `git absorb` did something dumb. The resulting function in the end works perfect though. I'll do a cleanup tomorrow.

tests/suite/known_triples.rs
rami3l: @djc Nice catch! It'd be much better to let it crash there.

build.rs
rami3l: > Leave off "to a newer version in `Cargo.toml`"? A `Cargo.lock` update might be sufficient and I don't think the added detail is necessary here.
Done.

> Do you know/understand why we need to emit the `rustc-env` value?
No idea, introduced in https://github.com/rust-lang/rustup/pull/1704 back in 2019 though. I highly suspect this has something to do with the current CI: 

https://github.com/search?q=repo%3Arust-lang%2Frustup+TARGET+language%3Ayaml&type=code

rami3l: > Leave off "to a newer version in `Cargo.toml`"?
> A `Cargo.lock` update might be sufficient and I don't think the added detail is necessary here.

@djc Done.

> Do you know/understand why we need to emit the `rustc-env` value?

This is required by a potential platform-specific `miri` use. See https://github.com/rust-lang/rust/blob/804421dff5542c9c7da5c60257b5dbc849719505/src/tools/miri/build.rs#L4-L7.

tests/suite/known_triples.rs
rami3l: @djc I have adjusted the test accordingly, and I have tested the in-place update locally to confirm that it's actually working.

djc (approved):
Nice!

PR #3819, opened 2024-05-10
Purge Fastly cache after release

jdno:
When releasing a new version of rustup, we need to purge the cached version on Fastly. This is done by calling Fastly's API using curl and purging the surrogate key that is used for rustup.

See https://github.com/rust-lang/simpleinfra/issues/415 for more details.

PR #3844, opened 2024-05-28
Document the new release process for Rustup

jdno:
The Rustup release process has historically been a manual process that involved copying files from S3 to the local machine and back to S3. This introduced a high risk of human error. When modifications to the existing release script became necessary, the decision was made to automate the release process (see #3819 for details).

The documentation in the dev-guide has been updated to cover the new release process, which is fully automated to produce `beta` releases using GitHub Actions and the [promote-release] tooling.

[promote-release]: https://github.com/rust-lang/promote-release

### DIFF
--- a/doc/dev-guide/src/release-process.md
+++ b/doc/dev-guide/src/release-process.md
@@ -1,66 +1,217 @@
-# Making a release
+# Release Process
 
-Before making a release, ensure that `rustup-init.sh` is behaving correctly,
+This document describes the process for making a new release of Rustup. It is
+split into the two sections. The first section explains the release process,
+while the second section documents for maintainers how to make a release.
+
+- [About the Release Process](#about-the-release-process)
+    - [Historic Context](#historic-context)
+    - [Beta and Stable Releases](#beta-and-stable-releases)
+    - [Release Artifacts](#release-artifacts)
+    - [Automation](#automation)
+- [How to Cut a Release](#how-to-cut-a-release)
+    - [1. Test `rustup-init.sh`](#1-test-rustup-initsh)
+    - [2. Update Version](#2-update-version)
+    - [3. Update Checksum](#3-update-checksum)
+    - [4. Sync `master` to `stable`](#4-sync-master-to-stable)
+    - [5. Test the `beta` Release](#5-test-the-beta-release)
+    - [6. Prepare Blog Post for `stable` Release](#6-prepare-blog-post-for-stable-release)
+    - [7. Request the `stable` Release](#7-request-the-stable-release)
+    - [8. Tag the `stable` Release](#8-tag-the-stable-release)
+
+## About the Release Process
+
+This section explains the release process for Rustup.
+
+### Historic Context
+
+The Rustup release process has evolved over time. In the past, the process
+involved more manual steps and large parts of it were executed locally. This
+required careful execution and coordination to avoid accidentally overwriting
+exiting releases, since the tooling did not implement sanity checks that would
+prevent destructive actions.
+
+A full step-by-step guide for the old process can be found in
+the [old developer guide](https://github.com/rust-lang/rustup/blob/1cf1b5a6d80c978e0dcaabbce5f10b3861612425/doc/dev-guide/src/release-process.md).
+The following summarizes how release artifacts were created and copied around.
+
+In the past, both beta and stable releases were started by merging a new commit
+into the `stable` branch in [rust-lang/rustup]. This started a new build on
+GitHub Actions that produced release
+artifacts ([source](https://github.com/rust-lang/rustup/blob/1cf1b5a6d80c978e0dcaabbce5f10b3861612425/.github/workflows/ci.yaml#L144-L151)),
+which were uploaded to `s3://dev-static-rust-lang-org/rustup/dist`. As new
+commits were merged into `stable`, they would overwrite the artifacts from prior
+builds.
+
+The release artifacts were then copied to the final location by running a script
+on a local machine. This script would download the artifacts from the
+`dev-static` bucket on S3 and upload them to the `static` bucket. The script
+also generated a new manifest with the given version, and uploaded a copy of the
+files to an archive directory.
+
+Given that the process was manual, involved copying files to the local machine,
+and waiting between creating a `beta` and a `stable` release, there was a risk
+of human error. When we had to update the script after four years without a
+change, we [decided](https://github.com/rust-lang/rustup/pull/3819) to redesign
+and automate the release process.
+
+### Beta and Stable Releases
+
+Rustup can be released to two different environments: `beta` and `stable`. The
+main difference between the two is that they use different values for the
+`RUSTUP_UPDATE_ROOT` environment variable:
+
+- A beta release is deployed on `https://dev-static.rust-lang.org/rustup`.
+- An official release is deployed on `https://static.rust-lang.org/rustup`.
+
+By switching between those two values, Rustup effectively provides two
+"self update channels", making beta testing possible with `rustup self update`.
+
+### Release Artifacts
+
+The release artifacts are produced by CI when a new commit is merged into the
+`stable` branch, and they are uploaded to the `dev-static` bucket on S3. There,
+they are put into a folder named after their commit hash, for example the
+artifacts for commit `1cf1b5a` would be uploaded to
+`s3://dev-static-rust-lang-org/rustup/builds/1cf1b5a`.
+
+When a new `beta` release is cut, the artifacts are copied to two new locations
+within the same bucket:
+
+- One copy is put into an archive named after the version, e.g.
+  `/rustup/archive/1.0.0`.
+- Another copy is put into the `/rustup/dist` folder, which is where clients
+  look for new versions.
+
+When a new `stable` release is cut, the artifacts are copied to the `static`
+bucket on S3, following the same process as the `beta` release:
+
+- One copy is archived into `/rustup/archive/1.0.0`.
+- Another copy is put into `/rustup/dist`.
+
+This ensures backwards compatibility with the old release process, while also
+ensuring that release artifacts are not overwritten by new builds.
+
+### Automation
+
+The interaction with the release artifacts is fully automated.
+
+First, artifacts are produced automatically by
+the [`CI`](https://github.com/rust-lang/rustup/blob/master/.github/workflows/ci.yaml)
+job on GitHub Actions when a new commit is merged into the `stable` branch, and
+are then automatically uploaded to their respective S3 bucket by the action as
+well.
+
+Second, when making a release, the artifacts are copied to their final locations
+by the [promote-release] tool. This reduces the risk of human error and ensures
+that the release process is consistent and reliable. The tool is also run in a
+secure environment on AWS CodeBuild, reducing the risk of leaking sensitive
+credentials that would give write access (past) releases.
+
+For a `beta` release, `promote-release` performs the following actions:
+
+1. Query GitHub's API to get the latest commit on the `stable` branch
+2. Confirm that `/rustup/builds/${commit}` exists in the `dev-static` bucket
+3. Get the new version number from the `stable` branch
+    1. Download `Cargo.toml` from `stable`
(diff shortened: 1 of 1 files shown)


jdno:
Here's a rendered version of the file: https://github.com/jdno/rust-rustup/blob/new-release-process/doc/dev-guide/src/release-process.md

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
+artifacts ([source](https://github.com/rust-lang/rustup/blob/1cf1b5a6d80c978e0dcaabbce5f10b3861612425/.github/workflows/ci.yaml#L144-L151)),
+which were uploaded to `s3://dev-static-rust-lang-org/rustup/dist`. As new
+commits were merged into `stable`, they would overwrite the artifacts from prior
+builds.
+
+The release artifacts were then copied to the final location by running a script
+on a local machine. This script would download the artifacts from the
+`dev-static` bucket on S3 and upload them to the `static` bucket. The script
+also generated a new manifest with the given version, and uploaded a copy of the
+files to an archive directory.
```
Mark-Simulacrum, line 50: This paragraph doesn't sound quite true to me - there was an intermediate step that uploaded something back to dev-static I think? Maybe just the manifest?

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
+- A beta release is deployed on `https://dev-static.rust-lang.org/rustup`.
+- An official release is deployed on `https://static.rust-lang.org/rustup`.
+
+By switching between those two values, Rustup effectively provides two
+"self update channels", making beta testing possible with `rustup self update`.
+
+### Release Artifacts
+
+The release artifacts are produced by CI when a new commit is merged into the
+`stable` branch, and they are uploaded to the `dev-static` bucket on S3. There,
```
Mark-Simulacrum: I think we should move this to a different bucket. CI shouldn't have direct access to the static buckets IMO, it also complicates any intermediate processing (e.g. signing or recompression).

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
-[Internals Forum]: https://internals.rust-lang.org
+2. Update `CHANGELOG.md` accordingly if necessary.
+
+Submit the changes in a PR and merge it.
+
+### 3. Update Checksum
+
+After merging the PR in the previous step, update the commit SHA checksum in
+`rustup-init.sh` to match the latest commit on `master`. Submit the change in a
+PR and merge it.
```
Mark-Simulacrum, line 163: This step sounds surprising to me. If someone updates this to some other commit, how will we notice? Is there an opportunity to adjust the script as part of CI publishing it?

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
+While you wait for green CI on `stable`, double-check the functionality of
+`rustup-init.sh` and `rustup-init` just in case.
+
+Ensure all of CI is green on the `stable` branch. Once it is, check through a
+representative proportion of the builds looking for the reported version
+statements to ensure that we definitely built something cleanly which reports as
+the right version number when run `--version`.
+
+Once the beta release has happened, post a new topic named "Seeking beta testers
+for Rustup $VER_NUM" on the [Internals Forum].
```
Mark-Simulacrum, line 191: We've automated this for rustc, maybe can reuse that for rustup...

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
+
+They will have to start a new CodeBuild job on AWS to run the `promote-release`
+tool for Rustup.
+
+### 8. Tag the `stable` Release
+
+Once the official release has happened, prepare and push a tag on the latest
+`stable` commit.
+
+- `git tag -as $VER_NUM -m $VER_NUM` (optionally without `-s` if not GPG
```
Mark-Simulacrum, line 209: Should we do this in promote release? We do for rust and cargo (also always signing the commits with the official key).

doc/dev-guide/src/release-process.md
rami3l: @Mark-Simulacrum This is the most painful point for me personally during the past 2 releases (v1.27.0 and v1.27.1, for which I actually handled the release process), and I have no idea why this have to be the case.

Well, when we migrated to GitHub Merge Queue in v1.27.0, we have almost broken the whole workflow (which relied on merges instead of rebases and thus assumed shasum stability; to keep the coherence with the old workflow, a 2nd PR is required with GHMQ enabled for script traceability, see https://github.com/rust-lang/rustup/pull/3653#issuecomment-1918320204 for more context). The current description is the closest I could get from the "spirit" of the original.

I'd love to see this step being automated. As long as the resulting `rustup-init.sh` script is traceable via the shasum when it's published onto `rustup.sh` etc., I personally don't see a particular reason we cannot leave a placeholder on `master` for example.

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
+and automate the release process.
+
+### Beta and Stable Releases
+
+Rustup can be released to two different environments: `beta` and `stable`. The
+main difference between the two is that they use different values for the
+`RUSTUP_UPDATE_ROOT` environment variable:
+
+- A beta release is deployed on `https://dev-static.rust-lang.org/rustup`.
+- An official release is deployed on `https://static.rust-lang.org/rustup`.
```
rami3l: I guess this adjustment will make it a better fit for the new context...

```suggestion
- A stable release is deployed on `https://static.rust-lang.org/rustup`.
```

doc/dev-guide/src/release-process.md
rami3l: @Mark-Simulacrum Will be cool. So far I have to ping `t-release` before bedtime, and manually do the tagging the next morning 😓

doc/dev-guide/src/release-process.md
jdno: Yes, the full process is a bit more involved. But I'm not sure if that is really relevant for this section. The reader can always review the script in the Git history.

doc/dev-guide/src/release-process.md
jdno: That's fine with me. How about we create a `rustup-builds` bucket that stores the artifacts, similar to the bucket for `rust-lang/rust`?

doc/dev-guide/src/release-process.md
jdno: We can look into automating this step, but I would do that after refactoring the release process as-is. This is already quite a big change and I don't want to slow down the implementation or add additional complexity to it.

doc/dev-guide/src/release-process.md
jdno: Happy to create a tracking issue for this so that we can look into it once we've refactored the tooling.

doc/dev-guide/src/release-process.md
Mark-Simulacrum: Sounds perfect.

doc/dev-guide/src/release-process.md
Mark-Simulacrum: Yeah, for sure. Mostly just flagging it since it sounded off.

rami3l (commented):
@jdno Should we remove the manual release script, as I assume it's not needed anymore?

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
-# Making a release
+# Release Process
 
-Before making a release, ensure that `rustup-init.sh` is behaving correctly,
+This document describes the process for making a new release of Rustup. It is
+split into the two sections. The first section explains the release process,
```
djc: "into two sections"?

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
+while the second section documents for maintainers how to make a release.
```
djc: "documents how to make a release"? The "for maintainers" part reads a little awkwardly.

(Also "make" feels a little vague here, can we make that more precise?)

doc/dev-guide/src/release-process.md
```diff
@@ -1,66 +1,217 @@
...
+## About the Release Process
+
+This section explains the release process for Rustup.
+
+### Historic Context
+
+The Rustup release process has evolved over time. In the past, the process
+involved more manual steps and large parts of it were executed locally. This
+required careful execution and coordination to avoid accidentally overwriting
+exiting releases, since the tooling did not implement sanity checks that would
```
djc: "exiting" -> "existing"

djc (approved):

djc:
Thanks for working on this!

doc/dev-guide/src/release-process.md
jdno: I've rewritten the into a bit, but I'm not sure how much better it is now. 😅

doc/dev-guide/src/release-process.md
jdno: The config for the new bucket has been created here: https://github.com/rust-lang/simpleinfra/pull/428

djc:
@jdno what's your plan for driving the new release process to completion? We're starting to think about how/when we want to publish our next release so it would be good to understand how much of a dependency we have on you and what that means for the schedule.

jdno:
I've started working on it, but I can't give a good estimate on how long it'll take. My hope is to get most of the implementation work done in the next two weeks.

I don't want to be a blocker, though, so my work is strictly additive right now. The current process will continue to work until we have a full replacement. So regarding dependencies or the schedule, there shouldn't be any.

doc/dev-guide/src/release-process.md
ia0: If I understand correctly what this mean, I agree it would be cool.

Currently I have an automated process to make sure the script that I download from rustup.rs is the same as a specific tagged release from this repository. Until a few hours ago it was v1.27.1. Now it should be a commit between e2d9e7edde6ee1a79e6e44f3b81065443c1fc6eb and master that would be tagged v1.28.0. But there is no such tag yet.

If I understand correctly, this tagging step is supposed to happen after the file is distributed. And this comment suggests to do it at the same time. Is that right?

joshka:
I'd strongly recommend adding roll back process for when a deployment causes failure like what happened in 1.28.0.

I'd like to echo and amplify the comments in https://github.com/rust-lang/rustup/issues/4211 about rolling back to make sure that there's more than one voice calling for this practice. The 1.28.0 release highlighted that Rustup is a load bearing piece of software. Having a failed deployment cause many downstream failures like this should have resulted in a rollback within about an hour, not a roll forward a day later.

I base this opinion having participated (as an owner and as someone affected by) in many post-mortems at a large internet company. The consensus engineering best practice is to rollback first, and think about how to solve the greater problem second. It's rare that this advice should be ignored, and there doesn't seem like there's a any good rationale that's been expressed here why that should have been so here. The users that migrated to using the new rustup changes would generally have been a very small number compared to those who were affected by this, and those users were actively aware of the changes. The impacted users were generally not made aware until this caused failure in their systems.

I want to explicitly state this comment is not intended to throw shade at all on anyone involved in 1.28. It is only meant to constructively improve the situation going forward.

rami3l:
> I'd strongly recommend adding roll back process for when a deployment causes failure like what happened in 1.28.0.

@joshka I second this. In fact, I [have already mentioned it on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/241545-t-release/topic/Rustup.201.2E28.2E0.20-.20testing.2Frelease/near/503397698):

> by the way, it’s not urgent, but can we have /archive/dist on our release server redirect to /archive or something? Anyway the goal is to trick rustup into thinking the archive is an actual release server, to provide a way for arbitrary rustup downgrade/upgrade/pinning without having to adapt the code on our side

I added that rustup could be modified to make use of the `/archive` path, however this modification itself would also need a new release to work.

Let's leave this thread for the release-related work: I've made https://github.com/rust-lang/rustup/issues/4240 for the rustup-related part of this problem.

PR #3845, opened 2024-05-28
Re-enable Ring support on Windows Arm64

Alovchin91:
Ring 0.17.x supports Windows on ARM so we can re-enable it.

Split-off from #3840


rami3l (approved):
Thanks!

PR #3847, opened 2024-05-30
Follow up #3367 to improve `run_future`

wtdcode:
Following up https://github.com/rust-lang/rustup/pull/3367 to slightly improve the `run_future` utils and add a few caveats. It gets merged before I'm trying to add my reviews so here is the separate PR.

### DIFF
--- a/src/utils/utils.rs
+++ b/src/utils/utils.rs
@@ -10,8 +10,7 @@ use home::env as home;
 use retry::delay::{jitter, Fibonacci};
 use retry::{retry, OperationResult};
 use sha2::Sha256;
-use tokio::runtime::Handle;
-use tokio::task;
+use tokio::runtime::{Handle, RuntimeFlavor};
 use url::Url;
 
 use crate::currentprocess::{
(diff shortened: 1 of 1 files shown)


rami3l:
To complete the context, this is a separate concern that @wtdcode has expressed to me in a DM right after https://github.com/rust-lang/rustup/pull/3367#issuecomment-2139401584 having worked on the very same functionality in https://github.com/bluealloy/revm/blob/6d5256eaa9212e17da5a9fd8e9131fc85a0e7ad9/crates/revm/src/db/ethersdb.rs#L35-L63.

In particular, the current implementation of this function relies on the fact that `handle.runtime_flavor()` will not return `RuntimeFlavor::CurrentThread` to work correctly. Currently this point is ensured in two places, but the link here is implicit, making it potentially difficult to find for future maintainers:

https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/Cargo.toml#L147

https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/rustup-macros/src/lib.rs#L115-L119

This PR tries to document more caveats on this point and cause `run_future()` to panic on a wrong runtime setup. For more info on this, see https://github.com/tokio-rs/tokio/discussions/4563.

djc:
Hmmm -- IMO we should just remove all uses of `run_future()` instead. Not sure why we still need it?

rami3l:
> Hmmm -- IMO we should just remove all uses of `run_future()` instead. Not sure why we still need it?

@djc Yes, that's the end goal for us, but looking at the removal of `deprecated()` in e92b85ec9e5cbdf5d37cf92fe90a1e77b8883b74 or the recently-restarted transition to `rustls` (#3790) I guess it's still better to be a bit more defensive :)

djc:
> > Hmmm -- IMO we should just remove all uses of `run_future()` instead. Not sure why we still need it?
> 
> @djc Yes, that's the end goal for us, but looking at the removal of `deprecated()` in [e92b85e](https://github.com/rust-lang/rustup/commit/e92b85ec9e5cbdf5d37cf92fe90a1e77b8883b74) or the recently-restarted transition to `rustls` (#3790) I guess it's still better to be a bit more defensive :)

Not sure how/why those are related to this issue?

rami3l:
> > > Hmmm -- IMO we should just remove all uses of `run_future()` instead. Not sure why we still need it?
> > 
> > 
> > @djc Yes, that's the end goal for us, but looking at the removal of `deprecated()` in [e92b85e](https://github.com/rust-lang/rustup/commit/e92b85ec9e5cbdf5d37cf92fe90a1e77b8883b74) or the recently-restarted transition to `rustls` (#3790) I guess it's still better to be a bit more defensive :)
> 
> Not sure how/why those are related to this issue?

@djc Those are two examples of a transient state ended up remaining much longer than originally expected (x_x), I mean it would be perfect if this time it ends sooner, but still I think this PR would be a nice addition for the moment being even if the lines it adds end up being deleted in months.

djc:
> > > > Hmmm -- IMO we should just remove all uses of `run_future()` instead. Not sure why we still need it?
> > > 
> > > 
> > > @djc Yes, that's the end goal for us, but looking at the removal of `deprecated()` in [e92b85e](https://github.com/rust-lang/rustup/commit/e92b85ec9e5cbdf5d37cf92fe90a1e77b8883b74) or the recently-restarted transition to `rustls` (#3790) I guess it's still better to be a bit more defensive :)
> > 
> > 
> > Not sure how/why those are related to this issue?
> 
> @djc Those are two examples of a transient state ended up remaining much longer than originally expected (x_x), I mean it would be perfect if this time it ends sooner, but still I think this PR would be a nice addition for the moment being even if it ends up being deleted in months.

To ask the question differently: why do you think we need to keep `utils::run_future()` around today?

rami3l:
> To ask the question differently: why do you think we need to keep `utils::run_future()` around today?

@djc I think we can agree that this is a temporary scaffolding function before we make the full async transition, as previously explained pretty clearly in https://github.com/rust-lang/rustup/pull/3367#issuecomment-1680225332:

> If that is changed naively by just making main() async, and reqwest async, then we get rust's async function tainting everything and will change all the relevant calls to async, but there are a number of places where types need to change to be compatible with async requirements; so the commit becomes huge and unwieldy very quickly.
> 
> So the approach I'm taking is some scaffolding that lets us work this incrementally, and once complete it will just be nice async code.

Apart from the signature generalization (which you might argue as some form of premature optimization), this PR is actually mostly documenting this function's limitations (multithread runtime only, and a limit of 512 for blocking threads), so that one might get some clue about a potential misuse.

I was not saying that I wish the transient state would last long... Actually I wish it could end today. But maybe I'm being overly cautious?

rbtcollins:
I think run_future is very close to gone right? I didn't actually intend to leave it in the migration to an async runtime.

rami3l:
Closing in favor of #3856. Thanks anyway @wtdcode!

Issue #3852, opened 2024-06-02
Tests relying on `reqwest::Client` may fail with "dispatch task is gone: runtime dropped the dispatch task"


rami3l:
### Verification

- [X] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue+is%3Aopen%2Cclosed and found no duplicates.
- [X] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

This is causing irregular failures for different CI platforms:
- Linux
  - https://github.com/rust-lang/rustup/actions/runs/9303236385/job/25605096671
  - https://github.com/rust-lang/rustup/actions/runs/9354772182/job/25748366479
  - https://github.com/rust-lang/rustup/actions/runs/9337787790/job/25700115067
  - https://github.com/rust-lang/rustup/actions/runs/9343224087/job/25712479916
- Windows
  - https://github.com/rust-lang/rustup/actions/runs/9335696949/job/25695378538

The error message here is identical (although the exact test that failed might differ):
```console
running 395 tests
...
test suite::cli_exact::check_updates_with_update ... FAILED
error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to '/tmp/rustup-updateD58GcC/release-stable.toml': failed to make network request: error sending request for url (https://static.rust-lang.org/rustup/release-stable.toml): client error (SendRequest): dispatch task is gone: runtime dropped the dispatch task
```

### Notes

Possibly related to https://github.com/hyperium/hyper/issues/2312, https://github.com/Azure/azure-sdk-for-rust/pull/1550 or https://github.com/fussybeaver/bollard/issues/387#issuecomment-2008133973.

I'm pretty sure that #3367 has caused this.

rami3l:
According to the error message format and my IDE searching the current codebase, this is a `RustupError::DownloadingFile`. This kind error message can only be emitted from here:

https://github.com/rust-lang/rustup/blob/6d4376180d10e35daec5453546e508eb4726fa52/src/utils/utils.rs#L154-L190

twuebi:
[...]

PR #3853, opened 2024-06-02
consistently add context with file path when parsing fails

### LINKED ISSUE #3851: all cargo commands error, because some manifest is broken
### Verification

### Problem

Independent of the directory and command I run cargo errors with the following error message:
```
rustup show active-toolchain
error: error parsing manifest: TOML parse error at line 1478, column 23
     |
1478 | [[pkg.rust.target.aarcwin.extensions]]
     |                       ^
invalid table header
expected `.`, `]]`
```

```
 cargo --help
error: error parsing manifest: TOML parse error at line 1478, column 23
     |
1478 | [[pkg.rust.target.aarcwin.extensions]]
     |                       ^
invalid table header
expected `.`, `]]`
```
in vscode the rust-analyzer log is a bit more helpfull with:

```
rust-analyzer failed to load workspace: Failed to load the project at /home/bennet/git/project/Cargo.toml: Failed to query rust toolchain version at /home/bennet/git/project, is your toolchain setup correctly?: cd "/home/bennet/git/project" && "/home/bennet/.cargo/bin/cargo" "--version" failed, exit status: 1
stderr:
error: error parsing manifest: TOML parse error at line 1478, column 23
     |
1478 | [[pkg.rust.target.aarc<!--a bunch of nul bytes-->win.extensions]]
     |                       ^
invalid table header
expected `.`, `]]`
```
<details>
<summary>It shows a bunch of null bytes where `&lt!--a bunch of nul bytes-->` is.</summary>
image
</details>

This is under WSL2 Ubuntu on Windows 11, I believe the corruption might be due to a recent export import of the WSL Image/distribution.
But I think cargo should handle this better, at least a filename of the manifest that failed to parse would be good.

### Steps

1. somehow corrupt the manifest file
2. run any of the command show above

### Possible Solution(s)

Include manifest path in error message

### Notes

_No response_

### Rustup version

```console
$> rustup --version
```

### Installed toolchains

```console
$> rustup show  
rustup home:  /home/bennet/.rustup

installed toolchains
--------------------

esp

active toolchain
----------------

(error: error parsing manifest, TOML parse error at line 1478, column 23
     |
1478 | [[pkg.rust.target.aarcwin.extensions]]
     |                       ^
invalid table header
expected `.`, `]]`
)
```

### OS version

```console
WSL2 under Windows 11 with

$> lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal
```

Skgland:
fixes #3851

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -641,11 +641,15 @@ impl Cfg {
             };
 
             if let Ok(contents) = contents {
-                let add_file_context = || format!("in {}", toolchain_file.to_string_lossy());
                 // XXX Should not return the unvalidated contents; but a new
                 // internal only safe struct
-                let override_file = Cfg::parse_override_file(contents, parse_mode)
-                    .with_context(add_file_context)?;
+                let override_file =
+                    Cfg::parse_override_file(contents, parse_mode).with_context(|| {
+                        RustupError::ParsingFile {
+                            name: "override",
+                            path: toolchain_file.clone(),
+                        }
+                    })?;
                 if let Some(toolchain_name_str) = &override_file.toolchain.channel {
                     let toolchain_name = ResolvableToolchainName::try_from(toolchain_name_str)?;
                     let default_host_triple = get_default_host_triple(settings);
--- a/src/dist/config.rs
+++ b/src/dist/config.rs
@@ -43,7 +43,7 @@ impl Config {
     }
 
     pub(crate) fn parse(data: &str) -> Result<Self> {
-        let value = toml::from_str(data).context("error parsing manifest")?;
+        let value = toml::from_str(data).context("error parsing config")?;
         Self::from_toml(value, "")
     }
 
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -1096,7 +1096,11 @@ pub(crate) async fn dl_v2_manifest(
                 return Ok(None);
             };
             let manifest_str = utils::read_file("manifest", &manifest_file)?;
-            let manifest = ManifestV2::parse(&manifest_str)?;
+            let manifest =
+                ManifestV2::parse(&manifest_str).with_context(|| RustupError::ParsingFile {
+                    name: "manifest",
+                    path: manifest_file.to_path_buf(),
+                })?;
 
             Ok(Some((manifest, manifest_hash)))
         }
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -313,8 +313,12 @@ impl Manifestation {
 
         // Read configuration and delete it
         let rel_config_path = prefix.rel_manifest_file(CONFIG_FILE);
-        let config_str = utils::read_file("dist config", &prefix.path().join(&rel_config_path))?;
-        let config = Config::parse(&config_str)?;
+        let abs_config_path = prefix.path().join(&rel_config_path);
+        let config_str = utils::read_file("dist config", &abs_config_path)?;
+        let config = Config::parse(&config_str).with_context(|| RustupError::ParsingFile {
+            name: "config",
+            path: abs_config_path,
+        })?;
         tx.remove_file("dist config", rel_config_path)?;
 
         for component in config.components {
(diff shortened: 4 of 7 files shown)


rami3l:
@Skgland Thanks for making this PR!

Would you mind explain in detail what is the motivation of this PR and what has been improved with a simple before/after comparison, for example?

PS: To make sure we are on the same page, regarding #3851 I would like to underline that the server incident causing bad manifests being downloaded is probably out of the scope of the Rustup project (so it possibly cannot be "fixed"), although improvements regarding error handling are welcomed.

rami3l:
Your CI failure is https://github.com/rust-lang/rustup/issues/3852, we will be investigating this on our side.

Skgland:
> Would you mind explain in detail what is the motivation of this PR and what has been improved with a simple before/after comparison, for example?
> 

Having a manifest parsing error while using cargo I initially assumed a mistake in the Cargo.toml,
though due to the problematic line in the error message that was quickly ruled out.
Next I tried to figure out where cargo was trying and failing to parse a manifest.
I assumed that this was maybe a manifest from the crates.io package registry cache, but searching for the line in `~/.cargo/` didn't bring up any results.
Only when I noticed rust-analyzer and `cargo --help` fail the same way did I come to the conclusion that the source is likely from the rustup wrapper of those binaries and not cargo itself and as such began searching in rustup.
This search could have been shorten had the manifest path been part of the error message, thus me not having to search for which file it was that had the error.

To recreate the error condition I inserted a NUL byte in a valid manifest file, resulting in the following error before/after my change.

Before: 

```
$> RUSTUP_HOME=home CARGO_HOME=home home/bin/cargo --help
error: error parsing manifest: TOML parse error at line 1286, column 24
     |
1286 | [[pkg.rust.target.aarch64-apple-darwin.components]]
     |                        ^
invalid table header
expected `.`, `]]`
```

After:

```
$> RUSTUP_HOME=home CARGO_HOME=home home/bin/cargo --help
error: could not parse manifest file: '/home/bennet/git/rustup/home/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/multirust-channel-manifest.toml': error parsing manifest: TOML parse error at line 1286, column 24
     |
1286 | [[pkg.rust.target.aarch64-apple-darwin.components]]
     |                        ^
invalid table header
expected `.`, `]]`
```

> PS: To make sure we are on the same page, regarding #3851 I would like to underline that the server incident causing bad manifests being downloaded is probably out of the scope of the Rustup project (so it possibly cannot be "fixed"), although improvements regarding error handling are welcomed.

Yes, I agree that the source of the manifest corruption is outside of rustup and as such out of scope. My goal with the Issue and this PR was to improve the error reporting. Also, I am pretty sure that there was no server incident involved, but I believe that the export and re-import of the WSL 2 distribution somehow corrupted the disk image file, resulting in the already installed manifest getting corrupted.

rami3l:
@Skgland Thanks for the clarification!

At Rustup the word "manifest" is used specifically to refer to a certain kind of file downloaded from the Rustup release server, and, as you've found out, named `multirust-channel-manifest.toml`.

This file is specifically used to determine what are available in this channel, however I have no idea how you got that wrong (since we have shasum checks), sorry about that...

https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/tests/suite/cli_v2.rs#L280-L300

This PR might have to wait for #3852 before we get our merge CI green again. If you'd like, in the meantime, you can add a regression test in the file above.

rami3l (approved):

Skgland:
Ok, just pushed a commit with a regression test

Skgland:
Hm, looks like my regression test fails on windows due to `\` vs. `/` in the path in the error message.

Skgland:
Do you have a suggestion for how to deal with the path-separator discrepancy on windows or should I skip the test on windows with e.g. `#[cfg(not(target_os = "windows"))]` the test?

rami3l:
> Hm, looks like my regression test fails on windows due to `\` vs. `/` in the path in the error message.

@Skgland Just use /. It's a feature...

rami3l:
> Do you have a suggestion for how to deal with the path-separator discrepancy on windows or should I skip the test on windows with e.g. `#[cfg(not(target_os = "windows"))]` the test?

@Skgland Please feel free to use what you consider appropriate for the problem. For example you can use two conditionally compiled assertions for Windows and Unix. Or you can compare the path components.

PR #3855, opened 2024-06-03
fix(download): work around `hyper` hang issue by adjusting `reqwest` config

### LINKED ISSUE #3852: Tests relying on `reqwest::Client` may fail with "dispatch task is gone: runtime dropped the dispatch task"

rami3l:
This PR tries to mitigate #3852 by applying https://github.com/Azure/azure-sdk-for-rust/pull/1550 's hack to all the `ClientBuilder`'s I can find across the repo.

The verification of the mitigation is not easy, since `async` execution is not so deterministic. Nevertheless, as this seems to be a widely encountered issue with `hyper` and `reqwest`, I suggest we merge this first and see how it goes.

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -345,6 +345,10 @@ pub mod reqwest_be {
 
     fn client_generic() -> ClientBuilder {
         Client::builder()
+            // HACK: set `pool_max_idle_per_host` to `0` to avoid an issue in the underlying
+            // `hyper` library that causes the `reqwest` client to hang in some cases.
+            // See <https://github.com/hyperium/hyper/issues/2312> for more details.
+            .pool_max_idle_per_host(0)
             .gzip(false)
             .proxy(Proxy::custom(env_proxy))
             .timeout(Duration::from_secs(30))
--- a/download/tests/read-proxy-env.rs
+++ b/download/tests/read-proxy-env.rs
@@ -61,6 +61,10 @@ async fn socks_proxy_request() {
     let url = Url::parse("http://192.168.0.1/").unwrap();
 
     let client = Client::builder()
+        // HACK: set `pool_max_idle_per_host` to `0` to avoid an issue in the underlying
+        // `hyper` library that causes the `reqwest` client to hang in some cases.
+        // See <https://github.com/hyperium/hyper/issues/2312> for more details.
+        .pool_max_idle_per_host(0)
         .proxy(Proxy::custom(env_proxy))
         .timeout(Duration::from_secs(1))
         .build()
--- a/download/tests/read-proxy-env.rs
+++ b/download/tests/read-proxy-env.rs
@@ -4,15 +4,16 @@ use std::env::{remove_var, set_var};
 use std::error::Error;
 use std::net::TcpListener;
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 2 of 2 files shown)


download/src/lib.rs
```diff
@@ -345,6 +345,10 @@ pub mod reqwest_be {
 
     fn client_generic() -> ClientBuilder {
         Client::builder()
+            // HACK: set `pool_max_idle_per_host` to `0` to avoid an issue in the underlying
```
djc, line 348: Can you explain why this is necessary/why there is no better solution for this?

rami3l: @djc The underlying issue is https://github.com/hyperium/hyper/issues/2312, see https://github.com/hyperium/hyper/issues/2312#issuecomment-778005053 for the first appearance of this hack. It is suspected that there is a hyper pool deadlock underneath, and this workaround has been used by many to successfully mitigate the issue (see all the external references below).

djc (approved):

download/src/lib.rs
djc: Do we have a good reproduction scenario?

rami3l: @djc Unfortunately not yet, and that's the tricky part of this thing I guess... If it really is a concurrency bug, then it'll be rather hard to track and to fix (that's also why that issue has been open for such a long time I guess, a previous project that I collaborated on has also hit it back in 2020, and the same hack was applied 🤔).

As far as our CI is concerned, we can see several failures in the past 5 days caused by this one, and they are all "solved" by relaunching the test task. The particular test(s) that failed varied from time to time, but I'm sure that all those tests are `reqwest`-related.

Anyway, this PR intentionally won't close #3852, and I'll keep an eye on https://github.com/hyperium/hyper/issues/2312 as well as our CI in the upcoming weeks.

rami3l: @djc The workaround seems to be working: our CI hasn't been failing for this reason in the past 5 days.

PR #3856, opened 2024-06-03
refactor: restrict `run_future()` usages to tests

rami3l:
Continuation of #3367, supersedes #3847.

This PR simply makes more function async, and thus the necessity of `run_future()` in non-test code is completely eliminated.

The only usages remaining are in `manifestation`'s tests, which is seemingly a bit harder than others considering lifetime restrictions and such.

### DIFF
--- a/src/dist/manifestation/tests.rs
+++ b/src/dist/manifestation/tests.rs
@@ -2,15 +2,16 @@
 // server (mocked on the file system)
 #![allow(clippy::type_complexity)]
 
-use std::cell::Cell;
 use std::collections::HashMap;
 use std::env;
 use std::fs;
 use std::path::Path;
 use std::str::FromStr;
 use std::sync::Arc;
+use std::{cell::Cell, future::Future};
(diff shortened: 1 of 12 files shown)


src/dist/manifestation/tests.rs
```diff
@@ -32,6 +33,24 @@ use crate::{
 
 const SHA256_HASH_LEN: usize = 64;
 
+/// Temporary thunk to support asyncifying from underneath.
+fn run_future<F, R, E>(f: F) -> Result<R, E>
```
djc, line 37: So can we guard with `#[cfg(test)]` now?

djc (approved):
Nice!

src/dist/manifestation/tests.rs
rami3l: @djc The whole module is under that cfg already, that's why there are no dead code warnings.

djc (approved):

PR #3858, opened 2024-06-04
feat: support for parsing beta versions with tags in the toolchain

### LINKED ISSUE #1329: Downloading specific betas should be easier
For Firefox, we have a script that kind of duplicates the behavior of rustup and that repacks rust for use on Firefox CI. In some cases, we need to pin a specific beta version, but failed to do so, so when 1.24.0-beta.1 replaced 1.23.0-beta.2 today, we ended up picking that, which, incidentally, broke Firefox CI because 1.24.0-beta.1 apparently breaks slog and servo.

Anyways, after some fiddling, it came out that it's possible to get a specific beta version... if you know when it was released.

That is,
```
$ rustup toolchain install beta-2017-12-08
```

installs 1.23.0-beta.2. You have to know that 1.23.0-beta.2 was released on that specific day, or scan all dates.

It would be much easier if it were possible to do:

```
$ rustup toolchain install 1.23.0-beta.2
```

which could rely on an hypothetical `https://static.rust-lang.org/dist/channel-rust-1.23.0-beta.2.toml` file.

roife:
Fix #1329.

This PR adds support for parsing toolchain names which containing beta channel and tags (like `1.23-beta`, `1.23.0-beta` or `1.23.0-beta.2`).

## Changes

- Add support for parsing beta versions with tags in the toolchain names and corresponding unit tests.
- Add scenario `BetaTag` and mock data.
- Add regression tests.

### DIFF
--- a/src/dist/dist.rs
+++ b/src/dist/dist.rs
@@ -181,7 +181,8 @@ impl FromStr for ParsedToolchainDesc {
                     "beta",
                     "stable",
                     // Allow from 1.0.0 through to 9.999.99 with optional patch version
-                    r"\d{1}\.\d{1,3}(?:\.\d{1,2})?",
+                    // and optional beta tag
+                    r"\d{1}\.\d{1,3}(?:\.\d{1,2})?(?:-beta(?:\.\d{1,2})?)?",
                 ]
                 .join("|")
             ))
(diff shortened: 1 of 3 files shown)


rami3l (approved):
Nice work 🎉

PR #3859, opened 2024-06-04
Store current_dir in Cfg

djc:
Stores the `current_dir` in `Cfg` to avoid having to pass it into a number of its methods. Eventually I think we can use this to completely get rid of `utils::current_dir()` and `Process::current_dir()` in favor of passing down the `current_dir` from the original `main()` in `rustup-init.rs`, but that last step is triggering some weird stuff that needs more investigation.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -403,10 +403,9 @@ pub(crate) fn list_toolchains(cfg: &Cfg, verbose: bool) -> Result<utils::ExitCod
         writeln!(process().stdout().lock(), "no installed toolchains")?;
     } else {
         let default_toolchain_name = cfg.get_default()?;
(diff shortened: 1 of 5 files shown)


rami3l (approved):
Looks legit, as long as this doesn't make certain tests harder to perform :)

> Eventually I think we can use this to completely get rid of `utils::current_dir()` and `Process::current_dir()` in favor of passing down the `current_dir` from the original `main()` in `rustup-init.rs`, but that last step is triggering some weird stuff that needs more investigation.

Would you mind sharing more info regarding this point?

djc:
If I make all of the changes to remove those functions, I get stack overflows from a whole bunch of tests.

PR #3860, opened 2024-06-05
Refactoring around use of `Process`

djc:
Various changes which I've made several times now that try to untangle the codebase mostly by inlining short and/or single-use methods. Preparation for moving more stuff out of the ambient `Process` global.

### DIFF
--- a/src/cli/proxy_mode.rs
+++ b/src/cli/proxy_mode.rs
@@ -1,5 +1,4 @@
-use std::{ffi::OsString, path::PathBuf};
-use std::process::ExitStatus;
+use std::{ffi::OsString, path::PathBuf, process::ExitStatus};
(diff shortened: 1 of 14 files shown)


src/cli/proxy_mode.rs
```diff
@@ -1,5 +1,5 @@
-use std::{ffi::OsString, path::PathBuf};
 use std::process::ExitStatus;
+use std::{ffi::OsString, path::PathBuf};
```
rami3l: Hmmm seems like a no-op. Can we merge these imports instead?

djc: Done.

rami3l (approved):
LGTM :)

PR #3863, opened 2024-06-07
Simplify notification handling setup

djc:
Here are some small changes that I feel make this code a little more coherent.

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -5,6 +5,7 @@ use std::io::Write;
 use std::os::windows::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 use std::process::Command;
+use std::sync::{Arc, Mutex};
 
 use anyhow::{anyhow, Context, Result};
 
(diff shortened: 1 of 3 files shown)


src/cli/self_update/windows.rs
```diff
@@ -179,14 +179,14 @@ pub(crate) async fn try_install_msvc(opts: &InstallOpts<'_>) -> Result<ContinueI
         .context("error creating temp directory")?;
 
     let visual_studio = tempdir.path().join("vs_setup.exe");
-    let download_tracker = DownloadTracker::new_with_display_progress(true);
-    download_tracker.lock().unwrap().download_finished();
+    let mut download_tracker = DownloadTracker::new_with_display_progress(true);
```
rami3l: I guess we might need to return to the original semantics (`Arc<Mutex<T>>`) at this particular place (https://stackoverflow.com/a/30560208), however this doesn't make the patch unjustified: it's still good to remove overhead at least in some parts of the codebase.

rami3l (approved):
Thanks!

src/cli/self_update/windows.rs
djc: Not exactly sure what that code is doing, it looks fishy...

PR #3864, opened 2024-06-08
Use serde to encode/decode various TOML formats

djc:
This saves a bunch of code (-190 lines) and is generally more idiomatic. It also saves about 8% of latency on proxy invocations, after (both best of 3 runs):

```
djc-2021 toml rustup $ RUSTUP_FORCE_ARG0=rustc hyperfine "./target/release/rustup-init --version"
Benchmark 1: ./target/release/rustup-init --version
  Time (mean ± σ):      32.6 ms ±   0.6 ms    [User: 23.3 ms, System: 5.5 ms]
  Range (min … max):    31.4 ms …  34.6 ms    86 runs
```

before:

```
djc-2021 master rustup $ RUSTUP_FORCE_ARG0=rustc hyperfine "./target/release/rustup-init --version"
Benchmark 1: ./target/release/rustup-init --version
  Time (mean ± σ):      35.3 ms ±   0.6 ms    [User: 25.6 ms, System: 5.8 ms]
  Range (min … max):    33.6 ms …  36.6 ms    79 runs
```


src/config.rs
```diff
@@ -490,11 +485,14 @@ impl Cfg {
                 }
 
                 self.settings_file.with_mut(|s| {
-                    DEFAULT_METADATA_VERSION.clone_into(&mut s.version);
+                    s.version = MetadataVersion::default();
                     Ok(())
                 })
             }
-            _ => Err(RustupError::UnknownMetadataVersion(current_version).into()),
+            MetadataVersion::V12 => {
```
rami3l: To me it's clear that the happy path in the original code is for `MetadataVersion::default()`, but now it's for `MetadataVersion::V12`. I know at the moment being it's all the same, but the meaning being conveyed seems to be a bit different, and that might cause problems when the default value changes 🤔

src/settings.rs
```diff
@@ -29,9 +27,13 @@ impl SettingsFile {
         }
     }
 
-    fn write_settings(&self) -> Result<()> {
-        let s = self.cache.borrow().as_ref().unwrap().clone();
-        utils::write_file("settings", &self.path, &s.stringify())?;
+    fn write_settings(&self) -> anyhow::Result<()> {
```
rami3l: Hmmm looks like we're already using `anyhow::Result` as `Result` earlier in the file? Same for 2 other places...

src/dist/config.rs
```diff
@@ -1,87 +1,56 @@
...
-        if !components.is_empty() {
-            result.insert("components".to_owned(), toml::Value::Array(components));
-        }
-        result
-    }
-
-    pub(crate) fn parse(data: &str) -> Result<Self> {
-        let value = toml::from_str(data).context("error parsing config")?;
-        Self::from_toml(value, "")
+    pub(crate) fn parse(data: &str) -> anyhow::Result<Self> {
```
rami3l: https://github.com/rust-lang/rustup/pull/3864/files#r1632151148

src/dist/config.rs
```diff
@@ -1,87 +1,56 @@
...
+        toml::from_str(data).context("error parsing config")
     }
 
-    pub(crate) fn stringify(self) -> String {
-        self.into_toml().to_string()
+    pub(crate) fn stringify(&self) -> anyhow::Result<String> {
```
rami3l: https://github.com/rust-lang/rustup/pull/3864/files#r1632151148

rami3l (approved):
Nice improvements overall! Somehow I thought we were using `serde` for such things already 🤣

src/config.rs
djc: I've reverted the logic to something closer to what was originally there.

PR #3865, opened 2024-06-09
Simplify installation code

djc:
Started with the first commit from #3766 and tried to improve the installation code a bit.


rami3l (approved):
Nice :D

PR #3868, opened 2024-06-12
refactor(test): execute all `#[rustup_macros::unit_test]`s within a `tokio` context

rami3l:
Split from #3803.

With this PR, all tests under the `#[rustup_macros::unit_test]` attribute now execute within a `tokio` runtime, eliminating the need of static `TRACER` and `TRACE_RUNTIME` variables. Also, the setup of `tracing` subscribers in both `bin` and `test` targets has also been unified, to prepare for the addition of another `stderr`-oriented subscriber in #3803.

## Rationale

Currently, the `tracing` subscriber registration will fail if a tokio runtime is not present. To mitigate this problem, static `TRACER` and `TRACE_RUNTIME` variables are added as an ad-hoc fallback if we're running a non-tokio unit test.

https://github.com/rust-lang/rustup/blob/3ba08da98221b2941aeb858e54cebc059859ffb7/src/test.rs#L230-L237

However, the fact that the subscriber state is **global** in tests has made some unnecessary complications especially WRT adding other `tracing` subscribers, such as the `stderr`-oriented tracing subscriber proposed in https://github.com/rust-lang/rustup/pull/3803#issuecomment-2090333256, which is bound to the **local** `currentprocess:process()` context (see https://github.com/rust-lang/rustup/pull/3803#issuecomment-2143703392).

I decided to go with local contexts all the way, and it did work. However, the main downside to this approach is that #2367 has to be reverted, since `run_inprocess()` calls `with_runtime()`, and for the latter to correctly alter `static PROCESS` in-process, a `tokio` runtime builder is required, which in this case will cause a "runtime-in-runtime" error once the builder is run.

Fortunately, no obvious slowdown has been observed WRT our CI so far. Also, as now `main()` seems to be the only caller to `with_runtime()`, this gets us closer to using `#[tokio::main]`.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -224,69 +224,32 @@ where
     f(&rustup_home)
 }
 
-#[cfg(feature = "otel")]
-use once_cell::sync::Lazy;
-
-/// A tokio runtime for the sync tests, permitting the use of tracing. This is
-/// never shutdown, instead it is just dropped at end of process.
-#[cfg(feature = "otel")]
-static TRACE_RUNTIME: Lazy<tokio::runtime::Runtime> =
-    Lazy::new(|| tokio::runtime::Runtime::new().unwrap());
-/// A tracer for the tests.
-#[cfg(feature = "otel")]
-static TRACER: Lazy<opentelemetry_sdk::trace::Tracer> = Lazy::new(|| {
-    use std::time::Duration;
-
-    use opentelemetry::{global, KeyValue};
-    use opentelemetry_otlp::WithExportConfig;
-    use opentelemetry_sdk::{
-        propagation::TraceContextPropagator,
-        trace::{self, Sampler},
-        Resource,
-    };
-    use tokio::runtime::Handle;
-    use tracing_subscriber::{layer::SubscriberExt, EnvFilter, Registry};
-
-    // Use the current runtime, or the sync test runtime otherwise.
-    let handle = match Handle::try_current() {
-        Ok(handle) => handle,
-        Err(_) => TRACE_RUNTIME.handle().clone(),
-    };
-    let _guard = handle.enter();
-
-    let tracer = opentelemetry_otlp::new_pipeline()
-        .tracing()
-        .with_exporter(
-            opentelemetry_otlp::new_exporter()
-                .tonic()
-                .with_timeout(Duration::from_secs(3)),
-        )
-        .with_trace_config(
-            trace::config()
-                .with_sampler(Sampler::AlwaysOn)
-                .with_resource(Resource::new(vec![KeyValue::new("service.name", "rustup")])),
-        )
-        .install_batch(opentelemetry_sdk::runtime::Tokio)
-        .unwrap();
-
-    global::set_text_map_propagator(TraceContextPropagator::new());
-    let env_filter = EnvFilter::try_from_default_env().unwrap_or(EnvFilter::new("INFO"));
-    let telemetry = tracing_opentelemetry::layer().with_tracer(tracer.clone());
-    let subscriber = Registry::default().with(env_filter).with(telemetry);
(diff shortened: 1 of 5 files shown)


rami3l:
Oops, another global state problem regarding `currentprocess::process()`... I have to be more careful regarding this one.

src/test.rs
```diff
@@ -224,85 +224,34 @@ where
     f(&rustup_home)
 }
 
-#[cfg(feature = "otel")]
-use once_cell::sync::Lazy;
+pub async fn before_test_async() -> Option<tracing::dispatcher::DefaultGuard> {
```
rami3l, line 227: This signature is temporary. In #3803, this will return a `tracing::dispatcher::DefaultGuard` directly.

src/test/mock/clitools.rs
```diff
@@ -703,54 +694,6 @@ impl Config {
         output
     }
 
-    #[cfg_attr(feature = "otel", tracing::instrument(skip_all))]
-    pub(crate) fn run_inprocess<I, A>(&self, name: &str, args: I, env: &[(&str, &str)]) -> Output
```
djc, line 707: Why is this necessary, and what is the effect on test suite execution time?

rami3l: @djc This has been explained in the final part of the post (https://github.com/rust-lang/rustup/pull/3868#issue-2347638914). I have to admit, it is a trade-off that I feel a bit uneasy to make.

rami3l: @djc I mean if we can get rid of `static PROCESS` in the future, then there should be a way to revive #2367.

djc: Okay, I've been working on what I think might be an alternative path today. Let me push that up in a bit so we can decide which way to go.

rami3l: @djc The thing is that I'm not quite sure about the goal #2367 was trying to reach. Is it for perf optimization? If it is, then the positive impact does not seem that large, at least so far, and it seems to me that it's adding complexity with the dance altering `static`s. Is it for something else? The more I change this part without knowing the original intent the more afraid I am of accidentally breaking something in the process: that's why I cc'd @rbtcollins.

OTOH I do have a feeling that the overall Rust testing ecosystem as it stands is hostile to injected global states (they like to use them directly/implicitly), so when working on #3803, I'm constantly thinking maybe we should get rid of that part altogether. Maybe that's too radical though?

djc: My alternative is in #3871.

rami3l: @djc Sorry, I'm a bit occupied today, but I assure you that I will find some time to have a closer look real soon.

djc: No worries!

src/test.rs
```diff
@@ -224,85 +224,34 @@ where
...
-        )
-        .with_trace_config(
-            trace::config()
-                .with_sampler(Sampler::AlwaysOn)
-                .with_resource(Resource::new(vec![KeyValue::new("service.name", "rustup")])),
-        )
-        .install_batch(opentelemetry_sdk::runtime::Tokio)
-        .unwrap();
+            global::set_text_map_propagator(TraceContextPropagator::new());
+            telemetry(telemetry_default_tracer().unwrap())
```
rami3l: Maybe this can be inlined into `telemetry().unwrap()` or even a lazy static `TELEMETRY`?

djc (approved):

src/test.rs
rami3l: The latter is not possible, since the output type of `telemetry()` must be generic (`impl Layer<S>`).
However, it seems that `telemetry_default_tracer` can be made static and hidden from `telemetry()`'s parameter list.

PR #3869, opened 2024-06-12
Fix misleading "uninstalled toolchain" notification

### LINKED ISSUE #3861: `rustup toolchain uninstall <toolchain>` reports toolchain has been uninstalled without doing anything
### Verification

### Problem

A `rustup toolchain uninstall <toolchain>` invocation reports that the `<toolchain>` has been uninstalled even if there is no such toolchain present in the first place.

### Steps

1. Install rustup.
2. Run `rustup toolchain uninstall <toolchain>` with literally anything in place of `<toolchain>`, "foo", 42, the smiley emoji (😄).
3. Observe the output
   ```
   rustup toolchain uninstall 😄
   info: no toolchain installed for '😄'
   info: toolchain '😄' uninstalled
   ```
Even though rustup recognizes that there is no such toolchain, it still reports that it has been uninstalled.

Now it's of course not a big issue in a silly example like above, but gives the false impression of removing an actual toolchain without doing so. For example,
```
> rustup toolchain install 1.70.0
> rustup toolchain list
```
Which shows
```
```
So naturally you run `rustup toolchain uninstall 1.70.0-x86_64-unknown-linux-gnu` and you get
```
info: no toolchain installed for '1.70.0-x86_64-unkown-linux-gnu'
info: toolchain '1.70.0-x86_64-unkown-linux-gnu' uninstalled
```
If you're not careful while reading, and miss that "no," looks like the toolchain has indeed been removed.

### Possible Solution(s)

This [unconditional check](https://github.com/rust-lang/rustup/blob/84495dc26c9e3d8557bed5c93d6aeaf448202631/src/toolchain/toolchain.rs#L361-L363) for the non-existence of the path even after possibly finding that the [toolchain doesn't exist](https://github.com/rust-lang/rustup/blob/84495dc26c9e3d8557bed5c93d6aeaf448202631/src/toolchain/toolchain.rs#L355-L357) may be the root cause.

### Notes

_No response_

### Rustup version

```console
```

### Installed toolchains

```console
rustup home:  /home/scimas/.rustup

aarch64-unknown-linux-gnu
wasm32-unknown-unknown
x86_64-unknown-linux-gnu

active toolchain
----------------

rustc 1.78.0 (9b00956e5 2024-04-29)
```

### OS version

```console
Linux fedora 6.8.10-200.fc39.x86_64
```

scimas:
Added a check to see if any filesystem modification was done. Fixes #3861.

### DIFF
--- a/src/toolchain/toolchain.rs
+++ b/src/toolchain/toolchain.rs
@@ -447,7 +447,7 @@ impl<'a> Toolchain<'a> {
             LocalToolchainName::Named(t) => t,
             LocalToolchainName::Path(_) => bail!("Cannot remove a path based toolchain"),
         };
-        match Self::exists(cfg, &(&name).into())? {
+        let fs_modified = match Self::exists(cfg, &(&name).into())? {
             true => {
                 (cfg.notify_handler)(Notification::UninstallingToolchain(&name));
                 let installed_paths = match &name {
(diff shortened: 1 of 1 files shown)


djc (approved):
Seems reasonable, thanks!

rami3l (approved):

PR #3871, opened 2024-06-12
Pass Process around explicitly

djc:
> OTOH I do have a feeling that the overall Rust testing ecosystem as it stands is hostile to global states/injections, so when working on https://github.com/rust-lang/rustup/pull/3803, I'm constantly thinking maybe we should get rid of that part altogether. Maybe that's too radical though?
_#3868_

I think that's kind of what this is? Doesn't seem too radical to me.

I haven't quite finished the tests and the I'm not sure yet what to do about the logging macros, but this seems viable (and should enable ripping out a bunch of stuff like `with()` and `with_runtime()` etc).

### DIFF
--- a/src/currentprocess.rs
+++ b/src/currentprocess.rs
@@ -15,8 +15,6 @@ use std::{
     sync::{Arc, Mutex},
 };
 
-#[cfg(feature = "test")]
-use rand::{thread_rng, Rng};
 use tracing_subscriber::util::SubscriberInitExt;
 
 pub mod filesource;
@@ -287,7 +285,6 @@ pub struct TestProcess {
     pub cwd: PathBuf,
     pub args: Vec<String>,
     pub vars: HashMap<String, String>,
-    pub id: u64,
     pub stdin: filesource::TestStdinInner,
     pub stdout: filesource::TestWriterInner,
     pub stderr: filesource::TestWriterInner,
@@ -305,18 +302,10 @@ impl TestProcess {
             cwd: cwd.as_ref().to_path_buf(),
             args: args.iter().map(|s| s.as_ref().to_string()).collect(),
             vars,
-            id: TestProcess::new_id(),
             stdin: Arc::new(Mutex::new(Cursor::new(stdin.to_string()))),
             stdout: Arc::new(Mutex::new(Vec::new())),
             stderr: Arc::new(Mutex::new(Vec::new())),
         }
-    }
-
-    fn new_id() -> u64 {
-        let low_bits: u64 = std::process::id() as u64;
-        let mut rng = thread_rng();
-        let high_bits = rng.gen_range(0..u32::MAX) as u64;
-        high_bits << 32 | low_bits
     }
 
     /// Extracts the stdout from the process
--- a/src/currentprocess.rs
+++ b/src/currentprocess.rs
@@ -15,6 +15,8 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+#[cfg(feature = "test")]
+use tracing::subscriber::DefaultGuard;
 use tracing_subscriber::util::SubscriberInitExt;
 
 pub mod filesource;
@@ -25,7 +27,7 @@ pub mod terminalsource;
 pub enum Process {
     OSProcess(OSProcess),
     #[cfg(feature = "test")]
-    TestProcess(TestProcess),
+    TestProcess(TestContext),
 }
 
 impl Process {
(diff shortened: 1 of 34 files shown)


djc:
Here the only place that still calls `process()` is in `cli::log`. Presumably we'll want to replace those soon with tracing macros anyway, so not a big deal? I think if we stack your changes from #3868 on top of this things will be much better.

rami3l (commented):
@djc Thanks for the work! Looks like 84e01b5dbde9204b5f936e3db0ff092b8b69003e is a rather mechanical change (replacing `process()` with `process`)? If that is the case, I think we can first merge the 2 commits before it.

Having noticed the fact that effectively `PROCESS` is only altered in `with()` and `with_runtime()`, both of which are used only in tests (except one single mention of `with_runtime()` in `rustup-init.rs`), my overall plan is that:
1. #3868 temporarily removes in-process tests, now subscribers can be totally local.
2. #3803 sets a global subscriber based on the current `static PROCESS` in normal business logic, and for tests it sets a local subscriber using `let _guard = tracing::subscriber::set_default(subscriber);`, where `subscriber` is a function of the current `PROCESS`. Clearly, the only injection points of the local subscriber would be inside the `with*()` functions (a new `PROCESS` will naturally require a new `stderr`-oriented subscriber). The `log` module now relies on the implicit subscriber in the current scope (global or local).
3. With this PR, `PROCESS` can be safely removed, since the `info!()` calls and friends only implicitly rely on the subscriber. The subscriber's dependency on the new `process` in this case can be specified elsewhere (globally, or locally using `with()`). Now `with()`'s only job is to install the hook and then change the current subscriber when the closure runs. since there's no `PROCESS` anymore and thus no more calls to `.on_thread_start()` and stuff to alter it, `with_runtime()` is no longer required. [^1] An async version of `with()` might still be desired though, but it won't have to depend on tokio runtime builder.
4. #3891 revives in-process tests using just `with()` or its async counterpart.

[^1]: I'm not sure about the parameters to be passed to the topmost tokio runtime builder yet. IIRC in `with_runtime()` 2 workers are used, in `#[test]` there's only one.

What do you think?

djc:
Sounds like a good direction! I'm not completely clear on which order you want to merge things?

> If that is the case, I think we can first merge the 2 commits before it.

I've submitted them separately as #3872. Once that is merged, I can rebase this.

rami3l:
@djc Looks like I have found a way out of the `tracing` struggle, but I'm not entirely sure if it will work. The only way to know, as before, is to actually write the patches :) How do you want to see the 4 consecutive changes above, in separate PRs (so if I end up failing we'll have to revert them) or in a big one (so if I end up failing nothing will happen, but it will be potentially more difficult to maintain the commit stack)?

rami3l:
> Sounds like a good direction! I'm not completely clear on which order you want to merge things?

@djc Just the top-down order from 1 to 4. It's a back-and-forth dance around #2367, but the latter is a small change and I believe we can afford reverting and then reviving it.

PS: Just in case it's not clear:
- (2.) relies on subscribers being made possibly local in (1.)
- (3.) relies on implicit `log` subscribers from (2.)
- (4.) relies on the removal of `with_runtime()` in (3.)

It's a bit late now here so I'll probably continue my work tomorrow, by polishing (1.) (i.e. #3868) first.

src/currentprocess.rs
```diff
@@ -140,37 +142,8 @@ impl home::env::Env for Process {
...
-        }
-        let _guard = crate::cli::log::tracing_subscriber(&process).set_default();
-        *p.borrow_mut() = Some(process);
-        let result = f();
-        *p.borrow_mut() = None;
-        result
-    })
-}
-
 fn ensure_hook() {
```
rami3l, line 147: A question: maybe I've missed it, but is `ensure_hook()` called anywhere? Somehow my search box didn't catch it 🤔

src/currentprocess.rs
```diff
@@ -279,55 +252,76 @@ impl Default for OSProcess {
...
 pub struct TestProcess {
-    pub cwd: PathBuf,
-    args: Vec<String>,
-    vars: HashMap<String, String>,
-    stdin: filesource::TestStdinInner,
-    stdout: filesource::TestWriterInner,
-    stderr: filesource::TestWriterInner,
+    pub process: Process,
+    #[allow(dead_code)] // guard is dropped at the end of the test
+    guard: DefaultGuard,
```
rami3l, line 260: Nit: I guess the linter actually wants you to write `_guard`, see https://github.com/search?q=org%3Arust-lang%20_guard%3A&type=code for more examples across the org.

rami3l (approved):
Thanks a lot for the patch! I've left some minor questions/suggestions, but great improvement overall :D

src/currentprocess.rs
djc: This PR doesn't get rid of `with_runtime()` yet, which uses it. Next on the list!

rami3l: > This PR doesn't get rid of `with_runtime()` yet, which uses it. Next on the list!

~~@djc Do you think it's still valuable to call it though? To be frank I'm not sure what this function is doing, just being curious...~~

Ah I'm back in front of my IDE and it's immediately clear: it has something to do with `PROCESS` XD

PR #3873, opened 2024-06-14
feat(log): unhide `tracing::instrument` from behind `feature = "otel"`

rami3l:
Cherry-picked from #3803 as discussed in https://github.com/rust-lang/rustup/pull/3803#issuecomment-2168125917.

## Note

This is done by simply replacing `cfg_attr\(feature = \"otel\", tracing::instrument(\((.*)\))?\)\]` with `tracing::instrument(level = "trace", $2)]`, then replacing `, \)\]` with `)]`.

### DIFF
--- a/doc/dev-guide/src/tracing.md
+++ b/doc/dev-guide/src/tracing.md
@@ -101,7 +101,7 @@ when enabled. Instrumenting a currently uninstrumented function is mostly simply
 done like so:
 
 ```rust
-#[tracing::instrument(level = "trace", err, skip_all)]
+#[tracing::instrument(level = "trace", err(level = "trace"), skip_all)]
 ```
 
 `skip_all` is not required, but some core structs don't implement Debug yet, and
--- a/doc/dev-guide/src/tracing.md
+++ b/doc/dev-guide/src/tracing.md
@@ -101,7 +101,7 @@ when enabled. Instrumenting a currently uninstrumented function is mostly simply
 done like so:
 
 ```rust
-#[cfg_attr(feature = "otel", tracing::instrument(err, skip_all))]
+#[tracing::instrument(level = "trace", err, skip_all)]
 ```
 
 `skip_all` is not required, but some core structs don't implement Debug yet, and
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,7 @@ jobs:
     if: ${{ contains('["pull_request", "merge_group"]', github.event_name) }} # skip-master skip-stable
     env:
       RUSTFLAGS: -Ctarget-feature=+crt-static
+      RUST_MIN_STACK: 16777216
     permissions:
       id-token: write
       contents: read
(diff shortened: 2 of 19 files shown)


rami3l:
The Windows CI is still red, but only a small number of tests are failing due to stack overflow compared to #3803's CI.

@roife has pointed me to the stack size restriction on Windows, saying that maybe it's not a stack overflow caused by infinite recursion, instead we might simply need a bigger stack. `rust-analyzer` has some code related to this restriction: https://github.com/rust-lang/rust-analyzer/blob/6b8b8ff4c56118ddee6c531cde06add1aad4a6af/crates/rust-analyzer/src/bin/main.rs#L147-L163

ChrisDenton:
On Windows, applications choose their stack size rather than being defined by the system. So another option is to increase the default stack size by using a linker argument (for msvc that's [`/stack`](https://learn.microsoft.com/en-us/cpp/build/reference/stack-stack-allocations?view=msvc-170)). I don't know what's easier.

rami3l:
@ChrisDenton Thanks a lot for your pointers!

Looks like unoptimized Rust is indeed hungry on stack space:

> The Rust compiler uses a 6MiB stack and then increase than to 20MiB (!!!) when optimizations are disabled. The default stack size on Linux is 8MiB, so that's likely how much is available for the main thread. Rust tends to need a ridiculous amount of stack space, especially without optimizations.
_https://github.com/rust-lang/rust/issues/17044#issuecomment-54713907_

I'll try your suggestion in https://users.rust-lang.org/t/stack-overflow-when-compiling-on-windows-10/50818/8 and https://github.com/rust-lang/rust/issues/85303 first.

ChrisDenton:
Oh, hm actually I think the test runner sets the stack size itself. I'm not sure if that is directly configurable. You might need to set the `RUST_MIN_STACK` environment variable.

rami3l:
> Oh, hm actually I think the test runner sets the stack size itself. I'm not sure if that is directly configurable. You might need to set the `RUST_MIN_STACK` environment variable.

@ChrisDenton Thanks a lot! That seems to work :)

PS: Sorry I got a bit sidetracked with other issues in the past few weeks 🙈

doc/dev-guide/src/tracing.md
```diff
@@ -101,7 +101,7 @@ when enabled. Instrumenting a currently uninstrumented function is mostly simply
 done like so:
 
 ```rust
-#[tracing::instrument(level = "trace", err, skip_all)]
+#[tracing::instrument(level = "trace", err(level = "trace"), skip_all)]
```
djc, line 104: What does this mean/do?

rami3l: This is for disabling error tracking for these functions unless the level is set to "TRACE". The default level is "ERROR" (see https://docs.rs/tracing/latest/tracing/attr.instrument.html#examples-2) which is way too high, and will interfere with the default logger: https://github.com/rust-lang/rustup/actions/runs/9867247214/job/27247322849

One thing to note is that the current "legacy" logging formatter has ignored spans altogether, this is okay for now (spans are not included in the original logger anyway), but might be worth fixing as well if we're moving towards a universal "--verbose" flag, probably in another PR. What truly matters now is that running `rustup` with `RUSTUP_LOG=rustup=TRACE` does emit the right spans and their associated errors (when they do occur).

djc (approved):

PR #3874, opened 2024-06-15
Let argument parser handle enum conversions

djc:


src/dist/dist.rs
```diff
@@ -611,6 +613,42 @@ pub enum Profile {
     Complete,
 }
 
+impl Profile {
+    pub(crate) fn names(sep: &str) -> String {
```
rami3l: You can also use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join for this like in https://github.com/rust-lang/rustup/pull/3874#discussion_r1641180721. 

Also, maybe it's a good idea to unify the names of both methods for the sake of coherence (e.g. `Self::join`), or, maybe we can just inline it if it's just a single-use method that can be reduced to a single line?

src/cli/self_update.rs
```diff
@@ -100,21 +102,48 @@ pub(crate) const NEVER_SELF_UPDATE: bool = true;
...
+    pub(crate) fn join(sep: &str) -> String {
+        let mut buf = String::new();
+        for mode in SelfUpdateMode::value_variants() {
+            if !buf.is_empty() {
+                buf.push_str(sep);
+            }
+            buf.push_str(mode.as_str());
+        }
+
+        buf
```
rami3l: You can use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join for this if you'd prefer, seems easier to parse to me :)

```suggestion
        Self::value_variants().iter().join(sep)
```

src/config.rs
```diff
@@ -425,11 +415,10 @@ impl Cfg {
 
     pub(crate) fn get_self_update_mode(&self) -> Result<SelfUpdateMode> {
         self.settings_file.with(|s| {
-            let mode = match &s.auto_self_update {
-                Some(mode) => mode.clone(),
+            Ok(match s.auto_self_update {
```
rami3l, line 413: Nit: How about using one of the `.unwrap_or*` methods here?

rami3l (approved):
This is exactly the kind of improvement I'd like to see after migrating to `clap_derive`, many thanks!

src/dist/dist.rs
djc: Fair, changed it!

src/cli/self_update.rs
djc: Good call, changed it.

src/config.rs
djc: Yeah, much better!

PR #3876, opened 2024-06-16
fix(log): use `RUSTUP_LOG` for internal `tracing` instead of `RUST_LOG`

### LINKED ISSUE #3875: Reserve `RUSTUP_LOG` for `tracing` usages instead of `RUST_LOG`

@weihanglo has raised concerns regarding `RUST_LOG` adding Rustup's log lines to `cargo run` when `cargo` is a `rustup` shim.

Cargo has changed to using `CARGO_LOG` for `cargo`-internal log lines for clarity in https://github.com/rust-lang/cargo/pull/6918, and considering the special relationship between the two projects, I think it's reasonable for us to do the same.

### Solution you'd like

Reserve `RUSTUP_LOG` for `tracing` usages inside of Rustup instead of the default `RUST_LOG`.

### Notes

_No response_

rami3l:
Closes #3875, plus some random trivial cleanups along the way.

### DIFF
--- a/src/currentprocess.rs
+++ b/src/currentprocess.rs
@@ -272,8 +272,8 @@ impl TestProcess {
             args: args.iter().map(|s| s.as_ref().to_string()).collect(),
             vars,
             stdin: Arc::new(Mutex::new(Cursor::new(stdin.to_string()))),
-            stdout: Arc::new(Mutex::new(Vec::new())),
-            stderr: Arc::new(Mutex::new(Vec::new())),
+            stdout: Arc::default(),
+            stderr: Arc::default(),
         })
     }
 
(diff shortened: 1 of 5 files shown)


src/currentprocess.rs
```diff
@@ -256,8 +256,7 @@ impl Default for OSProcess {
 #[cfg(feature = "test")]
 pub struct TestProcess {
     pub process: Process,
-    #[allow(dead_code)] // guard is dropped at the end of the test
-    guard: DefaultGuard,
+    _guard: DefaultGuard, // guard is dropped at the end of the test
```
rami3l, line 259: This is https://github.com/rust-lang/rustup/pull/3871#discussion_r1641254390.

djc: Yeah -- sorry, forgot to fix that.

djc (approved):

src/currentprocess.rs
rami3l: @djc No worries :)

PR #3877, opened 2024-06-16
Process cleanup

djc:
This is from last night. I think the custom test macros are unnecessary now, please review!


src/bin/rustup-init.rs
```diff
@@ -33,24 +32,19 @@ use rustup::errors::RustupError;
...
-        .block_on(async move {
-            match maybe_trace_rustup(&process).await {
-                Err(e) => {
-                    common::report_error(&e, &process);
-                    std::process::exit(1);
-                }
-                Ok(utils::ExitCode(c)) => std::process::exit(c),
-            }
-        });
+    match maybe_trace_rustup(&process).await {
```
rami3l: Just found out that we're already tracing 100% of the time so the function name makes no sense now. Given that it's yet another single-use function, I'm inclined to inlining it here :)

rami3l (approved):
Wow that's soooo impressive, friend!

Rustup is finally an "ordinary" `tokio` app now thanks to your work! Hooray 🎉

djc:
> Rustup is finally an "ordinary" `tokio` app now thanks to your work! Hooray 🎉

**Our** work, couldn't have done it without you!

src/bin/rustup-init.rs
djc: Done.

rami3l:
@djc Let's don't forget there's one remaining concern from https://github.com/rust-lang/rustup/pull/3847#issuecomment-2139584134: there are places where the original code was using `#[::tokio::test(flavor = "multi_thread", worker_threads = 1)]` (you just deleted these lines in this PR), while `worker_threads = 2` was used  in `with_runtime()` instead:

https://github.com/rust-lang/rustup/blob/0aa7a2a0511e7d3d731c77cb551831df84009ec9/src/test/mock/clitools.rs#L729-L738

I'm not quite sure about the exact intention of those configurations, but it's better to be watchful I think.

PR #3878, opened 2024-06-16
Use tracing macros directly

djc:


src/cli/log.rs
```diff
@@ -18,26 +18,6 @@ use opentelemetry_sdk::trace::Tracer;
 
 use crate::{currentprocess::Process, utils::notify::NotificationLevel};
 
-macro_rules! debug {
-    ( $ ( $ arg : tt ) * ) => ( ::tracing::trace ! ( $ ( $ arg ) * )  )
-}
-
-macro_rules! verbose {
-    ( $ ( $ arg : tt ) * ) => ( ::tracing::debug ! ( $ ( $ arg ) * )  )
-}
```
rami3l, line 27: The nomenclature trap is here. Every time `debug!()` is used, you should **replace it with `tracing::trace!()`**; every time `verbose!()` is used, you should replace it with `tracing::debug!()`. Simply importing `tracing::debug` is actually breaking the original semantics. Fortunately, a total search-replace in that exact order should be helpful.

djc: Thanks for the hint!

rami3l (approved):

PR #3879, opened 2024-06-16
docs(dev-guide/tracing): mention `RUSTUP_LOG` and console-based tracing

rami3l:
Split from #3803.

### DIFF
--- a/doc/dev-guide/src/linting.md
+++ b/doc/dev-guide/src/linting.md
@@ -1,37 +1,62 @@
 # Linting
 
-We use `cargo clippy` to ensure high-quality code and to enforce a set of best practices for Rust programming. However, not all lints provided by `cargo clippy` are relevant or applicable to our project.
+We use `cargo clippy` to ensure high-quality code and to enforce a set of best practices for Rust programming.
+However, not all lints provided by `cargo clippy` are relevant or applicable to our project.
 We may choose to ignore some lints if they are unstable, experimental, or specific to our project.
 If you are unsure about a lint, please ask us in the [rustup Discord channel](https://discord.com/channels/442252698964721669/463480252723888159).
 
 ## Manual linting
 
-When checking the codebase with [`clippy`](https://doc.rust-lang.org/stable/clippy/index.html), it is recommended to use:
+When checking the codebase with [`clippy`](https://doc.rust-lang.org/stable/clippy/index.html),
+it is recommended to use the following command:
 
 ```console
 $ cargo clippy --all --all-targets --all-features -- -D warnings
 ```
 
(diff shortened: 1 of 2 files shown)


doc/dev-guide/src/linting.md
```diff
@@ -14,24 +14,21 @@ $ cargo clippy --all --all-targets --all-features -- -D warnings
 
 ## Rust-Analyzer
 
-When using  [`rust-analyzer`](https://rust-analyzer.github.io/) integration in the IDE of your choice, you might want to set the `rust-analyzer.cargo.features` configuration to `"all"` (check the [`rust-analyzer` manual](https://rust-analyzer.github.io/manual.html#configuration) for more details).
+When using [`rust-analyzer`](https://rust-analyzer.github.io/)'s clippy integration in the IDE of your choice, apart from setting the `rust-analyzer.check.command` property to `"clippy"`, you might also want to set the `rust-analyzer.cargo.features` property to `"all"` (check the [`rust-analyzer` manual](https://rust-analyzer.github.io/manual.html#configuration) for more details).
```
djc: Pre-existing, but if you're goint to improve this:

* Setting the `check.command` is entirely independent of setting `features` to `all`, since the latter is useful even if you don't change the `command`.
* Would use a separate sentence for the pointer to Rust-Analyzer docs instead of wrapping it in parentheses.

doc/dev-guide/src/tracing.md
```diff
@@ -1,16 +1,86 @@
 # Tracing
 
-The feature "otel" can be used when building rustup to turn on Opentelemetry
-tracing with an OLTP GRPC exporter.
+Similar to other tools in the Rust ecosystem like Rustc and Cargo,
```
djc: I would spell "rustc" and "rustup" without capitals.

doc/dev-guide/src/tracing.md
```diff
@@ -1,16 +1,86 @@
...
-`protoc` must be installed, which can be downloaded from GitHub or installed via  package manager.
+2024-06-16T12:12:45.693350Z  INFO rustup::cli::rustup_mode: cleaning up downloads & tmp directories
+```
+
+### Caveats
+
+Due to historical reasons, the names of logging levels in Rustup's "legacy" `stderr`
+output and the names being used in the `tracing` crate are slightly different:
+
+| Legacy     | `tracing` |
```
djc: I think we should fix the "legacy" prefixes to align with the current/well-known names.

rami3l: @djc Split into #3890.

djc (approved):

PR #3880, opened 2024-06-16
Simplify toolchain construction

djc:
Here's another grab bag of refactoring that tries to improve cohesion in the system, oriented around how `Toolchain`s get created.


src/config.rs
```diff
@@ -708,6 +708,20 @@ impl<'a> Cfg<'a> {
         }
     }
 
+    #[cfg_attr(feature = "otel", tracing::instrument)]
+    pub(crate) async fn rustc_version(&mut self) -> Result<String> {
```
rami3l: I got a bit confused between `Cfg::rustc_version` and `Toolchain::rustc_version` when reviewing this and #3881 at the same time (x_x)

Maybe rename it to `Cfg::active_rustc_version` for clarity?

src/toolchain.rs
```diff
@@ -1,4 +1,474 @@
...
+        let mut cmd = Command::new(path);
+        self.set_env(&mut cmd);
+        Ok(cmd)
+    }
+
+    #[cfg(not(windows))]
+    pub(crate) fn man_path(&self) -> PathBuf {
+        let mut buf = PathBuf::from(&self.path);
+        buf.push("share");
+        buf.push("man");
```
rami3l: A tiny nit...

```suggestion
        buf.extend(["man", "share"]);
```

src/toolchain/names.rs
```diff
@@ -248,7 +248,7 @@ impl Display for MaybeOfficialToolchainName {
 /// ToolchainName can be used in calls to Cfg that alter configuration,
 /// like setting overrides, or that depend on configuration, like calculating
 /// the toolchain directory.
-#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
```
rami3l, line 251: While indeed the derived `Ord` instance are not used, I guess I was doing something idiomatic here (not entirely my fault, the function has been in existence way before #3518):

https://github.com/rust-lang/rustup/blob/f88394cbbbcc25ffff8e0a0fd465ae77ef22ccbf/src/toolchain/names.rs#L283-L306

How about making use of that to write an actual `Ord` instance?

djc: I looked at that but decided to postpone it because I think a `sort()` call that calls a function like this internally will be more expensive? The `.to_string()` call makes this a relatively expensive sorting key, and a `PartialOrd` or `Ord` implementation willl call it twice for each comparison made, whereas `sort_by_key()` can slightly optimize that (although we probably should be using `sort_by_cached_key()` here -- maybe using that directly with moving the key function into a `ToolchainName::sort_key()` method would be a good trade-off for now?).

So I think it would be okay to punt on that for this PR.

rami3l: @djc I agree, it's better if this function is not used in an `Ord` implementation, and something like `Ordering::then_with` should be used instead. Please feel free to move on without this point while I have a closer look into what should be done instead...

rami3l (approved):

src/toolchain/names.rs
rami3l: Split into #3884.

src/config.rs
djc: Fair, done.

src/toolchain.rs
djc: Done.

PR #3882, opened 2024-06-17
refactor(dist): hoist `dist::dist` into `dist`

rami3l:
This is a natural step in the spirit of #3880.

## Concerns

- [x] ~~It's time to decide between `foo.rs` and `foo/mod.rs`. (I'm slightly inclined towards `foo.rs`)~~ Split into #3883.


src/dist/mod.rs
```diff
@@ -1,17 +1,1339 @@
 //! Installation from a Rust distribution server
 
-pub use crate::dist::notifications::Notification;
-pub use dist::*;
-
-pub mod temp;
+use crate::{
```
djc, line 14: The imports look a bit disorganized. I think we should group as follows:

* `std` imports
* external crate imports
* `crate` imports (from outside this module only)
* Module declarations, each grouped with imports from that module

djc (approved):

djc:
> * It's time to decide between `foo.rs` and `foo/mod.rs`. (I'm slightly inclined towards `foo.rs`, cc @rbtcollins.)

I still find this still imposes non-trivial cognitive overhead because it means each module appears twice in the directory tree, and I have to decide which one I need to open.

src/dist/mod.rs
rami3l: Done, and thanks for the hint :)

rami3l:
> This is a natural step in the spirit of #3880.
> 
> ## Concerns
> * [ ]  It's time to decide between `foo.rs` and `foo/mod.rs`. (I'm slightly inclined towards `foo.rs`, cc @rbtcollins.)



> > * It's time to decide between `foo.rs` and `foo/mod.rs`. (I'm slightly inclined towards `foo.rs`, cc @rbtcollins.)
> 
> I still find this still imposes non-trivial cognitive overhead because it means each module appears twice in the directory tree, and I have to decide which one I need to open.

@djc Let's continue the discussion at #3883 since it looks not very appropriate to comment on a soon-to-be closed PR :P

djc (approved):

PR #3887, opened 2024-06-17
Refactor platform-dependent code in self_update

djc:


rami3l (approved):
Thanks a lot!

src/cli/self_update/windows.rs
```diff
@@ -84,6 +84,93 @@ pub(crate) fn choose_vs_install(process: &Process) -> Result<Option<VsInstallPla
...
+        md(term, MSVC_MANUAL_INSTALL_MESSAGE);
+        if !common::question_bool("\nContinue?", false, process)? {
+            info!("aborting installation");
+        }
+    }
+
+    Ok(())
+}
+
+#[cfg(windows)]
```
rami3l: The whole module is under that cfg now so it makes sense to remove these in this particular file:

```suggestion
```

djc: Yup!

PR #3889, opened 2024-06-17
chore: adjust source file structure

### LINKED ISSUE #3883: Style: Decide between `foo.rs` and `foo/mod.rs`

> It's time to decide between `foo.rs` and `foo/mod.rs`. (I'm slightly inclined towards `foo.rs`.)
_by @rami3l in https://github.com/rust-lang/rustup/pull/3882#issue-2356320731_

> I still find [`foo.rs` imposing] non-trivial cognitive overhead because it means each module appears twice in the directory tree, and I have to decide which one I need to open.
_by @djc in https://github.com/rust-lang/rustup/pull/3882#issuecomment-2172312913_

> Happy for us to have a discussion then about what style the project should take, but I've found the opposite - foo.rs means the module file path is always the same, whether or not the module has child modules, or is a leaf itself where a directory with just the module file would be tedious.
_by @rbtcollins in https://github.com/rust-lang/rustup/pull/3764#issuecomment-2048428811_

### Solution you'd like

Unify the style of source file organization across the codebase to either one of them, completely eliminating the trace of the other.

### Notes

_No response_

InfyniteHeap:
To make the module hierarchy of source files more uniform, I changed the file structure to make the hierarchy fits the "new method".

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -1205,6 +1205,8 @@ fn date_from_manifest_date(date_str: &str) -> Option<NaiveDate> {
 
 #[cfg(test)]
 mod tests {
+    use proptest::prelude::*;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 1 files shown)


rami3l:
@InfyniteHeap Thanks for contributing!

However, this aspect of the coding style remains undecided, see #3883. I'd love to merge this patch if we eventually chose `foo.rs` over `foo/mod.rs`.

rami3l (commented):
@InfyniteHeap Regardless of the result of the discussion, could you please squash all commits into one? Thanks in advance!

InfyniteHeap:
> @InfyniteHeap Regardless of the result of the discussion, could you please squash all commits into one? Thanks in advance!

Ok, and by the way I'll check whether there are something lost.

InfyniteHeap:
This PR will keep closed until #3883 has any conclusions or consensus.

The reason why this PR was temporarily closed is that if there are some conflicts between this PR and some newer PRs, this PR will not be smoothly merged. To avoid this circumstance, only when this issue has been resolved will I reopen this PR and commit changes.

PR #3891, opened 2024-06-18
test(clitools): revive `run_inprocess()`

rami3l:
This PR re-enables testing via the `run_inprocess()` function, which was originally introduced in #2367 but temporarily disabled in https://github.com/rust-lang/rustup/pull/3868.

Please note that this is a dirty fix as so far all in-process tests are still synchronous, and a `tokio` runtime builder has been re-introduced to handle the async-ness of `rustup_init::main()` as well as the need of initializing an `otel` subscriber in an async context.

### DIFF
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -12,8 +12,6 @@ use tracing_subscriber::{
 };
 
 #[cfg(feature = "otel")]
-use once_cell::sync::Lazy;
-#[cfg(feature = "otel")]
 use opentelemetry_sdk::trace::Tracer;
 
 use crate::{currentprocess::Process, utils::notify::NotificationLevel};
@@ -131,16 +129,16 @@ where
         EnvFilter::new("rustup=TRACE")
     };
     tracing_opentelemetry::layer()
-        .with_tracer(TELEMETRY_DEFAULT_TRACER.clone())
+        .with_tracer(telemetry_default_tracer())
         .with_filter(env_filter)
 }
 
 /// The default `opentelemetry` tracer used across Rustup.
 ///
 /// # Note
-/// The initializer function will panic if not called within the context of a [`tokio`] runtime.
+/// This function will panic if not called within the context of a [`tokio`] runtime.
 #[cfg(feature = "otel")]
-static TELEMETRY_DEFAULT_TRACER: Lazy<Tracer> = Lazy::new(|| {
+fn telemetry_default_tracer() -> Tracer {
     use std::time::Duration;
 
     use opentelemetry::KeyValue;
@@ -164,4 +162,4 @@ static TELEMETRY_DEFAULT_TRACER: Lazy<Tracer> = Lazy::new(|| {
         )
         .install_batch(opentelemetry_sdk::runtime::Tokio)
         .expect("error installing `OtlpTracePipeline` in the current `tokio` runtime")
-});
+}
--- a/src/test/mock/clitools.rs
+++ b/src/test/mock/clitools.rs
@@ -16,8 +16,11 @@ use std::{
 
 use enum_map::{enum_map, Enum, EnumMap};
 use once_cell::sync::Lazy;
+use tokio::runtime::Builder;
 use url::Url;
 
+use crate::cli::rustup_mode;
+use crate::currentprocess;
 use crate::test as rustup_test;
 use crate::test::const_dist_dir;
 use crate::test::this_host_triple;
@@ -676,8 +679,13 @@ impl Config {
         I: IntoIterator<Item = A> + Clone + Debug,
         A: AsRef<OsStr>,
     {
+        let inprocess = allow_inprocess(name, args.clone());
         let start = Instant::now();
-        let out = self.run_subprocess(name, args.clone(), env);
+        let out = if inprocess {
+            self.run_inprocess(name, args.clone(), env)
+        } else {
+            self.run_subprocess(name, args.clone(), env)
+        };
         let duration = Instant::now() - start;
         let output = SanitizedOutput {
             ok: matches!(out.status, Some(0)),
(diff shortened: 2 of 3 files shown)


src/cli/log.rs
rami3l, line 1: The change from a singleton to a function for the `otel` tracer is required in case the test binary decides to launch tests using another `tokio` runtime `B` instead of the current `A`, where the singleton previously initialized in `A` simply doesn't make any sense at all.

djc: Not sure I understand -- is this a theoretical risk?

src/test/mock/clitools.rs
```diff
@@ -694,6 +703,55 @@ impl Config {
...
+                    .into_boxed_str(),
+            );
+        }
+        let mut builder = Builder::new_multi_thread();
+        builder
+            .enable_all()
+            .worker_threads(2)
+            .max_blocking_threads(2);
+        let rt = builder.build().unwrap();
+        rt.block_on(async {
```
djc, line 734: That seems pretty straightforward!

djc:
> Please note that this is a dirty fix as so far all in-process tests are still synchronous,

What do you mean? It seems pretty mucht right to me.

src/cli/log.rs
rami3l: @djc No, this risk is practical. Simply put, if I don't do this then tests will fail in a random manner since the `otel` tracer will panic if it is used in a different runtime than the one it is initialized in. On my machine it gives like 2 passed tests out of 5, but the passed tests are different each time.

djc: As in, one runtime from one test and one from another?

rami3l: > As in, one runtime from one test and one from another?

@djc Yes, exactly. The error message was along the lines of "channel closed", so I imagine a message channel was created in the context of a specific async runtime to handle the log events for `otel`.

djc: Okay, that seems okay to me.

rami3l:
> > Please note that this is a dirty fix as so far all in-process tests are still synchronous,
> 
> What do you mean? It seems pretty mucht right to me.

@djc The previous implementation used a unique runtime and a static `TRACER` that is initialized in the former for all in-process tests. However once `process` is involved that seems no longer possible.

djc:
> > What do you mean? It seems pretty mucht right to me.
> 
> @djc The previous implementation used a unique runtime and a static `TRACER` that is initialized in the former for all in-process tests. However once `process` is involved that seems no longer possible.

Do you mean because I've tied the `DefaultGuard` into the `TestContext`? That's something we can just change if we want to. I suppose the issue here is that we have different kinds of tests that need the subscriber, these in-process tests but also "simpler"/more unit-style tests that benefit from the current setup. However we could for example add a third variant to `Process` for this maybe?

rami3l:
> Do you mean because I've tied the `DefaultGuard` into the `TestContext`? That's something we can just change if we want to. I suppose the issue here is that we have different kinds of tests that need the subscriber, these in-process tests but also "simpler"/more unit-style tests that benefit from the current setup. However we could for example add a third variant to `Process` for this maybe?

@djc Indeed, that would be interesting.

One trap I've encountered here is that currently `TestContext` cannot be constructed without an async context (or it'll panic), and that's why a big async block is required in the code above. Not surprising to me, but it's still a bit easy for me to accidentally make the mistake of initializing it at a random place and produce a panic (x_x)

djc:
> > Do you mean because I've tied the `DefaultGuard` into the `TestContext`? That's something we can just change if we want to. I suppose the issue here is that we have different kinds of tests that need the subscriber, these in-process tests but also "simpler"/more unit-style tests that benefit from the current setup. However we could for example add a third variant to `Process` for this maybe?
> 
> @djc Indeed, that would be interesting.
> 
> One trap I've encountered here is that currently `TestContext` cannot be constructed without an async context (or it'll panic), and that's why a big async block is required in the code above. Not surprising to me, but it's still a bit easy for me to accidentally make the mistake of initializing it at a random place and produce a panic (x_x)

We could explicitly stick a `Runtime` in it, but it feels like overkill? Ideally in my mind we would not have the `Runtime` construction in the code and instead annotate all tests that need it with `#[tokio::test]`, which seems like the idiomatic approach to me.

djc (approved):

rami3l:
> Ideally in my mind we would not have the Runtime construction in the code and instead annotate all tests that need it with #[tokio::test], which seems like the idiomatic approach to me.

@djc That's also where I want it to be, however the current state is still far from it.

djc:
> > Ideally in my mind we would not have the Runtime construction in the code and instead annotate all tests that need it with #[tokio::test], which seems like the idiomatic approach to me.
> 
> @djc That's also where I want it to be, however the current state is still far from it.

Why far? If we just remove the explicit runtime builder and add `#[tokio::test]` to the tests that need it, what breaks?

rami3l:
> > > Ideally in my mind we would not have the Runtime construction in the code and instead annotate all tests that need it with #[tokio::test], which seems like the idiomatic approach to me.
> 
> > 
> 
> > @djc That's also where I want it to be, however the current state is still far from it.
> 
> 
> 
> Why far? If we just remove the explicit runtime builder and add `#[tokio::test]` to the tests that need it, what breaks?

@djc Nothing in theory. However, as I've explained already, it's just that `run_inprocess()` is a bit deep down already in the call chain, and it's only there we start to observe asyncness (meaning only its callees can be async), so as for the current solution all its callers like `test()` and `.run()` don't require any changes.

OTOH, actually making the tests async (and adding `#[tokio::test]` to them) requires a nontrivial amount of effort to migrate those helper functions to async as well (which is worked around in this PR by pushing back the starting point of asyncness). But maybe you can have a try after this PR? I could be doing it wrong from the beginning, I'm not quite sure...

Anyway, I think this PR is ready to go as a first step. Please feel free to propose improvements in the subsequent ones.

PR #3892, opened 2024-06-18
refactor(toolchain/names): replace `toolchain_sort()` with `ToolchainName`'s `Ord` instance

### LINKED ISSUE #3884: Consider getting rid of `toolchain_sort()`
> While indeed the derived `Ord` instance are not used, I guess I was doing something idiomatic here (not entirely my fault, the function has been in existence way before #3518):

> https://github.com/rust-lang/rustup/blob/f88394cbbbcc25ffff8e0a0fd465ae77ef22ccbf/src/toolchain/names.rs#L283-L306

> How about making use of that to write an actual `Ord` instance?

_https://github.com/rust-lang/rustup/pull/3880#discussion_r1642092327_

rami3l:
Closes #3884.

This new implementation also tries to correct the handling of `X.Y`, `X.Y-beta.W` and `X.Y.Z-beta.W` versions' ordering by abusing [`semver::Comparator`](https://docs.rs/semver/latest/semver/struct.Comparator.html)'s `FromStr` implementation.

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -256,7 +256,7 @@ impl FromStr for ParsedToolchainDesc {
         // and an optional match of the date (2) and target (3)
         static TOOLCHAIN_CHANNEL_RE: Lazy<Regex> = Lazy::new(|| {
             Regex::new(&format!(
-                r"^({})(?:-(\d{{4}}-\d{{2}}-\d{{2}}))?(?:-(.+))?$",
+                r"^({})(?:-([0-9]{{4}}-[0-9]{{2}}-[0-9]{{2}}))?(?:-(.+))?$",
                 // The channel patterns we support
                 [
                     "nightly",
@@ -264,7 +264,7 @@ impl FromStr for ParsedToolchainDesc {
                     "stable",
                     // Allow from 1.0.0 through to 9.999.99 with optional patch version
                     // and optional beta tag
-                    r"\d{1}\.\d{1,3}(?:\.\d{1,2})?(?:-beta(?:\.\d{1,2})?)?",
+                    r"[0-9]{1}\.[0-9]{1,3}(?:\.[0-9]{1,2})?(?:-beta(?:\.[0-9]{1,2})?)?",
                 ]
                 .join("|")
             ))
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -120,7 +120,7 @@ pub enum DistError {
 
 #[derive(Debug, PartialEq)]
 struct ParsedToolchainDesc {
-    channel: String,
+    channel: Channel,
     date: Option<String>,
     target: Option<String>,
 }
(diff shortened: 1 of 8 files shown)


src/dist/mod.rs
```diff
@@ -146,12 +145,80 @@ pub struct PartialToolchainDesc {
...
 pub struct ToolchainDesc {
-    // Either "nightly", "stable", "beta", or an explicit version number
-    pub channel: String,
+    pub channel: ToolchainChannel,
     pub date: Option<String>,
     pub target: TargetTriple,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub enum ToolchainChannel {
```
djc: I would just name this `Channel` -- I don't think there are other kinds of channels in rustup that we need to disambiguate with?

src/dist/mod.rs
```diff
@@ -146,12 +145,80 @@ pub struct PartialToolchainDesc {
...
+    Stable,
+    Beta,
+    Nightly,
+    Ver(PartialVersion),
```
djc: Would suggest just naming the variant `Version`.

src/dist/mod.rs
```diff
@@ -146,12 +145,80 @@ pub struct PartialToolchainDesc {
...
+            "beta" => Ok(Self::Beta),
+            "nightly" => Ok(Self::Nightly),
+            ver => ver.parse().map(Self::Ver),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub struct PartialVersion {
+    pub major: u64,
```
djc, line 186: Would suggest using `u8`, `u16`, `u8` for these to use minimal types.

src/dist/mod.rs
```diff
@@ -146,12 +145,80 @@ pub struct PartialToolchainDesc {
...
+            write!(f, "-{}", self.pre)?;
+        }
+        Ok(())
+    }
+}
+
+impl FromStr for PartialVersion {
+    type Err = anyhow::Error;
+    fn from_str(ver: &str) -> Result<Self> {
+        let comparator: semver::Comparator =
```
djc: I usually try to avoid type annotations and think the spelling of `semver::Comparator::from_str(ver)` is preferable.

src/dist/mod.rs
```diff
@@ -229,7 +296,7 @@ impl FromStr for ParsedToolchainDesc {
             };
 
             Self {
-                channel: channel.to_owned(),
+                channel: channel.parse().unwrap(),
```
djc: (As such, I would also use `Channel::from_str()` here.

src/dist/mod.rs
```diff
@@ -1191,7 +1255,7 @@ mod tests {
             );
 
             let expected = ParsedToolchainDesc {
-                channel: channel.into(),
+                channel: channel.parse().unwrap(),
```
djc: And here.

djc (commented):
This looks great! I had started in the same direction but was a bit afraid of the time needed to get it right -- but would be awesome if you figure it out.

Left some early feedback.

src/dist/mod.rs
rami3l: @djc This is to align with https://docs.rs/semver/latest/semver/struct.Comparator.html 's field declarations to simplify the conversion from `Comparator` to `PartialVersion`. Should we truncate it explicitly?

djc: Hmm, I suppose it might not be worth it.

djc (approved):
Awesome!

PR #3893, opened 2024-06-18
Port tests to CliTestContext

djc:


tests/suite/cli_exact.rs
```diff
@@ -3,29 +3,24 @@
...
 #[test]
 fn update_once() {
-    test(&|config| {
-        config.with_scenario(Scenario::SimpleV2, &|config| {
-            config.expect_ok_ex(
-                &["rustup", "update", "nightly"],
-                for_host!(
-                    r"
+    let mut cx = CliTestContext::from(Scenario::None);
+    cx.config.with_scenario(Scenario::SimpleV2, &|config| {
```
rami3l: Just being curious, how do you plan to further simplify this case?

rami3l (approved):
LGTM :)

tests/suite/cli_exact.rs
djc: I think for most of these we can just pass `Scenario::SimpleV2` as the initial scenario... Trying now to remove `with_scenario()` after I made an initial pass porting all of these over to using `CliTestContext`.

rami3l: @djc I've rebased the PR to address formatting errors, and ~~I'm working on migrating~~ I have migrated `cli_v2` as well.

src/cli/self_update/windows.rs
```diff
@@ -807,16 +808,85 @@ pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
...
+            Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(None),
+            Err(e) => Err(e),
+        }
+    }
+
+    pub fn set_value(&self, new: Option<impl ToRegValue>) -> io::Result<()> {
+        self.set(new.map(|s| s.to_reg_value()).as_ref())
+    }
+
+    pub fn set(&self, new: Option<&RegValue>) -> io::Result<()> {
```
rami3l: Question: Do we really need to put both `.set()` and `.set_value()` in the public API?

rami3l (approved):
Thanks again for the tremendous amount of work (~14 000 diffLoC)! Wouldn't be possible without you!

src/cli/self_update/windows.rs
djc: We do not -- good point! Made `get()` and `set()` private.

djc:
Too bad this is like +1000 lines... all from adding async in the CLI tests. Oh well...

PR #3894, opened 2024-06-20
ci(all-features): add `-D warnings` to `cargo check-all-features`

rami3l:


src/currentprocess.rs
```diff
@@ -232,6 +231,8 @@ impl TestProcess {
 #[cfg(feature = "test")]
 impl From<TestContext> for TestProcess {
     fn from(inner: TestContext) -> Self {
+        use tracing_subscriber::util::SubscriberInitExt;
```
djc: So we import `DefaultGuard` at the top with a `feature = "test"` guard, and import this inside the item where we need it. I think we should try to be consistent? Personally I prefer to have the imports at the top, where they're out of the way of the actual code. What do you think?

djc (approved):
Nice!

src/currentprocess.rs
rami3l: @djc Okay, reverted!

PR #3895, opened 2024-06-20
fix(dist): throw an error when a `PartialVersion` string doesn't start with an ASCII digit

rami3l:
Continuation of #3892.

There is a logical bug in the previous implementation of `PartialVersion::from_str`: the parser will happily accept inputs like `^1.0` exactly because it was intended to be used that way (to parse version comparators instead of partial versions), it's just that it happens to have no effect thanks to where it's called in the current codebase. I think fixing it will make me feel safer :)

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -208,6 +208,17 @@ impl fmt::Display for PartialVersion {
 impl FromStr for PartialVersion {
     type Err = anyhow::Error;
     fn from_str(ver: &str) -> Result<Self> {
+        // `semver::Comparator::from_str` supports an optional operator
+        // (e.g. `=`, `>`, `>=`, `<`, `<=`, `~`, `^`, `*`) before the
+        // partial version, so we should exclude that case first.
+        if let Some(ch) = ver.chars().nth(0) {
+            if !ch.is_ascii_digit() {
+                return Err(anyhow!(
+                    "expected ASCII digit at the beginning of `{ver}`, found `{ch}`"
+                )
+                .context("error parsing `PartialVersion`"));
+            }
+        }
         let (ver, pre) = ver.split_once('-').unwrap_or((ver, ""));
         let comparator =
             semver::Comparator::from_str(ver).context("error parsing `PartialVersion`")?;
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -1205,6 +1205,8 @@ fn date_from_manifest_date(date_str: &str) -> Option<NaiveDate> {
 
 #[cfg(test)]
 mod tests {
+    use proptest::prelude::*;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/dist/mod.rs
```diff
@@ -208,6 +208,17 @@ impl fmt::Display for PartialVersion {
 impl FromStr for PartialVersion {
     type Err = anyhow::Error;
     fn from_str(ver: &str) -> Result<Self> {
+        // `semver::Comparator::from_str` supports an optional operator
+        // (e.g. `=`, `>`, `>=`, `<`, `<=`, `~`, `^`, `*`) before the
+        // partial version, so we should exclude that case first.
+        if let Some(ch) = ver.chars().nth(0) {
```
djc, line 216: Makes sense. Maybe add a quick unit test to cover this code?

src/dist/mod.rs
```diff
@@ -208,6 +208,17 @@ impl fmt::Display for PartialVersion {
...
+            if !ch.is_ascii_digit() {
+                return Err(anyhow!(
+                    "expected ASCII digit at the beginning of `{ver}`, found `{ch}`"
+                )
+                .context("error parsing `PartialVersion`"));
```
djc, line 221: In my mind `context()` is usually used to give more context in error messages that you don't control (from other libraries). I'm not sure the separation between the error message and the context makes sense here?

djc (approved):

src/dist/mod.rs
rami3l: @djc Oh, that's just a tiny trick I use to make the output more consistent. Here's a simple example:

```rust
use anyhow::anyhow;
use anyhow::Context;
use std::str::FromStr;

fn main() {
    foo().unwrap();
    /*
    thread 'main' panicked at src/main.rs:5:11:
    called `Result::unwrap()` on an `Err` value: error parsing `PartialVersion`
    
    Caused by:
        expected ASCII digit at the beginning of...
    */
    bar().unwrap();
    /*
    thread 'main' panicked at src/main.rs:14:11:
    called `Result::unwrap()` on an `Err` value: error parsing `PartialVersion`
    
    Caused by:
        invalid digit found in string
    */
}

fn foo() -> anyhow::Result<()> { // Throws an ad-hoc error
    Err(
        anyhow!("expected ASCII digit at the beginning of...")
            .context("error parsing `PartialVersion`"),
    )
}

fn bar() -> anyhow::Result<()> { // Throws an `std` error
    i32::from_str("bar").context("error parsing `PartialVersion`")?;
    Ok(())
}
```

Ideally if we had `try_block` I'd add the `.context()` after the `try{}`, however...

src/dist/mod.rs
rami3l: @djc Done! I figured it'll be easier to read (and more fun to write) if I write some doctests instead of unit tests, so voilà the first doctest in the whole project!

djc: Honestly I prefer unit tests? (Sorry!)

* They're out of the way instead of in the middle of other code
* No having to deal with the peculiarities of doctests
* Better syntax highlighting (for one thing, in GitHub PRs)

Doctests have their place, but IMO mostly in libraries where they serve more as tested documentation.

rami3l: @djc Alright, reverted.

djc (approved):

PR #3896, opened 2024-06-20
Replace `winreg` with `windows-registry`

### LINKED ISSUE #3779: rustup fails to open non-existing windows registry path
### Verification

### Problem

https://github.com/rust-lang/rustup/blob/1.27.0/src/cli/self_update/windows.rs#L420-L423

### Steps

```powershell
$ docker run --rm -it mcr.microsoft.com/windows/nanoserver:ltsc2022

C:\>reg query HKCU
HKEY_CURRENT_USER\Console
HKEY_CURRENT_USER\Control Panel
HKEY_CURRENT_USER\Network
HKEY_CURRENT_USER\SOFTWARE

C:\>curl --location --remote-name --remote-header-name "https://static.rust-lang.org/rustup/archive/1.27.0/x86_64-pc-windows-msvc/rustup-init.exe"

C:\>.\rustup-init.exe -y
warning: installing msvc toolchain without its prerequisites
error: Failed opening Environment key: The system cannot find the file specified. (os error 2)
```

### Possible Solution(s)

```powershell
C:\>reg add HKCU\Environment
```

### Notes

_No response_

### Rustup version

```console
rustup-init 1.27.0 (bbb9276d2 2024-03-08)
```

### Installed toolchains

```console
unknown
```

### OS version

```console
Microsoft Windows [Version 10.0.20348.2402]
```

InfyniteHeap:
Closes #3779.

APIs provided by `winreg` crate is not safe, simple and abstract enough. Besides that, it probably also contains protential bugs. Because of that, it is better to replace it with `windows-registry` crate, which is officially published and maintained by Microsoft.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,8 @@ zstd = "0.13"
 
 [target."cfg(windows)".dependencies]
 cc = "1"
-winreg = "0.52"
+windows-registry = "0.3.0"
+windows-result = "0.2.0"
 
 [target."cfg(windows)".dependencies.windows-sys]
 features = [
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -1,10 +1,8 @@
 use std::env::{consts::EXE_SUFFIX, split_paths};
 use std::ffi::{OsStr, OsString};
 use std::fmt;
-#[cfg(any(test, feature = "test"))]
-use std::io;
 use std::io::Write;
-use std::os::windows::ffi::{OsStrExt, OsStringExt};
+use std::os::windows::ffi::OsStrExt;
 use std::path::Path;
 use std::process::Command;
 use std::sync::{Arc, Mutex};
@@ -13,10 +11,11 @@ use std::sync::{LockResult, MutexGuard};
 
 use anyhow::{anyhow, Context, Result};
 use tracing::{info, warn};
-use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
 #[cfg(any(test, feature = "test"))]
-use winreg::types::{FromRegValue, ToRegValue};
-use winreg::{RegKey, RegValue};
+use windows_registry::Value;
+use windows_registry::{Key, CURRENT_USER, HSTRING};
+use windows_result::HRESULT;
+use windows_sys::Win32::Foundation::{ERROR_FILE_NOT_FOUND, ERROR_INVALID_DATA};
 
 use super::super::errors::*;
 use super::common;
@@ -465,7 +464,7 @@ pub(crate) fn do_add_to_path(process: &Process) -> Result<()> {
     do_add_to_programs(process)
 }
 
-fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
+fn _apply_new_path(new_path: Option<HSTRING>) -> Result<()> {
     use std::ptr;
     use windows_sys::Win32::Foundation::*;
     use windows_sys::Win32::UI::WindowsAndMessaging::{
@@ -477,17 +476,12 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
         None => return Ok(()), // No need to set the path
     };
 
-    let root = RegKey::predef(HKEY_CURRENT_USER);
-    let environment = root.open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)?;
+    let environment = CURRENT_USER.create("Environment")?;
 
     if new_path.is_empty() {
-        environment.delete_value("PATH")?;
+        environment.remove_value("PATH")?;
     } else {
-        let reg_value = RegValue {
-            bytes: to_winreg_bytes(new_path),
-            vtype: RegType::REG_EXPAND_SZ,
-        };
-        environment.set_raw_value("PATH", &reg_value)?;
+        environment.set_expand_hstring("PATH", &new_path)?;
     }
 
     // Tell other processes to update their environment
@@ -510,55 +504,49 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
 // Get the windows PATH variable out of the registry as a String. If
 // this returns None then the PATH variable is not a string and we
 // should not mess with it.
-fn get_windows_path_var() -> Result<Option<Vec<u16>>> {
-    use std::io;
-
-    let root = RegKey::predef(HKEY_CURRENT_USER);
-    let environment = root
-        .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+fn get_windows_path_var() -> Result<Option<HSTRING>> {
+    let environment = CURRENT_USER
+        .create("Environment")
         .context("Failed opening Environment key")?;
 
-    let reg_value = environment.get_raw_value("PATH");
+    let reg_value = environment.get_hstring("PATH");
     match reg_value {
-        Ok(val) => {
-            if let Some(s) = from_winreg_value(&val) {
-                Ok(Some(s))
-            } else {
-                warn!(
-                    "the registry key HKEY_CURRENT_USER\\Environment\\PATH is not a string. \
-                       Not modifying the PATH variable"
-                );
-                Ok(None)
-            }
+        Ok(val) => Ok(Some(val)),
+        Err(e) if e.code() == HRESULT::from_win32(ERROR_INVALID_DATA) => {
+            warn!(
+                "the registry key HKEY_CURRENT_USER\\Environment\\PATH is not a string. \
+                   Not modifying the PATH variable"
+            );
+            Ok(None)
         }
-        Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(Some(Vec::new())),
+        Err(e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => Ok(Some(HSTRING::new())),
         Err(e) => Err(e).context(CLIError::WindowsUninstallMadness),
     }
 }
 
 // Returns None if the existing old_path does not need changing, otherwise
 // prepends the path_str to old_path, handling empty old_path appropriately.
-fn _add_to_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>> {
+fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
     if old_path.is_empty() {
         Some(path_str)
     } else if old_path
         .windows(path_str.len())
-        .any(|path| path == path_str)
+        .any(|path| *path == *path_str)
     {
         None
     } else {
-        let mut new_path = path_str;
-        new_path.push(b';' as u16);
-        new_path.extend_from_slice(&old_path);
-        Some(new_path)
+        let mut new_path = path_str.to_os_string();
+        new_path.push(";");
+        new_path.push(old_path.to_os_string());
+        Some(HSTRING::from(new_path))
     }
 }
 
 // Returns None if the existing old_path does not need changing
-fn _remove_from_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>> {
+fn _remove_from_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
     let idx = old_path
         .windows(path_str.len())
-        .position(|path| path == path_str)?;
+        .position(|path| *path == *path_str)?;
     // If there's a trailing semicolon (likely, since we probably added one
     // during install), include that in the substring to remove. We don't search
     // for that to find the string, because if it's the last string in the path,
@@ -575,18 +563,17 @@ fn _remove_from_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>>
     if new_path.last() == Some(&(b';' as u16)) {
         new_path.pop();
     }
-    Some(new_path)
+    Some(HSTRING::from_wide(&new_path))
 }
 
-fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<Vec<u16>>>
+fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRING>>
 where
-    F: FnOnce(Vec<u16>, Vec<u16>) -> Option<Vec<u16>>,
+    F: FnOnce(HSTRING, HSTRING) -> Option<HSTRING>,
 {
     let windows_path = get_windows_path_var()?;
     let mut path_str = process.cargo_home()?;
     path_str.push("bin");
-    Ok(windows_path
-        .and_then(|old_path| f(old_path, OsString::from(path_str).encode_wide().collect())))
+    Ok(windows_path.and_then(|old_path| f(old_path, HSTRING::from(path_str.as_path()))))
 }
 
 pub(crate) fn do_remove_from_path(process: &Process) -> Result<()> {
@@ -597,16 +584,15 @@ pub(crate) fn do_remove_from_path(process: &Process) -> Result<()> {
 
 const RUSTUP_UNINSTALL_ENTRY: &str = r"Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup";
 
-fn rustup_uninstall_reg_key() -> Result<RegKey> {
-    Ok(RegKey::predef(HKEY_CURRENT_USER)
-        .create_subkey(RUSTUP_UNINSTALL_ENTRY)
-        .context("Failed creating uninstall key")?
-        .0)
+fn rustup_uninstall_reg_key() -> Result<Key> {
+    CURRENT_USER
+        .create(RUSTUP_UNINSTALL_ENTRY)
+        .context("Failed creating uninstall key")
 }
 
 pub(crate) fn do_update_programs_display_version(version: &str) -> Result<()> {
     rustup_uninstall_reg_key()?
-        .set_value("DisplayVersion", &version)
+        .set_string("DisplayVersion", version)
         .context("Failed to set `DisplayVersion`")
 }
 
@@ -616,11 +602,9 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
     let key = rustup_uninstall_reg_key()?;
 
     // Don't overwrite registry if Rustup is already installed
-    let prev = key
-        .get_raw_value("UninstallString")
-        .map(|val| from_winreg_value(&val));
-    if let Ok(Some(s)) = prev {
-        let mut path = PathBuf::from(OsString::from_wide(&s));
+    let prev = key.get_hstring("UninstallString");
+    if let Ok(s) = prev {
+        let mut path = PathBuf::from(s.to_os_string());
         path.pop();
         if path.exists() {
             return Ok(());
@@ -633,14 +617,9 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
     uninstall_cmd.push(path);
     uninstall_cmd.push("\" self uninstall");
 
-    let reg_value = RegValue {
-        bytes: to_winreg_bytes(uninstall_cmd.encode_wide().collect()),
-        vtype: RegType::REG_SZ,
-    };
-
-    key.set_raw_value("UninstallString", &reg_value)
+    key.set_hstring("UninstallString", &HSTRING::from(uninstall_cmd))
         .context("Failed to set `UninstallString`")?;
-    key.set_value("DisplayName", &"Rustup: the Rust toolchain installer")
+    key.set_string("DisplayName", "Rustup: the Rust toolchain installer")
         .context("Failed to set `DisplayName`")?;
     do_update_programs_display_version(env!("CARGO_PKG_VERSION"))?;
 
@@ -648,39 +627,10 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
 }
 
 pub(crate) fn do_remove_from_programs() -> Result<()> {
-    match RegKey::predef(HKEY_CURRENT_USER).delete_subkey_all(RUSTUP_UNINSTALL_ENTRY) {
+    match CURRENT_USER.remove_tree(RUSTUP_UNINSTALL_ENTRY) {
         Ok(()) => Ok(()),
-        Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => Ok(()),
+        Err(e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => Ok(()),
         Err(e) => Err(anyhow!(e)),
-    }
-}
-
-/// Convert a vector UCS-2 chars to a null-terminated UCS-2 string in bytes
-pub(crate) fn to_winreg_bytes(mut v: Vec<u16>) -> Vec<u8> {
-    v.push(0);
-    unsafe { std::slice::from_raw_parts(v.as_ptr().cast::<u8>(), v.len() * 2).to_vec() }
-}
-
-/// This is used to decode the value of HKCU\Environment\PATH. If that key is
-/// not REG_SZ | REG_EXPAND_SZ then this returns None. The winreg library itself
-/// does a lossy unicode conversion.
-pub(crate) fn from_winreg_value(val: &winreg::RegValue) -> Option<Vec<u16>> {
-    use std::slice;
-
-    match val.vtype {
-        RegType::REG_SZ | RegType::REG_EXPAND_SZ => {
-            // Copied from winreg
-            let mut words = unsafe {
-                #[allow(clippy::cast_ptr_alignment)]
-                slice::from_raw_parts(val.bytes.as_ptr().cast::<u16>(), val.bytes.len() / 2)
-                    .to_owned()
-            };
-            while words.last() == Some(&0) {
-                words.pop();
-            }
-            Some(words)
-        }
-        _ => None,
     }
 }
 
@@ -813,7 +763,7 @@ pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
 }
 
 #[cfg(any(test, feature = "test"))]
-pub fn get_path() -> io::Result<Option<RegValue>> {
+pub fn get_path() -> Result<Option<Value>> {
     USER_PATH.get()
 }
 
@@ -821,12 +771,12 @@ pub fn get_path() -> io::Result<Option<RegValue>> {
 pub struct RegistryGuard<'a> {
     _locked: LockResult<MutexGuard<'a, ()>>,
     id: &'static RegistryValueId,
-    prev: Option<RegValue>,
+    prev: Option<Value>,
 }
 
 #[cfg(any(test, feature = "test"))]
 impl<'a> RegistryGuard<'a> {
-    pub fn new(id: &'static RegistryValueId) -> io::Result<Self> {
+    pub fn new(id: &'static RegistryValueId) -> Result<Self> {
         Ok(Self {
             _locked: REGISTRY_LOCK.lock(),
             id,
@@ -859,50 +809,48 @@ pub struct RegistryValueId {
 
 #[cfg(any(test, feature = "test"))]
 impl RegistryValueId {
-    pub fn get_value<T: FromRegValue>(&self) -> io::Result<Option<T>> {
-        self.get()?.map(|v| T::from_reg_value(&v)).transpose()
+    pub fn get_value(&self) -> Result<Option<Value>> {
+        self.get()
     }
 
-    fn get(&self) -> io::Result<Option<RegValue>> {
-        let sub_key = RegKey::predef(HKEY_CURRENT_USER)
-            .open_subkey_with_flags(self.sub_key, KEY_READ | KEY_WRITE)?;
-        match sub_key.get_raw_value(self.value_name) {
+    fn get(&self) -> Result<Option<Value>> {
+        let sub_key = CURRENT_USER.create(self.sub_key)?;
+        match sub_key.get_value(self.value_name) {
             Ok(val) => Ok(Some(val)),
-            Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(None),
-            Err(e) => Err(e),
+            Err(e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => Ok(None),
+            Err(e) => Err(e.into()),
         }
     }
 
-    pub fn set_value(&self, new: Option<impl ToRegValue>) -> io::Result<()> {
-        self.set(new.map(|s| s.to_reg_value()).as_ref())
+    pub fn set_value(&self, new: Option<&Value>) -> Result<()> {
+        self.set(new)
     }
 
-    fn set(&self, new: Option<&RegValue>) -> io::Result<()> {
-        let sub_key = RegKey::predef(HKEY_CURRENT_USER)
-            .open_subkey_with_flags(self.sub_key, KEY_READ | KEY_WRITE)?;
+    fn set(&self, new: Option<&Value>) -> Result<()> {
+        let sub_key = CURRENT_USER.create(self.sub_key)?;
         match new {
-            Some(new) => sub_key.set_raw_value(self.value_name, new),
-            None => sub_key.delete_value(self.value_name),
+            Some(new) => Ok(sub_key.set_value(self.value_name, new)?),
+            None => Ok(sub_key.remove_value(self.value_name)?),
         }
     }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::os::windows::ffi::OsStringExt;
+
+    use windows_registry::Type;
+
     use super::*;
     use crate::process::TestProcess;
-
-    fn wide(str: &str) -> Vec<u16> {
-        OsString::from(str).encode_wide().collect()
-    }
 
     #[test]
     fn windows_install_does_not_add_path_twice() {
         assert_eq!(
             None,
             super::_add_to_path(
-                wide(r"c:\users\example\.cargo\bin;foo"),
-                wide(r"c:\users\example\.cargo\bin")
+                HSTRING::from(r"c:\users\example\.cargo\bin;foo"),
+                HSTRING::from(r"c:\users\example\.cargo\bin")
             )
         );
     }
@@ -914,16 +862,19 @@ mod tests {
             0x0101, // bogus trailing surrogate
             0x0000, // null
         ];
-        let cargo_home = wide(r"c:\users\example\.cargo\bin");
-        let final_path = [&cargo_home, &[b';' as u16][..], &initial_path].join(&[][..]);
+        let cargo_home = r"c:\users\example\.cargo\bin";
+        let mut final_path = OsString::from(cargo_home);
+        final_path.push(";");
+        final_path.push(OsString::from_wide(&initial_path));
 
         assert_eq!(
-            &final_path,
-            &super::_add_to_path(initial_path.clone(), cargo_home.clone(),).unwrap()
+            final_path,
+            super::_add_to_path(HSTRING::from_wide(&initial_path), HSTRING::from(cargo_home))
+                .unwrap()
         );
         assert_eq!(
-            &initial_path,
-            &super::_remove_from_path(final_path, cargo_home,).unwrap()
+            HSTRING::from_wide(&initial_path),
+            super::_remove_from_path(HSTRING::from(final_path), HSTRING::from(cargo_home)).unwrap()
         );
     }
 
@@ -931,56 +882,44 @@ mod tests {
     fn windows_path_regkey_type() {
         // per issue #261, setting PATH should use REG_EXPAND_SZ.
         let _guard = RegistryGuard::new(&USER_PATH);
-        let root = RegKey::predef(HKEY_CURRENT_USER);
-        let environment = root
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
-            .unwrap();
-        environment.delete_value("PATH").unwrap();
+        let environment = CURRENT_USER.create("Environment").unwrap();
+        environment.remove_value("PATH").unwrap();
 
         {
             // Can't compare the Results as Eq isn't derived; thanks error-chain.
             #![allow(clippy::unit_cmp)]
-            assert_eq!((), super::_apply_new_path(Some(wide("foo"))).unwrap());
+            assert_eq!(
+                (),
+                super::_apply_new_path(Some(HSTRING::from("foo"))).unwrap()
+            );
         }
-        let root = RegKey::predef(HKEY_CURRENT_USER);
-        let environment = root
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
-            .unwrap();
-        let path = environment.get_raw_value("PATH").unwrap();
-        assert_eq!(path.vtype, RegType::REG_EXPAND_SZ);
-        assert_eq!(super::to_winreg_bytes(wide("foo")), &path.bytes[..]);
+        let environment = CURRENT_USER.create("Environment").unwrap();
+        let path = environment.get_value("PATH").unwrap();
+        let path_hstring = environment.get_hstring("PATH").unwrap();
+        assert_eq!(path.ty(), Type::ExpandString);
+        assert_eq!(HSTRING::from("foo"), path_hstring);
     }
 
     #[test]
     fn windows_path_delete_key_when_empty() {
-        use std::io;
         // during uninstall the PATH key may end up empty; if so we should
         // delete it.
         let _guard = RegistryGuard::new(&USER_PATH);
-        let root = RegKey::predef(HKEY_CURRENT_USER);
-        let environment = root
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
-            .unwrap();
+        let environment = CURRENT_USER.create("Environment").unwrap();
         environment
-            .set_raw_value(
-                "PATH",
-                &RegValue {
-                    bytes: super::to_winreg_bytes(wide("foo")),
-                    vtype: RegType::REG_EXPAND_SZ,
-                },
-            )
+            .set_expand_hstring("PATH", &HSTRING::from("foo"))
             .unwrap();
 
         {
             // Can't compare the Results as Eq isn't derived; thanks error-chain.
             #![allow(clippy::unit_cmp)]
-            assert_eq!((), super::_apply_new_path(Some(Vec::new())).unwrap());
+            assert_eq!((), super::_apply_new_path(Some(HSTRING::new())).unwrap());
         }
-        let reg_value = environment.get_raw_value("PATH");
+        let reg_value = environment.get_value("PATH");
         match reg_value {
             Ok(_) => panic!("key not deleted"),
-            Err(ref e) if e.kind() == io::ErrorKind::NotFound => {}
-            Err(ref e) => panic!("error {e}"),
+            Err(e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => {}
+            Err(e) => panic!("error {e}"),
         }
     }
 
@@ -995,15 +934,10 @@ mod tests {
         );
 
         let _guard = RegistryGuard::new(&USER_PATH);
-        let root = RegKey::predef(HKEY_CURRENT_USER);
-        let environment = root
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+        let environment = CURRENT_USER.create("Environment").unwrap();
+        environment
+            .set_bytes("PATH", Type::Bytes, &[0x12, 0x34])
             .unwrap();
-        let reg_value = RegValue {
-            bytes: vec![0x12, 0x34],
-            vtype: RegType::REG_BINARY,
-        };
-        environment.set_raw_value("PATH", &reg_value).unwrap();
         // Ok(None) signals no change to the PATH setting layer
         assert_eq!(
             None,
@@ -1021,22 +955,19 @@ mod tests {
     fn windows_treat_missing_path_as_empty() {
         // during install the PATH key may be missing; treat it as empty
         let _guard = RegistryGuard::new(&USER_PATH);
-        let root = RegKey::predef(HKEY_CURRENT_USER);
-        let environment = root
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
-            .unwrap();
-        environment.delete_value("PATH").unwrap();
+        let environment = CURRENT_USER.create("Environment").unwrap();
+        environment.remove_value("PATH").unwrap();
 
-        assert_eq!(Some(Vec::new()), super::get_windows_path_var().unwrap());
+        assert_eq!(Some(HSTRING::new()), super::get_windows_path_var().unwrap());
     }
 
     #[test]
     fn windows_uninstall_removes_semicolon_from_path_prefix() {
         assert_eq!(
-            wide("foo"),
+            HSTRING::from("foo"),
             super::_remove_from_path(
-                wide(r"c:\users\example\.cargo\bin;foo"),
-                wide(r"c:\users\example\.cargo\bin"),
+                HSTRING::from(r"c:\users\example\.cargo\bin;foo"),
+                HSTRING::from(r"c:\users\example\.cargo\bin"),
             )
             .unwrap()
         )
@@ -1045,10 +976,10 @@ mod tests {
     #[test]
     fn windows_uninstall_removes_semicolon_from_path_suffix() {
         assert_eq!(
-            wide("foo"),
+            HSTRING::from("foo"),
             super::_remove_from_path(
-                wide(r"foo;c:\users\example\.cargo\bin"),
-                wide(r"c:\users\example\.cargo\bin"),
+                HSTRING::from(r"foo;c:\users\example\.cargo\bin"),
+                HSTRING::from(r"c:\users\example\.cargo\bin"),
             )
             .unwrap()
         )
--- a/tests/suite/cli_paths.rs
+++ b/tests/suite/cli_paths.rs
@@ -383,13 +383,19 @@ mod windows {
     use rustup::test::mock::clitools::{CliTestContext, Scenario};
     use rustup::test::{get_path, RegistryGuard, USER_PATH};
 
+    use windows_registry::{Value, HSTRING};
+
     #[tokio::test]
     /// Smoke test for end-to-end code connectivity of the installer path mgmt on windows.
     async fn install_uninstall_affect_path() {
         let mut cx = CliTestContext::new(Scenario::Empty).await;
         let _guard = RegistryGuard::new(&USER_PATH).unwrap();
         let cfg_path = cx.config.cargodir.join("bin").display().to_string();
-        let get_path_ = || get_path().unwrap().unwrap().to_string();
+        let get_path_ = || {
+            HSTRING::try_from(get_path().unwrap().unwrap())
+                .unwrap()
+                .to_string()
+        };
 
         cx.config.expect_ok(&INIT_NONE).await;
         assert!(
@@ -408,38 +414,39 @@ mod windows {
     #[tokio::test]
     /// Smoke test for end-to-end code connectivity of the installer path mgmt on windows.
     async fn install_uninstall_affect_path_with_non_unicode() {
-        use std::ffi::OsString;
         use std::os::windows::ffi::OsStrExt;
-        use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
-        use winreg::{RegKey, RegValue};
+
+        use windows_registry::{Type, CURRENT_USER};
 
         let mut cx = CliTestContext::new(Scenario::Empty).await;
         let _guard = RegistryGuard::new(&USER_PATH).unwrap();
         // Set up a non unicode PATH
-        let reg_value = RegValue {
-            bytes: vec![
-                0x00, 0xD8, // leading surrogate
-                0x01, 0x01, // bogus trailing surrogate
-                0x00, 0x00, // null
-            ],
-            vtype: RegType::REG_EXPAND_SZ,
-        };
-        RegKey::predef(HKEY_CURRENT_USER)
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+        let mut reg_value = Value::from([
+            0x00, 0xD8, // leading surrogate
+            0x01, 0x01, // bogus trailing surrogate
+            0x00, 0x00, // null
+        ]);
+        reg_value.set_ty(Type::ExpandString);
+        CURRENT_USER
+            .create("Environment")
             .unwrap()
-            .set_raw_value("PATH", &reg_value)
+            .set_value("PATH", &reg_value)
             .unwrap();
 
         // compute expected path after installation
-        let expected = RegValue {
-            bytes: OsString::from(cx.config.cargodir.join("bin"))
+        let mut expected = Value::from(
+            cx.config
+                .cargodir
+                .join("bin")
+                .as_os_str()
                 .encode_wide()
                 .flat_map(|v| vec![v as u8, (v >> 8) as u8])
                 .chain(vec![b';', 0])
-                .chain(reg_value.bytes.iter().copied())
-                .collect(),
-            vtype: RegType::REG_EXPAND_SZ,
-        };
+                .chain(reg_value.iter().copied())
+                .collect::<Vec<u8>>()
+                .as_slice(),
+        );
+        expected.set_ty(Type::ExpandString);
 
         cx.config.expect_ok(&INIT_NONE).await;
         assert_eq!(get_path().unwrap().unwrap(), expected);
--- a/tests/suite/cli_self_upd.rs
+++ b/tests/suite/cli_self_upd.rs
@@ -23,6 +23,8 @@ use rustup::test::{
 use rustup::test::{RegistryGuard, RegistryValueId, USER_PATH};
 use rustup::utils::{raw, utils};
 use rustup::{for_host, DUP_TOOLS, TOOLS};
+#[cfg(windows)]
+use windows_registry::Value;
 
 const TEST_VERSION: &str = "1.1.1";
 
@@ -350,12 +352,12 @@ async fn update_overwrites_programs_display_version() {
         .await;
 
     USER_RUSTUP_VERSION
-        .set_value(Some(PLACEHOLDER_VERSION))
+        .set_value(Some(&Value::from(PLACEHOLDER_VERSION)))
         .unwrap();
     cx.config.expect_ok(&["rustup", "self", "update"]).await;
     assert_eq!(
-        USER_RUSTUP_VERSION.get_value::<String>().unwrap().unwrap(),
-        version,
+        USER_RUSTUP_VERSION.get_value().unwrap().unwrap(),
+        Value::from(version)
     );
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1984,7 +1984,7 @@ dependencies = [
  "wasm-bindgen-futures",
  "wasm-streams",
  "web-sys",
- "windows-registry",
+ "windows-registry 0.2.0",
 ]
 
 [[package]]
@@ -2184,8 +2184,9 @@ dependencies = [
  "url",
  "wait-timeout",
  "walkdir",
+ "windows-registry 0.3.0",
+ "windows-result",
  "windows-sys 0.59.0",
- "winreg",
  "xz2",
  "zstd",
 ]
@@ -3182,7 +3183,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e400001bb720a623c1c69032f8e3e4cf09984deec740f007dd2b03ec864804b0"
 dependencies = [
  "windows-result",
- "windows-strings",
+ "windows-strings 0.1.0",
+ "windows-targets 0.52.6",
+]
+
+[[package]]
+name = "windows-registry"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bafa604f2104cf5ae2cc2db1dee84b7e6a5d11b05f737b60def0ffdc398cbc0a"
+dependencies = [
+ "windows-result",
+ "windows-strings 0.2.0",
  "windows-targets 0.52.6",
 ]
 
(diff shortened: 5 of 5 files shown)


src/cli/self_update/windows.rs
```diff
@@ -505,83 +512,111 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
+//     // Don't leave a trailing ; though, we don't want an empty string in the
+//     // path.
+//     if new_path.last() == Some(&(b';' as u16)) {
+//         new_path.pop();
+//     }
+//     Some(new_path)
+// }
+
 // Returns None if the existing old_path does not need changing
-fn _remove_from_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>> {
```
InfyniteHeap, line 558: This implementation has some bugs, hence the test item `windows_uninstall_removes_semicolon_from_path_prefix` will be failed.

I don't know how to fix it.

src/cli/self_update/windows.rs
```diff
@@ -505,83 +512,111 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
+        // .windows(path_str.len())
+        // .position(|path| path == path_str)?;
+        .find(&path_str)?;
     // If there's a trailing semicolon (likely, since we probably added one
     // during install), include that in the substring to remove. We don't search
     // for that to find the string, because if it's the last string in the path,
     // there may not be.
     let mut len = path_str.len();
-    if old_path.get(idx + path_str.len()) == Some(&(b';' as u16)) {
+    if old_path.chars().collect::<Vec<char>>()[idx + path_str.len() - 1] == ';' {
```
rami3l: Why are you using `.chars()` here? The previous `.find()` returns a byte index, not a char index, so indexing like this makes no sense (I assume you have some basic knowledge of how chars are laid out in [UTF8 strings](https://www.utf8-chartable.de/)).

Normally in this case, you should use [`.get()`](https://doc.rust-lang.org/std/primitive.str.html#method.get).  However, since you're checking for `';'`, I think you can take a shortcut here by calling `.as_bytes()` followed by a `.get()`, and finally compare it with `Some(b';')`.

src/cli/self_update/windows.rs
rami3l: @InfyniteHeap This seems to be a regular Rust function that is not related to Windows-specific features. I encourage you to try out some input String examples with, say, emojis, or Chinese characters, to make sure they work correctly at least in the valid UTF8 case.

src/cli/self_update/windows.rs
```diff
@@ -505,83 +512,111 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
         len += 1;
     }
 
-    let mut new_path = old_path[..idx].to_owned();
-    new_path.extend_from_slice(&old_path[idx + len..]);
+    let mut new_path = old_path[..idx].to_string();
```
rami3l: You should still be able to use `.to_owned()` here.

src/cli/self_update/windows.rs
InfyniteHeap: I previously correctly implemented this function (sorry, I mis-referred to this deleted one, I must refer to the added one😅), but today I met some accidents so what my changed had been lost, and I forgot how did I implemented it, so I used this one.

rami3l: @InfyniteHeap No worries! Just don't forget to squash these commits with back-and-forth changes into one when you mark the PR as ready for review, we'd like to see atomic commits (you can get inspired from the merged ones) :)

src/cli/self_update/test.rs
```diff
@@ -51,23 +56,21 @@ pub fn with_saved_reg_value(root: &RegKey, subkey: &str, name: &str, f: &mut dyn
 }
 
 #[cfg(windows)]
-fn get_reg_value(root: &RegKey, subkey: &str, name: &str) -> io::Result<Option<RegValue>> {
-    let subkey = root.open_subkey_with_flags(subkey, KEY_READ | KEY_WRITE)?;
-    match subkey.get_raw_value(name) {
+fn get_reg_value(root: &Key, subkey: &str, name: &str) -> io::Result<Option<String>> {
+    let subkey = root.create(subkey)?;
+    match subkey.get_string(name) {
```
rami3l: Ideally you should use `.get_value()` and `.set_value()` to replicate the functionalities of the original `get_reg_value()`.

InfyniteHeap: @rami3l One of the function callers is the `get_path()`. I think that the paths are always string, so I directly used `get_string()`.

rami3l: > @rami3l One of the function callers is the `get_path()`. I think that the paths are always string, so I directly used `get_string()`.

@InfyniteHeap If you are sure that it's the only call site, I suggest renaming the functions to `(get|set)_reg_string()` for clarity.

InfyniteHeap: > > @rami3l One of the function callers is the `get_path()`. I think that the paths are always string, so I directly used `get_string()`.
> 
> @InfyniteHeap If you are sure that it's the only call site, I suggest renaming the functions to `(get|set)_reg_string()` for clarity.

I re-checked this function's callers and I'm sure that those who calls this function will only get string.

src/cli/self_update/windows.rs
rami3l: @InfyniteHeap Sorry for not being clear enough previously, but you are allowed to have multiple commits per PR. The goal of atomic commits is to make it clearer to see how you made it to the final state step by step. For example, replacing `winreg` calls with `windows-registry` calls can go into a separate commit, and all function renaming actions can go into another one.

Cargo.toml
```diff
@@ -96,7 +96,9 @@ zstd = "0.13"
 
 [target."cfg(windows)".dependencies]
 cc = "1"
-winreg = "0.52"
+#winreg = "0.52"
```
djc: Let's get rid of the commented out stuff?

Cargo.toml
```diff
@@ -143,7 +145,10 @@ proptest = "1.1.0"
 tempfile = "3.8"
 termcolor = "1.2"
 thiserror = "1.0"
-tokio = { version = "1.26.0", default-features = false, features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.26.0", default-features = false, features = [
```
djc: Please avoid this reformatting.

src/cli/self_update/windows.rs
InfyniteHeap: > @InfyniteHeap Sorry for not being clear enough previously, but you are allowed to have multiple commits per PR. The goal of atomic commits is to make it clearer to see how you made it to the final state step by step. For example, replacing `winreg` calls with `windows-registry` calls can go into a separate commit, and all function renaming actions can go into another one.

So should I do something? Or just keep the current state?

src/cli/self_update/test.rs
```diff
@@ -3,10 +3,15 @@
 use std::{io, sync::Mutex};
 
 #[cfg(windows)]
-use winreg::{
-    enums::{HKEY_CURRENT_USER, KEY_READ, KEY_WRITE},
-    RegKey, RegValue,
-};
+// use winreg::{
```
djc: Remove the old stuff?

src/cli/self_update/test.rs
```diff
@@ -37,37 +42,35 @@ pub fn with_saved_path(f: &mut dyn FnMut()) {
...
 #[cfg(windows)]
-pub fn get_path() -> io::Result<Option<RegValue>> {
-    get_reg_value(&RegKey::predef(HKEY_CURRENT_USER), "Environment", "PATH")
+pub fn get_path() -> io::Result<Option<String>> {
+    get_reg_string(CURRENT_USER, "Environment", "PATH")
 }
 
 #[cfg(windows)]
-pub fn with_saved_reg_value(root: &RegKey, subkey: &str, name: &str, f: &mut dyn FnMut()) {
+pub fn with_saved_reg_string(root: &Key, subkey: &str, name: &str, f: &mut dyn FnMut()) {
```
djc: As @rami3l suggested, this renaming isn't really necessary for porting to windows-registry. As such, it should be in a separate commit. (If you're renaming a family of related functions those can probably all go in the same commit.)

src/cli/self_update/test.rs
```diff
@@ -37,37 +42,35 @@ pub fn with_saved_path(f: &mut dyn FnMut()) {
...
-fn get_reg_value(root: &RegKey, subkey: &str, name: &str) -> io::Result<Option<RegValue>> {
-    let subkey = root.open_subkey_with_flags(subkey, KEY_READ | KEY_WRITE)?;
-    match subkey.get_raw_value(name) {
+fn get_reg_string(root: &Key, subkey: &str, name: &str) -> io::Result<Option<String>> {
+    let subkey = root.create(subkey)?;
+    match subkey.get_string(name) {
         Ok(val) => Ok(Some(val)),
-        Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(None),
-        Err(e) => Err(e),
+        Err(ref e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => Ok(None),
```
djc: Are we sure this is the best way of expressing the condition? Can't we look at the error code in `e` "directly" somehow?

src/cli/self_update/windows.rs
```diff
@@ -19,8 +19,11 @@ use crate::dist::TargetTriple;
 use crate::utils::utils;
 use crate::utils::Notification;
 
-use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
-use winreg::{RegKey, RegValue};
+// use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
+// use winreg::{RegKey, RegValue};
```
djc: Remove the old stuff.

src/cli/self_update/windows.rs
```diff
@@ -460,35 +463,39 @@ pub(crate) fn do_add_to_path(process: &Process) -> Result<()> {
...
     };
 
     let new_path = match new_path {
         Some(new_path) => new_path,
         None => return Ok(()), // No need to set the path
     };
 
-    let root = RegKey::predef(HKEY_CURRENT_USER);
-    let environment = root.open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)?;
+    // let root = RegKey::predef(HKEY_CURRENT_USER);
```
djc: Remove the old stuff.

src/cli/self_update/windows.rs
```diff
@@ -460,35 +463,39 @@ pub(crate) fn do_add_to_path(process: &Process) -> Result<()> {
     do_add_to_programs(process)
 }
 
-fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
+fn _apply_new_path(new_path: Option<String>) -> Result<()> {
```
djc: Looks like this could be an `Option<&str>` (in the sense that the function body doesn't need to take/pass ownership)?

src/cli/self_update/windows.rs
```diff
@@ -460,35 +463,39 @@ pub(crate) fn do_add_to_path(process: &Process) -> Result<()> {
...
+        // environment.set_raw_value("PATH", &reg_value)?;
+
+        environment.set_string("PATH", &new_path)?;
     }
 
     // Tell other processes to update their environment
     #[allow(clippy::unnecessary_cast)]
     unsafe {
-        SendMessageTimeoutA(
+        SendMessageTimeoutW(
```
djc: Why are we switching this?

src/cli/self_update/windows.rs
```diff
@@ -505,83 +512,111 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
-            if let Some(s) = from_winreg_value(&val) {
-                Ok(Some(s))
-            } else {
-                warn!(
-                    "the registry key HKEY_CURRENT_USER\\Environment\\PATH is not a string. \
-                       Not modifying the PATH variable"
-                );
-                Ok(None)
-            }
+            // if let Some(s) = from_winreg_value(&val) {
```
djc: Old code.

djc (commented):
I haven't done a complete review yet, because I think there's a bunch of stuff left to clean up. The core change looks great though, thanks for working on this!

src/cli/self_update/test.rs
InfyniteHeap: I think we can't, because `Error` struct in `windows-result` doesn't hold the error code directly. Plus, the previous code also didn't expose its error code, so I made this change.

src/cli/self_update/windows.rs
InfyniteHeap: I noticed that there is a function called `CreateFileW()` in `src/cli/self_update/windows.rs`, so I want to unify them. But according to what @rami3l said, I'll put it into a separate PR.

src/cli/self_update/windows.rs
InfyniteHeap: In this function there is a code snippet which verifies whether this string is empty. Because this `is_empty()` is implemented for `String`, so I chose `String`.

djc: is_empty() is actually defined on &str and can be used via String's Deref<Target = &str> implementation.

src/cli/self_update/test.rs
rami3l: @djc HRESULT includes both Win32 and WinNT error codes, this `from_win32` is just an arithmetic mapping function without much overhead (plus it's a const function so can even be const propagated).

src/cli/self_update/windows.rs
rami3l: @djc *A is for ANSI string-based APIs and *W for UTF16-based ones. For this particular function there seems to be no difference between the two but I guess in terms of consistency *W is still slightly better: https://stackoverflow.com/a/7424550

src/cli/self_update/windows.rs
InfyniteHeap: > is_empty() is actually defined on &str and can be used via String's Deref<Target = &str> implementation.

Oh OK, got it!

InfyniteHeap: I attempted using `&str` but I met lifetime problems😅, so I continue using `String` unless I resolved these problems.

InfyniteHeap:
Well, it is only now that I know that changing branch name will automatically close PR. SORRY!

src/cli/self_update/windows.rs
```diff
@@ -505,103 +501,83 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
-            } else {
-                warn!(
-                    "the registry key HKEY_CURRENT_USER\\Environment\\PATH is not a string. \
-                       Not modifying the PATH variable"
-                );
-                Ok(None)
-            }
+    match environment.get_string("PATH") {
+        Ok(val) => Ok(Some(val)),
+        Err(ref e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => {
```
djc: I'm wondering if this code still makes sense? It seems surprising that getting the `PATH` from the registry would yield an `ERROR_FILE_NOT_FOUND` error...

src/cli/self_update/windows.rs
```diff
@@ -809,110 +749,86 @@ pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
...
-        assert_eq!(
-            &final_path,
-            &super::_add_to_path(initial_path.clone(), cargo_home.clone(),).unwrap()
-        );
-        assert_eq!(
-            &initial_path,
-            &super::_remove_from_path(final_path, cargo_home,).unwrap()
-        );
-    }
+    // FIXME: Must evaluate whether this test item should be removed.
```
djc: Should either remove this or fix it before merging I guess?

djc (commented):
Overall a really nice improvement!

src/cli/self_update/windows.rs
InfyniteHeap: I have the same question, but that fact is that I must use this code or tests will be failed with hint "File not found".

djc: Okay, that's enough for me to be satisfied that we still need it!

src/cli/self_update/windows.rs
InfyniteHeap: I considered a while and decided to remove it, because using `windows-registry` will never appear non-unicode characters, I thought.

src/cli/self_update/windows.rs
rami3l: @djc It actually can, see #3779 for such a case.

tests/suite/cli_paths.rs
```diff
@@ -415,46 +415,53 @@ mod windows {
     }
 
     #[test]
+    #[ignore = "FIXME: Must evaluate whether this test item should be removed."]
```
rami3l: I'm getting increasingly concerned about dropping ~~non-UTF8~~ non-Unicode support/tests. If you think about it, Rust's `str` is guaranteed to be valid UTF8, but we the string stored in the Windows PATH may not be one, and Rustup must be able to work correctly even if some other applications has added a ~~non-UTF8~~ non-Unicode path to the PATH, so the right abstraction to use in Microsoft's wrapper should be `OsStr`, not `str`.

`windows-registry` also uses the following translation function when it handles `&str` being passed in:

```rust
fn pcwstr<T: AsRef<str>>(value: T) -> Vec<u16> {
    value
        .as_ref()
        .encode_utf16()
        .chain(core::iter::once(0))
        .collect()
}
```
_https://github.com/microsoft/windows-rs/blob/2ec4e06d366c46ba9634316959f680ea3dc42a5a/crates/libs/registry/src/lib.rs#L45-L51_

This means the range of string values that we can write to the registry (valid UTF8 converted to UTF16) is a subset of the string values that can potentially be set in the registry (plain `Vec<u16>`s), which may lead to very bad consequences.

Also, as for getting `String` values from the registry:

```rs
            REG_SZ | REG_EXPAND_SZ => {
                let mut value = vec![0u16; len as usize / 2];

                let result = unsafe {
                    RegQueryValueExW(
                        self.0,
                        name.as_ptr(),
                        null(),
                        null_mut(),
                        value.as_mut_ptr() as _,
                        &mut len,
                    )
                };

                win32_error(result)?;
                Ok(Value::String(String::from_utf16_lossy(trim(&value))))
            }
```
_https://github.com/microsoft/windows-rs/blob/2ec4e06d366c46ba9634316959f680ea3dc42a5a/crates/libs/registry/src/key.rs#L165-L181_

... since it uses lossy conversion to UTF8, this simply cannot be a reliable source of truth. Imagine installing Rustup breaking some random user's PATH...

**TLDR: I think there's a API mistake in the `windows-registry` library that `winreg` didn't have. Before we address that issue, this PR cannot be merged.** (There's nothing fundamentally wrong from your side, don't worry @InfyniteHeap!)

@djc @ChrisDenton Is my analysis okay, are there any false claims about Windows (from a non-Windows dev)? What should be the next step moving forward? I think I'd still be glad to see such a change because, normally, Microsoft's library gets more support/social recognition and all...

djc: > TLDR: I think there's a API mistake in the `windows-registry` library that `winreg` didn't have.

Can you open an issue against upstream to check your reasoning? They're usually responsive.

djc: > ... since it uses lossy conversion to UTF8, this simply cannot be a reliable source of truth.

I'm not sure I agree? If the Windows registry API enforces valid UTF-16, then using `String::from_utf16_lossy()` is just the easiest `std` API to turn UTF-16 bytes into a `String`. I can't find clear references in https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-value-types but I wouldn't be surprised if it's the case.

rami3l: > If the Windows registry API enforces valid UTF-16, then using `String::from_utf16_lossy()` is just the easiest `std` API to turn UTF-16 bytes into a `String`.

@djc Exactly. However, the following statement in https://doc.rust-lang.org/std/ffi/struct.OsString.html makes me wonder if this is an appropriate occasion to interpret that `Vec<u16>` as valid UTF16 100% of the time:

> On Windows, strings are often arbitrary sequences of non-zero 16-bit values, interpreted as UTF-16 when it is valid to do so.

rami3l: @djc Oh, wait, the test data in the aforementioned test case is actually NOT valid UTF16LE:

https://github.com/rust-lang/rustup/blob/4c3ff9ce638e8402fbc2e8755d9e8a2cd148f6be/tests/suite/cli_paths.rs#L428-L435

```javascript
// Using JavaScript here for the sake of convenience...
console.log("\uD800\u0101\u0000".isWellFormed()) // false
```

ChrisDenton: Just catching up with this. I would only note that `Cargo` only really supports unicode paths (because they're stored as `String` in a lot of places). Tools like git or the linker also don't support broken UTF-16. Which is to say, I don't think this is necessarily a high priority. Something else will almost certainly break on invalid UTF-16 and arguably rustup should never be *writing* broken unicode to the registry even if it can read it.

rami3l: @ChrisDenton I see. Personally I still would like to wait for the final response from `windows-registry`'s side to actually make the decision, but thanks a lot for the information! If they decide that this change (to `OsStr`) is not worth it, let's just go with `&str`.

djc:
Sorry, but it turns out I had to make some changes to registry-related API in #3893. Hopefully it won't be too hard to rebase on top of my changes.

tests/suite/cli_paths.rs
ChrisDenton: The change seems fine to me. I just don't really see is as a big deal either way and not really a blocker. The only way non-unicode paths can happen on Windows is if something has gone badly wrong already (e.g. a broken conversion algorithm or some such). Unlike Unix, Windows paths are documented as being UTF-16 (the lack of validity checking is a kernel implementation detail). On Unix paths can be in any random encoding so it's more important there.

InfyniteHeap:
> Sorry, but it turns out I had to make some changes to registry-related API in #3893. Hopefully it won't be too hard to rebase on top of my changes.

Oh, I think I'd better replace dependency from scratch lest appear some unexpect problems.😅

InfyniteHeap:
This PR will keep closed until `windows-registry` updates.

rami3l:
@InfyniteHeap FYI https://github.com/microsoft/windows-rs/issues/3119 has been closed, so you might restart your work on this one!

InfyniteHeap:
> @InfyniteHeap FYI [microsoft/windows-rs#3119](https://github.com/microsoft/windows-rs/issues/3119) has been closed, so you might restart your work on this one!

Sounds great! I'll restart my work when new version of `windows-registry` is published on crates.io! :)

src/cli/self_update/windows.rs
```diff
@@ -511,27 +505,14 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
-        Ok(val) => {
-            if let Some(s) = from_winreg_value(&val) {
-                Ok(Some(s))
-            } else {
-                warn!(
-                    "the registry key HKEY_CURRENT_USER\\Environment\\PATH is not a string. \
-                       Not modifying the PATH variable"
-                );
-                Ok(None)
-            }
```
rami3l, line 532: This whole `Ok(None)` case has disappeared for no reason.

src/cli/self_update/windows.rs
```diff
@@ -929,62 +876,48 @@ mod tests {
 
     #[test]
     fn windows_path_regkey_type() {
-        // per issue #261, setting PATH should use REG_EXPAND_SZ.
```
rami3l, line 932: The actual semantics of `_apply_new_path()` and that of this test have been changed from saving a `REG_EXPAND_SZ` to `REG_SZ`, which looks like a clear regression to me (cc #261).

src/cli/self_update/windows.rs
```diff
@@ -929,62 +876,48 @@ mod tests {
...
         match reg_value {
             Ok(_) => panic!("key not deleted"),
-            Err(ref e) if e.kind() == io::ErrorKind::NotFound => {}
+            Err(ref e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => {}
             Err(ref e) => panic!("error {e}"),
         }
     }
 
     #[test]
+    #[ignore = "FIXME: I don't know how to make this test pass"]
```
rami3l: Please don't ignore tests like this; let them fail before it's either fixed or decided to be removed.

src/cli/self_update/windows.rs
```diff
@@ -995,15 +928,11 @@ mod tests {
         );
 
         let _guard = RegistryGuard::new(&USER_PATH);
-        let root = RegKey::predef(HKEY_CURRENT_USER);
-        let environment = root
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+
+        let environment = CURRENT_USER.create("Environment").unwrap();
+        environment
+            .set_hstring("PATH", &HSTRING::from_wide(&[0x12, 0x34]).unwrap())
```
rami3l: You should not set a `HSTRING` here, as the original code has been setting a binary (see https://docs.rs/windows-registry/latest/windows_registry/struct.Key.html#method.set_bytes).

src/cli/self_update/windows.rs
InfyniteHeap: I think `from_winreg_value()` and `to_winreg_bytes()` are useless, and `val` now has type `HSTRING`. Because of that, I removed these statements.

src/cli/self_update/windows.rs
InfyniteHeap: I briefly browsed the source code of `windows-registry` and found that it puts `REG_SZ` and `REG_EXPAND_SZ` together to handle strings.

src/cli/self_update/windows.rs
InfyniteHeap: I met this error when passing row bytes into registry:
屏幕截图 2024-07-05 105703
I think it could be fixed, but so far I haven't found a solutions to fix this error yet. To test other tests, I temporarily ignored this test.

InfyniteHeap: I met this error when passing raw bytes into registry:
屏幕截图 2024-07-05 105703
I think it could be fixed, but so far I haven't found solutions to fix this error yet. To test other tests, I temporarily ignored this test.

rami3l: @InfyniteHeap Add the `Ok(None)` case back in https://github.com/rust-lang/rustup/pull/3896/files/9977fd34910921f0e2e5e1236346d36f5f3ccaaf#r1666169848 and you'll see this error gone. This test is exactly for that part of the code.

src/cli/self_update/windows.rs
rami3l: > I think `from_winreg_value()` and `to_winreg_bytes()` are useless, and `val` now has type `HSTRING`. Because of that, I removed these statements.

@InfyniteHeap The thing is `PATH` can be of another type other than a string-like one, and you should be able to return `Ok(None)` in this case.

src/cli/self_update/windows.rs
rami3l: > I briefly browsed the source code of `windows-registry` and found that it puts `REG_SZ` and `REG_EXPAND_SZ` together to handle strings.

@InfyniteHeap I see. However this is not the reason not to "do the right thing" as some users might be affected with this change. You might consider calling raw APIs if necessary. Eventually this should become another feature request for `windows-registry`.

tests/suite/cli_paths.rs
```diff
@@ -410,43 +410,40 @@ mod windows {
...
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
-            .unwrap()
-            .set_raw_value("PATH", &reg_value)
+        let bytes = [
+            0x00, 0xD8, // leading surrogate
+            0x01, 0x01, // bogus trailing surrogate
+            0x00, 0x00, // null
+        ];
+        let key = CURRENT_USER.create("Environment").unwrap();
+        key.set_hstring("PATH", &HSTRING::from_wide(&bytes).unwrap())
```
InfyniteHeap, line 425: I guess `HSTRING::from_wide()` might eliminate the last two `0x00` and modify something, so the CI test will be failed.

src/cli/self_update/windows.rs
```diff
@@ -929,57 +912,79 @@ mod tests {
...
-            .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
-            .unwrap();
-        let path = environment.get_raw_value("PATH").unwrap();
-        assert_eq!(path.vtype, RegType::REG_EXPAND_SZ);
-        assert_eq!(super::to_winreg_bytes(wide("foo")), &path.bytes[..]);
+
+        let environment = CURRENT_USER.create("Environment").unwrap();
+        let path = environment.get_hstring("PATH").unwrap();
+
+        fn pcwstr<T: AsRef<str>>(value: T) -> Vec<u16> {
```
InfyniteHeap: This function was peeked from `windows-registry`. Maybe it should be removed and use `w!()` macro instead.

src/cli/self_update/windows.rs
```diff
@@ -929,57 +912,79 @@ mod tests {
...
+                .chain(core::iter::once(0))
+                .collect()
+        }
+
+        let handle = core::ptr::null_mut();
+        let mut ty = 0;
+        let mut len = 0;
+
+        unsafe {
+            RegCreateKeyExW(
```
InfyniteHeap, line 942: This function returns error code `0x800703F2`.

src/cli/self_update/windows.rs
rami3l: cc https://github.com/microsoft/windows-rs/issues/3148

src/cli/self_update/windows.rs
kennykerr: Yes, that would avoid a heap allocation.

InfyniteHeap: > Yes, that would avoid a heap allocation.

I tested but I cannot replace this function with `w!()`, because macros like `w!()` only accepts string literals, while this function accepts both literals and variables.

In addition, I tried modifying `w!()` macro to make it accepts variables, but failed to avoid heap allocations.😵

rami3l:
@InfyniteHeap According to https://github.com/microsoft/windows-rs/issues/3148#issuecomment-2263527164, you may now test how the upstream change works with this PR via `Cargo.toml`'s [`[patch]` section](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section).

InfyniteHeap:
> @InfyniteHeap According to [microsoft/windows-rs#3148 (comment)](https://github.com/microsoft/windows-rs/issues/3148#issuecomment-2263527164), you may now test how the upstream change works with this PR via `Cargo.toml`'s [`[patch]` section](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section).

`windows-rs` is not a single crate but a workspace with all windows crates included, but this way seems only accepts the git repo which is a single crate. So, I cloned `windows-rs` on my computer and manually set dependencies to test.

rami3l:
> > @InfyniteHeap According to [microsoft/windows-rs#3148 (comment)](https://github.com/microsoft/windows-rs/issues/3148#issuecomment-2263527164), you may now test how the upstream change works with this PR via `Cargo.toml`'s [`[patch]` section](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section).
> 
> `windows-rs` is not a single crate but a workspace with all windows crates included, but this way seems only accepts the git repo which is a single crate. So, I cloned `windows-rs` on my computer and manually set dependencies to test.

@InfyniteHeap Hmmm that shouldn't matter IIRC. I've done the same thing for `aws-lc-rs` just a few weeks ago, and that was also a monorepo. Cargo knows from the monorepo's  `Cargo.toml` how to get to the library by its name.

rami3l:
> Hmmm that shouldn't matter IIRC. I've done the same thing for `aws-lc-rs` just a few weeks ago, and that was also a monorepo. Cargo knows from the monorepo's `Cargo.toml` how to get to the library by its name.

@InfyniteHeap FWIW here's the commit: https://github.com/rust-lang/rustup/pull/3917/commits/3e7b23a4b0a2162c3271e10d2bd0365b3fd13755

InfyniteHeap:
The issue of `Cargo.toml` format will be resolved when code itself is ready to be merged into upstream repo.

tests/suite/cli_paths.rs
```diff
@@ -389,7 +391,9 @@ mod windows {
         let mut cx = CliTestContext::new(Scenario::Empty).await;
         let _guard = RegistryGuard::new(&USER_PATH).unwrap();
         let cfg_path = cx.config.cargodir.join("bin").display().to_string();
-        let get_path_ = || get_path().unwrap().unwrap().to_string();
+        let path = get_path().unwrap().unwrap();
+        let path = unsafe { from_raw_parts(path.as_ptr() as *const u16, path.len() / 2) };
```
InfyniteHeap: These two lines of code was partially peeked from [`winreg`](https://github.com/gentoo90/winreg-rs/blob/1c56127adf4e310b5c8ed2af89aa3f7b09d38371/src/types.rs#L34-L38).

tests/suite/cli_self_upd.rs
```diff
@@ -350,13 +350,12 @@ async fn update_overwrites_programs_display_version() {
...
+        .set_value(Some(PLACEHOLDER_VERSION.as_bytes()))
         .unwrap();
     cx.config.expect_ok(&["rustup", "self", "update"]).await;
-    assert_eq!(
-        USER_RUSTUP_VERSION.get_value::<String>().unwrap().unwrap(),
-        version,
-    );
+    assert_eq!(USER_RUSTUP_VERSION.get_value().unwrap().unwrap(), {
+        version.as_bytes()
+    });
```
InfyniteHeap, line 358: I don't know which side should I modify. CI failure pointed out that the left side is probably a UTF-16 string (seems also have a terminated `\0`?) which was represented with an `u8` array, while the right side is a pure UTF-8 string.

tests/suite/cli_paths.rs
djc: There must be a nicer way of doing this...

InfyniteHeap: Yes, and I am seeking for such this way. This is a temporary way.

src/cli/self_update/windows.rs
```diff
@@ -859,30 +844,28 @@ pub struct RegistryValueId {
...
+    fn set(&self, new: Option<&[u8]>) -> Result<()> {
+        let sub_key = CURRENT_USER.create(self.sub_key)?;
         match new {
-            Some(new) => sub_key.set_raw_value(self.value_name, new),
-            None => sub_key.delete_value(self.value_name),
+            Some(new) => Ok(sub_key.set_bytes(self.value_name, Type::String, new)?),
+            None => Ok(sub_key.remove_value(self.value_name)?),
         }
     }
 }
```
InfyniteHeap: This new implementation doesn't actually work as expected, because it cannot recover the proper key type after finishing testing.

InfyniteHeap:
Now all the CI tests about code itself are passed, so, I think it's suitable to review where should I modify.

src/cli/self_update/windows.rs
```diff
@@ -859,30 +845,28 @@ pub struct RegistryValueId {
...
+        self.set(ty, new)
     }
 
-    fn set(&self, new: Option<&RegValue>) -> io::Result<()> {
-        let sub_key = RegKey::predef(HKEY_CURRENT_USER)
-            .open_subkey_with_flags(self.sub_key, KEY_READ | KEY_WRITE)?;
-        match new {
-            Some(new) => sub_key.set_raw_value(self.value_name, new),
-            None => sub_key.delete_value(self.value_name),
+    fn set(&self, ty: Option<Type>, new: Option<&[u8]>) -> Result<()> {
```
rami3l: It looks like:
```rs
ty: Option<Type>, new: Option<&[u8]>,
```
... has introduced new input states (i.e. `None, Some(_)` and `Some(_), None`) that don't really make sense?

Would the following better convey what you intended to do?
```rs
new: Option<(Type, &[u8])>,
```

Same question to other places that use those two params.

src/cli/self_update/windows.rs
rami3l: @InfyniteHeap Sorry, is this still happening, or can we safely close this?

rami3l (commented):
@InfyniteHeap Thanks! Now this PR looks much better to me than the previous versions. However I still cannot merge it since you're using a Cargo patch.

For the moment being, to better prepare for this PR, it should be reorganized in the form of [atomic commits](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits) such that:

1. Every commit is a unit of change that only does one thing and does it well, and thus can be easily reverted;
2. The commits should be ordered properly so that every one gives a green CI.

You can refer to this comment (https://github.com/rust-lang/rustup/pull/3980#discussion_r1702752864) and the resulting commit history to better understand what is expected from your commit history.

Many thanks in advance!

Hint: In general, you should use Git's [interactive rebase](https://www.sitepoint.com/git-interactive-rebase-guide/) to rewrite the history.

InfyniteHeap:
> @InfyniteHeap Thanks! Now this PR looks much better to me than the previous versions. However I still cannot merge it since you're using a Cargo patch.
> 
> For the moment being, to better prepare for this PR, it should be reorganized in the form of [atomic commits](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits) such that:
> 
> 1. Every commit is a unit of change that only does one thing and does it well, and thus can be easily reverted;
> 2. The commits should be ordered properly so that every one gives a green CI.
> 
> You can refer to this comment (https://github.com/rust-lang/rustup/pull/3980#discussion_r1702752864) and the resulting commit history to better understand what is expected from your commit history.
> 
> Many thanks in advance!
> 
> Hint: In general, you should use Git's [interactive rebase](https://www.sitepoint.com/git-interactive-rebase-guide/) to rewrite the history.
> 
> 

Yes, I'll tidy up this PR once the new version of `windows-registry` is published.

As of the interactive rebase, I don't know why but when I rebasing commit histories, the git (exactly, the gnupg) always prompts me that I don't have a secret key, even though I indeed have one. I think whether it's because I am using a Windows computer. I'll try using WSL instead. That's why I frequently force-push (and this lead to this PR be frequently closed) commits. Apologize for this!

This is what I encountered:

屏幕截图 2024-08-05 120936

PS: I've installed gnupg from [this website](https://gpg4win.org/), and I successfully generated a gpg key. But it is strange that git itself cannot sign the commits, but github desktop and visual studio can.

rami3l:
@InfyniteHeap Have you tried docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key?platform=windows?

InfyniteHeap:
> @InfyniteHeap Have you tried docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key?platform=windows?

Yes, I tried (even several times), and that's what I annoyed: It still doesn't work! I also checked the `.gitconfig` file and can confirm that the key exists in it (if it doesn't exist, the github desktop and vs also cannot sign commits).

Sometimes, when I directly call `gpg` on my computer, it cannot be launched and the terminal shows me the IPC connecting errors.

InfyniteHeap:
> > @InfyniteHeap Have you tried docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key?platform=windows?
> 
> Yes, I tried (even several times), and that's what I annoyed: It still doesn't work! I also checked the `.gitconfig` file and confirm that the key exists in it (if it doesn't exist, the github desktop and vs also cannot sign commits).

Sometimes, when I directly call `gpg` on my computer, it cannot be launched and the terminal shows me the IPC connecting errors.

InfyniteHeap:
@rami3l I tested on WSL and succeeded in signing commits! Actually, I also cannot sign commits when first opening WSL, but after I referred to [this doc](https://gist.github.com/paolocarrasco/18ca8fe6e63490ae1be23e84a7039374), I eventually succeeded!😄

rami3l:
@InfyniteHeap Regarding the formatting errors, you're expected to use `taplo-cli` v0.9.3 for this. If you're using the VSCode extension, you can install a new version of `taplo-cli` yourself (via `cargo` or your package manager of choice) and choose not to use the (now outdated) bundled `taplo` in the TOML extension's configuration. `taplo`'s project owner is AFK and so new versions of the VSCode extension still cannot be published properly. This is the workaround I can think of for the moment being.

src/cli/self_update/windows.rs
```diff
@@ -859,30 +842,28 @@ pub struct RegistryValueId {
...
-            Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(None),
-            Err(e) => Err(e),
+            Err(e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => Ok(None),
+            Err(e) => Err(e.into()),
         }
     }
 
-    pub fn set_value(&self, new: Option<impl ToRegValue>) -> io::Result<()> {
-        self.set(new.map(|s| s.to_reg_value()).as_ref())
+    pub fn set_value(&self, new: Option<(Type, &[u8])>) -> Result<()> {
```
djc: Once this has the same argument type as `set()`, there's no point in retaining the method.

tests/suite/cli_paths.rs
rami3l: @djc I believe this is https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf16le_lossy, blocked on https://github.com/rust-lang/rust/issues/116258.

rami3l: @InfyniteHeap There's nothing wrong in bypassing nightly-exclusive APIs with unsafe, but adding a `TODO` comment will definitely help here :)

IIRC it so happens that Windows ARM64 and Windows x64 are all LE platforms, so a transmute like the one used above (`as *const u16`) should do the work 🤔

tests/suite/cli_self_upd.rs
```diff
@@ -350,12 +356,17 @@ async fn update_overwrites_programs_display_version() {
...
     cx.config.expect_ok(&["rustup", "self", "update"]).await;
+    let version = OsString::from(version)
+        .encode_wide()
+        .flat_map(|v| vec![v as u8, (v >> 8) as u8])
+        .chain([0, 0])
+        .collect::<Vec<u8>>();
     assert_eq!(
-        USER_RUSTUP_VERSION.get_value::<String>().unwrap().unwrap(),
-        version,
+        USER_RUSTUP_VERSION.get_value().unwrap().unwrap().as_ref(),
```
rami3l: It would also be helpful to `assert_eq` the `Type` of the `Value`s in your tests as the original code did. Same for the other test file that you've modified.

src/cli/self_update/windows.rs
rami3l: It would probably make sense to remove `(get|set)_value()` entirely in a separate commit now that they are not different from `get()` and `set()` respectively.

InfyniteHeap: I think I'd better open a seperate PR once this PR is merged.😵

rami3l: > I think I'd better open a seperate PR once this PR is merged.😵

@InfyniteHeap That'd be nice as well.

tests/suite/cli_paths.rs
rami3l: @InfyniteHeap Now that https://github.com/microsoft/windows-rs/pull/3190 is online, I think you should explore using the `HSTRING` type rather than using bytestrings directly.
See https://fgimian.github.io/blog/2022/10/23/working-with-strings-in-windows-rs/ for a brief introduction.

src/cli/self_update/windows.rs
```diff
@@ -633,47 +620,42 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
...
 
 /// This is used to decode the value of HKCU\Environment\PATH. If that key is
 /// not REG_SZ | REG_EXPAND_SZ then this returns None. The winreg library itself
 /// does a lossy unicode conversion.
-pub(crate) fn from_winreg_value(val: &winreg::RegValue) -> Option<Vec<u16>> {
-    use std::slice;
-
-    match val.vtype {
-        RegType::REG_SZ | RegType::REG_EXPAND_SZ => {
+pub(crate) fn from_winreg_value(val: Value) -> Option<Vec<u16>> {
```
rami3l: I wonder if `HSTRING` can also be used instead of `Vec<u16>` here to avoid `unsafe`. I'm not sure of this is feasible, just proposing a possibility... 🤔

InfyniteHeap: Maybe this can open a seperate PR? If this is feasible, I'll open one.

src/cli/self_update/windows.rs
InfyniteHeap: Yes, I know. I plan to open a seperate PR once this PR is merged.

rami3l:
@InfyniteHeap A gentle ping: this branch is in conflict now, which means you need to rebase this PR to get your CI checks back.

InfyniteHeap:
> @InfyniteHeap A gentle ping: this branch is in conflict now, which means you need to rebase this PR to get your CI checks back.

According to the conflict contents, I think this conflict will be automatically resolved once I use the newer version of `windows-registry` and remove the `[patch]` table. If it doesn't resolved, then I'll try rebasing this PR.

tests/suite/cli_paths.rs
kennykerr: Yes, there should be no reason to use `unsafe` for any registry value conversion.

InfyniteHeap:
> > @InfyniteHeap A gentle ping: this branch is in conflict now, which means you need to rebase this PR to get your CI checks back.
> 
> According to the conflict contents, I think this conflict will be automatically resolved once I use the newer version of `windows-registry` and remove the `[patch]` table. If it doesn't resolved, then I'll try rebasing this PR.

Ok, well, I've resolved it. :P

InfyniteHeap:
> > @InfyniteHeap A gentle ping: this branch is in conflict now, which means you need to rebase this PR to get your CI checks back.
> 
> According to the conflict contents, I think this conflict will be automatically resolved once I use the newer version of `windows-registry` and remove the `[patch]` table. If it doesn't resolved, then I'll try rebasing this PR.

Ok, well, I've resolved it. :P

InfyniteHeap:
This CI was failed, because while `reqwest` v0.12.7 (the newest version so far) adpoed `windows-registry` v0.2, this PR adopts the dependency coming from master branch. Comparing to `windows-registry` v0.2, the master-branch one had introduced some breaking API changes. So, I'll either try making `reqwest` use newer version of `windows-registry` after `windows-rs` publishing new release, or stick the version of `reqwest` to v0.12.5, which hadn't adopted `windows-registry` yet at that time.

kennykerr:
> This CI was failed, because while `reqwest` v0.12.7 (the newest version so far) adpoed `windows-registry` v0.2, this PR adopts the dependency coming from master branch. Comparing to `windows-registry` v0.2, the master-branch one had introduced some breaking API changes. So, I'll either try making `reqwest` use newer version of `windows-registry` after `windows-rs` publishing new release, or stick the version of `reqwest` to v0.12.5, which hadn't adopted `windows-registry` yet at that time.

You're probably running into this: https://github.com/microsoft/windows-rs/pull/3214

ChrisDenton:
It looks like `HSTRING` can simplify things and remove the need for `unsafe`. E.g.:

<details><summary>diff</summary>

```diff
diff --git a/src/cli/self_update/windows.rs b/src/cli/self_update/windows.rs
index 035d308c..c6c91fa0 100644
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -5,14 +5,13 @@ use std::io::Write;
 use std::os::windows::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 use std::process::Command;
-use std::slice;
 use std::sync::{Arc, Mutex};
 #[cfg(any(test, feature = "test"))]
 use std::sync::{LockResult, MutexGuard};

 use anyhow::{anyhow, Context, Result};
 use tracing::{info, warn};
-use windows_registry::{Key, Type, Value, CURRENT_USER};
+use windows_registry::{Key, Type, Value, CURRENT_USER, HSTRING};
 use windows_result::HRESULT;
 use windows_sys::Win32::Foundation::ERROR_FILE_NOT_FOUND;

@@ -480,7 +479,8 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
     if new_path.is_empty() {
         environment.remove_value("PATH")?;
     } else {
-        environment.set_bytes("PATH", Type::ExpandString, &to_winreg_bytes(new_path))?;
+        let new_path = HSTRING::from_wide(&new_path);
+        environment.set_expand_hstring("PATH", &new_path)?;
     }

     // Tell other processes to update their environment
@@ -620,12 +620,8 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
     uninstall_cmd.push(path);
     uninstall_cmd.push("\" self uninstall");

-    key.set_bytes(
-        "UninstallString",
-        Type::String,
-        &to_winreg_bytes(uninstall_cmd.encode_wide().collect()),
-    )
-    .context("Failed to set `UninstallString`")?;
+    key.set_hstring("UninstallString", &HSTRING::from(&uninstall_cmd))
+        .context("Failed to set `UninstallString`")?;
     key.set_string("DisplayName", "Rustup: the Rust toolchain installer")
         .context("Failed to set `DisplayName`")?;
     do_update_programs_display_version(env!("CARGO_PKG_VERSION"))?;
@@ -641,22 +637,13 @@ pub(crate) fn do_remove_from_programs() -> Result<()> {
     }
 }

-/// Convert a vector UCS-2 chars to a null-terminated UCS-2 string in bytes
-pub(crate) fn to_winreg_bytes(mut v: Vec<u16>) -> Vec<u8> {
-    v.push(0);
-    unsafe { slice::from_raw_parts(v.as_ptr().cast::<u8>(), v.len() * 2).to_vec() }
-}
-
 /// This is used to decode the value of HKCU\Environment\PATH. If that key is
 /// not REG_SZ | REG_EXPAND_SZ then this returns None. The winreg library itself
 /// does a lossy unicode conversion.
 pub(crate) fn from_winreg_value(val: Value) -> Option<Vec<u16>> {
     match val.ty() {
         Type::String | Type::ExpandString => {
-            // Copied from winreg
-            let mut words = unsafe {
-                slice::from_raw_parts(val.as_ptr().cast::<u16>(), val.as_ref().len() / 2).to_owned()
-            };
+            let mut words = Vec::from(val.as_wide());
             while words.last() == Some(&0) {
                 words.pop();
             }
@@ -922,7 +909,7 @@ mod tests {
         let environment = CURRENT_USER.create("Environment").unwrap();
         let path = environment.get_value("PATH").unwrap();
         assert_eq!(path.ty(), Type::ExpandString);
-        assert_eq!(super::to_winreg_bytes(wide("foo")), path.as_ref());
+        assert_eq!(HSTRING::from("foo").as_wide(), path.as_wide());
     }

     #[test]
@@ -932,11 +919,7 @@ mod tests {
         let _guard = RegistryGuard::new(&USER_PATH);
         let environment = CURRENT_USER.create("Environment").unwrap();
         environment
-            .set_bytes(
-                "PATH",
-                Type::ExpandString,
-                &super::to_winreg_bytes(wide("foo")),
-            )
+            .set_expand_hstring("PATH", &HSTRING::from("foo"))
             .unwrap();

         {
```
</diff>

A further improvement would be to use the `HSTRING` or `HStringBuilder` types throughout instead of `Vec<u16>`. But I've not investigated what that would look like.

ChrisDenton:
In case it helps, I created a branch that's rebased on master with my changes in case anyone wants to build on it: https://github.com/rust-lang/rustup/compare/master...ChrisDenton:rustup:winreg (note that it temporarily disables rustls because building their TLS is a right pain, feel free to drop that commit though).

InfyniteHeap:
> It looks like `HSTRING` can simplify things and remove the need for `unsafe`. E.g.:
> 
> diff

I initially want to do this after this PR is merged (see https://github.com/rust-lang/rustup/pull/3896#discussion_r1705266338). But if needed, I can also do this in the same PR.😊

djc:
Since upstream is waiting for feedback on their API before they release, I think we should try to remove all unsafe code related to the registry in this PR, and give them feedback on where this is impossible with their current API.

(Please rebase this branch on master instead of appending merge commits.)

InfyniteHeap:
These modifications are not completed yet, and I'll solve them later.

InfyniteHeap:
> Since upstream is waiting for feedback on their API before they release, I think we should try to remove all unsafe code related to the registry in this PR, and give them feedback on where this is impossible with their current API.
> 
> (Please rebase this branch on master instead of appending merge commits.)

I think this will be resolved when I squash the commits into a single one.

InfyniteHeap:
Oh no! I think I screwed this up!

InfyniteHeap:
It seems that the `openssl-sys` has some issues that prevent the commits from being passed.

rami3l:
> It seems that the `openssl-sys` has some issues that prevent the commits from being passed.

@InfyniteHeap You shouldn't update `Cargo.lock` for dependencies irrelevant to this PR. That's automatically handled by renovate on our side.

InfyniteHeap:
> > It seems that the `openssl-sys` has some issues that prevent the commits from being passed.
> 
> @InfyniteHeap You shouldn't update `Cargo.lock` for dependencies irrelevant to this PR. That's automatically handled by renovate on our side.

It seems that I should only open VS Code to let it automatically update `Cargo.lock` instead of manually run `cargo update`...

InfyniteHeap:
> > It seems that the `openssl-sys` has some issues that prevent the commits from being passed.
> 
> @InfyniteHeap You shouldn't update `Cargo.lock` for dependencies irrelevant to this PR. That's automatically handled by renovate on our side.

It seems that I should only open VS Code to let it automatically update `Cargo.lock` instead of manually running `cargo update`.

InfyniteHeap:
> > It seems that the `openssl-sys` has some issues that prevent the commits from being passed.
> 
> @InfyniteHeap You shouldn't update `Cargo.lock` for dependencies irrelevant to this PR. That's automatically handled by renovate on our side.

It seems that I should only open VS Code to let it automatically update `Cargo.lock` instead of manually running `cargo update`...

src/cli/self_update/windows.rs
```diff
@@ -510,83 +504,78 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
     let mut len = path_str.len();
-    if old_path.get(idx + path_str.len()) == Some(&(b';' as u16)) {
+    if old_path.as_wide().get(idx + path_str.len()) == Some(&(b';' as u16)) {
         len += 1;
     }
 
-    let mut new_path = old_path[..idx].to_owned();
-    new_path.extend_from_slice(&old_path[idx + len..]);
+    let mut new_path = old_path.as_wide()[..idx].to_owned();
+    new_path.extend_from_slice(&old_path.as_wide()[idx + len..]);
```
InfyniteHeap: @kennykerr I think it might be better to implement `Index` trait for `HSTRING`. What do you think?

src/cli/self_update/windows.rs
```diff
@@ -510,83 +504,78 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
     // Don't leave a trailing ; though, we don't want an empty string in the
     // path.
     if new_path.last() == Some(&(b';' as u16)) {
```
InfyniteHeap, line 563: Although I can convert `HSTRING` to `String` (with some probable losses) to judge which character is the last one in this string, I think adding `ends_with()` method for `HSTRING` seems better.

src/cli/self_update/windows.rs
```diff
@@ -510,83 +504,78 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
         .windows(path_str.len())
-        .position(|path| path == path_str)?;
+        .position(|path| path == path_str.as_wide())?;
     // If there's a trailing semicolon (likely, since we probably added one
     // during install), include that in the substring to remove. We don't search
     // for that to find the string, because if it's the last string in the path,
     // there may not be.
     let mut len = path_str.len();
-    if old_path.get(idx + path_str.len()) == Some(&(b';' as u16)) {
+    if old_path.as_wide().get(idx + path_str.len()) == Some(&(b';' as u16)) {
```
InfyniteHeap: It is better to implement `get()` method for `HSTRING` to avoid type conversions.

rami3l (commented):
This PR is getting even clearer than the previous iteration, modulo some eventual rebases (esp. after @kennykerr's possible API adjustments). Good job!

PS: As a favor, please consider listing @ChrisDenton as a [co-author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) of some (or all) of your commits when applicable :)

djc (approved):
This looks good to me, thanks for all the work!

kennykerr:
I could implement Deref in place of as_wide and that would provide all of this directly.

rami3l:
Current status: Waiting for a new upstream release to ship https://github.com/microsoft/windows-rs/issues/3148...

kennykerr:
Here's the `Deref` implementation for `HSTRING`. https://github.com/microsoft/windows-rs/pull/3291

src/cli/self_update/windows.rs
```diff
@@ -510,55 +505,49 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
 // Returns None if the existing old_path does not need changing, otherwise
 // prepends the path_str to old_path, handling empty old_path appropriately.
-fn _add_to_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>> {
+fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
     if old_path.is_empty() {
         Some(path_str)
     } else if old_path
         .windows(path_str.len())
-        .any(|path| path == path_str)
+        .any(|path| path == path_str.deref())
```
djc: Calling `deref()` "by hand" is pretty unidiomatic, we'd usually spell it more like `*path_str`.

src/cli/self_update/windows.rs
```diff
@@ -510,55 +505,49 @@ fn _apply_new_path(new_path: Option<Vec<u16>>) -> Result<()> {
...
     }
 }
 
 // Returns None if the existing old_path does not need changing
-fn _remove_from_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>> {
+fn _remove_from_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
     let idx = old_path
         .windows(path_str.len())
-        .position(|path| path == path_str)?;
+        .position(|path| path == path_str.deref())?;
```
djc: Here, too.

src/cli/self_update/windows.rs
```diff
@@ -575,18 +564,17 @@ fn _remove_from_path(old_path: Vec<u16>, path_str: Vec<u16>) -> Option<Vec<u16>>
...
 where
-    F: FnOnce(Vec<u16>, Vec<u16>) -> Option<Vec<u16>>,
+    F: FnOnce(HSTRING, HSTRING) -> Option<HSTRING>,
 {
     let windows_path = get_windows_path_var()?;
     let mut path_str = process.cargo_home()?;
     path_str.push("bin");
-    Ok(windows_path
-        .and_then(|old_path| f(old_path, OsString::from(path_str).encode_wide().collect())))
+    Ok(windows_path.and_then(|old_path| f(old_path, HSTRING::from(path_str.into_os_string()))))
```
djc: There seems to be a `From<&Path> for HSTRING`, so I think we can make this `HSTRING::from(&path_str)`?

src/cli/self_update/windows.rs
```diff
@@ -616,11 +603,9 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
...
 
     // Don't overwrite registry if Rustup is already installed
-    let prev = key
-        .get_raw_value("UninstallString")
-        .map(|val| from_winreg_value(&val));
-    if let Ok(Some(s)) = prev {
-        let mut path = PathBuf::from(OsString::from_wide(&s));
+    let prev = key.get_hstring("UninstallString");
+    if let Ok(s) = prev {
+        let mut path = PathBuf::from(s.to_os_string());
```
djc, line 607: I see we're also converting from `HSTRING` to `PathBuf`.

src/cli/self_update/windows.rs
```diff
@@ -916,71 +867,64 @@ mod tests {
         ];
         let cargo_home = wide(r"c:\users\example\.cargo\bin");
         let final_path = [&cargo_home, &[b';' as u16][..], &initial_path].join(&[][..]);
+        let final_path = HSTRING::from_wide(&final_path);
```
djc: @kennykerr do you think it would make sense to add `push_str()` to `HSTRING`?

djc (commented):
@kennykerr what's the rationale for using all-uppercase names for stuff like `HSTRING` and `HRESULT`? Are you trying to match the upstream names exactly? It looks rather unidiomatic.

src/cli/self_update/windows.rs
ChrisDenton: That would need to be on `HStringBuilder`. A `HSTRING` is an immutable string (more or less the equivalent of `Box<[u16]>` but without mutability) whereas `HStringBuilder` is a `Vec` equivalent.

djc: Ah, I hadn't noticed `HStringBuilder` yet -- so the type is called `HSTRING` but the builder is called `HStringBuilder`?

So I guess maybe we should be using a `HStringBuilder` here instead of joining `u16` slices?

kennykerr:
> what's the rationale for using all-uppercase names for stuff like HSTRING and HRESULT? Are you trying to match the upstream names exactly? It looks rather unidiomatic.

There are countless identifiers in the Windows API. It's just not very practical to come up with different names - it has been attempted. It also makes it a lot harder to cross-reference documentation and source code for equivalent types.

src/cli/self_update/windows.rs
InfyniteHeap: `PathBuf` doesn't implement `From<HSTRING>`, so I have to use this conversion temporarily.

InfyniteHeap:
What about these changes?

src/cli/self_update/windows.rs
djc: You can likely coerce the PathBuf to Path.

InfyniteHeap:
What about these changes?

djc (approved):
I think this is looking great. @InfyniteHeap and @kennykerr thanks for all the work! Would be nice to get that windows-registry release out so that we can move forward with this.

kennykerr:
Coming up: https://github.com/microsoft/windows-rs/pull/3293

src/cli/self_update/windows.rs
InfyniteHeap: Implicit deref coercion cannot be used there because `HSTRING::from()` method includes a lot of overloads.

kennykerr:
Version 0.3.0 of the [windows-registry](https://crates.io/crates/windows-registry) crate has been published. https://github.com/microsoft/windows-rs/releases/tag/0.60.0

djc:
Can you squash all of your changes into a single commit?

InfyniteHeap:
> Can you squash all of your changes into a single commit?

I've done this.

rami3l (approved):
LGTM, and thanks for your work @InfyniteHeap!

PS: You may proceed with cleanup PRs such as the one discussed in https://github.com/rust-lang/rustup/pull/3896#discussion_r1704153219 shortly after.

src/cli/self_update/windows.rs
InfyniteHeap: @rami3l Since this PR was merged, should I still make this change and create a PR?

PR #3899, opened 2024-06-24
Add missing period in warning while checking existing Rust installations

Uaitt:
It seems like there's a missing period in the warning that gets triggered if an user already has an existing installation of Rust.

Screenshot 2024-06-24 at 12 12 01


rami3l (approved):
Many thanks!

PR #3901, opened 2024-06-24
feat: add hint to run `rustup self` when err desc is `self`

### LINKED ISSUE #3812: `rustup update self` should suggest `rustup self update`

Xerxes-2:
Closes https://github.com/rust-lang/rustup/issues/3812.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -3,6 +3,7 @@ use std::fmt;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::ExitStatus;
+use std::str::FromStr;
 
 use anyhow::{anyhow, Error, Result};
 use clap::{builder::PossibleValue, Args, CommandFactory, Parser, Subcommand, ValueEnum};
@@ -135,7 +136,7 @@ enum RustupSubcmd {
     )]
     Update {
         /// Toolchain name, such as 'stable', 'nightly', or '1.8.0'. For more information see `rustup help toolchain`
-        #[arg(num_args = 1..)]
+        #[arg(num_args = 1.., value_parser = update_toolchain_value_parser)]
         toolchain: Vec<PartialToolchainDesc>,
 
         /// Don't perform self update when running the `rustup update` command
@@ -256,6 +257,14 @@ enum RustupSubcmd {
         #[arg(default_value = "rustup")]
         command: CompletionCommand,
     },
+}
+
+fn update_toolchain_value_parser(s: &str) -> Result<PartialToolchainDesc> {
+    PartialToolchainDesc::from_str(s).inspect_err(|_| {
+        if s == "self" {
+            info!("if you meant to update rustup itself, use `rustup self update`");
+        }
+    })
 }
 
 #[derive(Debug, Subcommand)]
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -457,7 +457,13 @@ impl<'a> Toolchain<'a> {
                     fs::remove_dir_all(&path)?;
                     true
                 } else {
+                    let name = name.to_string();
                     info!("no toolchain installed for '{name}'");
+                    if name == "self" {
+                        info!(
+                            "if you meant to uninstall rustup itself, use `rustup self uninstall`"
+                        );
+                    }
                     false
                 }
             }
--- a/tests/suite/cli_misc.rs
+++ b/tests/suite/cli_misc.rs
@@ -1180,3 +1180,27 @@ async fn toolchain_link_then_list_verbose() {
         .expect_stdout_ok(&["rustup", "toolchain", "list", "-v"], "/custom-1")
         .await;
 }
+
+#[tokio::test]
+async fn update_self_smart_guess() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+    let out = cx.config.run("rustup", &["update", "self"], &[]).await;
+    let invalid_toolchain = out.stderr.contains("invalid toolchain name");
+    if !out.ok && invalid_toolchain {
+        assert!(out
+            .stderr
+            .contains("if you meant to update rustup itself, use `rustup self update`"))
+    }
+}
+
+#[tokio::test]
+async fn uninstall_self_smart_guess() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+    let out = cx.config.run("rustup", &["uninstall", "self"], &[]).await;
+    let no_toolchain_installed = out.stdout.contains("no toolchain installed");
+    if out.ok && no_toolchain_installed {
+        assert!(out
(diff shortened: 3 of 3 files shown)


src/dist/mod.rs
```diff
@@ -319,7 +320,11 @@ impl FromStr for ParsedToolchainDesc {
         if let Some(d) = d {
             Ok(d)
         } else {
-            Err(RustupError::InvalidToolchainName(desc.to_string()).into())
+            let desc_string = desc.to_string();
+            if desc_string.as_str() == "self" {
+                info!("If you meant to modify rustup itself, try run `rustup self`");
+            }
+            Err(RustupError::InvalidToolchainName(desc_string).into())
```
djc: I think it would be better to do this near the top of the method? We can just check for `desc == "self"` and return the error immediately, which would save a bunch of work trying to parse.

rami3l: @djc My suggestion is that since `FromStr` is used in many places other than `update` and `uninstall` (such as `install`), this will cause many false positives, and instead we should modify the `value_parser` for these two commands only.

src/cli/rustup_mode.rs
```diff
@@ -258,6 +261,15 @@ enum RustupSubcmd {
     },
 }
 
+fn update_toolchain_value_parser(s: &str) -> clap::error::Result<PartialToolchainDesc> {
+    PartialToolchainDesc::from_str(s).map_err(|e| {
+        if s == "self" {
+            info!("if you meant to update rustup itself, use `rustup self update`");
+        }
+        clap::Error::raw(ErrorKind::ValueValidation, e)
```
rami3l: You should throw the original error instead of creating a new one. See https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err.

src/toolchain.rs
```diff
@@ -457,7 +457,13 @@ impl<'a> Toolchain<'a> {
                     fs::remove_dir_all(&path)?;
                     true
                 } else {
+                    let name = name.to_string();
                     info!("no toolchain installed for '{name}'");
+                    if name.as_str() == "self" {
```
rami3l: Will this work?

```suggestion
                    if &name == "self" {
```

Xerxes-2: Seems work even without `&`

rami3l (commented):
Thanks! IMHO there are two things left to do:

1. Squash all the feature changes together, as `rustup` encourages atomic commits, and the current changes are seemingly small enough to be squash into a single commit.
2. In a new commit, add some regression tests in `tests/suite/cli_misc.rs` to prevent future regressions.

src/toolchain.rs
rami3l: > Seems work even without `&`

@Xerxes-2 That's better then.

Xerxes-2:
> Thanks! IMHO there are two things left to do:
> 
> 1. Squash all the feature changes together, as `rustup` encourages atomic commits, and the current changes are seemingly small enough to be squash into a single commit.
> 2. In a new commit, add some regression tests in `tests/suite/cli_misc.rs` to prevent future regressions.

Fine, I might do it later cuz I'm now in a holiday.

src/dist/mod.rs
djc: But `self` is still invalid in all those places, right -- worst case this will provide a clearer error message?

rami3l: @djc I'm afraid that will make the error message less clear. `rustup install self` doesn't make sense and if the info leads the user to `rustup self` they won't find anything related to `install` there.

djc: Sure, but users are presumably also much less likely to invoke `rustup install self`. I still think it would be better to embed this check inside the `FromStr` impl where it is implicitly used in any future code that wants to consume toolchain names instead of remembering that we have to add an extra check everywhere, we can make the error message slightly more generic to make sure it doesn't lose clarity.

rami3l: @djc I still feel uneasy about adding side effects in conversions like `FromStr`... @ChrisDenton What do you think?

djc: I would argue that logging is pretty much "unobservable" as side effects go, so while I agree that in general side effects in something like a `FromStr` implementations could be a bit iffy, I think logging is an exception to this.

Would you have problems with adding logging to other kinds of "pure" functions?

rami3l: @djc Actually... it's not about purity, it's about the fact that conversions can be used in various places across the codebase, many of which are not easily detectable via call hierarchy graph `rust-analyzer` gives us, and I'm afraid of this potentially adding a log line somewhere unexpected. Since the very nature of this feature is to add a hint, I guess it'd be better to get the hint only when we are explicitly expecting for it.

For example, `rustup uninstall` uses `ResolvableToolchainName`, and its conversions relies on the conversions defined for `PartialToolchainDesc` first, and that of `CustomToolchainName` as a fallback. Let's say we're still using https://github.com/rust-lang/rustup/commit/a743b3adc19ec949761a236fe484b0e34f0f44fb, and it so happens that there's a custom toolchain named `self`, then the first conversion will definitely fail. However, before the fallback happens, the log line is already printed out.

Xerxes-2:
@rami3l Can I close this PR and open a new one?

rami3l:
> @rami3l Can I close this PR and open a new one?

@Xerxes-2 Yes, but why you'd want to do that instead of doing an interactive rebase directly on this one?

Xerxes-2:
> > @rami3l Can I close this PR and open a new one?
> 
> @Xerxes-2 Yes, but why you'd want to do that instead of doing an interactive rebase directly on this one?

Because I accidentally merged master into this branch, may not be able to squash together all the feature commits

rami3l:
> > > @rami3l Can I close this PR and open a new one?
> > 
> > @Xerxes-2 Yes, but why you'd want to do that instead of doing an interactive rebase directly on this one?
> 
> Because I accidentally merged master into this branch, may not be able to squash together all the feature commits

@Xerxes-2 AFAIK merge commits will be automatically dropped when doing an interactive rebase. See https://stackoverflow.com/a/4784082.

djc (approved):

rami3l (approved):

rami3l:
@Xerxes-2 Thanks again for helping out!

PR #3898, opened 2024-06-22
feat(download/rustls): use `aws-lc` instead of `ring`

rami3l:
Closes #3820 by replacing the `ring` backend with `aws-lc`.

Fortunately, with https://github.com/seanmonstar/reqwest/pull/2301, it's now possible to **totally disable `ring` at compile time**.

Tested locally on macOS 14.5 via the following command with positive outcome:
```console
> RUSTUP_LOG=TRACE RUSTUP_FORCE_ARG0=rustup rustup run stable cargo run -- update
```

PR #3903, opened 2024-06-25
feat(download/rustls): use `rustls-platform-verifier`

rami3l:
Split from #3898, as requested in https://github.com/rustls/rustls-platform-verifier/issues/101.

This in theory should not cause any CI breakages 🙏

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -294,12 +294,16 @@ pub mod reqwest_be {
     compile_error!("Must select a reqwest TLS backend");
 
     use std::io;
+    #[cfg(feature = "reqwest-rustls-tls")]
+    use std::sync::Arc;
     use std::time::Duration;
 
     use anyhow::{anyhow, Context, Result};
     #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
     use once_cell::sync::Lazy;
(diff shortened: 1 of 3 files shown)


download/src/lib.rs
```diff
@@ -293,16 +293,22 @@ pub mod reqwest_be {
     ))]
     compile_error!("Must select a reqwest TLS backend");
 
-    use std::io;
-    use std::time::Duration;
+    use std::{io, time::Duration};
```
djc: As in https://github.com/rust-lang/rustup/pull/3904#discussion_r1652138338.

djc (approved):
Nice!

rami3l:
@djc Unfortunately, this has broken our FreeBSD CI.

djc:
I guess that would be because of https://github.com/rustls/rustls-native-certs/issues/28. Apparently somehow we were previously pulling in webpki-roots (perhaps in addition rustls-native-certs)?

rami3l:
> I guess that would be because of [rustls/rustls-native-certs#28](https://github.com/rustls/rustls-native-certs/issues/28). Apparently somehow we were previously pulling in webpki-roots (perhaps in addition rustls-native-certs)?

@djc Sorry, let's move the discussion to #3908. I should not have replied on a merged PR, my bad.

PR #3904, opened 2024-06-25
ci(windows): run `cargo all-features`

rami3l:
Addresses the warning found in https://github.com/rust-lang/rustup/actions/runs/9655583919/job/26631715641?pr=3903:

```rust
warning: unused import: `self`
 --> src\cli\self_update\windows.rs:4:15
  |
4 | use std::io::{self, Write};
  |               ^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused imports: `LockResult`, `MutexGuard`
 --> src\cli\self_update\windows.rs:8:22
  |
8 | use std::sync::{Arc, LockResult, Mutex, MutexGuard};
  |                      ^^^^^^^^^^         ^^^^^^^^^^

warning: unused imports: `FromRegValue`, `ToRegValue`
  --> src\cli\self_update\windows.rs:23:21
   |
23 | use winreg::types::{FromRegValue, ToRegValue};
   |                     ^^^^^^^^^^^^  ^^^^^^^^^^

warning: `rustup` (lib) generated 3 warnings (run `cargo fix --lib -p rustup` to apply 3 suggestions)
```

The CI runtime overhead is +11min on Windows. It's not a bottleneck though, since our macOS x64 full CI takes even longer.

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -1,14 +1,22 @@
 use std::env::{consts::EXE_SUFFIX, split_paths};
 use std::ffi::{OsStr, OsString};
 use std::fmt;
-use std::io::{self, Write};
+#[cfg(any(test, feature = "test"))]
+use std::io;
+use std::io::Write;
 use std::os::windows::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 use std::process::Command;
-use std::sync::{Arc, LockResult, Mutex, MutexGuard};
+use std::sync::{Arc, Mutex};
+#[cfg(any(test, feature = "test"))]
+use std::sync::{LockResult, MutexGuard};
 
 use anyhow::{anyhow, Context, Result};
 use tracing::{info, warn};
+use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
+#[cfg(any(test, feature = "test"))]
+use winreg::types::{FromRegValue, ToRegValue};
+use winreg::{RegKey, RegValue};
 
 use super::super::errors::*;
 use super::common;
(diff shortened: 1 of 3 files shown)


src/cli/self_update/windows.rs
```diff
@@ -1,11 +1,16 @@
...
 use std::ffi::{OsStr, OsString};
 use std::fmt;
-use std::io::{self, Write};
+use std::io::Write;
 use std::os::windows::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 use std::process::Command;
-use std::sync::{Arc, LockResult, Mutex, MutexGuard};
+use std::sync::{Arc, Mutex};
+
```
djc: Here and in other PRs, I would prefer that we keep imports with compilation guards in the block that they belong to (so in this case, keeping them with the `std` imports). That also has the advantage that it keeps non-conditional imports and conditional imports from the same module together.

djc (approved):

src/cli/self_update/windows.rs
rami3l: @djc Would you mind elaborating on this? Maybe give an example? There are many different import styles being used in the project and it's hard to say which one is canonical... Of course we can change that from now on.

rami3l: @djc You could even change the style guide in the dev guide book to solidify the new style: for the moment being it's just `std-extern-crate` and nothing more.

djc: I just mean you're adding empty lines between imports that all come from `std`. I would make this change as follows:

```rust
use std::env::{consts::EXE_SUFFIX, split_paths};
use std::ffi::{OsStr, OsString};
use std::fmt;
use std::io;
use std::io::Write;
use std::os::windows::ffi::{OsStrExt, OsStringExt};
use std::path::Path;
use std::process::Command;
use std::sync::{Arc, Mutex};
use std::sync::{LockResult, MutexGuard};

// etc
```

and 

```rust
use winreg::enums::{RegType, HKEY_CURRENT_USER, KEY_READ, KEY_WRITE};
use winreg::types::{FromRegValue, ToRegValue};
use winreg::{RegKey, RegValue};
```

(And there's a pre-existing issue here where the winreg imports really blong in the extern block.)

PR #3907, opened 2024-06-25
Disable automatic self updates in CI environments

### LINKED ISSUE #2886: error: rustup is not installed at '/home/user/.cargo' when $RUSTUP_HOME is set to something other than the former
**Problem**
1. Rustup cannot update when `CARGO_HOME` is not the Linux default `~/.cargo`
1. Then `rustup self uninstall` fails to work
1. At https://www.rust-lang.org/tools/install, there are no further instructions, or a link to those instructions, on how to uninstall `rustup`.

**Steps**

```bash
20:02 $ echo $RUSTUP_HOME
/home/<snip>/.rustup
✔ ~/src/<snip> [port ↑·2|✚ 2…1]
20:02 $ echo $CARGO_HOME
/home/<snip>/.local/share/cargo
^[[A✔ ~/src/<snip> [port ↑·2|✚ 2…1]
20:02 $ which rustup
/home/<snip>/.local/share/cargo/bin/rustup
✔ ~/src/<snip> [port ↑·2|✚ 2…1]
20:02 $ rustup self update
error: rustup is not installed at '/home/<snip>/.cargo'
```

**Possible Solution(s)**
No idea. We seem to be in a Mexican-standoff.

**Notes**

Output of `rustup --version`:

```bash
$ rustup --version
rustup 1.23.1 (3df2264a9 2020-11-30)
info: The currently active `rustc` version is `rustc 1.56.0-nightly (5ad7389bd 2021-08-06)`
```

Output of `rustup show`:

```bash
20:14 $ rustup show
rustup home:  /home/<snip>/.rustup

installed toolchains
--------------------

active toolchain
----------------

rustc 1.56.0-nightly (5ad7389bd 2021-08-06)
```

### LINKED ISSUE #3824: Disable Rustup's self update by default if the `CI` environment variable is detected
Proposed as an alternative to #3821, mitigates #3709:

> Rustup could default to not auto update if the `CI` environment variable is present. This is simple both to implement and explain to users.

_Originally posted by @ChrisDenton in https://github.com/rust-lang/rustup/issues/3821#issuecomment-2106251789_

## Notes
- The relevant [user guide section](https://github.com/rust-lang/rustup/blob/454e8dc1bf31f20821cc8ac3228ea1f35b23d072/doc/user-guide/src/environment-variables.md) should be adjusted accordingly.
- This change might cause Rustup's own tests to fail, so removing `CI` from the environment variables is probably required before actually running those tests.

djc:
Fixes #2886, fixes #3824.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -418,6 +418,12 @@ impl<'a> Cfg<'a> {
     }
 
     pub(crate) fn get_self_update_mode(&self) -> Result<SelfUpdateMode> {
+        if self.process.var("CI").is_ok() && self.process.var("RUSTUP_CI").is_err() {
+            // If we're in CI (but not rustup's own CI, which wants to test this stuff!),
+            // disable automatic self updates.
+            return Ok(SelfUpdateMode::Disable);
+        }
+
         self.settings_file.with(|s| {
             Ok(match s.auto_self_update {
                 Some(mode) => mode,
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -77,5 +77,5 @@ build_test() {
 if [ -z "$SKIP_TESTS" ]; then
   target_cargo run --features test -- --dump-testament
   build_test build
-  build_test test
+  RUSTUP_CI=1 build_test test
 fi


src/config.rs
```diff
@@ -418,6 +418,10 @@ impl<'a> Cfg<'a> {
     }
 
     pub(crate) fn get_self_update_mode(&self) -> Result<SelfUpdateMode> {
+        if self.process.var("CI").is_ok() {
+            return Ok(SelfUpdateMode::Disable);
```
djc, line 424: Alternatively, we could return `CheckOnly`?

rami3l: @djc I guess disabled would still be better. It seems to me that the whole auto self update thing in the CI context is only introducing overheads.

djc: That's what I thought as well. 👍

rami3l (approved):
Thanks! LGTM, as long as the CI is doing well...

src/config.rs
```diff
@@ -418,6 +418,12 @@ impl<'a> Cfg<'a> {
     }
 
     pub(crate) fn get_self_update_mode(&self) -> Result<SelfUpdateMode> {
+        if self.process.var("CI").is_ok() {
+            if !self.process.var("RUSTUP_CI").is_ok() {
```
rami3l: Can we flatten this nested if?

smoelius:
FWIW, I think this is a really good idea. Thank you!

hoodmane:
I see that this fixes the problem but rustup hasn't had a release in 9 months. Could rustup make a release with this commit?

rami3l:
> I see that this fixes the problem but rustup hasn't had a release in 9 months. Could rustup make a release with this commit?

@hoodmane We are currently in the middle of releasing v1.28.0 (https://github.com/rust-lang/rustup/issues/3998, https://internals.rust-lang.org/t/seeking-beta-testers-for-rustup-v1-28-0). For the moment being we are actively collecting feedback on our beta release. Thank you for your patience :)

hoodmane:
How do I install rustup beta?

hoodmane:
I guess I should read your link and I'd find out.

hoodmane:
Thanks @rami3l!

PR #3909, opened 2024-06-25
Upload Rustup build artifacts to new S3 bucket

jdno:
We are implementing a [new release process](https://github.com/rust-lang/rustup/pull/3844) for Rustup, which changes slightly how build artifacts are uploaded. Going forward, every commit merged into `master` will produce a full set of release artifacts that will be stored in a new S3 bucket. The new bucket allows us to remove access for CI to the release bucket, which improves our security posture. And uploading every commit to `master` will make it easier to test new releases.

### DIFF
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -169,13 +169,17 @@ jobs: # skip-master skip-pr skip-stable
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-west-1
+      - name: Configure AWS credentials
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::890664054962:role/ci--rust-lang--rustup
+          aws-region: us-east-1
       - name: Deploy build to rustup-builds bucket for release team
-        # if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
         run: |
-          aws --debug s3 cp --recursive dist s3://rustup-builds/${{ github.sha }}
+          aws --debug s3 cp --recursive deploy/ s3://rustup-builds/${{ github.sha }}
         env:
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-east-1
       - name: Clear the cargo caches
         run: |
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -169,6 +169,14 @@ jobs: # skip-master skip-pr skip-stable
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-west-1
+      - name: Deploy build to rustup-builds bucket for release team
+        # if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
+        run: |
+          aws --debug s3 cp --recursive dist s3://rustup-builds/${{ github.sha }}
+        env:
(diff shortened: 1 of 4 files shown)


jdno:
The tracking issue for the new release process and its work can be found here: https://github.com/rust-lang/simpleinfra/issues/420

rami3l (approved):
Thanks a lot!

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -169,6 +169,18 @@ jobs: # skip-master skip-pr skip-stable
...
+      - name: Configure AWS credentials
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::890664054962:role/ci--rust-lang--rustup
+          aws-region: us-east-1
+      - name: Deploy build to rustup-builds bucket for release team
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
+        run: |
+          aws --debug s3 cp --recursive deploy/ s3://rustup-builds/${{ github.sha }}
```
rami3l, line 181: There has been a similar `aws s3 cp --recursive deploy/ s3://dev-static-rust-lang-org/rustup/` command right above. 
I guess this is intentional, in that you want to make additive changes for now and remove the old one only when you're sure the new approach works?

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -169,6 +169,18 @@ jobs: # skip-master skip-pr skip-stable
...
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-west-1
+      - name: Configure AWS credentials
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::890664054962:role/ci--rust-lang--rustup
+          aws-region: us-east-1
+      - name: Deploy build to rustup-builds bucket for release team
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
```
rami3l, line 179: Are you sure this is `master`? Currently our release builds (beta or not) are based on the `stable` branch. Examples should be found right above...

ci/actions-templates/linux-builds-template.yaml
jdno: Oh yeah, sorry! I forgot to mention this above. The changes right now are all additive as to not break the currently release process. Once the new one has been implemented and tested, I'll clean up the old steps.

ci/actions-templates/linux-builds-template.yaml
jdno: I was certain that in one of the many discussion around this topic it was suggested to upload every push on `master`. But as much as I'm looking for it, I can't find it anymore. So maybe I am mixing up something.

@Mark-Simulacrum Any preferences from the release team's perspective on whether to upload from `master` or `stable`?

Mark-Simulacrum: I'd prefer that we match what we do for rustc, and upload every commit on master too. We probably still need to upload from stable so that backports etc work.

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -169,6 +169,14 @@ jobs: # skip-master skip-pr skip-stable
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-west-1
+      - name: Deploy build to rustup-builds bucket for release team
+        # if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
```
djc: Do we want to leave these comments around?

djc (approved):
Nice!

ci/actions-templates/linux-builds-template.yaml
jdno: I'm not sure how backports in Rustup work, but the dev guide suggests fast-forward merges for regular releases. So both `master` and `stable` should point to the same commit and imo uploading only from `master` should be enough. But I'm happy to be corrected if my understanding is wrong...

rami3l: @jdno Yeah, since rustup is a not so active repo (compared to rust) we currently don't do backports at all.

ci/actions-templates/linux-builds-template.yaml
jdno: This was in a work-in-progress. The final version didn't include them anymore.

jdno:
Can I merge this and then iterate on the configuration if there are any issues?

djc:
> Can I merge this and then iterate on the configuration if there are any issues?

Sounds good to me!

jdno:
Hm, the merge queue build failed trying to fetch a base image from `rust-lang/rust`. This looks like a fluke, so I'm just gonna try again and see what happens... 🤷‍♂️

PR #3910, opened 2024-06-25
implements quiet flag in `rustup-init.sh`

### LINKED ISSUE #3350: shouldn't install script be quiet?

vic1707:
as discussed in #3350

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -27,6 +27,8 @@ set -u
 
 # If RUSTUP_UPDATE_ROOT is unset or empty, default it.
 RUSTUP_UPDATE_ROOT="${RUSTUP_UPDATE_ROOT:-https://static.rust-lang.org/rustup}"
+# Set quiet as a global for ease of use
+RUSTUP_QUIET=no
 
 # NOTICE: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
@@ -114,6 +116,9 @@ main() {
                 usage
                 exit 0
                 ;;
+            --quiet)
+                RUSTUP_QUIET=yes
+                ;;
             *)
                 OPTIND=1
                 if [ "${arg%%--*}" = "" ]; then
@@ -121,11 +126,14 @@ main() {
                     # don't attempt to interpret it.
                     continue
                 fi
-                while getopts :hy sub_arg "$arg"; do
+                while getopts :hqy sub_arg "$arg"; do
                     case "$sub_arg" in
                         h)
                             usage
                             exit 0
+                            ;;
+                        q)
+                            RUSTUP_QUIET=yes
                             ;;
                         y)
                             # user wants to skip the prompt --
@@ -140,18 +148,14 @@ main() {
         esac
     done
 
-    if $_ansi_escapes_are_valid; then
-        printf "\33[1minfo:\33[0m downloading installer\n" 1>&2
-    else
-        printf '%s\n' 'info: downloading installer' 1>&2
-    fi
+    say 'downloading installer'
 
     ensure mkdir -p "$_dir"
     ensure downloader "$_url" "$_file" "$_arch"
     ensure chmod u+x "$_file"
     if [ ! -x "$_file" ]; then
-        printf '%s\n' "Cannot execute $_file (likely because of mounting /tmp as noexec)." 1>&2
-        printf '%s\n' "Please copy the file to a location where you can execute binaries and run ./rustup-init${_ext}." 1>&2
+        err "Cannot execute $_file (likely because of mounting /tmp as noexec)." 
+        err "Please copy the file to a location where you can execute binaries and run ./rustup-init${_ext}." 
         exit 1
     fi
 
@@ -162,6 +166,7 @@ main() {
         # to explicitly connect /dev/tty to the installer's stdin.
         if [ ! -t 1 ]; then
             err "Unable to run interactively. Run with -y to accept defaults, --help for additional options"
+            exit 1;
         fi
 
         ignore "$_file" "$@" < /dev/tty
@@ -184,14 +189,14 @@ get_current_exe() {
     if test -L /proc/self/exe ; then
         _current_exe=/proc/self/exe
     else
-        printf '%s\n' "Warning: Unable to find /proc/self/exe. System architecture detection might be inaccurate." 1>&2
+        warn "Unable to find /proc/self/exe. System architecture detection might be inaccurate."
         if test -n "$SHELL" ; then
             _current_exe=$SHELL
         else
             need_cmd /bin/sh
             _current_exe=/bin/sh
         fi
-        printf '%s\n' "Warning: Falling back to $_current_exe." 1>&2
+        warn "Falling back to $_current_exe."
     fi
     echo "$_current_exe"
 }
@@ -213,6 +218,7 @@ get_bitness() {
         echo 64
     else
         err "unknown platform bitness"
+        exit 1;
     fi
 }
 
@@ -247,6 +253,7 @@ get_endianness() {
         echo "${cputype}${suffix_eb}"
     else
         err "unknown platform endianness"
+        exit 1
     fi
 }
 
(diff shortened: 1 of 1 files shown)


rami3l (dismissed):
LGTM, and thanks a lot!

rustup-init.sh
```diff
@@ -141,17 +149,17 @@ main() {
     done
 
     if $_ansi_escapes_are_valid; then
-        printf "\33[1minfo:\33[0m downloading installer\n" 1>&2
+        say "\33[1minfo:\33[0m downloading installer\n"
     else
-        printf '%s\n' 'info: downloading installer' 1>&2
+        say '%s\n' 'info: downloading installer'
```
rami3l: It seems that there are many places where this trailing `'%s\n'` doesn't make sense anymore (and is even wrong).

Tried on `dash`:

```bash
warn() {
    printf 'rustup: %s\n' "$1" >&2
}

warn '%s\n' "Cannot execute (likely because of mounting /tmp as noexec)."
```

This gives `rustup: %s\n`.

rustup-init.sh
```diff
@@ -705,8 +719,8 @@ check_help_for() {
                     ;;
                 *)
                     # Unknown product version, warn and continue
-                    echo "warn: Detected unknown macOS major version: $(sw_vers -productVersion)"
-                    echo "Warning TLS capabilities detection may fail"
+                    warn "warn: Detected unknown macOS major version: $(sw_vers -productVersion)"
+                    warn "Warning TLS capabilities detection may fail"
```
rami3l: Pre-existing issue, but while we're at it, maybe we can unify all `printf`/`echo` calls related to warnings here by replacing them with `warn`, and the `Warning:`/`Warning` prefixes with `warn:`?

vic1707: Yeah no problem 👌
Might as well do it at the warn function level 🙃

rustup-init.sh
vic1707: I'll clean these up as well 👌

rustup-init.sh
rami3l: > Might as well do it at the warn function level 🙃

@vic1707 That's okay as well, as long as you make sure the `say` calls aren't affected...

vic1707:
Hi @rami3l may I ask you for an updated review :grin: ?
I'd like to have your opinion on some of the choices I've made, in particular the fact that the `err` function no longer exits on its own.

rustup-init.sh
```diff
@@ -141,17 +149,17 @@ main() {
     done
 
     if $_ansi_escapes_are_valid; then
-        printf "\33[1minfo:\33[0m downloading installer\n" 1>&2
+        say "\33[1minfo:\33[0m downloading installer"
```
rami3l: You're printing out `rustup - INFO: info:` which is not very ideal...

Two things to be considered in general:
1. Use the normal `info: `, `warn: ` and `error: ` prefixes (so no `rustup -`) to align with the binary's logger. Keep the bolded style to make it clearer when ANSI escapes are available, exactly as demonstrated by this `if-else` structure.
2. Some warnings are long and span across several lines. It's expected to make it easier to read line by line if you're adding `warn: ` to the beginning of all of them, like in https://github.com/rust-lang/rustup/pull/3910#discussion_r1659810074. Alternatively, you can stick to the old output by keeping the prefix out of functions like `warn()`, it's your call.

rami3l (commented):
Thanks again! You're closer than ever to the finish line :)

rustup-init.sh
```diff
@@ -501,14 +512,14 @@ get_architecture() {
...
-                         echo "You will be able to add an x32 target after installation by running" 1>&2
-                         echo "  rustup target add x86_64-unknown-linux-gnux32" 1>&2
+                         err "This host is running an x32 userland; as it stands, x32 support is poor," 
+                         err "and there isn't a native toolchain -- you will have to install" 
+                         err "multiarch compatibility with i686 and/or amd64, then select one" 
+                         err "by re-running this script with the RUSTUP_CPUTYPE environment variable" 
+                         err "set to i686 or x86_64, respectively." 
+                         err 
+                         err "You will be able to add an x32 target after installation by running" 
+                         err "  rustup target add x86_64-unknown-linux-gnux32"
```
rami3l: If you're going for `error: ` on each line, then each line should make a sentence of its own. For example, you'd write this message like so:

```suggestion
                         err "This host is running an x32 userland, for which no native toolchain is provided."
                         err "You will have to install multiarch compatibility with i686 or amd64."
                         err "To do so, set the RUSTUP_CPUTYPE environment variable set to i686 or amd64 and re-run this script."
                         err "You will be able to add an x32 target after installation by running `rustup target add x86_64-unknown-linux-gnux32`." 
```

rustup-init.sh
```diff
@@ -294,19 +305,17 @@ ensure_loongarch_uapi() {
...
-            echo "warn: Cannot determine current system's ABI flavor, continuing anyway." >&2
-            echo >&2
-            echo 'Note that the official Rust distribution only works with the upstream' >&2
-            echo 'kernel ABI.  Installation will fail if your running kernel happens to be' >&2
-            echo 'incompatible.' >&2
+            warn "Cannot determine current system's ABI flavor, continuing anyway."
+            warn
+            warn 'Note that the official Rust distribution only works with the upstream'
+            warn 'kernel ABI.  Installation will fail if your running kernel happens to be'
+            warn 'incompatible.'
```
rami3l: So ideally those two sections should be reformatted to one sentence per line as well.

rustup-init.sh
vic1707: missed that one you're right, thought I had checked all of them 😓 ! 
I also thought that printing `rustup - info` and all was weird, being able to drop the `rustup -` part sounds good to me ^^

rustup-init.sh
vic1707: ok thx for the suggestion

rustup-init.sh
vic1707: no problem 👍

vic1707:
I think we're good now 🤔

rami3l (approved):
Nice work!

PR #3911, opened 2024-06-26
feat(rustup-init): set log level to `WARN` on `-q` if `RUSTUP_LOG` is unset

### LINKED ISSUE #3900: Use `LevelFilter::Warn` for `stderr` if `-q` is present in `rustup-init`
> I think it would be fine to make `-q` more general and use a `LevelFilter::Warn` if it's enabled.

_Originally posted by @djc in https://github.com/rust-lang/rustup/issues/3350#issuecomment-2186289993_

rami3l:
Closes #3900.

### DIFF
--- a/src/bin/rustup-init.rs
+++ b/src/bin/rustup-init.rs
@@ -21,6 +21,7 @@ use cfg_if::cfg_if;
 use rs_tracing::{
     close_trace_file, close_trace_file_internal, open_trace_file, trace_to_file_internal,
 };
+use tracing_subscriber::{reload::Handle, EnvFilter, Registry};
 
 use rustup::cli::common;
 use rustup::cli::proxy_mode;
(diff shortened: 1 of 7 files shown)


src/bin/rustup-init.rs
rami3l, line 1: @djc As requested. Sincerely I don't like this solution since we need to pass `console_filter: Handle<EnvFilter, Registry>` all the way down. Is there a better way though, considering the fact that we currently still initialize the subscriber before running `clap`?

djc: Perhaps we can just decide to initialize the subscriber after running clap?

rami3l: @djc The problem is the span info is already being logged even before entering clap, so it's not practical to initialze the subscriber that late.

djc: Can we change the the filter level for the subscriber after the fact? IMO that's arguably the best solution.

rami3l: @djc That's what this PR does! `console_filter: Handle<_, _>` is a reference to the filter that allows modifying its level at runtime with "a little overhead", as stated in the docs of `tracing_subscriber`.

djc: Ahh, yes -- sorry. I think this is reasonable for now. In the future, it might make sense to change `Process` to be a `struct` that contains the current `OSProcess`/`TestProcess` enum insight, at which point we could for example store the `Handle` in the `Process`.

djc (approved):

PR #3915, opened 2024-06-27
test(manifestation): replace `run_future()` with `TestContext`

rami3l:
As a continuation of https://github.com/rust-lang/rustup/pull/3367 and #3856, this PR makes all tests in `dist::manifestation::tests` that rely on `run_future()` async, thus the latter can be fully removed.

### DIFF
--- a/src/dist/manifestation/tests.rs
+++ b/src/dist/manifestation/tests.rs
@@ -3,15 +3,19 @@
 #![allow(clippy::type_complexity)]
 
 use std::{
-    cell::Cell, collections::HashMap, env, fs, future::Future, path::Path, str::FromStr, sync::Arc,
+    cell::Cell,
+    collections::HashMap,
+    env, fs,
+    path::{Path, PathBuf},
+    str::FromStr,
+    sync::Arc,
 };
 
 use anyhow::{anyhow, Result};
-use tokio::{runtime::Handle, task};
 use url::Url;
 
 use crate::{
-    currentprocess::{Process, TestProcess},
+    currentprocess::TestProcess,
     dist::{
         download::DownloadCfg,
         manifest::{Component, Manifest},
@@ -25,24 +29,6 @@ use crate::{
 };
 
 const SHA256_HASH_LEN: usize = 64;
-
-/// Temporary thunk to support asyncifying from underneath.
-fn run_future<F, R, E>(f: F) -> Result<R, E>
-where
-    F: Future<Output = Result<R, E>>,
-    E: std::convert::From<std::io::Error>,
-{
-    match Handle::try_current() {
-        Ok(current) => {
-            // hide the asyncness for now.
-            task::block_in_place(|| current.block_on(f))
-        }
-        Err(_) => {
-            // Make a runtime to hide the asyncness.
-            tokio::runtime::Runtime::new()?.block_on(f)
-        }
-    }
-}
 
 // Creates a mock dist server populated with some test data
 fn create_mock_dist_server(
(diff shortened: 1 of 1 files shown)


src/dist/manifestation/tests.rs
```diff
@@ -2,13 +2,15 @@
...
 #![allow(clippy::type_complexity)]
 
-use std::collections::HashMap;
-use std::env;
-use std::fs;
-use std::path::Path;
-use std::str::FromStr;
-use std::sync::Arc;
-use std::{cell::Cell, future::Future};
+use std::{
```
djc, line 5: Can you make a separate commit here that changes the formatting?

src/dist/manifestation/tests.rs
```diff
@@ -31,6 +33,7 @@ use crate::{
 const SHA256_HASH_LEN: usize = 64;
 
 /// Temporary thunk to support asyncifying from underneath.
+#[deprecated]
```
djc: I think we should squash this commit and the next one; they're not logically independent (this just adds unused code which would trigger warnings when running `cargo check` at this commit) and it's easier to review the new code when it's in the same commit as the old code being removed.

djc (commented):
Awesome! Would like some tweaks before I review this in a bit more detail...

rami3l:
@djc I don't know how you'd like to proceed with your reviews, but just FYI I find this PR a perfect fit for language-aware diffs like SemanticDiff: https://app.semanticdiff.com/gh/rust-lang/rustup/pull/3915/commits

src/dist/manifestation/tests.rs
```diff
@@ -388,108 +349,28 @@ fn rename_component_new() {
...
+    let cx = TestContext::from_dist_server(mock_dist_server, url, GZOnly);
+
+    let adds = [Component::new(
+        "bobo".to_string(),
+        Some(TargetTriple::new("x86_64-apple-darwin")),
+        true,
+    )];
+    // Install the basics from day 1
+    change_channel_date(&cx.url, "nightly", "2016-02-01");
+    cx.update_from_dist(&[], &[], false).await.unwrap();
```
djc, line 361: Probably not worth doing this now, but in the future, consider splitting the commits by first isolating the `TestContext` type to contain all the arguments given to these closures (generally by the same name) and passing the `TestContext` type to the testing closures as before (and perhaps also attaching the `update_from_dist()` function to that `TestContext` type). That would have made this a lot easier to review.

src/dist/manifestation/tests.rs
```diff
@@ -535,195 +400,205 @@ impl Compressions {
...
-    comps: Compressions,
-    f: &dyn Fn(&Url, &ToolchainDesc, &InstallPrefix, &DownloadCfg<'_>, &temp::Context),
-) {
-    server.write(
-        &[MockManifestVersion::V2],
-        comps.enable_xz(),
-        comps.enable_zst(),
-    );
+impl TestContext {
+    fn new(edit: Option<&dyn Fn(&str, &mut MockChannel)>, comps: Compressions) -> Self {
```
djc, line 414: I think this could benefit from a more descriptive name. I feel like now the relation between `new()` and `from_dist_server()` is the opposite from what I expect (`new()` is generally lower level than other constructors, but in this case it is arguably higher level.)

djc (approved):

djc:
> @djc I don't know how you'd like to proceed with your reviews, but just FYI I find this PR a perfect fit for language-aware diffs like SemanticDiff: https://app.semanticdiff.com/gh/rust-lang/rustup/pull/3915/commits

Ooohhh, that is pretty awesome. Should try to use that more.

src/dist/manifestation/tests.rs
rami3l: @djc Hmmm, `setup` used to be a short name and I changed it to `new` simply because of the "more uses = terser name" convention.

That said, what about `Self::from_edit()`? I'm really bad at naming 🤣

djc: Hmm, on review I feel like `new()` is probably reasonable here based on how it is used, even if it is a little counterintuitive to me. Let's just merge it?

PR #3918, opened 2024-06-29
download: make curl-backend buildable again

klensy:


klensy:
Oh, rustup broken too, `cargo b --no-default-features --features=curl-backend`:
```
error[E0599]: no variant or associated item named `Reqwest` found for enum `Backend` in the current scope
   --> src\utils\utils.rs:241:19
    |
241 |         (Backend::Reqwest(tls_backend), Notification::UsingReqwest)
    |                   ^^^^^^^ variant or associated item not found in `Backend`

```

rami3l:
@klensy Thanks for spotting the problem! However, the solution you've proposed seems inappropriate.

I tried changing this function:

https://github.com/rust-lang/rustup/blob/ad668d9a3b2fd8b9c8cfe644289f8257515c7724/download/src/lib.rs#L503-L510

... to `pub async fn` and `cargo clippy --no-default-features --features=curl-backend` has passed without any issue, so I encourage you to try if that works for you.

If it does, please feel free to replace the current fix with the one mentioned above :)

djc:
Would like to hear your use case for wanting to build rustup with only the curl backend!

rami3l:
I'm closing this now in favor of #3922. Please feel free to ping us back if you still have other concerns :)

klensy:
> Would like to hear your use case for wanting to build rustup with only the curl backend!

Oh, i just found that existing feature with curl backend wasn't worked, that's all, with something like `cargo check-all-features`.

PR #3920, opened 2024-07-02
feat(dist): add notes for `stable` and `beta` in `components_missing_msg()`

rami3l:
Prepares for `rls`' removal in the release channel.

Continuation of #3453, as requested by @onur-ozkan in https://github.com/rust-lang/rust/pull/126856#issuecomment-2197990717.

> The best thing we could do is to add a note saying some components may no longer be available (this note currently doesn't apply for stables), but:
>
> 1. We'll have to wait for v1.28 to ship it, and we're currently in the middle of the release cycle (we have limited bandwidth so we decided not to do backports), and
> 2. We should not hardcode the list of removed components in Rustup
> 
> _https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/excluding.20rls.20from.20the.20release/near/447968780_

~~The error message is not final, since I'd still like to add a link that provides more info about removed components (which is `rls` this time):~~ The message now includes a link to https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components.

> ... and for this I suggest adding a section in the Rust docs somewhere, and our notes should advise the user to check out the section above to see whether the component in question is missing as intended from what date (and if that is the case, they can choose to unblock the update with a clear consent by removing the component).
>
> But in any way we cannot simply skip RLS in an update if it's previously installed, since that would be a real backwards compatibility problem.
> _https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/excluding.20rls.20from.20the.20release/near/447968780_

## Concerns

- [x] ~~To which document should we add the list of removed components?~~ To the Rustup User Guide, see above for details.
- [x] ~~Should the note be applied to `nightly` as well?~~ No, as https://github.com/rust-lang/rustup-components-history/pull/54 will add a link to https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components on the components history page.

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -47,7 +47,7 @@ pub(crate) use triple::*;
 pub static DEFAULT_DIST_SERVER: &str = "https://static.rust-lang.org";
 
 const TOOLSTATE_MSG: &str =
-    "If you require these components, please install and use the latest successful build version,\n\
+    "If you require these components, please install and use the latest successfully built version,\n\
      which you can find at <https://rust-lang.github.io/rustup-components-history>.\n\nAfter determining \
      the correct date, install it with a command such as:\n\n    \
      rustup toolchain install nightly-2018-12-27\n\n\
@@ -93,7 +93,7 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
                 .join(", ");
             let _ = write!(
                 buf,
-                "some components unavailable for download for channel '{toolchain}': {cs_str}"
+                "some components are unavailable for download for channel '{toolchain}': {cs_str}"
             );
 
             if toolchain.starts_with("nightly") {
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -99,6 +99,23 @@ Then you can use the toolchain with commands such as:
 
     cargo +nightly-2018-12-27 build"
         );
+    } else if ["beta", "stable"].iter().any(|&p| toolchain.starts_with(p)) {
+        let _ = write!(
+            buf,
+            "\
+One or many components listed above might have been permanently removed from newer versions
+of the official Rust distribution due to deprecation.
+
+You can find the list of removed components at
+<https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components>.
+
+If you are updating an existing toolchain, after determining the deprecated component(s)
+in question, please remove them with a command such as:
+
+    rustup component remove --toolchain {toolchain} <COMPONENT>...
+
+After that, you should be able to continue with the update as usual.",
+        );
     }
 
     String::from_utf8(buf).unwrap()
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -75,15 +75,6 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
                 c.description(manifest),
                 toolchain,
             );
-
-            if toolchain.starts_with("nightly") {
-                let _ = write!(buf, "{nightly_tips}");
-            }
-
-            let _ = write!(
-                buf,
-                "If you don't need the component, you could try a minimal installation with:\n\n{suggestion}\n\n{TOOLSTATE_MSG}"
-            );
         }
         cs => {
             let cs_str = cs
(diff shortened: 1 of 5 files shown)


src/dist/mod.rs
```diff
@@ -45,7 +45,7 @@ pub static DEFAULT_DIST_SERVER: &str = "https://static.rust-lang.org";
 pub(crate) static DEFAULT_DIST_ROOT: &str = "https://static.rust-lang.org/dist";
 
 const TOOLSTATE_MSG: &str =
-    "If you require these components, please install and use the latest successful build version,\n\
+    "If you require the component(s), please install and use the latest successful build version,\n\
```
djc: It's a little hard to say without full context, but I think "these components" would read better?

src/dist/mod.rs
```diff
@@ -97,6 +97,20 @@ Then you can use the toolchain with commands such as:
 
     cargo +nightly-2018-12-27 build"
         );
+    } else if ["beta", "stable"].iter().any(|&p| toolchain.starts_with(p)) {
+        let _ = write!(
+            buf,
+            "\
+One or many components listed above might have been permanently removed from newer versions
```
djc, line 106: The wording here seems a little vague. Can we do something more specific, perhaps by checking for the specific `rls` component?

rami3l: @djc As explained above, I intend to learn from the message right above (regarding nightlies) and to add a link pointing to a list of removed components somewhere online (see the `Concerns` section of the original post). As a result, not mentioning `rls` explicitly here is intentional, since this gives the release team the freedom to remove components in the middle of any Rustup release cycle once this fix lands, as long as the online list can be updated in time.

src/dist/mod.rs
```diff
@@ -91,18 +72,45 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
...
+If you don't need the component(s), you could try a minimal installation with:
 
-            let _ = write!(
-                buf,
-                "If you don't need the components, you could try a minimal installation with:\n\n{suggestion}\n\n{TOOLSTATE_MSG}"
-            );
-        }
+    rustup toolchain add {toolchain} --profile minimal
+
+If you require the component(s), please install and use the latest successful build version,
```
rbtcollins:
```suggestion
If you require the component(s), please install and use the latest successfully built version,
```

src/dist/mod.rs
rami3l: @djc Fixed!

src/dist/mod.rs
rami3l: @djc I've decided to use https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components, and I've updated this message to include that link.

rami3l:
> * [ ]  Should the note be applied to `nightly` as well?

@onur-ozkan My apologies for getting a bit sidetracked previously! I would like to know how https://rust-lang.github.io/rustup-components-history will behave on the removal of `rls`, since we include this link in the message for nightly users.

IMHO, it should also include a link to https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components in case someone wonders why `rls` is gone in consecutive nightlies starting from a certain date.

onur-ozkan:
> @onur-ozkan My apologies for getting a bit sidetracked previously! I would like to know how https://rust-lang.github.io/rustup-components-history will behave on the removal of rls, since we include this link in the message for nightly users.

I couldn't find any hard-coded logic in https://github.com/rust-lang/rustup-components-history. I think it will just disappear without causing any issues.

> IMHO, it should also include a link to https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components in case someone wonders why rls is gone in consecutive nightlies starting from a certain date.

That sounds great but I don't know who maintains it.

rami3l:
> > IMHO, it should also include a link to https://rust-lang.github.io/rustup/devel/concepts/components.html#previous-components in case someone wonders why rls is gone in consecutive nightlies starting from a certain date.
> 
> That sounds great but I don't know who maintains it.

cc @Mark-Simulacrum @mexus

Mark-Simulacrum:
I'm happy to merge small PRs to rustup-components-history.

src/dist/mod.rs
```diff
@@ -99,6 +99,23 @@ Then you can use the toolchain with commands such as:
 
     cargo +nightly-2018-12-27 build"
         );
+    } else if ["beta", "stable"].iter().any(|&p| toolchain.starts_with(p)) {
```
djc, line 102: Should we do `beta-` and `stable-` instead to more precisely match the toolchain name boundary?

djc (approved):

src/dist/mod.rs
rami3l: @djc Hmmm, I don't think what you've suggested will work with `+stable`.

PR #3924, opened 2024-07-07
feat(cli): warn when removing the default/active toolchain

### LINKED ISSUE #3319: Give a warning before uninstalling the active/default toolchain, or the last/host target for a toolchain

rami3l:
Continuation of #3637, addresses the concern 3) in https://github.com/rust-lang/rustup/issues/3319.

Closes #3319.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1281,8 +1281,25 @@ async fn toolchain_link(
 }
 
 fn toolchain_remove(cfg: &mut Cfg<'_>, opts: UninstallOpts) -> Result<utils::ExitCode> {
+    let default_toolchain = cfg.get_default().ok().flatten();
+    let active_toolchain = cfg.find_active_toolchain().ok().flatten().map(|(it, _)| it);
+
     for toolchain_name in &opts.toolchain {
         let toolchain_name = toolchain_name.resolve(&cfg.get_default_host_triple()?)?;
+
+        if active_toolchain
+            .as_ref()
+            .is_some_and(|n| n == &toolchain_name)
+        {
+            warn!("removing the active toolchain; a toolchain override will be required for running Rust tools");
+        }
+        if default_toolchain
+            .as_ref()
+            .is_some_and(|n| n == &toolchain_name)
+        {
+            warn!("removing the default toolchain; proc-macros and build scripts might no longer build");
+        }
+
         Toolchain::ensure_removed(cfg, (&toolchain_name).into())?;
     }
     Ok(utils::ExitCode(0))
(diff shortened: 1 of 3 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1281,8 +1281,25 @@ async fn toolchain_link(
...
+    let active_toolchain = cfg.find_active_toolchain().ok().flatten().map(|(it, _)| it);
+
     for toolchain_name in &opts.toolchain {
         let toolchain_name = toolchain_name.resolve(&cfg.get_default_host_triple()?)?;
+
+        if active_toolchain
+            .as_ref()
+            .is_some_and(|n| n == &toolchain_name)
+        {
+            warn!("it seems that you are removing the active toolchain, is this intentional?");
```
djc: For a non-prompt warning, let's get rid of the question mark? Also is there a reason to soften the language with "it seems that"? I think this could just be "removing the active toolchain"?

src/cli/rustup_mode.rs
```diff
@@ -1281,8 +1281,25 @@ async fn toolchain_link(
...
+        }
+        if default_toolchain
+            .as_ref()
+            .is_some_and(|n| n == &toolchain_name)
+        {
+            warn!("after removing the default toolchain, proc-macros and build scripts might no longer build");
```
djc: Maybe slightly reword as "removing the default toolchain; proc-macros and build scripts might no longer build"?

djc (approved):

src/cli/rustup_mode.rs
rami3l: @djc I want to attract the user's attention and at least indicate that they are doing something probably undesirable. I've adjusted the wording nonetheless.

src/cli/rustup_mode.rs
rami3l: @djc This follows the style established by the previous two notes with #3637, but I'd be glad to shorten all three.

djc (approved):

PR #3923, opened 2024-07-07
feat(dist): refine suggestions regarding manifest checksum mismatches

rami3l:
Closes #3885.

>  I've made the following update to the error messages nonetheless:
> 
> ### Before
> 
> ```console
>     info: syncing channel updates for 'nightly-aarch64-apple-darwin'
>     info: update not yet available, sorry! try again later
>     error: toolchain 'nightly-aarch64-apple-darwin' is not installable
> ```
> 
> ### After
> 
> ```console
>     info: syncing channel updates for 'nightly-aarch64-apple-darwin'
>     warn: checksum failed for <SNIP>, expected: 'aaaaaaaaaa0f87d2df5e0cff356df307b6649a7b0b4effdf4b6dbd15070206de', calculated: '907f97b96e0f87d2df5e0cff356df307b6649a7b0b4effdf4b6dbd15070206de'
>     info: if you are on the official release server, this is most likely due to an update happening right now, please try again later
>     info: if you are on a third-party release server, this might indicate a problem with the server's configuration
>     error: toolchain 'nightly-aarch64-apple-darwin' is not installable
> ```
> 
> _https://github.com/rust-lang/rustup/issues/3885#issuecomment-2212318125_

PR #3925, opened 2024-07-08
Clean up usage of if_not_empty()

djc:
Triggered by #3923, this is a different way of slicing how `if_not_empty()` is used. What do you think?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,9 +1,9 @@
 use std::borrow::Cow;
 use std::fmt::{self, Debug, Display};
-use std::io;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
+use std::{env, io};
 
 use anyhow::{anyhow, bail, Context, Result};
 use serde::Deserialize;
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -1000,6 +993,15 @@ fn get_default_host_triple(s: &Settings, process: &Process) -> dist::TargetTripl
         .unwrap_or_else(|| dist::TargetTriple::from_host_or_build(process))
 }
 
+fn non_empty_var(name: &str, process: &Process) -> anyhow::Result<Option<String>> {
```
rami3l: Can we make this `Process::non_empty_var()` instead?

djc: I guess we could? Given that there are only two callers and both of them are in `config`, I'd rather keep it private to the module (unless you think there are other callers that could benefit).

rami3l: @djc Yup, keeping the `impl Process` block containing that single method local to this module would be perfect! We can move it out in another commit if it finds other usages later.

src/config.rs
```diff
@@ -288,30 +287,24 @@ impl<'a> Cfg<'a> {
...
+            None => {
                 // For backward compatibility
-                process
-                    .var("RUSTUP_DIST_ROOT")
-                    .ok()
-                    .and_then(utils::if_not_empty)
+                non_empty_var("RUSTUP_DIST_ROOT", process)?
                     .inspect(|url| trace!("`RUSTUP_DIST_ROOT` has been set to `{url}`"))
-                    .map_or(Cow::Borrowed(dist::DEFAULT_DIST_ROOT), Cow::Owned)
                     .as_ref()
```
rami3l, line 305: This line (`.as_ref()`) is probably not needed any more. It's a ghost from 8 years ago 😓

djc: Nope, we still need it. Otherwise the `map()` takes ownership of the value that was previously contained in the `Option`, but it only yields a reference to that value.

src/config.rs
djc: Hmm, not really a fan of keeping `impl` blocks (with the same bounds) in separate files... What's wrong with the free function setup?

rami3l (approved):

src/config.rs
rami3l: @djc It's just that a free `non_empty_var` function makes me feel less sure about what the function is doing without having to look at the implementation, whereas a method call makes it clear that it's a variation of `Process::var()`. Not a hard blocker really.

djc: Would renaming it to `non_empty_env_var()` (or some other name) help with that?

rami3l: > Would renaming it to `non_empty_env_var()` (or some other name) help with that?

@djc Yes, I think that would be fine as well. Thanks!

djc: Done!

djc:
Hope this doesn't disrupt your PR too much!

Issue #3708, opened 2024-03-12
Rustup 1.27.0 incompatible with older macOS


andersm:
### Problem

Trying to run `rustup self update` on macOS 10.14 fails with the error `dyld: cannot load 'rustup-init' (load command 0x80000034 is unknown)`.

Full output:
```
$ rustup self update
info: checking for self-update
info: downloading self-update
  rustup updated - (unknown) (from 1.26.0)

dyld: cannot load 'rustup-init' (load command 0x80000034 is unknown)
error: self-updated failed to replace rustup executable

$ uname -a
Darwin Caligari.local 18.7.0 Darwin Kernel Version 18.7.0: Tue Jun 22 19:37:08 PDT 2021; root:xnu-4903.278.70~1/RELEASE_X86_64 x86_64
```

### Steps

1. Run `rustup self update`
2. There is no step 2.

### Rustup version

```console
rustup 1.26.0 (5af9b9484 2023-04-05)
```

### Installed toolchains

```console
Default host: x86_64-apple-darwin
rustup home:  /Users/anders/.rustup

stable-x86_64-apple-darwin (default)
rustc 1.76.0 (07dca489a 2024-02-04)
```

ehuss:
Thanks for the report!

I believe rustup needs to be built with the environment variable `MACOSX_DEPLOYMENT_TARGET: 10.12`. See https://github.com/rust-lang/rust/issues/104570 for the details. Newer versions of Xcode technically don't support such an old version of macos, but I believe in this case it shouldn't be necessary to downgrade.

djc:
@ehuss are you suggesting we/the project should build/release a different binary release for macOS, or that the OP builds their own?

ehuss:
Yea, rustup itself needs to be built with that set, presumably in https://github.com/rust-lang/rustup/blob/master/ci/actions-templates/macos-builds-template.yaml.

You might also consider this for a point release since it is a tier-1 platform.

rami3l:
~~@andersm Does our [new build](https://github.com/rust-lang/rustup/actions/runs/8258781556/artifacts/1320871566) work for you? (You might need to log into GitHub in order to download.)~~

The build above is outdated, please use https://github.com/rust-lang/rustup/issues/3708#issuecomment-2081263739 instead.

andersm:
Yes, it seems to run fine.
```
$ rustup --version
rustup 1.27.0+7 (2eff946a6 2024-03-13)
[...]

PR #3931, opened 2024-07-09
Prepare deployment on master branch

jdno:
We added steps to the GitHub Actions workflows in #3909 to upload the build artifacts for the `master` branch as well as for the `stable` branch. But the scripts that prepare the `deploy/` directory were not set to run on `master`, causing the builds to fail.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -141,11 +141,11 @@ jobs:
             target/${{ matrix.target }}/release/rustup-init.exe
           retention-days: 7
       - name: Acquire the AWS tooling
-        if: github.event_name == 'push' && github.ref == 'refs/heads/stable' && matrix.mode == 'release'
+        if: github.event_name == 'push' && matrix.mode == 'release' && (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/master')
         run: |
           choco upgrade awscli
       - name: Prepare the dist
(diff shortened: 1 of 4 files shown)


rami3l:
@jdno Just being curious: will this allow some sort of publicly available "thunk" builds? Because that will also be a good move towards better transparency if it does :)

rami3l (approved):

jdno:
Do you mean that users will be able to install builds from the `master` branch? We haven't thought about it yet, but I see no reason why we can't make the S3 bucket publicly available so that users can do something like `RUSTUP_UPDATE_ROOT=https://builds.rustup.rs/${commit} rustup self update` (or another domain)... 🤔

rami3l:
> Do you mean that users will be able to install builds from the `master` branch? We haven't thought about it yet, but I see no reason why we can't make the S3 bucket publicly available so that users can do something like `RUSTUP_UPDATE_ROOT=https://builds.rustup.rs/${commit} rustup self update` (or another domain)... 🤔

~~@jdno Yup! I had to provide CI uploads to users in order to know if some recent fix has worked for them or not (e.g. https://github.com/rust-lang/rustup/issues/3708#issuecomment-1993590209). This would be a nice addition if possible, and it seems that there shouldn't be introducing too much overhead for you?~~

Moved to #3932.

rami3l:
I mean we can discuss this in a new thread.

PR #3933, opened 2024-07-09
Fix uploading of Windows build artifacts

jdno:
On Windows, the `dist` folder is uploaded at the end of the build. This wasn't taken into consideration when introducing the new S3 bucket in


jdno:
This should be the last one. The other artifacts got uploaded successfully to S3. 🥳

rami3l (approved):

PR #3934, opened 2024-07-10
Upload Windows artifacts into correct subdirectory

jdno:
The Windows artifacts are uploaded to S3 as part of every build on `master`, but due to slight differences in the upload command syntax between Linux and Windows currently end up at the root of the upload directory. Similar to the existing upload step for builds on `stable`, the command has been tweaked to place the files into the `dist` directory.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -166,7 +166,7 @@ jobs:
       - name: Deploy build to rustup-builds bucket for release team
         if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
         run: |
-          aws --debug s3 cp --recursive dist s3://rustup-builds/${{ github.sha }}
+          aws --debug s3 cp --recursive dist s3://rustup-builds/${{ github.sha }}/dist
         env:
           AWS_DEFAULT_REGION: us-east-1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -166,7 +166,7 @@ jobs:
       - name: Deploy build to rustup-builds bucket for release team
         if: github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.mode == 'release'
         run: |
-          aws --debug s3 cp --recursive dist s3://rustup-builds/${{ github.sha }}
+          aws --debug s3 cp --recursive dist s3://rustup-builds/${{ github.sha }}/dist
```
djc, line 169: This file is generated (see notice near the top)... should probably change the source?

jdno: I did. 😉 I added to the change to `windows-builds-template.yaml` and then ran the script to apply it to `ci.yaml`. Should be all good.

djc (approved):

.github/workflows/ci.yaml
djc: Ahh, sorry for the noise!

rami3l (approved):
Nice!

Issue #3936, opened 2024-07-11
Windows: "called `Option::unwrap()` on a `None` value"


Baiyuetribe:
### Verification

- [X] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue+is%3Aopen%2Cclosed and found no duplicates.
- [X] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

windows on Arm64 still not work.
```log
thread 'main' panicked at src\cli\self_update.rs:652:45:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Steps

reproduce
```sh
# my device: windows arm64
git clone https://github.com/rust-lang/rustup.git
cargo build --release
```
after,i got `rustup-init.exe`.
![image](https://github.com/rust-lang/rustup/assets/43716063/19f78eb0-3673-463d-b991-e32464dbdb4d)

### Rustup version

```console
latest
```

### Installed toolchains

```console
none
```

### OS version

```console
windows arm64
```

```[tasklist]
### Tasks
- [ ] https://github.com/rust-lang/rustup/pull/3938
```

rami3l:
@Baiyuetribe Sorry for your bad experience!

Do you have the `HOME` environment variable set (https://superuser.com/a/1617017)? Rustup depends on that to work.

rami3l:
Also, now that we're at it, maybe you can send a PR over **if that is indeed the root cause**? By adding a Windows-exclusive fallback to `USERPROFILE` for example.

The relevant code is here...
https://github.com/rust-lang/rustup/blob/79ba44c2814fa240cc64d1f4b1b1c324d75709ad/src/currentprocess.rs#L133

... and probably add a `.context()` call here:
https://github.com/rust-lang/rustup/blob/79ba44c2814fa240cc64d1f4b1b1c324d75709ad/src/currentprocess.rs#L51

... similar to what has been done below:
https://github.com/rust-lang/rustup/blob/79ba44c2814fa240cc64d1f4b1b1c324d75709ad/src/currentprocess.rs#L55

Many thanks in advance 🙇!

PS: Windows ARM64 is still a tier 2 target by far, meaning we still don't have the means to test it with our CI, we currently can only know that it builds correctly for Windows ARM64 (which it does).

ChrisDenton:
[...]

PR #3938, opened 2024-07-11
Fix home_dir() and current_dir() regression

djc:
In #3764 (specifically https://github.com/rust-lang/rustup/commit/204c8a9ca4e8a516865e381262a16b9eacb53901), I moved a bunch of code around and in the process mixed up the `OS` and `Test` cases. Fix my mistake.

Thanks @ChrisDenton for paying attention and bisecting!

As discussed in #3936.

### DIFF
--- a/doc/dev-guide/src/coding-standards.md
+++ b/doc/dev-guide/src/coding-standards.md
@@ -20,9 +20,9 @@ order. Any file that is not grouped like this can be rearranged whenever the
 file is touched - we're not precious about having it done in a separate commit,
 though that is helpful.
 
-## No direct use of process state outside rustup::currentprocess
+## No direct use of process state outside rustup::process
 
-The `rustup::currentprocess` module abstracts the global state that is
+The `rustup::process` module abstracts the global state that is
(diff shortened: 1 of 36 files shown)


ChrisDenton (approved):
Thanks djc!

rami3l (approved):
Thanks a lot! However the implementation of `var_os()` below is still a bit verbose... Can't we just call `Process::var_os()` directly without the `match`?

djc:
> Thanks a lot! However the implementation of `var_os()` below is still a bit verbose... Can't we just call `Process::var_os()` directly without the `match`?

Good call! Also tacked on two more commits with some quick renaming.

rami3l (approved):
LGTM modulo some `rustc`/`clippy` errors. Thanks again!

(fcb31a1bc40b6c636aa95ad0978477ae074ad3e6 has passed my manual tests via UTM WoA on my Mac.)

djc:
Thanks for cleaning this up for me, wasn't able to follow up yesterday!

PR #3939, opened 2024-07-11
Add help message for missing toolchain

### LINKED ISSUE #3573: improve error message for missing toolchain

Running `cargo +nightly fmt` without nightly installed produces

```
error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installed
```

but doesn't suggest any ways to fix the problem. This can be confusing for newcomers.

### Solution you'd like

It would be great if it suggested to run `rustup install nightly`.

### Notes

_No response_

jtr860830:
Closes #3573.

This patch adds a suggestion to run `rustup install <toolchain>` for missing toolchain.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -86,7 +86,14 @@ pub enum RustupError {
     RunningCommand { name: OsString },
     #[error("toolchain '{0}' is not installable")]
     ToolchainNotInstallable(String),
-    #[error("toolchain '{0}' is not installed")]
+    #[error(
+        "toolchain '{0}' is not installed{}",
+        if let ToolchainName::Official(t) = .0 {
+            format!("\nhelp: run `rustup toolchain install {t}` to install it")
+        } else {
+            String::new()
+        },
+    )]
     ToolchainNotInstalled(ToolchainName),
     #[error("path '{0}' not found")]
     PathToolchainNotInstalled(PathBasedToolchainName),
--- a/tests/suite/cli_exact.rs
+++ b/tests/suite/cli_exact.rs
@@ -691,6 +691,23 @@ info: installing component 'rust-std' for '{0}'
         .await;
 }
 
+// issue #3573
+#[tokio::test]
+async fn show_suggestion_for_missing_toolchain() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+    cx.config
+        .expect_err_ex(
+            &["cargo", "+nightly", "fmt"],
+            r"",
+            for_host!(
+                r"error: toolchain 'nightly-{0}' is not installed
+help: run `rustup toolchain install nightly-{0}` to install it
+"
+            ),
+        )
+        .await;
+}
+
 // issue #927
 #[tokio::test]
 async fn undefined_linked_toolchain() {


src/errors.rs
```diff
@@ -86,7 +86,7 @@ pub enum RustupError {
     RunningCommand { name: OsString },
     #[error("toolchain '{0}' is not installable")]
     ToolchainNotInstallable(String),
-    #[error("toolchain '{0}' is not installed")]
+    #[error("toolchain '{0}' is not installed.\nhelp: run `rustup install {0}` to install it.")]
```
rami3l:
```suggestion
    #[error("toolchain '{0}' is not installed.\nhelp: run `rustup toolchain install {0}` to install it.")]
```

rami3l (commented):
Thanks a lot for helping out! It'll be nice if you adjust our integration test accordingly as well :)

jtr860830:
> Thanks a lot for helping out! It'll be nice if you adjust our integration test accordingly as well :)

OK, I'll adjust it. Thx for replying!

rami3l:
> > Thanks a lot for helping out! It'll be nice if you adjust our integration test accordingly as well :)
> 
> OK, I'll adjust it. Thx for replying!

@jtr860830 No worries! Just remember to squash everything into a single commit since it's a single unit of change: we use [atomic commits](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits) across the repo.

The integration test you need to fix is `suite::cli_exact::undefined_linked_toolchain`, as indicated by the CI log.

rami3l (dismissed):
LGTM, and thanks again for helping out! :)

rami3l:
@jtr860830 Sorry, but I've changed my mind regarding this change, since it doesn't make the distinction between a typo (like `bogus`) and an actual toolchain that can be installed with `rustup toolchain install` (like `nightly-x86_64-unknown-linux-gnu`). Can we print out the second line only in the second case?

Hint: Use `PartialToolchainDesc::from_str()`.

jtr860830:
> @jtr860830 Sorry, but I've changed my mind regarding this change, since it doesn't make the distinction between a typo (like `bogus`) and an actual toolchain that can be installed with `rustup toolchain install` (like `nightly-x86_64-unknown-linux-gnu`). Can we print out the second line only in the second case?
> 
> Hint: Use `PartialToolchainDesc::from_str()`.

I get it. I'll improve it!

src/errors.rs
```diff
@@ -86,7 +86,10 @@ pub enum RustupError {
     RunningCommand { name: OsString },
     #[error("toolchain '{0}' is not installable")]
     ToolchainNotInstallable(String),
-    #[error("toolchain '{0}' is not installed")]
+    #[error("toolchain '{0}' is not installed.{}", match .0.to_string().parse::<PartialToolchainDesc>() {
```
rami3l: 1. `.to_string()` doesn't seem required here.
2. I prefer using `.map_or_else(String::new, ...)` over `match`.

src/errors.rs
```diff
@@ -86,7 +86,10 @@ pub enum RustupError {
...
+        Ok(desc) => format!("\nhelp: run `rustup toolchain install {}` to install it.", .0),
```
rami3l: I don't think you should be adding the `.` at the end of each line, that's not coherent with the rest of the codebase.

tests/suite/cli_exact.rs
```diff
@@ -699,7 +699,7 @@ async fn undefined_linked_toolchain() {
         .expect_err_ex(
             &["cargo", "+bogus", "test"],
             r"",
-            "error: toolchain 'bogus' is not installed\n",
+            "error: toolchain 'bogus' is not installed.\n",
```
rami3l: Now consider adding a new case which shows the second line, for example the one in the original issue.

src/errors.rs
jtr860830: I think `.to_string()` is required because it needs to cast `ToolchainName` to a `String`, or did I misunderstand something?

rami3l: > I think `.to_string()` is required because it needs to cast `ToolchainName` to a `String`, or did I misunderstand something?

@jtr860830 No, that's an oversight on my side, sorry. I didn't notice that we're passing a `ToolchainName`. In that case `.parse()` does work, but I guess there's a better way... I'll handle this.

rami3l (approved):
Sorry for the delay, and thanks again for your help!

jtr860830:
> Sorry for the delay, and thanks again for your help!

This PR has taught me a lot. Thx for your help!

PR #3941, opened 2024-07-12
refactor: clean up platform-specific logic

### LINKED ISSUE #2424: Finish disentangling Windows and Unix logic

### LINKED ISSUE #3936: Windows: "called `Option::unwrap()` on a `None` value"

rami3l:
Closes #2424, closes #3936.

> Looks like `do_pre_install_sanity_checks()` dates back from prehistoric times (v0.1.9, [0080271](https://github.com/rust-lang/rustup/commit/008027172d1c7566467cc653556b3b8c232bc147)), to defend against the old Rust distribution tools (namely "multirust.sh, rustc tarballs, and rustup.sh").
> 
> Now let's look at the files it verifies:
> 
> * `manifest-rustc` (without any prefix) is almost certainly gone: I cannot see this on my machine.
> * `uninstall.sh` is... gone as well? Not quite sure, but a solution like this will definitely not work with Windows.
> * `~/.rustup/rustup-version` is not for the Rust version of rustup, but rather its prototype `rustup.sh` (so there's no point in changing this to `rustup_home()`!), implemented in shell script, so is Unix only.
> 
> ... so this function serves no purpose, at least not on Windows. My suggestion for now is to revert [0080271](https://github.com/rust-lang/rustup/commit/008027172d1c7566467cc653556b3b8c232bc147) completely.
_https://github.com/rust-lang/rustup/issues/3936#issuecomment-2222363109_

### Concerns

- [x] This PR didn't remove all usages of `cfg!()` in the test code for the sake of convenience. (Otherwise a test writer might have to deal with multiple compilation errors in other platforms.)
- [ ] Should we replace `not(windows)` with `unix` in `cfg`s?


djc (approved):

djc:
Nice!

PR #3942, opened 2024-07-12
style: reformat all TOML files with `taplo`

rami3l:
As [discussed and approved](https://discord.com/channels/442252698964721669/463480252723888159/1261135626879766619) by @rbtcollins.

## Prior Art
- https://github.com/rust-lang/team/pull/1338

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1349,7 +1349,7 @@ jobs:
       - name: Run CI workflow generation checks
         run: |
           ./ci/actions-templates/gen-workflows.sh
-          git diff --quiet
+          git diff --exit-code
       - name: Prep cargo dirs
         run: |
           mkdir -p ~/.cargo/{registry,git}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1392,7 +1392,15 @@ jobs:
           shellcheck -x -s dash -- rustup-init.sh
           git ls-files -- '*.sh' | xargs shellcheck -x -s dash
           git ls-files -- '*.bash' | xargs shellcheck -x -s bash
-      - name: Run formatting checks
+      - name: Install taplo
+        uses: taiki-e/install-action@v2
+        with:
+          tool: taplo-cli@0.8
+      - name: Run TOML formatting checks
+        run: |
+          taplo fmt
+          git diff --exit-code
+      - name: Run Rust formatting checks
         run: |
           cargo fmt --all --check
       - name: Run cargo check and clippy
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,10 @@ build = "build.rs"
 [features]
 curl-backend = ["download/curl-backend"]
 default = [
-    "curl-backend",
-    "reqwest-backend",
-    "reqwest-native-tls",
-    "reqwest-rustls-tls",
+  "curl-backend",
+  "reqwest-backend",
+  "reqwest-native-tls",
+  "reqwest-rustls-tls",
 ]
 
 reqwest-backend = ["download/reqwest-backend"]
@@ -30,10 +30,10 @@ no-self-update = []
 
 # Include an Opentelemetry sink for tracing events
 otel = [
-    "dep:opentelemetry-otlp",
-    "dep:tracing-opentelemetry",
-    "dep:opentelemetry",
-    "dep:opentelemetry_sdk",
+  "dep:opentelemetry-otlp",
+  "dep:tracing-opentelemetry",
+  "dep:opentelemetry",
+  "dep:opentelemetry_sdk",
 ]
 
 # Exports code dependent on private interfaces for the integration test suite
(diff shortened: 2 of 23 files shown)


.github/workflows/ci.yaml
```diff
@@ -1349,7 +1349,7 @@ jobs:
       - name: Run CI workflow generation checks
         run: |
           ./ci/actions-templates/gen-workflows.sh
-          git diff --quiet
+          git diff
```
rami3l: This is to make it clearer that `git diff` actually checks for coherence, and will actually print out the inconsistencies if there are any.

Cargo.toml
```diff
@@ -143,7 +143,10 @@ proptest = "1.1.0"
 tempfile = "3.8"
 termcolor = "1.2"
 thiserror = "1.0"
-tokio = { version = "1.26.0", default-features = false, features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.26.0", default-features = false, features = [
```
djc: So one reason I've bounced off of this in the past is that IMO this is not an improvement. I think one line per dependency/features is a pretty strong default and having just a few dependencies that spread their features out over multiple lines does not make the `Cargo.toml` easier to read.

rami3l: > So one reason I've bounced off of this in the past is that IMO this is not an improvement. I think one line per dependency/features is a pretty strong default and having just a few dependencies that spread their features out over multiple lines does not make the `Cargo.toml` easier to read.

@djc Can we make max line lengths extra large, like 150?

rami3l: @djc In other words I don't care that much about the style per se, but rather that the formatting can be automated...

.github/workflows/ci.yaml
smoelius:
```suggestion
          git diff --exit-code
```
Sorry for the intrusion. `--quiet` [implies `--exit code`](https://git-scm.com/docs/git-diff). Plain `git diff` will exit with 0 even if there are differences.

rami3l: @smoelius Thanks a lot!

djc: @smoelius no intrusion at all, always happy to get your feedback!

Cargo.toml
djc: > > So one reason I've bounced off of this in the past is that IMO this is not an improvement. I think one line per dependency/features is a pretty strong default and having just a few dependencies that spread their features out over multiple lines does not make the `Cargo.toml` easier to read.
> 
> @djc Can we make max line lengths extra large, like 150?

Let's try it?

rami3l: @djc [Special rule](https://taplo.tamasfe.dev/configuration/file.html#rules) applied: `column_width = 160` for `dependencies` exclusively.

rami3l: @djc Sorry, I've forgot `workspace.`, `dev-` and `build-dependencies`, just fixed.

.github/workflows/ci.yaml
```diff
@@ -1392,7 +1392,15 @@ jobs:
           shellcheck -x -s dash -- rustup-init.sh
           git ls-files -- '*.sh' | xargs shellcheck -x -s dash
           git ls-files -- '*.bash' | xargs shellcheck -x -s bash
-      - name: Run formatting checks
+      - name: Install taplo
+        uses: taiki-e/install-action@v2
+        with:
+          tool: taplo-cli@0.8
```
rami3l, line 1398: The latest version is 0.9, but for now both Neovim/mason and VSCode are on 0.8, so I'll stick with that until later.

rami3l: Neovim/mason now has `taplo` 0.9 (https://github.com/mason-org/mason-registry/pull/6469), but I guess the PR will make more sense once the VSCode plugin has been updated to 0.9.

This might take a while since they'll have to wait for the maintainer to rotate the access token to the marketplace (https://github.com/tamasfe/taplo/issues/624)... ~~Marking as draft until this problem is resolved.~~

djc (approved):
Nice!

Suggest putting the reformatting commit before the one that adds the CI check.

rami3l:
> Suggest putting the reformatting commit before the one that adds the CI check.

@djc I've intentionally put the reformatting commit at the end so I can amend to it conveniently, but fair enough.

djc:
> > Suggest putting the reformatting commit before the one that adds the CI check.
> 
> @djc I've intentionally put the reformatting commit at the end so I can amend to it conveniently, but fair enough.

Makes sense -- can always reorder before the merge?

rami3l:
> Makes sense -- can always reorder before the merge?

@djc Yes, but that's a draft, so... Anyway, we are on the same page now.

rami3l:
@djc As there are a bit too many regressions/breakages in `taplo` v0.9, I've made the upgrade a separate follow-up issue in https://github.com/rust-lang/rustup/issues/3945. So far we are still using v0.8 for our CI.

PR #3952, opened 2024-07-16
fix(rustup-mode): return `ExitCode(1)` when `update()` fails

### LINKED ISSUE #3476: When `rustup update` fails, it doesn't return an exit code that indicate failure
### Problem

When `rustup update` fails, it doesn't return an exit code that indicate failure but instead returns 0.

### Steps

1. Be in a situation where `rustup update` fails, for example by disabling your network connection
2. Run `rustup update`
3. The command returns the exit code 0, or success

### Possible Solution(s)

Return an exit code different than 0 in case of failure

### Notes

_No response_

### Rustup version

```console
rustup 1.26.0 (2023-05-04)
info: The currently active `rustc` version is `rustc 1.72.0 (5680fa18f 2023-08-23)`
```

### Installed toolchains

```console
Not relevant
```

rami3l:
Closes #3476, verified via local testing.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -867,7 +867,7 @@ async fn update(cfg: &mut Cfg<'_>, opts: UpdateOpts) -> Result<utils::ExitCode> 
             common::self_update(|| Ok(()), cfg.process).await?;
         }
     } else {
-        exit_code = common::update_all_channels(cfg, self_update, opts.force).await?;
+        exit_code &= common::update_all_channels(cfg, self_update, opts.force).await?;
         info!("cleaning up downloads & tmp directories");
         utils::delete_dir_contents_following_links(&cfg.download_dir);
         cfg.tmp_cx.clean();
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -862,7 +862,7 @@ async fn update(cfg: &mut Cfg<'_>, opts: UpdateOpts) -> Result<utils::ExitCode> 
             }
         }
         if self_update {
-            common::self_update(|| Ok(utils::ExitCode(0)), cfg.process).await?;
+            common::self_update(|| Ok(()), cfg.process).await?;
         }
     } else {
         common::update_all_channels(cfg, self_update, opts.force).await?;
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -789,6 +789,8 @@ async fn check_updates(cfg: &Cfg<'_>) -> Result<utils::ExitCode> {
 }
 
 async fn update(cfg: &mut Cfg<'_>, opts: UpdateOpts) -> Result<utils::ExitCode> {
+    let mut exit_code = utils::ExitCode(0);
+
     common::warn_if_host_is_emulated(cfg.process);
     let self_update_mode = cfg.get_self_update_mode()?;
     // Priority: no-self-update feature > self_update_mode > no-self-update args.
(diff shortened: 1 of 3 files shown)


src/cli/rustup_mode.rs
```diff
@@ -862,10 +864,10 @@ async fn update(cfg: &mut Cfg<'_>, opts: UpdateOpts) -> Result<utils::ExitCode>
             }
         }
         if self_update {
-            common::self_update(|| Ok(utils::ExitCode(0)), cfg.process).await?;
+            common::self_update(|| Ok(()), cfg.process).await?;
         }
     } else {
-        common::update_all_channels(cfg, self_update, opts.force).await?;
+        exit_code = exit_code.and(common::update_all_channels(cfg, self_update, opts.force).await?);
```
rami3l: @djc Since `ExitCode::and` essentially simulates `&&` in bash, how about overriding `ops::BitAnd*`, and this will become a pretty standard `&=`?

src/utils/utils.rs
```diff
@@ -24,6 +24,16 @@ pub use crate::utils::utils::raw::{is_file, path_exists};
 #[derive(Debug, PartialEq, Eq)]
 pub struct ExitCode(pub i32);
 
+impl ExitCode {
+    pub fn and(self, other: Self) -> Self {
+        if self == Self(0) {
+            other
+        } else {
+            self
+        }
```
djc: Maybe rewrite this as

```rust
match self.0 == 0 {
    true => other,
    false => self,
}
```

djc (approved):
This seems okay. Personally I feel like use of the custom `ExitCode` is a bad abstraction. One idea I've played with is that we should change types to be `Result<T, Option<E>>`, where the `Some` represents an error that should be yielded by the top-level caller while `None` means the error has already been reported. (We then report `ExitCode::FAILURE` if the `Result` is `Err`.)

rami3l:
> This seems okay. Personally I feel like use of the custom `ExitCode` is a bad abstraction. One idea I've played with is that we should change types to be `Result<T, Option<E>>`, where the `Some` represents an error that should be yielded by the top-level caller while `None` means the error has already been reported. (We then report `ExitCode::FAILURE` if the `Result` is `Err`.)

@djc Indeed, in other projects that I've worked with my principle is that we should never expose the exit code with the single exception of the main module, where some `thiserror` is converted to `ExitStatus` via an `impl Termination`.

But that kind of change is clearly out of the scope of this PR...

src/utils/utils.rs
rami3l: > Maybe rewrite this as
> 
> ```rust
> match self.0 == 0 {
>     true => other,
>     false => self,
> }
> ```

@djc I don't like `match bool` very much personally, but there's a middle ground I guess we can all accept:

```rust
match self.0 {
  0 => other,
  _ => self,
}
```

djc: Ah yeah, that is even better!

src/cli/rustup_mode.rs
djc: Seems okay.

src/utils/utils.rs
```diff
@@ -24,6 +25,25 @@ pub use crate::utils::utils::raw::{is_file, path_exists};
 #[derive(Debug, PartialEq, Eq)]
 pub struct ExitCode(pub i32);
 
+impl BitAnd for ExitCode {
+    type Output = Self;
+
+    fn bitand(self, rhs: Self) -> Self::Output {
+        match self.0 {
```
djc, line 33: Maybe add a short comment explaining the conceptual logic?

```
// If `self` is `0` (success), yield `other`
```

(Also for the other impl.)

djc (approved):

src/utils/utils.rs
rami3l: @djc Done!

PR #3954, opened 2024-07-17
fix(utils): make `ExitCode` `#[must_use]`

rami3l:
This PR is a continuation of #3952 to "properly" address similar linting-related issues regarding `ExitCode`.

I should've done this yesterday, but it didn't occur to me till this morning... Still it's better late than never.


djc (approved):
Great idea!

PR #3955, opened 2024-07-17
build(windows): fix typo in `build.rs`

rami3l:
Cherry-picked from #3898.

The first commit is deliberately failing to demonstrate that the fix is actually working.

See https://github.com/aws/aws-lc-rs/issues/453#issuecomment-2199337475 for the whole context:

> Another crucial difference between our setups is that we use the following linker option on Windows MSVC as well:
> 
> ```rs
> println!("cargo:rustc-link-arg-bin=rustup-init=/WX");
> ```
> 
> This was originally added by @ChrisDenton to prevent `cargo` from silencing linkage warnings, which is probably why your CI happens to turn green.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -66,5 +66,5 @@ fn main() {
     // Rust hides linker warnings meaning mistakes may go unnoticed.
     // Turning them into errors forces them to be displayed (and the build to fail).
     // If we do want to ignore specific warnings then `/IGNORE:` should be used.
-    println!("cargo:cargo:rustc-link-arg-bin=rustup-init=/WX");
+    println!("cargo:rustc-link-arg-bin=rustup-init=/WX");
 }
--- a/build.rs
+++ b/build.rs
@@ -53,7 +53,7 @@ fn main() {
     // This will work on all supported Windows versions but it relies on
     // us using `SetDefaultDllDirectories` before any libraries are loaded.
     // See also: src/bin/rustup-init.rs
-    let delay_load_dlls = ["bcrypt", "powrprof", "secur32"];
+    let delay_load_dlls = ["bcrypt", "secur32"];
     for dll in delay_load_dlls {
         println!("cargo:rustc-link-arg-bin=rustup-init=/delayload:{dll}.dll");
     }


djc:
It doesn't look like we want to merge this as is? The commits here are a little strange to me.

rami3l:
> It doesn't look like we want to merge this as is? The commits here are a little strange to me.

@djc There is a typo in the build script that actually invalidated the check that @ChrisDenton wanted to add. If we're linking against `aws-lc` then this will happen more often...

ChrisDenton:
Hm, I don't think that's a typo? It's just that `powrprof.dll` is no longer being used when switching away from ring? Though maybe I'm misunderstanding the context.

build.rs
```diff
@@ -66,5 +66,5 @@ fn main() {
     // Rust hides linker warnings meaning mistakes may go unnoticed.
     // Turning them into errors forces them to be displayed (and the build to fail).
     // If we do want to ignore specific warnings then `/IGNORE:` should be used.
-    println!("cargo:rustc-link-arg-bin=rustup-init=/WX");
+    println!("cargo:cargo:rustc-link-arg-bin=rustup-init=/WX");
```
djc: But this seems to bring back the typo?

rami3l: @djc Nice catch! That's a Vim buffer incoherence problem on my side 🤦 Will fix.

ChrisDenton: Wait, where's that from? I feel like I'm seeing a different changes to you?

rami3l:
> Hm, I don't think that's a typo? It's just that `powrprof.dll` is no longer being used when switching away from ring? Though maybe I'm misunderstanding the context.

@ChrisDenton You accidentally wrote `cargo:cargo:rustc-link-arg-bin=rustup-init=/WX` with two `cargo`s, so the check didn't trigger.

ChrisDenton:
Ah thanks! I'm confused because when I clicked the "Files Changed" tab I don't see that change. Edit: so I was looking at completely the wrong thing, apparently 😅

build.rs
rami3l: @djc Fixed!

djc (approved):
These code changes look good to me, though I'll happily defer to @ChrisDenton on whether these changes are correct (in particular whether we still need delay loading for `powrprof`).

ChrisDenton (approved):
We don't seem to be using powrprof for anything and haven't been for at least a long while (I checked some old rustup-init versions). So I'm good with this.

PR #3959, opened 2024-07-22
Upgrade to opentelemetry 0.24

djc:


rami3l:
@djc #3961 is immortal. Only lockfile updates and grouped updates are expected to be immortal, so it's working as expected. BTW I've also noticed you cannot directly amend onto PRs from some point on; you'll have to create a superseding PR by yourself, but you can still cherry-pick stuff from there.

src/cli/log.rs
```diff
@@ -152,25 +152,28 @@ where
...
-            trace::config()
+            Config::default()
                 .with_sampler(Sampler::AlwaysOn)
                 .with_resource(Resource::new(vec![KeyValue::new("service.name", "rustup")])),
         )
         .install_batch(opentelemetry_sdk::runtime::Tokio)
-        .expect("error installing `OtlpTracePipeline` in the current `tokio` runtime")
+        .expect("error installing `OtlpTracePipeline` in the current `tokio` runtime");
+
+    global::set_tracer_provider(provider.clone());
```
rami3l, line 177: Thanks! However may I ask what this line is doing?

rami3l (approved):

src/cli/log.rs
djc: This was previously part of the code that was there before, but upstream no longer implicitly does this:

https://github.com/open-telemetry/opentelemetry-rust/issues/1592
https://github.com/open-telemetry/opentelemetry-rust/issues/1810

Issue #1236, opened 2017-08-14
Homebrew conflicts with rustup (detect and warn about it?)


kornelski:
On macOS if you install anything with Homebrew that has Rust as a dependency (including any program that has to be built from Rust source), Homebrew will install Rust system-wide and it will take precedence over rustup.

I've ran into this — it looked like if `rustup default` command didn't work properly. I took me a while to realize that binaries in `/usr/local/bin` are not from rustup.

Homebrew dislikes using any binaries and update mechanisms they don't control ([including rustup](https://github.com/Homebrew/homebrew-core/pull/9617)), so I think it's up to Rustup to improve this situation.

As a solution I suggest emitting a warning when `rustup default` command is used if the `PATH` contains *both* Rustup rustc and non-Rustup rustc, and non-Rustup rustc is first. That will catch this error, but won't get in the way of using Rustup without the `PATH` modified for Rustup.

skull-squadron:
A quick fix is `brew uninstall rust` or `brew unlink rust`.
rustup and rustup-init formulas are mutually-exclusive, and should mutually `conflicts` or whatever with each other. The latter seems to be a brew package of rustup.
Files issues about Homebrew there, because rustup properly detects an existing rust and exits.

kornelski:
To clarify, the situation I'm talking about has nothing to do with conflicts of Homebrew's packages. 

Consider this scenario:

1. Install rustup and use rustc from rustup (there's no Homebrew rustc, no conflict)
2. Install rustc from Homebrew (Homebrew doesn't care rustup is there, and does not report any problem)
3. Try to use `rustup default` to change rustc (doesn't appear to work)

I would like 3. to warn that it's been affected by another rustc in `PATH`, and it no longer works as usual.

While I mention Homebrew specifically, the problem is not unique to Homebrew. Identical problem can happen when you install another copy of rustc from any package manager.

rami3l:
Mitigated by https://github.com/Homebrew/homebrew-core/pull/177983 if you're using a `brew`-managed `rustup` installation.

Please note that now you need to add `$(brew --prefix rustup)/bin` to PATH manually to reach to the `rustup`-installed toolchain. Additionally, you might run `brew unlink rust` to make sure you won't accidentally trigger the `brew`-managed `rust` toolchain. 

I'll adjust the user guide shortly.

(Sorry, that was a misclick below...)

kornelski:
[...]

PR #3963, opened 2024-07-23
docs(user-guide): use `brew install rustup` instead of `rustup-init`

rami3l:
[Homebrew/homebrew-core#177983](https://github.com/Homebrew/homebrew-core/pull/177983) is using `brew install rustup` now, so I'm adjusting the docs accordingly:

> Please note that now you need to add `$(brew --prefix rustup)/bin` to PATH manually to reach to the `rustup`-installed toolchain. Additionally, you might run `brew unlink rust` to make sure you won't accidentally trigger the `brew`-managed `rust` toolchain.
_https://github.com/rust-lang/rustup/issues/1236#issuecomment-2245115828_

The original paragraph specific to `brew`-installed `rustup-init` has been moved and extended to a more general how-to description of the two methods being used by different package managers (thus not limited to `apt` and `brew`), so our users will hopefully have even fewer chances of getting lost.

### DIFF
--- a/doc/user-guide/src/installation/other.md
+++ b/doc/user-guide/src/installation/other.md
@@ -38,6 +38,39 @@ $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --profile
 > If you have encountered any problems installing `rustup` with a package manager,
 > please contact the package maintainer(s) for further information.
 
+### General tips
+
+Different package managers take slightly different approaches towards managing rustup.
+After installing rustup with your favorite package manager, there are usually two possibilities:
+
+- If your package manager has made the `rustup` command available
+  together with proxies for Rust tools such as `rustc` and `cargo`,
+  picking a default toolchain (e.g. `stable`) would usually be enough:
+
+  ```sh
+  $ rustup default stable
+  ```
+
+  > As of 2024/07/24, this is the case for
+  > [DNF](https://developer.fedoraproject.org/tech/languages/rust/further-reading.html).
+
+- If your package manager has only made the `rustup-init` command available, simply run:
+
+  ```sh
+  $ rustup-init
+  ```
+
+  This will allow you to perform the initial setup of `rustup`, populate all the proxies
+  managed by rustup, and install a default toolchain. When the installation is completed,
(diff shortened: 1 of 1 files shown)


doc/user-guide/src/installation/other.md
```diff
@@ -38,6 +38,21 @@ $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --profile
...
+- If the binary is named `rustup`, picking a default toolchain would usually be enough.
+  A popular choice is `stable`:
+
+  ```sh
+  rustup default stable
+  ```
+
+- If the binary is named `rustup-init`, simply execute `rustup-init` and follow the instructions.
+  When the installation is complete, make sure that `$HOME/.cargo/bin` is in your `$PATH`,
+  and you should be able to use `rustup` normally.
```
djc: I don't find this section to be super clear. What is this trying to express/convey?

rami3l: @djc Various package managers take slightly different approaches towards installing the rustup package:
- Some work like [`dnf`](https://developer.fedoraproject.org/tech/languages/rust/further-reading.html) where you're simply getting `rustup-init` and you'll have to run it manually to get `rustup` and other shims.
- Others work like `apt` or [`pacman`](https://wiki.archlinux.org/title/Rust#Arch_Linux_package) where you get a `rustup` binary directly together with the shims populated by their own automations, but you still have to pick a default toolchain to use them.

`brew`, in particular, has just moved from the first to the second, but I think it's still helpful for the user to at least have some clue about what they should probably do after running `package-manager install rustup`.
It's technically out of the scope of this project, but still I'd like to provide some help for more people now that we are here.

djc: Okay, I think something like this context should go in the docs rather than the more mechanical/opaque description you had before. Thanks for the explanation!

rami3l: > Okay, I think something like this context should go in the docs rather than the more mechanical/opaque description you had before. Thanks for the explanation!

@djc I agree. I'll try my best to adjust the wording and include more context.

rami3l: @djc Done! How do you like my new wording?

djc (approved):
Looks great, thanks!

PR #3971, opened 2024-07-30
Remove `once_cell` dependency and use `LazyLock` instead

InfyniteHeap:
Since Rust 1.80.0 was released, which brought `LazyCell` and `LazyLock`, plus the only purpose of introducing `once_cell` was to use `Lazy` for initializations of static variables, it is good to replace `Lazy` with `LazyLock`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -290,6 +290,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
 
 [[package]]
+name = "byteorder"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
+
+[[package]]
 name = "bytes"
 version = "1.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rustup"
-version = "1.27.1"
+version = "1.80.0"
```
rami3l: Hmmm I'm sure this is not what you want to do.

download/Cargo.toml
```diff
@@ -1,26 +1,24 @@
 [package]
 name = "download"
-version = "1.27.1"
+version = "1.80.0"
```
rami3l: This neither.

Cargo.toml
```diff
@@ -1,6 +1,5 @@
 [package]
 name = "rustup"
-version = "1.27.1"
```
rami3l: This is the version of the Rustup package, not the version of the compiler... Same thing for the download crate.

InfyniteHeap: Oh sorry! I got `version` and `rust-version` mixed up!

rami3l:
The CI failure might have something to do with CentOS 7's EOL on 2024/06/30. I've always wanted to use `cross` for our cross-build tests anyway, so might worth a try. I'll probably find some time in the next few weeks to have a look at it if the problem persists...

Cargo.lock
```diff
@@ -1946,9 +1954,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-platform-verifier-android"
-version = "0.1.0"
+version = "0.1.1"
```
rami3l, line 1958: Why are there irrelevant version bumps? I'm afraid this might interfere with what renovatebot is doing... `cargo rm` didn't generate this, right?

InfyniteHeap: `Cargo.lock` is automatically managed by Cargo... And this seems not a bump, because it only modified the patch number, according to semver.

InfyniteHeap:
> The CI failure might have something to do with CentOS 7's EOL on 2024/06/30. I've always wanted to use `cross` for our cross-build tests anyway, so might worth a try. I'll probably find some time in the next few weeks to have a look at it if the problem persists...

Ok, and it seems that I have to continue awaiting until this CI issue is resolved...

rami3l (approved):
@InfyniteHeap Looks like the CI is back online. Just squash your commits into one (read more about atomic commits [here](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits)), and we're good to go.

rami3l:
@InfyniteHeap Nice job!

PR #3975, opened 2024-07-31
fix: fix unreachable code lints on Android

rami3l:
Cherry-picked from #3898.

When compiling for Android, the linter will warn about unreachable branches such as the `(b"Linux", b"i686")` one, since it's already covered by the `(_, b"i686")` case.

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -489,19 +489,8 @@ impl TargetTriple {
                 )
             };
 
+            #[cfg(not(target_os = "android"))]
             let host_triple = match (sysname, machine) {
-                #[cfg(target_os = "android")]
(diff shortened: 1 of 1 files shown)


djc (approved):
Do we not have Android in CI, or do we not deny warnings in that CI?

rami3l:
> Do we not have Android in CI, or do we not deny warnings in that CI?

@djc That's a cross-compilation build-only action, and so far this is the only place where it's Android-specific (linter is run on separate tasks, currently only available for Linux and Windows). We hardly use `target_os` in this repo, as we have mostly Unix-compatible code plus some Windows-specific code, so the linting bandwidth is currently split between Linux and Windows.

PR #3980, opened 2024-08-03
fix(config): warn if host is incompatible with the toolchain in `rustup default`

### LINKED ISSUE #3972: `rustup default` doesn't show proper warning when architecture is not compatible
Fresh install rustup on my Windows Subsystem for Linux, choose `stable-x86_64-apple-darwin` toolchain
after installation, run `rustc` it reports `~/.rustup/toolchains/stable-x86_64-apple-darwin/bin/rustc: 1: Syntax error: "(" unexpected`

Or if you choose `stable-x86_64-pc-windows-gnu`, run `rustc` after installation,
it reports `error: the 'rustc' binary, normally provided by the 'rustc' component, is not applicable to the 'stable-x86_64-pc-windows-gnu' toolchain`

rami3l:
Closes #3972:

> FWIW we do emit this warning for `rustup toolchain install`:
> 
> https://github.com/rust-lang/rustup/blob/4207838d962204123f70705ff9f3e73c37b813df/src/cli/rustup_mode.rs#L820-L825
> 
> ```console
> > uname -mrs
> Darwin 23.5.0 arm64
> 
> > rustup toolchain install stable-unknown-linux-gnu --profile=minimal
> error: DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain.
> warn: toolchain 'stable-unknown-linux-gnu' may not be able to run on this system.
> warn: If you meant to build software to target that platform, perhaps try `rustup target add aarch64-unknown-linux-gnu` instead?
> info: syncing channel updates for 'stable-aarch64-unknown-linux-gnu'
> info: latest update on 2024-07-25, rust version 1.80.0 (051478957 2024-07-21)
> info: downloading component 'cargo'
> info: downloading component 'rust-std'
> ^C⏎                           
> ```
> 
> ... it could be that this is not covered by `rustup default` somehow. So adding the same warning to that path should address this issue.

_https://github.com/rust-lang/rustup/issues/3972#issuecomment-2266332004_
            
A unified `ensure_installed()` function also paves the way to addressing #2686.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -634,6 +634,22 @@ pub(crate) fn ignorable_error(
     }
 }
 
+/// Warns if rustup is trying to install a toolchain that might not be
+/// able to run on the host system.
+pub(crate) fn warn_if_host_is_incompatible(
+    toolchain: impl Display,
+    host_arch: &TargetTriple,
+    target_triple: &TargetTriple,
+    force_non_host: bool,
+) -> Result<()> {
+    if !force_non_host && !host_arch.can_run(target_triple)? {
+        error!("DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain.");
+        warn!("toolchain '{toolchain}' may not be able to run on this system.");
+        warn!("If you meant to build software to target that platform, perhaps try `rustup target add {target_triple}` instead?");
+    }
+    Ok(())
+}
+
 /// Warns if rustup is running under emulation, such as macOS Rosetta
 pub(crate) fn warn_if_host_is_emulated(process: &Process) {
     if TargetTriple::is_host_emulated() {
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -642,11 +642,12 @@ pub(crate) fn warn_if_host_is_incompatible(
     target_triple: &TargetTriple,
     force_non_host: bool,
 ) -> Result<()> {
-    if !force_non_host && !host_arch.can_run(target_triple)? {
-        error!("DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain.");
-        warn!("toolchain '{toolchain}' may not be able to run on this system.");
-        warn!("If you meant to build software to target that platform, perhaps try `rustup target add {target_triple}` instead?");
+    if force_non_host || host_arch.can_run(target_triple)? {
+        return Ok(());
     }
+    error!("DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain.");
+    warn!("toolchain '{toolchain}' may not be able to run on this system.");
+    warn!("If you meant to build software to target that platform, perhaps try `rustup target add {target_triple}` instead?");
     Ok(())
 }
 
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -15,17 +15,18 @@ use git_testament::{git_testament, render_testament};
 use tracing::{debug, error, info, trace, warn};
 
 use super::self_update;
-use crate::cli::download_tracker::DownloadTracker;
-use crate::dist::{
-    manifest::ComponentStatus, notifications as dist_notifications, TargetTriple, ToolchainDesc,
+use crate::{
+    cli::download_tracker::DownloadTracker,
+    config::Cfg,
+    dist::{
+        manifest::ComponentStatus, notifications as dist_notifications, TargetTriple, ToolchainDesc,
+    },
+    install::UpdateStatus,
+    notifications::Notification,
+    process::{terminalsource, Process},
+    toolchain::{DistributableToolchain, LocalToolchainName, Toolchain, ToolchainName},
+    utils::{notifications as util_notifications, notify::NotificationLevel, utils},
 };
-use crate::install::UpdateStatus;
-use crate::process::{terminalsource, Process};
-use crate::toolchain::{DistributableToolchain, LocalToolchainName, Toolchain, ToolchainName};
-use crate::utils::notifications as util_notifications;
-use crate::utils::notify::NotificationLevel;
-use crate::utils::utils;
-use crate::{config::Cfg, notifications::Notification};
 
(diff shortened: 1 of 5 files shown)


src/cli/common.rs
```diff
@@ -633,6 +634,22 @@ pub(crate) fn ignorable_error(
...
+/// Warns if rustup is trying to install a toolchain that might not be
+/// able to run on the host system.
+pub(crate) fn warn_if_host_is_incompatible(
+    toolchain: impl Display,
+    host_arch: &TargetTriple,
+    target_triple: &TargetTriple,
+    force_non_host: bool,
+) -> Result<()> {
+    if !force_non_host && !host_arch.can_run(target_triple)? {
+        error!("DEPRECATED: future versions of rustup will require --force-non-host to install a non-host toolchain.");
```
rami3l: I actually don't think this is a good way forward, since `host_arch.can_run()` relies on hardcoding and is fragile. The warnings below seem sufficient.

src/config.rs
```diff
@@ -777,40 +779,47 @@ impl<'a> Cfg<'a> {
...
-    ) -> Result<Toolchain<'_>> {
+        verbose: bool,
+    ) -> Result<(UpdateStatus, Toolchain<'_>)> {
+        if verbose {
+            (self.notify_handler)(Notification::LookingForToolchain(&toolchain));
+        }
         let components: Vec<_> = components.iter().map(AsRef::as_ref).collect();
         let targets: Vec<_> = targets.iter().map(AsRef::as_ref).collect();
-        let toolchain = match DistributableToolchain::new(self, toolchain.clone()) {
+        let profile = profile.map_or_else(|| self.get_profile(), Ok)?;
```
djc: This commit has a lot going on. I'd probably recommend splitting it up in set of smaller commits, maybe like this:

* Add `tracing::instrument` attribute here
* Add notifications, contingent on `verbose` flag
* More sophisticated profile selection if not specified
* Yield status from `ensure_installed()`

For the profile thing, IMO `map_or_else()` isn't great for this case, would be clearer to use an explicit `match` here:

```rust
let profile = match profile {
    Some(profile) => profile,
    None => self.get_profile()?,
};
```

src/cli/common.rs
```diff
@@ -15,17 +15,18 @@ use git_testament::{git_testament, render_testament};
 use tracing::{debug, error, info, trace, warn};
 
 use super::self_update;
-use crate::cli::download_tracker::DownloadTracker;
-use crate::dist::{
-    manifest::ComponentStatus, notifications as dist_notifications, TargetTriple, ToolchainDesc,
+use crate::{
```
djc, line 18: I think the import refactoring should get its own commit here is extensive enough that it merits its own commit.

src/cli/common.rs
```diff
@@ -633,6 +634,22 @@ pub(crate) fn ignorable_error(
...
 
+/// Warns if rustup is trying to install a toolchain that might not be
+/// able to run on the host system.
+pub(crate) fn warn_if_host_is_incompatible(
+    toolchain: impl Display,
+    host_arch: &TargetTriple,
+    target_triple: &TargetTriple,
+    force_non_host: bool,
+) -> Result<()> {
+    if !force_non_host && !host_arch.can_run(target_triple)? {
```
djc: Invert and early return to reduce rightward drift?

src/cli/common.rs
djc: Sounds good to me. Separate commit?

djc (commented):
I like the direction here, would mainly like a cleaner commit history here to make it easier to review the details.

rami3l:
@djc Thanks for your guidance! I'm a bit occupied with other stuff right now so I'll take another look later.

Just to be sure, ~~I actually prefer removing `--force-non-host` and the related deprecation notice~~ after [searching](https://github.com/search?q=%22--force-non-host%22&type=code) for the existing usages I think we should keep the flag itself, but remove the deprecation notice and leave the warnings as-is. Is that okay to you?

src/cli/common.rs
rami3l: @djc Let's get the main change merged first and then discuss how to handle this detail in another issue/PR, since it looks like we already have some users who rely on this flag: https://github.com/search?q=%22--force-non-host%22&type=code

src/config.rs
```diff
@@ -784,14 +786,18 @@ impl<'a> Cfg<'a> {
         components: Vec<String>,
         targets: Vec<String>,
         profile: Option<Profile>,
-    ) -> Result<Toolchain<'_>> {
+        verbose: bool,
+    ) -> Result<(UpdateStatus, Toolchain<'_>)> {
+        if verbose {
```
djc, line 800: The notifications aren't mentioned in the commit message, and don't seem to be related to the return type. Separate commit?

djc (approved):

src/config.rs
rami3l: @djc This is very weird, could be a rebase error on my side. Will fix.

djc: Yeah, those sometimes happen. Usually for me it happens when during a rebase I accidentally use `git commit --amend -a` when I actually meant `git rebase --continue`.

rami3l: @djc Oh, it looks like `git absorb` has messed up my history :( I'll be more careful next time.

rami3l: The original commit was there: https://github.com/rust-lang/rustup/commit/07777feac5e4b85846662335c541a2a57222238e

rami3l: I cherry-picked that one out. The history should be okay now.

rbtcollins:
> Just to be sure, I actually prefer removing --force-non-host and the related deprecation notice after [searching](https://github.com/search?q=%22--force-non-host%22&type=code) for the existing usages I think we should keep the flag itself, but remove the deprecation notice and leave the warnings as-is. Is that okay to you?

@rami3l  So the point of this was to support cross-rs properly (they need non-host toolchains), while stopping people that are just following their nose from getting stuck and confused. E.g. installing an arm64 toolchain on an amd64 machine and wondering why that breaks.

cross-rs should be updated by now I believe, so the appropriate action is to remove the warning and reject non-host toolchains unless the flag is provided.

Issue #3635, opened 2024-01-07
Tracking: Rustup should not perform implicit installations


jyn514:
### Problem you are trying to solve

here is a transcript of a session with rustup
```
PS C:\Users\jyn\src\example> rustup toolchain list                              
nightly-x86_64-pc-windows-msvc (default)
1.74-x86_64-pc-windows-msvc
stage1
PS C:\Users\jyn\src\example> rustup uninstall nightly
info: uninstalling toolchain 'nightly-x86_64-pc-windows-msvc'
info: toolchain 'nightly-x86_64-pc-windows-msvc' uninstalled
PS C:\Users\jyn\src\example> rustup toolchain list   
1.74-x86_64-pc-windows-msvc
stage1
PS C:\Users\jyn\src\example> rustc
info: syncing channel updates for 'nightly-x86_64-pc-windows-msvc'
info: latest update on 2024-01-07, rust version 1.77.0-nightly (b6a8c762e 2024-01-06)
info: downloading component 'cargo'
^C
PS C:\Users\jyn\src\example> rustc +nightly
error: toolchain 'nightly-x86_64-pc-windows-msvc' is not installed
PS C:\Users\jyn\src\example> rustup --version
rustup 1.26.0 (5af9b9484 2023-04-05)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: syncing channel updates for 'nightly-x86_64-pc-windows-msvc'
```

### Proposed Solution & Progress

[The following behaviors of the current `rustup` should be changed:]

- [x] uninstalling the default toolchain happens silently. i would expect either a warning (by the same logic as https://github.com/rust-lang/rustup/issues/3319), or for rustup to switch to a different default toolchain.
- [x] running `rustc` afterwards implicitly reinstalls the toolchain.
- [x] ~~I do not expect rustup to implicitly install in any case other than when `rust-toolchain.toml` is in the current directory. in fact, you can see that `rustc +nightly` does *not* implicitly install, only `rustc` by itself.~~ [@rami3l: We shouldn't do even that, instead we should let the user explicitly install the required toolchain with a command.]
- [x] running `rustup --version` also silently installs.
- [x] running `rustup show` also silently installs.

### Notes

rustup 1.26.0 (5af9b9484 2023-04-05)

### Tasks
- [ ] https://github.com/rust-lang/rustup/pull/3948

rami3l:
@jyn514 Thanks a lot for filing this issue! I see basically two points being made here:

[...]

PR #3983, opened 2024-08-06
feat(rustup-mode): install the active toolchain by default on `rustup toolchain install`

### LINKED ISSUE #2686: Command for "install stuff from `rust-toolchain.toml`"

rami3l:
Closes #2686 by extending the existing `rustup toolchain install` and `rustup install` commands, as discussed in https://github.com/rust-lang/rustup/issues/3635#issuecomment-2260336594:

> > What would be the semantics here? Would it only install if there is a `rust-toolchain.toml` file or would it install any missing default toolchain?
> 
> My understanding is that it is expected to do anything `rustc --version` currently does before actually running that command on the corresponding `rustc`, it's just that everything will become explicit now. So yes, this includes (and isn't limited to) installing the toolchain specified in the `rust-toolchain.toml` file.
> 
> > And presumably the `rustup install` alias.
> 
> Yes! As the two subcommands has been unified with #3596, that looks like a less painful move indeed. A single change for these two places!

TLDR: This PR makes `rustup toolchain install` and `rustup install` install the active toolchain, i.e. behave like the hypothetical `rustup toolchain ensure` command, when no toolchain is specified.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -107,7 +107,7 @@ enum RustupSubcmd {
         opts: UpdateOpts,
     },
 
-    /// Uninstall Rust toolchains
+    /// Uninstall the given toolchains
     #[command(hide = true)]
     Uninstall {
         #[command(flatten)]
@@ -309,7 +309,7 @@ enum ToolchainSubcmd {
         opts: UpdateOpts,
     },
 
-    /// Uninstall a toolchain
+    /// Uninstall the given toolchains
     #[command(alias = "remove")]
     Uninstall {
         #[command(flatten)]
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -584,7 +584,7 @@ pub async fn main(current_dir: PathBuf, process: &Process) -> Result<utils::Exit
 
     match subcmd {
         RustupSubcmd::DumpTestament => common::dump_testament(process),
-        RustupSubcmd::Install { opts } => update(cfg, opts).await,
+        RustupSubcmd::Install { opts } => update(cfg, opts, false).await,
         RustupSubcmd::Uninstall { opts } => toolchain_remove(cfg, opts),
         RustupSubcmd::Show { verbose, subcmd } => handle_epipe(match subcmd {
             None => show(cfg, verbose),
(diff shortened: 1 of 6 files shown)


djc (approved):

djc:
I think conceptually it would be a good next step to break `find_or_install_active_toolchain()` into find and install parts.

rami3l:
> I think conceptually it would be a good next step to break `find_or_install_active_toolchain()` into find and install parts.

@djc The whole story is actually more complex than that. We already have `find_active_toolchain()` but it doesn't equal to `find_or_install_active_toolchain()` minus `ensure_installed()` or something like that. Currently `find_or_install_active_toolchain()` doesn't even call `find_active_toolchain()`, it's a real mess that happens to get this job done.

My plan is that when removing implicit installations altogether, this new call site of that `find_or_install_active_toolchain()` will be the only one, and thus I can start inlining/refactoring it so that all the `verbose` logs and/or sanity checks can hopefully be included.

ChrisDenton:
Is there documentation of the new `rustup toolchain install` behaviour?

weihanglo:
While not really in the same situation, here is the background of why Cargo moved away from implicit `cargo install` to `cargo install --path .`: <https://github.com/rust-lang/cargo/issues/5327>.

ChrisDenton:
Hm, I think rustup is kinda in the exact opposite position. As that issue says, you (almost) never want to `cargo install` the crate in the current directory but you (almost) always want to `rustup install` the toolchain in the crate's rust-toolchain file if you're working on that crate.

rami3l:
> Is there documentation of the new `rustup toolchain install` behaviour?

@ChrisDenton This PR has already changed the `--help` output, would you mind sharing what you have in mind?

ChrisDenton:
I was thinking more of online documentation. Although maybe that should wait until a release. Something along the lines of advice for people using CI. Because presumably we want to guide people towards using this rather than relying on auto install as a side effect of other commands.

rami3l:
> I was thinking more of online documentation. Although maybe that should wait until a release. Something along the lines of advice for people using CI. Because presumably we want to guide people towards using this rather than relying on auto install as a side effect of other commands.

@ChrisDenton The plan is to get rid of implicit installations immediately (#3985), and I promise that will be mentioned in highlight when we cut a new release 🙏

PR #3986, opened 2024-08-09
refactor(config)!: simplify logic

rami3l:
Continuation of #3985.

I originally wanted to unify `async fn find_or_install_active_toolchain()` and `fn find_active_toolchain()`, but since this essentially involves plugging an async part in the middle of the latter, I finally backed out on this change. Instead, I rewrote the former to look more like the latter, thanks to the fact that I'm now free to change what the former returns.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -711,13 +711,9 @@ impl<'a> Cfg<'a> {
         &self,
         name: Option<ResolvableToolchainName>,
     ) -> Result<Toolchain<'_>> {
-        let toolchain = match name {
-            Some(name) => {
-                let desc = name.resolve(&self.get_default_host_triple()?)?;
-                Some(desc.into())
-            }
-            None => None,
-        };
+        let toolchain = name
+            .map(|name| anyhow::Ok(name.resolve(&self.get_default_host_triple()?)?.into()))
+            .transpose()?;
         self.local_toolchain(toolchain)
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -741,18 +741,18 @@ impl<'a> Cfg<'a> {
 
     #[tracing::instrument(level = "trace", skip_all)]
     pub(crate) async fn find_or_install_active_toolchain(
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -744,45 +744,32 @@ impl<'a> Cfg<'a> {
         &self,
         verbose: bool,
     ) -> Result<(LocalToolchainName, ActiveReason)> {
-        match self.find_override_config()? {
-            Some((override_config, reason)) => match override_config {
-                OverrideCfg::PathBased(path_based_name) => {
-                    let toolchain = Toolchain::with_reason(self, path_based_name.into(), &reason)?;
```
rami3l, line 754: Since we're only getting `LocalToolchainName` now, it's no longer necessary to construct `Toolchain` with `with_reason()`. Instead, `override_config.into_local_toolchain_name()` should be enough.

rami3l: Oops, actually not when a custom toolchain is active. I'll revert this for now...

src/config.rs
```diff
@@ -744,45 +744,32 @@ impl<'a> Cfg<'a> {
...
-                    let reason = ActiveReason::Default;
-                    let toolchain = self
-                        .ensure_installed(&toolchain_desc, vec![], vec![], None, verbose)
-                        .await?
-                        .1;
-                    Ok((toolchain.name().clone(), reason))
-                }
-            },
+        if let Some((override_config, reason)) = self.find_override_config()? {
+            let toolchain = override_config.clone().into_local_toolchain_name();
```
rami3l, line 748: Since we're only getting LocalToolchainName now, calling `override_config.into_local_toolchain_name()` should be enough. There's no need for more complicated pattern matching other than checking whether the toolchain is official: if it is, we ensure its installation via `ensure_installed()`; if it isn't, we ensure that the override reason is valid via `Toolchain::with_reason()`.

djc (approved):

PR #3987, opened 2024-08-10
feat(cli)!: set log level to `INFO`/`DEBUG` on `--quiet`/`--verbose` if `RUSTUP_LOG` is unset

rami3l:
In the current setting, `debug!()` lines are not printed out even if I use `RUSTUP_LOG=rustup=DEBUG`. This logic used to make sense when we didn't have support for directives, but now it feels a bit awkward.

As a solution, I decided to follow the basic idea of #3911:
- `RUSTUP_LOG` decides ultimately when it's present.
- When `RUSTUP_LOG` is unset, `--quiet` or `--verbose` gets to decide the log verbosity.

Overall, now `--quiet` and `--verbose` behave more like what would be expected from general CLI applications.

This applies to both `rustup-mode` (i.e. `rustup`) and `setup-mode` (i.e. `rustup-init`).

### DIFF
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -69,7 +69,7 @@ where
         (logger.compact().with_filter(env_filter).boxed(), handle)
     } else {
         // Receive log lines from Rustup only.
-        let (env_filter, handle) = reload::Layer::new(EnvFilter::new("rustup=DEBUG"));
+        let (env_filter, handle) = reload::Layer::new(EnvFilter::new("rustup=INFO"));
         (
             logger
                 .event_format(EventFormatter)
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -545,7 +545,7 @@ pub async fn main(current_dir: PathBuf, process: &Process) -> Result<utils::Exit
         Err(err) if err.kind() == DisplayVersion => {
             write!(process.stdout().lock(), "{err}")?;
             info!("This is the version for the rustup toolchain manager, not the rustc compiler.");
-            let mut cfg = common::set_globals(current_dir, false, true, process)?;
+            let mut cfg = common::set_globals(current_dir, true, process)?;
             match cfg.active_rustc_version() {
                 Ok(Some(version)) => info!("The currently active `rustc` version is `{version}`"),
                 Ok(None) => info!("No `rustc` is currently active"),
@@ -570,7 +570,7 @@ pub async fn main(current_dir: PathBuf, process: &Process) -> Result<utils::Exit
         }
     };
 
-    let cfg = &mut common::set_globals(current_dir, matches.verbose, matches.quiet, process)?;
+    let cfg = &mut common::set_globals(current_dir, matches.quiet, process)?;
 
     if let Some(t) = &matches.plus_toolchain {
         cfg.set_toolchain_override(t);
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -10,10 +10,11 @@ use clap::{builder::PossibleValue, Args, CommandFactory, Parser, Subcommand, Val
 use clap_complete::Shell;
 use itertools::Itertools;
 use tracing::{info, trace, warn};
+use tracing_subscriber::{reload::Handle, EnvFilter, Registry};
 
 use crate::{
     cli::{
-        common::{self, PackageUpdate},
+        common::{self, update_console_filter, PackageUpdate},
         errors::CLIError,
         help::*,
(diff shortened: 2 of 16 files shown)


src/cli/log.rs
```diff
@@ -69,7 +69,7 @@ where
         (logger.compact().with_filter(env_filter).boxed(), handle)
     } else {
         // Receive log lines from Rustup only.
-        let (env_filter, handle) = reload::Layer::new(EnvFilter::new("rustup=DEBUG"));
+        let (env_filter, handle) = reload::Layer::new(EnvFilter::new("rustup=INFO"));
```
rami3l, line 72: Since `debug!()` for notifications is no longer under `if verbose`, we'll have to change this to `INFO` to prevent `debug` lines from showing up in the default state. Instead, we use `Handle<_, _>` to change the log level when we detect flags like `--verbose` and `--quiet`.

src/cli/setup_mode.rs
```diff
@@ -25,12 +25,12 @@ use crate::{
     before_help = format!("rustup-init {}", common::version()),
 )]
 struct RustupInit {
-    /// Enable verbose output
-    #[arg(short, long)]
+    /// Enable verbose output, limit console logger level to 'DEBUG' if 'RUSTUP_LOG' is unset
```
djc: Conceptually it feels like CLI flags should have higher specificity than environment variables. I think the reason not to do so here is because the environment variable is more expressive (in the sense that it can set levels by crate etc), right?

The docstring seems a little verbose/awkward, maybe just "Set log level to 'DEBUG' (if 'RUSTUP_LOG' is unset)"?

Why use `overrides_with` here instead of `conflicts_with`? The meaning of `overrides_with` is not clear to me in case both are specified -- maybe it means last wins?

src/cli/setup_mode.rs
```diff
@@ -115,10 +115,17 @@ pub async fn main(
         warn!("{}", common::WARN_COMPLETE_PROFILE);
     }
 
-    if quiet && process.var("RUSTUP_LOG").is_err() {
-        console_filter
-            .modify(|it| *it = EnvFilter::new("rustup=WARN"))
-            .expect("error reloading `EnvFilter` for console_logger");
+    if process.var("RUSTUP_LOG").is_err() {
+        if quiet {
+            console_filter
```
djc: Wonder whether we could deduplicate this a bit more.

src/cli/rustup_mode.rs
```diff
@@ -68,12 +69,12 @@ fn handle_epipe(res: Result<utils::ExitCode>) -> Result<utils::ExitCode> {
     after_help = RUSTUP_HELP,
 )]
 struct Rustup {
-    /// Enable verbose output
-    #[arg(short, long)]
+    /// Enable verbose output, limit console logger level to 'DEBUG' if 'RUSTUP_LOG' is unset
```
djc: As for setup-mode.

src/cli/setup_mode.rs
rami3l: > Conceptually it feels like CLI flags should have higher specificity than environment variables. I think the reason not to do so here is because the environment variable is more expressive (in the sense that it can set levels by crate etc), right?

@djc Yes, I'd like to say that this is `RUSTUP_LOG` instead of something like [`CARGO_TERM_VERBOSE`](https://doc.rust-lang.org/cargo/reference/config.html#termverbose), so they should not be directly comparable.

> The docstring seems a little verbose/awkward, maybe just "Set log level to 'DEBUG' (if 'RUSTUP_LOG' is unset)"?

Will fix.

> Why use `overrides_with` here instead of `conflicts_with`? The meaning of `overrides_with` is not clear to me in case both are specified -- maybe it means last wins?

Yes, this was stated in the PR description. However looking at what `cargo` currently does with `--quiet` and `--verbose`, I guess it'll still be better to revert this change...

src/cli/setup_mode.rs
rami3l: @djc Done.

djc: I would suggest like this:

```rust
let filter = match (quiet, verbose) {
    (true, _) => Some("rustup=WARN"),
    (_, true) => Some("rustup=DEBUG"),
    (_, _) => None,
};

if let Some(filter) = filter {
    console_filter.modify(|it| *it = EnvFilter::new(filter)).expect(..);
}
```

djc (approved):

djc:
Sorry for the slow feedback, this is looking good!

src/cli/setup_mode.rs
rami3l: @djc Sure thing!

PR #3991, opened 2024-08-14
fix(manifest): consider possible renames in `Component::try_new()`

rami3l:
Continuation of #3601. Addresses the following concern:

> I saw the fix of this issue was added into milestone 1.27.1.
> But I still met this bug with rustup 1.27.1.
> 
> Execute:
> ```
> rustup component remove llvm-tools-x86_64-unknown-linux-gnu
> ```
> 
> Output:
> ```
> error: toolchain 'stable-x86_64-unknown-linux-gnu' does not contain component 'llvm-tools-x86_64-unknown-linux-gnu' for target 'x86_64-unknown-linux-gnu'
> ```
> 
> But the following command works well:
> ```
> rustup component remove --target x86_64-unknown-linux-gnu llvm-tools
> ```

_by @yangby-cryptape in https://github.com/rust-lang/rustup/issues/3166#issuecomment-2285774658_

### DIFF
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -498,19 +498,11 @@ impl Component {
         distributable: &DistributableToolchain<'_>,
         fallback_target: Option<&TargetTriple>,
     ) -> Result<Self> {
+        let manifest = distributable.get_manifest()?;
         for component_status in distributable.components()? {
-            let short_name = component_status.component.short_name_in_manifest();
-            let target = component_status.component.target.as_ref();
-
-            if name.starts_with(short_name)
-                && target.is_some()
-                && name == format!("{}-{}", short_name, target.unwrap())
-            {
-                return Ok(Component::new(
-                    short_name.to_string(),
-                    target.cloned(),
-                    false,
-                ));
+            let component = component_status.component;
+            if name == component.name_in_manifest() || name == component.name(&manifest) {
+                return Ok(component);
             }
         }
 
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -1543,6 +1543,38 @@ async fn add_component_by_target_triple() {
 }
 
 #[tokio::test]
+async fn add_component_by_target_triple_renamed_from() {
+    let mut cx = CliTestContext::new(Scenario::SimpleV2).await;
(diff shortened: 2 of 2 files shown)


src/dist/manifest.rs
```diff
@@ -498,19 +498,11 @@ impl Component {
...
-                && target.is_some()
-                && name == format!("{}-{}", short_name, target.unwrap())
-            {
-                return Ok(Component::new(
-                    short_name.to_string(),
-                    target.cloned(),
-                    false,
-                ));
+            let component = component_status.component;
+            if name == component.name_in_manifest() || name == component.name(&manifest) {
```
rami3l, line 504: `component.name_in_manifest()` and `component.name()` give the _full name_ of the component after and before a rename respectively (e.g. `llvm-tools-preview-aarch64-apple-darwin` and `llvm-tools-aarch64-apple-darwin`).

As they already take the optional target into account, there's no need to build the component name in-place with `format!("{}-{}", short_name, target.unwrap())`.

Also, when the target is `None`, `component.name()` is the same as `component.short_name()`. In this case, we shouldn't let it slip to the fallback case where `fallback_target` is used.

src/dist/manifest.rs
```diff
@@ -498,19 +498,11 @@ impl Component {
         distributable: &DistributableToolchain<'_>,
         fallback_target: Option<&TargetTriple>,
     ) -> Result<Self> {
+        let manifest = distributable.get_manifest()?;
```
djc, line 501: It looks like `distributable.components()` also contains a call to `distributable.get_manifest()`, so doing that twice might be kind of wasteful? Is it worth doing that differently?

djc (approved):

src/dist/manifest.rs
rami3l: @djc For the moment I'd like to keep it that way for readability, as this really isn't a hot function. I agree that the implementation is weird underneath, but as soon as I replace that with an inlined `.components()` I realized `.get_manifest()` actually calls `.get_manifestation()` which belongs to another layer of problems to be solved... Let's keep it this way first?

It definitely requires some refactoring afterwards anyway.

djc: Works for me!

PR #3999, opened 2024-08-19
Upgrade pulldown-cmark to 0.12

djc:
Replaces #3993. I'm not entirely sure if the definition list handling makes sense, but since we don't have any definition lists right now it's probably not a big deal -- we can improve the formatting later.

See discussion in https://github.com/pulldown-cmark/pulldown-cmark/issues/67 and https://talk.commonmark.org/t/description-list/289/2.


rami3l (approved):
Thanks for your efforts! Meanwhile I'm wondering whether this markdown module has been covered by a third-party library... 🤔

PR #4021, opened 2024-09-16
Update semver-compatible dependencies, except openssl-src

djc:
Replaces #4020 with a version that doesn't update openssl-src, because of the [breakage](https://github.com/openssl/openssl/pull/25367) from their 3.3.2 release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,9 +86,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.87"
+version = "1.0.89"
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -2048,9 +2048,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.12"
+version = "0.23.13"
```
djc, line 2051: This is actually pretty nice because aws-lc-rs 1.9.0 comes with optional prebuilt ASM, which rustls now enables by default.

rami3l (approved):

PR #4023, opened 2024-09-16
Try symlinking proxies first, falling back to hardlinking if that fails

### LINKED ISSUE #4022: Consider using symlinks for rustup proxies
This was brought up elsewhere so I'm writing an issue here to note the conversation. I'm happy to implement this if it's acceptable (though I don't mind if someone else wants to claim it).

Currently rustup proxies are created using hard links with a fallback to soft links if creating hard links aren't supported by the filesystem or the platform (e.g. android). This proposal is to swap the order so that proxies are created using soft links with a fallback to hard links if creating soft links aren't supported by the filesystem or platform (e.g. Windows in the default configuration).

Hard links have the issue that they are, for most purposes, normal files. So many copy utilities (including `cp`) will just naively copy the file. This can be a problem when copying the rustup directory as all the hard linked proxies will be turned into literal copies of rustup.

Using symlinks for this use case is common in Unix-like environments. E.g. busybox is a canonical example. I have not been able to find any real downside to this approach (though I may have missed something).

ChrisDenton:
This reverse the previous order.

Closes #4022

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -731,8 +731,8 @@ pub(crate) fn install_proxies(process: &Process) -> Result<()> {
     let mut tool_handles = Vec::new();
     let mut link_afterwards = Vec::new();
 
-    // Try to hardlink all the Rust exes to the rustup exe. Some systems,
-    // like Android, does not support hardlinks, so we fallback to symlinks.
+    // Try to symlink all the Rust exes to the rustup exe. Some systems,
+    // like Windows, do not always support symlinks, so we fallback to hard links.
     //
     // Note that this function may not be running in the context of a fresh
     // self update but rather as part of a normal update to fill in missing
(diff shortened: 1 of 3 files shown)


djc (approved):

rami3l (commented):
@ChrisDenton This PR looks good overall, I second @djc.

I have just one concern about what this would mean for existing users with existing hard links. Will this cause problems WRT upgrades?

ChrisDenton:
Oh, looking at this I did find one issue, The [`raw::hardlink`](https://github.com/rust-lang/rustup/blob/ccc668ccf852b7f37a4072150a6dd2aac5844d38/src/utils/raw.rs#L229) function sneakily includes a load bearing `let _ = fs::remove_file(dest);` call whereas there's no symlink equivalent. I should probably move that to a higher level rather than it being so implicit.

Otherwise there shouldn't be an issue as far as I can tell from the `install_proxies` function. We use the `same-file` crate to test if two handles/fds point to the same file. And we open the handle by following symlinks so it'll work in either case.

rami3l (approved):
Thanks for the explanation! I guess we can try this out as a part of the upcoming release.

PR #4025, opened 2024-09-18
Note that selecting VS lang packs is optional

ChrisDenton:
Rust used to mangle linker output for non-ascii error messages. The workaround was to install the Visual Studio English Language Pack so that most error messages would be mostly ascii (file names aside). However this was [fixed](https://github.com/rust-lang/rust/pull/110586) in Rust 1.71 (released July 2023) so it's no longer required. We should still recommend installing the English pack both for the sake of compatibility and because rustc's output is itself still in English unless a nightly feature is used.


djc (approved):

rami3l (approved):

rami3l:
@ChrisDenton Thanks a lot :D

PR #4026, opened 2024-09-18
docs(README): Point out where to find nightly/master docs.

LunarLambda:
Change was briefly discussed with @rami3l over Discord.

I'm not sure if this is the most clear wording, and whether the book itself should point out where to find the nightly/master version. Can adjust both if needed.


rami3l (approved):
Thanks a lot! Regarding the book itself I'm not very sure; how are other books across the org doing -- I assume nothing special?

LunarLambda:
yeah nothing special I believe.

rami3l:
Alright, let's merge this first :)

PR #4032, opened 2024-09-25
Update remove_dir_all

### LINKED ISSUE #4022: Consider using symlinks for rustup proxies

ChrisDenton:
Fixes #4022 (specifically https://github.com/rust-lang/rustup/issues/4022#issuecomment-2362612669)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1926,9 +1926,9 @@ checksum = "7a66a03ae7c801facd77a29370b4faec201768915ac14a721ba36f20bc9c209b"
 
 [[package]]
 name = "remove_dir_all"
-version = "0.8.3"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c914caef075f03e9d5c568e2e71b3d3cf17dc61a5481ff379bb744721be0a75a"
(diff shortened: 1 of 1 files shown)


djc (approved):

rami3l (approved):

rami3l:
Thanks a lot for the investigation @ChrisDenton! Let's merge it and see how it goes on our CI.

---

It seems to work perfectly! Thanks again!

PR #4034, opened 2024-09-26
Remove unnecessary methods

InfyniteHeap:
Coming from this: https://github.com/rust-lang/rustup/pull/3896#discussion_r1704153219. The `get/set_value()` are actually the alias of `get/set()`. So, it should be removed.


rami3l (approved):
LGTM :)

PR #4044, opened 2024-10-07
Update semver-compatible dependencies

djc:
Supersedes #4043. This is still a thing:

> Since we're waiting for a potential OpenSSL v3.3.3, [openssl/openssl#25367 (comment)](https://github.com/openssl/openssl/pull/25367#issuecomment-2338141250) suggestion of running `cargo update openssl-src --precise 300.3.1+3.3.1` still applies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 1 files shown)


rami3l (approved):
Thanks a lot :)

PR #4056, opened 2024-10-21
Update semver-compatible dependencies, except openssl-src

djc:
Alternative to #4055. Still holding back openssl-src, as in #4049 and earlier.


rami3l (approved):
Many thanks :D

ChrisDenton (approved):
Makes sense to me

PR #4041, opened 2024-10-06
dist: bump `rustup` version to `1.28.0`

rami3l:
This is the 1st PR for the `1.28.0` release according to our [release process](https://rust-lang.github.io/rustup/dev-guide/release-process.html#making-a-release).

## Checklist

- [x] Bump version in the manifest.
- [ ] Update the changelog.
- [ ] ~~Draft the release blog post~~ (Not required, since we're doing a beta release).

---

## [1.28.0] - 2024-10-07

This new release might be one of rustup's biggest updates in quite a while. In the making since mid-2023, it has combined a significant amount of new features and internal cleanups.

The headlines for this release are:

1. **BREAKING:** Rustup will no longer automatically install the active toolchain if it is currently uninstalled [pr#3225] [pr#3948] [pr#3985]
   - To ensure its installation, run `rustup toolchain install` with **no** arguments [pr#3983]
   - **TIP:** The following command installs the active toolchain both before and after this change
      ```bash
      rustup show active-toolchain || rustup toolchain install
      # Or, on older versions of PowerShell:
      rustup show active-toolchain; if ($LASTEXITCODE -ne 0) { rustup toolchain install }
      ```
2. **BREAKING:** When creating executable proxy links, Rustup will now try symlinks first and fall back to hardlinks, as opposed to the hardlink-first policy in previous versions [pr#4023]
3. **BREAKING:** Installing a host-incompatible toolchain via `rustup toolchain install` or `rustup default` will now be rejected without explicitly adding `--force-non-host` [pr#4028]
4. As part of rustup's downloader simplification plan, `reqwest-rustls-tls` is now our new default download backend [pr#3798]
   - `RUSTUP_USE_CURL` and `RUSTUP_USE_RUSTLS` environment variables remain valid in this new version, but the default value of `RUSTUP_USE_RUSTLS` has been changed to `1`
   - Rustup now uses `rustls-platform-verifier` when `rustls` is activated for better platform compatibility [pr#3903]
   - Rustup's TLS backend when using `rustls` has been changed from `ring` to `aws-lc-rs` to support P521 signatures [pr#3898]
   - The `reqwest-default-tls` feature flag is now renamed `reqwest-native-tls` [pr#3888]
[...]

PR #4061, opened 2024-10-24
chore: use more workspace-wide configurations

rami3l:
This PR does the following cleanup WRT `Cargo.toml`:

* Updated `Cargo.toml` to use workspace settings for `version`, `edition`, and `license` fields (cherry-picked from #4041) [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L3-L8) [[2]](diffhunk://#diff-96c60da7e4be39230f90e0d59f4bc52cbbc12cb8330c5df73313c4e63dd3e014L3-R5)
* Added workspace linting configurations for Rust idioms and clippy lints. (`Cargo.toml`, `download/Cargo.toml`) [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R164-R171) [[2]](diffhunk://#diff-96c60da7e4be39230f90e0d59f4bc52cbbc12cb8330c5df73313c4e63dd3e014R35-R37)

This also prepares the repo for its upcoming migration to Rust 2024 by making it easier to enable `rust-2024-compatibility` across the project. With this change, that will look like:

```toml
[workspace.lints.rust]
rust_2018_idioms = { level = "deny", priority = -1 }
rust-2024-compatibility = { level = "warn", priority = -2 }
```

### DIFF
--- a/src/diskio/threaded.rs
+++ b/src/diskio/threaded.rs
@@ -268,6 +268,7 @@ impl<'a> Executor for Threaded<'a> {
                 prev_files as u64,
             ));
         }
+        #[allow(clippy::print_stderr)]
         if prev_files > 50 {
             eprintln!("{prev_files} deferred IO operations");
         }
--- a/src/test/mock/clitools.rs
+++ b/src/test/mock/clitools.rs
@@ -668,8 +668,11 @@ impl Config {
             println!("expected.ok: true");
             print_indented("expected.stdout", stdout);
             print_indented("expected.stderr", stderr);
-            dbg!(out.stdout == stdout);
-            dbg!(out.stderr == stderr);
+            #[allow(clippy::dbg_macro)]
+            {
+                dbg!(out.stdout == stdout);
+                dbg!(out.stderr == stderr);
+            }
             panic!();
         }
     }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
 name = "rustup"
-version = "1.27.1"
-edition = "2021"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
 description = "Manage multiple rust installations with ease"
 homepage = "https://github.com/rust-lang/rustup"
 keywords = ["rustup", "multirust", "install", "proxy"]
-license = "MIT OR Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/rust-lang/rustup"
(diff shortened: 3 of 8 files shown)


ChrisDenton (approved):
Workspace inheritance is more verbose than I'd like but only setting information in one place is definitely an advantage.

src/diskio/threaded.rs
```diff
@@ -268,6 +268,7 @@ impl<'a> Executor for Threaded<'a> {
                 prev_files as u64,
             ));
         }
+        #[allow(clippy::print_stderr)]
```
djc, line 271: Why should this be exempted from the logging policy?

src/test/mock/clitools.rs
```diff
@@ -668,8 +668,11 @@ impl Config {
             println!("expected.ok: true");
             print_indented("expected.stdout", stdout);
             print_indented("expected.stderr", stderr);
-            dbg!(out.stdout == stdout);
```
djc, line 671: IMO `dbg!()` in test-only code can actually be pretty useful, since it will typically only be shown when a test fails anyway. So I would recommend that we `allow(dbg_macro)` for all of the `test` module.

(Apparently clippy already allows its use directly in tests.)

src/diskio/threaded.rs
rami3l: @djc Nice catch! It's not a deliberate exemption actually, it's just that I found it a bit hard to determine where to inject `process` in here, but I could be doing it wrong.

Update: It just occurred to me that, as `process`-injected `stderr` is part of the test result, it might make some sense to use the logger here with a lower priority level instead, so that this line won't mess with that?

src/test/mock/clitools.rs
rami3l: @djc That's a pretty good suggestion. Will do.

src/diskio/threaded.rs
djc: Sounds okay to me.

src/test/mock/clitools.rs
rami3l: Addressed in #4063.

src/diskio/threaded.rs
rami3l: Addressed in #4063.

Issue #4069, opened 2024-11-08
add lint names to `rustup doc` topic list


lolbinarycat:
### Problem you are trying to solve

it is inconvenient to try to navigate to the page for a specific lint in the local documentation. 

### Solution you'd like

typing `rustup doc lintname` takes you to the page for that lint.  this would work for rustc lints, rustdoc lints, clippy lints, and potentially other tool lints.

for example, `rustup doc rustdoc::broken_intra_doc_links` would take you to `file:$TOOLCHAIN/share/doc/rust/html/rustdoc/lints.html#broken_intra_doc_links`.

rami3l:
@lolbinarycat Thanks for filing this issue!

Indeed, `rustup doc` has kind of fallen into disrepair over the years, so the user experience might not be optimal.

Your proposal is perfectly valid, it's just that I might not implement it as described above now that we have `rustup doc --rustdoc` for that already.

The relevant snippet is here:

https://github.com/rust-lang/rustup/blob/708ffd6aeaa84d291d2a16cfd99bb45ae7e1e575/src/cli/rustup_mode.rs#L1499-L1505

... which is not idiomatic, of course, since whenever `topic` (e.g. `std::mem`) is present, we won't even look at `doc_path` (e.g. `--rustdoc`).

So I think we should consider these two altogether, for instance allowing `rustup doc --rustdoc lints`, and that's pretty easy.

However, there is already enough magic in the current implementation in the way it searches for stdlib items, so we should never explicitly point to lints just like we did with stdlib... I wonder how this will eventually end up like.
Does `rustup doc --rustdoc lints::broken_intra_doc_links` look okay to you?

I mean, we can try with the following semantics with `lints::broken_intra_doc_links`:

1. Look for `lints/broken_intra_doc_links/index.html`.
2. Look for `lints/broken_intra_doc_links.html`.
3. Look for `lints.html`. If it exists, resolve to `lints.html#broken_intra_doc_links`.
4. Throw an error.

PS: I have manually verified that this algorithm, if implemented correctly, will work with `rustc`, `rustdoc` and `clippy` books.

lolbinarycat:
> However, there is already enough magic in the current implementation in the way it searches for stdlib items, so we should never explicitly point to lints just like we did with stdlib...

Aren't search engines generally supposed to be "magic"?

Although, ideally, it would be possible to unify this logic with something that does the same for documentation generated by `cargo doc`...

rami3l:
[...]

PR #4070, opened 2024-11-09
feat(cli/rustup-mode): allow `rustup doc` with both a flag and a topic

### LINKED ISSUE #3345: `rustup doc --book` to directly link to subsections and paragraphs of the book

rami3l:
This PR is my stab at the concern in https://github.com/rust-lang/rustup/issues/4069.

However, I think `rustup` should not magically detect and acknowledge lints, so I've made a more neutral solution that might apply to more topics:

> Does `rustup doc --rustdoc lints::broken_intra_doc_links` look okay to you?
> 
> I mean, we can try with the following semantics with `lints::broken_intra_doc_links`:
> 
> 1. Look for `lints/broken_intra_doc_links/index.html`.
> 2. Look for `lints/broken_intra_doc_links.html`.
> 3. Look for `lints.html`. If it exists, resolve to `lints.html#broken_intra_doc_links`.
> 4. Throw an error.
> 
> PS: I have manually verified that this algorithm, if implemented correctly, will work with `rustc`, `rustdoc` and `clippy` books.
_https://github.com/rust-lang/rustup/issues/4069#issuecomment-2466043060_

## Example

```console
> rustup doc --rustc platform-support::armv6k-nintendo-3ds::building-the-target
```

## Concerns

- [x] ~~Should we add test cases for this PR, and how may we add them?~~ Added `--path` tests to ensure that we can get the doc paths right.
- [ ] Due to a restriction with our upstream [Seeker14491/opener](https://github.com/Seeker14491/opener) and/or the underlying macOS `open` command, anchors (such as `#building-the-target` in this case) are unfortunately ignored when using this syntax on macOS.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1513,19 +1513,19 @@ async fn doc(
     if path_only {
         let doc_path = toolchain.doc_path(doc_url)?;
         writeln!(cfg.process.stdout().lock(), "{}", doc_path.display())?;
-        Ok(utils::ExitCode(0))
-    } else {
-        if let Some(name) = topic {
-            writeln!(
-                cfg.process.stderr().lock(),
-                "Opening docs named `{name}` in your browser"
-            )?;
-        } else {
-            writeln!(cfg.process.stderr().lock(), "Opening docs in your browser")?;
-        }
-        toolchain.open_docs(doc_url)?;
-        Ok(utils::ExitCode(0))
+        return Ok(utils::ExitCode(0));
     }
+
+    if let Some(name) = topic {
+        writeln!(
+            cfg.process.stderr().lock(),
+            "Opening docs named `{name}` in your browser"
+        )?;
+    } else {
+        writeln!(cfg.process.stderr().lock(), "Opening docs in your browser")?;
+    }
+    toolchain.open_docs(doc_url)?;
+    Ok(utils::ExitCode(0))
 }
 
 #[cfg(not(windows))]
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1531,7 +1531,7 @@ async fn doc(
     } else {
         writeln!(cfg.process.stderr().lock(), "Opening docs in your browser")?;
     }
-    toolchain.open_docs(&doc_path)?;
+    toolchain.open_docs(&doc_path, None)?;
     Ok(utils::ExitCode(0))
 }
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::env::consts::EXE_SUFFIX;
 use std::fmt;
 use std::io::Write;
@@ -1425,7 +1426,7 @@ macro_rules! docs_data {
         }
 
         impl DocPage {
-            fn path(&self) -> Option<&'static str> {
+            fn path_str(&self) -> Option<&'static str> {
                 $( if self.$ident { return Some($path); } )+
                 None
             }
@@ -1463,8 +1464,12 @@ docs_data![
 ];
 
 impl DocPage {
+    fn path(&self) -> Option<&'static Path> {
+        self.path_str().map(Path::new)
+    }
+
     fn name(&self) -> Option<&'static str> {
-        Some(self.path()?.rsplit_once('/')?.0)
+        Some(self.path_str()?.rsplit_once('/')?.0)
     }
 }
 
@@ -1502,18 +1507,18 @@ async fn doc(
         }
     };
 
-    let topical_path: PathBuf;
-
     let doc_url = if let Some(topic) = topic {
-        topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
-        topical_path.to_str().unwrap()
+        Cow::Owned(topical_doc::local_path(
+            &toolchain.doc_path("").unwrap(),
+            topic,
+        )?)
     } else {
         topic = doc_page.name();
-        doc_page.path().unwrap_or("index.html")
+        Cow::Borrowed(doc_page.path().unwrap_or(Path::new("index.html")))
     };
 
     if path_only {
-        let doc_path = toolchain.doc_path(doc_url)?;
+        let doc_path = toolchain.doc_path(&doc_url)?;
         writeln!(cfg.process.stdout().lock(), "{}", doc_path.display())?;
         return Ok(utils::ExitCode(0));
     }
@@ -1526,7 +1531,7 @@ async fn doc(
     } else {
         writeln!(cfg.process.stderr().lock(), "Opening docs in your browser")?;
     }
-    toolchain.open_docs(doc_url)?;
+    toolchain.open_docs(&doc_url)?;
     Ok(utils::ExitCode(0))
 }
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1507,7 +1507,7 @@ async fn doc(
         }
     };
 
-    let doc_url = if let Some(topic) = topic {
+    let doc_path = if let Some(topic) = topic {
         Cow::Owned(topical_doc::local_path(
             &toolchain.doc_path("").unwrap(),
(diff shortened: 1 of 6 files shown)


rami3l:
PS: The CI failures seem irrelevant to the changes being made in this PR.

rami3l:
> PS: The CI failures seem irrelevant to the changes being made in this PR.

This issue has been resolved in https://github.com/rust-lang/rust/pull/133532.

src/cli/topical_doc.rs
```diff
@@ -19,13 +20,10 @@ fn index_html(doc: &DocData<'_>, wpath: &Path) -> Option<PathBuf> {
...
-    let mut v = Vec::new();
-    for entry in entries {
-        let entry = entry?;
-        v.push(entry.file_name());
-    }
-    Ok(v)
+    fs::read_dir(dir)
+        .with_context(|| format!("Failed to read_dir {dir:?}"))?
+        .map(|f| anyhow::Ok(f?.file_name()))
+        .try_collect()
```
djc: Can we just use `Ok` instead of `anyhow::Ok`? I also think we probably don't need `try_collect()` here and can use `collect()` instead.

src/cli/topical_doc.rs
```diff
@@ -118,17 +116,12 @@ pub(crate) fn local_path(root: &Path, topic: &str) -> Result<PathBuf> {
...
+                index_html(&doc, &work_path).context(anyhow!("No document for '{}'", doc.topic))?
+            }
             _ => {
-                let std = PathBuf::from("std");
-                let search_keywords = match forced_keyword {
-                    Some(k) => k,
-                    None => keywords_top,
-                };
-                search_path(&doc, &std, &search_keywords)?
+                let search_keywords = forced_keyword.unwrap_or(keywords_top);
```
djc: Nit: maybe inline the `search_keywords` binding?

src/cli/rustup_mode.rs
```diff
@@ -1502,18 +1507,15 @@ async fn doc(
         }
     };
 
-    let topical_path: PathBuf;
-
-    let doc_url = if let Some(topic) = topic {
-        topical_path = topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?;
-        topical_path.to_str().unwrap()
+    let doc_url: Cow<'_, Path> = if let Some(topic) = topic {
```
djc: I'd probably prefer `Cow::Borrowed`/`Cow::Owned` in the arms instead of the type annotation with `.into()` calls.

src/toolchain.rs
```diff
@@ -408,16 +408,21 @@ impl<'a> Toolchain<'a> {
         buf
     }
 
-    pub fn doc_path(&self, relative: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
+    pub fn doc_path(&self, maybe_relative: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
```
djc: At this point, not sure the `maybe_relative` name is useful over something simpler like just `path`?

src/cli/rustup_mode.rs
```diff
@@ -1471,6 +1471,32 @@ impl DocPage {
     fn name(&self) -> Option<&'static str> {
         Some(self.path_str()?.rsplit_once('/')?.0)
     }
+
+    fn resolve<'t>(&self, root: &Path, topic: &'t str) -> Option<(PathBuf, Option<&'t str>)> {
+        // Save the components in case the last one is used later with `parent_html`.
+        let components = topic.split("::").collect::<Vec<_>>();
```
djc: Maybe review how it looks if you don't do this, but instead do `base.extend(topic.split("::"))` (with any necessary variations to make that work) and then `topic.rsplit_once("::")` for the `parent_html` case?

src/cli/rustup_mode.rs
```diff
@@ -1507,11 +1533,22 @@ async fn doc(
         }
     };
 
-    let doc_path: Cow<'_, Path> = if let Some(topic) = topic {
-        topical_doc::local_path(&toolchain.doc_path("").unwrap(), topic)?.into()
-    } else {
-        topic = doc_page.name();
-        doc_page.path().unwrap_or(Path::new("index.html")).into()
+    let (doc_path, fragment): (Cow<'_, Path>, _) = match (topic, doc_page.name()) {
```
djc: Same here with the type annotations/`into()`.

tests/suite/cli_rustup.rs
```diff
@@ -2640,16 +2641,20 @@ async fn docs_topical_with_path() {
         .expect_ok(&["rustup", "toolchain", "install", "nightly"])
         .await;
 
-    for (topic, path) in mock::topical_doc_data::test_cases() {
-        let mut cmd = clitools::cmd(&cx.config, "rustup", ["doc", "--path", topic]);
+    for (args, path) in mock::topical_doc_data::test_cases() {
+        let mut cmd = clitools::cmd(
+            &cx.config,
+            "rustup",
+            chain!(["doc", "--path"], args.iter().cloned()),
```
djc: Maybe something like `["doc", "--path"].into_iter().chain(args.iter())`?

djc (approved):
LGTM! (I assume you'll rebase this to ditch the merge commit?)

src/toolchain.rs
rami3l: Personally, if I were to read `fn doc_path(&self, path: impl AsRef<Path>)`, I would not have known what to pass as `path`. `maybe_relative` is definitely a long name and I kind of dislike it. However I don't think `path` is a better option.

rami3l: I think it's better to just call it `relative`, since implicitly we still have absolute + absolute = absolute, which is also the convention for e.g. `std`.

ChrisDenton: On that note, could we just store `self.html_root = doc_dir.extend(["share", "doc", "rust", "html"]);` or so then `doc_path` is just `self.html_root.join(relative)`.

rami3l: @djc Hmmm maybe you have marked this conversation as resolved by mistake? Anyway, I think what @ChrisDenton has pointed out is a pre-existing issue, and looking at the current source I'm not sure what the best thing to do currently is.

We could cache the extended base path somewhere, but there's nothing stopping us from changing the `Toolchain` instance mid-way so the cache should be invalidated...

rami3l:
@djc Do you think we can merge this change now, so that it can join the others in the upcoming beta? 🙏

djc:
I already approved it, so feel free to merge unless there's something in particular you want me to look at.

rami3l:
> I already approved it, so feel free to merge unless there's something in particular you want me to look at.

@djc Thanks! There's probably https://github.com/rust-lang/rustup/pull/4070#discussion_r1876050391 although I don't think it should be blocking it being merged. Let's do the merge first and see.

PR #4078, opened 2024-11-15
feat: add nushell support

### LINKED ISSUE #3472: [Feature] add to Nushell's PATH as well
I believe Nushell (made in the very Rust) deserve this consideration, as Bash does already.

LittleJianCH:
Closes https://github.com/rust-lang/rustup/issues/3472.

I tried to add nushell support referencing #3108.

But it's hard to write tests for it, so I could only manually use it a few times for testing.

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -77,6 +77,7 @@ fn enumerate_shells() -> Vec<Shell> {
         Box::new(Bash),
         Box::new(Zsh),
         Box::new(Fish),
+        Box::new(Nu),
     ]
 }
 
@@ -252,6 +253,53 @@ impl UnixShell for Fish {
 
     fn source_string(&self, process: &Process) -> Result<String> {
         Ok(format!(r#"source "{}/env.fish""#, cargo_home_str(process)?))
+    }
+}
+
+struct Nu;
+
+impl UnixShell for Nu {
+    fn does_exist(&self, process: &Process) -> bool {
+        // nu has to either be the shell or be callable for nu setup.
+        matches!(process.var("SHELL"), Ok(sh) if sh.contains("nu"))
+            || utils::find_cmd(&["nu"], process).is_some()
+    }
+
+    fn rcfiles(&self, process: &Process) -> Vec<PathBuf> {
+        let mut paths = vec![];
+
+        if let Ok(p) = process.var("XDG_CONFIG_HOME") {
+            let path = PathBuf::from(p).join("nushell/");
+            paths.push(path.join("env.nu"));
+            paths.push(path.join("config.nu"));
+        }
+
+        if let Some(p) = process.home_dir() {
+            let path = p.join(".config/nushell/");
+            paths.push(path.join("env.nu"));
+            paths.push(path.join("config.nu"));
+        }
+        paths
+    }
+
+    fn update_rcs(&self, process: &Process) -> Vec<PathBuf> {
+        let mut rcs = self.rcfiles(process);
+        if rcs.len() == 4 {
+            // The first two rcfile takes precedence (XDG_CONFIG_HOME).
+            rcs.truncate(2);
+        }
+        rcs
+    }
+
+    fn env_script(&self) -> ShellScript {
+        ShellScript {
+            name: "env.nu",
+            content: include_str!("env.nu"),
+        }
+    }
+
+    fn source_string(&self, process: &Process) -> Result<String> {
+        Ok(format!(r#"source "{}/env.nu""#, cargo_home_str(process)?))
     }
 }
 
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -384,6 +384,7 @@ the corresponding `env` file under {cargo_home}.
 This is usually done by running one of the following (note the leading DOT):
     . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
     source "{cargo_home}/env.fish"  # For fish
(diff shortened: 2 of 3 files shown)


src/cli/self_update/shell.rs
```diff
@@ -255,6 +256,50 @@ impl UnixShell for Fish {
...
+            .and(process.home_dir())
+            .iter()
+            .flat_map(|path| {
+                let mut p0 = path.clone();
+                let mut p1 = path.clone();
+                p0.push("nushell/env.nu");
+                p1.push("nushell/config.nu");
+                [p0, p1]
+            })
+            .collect()
```
djc: This doesn't seem like a clear formulation of what the logic is supposed to be. Maybe something like this?

```rust
let mut paths = Vec::new();
if let Ok(config) = process.var("XDG_CONFIG_HOME") {
    let mut path = PathBuf::from(config);
    path.push("nushell/env.nu");
    paths.push(path);
}

if let Some(home) = process.home_dir() {
    home.push("nushell/config.nu");
    paths.push(path);
}

paths
```

LittleJianCH: Nushell will try to look for `XDG_CONFIG_HOME` firstly. If it doesn't existed, Nushell will go to `$HOME/.config` instead. Then `env.nu` and `config.nu` are in the same config dir. 

So the logic should be
```rust
let config_dir = process
            .var("XDG_CONFIG_HOME")
            .ok()
            .and(process.home_dir().map(|dir| dir.join("/.config")));

config_dir.map_or(vec![], |dir| {
  vec![dir.join("env.nu"), dir.join("config.nu")]
})
```

I will fix it soon. BTW, should I just add a new commit in this branch?

djc: That code checks if the `XDG_CONFIG_HOME` variable is defined but not whether the directory exists, is that what you meant? (I still think this code sample isn't very easy to follow due to overuse of combinators.)

Ideally please squash additional changes into your existing commit.

LittleJianCH: Yes. Should I check the case that `XDG_CONFIG_HOME` variable is defined but the directory doesn't exist, since the `XDG_CONFIG_HOME` is not only for Nushell?

And I will rewrite the code using less combinators to make it easier to read.

rami3l: > Nushell will try to look for `XDG_CONFIG_HOME` firstly. If it doesn't existed, Nushell will go to `$HOME/.config` instead. Then `env.nu` and `config.nu` are in the same config dir.

@LittleJianCH That is to say, I believe, that the resolution logic is the same logic as `fish`. So this might be useful to you:

https://github.com/rust-lang/rustup/blob/c8eddcb4c1dc9aea8067fa73179ce1791aa24d1d/src/cli/self_update/shell.rs#L221-L236

src/cli/self_update/shell.rs
```diff
@@ -92,7 +93,7 @@ pub(crate) trait UnixShell {
     fn does_exist(&self, process: &Process) -> bool;
 
     // Gives all rcfiles of a given shell that Rustup is concerned with.
-    // Used primarily in checking rcfiles for cleanup.
+    // Used primarily in checking rcfiles c
```
rami3l: Might be a typo?

src/cli/self_update/shell.rs
```diff
@@ -255,6 +256,60 @@ impl UnixShell for Fish {
...
+            })
+            .collect();
+
+        [p0, p1].concat()
+    }
+
+    fn update_rcs(&self, process: &Process) -> Vec<PathBuf> {
+        let rcs = self.rcfiles(process);
+        if rcs.len() == 4 {
+            // The first two rcfile takes precedence(XDG_CONFIG_HOME).
```
rami3l:
```suggestion
            // The first two rcfile takes precedence (XDG_CONFIG_HOME).
```

src/cli/self_update/shell.rs
```diff
@@ -255,6 +256,60 @@ impl UnixShell for Fish {
...
+            rcs[0..2].to_vec()
```
rami3l: You can use [`vec.resize(2)`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize) to save an allocation :)

LittleJianCH: I'll fix them soon.

src/cli/self_update/shell.rs
```diff
@@ -255,6 +256,61 @@ impl UnixShell for Fish {
...
+        let p1: Vec<_> = process
+            .home_dir()
+            .iter()
+            .flat_map(|p| {
+                let path = PathBuf::from(p).join(".config/nushell/");
+                [path.join("env.nu"), path.join("config.nu")]
+            })
+            .collect();
+
+        [p0, p1].concat()
```
rami3l: Sorry for the back and forth, but one more thing: the `.flat_map()` thing still looks a bit too clever.

Simple procedural code will read even better:

```rs
let mut paths = vec![];

if let Ok(path) = process.var("XDG_CONFIG_HOME") {
  let path = PathBuf::from(path).join("nushell/");
  paths.push(path.join("env.nu"));
  paths.push(path.join("config.nu"));
}

if let Some(path) = process.home_dir() {
  let path = PathBuf::from(path).join("nushell/");
  paths.push(path.join("env.nu"));
  paths.push(path.join("config.nu"));
}

paths

rami3l (approved):
Many thanks :D

djc (approved):

PR #4080, opened 2024-11-16
feat: align website style with www.rust-lang.org for consistency

### LINKED ISSUE #3191: Rustup.rs needs some rust theming for credibility
I hate to open a discussion that I know has the potential to turn into too much bikeshedding, but I think that there could be some improvements done here. Current site:

image

Barebones is great and not much is needed for the purpose of this site. However: there's no logo, no theme, no links to a "getting started" or usage guide, and imho looks a bit like the result of a modern website when you disable javascript. There just isn't anything that visually suggests "this is an official rust project", to match the text that literally says it.

I am not suggesting large changes, just enough to make it a bit more coherent with other official Rust websites.

To do so, I think rustup.rs could generally match rustdoc, mdbook, rust-lang.org, or foundation.rust-lang.org themes (why do they all have completely different themes...). I think rust-lang.org is the best bet here: https://www.rust-lang.org/tools/install as a basis. From there:

- Keep the fonts
- Slightly shrink the 8rem header
- (maybe) make the install box a large green div to parallel this page
- Keep the logo in the top left
- Add a link to the "getting started" guide so people have somewhere to go after installing

And that's it. No more questioning whether you're on the right website when you visit.

(I don't mind heading something up here if there's no strong opposition)

@rustbot label +website +enhancement

0xPoe:
👋 Long time no see!

close https://github.com/rust-lang/rustup/issues/3191

image

I tried to match the colors and fonts from www.rust-lang.org and make the design more consistent and modern.

1. Use the font from www.rust-lang.org to render the title.  
2. Change the box background color to white for a cleaner and more modern look.  
3. Ensure the website is more responsive and resizable across various screen sizes.  
4. Update the link color from `#428bca` to `#0b7261`, matching the color scheme of www.rust-lang.org.  

You can use a server to host the www dir to check it. For example:
```shell
npm install http-server -g

cd www

http-server    
```

If you have any ideas about making it more beautiful, please let me know and feel free to comment.


0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback.)

rami3l (commented):
Thanks a bunch for making this!

Sorry for my not being a webdev expert, but TBH the deep-colored box looks weird to me. https://www.rust-lang.org uses full-width stripes instead of boxes, so I imagine something like:

image

In terms of the use of fonts and colors, I have a more radical proposal: maybe consider borrowing the style of the first and the last stripe?

Modding the https://www.rust-lang.org page for illustration purposes below (the "sponsor" section should be replaced by the command with a copy button of `#ffc832`):

image

image

0xPoe:
https://github.com/user-attachments/assets/2244ab9e-9e95-4b81-9a77-0aba7f891c16





> but TBH the deep-colored box looks weird to me.

I totally agree. It looks a little bit weird.

I have updated it.

1. I will try to use the layout of www.rust-lang.org.  
2. I kept the box layout because we don't have enough text content; using full width would look too empty.  
3. I also tried adapting the same navigator from www.rust-lang.org, but it doesn’t work well with our content. So, I prefer not to include it.

I also formatted the index HTML in a separate commit and added the missing closing body tag. Let me know if you'd prefer not to format it until we establish a standard formatter. I used the default HTML formatter in VS Code.

0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback.)

0xPoe:
The font was copied from the [www.rust-lang.org](http://www.rust-lang.org/).

0xPoe:
I also tested it on screens of various sizes:

https://github.com/user-attachments/assets/857bd7ef-fcea-4a9e-8986-755164aacadb



image

0xPoe:
Also on my cell phone:
image

rami3l (approved):
> 1. I will try to use the layout of [www.rust-lang.org](http://www.rust-lang.org).
> 2. I kept the box layout because we don't have enough text content; using full width would look too empty.
> 3. I also tried adapting the same navigator from [www.rust-lang.org](http://www.rust-lang.org), but it doesn’t work well with our content. So, I prefer not to include it.

@Rustin170506 Thanks! That looks much better indeed, but still, I think the "Run the following" line should be joining the "Get Started" one, that is, both left-aligned, either outside or inside the box (I guess outside would look more natural).

> I also formatted the index HTML in a separate commit and added the missing closing body tag. Let me know if you'd prefer not to format it until we establish a standard formatter. I used the default HTML formatter in VS Code.

I have no problem with that, as long as the formatting is made in a single commit for maximal clarity.

cc @djc @rbtcollins

0xPoe:
> think the "Run the following" line should be joining the "Get Started" one, that is, both left-aligned, either outside or inside the box (I guess outside would look more natural).

The reason I didn't do that is that when you click `display all...`, the content changes, and more items are added to the box. So it doesn't make sense to link only the first one to `Get Started`.

0xPoe:
@rami3l What do you think if we remove that `get started`? It seems a little bit weird.

image

It seems removing it would make it simpler and clearer.

rami3l:
@Rustin170506 Hmmm honestly speaking I'm not sure... Let me take some time to think about it.

0xPoe:
> @Rustin170506 Hmmm honestly speaking I'm not sure... Let me take some time to think about it.

No rush—please take your time. I believe we should consider removing it because the main content is concentrated in the middle of the page. The short `Get Started` section feels out of place within the overall layout. Its brevity makes the page look unbalanced and less cohesive.

The main reason I didn’t follow the two-column layout from [www.rust-lang.org](http://www.rust-lang.org/) is that we don’t have enough content to fully utilize two columns. Keeping the content centered makes the layout cleaner and more visually balanced.

djc:
IMO this PR is a little tricky because it contains a whole bunch of changes at the same time. Although I understand that design maybe a little different in that it might be harder to get to a good (consistent) design by smaller pieces, I think in this context it would be helpful to split the changes here into a series of smaller ones that can be judged separately.

For example, I do really like the idea of using the font and accent color from the Rust website here, but we can make smaller/targeted changes for that.

There also seem to be a whole bunch of non-functional formatting changes here which IMO it would be good to isolate separately and decide how we want to handle CSS formatting (ideally with automated tools) -- personally for example the change from ` > ` to `>` in selectors does not look like an improvement to me.

For other structural changes to the HTML, I'd also like to see those isolated so we can discuss their merits separately.

0xPoe:
> There also seem to be a whole bunch of non-functional formatting changes here which IMO it would be good to isolate separately and decide how we want to handle CSS formatting (ideally with automated tools) -- personally for example the change from `>` to `>` in selectors does not look like an improvement to me.

I can separate these changes to a separate commit. Or we can chose to do not include any format changes at all. I can do it in the following PRs.

djc:
Separate commits are helpful but, to the extent it doesn't cause outsized rebase hazards I think I would prefer separate PRs here so we can merge changes independently of each other.

0xPoe:
> There also seem to be a whole bunch of non-functional formatting changes here which IMO it would be good to isolate separately and decide how we want to handle CSS formatting (ideally with automated tools) -- personally for example the change from `>` to `>` in selectors does not look like an improvement to me.

I have created an issue for it. https://github.com/rust-lang/rustup/issues/4091 If you have more thought on it, we can discuss it in the issue.

I will cleanup this PR later.

0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback.)

0xPoe:
@rami3l @djc I believe this PR is ready for another review. If you'd prefer to split it further, I think the only thing we could extract is the color styling for the `<a>` tag. The rest pertains to the overall look of the website, and I don't think it makes sense to separate those into different PRs.

www/index.html
```diff
@@ -19,12 +19,13 @@
   click or press "n" to cycle platforms
 </a>
 
-<p id="pitch">
-  <em>rustup</em> is an installer for<br/>
-  the systems programming language
-  <a href="https://www.rust-lang.org">Rust</a>
-</p>
-
+<main>
```
djc, line 18: Why introduce a `main` element here? Why a `header`?

www/index.html
```diff
@@ -223,8 +224,9 @@
   </div>
 </div>
 
-<p>
-  Need help?<br>Ask on <a href="https://discord.gg/rust-lang">#beginners in the Rust Discord</a><br>
+<p id="help">
```
djc, line 223: With the larger changes, I'd like to be able to see a single commit that has the `id="help"` added here and also the CSS changes you applied here.

www/index.html
```diff
@@ -223,8 +224,9 @@
...
+  Need help?<br>
+  Ask on <a href="https://discord.gg/rust-lang">#beginners in the Rust Discord</a><br>
```
djc: Breaking the second sentence into a separate line is a trivial formatting change, and I'd like to isolate all of those in a separate commit.

www/rustup.css
```diff
@@ -31,49 +31,69 @@
     src: local('Work Sans Medium'), url("fonts/WorkSans-Medium.ttf") format('truetype');
 }
 
+@font-face {
```
djc, line 34: I think the font changes should constitute a separate commit.

www/rustup.css
```diff
@@ -31,49 +31,69 @@
...
+        url("fonts/AlfaSlabOne-Regular.woff2") format('woff2'),
+        url("fonts/AlfaSlabOne-Regular.woff") format('woff');
+}
+
 body {
-    margin-top: 2em;
     background-color: white;
-    color: #515151;
-    font-family: "Fira Sans","Helvetica Neue",Helvetica,Arial,sans-serif;
+    font-family: "Fira Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
```
djc: Trivial formatting changes.

www/rustup.css
```diff
@@ -31,49 +31,69 @@
...
     font-weight: 300;
     font-size: 25px;
+    margin: 0;
```
djc: Why?

www/rustup.css
```diff
@@ -31,49 +31,69 @@
...
+header > div > h2 {
+    color: black;
+    font-size: 2rem;
+    font-weight: 300;
 }
 
 a {
-    color: #428bca;
-    text-decoration: none;
+    color: #0b7261;
```
djc, line 81: Why change the link color/decoration?

www/rustup.css
```diff
@@ -85,7 +105,7 @@ body#idx p.other-platforms-help {
 }
 
 .instructions {
-    background-color: rgb(250, 250, 250);
+    color: black;
```
djc: Why?

djc (commented):
> The rest pertains to the overall look of the website, and I don't think it makes sense to separate those into different PRs.

I don't think I agree, I'd like to better understand/discuss the motivations for a bunch of the minor changes you've made here.

www/rustup.css
0xPoe: > Change the box background color to white for a cleaner and more modern look.

www/rustup.css
0xPoe: > Update the link color from #428bca to #0b7261, matching the color scheme of [www.rust-lang.org](http://www.rust-lang.org/).

djc: Okay, so I think it makes sense to have at least one commit per bullet point, more strongly linking the commit message to the changes at a more granular level.

www/index.html
0xPoe: I followed the HTML structure from [www.rust-lang.org](http://www.rust-lang.org/). It places the large `Rust` in the header inside a `main` element. The `main` element is introduced here to ensure it only occupies 80% of the body’s width.

www/index.html
0xPoe: Ok. I will do it.

www/rustup.css
0xPoe: Done.

www/rustup.css
0xPoe: Done

www/rustup.css
0xPoe: Removed. Good catch!

www/rustup.css
0xPoe: Rverted.

www/index.html
0xPoe: Reverted.

0xPoe:
@djc I’ve split each change into a single commit. Apologies for any confusion caused by unclear changes.

www/index.html
```diff
@@ -128,7 +129,6 @@
   </div>
 
   <!-- duplicate the default cross-platform instructions -->
-  <div>
```
0xPoe, line 131: The reason we need to remove it is to ensure the content in the instructions is centered. Compared to the default `platform-instructions-unix`, we have this extra div here. Removing it will help ensure the style is applied correctly.

www/index.html
```diff
@@ -176,7 +175,6 @@
 </div>
 
 <div id="platform-instructions-default" class="instructions">
-  <div>
```
0xPoe, line 179: ditto

www/rustup.css
```diff
@@ -45,34 +54,45 @@ pre {
...
-    text-decoration: none;
+header > div > h1 {
+    font-family: "Alfa Slab One", serif;
+    color: black;
+    font-size: 4rem;
+    margin-bottom: 0;
+    margin-top: 1rem;
+    line-height: 1;
+    font-weight: 300;
+    letter-spacing: 1px;
```
0xPoe, line 71: These(most of them) are copied from the Rust website.

www/rustup.css
```diff
@@ -45,34 +54,45 @@ pre {
...
+a {
+    color: #0b7261;
+    text-decoration: underline;
 }
 
-body#idx > #pitch {
-    width: 35rem;
+a:hover {
+    color: #0d8b75;
+    text-decoration: underline;
```
0xPoe, line 87: This also is the style(underline) from rust website

www/rustup.css
```diff
@@ -95,12 +114,12 @@ body#idx p.other-platforms-help {
 }
 
 .instructions > * {
-    width: 45rem;
```
0xPoe, line 98: We no longer need the fixed width. We’re using percentages for better responsiveness.

www/rustup.css
```diff
@@ -111,29 +130,28 @@ hr {
...
 
 .rustup-command::before {
-    color: #999;
+    color: black;
     content: " $ ";
 	margin-left: 15px;
 }
 
 .rustup-command {
-    background-color: #515151;
```
0xPoe, line 124: The white box looks more modern to me.

www/rustup.css
```diff
@@ -144,6 +162,8 @@ hr {
 #platform-instructions-unknown div.copy-container {
     display: flex;
     align-items: center;
+    width: 90%;
```
0xPoe, line 165: Make sure it can be responsive.

0xPoe (commented):
@djc Here are the reasons behind the decisions I made. I hope this helps you with the review.

djc:
@Rustin170506 I was going to review the current state of the PR this morning, are you sure you want to retract it? Seems like a waste to me!

0xPoe:
> I was going to review the current state of the PR this morning, are you sure you want to retract it? Seems like a waste to me!

Thank you for your time! I'm a bit hesitant about this change. As the issue author mentioned, it's hard to define what makes a "good-looking website." So I'm a bit worried that this change (from a UI design perspective) might be seen by others not as an improvement but as a step backward.

0xPoe:
Maybe we should wait until a designer (or someone with more web design experience) is willing to provide a completely new design for the website before making adjustments to the implementation.

djc:
I think any changes that move the rustup website closer to the main Rust website in fairly obvious ways would be a clear improvement, so I still think there's value in incremental changes like this.

0xPoe:
> I think any changes that move the rustup website closer to the main Rust website in fairly obvious ways would be a clear improvement, so I still think there's value in incremental changes like this.

Cool! I've reopened it. If there's anything I can do to help review the changes, please let me know!

www/rustup.css
```diff
@@ -31,6 +31,15 @@
     src: local('Work Sans Medium'), url("fonts/WorkSans-Medium.ttf") format('truetype');
 }
 
+@font-face {
+    font-family: 'Alfa Slab One';
```
djc, line 35: Can you squash this change into the previous commit? Otherwise the previous commit is left incomplete, which seems unfortunate.

www/rustup.css
djc: I don't know, none of the color changes in this commit seem particularly well-motivated to me. Are they related/helpful/necessary with the other changes in this PR?

djc (approved):

www/rustup.css
0xPoe: It is just my taste. I think the white box looks more modern to me.
image
vs
image

0xPoe: I am OK to drop this change. Please let me know if you want to keep the original style.

www/rustup.css
0xPoe: Squashed. Thank you!

rami3l:
@Rustin170506 Don't worry about the CI failure; I'm working on this (https://github.com/rust-lang/rustup/pull/4098#issuecomment-2503418305).

www/rustup.css
djc: Okay, let's keep it.

rami3l (approved):
This looks pretty nice! Thanks a lot for your help :)

rami3l:
@djc I guess it's not against the rules to merge this immediately so that it can get to the beta as well (after #4098, of course)?

0xPoe:

PR #4093, opened 2024-11-20
Add aliases for remove/uninstall/unset commands

joshtriplett:
After one too many times of running `rustup toolchain rm` or `rustup
toolchain delete` and being surprised that it doesn't work, add some
aliases so that it does.


djc (approved):
Makes sense to me!

rami3l (approved):
Thanks a lot :)

PR #4098, opened 2024-11-25
Prepare 1.28.0: step 2

djc:
Step 2 in our [release process](https://rust-lang.github.io/rustup/dev-guide/release-process.html#making-a-release):

- [x] Update the commit shasum in `rustup-init.sh` to match the latest commit on `master`.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -33,7 +33,7 @@ RUSTUP_QUIET=no
 # NOTICE: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat <<EOF
-rustup-init 1.27.1 (a8e4f5c64 2024-04-24)
+rustup-init 1.28.0 (b6fdf2fa8 2024-11-25)
 
 The installer for rustup
 
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -154,8 +154,8 @@ main() {
     ensure downloader "$_url" "$_file" "$_arch"
     ensure chmod u+x "$_file"
     if [ ! -x "$_file" ]; then
-        err "Cannot execute $_file (likely because of mounting /tmp as noexec)." 
-        err "Please copy the file to a location where you can execute binaries and run ./rustup-init${_ext}." 
+        err "Cannot execute $_file (likely because of mounting /tmp as noexec)."
+        err "Please copy the file to a location where you can execute binaries and run ./rustup-init${_ext}."
         exit 1
     fi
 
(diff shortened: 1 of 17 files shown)


rustup-init.sh
```diff
@@ -33,7 +33,7 @@ RUSTUP_QUIET=no
 # NOTICE: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat <<EOF
-rustup-init 1.27.1 (a8e4f5c64 2024-04-24)
+rustup-init 1.28.0 (b6fdf2fa 2024-11-25)
```
ChrisDenton: Would it not be better that the release commit include the full changelog? Even though it has no effect on the binary itself.

ChrisDenton (approved):
Looks good to me. I have one query but I don't want to block on it.

rustup-init.sh
rami3l: @chrisdenton Ideally we'd be able to do that, but not just yet. The plan is that this will be a placeholder for the release team to fill in. For now we just want the traceability (and since we're preparing for the beta, even this can be broken as long as it gets fixed in the final release).

rami3l (approved):
@djc Thanks a lot!

BTW are you planning to cut a beta release this week?

djc:
> @djc Thanks a lot!
> 
> BTW are you planning to cut a beta release this week?

I plan to keep working towards it, don't have any kind of deadline in mind.

rustup-init.sh
djc: Also given that this is just the beta release, I don't think it is as important for right now.

rami3l: @djc Sorry for the interruption: you're breaking the output format by truncating the hash by one more digit :o

rami3l (approved):

rami3l:
@djc These CI errors have also influenced https://github.com/rust-lang/rustup/pull/4070#issuecomment-2500886187.

```console
#5 0.935 E: The repository 'http://archive.ubuntu.com/ubuntu lunar-updates Release' no longer has a Release file.
```

`lunar` means Ubuntu 23.04 which have, I believe, already been EOL for quite a while now. Not sure why our [upstream](https://github.com/rust-lang/rust/blob/39cb3386ddc6c71657418be28dbb3987eea4aa4b/src/ci/docker/host-x86_64/arm-android/Dockerfile#L1) still hasn't updated it...

rami3l:
> @djc These CI errors have also influenced [#4070 (comment)](https://github.com/rust-lang/rustup/pull/4070#issuecomment-2500886187).
> 
> ```
> #5 0.935 E: The repository 'http://archive.ubuntu.com/ubuntu lunar-updates Release' no longer has a Release file.
> ```
> 
> `lunar` means Ubuntu 23.04 which have, I believe, already been EOL for quite a while now. Not sure why our [upstream](https://github.com/rust-lang/rust/blob/39cb3386ddc6c71657418be28dbb3987eea4aa4b/src/ci/docker/host-x86_64/arm-android/Dockerfile#L1) still hasn't updated it...

It took quite a while, but my https://github.com/rust-lang/rust/pull/133532 has finally been merged! I guess we can restart the release process now.

rami3l:
Oh, it's quite unfortunate since we're getting a Rust version bump in the middle of this PR, so we're introducing unresolved lints to the codebase...

PR #4105, opened 2024-12-04
ci(stable): fix build issues to prepare for v1.28.0 release

rami3l:
Before merging, we should also:

- [x] Wait for https://github.com/aws/aws-lc-rs/issues/620 to land in a new release.
- [x] Drop the `DROP ME` commit.

---

FWIW, because of `rustls`'s migration to `aws-lc-rs`, with this PR:
- The `reqwest/rustls/aws-lc-rs` combo has become the default download backend on Linux for `riscv*`, `s390x*` and `ppc32` platforms.
- `netbsd` and `illumos` builds have switched to `reqwest/rustls/native-tls` by default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,9 +170,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.11.1"
+version = "1.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f47bb8cc16b669d267eeccf585aea077d0882f4777b1c1f740217885d6e6e5a3"
+checksum = "f409eb70b561706bf8abba8ca9c112729c481595893fd06a2dd9af8ed8441148"
 dependencies = [
  "aws-lc-sys",
  "paste",
@@ -181,9 +181,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.23.1"
+version = "0.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2101df3813227bbaaaa0b04cd61c534c7954b22bd68d399b440be937dc63ff7"
+checksum = "8478a5c29ead3f3be14aff8a202ad965cf7da6856860041bfca271becf8ba48b"
 dependencies = [
  "bindgen",
  "cc",
--- a/ci/docker/x86_64-unknown-freebsd/Dockerfile
+++ b/ci/docker/x86_64-unknown-freebsd/Dockerfile
@@ -1,4 +1,8 @@
 FROM rust-x86_64-unknown-freebsd
 
+# Building `aws-lc-rs` for FreeBSD on Linux depends on `gcc-multilib`, `libclang` and `bindgen`.
+# See: https://aws.github.io/aws-lc-rs/requirements/linux
+RUN apt-get update && apt-get install -qy gcc-multilib libclang-dev
+
 ENV CC_x86_64_unknown_freebsd=x86_64-unknown-freebsd12-clang \
     CARGO_TARGET_X86_64_UNKNOWN_FREEBSD_LINKER=x86_64-unknown-freebsd12-clang
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -7,6 +7,18 @@ export RUST_BACKTRACE=1
 rustc -vV
 cargo -vV
 
+if [ -n "$INSTALL_BINDGEN" ]; then
+  if ! curl --proto '=https' --tlsv1.2 -LsSf https://github.com/rust-lang/rust-bindgen/releases/latest/download/bindgen-cli-installer.sh | sh -s -- --no-modify-path \
+    | grep "everything's installed!";
+    # Ignoring exit code since the script might fail to write the receipt after a successful installation.
+  then
+    cargo install --force --locked bindgen-cli
+  fi
+  mkdir "$CARGO_HOME"/bin/bindgen-cli
+  mv "$CARGO_HOME"/bin/bindgen "$CARGO_HOME"/bin/bindgen-cli/
+  export PATH="$CARGO_HOME/bin/bindgen-cli:$PATH"
+fi
+
 
 FEATURES=('--no-default-features' '--features' 'curl-backend,reqwest-native-tls')
 case "$(uname -s)" in
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -15,11 +15,12 @@ case "$(uname -s)" in
 esac
 
 case "$TARGET" in
-  # these platforms aren't supported by ring:
-  powerpc* ) ;;
+  # these platforms aren't supported by aws-lc-rs:
+  powerpc64* ) ;;
   mips* ) ;;
-  riscv* ) ;;
-  s390x* ) ;;
+  loongarch* ) ;;
+  *netbsd* ) ;;
+  *illumos* ) ;;
   # default case, build with rustls enabled
   * ) FEATURES+=('--features' 'reqwest-rustls-tls') ;;
 esac
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -874,6 +874,7 @@ jobs:
           - x86_64-linux-android           # skip-pr skip-master
           - riscv64gc-unknown-linux-gnu    # skip-pr skip-master
           - loongarch64-unknown-linux-gnu  # skip-pr skip-master
+          - loongarch64-unknown-linux-musl # skip-pr skip-master
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -621,6 +621,7 @@ jobs:
             --env LIBZ_SYS_STATIC=1 \
             --env SKIP_TESTS="${SKIP_TESTS}" \
             --env TARGET="${TARGET}" \
+            --env INSTALL_BINDGEN=1 \
             --init \
             --rm \
             --tty \
@@ -785,6 +786,7 @@ jobs:
             --env LIBZ_SYS_STATIC=1 \
             --env SKIP_TESTS="${SKIP_TESTS}" \
             --env TARGET="${TARGET}" \
+            --env INSTALL_BINDGEN=1 \
             --init \
             --rm \
             --tty \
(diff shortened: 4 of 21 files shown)


heiher:
We need to temporarily disable the `rustls` feature for `loongarch64-linux-{gnu,musl}` until the `libclang-dev` adds support the LoongArch target.

https://github.com/rust-lang/rustup/blob/14470b2b14c10e1578d0d343c4a4f8445cd96a0f/ci/run.bash#L17-L25
https://github.com/rust-lang/rustup/blob/14470b2b14c10e1578d0d343c4a4f8445cd96a0f/ci/run.bash#L42-L50

rami3l:
> We need to temporarily disable the `rustls` feature for `loongarch64-linux-{gnu,musl}` until the `libclang-dev` adds support the LoongArch target.

@heiher Thanks for the heads up! I guess removing `reqwest-rustls-tls` would be enough as a last resort (https://github.com/rust-lang/rustup/issues/4094#issuecomment-2493605991). In the meantime, I'll try to configure `bindgen-cli` for the rest of the failures.

heiher:
I tested the [`rustup-init-loongarch64-unknown-linux-gnu`](https://github.com/rust-lang/rustup/actions/runs/12162420263), and the following error was thrown when downloading:

```
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/home/hev/.rustup/tmp/nf3c6b3n2mqj5w73_file': failed to make network request: download backend 'reqwest rustls' unavailable
```

rami3l:
> I tested the [`rustup-init-loongarch64-unknown-linux-gnu`](https://github.com/rust-lang/rustup/actions/runs/12162420263), and the following error was thrown when downloading:
> 
> ```
> error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/home/hev/.rustup/tmp/nf3c6b3n2mqj5w73_file': failed to make network request: download backend 'reqwest rustls' unavailable
> ```

@heiher Despite being not idiomatic, that's the expected behavior currently, as `reqwest-rustls-tls` is the new default now, so you'll need to export `RUSTUP_USE_RUSTLS=0`:

https://github.com/rust-lang/rustup/blob/14470b2b14c10e1578d0d343c4a4f8445cd96a0f/src/utils/mod.rs#L251-L253

... it'd definitely be nice to add `cfg!()` to the predicate before actually shipping this build, I believe.

heiher:
> > I tested the [`rustup-init-loongarch64-unknown-linux-gnu`](https://github.com/rust-lang/rustup/actions/runs/12162420263), and the following error was thrown when downloading:
> > ```
> > error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/home/hev/.rustup/tmp/nf3c6b3n2mqj5w73_file': failed to make network request: download backend 'reqwest rustls' unavailable
> > ```
> 
> @heiher Despite being not idiomatic, that's the expected behavior currently, as `reqwest-rustls-tls` is the new default now, so you'll need to export `RUSTUP_USE_RUSTLS=0`:
> 
> https://github.com/rust-lang/rustup/blob/14470b2b14c10e1578d0d343c4a4f8445cd96a0f/src/utils/mod.rs#L251-L253
> 
> ... it'd definitely be nice to add `cfg!()` to the predicate before actually shipping this build, I believe.

@rami3l Thank you. Could you include this patch together as well?

```diff
diff --git a/src/utils/mod.rs b/src/utils/mod.rs
index dae45d4b..d9a8ee51 100644
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -250,7 +250,8 @@ async fn download_file_(
         .is_some_and(|it| it != "0");
     let use_rustls = process
         .var_os("RUSTUP_USE_RUSTLS")
-        .is_none_or(|it| it != "0");
+        .is_none_or(|it| it != "0")
+        && cfg!(feature = "reqwest-rustls-tls");
     let (backend, notification) = if use_curl_backend {
         (Backend::Curl, Notification::UsingCurl)
     } else {
```

rami3l:
@heiher Done! I guess it should be working now?

heiher:
> @heiher Done! I guess it should be working now?

Acked. :+1:

rami3l:
Blocking on https://github.com/aws/aws-lc-rs/issues/620.

rami3l:
Oops, it looks like https://github.com/rust-lang/rust-bindgen/releases/tag/v0.71.0 has suddenly broken everything...

Not only are the prebuilt binaries and installer missing (as reported in https://github.com/rust-lang/rust-bindgen/issues/3038, though our fallback to `cargo install` did work perfectly), but all `bindgen-cli` related workflows, namely those that depend on `aws-lc`, are now failing.

rami3l:
I could lock the `bindgen-cli` version back to the old one, though I do think it's better to figure out what happened first.

rami3l:
I strongly suspect that https://github.com/rust-lang/rust-bindgen/issues/3039 has caused the build failure; checking `--version` upfront seems to be pretty standard WRT calling any subprocess.

Update: theory confirmed (https://github.com/aws/aws-lc-rs/blob/a7a70cbe163b22f6a4cd438ad2d0c668e2fbdc58/aws-lc-sys/builder/main.rs#L700).

rami3l:
As https://github.com/aws/aws-lc-rs/pull/627 is now open, I'll request the team's review first.

Of course, I'll finish the checklist before actually merging this PR.

ChrisDenton:
I'm not yet familiar with rustup's CI setup. Is skip-master kept because we still expect master to fail after this PR is merged?

rami3l:
> I'm not yet familiar with rustup's CI setup. Is skip-master kept because we still expect master to fail after this PR is merged?

@ChrisDenton We don't run everything on master (that's how things had always been done way before my CI refactoring last year), and essentially the only moment we need to make sure all workflows are green is during a release, which is related to the stable CI rather than the master one. 

I've already triggered the stable CI last week, and it failed; according to the parts that have failed during that run, I have temporarily enabled some workflows just for this PR. They will be disabled again before this PR is merged.

rami3l:
> I'm not yet familiar with rustup's CI setup. Is skip-master kept because we still expect master to fail after this PR is merged?

@ChrisDenton BTW if you're wondering why the master CI is currently failing, that's not related to the build-time behavior, but rather `aws-cli`, so essentially not an issue on our side, unfortunately :(

**Update:** I've made https://github.com/rust-lang/rustup/pull/4111.

rami3l:
@djc `aws-lc-*` deps have been updated and are verified to build in https://github.com/rust-lang/rustup/actions/runs/12384929706 (don't mind the illumos build failure, I've now switched `aws-lc-rs` off for it while rebasing). As such, this PR is ready to be merged.

Cargo.lock
```diff
@@ -170,9 +170,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.11.1"
+version = "1.12.0"
```
djc, line 173: If we require these, should we bump the dependency in `Cargo.toml`?

src/utils/mod.rs
```diff
@@ -256,16 +256,19 @@ async fn download_file_(
         (Backend::Curl, Notification::UsingCurl)
     } else {
         let tls_backend = if use_rustls {
-            TlsBackend::Rustls
-        } else {
-            #[cfg(feature = "reqwest-native-tls")]
+            #[cfg(feature = "reqwest-rustls-tls")]
```
djc: See https://github.com/rust-lang/rustup/pull/4118.

ci/docker/x86_64-unknown-freebsd/Dockerfile
```diff
@@ -1,4 +1,8 @@
 FROM rust-x86_64-unknown-freebsd
 
+# Building `aws-lc-rs` for Linux depends on `gcc-multilib`, `libclang` and `bindgen`.
```
djc: It feels confusing to me that we mention "Linux" in these FreeBSD/Illumos/NetBSD Dockerfiles. Maybe use the proper target OS name in the first line?

Should also suggest upstream that they clarify this page in their documentation that it applies to other Unix-likes too, ideally.

ci/run.bash
```diff
@@ -15,11 +15,12 @@ case "$(uname -s)" in
...
-  # these platforms aren't supported by ring:
-  powerpc* ) ;;
+  # these platforms aren't supported by aws-lc-rs:
+  powerpc64* ) ;;
   mips* ) ;;
-  riscv* ) ;;
-  s390x* ) ;;
+  loongarch* ) ;;
+  *netbsd* ) ;;
+  *illumos* ) ;;
```
djc, line 35: So does aws-lc-rs work on non-64-bit PowerPC, RISC-V and S390x?

djc:
Thanks for all the work on this so far!

Cargo.lock
rami3l: @djc No, because there's no such direct dependency.

ci/run.bash
rami3l: @djc Yes, each addition/deletion has been manually tested on CI: see https://github.com/rust-lang/rustup/actions/runs/12384929706 for everything except illumos; see https://github.com/rust-lang/rustup/actions/runs/12237451661/job/34133256660 for illumos.

ci/docker/x86_64-unknown-freebsd/Dockerfile
rami3l: @djc You're right. This info was extracted from their doc page and probably that very page needs to be updated.

src/utils/mod.rs
rami3l: ~~@djc Acknowledged. Blocking on #4118.~~

rami3l: @djc Wait, if I drop https://github.com/rust-lang/rustup/pull/4105/commits/c4f166ed4d979b7843e2dc835a5bf6fbea01f76a, this PR will be functionally orthogonal with #4118, so it doesn't really have to block on #4118; on the contrary we'll just have to work in parallel (you on #4118 and I on https://github.com/rust-lang/rustup/pull/4105#discussion_r1889971879) and call it a day?

djc: Yes, makes sense!

ci/docker/x86_64-unknown-freebsd/Dockerfile
rami3l: @djc Fixed on our side. I'll inform `aws-lc-rs` shortly.

djc (approved):

PR #3989, opened 2024-08-12
ci(macos): install `awscli` from `brew`

rami3l:
Coming from https://github.com/aws/aws-cli/issues/8858 and https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Strange.20error.20with.20Rustup's.20release.20process.20on.20macOS.20ARM64, it looks like GitHub Actions is installing the x64 version of `awscli` on macOS ARM64, causing weird segfaults on upload.

This PR tries to install the native `awscli` from `brew` instead. It's not guaranteed to work, but AWS might no accept reports running x64 `awscli` on ARM64 in the first place, so this should be a reasonable move forward.

---

Tested with [d696700](https://github.com/rust-lang/rustup/commit/d696700bb39920a8577a05db73beb35ff4a7279d):

```console
> brew install --overwrite awscli
  which aws
  aws --version
  shell: /bin/bash -e {0}
  env:
    MACOSX_DEPLOYMENT_TARGET: 11
==> Downloading https://ghcr.io/v2/homebrew/core/awscli/manifests/2.17.22
==> Fetching awscli
==> Downloading https://ghcr.io/v2/homebrew/core/awscli/blobs/sha256:5e99263aa00c6f2dfb113b929e40f9c412e7ac7c000fc98d0a231c08b5782c71
==> Pouring awscli--2.17.22.arm64_sonoma.bottle.tar.gz
==> Caveats
The "examples" directory has been installed to:
  /opt/homebrew/share/awscli/examples

Bash completion has been installed to:
  /opt/homebrew/etc/bash_completion.d
==> Summary
🍺  /opt/homebrew/Cellar/awscli/2.17.22: 14,541 files, 127.9MB
/opt/homebrew/bin/aws
aws-cli/2.17.22 Python/3.11.9 Darwin/23.6.0 source/arm64
```
_https://github.com/rust-lang/rustup/actions/runs/10351189934/job/28649168983_

```console
> brew install --overwrite awscli
  which aws
  aws --version
  shell: /bin/bash -e {0}
  env:
    MACOSX_DEPLOYMENT_TARGET: 10.12
==> Downloading https://ghcr.io/v2/homebrew/core/awscli/manifests/2.17.27
==> Fetching dependencies for awscli: cffi
==> Downloading https://ghcr.io/v2/homebrew/core/cffi/manifests/1.17.0
==> Fetching cffi
[...]

PR #4111, opened 2024-12-12
ci(windows): don't install `awscli` via `choco`

rami3l:
TLDR: Addresses https://github.com/rust-lang/rustup/pull/4105#issuecomment-2537797422.

It's not our problem, but `choco upgrade awscli` has been causing our `master` CI to fail since a few days ago:

```console
Progress: 100% - Completed download of C:\Users\runneradmin\AppData\Local\Temp\chocolatey\awscli\2.22.13\AWSCLIV2-2.22.13.msi (40.56 MB).
Download of AWSCLIV2-2.22.13.msi (40.56 MB) completed.
Hashes match.
Installing awscli...
WARNING: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\awscli_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error.
ERROR: Running ["C:\Windows\System32\msiexec.exe" /i "C:\Users\runneradmin\AppData\Local\Temp\chocolatey\awscli\2.22.13\AWSCLIV2-2.22.13.msi" /qn /norestart /l*v "C:\Users\runneradmin\AppData\Local\Temp\chocolatey\awscli.2.22.13.MsiInstall.log"] was not successful. Exit code was '1603'. Exit code indicates the following: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\awscli_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error..
The upgrade of awscli was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\awscli\tools\chocolateyinstall.ps1'.
 See log for details.
Chocolatey upgraded 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
Failures
 - awscli (exited 1603) - Error while running 'C:\ProgramData\chocolatey\lib\awscli\tools\chocolateyinstall.ps1'.
 See log for details.
Error: Process completed with exit code 1.
```

... and I guess the only reasonable thing to do now is stop using that command. Instead, we should just check whether `awscli` is already installed, like we currently do on Linux.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,10 +146,11 @@ jobs:
           path: |
             target/${{ matrix.target }}/release/rustup-init.exe
           retention-days: 7
-      - name: Acquire the AWS tooling
+      - name: Ensure that awscli is installed
         if: github.event_name == 'push' && matrix.mode == 'release' && (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/master')
         run: |
-          choco upgrade awscli
+          Get-Command aws
+          aws --version
       - name: Prepare the dist
         if: github.event_name == 'push' && matrix.mode == 'release' && (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/master')
         run: |
@@ -309,10 +310,11 @@ jobs:
           path: |
             target/${{ matrix.target }}/release/rustup-init.exe
           retention-days: 7
-      - name: Acquire the AWS tooling
+      - name: Ensure that awscli is installed
         if: github.event_name == 'push' && matrix.mode == 'release' && (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/master')
         run: |
-          choco upgrade awscli
+          Get-Command aws
+          aws --version
       - name: Prepare the dist
         if: github.event_name == 'push' && matrix.mode == 'release' && (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/master')
         run: |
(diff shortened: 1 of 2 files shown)


ChrisDenton (approved):
Makes sense to me. In any case, if it's already installed then we don't need to do anything.

rami3l:
> Makes sense to me. In any case, if it's already installed then we don't need to do anything.

@ChrisDenton I believe it was there because sometimes inconsistent `awscli` versions between different runners will also cause problems for us... Now I really don't know since GitHub is constantly changing the list of software being installed and how each piece of software is managed in the system. Let's hope the actual release will be alright 🙏 I've scheduled it for this weekend.

ChrisDenton:
> I believe it was there because sometimes inconsistent awscli versions between different runners will also cause problems for us...

Would that not be an issue for Linux runners too?

rami3l:
> > I believe it was there because sometimes inconsistent awscli versions between different runners will also cause problems for us...
> 
> Would that not be an issue for Linux runners too?

@ChrisDenton That's a very good question 🤣 
- https://github.com/rust-lang/rustup/pull/3786
- https://github.com/rust-lang/rustup/pull/3989

PR #4118, opened 2024-12-18
Implement more complete backend selection

djc:
Feedback from #4105.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -214,7 +214,7 @@ async fn download_file_(
 ) -> Result<()> {
     #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
     use download::TlsBackend;
-    use download::{download_to_path_with_backend, Backend, Event};
+    use download::{Backend, Event};
     use sha2::Digest;
     use std::cell::RefCell;
 
@@ -309,9 +309,9 @@ async fn download_file_(
         Backend::Reqwest(_) => Notification::UsingReqwest,
     });
 
-    let res =
-        download_to_path_with_backend(backend, url, path, resume_from_partial, Some(callback))
-            .await;
+    let res = backend
+        .download_to_path(url, path, resume_from_partial, Some(callback))
+        .await;
 
     notify_handler(Notification::DownloadFinished);
 
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -12,6 +12,8 @@ use anyhow::{anyhow, bail, Context, Result};
 use retry::delay::{jitter, Fibonacci};
 use retry::{retry, OperationResult};
 use sha2::Sha256;
+#[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
+use tracing::info;
 use url::Url;
 
 use crate::errors::*;
@@ -210,8 +212,9 @@ async fn download_file_(
     notify_handler: &dyn Fn(Notification<'_>),
     process: &Process,
 ) -> Result<()> {
-    use download::download_to_path_with_backend;
-    use download::{Backend, Event, TlsBackend};
+    #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
+    use download::TlsBackend;
+    use download::{download_to_path_with_backend, Backend, Event};
     use sha2::Digest;
     use std::cell::RefCell;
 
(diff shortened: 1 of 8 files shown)


src/utils/mod.rs
```diff
@@ -246,30 +247,52 @@ async fn download_file_(
...
-            {
-                TlsBackend::NativeTls
+    let use_curl_backend = process.var_os("RUSTUP_USE_CURL").map(|it| it != "0");
+    let use_rustls = process.var_os("RUSTUP_USE_RUSTLS").map(|it| it != "0");
+
+    let backend = match (use_curl_backend, use_rustls) {
+        // If environment specifies a backend that's unavailable, error out
+        #[cfg(not(feature = "reqwest-rustls-tls"))]
+        (_, Some(true)) => {
+            return Err(anyhow!(
```
rami3l, line 259: IIRC `return Err(anyhow!())` is the same thing as `bail!()`.

src/utils/mod.rs
```diff
@@ -246,30 +247,54 @@ async fn download_file_(
...
+                "RUSTUP_USE_RUSTLS is set, but reqwest-rustls-tls is not enabled"
+            ));
+        }
+        #[cfg(not(feature = "reqwest-native-tls"))]
+        (_, Some(false)) => {
+            return Err(anyhow!(
+                "RUSTUP_USE_RUSTLS is set to false, but reqwest-native-tls is not enabled"
```
rami3l: I think the wording can be improved here, since "reqwest-native-tls is not enabled" doesn't necessarily lead me to compile-time configuration.

What about "this rustup distribution was not built with the `reqwest-native-tls` optional feature"?

src/utils/mod.rs
```diff
@@ -246,30 +247,54 @@ async fn download_file_(
...
+            if use_curl_backend == Some(true) {
+                info!("RUSTUP_USE_CURL is set and RUSTUP_USE_RUSTLS is set to off, using reqwest with native-tls");
             }
-            #[cfg(not(feature = "reqwest-native-tls"))]
-            {
-                TlsBackend::Rustls
+            Backend::Reqwest(TlsBackend::NativeTls)
+        }
+        #[cfg(feature = "reqwest-rustls-tls")]
+        (_, _) => {
```
rami3l: The long match here is still a bit difficult to wrap my head around. I admit that the actual logic is already complex here, but I think you can at least be more explicit about `_` cases... For instance, can this be written as something like `(_, Some(true) | None)`?

src/utils/mod.rs
djc: Yeah -- I prefer the former because it's explicit about the `return` and most macro calls do not wrap control flow.

src/utils/mod.rs
ChrisDenton: Maybe using a big match statement isn't ideal for this? The two values are almost entirely independent, except for when both are set.

src/utils/mod.rs
djc: Expanded these to be more explicit.

src/utils/mod.rs
djc: Done. The compiler will make sure that the arms are exhaustive so it feels redundant to restate all the variants covered by this arm.

djc: > Maybe using a big match statement isn't ideal for this? The two values are almost entirely independent, except for when both are set.

Hmm, that's fair, let me try another approach.

djc: Reworked this. Also did a bit more work to make sure that the `match` expressions are exhaustive in most combinations of the Cargo features.

ChrisDenton (commented):
I think this is better now but still a bit less clear then I'd like. Maybe that's unavoidable. Perhaps using `cfg!` instead of `#[cfg]` would help but perhaps not.

As it is, I'd be ok approving this but I'd want to hear @rami3l's thoughts first.

rami3l (approved):
@djc Thanks for your efforts!

I do think the new version looks much better for two reasons:
1. It's clear why each conditional branch is doing certain things (and thus making the review easier).
2. If I'd like remove a feature flag (I hope this won't be too far in the future), I can easily figure out what parts of the code to rip off.

@ChrisDenton We have 7 `cfg`s (2 ^ 3 backend stacks - 1 won't build, if I'm not mistaken) to consider in this piece of code, so unfortunately it will be very long if it was made absolutely plain. My guess is that we'll start to deprecate backends shortly, so this seems bearable.

src/utils/mod.rs
```diff
@@ -246,30 +250,77 @@ async fn download_file_(
...
+            "RUSTUP_USE_RUSTLS is set to false, but this rustup distribution was not built with the reqwest-native-tls feature"
+        ));
+    }
+
+    let backend = match use_curl_backend {
+        #[cfg(feature = "reqwest-backend")]
+        None | Some(false) => Backend::Reqwest(match use_rustls {
+            #[cfg(feature = "reqwest-rustls-tls")]
+            Some(true) | None => {
+                if use_curl_backend == Some(true) {
```
rami3l, line 280: Looks like a typo?

ChrisDenton (approved):
> We have 7 cfgs (2 ^ 3 backend stacks - 1 won't build, if I'm not mistaken) to consider in this piece of code, so unfortunately it will be very long if it was made absolutely plain. My guess is that we'll start to deprecate backends shortly, so this seems bearable.

That's fair.

src/utils/mod.rs
rami3l: Looks like a typo here? You matched `use_curl_backend` to be `None | Some(false)` so it can't be `Some(true)`?

src/utils/mod.rs
```diff
@@ -246,30 +250,77 @@ async fn download_file_(
...
+                    info!(
+                        "both RUSTUP_USE_CURL and RUSTUP_USE_RUSTLS are set, using reqwest with rustls"
+                    );
+                }
+                TlsBackend::Rustls
+            }
             #[cfg(feature = "reqwest-native-tls")]
-            {
+            Some(false) => {
+                if use_curl_backend == Some(true) {
```
rami3l: I think the same thing happens here.

rami3l:
> Also did a bit more work to make sure that the match expressions are exhaustive in most combinations of the Cargo features.

> I think this is better now but still a bit less clear then I'd like. Maybe that's unavoidable. Perhaps using `cfg!` instead of `#[cfg]` would help but perhaps not.

@djc I think this is where `cfg_if` shines with its `else` conditions. I know you had avoided using it and I don't know if it will work at all here, but maybe you can give it a shot?

src/utils/mod.rs
djc: You're right. So this is the reason I think a big `match` expression is actually a better approach, because it makes these cases more explicit.

djc:
> > I think this is better now but still a bit less clear then I'd like. Maybe that's unavoidable. Perhaps using `cfg!` instead of `#[cfg]` would help but perhaps not.
> 
> @djc I think this is where `cfg_if` shines with its `else` conditions. I know you had avoided using it and I don't know if it will work at all here, but maybe you can give it a shot?

I agree that this might not be the optimal encoding but I'm also inclined to spend a bunch more time on it -- this takes care of covering all the cases with reasonably good feedback for the user, and I think that's a good improvement over the status quo. For me, the code isn't bad enough that it's worth investing much more in.

src/utils/mod.rs
rami3l: > You're right. So this is the reason I think a big `match` expression is actually a better approach, because it makes these cases more explicit.

@djc I'd say it's okay to keep the match as long as it's clear where every conditional branch comes from.

djc: > > You're right. So this is the reason I think a big `match` expression is actually a better approach, because it makes these cases more explicit.
> 
> @djc I'd say it's okay to keep the match as long as it's clear where every conditional branch comes from.

The `_` is actually load-bearing if you factor in variance due to Cargo features.

djc:
So brought this back to a single `match` (for the exhaustiveness checking) and checked compilation under most of the potentially relevant feature combinations -- there are now 8 arms in 3 groups which I think makes them easier to reason about? Also cleaned up the download crate a little to get more compile-time errors under different feature combinations.

Dakota48423 (approved):

ChrisDenton (approved):
I'm fine with this version and in any case I agree it's not worth spending more time on tbh.

src/utils/mod.rs
rami3l: @djc Thanks! This is looking much cleaner now.

rami3l (approved):

PR #4122, opened 2024-12-21
fix(rustup-init/sh): don't emit "unknown macOS major version" for macOS v11+

### LINKED ISSUE #4119: rustup-init.sh: unknown macOS major version
In `check_help_for()`, there is a branch for macOS that treats all versions newer than Big Sur as "unknown macOS major version."

Normally this branch is not executed, but before commit a9c4362a1b256ce56ef6a567b0ce46a474b8cd28 there was a brief period where it could be reached when `curl` is too new.

Is the check still necessary?

https://github.com/rust-lang/rustup/blob/dc9d072c34eccbee2d9249daea177b9c4263c5da/rustup-init.sh#L732-L739

By the way, `:` is missing on the last line of `warn`, which is inconsistent and might worth fixing.

rami3l:
Closes #4119.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -733,14 +733,13 @@ check_help_for() {
                         return 1
                     fi
                     ;;
-                11)
-                    # We assume Big Sur will be OK for now
-                    ;;
                 *)
-                    # Unknown product version, warn and continue
-                    warn "Detected unknown macOS major version: $_os_version"
-                    warn "TLS capabilities detection may fail"
-                    ;;
+                    if ! { [ "$_os_major" -eq "$_os_major" ] 2>/dev/null && [ "$_os_major" -ge 11 ]; }; then
+                        # Unknown product version, warn and continue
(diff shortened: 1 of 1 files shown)


rustup-init.sh
```diff
@@ -719,24 +719,27 @@ check_help_for() {
...
                     ;;
-                11.*)
-                    # We assume Big Sur will be OK for now
-                    ;;
                 *)
-                    # Unknown product version, warn and continue
-                    warn "Detected unknown macOS major version: $(sw_vers -productVersion)"
-                    warn "TLS capabilities detection may fail"
-                    ;;
+                    if ! { [ "$_os_major" -eq "$_os_major" ] 2>/dev/null && [ "$_os_major" -ge 11 ]; }; then
```
rami3l, line 737: `[[ ]]` doesn't exist on POSIX shell so we have to use this weird `[ "$_os_major" -eq "$_os_major" ]` trick to check for numerics first...

djc (approved):

PR #4124, opened 2024-12-22
drop unused dep or make them optional

klensy:


djc (approved):
Thanks, looks nice!

rami3l (approved):
Many thanks!

PR #4130, opened 2024-12-29
feat(cli): show the toolchain path with `rustup show [active-toolchain] --verbose`

### LINKED ISSUE #4129: `rustup show active-toolchain` should provide the toolchain path

> Let me guess: you actually just want to get the active toolchain path with this snippet...
> 
> https://github.com/corrosion-rs/corrosion/blob/b88ec09915a6b12d1aaec7e451ecf0e41a831030/cmake/FindRust.cmake#L423-L428
> 
> ... however you might have tried the following options and concluded that neither is satisfactory:
> - `rustup show active-toolchain`. This seems like the most obvious option, however it is somehow not willing to give you the toolchain path...
> - `rustup toolchain list --verbose`. This does give you the toolchain paths, but you have to manually deduce what the active toolchain is :|

_https://github.com/corrosion-rs/corrosion/issues/590#issuecomment-2564138875_
            

### Solution you'd like

`rustup show active-toolchain` should provide the toolchain path, or optionally with `--verbose`.

### Notes

_No response_

rami3l:
Closes #4129.

This PR also tries to make `--verbose`'s effect on `rustup show`'s output more like that of `rustup show active-toolchain`.

### DIFF
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -661,7 +661,6 @@ nightly-{0} (active, default)
 active toolchain
 ----------------
 name: nightly-{0}
-compiler: 1.3.0 (hash-nightly-2)
 active because: it's the default toolchain
 installed targets:
   {0}
@@ -735,7 +734,6 @@ nightly-{0} (active, default)
 active toolchain
 ----------------
 name: nightly-{0}
-compiler: 1.3.0 (hash-nightly-2)
 active because: it's the default toolchain
 installed targets:
   {0}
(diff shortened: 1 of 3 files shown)


tests/suite/cli_rustup.rs
```diff
@@ -661,7 +661,6 @@ nightly-{0} (active, default)
 active toolchain
 ----------------
 name: nightly-{0}
-compiler: 1.3.0 (hash-nightly-2)
```
djc, line 664: I would be inclined to show the version even without --verbose.

djc (approved):

tests/suite/cli_rustup.rs
rami3l: @djc The original rationale for putting the rustc version in `--verbose` is that it requires a subprocess call. See also `rustup --version --verbose`.

djc: Fair enough!

djc: Fair enough!

PR #4141, opened 2025-01-04
fix(cli/rustup-mode): err when trying to show the default/active toolchain but there is none

### LINKED ISSUE #4140: `rustup default` no longer exits with `1` with no default toolchain is set
### Verification

### Problem

(Edited by @rami3l to better reflect the original issue)

Coming from <https://internals.rust-lang.org/t/seeking-beta-testers-for-rustup-v1-28-0/22060>.

With rustup v1.27.1, the following shell snippet can be used to force the installation of the default toolchain, or fall back on `stable` when none is present:

```bash
rustup default || rustup default stable
```

When testing out the upcoming 1.28 release, however, I found that `rustup default` is returning `0` even there's no such thing as the default toolchain, resulting in an unchanged default toolchain state.

I've noticed the suggestion provided in the post:

```
rustup show active-toolchain || rustup toolchain install
```

... but it doesn't really help with the use case above, since regardless in 1.27.1, the current stable branch (1.28-beta), it seems to always exit with code 0.

* `1.27.1` https://github.com/rust-lang/rustup/blob/54dd3d00fd20e64dc522517a0d7be4285570a2f1/src/cli/rustup_mode.rs#L1227-L1257
* `stable` https://github.com/rust-lang/rustup/blob/7ccf717e6e1aee46f65cc6fea4132a3f0e37593b/src/cli/rustup_mode.rs?plain=1#L1100-L1105

### Steps

1. `rustup default none`
2. `rustup default || rustup default stable`

### Possible Solution(s)

Should we find another workaround?

### Notes

_No response_

### Rustup version

```console
* rustup 1.27.1 (1980-01-01)
* rustup 1.28.0 (7ccf717e6e1aee46f65cc6fea4132a3f0e37593b)
```

### Installed toolchains

rustup 1.27.1

```
installed toolchains
--------------------

```

rustup 1.28.0

```
installed toolchains
--------------------

active toolchain
----------------
no active toolchain
```

### OS version

Both Darwin and Linux

rami3l:
Closes #4140.

## Concerns

- [ ] When should `rustup show` exit with `1` (https://github.com/rust-lang/rustup/issues/4140#issuecomment-2570016612)?

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -753,15 +753,10 @@ async fn default_(
             }
         }
     } else {
-        match cfg.get_default()? {
-            Some(default_toolchain) => {
-                writeln!(cfg.process.stdout().lock(), "{default_toolchain} (default)")?;
-            }
-            None => writeln!(
-                cfg.process.stdout().lock(),
-                "no default toolchain is configured"
-            )?,
-        }
+        let default_toolchain = cfg
+            .get_default()?
+            .ok_or_else(|| anyhow!("no default toolchain is configured"))?;
+        writeln!(cfg.process.stdout().lock(), "{default_toolchain} (default)")?;
     }
 
     Ok(utils::ExitCode(0))
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1096,7 +1096,7 @@ fn show_active_toolchain(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode
                 )?;
             }
         }
-        None => return Err(anyhow!("there isn't an active toolchain")),
+        None => return Err(anyhow!("no active toolchain")),
     }
     Ok(utils::ExitCode(0))
 }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1096,10 +1096,7 @@ fn show_active_toolchain(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode
                 )?;
             }
         }
-        None => writeln!(
-            cfg.process.stdout().lock(),
-            "There isn't an active toolchain"
-        )?,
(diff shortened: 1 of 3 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1096,10 +1096,7 @@ fn show_active_toolchain(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode
                 )?;
             }
         }
-        None => writeln!(
-            cfg.process.stdout().lock(),
-            "There isn't an active toolchain"
-        )?,
+        None => return Err(anyhow!("there isn't an active toolchain")),
```
djc: Maybe tweak this wording to "there is no active toolchain" while you're in here?

djc (approved):
Nice!

src/cli/rustup_mode.rs
ChrisDenton: Simply "no active toolchain" would be fine, I think? That should be rendered as `error: no active toolchain` and would be consistent with the other no active toolchain error message.

djc: Yup, that's probably better.

rami3l: @djc @ChrisDenton Thanks for the suggestions! I've updated the error message.

ChrisDenton (approved):

weihanglo:
Thanks for the swift fix, people.

I'm not familiar with the release process, though I guess this will be backported to the stable branch for 1.28?

rami3l:
> I'm not familiar with the release process, though I guess this will be backported to the stable branch for 1.28?

@weihanglo For now `stable` is just a snapshot of `master` at some point in time, so, indeed, it will turn stable when we want to update that snapshot and cut the actual release. For now, you can try build from source or take advantage of our [CI artifacts](https://github.com/rust-lang/rustup/actions/runs/12657039661) :)

PR #4146, opened 2025-01-07
Update Windows dependencies

kennykerr:
The latest release of the `windows-registry` and `windows-result` crates are now available. 

https://github.com/microsoft/windows-rs/releases/tag/0.61.0

This also syncs the version of the `windows-registry` crate with the version used to build `reqwest`. https://github.com/seanmonstar/reqwest/pull/2518

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2317,8 +2317,8 @@ dependencies = [
  "url",
  "wait-timeout",
  "walkdir",
- "windows-registry 0.3.0",
- "windows-result",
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks! Would be nice if you can clean up the commit message a bit (we merge via rebase -- something like "Update Windows dependencies" would do nicely).

kennykerr:
Renamed - there really should be an easier way to rename commits. 🤯

Issue #4099, opened 2024-11-25
Shall `wasm32-wasi`-related messages be improved?


vi:
### Problem you are trying to solve

```
$ rustup update nightly
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: latest update on 2024-11-25, rust version 1.85.0-nightly (28fc2ba71 2024-11-24)
info: skipping nightly which is missing installed component 'rust-std-wasm32-wasi'
info: syncing channel updates for 'nightly-2024-11-24-x86_64-unknown-linux-gnu'
info: syncing channel updates for 'nightly-2024-11-23-x86_64-unknown-linux-gnu'
info: latest update on 2024-11-23, rust version 1.85.0-nightly (a47555110 2024-11-22)
info: skipping nightly which is missing installed component 'rust-std-wasm32-wasi'
...
```

Nightly update falls back (probably because of [WASI target changes](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html)).

### Solution you'd like

Shall `rustup` include a helpful message for `wasm32-wasi` specifically, mentioning that now one should write `wasm32-wasip1`?

Or maybe even automatically migrate to it.

rami3l:
@vi Thanks for filing this issue!

I think it's a bit dangerous to hardcode anything from our side, but given the repeated requests (which led to patches like #3453 and #3920), maybe we could borrow from Homebrew [the "caveats" section](https://docs.brew.sh/Cask-Cookbook#stanza-caveats) which is controlled by the release server instead?

ngoldbaum:
I just hit this today now that the change to `wasm32-wasip1` stabilized in rust 1.84. If anyone is finding this thread and just wants a fix, you can do 

```bash
rustup target remove wasm32-wasi
rustup update
rustup target add wasm32-wasip1
```

I agree that it would be nice to have an explicit helpful message containing something like these steps.

rami3l:
@ngoldbaum It should be noted though, if I were to directly add this note in Rustup today, it will not be delivered to the users in time, since we expect to release about once or twice a year, and the upstream is changing really fast (one stable release per 6 weeks).

[...]

PR #4149, opened 2025-01-10
docs(user-guide/components): add deprecation notice for `wasm32-wasi`

rami3l:
Partially addresses https://github.com/rust-lang/rustup/issues/4099.

### DIFF
--- a/doc/user-guide/src/concepts/components.md
+++ b/doc/user-guide/src/concepts/components.md
@@ -71,6 +71,9 @@ These components have been deprecated and are not published in new Rust releases
 * `rls` --- [RLS] is a language server that is deprecated and has been replaced
   by rust-analyzer.
 * `rust-analysis` --- Metadata about the standard library, used by [RLS].
+* The `wasm32-wasi` target --- It has been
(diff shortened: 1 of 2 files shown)


doc/user-guide/src/concepts/components.md
```diff
@@ -71,6 +71,9 @@ These components have been deprecated and are not published in new Rust releases
 * `rls` --- [RLS] is a language server that is deprecated and has been replaced
   by rust-analyzer.
 * `rust-analysis` --- Metadata about the standard library, used by [RLS].
+* The `wasm32-wasi` target --- It is now
+  [deprecated](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html)
```
ChrisDenton: Hm it's not really deprecated, it has been renamed, no?

rami3l: @ChrisDenton You are right. Updated.

ChrisDenton (approved):
Whatever else we, do this looks like a start. It would be good to have some upstream support for renames but I guess it's a very rare occurrence.

PR #4154, opened 2025-01-16
Use ARM based runners for ARM CI targets

Kobzol:
This should hopefully resolve CI errors.


djc (approved):

Kobzol:
Looks like it works now, at least CI is green. The CI log contained this:
```
ERROR: there isn't a download for your platform aarch64-unknown-linux-gnu
```
not sure if that is something that we should worry about?

rami3l (approved):
@Kobzol Thanks a lot!

~~OTOH I think we can give the public runners a try? I guess it'd be fine if we are all on ARM64 but with different runner configurations...~~ I've opened #4156.

rami3l:
> Looks like it works now, at least CI is green. The CI log contained this:
> 
> ```
> ERROR: there isn't a download for your platform aarch64-unknown-linux-gnu
> ```
> 
> not sure if that is something that we should worry about?

@Kobzol The `bindgen-cli` downloader is complaining about not being able to find anything to download... but the fallback seems to work correctly, so that's good news.

```console
ERROR: there isn't a download for your platform aarch64-unknown-linux-gnu
+ cargo install --force --locked bindgen-cli
```

I think with the announcement of ARM64 Linux runners they will be able to push the corresponding artifacts pretty soon, hopefully :)

PR #4156, opened 2025-01-17
ci(linux): use public ARM64 Linux runners

rami3l:
Continuation of #4154.

After reading https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Free.20Linux.20ARM64.20runners.20on.20GHA.20now.20in.20beta, it looks like our upstream (https://github.com/rust-lang/rust) has some issues WRT the new runner's virtual environment.

However, this shouldn't affect us in theory because we don't use that environment :)

**Update:** The new runners are blazingly ⚡ fast! They can complete the full test suite *within 6 minutes* where it usually will take our ARM64 Mac runners ~12 minutes to finish.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -543,6 +543,7 @@ jobs:
             # to generate their Dockerfiles.
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-24.04-arm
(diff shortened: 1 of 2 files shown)


djc (approved):
Nice!

PR #4157, opened 2025-01-17
ci(deploy-docs): install `mdbook` with `install-action`

rami3l:
This PR uses the `install-action` to install `mdbook`, giving us access to the new features and/or fixes from the upstream.

We're already depending on the very same action to install `taplo`:

https://github.com/rust-lang/rustup/blob/3fc52e98e6190843bd549e46f457acaa97f6df8d/ci/actions-templates/centos-fmt-clippy-template.yaml#L68-L71


.github/workflows/deploy-docs.yaml
```diff
@@ -20,11 +21,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Install mdbook
-        run: |
-          mkdir mdbook
-          curl -Lf https://github.com/rust-lang/mdBook/releases/download/v0.4.37/mdbook-v0.4.37-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=./mdbook
-          echo "`pwd`/mdbook" >> $GITHUB_PATH
+      - name: Install taplo
```
ChrisDenton:
```suggestion
      - name: Install mdbook
```

ChrisDenton:
Maybe it could also be used here: https://github.com/rust-lang/rustup/blob/3fc52e98e6190843bd549e46f457acaa97f6df8d/.github/workflows/ci.yaml#L1525-L1529

And here: https://github.com/rust-lang/rustup/blob/3fc52e98e6190843bd549e46f457acaa97f6df8d/ci/actions-templates/test-docs-template.yaml#L15-L19

rami3l:
> Maybe it could also be used here:
> 
> https://github.com/rust-lang/rustup/blob/3fc52e98e6190843bd549e46f457acaa97f6df8d/.github/workflows/ci.yaml#L1525-L1529
> 
> And here:
> 
> https://github.com/rust-lang/rustup/blob/3fc52e98e6190843bd549e46f457acaa97f6df8d/ci/actions-templates/test-docs-template.yaml#L15-L19

@ChrisDenton Thanks for pointing that out! I actually think these two jobs should be merged so these books can be tested on PR using the same workflow. I'll try to do that and delete this job...

.github/workflows/deploy-docs.yaml
rami3l: @ChrisDenton Yup that's a typo on my side. Will fix.

.github/workflows/ci.yaml
```diff
@@ -1560,7 +1534,6 @@ jobs:
       - build-linux-stable
       - build-macos-aarch64
       - build-macos-x86_64
-      - doc
```
rami3l: Wait, probably I should move the resulting `doc` job in `ci.yaml`.

ChrisDenton (approved):

ehuss:
FWIW, I do not recommend doing this without pinning to a specific version. It is not unusual that new releases break things, and it's best to test new versions and look at how it renders before updating. (Fortunately the rustup book is pretty basic, so problems are less likely, but I still wouldn't recommend it.)

rami3l:
> FWIW, I do not recommend doing this without pinning to a specific version. It is not unusual that new releases break things, and it's best to test new versions and look at how it renders before updating. (Fortunately the rustup book is pretty basic, so problems are less likely, but I still wouldn't recommend it.)

@ehuss Thanks for the feedback! I have a question about this though:

When making this PR I thought, at least for me, it would be the best if we use the same version for local preview (I use the `watch` command when editing the book on my machine) and for deployment, and IIRC I've already noticed a few updates (but no breakages so far) with my local render compared to the online version...

Are we supposed to use like a one-year-old `mdbook` version locally, or is there a convenient way to check within our CI pipeline? Also, what kind of breakages do you have in mind?

PS: Pinning the version would simply be two extra lines in this new setup, I've done this before.

ehuss:
Yea, it's a bit tricky to test locally with the exact same version that is used in this repo. Generally my workflow is:

1. Update locally to the latest mdbook
2. Render locally, make sure it looks ok
3. Push a PR to update to the latest version
4. (Optional) Verify the new rendering at https://rust-lang.github.io/rustup/devel/

Or if you have an automated update (like with renovatebot):
1. Renovate opens a PR to update mdbook
2. Update mdbook locally
3. Render locally, make sure it looks ok
4. Merge renovatebot's PR

So, you can generally stay on the latest version of mdbook (particularly with renovatebot), but the actual transition happens through a PR, not implicitly when new versions are published.

It's tough to say exactly what kinds of breakage might happen. Some past examples might be https://github.com/rust-lang/mdBook/issues/2482, https://github.com/rust-lang/mdBook/issues/2479, https://github.com/rust-lang/mdBook/pull/2473, https://github.com/rust-lang/mdBook/pull/2383, https://github.com/rust-lang/mdBook/issues/2472, https://github.com/rust-lang/mdBook/issues/1949, https://github.com/rust-lang/mdBook/issues/1743, https://github.com/rust-lang/mdBook/issues/1235, etc. There have also been issues with these third-party installers having problems or bugs or bad interactions or assumptions with how the releases are structured.

PR #4160, opened 2025-01-18
ci(schedule): promote to use the `stable` job list

### LINKED ISSUE #4155: Make the CI cron task more frequent
With respect to our recent (mostly upstream-related) CI breakages, @kobzol suggested:

> maybe the cron could run once per days instead of once per week to catch these breakages more quickly?

rami3l:
Closes https://github.com/rust-lang/rustup/issues/4155.

There is no way to verify this change other than merging it and observe, but I have skimmed through the steps and haven't found any signs that this PR will change anything other than the job list (for example, will it cause the S3 releasing-related behaviors on `schedule` to behave like `push on stable`?), but please feel free to double check while reviewing.

As usual, please ignore `ci.yaml` since it's a generated file.


ChrisDenton (approved):
This look right to me but maybe wait for djc if you want to be cautious before running in CI?

djc (approved):
LGTM!

rami3l:
Looks like it's running fine (https://github.com/rust-lang/rustup/actions/runs/12858770122)!

PR #4162, opened 2025-01-21
ci(doc): pin `mdbook` and let `renovatebot` update it

rami3l:
Implements https://github.com/rust-lang/rustup/pull/4157#issuecomment-2602623605.

This basically mirrors https://github.com/astral-sh/uv/pull/7888, however I have no idea if that magic comment will work at all...

If it works, then locking `taplo` might be a good idea as well.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,8 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
     "extends": [
-        "config:recommended"
+        "config:recommended",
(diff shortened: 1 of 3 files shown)


djc (approved):
Nice!

> This basically mirrors [astral-sh/uv#7888](https://github.com/astral-sh/uv/pull/7888), however I have no idea if that magic comment will work at all...

Only one way to find out?

ChrisDenton (approved):
Looks good to me. The test failures look unrelated?

PR #4179, opened 2025-02-09
fix(component): normalize path separators

rami3l:
Supersedes #3077.

This solution is based on the analysis made in https://github.com/rust-lang/rustup/pull/3077#discussion_r1909829756.

~~PS: Sincerely, I think this `ComponentPart` needs to be changed to something like:~~

```rs
struct ComponentPart {
  pub kind: String,
  pub path: PathBuf,
}
```

~~... I'd like to work on stylistic changes like this once I'm done with updating the test cases (some of them are of course expected to break).~~
Done.

### DIFF
--- a/src/dist/component/components.rs
+++ b/src/dist/component/components.rs
@@ -198,13 +198,16 @@ impl ComponentPart {
     const PATH_SEP_MAIN: &str = std::path::MAIN_SEPARATOR_STR;
 
     pub(crate) fn encode(&self) -> String {
+        let mut buf = self.kind.to_string();
+        buf.push(':');
         // Lossy conversion is safe here because we assume that `path` comes from
         // `ComponentPart::decode()`, i.e. from calling `Path::from()` on a `&str`.
         let mut path = self.path.to_string_lossy();
         if Self::PATH_SEP_MAIN != Self::PATH_SEP_MANIFEST {
             path = Cow::Owned(path.replace(Self::PATH_SEP_MAIN, Self::PATH_SEP_MANIFEST));
         };
-        format!("{}:{path}", self.kind)
+        buf.push_str(&path);
+        buf
     }
 
     pub(crate) fn decode(line: &str) -> Option<Self> {
--- a/src/dist/component/components.rs
+++ b/src/dist/component/components.rs
@@ -3,8 +3,11 @@
 //! installation / uninstallation process.
 
 use std::borrow::Cow;
+use std::convert::Infallible;
+use std::fmt;
 use std::io::BufWriter;
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 
 use anyhow::{bail, Result};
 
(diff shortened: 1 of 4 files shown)


src/dist/component/components.rs
```diff
@@ -145,15 +154,38 @@ impl<'a> ComponentBuilder<'a> {
 }
 
 #[derive(Debug)]
-pub struct ComponentPart(pub String, pub PathBuf);
+pub struct ComponentPart {
+    /// Kind of the [`ComponentPart`], such as `"file"` or `"dir"`.
+    pub kind: String,
```
ChrisDenton: It's pre-existing but could this be an enum?

src/dist/prefix.rs
```diff
@@ -2,7 +2,14 @@ use std::path::{Path, PathBuf};
...
 use crate::utils;
 
-const REL_MANIFEST_DIR: &str = "lib/rustlib";
+/// The relative path to the manifest directory in a Rust installation,
+/// with path components separated by [`std::path::MAIN_SEPARATOR`].
+const REL_MANIFEST_DIR: &str = match std::path::MAIN_SEPARATOR {
+    '/' => "lib/rustlib",
+    '\\' => r"lib\rustlib",
+    _ => panic!("unknown `std::path::MAIN_SEPARATOR`"),
+};
```
ChrisDenton, line 11: I feel like we should be able to do this better but I couldn't come up with anything better than a macro that just wasn't worth it for one use. So on reflection this is fine.

rami3l: @ChrisDenton I've successfully implemented [const-concatenation](https://github.com/rami3l/rustup/blob/a807ea59ec2ced51eaafc2fd7c183357dba1176f/src/dist/prefix.rs#L5-L42) but that has been way too verbose for only this, so I truncated it to what you see now.

src/dist/component/components.rs
rami3l: @ChrisDenton Thanks! I've made [`b4323b7` (#4179)](https://github.com/rust-lang/rustup/pull/4179/commits/b4323b7e9e93f949984e895d43d483f3496012a3).

Meanwhile, I chose to keep an `Unknown` variant to maintain bidirectional conversions:
- I could have used `kind: Result<ComponentPartKind, String>` but I don't think we are really using any `Result`-related facilities in that particular case.
- A hard `FromStr` error would look weird to me as well, since the original code handles unknown `ComponentPartKind`s much later in the business logic, and I don't want this fix focused on path normalization to get to involved with all that.

ChrisDenton (approved):
Thanks! I think this will be more consistent between platforms.

src/dist/component/components.rs
```diff
@@ -148,12 +149,28 @@ impl<'a> ComponentBuilder<'a> {
...
+
     pub(crate) fn encode(&self) -> String {
-        format!("{}:{}", &self.0, &self.1.to_string_lossy())
+        let mut path = self.1.to_string_lossy();
+        if Self::PATH_SEP_MAIN != Self::PATH_SEP_MANIFEST {
+            // Lossy conversion is safe here because we assume that `path` comes from
+            // `ComponentPart::decode()`, i.e. from calling `Path::from()` on a `&str`.
+            path = Cow::Owned(path.replace(Self::PATH_SEP_MAIN, Self::PATH_SEP_MANIFEST));
+        };
+        format!("{}:{path}", self.0)
```
djc: It seems not that hard to avoid the extra allocation from concatenating here, maybe try that?

djc: (Sorry for the late/slow review.)

rami3l: @djc That is a good point! This still looks like an existing issue though, so I have fixed it in a separate commit.

Theoretically this can be done with a single allocation (since `.replace()` is implemented via `.push_str()` on a new buffer), but I haven't found anything like `.push_replaced_str()` so I've left the `Cow<'_, str>` as-is for more readability.

djc (approved):

PR #4190, opened 2025-02-20
style(edition): migrate the codebase to the 2024 Style Edition

rami3l:
This PR is the first part of rustup's migration to Rust 2024, which contains style-related changes only. Notably, macro invocations are now correctly formatted by `rustfmt`.

A second PR will follow shortly once this one gets merged, which will perform a semi-auto migration based on  #4061's plan:

> ```toml
> [workspace.lints.rust]
> rust_2018_idioms = { level = "deny", priority = -1 }
> rust-2024-compatibility = { level = "warn", priority = -2 }
> ```

PS: The newly-added `rustfmt` config is here for migration purposes. It should be deleted by the end of the second PR.

PPS: The CI should pass once the 2024 Edition lands in stable in a few hours.

### DIFF
--- a/download/src/lib.rs
+++ b/download/src/lib.rs
@@ -327,8 +327,8 @@ pub mod reqwest_be {
     use std::sync::LazyLock;
     use std::time::Duration;
 
-    use anyhow::{anyhow, Context, Result};
-    use reqwest::{header, Client, ClientBuilder, Proxy, Response};
+    use anyhow::{Context, Result, anyhow};
+    use reqwest::{Client, ClientBuilder, Proxy, Response, header};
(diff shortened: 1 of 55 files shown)


ChrisDenton (approved):
Seems fine. Though did something about maximum line length change? In any case using the new style seems like a good idea.

rami3l:
@ChrisDenton I believe what's really happening is that `rustfmt` is now trying its best to make macro invocations behave like function invocations, and there used to be many macro invocation lines that are too wide but got ignored.

ChrisDenton:
Ah that makes sense.

PR #4191, opened 2025-02-22
build: migrate the codebase to the 2024 Language Edition

rami3l:
Continuation of #4190.

This is a semi-automatic migration due to the sheer amount of false positives in the `rust-2024-compatibility` lint group and the `if_let_rescope` lint in particular. Some steps have more details in the corresponding commit messages.

You can get a clearer version of the PR diff via [_semanticdiff_](https://app.semanticdiff.com/gh/rust-lang/rustup/pull/4191).

## Concerns

- [ ] Did I get all the `if-let` correctly migrated?
  - To make the review easier, I've made #4192 to include all the suggestions made by `clippy` where there are a lot of false positives.
  -  Please feel free to point out if I have missed a significant drop somewhere.

### DIFF
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -20,7 +20,7 @@ use crate::{process::Process, utils::notify::NotificationLevel};
 pub fn tracing_subscriber(
     process: &Process,
 ) -> (
-    impl tracing::Subscriber,
+    impl tracing::Subscriber + use<>,
     reload::Handle<EnvFilter, Registry>,
 ) {
     #[cfg(feature = "otel")]
@@ -45,7 +45,7 @@ pub fn tracing_subscriber(
 /// When the `RUSTUP_LOG` environment variable is present, a standard [`tracing_subscriber`]
 /// formatter will be used according to the filtering directives set in its value.
 /// Otherwise, this logger will use [`EventFormatter`] to mimic "classic" Rustup `stderr` output.
-fn console_logger<S>(process: &Process) -> (impl Layer<S>, reload::Handle<EnvFilter, S>)
+fn console_logger<S>(process: &Process) -> (impl Layer<S> + use<S>, reload::Handle<EnvFilter, S>)
 where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
@@ -129,7 +129,7 @@ impl NotificationLevel {
 /// A [`tracing::Subscriber`] [`Layer`][`tracing_subscriber::Layer`] that corresponds to Rustup's
 /// optional `opentelemetry` (a.k.a. `otel`) feature.
 #[cfg(feature = "otel")]
-fn telemetry<S>(process: &Process) -> impl Layer<S>
+fn telemetry<S>(process: &Process) -> impl Layer<S> + use<S>
 where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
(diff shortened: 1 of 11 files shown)


src/diskio/immediate.rs
```diff
@@ -203,25 +200,24 @@ impl IncrementalFileWriter {
...
-                    trace_scoped!("write_segment", "name": self.path_display, "len": chunk.len());
-                    file.write_all(&chunk)?;
-
-                    state.completed_chunks.push(chunk.len());
-                }
-                Ok(true)
+        let Some(ref mut state) = *state else {
+            unreachable!()
+        };
+        if let Some(ref mut file) = self.file.as_mut() {
```
djc: Do a let-else here, too?

src/dist/mod.rs
```diff
@@ -335,11 +335,7 @@ impl FromStr for ParsedToolchainDesc {
             }
         });
 
-        if let Some(d) = d {
-            Ok(d)
-        } else {
-            Err(RustupError::InvalidToolchainName(desc.to_string()).into())
-        }
+        d.ok_or_else(|| RustupError::InvalidToolchainName(desc.to_string()).into())
```
djc: Nit: this is a little dense for my taste, I think a `match` expression would be easier to read.

src/cli/log.rs
```diff
@@ -20,7 +20,7 @@ use crate::{process::Process, utils::notify::NotificationLevel};
 pub fn tracing_subscriber(
     process: &Process,
 ) -> (
-    impl tracing::Subscriber,
+    impl tracing::Subscriber + use<'_>,
```
djc: Are the `use` blocks necessary for compilation to succeed?

djc (approved):

djc:
Nice work!

src/cli/log.rs
rami3l: Yes, `impl Foo + use<A>` is there to indicate that `impl Foo` is only involved with `A` and nothing else, otherwise the 2024 Language Edition will deduce that you'll also need all unused generic parameters such as `'self`. In our case this will cause overly restrictive bounds.

rami3l: Wait, I've probably done it wrong: the bound on `telemetry()` is the one that should be fixed, not the one on the final logger. Maybe the clippy lints are right this time: I'll have another look.

src/diskio/immediate.rs
rami3l: Done.

src/dist/mod.rs
rami3l: Fair enough...

src/cli/log.rs
rami3l: Yes, `clippy` is doing it right this time :)

src/cli/log.rs
```diff
@@ -129,7 +129,7 @@ impl NotificationLevel {
 /// A [`tracing::Subscriber`] [`Layer`][`tracing_subscriber::Layer`] that corresponds to Rustup's
 /// optional `opentelemetry` (a.k.a. `otel`) feature.
 #[cfg(feature = "otel")]
-fn telemetry<S>(process: &Process) -> impl Layer<S>
+fn telemetry<S>(process: &Process) -> impl Layer<S> + use<S>
```
rami3l, line 132: @djc So basically `impl Layer<S> + use<S>` means "this `impl Layer<S>` has nothing to do with `'process`" (this function is generic over both `'process` and `S`, so we `use<S>` to exclude `'process`), and the same thing goes for other fixes in this commit.

rami3l:
I just did one more check and have again found no missing significant drops. I think it should be safe to merge this one now.

PR #4192, opened 2025-02-22
DON'T MERGE[ci skip]: demonstrate `if_let_rescope` auto migration

rami3l:
Complements https://github.com/rust-lang/rustup/pull/4191 to show the `if-let` migrations that I **didn't** do.

For the sake of simplicity, the first commit of #4191 has also been cherry-picked into this branch. After that, a `cargo clippy --fix` is performed after adjusting the `Cargo.toml` to activate the `if_let_rescope` lint.

Notably, according to https://github.com/rust-lang/rust/issues/133167#issuecomment-2674679030, `if let Some()` and other similar cases **where the `else` drop is not significant** are not affected by the Edition change.

Please feel free to point out if I have missed a significant drop somewhere in #4191.

You can get a clearer version of the PR diff via [_semanticdiff_](https://app.semanticdiff.com/gh/rust-lang/rustup/pull/4192).

### DIFF
--- a/src/diskio/immediate.rs
+++ b/src/diskio/immediate.rs
@@ -81,21 +81,20 @@ impl Executor for ImmediateUnpacker {
                     // If there is a pending error, return it, otherwise stash the
                     // Item for eventual return when the file is finished.
                     let mut guard = self.incremental_state.lock().unwrap();
-                    if let Some(ref mut state) = *guard {
-                        if state.err.is_some() {
-                            let err = state.err.take().unwrap();
-                            item.result = err;
-                            item.finish = item
-                                .start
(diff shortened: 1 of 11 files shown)


src/diskio/immediate.rs
```diff
@@ -181,30 +181,30 @@ impl IncrementalFileWriter {
         if (self.state.lock().unwrap()).is_none() {
             return false;
         }
-        let chunk = if let FileBuffer::Immediate(v) = chunk {
```
rami3l, line 184: The `else` drop is significant here, however the `else` branch only has an `unreachable!()`.

PR #4193, opened 2025-02-23
refactor: improve binary suffix stripping

Timmmm:
Use `strip_suffix()` which is clearer, shorter, and won't break if the suffix is in the middle of the filename (unlikely, but still). I also renamed `binary_prefix` to something clearer, since it isn't actually a prefix; it's the main part.


djc (approved):

djc:
Thanks!

Issue #4195, opened 2025-02-24
Reorganize `Process` to include extra global states


rami3l:
This was part of the original plan after #3803, but we probably have put it aside for a while:

In https://github.com/rust-lang/rustup/pull/4182#issuecomment-2663126604, I've noticed that the current `Process` abstraction should probably be extended to allow the existence of a `tracing_subscriber` global state, so that we can handle graceful shutdown more easily.

Having noticed the clear pattern here:

https://github.com/rust-lang/rustup/blob/52e6070c98ca8f4c26fce9107ff8310323dfa700/src/bin/rustup-init.rs#L44-L53

... it should probably be modeled as an RAII guard.

I have also noticed that this should resolve another problem, i.e. the corresponding construct in tests is simply no longer used, as these two functions no longer have a caller:

https://github.com/rust-lang/rustup/blob/52e6070c98ca8f4c26fce9107ff8310323dfa700/src/test.rs#L235-L250

This is most probably caused by the removal of `rustup_macros` (https://github.com/rust-lang/rustup/pull/3877/commits/d60f0058503954f57786d16e15a2d42f6798e141): indeed, the tests are passing on the CI, but sometimes one might want to run tests with `otel` on as well.

@djc What do you think?

djc:
Right, I remember we discussed with @rbtcollins whether there was value in the OpenTelemetry support and he argued that it was very helpful for him the few times he needed it. I'm basically still not sure there is much value in the OpenTelemetry support that we've got and apparently may have broken it inadvertently.

I feel like, unless it's used regularly -- and it hasn't been for a few years at this point I guess -- it may be better to remove the feature altogether because it definitely has maintenance costs. Can always add it back when we need it, and I'm not sure fixing it now when we don't is a good way to use our time.

rami3l:
> Right, I remember we discussed with @rbtcollins whether there was value in the OpenTelemetry support and he argued that it was very helpful for him the few times he needed it. I'm basically still not sure there is much value in the OpenTelemetry support that we've got and apparently may have broken it inadvertently.
> 
> I feel like, unless it's used regularly -- and it hasn't been for a few years at this point I guess -- it may be better to remove the feature altogether because it definitely has maintenance costs. Can always add it back when we need it, and I'm not sure fixing it now when we don't is a good way to use our time.

[...]

PR #4199, opened 2025-03-01
chore(deps): update opentelemetry

rami3l:
Supersedes #4182.

Following https://github.com/rust-lang/rustup/pull/4182#issuecomment-2663126604, I realized that since `otel` is not an essential part of our project (it's used in benchmarks for the timeline view only), the following change has brought us no benefits:

> `opentelemetry::global::shutdown_tracer_provider()` is removed.
> Now, you should explicitly call shutdown() on the created tracer provider.
> [..]
> This now makes shutdown consistent across signals.
_https://github.com/open-telemetry/opentelemetry-rust/blob/main/docs/migration_0.28.md#tracing-shutdown-changes_

... so it'd be nice to just stick to the original behavior of `shutdown_tracer_provider()`, which is simply replacing the global tracer provider with `NoopTracerProvider::new()` and dropping the replaced provider immediately:

```rs
pub fn set_tracer_provider<P, T, S>(new_provider: P) -> GlobalTracerProvider
where
    S: trace::Span + Send + Sync + 'static,
    T: trace::Tracer<Span = S> + Send + Sync + 'static,
    P: trace::TracerProvider<Tracer = T> + Send + Sync + 'static,
{
    let mut tracer_provider = GLOBAL_TRACER_PROVIDER
        .write()
        .expect("GLOBAL_TRACER_PROVIDER RwLock poisoned");
    mem::replace(
        &mut *tracer_provider,
        GlobalTracerProvider::new(new_provider),
    )
}

pub fn shutdown_tracer_provider() {
    let mut tracer_provider = GLOBAL_TRACER_PROVIDER
        .write()
        .expect("GLOBAL_TRACER_PROVIDER RwLock poisoned");
    let _ = mem::replace(
        &mut *tracer_provider,
        GlobalTracerProvider::new(NoopTracerProvider::new()),
    );
}
```
_https://github.com/open-telemetry/opentelemetry-rust/blob/99d24b7b8ca25652b955678a1109af0f9d65e242/opentelemetry/src/global/trace.rs#L398-L430_

Regarding the setup/shutdown ergonomics, I've made a simple guard to bring back `(before|after)_test_async()` with minimal efforts, so this has also addressed the test-related part of https://github.com/rust-lang/rustup/issues/4195, although it'd probably still be nice to be able to fetch e.g. logger filter levels directly from within `Process`, so the remainder of that issue stays valid.

### DIFF
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -153,12 +153,11 @@ where
 fn telemetry_default_tracer() -> Tracer {
     use std::time::Duration;
 
-    use opentelemetry::{KeyValue, global, trace::TracerProvider as _};
+    use opentelemetry::{global, trace::TracerProvider as _};
     use opentelemetry_otlp::WithExportConfig;
     use opentelemetry_sdk::{
         Resource,
-        runtime::Tokio,
-        trace::{Sampler, TracerProvider},
+        trace::{Sampler, SdkTracerProvider},
     };
 
     let exporter = opentelemetry_otlp::SpanExporter::builder()
@@ -167,10 +166,10 @@ fn telemetry_default_tracer() -> Tracer {
         .build()
         .unwrap();
 
-    let provider = TracerProvider::builder()
+    let provider = SdkTracerProvider::builder()
         .with_sampler(Sampler::AlwaysOn)
-        .with_resource(Resource::new(vec![KeyValue::new("service.name", "rustup")]))
-        .with_batch_exporter(exporter, Tokio)
+        .with_resource(Resource::builder().with_service_name("rustup").build())
+        .with_batch_exporter(exporter)
         .build();
 
     global::set_tracer_provider(provider.clone());
--- a/src/cli/log.rs
+++ b/src/cli/log.rs
@@ -175,3 +175,27 @@ fn telemetry_default_tracer() -> Tracer {
     global::set_tracer_provider(provider.clone());
     provider.tracer("tracing-otel-subscriber")
 }
+
+#[cfg(feature = "otel")]
+#[must_use]
+pub struct GlobalTelemetryGuard {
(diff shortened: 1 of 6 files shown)


src/cli/log.rs
```diff
@@ -175,3 +175,35 @@ fn telemetry_default_tracer() -> Tracer {
...
     provider.tracer("tracing-otel-subscriber")
 }
+
+#[cfg(feature = "otel")]
+pub struct OtelGlobalGuard {
+    _private: (),
+}
+
+#[cfg(feature = "otel")]
+impl OtelGlobalGuard {
```
djc: Nit: I don't see much point in having both `new()` and a `Default` impl. Given that the construction is manipulating some global state, maybe drop the `Default` impl and rename `new()` to `install()` or something? Maybe also change the name to better reflect what it does?

djc (approved):
Good stuff!

src/cli/log.rs
rami3l: @djc That's https://rust-lang.github.io/rust-clippy/master/#new_without_default. I'm okay with making an exception here.

OTOH `install()` looks like we are directly manipulating the global state rather than doing so in a scope. I agree that `new()` doesn't sound nice either. I'll try to come up with something else.

rami3l: @djc Renamed to `log::set_global_telemetry()` and `log::GlobalTelemetryGuard` to stay closer to the original naming logic ("telemetry" instead of "otel"): an associated function on `*Guard` does look weird to me.

rami3l: @djc If you are okay with this change, then I think we're ready to go. Ideally this should be the final non-trivial change in the 1.28.0 milestone.

PR #4175, opened 2025-02-03
Warn about `CARGO` environment variable in cargo proxy

smoelius:
Addresses rust-lang/clippy#15099

To recount the problem (copied from https://github.com/rust-lang/cargo/issues/15099#issuecomment-2621296667):
- The `cargo` proxy runs, selects the default `cargo`, and sets the `CARGO` environment variable to default toolchain's `cargo`.
- That `cargo` invocations runs some other code, which invokes `cargo +<toolchain> <subcommand>`.
- `<subcommand>` reads the `CARGO` environment variable and runs the default toolchain's `cargo` rather than `<toolchain>`'s `cargo`.

This PR addresses the problem by warning when the  `CARGO` environment variable does not match what `cargo +<toolchain> <subcommand>` will run.

This PR is currently organized as eight commits:
- The first commit applies the fix.
- The next five commits add infrastructure to support the seventh commit.
- The seventh commit adds a test to address all scenarios in https://github.com/rust-lang/cargo/issues/15099#issuecomment-2624230906.
- The eighth commit is a small bug fix to `Config::cmd`.
- The ninth commit incorporates feedback from https://github.com/rust-lang/cargo/issues/15099#issuecomment-2632618041 and onward.

Nits are welcome.

PR #4200, opened 2025-03-01
Do not append `EXE_SUFFIX` in `Config::cmd`

smoelius:
Previously, `Config::cmd` would always append an `EXE_SUFFIX`. However, this would produce an invalid path if the suffix was already present.

Extracted from #4175

### DIFF
--- a/src/test/mock/clitools.rs
+++ b/src/test/mock/clitools.rs
@@ -537,7 +537,7 @@ impl Config {
         I: IntoIterator<Item = A>,
         A: AsRef<OsStr>,
     {
-        let exe_path = self.exedir.join(format!("{name}{EXE_SUFFIX}"));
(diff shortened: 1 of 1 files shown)


rami3l:
## Note

See [`2d94925` (#4175)](https://github.com/rust-lang/rustup/pull/4175/commits/2d949254be4d98e5cd728b4d1792a8d5e9a5e574) for the full context: it's about extracting a binary path from elsewhere and running it in `Config::cmd()`.

I believe this is something that we should support and, on Windows, it's a real bug fix, since IIRC it does support `foo` and `foo.exe` as valid references to `foo.exe` (all our previous tests called it `foo` which at this step will become `foo.exe`), but `foo.exe.exe` is a hard error.

rami3l (approved):

PR #4201, opened 2025-03-01
Add `TryFrom<Output>` for `SanitizedOutput`

smoelius:
`SanitizedOutput`s have `String` `stdout` and `stderr` fields. This makes `SanitizedOutput`s easier to display than `Output`s.

This commit adds a `TryFrom<Output>` implementation for `SanitizedOutput`. In this way, a test can run a command to produce an `Output`, and from that `Output` produce a `SanitizedOutput`.

Extracted from #4175


rami3l (approved):
Thanks a lot :)

PR #4202, opened 2025-03-01
Add `Config::run_subprocess_cmd`

smoelius:
The function `Config::run_subprocess` takes a program name and a list of arguments, constructs a command from them, and then runs the command.

The function this commit adds is similar, but it accepts an already constructed command.

Extracted from #4175

### DIFF
--- a/src/test/mock/clitools.rs
+++ b/src/test/mock/clitools.rs
@@ -839,7 +839,10 @@ impl Config {
         for env in env {
             cmd.env(env.0, env.1);
         }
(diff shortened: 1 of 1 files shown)


rami3l (dismissed):
Thanks for splitting all those PRs!

smoelius:
I won't be insulted if you reject this. Thank you for merging the other three. :)

djc:
Yeah, by itself this doesn't seem a clear improvement. Thanks for your contributions, though!

PR #4203, opened 2025-03-01
Have mocked cargo better adhere to cargo conventions

smoelius:
Currently, when the mocked cargo command is passed `--recursive-cargo-subcommand`, it runs:

```
cargo-foo --recursive-cargo
```

However, cargo subcommands are normally passed their subcommand name as the first argument. Thus, one would expect the following to be run:

```
cargo-foo foo --recursive-cargo
```

This commit changes the mocked cargo command to do the latter. It also adds a check to ensure that the "subcommand name as first argument" behavior is respected.

This is, admittedly, a rather pedantic change.

Extracted from #4175

### DIFF
--- a/src/test/mock/mock_bin_src.rs
+++ b/src/test/mock/mock_bin_src.rs
@@ -5,7 +5,16 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


rami3l (approved):
LGTM :)

PR #4207, opened 2025-03-02
Test module cleanup

djc:


src/test/clitools.rs
```diff
@@ -474,7 +474,7 @@ struct ConstState {
 /// environments. In doing this we can ensure that new test environment creation
 /// does not result in ETXTBSY because the FDs in question happen to be in
 /// newly `fork()`d but not yet `exec()`d subprocesses of other tests.
-pub static CMD_LOCK: LazyLock<RwLock<usize>> = LazyLock::new(|| RwLock::new(0));
```
rami3l, line 138: Following the convention that each commit message starts with a verb, I assume you meant "privatize".

src/test/clitools.rs
```diff
@@ -445,6 +445,127 @@ pub enum Scenario {
     MissingComponentMulti,
 }
 
+impl Scenario {
+    // Creates a mock dist server populated with some test data
+    #[tracing::instrument(level = "trace", skip_all)]
+    fn write(&self, path: &Path) {
```
rami3l: Nit: Naming convention-wise it looks like `fn write(&self, stuff)` is writing something into `self` (see this [example in `std`](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write)). Since it's the opposite case here, I propose changing the name to, say, `.write_to(path)`.

rami3l (approved):
Thanks a whole lot for this impressive amount of cleanup work :D

src/test/clitools.rs
djc: Yup!

src/test/clitools.rs
djc: Good point, I considered the same change but failed to follow through.

src/test/clitools.rs
```diff
@@ -445,6 +445,127 @@ pub enum Scenario {
     MissingComponentMulti,
 }
 
+impl Scenario {
```
rami3l, line 456: ... just fix this outdate commit message and we are ready to go.

PR #4214, opened 2025-03-04
fix!(config): re-enable active toolchain installation in `Cfg::find_active_toolchain()` with optional opt-out

rami3l:
> [!NOTE]
> Following @djc's previous statement at https://github.com/rust-lang/rustup/issues/4211#issuecomment-2695620307, I sincerely apologize for all the previous issues related to the abrupt breaking change without enough clarity and transparency, and, if a well-organized public discussion allows this change to happen, a potential experimental stage with a properly-extended migration period.

This PR tries to mitigate #4211 by introducing a new environment variable `RUSTUP_AUTO_INSTALL` that defaults to `1` ~~and is subsumed under `CI`~~. This environment variable, when evaluated to `true`, will try to install the active toolchain in `Cfg::find_active_toolchain()`.

@rust-lang/rustup please feel free to change this PR as you see fit so that we can land the fix as promptly as possible.

## Concerns

Remaining work around this PR might include:
- [ ] Introducing a proper warning message about the originally-planned change in behavior;
- [x] Making sure that the tests pass properly;
- [ ] Ensuring that both the old way and the new way are working as intended.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -516,15 +516,63 @@ impl<'a> Cfg<'a> {
 
     pub(crate) async fn find_active_toolchain(
         &self,
+        force_install_active: Option<bool>,
     ) -> Result<Option<(LocalToolchainName, ActiveReason)>> {
-        Ok(
-            if let Some((override_config, reason)) = self.find_override_config()? {
-                Some((override_config.into_local_toolchain_name(), reason))
-            } else {
-                self.get_default()?
-                    .map(|x| (x.into(), ActiveReason::Default))
-            },
-        )
+        let (components, targets, profile, toolchain, reason) = match self.find_override_config()? {
+            Some((
+                OverrideCfg::Official {
+                    components,
+                    targets,
+                    profile,
+                    toolchain,
+                },
+                reason,
+            )) => (components, targets, profile, toolchain, reason),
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -511,20 +511,69 @@ impl<'a> Cfg<'a> {
...
+                    profile,
+                    ref toolchain,
+                },
+                reason,
+            )) => {
+                let should_install_active = force_install_active.unwrap_or_else(|| {
+                    self.process
+                        .var("RUSTUP_AUTO_INSTALL")
+                        .map_or(true, |it| it != "0")
+                        || self.process.var("CI").is_ok() && self.process.var("RUSTUP_CI").is_err()
```
ChrisDenton:
```suggestion
```

Let's not do this for now. The `RUSTUP_AUTO_INSTALL` env being set to `true` if missing is enough for the initial fix and in any case the env, if set, should probably override any implicit behaviours.

ChrisDenton (approved):
This looks great to me. There are things I could quibble with but given the circumstances I don't think they need to be addressed in this PR.

djc:
I've split this into two commits to make it a little easier to review, and have made some stylistic tweaks. LGTM!

djc (approved):

PR #4230, opened 2025-03-05
Correct Nushell source command

Yethal:
Nushell uses `$env.HOME` instead of `$HOME` syntax for environment variables. However since `source` is a parser keyword it cannot access environment variables which is why it should use `$nu.home-path` instead
Additionally, the original command will cause nushell to error out on next config evaluation

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -33,7 +33,7 @@ RUSTUP_QUIET=no
 # NOTICE: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat <<EOF
-rustup-init 1.28.1 (83ac5d9ec 2025-03-04)
+rustup-init 1.28.1 (bb9441b61 2025-03-05)
 
 The installer for rustup
 
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -33,7 +33,7 @@ RUSTUP_QUIET=no
 # NOTICE: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat <<EOF
-rustup-init 1.28.1 (bb9441b61 2025-03-05)
+rustup-init 1.28.1 (83ac5d9ec 2025-03-04)
 
 The installer for rustup
 
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -382,9 +382,9 @@ macro_rules! post_install_msg_unix_source_env {
 the corresponding `env` file under {cargo_home}.
 
 This is usually done by running one of the following (note the leading DOT):
-    . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
-    source "{cargo_home}/env.fish"  # For fish
-    source "{cargo_home}/env.nu"    # For nushell
+    . "{cargo_home}/env"                       # For sh/bash/zsh/ash/dash/pdksh
+    source "{cargo_home}/env.fish"             # For fish
+    source $"($nu.home-path)/.cargo/env.nu"    # For nushell
 "#
     };
 }
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -299,7 +299,8 @@ impl UnixShell for Nu {
     }
 
     fn source_string(&self, process: &Process) -> Result<String> {
-        Ok(format!(r#"source "{}/env.nu""#, cargo_home_str(process)?))
+        let cargo_home_nushell = (cargo_home_str(process)?).replace("$HOME", "($nu.home-path)");
+        Ok(format!(r#"source $"{}/env.nu""#, cargo_home_nushell))
     }
 }
 


djc (approved):

src/cli/self_update.rs
```diff
@@ -382,9 +382,9 @@ macro_rules! post_install_msg_unix_source_env {
 the corresponding `env` file under {cargo_home}.
 
 This is usually done by running one of the following (note the leading DOT):
-    . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
-    source "{cargo_home}/env.fish"  # For fish
-    source "{cargo_home}/env.nu"    # For nushell
+    . "{cargo_home}/env"                       # For sh/bash/zsh/ash/dash/pdksh
+    source "{cargo_home}/env.fish"             # For fish
+    source $"($nu.home-path)/.cargo/env.nu"    # For nushell
```
rami3l, line 387: Note that this is not necessary the `$CARGO_HOME`! It's just its default value...

rami3l: AIUI if `$CARGO_HOME` is something else then this will get out of sync with other commands.

rami3l (changes requested):
AIUI if `$CARGO_HOME` is something else then this will get out of sync with other commands.

I believe we should add a more friendly note instead of just using the default value, if the equivalent cannot be achieved in nushell.

rami3l:
It looks like we need something like https://www.nushell.sh/cookbook/foreign_shell_scripts.html but I'm not so sure... cc @LittleJianCH

src/cli/self_update.rs
Yethal: Is there a scenario where $CARGO_HOME is set to a value different than default on a system and a human user will actually read this prompt? Or is it something that's only overriden in CI and such?

rami3l: @Yethal A human user will actually read the prompt. We allow `$CARGO_HOME` to be set to something else beforehand, actually IIRC we have suggested in an earlier note (before the actual installation, see https://github.com/search?q=repo%3Arust-lang%2Frustup+%22This+can+be+modified+with+the%22&type=code) that if you want it to be elsewhere you can set that and return to the installer.

Yethal: So will it suffice to add a note that if `$CARGO_HOME` was overriden then the path will need to be adjusted accordingly?

rami3l: @Yethal I am still inclined to doing this properly.

For example, you probably can get inspired from this logic:

https://github.com/rust-lang/rustup/blob/21d4b2777f5d77f9621912208cea29062fbefaf6/src/cli/self_update.rs#L469-L487

... and figure out a proper way to print the exact path for `env.nu` for nushell. Maybe instead of using `{cargo_home}`, you can add an extra format argument `{cargo_home_nushell}`.

The only detail I don't know is how `$nu.home-path` is defined in nushell. However the spirit stays the same: if dirname is `($nu.home-path)/.cargo` you can print `source $"($nu.home-path)/.cargo/env.nu"`, otherwise just print out the full path.

rami3l: Ok I think I have got the full picture. Nushell uses https://docs.rs/dirs/latest/dirs/fn.home_dir.html meanwhile we use https://docs.rs/home/latest/home/fn.home_dir.html for the `$HOME` definition, so it's a perfect match on Unix, meaning you can possibly issue `cargo_home_nushell = cargo_home.replace("$HOME", "($nu.home-path)")` and call it a day (of course, you'll have to adjust the template string accordingly to include that initial `$` as well).

However, I'm not so sure about Windows as the former uses `{FOLDERID_Profile}` where as the latter uses `%USERPROFILE%` and tries to fall back on using `{FOLDERID_Profile}`. Would that cause inaccurate results?

cc @ChrisDenton

Yethal: Afaict Rustup doesn't support shell customization on Windows anyway.

rami3l: @Yethal I see what you mean, however there is a slight chance that the user has manually set `%USERPROFILE` to something else. Maybe a special branch needs to be added in that case.

rustup-init.sh
```diff
@@ -33,7 +33,7 @@ RUSTUP_QUIET=no
 # NOTICE: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat <<EOF
-rustup-init 1.28.1 (83ac5d9ec 2025-03-04)
```
rami3l: You don't need to update this file: it is part of our release process.

src/cli/self_update.rs
Yethal: If user has manually set `%USERPROFILE%` to something other than their profile then I'd say it's on them if the prompt is incorrect

ChrisDenton: Might this be considered a nushell bug? I think they could use `std::env::home_dir` since the implementation was fixed in 1.85 (though it won't be undeprecated until 1.87).

rustup-init.sh
Yethal: I reverted that

src/cli/self_update.rs
Yethal: What, the fact that environment variables are not parse-time safe? To the best of my knowledge this is by design.

rami3l: > What, the fact that environment variables are not parse-time safe? To the best of my knowledge this is by design.

I believe @ChrisDenton meant that nushell relying on [dirs/fn.home_dir.html](https://docs.rs/dirs/latest/dirs/fn.home_dir.html)'s behavior could be considered a bug in this case... So you are basically free to ignore the "`%USERPROFILE` set to something else" case for now.

ChrisDenton: No, I mean the fact that it only uses `{FOLDERID_Profile}` and doesn't try `%USERPROFILE%` first. That would at least mean the behaviour is consistent.

Yethal: If the behavior was fixed in 1.85 then it might be too early for Nushell since the policy is to stay two versions behind latest stable

alluring-mushroom:
@rami3l I don't think you need the tricks in that link. The canonical way in nushell to resolve dynamic solutions is to place the code that *generates* in `nushell/env.nu` and the code that *sources* in `nushell/config.nu`. For example, most completion scripts work like this.

Yethal:
An alternative approach would be to place the file in one of nushell's autoload directories so it could be sourced by name only with no relative path

rami3l:
> @rami3l I don't think you need the tricks in that link. The canonical way in nushell to resolve dynamic solutions is to place the code that *generates* in `nushell/env.nu` and the code that *sources* in `nushell/config.nu`. For example, most completion scripts work like this.

@alluring-mushroom I agree, and I now think plain substring substitutions would be fine: it's mostly about pretty-printing the absolute path by truncating it after all.

Yethal:
@rami3l What else do I need to do to get this merged and/or the bug being resolved in any other way?

rami3l:
@Yethal Sorry for the late reply. I haven't forgotten about this PR, it's just that I've been busy with other stuff for a while. That said, I'd definitely want it to become part of v1.28.2.

After a second look, I think although I suggested a string replacement previously, that replacement should not happen when printing the note. Instead, it should happen in the form of a refactored `cargo_home()` function that either outputs a path with `HOME`/`USERPROFILE` or one with the correct nushell-type of path.

Anyway, the overall outcome I'd like to see is:
- Accurate hints (modulo the `HOME` difference on Windows as discussed above)
- Use the full path if and only if necessary

PS: If you think that's too much for you, I can offer to fix this patch with your consent.

rami3l:
So, as a quick summary, this `$HOME`-adding logic happens in two places:

1. https://github.com/rust-lang/rustup/blob/364f72a22f279abfa33e333fe45644210dd0c0ec/src/cli/self_update.rs#L469
  However this doesn't affect any note on Windows, so we can safely replace that note for Unix only.

2. https://github.com/rust-lang/rustup/blob/364f72a22f279abfa33e333fe45644210dd0c0ec/src/cli/self_update/shell.rs#L54
  This has more impact as it actually influences two places:
  i. https://github.com/rust-lang/rustup/blob/364f72a22f279abfa33e333fe45644210dd0c0ec/src/cli/self_update/shell.rs#L42-L51
  ii. https://github.com/rust-lang/rustup/blob/364f72a22f279abfa33e333fe45644210dd0c0ec/src/cli/self_update/shell.rs#L301-L303

  ... and as we must consider Windows in such scenarios now, I suggest using the full paths for nushell and don't bother with `cargo_home_str()` any more.

rami3l:
@Yethal I understand that the command sourcing `env.nu` is wrong. However, does the generated `env.nu` work for you, or it needs correction as well?

Yethal:
> @Yethal I understand that the command sourcing `env.nu` is wrong. However, does the generated `env.nu` work for you, or it needs correction as well?

Opposite actually, command sourcing is ok but the generated env.nu is incorrect

rami3l:
Closing in favor of #4265.

PR #4244, opened 2025-03-10
Allow copy&paste of documented registry keys into Regedit

eugenesvk:
Current documented registry path can't be easily used as a lookup in the default Windows Regedit registry editing tool
  - Forward slashes are not accepted by Regedit
  - nor do the full paths that include registry keys, only their parent address

This edits the message format so you could copy&paste

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -360,8 +360,8 @@ modifying the profile file{plural} located at:
 macro_rules! pre_install_msg_win {
     () => {
         pre_install_msg_template!(
-            "This path will then be added to your `PATH` environment variable by
-modifying the `HKEY_CURRENT_USER/Environment/PATH` registry key."
+            r#"This path will then be added to your `PATH` environment variable by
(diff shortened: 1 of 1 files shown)


src/cli/self_update.rs
```diff
@@ -361,7 +361,7 @@ macro_rules! pre_install_msg_win {
     () => {
         pre_install_msg_template!(
             "This path will then be added to your `PATH` environment variable by
-modifying the `HKEY_CURRENT_USER/Environment/PATH` registry key."
+modifying the `PATH` registry key at `HKEY_CURRENT_USER\\Environment`."
```
rami3l: Nit: Maybe use a raw string here so that the `\` looks clearer?

rami3l (approved):

src/cli/self_update.rs
eugenesvk: changed

djc (approved):

PR #4247, opened 2025-03-11
document RUSTUP_AUTO_INSTALL

ahl:
I noticed that `RUSTUP_AUTO_INSTALL` was absent from the docs. I wasn't sure of the ordering of the entries in this file so I just put it at the end. And if the specific documentation isn't right, feel free to edit.

Also! if this is just an unneeded PR, please feel free to close! Thanks!


djc (approved):
Thanks, this seems good!

rami3l (approved):

PR #4250, opened 2025-03-12
Remove validation for custom toolchains when reading rust-toolchain.toml

### LINKED ISSUE #4248: Reconsider the "toolchain options are ignored for a custom toolchain" error

The rustup 1.28 release [introduced new validation](https://github.com/rust-lang/rustup/commit/e1306b390820826d967f3bf367cb8b83ab1b16cd#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR226-R237) for custom toolchains listed in `rust-toolchain.toml` files. At Microsoft, we use an internal build of the Rust compiler that is distributed as a custom toolchain (not managed by rustup but linked to from `~/.rustup/toolchains`) so it's common for our developers to have `rust-toolchain.toml` files that look like:

```toml
[toolchain]
channel = "ms-1.84.1"
targets = [
    "aarch64-pc-windows-msvc",
    "x86_64-pc-windows-msvc",
]
```

We have tooling that handles installing the Rust build and appropriate targets. For our developers, this is nice as it aligns very closely to working with the standard Rust toolchain build, which we want to support as much as possible.

With the update to rustup 1.28 however, any invocation of `rustc`, `cargo`, etc is met with:

```
> cargo --version
error: toolchain options are ignored for a custom toolchain (ms-1.84.1)
```

even when the appropriate custom toolchain is already configured and ready to be used. 

### Solution you'd like

Ideally, it would be great if this error could only be shown in the case where required components or targets are missing from the custom toolchain! 

I understand that is probably difficult to determine ahead of time so some other solutions might be:

1. Revert the additional validation for custom toolchain options.
    - If this is acceptable, I would be happy to prepare a PR with this change.
1. Allow the user to ignore this error entirely somehow (preferably via a configuration flag in `~/.rustup/settings.toml` or an environment variable).
1. Downgrade the error to a warning or informational message.
   - This is not highly preferred because it will still appear frequently to our developers while not being useful/actionable. 

Thanks for your consideration and efforts to improve rustup! 

### Notes

_No response_

wesleywiser:
This validation was introduced in e1306b390820826d967f3bf367cb8b83ab1b16cd However, it breaks a number of scenarios that previously worked without issue such as #4248. As requested by rustup maintainers in that thread, I'm taking out the extra validation added in that commit, leaving the rest of the code as-is.

Testing confirms this fixes #4248.

### DIFF
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -2954,3 +2954,70 @@ async fn warn_on_duplicate_rust_toolchain_file() {
         )
         .await;
 }
+
+#[tokio::test]
+async fn custom_toolchain_with_components_toolchains_profile_does_not_err() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+
+    let cwd = cx.config.current_dir();
+    let toolchain_file = cwd.join("rust-toolchain.toml");
+
+    // install a toolchain so we can make a custom toolchain that links to it
+    cx.config
+        .expect_stderr_ok(
+            &[
+                "rustup",
+                "toolchain",
+                "install",
+                "nightly",
+                "--profile=minimal",
+                "--component=cargo",
+            ],
+            for_host!(
+                "\
+info: syncing channel updates for 'nightly-{0}'
+info: latest update on 2015-01-02, rust version 1.3.0 (hash-nightly-2)
+info: downloading component 'cargo'
+info: downloading component 'rustc'
+info: installing component 'cargo'
+info: installing component 'rustc'
+info: default toolchain set to 'nightly-{0}'"
+            ),
+        )
+        .await;
+
+    // link the toolchain
+    let toolchains = cx.config.rustupdir.join("toolchains");
+    raw::symlink_dir(
+        &toolchains.join(for_host!("nightly-{0}")),
+        &toolchains.join("my-custom"),
+    )
+    .expect("failed to symlink");
+
+    raw::write_file(
+        &toolchain_file,
+        r#"
+[toolchain]
+channel = "my-custom"
+components = ["rustc-dev"]
+targets = ["x86_64-unknown-linux-gnu"]
+profile = "minimal"
+"#,
+    )
+    .unwrap();
+
+    cx.config
+        .expect_stdout_ok(
+            &["rustup", "show", "active-toolchain"],
+            &format!("my-custom (overridden by '{0}')", toolchain_file.display(),),
+        )
+        .await;
+
+    cx.config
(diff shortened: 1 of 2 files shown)


djc (approved):
Makes sense to me. Will wait for @rami3l to have a look.

I guess alternatively we could show a warning, and then store a setting that we've shown the warning, or something? (Or, store the date when we've last shown the warning and only show it again after x days?)

rami3l:
@wesleywiser Thanks a lot for your comprehension and your quick patch!

> I was not able to figure out a way to write an automated test for this scenario since the rustup CLI needs the custom toolchain to actually exist and unit testing looked infeasible due to use of the `Cfg` type.

Sorry for the back and forth, but I would still like to see an automated test for this scenario. Is it possible to trick rustup into believing the existence of that toolchain by installing e.g. `stable` with the components you want, and *rename* it  directly under `$RUSTUP_HOME/toolchains` (written as `cx.config.rustupdir.join("toolchains")` in our test suites)?

wesleywiser:
Thanks @rami3l for that suggestion! I've added a test that covers this 🙂

tests/suite/cli_rustup.rs
```diff
@@ -2954,3 +2954,68 @@ async fn warn_on_duplicate_rust_toolchain_file() {
...
+async fn custom_toolchain_with_components_toolchains_profile_does_not_err() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+
+    let cwd = cx.config.current_dir();
+    let toolchain_file = cwd.join("rust-toolchain.toml");
+
+    // install a toolchain so we can make a custom toolchain that links to it
+    cx.config
+        .expect_stderr_ok(
+            &["rustup", "toolchain", "install", "nightly"],
```
rami3l: Maybe we can use `--profile=minimal` to save some time here?

tests/suite/cli_rustup.rs
```diff
@@ -2954,3 +2954,68 @@ async fn warn_on_duplicate_rust_toolchain_file() {
...
+            &format!(
```
rami3l: We have an extra `for_host!()` macro here. It's far from idiomatic but using consistent APIs will help us with further evolution of our testing code (idem for other places using the same pattern). Thanks!

rami3l (approved):

rami3l:
@wesleywiser About the message that @djc has mentioned, I think it will probably be worthwhile to add a warning clarifying that we are not doing any checks, as demonstrated in the new test. That said, it will definitely not be a blocker anyway. Thanks again for your contribution!

@djc I'm afraid a "last emitted warning" check might not be enough now that we are considering some workstation with multiple repos/toolchain combinations...

PR #4254, opened 2025-03-14
Add `set auto-install disable`

ChrisDenton:
The rustup setting equivalent to the environment variable `RUSTUP_AUTO_INSTALL`.

The environment variable will take precedence if it's set, otherwise the global setting will be used.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -10,6 +10,7 @@ use thiserror::Error as ThisError;
 use tokio_stream::StreamExt;
 use tracing::trace;
 
+use crate::dist::AutoInstallMode;
 use crate::{
     cli::{common, self_update::SelfUpdateMode},
     dist::{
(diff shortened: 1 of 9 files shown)


src/config.rs
```diff
@@ -375,6 +376,29 @@ impl<'a> Cfg<'a> {
...
+        self.settings_file.with_mut(|s| {
+            s.auto_install = Some(mode);
+            Ok(())
+        })?;
+        (self.notify_handler)(Notification::SetAutoInstall(mode.as_str()));
+        Ok(())
+    }
+
+    pub(crate) fn should_auto_install(&self) -> Result<bool> {
+        // If auto install is disabled in settings then the environment variable
```
rami3l: Just to be sure, is this coherent with what we have for now? Somehow I'd expect the env var to have the higher precedence here...

https://github.com/rust-lang/rustup/blob/35d97428d259fcc3dabf1dd1c10b65ae5c0d0be2/doc/user-guide/src/overrides.md?plain=1#L7-L12

ChrisDenton: Yeah, I wasn't sure about that. My thought was to make it act like 1.28.0 (where the environment var didn't exist). But rethinking it, I've changed my mind and now I think the environment var should take precedence as it is more consistent.

ChrisDenton: Ok, done.

src/config.rs
```diff
@@ -375,6 +376,24 @@ impl<'a> Cfg<'a> {
...
+            s.auto_install = Some(mode);
+            Ok(())
+        })?;
+        (self.notify_handler)(Notification::SetAutoInstall(mode.as_str()));
+        Ok(())
+    }
+
+    pub(crate) fn should_auto_install(&self) -> Result<bool> {
+        if let Ok(mode) = self.process.var("RUSTUP_AUTO_INSTALL") {
+            Ok(mode != "0")
```
ChrisDenton, line 390: Should I be smarter here? So if auto-install is disabled then it checks `mode != "1"` instead of `mode != "0"`. Basically to ensure it defaults to the setting if the environment variable is not recognised.

ChrisDenton: On the other hand, that would be more awkward. I'm just worried about `RUSTUP_AUTO_INSTALL=false` having surprising behaviour if auto-install is disabled in the settings.

rami3l: @ChrisDenton I understand your concerns, idiomatically we should have used something like `str2bool()`, yes. However since we're documenting `1` and `0` ONLY, I personally don't think that would be that much of a problem here. In the past, I believe, the convention is that, if a flag is enabled by default, the only way to disable it is to set it to a `0`.

rami3l (approved):
Nice work :)

src/config.rs
ChrisDenton: Fair enough!

PR #4255, opened 2025-03-14
Use `cursor: pointer` for copy button on website

Noratrieb:
It's clickable, so it should look clickable as well.


ChrisDenton (approved):
Sure, makes sense to me.

PR #4258, opened 2025-03-18
fix(errors)!: improve error messages for `RustupError::ToolchainNotInstalled`

### LINKED ISSUE #4212: Recommended `toolchain install` command line isn't exactly right
### Verification

### Problem

In the behavior introduced in the latest release, the error message if you don't have the toolchain installed is not actually correct if your rust-toolchain.toml file formerly triggered the installation of more than just a toolchain. This is common in cross-compile situations with the `targets` key.

rustup 1.28 tells me:

```
[cbiffle@gwydion soundlaser-rs]$ cargo build
error: toolchain '1.85.0-x86_64-unknown-linux-gnu' is not installed
help: run `rustup toolchain install 1.85.0-x86_64-unknown-linux-gnu` to install it
```

...but that's not actually correct, because that does not install the `targets` listed in the toml. `cargo build` fails.

On the other hand, `rustup toolchain install` with no arguments does the right thing.

### Steps

1. Update rustup to 1.28.
2. Check out, say, https://github.com/cbiffle/soundlaser-rs
3. Make sure you don't have 1.85.0 installed.
4. Run `cargo build`.
5. Now, run the exact command from the error that displays.
6. Now, try `cargo build` again.

### Possible Solution(s)

In the event that the command is failing in a directory with a (valid) `rust-toolchain.toml`, perhaps the error message should suggest the no-argument version of `rustup toolchain install`.

### Notes

_No response_

### Rustup version

```console
rustup 1.28.0 (6e19fbec7 2025-03-02)
info: The currently active `rustc` version is `rustc 1.85.0 (4d91de4e4 2025-02-17)`
```

### Installed toolchains

```console
Many.
```

### OS version

```console
Linux gwydion 6.9.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 26 May 2024 01:30:29 +0000 x86_64 GNU/Linux
```

rami3l:
Closes #4212.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -763,7 +763,7 @@ async fn default_(
             }
         };
 
-        if let Some((toolchain, reason)) = cfg.maybe_ensure_active_toolchain(None).await? {
+        if let Some((toolchain, reason)) = cfg.active_toolchain()? {
             if !matches!(reason, ActiveReason::Default) {
                 info!("note that the toolchain '{toolchain}' is currently in use ({reason})");
             }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -899,9 +899,7 @@ async fn update(
             exit_code &= common::self_update(|| Ok(()), cfg.process).await?;
         }
     } else if ensure_active_toolchain {
-        let (toolchain, reason) = cfg
-            .find_or_install_active_toolchain(force_non_host, true)
-            .await?;
+        let (toolchain, reason) = cfg.ensure_active_toolchain(force_non_host, true).await?;
         info!("the active toolchain `{toolchain}` has been installed");
         info!("it's active because: {reason}");
     } else {
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -763,7 +763,7 @@ async fn default_(
             }
         };
 
-        if let Some((toolchain, reason)) = cfg.find_active_toolchain(None).await? {
+        if let Some((toolchain, reason)) = cfg.maybe_ensure_active_toolchain(None).await? {
             if !matches!(reason, ActiveReason::Default) {
                 info!("note that the toolchain '{toolchain}' is currently in use ({reason})");
             }
@@ -974,7 +974,7 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
     let installed_toolchains = cfg.list_toolchains()?;
     let active_toolchain_and_reason: Option<(ToolchainName, ActiveReason)> =
         if let Ok(Some((LocalToolchainName::Named(toolchain_name), reason))) =
-            cfg.find_active_toolchain(None).await
+            cfg.maybe_ensure_active_toolchain(None).await
         {
             Some((toolchain_name, reason))
         } else {
@@ -1094,7 +1094,7 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
 
 #[tracing::instrument(level = "trace", skip_all)]
 async fn show_active_toolchain(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
-    match cfg.find_active_toolchain(None).await? {
+    match cfg.maybe_ensure_active_toolchain(None).await? {
         Some((toolchain_name, reason)) => {
             let toolchain = Toolchain::with_reason(cfg, toolchain_name.clone(), &reason)?;
             if verbose {
@@ -1332,7 +1332,7 @@ async fn toolchain_link(
 async fn toolchain_remove(cfg: &mut Cfg<'_>, opts: UninstallOpts) -> Result<utils::ExitCode> {
     let default_toolchain = cfg.get_default().ok().flatten();
     let active_toolchain = cfg
-        .find_active_toolchain(Some(false))
+        .maybe_ensure_active_toolchain(Some(false))
         .await
         .ok()
         .flatten()
--- a/src/config.rs
+++ b/src/config.rs
@@ -526,61 +526,33 @@ impl<'a> Cfg<'a> {
         &self,
         force_install_active: Option<bool>,
     ) -> Result<Option<(LocalToolchainName, ActiveReason)>> {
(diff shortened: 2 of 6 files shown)


src/toolchain.rs
```diff
@@ -107,7 +108,10 @@ impl<'a> Toolchain<'a> {
         let path = cfg.toolchain_path(&name);
         if !Toolchain::exists(cfg, &name)? {
             return Err(match name {
-                LocalToolchainName::Named(name) => RustupError::ToolchainNotInstalled { name },
+                LocalToolchainName::Named(name) => {
+                    let is_active = matches!(cfg.active_toolchain(), Ok(Some((t, _))) if t == name);
```
rami3l, line 112: The following secretly calls `Toolchain::new()`, causing an infinite recursion.

https://github.com/rust-lang/rustup/blob/671b874c943e1f23f69adcf7b906fb1fe1077f27/src/config.rs#L704-L706

src/config.rs
```diff
@@ -526,61 +526,30 @@ impl<'a> Cfg<'a> {
...
-                    distributable
-                        .update(&components, &targets, profile.unwrap_or_default())
-                        .await?;
-                }
-            }
-            Err(e) => return Err(e.into()),
-        };
-
-        Ok(Some(((&toolchain).into(), reason)))
+    pub(crate) fn active_toolchain(&self) -> Result<Option<(LocalToolchainName, ActiveReason)>> {
```
rami3l, line 547: We have to extract a sync function here because v1.28.1 has made the `Cfg::find_active_toolchain()` async again, but we really need to find the active toolchain in a sync context in the commits that follow... (sigh)

rami3l: Also, I suggest renaming `Cfg::find_active_toolchain()` to something else. Any ideas?

src/toolchain.rs
rami3l: Addressed by manually inlining the call and removing unnecessary data dependencies, as stated in [`014a6d0` (#4258)](https://github.com/rust-lang/rustup/pull/4258/commits/014a6d0b8b3a3ba05666daedf04516e1266bb1cd#r2005543171).

src/config.rs
ChrisDenton: To be honest I'm struggling to think of a good name.

src/config.rs
```diff
@@ -526,61 +526,33 @@ impl<'a> Cfg<'a> {
...
-            Ok(mut distributable) => {
-                if !distributable.components_exist(&components, &targets)? {
-                    distributable
-                        .update(&components, &targets, profile.unwrap_or_default())
-                        .await?;
-                }
-            }
-            Err(e) => return Err(e.into()),
-        };
+        match self.find_or_install_active_toolchain(true, false).await {
```
rami3l, line 538: This has also exposed an existing bug in v1.28.1 that `Cfg::find_active_toolchain()` ignores `--force-non-host` at all. Will fix.

src/config.rs
rami3l: @ChrisDenton I'm currently thinking about something like:

| Before | After | Remarks |
|---|---|---|
| find_active_toolchain() | maybe_ensure_active_toolchain() | Dispatches to one of the following (async) |
| find_active_toolchain() (old implementation in 1.28.0) | active_toolchain() | Queries the active toolchain (sync) |
| find_or_install_active_toolchain() | ensure_active_toolchain() | Ensures the installation of the active toolchain (async) |

... how does that sound?

ChrisDenton: That sounds good to me. I think those names do make the relationship between them clearer to me at least.

src/cli/rustup_mode.rs
```diff
@@ -763,7 +763,7 @@ async fn default_(
             }
         };
 
-        if let Some((toolchain, reason)) = cfg.maybe_ensure_active_toolchain(None).await? {
+        if let Some((toolchain, reason)) = cfg.active_toolchain()? {
```
rami3l, line 766: Since we're just emitting an `info!()`, there's no need to install anything.

ChrisDenton (approved):
lgtm, thanks!

src/config.rs
```diff
@@ -526,61 +526,33 @@ impl<'a> Cfg<'a> {
...
-
         let should_install_active = if let Some(force) = force_install_active {
             force
         } else {
             self.should_auto_install()?
         };
-
         if !should_install_active {
-            return Ok(Some(((&toolchain).into(), reason)));
+            return self.active_toolchain();
```
rami3l, line 535: If `should_install_active` is set to `false`, we return `self.active_toolchain()` which was the old behavior of `find_active_toolchain()` in v1.28.0, corresponding to the last 2 early-returning branch of the previous `match` that gets deleted in this change:

https://github.com/rust-lang/rustup/blob/e2d9e7edde6ee1a79e6e44f3b81065443c1fc6eb/src/config.rs#L517-L528

src/config.rs
```diff
@@ -526,61 +526,33 @@ impl<'a> Cfg<'a> {
...
         }
 
-        let components = components.iter().map(AsRef::as_ref).collect::<Vec<_>>();
```
rami3l, line 559: If `should_install` is set to `true`, we delegate the installation to `find_or_install_active_toolchain()` which has the same logic:

https://github.com/rust-lang/rustup/blob/e2d9e7edde6ee1a79e6e44f3b81065443c1fc6eb/src/config.rs#L758-L765
https://github.com/rust-lang/rustup/blob/e2d9e7edde6ee1a79e6e44f3b81065443c1fc6eb/src/config.rs#L813-L844

src/config.rs
```diff
@@ -675,17 +675,13 @@ impl<'a> Cfg<'a> {
 
                     // XXX: this awkwardness deals with settings file being locked already
                     let toolchain_name = toolchain_name.resolve(&default_host_triple)?;
-                    match Toolchain::new(self, (&toolchain_name).into()) {
```
rami3l, line 706: I have manually inlined `Toolchain::new()` here and removed the statement causing an infinite recursion, because it was not needed; in fact, to allow any `RustupError::ToolchainNotInstalled` to happen with this call, the toolchain must not exist:

https://github.com/rust-lang/rustup/blob/f9edccde032ea3109e0b78c891dd74852c0b25ff/src/toolchain.rs#L106-L115

PR #4259, opened 2025-03-19
feat: allow version pinning & arbitrary downgrades via `RUSTUP_VERSION`

### LINKED ISSUE #1158: Make downgrading via the archives work
I expect this to work:

```
RUSTUP_UPDATE_ROOT=https://dev-static.rust-lang.org/rustup/archive/1.2.0 rustup self update
```

but it does not:

```
error: could not download file from 'https://dev-static.rust-lang.org/rustup/archive/1.2.0/release-stable.toml' to '/tmp/rustup-update.6RjOKF0hq6Ji/release-stable.toml'
info: caused by: http request returned an unsuccessful status code: 404
```

I suspect it used to but broke once the 'release-stable.toml' file was added.

Probably just use another environment variable like `RUSTUP_UPDATE_VERSION`, which circumvents the 'release-stable.toml' file and appends the appropriate archive path to the update root.

This could be useful in the case of deploying a bad rustup (where the updater still works).

rami3l:
Closes #1158.

I think the following console log should suffice to explain what this PR does (no tests for the `.sh` script, but works on my machine):

```console
> RUSTUP_VERSION=1.26.0 ./rustup-init.sh --version
info: `RUSTUP_VERSION` has been set to `1.26.0`
info: downloading installer
```

The same environment variable now also applies to `rustup self update`, which does come with a proper test.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1036,9 +1036,12 @@ pub(crate) fn dist_root_server(process: &Process) -> Result<String> {
             s
         }
         // For backwards compatibility
-        else if let Some(root) = non_empty_env_var("RUSTUP_DIST_ROOT", process)? {
+        else if let Some(mut root) = non_empty_env_var("RUSTUP_DIST_ROOT", process)? {
             trace!("`RUSTUP_DIST_ROOT` has been set to `{root}`");
-            root.trim_end_matches("/dist").to_owned()
+            if let Some(stripped) = root.strip_suffix("/dist") {
+                root.truncate(stripped.len());
+            }
+            root
         } else {
             dist::DEFAULT_DIST_SERVER.to_owned()
         },
--- a/src/config.rs
+++ b/src/config.rs
@@ -1027,22 +1027,22 @@ impl<'a> Cfg<'a> {
     }
 }
 
+/// The root path of the release server, without the `/dist` suffix.
+/// By default, it points to [`dist::DEFAULT_DIST_SERVER`].
 pub(crate) fn dist_root_server(process: &Process) -> Result<String> {
-    Ok(match non_empty_env_var("RUSTUP_DIST_SERVER", process)? {
-        Some(s) => {
+    Ok(
+        if let Some(s) = non_empty_env_var("RUSTUP_DIST_SERVER", process)? {
             trace!("`RUSTUP_DIST_SERVER` has been set to `{s}`");
             s
         }
-        None => {
-            // For backward compatibility
-            non_empty_env_var("RUSTUP_DIST_ROOT", process)?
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -1027,22 +1027,19 @@ impl<'a> Cfg<'a> {
...
+        trace!("`RUSTUP_DIST_SERVER` has been set to `{s}`");
+        return Ok(s);
+    }
+    // For backwards compatibility
+    if let Some(root) = non_empty_env_var("RUSTUP_DIST_ROOT", process)? {
+        trace!("`RUSTUP_DIST_ROOT` has been set to `{root}`");
+        return Ok(root.trim_end_matches("/dist").to_owned());
+    }
+    Ok(dist::DEFAULT_DIST_SERVER.to_owned())
 }
```
ChrisDenton: Could this use `else if` and `else`?

rami3l: @ChrisDenton No problem!

ChrisDenton (approved):

src/config.rs
```diff
@@ -1027,22 +1027,22 @@ impl<'a> Cfg<'a> {
...
-                .as_ref()
-                .map(|root| root.trim_end_matches("/dist"))
-                .unwrap_or(dist::DEFAULT_DIST_SERVER)
-                .to_owned()
-        }
-    })
+        // For backwards compatibility
+        else if let Some(root) = non_empty_env_var("RUSTUP_DIST_ROOT", process)? {
+            trace!("`RUSTUP_DIST_ROOT` has been set to `{root}`");
+            root.trim_end_matches("/dist").to_owned()
```
rami3l: `.trim_end_matches()` repeatedly removes the suffix. I guess that's not what the original meant to write?

ChrisDenton: Hm, oh yes that's wrong. I guess it's unlikely to make much difference in practice (`/dist/dist` would be unusual) but that should be fixed.

ChrisDenton: I mean, assuming there's no backwards compatibility reason to keep it behaving the same (and I don't think there is)?

rami3l: @ChrisDenton Looking at the original commit (https://github.com/rust-lang/rustup/commit/b7e2a6ebd4990eb63aa057ee0e001479119d17ee) I believe this was just a mistake in the original implementation.

rami3l: I don't believe many are using it, since that commit was from nearly 10 years ago (so `RUSTUP_DIST_ROOT` has long been deprecated I think). I just checked and it seems to be the only really problematic use of `.trim_*_matches()` in the business logic.

PR #4265, opened 2025-03-24
fix(cli/self-update): fix Nushell-related suggestions and scripts

### LINKED ISSUE #4242: Nushell environment script (`env.nu`) is using the wrong string
### Verification

### Problem

I just installed rustup via `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`

I use nushell, so when I next opened nushell it was very broken and both `nushell/env.nu` and `nushell/config.nu` contained `source "$HOME/.cargo/env.nu`

I fixed this with a `source ~/.cargo/env.nu`. You could equally use something like `source ($nu.home-path | join `.cargo/env.nu)`

cargo still couldn't be found, and I noticed the `.cargo/env.nu` was this:
```nushell
if ("$HOME/.cargo/bin" not-in ($env.Path | split row (char esep))) {
  $env.Path = ($env.Path | prepend "$HOME/.cargo/bin")
}
```

Which resulted in a path of `PATH=/home/user/$HOME/.cargo/bin:/home/user/.local/bin: ...`

I haven't investigated what has caused this yet, but I would be happy to do so a bit later

### Steps

1. Prepare a system with nushell but not cargo
2. Install cargo via rustip (`curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`)
3. Open nushell and observe its errors

### Possible Solution(s)

I assume some string interpolation has changed somewhere?

### Notes

_No response_

### Rustup version

```console
rustup 1.28.1 (f9edccde0 2025-03-05)
info: The currently active `rustc` version is `rustc 1.87.0-nightly (b74da9613 2025-03-06)`
```

### Installed toolchains

```console
rustup home:  /home/user/.rustup

installed toolchains
--------------------
nightly-x86_64-unknown-linux-gnu (active, default)

active toolchain
----------------
name: nightly-x86_64-unknown-linux-gnu
installed targets:
  x86_64-unknown-linux-gnu
```

### OS version

```console
ubuntu 24.04
```

rami3l:
Supersedes #4230. Closes #4242.

PS: Thanks for your contribution, @yethal! I've made you the co-author of ~~two~~ three main commits in this PR.

@alluring-mushroom @yethal thanks again for your report! I don't use Nushell personally and I'm not sure if this has covered all the corner cases (it would be difficult to test automatically due to the amount of different setups we need to support). Would you mind trying out this patch and see if the fix works for you?

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -31,13 +31,13 @@
 //! and racy on Windows.
 
 use std::borrow::Cow;
-use std::env::consts::EXE_SUFFIX;
+use std::env::{self, consts::EXE_SUFFIX};
+use std::fmt;
 use std::fs;
 use std::io::Write;
 use std::path::{Component, MAIN_SEPARATOR, Path, PathBuf};
 use std::process::Command;
 use std::str::FromStr;
-use std::{env, fmt};
 
 use anyhow::{Context, Result, anyhow};
 use cfg_if::cfg_if;
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -77,6 +77,9 @@ use unix::{delete_rustup_and_cargo_home, do_add_to_path, do_remove_from_path};
 #[cfg(unix)]
 pub(crate) use unix::{run_update, self_replace};
 
+#[cfg(unix)]
+use self::shell::{Nu, UnixShell};
+
 #[cfg(windows)]
 mod windows;
 #[cfg(windows)]
@@ -384,7 +387,7 @@ the corresponding `env` file under {cargo_home}.
 This is usually done by running one of the following (note the leading DOT):
     . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
     source "{cargo_home}/env.fish"  # For fish
-    source "{cargo_home}/env.nu"    # For nushell
+    source $"{cargo_home_nushell}/env.nu"  # For nushell
 "#
     };
 }
@@ -578,13 +581,20 @@ pub(crate) async fn install(
         format!(post_install_msg_win!(), cargo_home = cargo_home)
     };
     #[cfg(not(windows))]
+    let cargo_home_nushell = Nu.cargo_home_str(process)?;
+    #[cfg(not(windows))]
     let msg = if no_modify_path {
         format!(
             post_install_msg_unix_no_modify_path!(),
-            cargo_home = cargo_home
+            cargo_home = cargo_home,
+            cargo_home_nushell = cargo_home_nushell,
         )
     } else {
-        format!(post_install_msg_unix!(), cargo_home = cargo_home)
+        format!(
+            post_install_msg_unix!(),
+            cargo_home = cargo_home,
+            cargo_home_nushell = cargo_home_nushell,
+        )
     };
     md(&mut term, msg);
 
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -40,11 +40,6 @@ pub(crate) struct ShellScript {
 }
 
 // TODO: Update into a bytestring.
-pub(crate) fn cargo_home_str(process: &Process) -> Result<Cow<'static, str>> {
-    cargo_home_str_with_home("$HOME", process)
-}
-
-// TODO: Update into a bytestring.
 fn cargo_home_str_with_home(home: &str, process: &Process) -> Result<Cow<'static, str>> {
     let path = process.cargo_home()?;
 
@@ -101,13 +96,17 @@ pub(crate) trait UnixShell {
         }
     }
 
+    fn cargo_home_str(&self, process: &Process) -> Result<Cow<'static, str>> {
+        cargo_home_str_with_home("$HOME", process)
+    }
+
     fn source_string(&self, process: &Process) -> Result<String> {
-        Ok(format!(r#". "{}/env""#, cargo_home_str(process)?))
+        Ok(format!(r#". "{}/env""#, self.cargo_home_str(process)?))
     }
 
     fn write_script(&self, script: &ShellScript, process: &Process) -> Result<()> {
         let home = process.cargo_home()?;
-        let cargo_bin = format!("{}/bin", cargo_home_str(process)?);
+        let cargo_bin = format!("{}/bin", self.cargo_home_str(process)?);
         let env_name = home.join(script.name);
         let env_file = script.content.replace("{cargo_bin}", &cargo_bin);
         utils::write_file(script.name, &env_name, &env_file)?;
@@ -115,7 +114,8 @@ pub(crate) trait UnixShell {
     }
 }
 
-struct Posix;
+pub(super) struct Posix;
+
 impl UnixShell for Posix {
     fn does_exist(&self, _: &Process) -> bool {
         true
@@ -255,7 +255,10 @@ impl UnixShell for Fish {
     }
 
     fn source_string(&self, process: &Process) -> Result<String> {
-        Ok(format!(r#"source "{}/env.fish""#, cargo_home_str(process)?))
+        Ok(format!(
+            r#"source "{}/env.fish""#,
+            self.cargo_home_str(process)?
+        ))
     }
 }
 
@@ -302,7 +305,10 @@ impl UnixShell for Nu {
     }
 
     fn source_string(&self, process: &Process) -> Result<String> {
-        Ok(format!(r#"source "{}/env.nu""#, cargo_home_str(process)?))
+        Ok(format!(
+            r#"source "{}/env.nu""#,
+            self.cargo_home_str(process)?
+        ))
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/cli/self_update/shell.rs
```diff
@@ -256,7 +260,20 @@ impl UnixShell for Fish {
     }
 }
 
-struct Nu;
+pub(super) struct Nu;
+
+impl Nu {
+    pub(super) fn translate_cargo_home(cargo_home: &mut String) {
```
rami3l: It looks like `cargo_home_str` should probably be a method in the `UnixShell` trait with a default implementation.

rami3l: According to @alluring-mushroom's report, it could be that the content of `env.nu` is also wrong because it contains `$HOME`: the `UnixShell` trait might need a more thorough rework.

Yethal:
@rami3l Just ran this in a VM
Preinstall prompt:
```
This path will then be added to your PATH environment variable by
modifying the profile files located at:

  /home/vagrant/.profile
  /home/vagrant/.bashrc
  /home/vagrant/.config/nushell/env.nu
  /home/vagrant/.config/nushell/config.nu
```
Post install prompt:
```
...

source $"($nu.home-path)/.cargo/env.nu"  # For nushell
```

Generated env.nu:
```
if ("$HOME/.cargo/bin" not-in ($env.Path | split row (char esep))) {
  $env.Path = ($env.Path | prepend "$HOME/.cargo/bin")
}
```

$env.PATH contents:
```
╭────┬──────────────────────────╮
│  0 │ $HOME/.cargo/bin         │
│  1 │ /home/vagrant/.cargo/bin │
│  2 │ /usr/local/sbin          │
│  3 │ /usr/local/bin           │
│  4 │ /usr/sbin                │
│  5 │ /usr/bin                 │
│  6 │ /sbin                    │
│  7 │ /bin                     │
│  8 │ /usr/games               │
│  9 │ /usr/local/games         │
│ 10 │ /snap/bin                │
╰────┴──────────────────────────╯
```
Unfortunately generated env.nu still uses posix-like `$HOME` syntax instead of nu-specific `$"($nu.home-path)/.cargo/bin"` so this still doesn't work as advertised (but no longer breaks existing configs so there's that)
Other than that you may want to consider only appending to `config.nu` since `env.nu` is about to be deprecated

rami3l:
@yethal Thanks for your quick response. As per https://github.com/rust-lang/rustup/pull/4265#discussion_r2010452958 this is the expected behavior (for now); will fix.

alluring-mushroom:
@Yethal 
> `env.nu` is about to be deprecated 

Thank you for this, I hadn't been following the recent changes.

@rami3l if you look at the [documentation for the next release](https://www.nushell.sh/book/configuration.html#using-constants) 

It shows out nushell intends for it to be done in the future

src/cli/self_update/shell.rs
rami3l: Fixed.

rami3l:
@Yethal @alluring-mushroom Sorry for the back and forth, but I have tried to apply the suggestions from both of you. Would you mind giving this PR another shot?

Yethal:
```
~> $env.PATH
╭───┬──────────────────────────╮
│ 0 │ /home/vagrant/.cargo/bin │
│ 1 │ /usr/local/sbin          │
│ 2 │ /usr/local/bin           │
│ 3 │ /usr/sbin                │
│ 4 │ /usr/bin                 │
│ 5 │ /sbin                    │
│ 6 │ /bin                     │
│ 7 │ /usr/games               │
│ 8 │ /usr/local/games         │
│ 9 │ /snap/bin                │
╰───┴──────────────────────────╯
```
@rami3l We're good. Path is properly appended so the issue is resolved.

For easier multiplatorm support (once that happens) you may want to switch over to using the `path add` function from the Nushell standard library which will take care of deduplicating Path entries for you as well as `path join` for platform specific path separators instead of hardcoding `/`. Code works fine as is though so treat this as a suggestion and not something that needs to be implemented right now.

```
use std/util "path add"
path add ($nu.home-path | path join .cargo bin)
```

rami3l:
@Yethal Thanks for your feedback! I'm glad to see that this PR is working for you.

Regarding your two suggestions, I think since we are handling path joining from Rust's side, it might be better to postpone the change regarding `path join`.

On the other hand, your `path add` suggestion does look like a good improvement that we can integrate right now. Regarding the final `env.nu` template, I'm thinking about something like the following:

```diff
-if ($"{cargo_bin}" not-in ($env.Path | split row (char esep))) {
-  $env.Path = ($env.Path | prepend $"{cargo_bin}")
-}
+use std/util "path add"
+path add $"{cargo_bin}"
```

... I hope that looks reasonable to you. Besides, do you think this `use std/util "path add"` API is stable enough for the inclusion?

Yethal:
> @Yethal Thanks for your feedback! I'm glad to see that this PR is working for you.
> 
> Regarding your two suggestions, I think since we are handling path joining from Rust's side, it might be better to postpone the change regarding `path join`.
> 
> On the other hand, your `path add` suggestion does look like a good improvement that we can integrate right now. Regarding the final `env.nu` template, I'm thinking about something like the following:
> 
> ```diff
> -if ($"{cargo_bin}" not-in ($env.Path | split row (char esep))) {
> -  $env.Path = ($env.Path | prepend $"{cargo_bin}")
> -}
> +use std/util "path add"
> +path add $"{cargo_bin}"
> ```
> 
> ... I hope that looks reasonable to you. Besides, do you think this `use std/util "path add"` API is stable enough for the inclusion?

Yes, this looks reasonable. `path add` is part of Nushell's standard library so I'd say it's stable enough

src/cli/self_update/shell.rs
```diff
@@ -41,14 +41,19 @@ pub(crate) struct ShellScript {
...
+fn cargo_home_str_with_home(home: &str, process: &Process) -> Result<Cow<'static, str>> {
     let path = process.cargo_home()?;
 
     let default_cargo_home = process
         .home_dir()
         .unwrap_or_else(|| PathBuf::from("."))
         .join(".cargo");
     Ok(if default_cargo_home == path {
-        "$HOME/.cargo".into()
+        Cow::Owned(home.to_owned() + "/.cargo")
```
djc: Nit: I think it would be clearer to write this as `Cow::Owned(format!("{home}/.cargo"))`.

src/cli/self_update/shell.rs
```diff
@@ -101,21 +96,27 @@ pub(crate) trait UnixShell {
         }
     }
 
+    fn cargo_home_str(&self, process: &Process) -> Result<Cow<'static, str>> {
+        let home = "$HOME";
```
djc, line 103: Nit: don't see much point in the `home` binding in the context of this commit. Suggest inlining it here.

src/cli/self_update.rs
```diff
@@ -578,13 +581,21 @@ pub(crate) async fn install(
...
     };
     #[cfg(not(windows))]
-    let msg = if no_modify_path {
-        format!(
-            post_install_msg_unix_no_modify_path!(),
-            cargo_home = cargo_home
-        )
-    } else {
-        format!(post_install_msg_unix!(), cargo_home = cargo_home)
+    let msg = {
```
djc: Nit: suggest skipping the outer scope here.

src/cli/self_update/unix.rs
```diff
@@ -174,15 +174,15 @@ fn remove_legacy_paths(process: &Process) -> Result<()> {
     remove_legacy_source_command(
         format!(
             "export PATH=\"{}/bin:$PATH\"\n",
-            shell::cargo_home_str(process)?
+            Posix.cargo_home_str(process)?
```
djc, line 177: Should we pass some kind of shell into here? Why is it okay to hardcode the shell here?

djc (commented):
Good stuff!

src/cli/self_update/shell.rs
rami3l: The point will show up later... It's one of the side effects of rewriting history I think :)

src/cli/self_update/unix.rs
rami3l: In legacy error notes we *assume* POSIX-style paths. You can click to expand the context to get a better view.

src/cli/self_update/shell.rs
rami3l: Done.

src/cli/self_update/shell.rs
rami3l: The point will show up later in [`908285b` (#4265)](https://github.com/rust-lang/rustup/pull/4265/commits/908285bd067badcd1517999c5cfc0ad6914e0127)... Either way it won't make that much of a difference, so I've changed that according to your suggestion.

src/cli/self_update/unix.rs
djc: Fair enough!

djc (approved):

PR #4270, opened 2025-03-29
Show toolchain paths in `rustup show -v` output

### LINKED ISSUE #3408: rustup show should include toolchain link paths (maybe only with `show -v`)

I often need a reminder about which paths in my local development map to which rustup toolchain identifiers.

As far as I can tell, the only way to do this today is for me to chdir into the hidden directory rustup uses to represent the toolchains and list them, so that I can see the symbolic links it is using. I.e., there is no way to do it from the command line via the `rustup` tool itself; instead I am relying on knowledge of rustup's internal architecture to find what I need.

### Solution you'd like

It would be nice if the `rustup show` command actually *told me* about the toolchains that were set up via `rustup toolchain link` and also told me what path they link to.

### Notes

_No response_

CLoaKY233:
### Changes
1. Modified the `show` function in `src/cli/rustup_mode.rs` to display the path for each installed toolchain in verbose mode
2. Now all toolchains show their paths in the detailed output, not just the active toolchain

### Before
```
installed toolchains
--------------------
  rustc 1.85.1 (4eb161250 2025-03-15)

1.81-x86_64-pc-windows-msvc
  rustc 1.81.0 (eeb90cda1 2024-09-04)
```

### After
```
installed toolchains
--------------------
  rustc 1.85.1 (4eb161250 2025-03-15)
  path: C:\Users\laysh\.rustup\toolchains\stable-x86_64-pc-windows-msvc

1.81-x86_64-pc-windows-msvc
  rustc 1.81.0 (eeb90cda1 2024-09-04)
  path: C:\Users\laysh\.rustup\toolchains\1.81-x86_64-pc-windows-msvc
```
Fixes #3408
---

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1007,7 +1007,6 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
         print_header::<Error>(&mut t, "installed toolchains")?;
 
         let default_toolchain_name = cfg.get_default()?;
-
         let last_index = installed_toolchains.len().wrapping_sub(1);
         for (n, toolchain_name) in installed_toolchains.into_iter().enumerate() {
             let is_default_toolchain = default_toolchain_name.as_ref() == Some(&toolchain_name);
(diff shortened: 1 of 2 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1026,14 +1025,11 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
                 let toolchain = Toolchain::new(cfg, toolchain_name.into())?;
                 writeln!(
                     cfg.process.stdout().lock(),
-                    "  {}",
-                    toolchain.rustc_version()
+                    "  {}\n  path: {}",
+                    toolchain.rustc_version(),
+                    toolchain.path().display()
                 )?;
-                // To make it easy to see which rustc belongs to which
```
rami3l, line 1032: I'd prefer keeping this new line and modify the expected test instead :)

rami3l (commented):
Thanks a lot for the contribution!

Before merging this PR, could you please squash these changes into one commit? Many thanks in advance 🙏

src/cli/rustup_mode.rs
CLoaKY233: Updated as you requested :D

CLoaKY233:
@rami3l changes made, and commits squashed, any changes you suggest?

djc (approved):
Makes sense to me, thanks!

rami3l (approved):

rami3l:
@CLoaKY233 Thanks for your contribution, and LGTM!

PR #4272, opened 2025-03-30
Tweak rustup help output

### LINKED ISSUE #4269: `rustup` is an installer, but `install` is not discoverable from `rustup --help`
### Verification

### Problem

The issue is about discoverability of `rustup`.

`rustup` is a "toolchain INSTALLER".  However:

```
$ rustup --help 2>&1 | grep install                                                                                                          127
The Rust toolchain installer
  show         Show the active and installed toolchains or profiles
  toolchain    Modify or query the installed toolchains
  component    Modify a toolchain's installed components
  self         Modify the rustup installation
    Rustup installs The Rust Programming Language from the official
```

There is no command to install anything.  There are commands to modify and show and query toolchains, and modify a toolchain, and modify rustup itself.

But no command to INSTALL a toolchain!  There should be something in `--help` that indicate which command is used to install a toolchain.

Image

### Steps

rustup --help

### Possible Solution(s)

Show an example installing the latest toolchain in the help.

### Notes

_No response_

### Rustup version

```console
1.27.1
```

### Installed toolchains

```console
there is nothing installed
```

### OS version

```console
Nixos
```

djc:
```diff
--- old.txt     2025-03-30 10:49:43
+++ new.txt     2025-03-30 10:49:30
@@ -1,15 +1,15 @@
-rustup 1.28.1+35 (626178865 2025-03-25) dirty 1 modification
+rustup 1.28.1+37 (39efc4949 2025-03-30)
 
 The Rust toolchain installer
 
 Usage: rustup[EXE] [OPTIONS] [+toolchain] [COMMAND]
 
 Commands:
+  toolchain    Install, uninstall, or list toolchains
+  default      Set the default toolchain
   show         Show the active and installed toolchains or profiles
   update       Update Rust toolchains and rustup
   check        Check for updates to Rust toolchains and rustup
-  default      Set the default toolchain
-  toolchain    Modify or query the installed toolchains
   target       Modify a toolchain's supported targets
   component    Modify a toolchain's installed components
   override     Modify toolchain overrides for directories
```

Fixes #4269.

### DIFF
--- a/tests/suite/cli-ui/rustup/rustup_help_cmd_stdout.toml
+++ b/tests/suite/cli-ui/rustup/rustup_help_cmd_stdout.toml
@@ -9,7 +9,7 @@ The Rust toolchain installer
 Usage: rustup[EXE] [OPTIONS] [+toolchain] [COMMAND]
 
 Commands:
-  toolchain    Modify or query the installed toolchains
+  toolchain    Install, uninstall, or list toolchains
   default      Set the default toolchain
   show         Show the active and installed toolchains or profiles
   update       Update Rust toolchains and rustup
--- a/tests/suite/cli-ui/rustup/rustup_help_cmd_stdout.toml
+++ b/tests/suite/cli-ui/rustup/rustup_help_cmd_stdout.toml
@@ -9,11 +9,11 @@ The Rust toolchain installer
 Usage: rustup[EXE] [OPTIONS] [+toolchain] [COMMAND]
 
 Commands:
+  toolchain    Modify or query the installed toolchains
+  default      Set the default toolchain
   show         Show the active and installed toolchains or profiles
   update       Update Rust toolchains and rustup
   check        Check for updates to Rust toolchains and rustup
-  default      Set the default toolchain
-  toolchain    Modify or query the installed toolchains
   target       Modify a toolchain's supported targets
   component    Modify a toolchain's installed components
   override     Modify toolchain overrides for directories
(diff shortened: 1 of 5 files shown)


rami3l (approved):
That's much cleaner, thanks!

tests/suite/cli-ui/rustup/rustup_help_cmd_stdout.toml
```diff
@@ -9,11 +9,11 @@ The Rust toolchain installer
 Usage: rustup[EXE] [OPTIONS] [+toolchain] [COMMAND]
 
 Commands:
+  toolchain    Install, uninstall, or list toolchains
   show         Show the active and installed toolchains or profiles
   update       Update Rust toolchains and rustup
   check        Check for updates to Rust toolchains and rustup
   default      Set the default toolchain
```
rami3l: Looking at the diff it makes me wonder if `default`/`target`/`component` should be moved to the top as well since they basically make up the main concerns of this project. It'd be fine just to put them under `toolchain` I guess.

djc: I thought about that -- I could see your point for `default` but I think `target` and `component` are probably much less likely to be used by many users, especially inexperienced users just getting started?

Like, I think the vast majority of users probably installs a default toolchain and then updates it over time, and doesn't do much else.

rami3l: > I thought about that -- I could see your point for `default` but I think `target` and `component` are probably much less likely to be used by many users, especially inexperienced users just getting started?

Thanks for the quick reply! I think it will make sense as well to just move `default` like you've pointed out. After all, this is just my additional remark, I don't want it to be a blocker :)

ChrisDenton (approved):
This is much nicer!

PR #4275, opened 2025-04-03
Update `CARGO` environment variable if it is set to a rustup proxy

### LINKED ISSUE #4274: `CARGO` environment variable, if set to a proxy, should be changed to match the selected toolchain

This is a follow up to https://github.com/rust-lang/cargo/issues/15099

In short, it used to be that setting the `CARGO` environment variable to, e.g. `$CARGO_HOME/bin/cargo` prior to invoking `cargo` could result in mismatched toolchains when `cargo` is managed by rustup, E.g. `cargo +nightly` will run the nightly cargo but later use of the `CARGO` environment variable would mean running the default toolchain instead (which may be different).

That issue is fixed for newer versions of Cargo but rustup obviously supports installing old versions. Should we make a best effort to ensure the `CARGO` environment variable is correctly set to the right toolchain when running via a proxy?

### Solution you'd like

If running `cargo` via rustup and if the `CARGO` environment variable is set to a rustup proxy (e.g. as in rustup managed `$CARGO_HOME/bin/cargo`) then set `CARGO` to the actual path for `cargo`.

### Notes

This mimics newer cargo behaviour. We shouldn't go out of our way to fix this in all cases but if a minimal change works around this issue in most cases then it would still be useful, I think.

ChrisDenton:
This is done only if we know the absolute path to the real `cargo` and only if we're sure that `CARGO` points to a proxy.

Fixes #4274

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -423,6 +423,19 @@ impl<'a> Toolchain<'a> {
         };
         let mut cmd = Command::new(path);
         self.set_env(&mut cmd);
+
+        // If we're running cargo and the `CARGO` environment variable is set
+        // to a rustup proxy then change `CARGO` to be the real cargo binary,
+        // but only if we know the absolute path to cargo.
+        // This works around an issue with old versions of cargo not updating
+        // the environment variable itself.
+        if Path::new(&binary).file_stem() == Some("cargo".as_ref()) && path.is_absolute() {
+            if let Some(cargo) = self.cfg.process.var_os("CARGO") {
+                if fs::read_link(&cargo).is_ok_and(|p| p.file_stem() == Some("rustup".as_ref())) {
+                    cmd.env("CARGO", path);
+                }
+            }
+        }
         Ok(cmd)
     }
 
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -684,6 +684,7 @@ async fn setup_test_state(test_dist_dir: tempfile::TempDir) -> (tempfile::TempDi
     //   **in this process** when the tests are still running.
     unsafe {
         // Unset env variables that will break our testing
+        env::remove_var("CARGO");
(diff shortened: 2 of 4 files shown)


rami3l (approved):
Thanks!

Not a blocker, but do we have a way to reliably test this though?

src/toolchain.rs
```diff
@@ -423,6 +423,22 @@ impl<'a> Toolchain<'a> {
...
+
+        // If we're running cargo and the `CARGO` environment variable is set
+        // to a rustup proxy then change `CARGO` to be the real cargo binary,
+        // but only if we know the absolute path to cargo.
+        // This works around an issue with old versions of cargo not updating
+        // the environment variable itself.
+        if Path::new(&binary).file_stem() == Some("cargo".as_ref()) && path.is_absolute() {
+            if let Some(cargo) = self.cfg.process.var_os("CARGO") {
+                if fs::read_link(&cargo)
+                    .map(|p| p.file_stem() == Some("rustup".as_ref()))
```
rami3l: Nit: I believe there's a new `.is_some_and()` which can replace this `.map().unwrap_or(false)`.

ChrisDenton:
> Not a blocker, but do we have a way to reliably test this though?

Hm. Not easily without some bigger changes to the testing framework... actually I guess I can manually create a symlink to a mock cargo in a test. That might work.

src/toolchain.rs
djc: This should really not be a separate commit.

djc (approved):

src/toolchain.rs
rami3l: @djc Emmm I guess we can go back and squash it?

rami3l: @djc Squashed. Returning to the merge queue.

PR #4277, opened 2025-04-04
Simplify download abstractions and start warning about curl usage

### LINKED ISSUE #993: Replace custom download crate with reqwest
reqwest is mature enough that I don't feel the need to keep maintaining an abstraction layer, and would rather be using pure Rust instead of curl.

Make sure that it continues obeying the relevant environment variables for setting up proxies.

djc:
This is another step towards #3790. Also in #2626 it was noted that curl is imposing substantial overhead on startup time, even when it is not being used.


rami3l (approved):
LGTM! Just don't forget to adjust the target features in `run.bash` accordingly...

ci/run.bash
```diff
@@ -50,21 +50,6 @@ target_cargo() {
 
 target_cargo build
 
-download_pkg_test() {
-  features=('--no-default-features' '--features' 'curl-backend,reqwest-native-tls')
```
rami3l, line 54: I believe this logic is still required at least for now: the `--features test` below won't suffice.

ChrisDenton (approved):
lgtm but will this require any changes in our release process?

djc:
> lgtm but will this require any changes in our release process?

Why would it? I think we still build the rustup crate, whose features have not changed.

ci/run.bash
djc: Adapted this -- want to take another look?

rami3l: @djc This is looking good, thanks :)

rami3l (approved):

PR #4283, opened 2025-04-09
docs(user-guide): cover correct `$PATH` configuration for proxies

rami3l:
As a prerequisite of merging my https://github.com/Homebrew/homebrew-core/pull/218316 (and thus addressing https://github.com/rust-lang/rustup/issues/1236#issuecomment-2780141056), this addition has been suggested by @mikemcquaid (the current lead of homebrew) in https://github.com/Homebrew/homebrew-core/pull/218316#discussion_r2032633321:

> the actual instructions should be in the `rustup` formula and/or the official Rust documentation (and provide a link to there instead).
> 
> As-is the existing text is:
> - far too long (the longer the text, the less likely any of it is to be read)
> - unclear on what should be added where; there's no specifics

### DIFF
--- a/doc/user-guide/src/installation/already-installed-rust.md
+++ b/doc/user-guide/src/installation/already-installed-rust.md
@@ -7,6 +7,8 @@ will complain that you already have Rust installed in `/usr` and refuse to
 install. However, you can install Rust via `rustup` and have it coexist with
 your packaged Rust toolchain.
 
+## Set up rustup with an existing Rust toolchain
+
 When you initially install Rust with `rustup`, pass the `-y` option to make it
 ignore the packaged Rust toolchain and install a `rustup`-managed toolchain
 into `~/.cargo/bin`. Add that directory to your `$PATH` (or let `rustup` do it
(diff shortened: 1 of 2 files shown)


djc (approved):

doc/user-guide/src/installation/already-installed-rust.md
```diff
@@ -21,12 +23,64 @@ You can then use "system" as a `rustup` toolchain, just like "nightly".
...
+lines in your shell configuration file(s).
+
+To make it clearer, let's look at the example of a common conflict between the
+regular rustup downloaded from [rustup.rs] and homebrew-installed `rust` on macOS.
+In your `.profile`, you can probably find these two lines in this order:
+
+```bash
+. $HOME/.cargo/env
+eval $(/opt/homebrew/bin/brew shellenv)
+```
```
ChrisDenton: Maybe this example and text should be written to emphasise the *correct* way to do it rather than showing the wrong way? If someone is scanning the text then they may end up with the wrong idea unless they read more closely. I mean something like "the right way to do this is... if the lines are the other way around then..."

doc/user-guide/src/installation/already-installed-rust.md
```diff
@@ -21,12 +23,64 @@ You can then use "system" as a `rustup` toolchain, just like "nightly".
...
+
+- ```text
+  cargo 1.85.1 (d73d2caf9 2024-12-31)
+  ```
+
+- ```text
+  error: toolchain 'stable' is not installed
+  ```
+
+[rustup.rs]: https://rustup.rs
```
MikeMcQuaid: I'd suggest perhaps using `which cargo` and/or `echo $PATH` to help debug these issues.

MikeMcQuaid (approved):
Thanks for this! One suggestion of how you could potentially make it easier to debug this.

doc/user-guide/src/installation/already-installed-rust.md
rami3l: @MikeMcQuaid `which cargo` or `readlink $(which cargo)` won't always work, unfortunately, because 1) I've seen different distros putting the proxy in weird places to the point where it can be unclear whether that's rustup-related; 2) the link could in certain cases be a hard link rather than a symlink, as IIRC @ChrisDenton has previously pointed out in another issue.

OTOH `echo $PATH` would do, definitely, however both rustup and homebrew use indirections to introduce these paths, so it's not definitive either 😅 It's a good addition though!

MikeMcQuaid: @rami3l your project so, up to you, but my experience from Homebrew has been it's worth asking "what proportion of people will these instructions be right for?" rather than "will these instructions be right for everyone?".

If they are useful for e.g. 90-99% of people: it's probably worth being explicit about what commands which resolve their issues (as concisely as possible) to minimise your support burden on the project.

If they are useful for <=50% of people, yeh, those instructions are going to waste more time than they will save.

I don't have enough context to suggest which buckets these will fit into and defer to you here but hope the advice is useful!

doc/user-guide/src/installation/already-installed-rust.md
rami3l: @ChrisDenton Nice suggestion, and fixed!

doc/user-guide/src/installation/already-installed-rust.md
rami3l: @MikeMcQuaid I've taken `echo $PATH` into consideration. Thanks again for the review!

ChrisDenton (approved):
Looks clear to me, thanks!

PR #4290, opened 2025-04-16
ci(windows): enable full test suite for ARM64 build

rami3l:
This takes advantage of the new (free) ARM64 Windows runners: https://blogs.windows.com/windowsdeveloper/2025/04/14/github-actions-now-supports-windows-on-arm-runners-for-all-public-repos/


rami3l:
Not sure why our ~~installer~~ setup script (`echo "%USERPROFILE%\.cargo\bin" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8`) cannot add the `$PATH` correctly (the same stuff works alright on Windows Server x64). I'm afraid this will make stuff go wrong in `run.bash` anyway.

---

Update: Copilot suggested trying `$env.USERPROFILE`. Will give it a try real soon.

rami3l:
Our compatibility with Windows ARM64 seems to be quite good: the full test suite has passed for a few times in a row.

rami3l:
I can think of one potential risk of merging this PR: there could be a slight possibility that `aws` is not installed on this new machine... I'll use a quick droppable delta to verify this.

---

Update: Good news! It's actually installed already:

```console
Run Get-Command aws
  Get-Command aws
  aws --version
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    RUSTFLAGS: -Ctarget-feature=+crt-static
    RUST_MIN_STACK: 16777216

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     aws.exe                                            0.0.0.0    C:\Program Files\Amazon\AWSCLIV2\aws.exe
aws-cli/2.24.8 Python/3.12.6 Windows/11 exe/ARM64
```

_https://github.com/rust-lang/rustup/actions/runs/14490128412/job/40644490076?pr=4290_

ChrisDenton (approved):
> Our compatibility with Windows ARM64 seems to be quite good: the full test suite has passed for a few times in a row.

That's great to hear! I think for rustup problems with platform detection would be the main thing likely to be affected by an ARM64 host but it seems that's not an issue.

djc (approved):
Nice!

rami3l:
> I think for rustup problems with platform detection would be the main thing likely to be affected by an ARM64 host but it seems that's not an issue.

@ChrisDenton Thanks for pointing that out; to be frank I'm not quite sure the platform detection is good though, it looks like when we install rustup in this CI everything is hardcoded, and I don't quite remember if we have tests on platform detection.

That said, I've heard from a few users about their experience on ARM64 and I'm not aware of anything about that yet either.

---

Update: Someone has obviously done that testing manually on their machine when adding support for ARM64 Windows: https://github.com/rust-lang/rustup/pull/3010

PR #4295, opened 2025-04-18
Use `RUSTUP_HARDLINK_PROXIES` environment variable to hard link proxies

ChrisDenton:
I've documented it but marked it as unstable for now as we may want to change defaults in the future.

### DIFF
--- a/doc/user-guide/src/environment-variables.md
+++ b/doc/user-guide/src/environment-variables.md
@@ -57,7 +57,12 @@
   toolchain when it is absent. Set this value to `0` to disable automatic
   installation.
 
+- `RUSTUP_HARDLINK_PROXIES` *unstable*. When set, rustup will not attempt to
+  symlink proxies and instead always use hardlinks. If you find this fixes
+  a problem, then please report the issue on the [rustup issue tracker].
+
 [directive syntax]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
 [dc]: https://docs.docker.com/storage/storagedriver/overlayfs-driver/#modifying-files-or-directories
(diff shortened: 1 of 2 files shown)


rami3l (approved):

djc (approved):
Don't love the proliferation of environment variables to work around broken defaults. @ChrisDenton do you feel the change in 1.28 has been worth it?

ChrisDenton:
It does solve an issue on Windows where attempting to replace proxies could error. Of the issues reported so far, I think most could be fairly categorised as teething problems, but the latest one is giving me more reason to reconsider. But I'm not yet sure what's causing it. If it's a particularly unusual environment then having a special configuration for that seems reasonable. If we can detect that environment automatically then even better. If it's something more general then making symlinks opt-in on Windows would make sense (I would like it to at least be an option for the reason I mentioned earlier).

ChrisDenton:
Actually, I think it would be simple enough to try `File::open` on a proxy (which follows symlinks) after they're installed and then if that fails, restart the process but forcing hardlinks.

The environment variable could still be useful as an override, at least temporarily, in case the above doesn't work for some reason. Because this is not something we can test in our CI.

ChrisDenton:
I wasn't sure whether to restrict the fallback to Windows only or if this should be used on all platforms?

rami3l:
> I wasn't sure whether to restrict the fallback to Windows only or if this should be used on all platforms?

@ChrisDenton I'd say just do it for all platforms. We've been making some unnecessary platform differences already (such as the shell env file thing; it could have been useful on Windows as well)...

ChrisDenton:
Ok, done!

rami3l (approved):

PR #4297, opened 2025-04-19
Switch flate2 to use the zlib-rs backend

joshtriplett:
This backend is written in Rust, but has substantially higher performance. And, as of the latest version, zlib-rs does not export any C zlib symbols, so it won't conflict with other implementations of zlib in the same process; thus, it doesn't conflict with linking to C libraries that use zlib.


djc (approved):

rami3l (approved):
Looks great to me, many thanks!

joshtriplett:
I don't think I understand why the "conclusion" CI job failed here.

PR #4307, opened 2025-04-25
Log original download errors immediately

djc:
To help diagnose errors like in #4305.

### DIFF
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -3,13 +3,13 @@
 use std::fs::remove_file;
 use std::path::Path;
 
+use anyhow::Context;
 #[cfg(any(
     not(feature = "curl-backend"),
     not(feature = "reqwest-rustls-tls"),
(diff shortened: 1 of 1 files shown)


src/download/mod.rs
```diff
@@ -522,31 +522,36 @@ mod reqwest_be {
...
+    ) -> anyhow::Result<()> {
         // Short-circuit reqwest for the "file:" URL scheme
         if download_from_file_url(url, resume_from, callback)? {
             return Ok(());
         }
 
-        let res = request(url, resume_from, client)
-            .await
-            .context("failed to make network request")?;
+        let res = match request(url, resume_from, client).await {
```
rami3l: Nit: Looks like a pretty good use case of `.inspect_err()` here.

rami3l (approved):

src/download/mod.rs
djc: Yup, good call!

PR #4312, opened 2025-04-25
Set ALPN protocols for rustls backend

### LINKED ISSUE #4311: rustup ought to tell rustls an ALPN

Noted in #4305 -- because rustup passes a pre-built rustls configuration into reqwest, it is responsible for configuring the `rustls::ClientConfig::alpn_protocols` field. This probably has no significant impact given the simple workloads of rustup, but disabling ALPN (and therefore HTTP2) is probably an unintended variance from the curl backend.

### Solution you'd like

Set ALPN so HTTP2 can be negotiated.

### Notes

_No response_

djc:
Fixes #4311.

cc @ctz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2047,6 +2047,7 @@ dependencies = [
  "futures-channel",
  "futures-core",
  "futures-util",
+ "h2",
  "http",
  "http-body",
  "http-body-util",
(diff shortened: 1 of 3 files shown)


rami3l:
@djc Would we want @muja to test this new branch again?

--- 

Update: Hmmm the CI is failing now, interesting...

muja:
I built & ran this branch, still the same error.

ctz:
I am pretty confident this is not the root cause of #4305

djc:
> Update: Hmmm the CI is failing now, interesting...

We didn't actually enable `http2` for our reqwest dependency. Should be better now.

rami3l (approved):

PR #4317, opened 2025-04-28
Deprecate native-tls as well

djc:
@shepmaster suggested in https://github.com/rust-lang/blog.rust-lang.org/pull/1593#discussion_r2063547165 that we should avoid people jumping from cURL to native-tls. I think we haven't seen a case where native-tls was better than rustls since we made rustls the default in 1.28.0, so I think it might make sense to deprecate this at the same time.


rami3l (approved):
Makes sense to me!

PR #4324, opened 2025-05-07
feat(toolchain): consider external `rust-analyzer` when calling a proxy

### LINKED ISSUE #3299: Is the rust-analyzer proxy supposed to be installed on tier-2 and lower platforms?
### Problem

On platforms like aarch64-apple-darwin or x86_64-unknown-freebsd, when I install Rust via rustup, and then install the rust-analyzer component, rust-analyzer isn't installed (no binary, no proxy). It seems to only install on tier 1 platforms (I'm afraid I can only test on x86_64-unknown-linux-gnu). Is this deliberate?

### Steps

1. Install Rust via rustup on a tier 2 platform
2. Install the rust-analyzer component
3. Try to run rust-analyzer

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
rustup 1.25.2 (17db695f1 2023-02-01)
```

### Installed toolchains

```console
rustup home:  /Users/luna/.rustup

installed toolchains
--------------------

nightly-aarch64-apple-darwin

active toolchain
----------------

rustc 1.68.2 (9eb3afe9e 2023-03-27)
```

### LINKED ISSUE #3846: Binary named `rust-analyzer` added to cargo bin path even when component is not added
### Verification

### Problem

Rustup keeps adding a binary named `rust-analyzer` to the .cargo/bin folder. But I didn't add it because I want homebrew to manage `rust-analyzer` version, because I want the latest version. But since there is a `rust-analyer` in my path at `~/.cargo/bin/rust-analyzer` my editor tries to use it. 

Opening it i get a message

```
error: Unknown binary 'rust-analyzer' in official toolchain 'stable-aarch64-apple-darwin'.
To install, run `rustup component add rust-analyzer`
```

This is very annoying because I keep having to delete this dummy binary everytime i run rustup so my editor can find the actual binary. Is there any reason to create these dummy binaries? Can I disable this?

### Steps

1. Install rustup with with minimal 

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
```

### Installed toolchains

```console
rustup home:  /Users/praveen/.rustup

rustc 1.78.0 (9b00956e5 2024-04-29)
```

### OS version

```console
MacOS Version 14.5 (23F79)
```

rami3l:
Closes #3846.

Following the reasoning in https://github.com/rust-lang/rustup/issues/3846#issuecomment-2597469049, this PR make rustup prioritize rustup-hosted RA and only delegates to the outside as a final rescue. This way, it will also close https://github.com/rust-lang/rustup/issues/3299:

```console
> rust-analyzer --version
info: `rust-analyzer` is unavailable for the active toolchain
info: falling back to "~/.local/share/nvim/mason/bin/rust-analyzer"
rust-analyzer 0.3.2449-standalone (3b57c00151 2025-05-04)
```

This approach has also been approved by @davidbarsky in [#t-rustup > More frequent rust-analyzer updates in rustup @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/490103-t-rustup/topic/More.20frequent.20rust-analyzer.20updates.20in.20rustup/near/514149302) as a temporary measure before the new update cadence for RA on our release server is settled.

## Concerns

- [x] ~~Should we emit a log line indicating the fallback is happening? Idem for the pre-existing `cargo` fallback.~~ Added.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -1,3 +1,5 @@
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt as _;
 use std::{
     env::{self, consts::EXE_SUFFIX},
     ffi::{OsStr, OsString},
@@ -12,6 +14,7 @@ use std::{
 
 use anyhow::{Context, anyhow, bail};
 use fs_at::OpenOptions;
+use same_file::is_same_file;
 use tracing::info;
 use url::Url;
 use wait_timeout::ChildExt;
@@ -331,6 +334,8 @@ impl<'a> Toolchain<'a> {
         // perhaps a trait that create command layers on?
         if let Some(cmd) = self.maybe_do_cargo_fallback(binary)? {
             return Ok(cmd);
+        } else if let Some(cmd) = self.maybe_do_rust_analyzer_fallback(binary)? {
+            return Ok(cmd);
         }
 
         self.create_command(binary)
@@ -368,6 +373,52 @@ impl<'a> Toolchain<'a> {
             }
         }
 
+        Ok(None)
+    }
+
+    /// Tries to find `rust-analyzer` on the PATH when the active toolchain does
+    /// not have `rust-analyzer` installed.
+    ///
+    /// This happens from time to time often because the user wants to use a
+    /// more recent build of RA than the one shipped with rustup, or because
+    /// rustup isn't shipping RA on their host platform at all.
+    ///
+    /// See the following issues for more context:
+    /// - <https://github.com/rust-lang/rustup/issues/3299>
+    /// - <https://github.com/rust-lang/rustup/issues/3846>
+    fn maybe_do_rust_analyzer_fallback(&self, binary: &str) -> anyhow::Result<Option<Command>> {
+        if binary != "rust-analyzer" && binary != "rust-analyzer.exe"
+            || self.binary_file("rust-analyzer").exists()
+        {
+            return Ok(None);
+        }
+
+        let proc = self.cfg.process;
+        let Some(path) = proc.var_os("PATH") else {
+            return Ok(None);
+        };
+
+        let me = env::current_exe()?;
+
+        // Try to find the first `rust-analyzer` under the `$PATH` that is both
+        // an existing file and not the same file as `me`, i.e. not a rustup proxy.
+        for mut p in env::split_paths(&path) {
+            p.push(binary);
+            let is_external_ra = p.is_file()
+                // We report `true` on `is_same_file()` error to prevent an invalid `p`
+                // from becoming the candidate.
+                && !is_same_file(&me, &p).unwrap_or(true);
+            // On Unix, we additionally check if the file is executable.
+            #[cfg(unix)]
+            let is_external_ra = is_external_ra
+                && p.metadata()
+                    .is_ok_and(|meta| meta.permissions().mode() & 0o111 != 0);
+            if is_external_ra {
+                let mut ra = Command::new(p);
+                self.set_env(&mut ra);
+                return Ok(Some(ra));
+            }
+        }
         Ok(None)
     }
 
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -333,8 +333,12 @@ impl<'a> Toolchain<'a> {
         // Should push the cargo fallback into a custom toolchain type? And then
         // perhaps a trait that create command layers on?
         if let Some(cmd) = self.maybe_do_cargo_fallback(binary)? {
+            info!("`cargo` is unavailable for the active toolchain");
+            info!("falling back to {:?}", cmd.get_program());
             return Ok(cmd);
         } else if let Some(cmd) = self.maybe_do_rust_analyzer_fallback(binary)? {
+            info!("`rust-analyzer` is unavailable for the active toolchain");
+            info!("falling back to {:?}", cmd.get_program());
             return Ok(cmd);
         }
 
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -343,7 +343,7 @@ impl<'a> Toolchain<'a> {
 
     // Custom toolchains don't have cargo, so here we detect that situation and
     // try to find a different cargo.
-    pub(crate) fn maybe_do_cargo_fallback(&self, binary: &str) -> anyhow::Result<Option<Command>> {
+    fn maybe_do_cargo_fallback(&self, binary: &str) -> anyhow::Result<Option<Command>> {
         if binary != "cargo" && binary != "cargo.exe" {
             return Ok(None);
         }
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -329,13 +329,8 @@ impl<'a> Toolchain<'a> {
     pub(crate) fn command(&self, binary: &str) -> anyhow::Result<Command> {
         // Should push the cargo fallback into a custom toolchain type? And then
         // perhaps a trait that create command layers on?
-        if !matches!(
-            self.name(),
-            LocalToolchainName::Named(ToolchainName::Official(_))
-        ) {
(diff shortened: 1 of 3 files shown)


src/toolchain.rs
```diff
@@ -374,6 +374,37 @@ impl<'a> Toolchain<'a> {
...
+            let is_external_ra = p.is_file()
+                // We report `true` on `is_same_file()` error to prevent an invalid `p`
+                // from becoming the candidate.
+                && !is_same_file(&me, &p).unwrap_or(true);
+            if !is_external_ra {
+                return None;
+            }
+            let mut ra = Command::new(p);
+            self.set_env(&mut ra);
+            Some(ra)
```
weihanglo: Should we add a log here? I think it is quite important to know rustup special-cased this when debugging?

src/toolchain.rs
```diff
@@ -374,6 +374,37 @@ impl<'a> Toolchain<'a> {
...
+        let Some(path) = proc.var_os("PATH") else {
+            return Ok(None);
+        };
+
+        let me = env::current_exe()?;
+
+        // Try to find the first `rust-analyzer` under the `$PATH` that is both
+        // an existing file and not the same file as `me`, i.e. not a rustup proxy.
+        Ok(env::split_paths(&path).find_map(|mut p| {
+            p.push(binary);
```
weihanglo, line 410: We might want to check the executable bit of this path. And maybe show the first nonexecutable file path when nothing found.

ChrisDenton: I'm not too sure about that. If something is in PATH the expectation is that it's executable. It's up to the OS to determine if it actually is or not and I don't think it's useful to second guess it.

rami3l: I agree this should look more like `which` in that only executables can be considered candidates (provided that it's as platform-agnostic as it gets). However I'm not very sure about the second half...

weihanglo: And most shells have the behavior, as well as [execvp](https://linux.die.net/man/3/execvp) on Linux.

~~The other way is falling back to call `rust-analyzer` directly and let OS do whatever it wants, so we don't need to mimic the PATH lookup logic.~~

edit: the fallback won't work. It'll become an infinite loop… 😆

ChrisDenton: From experience I'm very leery of look-before-you-leap because even if you ignore toctou, there's opportunities for things to go wrong on weird filesystems or setups.

Btw, I just tried in bash and it just runs a script without the executable bit (well, it fails, but it tries) and does not read the rest of the PATH.

rami3l: @ChrisDenton @weihanglo Thanks for the detailed discussion touching both sides of the argument. I just read the relevant [`which-rs`' code](https://github.com/harryfei/which-rs/blob/eef199824a0cf1596e8afbe9e7a5e6a793486cad/src/checker.rs#L14-L44) on this and what it's doing is basically checking `x` on Unix exclusively. I guess that'd be a reasonable middle ground for this particular case (and stuff like a TOCTOU error is not really important here I believe)...

Also, quoting the previous link on [execvp](https://linux.die.net/man/3/execvp):

> The execlp(), execvp(), and execvpe() functions duplicate the actions of the shell in searching for an executable file...

rami3l: > Btw, I just tried in bash and it just runs a script without the executable bit (well, it fails, but it tries) and does not read the rest of the PATH.

I just tried the same on my Mac and I believe that's not the case at least on my machine: a simple `touch rust-analyzer` can change the interpretation of calling that command depending on `chmod +x` and `chmod -x`.

ChrisDenton: Seems like there's a difference between bash and which:

```
$ which biscuits
/home/chris/.cargo/bin/biscuits
$ biscuits
-bash: /home/chris/temp/biscuits: Permission denied
```

weihanglo: @ChrisDenton 
Did you have another biscuits executable with a lower precedence in your PATH?

That was not my case with bash 5.2.37 on both macOS 15.4.1 and Linux 5.10.235. Not sure what was different 🤔

weihanglo: Oh I see. `/home/chris/.cargo/bin/biscuits` is the executable

src/toolchain.rs
rami3l: I added log lines for both fallbacks, in stderr just in case stdout is somehow relied on.

src/toolchain.rs
rami3l: I'm not quite sure, but anyway this is what I've tried, and it seems to be coherent with what `which` gives (this `~/.moon/bin` has higher precedence than the `rust-analyzer` proxy, of course):

```console
> printf '#!/bin/sh\necho Hello World' > ~/.moon/bin/rust-analyzer
> chmod +x ~/.moon/bin/rust-analyzer
> rust-analyzer
Hello World
> chmod -x ~/.moon/bin/rust-analyzer
> rust-analyzer
error: Unknown binary 'rust-analyzer' in official toolchain 'stable-aarch64-apple-darwin'.
```

~~Maybe that's because I'm using `fish`?~~ I have observed the same behavior using `zsh` as well.

rami3l: I've updated my solution to align with [`which-rs`](https://github.com/harryfei/which-rs/blob/eef199824a0cf1596e8afbe9e7a5e6a793486cad/src/checker.rs#L14-L44) as mentioned above.

src/toolchain.rs
```diff
@@ -371,6 +376,42 @@ impl<'a> Toolchain<'a> {
         Ok(None)
     }
 
+    fn maybe_do_rust_analyzer_fallback(&self, binary: &str) -> anyhow::Result<Option<Command>> {
```
djc, line 393: Add a docstsring that briefly describes why this is useful/necessary and what the goal/semantics of the function should be?

src/toolchain.rs
```diff
@@ -371,6 +376,42 @@ impl<'a> Toolchain<'a> {
...
+        let proc = self.cfg.process;
+        let Some(path) = proc.var_os("PATH") else {
+            return Ok(None);
+        };
+
+        let me = env::current_exe()?;
+
+        // Try to find the first `rust-analyzer` under the `$PATH` that is both
+        // an existing file and not the same file as `me`, i.e. not a rustup proxy.
+        Ok(env::split_paths(&path).find_map(|mut p| {
```
djc: Given that there's a pretty long closure being passed to `find_map()` and not much else is happening here, suggest writing a `for`-loop instead of an iterator combinator.

src/toolchain.rs
```diff
@@ -333,8 +333,12 @@ impl<'a> Toolchain<'a> {
         // Should push the cargo fallback into a custom toolchain type? And then
         // perhaps a trait that create command layers on?
         if let Some(cmd) = self.maybe_do_cargo_fallback(binary)? {
+            info!("`cargo` is unavailable for the active toolchain");
+            info!("falling back on {:?}", cmd.get_program());
```
djc: Suggest that "falling back" usually pairs with "to" rather than "on"?

rami3l: @djc I started with `falling back to` and my spell checker is not happy with it: https://en.wiktionary.org/wiki/fall_back_on

~~Probably that's a common mistake or something.~~ A quick search in our org shows mixed usage of both forms, so probably it doesn't matter that much.

src/toolchain.rs
rami3l: I personally think the readability is about the same either way, but I've fixed it anyway.

src/toolchain.rs
rami3l: @djc Nice point, and fixed!

src/toolchain.rs
rami3l: Reverted to `falling back to` since it looks more popular across our org.

djc (approved):

PR #4330, opened 2025-05-12
toolchain: hoist binary name conditionals out of fallback functions

djc:
Follow up from #4324. Conceptually I feel like this makes more sense? Not a big win, though.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -332,14 +332,18 @@ impl<'a> Toolchain<'a> {
     pub(crate) fn command(&self, binary: &str) -> anyhow::Result<Command> {
         // Should push the cargo fallback into a custom toolchain type? And then
         // perhaps a trait that create command layers on?
-        if let Some(cmd) = self.maybe_do_cargo_fallback(binary)? {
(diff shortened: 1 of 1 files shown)


src/toolchain.rs
```diff
@@ -332,35 +332,36 @@ impl<'a> Toolchain<'a> {
...
         // perhaps a trait that create command layers on?
-        if let Some(cmd) = self.maybe_do_cargo_fallback(binary)? {
-            info!("`cargo` is unavailable for the active toolchain");
-            info!("falling back to {:?}", cmd.get_program());
-            return Ok(cmd);
-        } else if let Some(cmd) = self.maybe_do_rust_analyzer_fallback(binary)? {
-            info!("`rust-analyzer` is unavailable for the active toolchain");
-            info!("falling back to {:?}", cmd.get_program());
-            return Ok(cmd);
+        if binary == "cargo" || binary == "cargo.exe" {
```
rami3l: I guess a match with `|` patterns maybe clearer...

rami3l (approved):

PR #4331, opened 2025-05-12
Log versions during self updates

### LINKED ISSUE #4327: Self-update does not report the new version downloaded

When I ran my regular `rustup update` I got the new and amazing `rustup` version 1.28.2. Unfortunately, I didn't even know it, because it wasn't included in the summary:

```
...
info: downloading self-update

  stable-aarch64-apple-darwin unchanged - rustc 1.86.0 (05f9846f8 2025-03-31)
   nightly-aarch64-apple-darwin updated - rustc 1.89.0-nightly (dcecb9917 2025-05-09) (from rustc 1.88.0-nightly (e643f59f6 2025-04-07))

```

I had to do a web search to figure out it was updated, and then was able to run `rustup self update` to see I now had version 1.28.2.

### Solution you'd like

I'd like the rustup update to show in the list of things updated, and if it isn't updated, it could just say "rustup unchanged" just like it does for stable in my example output above.

### Notes

_No response_

djc:
Fixes #4327.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,6 +407,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.15.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "054ccb5b10f9f2cbf51eb355ca1d05c2d279ce1804688d0db74b4733a5aeafd8"
+dependencies = [
+ "encode_unicode",
+ "libc",
+ "once_cell",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "content_inspector"
 version = "0.2.4"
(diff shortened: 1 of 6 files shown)


rami3l (approved):
LGTM, thanks a lot :)

rami3l:
I just noticed that during `rustup self update` we already have this message:

https://github.com/rust-lang/rustup/blob/7ba0abf4ced87c239393d83971c1df017c45c40a/src/cli/self_update.rs#L1063-L1067

... so it might be interesting to just reuse that logic here.

djc:
> I just noticed that during `rustup self update` we already have this message:

I noticed that too, but I think just showing these versions as soon as we have them is probably better.

PR #4333, opened 2025-05-16
feat(custom-toolchains): `rustup show` now reporting installed targets

### LINKED ISSUE #4251: `rustup show` doesn't correctly report installed targets for custom toolchains
Following @wesleywiser's report in https://github.com/rust-lang/rustup/issues/4248, I realized that custom toolchains with optional components is a use case that needs more attention.

Minimal reproduction below:

```console
> rustup toolchain install nightly-2023-09-30
[..]

> mv $RUSTUP_HOME/toolchains/nightly-2023-09-30-aarch64-apple-darwin $RUSTUP_HOME/toolchains/stuff
[..]

> rustup +stuff show --verbose
[..]
active toolchain
----------------
name: stuff
active because: overridden by +toolchain on the command line
compiler: rustc 1.74.0-nightly (8ce4540bd 2023-09-29)
path: [$RUSTUP_HOME]/toolchains/stuff
installed targets:
```

FranciscoTGouveia:
Closes #4251.

Hi everyone! This is my first PR on _rustup_, part of my community bonding period of GSoC (on the _"Making Rustup Concurrent"_ project).

I'd appreciate your feedback on a few decisions I made, such as:
+ Assuming the existence of a `rustlib/components file`, even for custom toolchains.
+ Placing the function for listing custom targets in `toolchain.rs`.

Additionally, before I add a test, I’d like to ask where it would be most appropriate to place it: should it go in `tests/cli_v2.rs` or `tests/cli_rustup.rs`?

Thank you in advance for your time and input!

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -22,7 +22,11 @@ use wait_timeout::ChildExt;
 use crate::{
     RustupError,
     config::{ActiveReason, Cfg, InstalledPath},
-    dist::PartialToolchainDesc,
+    dist::{
+        PartialToolchainDesc, TargetTriple,
+        component::{Component, Components},
+        prefix::InstallPrefix,
+    },
     env_var, install,
     notifications::Notification,
     utils::{self, raw::open_dir_following_links},
@@ -574,5 +578,32 @@ impl<'a> Toolchain<'a> {
             (cfg.notify_handler)(Notification::UninstalledToolchain(&name));
         }
         Ok(())
+    }
+
+    /// Get the list of installed components for any toolchain
+    ///
+    /// NB: An assumption is made that custom toolchains always have a `rustlib/components` file
+    pub fn installed_components(&self) -> anyhow::Result<Vec<Component>> {
+        let prefix = InstallPrefix::from(self.path.clone());
+        let components = Components::open(prefix)?;
+        components.list()
+    }
+
+    /// Get the list of installed targets for any toolchain
+    pub fn installed_targets(&self) -> anyhow::Result<Vec<TargetTriple>> {
+        let targets = self
+            .installed_components()?
+            .into_iter()
+            .filter_map(|c| {
+                if c.name().starts_with("rust-std-") {
+                    Some(TargetTriple::new(
+                        c.name().trim_start_matches("rust-std-").to_string(),
+                    ))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        Ok(targets)
     }
 }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -987,17 +987,24 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
 
     let active_toolchain_targets: Vec<TargetTriple> = active_toolchain_name
         .and_then(|atn| match atn {
-            ToolchainName::Official(desc) => DistributableToolchain::new(cfg, desc.clone()).ok(),
-            // So far, it is not possible to list targets for a custom toolchain.
-            ToolchainName::Custom(_) => None,
-        })
-        .and_then(|distributable| distributable.components().ok())
-        .map(|cs_vec| {
-            cs_vec
-                .into_iter()
-                .filter(|c| c.installed && c.component.short_name_in_manifest() == "rust-std")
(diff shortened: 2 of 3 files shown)


src/dist/component/mod.rs
```diff
@@ -10,7 +10,7 @@ mod transaction;
 // The representation of a package, its components, and installation
 mod package;
 // The representation of *installed* components, and uninstallation
-mod components;
+pub mod components;
```
rami3l: I believe this change is not necessary since it is placed right after `pub use components::*` :)

src/toolchain.rs
```diff
@@ -575,4 +578,27 @@ impl<'a> Toolchain<'a> {
         }
         Ok(())
     }
+
+    /// Get the list of installed targets for any toolchain
+    ///
+    /// NB: An assumption is made that custom toolchains always have a `rustlib/components` file
+    pub fn list_targets(&self) -> anyhow::Result<Vec<TargetTriple>> {
```
rami3l: This function looks good, but I think it might be too specialized. You can split it into two functions, one named `installed_targets()` and another `installed_components()`, with the former reusing part of the result of the latter.

rami3l (commented):
Nice changes overall!

Pre-existing issue, but it'd probably be better if we make this new behavior more consistent across the board if `rustup component list` and `rustup target list` come into play as well. You can give the same outputs with or without the flag `--installed` if it's a custom toolchain, I believe.

As for the use of the `rustlib/components` file, I believe it's the best thing to do now, but do make sure the user experience is not heavily damaged when the file is missing (meaning you need to add a test case for the failing case as well).

`cli_v1` and `cli_v2` are for tests specific to a certain manifest version. Since this machinery is a fallback in the manifest-free case, it's better to put it in another file.

Many thanks in advance 🙏

src/toolchain.rs
FranciscoTGouveia: Thank you. I have now split it into two different functions (one reusing the other).

src/dist/component/mod.rs
FranciscoTGouveia: Yes, it is not necessary. Sorry about it. Already fixed! :)

FranciscoTGouveia:
> Pre-existing issue, but it'd probably be better if we make this new behavior more consistent across the board if `rustup component list` and `rustup target list` come into play as well.

Should I open a new issue/PR for this in order to keep this PR as small/atomic as possible, or should I do it here?

> As for the use of the `rustlib/components` file, I believe it's the best thing to do now, but do make sure the user experience is not heavily damaged when the file is missing (meaning you need to add a test case for the failing case as well).

In the failing case should I display any kind of error or should I keep the previous behavior of simply not reporting any installed target?

rami3l:
> Should I open a new issue/PR for this in order to keep this PR as small/atomic as possible, or should I do it here?

I'd prefer doing that in a separate PR that follows this one. Thanks :)

> > As for the use of the `rustlib/components` file, I believe it's the best thing to do now, but do make sure the user experience is not heavily damaged when the file is missing (meaning you need to add a test case for the failing case as well).
> 
> In the failing case should I display any kind of error or should I keep the previous behavior of simply not reporting any installed target?

It's okay to keep the previous behavior in the error case. We can always add that back later if this confuses some of our users.

src/toolchain.rs
```diff
@@ -575,4 +579,33 @@ impl<'a> Toolchain<'a> {
...
+    ///
+    /// NB: An assumption is made that custom toolchains always have a `rustlib/components` file
+    pub fn installed_components(&self) -> anyhow::Result<Vec<Component>> {
+        let prefix = InstallPrefix::from(self.path.clone());
+        let components = Components::open(prefix)?;
+        components.list()
+    }
+
+    /// Get the list of installed targets for any toolchain
+    ///
```
rami3l: Nit: did you mean to write something here on the consecutive blank lines though?

rami3l (approved):

src/toolchain.rs
FranciscoTGouveia: No, I just aimed to be consistent with already existent functions (e.g. `ensure_removed`); this can be removed.

rami3l: @FranciscoTGouveia Fair point, although I have no idea why that has been the case... just do the removal here and I'll merge it, many thanks 🙏

PR #4334, opened 2025-05-16
refactor(test): add new APIs for easier snapshot testing

rami3l:
This PR is the foundation of #4359.

RA-powered inline snapshot updates seem to work without issues in VSCode with the given demo (thanks @roife!):

image

## Concerns

- [ ] Could the API design use some improvements (naming, signature, ...)?
- [x] Should we start to mark the old APIs as deprecated as Cargo once did with theirs?
- [x] What redactions should we include by default (`this_host_triple`, `tempdir`, ...)?
  - This can be added as we go; it does look like `this_host_triple` is a good inclusion at the start; and I've made it so that each test can add their own redactions.

### DIFF
--- a/tests/suite/cli_exact.rs
+++ b/tests/suite/cli_exact.rs
@@ -8,6 +8,7 @@ use rustup::test::{
     CROSS_ARCH1, CROSS_ARCH2, CliTestContext, MULTI_ARCH1, Scenario, this_host_triple,
 };
 use rustup::utils::raw;
+use snapbox::str;
 
 #[tokio::test]
 async fn update_once() {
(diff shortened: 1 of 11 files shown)


ChrisDenton:
I really like this. Using a more principled testing framework makes a lot of sense rather than maintaining a bespoke solution. And it does seem to be a mostly drop-in replacement but  allows for more exact matching (no need for `contains`) and easier updating of expected output when it's intentionally changed.

It'd be great if we can figure out how to make the old APIs deprecated while still making CI pass but I think this is worth doing even if that proves challenging.

I don't have a strong opinion on naming. Maybe "redactions" could be "replacements"? Because it replaces the matched string.

rami3l:
@ChrisDenton Many thanks for the kind words! `Redaction` is snapbox terminology so I'll probably stick with that...

My plan is that we can merge this as-is first (with `#![allow(deprecated)]`), and then move on with the migration one bit at a time.

tests/suite/cli_exact.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(deprecated)]
```
djc: Nit: move these below the module-level docstrings, please?

tests/suite/cli_self_upd.rs
```diff
@@ -21,6 +21,7 @@ use rustup::test::{
 use rustup::test::{RegistryGuard, RegistryValueId, USER_PATH};
 use rustup::utils::{self, raw};
 use rustup::{DUP_TOOLS, TOOLS, for_host};
+use snapbox::str;
```
djc, line 25: I don't love this because I usually expect `str` to be `std::str`. Maybe alias this or just use `snapbox::str!()` in callers?

djc (approved):
Looks like a good direction!

tests/suite/cli_self_upd.rs
rami3l: @djc It's just a macro so they won't really clash, right?

On the one hand, as I've just tested, we can even `use std::str` on top of `use snapbox::str` without problems. At least in the corresponding `cargo` code it's also been imported like this...

On the other hand, the `str![[r#"..."#]]` special syntax is particular enough that I don't personally believe it can be confused with any `std` API even considering a potential name clash.

djc: > On the other hand, the `str![[r#"..."#]]` special syntax is particular enough that I don't personally believe it can be confused with any `std` API even considering a potential name clash.

This is why, without context, I would like the question to be answered, "where is the special syntax coming from".

rami3l: @djc That's fair; I've made https://github.com/rust-lang/rustup/pull/4342.

PR #4338, opened 2025-05-17
feat(self_update): add proxy sanity checks

### LINKED ISSUE #4336: Sanity check for proxies created during the installation phase
As suggested by @ChrisDenton:

Previous issues such as https://github.com/rust-lang/rustup/issues/4291 proved that the symlink approach to proxies on Windows could be problematic from time to time.

Maybe we could add a sanity check right after the installation phase for the proxie(s) with a backing binary, say running `cargo --version` or `rustc --version`, and see if the proxy can actually work and produce a reasonable output. If it doesn't, we should prompt the user to file an issue in https://github.com/rust-lang/rustup/issues and exit immediately with an error.

## Note
- It's unfortunately not possible to assume any backing binaries to be available at all times. Not all toolchains have `cargo`, and `rustup-init --default-toolchain=none` can even leave you without `rustc`.
- For packagers who would like to skip the installation phase entirely, the validation verification of proxies should not be our responsibility.

manyinsects:
Sometimes, proxies can be incorrectly installed for whatever reason. This adds a new check after initial component installation to call all proxies with --version, and to fail if one of them doesn't report success.

Closes #4336 (but there should probably be a followup issue for potentially self-repairing broken proxies?).

~~There are two test failures in my local testing:~~ this has been resolved (for now...?) by limiting to only core components

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -853,6 +853,29 @@ fn install_proxies_with_opts(process: &Process, force_hard_links: bool) -> Resul
     Ok(())
 }
 
+fn check_proxy_sanity(process: &Process, components: &[&str], desc: &ToolchainDesc) -> Result<()> {
+    let bin_path = process.cargo_home()?.join("bin");
+
+    // Sometimes linking a proxy produces an unpredictable result, where the proxy
+    // is in place, but manages to not call rustup correctly. One way to make sure we
+    // don't run headfirst into the wall is to at least try and run our freshly
+    // installed proxies, to see if they return some manner of reasonable output.
+    // We limit ourselves to the most common two installed components (cargo and rustc),
+    // because their binary names also happen to match up, which is not necessarily
+    // a given.
+    for component in components.iter().filter(|c| ["cargo", "rustc"].contains(c)) {
+        let cmd = Command::new(bin_path.join(format!("{component}{EXE_SUFFIX}")))
+            .args([&format!("+{desc}"), "--version"])
+            .status();
+
+        if !cmd.is_ok_and(|status| status.success()) {
+            return Err(RustupError::BrokenProxy.into());
+        }
+    }
+
+    Ok(())
+}
+
 async fn maybe_install_rust(
     current_dir: PathBuf,
     quiet: bool,
@@ -906,6 +929,8 @@ async fn maybe_install_rust(
             .await?
             .0
         };
+
+        check_proxy_sanity(process, components, desc)?;
 
         cfg.set_default(Some(&desc.into()))?;
         writeln!(process.stdout().lock())?;
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -141,6 +141,12 @@ pub enum RustupError {
         target: TargetTriple,
         suggestion: Option<String>,
     },
+    #[error(
+        "rustup executable proxies don't seem to work\n\
+        help: this might be a bug in rustup, please open a new issue here:\n\
+        help: https://github.com/rust-lang/rustup/issues/new"
+    )]
+    BrokenProxy,
     #[error("unknown metadata version: '{0}'")]
     UnknownMetadataVersion(String),
     #[error("manifest version '{0}' is not supported")]


src/cli/self_update.rs
```diff
@@ -853,6 +853,30 @@ fn install_proxies_with_opts(process: &Process, force_hard_links: bool) -> Resul
...
 }
 
+fn check_proxy_sanity(process: &Process, components: &[&str]) -> Result<()> {
+    let bin_path = process.cargo_home()?.join("bin");
+
+    // Sometimes linking a proxy produces an unpredictable result, where the proxy
+    // is in place, but manages to not call rustup correctly. One way to make sure we
+    // don't run headfirst into the wall is to at least try and run our freshly
+    // installed proxies, to see if they return some manner of reasonable output.
+    for component in components {
```
rami3l: I believe it should be expressed through this code that component names and binary names are not necessarily identical, but if you limit it to a representative portion like `rustc` and `cargo` then they happen to collide, but that also has to go into the comments. BTW I don't particularly like the idea of checking all components, maybe just `rustc` and `cargo` would be enough?

src/errors.rs
```diff
@@ -141,6 +141,11 @@ pub enum RustupError {
         target: TargetTriple,
         suggestion: Option<String>,
     },
+    #[error(
+        "unable to correctly link proxies. this might be a bug in rustup, please open a new issue here:\n\
```
rami3l: The usual trick for style consistency here is that you don't write dots. Probably:

```rs
"unable to correctly link proxies\nnote: this might be a bug in rustup, please file an issue:\nnote: https://github.com/rust-lang/rustup/issues/new"
```

The above is just a suggestion; it'd be still nice to use the `\` to keep the readability.

src/cli/self_update.rs
```diff
@@ -853,6 +853,30 @@ fn install_proxies_with_opts(process: &Process, force_hard_links: bool) -> Resul
...
+        let cmd = Command::new(bin_path.join(format!("{}{EXE_SUFFIX}", component)))
+            .args(["--version"])
+            .status();
+
+        if let Ok(status) = cmd {
```
rami3l: Nit: `if !cmd.is_ok_and(|status| status.success())` might be better in terms of deduplication.

rami3l (changes requested):
Not bad for the first PR, thanks a lot :)

The existing error messages might indicate that the default toolchain is unset in certain situations, so instead of calling `<bin> --version`, maybe call `<bin> +<toolchain> --version`?

src/cli/self_update.rs
manyinsects: Sure, I can limit it to those two, provided they've actually been installed.

manyinsects:
this should be ready for re-review :)

src/cli/self_update.rs
```diff
@@ -853,6 +853,29 @@ fn install_proxies_with_opts(process: &Process, force_hard_links: bool) -> Resul
...
+
+    // Sometimes linking a proxy produces an unpredictable result, where the proxy
+    // is in place, but manages to not call rustup correctly. One way to make sure we
+    // don't run headfirst into the wall is to at least try and run our freshly
+    // installed proxies, to see if they return some manner of reasonable output.
+    // We limit ourselves to the most common two installed components (cargo and rustc),
+    // because their binary names also happen to match up, which is not necessarily
+    // a given.
+    for component in components.iter().filter(|c| ["cargo", "rustc"].contains(c)) {
+        let cmd = Command::new(bin_path.join(format!("{}{EXE_SUFFIX}", component)))
```
rami3l: Nit: You could use the new-style `format!()` here :)

```suggestion
        let cmd = Command::new(bin_path.join(format!("{component}{EXE_SUFFIX}")))
```

src/cli/self_update.rs
```diff
@@ -853,6 +853,29 @@ fn install_proxies_with_opts(process: &Process, force_hard_links: bool) -> Resul
...
+            .args([&format!("+{}", desc), "--version"])
```
rami3l:
```suggestion
            .args([&format!("+{desc}"), "--version"])
```

src/errors.rs
```diff
@@ -141,6 +141,12 @@ pub enum RustupError {
         target: TargetTriple,
         suggestion: Option<String>,
     },
+    #[error(
+        "unable to correctly link proxies\n\
+        this might be a bug in rustup, please open a new issue here:\n\
+        https://github.com/rust-lang/rustup/issues/new"
```
rami3l:
```suggestion
        help: this might be a bug in rustup, please open a new issue here:\n\
        help: https://github.com/rust-lang/rustup/issues/new"
```

rami3l (approved):
Thanks a lot! I've left a few tiny nits but the overall it looks pretty nice!

cc @ChrisDenton I don't currently think we need to add tests for this particular PR, so modulo the comments I think it's okay to merge it as-is and see how it goes?

src/errors.rs
```diff
@@ -141,6 +141,12 @@ pub enum RustupError {
         target: TargetTriple,
         suggestion: Option<String>,
     },
+    #[error(
+        "unable to correctly link proxies\n\
+        help: this might be a bug in rustup, please open a new issue here:\n\
+        help: https://github.com/rust-lang/rustup/issues/new"
+    )]
+    InvalidProxyInstall,
```
djc: I'd avoid this wording and variant name since technically the proxy install completed without error -- it's just that the installed proxy does not actually work.

rami3l: @djc How about:

```suggestion
    BrokenProxy,
```

... and

 `"the rustup executable proxies don't seem to work"`

PR #4340, opened 2025-05-18
rustup check: add exit status and no-self-update logic

tjkirch:
First:
```
rustup check: adopt no-self-update logic

The 'self update' subcommand checks whether self updates have been
disabled, but 'check' did not, meaning the check output could include
updates you're powerless to accept.

This adds the logic from self update to check, no longer showing updates
that can't be applied by rustup, along with its --no-self-update flag,
in case you *are* able to update rustup but don't want to check it for
updates at the moment.
```

Then:
```
rustup check: set exit status based on available updates

This helps with interfaces above rustup, for example prompting a user to
update, but only if updates are actually available.

If any updates are available, exits 0, otherwise exits 1.  This allows
"natural" shell commands like 'if rustup check; then <update>'.  It also
mirrors existing exit codes for update, where rustup exits 1 if updates
are not permitted by the distribution or because of errors.
```

My use case is as described in the second commit: building a little interface on top of rustup to help a user stay updated, but not prompting them if there's not actually an update available.  It was fairly easy to add an exit code to `rustup check` to represent what it found about available updates.

I'm using a distribution-managed rustup, so seeing rustup updates in this process isn't actually useful; I see they're available but can't `rustup self update` to take care of them.  I thought it made sense to use the same logic as in `rustup self update` to determine whether to show self updates in `check`.  And since `self update` has an argument to force hiding that output (even if you _can_ update) I did the same in `check`.

---

One potential downside here is the change in behavior of the exit status of `check`, which is currently 0 regardless of available updates, unless rustup had an error.  I believe this to be minor, given that an exit of 1 in existing code didn't really give an option for automated recovery, which matches the "nothing to do" result of 1 for not having available updates with this change.  But I'd understand feedback about wanting a different exit code, or requiring a new flag to activate use of these exit codes.

Another potential downside is the removal of some output of `check` in the no-self-update case.  I see this as minor for a similar reason; there was nothing you could ask rustup to do with that output.  If someone is using it to automate checks for rustup updates without actually using rustup for updating itself (?) it seems there are better ways.

---

To test, I found that I could force an available toolchain update by removing its manifest file, and force a rustup update by lowering its version in Cargo.toml and reinstalling it to the test prefix.  So, I tested each combination - no updates, toolchain update, rustup update, and toolchain + rustup update.  I also tested with the no-self-update feature to confirm rustup updates are no longer shown, if desired.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -177,7 +177,10 @@ enum RustupSubcmd {
     },
 
     /// Check for updates to Rust toolchains and rustup
-    Check,
+    Check {
+        #[command(flatten)]
+        opts: CheckOpts,
+    },
 
     /// Modify a toolchain's supported targets
     Target {
@@ -336,6 +339,13 @@ enum ToolchainSubcmd {
         /// Path to the directory
         path: PathBuf,
     },
+}
+
+#[derive(Debug, Default, Args)]
+struct CheckOpts {
+    /// Don't check for self update when running the `rustup check` command
+    #[arg(long)]
+    no_self_update: bool,
 }
 
 #[derive(Debug, Default, Args)]
@@ -650,7 +660,7 @@ pub async fn main(
             }
             ToolchainSubcmd::Uninstall { opts } => toolchain_remove(cfg, opts).await,
         },
-        RustupSubcmd::Check => check_updates(cfg).await,
+        RustupSubcmd::Check { opts } => check_updates(cfg, opts).await,
         RustupSubcmd::Default {
             toolchain,
             force_non_host,
@@ -779,7 +789,7 @@ async fn default_(
     Ok(utils::ExitCode(0))
 }
 
-async fn check_updates(cfg: &Cfg<'_>) -> Result<utils::ExitCode> {
+async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode> {
     let mut t = cfg.process.stdout().terminal(cfg.process);
     let channels = cfg.list_channels()?;
 
@@ -815,7 +825,18 @@ async fn check_updates(cfg: &Cfg<'_>) -> Result<utils::ExitCode> {
         }
     }
 
-    check_rustup_update(cfg.process).await?;
+    let self_update_mode = cfg.get_self_update_mode()?;
+    // Priority: no-self-update feature > self_update_mode > no-self-update args.
+    // Check for update only if rustup does **not** have the no-self-update feature,
+    // and auto-self-update is configured to **enable**
+    // and has **no** no-self-update parameter.
+    let self_update = !self_update::NEVER_SELF_UPDATE
+        && self_update_mode == SelfUpdateMode::Enable
+        && !opts.no_self_update;
+
+    if self_update {
+        check_rustup_update(cfg.process).await?;
+    }
 
     Ok(utils::ExitCode(0))
 }
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -19,7 +19,7 @@ use crate::{
         common::{self, PackageUpdate, update_console_filter},
         errors::CLIError,
         help::*,
-        self_update::{self, SelfUpdateMode, check_rustup_update},
+        self_update::{self, RustupUpdateAvailable, SelfUpdateMode, check_rustup_update},
         topical_doc,
     },
     command,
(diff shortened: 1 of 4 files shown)


tjkirch:
^ This push fixes the relevant tests to expect the new behavior/output.  (I wasn't able to run tests earlier, but found that it was just flaky network tests and I could rerun them a few times.)

src/cli/rustup_mode.rs
```diff
@@ -177,7 +177,11 @@ enum RustupSubcmd {
     },
 
     /// Check for updates to Rust toolchains and rustup
-    Check,
+    Check {
+        /// Don't check for self update when running the `rustup check` command
+        #[arg(long)]
+        no_self_update: bool,
```
rami3l: I believe it's better to just flatten `CheckOpts` into `Check`, like how `UpdateOpts` is flattened into `ToolchainSubcmd::Install`.

src/test/clitools.rs
```diff
@@ -188,6 +188,17 @@ impl Config {
         }
     }
 
+    /// Expect an err status and a string in stdout
+    pub async fn expect_stdout_err(&self, args: &[&str], expected: &str) {
```
rami3l: I am really sorry for the pre-existing test APIs, they give a pretty cumbersome vibe and thus I won't recommend adding any more APIs like this (they have already been deprecated, as a matter of fact). The good news is that you can rebase onto the latest master and get to use the new set of APIs, as demonstrated in https://github.com/rust-lang/rustup/pull/4343 :)

rami3l (approved):
Nice catch about the inconsistencies, and many thanks for the PR :)

tjkirch:
^ This push attempts to address the requests above by flattening CheckOpts and rebasing onto master to use the new testing APIs.

src/cli/rustup_mode.rs
tjkirch: Done, thanks!  (I guess I was being too literal in copying the nearby Update code.)

src/test/clitools.rs
tjkirch: Done, thanks.  (I have to admit that I found the prior APIs more straightforward, and I had a harder time getting tests to pass with the new APIs.  Perhaps the visible result of this is the extra newline in expected test output in a couple of places, because a newline is being stripped somewhere.  I saw that snapbox has newline filtering, but I don't see it being applied, so I'm not sure what happened exactly.)

rami3l: Oh, you are basically not expected to write the intended output; you run these tests with snapshot overwrite mode on (`SNAPSHOTS=overwrite cargo test`) so the machine will write the intended output for you. If you use VSCode, this should be a code action you can click on to activate.

That's a good point though: it's probably time to start adding "how to use" comments in that regard.

tjkirch: Ah, I guess I didn't see how to do that.  I only see `snapbox::str` being used; I don't see any saved output files (other than the TOML files storing `--help` output) so I didn't think a feature like that was in use.  Did I wind up with the correct thing, or do I need to use snapbox differently?

rami3l: @tjkirch The sections bounded by `snapbox::str![[""]]` are actually snapshots (yes, they are embedded in the current source file), so with every "expected output" section you have, leave `snapbox::str![[""]]` and let snapbox update the actual output for you. If you need something to be formatted with in that section, call `.extend_redactions()` before calling `.is_ok()`/`.is_err()`/`.with_stdout()` etc. See https://github.com/rust-lang/rustup/pull/4340#discussion_r2101622335 for an example.

tests/suite/cli_exact.rs
```diff
@@ -224,15 +228,20 @@ async fn check_updates_self_no_change() {
...
-"
-            ),
-        )
+        .run("rustup", ["set", "auto-self-update", "enable"], &[])
         .await;
+
+    cx.config
+        .expect(["rustup", "check"])
+        .await
+        .is_err()
```
rami3l, line 241: Add `.extend_redactions([("[VERSION]", current_version)])` before `.is_err()`, probably inlining `current_version` as well if it's not used anywhere else.

src/test/clitools.rs
rami3l: @tjkirch If you use VSCode + Rust Analyzer, then updating the snapshots would be as simple as a button click ("Update Tests"):

image

If you are using something else, then you should be able to do something similar (I've just added support for this in Neovim with https://github.com/mrcjkb/rustaceanvim/pull/771), or you can always fall back to running the `SNAPSHOTS=overwrite cargo test` by hand.

tjkirch: I changed to `snapbox::str![[""]]` and used `SNAPSHOTS=overwrite` and it was set back to what I had, so I guess the existing commits were OK for this part of the request.

tests/suite/cli_exact.rs
tjkirch: Done, thanks.

tjkirch:
^ This push updates to use `extend_redactions` in `check_updates_self_no_change` as requested.

rami3l (approved):
LGTM :)

PR #4343, opened 2025-05-21
test(cli-self-upd): migrate to `.expect()` APIs

rami3l:
Part of #4359.

[_SemanticDiff_](https://app.semanticdiff.com/gh/rust-lang/rustup/pull/4343/commits)

### DIFF
--- a/tests/suite/cli_self_upd.rs
+++ b/tests/suite/cli_self_upd.rs
@@ -1,7 +1,5 @@
 //! Testing self install, uninstall and update
 
-#![allow(deprecated)]
-
 use std::env;
 use std::env::consts::EXE_SUFFIX;
 use std::fs;
@@ -21,7 +19,7 @@ use rustup::test::{
 #[cfg(windows)]
 use rustup::test::{RegistryGuard, RegistryValueId, USER_PATH};
 use rustup::utils::{self, raw};
-use rustup::{DUP_TOOLS, TOOLS, for_host};
+use rustup::{DUP_TOOLS, TOOLS};
 #[cfg(windows)]
 use windows_registry::Value;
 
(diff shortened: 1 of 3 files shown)


tests/suite/cli_self_upd.rs
```diff
@@ -949,11 +1073,14 @@ async fn update_does_not_overwrite_rustfmt() {
...
     cx.config
-        .expect_stderr_ok(
-            &["rustup", "self", "update"],
-            "`rustfmt` is already installed",
-        )
-        .await;
+        .expect(["rustup", "self", "update"])
+        .await
+        .with_stderr(snapbox::str![[r#"
+info: checking for self-update (current version: 1.28.2)
```
ChrisDenton: Do we want to hard code the version here?

ChrisDenton (commented):
Looks good to me. I just have one question.

tests/suite/cli_self_upd.rs
rami3l: @ChrisDenton Nice catch! Indeed that should've been a redaction. Let me see how to add it properly though...

rami3l: @ChrisDenton I've added a new global redaction `[CURRENT_VERSION]` for this since it also looks quite common.

ChrisDenton: Thanks! That does seem like a good idea, not just for the current tests but also possible future ones.

ChrisDenton (approved):

rami3l:
^ I have re-organized the commits so the migration of old tests to snapshot ones and the migration of existing snapshot tests to new redactions are properly separated.

PR #4347, opened 2025-05-25
feat(custom-toolchains): targets and components are now reported for custom toolchains

FranciscoTGouveia:
Continuation of #4333.

Instead of throwing an error, `rustup target list` and `rustup component list` now display the installed targets and components, respectively, for custom toolchains.

Similar to the previous PR, this assumes that there may not be a manifest file for custom toolchains. As a result, there is a minor inconsistency: using `target/component list` without the `--installed` flag always yields the same result as using it with the flag.

Thank you in advance for your time and input!

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -19,14 +19,12 @@ use super::self_update;
 use crate::{
     cli::download_tracker::DownloadTracker,
     config::Cfg,
-    dist::{
-        TargetTriple, ToolchainDesc, manifest::ComponentStatus, notifications as dist_notifications,
-    },
+    dist::{TargetTriple, ToolchainDesc, notifications as dist_notifications},
     errors::RustupError,
     install::UpdateStatus,
     notifications::Notification,
     process::{Process, terminalsource},
-    toolchain::{DistributableToolchain, LocalToolchainName, Toolchain, ToolchainName},
+    toolchain::{LocalToolchainName, Toolchain, ToolchainName},
     utils::{self, notifications as util_notifications, notify::NotificationLevel},
 };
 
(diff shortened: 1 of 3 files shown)


tests/suite/cli_v2.rs
```diff
@@ -837,18 +837,25 @@ async fn list_targets_v1_toolchain() {
...
-        .await;
+    cx.config.expect_ok(&["rustup", "default", "stable"]).await;
+    let stable_path = cx
+        .config
+        .rustupdir
+        .join("toolchains")
+        .join(format!("stable-{}", this_host_triple()));
     cx.config
-        .expect_ok(&["rustup", "default", "default-from-path"])
+        .expect_ok(&[
```
rami3l: I'd prefer using the new `.expect()` API as demonstrated in #4343 from now on :)

src/cli/common.rs
```diff
@@ -406,6 +406,45 @@ pub(super) fn list_items(
     Ok(utils::ExitCode(0))
 }
 
+pub(super) fn list_items_custom(
```
rami3l: I'm afraid this is still too repetitive. Can `list_items()` take a list of components instead so that you can reuse that part of the logic?

tests/suite/cli_v2.rs
FranciscoTGouveia: Thank you for pointing that out!
I think it should already be more coherent with the new `.expect()` API :)

src/cli/common.rs
```diff
@@ -381,70 +379,26 @@ where
 }
 
 pub(super) fn list_items(
-    distributable: DistributableToolchain<'_>,
-    f: impl Fn(&ComponentStatus) -> Option<&str>,
+    items: impl Iterator<Item = (String, bool)>,
```
rami3l, line 386: ~~How about a named `struct ListItem { name: String, installed: bool }` to make it clearer what `(String, bool)` actually means here?~~

src/cli/common.rs
```diff
@@ -381,70 +379,26 @@ where
...
 ) -> Result<utils::ExitCode> {
     let mut t = process.stdout().terminal(process);
-    for component in distributable.components()? {
-        let Some(name) = f(&component) else { continue };
-        match (component.available, component.installed, installed_only) {
-            (false, _, _) | (_, false, true) => continue,
-            (true, true, false) if !quiet => {
+    for (name, installed) in items {
+        if !installed_only || installed {
+            if installed && !installed_only && !quiet {
```
rami3l: Nit: How about using `if installed && !installed_only && !quiet { .. } else if installed || !installed_only { .. }` to prevent deep nesting?

rami3l (approved):
LGTM modulo the comments, nice job!

PS: When addressing review comments, please squash relevant changes back to the respective original commit :)

src/cli/common.rs
djc: IMO that's quite a bit of syntactic overhead for this. Instead, suggest a docstring that explains how the iterated values get used?

rami3l (approved):

djc:
Looks to me like the tests might be broken after the first of these commits? Suggest keeping test fixes in the same commit as the code changes that need them -- or squash merging.

PR #4351, opened 2025-05-30
Custom list tweaks

djc:
Follow up on #4347.

cc @FranciscoTGouveia


rami3l (approved):
Many thanks for the improvements!

FranciscoTGouveia:
Hello @djc!
Thank you for this PR, and apologies for not thinking of this earlier.
You're absolutely right, this approach is much cleaner.
I'll do my best to avoid situations like this in future contributions :)

PR #4361, opened 2025-06-02
Clean up installed components/targets list tweaks

djc:
Follow up on #4333.

IMO these long and nested combinator chains are very hard to understand, so it's better to avoid them by writing more boring code, even if it's longer.

cc @FranciscoTGouveia

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1016,28 +1016,20 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
         .map(|atar| (&atar.0, &atar.1))
         .unzip();
 
-    let active_toolchain_targets: Vec<TargetTriple> = active_toolchain_name
-        .and_then(|atn| match atn {
-            ToolchainName::Official(desc) => DistributableToolchain::new(cfg, desc.clone())
-                .ok()
-                .and_then(|distributable| distributable.components().ok())
-                .map(|cs_vec| {
-                    cs_vec
-                        .into_iter()
-                        .filter(|c| {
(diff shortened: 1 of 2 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1016,28 +1016,24 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<utils::ExitCode> {
...
-            }
-        })
-        .unwrap_or_default();
+    let active_toolchain_targets = match active_toolchain_name {
+        Some(ToolchainName::Official(desc)) => DistributableToolchain::new(cfg, desc.clone())?
+            .components()?
+            .into_iter()
+            .filter_map(
+                |c| match (c.installed, c.component.short_name_in_manifest().as_str()) {
+                    (true, "rust-std") => {
```
rami3l: Nit: Personally I guess `"rust-std" if c.installed` is even better in terms of readability than matching a pair.

rami3l (approved):
Many thanks and LGTM!

PS: I'd like to point out that this .and_then() thing is existing code so not necessarily related to https://github.com/rust-lang/rustup/pull/4333. It is true though that we should consider overall readability in addition to direct modifications.

djc:
> PS: I'd like to point out that this `.and_then()` thing is existing code so not necessarily related to #4333.

Yes, the top-level `and_then()` might have been there before, but this added a bunch of nesting and chaining that made it pretty unclear at a glance what was going on.

src/cli/rustup_mode.rs
djc: Changed it to:

```rust
(c.installed && c.component.short_name_in_manifest() == "rust-std")
    .then(|| c.component.target.expect("rust-std should have a target"))
```

PR #4362, opened 2025-06-02
Increase Windows main thread stack size to 2mb

ChrisDenton:
This works around an issue with increased stack space usage in debug mode with clap 4.5.3 (see https://github.com/rust-lang/rustup/pull/4357).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -67,4 +67,9 @@ fn main() {
     // Turning them into errors forces them to be displayed (and the build to fail).
     // If we do want to ignore specific warnings then `/IGNORE:` should be used.
     println!("cargo::rustc-link-arg-bin=rustup-init=/WX");
+
+    // Increase the stack size to 2 mb.
+    // Works around issue with the clap parser using a lot of stack space in debug mode.
+    // See https://github.com/clap-rs/clap/issues/5134
+    println!("cargo::rustc-link-arg-bin=rustup-init=/STACK:0x200000");
 }


rami3l (commented):
I don't want to be a blocker, but it looks weird to me because I believe yourself have suggested https://github.com/rust-lang/rustup/commit/bc0acdbc8917e537a6d8ea9e82ba827179aa52b1. How is that change related to this new one? (Is `RUST_MIN_STACK` implicitly not for the main thread stack?)

rami3l (approved):

ChrisDenton:
`RUST_MIN_STACK` is a runtime argument that affects the stack size the standard library uses when spawning new threads. This PR affects the OS created main thread.

djc:
> Specifically the derived `augment_args` on [UpdateOpts](https://github.com/rust-lang/rustup/blob/4de10367b91e874ff12e593725f85c2969e45e21/src/cli/rustup_mode.rs#L352) requires 50,808 bytes of stack (half the default stack size on Windows) and it appears twice on the stack so it overflows.

Could we instead avoid this being duplicated on the twice? Seems unnecessary...

djc:
> > Specifically the derived `augment_args` on [UpdateOpts](https://github.com/rust-lang/rustup/blob/4de10367b91e874ff12e593725f85c2969e45e21/src/cli/rustup_mode.rs#L352) requires 50,808 bytes of stack (half the default stack size on Windows) and it appears twice on the stack so it overflows.
> 
> Could we instead avoid this being duplicated on the twice? Seems unnecessary...

I guess that happens in the derived code, so it's not trivial.

djc (approved):

ChrisDenton:
Yeah this is a clap issue and should be fixed over there.

Issue #4359, opened 2025-06-02
Tracking: Migrate to snapshot-based integration testing APIs


rami3l:
Borrowing from [Cargo's previous effort](https://github.com/rust-lang/cargo/issues/14039), this proposal tries to integrate a new testing DSL for better overall dev experience.

The idea is to make use of your IDE (or CLI, if you prefer) to update snapshots directly, instead of writing your expectations by hand:

<img width="529" alt="image" src="https://github.com/user-attachments/assets/6fa7bc62-6fa8-4efd-9670-107e25af5c76" />

I have a more ambitious goal of removing dependency injections altogether, but we must start from somewhere, so here it is.

## Implementation History
- #4334
- #4342 
- #4343 
- #4352
- #4353
- #4354
- #4363
- #4365
- #4366 
- #4367
- #4376

PR #4365, opened 2025-06-03
test(cli-rustup): migrate to `.expect()` APIs

rami3l:
Part of #4359.

> [!WARNING]
> This PR is very long. ~~I won't suggest reviewing it until it is marked ready.~~

[_SemanticDiff_](https://app.semanticdiff.com/gh/rust-lang/rustup/pull/4365/commits)

### DIFF
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -101,6 +101,15 @@ impl Assert {
         self
     }
 
+    pub fn remove_redactions(&mut self, vars: impl IntoIterator<Item = &'static str>) -> &mut Self {
+        for var in vars {
+            self.redactions
+                .remove(var)
+                .expect("invalid redactions detected");
+        }
+        self
+    }
+
     /// Asserts that the command exited with an ok status.
     pub fn is_ok(&self) -> &Self {
         assert!(self.output.ok);
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -224,20 +224,32 @@ impl Config {
     /// Returns an [`Assert`] object to check the output of running the command
     /// specified by `args` under the default environment.
     #[must_use]
-    pub async fn expect(&self, args: impl AsRef<[&str]>) -> Assert {
+    pub async fn expect<S: AsRef<OsStr> + Clone + Debug>(&self, args: impl AsRef<[S]>) -> Assert {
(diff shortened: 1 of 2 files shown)


src/test/clitools.rs
```diff
@@ -224,20 +224,32 @@ impl Config {
     /// Returns an [`Assert`] object to check the output of running the command
     /// specified by `args` under the default environment.
     #[must_use]
-    pub async fn expect(&self, args: impl AsRef<[&str]>) -> Assert {
+    pub async fn expect<S: AsRef<OsStr> + Clone + Debug>(&self, args: impl AsRef<[S]>) -> Assert {
```
djc, line 236: Pretty sure you can inline this as `args: impl AsRef<[impl AsRef<OsStr> + Clone + Debug]>`, which IMO is easier to read. It's not obvious to me why you need the `Clone` and `Debug` bounds?

djc (approved):

src/test/clitools.rs
rami3l: > Pretty sure you can inline this as `args: impl AsRef<[impl AsRef<OsStr> + Clone + Debug]>`, which IMO is easier to read.

@djc I would've written it that way if it's allowed:

```rs
error[E0666]: nested `impl Trait` is not allowed
 --> src/lib.rs:2:26
  |
2 | fn foo(args: impl AsRef<[impl AsRef<OsStr> + Clone + Debug]>) { 
  |              ------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^--
  |              |           |
  |              |           nested `impl Trait` here
  |              outer `impl Trait`
```

> It's not obvious to me why you need the `Clone` and `Debug` bounds?

They seem to be required by the upstream:

https://github.com/rust-lang/rustup/blob/5d31b52d41da5791569f3a772bb85c788862ed84/src/test/clitools.rs#L429-L431
https://github.com/rust-lang/rustup/blob/5d31b52d41da5791569f3a772bb85c788862ed84/src/test/clitools.rs#L436-L440
https://github.com/rust-lang/rustup/blob/5d31b52d41da5791569f3a772bb85c788862ed84/src/test/clitools.rs#L445

rami3l:
@djc Many thanks for the reviewing efforts! The remaining patches should be significantly shorter :]

PR #4372, opened 2025-06-05
test(download): serialize tests with proxy-sensitive URLs

cuviper:
There was already a guard for tests that *set* a proxy variable in the
environment, but this is also relevant for tests that would implicitly
read those variables while downloading a local `serve_file` URL.

### DIFF
--- a/src/download/tests.rs
+++ b/src/download/tests.rs
@@ -1,8 +1,10 @@
 use std::convert::Infallible;
+use std::env::remove_var;
 use std::fs;
 use std::io;
 use std::net::SocketAddr;
 use std::path::Path;
+use std::sync::LazyLock;
 use std::sync::mpsc::{Sender, channel};
 use std::thread;
 
@@ -20,7 +22,7 @@ mod curl {
 
     use url::Url;
 
-    use super::{serve_file, tmp_dir, write_file};
+    use super::{scrub_env, serve_file, tmp_dir, write_file};
     use crate::download::{Backend, Event};
 
     #[tokio::test]
@@ -43,6 +45,7 @@ mod curl {
 
     #[tokio::test]
     async fn callback_gets_all_data_as_if_the_download_happened_all_at_once() {
+        let _guard = scrub_env().await;
         let tmpdir = tmp_dir();
         let target_path = tmpdir.path().join("downloaded");
         write_file(&target_path, "123");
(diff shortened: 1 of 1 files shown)


cuviper:
Without this change, I could pretty reliably induce this race by narrowing the testsuite, so they're more likely to run in parallel.

```
$ cargo test --features test --lib -- download::tests::
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running unittests src/lib.rs (target/debug/deps/rustup-d6143b76edd93be0)

running 6 tests
test download::tests::curl::partially_downloaded_file_gets_resumed_from_byte_offset ... ok
test download::tests::curl::callback_gets_all_data_as_if_the_download_happened_all_at_once ... FAILED
test download::tests::reqwest::resume_partial_from_file_url ... ok
test download::tests::reqwest::callback_gets_all_data_as_if_the_download_happened_all_at_once ... FAILED
test download::tests::reqwest::socks_proxy_request ... FAILED
test download::tests::reqwest::read_basic_proxy_params ... ok

failures:

---- download::tests::curl::callback_gets_all_data_as_if_the_download_happened_all_at_once stdout ----

thread 'download::tests::curl::callback_gets_all_data_as_if_the_download_happened_all_at_once' panicked at src/download/tests.rs:85:14:
Test download failed: error during download

Caused by:
    [97] proxy handshake error (connection to proxy closed)

---- download::tests::reqwest::callback_gets_all_data_as_if_the_download_happened_all_at_once stdout ----

thread 'download::tests::reqwest::callback_gets_all_data_as_if_the_download_happened_all_at_once' panicked at src/download/tests.rs:252:14:
Test download failed: error downloading file

Caused by:
    0: error sending request
    1: client error (Connect)
    2: socks connect error: unexpected end of file

---- download::tests::reqwest::socks_proxy_request stdout ----

thread 'download::tests::reqwest::socks_proxy_request' panicked at src/download/tests.rs:187:13:
  left: 3
 right: 1


failures:
    download::tests::curl::callback_gets_all_data_as_if_the_download_happened_all_at_once
    download::tests::reqwest::callback_gets_all_data_as_if_the_download_happened_all_at_once
    download::tests::reqwest::socks_proxy_request

test result: FAILED. 3 passed; 3 failed; 0 ignored; 0 measured; 114 filtered out; finished in 0.01s

error: test failed, to rerun pass `--lib`
```

Note that the `socks_proxy_request` saw 3 requests when it expected 1, because the two callback tests were also using that proxy environment variable.

rami3l:
@cuviper Thanks a lot for your investigation! This does look like one of the reasons why Rustup's CI is not as stable as expected. I'll have a look at your patch real soon :)

rami3l (approved):
LGTM!

Nit: Before merging, it'd be probably better to change the commit message to "all download tests" instead of just "all tests"... I know by "proxy-sensitive URLs" it implies download tests, but it looks a bit misleading at the first glance 💦 ~~Many thanks in advance~~ I decided to do that myself, never mind :)

djc (approved):
Nice, thanks!

PR #4380, opened 2025-06-09
add Solaris support

### LINKED ISSUE #2987: SunOS/Solaris/Illumos support

Follow up to a closed issue  [SunOS/Solaris/Illumos support #1320](https://github.com/rust-lang/rustup/issues/1320)  

As of now Solaris ships with rustc 1.53.0 and cargo 1.53 and now "Common Build Environment (CBE) releases for the Oracle Solaris 11.4" the use of it could increase.

Also, more applicationes are using rust and rustup i.e. https://www.clamav.net/ now requires rust to deploy the very popular Virus scanner.

#  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh 
rustup: unrecognized OS type: SunOS

Thanks again for your great work.

Andrew

### Solution you'd like

rustup to work with SunOS

### Notes

_No response_

psumbera:
Rustup should be built also for Solaris (after https://github.com/rust-lang/rust/pull/138699).

### DIFF
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -29,6 +29,7 @@ jobs: # skip-master skip-pr skip-stable
           - i686-unknown-linux-gnu         # skip-pr skip-master
           - arm-unknown-linux-gnueabi      # skip-pr skip-master
           - arm-unknown-linux-gnueabihf    # skip-pr skip-master
+          - x86_64-pc-solaris              # skip-pr skip-master
           - x86_64-unknown-freebsd         # skip-pr skip-master
           - x86_64-unknown-netbsd          # skip-pr skip-master
           - x86_64-unknown-illumos         # skip-pr skip-master
@@ -36,6 +37,7 @@ jobs: # skip-master skip-pr skip-stable
           - powerpc64le-unknown-linux-gnu  # skip-pr skip-master
           - powerpc64le-unknown-linux-musl # skip-pr skip-master
           - s390x-unknown-linux-gnu        # skip-pr skip-master
+          - sparcv9-sun-solaris            # skip-pr skip-master
           - arm-linux-androideabi          # skip-pr skip-master
           - armv7-linux-androideabi        # skip-pr skip-master
           - x86_64-linux-android           # skip-pr skip-master
--- a/ci/docker/x86_64-pc-solaris/Dockerfile
+++ b/ci/docker/x86_64-pc-solaris/Dockerfile
@@ -1,0 +1,12 @@
+FROM rust-x86_64-pc-solaris
+
+# Building `aws-lc-rs` for Solaris on Linux depends on `gcc-multilib`, `libclang` and `bindgen`.
+# See: https://aws.github.io/aws-lc-rs/requirements/linux
+RUN apt-get update && apt-get install -qy gcc-multilib libclang-dev
+
+ENV \
+  AR_x86_64_pc_solaris=x86_64-solaris-ar \
+  RANLIB_x86_64_pc_solaris=x86_64-solaris-ranlib \
+  CC_x86_64_pc_solaris=x86_64-solaris-gcc \
+  CXX_x86_64_pc_solaris=x86_64-solaris-g++ \
+  CARGO_TARGET_X86_64_PC_SOLARIS_LINKER=x86_64-solaris-gcc
--- a/doc/user-guide/src/installation/other.md
+++ b/doc/user-guide/src/installation/other.md
@@ -159,10 +159,14 @@ You can manually download `rustup-init` for a given target from
   - [sha256 file](https://static.rust-lang.org/rustup/dist/powerpc64le-unknown-linux-musl/rustup-init.sha256)
 - [s390x-unknown-linux-gnu](https://static.rust-lang.org/rustup/dist/s390x-unknown-linux-gnu/rustup-init)
   - [sha256 file](https://static.rust-lang.org/rustup/dist/s390x-unknown-linux-gnu/rustup-init.sha256)
+- [sparcv9-sun-solaris](https://static.rust-lang.org/rustup/dist/sparcv9-sun-solaris/rustup-init)
+  - [sha256 file](https://static.rust-lang.org/rustup/dist/sparcv9-sun-solaris/rustup-init.sha256)
 - [x86_64-apple-darwin](https://static.rust-lang.org/rustup/dist/x86_64-apple-darwin/rustup-init)
   - [sha256 file](https://static.rust-lang.org/rustup/dist/x86_64-apple-darwin/rustup-init.sha256)
 - [x86_64-linux-android](https://static.rust-lang.org/rustup/dist/x86_64-linux-android/rustup-init)
   - [sha256 file](https://static.rust-lang.org/rustup/dist/x86_64-linux-android/rustup-init.sha256)
+- [x86_64-pc-solaris](https://static.rust-lang.org/rustup/dist/x86_64-pc-solaris/rustup-init)
+  - [sha256 file](https://static.rust-lang.org/rustup/dist/x86_64-pc-solaris/rustup-init.sha256)
 - [x86_64-pc-windows-gnu](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-gnu/rustup-init.exe)[^msys2]
   - [sha256 file](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-gnu/rustup-init.exe.sha256)
 - [x86_64-pc-windows-msvc](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe)[^msvc]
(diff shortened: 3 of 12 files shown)


rami3l:
@psumbera Thanks for your contribution! Have you (temporarily) removed `skip-pr` to get a CI run to prove the correctness of this PR though?

psumbera:
> @psumbera Thanks for your contribution! Have you (temporarily) removed `skip-pr` to get a CI run to prove the correctness of this PR though?

No. To be honest I'm trying to find out how it's working...

So I removed them from `ci/actions-templates/linux-builds-template.yaml` for now.

rami3l:
@psumbera As a friendly hint, we often do it like so: do the normal PR first, add a `DROP ME` commit on top that does the temporary hack, and when the CI turns green, rebase the PR and drop that commit.

rami3l:
> > @psumbera Thanks for your contribution! Have you (temporarily) removed `skip-pr` to get a CI run to prove the correctness of this PR though?
> 
> No. To be honest I'm trying to find out how it's working...
> 
> So I removed them from `ci/actions-templates/linux-builds-template.yaml` for now.

@psumbera That's right, but you'll also need to run the [shell script](https://github.com/rust-lang/rustup/blob/5cc22865075c7cc8ded61b1582db5c62d37a2790/ci/actions-templates/gen-workflows.sh) to update the GitHub CI YAML to keep them in sync, otherwise the CI will fail anyway.

ci/docker/x86_64-pc-solaris/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM rust-x86_64-pc-solaris
+
+# Building `aws-lc-rs` for Illumos on Linux depends on `gcc-multilib`, `libclang` and `bindgen`.
```
rami3l: ... for Solaris.

doc/user-guide/src/installation/other.md
```diff
@@ -167,6 +169,8 @@ You can manually download `rustup-init` for a given target from
   - [sha256 file](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-gnu/rustup-init.exe.sha256)
 - [x86_64-pc-windows-msvc](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe)[^msvc]
   - [sha256 file](https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe.sha256)
+- [x86_64-pc-solaris](https://static.rust-lang.org/rustup/dist/x86_64-pc-solaris/rustup-init)
```
rami3l: Although not machine-enforced, we'd still wish to maintain the alphabetical order.

src/dist/mod.rs
```diff
@@ -532,7 +533,12 @@ impl TargetTriple {
                 (b"NetBSD", b"x86_64") => Some("x86_64-unknown-netbsd"),
                 (b"NetBSD", b"i686") => Some("i686-unknown-netbsd"),
                 (b"DragonFly", b"x86_64") => Some("x86_64-unknown-dragonfly"),
-                (b"SunOS", b"i86pc") => Some("x86_64-unknown-illumos"),
+                (b"SunOS", b"i86pc") => Some(if version.starts_with(b"11.4") {
```
rami3l: Sorry, but `version.starts_with(b"11.4")` looks way too hacky. Is it possible to reuse the logic from our shell script instead (of course, write it in Rust)?

https://github.com/rust-lang/rustup/blob/c4adcc262d54c55f9a9885406da3899079b0cd90/rustup-init.sh#L367-L373

rami3l (changes requested):

src/dist/mod.rs
psumbera: `-o` / `--operating-system` is only supported via `uname(1)` command. Not using system call `uname(2)`. So I would have to call external program and parse its output.

On the other hand. Rust is supported only for Solaris 11.4 (https://doc.rust-lang.org/beta/rustc/platform-support/solaris.html). And there is no new Solaris releas. It's meant to be 11.4 forever :-)

Not sure where this code is used. Maybe I can use somehow `#[cfg(target_os = "solaris")]` and `#[cfg(target_os = "illumos")]`?

rami3l: @psumbera Thanks for the clarification! I believe the runtime behavior is here to rule out "compatibility mode", that is, running a rustup binary built for another platform for whatever reason. In this case, rustup is expected to "see through" the binary runtime and find out the target triple that belongs to the current machine.

Thus, the question that we should be asking is: is there some sort of binary compatibility between Solaris and Illumos? If not then conditional compilation should be fine. I'm not really familiar with these operating systems, so...

rami3l: @psumbera OTOH if we have to detect this at runtime and given that no other operating system will be clashing with Solaris 11.4 on these criteria, then I can accept the code as it is.

psumbera: There is no binary compatibility between `Solaris` and `Illumos`.

src/dist/mod.rs
```diff
@@ -497,7 +497,7 @@ impl TargetTriple {
             let mut sys_info;
             let (sysname, machine) = unsafe {
                 sys_info = mem::zeroed();
-                if libc::uname(&mut sys_info) != 0 {
+                if libc::uname(&mut sys_info) == -1 {
```
rami3l, line 500: Would you mind clarifying the necessity of this change?

src/dist/mod.rs
```diff
@@ -532,7 +532,6 @@ impl TargetTriple {
                 (b"NetBSD", b"x86_64") => Some("x86_64-unknown-netbsd"),
                 (b"NetBSD", b"i686") => Some("i686-unknown-netbsd"),
                 (b"DragonFly", b"x86_64") => Some("x86_64-unknown-dragonfly"),
-                (b"SunOS", b"i86pc") => Some("x86_64-unknown-illumos"),
```
rami3l, line 535: Nit: Wouldn't it be simpler if we just put `#[cfg()]`s on match arms instead in this case? I'm a bit afraid to create more "android" blocks; that one block is required because on Android `(_, xxx)` shadows many other arms... Otherwise it seems easier to manage all of them in one place despite many branches being dead.

src/dist/mod.rs
psumbera: On Solaris it can return for example 1 as success:
https://docs.oracle.com/cd/E88353_01/html/E37841/uname-2.html
Probably also on Illumos:
https://www.smartos.org/man/2/uname

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -23,6 +23,8 @@ jobs: # skip-master skip-pr skip-stable
           - armv7-unknown-linux-gnueabihf
           - aarch64-linux-android
           - aarch64-unknown-linux-gnu
+          - x86_64-pc-solaris
```
rami3l: Please don't forget to bring back `# skip-pr skip-master` :)

rami3l (approved):
Many thanks for this PR, and LGTM (modulo the comments)!

src/dist/mod.rs
psumbera: Ops, I think I'm confused here. Can you please more elaborate? Thank you!

rami3l: @psumbera I was thinking about something like:

```rs
#[cfg(target_os = "solaris")]
(b"SunOS", b"i86pc") => Some("x86_64-pc-solaris"),
```

... instead of a separate `match` block.

rami3l:
@psumbera Many thanks again for your contribution!

As a follow-up question, is browser detection on https://dev.rustup.rs working alright on Solaris? (I assume it's working alright on x64, but you are adding support for `sparcv9` in this PR as well which seems to be an architecture never supported before by rustup, so I'm a bit concerned.)

For example, for other unices we have:
https://github.com/rust-lang/rustup/blob/452c8e5c79c0da3530c2e8b5754156e5878777a9/www/rustup.js#L36-L40

psumbera:
> As a follow-up question, is browser detection on https://dev.rustup.rs working alright on Solaris? (I assume it's working alright on x64, but you are adding support for `sparcv9` in this PR as well which seems to be an architecture never supported before by rustup, so I'm a bit concerned.)

So far I wasn't able to test it. `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs` still returns the script without my changes.

rami3l:
> So far I wasn't able to test it. 

@psumbera Sorry for not being clear enough previously; I meant the Web frontend part of the detection specifically, since we'd normally assume a host platform (anything tier 1, or tier 2 with host tools) can run a browser and the user might want to go to the website for installation instructions.

> `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs` still returns the script without my changes.

This is expected; merging the PR only gets it into our code base and the actual deployment will happen in the next release. Although I cannot say for sure when that will come, I'd say that we are trying our best to make releases happen more often.

psumbera:
> As a follow-up question, is browser detection on https://dev.rustup.rs working alright on Solaris? (I assume it's working alright on x64, but you are adding support for `sparcv9` in this PR as well which seems to be an architecture never supported before by rustup, so I'm a bit concerned.)

See attached screen shot what I get on Solaris SPARC in Firefox 128.9.0. I think it's ok.

Screenshot 2025-06-11 at 9 25 13

PR #4382, opened 2025-06-11
Fix rustup-init.sh cputype check for sparcv9

psumbera:
This is follow up after https://github.com/rust-lang/rustup/pull/4380.

@rami3l sorry for not seeing it in first place.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -415,7 +415,7 @@ get_architecture() {
             ;;
 
         SunOS)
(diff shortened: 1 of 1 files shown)


rami3l (approved):

rami3l:
@psumbera No worries, and thanks for the fix! As we don't enable the full test suite for Tier 2, in this case we rely on the contributor's judgement.

PR #4384, opened 2025-06-14
Emphasize that `llvm-tools` dist component is not subject to compiler stability guarantees

jieyouxu:
See discussions over at [#t-compiler > &#96;llvm-tools-preview&#96; dist component](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/.60llvm-tools-preview.60.20dist.20component/with/524067760).


rami3l (approved):
Looks reasonable to me. Thanks!

Issue #3132, opened 2023-01-03
Check channel updates in parallel


gyscos:
### Problem you are trying to solve

In part thanks to `cargo-msrv`, I have ~20 (and growing!) pinned stable versions of rust installed through rustup.
Every time I run `rustup update`, it seems to sequentially check for any update on each of these 20 versions, making this operation much slower than with only a couple channels.

I believe checking each of these 20 channels in parallel might improve the situation.

Here is a recording of running `rustup update`:

https://asciinema.org/a/yo7PjqQRII2YCrvG1UEstTmrc

### Solution you'd like

Rather than checking each channel sequentially, checking them in parallel would improve the runtime of the command.

Note that this proposal is mostly for _checking_ for updates. I do not suggest actually downloading and installing updates in parallel, though it might be an option as well.

rbtcollins:
Completely agree. In fact we'd like to move the core to async, but noone has started on this yet.

r3bu1ld3r:
Hi! If you don't mind, I'd like to work on this

djc:
@r3bu1ld3r sounds great, let us know if you need any guidance.

rami3l:
Closed by #4388.

Starbrightcandy:
### 

> pkg:cargo/log@0.4.29

PR #4388, opened 2025-06-18
Check for updates concurrently

FranciscoTGouveia:
This change is related to #3132.

Currently, the `rustup check` command verifies updates for each toolchain sequentially.
This patch modifies that behavior to perform the checks concurrently.

To implement this change, the `futures_util` crate was introduced. I would appreciate feedback on this choice, and whether there might be a preferable alternative for handling concurrency.

My first attempt at adding concurrency caused all output to be printed only after every channel had been checked, which deviated from the current behavior of the command.
To preserve the existing behavior (reporting progress as each toolchain is checked), I tried printing concurrently as well.
However, this led to race conditions, resulting in inconsistent output order.

To address this, I used the `indicatif` crate (cc @djc) to display progress bars while maintaining the correct output order.

This PR will soon be updated to ensure progress bars are correctly written to _stdout_ in order to pass the tests.

Regarding performance, using `hyperfine`, a **2.2× speedup** was observed compared to the sequential version. These benchmarks were run 100 times (with 5 warmup runs) over a 50 Mbps connection.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ cfg-if = "1.0"
 chrono = { version = "0.4", default-features = false, features = ["std"] }
 clap = { version = "4", features = ["derive", "wrap_help"] }
 clap_complete = "4"
+console = "0.16"
 curl = { version = "0.4.44", optional = true }
 effective-limits = "0.5.5"
 enum-map = "2.5.0"
@@ -54,6 +55,7 @@ flate2 = { version = "1.1.1", default-features = false, features = ["zlib-rs"] }
 fs_at = "0.2.1"
 git-testament = "0.2"
 home = "0.5.4"
+indicatif = "0.18"
 itertools = "0.14"
 libc = "0.2"
 opener = "0.8.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ enum-map = "2.5.0"
 env_proxy = { version = "0.4.1", optional = true }
 flate2 = { version = "1.1.1", default-features = false, features = ["zlib-rs"] }
 fs_at = "0.2.1"
+futures-util = "0.3.31"
 git-testament = "0.2"
 home = "0.5.4"
 indicatif = "0.18"
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -6,11 +6,15 @@ use std::{
     path::{Path, PathBuf},
     process::ExitStatus,
     str::FromStr,
+    time::Duration,
 };
 
 use anyhow::{Context, Error, Result, anyhow};
 use clap::{Args, CommandFactory, Parser, Subcommand, ValueEnum, builder::PossibleValue};
 use clap_complete::Shell;
+use console::style;
+use futures_util::stream::StreamExt;
+use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
 use itertools::Itertools;
 use tracing::{info, trace, warn};
 use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
@@ -34,7 +38,7 @@ use crate::{
     install::{InstallMethod, UpdateStatus},
     process::{
         Process,
-        terminalsource::{self, ColorableTerminal},
+        terminalsource::{self, ColorChoice, ColorableTerminal},
     },
     toolchain::{
         CustomToolchainName, DistributableToolchain, LocalToolchainName,
@@ -792,45 +796,99 @@ async fn default_(
 }
 
 async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode> {
+    let t = cfg.process.stdout().terminal(cfg.process);
+    let is_a_tty = t.is_a_tty();
+    let use_colors = matches!(t.color_choice(), ColorChoice::Auto | ColorChoice::Always);
     let mut update_available = false;
-
-    let mut t = cfg.process.stdout().terminal(cfg.process);
     let channels = cfg.list_channels()?;
+    let num_channels = channels.len();
+    // Ensure that `.buffered()` is never called with 0 as this will cause a hang.
+    // See: https://github.com/rust-lang/futures-rs/pull/1194#discussion_r209501774
+    if num_channels > 0 {
+        let multi_progress_bars = if is_a_tty {
+            MultiProgress::with_draw_target(ProgressDrawTarget::term_like(Box::new(t)))
+        } else {
+            MultiProgress::with_draw_target(ProgressDrawTarget::hidden())
+        };
+        let channels = tokio_stream::iter(channels.into_iter()).map(|(name, distributable)| {
+            let pb = multi_progress_bars.add(ProgressBar::new(1));
+            pb.set_style(
+                ProgressStyle::with_template("{msg:.bold} - Checking... {spinner:.green}")
+                    .unwrap()
+                    .tick_chars("⠁⠂⠄⡀⢀⠠⠐⠈ "),
+            );
+            pb.set_message(format!("{name}"));
+            pb.enable_steady_tick(Duration::from_millis(100));
+            async move {
+                let current_version = distributable.show_version()?;
+                let dist_version = distributable.show_dist_version().await?;
+                let mut update_a = false;
 
-    for channel in channels {
-        let (name, distributable) = channel;
-        let current_version = distributable.show_version()?;
-        let dist_version = distributable.show_dist_version().await?;
-        let _ = t.attr(terminalsource::Attr::Bold);
-        write!(t.lock(), "{name} - ")?;
-        match (current_version, dist_version) {
-            (None, None) => {
-                let _ = t.fg(terminalsource::Color::Red);
-                writeln!(t.lock(), "Cannot identify installed or update versions")?;
+                let mut styled_name = style(format!("{name} - "));
+                if use_colors {
+                    styled_name = styled_name.bold();
+                }
+                let message = match (current_version, dist_version) {
+                    (None, None) => {
+                        let mut m = style("Cannot identify installed or update versions");
+                        if use_colors {
+                            m = m.red().bold();
+                        }
+                        format!("{styled_name}{m}")
+                    }
+                    (Some(cv), None) => {
+                        let mut m = style("Up to date");
+                        if use_colors {
+                            m = m.green().bold();
+                        }
+                        format!("{styled_name}{m} : {cv}")
+                    }
+                    (Some(cv), Some(dv)) => {
+                        let mut m = style("Update available");
+                        if use_colors {
+                            m = m.yellow().bold();
+                        }
+                        update_a = true;
+                        format!("{styled_name}{m} : {cv} -> {dv}")
+                    }
+                    (None, Some(dv)) => {
+                        let mut m = style("Update available");
+                        if use_colors {
+                            m = m.yellow().bold();
+                        }
+                        update_a = true;
+                        format!("{styled_name}{m} : (Unknown version) -> {dv}")
+                    }
+                };
+                pb.set_style(ProgressStyle::with_template(message.as_str()).unwrap());
+                pb.finish();
+                Ok::<(bool, String), Error>((update_a, message))
             }
-            (Some(cv), None) => {
-                let _ = t.fg(terminalsource::Color::Green);
-                write!(t.lock(), "Up to date")?;
-                let _ = t.reset();
-                writeln!(t.lock(), " : {cv}")?;
-            }
-            (Some(cv), Some(dv)) => {
+        });
+
+        // If we are running in a TTY, we can use `buffer_unordered` since
+        // displaying the output in the correct order is already handled by
+        // `indicatif`.
+        let channels = if is_a_tty {
+            channels
+                .buffer_unordered(num_channels)
+                .collect::<Vec<_>>()
+                .await
+        } else {
+            channels.buffered(num_channels).collect::<Vec<_>>().await
+        };
+
+        let t = cfg.process.stdout().terminal(cfg.process);
+        for result in channels {
+            let (update_a, message) = result?;
+            if update_a {
                 update_available = true;
-                let _ = t.fg(terminalsource::Color::Yellow);
-                write!(t.lock(), "Update available")?;
-                let _ = t.reset();
-                writeln!(t.lock(), " : {cv} -> {dv}")?;
             }
-            (None, Some(dv)) => {
-                update_available = true;
-                let _ = t.fg(terminalsource::Color::Yellow);
-                write!(t.lock(), "Update available")?;
-                let _ = t.reset();
-                writeln!(t.lock(), " : (Unknown version) -> {dv}")?;
+            if !is_a_tty {
+                writeln!(t.lock(), "{message}")?;
             }
         }
     }
-
     let self_update_mode = cfg.get_self_update_mode()?;
     // Priority: no-self-update feature > self_update_mode > no-self-update args.
     // Check for update only if rustup does **not** have the no-self-update feature,
--- a/src/process/terminalsource.rs
+++ b/src/process/terminalsource.rs
@@ -1,3 +1,5 @@
+use console::Term;
+use indicatif::TermLike;
 use std::{
     io::{self, Write},
     mem::MaybeUninit,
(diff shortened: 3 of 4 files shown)


FranciscoTGouveia:
After taking a closer look at the code, I realized that `buffered` can be replaced with `buffer_unordered` when running in a _tty_, since `indicatif` already ensures that output appears in the correct order.

To see if this had a practical impact, I re-ran the benchmarks (using the same settings as before) and observed a **3.1x** speedup.

rami3l (approved):
Nice work overall 🎉 I have to suggest some other changes regarding some small details but modulo those everything seems to be going pretty well. Looking forward to landing this patch!

src/cli/rustup_mode.rs
```diff
@@ -812,7 +812,7 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
                         .unwrap()
                         .tick_chars("⠁⠂⠄⡀⢀⠠⠐⠈ "),
                 );
-                pb.set_message(format!("{} - Checking...", name));
+                pb.set_message(format!("{name} - Checking..."));
```
rami3l: These clippy lints seem to largely originate from your earlier modifications in this PR; in this case it is recommended to squash them back to the respective original commit(s). You can use the [`git absorb`](https://andrewlock.net/super-charging-git-rebase-with-git-absorb/) plugin to do this more easily if you are using git.

src/cli/rustup_mode.rs
```diff
@@ -795,6 +795,11 @@ async fn default_(
 async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode> {
     let t = cfg.process.stdout().terminal(cfg.process);
     let is_a_tty = t.is_a_tty();
+    let use_colors = match cfg.process.var("RUSTUP_TERM_COLOR").as_deref() {
```
rami3l: Do you think it's useful to expose the `ColorChoice` we have already calculated in `ColorableTerminal::new()` here like what we have done in `.is_a_tty()`, instead of repeating the fetching of `RUSTUP_TERM_COLOR` all over the place? I'd expect that to be reused in many other places in the future so...

src/cli/rustup_mode.rs
```diff
@@ -824,28 +825,38 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
...
                 };
-                pb.set_style(ProgressStyle::with_template("{msg}").unwrap());
-                pb.finish_with_message(message.clone());
+                pb.set_style(ProgressStyle::with_template(message.as_str()).unwrap());
+                pb.finish();
                 Ok::<(bool, String), Error>((update_a, message))
             },
         );
 
+        // If we are running in a TTY, we can use `buffer_unordered` since
```
rami3l, line 869: I suspect this PR can be entirely or partially squashed back to one or more of the previous commits following the same principle... This comment for example feels a bit off in this very commit.

src/cli/rustup_mode.rs
```diff
@@ -792,12 +792,17 @@ async fn default_(
...
 async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode> {
+    let t = cfg.process.stdout().terminal(cfg.process);
+    let is_a_tty = t.is_a_tty();
     let mut update_available = false;
     let channels = cfg.list_channels()?;
     let num_channels = channels.len();
     if num_channels > 0 {
-        let mp = MultiProgress::with_draw_target(ProgressDrawTarget::stdout());
-
+        let mp = if is_a_tty {
```
rami3l: Bonus: `cargo` uses the [`CARGO_TERM_PROGRESS_WHEN`](https://doc.rust-lang.org/cargo/reference/config.html#termprogresswhen) environment variable to control the drawing of the progress bar. I think it'd be interesting to mirror that here as well, maybe in a separate commit or PR.

src/process/terminalsource.rs
```diff
@@ -223,6 +232,81 @@ impl io::Write for ColorableTerminalLocked {
     }
 }
 
+impl TermLike for ColorableTerminal {
+    fn width(&self) -> u16 {
+        Term::stdout().size().1
```
rami3l, line 243: Bonus: `cargo` uses the [`CARGO_TERM_PROGRESS_WIDTH`](https://doc.rust-lang.org/cargo/reference/config.html#termprogresswidth) environment variable to control the terminal width being detected. I think it'd be interesting to mirror that here as well, maybe in a separate commit or PR.

src/cli/rustup_mode.rs
FranciscoTGouveia: Yes, I had the same thought. Fetching the `RUSTUP_TERM_COLOR` environment variable in multiple places could definitely hurt readability. I was just a bit hesitant to add yet another method to `ColorableTerminal`, but I completely agree with your suggestion. Thank you!

src/cli/rustup_mode.rs
```diff
@@ -793,44 +795,67 @@ async fn default_(
...
-            (None, Some(dv)) => {
-                update_available = true;
-                let _ = t.fg(terminalsource::Color::Yellow);
-                write!(t.lock(), "Update available")?;
-                let _ = t.reset();
-                writeln!(t.lock(), " : (Unknown version) -> {dv}")?;
-            }
-        }
+    let num_channels = channels.len();
+    if num_channels > 0 {
```
rami3l, line 807: Probably worth putting the link (https://github.com/rust-lang/futures-rs/pull/1194#discussion_r209501774) here in the comments saying that `.buffered(0)` would cause a hang.

src/cli/rustup_mode.rs
```diff
@@ -793,44 +795,67 @@ async fn default_(
...
+
+                let (message, style) = match (current_version, dist_version) {
+                    (None, None) => {
+                        let msg =
+                            format!("{} - Cannot identify installed or update versions", name);
+                        let style = ProgressStyle::with_template("{msg}").unwrap();
+                        (msg, style)
+                    }
+                    (Some(cv), None) => {
+                        let msg = format!("{} - Up to date : {}", name, cv);
```
rami3l: IIRC `clippy` would fail on this commit, I suggest that you do another rebase to make sure?

src/cli/rustup_mode.rs
```diff
@@ -797,6 +797,8 @@ async fn default_(
 async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode> {
     let t = cfg.process.stdout().terminal(cfg.process);
     let is_a_tty = t.is_a_tty();
+    let use_colors = t.color_choice() == terminalsource::ColorChoice::Auto
```
rami3l: Nit: `matches!()` would be a better choice here.

src/cli/rustup_mode.rs
```diff
@@ -793,44 +795,67 @@ async fn default_(
...
-                update_available = true;
-                let _ = t.fg(terminalsource::Color::Yellow);
-                write!(t.lock(), "Update available")?;
-                let _ = t.reset();
-                writeln!(t.lock(), " : (Unknown version) -> {dv}")?;
-            }
-        }
+    let num_channels = channels.len();
+    if num_channels > 0 {
+        let mp = MultiProgress::with_draw_target(ProgressDrawTarget::stdout());
```
djc: I suggest you pick a more explicit variable name than `mp` for this.

src/cli/rustup_mode.rs
```diff
@@ -793,44 +795,67 @@ async fn default_(
...
+                        .unwrap()
+                        .tick_chars("⠁⠂⠄⡀⢀⠠⠐⠈ "),
+                );
+                pb.set_message(format!("{} - Checking...", name));
+                pb.enable_steady_tick(std::time::Duration::from_millis(100));
+                pb
+            })
+            .collect();
+
+        let channels = tokio_stream::iter(channels.into_iter().zip(progress_bars))
```
djc: Why are `progress_bars` and `channels` created in separate loops? Suggest just using for-loops here instead of a trivial combinator chain with a long `map()` closure.

src/cli/rustup_mode.rs
```diff
@@ -794,12 +794,17 @@ async fn default_(
...
 async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode> {
+    let t = cfg.process.stdout().terminal(cfg.process);
+    let is_a_tty = t.is_a_tty();
     let mut update_available = false;
     let channels = cfg.list_channels()?;
     let num_channels = channels.len();
     if num_channels > 0 {
-        let mp = MultiProgress::with_draw_target(ProgressDrawTarget::stdout());
-
+        let mp = if is_a_tty {
```
djc: Given that this commit ends up modifying a lot of the code introduced in the previous commit, suggest that this should be squashed into the previous one.

Cargo.toml
```diff
@@ -52,8 +52,10 @@ enum-map = "2.5.0"
 env_proxy = { version = "0.4.1", optional = true }
 flate2 = { version = "1.1.1", default-features = false, features = ["zlib-rs"] }
 fs_at = "0.2.1"
+futures-util = "0.3.31"
 git-testament = "0.2"
 home = "0.5.4"
+indicatif = "0.17"
```
djc: Note that the indicatif 0.17 version that used console 0.16 was yanked, should use indicatif 0.18 here.

src/cli/rustup_mode.rs
```diff
@@ -810,11 +811,11 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
             .map(|(name, _)| {
                 let pb = mp.add(ProgressBar::new(1));
                 pb.set_style(
-                    ProgressStyle::with_template("{msg} {spinner:.green}")
+                    ProgressStyle::with_template("{msg:.bold} - Checking... {spinner:.green}")
```
djc: This too is primarily changing code introduced in the previous commit, just squash instead.

src/cli/rustup_mode.rs
djc: Why? This seems fine to me.

Do suggest importing `ColorChoice` from `terminalsource`.

src/cli/rustup_mode.rs
```diff
@@ -793,44 +795,67 @@ async fn default_(
...
+            .iter()
+            .map(|(name, _)| {
+                let pb = mp.add(ProgressBar::new(1));
+                pb.set_style(
+                    ProgressStyle::with_template("{msg} {spinner:.green}")
+                        .unwrap()
+                        .tick_chars("⠁⠂⠄⡀⢀⠠⠐⠈ "),
+                );
+                pb.set_message(format!("{} - Checking...", name));
+                pb.enable_steady_tick(std::time::Duration::from_millis(100));
```
djc: Nit: please import `Duration`.

djc (commented):
Did you benchmark the performance impact of one or more rustup commands?

Would also be good to add an animation/video of the new output.

src/cli/rustup_mode.rs
rami3l: @djc I was thinking about something like `matches!(t.color_choice(), ColorChoice::Auto | ColorChoice::Always)`.

djc: Ahh, yes, that would be better.

FranciscoTGouveia:
In order to evaluate the performance impact of these changes, I ran a series of benchmarks using `hyperfine`.
My setup included a 50 Mbps internet connection, 16 GB of RAM, and an i7-1165G7 processor.

I compared my modified version of the `rustup check` command against the current implementation on the master branch.
Each benchmark was executed 100 times, with 5 warmup runs beforehand.
To ensure a meaningful comparison, I checked **20** different toolchains for updates in each run.

The results showed a **3.3×** speedup over the current implementation.
This exceeded my initial expectations, thanks to the realization that, since `indicatif` already handles the output in the correct order, we could safely replace `buffered` with `buffer_unordered`.

To give a better understanding of what this change implies regarding the user experience, I leave below a short animation.
On the left, you’ll see the current (sequential) behavior, and on the right, the new (concurrent) behavior:

showcase

src/cli/rustup_mode.rs
```diff
@@ -792,45 +796,99 @@ async fn default_(
...
-                let _ = t.fg(terminalsource::Color::Red);
-                writeln!(t.lock(), "Cannot identify installed or update versions")?;
-            }
-            (Some(cv), None) => {
-                let _ = t.fg(terminalsource::Color::Green);
-                write!(t.lock(), "Up to date")?;
-                let _ = t.reset();
-                writeln!(t.lock(), " : {cv}")?;
+    let num_channels = channels.len();
+    // Ensure that `.buffered()` is never called with 0 as this will cause an hang.
```
rami3l: Nit: ... a hang.

src/cli/rustup_mode.rs
```diff
@@ -792,45 +796,99 @@ async fn default_(
...
+            channels
+                .buffer_unordered(num_channels)
+                .collect::<Vec<_>>()
+                .await
+        } else {
+            channels.buffered(num_channels).collect::<Vec<_>>().await
+        };
+
+        let t = cfg.process.stdout().terminal(cfg.process);
+        for result in channels.into_iter() {
```
rami3l: Nit: `.into_iter()` seems redundant here.

rami3l (approved):
That's much clearer, thanks!

src/cli/rustup_mode.rs
rami3l: @djc We are mapping over a stream rather than an iterator, so doing a plain `for` loop might not help here. Depending on the current configuration more combinators will be involved down the line, notably with `.buffered()` and `.buffer_unordered()`.

djc:
It seems like a substantial downside to me that the new version takes so long to show any output at all. IMO you should look into fixing that.

rami3l:
> It seems like a substantial downside to me that the new version takes so long to show any output at all. IMO you should look into fixing that.

@djc I believe you are reading the gif wrong? This is not a side-by-side comparison; these two commands are executed one after another, both `time`d so that you can see the second is faster.

@FranciscoTGouveia It's true that we would like to eliminate all such confusions in the final demo to be used in the blog post/release notes or whatever it may be.

FranciscoTGouveia:
I’m sorry, I should have showcased the implementation a bit better.

In reality, the new version displays output instantaneously. The delay you see between when the left side finishes and the right side starts is simply because I only ran the new version after the base version completed. The pause was caused by me switching panes in `tmux` during the recording.
Below is an updated version of the animation that I hope better illustrates the behavior:

showcase_final

djc:
Ahh, awesome, that's much more impressive!

src/cli/rustup_mode.rs
```diff
@@ -792,45 +796,99 @@ async fn default_(
...
-            (Some(cv), None) => {
-                let _ = t.fg(terminalsource::Color::Green);
-                write!(t.lock(), "Up to date")?;
-                let _ = t.reset();
-                writeln!(t.lock(), " : {cv}")?;
+    let num_channels = channels.len();
+    // Ensure that `.buffered()` is never called with 0 as this will cause a hang.
+    // See: https://github.com/rust-lang/futures-rs/pull/1194#discussion_r209501774
+    if num_channels > 0 {
+        let multi_progress_bars = if is_a_tty {
```
djc, line 808: Nit: extract the `MultiProgress::with_draw_target()` out of the `if` (also, I think a `match` would look better than an `if` for this).

rami3l:
As @FranciscoTGouveia and I have agreed to address the bonus points (https://github.com/rust-lang/rustup/pull/4388#discussion_r2187891633, https://github.com/rust-lang/rustup/pull/4388#discussion_r2187892015) in a later commit, I think this PR is ready to go.

Nice work @FranciscoTGouveia !

src/cli/rustup_mode.rs
rami3l: @djc Oops that has happened after the merge so... @FranciscoTGouveia Maybe you could fix it in a new PR? 🙏

FranciscoTGouveia: Yes, of course! Thank you for pointing this out! :)

FranciscoTGouveia: Doesn't this get overshadowed by this [comment](https://github.com/rust-lang/rustup/pull/4388#discussion_r2187891633)?

djc: Potentially! If so, feel free to ignore my comment.

FranciscoTGouveia: I realize now that it doesn’t overshadow the previous point, I am sorry.
I have made that change in the follow-up [PR](https://github.com/rust-lang/rustup/pull/4406).

PR #4368, opened 2025-06-04
Fix CI image names for downloading ARM and PowerPC artifacts

Kobzol:
The split happened in https://github.com/rust-lang/rust/pull/141078 and https://github.com/rust-lang/rust/pull/141042.

@marcoieni We should probably also cross check `rustup` everytime we update our dist job names.

Context: [#t-infra > Rustup CI fails to install Docker images again (Jun 2025)](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Rustup.20CI.20fails.20to.20install.20Docker.20images.20again.20.28Jun.202025.29/with/522266007)

PR #4394, opened 2025-06-27
ci(run): install `codegen-cli` with `cargo-binstall`

rami3l:
This pull request updates the installation process for `bindgen-cli` in the `ci/run.bash` file to use `cargo-binstall`, which provides a faster and more efficient installation method.

This is necessary because since https://github.com/rust-lang/rust-bindgen/releases/tag/v0.72.0, the original quick installation script is no longer present.

### DIFF
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -8,12 +8,9 @@ rustc -vV
 cargo -vV
 
 if [ -n "$INSTALL_BINDGEN" ]; then
-  if ! curl --proto '=https' --tlsv1.2 -LsSf https://github.com/rust-lang/rust-bindgen/releases/latest/download/bindgen-cli-installer.sh | sh -s -- --no-modify-path \
-    | grep "everything's installed!";
-    # Ignoring exit code since the script might fail to write the receipt after a successful installation.
-  then
-    cargo install --force --locked bindgen-cli
-  fi
+  # Install `cargo-binstall` first for faster installation.
+  curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
+  cargo binstall -y --force --locked bindgen-cli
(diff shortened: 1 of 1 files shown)


djc (approved):

rami3l:
One very weird (although not directly related to the optimization this PR is doing) thing is that `bindgen-cli` is failing to build (on master) in both `aarch64-unknown-linux-musl` and `powerpc64le-unknown-linux-musl` workflows and nothing else.

Furthermore, the error message always says:

```console
error: failed to compile `bindgen-cli v0.72.0`, intermediate artifacts can be found at `/checkout/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  cannot produce proc-macro for `clap_derive v4.1.0` as the target `x86_64-unknown-linux-gnu` does not support these crate types
```

We do know that `clap_derive` supports `x86_64-unknown-linux-gnu`. What could possibly go wrong?

PS: `cargo -vV` gives `host: x86_64-unknown-linux-gnu` and `os: Ubuntu 22.4.0 [64-bit]` for both workflows.

@0xPoe have you seen this error message before?
Also cc @Kobzol for the potential link with #4368...

Kobzol:
Not sure where the issue is, but maybe using e.g. rust-cache could be an alternative zo cargo-binstall.

rami3l:
> Not sure where the issue is, but maybe using e.g. rust-cache could be an alternative zo cargo-binstall.

@Kobzol Thanks for the reply! To be clear, however, the error I've mentioned above seems to be pre-existing and not directly related to this patch, as we will still need to get a recent `codegen-cli` installed in `aarch64-unknown-linux-musl` and `powerpc64le-unknown-linux-musl` in some way in the first place.

I'll be merging this patch first (since we need that installation anyway), and let's move the rest of the discussion to https://github.com/rust-lang/rustup/issues/4396.

PR #4395, opened 2025-06-27
feat(www): improve "copy" button style compatibility with Chromium

rami3l:
#4080 works pretty well on Firefox and Safari. However, recent Chromium builds fail to handle the "solid" button style correctly.

Given the wide use of Chromium-based browsers, this PR has removed the "solid" style together with the button border for better consistency across platforms. The position of the SVG icon and that of the caption have also been slightly adjusted to be more coherent with the new style.

Below is a before/after comparison on both Chromium and Firefox:

|          | Before                                                                                    | After                                                                                     |
|----------|-------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| Chromium | image | image |
| Firefox  | image | image |

### DIFF
--- a/www/rustup.css
+++ b/www/rustup.css
@@ -175,9 +175,7 @@ hr {
     height: 58px;
     margin: 0;
     padding: 0 5px 0 5px;
-    border-radius: 0 3px 3px 0;
-    border-left-width: 0px;
-    border-left-style: solid;
+    border-width: 0px;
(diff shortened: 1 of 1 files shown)


ChrisDenton (approved):
It's weird that chromium changed behaviour here but this looks like a reasonable alternative to me.

PR #4397, opened 2025-06-28
style: migrate the codebase to `let-chains`

rami3l:
This PR has mainly converted the old code to use the new `let-chains` thanks to the `clippy::collapsible_if` lint. The overall readability seems to have been improved in a bunch of places.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -49,10 +49,10 @@ fn handle_epipe(res: Result<utils::ExitCode>) -> Result<utils::ExitCode> {
     match res {
         Err(e) => {
             let root = e.root_cause();
-            if let Some(io_err) = root.downcast_ref::<std::io::Error>() {
-                if io_err.kind() == std::io::ErrorKind::BrokenPipe {
-                    return Ok(utils::ExitCode(0));
-                }
+            if let Some(io_err) = root.downcast_ref::<std::io::Error>()
+                && io_err.kind() == std::io::ErrorKind::BrokenPipe
+            {
+                return Ok(utils::ExitCode(0));
             }
             Err(e)
         }
@@ -774,10 +774,10 @@ async fn default_(
             }
         };
 
-        if let Some((toolchain, reason)) = cfg.active_toolchain()? {
-            if !matches!(reason, ActiveReason::Default) {
-                info!("note that the toolchain '{toolchain}' is currently in use ({reason})");
-            }
+        if let Some((toolchain, reason)) = cfg.active_toolchain()?
+            && !matches!(reason, ActiveReason::Default)
+        {
+            info!("note that the toolchain '{toolchain}' is currently in use ({reason})");
         }
     } else {
         let default_toolchain = cfg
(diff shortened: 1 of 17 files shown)


src/cli/rustup_mode.rs
```diff
@@ -49,10 +49,10 @@ fn handle_epipe(res: Result<utils::ExitCode>) -> Result<utils::ExitCode> {
     match res {
         Err(e) => {
             let root = e.root_cause();
-            if let Some(io_err) = root.downcast_ref::<std::io::Error>() {
-                if io_err.kind() == std::io::ErrorKind::BrokenPipe {
-                    return Ok(utils::ExitCode(0));
-                }
+            if let Some(io_err) = root.downcast_ref::<std::io::Error>()
```
djc: Pre-existing nit: let's import `io` to save some tokens?

src/config.rs
```diff
@@ -659,17 +659,15 @@ impl<'a> Cfg<'a> {
...
-                            // Permit fully qualified names IFF the toolchain is installed. TODO(robertc): consider
-                            // disabling this and backing out https://github.com/rust-lang/rustup/pull/2141 (but provide
-                            // the base name in the error to help users)
-                            let resolved_name = &ToolchainName::try_from(toolchain_name_str)?;
-                            if !self.list_toolchains()?.iter().any(|s| s == resolved_name) {
-                                return Err(anyhow!(format!(
-                                    "target triple in channel name '{name}'"
-                                )));
-                            }
+                    if let ResolvableToolchainName::Official(ref name) = toolchain_name
```
djc: Pre-existing nit: use match ergonomics to drop the `ref`?

src/download/mod.rs
```diff
@@ -469,18 +469,17 @@ mod curl {
...
-                                let msg = Event::DownloadContentLengthReceived(s + resume_from);
-                                match callback(msg) {
-                                    Ok(()) => (),
-                                    Err(e) => {
-                                        *cberr.borrow_mut() = Some(e);
-                                        return false;
-                                    }
-                                }
+                    let prefix = "content-length: ";
+                    if let Ok(data) = str::from_utf8(header)
```
djc: This is cool and stuff but I feel like it would be more readable as a series of early returns with let-else.

src/download/mod.rs
```diff
@@ -469,18 +469,17 @@ mod curl {
...
+                        && data.to_ascii_lowercase().starts_with(prefix)
+                        && let Ok(s) = data[prefix.len()..].trim().parse::<u64>()
+                    {
+                        let msg = Event::DownloadContentLengthReceived(s + resume_from);
+                        match callback(msg) {
```
djc: In particular, this `match` could be a let-else.

djc (approved):
Nice!

src/download/mod.rs
rami3l: Actually, if-let would be a better choice in this case, since the happy path does nothing.

src/download/mod.rs
rami3l: @djc I think the new version is a bit error prone because it repeats the default return value `true` so many times. What do you think?

PS: It's quite unfortunate that we can't use `let-chains` with `let-else` yet...

ChrisDenton (approved):
lgtm!

src/download/mod.rs
rami3l: It occurs to me that this `curl`-related code will be deleted sooner or later anyway, so I'm not going to cause much churn here blocking on this tiny detail :)

djc: Hopefully sooner!

PR #4404, opened 2025-07-08
Simplify updates

djc:
Slow followup from

- #4340

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -291,24 +291,20 @@ pub(crate) async fn update_all_channels(
         info!("no updatable toolchains installed");
     }
 
-    let show_channel_updates = || {
-        if !toolchains.is_empty() {
-            writeln!(cfg.process.stdout().lock())?;
+    if !toolchains.is_empty() {
+        writeln!(cfg.process.stdout().lock())?;
 
-            let t = toolchains
-                .into_iter()
-                .map(|(p, s)| (PackageUpdate::Toolchain(p), s))
-                .collect();
-            show_channel_updates(cfg, t)?;
-        }
-        Ok(())
-    };
+        let t = toolchains
+            .into_iter()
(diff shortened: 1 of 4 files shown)


src/cli/common.rs
```diff
@@ -291,24 +291,20 @@ pub(crate) async fn update_all_channels(
...
-        Ok(())
-    };
+    if !toolchains.is_empty() {
+        writeln!(cfg.process.stdout().lock())?;
+
+        let t = toolchains
+            .into_iter()
+            .map(|(p, s)| (PackageUpdate::Toolchain(p), s))
+            .collect();
+        show_channel_updates(cfg, t)?;
```
rami3l, line 300: Sorry, but I'm a bit confused about this... Do you know why are we showing channel updates during `self update` in the first place? That might help with understanding the current behavior...

rami3l (approved):
LGTM as most changes are pretty much innocent, modulo `Show channel updates even if self update is not allowed` which I think might need some more discussion 🤔

src/cli/common.rs
djc: https://github.com/rust-lang/rustup/pull/1605 mentions

> This will make sure people who do `rustup update stable` still get a new rustup

rami3l: @djc Thanks for mentioning that PR!

After another it looks like this change is basically inlining the `show_channel_updates()` calls, and the only difference is that in the original version might do other fallible things before printing the channel updates.

Given that, maybe I have misunderstood your original commit message, which gave me the impression of something related to `--no-self-update`. How about "Show channel updates even if self update is not *permitted*", so it's clear that it's related to `SelfUpdatePermission`?

rami3l: @djc If you are too occupied, I can offer to do this for you on your permission :)

PR #4406, opened 2025-07-09
Introduce `RUSTUP_TERM_WIDTH` and `RUSTUP_TERM_PROGRESS_WHEN`

FranciscoTGouveia:
Follow-up on #4388 .

As discussed in these comments, [1](https://github.com/rust-lang/rustup/pull/4388#discussion_r2187891633) and [2](https://github.com/rust-lang/rustup/pull/4388#discussion_r2187892015), new environment variables are introduced.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -805,11 +805,15 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
     // Ensure that `.buffered()` is never called with 0 as this will cause a hang.
     // See: https://github.com/rust-lang/futures-rs/pull/1194#discussion_r209501774
     if num_channels > 0 {
-        let multi_progress_bars = if is_a_tty {
-            MultiProgress::with_draw_target(ProgressDrawTarget::term_like(Box::new(t)))
-        } else {
-            MultiProgress::with_draw_target(ProgressDrawTarget::hidden())
-        };
+        let multi_progress_bars =
+            MultiProgress::with_draw_target(match cfg.process.var("RUSTUP_TERM_PROGRESS_WHEN") {
+                Ok(s) if s.eq_ignore_ascii_case("always") => {
+                    ProgressDrawTarget::term_like(Box::new(t))
+                }
+                Ok(s) if s.eq_ignore_ascii_case("never") => ProgressDrawTarget::hidden(),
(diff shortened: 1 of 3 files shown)


src/cli/rustup_mode.rs
```diff
@@ -805,10 +805,17 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
     // Ensure that `.buffered()` is never called with 0 as this will cause a hang.
     // See: https://github.com/rust-lang/futures-rs/pull/1194#discussion_r209501774
     if num_channels > 0 {
-        let multi_progress_bars = if is_a_tty {
-            MultiProgress::with_draw_target(ProgressDrawTarget::term_like(Box::new(t)))
-        } else {
-            MultiProgress::with_draw_target(ProgressDrawTarget::hidden())
+        let multi_progress_bars = match cfg.process.var("RUSTUP_TERM_PROGRESS_WHEN") {
+            Ok(s) if s.eq_ignore_ascii_case("always") => {
+                MultiProgress::with_draw_target(ProgressDrawTarget::term_like(Box::new(t)))
```
rami3l, line 810: Nit: @djc's previous comment applies: extract `MultiProgress::with_draw_target()` to outside of this `match`.

src/process/terminalsource.rs
```diff
@@ -107,10 +108,16 @@ impl ColorableTerminal {
             #[cfg(all(test, feature = "test"))]
             StreamSelector::TestTtyWriter(w) => TerminalInner::TestWriter(w, choice),
         };
+        let width = process
```
djc, line 112: This can be simplified by deferring the conversion from `Result` to `Option` to later, I think:

```rust
let width = process
     .var("RUSTUP_TERM_WIDTH")
     .and_then(|s| u16::from_str(&s))
     .ok()
     .filter(|&n| n > 0);
```

Avoiding the turbofish by using the direct trait method also seems nicer to me.

djc (approved):

src/process/terminalsource.rs
rami3l: @djc That's not the same kind of error, I guess, which makes them not collapsible, unlike `Option`?

djc: Hmm, fair.

rami3l (approved):
LGTM, and many thanks :)

PR #4412, opened 2025-07-14
Further refactoring of self update logic

djc:
Previously:

- #4404 

Followup from:

- #4340

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -3,8 +3,6 @@
 use std::cell::RefCell;
 use std::fmt::Display;
 use std::fs;
-#[cfg(not(windows))]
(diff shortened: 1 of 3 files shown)


rami3l:
Thanks for this patch! 🙏 

The changes you made seemed innocent at first glance. However, the changes in the order of execution made me a little concerned, which is later confirmed by the test failures... 🤔

djc:
Scoped this down to just the first commit -- I don't think I'm motivated to pursue the rest of these changes in the near future.

rami3l (approved):
> I don't think I'm motivated to pursue the rest of these changes in the near future.

Many thanks all the same 🙏

PR #4414, opened 2025-07-15
opt(err): show renaming file error source

Binlogo:
## Problem you are trying to solve

We occasionally encounter the following error:

```
error: component download failed for cargo-aarch64-apple-darwin: could not rename downloaded file from '/Users/xxx/.rustup/downloads/344e710d......76ee75be1fff0c7bf3.partial' to '/Users/xxx/.rustup/downloads/344e710d......76ee75be1fff0c7bf3'
```

This issue is difficult to reproduce, so it's important to add a source for reporting first.

## Description

This pull request introduces a `RenamingFile` error with a `source` to improve error handling for file renaming operations in the `RustupError` enum. It updates the codebase to use this new error type, enhancing the clarity and consistency of error reporting.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -2,7 +2,6 @@
 
 use std::ffi::OsString;
 use std::fmt::Debug;
-#[cfg(not(windows))]
 use std::io;
 use std::io::Write;
 use std::path::PathBuf;
@@ -77,6 +76,13 @@ pub enum RustupError {
     RemovingDirectory { name: &'static str, path: PathBuf },
     #[error("could not remove '{name}' file: '{}'", .path.display())]
     RemovingFile { name: &'static str, path: PathBuf },
+    #[error("could not rename '{name}' file from '{src}' to '{dest}': {source}")]
+    RenamingFile {
+        name: &'static str,
+        src: PathBuf,
+        dest: PathBuf,
+        source: io::Error,
+    },
     #[error("{}", component_unavailable_msg(.components, .manifest, .toolchain))]
     RequestedComponentsUnavailable {
         components: Vec<Component>,
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -471,13 +471,14 @@ where
             },
         },
     )
-    .with_context(|| {
-        format!(
-            "could not rename {} file from '{}' to '{}'",
+    .map_err(|e| {
+        RustupError::RenamingFile {
             name,
-            src.display(),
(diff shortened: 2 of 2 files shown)


djc (commented):
Should the `std::error::Error::source()` method be implemented to yield the `source` error for this?

Binlogo:
> Should the `std::error::Error::source()` method be implemented to provide the `source` error for this?

@djc Thanks for asking, but I'm not sure, actually. I noticed that `SettingPermissions` is using `io::Error` here, so I followed the same pattern. Do you have any suggestions?

Binlogo:
> Should the `std::error::Error::source()` method be implemented to yield the `source` error for this?

According to the thiserror [documentation](https://docs.rs/thiserror/latest/thiserror/#details), it states:

> The Error trait’s source() method is implemented to return whichever field has a #[source] attribute or is named source, if any. This is for identifying the underlying lower level error that caused your error.
> 
> The #[from] attribute always implies that the same field is #[source], so you don’t ever need to specify both attributes.
> 
> Any error type that implements std::error::Error or dereferences to dyn std::error::Error will work as a source.
> ```
> #[derive(Error, Debug)]
> pub struct MyError {
>     msg: String,
>     #[source]  // optional if field name is `source`
>     source: anyhow::Error,
> }
> ```

Is it still necessary to implement the std::error::Error::source() method?

djc:
Sure, but your PR does not add any `#[source]` attribute, right?

Binlogo:
> Sure, but your PR doesn't add any `#[source]` attribute, right?

This PR didn't add `#[source]` because this macro is optional if the field name is `source`, as the documentation mentions.

So it should implemented `std::error::Error::source()` method by `thiserror` crate.

djc:
> > Sure, but your PR doesn't add any `#[source]` attribute, right?
> 
> This PR didn't add `#[source]` because this macro is optional if the field name is `source`, as the documentation mentions.
> 
> So it should implemented `std::error::Error::source()` method by `thiserror` crate.

Ahh, okay, thanks.

rami3l (approved):
Looks nice to me, thanks!

@djc Do you happen to have remaining concerns or all of them have been addressed already?

djc (approved):

djc:
@Binlogo just squash your commits please -- the second commit really belongs with the first one.

Binlogo:
@djc Squashed. Thanks for the review.

PR #4413, opened 2025-07-15
Set a maximum thread limit for the `remove_dir_all` crate

ChrisDenton:
Under the hood it uses rayon so we preinitialise rayon's global pool.

Follow up to #4407
Fixes #4403

PR #4407, opened 2025-07-09
Limit the default number of I/O threads

ChrisDenton:
This puts an upper bound on the number of I/O threads spawned. I'd appreciate if other people could try setting `RUSTUP_IO_THREADS` on their machine and seeing how much real-world impact setting a high number has.

Fixes #4403

PR #4417, opened 2025-07-18
Limit Tokio worker threads to I/O thread count

### LINKED ISSUE #4416: Error: "could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
Error message:
```
   default host triple: x86_64-unknown-linux-gnu
     default toolchain: stable (default)
               profile: default
  modify PATH variable: yes

1) Proceed with standard installation (default - just press enter)
2) Customize installation
3) Cancel installation
>

info: default host triple is x86_64-unknown-linux-gnu
error: failed to download file error=Reqwest(reqwest::Error { kind: Request, url: "https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256", source: TimedOut })
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256' to '/home/c/cm34324/.rustup/tmp/edwvukhsujhopr0i_file': error downloading file: error sending request for url (https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256): operation timed out
```

At first glance this is very strange issue, because **_curl_** or **_wget_** download that file without any error.

But I believe, that issue have same root cause as in #4403!
Before, I am using _rustup-**1.27.1**_ and have issue described in #4403
After fix #4403 in last _rustup_ ( > **_1.28.2_** ) I got this issue.
According my investigation, that strange behavior appears after this commit:
```
commit 11381dbf94222c46d4fc3de610d041eca85c65e7
Author: Robert Collins <robertc@robertcollins.net>
Date:   Mon May 29 20:00:22 2023 +0200

    Convert main to using a tokio runtime always
```

And again, as in #4403, one of crate ( now _**tokio**_  instead of **_rayon_**) try to use max number of CPUs 
on system (**104**) , but host set limit of CPUs using **40** (via _ulimit_)

If you check **_tokio_** code for function **num_cpus** in _src/loom/std/mod.rs_ , you can see that
this crate try to use **_std::thread::available_parallelism()_** - it not working correctly in that Linux system w/o this patchset: 
https://github.com/rust-lang/rust/pull/143614/commits

**_rayon_** crate using the same call, and follow to issue #4403 

Sure, you can fix this issue by temporary patch, like in #4413, but bugs will appears in another crates and places
Please, fix a root of such issues and test your code on conditions that described here and in #4403  (**_CPUs > ulimit_**)  !

One more moment: error message absolutely not related to real issue, no any exceptions  (where messages from system, that resources is exceeded), there are no clues where to start investigation

djc:
Fixes #4416. Previously:

- #4407 
- #4413

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -7,8 +7,6 @@ use std::io::{self, BufReader, Write};
 use std::ops::{BitAnd, BitAndAssign};
 use std::path::{Path, PathBuf};
 use std::process::ExitStatus;
-use std::str::FromStr;
-use std::thread;
 
 use anyhow::{Context, Result, anyhow, bail};
 use retry::delay::{Fibonacci, jitter};
@@ -27,26 +25,6 @@ pub(crate) mod notifications;
 pub(crate) mod notify;
 pub mod raw;
 pub(crate) mod units;
-
-pub fn io_thread_count(process: &Process) -> anyhow::Result<usize> {
-    if let Ok(n) = process.var("RUSTUP_IO_THREADS") {
-        let threads = usize::from_str(&n).context(
-            "invalid value in RUSTUP_IO_THREADS -- must be a natural number greater than zero",
-        )?;
-        match threads {
-            0 => bail!("RUSTUP_IO_THREADS must be a natural number greater than zero"),
-            _ => return Ok(threads),
-        }
-    };
-
-    Ok(match thread::available_parallelism() {
-        // Don't spawn more than 8 I/O threads unless the user tells us to.
-        // Feel free to increase this value if it improves performance.
-        Ok(threads) => Ord::min(threads.get(), 8),
-        // Unknown for target platform or no permission to query.
-        Err(_) => 1,
-    })
-}
 
 #[must_use]
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


src/utils/mod.rs
```diff
@@ -27,14 +28,19 @@ pub(crate) mod notify;
 pub mod raw;
 pub(crate) mod units;
 
-pub fn io_thread_count() -> usize {
-    match thread::available_parallelism() {
+pub fn io_thread_count(process: &Process) -> anyhow::Result<usize> {
+    if let Ok(n) = process.var("RUSTUP_IO_THREADS") {
+        return usize::from_str(&n)
+            .context("invalid value in RUSTUP_IO_THREADS. Must be a natural number");
```
rami3l: I'm afraid that has to be a *positive* number (`NonZero<usize>`), or you have something else planned for `0`?

rami3l (approved):
Many thanks!

src/utils/mod.rs
djc: Good point, adjusted.

rami3l (approved):

djc:
Since @ChrisDenton worked on this previously, I'll give him some time to give feedback before merging this.

ChrisDenton (approved):
Looks great to me, thanks!

Issue #3828, opened 2024-05-15
#3827 has broken the ETA format when downloading/installing components


rami3l:
> @djc Just tried the latest build and this looks incorrect.
>
> <img width="918" alt="image" src="https://github.com/rust-lang/rustup/assets/33851577/0a41d31a-8ffc-4fab-8d79-100383114e61">

_Originally posted by @rami3l in https://github.com/rust-lang/rustup/issues/3827#issuecomment-2112434768_

This regression has been discovered in the following version:

```console
> rustup --version
rustup 1.27.1 :: 1.27.0+133 (eca50133c 2024-05-15)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.78.0 (9b00956e5 2024-04-29)`
```

rami3l:
Suggestion: we should stick to the old format:

https://github.com/rust-lang/rustup/blob/9caf4fc77417c03aaf56aecee449e48d16485896/src/cli/download_tracker.rs#L188-L196

... which means there should be an `ETA: {}` somewhere.

My suggestion `write!(f, " ETA: {}", self.0.display()),` in https://github.com/rust-lang/rustup/pull/3827#discussion_r1601229010 should do exactly this.

djc:
Why should we stick to the old format? I agree with #3826 that showing it after finishing is pointless.

rami3l:
> Why should we stick to the old format? I agree with #3826 that showing it after finishing is pointless.

@djc Sorry, I meant stick to the old format when downloading. The expected result should be something like `in 1s ETA: Unknown`.

rami3l:
For example if ETA is 3s, the current version will print `in 1s3s`, which is clearly not ideal.

djc:
Ahh, no, I only applied half of your suggestion -- sorry, was too hasty; fix in #3829.

I was confused for a while by your screenshot showing "Unknown", but that seems unrelated.

rami3l:
> Ahh, no, I only applied half of your suggestion -- sorry, was too hasty; fix in #3829.
> 
> I was confused for a while by your screenshot showing "Unknown", but that seems unrelated.

@djc Sorry, that `Unknown` is caused by a division-by-zero I believe. I probably won't be able to make a better screenshot: it all goes too fast on my machine :]

djc:
All of this code looks pretty shitty and would like to use indicatif instead (#1835), but that'll have to wait for another day.

[...]

PR #4426, opened 2025-07-24
Refactor the DownloadTracker in favor of `indicatif`

### LINKED ISSUE #1835: Investigate using a community maintained progress bar

FranciscoTGouveia:
Closes #1835.

This PR refactors the current manual `DownloadTracker` progress reporting, replacing it with an `indicatif`-based implementation, as previously discussed in #1835 and #3828.

This change is a first step toward enabling concurrent component downloads, as the `MultiProgress` construct of `indicatif` will very helpful to concurrently report progress.

In terms of performance, benchmarks using `hyperfine` showed no noticeable difference.
Benchmarks were run 20 times (with 2 warmup runs) over a 50 Mbps connection, each downloading an entire toolchain.

Thanks to @0xPoe and his previous [PR](https://github.com/rust-lang/rustup/pull/3349), which served as the basis for this change.

Prior art: #2610, #2828, #3349.

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -1,68 +1,34 @@
-use std::collections::VecDeque;
-use std::fmt;
-use std::io::Write;
-use std::time::{Duration, Instant};
+use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
+use std::time::Duration;
 
 use crate::dist::Notification as In;
 use crate::notifications::Notification;
-use crate::process::{Process, terminalsource};
+use crate::process::Process;
 use crate::utils::Notification as Un;
-use crate::utils::units::{Size, Unit, UnitMode};
-
-/// Keep track of this many past download amounts
-const DOWNLOAD_TRACK_COUNT: usize = 5;
 
 /// Tracks download progress and displays information about it to a terminal.
 ///
 /// *not* safe for tracking concurrent downloads yet - it is basically undefined
 /// what will happen.
 pub(crate) struct DownloadTracker {
-    /// Content-Length of the to-be downloaded object.
-    content_len: Option<usize>,
-    /// Total data downloaded in bytes.
-    total_downloaded: usize,
-    /// Data downloaded this second.
-    downloaded_this_sec: usize,
-    /// Keeps track of amount of data downloaded every last few secs.
-    /// Used for averaging the download speed. NB: This does not necessarily
-    /// represent adjacent seconds; thus it may not show the average at all.
-    downloaded_last_few_secs: VecDeque<usize>,
-    /// Time stamp of the last second
-    last_sec: Option<Instant>,
-    /// Time stamp of the start of the download
-    start_sec: Option<Instant>,
-    term: terminalsource::ColorableTerminal,
-    /// Whether we displayed progress for the download or not.
-    ///
-    /// If the download is quick enough, we don't have time to
-    /// display the progress info.
-    /// In that case, we do not want to do some cleanup stuff we normally do.
-    ///
-    /// If we have displayed progress, this is the number of characters we
-    /// rendered, so we can erase it cleanly.
-    displayed_charcount: Option<usize>,
-    /// What units to show progress in
-    units: Vec<Unit>,
-    /// Whether we display progress
-    display_progress: bool,
-    stdout_is_a_tty: bool,
+    /// MultiProgress bar for the downloads.
+    multi_progress_bars: MultiProgress,
+    /// ProgressBar for the current download.
+    progress_bar: ProgressBar,
 }
 
 impl DownloadTracker {
     /// Creates a new DownloadTracker.
     pub(crate) fn new_with_display_progress(display_progress: bool, process: &Process) -> Self {
+        let multi_progress_bars = MultiProgress::with_draw_target(if display_progress {
+            process.progress_draw_target()
+        } else {
+            ProgressDrawTarget::hidden()
+        });
+
         Self {
-            content_len: None,
-            total_downloaded: 0,
-            downloaded_this_sec: 0,
-            downloaded_last_few_secs: VecDeque::with_capacity(DOWNLOAD_TRACK_COUNT),
-            start_sec: None,
-            last_sec: None,
-            term: process.stdout().terminal(process),
-            displayed_charcount: None,
-            units: vec![Unit::B],
-            display_progress,
-            stdout_is_a_tty: process.stdout().is_a_tty(process),
+            multi_progress_bars,
+            progress_bar: ProgressBar::hidden(),
         }
     }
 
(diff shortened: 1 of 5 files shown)


rami3l:
For the reference, the current style is like so on my machine:

image

... and the original goes like:

```
 20.2 MiB /  20.2 MiB (100 %)  20.1 MiB/s in  1s  
```

rami3l:
@FranciscoTGouveia I am thinking, the percentage has been replaced by the bar, but the speed meter might still be worth it (and I don't think the timer is that important)?

```
[███████████████████████████████████████░] 20.09 MiB/20.22 MiB (20.1 MiB/s, ETA: 0s) 
```

Also, I actually prefer the legacy progress bar style, something closer to `pacman` or `apt`:

```
[####################################    ] 20.09 MiB/20.22 MiB (20.1 MiB/s, ETA: 0s)
```

@djc @ChrisDenton What would you like (and don't like) to see in this interface that you might see every now and then?

src/cli/download_tracker.rs
```diff
@@ -1,291 +1,87 @@
...
+    progress_bar: ProgressBar,
 }
 
 impl DownloadTracker {
     /// Creates a new DownloadTracker.
     pub(crate) fn new_with_display_progress(display_progress: bool, process: &Process) -> Self {
+        let t = process.stdout().terminal(process);
+        let is_a_tty = t.is_a_tty();
+        let multi_progress_bars =
+            MultiProgress::with_draw_target(match process.var("RUSTUP_TERM_PROGRESS_WHEN") {
```
rami3l: Doesn't this overlap with some of our previous code? Could we extract a function first before actually adding this feature?

rami3l (approved):
Thanks for this PR and your prompt update afterwards!

It should be noted though that given the current commit guidelines, any back-and-forth in a single PR should be squashed back to the original commit.

Also, I'd like to wait for my teammates to see if the current style goes for them:

Image

ChrisDenton:
I'll admit to not having any strong opinions here but the current style looks fine to me. The ETA can be useful on slower connections so it's clear if you should go and do something else for 20 minutes. Also when the CDN is having issues. Admittedly this can be discerned by looking at the download rate if you're thinking about it.

src/cli/download_tracker.rs
FranciscoTGouveia: Yes, this does overlap with the beginning of the `check_updates` function in `src/cli/rustup_mode.rs`.
Would it make sense to implement a new method on `Process` to encapsulate this logic?
That way, the highlighted code could be replaced with a simple function call.

PS: I’ve already squashed the latest commit (the one switching to the legacy progress bar).

src/cli/download_tracker.rs
```diff
@@ -1,291 +1,80 @@
...
+    multi_progress_bars: MultiProgress,
+    /// ProgressBar for the current download.
+    progress_bar: ProgressBar,
 }
 
 impl DownloadTracker {
     /// Creates a new DownloadTracker.
     pub(crate) fn new_with_display_progress(display_progress: bool, process: &Process) -> Self {
+        let multi_progress_bars = MultiProgress::with_draw_target(process.progress_draw_target());
+        if !display_progress {
```
rami3l: Nit: When `display_progress` is `false` we would like to prevent unnecessary calculation. How about the following instead (unformatted, don't commit directly)?


```suggestion
        let multi_progress_bars = MultiProgress::with_draw_target(if display_progress { process.progress_draw_target() } else { ProgressDrawTarget::hidden() });
```

src/cli/rustup_mode.rs
```diff
@@ -806,14 +805,7 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
...
         let multi_progress_bars =
-            MultiProgress::with_draw_target(match cfg.process.var("RUSTUP_TERM_PROGRESS_WHEN") {
-                Ok(s) if s.eq_ignore_ascii_case("always") => {
-                    ProgressDrawTarget::term_like(Box::new(t))
-                }
-                Ok(s) if s.eq_ignore_ascii_case("never") => ProgressDrawTarget::hidden(),
-                _ if is_a_tty => ProgressDrawTarget::term_like(Box::new(t)),
-                _ => ProgressDrawTarget::hidden(),
-            });
+            MultiProgress::with_draw_target(cfg.process.progress_draw_target());
```
rami3l, line 808: When extracting existing logic, the recommended approach would be doing the extraction first in a separate commit, and then reuse the newly-extracted function in your new code. Thus, the order of the 2 current commits should be the other way around. This way the per-commit diff would be smaller as well.

rami3l (approved):
Thanks again for the work, your patch works pretty well on my machine!

LGTM modulo some minor suggestions :)

djc:
> Also, I actually prefer the legacy progress bar style, something closer to pacman or apt:

Can you explain why?

src/cli/download_tracker.rs
```diff
@@ -1,291 +1,87 @@
...
-                    if self.downloaded_last_few_secs.len() == DOWNLOAD_TRACK_COUNT {
-                        self.downloaded_last_few_secs.pop_back();
-                    }
-                    self.downloaded_last_few_secs
-                        .push_front(self.downloaded_this_sec);
-                    self.downloaded_this_sec = 0;
-                }
-            }
+        if self.progress_bar.is_hidden() && self.progress_bar.elapsed() >= Duration::from_secs(1) {
+            self.multi_progress_bars.add(self.progress_bar.clone());
```
djc, line 71: Why do this?

src/utils/units.rs
```diff
@@ -29,16 +28,13 @@ impl Size {
 impl Display for Size {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Size::B(size, unitmode) => {
+            Size::B(size, _) => {
```
djc: Should `Size` variants just be changed to omit the `UnitMode`? It doesn't look like we're still using it?

src/process.rs
```diff
@@ -151,6 +152,24 @@ impl Process {
...
+                        ProgressDrawTarget::term_like(Box::new(t))
+                    }
+                    Ok(s) if s.eq_ignore_ascii_case("never") => ProgressDrawTarget::hidden(),
+                    _ if t.is_a_tty() => ProgressDrawTarget::term_like(Box::new(t)),
+                    _ => ProgressDrawTarget::hidden(),
+                }
+            }
+            #[cfg(feature = "test")]
+            Process::TestProcess(_) => ProgressDrawTarget::hidden(),
+        }
```
djc: Suggest restructuring like this:

```rust
if Self::TestProcess(_) = self {
     return ProgressDrawTarget::hidden();
}

let t = ..
```

rami3l: @djc I can imagine some difficulties if we are to explicitly state the `OsProcess` variant...  Do you mean marking the if let with a `cfg` and no longer explicitly stating the variant below? In that case its implicit use isn't immediately obvious if we were to adjust the available variants of this enum.

rami3l:
> > Also, I actually prefer the legacy progress bar style, something closer to pacman or apt:
> 
> Can you explain why?

Being a universal tool, we would expect rustup UI to work correctly even if our host doesn't come with proper support for Unicode (this doesn't mean non-ASCII chars are not displayed; but it could mean that they can be shown in a surprising way to be no longer suitable for UI use).

Cargo, for example, has to [detect this in case it needs to provide a fallback](https://github.com/rust-lang/cargo/blob/9b296973b425ffb159e12cf3cd56580fd5c85382/src/cargo/core/shell.rs#L578) (thanks, @weihanglo!). [The same is true for Python's `rich`](https://github.com/Textualize/rich/blob/2dca1b70359dac61e1bbfb6f14ebe19a5ab79c3d/rich/progress_bar.py#L170) and more. IMHO this is not providing much additional value to our simplification plan, so going with a more traditional solution means we don't have to consider such complexities.

As for the actual character set being used, `"## "` is my personal preference, but `"=> "` is another option that I have considered so the progress bar could look more like cargo. [UV seems to be using `"-- "`](https://github.com/astral-sh/uv/blob/0a51489ec4ccc863bc84bc8bddd244859defbf96/crates/uv/src/commands/reporters.rs#L259) to be closer to `rich` and friends.

src/cli/download_tracker.rs
rami3l: @djc This is to be coherent with the old rustup logic where if a progress is considered to have finished instantly, the corresponding progress bar will remain invisible all along.

Also, the `.clone()` action is cheap because [`ProgressBar` has referential semantics](https://docs.rs/indicatif/0.18.0/src/indicatif/progress_bar.rs.html#25-29).

src/process.rs
djc: > @djc I can imagine some difficulties if we are to explicitly state the `OsProcess` variant... Do you mean marking the if let with a `cfg` and no longer explicitly stating the variant below? In that case its implicit use isn't immediately obvious if we were to adjust the available variants of this enum.

I'm fine with doing an explicit `match` will fall-through for the `OsProcess(_)` arm, too. My main point is that we can reduce rightward drift for the code that builds a `ProgressDrawTarget` for the `OsProcess` variant.

rami3l: @djc I see what you mean. However I'm not quite sure if clippy will be happy with the two-case match where one of them just does `OsProcess(_) => ()`, meaning a manual override might be required... Let's see.

src/cli/download_tracker.rs
FranciscoTGouveia: Yes, this is definitely cleaner. Thank you! :)

src/cli/rustup_mode.rs
FranciscoTGouveia: Thank you for noticing this. I have already re-ordered the commits. Many thanks!

src/utils/units.rs
FranciscoTGouveia: I have also noticed this. Having deleted `Rate`, the `UnitMode` enum does not really make sense anymore.
Should I address this and delete the `UnitMode` enum and just leave `Norm`?

FranciscoTGouveia: Done! :)

src/process.rs
FranciscoTGouveia: Thank you for pointing this out, @djc!
I believe I’ve now addressed the rightward drift.

FranciscoTGouveia:
Regarding the use of [MultiProgress](https://docs.rs/indicatif/latest/indicatif/struct.MultiProgress.html), this construct was chosen with two goals in mind:

- **Avoiding redundant terminal handling:** Instead of creating a separate `ColorableTerminal` for each `ProgressBar`, we set the _draw target_ (only) once during the creation of the `MultiProgress`. Moreover, each individual _progress bar_ is initially hidden, and when added to the `MultiProgress`, its draw target is updated automatically to match.

- **Preparing for concurrent downloads:** This design sets up the `DownloadTracker` to support multiple progress bars simultaneously, which aligns with my goal of submitting a patch to download components concurrently.

src/cli/download_tracker.rs
rami3l: See also https://github.com/rust-lang/rustup/pull/4426#issuecomment-3132730728 for the reason of using `multi_progress_bars` in the first place.

PR #4428, opened 2025-07-24
ci(macos): run x64 workflows with Rosetta 2

rami3l:
In response to [#t-infra > macos-13 shutting down](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/macos-13.20shutting.20down), this PR tries to find out what will happen to our CI/CD when the only x64 macOS environment is no longer available on GitHub Actions, as well as what we can do about this situation.

cc https://github.com/rust-lang/rfcs/pull/3841


djc (approved):
Nice work!

PR #4429, opened 2025-07-26
feat: improve error message for `rustup which`

### LINKED ISSUE #3321: Give a more helpful error message for `rustup which rustfmt` if it's not installed

These errors are inconsistent:
```
; rustup which rustfmt --toolchain beta
error: not a file: '/home/jyn/.local/lib/rustup/toolchains/beta-x86_64-unknown-linux-gnu/bin/rustfmt'
; rustfmt +beta
error: 'rustfmt' is not installed for the toolchain 'beta-x86_64-unknown-linux-gnu'
To install, run `rustup component add rustfmt --toolchain beta-x86_64-unknown-linux-gnu`
```

### Solution you'd like

The second error here ('rustfmt not installed') seems more helpful; it would be nice to show that consistently.

### Notes

```
; rustup --version
```

Note that this error message also happens for commands which aren't proxied:
```
; rustup which ls --toolchain beta
error: not a file: '/home/jyn/.local/lib/rustup/toolchains/beta-x86_64-unknown-linux-gnu/bin/ls'
```
I'm not sure the right thing to do here; maybe the current error is good enough?

Bogay:
Close #3321 

This PR improves the error message for `rustup which` by providing a suggested command to install the missing component, provided that we know which one it is.

## Limitations

Some edge cases are not coverd, e.g. if a binary is expected to be provided by a component but not found on disk. (which is handled in `DistributableToolchain::recursion_error`)

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1045,7 +1045,7 @@ async fn which(
     };
 
     Err(anyhow!(
-        "'{binary}' is not installed for the toolchain '{toolchain_name}'.\nTo install, run `rustup component add {selector}{component_name}`"
+        "'{binary}' is not installed for the toolchain '{toolchain_name}'.\nhelp: run `rustup component add {selector}{component_name}` to install it"
     ))
 }
 
@@ -1714,11 +1714,8 @@ async fn doc(
             .as_slice()
     {
         info!(
-            "`rust-docs` not installed in toolchain `{}`",
-            distributable.desc()
-        );
-        info!(
-            "To install, try `rustup component add --toolchain {} rust-docs`",
+            "`rust-docs` not installed in toolchain `{}`\nhelp: run `rustup component add --toolchain {} rust-docs` to install it",
+            distributable.desc(),
             distributable.desc()
         );
         return Err(anyhow!(
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -19,7 +19,6 @@ use itertools::Itertools;
 use tracing::{info, trace, warn};
 use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
 
-use crate::dist::AutoInstallMode;
 use crate::{
     cli::{
         common::{self, PackageUpdate, update_console_filter},
@@ -28,10 +27,10 @@ use crate::{
         self_update::{self, SelfUpdateMode, check_rustup_update},
         topical_doc,
     },
-    command,
+    command, component_for_bin,
     config::{ActiveReason, Cfg},
     dist::{
-        PartialToolchainDesc, Profile, TargetTriple,
+        AutoInstallMode, PartialToolchainDesc, Profile, TargetTriple,
         manifest::{Component, ComponentStatus},
     },
     errors::RustupError,
@@ -1026,12 +1025,28 @@ async fn which(
     binary: &str,
     toolchain: Option<ResolvableToolchainName>,
 ) -> Result<utils::ExitCode> {
-    let binary_path = cfg.resolve_toolchain(toolchain).await?.binary_file(binary);
+    let toolchain = cfg.resolve_toolchain(toolchain).await?;
+    let binary_path = toolchain.binary_file(binary);
+    if utils::is_file(&binary_path) {
+        writeln!(cfg.process.stdout().lock(), "{}", binary_path.display())?;
+        return Ok(utils::ExitCode(0));
+    }
 
-    utils::assert_is_file(&binary_path)?;
+    let toolchain_name = toolchain.name();
+    let Some(component_name) = component_for_bin(binary) else {
+        return Err(anyhow!(
+            "unknown binary '{binary}' in toolchain '{toolchain_name}'",
+        ));
+    };
 
-    writeln!(cfg.process.stdout().lock(), "{}", binary_path.display())?;
-    Ok(utils::ExitCode(0))
+    let selector = match cfg.active_toolchain() {
+        Ok(Some((t, _))) if &t == toolchain_name => String::new(),
+        _ => format!("--toolchain {} ", toolchain.name()),
+    };
+
+    Err(anyhow!(
+        "'{binary}' is not installed for the toolchain '{toolchain_name}'.\nTo install, run `rustup component add {selector}{component_name}`"
+    ))
 }
 
 #[tracing::instrument(level = "trace", skip_all)]
--- a/src/toolchain/distributable.rs
+++ b/src/toolchain/distributable.rs
@@ -453,7 +453,7 @@ impl<'a> DistributableToolchain<'a> {
                     _ => format!("--toolchain {} ", self.toolchain.name()),
                 };
                 Err(anyhow!(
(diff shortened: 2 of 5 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1034,12 +1034,30 @@ async fn which(
...
+    if utils::is_file(&binary_path) {
+        writeln!(cfg.process.stdout().lock(), "{}", binary_path.display())?;
+        return Ok(utils::ExitCode(0));
+    }
 
-    utils::assert_is_file(&binary_path)?;
+    let toolchain_name = toolchain.name();
+    let Some(component_name) = component_for_bin(binary) else {
+        return Err(anyhow!(
+            "Unknown binary '{binary}' in toolchain '{toolchain_name}'.",
```
djc: Nit: should drop the capital at the start and period at the end here for consistency.

src/cli/rustup_mode.rs
```diff
@@ -1034,12 +1034,30 @@ async fn which(
...
+    let active = matches!(cfg.active_toolchain(), Ok(Some((t, _))) if &t == toolchain_name);
+    let selector = if active {
+        String::new()
+    } else {
+        format!("--toolchain {} ", toolchain.name())
+    };
+
+    return Err(anyhow!(
+        "'{binary}' is not installed for the toolchain '{toolchain_name}'.\nTo install, run `rustup component add {selector}{component_name}`"
+    ));
```
djc: Nit: idiomatically, no need to use `return` here, can just make `Err(..)` the tail expression.

src/cli/rustup_mode.rs
```diff
@@ -1034,12 +1034,30 @@ async fn which(
...
+    };
 
-    writeln!(cfg.process.stdout().lock(), "{}", binary_path.display())?;
-    Ok(utils::ExitCode(0))
+    let active = matches!(cfg.active_toolchain(), Ok(Some((t, _))) if &t == toolchain_name);
+    let selector = if active {
+        String::new()
+    } else {
+        format!("--toolchain {} ", toolchain.name())
+    };
```
djc: This seems little roundabout, what about something more like:

```rust
let selector = match cfg.active_toolchain() {
     Ok(Some((t, _))) if t == toolchain_name => String::new(),
     _ => format!("--toolchain {} ", toolchain.name()),
};
```

src/cli/rustup_mode.rs
```diff
@@ -47,6 +46,7 @@ use crate::{
     },
     utils::{self, ExitCode},
 };
+use crate::{component_for_bin, dist::AutoInstallMode};
```
djc: If you're going to move this around, would be nicer to move both of these into the big `crate` block just above. 👍

djc (commented):
Seems like a good start!

Obviously this should pass CI -- would also be great if you could make sure there are tests to cover the new code paths.

djc (approved):
LGTM! Would be good to squash your commits into a single one, and mark the PR as ready for review.

rami3l (commented):
Thanks @djc for doing the review! I have one question though:

What do you think about the edge case mentioned [here](https://github.com/rust-lang/rustup/issues/3321#issuecomment-3122259945)?

djc:
> What do you think about the edge case mentioned here?

The situation with this PR merged seems strictly better to me than the current state on master. Maybe there are still further improvements to be made but it's not obvious to me they should necessarily be a part of this PR.

rami3l:
> > What do you think about the edge case mentioned here?
> 
> The situation with this PR merged seems strictly better to me than the current state on master. Maybe there are still further improvements to be made but it's not obvious to me they should necessarily be a part of this PR.

@djc I agree. That said, it might still be worth it if @Bogay can put these considerations at the top of this thread (in a separate section, say `Limitations`) instead of a separate commit comment :)

Bogay:
I have put these considerations to the PR description and squash my commits into one. I think it's ready for review now. Regarding the edge case I mentioned, I am not sure whether I should try to solve it? or wait for someone to encounter it and open an issue.

src/cli/rustup_mode.rs
```diff
@@ -1034,12 +1033,28 @@ async fn which(
...
-    utils::assert_is_file(&binary_path)?;
+    let selector = match cfg.active_toolchain() {
+        Ok(Some((t, _))) if &t == toolchain_name => String::new(),
+        _ => format!("--toolchain {} ", toolchain.name()),
+    };
 
-    writeln!(cfg.process.stdout().lock(), "{}", binary_path.display())?;
-    Ok(utils::ExitCode(0))
+    Err(anyhow!(
+        "'{binary}' is not installed for the toolchain '{toolchain_name}'.\nTo install, run `rustup component add {selector}{component_name}`"
```
rami3l: Pre-existing issue, but [``\nhelp: run `...` to install it``](https://github.com/rust-lang/rustup/issues/3321#issuecomment-3128153195) seems like a more coherent option in terms of overall styling of these messages 🤔 

It would be nice if, in a separate commit, the places still using `\nTo install, ...` are also adjusted to this style.

src/cli/rustup_mode.rs
```diff
@@ -1711,7 +1726,7 @@ async fn doc(
             distributable.desc()
         );
         info!(
-            "To install, try `rustup component add --toolchain {} rust-docs`",
+            "run `rustup component add --toolchain {} rust-docs` to install it",
```
Bogay: Because `info!` add `info: ` prefix to output message and I think `info: help: ` looks a little weird. I don't add the `help: ` prefix here. Not sure whether it's better to merge these two lines info the error message below?

rami3l (approved):
LGTM, and thanks for the patch!

@djc Can you confirm that your concerns have been properly addressed as well?

rami3l:
> I have put these considerations to the PR description and squash my commits into one. I think it's ready for review now. Regarding the edge case I mentioned, I am not sure whether I should try to solve it? or wait for someone to encounter it and open an issue.

@Bogay If you think a further PR is possible, I would greatly appreciate it :)

Bogay (commented):
I've updated the relevant error message to suggest how users can install the missing component. For the long term, I'm thinking we should find a more consistent way to manage error messages across the project. Do you have any thoughts on how we could approach this, perhaps with custom error types or utility functions?

src/cli/rustup_mode.rs
rami3l: Nice catch! Yeah, I think in that case it'd be better to just do a single `info!()` call instead of two...

Bogay: Updated. We have only one `info!()` call now 👍

rami3l:
@Bogay It looks like both I and my teammate are having limited availability issues recently, but I haven't forgotten about this PR. So sorry for the long wait! 🙇

rami3l:
> I've updated the relevant error message to suggest how users can install the missing component. For the long term, I'm thinking we should find a more consistent way to manage error messages across the project. Do you have any thoughts on how we could approach this, perhaps with custom error types or utility functions?

@Bogay The difficulty here is that we have many messages that are used only for a few times across the project that are added over the course of 10+ years. We do have an enumeration of the most common error messages in the "notification" system, but in  fact ad-hoc error messages are just more practical for most places. 🙏

Issue #3018, opened 2022-06-27
Download components in parallel while installing a toolchain


ghost:
### Problem you are trying to solve

Not exactly a problem, rather a feature. It could make downloading a toolchain much faster, really helpful while bisecting Rustc.

### Solution you'd like

I would want Rustup to download components in parallel while downloading a toolchain.

### Notes

I'd love to contribute!

kinnison:
I'd love to see a solution for this implemented, we might need to switch to a better progress bar implementation to support multiple parallel downloads though.

fisherdarling:
Hi! I'd like to work on this :) It's been about ~5 years~ 3 years since I last looked at the rustup repo, but I think I can implement this.

I'll definitely be asking many questions :D

fisherdarling:
Ok, so this is rather complex. The primary problem is the `&dyn Fn(Notification<'_>)` callback threaded through the codebase and the fact that `Notification` has a lifetime.

I'm going to see if I can find a good place to split the dist::Notifications into something that's more Send/Sync friendly.

fisherdarling:
@gimbles Could you give me a command that you run that you hope parallel downloads would speed up? I have a fork that downloads in parallel and I'd like to get some example commands to benchmark.

ghost:
> Could you give me a command that you run that you hope parallel downloads would speed up? I have a fork that downloads in parallel and I'd like to get some example commands to benchmark.

@fisherdarling Downloading components in parallel would really speed up bisecting Rustc, take a peek at cargo-bisect-rustc and their examples. `cargo-msrv` will also benefit from this. Really, just install a bunch of toolchains. =)

fisherdarling:
Sounds good! I'll check that out :) I'm actually having problems showing that it's actually changing anything since I'm already saturating my download bandwidth, adding parallelism to the mix doesn't change anything for me.

However, parallelizing installs would be a huge speed up.

rbtcollins:
We're already highly parallel in install on everything except decompression

rami3l:
Closed by #4436.

Issue #731, opened 2016-09-23
Should download and install concurrently


nagisa:
Currently rustup is scarily sequential. It could easily download more stuff while installing the previously downloaded stuff.

brson:
I don't know if it would be profitable to do multiple network requests in parallel, or multiple file ops in parallel, but it would definitely be profitable to be doing networking while doing file I/O.

Today rustup does installation in two phases: first it acquires all the resources off the network; then it installs them. It does this to eliminate the uncertainty of the network failing during installation.

The actual file I/O though in rustup _does_ have a transactional system that is _supposed_ to be able to roll back, and it definitely does in the test suite, and I've seen it do rollbacks live. I am not super-confident that it is bulletproof, though we could try interleaving downloading and installation and see how it goes.

Adding parallelism here would make status messages nondeterministic and more confusing.

There are definitely opportunities for improvement here, though I'm not sure I'm ready to pull the trigger yet without thinking about the constraints more. If somebody wanted to give it a shot I'd be happy to review.

nagisa:
> I don't know if it would be profitable to do multiple network requests in parallel

~~I think the reason I proposed this in the first place is that rustc’s CDN is terribly bad with not yet cached stuff.~~ Rereading original report, I guess it wasn’t the reason, but it is still worth doing it IMO.

For reference, on my 100Mbps connection, downloading the tarball for `x86_64-unknown-linux-gnu` or `x86_64-pc-windows-msvc` is pretty fast (~8-11MB/s), but downloading artefacts for something like `powerpc64-unknown-linux-gnu` will run at some abysmal sub-1MB/s (yes, it is THAT terrible). Having 3 parallel network streams (rustc/libstd/cargo) downloading at 3MB/s combined would cut the installation time by 3. Using HTTP range requests could improve the download times even further.

As for output status/messages, maybe something like [indicatif](https://github.com/mitsuhiko/indicatif) could work?

rbtcollins:
One thing I think thats worth noting is that the failure modes leading to a partial install go up with concurrent download-and-install : the transactional system is only an approximation of one - it has no write ahead journal, nor any ability to recover after interruptions.

My recommendation would be to eliminate the transactional system in favour of a interrupt-safe eventually correct system: use the manifests and installed metadata to (coarsely, not per-individual-file!) cleanup after interrupted executions; this would permit streaming installation where the archive doesn't have to get written to disk at all.

kinnison:
[...]

PR #4436, opened 2025-08-03
Concurrently download components of a toolchain

FranciscoTGouveia:
As of now, each component of a toolchain is downloaded sequentially, leaving room for performance improvements.
This is particularly notorious when considering interleaving downloads with installation, as discussed in #731.

Following the work made in #4426, as the `DownloadTracker` is now `indicatif`-based, doing the progress reporting for concurrent downloads is much easier.
As such, this PR comes to enable downloading different components concurrently (possibly closing #3018).

Performance-wise, benchmarks using `hyperfine` showed a small (**1.1x**) speedup -- which doesn't seem noticeable but lays the foundations for a future PR to interleave the download with the installation (of components).
The benchmarks were ran 50 times (with 5 warmup runs) over a 50 Mbps connection, each downloading an entire toolchain.

Many thanks to @rami3l and his [PR](https://github.com/rust-lang/rustup/pull/4435) for refactoring part of the test suite for the purpose of this PR.

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -1,5 +1,5 @@
 use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
-use std::time::Duration;
+use std::collections::HashMap;
 
 use crate::dist::Notification as In;
 use crate::notifications::Notification;
@@ -13,8 +13,8 @@ use crate::utils::Notification as Un;
 pub(crate) struct DownloadTracker {
     /// MultiProgress bar for the downloads.
     multi_progress_bars: MultiProgress,
-    /// ProgressBar for the current download.
-    progress_bar: ProgressBar,
+    /// Mapping of URLs being downloaded to their corresponding progress bars.
+    file_progress_bars: HashMap<String, ProgressBar>,
 }
 
 impl DownloadTracker {
@@ -28,22 +28,35 @@ impl DownloadTracker {
 
         Self {
             multi_progress_bars,
-            progress_bar: ProgressBar::hidden(),
+            file_progress_bars: HashMap::new(),
         }
     }
 
     pub(crate) fn handle_notification(&mut self, n: &Notification<'_>) -> bool {
         match *n {
-            Notification::Install(In::Utils(Un::DownloadContentLengthReceived(content_len))) => {
-                self.content_length_received(content_len);
+            Notification::Install(In::Utils(Un::DownloadContentLengthReceived(
+                content_len,
+                url,
+            ))) => {
+                if let Some(url) = url {
+                    self.content_length_received(content_len, url);
+                }
                 true
             }
-            Notification::Install(In::Utils(Un::DownloadDataReceived(data))) => {
-                self.data_received(data.len());
+            Notification::Install(In::Utils(Un::DownloadDataReceived(data, url))) => {
+                if let Some(url) = url {
+                    self.data_received(data.len(), url);
+                }
                 true
             }
-            Notification::Install(In::Utils(Un::DownloadFinished)) => {
-                self.download_finished();
+            Notification::Install(In::Utils(Un::DownloadFinished(url))) => {
+                if let Some(url) = url {
+                    self.download_finished(url);
+                }
+                true
+            }
+            Notification::Install(In::DownloadingComponent(component, _, _, url)) => {
+                self.create_progress_bar(component.to_owned(), url.to_owned());
                 true
             }
             Notification::Install(In::Utils(Un::DownloadPushUnit(_))) => true,
@@ -53,30 +66,44 @@ impl DownloadTracker {
         }
     }
 
-    /// Sets the length for a new ProgressBar and gives it a style.
-    pub(crate) fn content_length_received(&mut self, content_len: u64) {
-        self.progress_bar.set_length(content_len);
-        self.progress_bar.set_style(
+    /// Creates a new ProgressBar for the given component.
+    pub(crate) fn create_progress_bar(&mut self, component: String, url: String) {
+        let pb = ProgressBar::hidden();
+        pb.set_style(
             ProgressStyle::with_template(
-                "[{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
+                "{msg:>12.bold}  [{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
             )
             .unwrap()
             .progress_chars("## "),
         );
+        pb.set_message(component);
+        self.multi_progress_bars.add(pb.clone());
+        self.file_progress_bars.insert(url, pb);
+    }
+
+    /// Sets the length for a new ProgressBar and gives it a style.
+    pub(crate) fn content_length_received(&mut self, content_len: u64, url: &str) {
+        if let Some(pb) = self.file_progress_bars.get(url) {
+            pb.set_length(content_len);
+        }
     }
 
     /// Notifies self that data of size `len` has been received.
-    pub(crate) fn data_received(&mut self, len: usize) {
-        if self.progress_bar.is_hidden() && self.progress_bar.elapsed() >= Duration::from_secs(1) {
-            self.multi_progress_bars.add(self.progress_bar.clone());
+    pub(crate) fn data_received(&mut self, len: usize, url: &str) {
+        if let Some(pb) = self.file_progress_bars.get(url) {
+            pb.inc(len as u64);
         }
-        self.progress_bar.inc(len as u64);
     }
 
     /// Notifies self that the download has finished.
-    pub(crate) fn download_finished(&mut self) {
-        self.progress_bar.finish_and_clear();
-        self.multi_progress_bars.remove(&self.progress_bar);
-        self.progress_bar = ProgressBar::hidden();
+    pub(crate) fn download_finished(&mut self, url: &str) {
+        let Some(pb) = self.file_progress_bars.get(url) else {
+            return;
+        };
+        pb.set_style(
+            ProgressStyle::with_template("{msg:>12.bold}  downloaded {total_bytes} in {elapsed}")
+                .unwrap(),
+        );
+        pb.finish();
     }
 }
--- a/src/diskio/threaded.rs
+++ b/src/diskio/threaded.rs
@@ -263,10 +263,11 @@ impl Executor for Threaded<'_> {
         // pretend to have bytes to deliver.
         let mut prev_files = self.n_files.load(Ordering::Relaxed);
         if let Some(handler) = self.notify_handler {
-            handler(Notification::DownloadFinished);
+            handler(Notification::DownloadFinished(None));
             handler(Notification::DownloadPushUnit(Unit::IO));
             handler(Notification::DownloadContentLengthReceived(
                 prev_files as u64,
+                None,
             ));
         }
         if prev_files > 50 {
@@ -284,12 +285,15 @@ impl Executor for Threaded<'_> {
             current_files = self.n_files.load(Ordering::Relaxed);
             let step_count = prev_files - current_files;
             if let Some(handler) = self.notify_handler {
-                handler(Notification::DownloadDataReceived(&buf[0..step_count]));
+                handler(Notification::DownloadDataReceived(
+                    &buf[0..step_count],
+                    None,
+                ));
             }
         }
         self.pool.join();
         if let Some(handler) = self.notify_handler {
-            handler(Notification::DownloadFinished);
+            handler(Notification::DownloadFinished(None));
             handler(Notification::DownloadPopUnit);
         }
         // close the feedback channel so that blocking reads on it can
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -6,8 +6,10 @@ mod tests;
 
 use std::path::Path;
 
-use anyhow::{Context, Result, anyhow, bail};
+use anyhow::{Context, Error, Result, anyhow, bail};
+use futures_util::stream::StreamExt;
 use tokio_retry::{RetryIf, strategy::FixedInterval};
+use tracing::info;
 
 use crate::dist::component::{
     Components, Package, TarGzPackage, TarXzPackage, TarZStdPackage, Transaction,
@@ -153,6 +155,7 @@ impl Manifestation {
         let mut things_to_install: Vec<(Component, CompressionKind, File)> = Vec::new();
         let mut things_downloaded: Vec<String> = Vec::new();
         let components = update.components_urls_and_hashes(new_manifest)?;
+        let components_len = components.len();
 
         const DEFAULT_MAX_RETRIES: usize = 3;
         let max_retries: usize = download_cfg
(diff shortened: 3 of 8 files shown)


FranciscoTGouveia:
For reference, the current download of components (sequential) looks like this:

before-ezgif com-speed

... and the changes introduced by this PR go like this:

after-ezgif com-speed

Please note that both animations were sped up and they are not meaningful for comparing performances.

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,61 @@ impl DownloadTracker {
...
     /// Notifies self that the download has finished.
-    pub(crate) fn download_finished(&mut self) {
-        self.progress_bar.finish_and_clear();
-        self.multi_progress_bars.remove(&self.progress_bar);
-        self.progress_bar = ProgressBar::hidden();
+    pub(crate) fn download_finished(&mut self, file: &str) {
+        if let Some(pb) = self.find_progress_bar(file) {
+            pb.set_style(
+                ProgressStyle::with_template(
+                    "{msg:>12.bold}  downloaded {total_bytes} in {elapsed}.",
```
rami3l, line 112: Nit: `.` should be omitted to be coherent with the general notification style.

rami3l: Nit: `.` should be omitted to be coherent with the general notification style.

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,61 @@ impl DownloadTracker {
         }
     }
 
-    /// Sets the length for a new ProgressBar and gives it a style.
-    pub(crate) fn content_length_received(&mut self, content_len: u64) {
-        self.progress_bar.set_length(content_len);
-        self.progress_bar.set_style(
+    /// Helper function to find the progress bar for a given file.
+    fn find_progress_bar(&mut self, file: &str) -> Option<&mut ProgressBar> {
```
rami3l: Nit: `progress_bar()` would be a better name according to the [API style guide](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,61 @@ impl DownloadTracker {
...
+        // During the installation this function can be called with an empty file/URL.
+        if file.is_empty() {
+            return None;
+        }
+        let component = self
+            .file_progress_bars
+            .keys()
+            .find(|comp| file.contains(*comp))
+            .cloned()?;
```
rami3l: It should not be necessary to clone this component given that we will immediately refer to it.

src/dist/manifestation.rs
```diff
@@ -162,41 +165,69 @@ impl Manifestation {
...
+                }
+            });
+        if components_len > 0 {
+            let results = component_stream
+                .buffered(components_len)
+                .collect::<Vec<_>>()
+                .await;
+            for result in results {
+                match result {
+                    Ok((component, format, downloaded_file, hash)) => {
```
rami3l: Nit: Isn't this just `let (...) = result?;`?

src/dist/manifestation.rs
```diff
@@ -162,41 +165,69 @@ impl Manifestation {
...
+                                }
+                                _ => false,
+                            }
+                        },
+                    )
+                    .await
+                    .with_context(|| {
+                        RustupError::ComponentDownloadFailed(component.name(new_manifest))
+                    })?;
+                    Ok::<(Component, CompressionKind, File, String), Error>((
```
rami3l: Can this annotation be simpler, say `Ok::<_, Error>`?

src/utils/notifications.rs
```diff
@@ -14,11 +14,11 @@ pub enum Notification<'a> {
     RemovingDirectory(&'a str, &'a Path),
     DownloadingFile(&'a Url, &'a Path),
     /// Received the Content-Length of the to-be downloaded data.
-    DownloadContentLengthReceived(u64),
+    DownloadContentLengthReceived(u64, &'a str),
```
rami3l: Would you mind explaining the meaning of `""` here and why they are required? It might be an interesting addition to the doc comments as well.

rami3l (commented):
A note to myself: we also have to consider the restriction of concurrent downloads similar to what we have done with threadpooled diskio.

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,61 @@ impl DownloadTracker {
...
+        let pb = ProgressBar::hidden();
+        pb.set_style(
             ProgressStyle::with_template(
-                "[{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
+                "{msg:>12.bold}  [{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
             )
             .unwrap()
             .progress_chars("## "),
         );
+        pb.set_message(component.to_string());
```
rami3l: Nit: `.to_string()` doesn't seem necessary here...

src/dist/manifestation.rs
FranciscoTGouveia: I believe we can simply drop the annotation.

FranciscoTGouveia: As I was addressing the previous nit, I have came to the conclusion that this **is** necessary. Sorry for the confusion.

src/utils/notifications.rs
FranciscoTGouveia: Since we are downloading the components concurrently, we can’t know which download the *Content-Length* refers to. Therefore, in addition to receiving the length itself, we also need the component’s *URL* so we can match it to the corresponding progress bar.

Note: I've already added a doc comment. Thank you!

src/cli/download_tracker.rs
FranciscoTGouveia: I may be misreading the situation here, but I think we actually need the `.to_string()` here.

This is due to the fact that the [`set_message()`](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.set_message) expects `Into<Cow<'static, str>>`. and we would try to pass a `&str` with a non-`'static` lifetime.

src/utils/notifications.rs
FranciscoTGouveia: Before resolving this conversation, I will investigate the correct use of this notification and the addition of this extra parameter to ensure that, even though the installations are not concurrent, it did not have any negative effect on them (which, unfortunately, cannot be guaranteed by the test suite).

FranciscoTGouveia:
After my investigation stemming from this [comment](https://github.com/rust-lang/rustup/pull/4436#discussion_r2266964528), I have concluded that the changes presented on this PR are **not** consistent with the previous behavior of displaying a progress bar during the installation of the component.

Would this be necessary, especially considering that the next step will be to interleave the downloads with the installation?

rami3l:
> I have concluded that the changes presented on this PR are not consistent with the previous behavior of displaying a progress bar during the installation of the component.
> 
> Would this be necessary, especially considering that the next step will be to interleave the downloads with the installation?

@FranciscoTGouveia No, I don't think it is necessary for that part of the behavior to stay unchanged, as you have already mentioned. Of course, if it is easy to add it back, then it might still be interesting to do so.

src/dist/manifestation.rs
```diff
@@ -162,41 +165,60 @@ impl Manifestation {
...
-            )
-            .await
-            .with_context(|| RustupError::ComponentDownloadFailed(component.name(new_manifest)))?;
-
-            things_downloaded.push(hash);
+        }
 
-            things_to_install.push((component, format, downloaded_file));
+        let component_stream =
+            tokio_stream::iter(components.into_iter()).map(|(component, format, url, hash)| {
```
rami3l, line 178: We also have to consider the restriction of concurrent downloads similar to what we have done with threadpooled diskio.

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,60 @@ impl DownloadTracker {
...
-    /// Sets the length for a new ProgressBar and gives it a style.
-    pub(crate) fn content_length_received(&mut self, content_len: u64) {
-        self.progress_bar.set_length(content_len);
-        self.progress_bar.set_style(
+    /// Helper function to find the progress bar for a given file.
+    fn progress_bar(&mut self, file: &str) -> Option<&ProgressBar> {
+        // During the installation this function can be called with an empty file/URL.
+        if file.is_empty() {
+            return None;
+        }
```
djc: Why does that happen? Seems pretty iffy.

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,60 @@ impl DownloadTracker {
...
+        let component = self
+            .file_progress_bars
+            .keys()
+            .find(|comp| file.contains(*comp))?;
+
+        self.file_progress_bars.get(component)
```
djc: Why this elaborate setup? Are `file` values not the same as the `component`s?

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,60 @@ impl DownloadTracker {
...
-        self.progress_bar.inc(len as u64);
     }
 
     /// Notifies self that the download has finished.
-    pub(crate) fn download_finished(&mut self) {
-        self.progress_bar.finish_and_clear();
-        self.multi_progress_bars.remove(&self.progress_bar);
-        self.progress_bar = ProgressBar::hidden();
+    pub(crate) fn download_finished(&mut self, file: &str) {
+        if let Some(pb) = self.progress_bar(file) {
```
djc: Nit: `let Some(pb) = .. else { return };`.

src/cli/download_tracker.rs
```diff
@@ -53,30 +60,60 @@ impl DownloadTracker {
...
+            pb.set_style(
+                ProgressStyle::with_template(
+                    "{msg:>12.bold}  downloaded {total_bytes} in {elapsed}",
+                )
+                .unwrap(),
+            );
+            let msg = pb.message();
+            pb.finish_with_message(msg);
```
djc: This seems weird? Could instead probably use one of the other `ProgressFinish` modes.

src/cli/download_tracker.rs
djc: IMO `to_owned()` more clearly expresses what you're trying to do (and might be slightly more efficient). In this case since `create_progress_bar()` needs two copies, suggest taking the argument as `String` and cloning it for use in `set_message()`.

src/diskio/threaded.rs
```diff
@@ -263,10 +263,11 @@ impl Executor for Threaded<'_> {
         // pretend to have bytes to deliver.
         let mut prev_files = self.n_files.load(Ordering::Relaxed);
         if let Some(handler) = self.notify_handler {
-            handler(Notification::DownloadFinished);
+            handler(Notification::DownloadFinished(""));
```
djc: This seems strange. Why pass in an empty string?

src/diskio/threaded.rs
```diff
@@ -263,10 +263,11 @@ impl Executor for Threaded<'_> {
...
             handler(Notification::DownloadPushUnit(Unit::IO));
             handler(Notification::DownloadContentLengthReceived(
                 prev_files as u64,
+                "",
```
djc: Same here. What does the empty string mean? Should the type be `Option<_>` instead?

src/dist/manifestation.rs
```diff
@@ -162,41 +165,60 @@ impl Manifestation {
...
+                        },
+                    )
+                    .await
+                    .with_context(|| {
+                        RustupError::ComponentDownloadFailed(component.name(new_manifest))
+                    })?;
+                    Ok::<_, Error>((component, format, downloaded_file, hash))
+                }
+            });
+        if components_len > 0 {
```
djc, line 191: If this is the only use of `components_len`, suggest storing a `bool` instead.

src/dist/manifestation.rs
djc: This closure is complicated enough that I think it should be extracted into a separate function or `struct`.

FranciscoTGouveia: > We also have to consider the restriction of concurrent downloads similar to what we have done with threadpooled diskio.

I will prepare a PR in the near future to address this (possibly before this one gets merged).
Would introducing an environment variable (e.g., `RUSTUP_CONCURRENT_DOWNLOADS`), similar to `RUSTUP_IO_THREADS`, be a good approach?

**NB:** This could also be extended to the `rustup check` command to maintain consistency.

---

> This closure is complicated enough that I think it should be extracted into a separate function or `struct`.

I am somewhat reluctant to do this, as my upcoming work (installing a component as soon as its download completes) will also require changes to this function.
However, for the sake of keeping the changes coherent in this PR, I will ensure it is extracted. Thank you!

src/cli/download_tracker.rs
FranciscoTGouveia: Thank you for bringing this up. I am now realizing that I should have explained these somewhat "questionable" changes in the PR description.

This check exists because, as you noticed, I am passing an empty string in the installation notifications. With this early check in place, I can avoid traversing the *HashMap* (which happens frequently) when I already know it will fail.

That said, I agree that passing an empty string as a “_filler_” argument is not ideal; as you suggested, using an `Option<_>` would possible be a better choice.

I will revisit this comment after addressing all the other ones.

src/cli/download_tracker.rs
FranciscoTGouveia: No, a `component` refers to the name of the *component* itself (e.g. `rust-std` or `rustfmt`) whereas the `file` is the actual download *URL* (e.g. `https://static.rust-lang.org/dist/2025-08-13/rust-std-nightly-x86_64-unknown-linux-gnu.tar.xz`).

This distinction exists because, when starting the downloads -- which must be initiated _in-order_ to output the progress bars correctly -- the `DownloadingComponent` notification being used, only carries the component's name and not the _URL_.

However, it is important to note that the _URL_ is available at the time that the `DownloadingComponent` notification is triggered. If we decided to add an extra parameter to this notification, the keys of the _HashMap_ could be the _URL_'s themselves, avoiding the need for this elaborate setup.

What are your thoughts on this? Should I keep it this way, or change the `DownloadingComponent` notification to also carry the _URL_ and dropping this setup of travesing the _HashMap_? (cc @rami3l)

djc: It seems better to me to use a precise key. If I understand correctly, if the `file` (which should really be named `url` if it represents a URL) contains the `component` in a different part than the final path segment, the matching here would get things pretty confused.

FranciscoTGouveia: Yes, I will rename it from `file` to `url`, and store a precise key instead so that I can drop this setup. Thanks!

src/cli/download_tracker.rs
FranciscoTGouveia: I didn't know about this, many thanks :)

src/cli/download_tracker.rs
FranciscoTGouveia: Yes, thank you! I had initially thought that changing the style of the *ProgressBar* would somehow clear the stored `message` (which, in hindsight, doesn’t make sense, as I was only retrieving it after changing styles).
I have now replaced it with `pb.finish()`.

src/cli/download_tracker.rs
FranciscoTGouveia: This isn't necessary as we are currently storing the precise keys as discussed [here](https://github.com/rust-lang/rustup/pull/4436#discussion_r2274495555).
This way, we can drop this helper function and simply call `.get()`.

src/cli/download_tracker.rs
FranciscoTGouveia: As discussed [here](https://github.com/rust-lang/rustup/pull/4436#discussion_r2274495555), I do not need to use two copies as the `component` is no longer used as a key in the *HashMap*.
However, I may avoid the `.to_string()` by passing the arguments (now two, `component` and `url`) as `String`'s instead.

src/diskio/threaded.rs
FranciscoTGouveia: Thank you once again for bringing up a somewhat “questionable” addition.

The rationale behind this change is that I needed to add a new parameter to the `DownloadFinished` notification (as well as to `DownloadContentLengthReceived` and `DownloadDataReceived`), and in this particular situation, I did not have any URL to pass as an argument.

As a result, I opted to pass an empty string.
However, we can certainly reconsider this and use an `Option<_>`. instead. Would it be preferable?

src/diskio/threaded.rs
FranciscoTGouveia: The reasoning is explained above. Thank you.

src/dist/manifestation.rs
FranciscoTGouveia: No, `components_len` is used as an argument to [`.buffered()`](https://docs.rs/futures-util/0.3.31/futures_util/stream/trait.StreamExt.html#method.buffered) to define the number of components to be downloaded at a same point in time.
Therefore, I don’t believe we can make this change.

src/diskio/threaded.rs
djc: > However, we can certainly reconsider this and use an `Option<_>`. instead. Would it be preferable?

Definitely. Using sentinel values is generally to be avoided.

(If you know that something is questionable, it would be great if you can call it out yourself in a PR review comment.)

FranciscoTGouveia:
Thank you for all the reviews comments.
Before anything I would like to apologize for not explaining my questionable decisions in the PR description from the start.
I will definitely have this in mind in my future contributions.

Additionally, I would like to highlight one point:
- The downloads are being executed using `.buffered()` rather than `.buffer_unordered()`. A similar situation arose [here](https://github.com/rust-lang/rustup/pull/4388#issuecomment-3046618948), however, in this case, the speedup is somewhat despicable (only a 1.1x speedup). Additionally, making this change would change the order that the installations are made in, which I think is unnecessary for this PR. In a future contribution aiming to interleave the downloads with installations, if the installation process is modified, we can revisit this idea.

src/dist/notifications.rs
```diff
@@ -23,7 +23,7 @@ pub enum Notification<'a> {
     ExtensionNotInstalled(&'a str),
     NonFatalError(&'a anyhow::Error),
     MissingInstalledComponent(&'a str),
-    DownloadingComponent(&'a str, &'a TargetTriple, Option<&'a TargetTriple>),
+    DownloadingComponent(&'a str, &'a TargetTriple, Option<&'a TargetTriple>, &'a str),
```
rami3l, line 27: Nit: I understand the last component is the `url` but I'm afraid we will forget about it very soon. Maybe adding another doc comment as you did in other places?

rami3l (commented):
This is looking much better to me! Many thanks @FranciscoTGouveia for the work and @djc for a second round of review :)

I've left just a final nit but otherwise I believe this PR is quite ready to be merged 🎉

rami3l (approved):

src/cli/download_tracker.rs
```diff
@@ -28,22 +28,35 @@ impl DownloadTracker {
...
-            Notification::Install(In::Utils(Un::DownloadFinished)) => {
-                self.download_finished();
+            Notification::Install(In::Utils(Un::DownloadFinished(url))) => {
+                if let Some(url) = url {
+                    self.download_finished(url);
+                }
+                true
+            }
+            Notification::Install(In::DownloadingComponent(component, _, _, url)) => {
+                self.create_progress_bar(component.to_string(), url.to_string());
```
djc: Nit: prefer to use `.to_owned()` for `component`, which is `&str` and thus doesn't need the full `Display` treatment.

src/dist/manifestation.rs
djc: This is still open -- what's the plan for extracting the closure?

djc:
(Also, CI failures...)

src/dist/manifestation.rs
FranciscoTGouveia: Done!

FranciscoTGouveia:
I believe it was a spurious failure.
All comments have now been addressed, and the *CI* is passing :)

PR #4440, opened 2025-08-10
Introduce `RUSTUP_DOWNLOAD_TIMEOUT` to override the download timeout

### LINKED ISSUE #4439: Download TimedOut on company firewall
### Verification

### Problem

Our company firewall takes a while to unpack and verify the files downloaded through rustup.

```
$ rustup update
info: latest update on 2025-08-07, rust version 1.89.0 (29483883e 2025-08-04)
info: downloading component 'rust-src'
error: failed to download file error=Reqwest(reqwest::Error { kind: Request, url: "https://static.rust-lang.org/dist/2025-08-07/rust-docs-1.89.0-x86_64-unknown-linux-gnu.tar.xz", source: TimedOut })
info: retrying download for 'https://static.rust-lang.org/dist/2025-08-07/rust-docs-1.89.0-x86_64-unknown-linux-gnu.tar.xz'
^C⏎                                                                                                                     
```

I found that `rust-docs` is especially susceptible. I timed its download, and it took 71 seconds before it was approved and the first bytes were transferred in the download.

The timeout is currently hardcoded to 30 seconds:

https://github.com/rust-lang/rustup/blob/e43263a739b2eddb8f0ac28965316fc9323f84d9/src/download/mod.rs#L589

### Steps

Run rustup on a network with an HTTPS proxy that delays downloads by verifying them. I unfortunately don't have any details on my corporations firewall.

### Possible Solution(s)

Ideally the timeout should be increased, and/or be made configurable.

### Notes

A workaround is to use the curl backend, which has a higher timeout.
```
$ RUSTUP_USE_CURL=1 rustup update
warn: RUSTUP_USE_CURL is set; the curl backend is deprecated,
            please file an issue if the default download backend does not work for your use case
warn: RUSTUP_USE_CURL is set; the curl backend is deprecated,
            please file an issue if the default download backend does not work for your use case
info: latest update on 2025-08-07, rust version 1.89.0 (29483883e 2025-08-04)
info: downloading component 'rust-src'
warn: RUSTUP_USE_CURL is set; the curl backend is deprecated,
            please file an issue if the default download backend does not work for your use case
...
```

### Rustup version

```console
1.28.2
```

### Installed toolchains

```console
```

### OS version

```console
WSL2 Debian 12 bookworm
```

FranciscoTGouveia:
Fixes #4439.

Besides increasing the default timeout from 30 seconds to 3 minutes, a new environment variable (`RUSTUP_DOWNLOAD_TIMEOUT`) is introduced to allow users to override this value.

### DIFF
--- a/doc/user-guide/src/environment-variables.md
+++ b/doc/user-guide/src/environment-variables.md
@@ -66,6 +66,9 @@
 
 - `RUSTUP_TERM_WIDTH` (default: none). Allows to override the terminal width for progress bars.
 
+- `RUSTUP_DOWNLOAD_TIMEOUT` *unstable* (default: 180). Allows to override the default
+  timeout (in seconds) for downloading components.
+
 [directive syntax]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
 [dc]: https://docs.docker.com/storage/storagedriver/overlayfs-driver/#modifying-files-or-directories
 [override]: overrides.md
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -1,7 +1,10 @@
 //! Easy file downloading
 
 use std::fs::remove_file;
+use std::num::NonZeroU64;
 use std::path::Path;
+use std::str::FromStr;
+use std::time::Duration;
 
 use anyhow::Context;
 #[cfg(any(
@@ -194,6 +197,13 @@ async fn download_file_(
         _ => Backend::Curl,
     };
 
+    let timeout = Duration::from_secs(match process.var("RUSTUP_DOWNLOAD_TIMEOUT") {
+        Ok(s) => NonZeroU64::from_str(&s).context(
+            "invalid value in RUSTUP_DOWNLOAD_TIMEOUT -- must be a natural number greater than zero",
+        )?.get(),
+        Err(_) => 180,
+    });
+
     notify_handler(match backend {
         #[cfg(feature = "curl-backend")]
         Backend::Curl => Notification::UsingCurl,
@@ -202,7 +212,7 @@ async fn download_file_(
     });
 
     let res = backend
-        .download_to_path(url, path, resume_from_partial, Some(callback))
+        .download_to_path(url, path, resume_from_partial, Some(callback), timeout)
         .await;
 
     notify_handler(Notification::DownloadFinished);
@@ -241,9 +251,10 @@ impl Backend {
         path: &Path,
         resume_from_partial: bool,
         callback: Option<DownloadCallback<'_>>,
+        timeout: Duration,
     ) -> anyhow::Result<()> {
         let Err(err) = self
-            .download_impl(url, path, resume_from_partial, callback)
+            .download_impl(url, path, resume_from_partial, callback, timeout)
             .await
         else {
             return Ok(());
@@ -265,6 +276,7 @@ impl Backend {
         path: &Path,
         resume_from_partial: bool,
         callback: Option<DownloadCallback<'_>>,
+        timeout: Duration,
     ) -> anyhow::Result<()> {
         use std::cell::RefCell;
         use std::fs::OpenOptions;
@@ -324,7 +336,7 @@ impl Backend {
         let file = RefCell::new(file);
 
         // TODO: the sync callback will stall the async runtime if IO calls block, which is OS dependent. Rearrange.
-        self.download(url, resume_from, &|event| {
+        self.download(url, resume_from, timeout, &|event| {
             if let Event::DownloadDataReceived(data) = event {
                 file.borrow_mut()
                     .write_all(data)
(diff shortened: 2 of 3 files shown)


src/download/mod.rs
```diff
@@ -322,9 +326,15 @@ impl Backend {
         };
 
         let file = RefCell::new(file);
+        let timeout = Duration::from_secs(match process.var("RUSTUP_DOWNLOAD_TIMEOUT") {
+            Ok(s) => s.parse::<NonZeroU64>().context(
+                "invalid value in RUSTUP_DOWNLOAD_TIMEOUT -- must be a natural number greater than zero",
+            ).unwrap().get(),
```
rami3l: Why are we using an `.unwrap()` here?

FranciscoTGouveia: To get the `NonZeroU64` out of the `Result`; but after reading about good practices, I realized that using `?` instead of `.unwrap()` is preferable.
Is this what you were referring to?

djc: FYI, IMO it's nicer to write `NonZeroU64::from_str(s)` compared to `s.parse::<NonZeroU64>()`.

src/download/mod.rs
```diff
@@ -622,21 +634,24 @@ mod reqwest_be {
...
+            .build()
+            .map_err(DownloadError::Reqwest)?;
+
+        let _ = CLIENT_NATIVE_TLS.set(client);
+
+        Ok(CLIENT_NATIVE_TLS.get().unwrap())
+    }
+
+    #[cfg(feature = "reqwest-native-tls")]
+    static CLIENT_NATIVE_TLS: OnceLock<Client> = OnceLock::new();
```
djc: The change from `LazyLock` to `OnceLock` seems unrelated and doesn't look like an improvement to me.

djc (commented):
Nice to see that you are now passing `Duration` by value!

doc/user-guide/src/environment-variables.md
```diff
@@ -66,6 +66,9 @@
 
 - `RUSTUP_TERM_WIDTH` (default: none). Allows to override the terminal width for progress bars.
 
+- `RUSTUP_DOWNLOAD_TIMEOUT` (default: 180). Allows to override the default
```
rami3l: Nit: You probably need to mark this variable as unstable.

src/download/mod.rs
FranciscoTGouveia: Thank you for the suggestion, this way I can avoid the *turbofish*.
However, doesn’t this require an additional import (`use std::str::FromStr`)?
Would it be acceptable to include the extra import in this case?

djc: Yes, just add the extra import at the top of the module with the other `std` ones.

src/download/mod.rs
FranciscoTGouveia: Thank you for pointing this out!
The motivation for this change was to allow me to pass an argument upon the creation of the *client*.
Since `LazyLock` does not allow passing any arguments (the `timeout`), I decided to switch to an `OnceLock`.
Is there another way to achieve this that I might be overlooking?

djc: Ah yes, I guess that makes sense -- sorry for missing that. I think this would be easier to review if you extract the move from `LazyLock` to `OnceLock` into a separate commit.

FranciscoTGouveia: I will now divide my changes in two separate commits.

src/download/mod.rs
```diff
@@ -620,12 +635,13 @@ mod reqwest_be {
     static CLIENT_RUSTLS_TLS: OnceLock<Client> = OnceLock::new();
 
     #[cfg(feature = "reqwest-native-tls")]
-    pub(super) fn native_tls_client() -> Result<&'static Client, DownloadError> {
+    pub(super) fn native_tls_client(timeout: Duration) -> Result<&'static Client, DownloadError> {
```
rami3l, line 639: It might be worth commenting that the timeout being passed in does nothing when `CLIENT_NATIVE_TLS` is already initialized (idem for `CLIENT_RUSTLS_TLS`)?

rami3l (approved):

src/download/mod.rs
FranciscoTGouveia: Yes, I agree -- the early check may execute and the `timeout` is never utilized.
However, looking at the uses of `CLIENT_NATIVE_TLS`, I do not think that there is any other place where it is initialized -- rendering the check useless.

Am I overlooking something here?

Anyways, I have already added a comment. Thank you.

rami3l: @FranciscoTGouveia You are right, but I believe when people read this code out of context, local reasoning could be tricky as they don't typically take all existing callers into account; rather they might think of all possible callers.

src/download/mod.rs
```diff
@@ -622,21 +620,23 @@ mod reqwest_be {
...
-        // the TLS library returning a null pointer as well.
-        catcher().unwrap()
-    });
+        let client = client_generic()
+            .user_agent(super::REQWEST_DEFAULT_TLS_USER_AGENT)
+            .build()
+            .map_err(DownloadError::Reqwest)?;
+
+        let _ = CLIENT_NATIVE_TLS.set(client);
+
```
djc, line 652: Add a comment here similar to what's in `rustls_client()`?

djc (approved):

src/download/mod.rs
FranciscoTGouveia: I’ve just updated my branch, but it was too late, as the PR had already been merged.
How should I proceed?

djc: Just submit a new PR with the change, please.

FranciscoTGouveia: Is it acceptable for the PR to contain only the comment?

PR #4449, opened 2025-08-17
ci(check): make installation of `taplo-cli` faster

AudaciousAxiom:
As discussed with the author of [`taiki-e/install-action`](https://github.com/taiki-e/install-action) in https://github.com/taiki-e/install-action/issues/1100, `taplo` should be used instead of `taplo-cli` as the tool name for the `install-action` Action, otherwise `install-action` falls back on using [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall), which itself seems to currently fall back on using QuickInstall. This makes the installation of `taplo-cli` in CI very slow (it currently takes more than two minutes).

This PR fixes the tool name and disables the fallback on `cargo-binstall` as that would then fail anyway, because `cargo-binstall` would try to install the `taplo` *lib* crate, instead of the `taplo-cli` bin crate.

---

Feel free to push to this PR's branch as needed.

### DIFF
--- a/ci/actions-templates/centos-fmt-clippy-template.yaml
+++ b/ci/actions-templates/centos-fmt-clippy-template.yaml
@@ -68,7 +68,8 @@ jobs: # skip-all
       - name: Install taplo
         uses: taiki-e/install-action@v2
         with:
-          tool: taplo-cli
+          tool: taplo
+          fallback: none
       - name: Run TOML formatting checks
         run: |
           taplo fmt
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1464,7 +1464,8 @@ jobs:
       - name: Install taplo
         uses: taiki-e/install-action@v2
         with:
-          tool: taplo-cli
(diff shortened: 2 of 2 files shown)


ci/actions-templates/centos-fmt-clippy-template.yaml
```diff
@@ -68,7 +68,8 @@ jobs: # skip-all
       - name: Install taplo
         uses: taiki-e/install-action@v2
         with:
-          tool: taplo-cli
+          tool: taplo
+          fallback: none
```
AudaciousAxiom, line 72: I edited this file and then ran `./ci/actions-templates/gen-workflows.sh` to update `.github/workflows/ci.yaml` accordingly; hopefully that was the expected workflow.

AudaciousAxiom:
Checking [the CI logs](https://github.com/rust-lang/rustup/actions/runs/17018289984/job/48243770914?pr=4449), `taplo-cli` is now installed in just one second as expected.

djc (approved):

PR #4450, opened 2025-08-17
Introduce `RUSTUP_CONCURRENT_DOWNLOADS` to control concurrency

FranciscoTGouveia:
This PR (based on [this](https://github.com/rust-lang/rustup/pull/4436#discussion_r2274027447) comment) introduces a new environment variable, `RUSTUP_CONCURRENT_DOWNLOADS`.

This allows users to configure how many components are downloaded concurrently -- or, in the case of `rustup check`, how many toolchains are checked concurrently.

### DIFF
--- a/doc/user-guide/src/environment-variables.md
+++ b/doc/user-guide/src/environment-variables.md
@@ -69,6 +69,9 @@
 - `RUSTUP_DOWNLOAD_TIMEOUT` *unstable* (default: 180). Allows to override the default
   timeout (in seconds) for downloading components.
 
+- `RUSTUP_CONCURRENT_DOWNLOADS` *unstable* (default: the number of components to download). Controls the number of
+  downloads made concurrently.
+
 [directive syntax]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
 [dc]: https://docs.docker.com/storage/storagedriver/overlayfs-driver/#modifying-files-or-directories
 [override]: overrides.md
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -799,7 +799,8 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
     let use_colors = matches!(t.color_choice(), ColorChoice::Auto | ColorChoice::Always);
     let mut update_available = false;
     let channels = cfg.list_channels()?;
-    let num_channels = channels.len();
+    let num_channels = cfg.process.concurrent_downloads().unwrap_or(channels.len());
+
     // Ensure that `.buffered()` is never called with 0 as this will cause a hang.
     // See: https://github.com/rust-lang/futures-rs/pull/1194#discussion_r209501774
     if num_channels > 0 {
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -154,7 +154,10 @@ impl Manifestation {
         let mut things_to_install: Vec<(Component, CompressionKind, File)> = Vec::new();
         let mut things_downloaded: Vec<String> = Vec::new();
         let components = update.components_urls_and_hashes(new_manifest)?;
-        let components_len = components.len();
+        let num_channels = download_cfg
+            .process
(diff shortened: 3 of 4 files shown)


doc/user-guide/src/environment-variables.md
```diff
@@ -69,6 +69,9 @@
 - `RUSTUP_DOWNLOAD_TIMEOUT` *unstable* (default: 180). Allows to override the default
   timeout (in seconds) for downloading components.
 
+- `RUSTUP_CONCURRENT_DOWNLOADS` *unstable* (default: the number of components to download). Controls the number of
+  downloads made concurrently.
```
FranciscoTGouveia, line 73: Is this the best wording?
In reality, this environment variable targets not only the download of components, but also checking toolchains (with `rustup check`).

src/cli/rustup_mode.rs
```diff
@@ -799,7 +800,13 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
...
     let mut update_available = false;
     let channels = cfg.list_channels()?;
-    let num_channels = channels.len();
+    let num_channels = match cfg.process.var("RUSTUP_CONCURRENT_DOWNLOADS") {
+        Ok(s) => NonZeroU64::from_str(&s).context(
+            "invalid value in RUSTUP_CONCURRENT_DOWNLOADS -- must be a natural number greater than zero"
+        )?.get() as usize,
+        Err(_) => channels.len(),
+    };
+
```
FranciscoTGouveia: This snippet is repeated twice in this PR.

Considering that more concurrency-related changes may be added to `rustup` in the future, would it make sense to add a new method to the `Process` construct that takes a `usize` argument, representing the number of channels, in case an error occurs while reading this new environment variable?

djc: I think it makes sense to add a method to `Process` for this, but I would have it yield an `Option` instead of passing in a `usize` argument.

FranciscoTGouveia: Done!

src/cli/rustup_mode.rs
```diff
@@ -800,11 +799,9 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
     let use_colors = matches!(t.color_choice(), ColorChoice::Auto | ColorChoice::Always);
     let mut update_available = false;
     let channels = cfg.list_channels()?;
-    let num_channels = match cfg.process.var("RUSTUP_CONCURRENT_DOWNLOADS") {
-        Ok(s) => NonZeroU64::from_str(&s).context(
-            "invalid value in RUSTUP_CONCURRENT_DOWNLOADS -- must be a natural number greater than zero"
-        )?.get() as usize,
-        Err(_) => channels.len(),
+    let num_channels = match cfg.process.concurrent_downloads() {
```
djc: Suggest `.unwrap_or(channels.len())` for this.

djc (approved):
Given that this ends up modifying much of the code introduced in the first commit in the second commit, suggest squashing these commits -- they're not really independent.

FranciscoTGouveia:
During a [conversation](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/Project.3A.20Make.20Rustup.20Concurrent/near/534838463) about this PR, it was noted that enabling concurrent downloads by default may introduce a slight performance regression for some users.

With that said, would it be preferable to revert the default behavior of _max-concurrency_ (defined in #4436) and leave it up to the user to opt in to concurrency, or should we keep it and address this, if necessary, in a separate issue or PR?

djc:
> During a [conversation](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/Project.3A.20Make.20Rustup.20Concurrent/near/534838463) about this PR, it was noted that enabling concurrent downloads by default may introduce a slight performance regression for some users.
> 
> With that said, would it be preferable to revert the default behavior of _max-concurrency_ (defined in #4436) and leave it up to the user to opt in to concurrency, or should we keep it and address this, if necessary, in a separate issue or PR?

Let's do that in a separate PR.

PR #4452, opened 2025-08-18
fix(deps): update rust crate tar to 0.4.44 as per Cargo.lock

yjqg6666:
In Cargo.lock, the version of crate tar dependency is 0.4.44, while in Cargo.toml it's version 0.4.26. Fix it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.10"
 sharded-slab = "0.1.1"
 strsim = "0.11"
-tar = "0.4.26"
+tar = "0.4.44"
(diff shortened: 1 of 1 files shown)


rami3l:
@yjqg6666 Thanks for contributing!

However I don't think there is anything to be fixed; the version in `Cargo.toml` gives the bottom-line requirements of a dependency, whereas the `Cargo.lock` version gives the recommended version combination. When we update dependencies, `Cargo.lock` is bumped automatically like in https://github.com/rust-lang/rustup/pull/4451, leaving `Cargo.toml` unchanged.

As such, I'm closing this PR. Sorry! 🙇

yjqg6666:
Thanks for the explanation.

PR #4453, opened 2025-08-22
chore(deps): disable default features for zstd

klensy:
Zstd not really currently used(as i understand), so no need to support legacy/dict features. Slightly reduces binary size.


djc (approved):

djc:
Thanks!

PR #4454, opened 2025-08-24
feat(config): warn user if auto-install is enabled

### LINKED ISSUE #4445: Show that a toolchain download is due to auto installation

I'm in directory [`atuin/`](https://github.com/atuinsh/atuin), trying to run the tool [bacon](https://github.com/Canop/bacon), and wanted to check if bacon has a newer version because it's `cargo install`-ed:

```
atuin  [!]
❯ cargo info bacon
info: syncing channel updates for '1.88-x86_64-unknown-linux-gnu'
info: latest update on 2025-06-26, rust version 1.88.0 (6b00bc388 2025-06-23)
info: downloading component 'rustfmt'
info: installing component 'clippy'
error: the 'cargo' binary, normally provided by the 'cargo' component, is not applicable to the '1.88-x86_64-unknown-linux-gnu' toolchain
```

Why is cargo suddenly updating the toolchain?

### Solution you'd like

Show an obvious notice that it's downloading a toolchain due to rust-toolchain.toml.

### Notes

_No response_

rami3l:
Requires #4840; closes https://github.com/rust-lang/rustup/issues/4445.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,7 +8,7 @@ use anyhow::{Context, Result, anyhow, bail};
 use serde::Deserialize;
 use thiserror::Error as ThisError;
 use tokio_stream::StreamExt;
-use tracing::trace;
+use tracing::{info, trace, warn};
 
 use crate::dist::AutoInstallMode;
 use crate::{
@@ -387,12 +387,26 @@ impl<'a> Cfg<'a> {
     }
 
     pub(crate) fn should_auto_install(&self) -> Result<bool> {
-        if let Ok(mode) = self.process.var("RUSTUP_AUTO_INSTALL") {
-            Ok(mode != "0")
-        } else {
-            self.settings_file
-                .with(|s| Ok(s.auto_install != Some(AutoInstallMode::Disable)))
+        let should_auto = match self.process.var("RUSTUP_AUTO_INSTALL") {
+            Ok(mode) => mode != "0",
+            Err(_) => self
+                .settings_file
+                .with(|s| Ok(s.auto_install != Some(AutoInstallMode::Disable)))?,
+        };
+        if !should_auto {
+            return Ok(false);
         }
+
+        // We also need to suppress this warning if we're deep inside a recursive call.
+        let recursions = self.process.var("RUSTUP_RECURSION_COUNT");
+        if recursions.is_ok_and(|it| it != "0") {
+            return Ok(true);
+        }
+
+        warn!("auto-install is enabled, active toolchain will be installed if absent");
+        warn!("this might cause rustup commands to take longer time to finish than expected");
+        info!("you may opt out with `RUSTUP_AUTO_INSTALL=0` or `rustup set auto-install disable`");
+        Ok(true)
     }
 
     // Returns a profile, if one exists in the settings file.
--- a/src/config.rs
+++ b/src/config.rs
@@ -746,10 +746,7 @@ impl<'a> Cfg<'a> {
 
     async fn local_toolchain(&self, name: Option<LocalToolchainName>) -> Result<Toolchain<'_>> {
         match name {
-            Some(tc) => {
-                let install_if_missing = self.should_auto_install()?;
-                Toolchain::from_local(tc, install_if_missing, self).await
-            }
+            Some(tc) => Toolchain::from_local(tc, || self.should_auto_install(), self).await,
             None => {
                 let tc = self
                     .maybe_ensure_active_toolchain(None)
--- a/src/config.rs
+++ b/src/config.rs
@@ -717,13 +717,10 @@ impl<'a> Cfg<'a> {
         name: Option<(LocalToolchainName, ActiveSource)>,
     ) -> Result<(Toolchain<'_>, ActiveSource)> {
         match name {
-            Some((tc, source)) => {
-                let install_if_missing = self.should_auto_install()?;
-                Ok((
-                    Toolchain::from_local(tc, install_if_missing, self).await?,
-                    source,
-                ))
-            }
+            Some((tc, src)) => Ok((
+                Toolchain::from_local(tc, || self.should_auto_install(), self).await?,
+                src,
+            )),
             None => {
                 let (tc, source) = self
                     .maybe_ensure_active_toolchain(None)
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -280,6 +280,9 @@ impl Config {
         // Clear current recursion count to avoid messing up related logic
         cmd.env("RUST_RECURSION_COUNT", "");
 
+        // Clear override for auto installation of active toolchain unless explicitly requested
+        cmd.env("RUSTUP_AUTO_INSTALL", "");
+
         // Pass `RUSTUP_CI` over to the test process in case it is required downstream
         if let Some(ci) = env::var_os("RUSTUP_CI") {
             cmd.env("RUSTUP_CI", ci);
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -291,7 +291,7 @@ impl Config {
     /// specified by `args` under the default environment.
     #[must_use]
     pub async fn expect<S: AsRef<OsStr> + Clone + Debug>(&self, args: impl AsRef<[S]>) -> Assert {
-        self.expect_with_env(args, &[]).await
+        self.expect_with_env(args, []).await
     }
 
     /// Returns an [`Assert`] object to check the output of running the command
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -277,6 +277,9 @@ impl Config {
             "/bogus-config-file.toml",
         );
 
+        // Clear current recursion count to avoid messing up related logic
+        cmd.env("RUST_RECURSION_COUNT", "");
+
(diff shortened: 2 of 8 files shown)


src/toolchain.rs
```diff
@@ -53,15 +53,15 @@ pub(crate) struct Toolchain<'a> {
 impl<'a> Toolchain<'a> {
     pub(crate) async fn from_local(
         name: LocalToolchainName,
-        install_if_missing: bool,
+        install_if_missing: impl Fn() -> anyhow::Result<bool>,
```
rami3l, line 56: This change is made so that the `self.should_auto_install()` call is postponed to L64, and that the false positive in L60 can be suppressed.

src/test/clitools.rs
```diff
@@ -277,6 +277,12 @@ impl Config {
             "/bogus-config-file.toml",
         );
 
+        // Clear current recursion count to avoid messing up related logic
+        cmd.env("RUST_RECURSION_COUNT", "");
+
+        // Disable auto installation of active toolchain unless explicitly requested
+        cmd.env("RUSTUP_AUTO_INSTALL", "0");
```
djc: Not sure about this. IMO we should keep the defaults in the tests the same way they are when running real-world `rustup`, otherwise things get confusing.

src/config.rs
```diff
@@ -387,12 +387,26 @@ impl<'a> Cfg<'a> {
     }
 
     pub(crate) fn should_auto_install(&self) -> Result<bool> {
-        if let Ok(mode) = self.process.var("RUSTUP_AUTO_INSTALL") {
-            Ok(mode != "0")
-        } else {
-            self.settings_file
-                .with(|s| Ok(s.auto_install != Some(AutoInstallMode::Disable)))
+        let res = match self.process.var("RUSTUP_AUTO_INSTALL") {
```
djc: Suggest not naming this `res` (which I usually read as short for `result` and expect to be of type `Result`) but `auto_install` (or maybe `auto`).

src/config.rs
```diff
@@ -387,12 +387,26 @@ impl<'a> Cfg<'a> {
...
+            Ok(mode) => mode != "0",
+            Err(_) => self
+                .settings_file
+                .with(|s| Ok(s.auto_install != Some(AutoInstallMode::Disable)))?,
+        };
+        if res
```
djc: Suggest making this more linear to avoid the messy layout:

```rust
if !res {
    return Ok(res);
}

let recursing = self.process.var("RUST_RECURSION_COUNT");
if matches!(recursing.as_deref(), Ok("1")) {
     return Ok(res);
}

warn!(..);
Ok(res)
```

src/config.rs
```diff
@@ -387,12 +387,26 @@ impl<'a> Cfg<'a> {
...
+            // We also need to suppress this warning if we're deep inside a recursive call.
+            && matches!(
+                self.process.var("RUST_RECURSION_COUNT").as_deref(),
+                Err(_) | Ok("0"),
+            )
+        {
+            warn!("auto-install is enabled, active toolchain will be installed if absent");
```
djc: I think this is too strong until we have a clearer plan about how/when we are actually going to disable it by default. So I would only want to show a message when the active toolchain is actually absent, and make it more informational. I do like showing the suggestion on how to disable the behavior!

src/test/clitools.rs
rami3l: @djc FYI this line should be a proof, not a requirement: the changes in this commit modulo this line will continue to work even if `RUSTUP_AUTO_INSTALL` is set to `1` or not set at all. How about unsetting it here?

djc: Unsetting it seems okay to me, as that should make it similar to the default behavior.

src/config.rs
rami3l: Rephrased!

src/test/clitools.rs
rami3l: @djc Hmmm technically those fixes did work but now we are creating a lot of noises in the snapshot tests 🤔
I will try to find a way out of this.

rami3l: @djc ~~How about introducing another env var to remove this warning without disabling auto install?~~

Update: Okay, I think this means that the evaluation of `self.should_auto_install()` is not lazy enough. I'll try to fix that first.

rami3l: I just had another look, it turns out that even `rustup target remove` will call `DistributableToolchain::from_partial()` which in turn calls `Cfg::local_toolchain()` and triggers the warning added here, meaning if the active toolchain is not installed, even removing a target from it will cause the active toolchain be installed first.

rami3l: @djc The old behavior is so confusing to the point where I start to think that the test failures are actually exposing the inconsistencies in it rather than showing a flaw in my implementation...

If we ship it like this I am afraid that we will spam the users. How about making it less frequent by using a global flag to ensure it's only printed once per process, and providing an extra flag to disable these messages?

If you agree with that, we can potentially use that flag to make the snapshot tests pass, as I don't think the changes to be made to the snapshots are very useful in the scope of this PR.

djc: > If we ship it like this I am afraid that we will spam the users. How about making it less frequent by using a global flag to ensure it's only printed once per process, and providing an extra flag to disable these messages?

I'm struggle to remember all the context here. Can you be more specific about what you're proposing? (Maybe with some "screenshots"/mock shell transcripts?)

rami3l: @djc Of course. To elaborate, what I was suggesting is that, while trying to fix this issue, it seemed that the warning message is happening more frequently than necessary, however it isn't, because we now have somewhat surprising behaviors like this:

```console
> # beta is not installed

> RUSTUP_AUTO_INSTALL=1 rustup +beta target remove aarch64-apple-darwin
info: syncing channel updates for 'beta-aarch64-apple-darwin'
info: latest update on 2025-09-28, rust version 1.91.0-beta.4 (aa7859c0d 2025-09-27)
info: downloading component(s)
[...]
```

... and every time this happens, the warning message will show up with this patch.

I am afraid rectifying this will bring breaking changes, and at the same time, shipping this patch as-is will, as I said earlier, spam the users.

So I am suggesting that, if we want to land this, maybe we should find a way to automatically and/or manually disable these messages:

> How about making it less frequent by using a global flag to ensure it's only printed once per process, and providing an extra flag to disable these messages?

For example `RUSTUP_AUTO_INSTALL_WARN=0` will stop these messages from showing up, and a static variable `static HAS_WARNED_AUTO_INSTALL` will prevent these messages from showing up twice in a single run. I admit these are not very good solutions, but I'm proposing this here so that we might have a direction to move forward. Maybe you have better options in mind? I'd love to hear 👂

djc: > because we now have somewhat surprising behaviors like this:

If you mean that the surprising behavior is that a toolchain is installed when a command is invoked to remove a target, I guess we should fix that separately/maybe first?

> I am afraid rectifying this will bring breaking changes, and at the same time, shipping this patch as-is will, as I said earlier, spam the users.

I think you've gotten a bit too risk averse after our auto install adventure. But spamming definitely bad.

> For example `RUSTUP_AUTO_INSTALL_WARN=0` will stop these messages from showing up, and a static variable `static HAS_WARNED_AUTO_INSTALL` will prevent these messages from showing up twice in a single run. I admit these are not very good solutions, but I'm proposing this here so that we might have a direction to move forward. Maybe you have better options in mind? I'd love to hear 👂

I'm definitely not excited about an environment variable to "configure" this behavior but `static` that will allow for rate limiting this makes sense to me. A little spam might be a good thing incentive to turn off auto installs!

rami3l: > I think you've gotten a bit too risk averse after our auto install adventure.

@djc Yes you can probably say that... Be assured that this behavior is a faithful reproduction of that in the old version:

```console
> docker run --rm -it rust:alpine sh

> rustup --version
info: The currently active `rustc` version is `rustc 1.81.0 (eeb90cda1 2024-09-04)`

> rustup +beta target remove aarch64-apple-darwin
info: syncing channel updates for 'beta-aarch64-unknown-linux-musl'
1007.6 KiB / 1007.6 KiB (100 %) 567.4 KiB/s in  1s ETA:  0s
info: latest update on 2025-09-28, rust version 1.91.0-beta.4 (aa7859c0d 2025-09-27)
^C
```

I still think we should rectify the behavior and I don't mind doing a separate PR first for this, but indeed constantly breaking others' workflow did put me into self-doubt every now and then.

djc: I don't think it's constantly and when it did happen we figured out a way forward pretty quickly. But, it's definitely time to start planning for a release again so our releases don't get too big. As team lead, I think you should take the lead on that.

rami3l: @djc Split into #4836. Extra comments/suggestions are welcome :)

rami3l:
Looking back, my current design decisions are slightly wrong. I should have used `UpdateStatus` and issue a warning only if:

- `RUSTUP_AUTO_INSTALL` is enabled.
- `RUST_RECURSION_COUNT` is unset or set to `0`.
- `UpdateStatus` is set to `Installed`.

I'll come back later and see if I can get a new design working.

rami3l:
Closing in favor of #4859.

PR #4455, opened 2025-08-24
Introduce a `Semaphore` to fully honor `RUSTUP_CONCURRENT_DOWNLOADS`

FranciscoTGouveia:
Follow-up on #4450.

Although introducing the `RUSTUP_CONCURRENT_DOWNLOADS` environment variable allows the user to opt in to concurrency, there may be moments where this variable is not fully honored.

An example would be when installing a toolchain (6 components) and the user had `RUSTUP_CONCURRENT_DOWNLOADS=2`.
This would create some situations where there would only be one download being done, where two could have been downloaded instead; this situation has been reported [here](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/Project.3A.20Make.20Rustup.20Concurrent/near/535051136).

This is not a surprise as per the nature of the [`.buffered()`](https://docs.rs/futures-util/0.3.31/futures_util/stream/trait.StreamExt.html#method.buffered) method which, unfortunately, cannot ensure that `n` futures will always be buffered at a same point in time.

To overcome this, and ensure that the environment variable is honored at all times, I decided to introduce a [`Semaphore`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html) that guarantees that there are always `n` downloads happening concurrently.

---

For reference, I leave below a small animation of the old (faulty) behavior:

faulty

... and the new (corrected) behavior:

fixed

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -6,6 +6,7 @@ use std::{
     path::{Path, PathBuf},
     process::ExitStatus,
     str::FromStr,
+    sync::Arc,
     time::Duration,
 };
 
@@ -16,6 +17,7 @@ use console::style;
 use futures_util::stream::StreamExt;
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use itertools::Itertools;
+use tokio::sync::Semaphore;
 use tracing::{info, trace, warn};
 use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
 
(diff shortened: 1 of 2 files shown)


rami3l (approved):
That's quite straightforward, thanks!

It might be interesting to introduce this to [`check_updates()`](https://github.com/rust-lang/rustup/blob/2d429e8e172d5854b6dd7244ecbb0dc3da88a678/src/cli/rustup_mode.rs#L797) as well.

rami3l (approved):

FranciscoTGouveia:
I think this was just a spurious failure :/

PR #4457, opened 2025-08-25
Specify --speed-limit and --speed-time for the fetch script

grahamc:
This change will cause curl to cancel and retry the connection if the connection to the remote is exceedingly slow.

Note this change does apply some "policy" by expecting the user's internet connection to be at least 20% the speed of modern DSL: If the connection can't sustain at least 250,000 bytes per second for 15 seconds, curl will disconnect and try again.

I don't know if the rust project has data on users in that threshold or not, but this is upstreaming a change from our fork of this script in DeterminateSystems/nix-installer.

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -791,17 +791,27 @@ check_help_for() {
 
 # Check if curl supports the --retry flag, then pass it to the curl invocation.
 check_curl_for_retry_support() {
-    local _retry_supported=""
+    local _retry_part=""
+    local _continue_part=""
+    local _speed_limit_part=""
+
     # "unspecified" is for arch, allows for possibility old OS using macports, homebrew, etc.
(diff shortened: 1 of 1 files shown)


rami3l:
@grahamc Thanks for submitting this patch! Would you mind providing more info about the rationale of this change though?

Looking at the solution you are trying to resolve, I'm a bit worried that this will not be a generally good default for our users around the globe. Even if we are adding support for this, we might want to add an extra env var for opting in and specifying the low water mark.

grahamc:
@rami3l We've seen a number of users end up with a stalled download due to a bad route or similar. The connection is alive enough to not trigger timeouts and fail, but not alive enough to complete in a reasonable amount of time. The goal of this PR is to cause it to disconnect and retry on a fresh connection.

PR #4459, opened 2025-08-26
feat(config): add support for `tcsh` shell

### LINKED ISSUE #3413: Add support for tcsh shell
### Problem

On FreeBSD, rustup will generate a .cargo/env that is sourced in .profile. There are two issues here.

1.  tcsh does not source .profile on run, it sources .login
2. Manually sourcing .cargo/env will give a syntax error that too many arguments were given to `case`

### Steps

1. Install FreeBSD on a fresh machine with tcsh as the set shell (I used a VM)
2. Install rustup using the given directions on the website
3. You can either try to source .cargo/env or exit and reenter the shell
4. Try running any rustup commands

### Possible Solution(s)

- Modify .cargo/env to not use BASH specific extensions (I wouldn't quite know what specific syntax you should be using)
- Add sourcing .cargo/env to .login if present,

### Notes

_No response_

### Rustup version

```console
Rustup 1.26.0
```

### Installed toolchains

```console
rustc 1.71.0
```

cachebag:
Fixes #3413

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -399,6 +399,7 @@ This is usually done by running one of the following (note the leading DOT):
     . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
     source "{cargo_home}/env.fish"  # For fish
     source $"{cargo_home_nushell}/env.nu"  # For nushell
+    source "{cargo_home}/env.tcsh"  # For tcsh
 "#
     };
 }
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -67,6 +67,7 @@ fn enumerate_shells() -> Vec<Shell> {
         Box::new(Zsh),
         Box::new(Fish),
         Box::new(Nu),
+        Box::new(Tcsh),
     ]
 }
 
@@ -315,6 +316,55 @@ impl UnixShell for Nu {
 
     fn cargo_home_str(&self, process: &Process) -> Result<Cow<'static, str>> {
         cargo_home_str_with_home("($nu.home-path)", process)
+    }
+}
+
+struct Tcsh;
+
+impl UnixShell for Tcsh {
+    fn does_exist(&self, process: &Process) -> bool {
+        matches!(process.var("SHELL"), Ok(sh) if sh.contains("tcsh"))
+            || utils::find_cmd(&["tcsh"], process).is_some()
+    }
+
+    fn rcfiles(&self, process: &Process) -> Vec<PathBuf> {
+        let mut paths = vec![];
+
+        if let Some(home) = process.home_dir() {
+            paths.push(home.join(".tcshrc"));
+            paths.push(home.join(".cshrc"));
+        }
+
+        paths
+    }
+
+    fn update_rcs(&self, process: &Process) -> Vec<PathBuf> {
+        for f in self.rcfiles(process) {
+            if f.is_file() {
+                return vec![f];
+            }
+        }
+
+        // If neither exists, default to ~/.tcshrc
+        if let Some(home) = process.home_dir() {
+            return vec![home.join(".tcshrc")];
+        }
+
+        vec![]
+    }
+
+    fn env_script(&self) -> ShellScript {
+        ShellScript {
+            name: "env.tcsh",
+            content: include_str!("env.tcsh"),
+        }
+    }
+
+    fn source_string(&self, process: &Process) -> Result<String> {
+        Ok(format!(
+            r#"source "{}/env.tcsh""#,
+            self.cargo_home_str(process)?
+        ))
     }
 }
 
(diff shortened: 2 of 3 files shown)


djc:
Do you actually use tcsh? I'm a little wary of adding a feature when there was only a single issue report about it 2 years ago.

cachebag:
@djc 
I don't, no. I downloaded FreeBSD just for the issue. Frankly I was just looking for an issue I could get my feet wet with, this is my first PR in any OSS project.

rustup-init.sh
```diff
@@ -414,14 +414,6 @@ get_architecture() {
             _ostype=unknown-illumos
             ;;
 
-        SunOS)
```
rami3l: This looks quite fishy... Why did you want to delete support for SunOS?

PS: If it's accidental, consider rewriting the history with `git rebase` so that this part of the changes are gone: it serves no purpose as for now.

cachebag: I thought that I had already reverted this change. I'm not quite sure what happened with `rust-init.sh` there were various changes made to the file unbeknownst to me, I'm assuming this happened when I installed `rustup` on my FreeBSD VM to test the build and changes I made.

I never actually touched the file itself, only the new file I made and the `struct` and `impl` in `shell.rs` as well as the instructions for `tcsh` in `self_update.rs`.

cachebag: @rami3l Okay I believe it's all cleaned up now. Sorry about that.

src/cli/self_update/shell.rs
```diff
@@ -318,6 +319,56 @@ impl UnixShell for Nu {
...
+            paths.push(home.join(".tcshrc"));
+            paths.push(home.join(".cshrc"));
+        }
+
+        paths
+    }
+
+    fn update_rcs(&self, process: &Process) -> Vec<PathBuf> {
+        let files = self.rcfiles(process);
+        for f in &files {
```
rami3l: Nit: `files` is not really used in the later sections, consider going with `for f in self.rcfiles(process)` to save a `.clone()`.

rami3l (commented):
Many thanks! This patch is mostly ready, but before merging it though I would like to clarify something about our committing convention:

It looks like you are only adding support to the `self_update` module, so squashing everything in single commit should be fine; OTOH we use rebase merges, so instead of using merge commits to sync master back to the topic branch, rebase the topic branch directly onto the latest master.

Of course we would still like to find some real users to test it out before it gets merged, do you happen to know any external viewers, or do you think it's okay to ping some potential ones from the original issue?

cachebag:
@rami3l 
Finished, hopefully I understood that correctly. I can ping those from the initial issue, but I do also know a few `rustup` users that aren't involved in the repo. How would you like me to report back when and if they test it out and confirm it works? (Note: I also did spin up a new FreeBSD VM and tested this out if you wanted those results).

rami3l:
> @rami3l Finished, hopefully I understood that correctly. I can ping those from the initial issue, but I do also know a few `rustup` users that aren't involved in the repo. How would you like me to report back when and if they test it out and confirm it works? (Note: I also did spin up a new FreeBSD VM and tested this out if you wanted those results).

@cachebag There is no specific report format to be followed, but basically you need to ensure that the PATH injection works both upon installation (via `source`) and upon shell restart.

It would be nice if at least two (including you) reporting the solution is working as they'd expect; relevant screenshots are welcome if applicable.

This is to ensure the quality of this part of the code because it is not gated by any unit tests as you might have noticed, and we are basically relying on users to crowdsource shell support. Many thanks in advance!

cachebag:
@rami3l 
## Fresh install on FreeBSD VM (via UTM on Mac)
### **Install**
image

---
### **PATH check**
image
image

---

One of my peers should be testing this on their machine tonight and posting directly here.

Nowarr:
@rami3l
### Fresh install on tcsh 
Image 8-28-25 at 8 11 PM
### PATH injection
Image 8-28-25 at 8 20 PM

src/cli/self_update.rs
```diff
@@ -399,6 +399,7 @@ This is usually done by running one of the following (note the leading DOT):
     . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
     source "{cargo_home}/env.fish"  # For fish
     source $"{cargo_home_nushell}/env.nu"  # For nushell
+    source "{cargo_home}/env.tcsh"         # For tcsh
```
rami3l: Nit: Given your last screenshot it might be worthwhile to align this `#` with the fish line so that in most cases these columns are aligned. `nushell` is some sort of an outlier so forget about it...

rami3l: Also, did this suggestion actually work in your "fresh installation" scenario?

cachebag: @rami3l It did, it wasn't shown but @Nowarr initially used that command in order to correctly inject the path, which is what you see in his ss.

cachebag: @rami3l Should be good now. Thank you!

rami3l (approved):

PR #4460, opened 2025-08-27
Report the real elapsed time of a download instead of the cumulative time

FranciscoTGouveia:
Related to the introduction of the `RUSTUP_CONCURRENT_DOWNLOADS` environment variable in #4450.

This fixes the erroneous behavior of making the progress bar start ticking for all downloads at once, even when we are not in a situation of max concurrency.

**PS:** This situation can be observed in [this](https://github.com/rust-lang/rustup/pull/4455#issue-3349739692) animation (the first one).

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -81,6 +81,7 @@ impl DownloadTracker {
     /// Sets the length for a new ProgressBar and gives it a style.
     pub(crate) fn content_length_received(&mut self, content_len: u64, url: &str) {
         if let Some(pb) = self.file_progress_bars.get(url) {
+            pb.reset();
             pb.set_length(content_len);
         }
     }


src/cli/download_tracker.rs
```diff
@@ -81,6 +81,7 @@ impl DownloadTracker {
     /// Sets the length for a new ProgressBar and gives it a style.
     pub(crate) fn content_length_received(&mut self, content_len: u64, url: &str) {
         if let Some(pb) = self.file_progress_bars.get(url) {
+            pb.reset();
```
djc, line 84: Is it still useful that we're building the `ProgressBar` for `DownloadingComponent`, rather than starting at `DownloadContentLengthReceived`?

FranciscoTGouveia: I believe this is justifiable since we don’t have access to the component’s name when creating a `DownloadContentLengthReceived` notification -- only the _URL_.

With that in mind, and wishing to avoid parsing the _URL_ to extract the component’s name, I think it’s best to keep the creation of the `ProgressBar` within the `DownloadingComponent` notification.

Additionally, I’d like to note that when retrying a download, the `DownloadingComponent` notification is not sent again. Moving the creation of the `ProgressBar` to the `DownloadContentLengthReceived` notification could complicate the logic in the `DownloadTracker`, particularly when handling download retries.

FranciscoTGouveia: Moreover, I’d like to add that creating the `ProgressBar` within `DownloadingComponent` is the only way to ensure the output remains in the correct order when handling concurrent downloads.

djc (approved):

djc:
I think this looks okay. It seems a little unintuitive to me to wait until the `DownloadContentLengthReceived` since it would ignore any latency from waiting for the header response, but I guess in practice that's probably a trivial part of the download time anyway.

rami3l (approved):
I think the justification is clear enough, so I agree to merge it first and see how it goes :)

PR #4461, opened 2025-08-27
Replace non_empty_env_var() with Process::var_opt()

djc:
Followup from

- #4422

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,8 +1,8 @@
 use std::fmt::{self, Debug, Display};
+use std::io;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
-use std::{env, io};
 
 use anyhow::{Context, Result, anyhow, bail};
 use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -282,11 +282,12 @@ impl<'a> Cfg<'a> {
...
-        let env_override = non_empty_env_var("RUSTUP_TOOLCHAIN", process)?
-            .map(ResolvableLocalToolchainName::try_from)
-            .transpose()?
-            .map(|t| t.resolve(&default_host_triple))
-            .transpose()?;
+        let env_override = match process.var_opt("RUSTUP_TOOLCHAIN")? {
+            Some(tc) => {
+                Some(ResolvableLocalToolchainName::try_from(&tc)?.resolve(&default_host_triple)?)
+            }
+            None => None,
```
djc, line 289: The `Some(_) => Some(_)` and `None => None` branches are a little ugly, but this saves on the dual `.map().transpose()?` calls so I'm inclined to still prefer this.

src/config.rs
```diff
@@ -982,22 +983,20 @@ impl<'a> Cfg<'a> {
...
-        } else {
-            dist::DEFAULT_DIST_SERVER.to_owned()
-        },
-    )
+    if let Some(s) = process.var_opt("RUSTUP_DIST_SERVER")? {
+        trace!("`RUSTUP_DIST_SERVER` has been set to `{s}`");
+        return Ok(s);
+    }
+
+    let Some(mut root) = process.var_opt("RUSTUP_DIST_ROOT")? else {
```
rami3l, line 992: It might still be interesting to keep the `// For backwards compatibility` comment, maybe right above this line.

rami3l (approved):

src/config.rs
djc: Fair, done.

PR #4466, opened 2025-08-31
Correct `DownloadTracker` reporting on retries and failures

FranciscoTGouveia:
After refactoring the `DownloadTracker` to use `indicatif` and enabling concurrent downloads, a few *UX* issues arose.

This PR addresses those issues, specifically fixing the cluttered output that occurred whenever a download was retried or failed.

An example of the cluttered output described above is shown below:

image

### DIFF
--- a/src/cli/download_tracker.rs
+++ b/src/cli/download_tracker.rs
@@ -1,5 +1,6 @@
 use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
 use std::collections::HashMap;
+use std::time::{Duration, Instant};
 
 use crate::dist::Notification as In;
 use crate::notifications::Notification;
@@ -14,7 +15,12 @@ pub(crate) struct DownloadTracker {
     /// MultiProgress bar for the downloads.
     multi_progress_bars: MultiProgress,
     /// Mapping of URLs being downloaded to their corresponding progress bars.
-    file_progress_bars: HashMap<String, ProgressBar>,
+    /// The `Option<Instant>` represents the instant where the download is being retried,
+    /// allowing us delay the reappearance of the progress bar so that the user can see
+    /// the message "retrying download" for at least a second.
+    /// Without it, the progress bar would reappear immediately, not allowing the user to
+    /// correctly see the message, before the progress bar starts again.
+    file_progress_bars: HashMap<String, (ProgressBar, Option<Instant>)>,
 }
 
 impl DownloadTracker {
@@ -83,12 +89,12 @@ impl DownloadTracker {
         );
         pb.set_message(component);
         self.multi_progress_bars.add(pb.clone());
-        self.file_progress_bars.insert(url, pb);
+        self.file_progress_bars.insert(url, (pb, None));
     }
 
     /// Sets the length for a new ProgressBar and gives it a style.
     pub(crate) fn content_length_received(&mut self, content_len: u64, url: &str) {
-        if let Some(pb) = self.file_progress_bars.get(url) {
+        if let Some((pb, _)) = self.file_progress_bars.get(url) {
             pb.reset();
             pb.set_length(content_len);
         }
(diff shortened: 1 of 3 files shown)


src/cli/download_tracker.rs
```diff
@@ -105,4 +112,30 @@ impl DownloadTracker {
...
+    }
+
+    /// Notifies self that the download is being retried.
+    pub(crate) fn retrying_download(&mut self, url: &str) {
+        let Some(pb) = self.file_progress_bars.get(url) else {
+            return;
+        };
+        pb.set_style(
+            ProgressStyle::with_template(
+                "{msg:>12.bold}  [{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
```
rami3l: Hmmm this looks reasonable impl-wise but I'm afraid the user will not be very clear of what will happen after, say, `download failed after 2s`. Can we somehow delay the reappearance of the progress bar to a later notification, if possible, and saying something like `retrying download` here instead?

rami3l (changes requested):
Many thanks for this PR :)

src/cli/download_tracker.rs
FranciscoTGouveia: Thanks for the comment!

Regarding the part highlighted in this comment, you’re right -- it should explicitly inform the user that the download is being retried. I’ll fix that.
My initial _rationale_ was that, since the progress bar already resets to the beginning, it would be clear that the download was starting over.

As for delaying the reappearance of the progress bar to give users time to see the retrying download message, I’ll look into that now.

src/download/mod.rs
```diff
@@ -218,7 +218,12 @@ async fn download_file_(
         .download_to_path(url, path, resume_from_partial, Some(callback), timeout)
         .await;
 
-    notify_handler(Notification::DownloadFinished(Some(url.as_str())));
+    // The notification should only be sent if the download was successful (i.e. didn't timeout)
+    if res.is_ok() {
+        notify_handler(Notification::DownloadFinished(Some(url.as_str())));
+    } else {
+        notify_handler(Notification::DownloadFailed(url.as_str()));
+    }
```
djc: Nit: suggest rewriting this as

```rust
notify_handler(match &res {
     Ok(_) => Notification::DownloadFinished(Some(url.as_str()))),
     Err(_) => Notification::DownloadFailed(url.as_str()),
})
```

FranciscoTGouveia: Thanks!

src/cli/download_tracker.rs
FranciscoTGouveia: I’ve already addressed this.
The reappearance of the `ProgressBar` is now delayed by 1 second -- do you think that’s sufficient?

rami3l: @FranciscoTGouveia Yes, it looks good to me this time, thanks again :)

src/cli/download_tracker.rs
```diff
@@ -14,7 +15,7 @@ pub(crate) struct DownloadTracker {
     /// MultiProgress bar for the downloads.
     multi_progress_bars: MultiProgress,
     /// Mapping of URLs being downloaded to their corresponding progress bars.
```
rami3l, line 17: You'll need to update the docstring as well :)

rami3l (approved):
LGTM modulo a tiny nit :)

src/cli/download_tracker.rs
```diff
@@ -83,27 +84,40 @@ impl DownloadTracker {
...
     pub(crate) fn data_received(&mut self, len: usize, url: &str) {
-        if let Some(pb) = self.file_progress_bars.get(url) {
-            pb.inc(len as u64);
+        let Some((pb, retry_time)) = self.file_progress_bars.get_mut(url) else {
+            return;
+        };
+        pb.inc(len as u64);
+        if let Some(retry_instant) = retry_time
+            && retry_instant.elapsed() > Duration::from_secs(1)
+        {
```
djc: Invert this to reduce rightward drift:

```rust
match retry_time {
     Some(retry_time) if retry_instant.elapsed() > Duration::from_secs(1) => {}
     _ => return,
}
```

djc (approved):

src/cli/download_tracker.rs
rami3l: @FranciscoTGouveia I think you have misunderstood the comment above. Since `retry_time` is not really used, either the one suggested above or `if !retry_time.is_some_and(..)` with early return will be clearer.

FranciscoTGouveia: Sorry about that! :sweat_smile: 
I’ve already addressed it, opting for an early return.

src/cli/download_tracker.rs
rami3l: @FranciscoTGouveia In addition to what it does, you should also write what it is which is equally important. Something similar to "The `Option<T>` represents the instant of the last retry..." but in your own words.

PR #4469, opened 2025-09-03
feat(install): warn if default linker (`cc`) is missing from PATH

### LINKED ISSUE #4356: Lack of linker is not reported
### Verification

### Problem

This may be a very basic issue, but may be hard to properly fix for new users. As a Rust novice myself, I'm not sure if this is a problem with installer or docs. When following the [Getting Started](https://www.rust-lang.org/learn/get-started), as a user, I am met with an obtuse error before running a `Hello, World!` demo:

```shell
# cargo run
   Compiling hello-rust v0.1.0 (/hello-rust)
error: linker `cc` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `hello-rust` (bin "hello-rust") due to 1 previous error
```

### Steps

The simplest way is to use Docker with a minimal Debian image, created ad-hoc using `docker run --rm -it debian:testing bash`:

1. `apt update && apt install curl`
2. `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
     ```
     info: downloading installer
     
     Welcome to Rust!
     #.....
     Rust is installed now. Great!
     ```
3. `. "$HOME/.cargo/env"` *(as nicely guided by the installer!)*
4. `cargo new hello-rust && cd hello-rust`
5. `cargo run` **<== produces error as shown above**
6. `apt install build-essential`
7. `cargo run` **<== runs correctly**
     ```
        Compiling hello-rust v0.1.0 (/hello-rust)
         Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
          Running `target/debug/hello-rust`
     Hello, world!
     ```

### Possible Solution(s)

When running on Debian, it's as simple as installing `build-essential`, which is a nice a bold shortcut to get all typical tools. However, I think Rustup should probably check for a missing linker and at least suggest it is missing?

### Rustup version

```console
info: The currently active `rustc` version is `rustc 1.87.0 (17067e9ac 2025-05-09)`
```

### Installed toolchains

```console
rustup home:  /root/.rustup

installed toolchains
--------------------

active toolchain
----------------
installed targets:
  x86_64-unknown-linux-gnu
```

### OS version
Debian 13.0 from `debian:testing` image

```console
# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.0
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

cachebag:
Closes #4356 

Test was run in the same environment outlined in the original issue (using `docker run --rm -it debian:testing bash`) and simulates an environment where `cc` is not installed.
<br>
image

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -610,6 +610,9 @@ pub(crate) async fn install(
     };
     md(&mut term, msg);
 
+    #[cfg(unix)]
+    warn_if_default_linker_missing(process);
+
     #[cfg(windows)]
     if !no_prompt {
         // On windows, where installation happens in a console
@@ -740,6 +743,24 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
         opts.profile,
         if !opts.no_modify_path { "yes" } else { "no" }
     )
+}
+
+#[cfg(unix)]
+fn warn_if_default_linker_missing(process: &Process) {
+    // Search for `cc` in PATH
+    if let Some(path) = process.var_os("PATH") {
+        let cc_binary = format!("cc{}", env::consts::EXE_SUFFIX);
+
+        for mut p in env::split_paths(&path) {
+            p.push(&cc_binary);
+            if p.is_file() {
+                return;
+            }
+        }
+    }
+
+    warn!("no default linker (`cc`) was found in your PATH");
+    warn!("many Rust crates require a system C toolchain to build");
 }
 
 fn install_bins(process: &Process) -> Result<()> {
--- a/tests/suite/cli_inst_interactive.rs
+++ b/tests/suite/cli_inst_interactive.rs
@@ -626,3 +626,29 @@ warn: It looks like you have an existing rustup settings file at:
 ...
 "#]]);
 }
+
+#[cfg(unix)]
+#[tokio::test]
+async fn install_warns_if_default_linker_missing() {
+    // Fake PATH pointing to an empty directory so `cc` cannot be found
+    let temp_dir = tempfile::Builder::new()
+        .prefix("emptybin")
+        .tempdir()
+        .unwrap();
+    let temp_dir_path = temp_dir.path().to_str().unwrap();
+
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+    cx.config
+        .expect_with_env(
+            ["rustup-init", "-y", "--no-modify-path"],
+            [("PATH", temp_dir_path)], // override PATH to hide system `cc`
+        )
+        .await
+        .is_ok()
+        .with_stderr(snapbox::str![[r#"
+...
+warn: no default linker (`cc`) was found in your PATH
+warn: many Rust crates require a system C toolchain to build
+...
+"#]]);
+}


src/cli/self_update.rs
```diff
@@ -742,6 +744,20 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+fn warn_if_default_linker_missing(_process: &Process) {
+    match Command::new("cc").arg("--version").output() {
+        Ok(output) if output.status.success() => {
+            // Found a working `cc`, nothing to do.
+        }
+        _ => {
+            warn!("No default linker (`cc`) was found in your PATH.");
+            warn!("Many Rust crates require a system C toolchain to build.");
+            warn!("For example, on Debian/Ubuntu you can install it with:");
+            warn!("    sudo apt install build-essential");
```
djc: I don't think we want to give distro-specific advise in our log messages.

djc (approved):
LGTM modulo a small nit, thanks!

src/cli/self_update.rs
cachebag: Makes sense, thanks. Fixed

src/cli/self_update.rs
```diff
@@ -742,6 +744,18 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
     )
 }
 
+fn warn_if_default_linker_missing(_process: &Process) {
```
rami3l: It looks like the `_process` param is unused, what about removing it?

src/cli/self_update.rs
```diff
@@ -742,6 +744,18 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+    match Command::new("cc").arg("--version").output() {
+        Ok(output) if output.status.success() => {
```
rami3l: I thought the [original suggestion](https://github.com/rust-lang/rustup/issues/4356#issuecomment-2928041615) said checking the `PATH` for `cc`, not running `cc`. I thus suggest extracting the relevant function from [here](https://github.com/rust-lang/rustup/blob/fc0a76bd280293da1bdb8fab484d6468109d6ccd/src/toolchain.rs#L416-L434).

src/cli/self_update.rs
```diff
@@ -742,6 +744,18 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+            // Found a working `cc`, nothing to do.
+        }
+        _ => {
+            warn!("No default linker (`cc`) was found in your PATH.");
```
rami3l: For consistency, don't put capitals at the beginning or dots at the end.

src/cli/self_update.rs
rami3l: Also, this makes your new test case actually workable because you can override `PATH` in the test...

cachebag: I see what you mean, okay I redid the method. Thank you for pointing this out

src/cli/self_update.rs
cachebag: I ended up using it via your other suggestion- initially I felt it right to keep the code consistent but removing it makes sense

ChrisDenton:
This shouldn't run on Windows. We already have code to handle a missing linker (which is called `link` not `cc`).

Also I'm assuming `cc` is the default for all other hosts but I don't know for certain if that's true.

rami3l:
> This shouldn't run on Windows. We already have code to handle a missing linker (which is called `link` not `cc`).
> 
> Also I'm assuming `cc` is the default for all other hosts but I don't know for certain if that's true.

@ChrisDenton I have second thoughts as well, since (if I am reading it right) ~~starting from Rust v1.90 cc will no longer be a hard requirement on Linux x64 (edit: GNU ABI) because it will use the builtin linker instead, making this PR obsolete~~:

~~https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/~~

~~Considering most users who have encountered this issue are indeed using Linux x64, I'm afraid the best option in this case is to do nothing.~~

Edit: The reasoning is inexact. See https://github.com/rust-lang/rustup/pull/4469#issuecomment-3251983199.

cachebag:
@rami3l 

What about non-x86_64 Linux systems? It's probably like 1% of Linux users, so if that's not pertinent due to `lld` fixing the other 99% of cases I'll close this, it makes sense. 

I didn't know that, that's interesting.

rami3l:
@Kobzol I just tried the latest beta release and it looks like `cc` is still required; am I misunderstanding something here regarding the new default linker thing?

Also cc https://github.com/rust-lang/rust/issues/46998.

Kobzol:
Linking is still done through `cc` on Linux, that's just kind of how it works :sweat_smile: It's just that the actually used linker is LLD, and not BFD.

lqd:
`cc` is still the *linker driver*, yup, it knows how to pass the additional object files from the system. It's not like windows where link.exe or lld.exe are the `-Clinker`, on linux it's just some additional configuration to `-Clinker=cc` to tell *it* which linker to use.

rami3l:
@Kobzol @lqd Thanks for the clarification :)

@cachebag I'm really sorry for the false alarm, would you mind pushing this patch towards the finish line? 🙏

cachebag:
> @Kobzol @lqd Thanks for the clarification :)
> 
> 
> 
> @cachebag I'm really sorry for the false alarm, would you mind pushing this patch towards the finish line? 🙏 

@rami3l 
Yes no worries. I did make a few changes based on your suggestions- removed the windows check and just extracted the function you mentioned, which are posted. Was there anything more that should be done beyond that?

rami3l:
> Yes no worries. I did make a few changes based on your suggestions- removed the windows check and just extracted the function you mentioned, which are posted. Was there anything more that should be done beyond that?

@cachebag Did you forget to push? Because I see the current version in an incoherent state, e.g. no special casing for Windows, no updated test snapshot, ...

src/cli/self_update.rs
rami3l: @cachebag Hmmm sorry for not being clear enough, but the point here is not to use `Command::new()`; rather in the link [here](https://github.com/rust-lang/rustup/blob/fc0a76bd280293da1bdb8fab484d6468109d6ccd/src/toolchain.rs#L416-L434) there is a piece of code that searches a bin directly inside `PATH` and you can kind of copy and modify to look for `cc` here.

rami3l: After a second look I concluded that extracting a function from there was not necessary, but you can still use `env::split_paths(&path)` with `path` fetched from `process.env_os("PATH")`.

cachebag:
> > Yes no worries. I did make a few changes based on your suggestions- removed the windows check and just extracted the function you mentioned, which are posted. Was there anything more that should be done beyond that?
> 
> @cachebag Did you forget to push? Because I see the current version in an incoherent state, e.g. no special casing for Windows, no updated test snapshot, ...

Should be good now- my mistake I did forget to push earlier but I adjusted it now to match the behavior of the test case I wrote for this patch.

src/cli/self_update.rs
```diff
@@ -610,6 +610,8 @@ pub(crate) async fn install(
     };
     md(&mut term, msg);
 
+    warn_if_default_linker_missing(process);
```
rami3l, line 614: As @ChrisDenton has mentioned, there is no need to run this code in Windows, so `cfg(unix)` for both the declaration and the call.

src/cli/self_update.rs
```diff
@@ -742,6 +744,30 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+fn warn_if_default_linker_missing(process: &Process) {
+    // Search for `cc` in PATH
+    if let Some(path) = process.var_os("PATH") {
+        let cc_binary = format!("cc{}", env::consts::EXE_SUFFIX);
+
+        for mut p in env::split_paths(&path) {
+            p.push(&cc_binary);
+            if p.is_file() {
+                // Try to run the found `cc` to see if it works
+                match Command::new(&p).arg("--version").output() {
```
rami3l: I believe there is no need to pen test `cc` here with `Command::new()`; the actual Rust project will do the real thing for us and we only need to emit a warning indicating a suspicious behavior.

tests/suite/cli_inst_interactive.rs
```diff
@@ -626,3 +626,30 @@ warn: It looks like you have an existing rustup settings file at:
...
+            ["rustup-init", "-y", "--no-modify-path"],
+            [("PATH", temp_dir_path)], // override PATH to hide system `cc`
+        )
+        .await
+        .is_ok()
+        .with_stderr(snapbox::str![[r#"
+...
+warn: No default linker (`cc`) was found in your PATH.
+warn: Many Rust crates require a system C toolchain to build.
+warn: For example, on Debian/Ubuntu you can install it with:
```
rami3l: The snapshot is still out of date.

tests/suite/cli_inst_interactive.rs
```diff
@@ -626,3 +626,30 @@ warn: It looks like you have an existing rustup settings file at:
 ...
 "#]]);
 }
+
+#[tokio::test]
```
rami3l, line 631: Also, the changes are simple enough so a single commit will suffice.

rami3l (changes requested):

rami3l (approved):
@cachebag Many thanks :)

rami3l:
> `cc` is still the _linker driver_, yup, it knows how to pass the additional object files from the system. It's not like windows where link.exe or lld.exe are the `-Clinker`, on linux it's just some additional configuration to `-Clinker=cc` to tell _it_ which linker to use.

@lqd Just to be sure, do you happen to know where can I find this information about the platform-linker driver correspondence? I'm still a bit afraid of emitting the warning for more platforms than we should 🤔

tests/suite/cli_inst_interactive.rs
```diff
@@ -626,3 +626,28 @@ warn: It looks like you have an existing rustup settings file at:
...
```
rami3l, line 631: This test also needs to be unix-only.

lqd:
> do you happen to know where can I find this information about the platform-linker driver correspondence

Hum. In general, the target `compiler/rustc_target/src/spec` themselves have information that should give the default linker, but that can just be the linker flavor in particular. That can of course be overridden on the CLI...

I don't know if we have such a thing, but https://doc.rust-lang.org/nightly/nightly-rustc/rustc_target/spec/enum.LinkerFlavor.html should give you an idea.

rami3l:
@lqd Thanks for the explanation!

From the link that you have mentioned above I conclude that the false alarms, if they do exist, are pretty much constrained given that we don't interface with WASM, Emscripten, Nvidia PTX and similar, so it should in most cases be a simple Windows/Unix distinction.

As such, I am merging this PR first to see how it goes.

PR #4471, opened 2025-09-05
Interleave downloads with installations when downloading a toolchain

FranciscoTGouveia:
Continuation of #4436.

This change aims to speed up the overall performance of the `rustup [toolchain] install` command by allowing the installation of a component to start immediately after its download completes.
This approach is particularly beneficial for slower connections, as it enables installations to progress in parallel with ongoing downloads -- so by the time all downloads are finished, all installations should be completed as well.

**NB:** This PR is not yet ready for review. I believe this change should be backed by benchmarks and user feedback before moving forward. I’ll share those here in the next couple of days.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,6 +2,7 @@ use std::fmt::{self, Debug, Display};
 use std::io;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::sync::Arc;
 
 use anyhow::{Context, Result, anyhow, bail};
 use serde::Deserialize;
@@ -238,7 +239,7 @@ pub(crate) struct Cfg<'a> {
     pub toolchains_dir: PathBuf,
     update_hash_dir: PathBuf,
     pub download_dir: PathBuf,
-    pub tmp_cx: temp::Context,
+    pub tmp_cx: Arc<temp::Context>,
     pub toolchain_override: Option<ResolvableToolchainName>,
     env_override: Option<LocalToolchainName>,
     pub dist_root_url: String,
@@ -299,7 +300,10 @@ impl<'a> Cfg<'a> {
         };
 
         let dist_root_server = dist_root_server(process)?;
-        let tmp_cx = temp::Context::new(rustup_dir.join("tmp"), dist_root_server.as_str());
+        let tmp_cx = Arc::new(temp::Context::new(
+            rustup_dir.join("tmp"),
+            dist_root_server.as_str(),
+        ));
         let dist_root = dist_root_server + "/dist";
 
         let cfg = Self {
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -28,13 +28,13 @@ pub(crate) const VERSION_FILE: &str = "rust-installer-version";
 
 pub trait Package: fmt::Debug {
     fn contains(&self, component: &str, short_name: Option<&str>) -> bool;
-    fn install<'a>(
+    fn install(
         &self,
         target: &Components,
         component: &str,
         short_name: Option<&str>,
-        tx: Transaction<'a>,
-    ) -> Result<Transaction<'a>>;
+        tx: Transaction,
+    ) -> Result<Transaction>;
     fn components(&self) -> Vec<String>;
 }
 
@@ -78,13 +78,13 @@ impl Package for DirectoryPackage {
                 false
             }
     }
-    fn install<'a>(
+    fn install(
         &self,
         target: &Components,
         name: &str,
         short_name: Option<&str>,
-        tx: Transaction<'a>,
-    ) -> Result<Transaction<'a>> {
+        tx: Transaction,
+    ) -> Result<Transaction> {
         let actual_name = if self.components.contains(name) {
             name
         } else if let Some(n) = short_name {
@@ -139,7 +139,7 @@ impl Package for DirectoryPackage {
 pub(crate) struct TarPackage(DirectoryPackage, temp::Dir);
 
 impl TarPackage {
-    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg<'_>) -> Result<Self> {
+    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg) -> Result<Self> {
         let temp_dir = dl_cfg.tmp_cx.new_directory()?;
         let mut archive = tar::Archive::new(stream);
         // The rust-installer packages unpack to a directory called
@@ -159,7 +159,7 @@ impl TarPackage {
 fn unpack_ram(
     io_chunk_size: usize,
     effective_max_ram: Option<usize>,
-    dl_cfg: &DownloadCfg<'_>,
+    dl_cfg: &DownloadCfg,
 ) -> usize {
     const RAM_ALLOWANCE_FOR_RUSTUP_AND_BUFFERS: usize = 200 * 1024 * 1024;
     let minimum_ram = io_chunk_size * 2;
@@ -285,7 +285,7 @@ enum DirStatus {
 fn unpack_without_first_dir<R: Read>(
     archive: &mut tar::Archive<R>,
     path: &Path,
-    dl_cfg: &DownloadCfg<'_>,
+    dl_cfg: &DownloadCfg,
 ) -> Result<()> {
     let entries = archive.entries()?;
     let effective_max_ram = match effective_limits::memory_limit() {
@@ -296,7 +296,7 @@ fn unpack_without_first_dir<R: Read>(
         }
     };
     let unpack_ram = unpack_ram(IO_CHUNK_SIZE, effective_max_ram, dl_cfg);
-    let mut io_executor: Box<dyn Executor> = get_executor(unpack_ram, dl_cfg.process)?;
+    let mut io_executor: Box<dyn Executor> = get_executor(unpack_ram, &dl_cfg.process)?;
 
     let mut directories: HashMap<PathBuf, DirStatus> = HashMap::new();
     // Path is presumed to exist. Call it a precondition.
@@ -530,13 +530,13 @@ impl Package for TarPackage {
     fn contains(&self, component: &str, short_name: Option<&str>) -> bool {
         self.0.contains(component, short_name)
     }
-    fn install<'b>(
+    fn install(
         &self,
         target: &Components,
         component: &str,
         short_name: Option<&str>,
-        tx: Transaction<'b>,
-    ) -> Result<Transaction<'b>> {
+        tx: Transaction,
+    ) -> Result<Transaction> {
         self.0.install(target, component, short_name, tx)
     }
     fn components(&self) -> Vec<String> {
@@ -548,7 +548,7 @@ impl Package for TarPackage {
 pub(crate) struct TarGzPackage(TarPackage);
 
 impl TarGzPackage {
-    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg<'_>) -> Result<Self> {
+    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg) -> Result<Self> {
         let stream = flate2::read::GzDecoder::new(stream);
         Ok(TarGzPackage(TarPackage::new(stream, dl_cfg)?))
     }
@@ -558,13 +558,13 @@ impl Package for TarGzPackage {
     fn contains(&self, component: &str, short_name: Option<&str>) -> bool {
         self.0.contains(component, short_name)
     }
-    fn install<'b>(
+    fn install(
         &self,
         target: &Components,
         component: &str,
         short_name: Option<&str>,
-        tx: Transaction<'b>,
-    ) -> Result<Transaction<'b>> {
+        tx: Transaction,
+    ) -> Result<Transaction> {
         self.0.install(target, component, short_name, tx)
     }
     fn components(&self) -> Vec<String> {
@@ -576,7 +576,7 @@ impl Package for TarGzPackage {
 pub(crate) struct TarXzPackage(TarPackage);
 
 impl TarXzPackage {
-    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg<'_>) -> Result<Self> {
+    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg) -> Result<Self> {
         let stream = xz2::read::XzDecoder::new(stream);
         Ok(TarXzPackage(TarPackage::new(stream, dl_cfg)?))
     }
@@ -586,13 +586,13 @@ impl Package for TarXzPackage {
     fn contains(&self, component: &str, short_name: Option<&str>) -> bool {
         self.0.contains(component, short_name)
     }
-    fn install<'b>(
+    fn install(
         &self,
         target: &Components,
         component: &str,
         short_name: Option<&str>,
-        tx: Transaction<'b>,
-    ) -> Result<Transaction<'b>> {
+        tx: Transaction,
+    ) -> Result<Transaction> {
         self.0.install(target, component, short_name, tx)
     }
     fn components(&self) -> Vec<String> {
@@ -604,7 +604,7 @@ impl Package for TarXzPackage {
 pub(crate) struct TarZStdPackage(TarPackage);
 
 impl TarZStdPackage {
-    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg<'_>) -> Result<Self> {
+    pub(crate) fn new<R: Read>(stream: R, dl_cfg: &DownloadCfg) -> Result<Self> {
         let stream = zstd::stream::read::Decoder::new(stream)?;
         Ok(TarZStdPackage(TarPackage::new(stream, dl_cfg)?))
     }
@@ -614,13 +614,13 @@ impl Package for TarZStdPackage {
     fn contains(&self, component: &str, short_name: Option<&str>) -> bool {
         self.0.contains(component, short_name)
     }
-    fn install<'b>(
+    fn install(
         &self,
         target: &Components,
         component: &str,
         short_name: Option<&str>,
-        tx: Transaction<'b>,
-    ) -> Result<Transaction<'b>> {
+        tx: Transaction,
+    ) -> Result<Transaction> {
         self.0.install(target, component, short_name, tx)
     }
     fn components(&self) -> Vec<String> {
--- a/src/dist/component/transaction.rs
+++ b/src/dist/component/transaction.rs
@@ -9,8 +9,11 @@
 //! FIXME: This uses ensure_dir_exists in some places but rollback
 //! does not remove any dirs created by it.
 
-use std::fs::File;
-use std::path::{Path, PathBuf};
+use std::{
+    fs::File,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
 use anyhow::{Context, Result, anyhow};
 use tracing::{error, info};
@@ -34,16 +37,16 @@ use crate::utils;
 ///
 /// All operations that create files will fail if the destination
 /// already exists.
-pub struct Transaction<'a> {
+pub struct Transaction {
     prefix: InstallPrefix,
     changes: Vec<ChangedItem>,
-    tmp_cx: &'a temp::Context,
+    tmp_cx: Arc<temp::Context>,
     committed: bool,
-    process: &'a Process,
+    process: Arc<Process>,
 }
 
-impl<'a> Transaction<'a> {
-    pub fn new(prefix: InstallPrefix, tmp_cx: &'a temp::Context, process: &'a Process) -> Self {
+impl Transaction {
+    pub fn new(prefix: InstallPrefix, tmp_cx: Arc<temp::Context>, process: Arc<Process>) -> Self {
         Transaction {
             prefix,
             changes: Vec::new(),
@@ -92,8 +95,13 @@ impl<'a> Transaction<'a> {
     /// Remove a file from a relative path to the install prefix.
     pub fn remove_file(&mut self, component: &str, relpath: PathBuf) -> Result<()> {
         assert!(relpath.is_relative());
-        let item =
-            ChangedItem::remove_file(&self.prefix, component, relpath, self.tmp_cx, self.process)?;
+        let item = ChangedItem::remove_file(
+            &self.prefix,
+            component,
+            relpath,
+            self.tmp_cx.clone(),
+            &self.process,
+        )?;
         self.change(item);
         Ok(())
     }
@@ -102,8 +110,13 @@ impl<'a> Transaction<'a> {
     /// install prefix.
     pub fn remove_dir(&mut self, component: &str, relpath: PathBuf) -> Result<()> {
         assert!(relpath.is_relative());
-        let item =
-            ChangedItem::remove_dir(&self.prefix, component, relpath, self.tmp_cx, self.process)?;
+        let item = ChangedItem::remove_dir(
+            &self.prefix,
+            component,
+            relpath,
+            self.tmp_cx.clone(),
+            &self.process,
+        )?;
         self.change(item);
         Ok(())
     }
@@ -129,7 +142,7 @@ impl<'a> Transaction<'a> {
     /// This is used for arbitrarily manipulating a file.
     pub fn modify_file(&mut self, relpath: PathBuf) -> Result<()> {
         assert!(relpath.is_relative());
-        let item = ChangedItem::modify_file(&self.prefix, relpath, self.tmp_cx)?;
+        let item = ChangedItem::modify_file(&self.prefix, relpath, self.tmp_cx.clone())?;
         self.change(item);
         Ok(())
     }
@@ -142,7 +155,7 @@ impl<'a> Transaction<'a> {
         src: &Path,
     ) -> Result<()> {
         assert!(relpath.is_relative());
-        let item = ChangedItem::move_file(&self.prefix, component, relpath, src, self.process)?;
+        let item = ChangedItem::move_file(&self.prefix, component, relpath, src, &self.process)?;
         self.change(item);
         Ok(())
     }
@@ -150,26 +163,26 @@ impl<'a> Transaction<'a> {
     /// Recursively move a directory to a relative path of the install prefix.
     pub(crate) fn move_dir(&mut self, component: &str, relpath: PathBuf, src: &Path) -> Result<()> {
         assert!(relpath.is_relative());
-        let item = ChangedItem::move_dir(&self.prefix, component, relpath, src, self.process)?;
+        let item = ChangedItem::move_dir(&self.prefix, component, relpath, src, &self.process)?;
         self.change(item);
         Ok(())
     }
 
-    pub(crate) fn temp(&self) -> &'a temp::Context {
-        self.tmp_cx
+    pub(crate) fn temp(&self) -> Arc<temp::Context> {
+        self.tmp_cx.clone()
     }
 }
 
 /// If a Transaction is dropped without being committed, the changes
 /// are automatically rolled back.
-impl Drop for Transaction<'_> {
+impl Drop for Transaction {
     fn drop(&mut self) {
         if !self.committed {
             info!("rolling back changes");
             for item in self.changes.iter().rev() {
                 // ok_ntfy!(self.notify_handler,
                 //          Notification::NonFatalError,
-                match item.roll_back(&self.prefix, self.process) {
+                match item.roll_back(&self.prefix, &self.process) {
                     Ok(()) => {}
                     Err(e) => error!("{e}"),
                 }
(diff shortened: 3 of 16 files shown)


src/dist/manifestation.rs
```diff
@@ -246,17 +213,94 @@ impl Manifestation {
...
+                    hashes
+                },
+                async {
+                    let mut current_tx = tx;
+                    let mut counter = 0;
+                    loop {
+                        if counter >= total_components {
+                            break;
+                        }
+                        if let Some(message) = download_rx.recv().await {
```
rami3l: Could it be that the counter never gets incremented because `.recv().await` returns `None`?

Did you mean something like this instead?

```rs
while counter < total_components && let Some(_) = ... {  ...; counter += 1; }
```

src/dist/manifestation.rs
```diff
@@ -246,17 +213,94 @@ impl Manifestation {
...
+                                        installer_file,
+                                        tmp_cx,
+                                        download_cfg,
+                                        new_manifest,
+                                        current_tx,
+                                    ) {
+                                        Ok(new_tx) => {
+                                            current_tx = new_tx;
+                                        }
+                                        Err(e) => {
```
rami3l: If I'm not mistaken, `Err(e) => return Err(e)` is just `?` 🤔

tests/suite/cli_rustup.rs
```diff
@@ -35,7 +35,6 @@ async fn rustup_stable() {
         .with_stderr(snapbox::str![[r#"
 info: syncing channel updates for 'stable-[HOST_TRIPLE]'
 info: latest update on 2015-01-02, rust version 1.1.0 (hash-stable-1.1.0)
-info: downloading component[..]
```
rami3l, line 38: Possibly squash this commit with the one that has broken these tests.

src/dist/notifications.rs
```diff
@@ -122,6 +124,13 @@ impl Display for Notification<'_> {
                     write!(f, "installing component '{}' for '{}'", c, t.unwrap())
                 }
             }
+            ComponentInstalled(c, h, t) => {
+                if Some(h) == t.as_ref() || t.is_none() {
```
rami3l: Nit: Now that we are at it, I would probably give an extra commit to avoid the `t.unwrap()` here by flipping the `if` and writing something like:

```rs
if let Some(t) = t && t != h { ... } else { ... }
```

src/cli/download_tracker.rs
```diff
@@ -170,8 +174,11 @@ impl DownloadTracker {
             && let Some((pb, _)) = self.file_progress_bars.get(&key)
         {
             pb.set_style(
-                ProgressStyle::with_template(
-                    "{msg:>12.bold}  downloaded {total_bytes} in {elapsed} installing now {spinner:.green}",
+                ProgressStyle::with_template( if pb.position() != 0 {
+                    "{msg:>12.bold}  downloaded {total_bytes} in {elapsed} installing now {spinner:.green}"
+                } else {
+                    "{msg:>12.bold}  component already downloaded, installing now {spinner:.green}"
```
rami3l: Nit: the `, installing now`, ` installing now` and ` and installed` might seem inconsistent. How about just ` and installed` and ` and installing`? I believe the spinner can compensate for the minor difference.

src/dist/notifications.rs
FranciscoTGouveia: I did this based on how other Notifications were being handled.
Should I refactor all of them now, leave that for another PR, or just change this one?

rami3l: @FranciscoTGouveia It should be worthwhile to refactor them all in a separate commit (before your addition) lest we forget about it :)

src/dist/manifestation.rs
```diff
@@ -224,79 +224,104 @@ impl Manifestation {
...
+                            let new_manifest = Arc::clone(&new_manifest);
+                            let tmp_cx = download_cfg.tmp_cx.clone();
+                            let download_cfg = Arc::clone(&download_cfg);
+                            move || {
+                                let download_cfg = (&*download_cfg).into();
+                                this.install_component(
+                                    component,
+                                    format,
+                                    installer_file,
+                                    Arc::clone(&tmp_cx),
```
rami3l: Note: Looks like there are extra clones already... So in addition to making sure the new implementation is working correctly, it would also be useful to audit the use of `Arc::clone()`.

FranciscoTGouveia: Thanks for noticing this!
I will revisit all uses of `Arc::clone()` now.

FranciscoTGouveia:
As noted in the opening message of the PR, I now present some benchmarks that will enable us to move forward with this change.
As stated before, this will bump the install and update commands' performance, as when the downloads finish, the installations have already executed.

The results are presented in the plot below, for both low-speed and high-speed internet connections.

image

These results were obtained by running the `rustup toolchain install nightly` command 10x times for each combination of variables (`RUSTUP_CONCURRENT_DOWNLOADS`, interleaved or not, and internet connection).
Please note that the *y axis* do not start on zero, and cannot be directly compared between themselves.

---
Below, I also leave a small animation of the previous and current behavior of the `rustup toolchain install` command to illustrate the *UX* changes that this PR also implies.
On the left you can see the previous version and on the right, the version with the interleaved installations.
Please note that this is just an animation and does not have the purpose of showing any kind of speedup.

animation4471-ezgif com-speed

Thanks for taking the time to review this PR!

src/config.rs
```diff
@@ -330,13 +330,13 @@ impl<'a> Cfg<'a> {
     /// construct a download configuration
     pub(crate) fn download_cfg(
         &'a self,
-        notify_handler: &'a dyn Fn(crate::dist::Notification<'_>),
+        notify_handler: Arc<dist::notifications::NotifyHandler>,
     ) -> DownloadCfg<'a> {
         DownloadCfg {
             dist_root: &self.dist_root_url,
-            tmp_cx: &self.tmp_cx,
+            tmp_cx: Arc::new(self.tmp_cx.clone()),
```
rami3l: What if we make `self.tmp_cx` an `Arc<>` directly?

src/config.rs
```diff
@@ -330,13 +330,13 @@ impl<'a> Cfg<'a> {
...
             download_dir: &self.download_dir,
-            notify_handler,
+            notify_handler: Arc::clone(&notify_handler),
```
rami3l: This `Arc::clone()` is redundant.

src/dist/component/package.rs
```diff
@@ -551,36 +552,36 @@ impl Package for TarPackage<'_> {
...
-        notify_handler: Option<&'a dyn Fn(Notification<'_>)>,
+        tmp_cx: Arc<temp::Context>,
+        notify_handler: Option<&dyn Fn(Notification<'_>)>,
         process: &Process,
     ) -> Result<Self> {
         let stream = flate2::read::GzDecoder::new(stream);
         Ok(TarGzPackage(TarPackage::new(
             stream,
-            tmp_cx,
+            Arc::clone(&tmp_cx),
```
rami3l: This `Arc::clone()` is redundant.

src/dist/component/package.rs
```diff
@@ -589,36 +590,36 @@ impl Package for TarGzPackage<'_> {
...
-        notify_handler: Option<&'a dyn Fn(Notification<'_>)>,
+        tmp_cx: Arc<temp::Context>,
+        notify_handler: Option<&dyn Fn(Notification<'_>)>,
         process: &Process,
     ) -> Result<Self> {
         let stream = xz2::read::XzDecoder::new(stream);
         Ok(TarXzPackage(TarPackage::new(
             stream,
-            tmp_cx,
+            Arc::clone(&tmp_cx),
```
rami3l: This `Arc::clone()` is redundant.

src/dist/component/package.rs
```diff
@@ -627,36 +628,36 @@ impl Package for TarXzPackage<'_> {
...
-        notify_handler: Option<&'a dyn Fn(Notification<'_>)>,
+        tmp_cx: Arc<temp::Context>,
+        notify_handler: Option<&dyn Fn(Notification<'_>)>,
         process: &Process,
     ) -> Result<Self> {
         let stream = zstd::stream::read::Decoder::new(stream)?;
         Ok(TarZStdPackage(TarPackage::new(
             stream,
-            tmp_cx,
+            Arc::clone(&tmp_cx),
```
rami3l: This `Arc::clone()` is redundant.

src/dist/download.rs
```diff
@@ -16,15 +17,48 @@ use crate::utils;
...
-    pub tmp_cx: &'a temp::Context,
+    pub tmp_cx: Arc<temp::Context>,
     pub download_dir: &'a PathBuf,
-    pub notify_handler: &'a dyn Fn(Notification<'_>),
+    pub notify_handler: Arc<NotifyHandler>,
     pub process: &'a Process,
 }
 
+#[derive(Clone)]
+pub struct OwnedDownloadCfg {
```
rami3l: Consider removing `OwnedDownloadCfg` and make `DownloadCfg` use `Arc<str>`, `Arc<Path>` and `Arc<Process>` instead, so the lifetime parameters can be removed.

src/dist/temp.rs
```diff
@@ -185,7 +191,7 @@ impl Context {
                 fs::File::create(&temp_file)
                     .with_context(|| CreatingError::File(PathBuf::from(&temp_file)))?;
                 return Ok(File {
-                    cfg: self,
+                    cfg: Arc::clone(&self),
```
rami3l: This `Arc::clone()` is redundant.

src/dist/temp.rs
```diff
@@ -159,18 +165,18 @@ impl Context {
                 fs::create_dir(&temp_dir)
                     .with_context(|| CreatingError::Directory(PathBuf::from(&temp_dir)))?;
                 return Ok(Dir {
-                    cfg: self,
+                    cfg: Arc::clone(&self),
```
rami3l: This `Arc::clone()` is redundant.

src/dist/temp.rs
```diff
@@ -1,8 +1,11 @@
-use std::fmt::{self, Display};
 use std::fs;
 use std::io;
 use std::ops;
 use std::path::{Path, PathBuf};
+use std::{
```
rami3l: Consider splitting the import here for consistency.

src/dist/manifestation.rs
```diff
@@ -224,79 +224,101 @@ impl Manifestation {
...
-                            current_tx,
-                        )?;
-                        (download_cfg.notify_handler)(Notification::ComponentInstalled(
-                            &component.short_name(new_manifest),
+                        let component_name = component.short_name(&new_manifest);
+                        let notify_handler = Arc::clone(&download_cfg.notify_handler);
+                        let new_tx = {
+                            let this = Arc::clone(&self);
+                            let new_manifest = Arc::clone(&new_manifest);
+                            let tmp_cx = download_cfg.tmp_cx.clone();
```
rami3l: Nit: It is preferable to use explicit `Arc::clone()` for such variables.

src/cli/common.rs
```diff
@@ -126,14 +125,14 @@ pub(crate) fn read_line(process: &Process) -> Result<String> {
 
 pub(super) struct Notifier {
     tracker: Mutex<DownloadTracker>,
-    ram_notice_shown: RefCell<bool>,
+    ram_notice_shown: RwLock<bool>,
```
rami3l: Given the sheer upcoming size of that one particular commit, it might be better to split all changes regarding the replacement of RefCell to another commit, just to make the review process easier.

rami3l (commented):
Modulo the previously requested changes and the two extra nits, I think this PR is quite ready. Thanks a lot, @FranciscoTGouveia!

src/dist/notifications.rs
FranciscoTGouveia: If you agree, I would like to leave this for a follow-up PR, right after this one is merged.
So I suggest leaving this as unresolved for now.

src/dist/manifestation.rs
```diff
@@ -591,6 +541,77 @@ impl Manifestation {
...
+
+    fn install_component<'a>(
+        &self,
+        component: Component,
+        format: CompressionKind,
+        installer_file: File,
+        tmp_cx: &temp::Context,
+        download_cfg: &DownloadCfg<'_>,
+        new_manifest: &Manifest,
+        tx: Transaction<'a>,
```
djc: I think the number of arguments is here too large, and this should be rethought to have a named field `struct` that can hold all/most of the arguments.

I think the pre-existing code here could benefit from some refactoring as well, for example by packaging together the arguments passed into the compression wrappers. The uninitialized variables structure also sounds out as pretty unidiomatic.

src/dist/manifestation.rs
```diff
@@ -248,17 +215,76 @@ impl Manifestation {
...
+                mpsc::channel::<Result<(Component, CompressionKind, File)>>(total_components);
+
+            let semaphore = Arc::new(Semaphore::new(concurrent_downloads));
+            let component_stream =
+                tokio_stream::iter(components.into_iter()).map(|(component, format, url, hash)| {
+                    let sem = semaphore.clone();
+                    let download_tx_cloned = download_tx.clone();
+                    async move {
+                        let _permit = sem.acquire().await.unwrap();
+                        self.download_component(
```
djc: Pre-existing, but this argument list is also just too long, and I think we should find a different way to express this before layering on more complexity.

src/dist/manifestation.rs
```diff
@@ -248,17 +215,76 @@ impl Manifestation {
...
+            // Create a channel to communicate whenever a download is done and the component can be installed
+            // The `mpsc` channel was used as we need to send many messages from one producer (download's thread) to one consumer (install's thread)
+            // This is recommended in the official docs: https://docs.rs/tokio/latest/tokio/sync/index.html#mpsc-channel
+            let total_components = components.len();
+            let (download_tx, mut download_rx) =
+                mpsc::channel::<Result<(Component, CompressionKind, File)>>(total_components);
+
+            let semaphore = Arc::new(Semaphore::new(concurrent_downloads));
+            let component_stream =
+                tokio_stream::iter(components.into_iter()).map(|(component, format, url, hash)| {
```
djc: Suggest writing this closure as a named function.

src/dist/manifestation.rs
```diff
@@ -248,17 +215,76 @@ impl Manifestation {
...
+                            new_manifest,
+                            download_tx_cloned,
+                        )
+                        .await
+                    }
+                });
+
+            let mut stream = component_stream.buffered(components_len);
+            let download_handle = async {
+                let mut hashes = Vec::new();
```
djc: Suggest writing this as a named function.

src/dist/manifestation.rs
```diff
@@ -248,17 +215,76 @@ impl Manifestation {
...
+                            hashes.push(hash);
+                        }
+                        Err(e) => {
+                            let _ = download_tx.send(Err(e)).await;
+                        }
+                    }
+                }
+                hashes
+            };
+            let install_handle = async {
```
djc: Suggest writing this as a named function.

src/cli/download_tracker.rs
```diff
@@ -151,4 +155,24 @@ impl DownloadTracker {
...
     }
+
+    /// Notifies that the downloaded component is being installed.
+    pub(crate) fn installing_component(&mut self, component: &str) {
+        let key = self
+            .file_progress_bars
+            .keys()
+            .find(|comp| comp.contains(component))
+            .cloned();
+        if let Some(key) = key
```
djc: Nit: invert these for early returns.

src/cli/download_tracker.rs
```diff
@@ -168,7 +171,28 @@ impl DownloadTracker {
...
+                )
+                .unwrap()
+                .tick_chars(r"|/-\ "),
+            );
+            pb.enable_steady_tick(Duration::from_millis(100));
+        }
+    }
+
+    /// Notifies self that the component has been installed.
+    pub(crate) fn component_installed(&mut self, component: &str) {
```
djc: Can we share the duplicate code?

src/dist/notifications.rs
```diff
@@ -122,6 +124,15 @@ impl Display for Notification<'_> {
...
                 }
             }
+            ComponentInstalled(c, h, t) => {
+                if let Some(t) = t
+                    && t != h
+                {
+                    write!(f, "installing component '{c}' for '{t}'")
+                } else {
+                    write!(f, "installing component '{c}'")
+                }
```
djc: Nit: I think this would look better as a `match`:

```rust
match t {
    Some(t) if != h => write!(f, ..),
    _ => write!(f, ..),
}
```

src/cli/download_tracker.rs
```diff
@@ -133,8 +133,12 @@ impl DownloadTracker {
...
         };
         pb.set_style(
-            ProgressStyle::with_template("{msg:>12.bold}  downloaded {total_bytes} in {elapsed}")
-                .unwrap(),
+            ProgressStyle::with_template(if pb.position() != 0 {
+                "{msg:>12.bold}  downloaded {total_bytes} in {elapsed}"
+            } else {
+                "{msg:>12.bold}  component already downloaded"
+            })
+            .unwrap(),
```
djc: Again there's a lot of duplication here, suggest adding a little abstraction to make it more obvious what happens here.

src/dist/component/transaction.rs
```diff
@@ -34,21 +34,21 @@ use crate::utils;
...
 /// All operations that create files will fail if the destination
 /// already exists.
-pub struct Transaction<'a> {
+pub struct Transaction {
     prefix: InstallPrefix,
-    changes: Vec<ChangedItem<'a>>,
-    tmp_cx: &'a temp::Context,
-    notify_handler: &'a dyn Fn(Notification<'_>),
+    changes: Vec<ChangedItem>,
+    tmp_cx: Arc<temp::Context>,
```
djc, line 43: Seems to me that `tmp_cx`, `notify_handler` and `process` should be bundled up together.

djc:
Very excited about these performance improvements! Personally I don't really feel comfortable with the added complexity in the current changes -- I think there should be some more refactoring before this can be cleanly added.

src/dist/notifications.rs
rami3l: Okay, let's do https://github.com/rust-lang/rustup/pull/4471#discussion_r2362667381 for the particular change in this PR, and then move the rest to that style in a subsequent PR.

rbtcollins:
> These results were obtained by running the rustup toolchain install nightly command 10x times for each combination of variables (RUSTUP_CONCURRENT_DOWNLOADS, interleaved or not, and internet connection).
Please note that the y axis do not start on zero, and cannot be directly compared between themselves.

As I read these results:
low speed connections : 
interleaved gives 65 seconds to 59 seconds or a 10% reduction, somewhat stable for 2 or more concurrent downloads.

high speed connections:
interleaved gives 18-12 for low concurrency and 15-13 for higher concurrency, or 33% reduction for 1 or 2 concurrent downloads, and a 13% reduction for higher concurrency, with best performance at 2 concurrent downloads, and performance decreasing for successively more downloads with interleaving, but increasing for more concurrent downloads w/out interleaving.


I suspect some interaction with memory limits showing up in the unpack phase there; I suggest raising the working memory much higher, if you're on a machine that has capacity, to see if thats the case.

But also it suggests the default should probably be quite conservative, as we still need to support unpacking on Rasberry PI class machines.

src/dist/manifestation.rs
FranciscoTGouveia: Thanks for the comment and for the associated PR to simplify this!
I believe that that PR addressed everything; also, *clippy* is no longer complaining.
If you agree, we can resolve this conversation.

src/dist/manifestation.rs
FranciscoTGouveia: Similar to the previous comment, I believe PR 4489, which you’ve already merged, takes care of this.
If you agree, we can resolve this conversation.

src/dist/manifestation.rs
```diff
@@ -249,16 +216,77 @@ impl Manifestation {
...
-                installer_file,
-                tmp_cx,
-                download_cfg,
-                new_manifest,
-                tx,
-            )?;
+        if components_len > 0 {
+            // Create a channel to communicate whenever a download is done and the component can be installed
+            // The `mpsc` channel was used as we need to send many messages from one producer (download's thread) to one consumer (install's thread)
+            // This is recommended in the official docs: https://docs.rs/tokio/latest/tokio/sync/index.html#mpsc-channel
```
djc, line 229: I can't find it now, but I thought I wrote a comment on wondering why we're using a channel here, when we could instead compose the futures that download and install together.

rami3l: Hmmm would you mind clarifying a little more what you meant by composing the futures?

The current constraints as we previously agreed was that the downloads are concurrent but the installations happening alongside them aren't concurrent within themselves, so I think a channel would be ideal for that case in favor of, say, a mutex.

djc: So you're saying, we only do one installation at a time? If that's the case, I would suggest something more like a `FuturesUnordered` containing all the downloads, and then pulling from that to run the installations.

src/cli/download_tracker.rs
FranciscoTGouveia: I believe the recent PRs involving the simplification of the notification system already addressed this.
 If you agree, we can resolve this conversation.

djc: Sure, feel free to resolve any of my comments that you think are no longer applicable!

src/dist/manifestation.rs
rami3l: @djc I discussed with @FranciscoTGouveia and we would like to know: how about leaving this simplification to another PR now that we already have something working and ready to be test-driven? Leaving this outside of mainline for long would definitely mean fewer chances to test the latest changes...

djc: Yeah, I can live with that.

FranciscoTGouveia:
Hello everyone!

I think all review comments should be addressed by now with [this](https://github.com/rust-lang/rustup/pull/4471#discussion_r2385119158) and [this](https://github.com/rust-lang/rustup/pull/4471#discussion_r2362669804) comments being left for a future PR -- the justification can be seen in the comments.

> I suspect some interaction with memory limits showing up in the unpack phase there; I suggest raising the working memory much higher, if you're on a machine that has capacity, to see if thats the case.
> 
> But also it suggests the default should probably be quite conservative, as we still need to support unpacking on Rasberry PI class machines.

First of all, thank you for your comment @rbtcollins. I have indeed reran the benchmarks with a higher memory and it didn't show any difference. For reference, it was a machine equipped with an Intel Xeon E5-2630 v2 processor and 64 GB of RAM.

Moreover, I would like to thank @djc for all the simplification work that he did, which made this PR much easier to implement. :smile:

src/cli/download_tracker.rs
FranciscoTGouveia: As discussed with @rami3l, I plan to redesign the progress bar styling so that they become column-aligned.

Additionally, I believe it would be useful to introduce another status update to inform the user when a download has completed and is waiting to be installed, rather than simply displaying a completed progress bar, which can be confusing.

With that in mind, this upcoming redesign will affect the highlighted code in this comment, making any changes at this stage redundant.
Since the redesign should be carefully discussed, I suggest leaving this for a future PR.

src/config.rs
```diff
@@ -238,13 +239,13 @@ pub(crate) struct Cfg<'a> {
...
     pub download_dir: PathBuf,
-    pub tmp_cx: temp::Context,
+    pub tmp_cx: Arc<temp::Context>,
     pub toolchain_override: Option<ResolvableToolchainName>,
     pub env_override: Option<LocalToolchainName>,
     pub dist_root_url: String,
     pub quiet: bool,
     pub current_dir: PathBuf,
-    pub process: &'a Process,
+    pub process: &'a Process, // TODO: this may need to be inside an Arc.
```
rami3l: Is this TODO finally addressed? Idem for other TODO comments in the diff.

src/dist/component/transaction.rs
```diff
@@ -9,8 +9,8 @@
 //! FIXME: This uses ensure_dir_exists in some places but rollback
 //! does not remove any dirs created by it.
 
-use std::fs::File;
 use std::path::{Path, PathBuf};
+use std::{fs::File, sync::Arc};
```
rami3l: Nit: Maybe just merge these imports into one?

src/dist/manifestation.rs
```diff
@@ -499,6 +474,131 @@ impl Manifestation {
...
+        dist_server: &str,
+        max_retries: usize,
+        altered: bool,
+    ) -> impl futures_util::Stream<Item = impl Future<Output = Result<String>>> {
+        tokio_stream::iter(components).map({
+            move |bin| {
+                let sem = semaphore.clone();
+                let download_tx = download_tx.clone();
+                let new_manifest = Arc::clone(&new_manifest);
+                let download_cfg = download_cfg.clone();
```
rami3l: Nit: We might as well use `Arc::clone()` for `download_cfg`/`dl_cfg` for the sake of consistency. This happens in a few other places as well IIRC. Suggest checking for the `remove lifetimes` commit in particular.

src/dist/manifestation.rs
```diff
@@ -419,7 +390,7 @@ impl Manifestation {
         &self,
         new_manifest: &[String],
         update_hash: Option<&Path>,
-        dl_cfg: &DownloadCfg<'_>,
+        dl_cfg: Arc<DownloadCfg>, // TODO: might need to be wrapped in an Arc.
```
rami3l: Nit: Redundant TODO.

rami3l (approved):
LGTM modulo some questions/nits :)

src/dist/manifestation.rs
rami3l: @djc I think `Self::foo()` that @FranciscoTGouveia came up might a bit surprising here. Maybe you meant using regular function-in-function declarations?

Idem for https://github.com/rust-lang/rustup/pull/4471#discussion_r2362662131 and https://github.com/rust-lang/rustup/pull/4471#discussion_r2362662465.

src/config.rs
FranciscoTGouveia: I am very sorry to have left this -- it was just some notes to self during the rebase. I apologize.

src/dist/manifestation.rs
djc: While `Arc::clone()` was popular a while ago as a method of demonstrating that it's just a cheap clone, I think much of the ecosystem has moved away from it in favor of just using `.clone()`, which is more concise.

src/dist/manifestation.rs
```diff
@@ -499,6 +472,50 @@ impl Manifestation {
 
         Ok(tx)
     }
+
+    fn install_component<'a>(
```
djc: This seems more strongly attached to `ComponentBinary` than to `Manifestation`, so would suggest attaching it to that type. Suggest ordering arguments from most specific to least specific, so `tx` probably comes after `installer_file` and `new_manifest` probably comes before `download_cfg`.

src/dist/manifestation.rs
djc: I really think the named functions should be introduced with this commit, rather than at the end.

djc (commented):
So while I did say I was fine with doing spawning/channel approach, after looking at the "remove some lifetime parameters" commit I am no longer sure if that's the best approach.

Effectively it means that whoever goes and puts in place the right architecture will basically have to undo all of those changes which IMO should not be necessary.

djc:
(It might be useful to start split some of the commits out of this PR, since it's obviously complex/contentious -- would be useful to start landing some parts to avoid more rebasing. At least the first commit is an obvious candidate.)

src/dist/manifestation.rs
rami3l: @djc I somehow wasn't well informed then, my apologies. That should be a trivial replacement for @FranciscoTGouveia right?

src/cli/download_tracker.rs
djc: Sure.

src/dist/manifestation.rs
FranciscoTGouveia: Is this change strictly related to the highlighted lines of code, or should I go ahead and update all occurrences?

djc: Change the new uses in these commits. If you want to do a sweep across the codebase that would also be nice IMO but should be in a separate PR.

djc:
I think the correcting faulty output commits are also independent of the other tricky parts?

FranciscoTGouveia:
> I think the correcting faulty output commits are also independent of the other tricky parts?

Yes, they are. Would it be okay if I opened a PR for that now to help streamline this one?

src/dist/manifestation.rs
FranciscoTGouveia: Noted. I will make sure that this PR does not introduce any `Arc::clone()`, and I will create another PR to sweep the database soon.

djc:
> Yes, they are. Would it be okay if I opened a PR for that now to help streamline this one?

Definitely!

src/dist/manifestation.rs
FranciscoTGouveia: I have checked the codebase for other instances of `Arc::clone()`, but there were none.

rami3l:
Closing as superseded by #4605. I'd still like to express my great appreciation for you and for all the work you were able to do in the concurrent rustup GSoC and even more. It really means a lot for us to prove what is possible with this project, and I'm glad that we have been able to push the project through the finish line together :)

PR #4473, opened 2025-09-07
fix(cli/rustup_mode): use ASCII-compatible spinner

rami3l:
Continuation of #4388.

When reviewing #4471 (cc @FranciscoTGouveia) I realized that the spinner should also be ASCII-compatible, i.e. my previous comment on the progress bar (https://github.com/rust-lang/rustup/pull/4426#issuecomment-3124288998) applies to spinners as well.

https://stackoverflow.com/a/12305221 proposes some options that meet this criterion:

- _The Bar_: `r"|/-\ "`
- _The Ballon_: `".oO@* "`

... and even if  I decided to go with _The Bar_ for the familiarity (personally I have been seeing that a lot on Windows since I first learnt to use a PC), but please feel free to propose other options.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -815,7 +815,7 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
             pb.set_style(
                 ProgressStyle::with_template("{msg:.bold} - Checking... {spinner:.green}")
                     .unwrap()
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Could also adapt to whether Unicode is available if we have a good way of detecting that.

rami3l:
> Could also adapt to whether Unicode is available if we have a good way of detecting that.

Yup, I guess we can figure that out in another PR if possible :)

PR #4474, opened 2025-09-07
feat: adjust default concurrent download count

rami3l:
I am currently test-driving the new concurrent downloads implementation and would like to try out different download counts. However when doing so I noticed that the previous decision of make both `rustup check` and `rustup toolchain` share a common default concurrent download count might not be optimal.

Thus, with this PR:

1. We check updates for all channels during `rustup check` unless the download count is overridden to 1.
2. We make the default download count 2 to hit a better balance between concurrent downloads and error recovery (this can be postponed to another PR if preferred).

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -802,7 +802,13 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
     let mut update_available = false;
     let channels = cfg.list_channels()?;
     let channels_len = channels.len();
-    let concurrent_downloads = cfg.process.concurrent_downloads().unwrap_or(channels_len);
+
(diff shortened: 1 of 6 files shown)


djc (approved):

src/cli/rustup_mode.rs
```diff
@@ -806,7 +806,7 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<utils::ExitCode
 
     // Ensure that `.buffered()` is never called with 0 as this will cause a hang.
     // See: https://github.com/rust-lang/futures-rs/pull/1194#discussion_r209501774
-    if num_channels > 0 {
```
rami3l, line 809: This will actually hang when `.concurrent_downloads()` returns non-zero but `channels_len` is zero. This PR happens to help manifest this regression (the invariant stated in the comment above has been broken since https://github.com/rust-lang/rustup/pull/4455, my apologies for not finding it out earlier).

PR #4481, opened 2025-09-09
reduce tracing lvl to debug for release build

klensy:
This cuts ~300kb for `x86_64-pc-windows-msvc` release build.

Similar to fix in https://github.com/rust-lang/rust/pull/146188 (not really, there wasn't `max_level_*` feature set), docs: https://docs.rs/tracing/latest/tracing/level_filters/index.html

Can be changed to `release_max_level_info`, but it don't decrease size much (at least on `x86_64-pc-windows-msvc`). This change disables `trace` level for tracing for release build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ tokio = { version = "1.26.0", default-features = false, features = ["macros", "r
 tokio-retry = "0.3.0"
 tokio-stream = "0.1.14"
 toml = "0.9"
-tracing = "0.1"
+tracing = { version = "0.1", features = ["max_level_trace", "release_max_level_debug"] }
 tracing-log = "0.2"
 tracing-opentelemetry = { version = "0.31", optional = true }
 tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }


djc:
I'm not sure I'm a fan of this. One of the attractions in the first place of getting tracing working well within rustup was being able to get detailed logs from people whose rustup installation is misbehaving, and this would make that harder. What percentage is 300 kB? Why does this matter to you?

klensy:
I've found 8 uses of `trace!`, most of them don't deserve be at trace level, IMHO (forgot about instrumenting, yes, my bad). And this is how `tracing` worked (it worked actually with `release_max_level_info`, not debug with current config) before bug was introduced. As for size - just free improvement.

djc:
> I've found 8 uses of `trace!`, most of them don't deserve be at trace level, IMHO (forgot about instrumenting, yes, my bad).

That's a different PR though -- open to reviewing that.

> And this is how `tracing` worked (it worked actually with `release_max_level_info`, not debug with current config) before bug was introduced. As for size - just free improvement.

So there was a bug before and it got fixed, so now it works as intended, and you want to change the intention because you liked the previous behavior better?

klensy:
Looking second time, rustup is unaffected, because it didn't configured tracing levels via features at all, so they was always at trace level. Should in that case just set `trace` for debug/release (to have some configuration, instead of relying on defaults)?

djc:
> Looking second time, rustup is unaffected, because it didn't configured tracing levels via features at all, so they was always at trace level. Should in that case just set `trace` for debug/release (to have some configuration, instead of relying on defaults)?

I'm comfortable relying on this particular default, and don't think there's anything to do here.

rami3l:
@djc I second your opinion; I am still not convinced that the potential benefits of this change are worthwhile for us to push it forward...

rami3l:
@klensy I'm closing this issue for now, sorry about that 🙏

PR #4487, opened 2025-09-17
Fix link in the bug reporting template

LingMan:
The existing link added the search parameter `is:open,closed` which results in no issues being found at all. Simply leaving it out will show all results, open and closed, which was most likely the intention here.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -12,7 +12,7 @@ body:
       label: Verification
       description: Please verify that you've followed these steps.
       options:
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

rami3l:
> The existing link added the search parameter `is:open,closed` which results in no issues being found at all. Simply leaving it out will show all results, open and closed, which was most likely the intention here.

@LingMan Thanks for this PR!

`is:open,closed` was specifically there to work around a GitHub bug where leaving it out will only show the open issues bot not the closed ones. It seems like the original bug has been fixed but a new regression has been introduced? 🤔

PR #4489, opened 2025-09-21
Simplify component downloads

djc:
From reviewing

- #4471 

cc @FranciscoTGouveia

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -11,6 +11,7 @@ use futures_util::stream::StreamExt;
 use std::sync::Arc;
 use tokio::sync::Semaphore;
 use tracing::info;
+use url::Url;
 
 use crate::dist::component::{
     Components, Package, TarGzPackage, TarXzPackage, TarZStdPackage, Transaction,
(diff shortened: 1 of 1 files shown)


src/dist/manifestation.rs
```diff
@@ -173,44 +174,44 @@ impl Manifestation {
...
-                        tmp_cx,
-                        download_cfg,
-                        max_retries,
-                        new_manifest,
-                    )
+        let component_stream = tokio_stream::iter(components.into_iter()).map(|bin| {
+            let sem = semaphore.clone();
+            async move {
+                let _permit = sem.acquire().await.unwrap();
+                let url = if altered {
```
rami3l, line 191: Nit: Consider moving this `utils::parse_url()` outside of `if-else`?

rami3l (approved):
That's much clearer indeed, thanks :)

src/dist/manifestation.rs
djc: I thought about that but it doesn't quite work. The first arm is a `String`, and we can't yield a reference to a locally owned `String`. `parse_url()` takes a reference anyway and returns an owned `Url`, so cloning `bin.binary.url` feels like a waste.

rami3l:
@djc I think GitHub CI is experiencing some network problems right now. Maybe try merging later?

PR #4496, opened 2025-09-22
Flatten notification enums

djc:
We've had 4 different `Notification` enums:

- `dist::notifications::Notification`
- `dist::temp::Notification`
- `utils::notifications::Notification`
- `notifications::Notification`

While these carried somewhat different variants (although there was some overlap as well!), the complexity of having these separate types does not seem worth it. (And in fact, the existence of the `Notification` type itself doesn't seem all that valuable.) Merge all the variants into the top-level type.


src/notifications.rs
```diff
@@ -92,7 +134,97 @@ impl Display for Notification<'_> {
...
+            FileAlreadyDownloaded => write!(f, "reusing previously downloaded file"),
+            CachedFileChecksumFailed => write!(f, "bad checksum for cached download"),
+            RollingBack => write!(f, "rolling back changes"),
+            ExtensionNotInstalled(c) => write!(f, "extension '{c}' was not installed"),
+            NonFatalError(e) => write!(f, "{e}"),
+            MissingInstalledComponent(c) => {
+                write!(f, "during uninstall component {c} was not found")
+            }
+            DownloadingComponent(c, h, t, _) => {
+                if Some(h) == t.as_ref() || t.is_none() {
```
rami3l, line 191: Pre-existing, but now that we are at it, how about using one extra commit to address https://github.com/rust-lang/rustup/pull/4471#discussion_r2362667381?

src/notifications.rs
```diff
@@ -40,29 +95,58 @@ pub(crate) enum Notification<'a> {
...
-}
-
 impl Notification<'_> {
     pub(crate) fn level(&self) -> NotificationLevel {
         use self::Notification::*;
         match self {
-            Install(n) => n.level(),
-            Utils(n) => n.level(),
-            Temp(n) => n.level(),
+            ChecksumValid(_)
```
rami3l, line 102: The original code (that has been deleted in this PR) seems to be grouping this by notification level first and then by order of variants. Should we do the same with the merged version?

rami3l (approved):
Many thanks! It will definitely make #4471 even clearer for @FranciscoTGouveia, so looking forward to landing it :)

src/notifications.rs
djc: I intend to delete most of `Notification` in a next PR, so I don't think there's a lot of value in style tweaks.

src/notifications.rs
djc: Done.

Issue #4421, opened 2025-07-21
'rustup' detects inappropriate default host and incorrectly handles full paths set in *_HOME* environment variables


Petr-Lazecky:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue+is%3Aopen%2Cclosed and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

When 'rustup' is run within Cygwin environment on the Windows, it detects `x86_64-pc-windows-msvc` as default host. One would expect `x86_64-pc-windows-gnu` on Cygwin where gcc is typical primary toolchain. This can be handled by passing `--default-host` to install script so it is really minor problem but IMHO a little bit of inconvenience, depending on how one looks into it.

Also...

When one tries to pass CARGO_HOME and RUSTUP_HOME variables to control where rust should be installed within Cygwin environment on Windows *and* user sets absolute paths for these variables, installation script "mangles"   these paths by prefixing them with "C:/" making them invalid (setting up rust somewhere "outside" of the Cygwin environment, typically in the C:\ root which is not what user really wants.

### Issue 1:

1. Environment: Windows 10, Cygwin environment 
2. Launch Cygwin environment
3. Execute standard Linux command to install rustup:
     
    `curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh -s`

**Observed:**

    `default host triple: x86_64-pc-windows-msvc`

**Expected:**

    `default host triple: x86_64-pc-windows-gnu`

### Issue 2:

1. Environment: Windows 10, Cygwin environment 
2. Launch Cygwin environment
3. Execute command by passing CARGO_HOME and RUSTUP_HOME environment variables referencing full paths within Cygwin environment where rust should be installed:

`CARGO_HOME="/home/user1/Rxes/rust/cargo" RUSTUP_HOME="/home/user1/Rxes/rust/env" bash -c "curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh -s -- --default-host x86_64-pc-windows-gnu"`

**Observed:**

Instalation message contains this:

```
The Cargo home directory is located at:
  C:/home/user1/Rxes/rust/cargo
...
The cargo, rustc, rustup and other commands will be added to
Cargo's bin directory, located at:
  C:/home/user1/Rxes/rust/cargo/bin
```

**Expected:**

[...]

PR #4497, opened 2025-09-22
fix(install): default to GNU host in Cygwin/MSYS/MinGW environments (#4421)

cachebag:
Addresses first mentioned issue in #4421, modifying the install script to default to `x86_64-pc-windows-gnu` under Cygwin/MSYS when no --default-host is provided. 

The second issue requires changes in the compiled rustup-init binary (src/), which is out of scope for now

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -80,6 +80,13 @@ main() {
     local _arch="$RETVAL"
     assert_nz "$_arch" "arch"
 
+    local _default_host_override=""
+    case "$_arch" in
+        *windows*)
+            _default_host_override="--default-host=$_arch"
+            ;;
+    esac
+
     local _ext=""
     case "$_arch" in
         *windows*)
@@ -177,9 +184,9 @@ main() {
             exit 1;
         fi
 
-        ignore "$_file" "$@" < /dev/tty
+        ignore "$_file" ${_default_host_override:+"$_default_host_override"} "$@" < /dev/tty
     else
-        ignore "$_file" "$@"
+        ignore "$_file" ${_default_host_override:+"$_default_host_override"} "$@"
     fi
 
     local _retval=$?


rustup-init.sh
```diff
@@ -179,6 +179,21 @@ main() {
...
+        case "$(uname -s)" in 
+            CYGWIN*|MSYS*|MINGW*)
+                _has_default_host_arg=false
+                for arg in "$@"; do 
+                    case "$arg" in 
+                        --default-host*) _has_default_host_arg=true ;;
+                    esac 
+                done 
+                if [ "$_has_default_host_arg" = false ]; then 
+                    set -- "$@" --default-host x86_64-pc-windows-gnu
```
rami3l: Should we consider the possibility of [`aarch64-pc-windows-gnullvm`](https://doc.rust-lang.org/nightly/rustc/platform-support/windows-gnullvm.html) as well? It's a tier 2 target for now.

cc @ChrisDenton

rami3l (commented):
Thank you very much :)

However, 4221 is not an issue. Maybe you've got the number wrong?

rustup-init.sh
```diff
@@ -179,6 +179,21 @@ main() {
 
         ignore "$_file" "$@" < /dev/tty
     else
+        # Default to GNU toolchain under Cygwin/MSYS/MinGW environments
+        case "$(uname -s)" in
```
rami3l: You should move whatever default host detection logic you may have into the `get_architecture()` function, otherwise it might break the `RUSTUP_INIT_SH_PRINT=arch` functionality.

cachebag:
> However, 4221 is not an issue. Maybe you've got the number wrong?

@rami3l 
Thanks for letting me know! Lmk what else may be needed

rustup-init.sh
ChrisDenton: Hm... for arm we should probably not touch the default for now, at least until support for gnu on the target is more mainstream. The gnullvm target was only fairly recently promoted to having host tooling and iirc requires installing llvm tools (instead of gnu ones). We could revisit this in the future.

rami3l (commented):
How is this different from the existing functionality below?

https://github.com/rust-lang/rustup/blob/7365ab767a586c13f4c48e89ef18664c0705b37b/rustup-init.sh#L425-L427

cachebag:
@rami3l 

> How is this different from the existing functionality below?

I presume that it's essentially the safety net to this, no? The idea here is that the issue at hand was detecting `x86_64-pc-windows-msvc` instead of `x86_64-pc-windows-gnu` in Cygwin despite the existing detection. I figured that a blanket override ensures that shell script users on Windows get GNU regardless of detection edge cases.

Do we address/refactor the existing detection instead?

rami3l:
> > How is this different from the existing functionality below?
> 
> I presume that it's essentially the safety net to this, no? The idea here is that the issue at hand was detecting `x86_64-pc-windows-msvc` instead of `x86_64-pc-windows-gnu` in Cygwin despite the existing detection. I figured that a blanket override ensures that shell script users on Windows get GNU regardless of detection edge cases.
> 
> Do we address/refactor the existing detection instead?

Yes, if this PR is made to address https://github.com/rust-lang/rustup/issues/4421#issuecomment-3316574199. I don't have a Cygwin test environment, but the detection seems right already to me, so this PR seems unnecessary in its current state.

cc @ChrisDenton for further verification.

cachebag:
@rami3l
After looking at this with fresher eyes, I see that I shouldn't simply try to force GNU after detection.
When on Windows, the script now sets `_default_host="--default-host $_arch"` and checks if the user already passed `--default-host` (if so, it clears it to respect user preference). I believe this is what @ChrisDenton was [saying here ](https://github.com/rust-lang/rustup/issues/4421#issuecomment-3316574199)and I apologize for misunderstanding.

> Hm, I would have to look into this more before answering that definitively but I think we can address this by the shell script passing --default-host to rustup (assuming it's not set by the user) and not need any bigger changes.

rustup-init.sh
```diff
@@ -80,6 +80,21 @@ main() {
     local _arch="$RETVAL"
     assert_nz "$_arch" "arch"
 
+    local _default_host=""
+    case "$_arch" in
+        *windows*)
+            _default_host="--default-host $_arch"
```
rami3l: I think this is okay, but the variable name here is a bit confusing. Maybe this variable can be renamed to `_default_host_override` to emphasize the fact that it's really an override.

rami3l: I think another trick can be used to make this simpler: just call this `--default-host=$_arch` so you can avoid `SC2086`. I promise that the host name doesn't contain whitespaces or whatsoever.

rustup-init.sh
```diff
@@ -80,6 +80,21 @@ main() {
...
+            for arg in "$@"; do
```
rami3l: This is not necessary either, IIRC according to the GNU coreutils convention, if we pass `--default-host=A --default-host=B`, then only `B` will be recognized. So just putting the `$_default_host_override` before the rest of the args should be fine.

rami3l (changes requested):

rami3l (commented):
@cachebag I'm sorry for the long delay, but your current version is much cleaner now, thanks a lot!

rustup-init.sh
rami3l: @cachebag Did you forget about this one, or was I analyzing this wrong?

cachebag: Latest commit should reflect I was just trying to figure out how to quote `_default_host_override` without passing an empty string @rami3l

rami3l (approved):
@cachebag I think this code looks okay to merge, but just to be sure, has it been tested to work though?

cachebag:
> @cachebag I think this code looks okay to merge, but just to be sure, has it been tested to work though?
@rami3l 

I unfortunately don't run Windows on any devices. I can ping the original author of the issue to try these changes out or quickly install a VM to test this, I think that should work.

rami3l:
@Petr-Lazecky Would you mind testing out this new installation script on Windows to see if the host detection has changed as expected?

cc @ChrisDenton

cachebag:
@rami3l 

Was able to test this on one of my clients' machines. 

image

rami3l:
@cachebag Looks cool! Merging...

PR #4500, opened 2025-09-24
tests: use DistContext for dist::component tests

djc:


rami3l (approved):
These changes are very clear and simple to review, thanks!

PS: I'm requesting an additional review from Copilot in case it can find some accidental typos... But otherwise everything looks great from my side :)

djc:
> These changes are very clear and simple to review, thanks!
> 
> PS: I'm requesting an additional review from Copilot in case it can find some accidental typos... But otherwise everything looks great from my side :)

I don't agree with Copilot's nit picks, so will just merge this!

PR #4499, opened 2025-09-23
Start removing Notification variants

djc:

PR #4501, opened 2025-09-26
Remove more Notification variants

djc:
Continuing the work started in

- #4499

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -122,13 +122,7 @@ impl Manifestation {
 
         // Create the lists of components needed for installation
         let config = self.read_config()?;
-        let mut update = Update::build_update(
-            self,
-            new_manifest,
-            &changes,
-            &config,
-            &download_cfg.notify_handler,
-        )?;
+        let mut update = Update::build_update(self, new_manifest, &changes, &config)?;
 
         if update.nothing_changes() {
             return Ok(UpdateStatus::Unchanged);
@@ -539,7 +533,6 @@ impl Update {
         new_manifest: &Manifest,
         changes: &Changes,
         config: &Option<Config>,
-        notify_handler: &dyn Fn(Notification<'_>),
     ) -> Result<Self> {
         // The package to install.
         let rust_package = new_manifest.get_package("rust")?;
@@ -601,9 +594,17 @@ impl Update {
                 if !starting_list.contains(component) {
                     result.components_to_install.push(component.clone());
                 } else if changes.explicit_add_components.contains(component) {
-                    notify_handler(Notification::ComponentAlreadyInstalled(
-                        &component.description(new_manifest),
-                    ));
+                    match &component.target {
+                        Some(t) if t != &manifestation.target_triple => info!(
+                            "component {} for target {} is up to date",
+                            component.short_name(new_manifest),
+                            t,
+                        ),
+                        _ => info!(
+                            "component {} is up to date",
+                            component.short_name(new_manifest)
+                        ),
+                    }
                 }
             }
         } else {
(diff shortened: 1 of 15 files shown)


tests/suite/cli_v2.rs
```diff
@@ -1630,7 +1630,7 @@ async fn add_target_again() {
         .expect(["rustup", "target", "add", CROSS_ARCH1])
         .await
         .with_stderr(snapbox::str![[r#"
-info: component 'rust-std' for target '[CROSS_ARCH_I]' is up to date
+info: component is up to date component=rust-std target=[CROSS_ARCH_I]
```
rami3l: Are you sure about this? Message changes like this are making the readability worse, thus this might be a UX regression when it comes to `info!()`, `warn!()` and `error!()`...

rami3l (dismissed):

tests/suite/cli_v2.rs
rami3l: Or, are you moving towards the machine-readable output this way? If that's the case I'm afraid we need to update our user-facing message format:

https://github.com/rust-lang/rustup/blob/7365ab767a586c13f4c48e89ef18664c0705b37b/src/cli/log.rs#L112

... to make it clearer, say "info: component is up to date **for** component=rust-std target=[CROSS_ARCH_I]" to stay closer to the original vibe.

djc: Yeah, I was hoping to make this more like structured logging. However, I agree that this is a bit of a UX regression, so I'll revert some of the message changes to be more human-friendly.

rami3l: @djc Now that you've brought this up, I think there are actually two ways of doing https://github.com/rust-lang/rustup/issues/450. The first is giving the notification system an alternative output format (corrosion maintainer specifically asked for JSON for better CMake compatibility, so the resulting output would be JSONL), the second would be returning a specifically-designed single JSON value.

If we are moving on with the first option then your work of extracting field names is actually valuable (as I can then further abuse `tracing_subscriber` to output JSON directly), so that is why I proposed the compromise above :)

What do you think about this?

cc @ChrisDenton

ChrisDenton: I find the first option appealing as it seems much more generally useful and I think it's more in line with what tools such as cargo provide. I'm less sure about hacking around `tracing_subscriber` to provide it but I'd want a better idea of what that would look like. Maybe it's ok.

rami3l: @chrisdenton Back in the v1.28 release cycle I've reproduced the regular rustup log messages with `tracing_subscriber` while retaining the previous format, and almost everyone should be using it now without major issues (some aren't even aware of the fact?). 

My option 1 is just a natural extension of that idea.

Of course, special care must be taken to handle the direct prints in say `rustup show`, but I consider that to be a minority.

djc: I don't really want to increase the scope of this PR beyond what it already does, so I think I'll just revert to more human-readable messages. IMO the UX is definitely more important than a currently unimplemented potential future machine-readable mechanism.

rami3l (approved):
LGTM! I'm still not sure if anyone is relying on parsing our output though since the format has changed, maybe we have to go through a full release to know that.

src/dist/manifestation.rs
```diff
@@ -463,12 +479,7 @@ impl Manifestation {
         let (installer_file, installer_hash) = dl.unwrap();
 
         let prefix = self.installation.prefix();
-
-        notify_handler(Notification::InstallingComponent(
-            "rust",
-            &self.target_triple,
-            Some(&self.target_triple),
-        ));
+        info!(component = "rust", "installing component");
```
rami3l: This will be covered by #4471, not sure if it should be added back in that PR during rebase.

cc @FranciscoTGouveia

PR #4507, opened 2025-09-30
refactor: Switch logging to anstyle

epage:
This removes a use of `termcolor` where it isn't offering much value and instead uses `anstyle` for styling which is already a dependency through clap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2203,6 +2203,7 @@ dependencies = [
 name = "rustup"
 version = "1.28.2"
 dependencies = [
+ "anstyle",
  "anyhow",
  "cc",
  "cfg-if 1.0.3",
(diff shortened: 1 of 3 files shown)


djc (approved):
Looks okay.

Unless they're still ambiguous, would prefer to have `Style` and `Color` imported rather than qualified.

epage:
> Unless they're still ambiguous, would prefer to have Style and Color imported rather than qualified.

Habit. I've updated it to match the local style

djc:
Do you also have stuff in the clap stack that can replace the remaining use of termcolor?

https://github.com/rust-lang/rustup/blob/master/src/process/terminalsource.rs#L13

epage:
Yes, my goal is to remove termcolor and allow clap to be colored. These were just smaller steps on the way.

rami3l (approved):

PR #4510, opened 2025-10-02
Remove more Notification variants (part 3)

djc:
Previously:

- #4501 
- #4499


rami3l (approved):
Thank you so much for the effort to simplify all this :)

PR #4513, opened 2025-10-03
dist: bump `rustup` version to v1.29.0

rami3l:
This is the 1st PR for the `1.29.0` beta release according to our [release process](https://rust-lang.github.io/rustup/dev-guide/release-process.html#making-a-release).

### Checklist

- [x] Bump version in the manifest.
- [x] Update the changelog.
- [ ] ~~Draft the release blog post.~~ (Not required, since this is a beta release.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,256 @@
 # Changelog
 
+## [1.29.0] - Unreleased
+
+This new release of rustup comes with significant changes.
+
+The headline feature is better concurrency for certain operations,
+implemented by @FranciscoTGouveia with the help from the team as a part of
+the GSoC 2025 event:
+
+- Improved the UX for the `rustup update` and `rustup toolchain` subcommands by
+  downloading components concurrently and unpacking during downloads, which
+  should speed up the toolchain installation process. Meanwhile, progress bars
+  have been updated accordingly to and provide better feedback about where time
+  is being spent. [pr#4426] [pr#4436] [pr#4471] [pr#4605]
+
+  - You can use the `RUSTUP_CONCURRENT_DOWNLOADS` environment variable to
+    adjust the number of concurrent downloads to better reflect the needs of
+    your specific environment (the current default is `2`). Particularly, you
+    may set the value to `1` to restore the sequential downloading behavior.
+    [pr#4455]
+
+- Improved the UX for the `rustup check` subcommand by checking for toolchain
+  updates concurrently and updating the progress reporting UI. [pr#4388]
+
+rustup now officially supports the following host platforms:
+
+- `sparcv9-sun-solaris` [pr#4380]
+- `x86_64-pc-solaris` [pr#4380]
+
+In addition, basic support for more shells has been added, including:
+
+- `tcsh` [pr#4459]
+- `xonsh` [pr#4626]
+
+This release also comes with other quality-of-life improvements, to name a few:
+
+- When running rust-analyzer via a proxy, rustup will consider the
+  `rust-analyzer` binary from `PATH` when the rustup-managed one is not found.
+  [pr#4324]
+
+- Empty environment variables are now treated as unset. This should help with
+  resetting configuration values to default when an override is present.
+  [pr#4422]
+
+[1.29.0]: https://github.com/rust-lang/rustup/releases/tag/1.29.0
+[pr#4324]: https://github.com/rust-lang/rustup/pull/4324
+[pr#4380]: https://github.com/rust-lang/rustup/pull/4380
+[pr#4388]: https://github.com/rust-lang/rustup/pull/4388
+[pr#4422]: https://github.com/rust-lang/rustup/pull/4422
+[pr#4426]: https://github.com/rust-lang/rustup/pull/4426
+[pr#4436]: https://github.com/rust-lang/rustup/pull/4436
+[pr#4455]: https://github.com/rust-lang/rustup/pull/4455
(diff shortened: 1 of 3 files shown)


rami3l:
PS: I understand that this version is not final, but any suggestion regarding the wording/phrasing/format of the changelog is greatly appreciated :)

CHANGELOG.md
```diff
@@ -1,5 +1,177 @@
...
+- `rustup update` and `rustup toolchain` subcommands will download toolchain
+  components concurrently. [pr#4436] 
+
+  - You can use the `RUSTUP_CONCURRENT_DOWNLOADS`
+    environment variable to adjust the number of concurrent downloads to better
+    reflect the needs of your specific environment. [pr#4455]
+
+- `rustup` now uses `indicatif`-based progress bars for toolchain update
+  checks and downloads. This provides not only a more async-friendly UI solution,
+  but also a more consistent and visually-appealing experience for various
```
djc: Nit: I think "visually appealing" would usually be spelled as 2 words.

CHANGELOG.md
```diff
@@ -1,5 +1,177 @@
...
+
+The headline feature is better concurrency for certain operations,
+implemented by @FranciscoTGouveia within the scope of the GSoC 2025 event:
+
+- `rustup check` will check for toolchain updates concurrently. [pr#4388]
+
+- `rustup update` and `rustup toolchain` subcommands will download toolchain
+  components concurrently. [pr#4436] 
+
+  - You can use the `RUSTUP_CONCURRENT_DOWNLOADS`
```
djc: Is this discoverable in some other way? If not, how do we make it discoverable?

rami3l: @djc I think we must achieve a careful balance between not spamming the users and improving discoverability. I'm not sure about other potential solutions, but printing a hint every time the download is activated seems like a no-go to me (meaning myself as a user would not want that, and other tools like APT and Docker Compose seemingly are giving no similar hints).

Idem for other options regarding the download.

rami3l: @djc Some applications show random tips for such cases (I'm thinking of certain video games), maybe we can apply the same techniques somehow and  allow it to be hidden at the same time?

The rationale is that these tips are not essential to 99% of the OOTB experience of our app; if users are interested in tuning for their specific environment then after a certain period of time they will know enough information about it.

djc: Pff, not sure I'd be a fan of random tips. I'm thinking more of relevant error messages, the `help` stuff for relevant CLI commands and maybe other places in the documentation.

rami3l: Made #4554.

rami3l:
@ChrisDenton Would you mind having a look? Many thanks in advance :)

ChrisDenton (approved):
Wording looks good to me.

CHANGELOG.md
```diff
@@ -1,5 +1,261 @@
...
+the GSoC 2025 event:
+
+- `rustup check` will check for toolchain updates concurrently. [pr#4388]
+
+- `rustup update` and `rustup toolchain` subcommands will download toolchain
+  components concurrently. [pr#4436] 
+
+  - You can use the `RUSTUP_CONCURRENT_DOWNLOADS` environment variable to
+    adjust the number of concurrent downloads to better reflect the needs of
+    your specific environment. Particularly, you may set the value to `1` to
```
djc: Suggest we mention the default value here?

CHANGELOG.md
```diff
@@ -1,5 +1,261 @@
...
+
+This new release of rustup comes with significant changes.
+
+The headline feature is better concurrency for certain operations,
+implemented by @FranciscoTGouveia with the help from the team as a part of
+the GSoC 2025 event:
+
+- `rustup check` will check for toolchain updates concurrently. [pr#4388]
+
+- `rustup update` and `rustup toolchain` subcommands will download toolchain
```
djc: I feel like this is more important/a bigger win for most usage than the `rustup check` changes, and so merits moving up?

CHANGELOG.md
```diff
@@ -1,5 +1,261 @@
...
+    restore the sequential downloading behavior. [pr#4455]
+
+- `rustup update` and `rustup toolchain` subcommands will unpack components
+  of a toolchain while others are being downloaded. [pr#4471] [pr#4605]
+
+- `rustup` now uses `indicatif`-based progress bars for toolchain update
+  checks and downloads. This provides not only a more async-friendly UI solution,
+  but also a more consistent and visually appealing experience for various
+  commands like `rustup check`, `rustup update`, and `rustup toolchain install`.
+  [pr#4388] [pr#4426]
```
djc: It feels like these are really all part of the same improvement in experience, and I don't think it makes sense to have separate bullet points from the user's point of view.

rami3l: @djc Just to be sure, are you suggesting using a short paragraph like in https://brew.sh/2025/11/12/homebrew-5.0.0 to cover all of the GSoC project, or just omitting this particular bullet point?

djc: I'm suggesting that "download toolchain components concurrently" (including adjustment), "unpack components [..] while [..] being downloaded" and "used [..] progress bars" feel like they're conceptually part of the same feature, and should be grouped as a single paragraph. Something like:

> - Improved the UX for `rustup update` and `rustup toolchain` subcommands to implement concurrency and use progress bars to reflect updates. This should speed up the process and provide better feedback about where time is being spent.

djc (approved):

PR #4514, opened 2025-10-03
fix(process): Ensure stdout/stderr lock is held across calls 

epage:
While `std::io::Write` only has a few required methods,
while implementing `anstream` I found its better to implement
as many as possible when dealing with `stdout` or `stderr`
so that the implicitly acquired lock is aquired once for the call,
rather than multiple times as the inherent methods break down a single higher
level calls into multiple lower level calls.

This is likely not a problem right now but the different layers of APIs, especially in interacting with `tracing`, could hit problems with this in the future.

### DIFF
--- a/src/process/terminalsource.rs
+++ b/src/process/terminalsource.rs
@@ -69,6 +69,16 @@ enum TerminalInner {
     TestWriter(TestWriter, ColorChoice),
 }
 
+impl TerminalInner {
+    fn as_write(&mut self) -> &mut dyn io::Write {
+        match self {
+            TerminalInner::StandardStream(s, _) => s,
+            #[cfg(feature = "test")]
+            TerminalInner::TestWriter(w, _) => w,
+        }
+    }
+}
+
 pub struct ColorableTerminalLocked {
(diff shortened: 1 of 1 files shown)


src/process/terminalsource.rs
```diff
@@ -194,6 +194,26 @@ impl ColorableTerminal {
     }
 }
 
+impl TerminalInner {
```
djc: Nit: ordering in this module is a bit of a mess, but please keep impls directly below their respective type's definitions.

djc (approved):
Makes sense, thanks!

src/process/terminalsource.rs
epage: Was trying to be consistent with the file which seemed to be types before impls (mostly). But fine with changing this.

djc: Yes, I think that would be nice. I'll clean up the other items after this is merged.

(Note that CI is failing on a formatting issue.)

rami3l (approved):
Many thanks! Just address whatever suggestions @djc may have and I think this PR is good to go 🙏

PR #4517, opened 2025-10-06
refactor: Simplify abstracted stdout/stderr

epage:
To write colored output on a locked stream, you had to:
1. Call `process.stdout()` to get a `Box<dyn Writer>`
2. Call `Writer::terminal(process)` to get a `ColorableTerminal` (enum)
3. Call `ColorableTerminal::lock` to get a `ColorableTerminalLocked` (enum)

This cuts out the first step, making sure everything goes through `ColorableTerminal`.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -148,19 +148,25 @@ impl Process {
         }
     }
 
-    pub(crate) fn stdout(&self) -> Box<dyn filesource::Writer> {
+    pub(crate) fn stdout(&self) -> terminalsource::ColorableTerminal {
         match self {
-            Process::OsProcess(_) => Box::new(io::stdout()),
+            Process::OsProcess(_) => terminalsource::ColorableTerminal::stdout(self),
             #[cfg(feature = "test")]
-            Process::TestProcess(p) => Box::new(filesource::TestWriter(p.stdout.clone())),
+            Process::TestProcess(p) => terminalsource::ColorableTerminal::test(
+                filesource::TestWriter(p.stdout.clone()),
+                self,
+            ),
         }
     }
 
-    pub(crate) fn stderr(&self) -> Box<dyn filesource::Writer> {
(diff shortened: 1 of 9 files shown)


src/process.rs
```diff
@@ -148,19 +148,25 @@ impl Process {
         }
     }
 
-    pub(crate) fn stdout(&self) -> Box<dyn filesource::Writer> {
+    pub(crate) fn stdout(&self) -> terminalsource::ColorableTerminal {
```
djc, line 151: Nit: still preferring importing instead of qualifying.

djc (approved):
Great stuff!

src/process/filesource.rs
```diff
@@ -174,7 +174,7 @@ mod test_support {
         }
 
         fn terminal(&self, process: &Process) -> ColorableTerminal {
-            ColorableTerminal::new(StreamSelector::TestWriter(self.clone()), process)
+            ColorableTerminal::test(self, process)
```
rami3l: Question: is it true that this line has been changed here first and a second time in the following commit? Would that mean at least one of the commits are not compiling, thus violating the rules of atomic commits?

src/process.rs
epage: I'm maintaning existing style within the code. If you want contributors to contribute in a different style, please change the style of the code.

src/process/filesource.rs
epage: You're right; I forgot to test my history edits.

epage: Fixed

rami3l: @epage Sorry, but I don't see the fix being applied here? I think `.clone()` is needed from the start if I'm not mistaken. Maybe you are not checking with the `test` feature on?

epage: Huh, somehow I forgot to push.  Sorry about that.

rami3l (approved):
Thanks a lot for this patch :)

PR #4518, opened 2025-10-06
Set `RUSTUP_TOOLCHAIN_SOURCE`

smoelius:
This PR is in pursuit of https://github.com/rust-lang/cargo/issues/11036. Specifically, it sets an environment variable `RUST_TOOLCHAIN_SOURCE`. The variable's content will be used to issue a warning when a `cargo install` is performed with a non-default toolchain.

The variable's content is one of `cli`, `env`, `override`, `config`, or `default`.

Nits are welcome.

cc: @epage

### DIFF
--- a/doc/user-guide/src/environment-variables.md
+++ b/doc/user-guide/src/environment-variables.md
@@ -72,6 +72,8 @@
 - `RUSTUP_CONCURRENT_DOWNLOADS` *unstable* (default: the number of components to download). Controls the number of
   downloads made concurrently.
 
+- `RUSTUP_TOOLCHAIN_SOURCE` *unstable*. Set by rustup to tell proxied tools how `RUSTUP_TOOLCHAIN` was determined.
+
 [directive syntax]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
 [dc]: https://docs.docker.com/storage/storagedriver/overlayfs-driver/#modifying-files-or-directories
 [override]: overrides.md
--- a/src/cli/proxy_mode.rs
+++ b/src/cli/proxy_mode.rs
@@ -5,6 +5,7 @@ use anyhow::Result;
 use crate::{
     cli::{common::set_globals, job, self_update},
     command::run_command_for_dir,
+    config::ActiveSource,
     process::Process,
     toolchain::ResolvableLocalToolchainName,
 };
@@ -24,6 +25,7 @@ pub async fn main(arg0: &str, current_dir: PathBuf, process: &Process) -> Result
         .filter(|arg| arg.starts_with('+'))
         .map(|name| ResolvableLocalToolchainName::try_from(&name.as_ref()[1..]))
         .transpose()?;
+    let toolchain_specified = toolchain.is_some();
 
     // Build command args now while we know whether or not to skip arg 1.
     let cmd_args: Vec<_> = process
@@ -32,9 +34,15 @@ pub async fn main(arg0: &str, current_dir: PathBuf, process: &Process) -> Result
         .collect();
 
     let cfg = set_globals(current_dir, true, process)?;
-    let cmd = cfg
-        .resolve_local_toolchain(toolchain)
-        .await?
-        .command(arg0)?;
+    let toolchain = cfg.resolve_local_toolchain(toolchain).await?;
+    let mut cmd = toolchain.command(arg0)?;
+    if toolchain_specified {
+        cmd.env(
+            "RUSTUP_TOOLCHAIN_SOURCE",
+            ActiveSource::CommandLine.to_string(),
+        );
+    } else if let Ok(Some((_, source))) = cfg.active_toolchain() {
+        cmd.env("RUSTUP_TOOLCHAIN_SOURCE", source.to_string());
+    }
     run_command_for_dir(cmd, arg0, &cmd_args)
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,4 @@
-use std::fmt::{self, Debug, Display};
+use std::fmt::{self, Debug};
 use std::io;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -102,14 +102,16 @@ pub(crate) enum ActiveSource {
     ToolchainFile(PathBuf),
 }
 
-impl Display for ActiveSource {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::result::Result<(), fmt::Error> {
+impl ActiveSource {
+    pub fn to_reason(&self) -> String {
         match self {
-            Self::Default => write!(f, "it's the default toolchain"),
-            Self::Environment => write!(f, "overridden by environment variable RUSTUP_TOOLCHAIN"),
-            Self::CommandLine => write!(f, "overridden by +toolchain on the command line"),
-            Self::OverrideDB(path) => write!(f, "directory override for '{}'", path.display()),
-            Self::ToolchainFile(path) => write!(f, "overridden by '{}'", path.display()),
+            Self::Default => String::from("it's the default toolchain"),
+            Self::Environment => {
+                String::from("overridden by environment variable RUSTUP_TOOLCHAIN")
+            }
+            Self::CommandLine => String::from("overridden by +toolchain on the command line"),
+            Self::OverrideDB(path) => format!("directory override for '{}'", path.display()),
+            Self::ToolchainFile(path) => format!("overridden by '{}'", path.display()),
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -92,9 +92,9 @@ impl<T: Into<String>> From<T> for OverrideFile {
     }
 }
 
-// Represents the reason why the active toolchain is active.
+// Represents the source that determined the current active toolchain.
 #[derive(Debug)]
-pub(crate) enum ActiveReason {
+pub(crate) enum ActiveSource {
     Default,
     Environment,
     CommandLine,
@@ -102,7 +102,7 @@ pub(crate) enum ActiveReason {
     ToolchainFile(PathBuf),
 }
 
-impl Display for ActiveReason {
+impl Display for ActiveSource {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::result::Result<(), fmt::Error> {
         match self {
             Self::Default => write!(f, "it's the default toolchain"),
@@ -505,7 +505,7 @@ impl<'a> Cfg<'a> {
     pub(crate) async fn maybe_ensure_active_toolchain(
         &self,
         force_ensure: Option<bool>,
-    ) -> Result<Option<(LocalToolchainName, ActiveReason)>> {
+    ) -> Result<Option<(LocalToolchainName, ActiveSource)>> {
         let should_ensure = if let Some(force) = force_ensure {
             force
         } else {
@@ -524,23 +524,23 @@ impl<'a> Cfg<'a> {
         }
     }
 
-    pub(crate) fn active_toolchain(&self) -> Result<Option<(LocalToolchainName, ActiveReason)>> {
+    pub(crate) fn active_toolchain(&self) -> Result<Option<(LocalToolchainName, ActiveSource)>> {
         Ok(
-            if let Some((override_config, reason)) = self.find_override_config()? {
-                Some((override_config.into_local_toolchain_name(), reason))
+            if let Some((override_config, source)) = self.find_override_config()? {
+                Some((override_config.into_local_toolchain_name(), source))
             } else {
                 self.get_default()?
-                    .map(|x| (x.into(), ActiveReason::Default))
+                    .map(|x| (x.into(), ActiveSource::Default))
             },
         )
     }
 
-    fn find_override_config(&self) -> Result<Option<(OverrideCfg, ActiveReason)>> {
-        let override_config: Option<(OverrideCfg, ActiveReason)> =
+    fn find_override_config(&self) -> Result<Option<(OverrideCfg, ActiveSource)>> {
+        let override_config: Option<(OverrideCfg, ActiveSource)> =
             // First check +toolchain override from the command line
             if let Some(name) = &self.toolchain_override {
                 let override_config = name.resolve(&self.get_default_host_triple()?)?.into();
-                Some((override_config, ActiveReason::CommandLine))
+                Some((override_config, ActiveSource::CommandLine))
             }
             // Then check the RUSTUP_TOOLCHAIN environment variable
             else if let Some(name) = &self.env_override {
@@ -548,15 +548,15 @@ impl<'a> Cfg<'a> {
                 // custom, distributable, and absolute path toolchains otherwise
                 // rustup's export of a RUSTUP_TOOLCHAIN when running a process will
                 // error when a nested rustup invocation occurs
-                Some((name.clone().into(), ActiveReason::Environment))
+                Some((name.clone().into(), ActiveSource::Environment))
             }
             // Then walk up the directory tree from 'path' looking for either the
             // directory in the override database, or a `rust-toolchain{.toml}` file,
             // in that order.
-            else if let Some((override_cfg, active_reason)) = self.settings_file.with(|s| {
+            else if let Some((override_cfg, active_source)) = self.settings_file.with(|s| {
                     self.find_override_from_dir_walk(&self.current_dir, s)
                 })? {
-                Some((override_cfg, active_reason))
+                Some((override_cfg, active_source))
             }
             // Otherwise, there is no override.
             else {
@@ -570,13 +570,13 @@ impl<'a> Cfg<'a> {
         &self,
         dir: &Path,
         settings: &Settings,
-    ) -> Result<Option<(OverrideCfg, ActiveReason)>> {
+    ) -> Result<Option<(OverrideCfg, ActiveSource)>> {
         let mut dir = Some(dir);
 
         while let Some(d) = dir {
             // First check the override database
             if let Some(name) = settings.dir_override(d) {
-                let reason = ActiveReason::OverrideDB(d.to_owned());
+                let source = ActiveSource::OverrideDB(d.to_owned());
                 // Note that `rustup override set` fully resolves it's input
                 // before writing to settings.toml, so resolving here may not
                 // be strictly necessary (could instead model as ToolchainName).
@@ -586,7 +586,7 @@ impl<'a> Cfg<'a> {
                 let toolchain_name = ResolvableToolchainName::try_from(name)?
                     .resolve(&get_default_host_triple(settings, self.process))?;
                 let override_cfg = toolchain_name.into();
-                return Ok(Some((override_cfg, reason)));
+                return Ok(Some((override_cfg, source)));
             }
 
             // Then look for 'rust-toolchain' or 'rust-toolchain.toml'
@@ -670,9 +670,9 @@ impl<'a> Cfg<'a> {
                     }
                 }
 
-                let reason = ActiveReason::ToolchainFile(toolchain_file);
+                let source = ActiveSource::ToolchainFile(toolchain_file);
                 let override_cfg = OverrideCfg::from_file(self, override_file)?;
-                return Ok(Some((override_cfg, reason)));
+                return Ok(Some((override_cfg, source)));
             }
 
             dir = d.parent();
@@ -766,8 +766,8 @@ impl<'a> Cfg<'a> {
         &self,
         force_non_host: bool,
         verbose: bool,
-    ) -> Result<(LocalToolchainName, ActiveReason)> {
-        if let Some((override_config, reason)) = self.find_override_config()? {
+    ) -> Result<(LocalToolchainName, ActiveSource)> {
+        if let Some((override_config, source)) = self.find_override_config()? {
             let toolchain = override_config.clone().into_local_toolchain_name();
             if let OverrideCfg::Official {
                 toolchain,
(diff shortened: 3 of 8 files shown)


smoelius:
> The variable's content has the form `source toolchain`

This is redundant, because the latter component is `RUSTUP_TOOLCHAIN`. I'll fix this PR in the next few hours.

EDIT: Done.

src/toolchain.rs
```diff
@@ -605,3 +613,16 @@ impl<'a> Toolchain<'a> {
             .collect())
     }
 }
+
+/// Set the `RUSTUP_TOOLCHAIN_SOURCE` environment variable if not already set.
+///
+/// `RUSTUP_TOOLCHAIN_SOURCE` indicates how the toolchain was determined. The environment
+/// variable could have been set in proxy_mode.rs, in which case it should not be changed.
```
epage: Is there a reason  we let the original through rather than override it?

src/toolchain.rs
```diff
@@ -605,3 +613,16 @@ impl<'a> Toolchain<'a> {
...
+pub(crate) fn maybe_set_env_source(cmd: &mut Command, f: impl FnOnce() -> Option<ActiveReason>) {
+    if env::var_os("RUSTUP_TOOLCHAIN_SOURCE").is_some() {
+        return;
+    }
+    if let Some(reason) = f() {
+        cmd.env("RUSTUP_TOOLCHAIN_SOURCE", reason.to_source());
```
epage: I assume you'll need to update https://github.com/rust-lang/rustup/blob/main/doc/user-guide/src/environment-variables.md

epage: As mentioned in rust-lang/cargo#11036, `RUSTUP_TOOLCHAIN_SOURCE` could be taken as pointing to something inside the toolchain or pointing to what caused the toolchain to be set, not just the kind of toolchain source.

epage: Or is that only env variables that rustup reads?

Are the ones it sets defined anywhere?

src/config.rs
```diff
@@ -104,16 +104,13 @@ pub(crate) enum ActiveReason {
...
-            Self::Default => format!("default {name}"),
-            Self::Environment => format!("env {name}"),
-            Self::CommandLine => format!("cli {name}"),
-            Self::OverrideDB(_) => format!("override {name}"),
-            Self::ToolchainFile(_) => format!("config {name}"),
+            Self::Default => "default",
+            Self::Environment => "env",
+            Self::CommandLine => "cli",
+            Self::OverrideDB(_) => "override",
+            Self::ToolchainFile(_) => "config",
```
epage: > This is redundant, because the latter component is RUSTUP_TOOLCHAIN. I'll fix this PR in the next few hours.

Huh, `RUSTUP_TOOLCHAIN` is documented as something the user can set; I didn't realize that rustup sets it on child processes but https://github.com/rust-lang/rustup/blob/061cf7830a61eaefaddddb8454232186d26fd2d1/src/toolchain.rs#L184

src/toolchain.rs
smoelius: I think that's the behavior we want.

In my installation, all the proxies link to rustup:
```
lrwxrwxrwx 1 smoelius smoelius       32 Jan 31  2025 cargo -> /home/smoelius/.cargo/bin/rustup
```
So I think we want it to handle the `+` args and we should just let those through: https://github.com/rust-lang/rustup/blob/061cf7830a61eaefaddddb8454232186d26fd2d1/src/cli/proxy_mode.rs#L19-L26

I am happy to be corrected if I am misunderstanding.

src/toolchain.rs
smoelius: I think both are in that list. I updated the list.

smoelius: I take your point. But I thought we wanted to let the rustup team decide on the variable name. So I have tried to remain agnostic on that front.

src/toolchain.rs
epage: What happens if run `cargo +beta test` and one of my tests invokes `cargo doc` within a directory  that has a `rustup-toolchain.toml` for nightly?

smoelius: `RUSTUP_TOOLCHAIN_SOURCE` should be set to `cli`. But I don't think I understand the point of your question.

epage: Within the `cargo doc` call, the source of the toolchain is `config` but `RUSTUP_TOOLCHAIN_SOURCE` will incorrectly report back `cli` because the outer most rustup call had it set on the CLI.

To me, it seems like we should only inherit `RUSTUP_TOOLCHAIN_SOURCE` if we are inheriting the parent's toolchain.

smoelius: > Within the `cargo doc` call, the source of the toolchain is `config`

I don't mean to question you but are you sure that's right?

I just tested in a configuration like you described with this program:
```rs
//! ```
//! let stdout = std::process::Command::new("env").output().unwrap().stdout;
//! std::fs::write("env.txt", stdout).unwrap();
//! ```

#[test]
fn test() {
    std::process::Command::new("cargo").arg("doc").status().unwrap();
}
```
And here is what I saw written to the `env.txt` file:
```
RUSTUP_HOME=/home/smoelius/.rustup
RUSTUP_TOOLCHAIN=beta-x86_64-unknown-linux-gnu
RUSTUP_TOOLCHAIN_SOURCE=cli
```
If I run with just `cargo test` instead of `cargo +beta test`, I observe what you describe.

epage: > If I run with just cargo test instead of cargo +beta test, I observe what you describe.

So my exact example didn't work but the principle still applies.

doc/user-guide/src/environment-variables.md
```diff
@@ -72,6 +72,8 @@
 - `RUSTUP_CONCURRENT_DOWNLOADS` *unstable* (default: the number of components to download). Controls the number of
   downloads made concurrently.
 
+- `RUSTUP_TOOLCHAIN_SOURCE`. Set by rustup to tell downstream tools how `RUSTUP_TOOLCHAIN` was determined.
```
rami3l: I think this will be an undocumented environment variable at least at first?

cc @djc

src/config.rs
```diff
@@ -102,6 +102,19 @@ pub(crate) enum ActiveReason {
...
 }
 
+impl ActiveReason {
+    /// Format `ActiveReason` for setting the `RUSTUP_TOOLCHAIN_SOURCE` environment variable.
+    pub fn to_source(&self) -> &str {
+        match self {
+            Self::Default => "default",
+            Self::Environment => "env",
+            Self::CommandLine => "cli",
+            Self::OverrideDB(_) => "override",
```
rami3l: I think the naming here is a bit dangerous: it's not completely clear that this is a per-dir global config thing, i.e. `rustup override set`. Idem for `config`. Maybe we can call it something else? Something closer to the original?

src/toolchain.rs
djc: > To me, it seems like we should only inherit `RUSTUP_TOOLCHAIN_SOURCE` if we are inheriting the parent's toolchain.

This makes sense to me as the least surprising behavior.

doc/user-guide/src/environment-variables.md
djc: Hmm, not sure. What's your rationale?

src/config.rs
djc: `user-config` and `path-override`?

src/config.rs
```diff
@@ -102,6 +102,19 @@ pub(crate) enum ActiveReason {
     ToolchainFile(PathBuf),
 }
 
+impl ActiveReason {
+    /// Format `ActiveReason` for setting the `RUSTUP_TOOLCHAIN_SOURCE` environment variable.
+    pub fn to_source(&self) -> &str {
```
djc: So it seems we have this `ActiveReason` enum which pretty much represents the same core concept. We should probably rename either `ActiveReason` (to `ActiveSource` or maybe `ToolchainSource`?) or `RUSTUP_TOOLCHAIN_SOURCE` (to `RUSTUP_TOOLCHAIN_REASON`).

src/toolchain.rs
```diff
@@ -181,6 +181,14 @@ impl<'a> Toolchain<'a> {
 
         env_var::inc("RUST_RECURSION_COUNT", cmd, self.cfg.process);
 
+        maybe_set_env_source(cmd, || {
```
djc: Why do we need it here? Seems like potentially doing this only in the high-level proxy mode binary should be sufficient?

djc:
Thanks for working on this!

(Please squash your commits -- except for a potential `ActiveReason` rename, which would be nice to keep in a separate commit or PR.)

doc/user-guide/src/environment-variables.md
rami3l: @djc I assume this to be a private contract between `rustup` and `cargo` for now, or am I misunderstanding the context? As a middle ground we could also make it `unstable`.

djc: Since we want there to be a contract between rustup and cargo, I think the documentation would serve as a useful way to describe what the contract is supposed to be at a high level. We can definitely qualify it as "unstable", especially until its use in cargo is stabilized.

src/config.rs
rami3l: @djc Maybe `toolchain-file` and `path-override`?

djc: Yes, that's better -- clearly I was misunderstanding `config`!

src/toolchain.rs
smoelius: > > To me, it seems like we should only inherit `RUSTUP_TOOLCHAIN_SOURCE` if we are inheriting the parent's toolchain.
> 
> This makes sense to me as the least surprising behavior.

Should I document this somewhere?

doc/user-guide/src/environment-variables.md
smoelius: I marked the variable as unstable in `environment-variables.md`.

src/toolchain.rs
smoelius: That seems to work!

Should I leave a comment in this file?

My intuition told me to look here. I wonder if others will do the same since this is where other environment variables are set.

smoelius:
> (Please squash your commits -- except for a potential `ActiveReason` rename, which would be nice to keep in a separate commit or PR.)

I will once the dust has settled. :slightly_smiling_face:

src/config.rs
smoelius: > So it seems we have this `ActiveReason` enum which pretty much represents the same core concept. We should probably rename either `ActiveReason` (to `ActiveSource` or maybe `ToolchainSource`?) or `RUSTUP_TOOLCHAIN_SOURCE` (to `RUSTUP_TOOLCHAIN_REASON`).

What would you prefer?

src/toolchain.rs
djc: > > > To me, it seems like we should only inherit `RUSTUP_TOOLCHAIN_SOURCE` if we are inheriting the parent's toolchain.
> > 
> > This makes sense to me as the least surprising behavior.
> 
> Should I document this somewhere?

Yes, sounds good. In the existing documentation thingy?

src/config.rs
djc: I think "source" is a better conceptual match than "reason". @rami3l thoughts? And yourself?

src/toolchain.rs
djc: Hmm, not sure. If you think there's a logical place to comment...

smoelius:
I made one small change ("downstream" -> "proxied") and squashed all of my commits down to one.

I think there are still three outstanding points, which could be separate PRs:
- Document that we should only inherit `RUSTUP_TOOLCHAIN_SOURCE` if we are inheriting the parent's toolchain (https://github.com/rust-lang/rustup/pull/4518#discussion_r2417771456). My attempts at this felt unnatural, so I left this alone for now.
- Mention `RUSTUP_TOOLCHAIN_SOURCE` in `src/toolchain.rs` (https://github.com/rust-lang/rustup/pull/4518#discussion_r2417774800). I propose adding "Note that `RUSTUP_TOOLCHAIN_SOURCE`  is set in `src/cli/proxy_mode.rs`." or similar to to `set_env`'s rustdoc comment.
- Possibly rename `ActiveReason` (https://github.com/rust-lang/rustup/pull/4518#discussion_r2417773257). I'd like to let @rami3l comment, as I have no strong opinion.

src/config.rs
rami3l: @smoelius To be honest "source" makes me think about source repositories in the context of installers (such as APT). But it could very well be just me.

`ActiveSource` is fine to me since it adds more context, but just be careful because there are a bunch of `*reason`s to be changed across the codebase, and the current `Display` implementation should conceptually replaced by this `to_source()`.

OTOH if we are going that route, the original `Display` might have to go to somewhere else for the sake of completeness, meaning I find it a bit weird to output a sentence in the `impl Display for ActiveSource`. Is that used very often in the codebase?

smoelius: I agree that `ActiveSource` sounds sensible and that the current `to_source` function could serve as its `Display` implementation. I also think `ActiveReason`'s current `Display` implementation would make a good `to_reason` function.

I don't know how often `ActiveReason`'s current `Display` implementation is used. But I think commenting it out and looking at the compile errors should reveal that.

Make these changes in a separate PR?

djc: > Make these changes in a separate PR?

That would work -- separate commit in this PR is also fine IMO.

smoelius:
I think all unchanged uses of "reason" are in license files, markdown files, or comments. But I apologize if I missed anything.

rami3l (commented):
I don't have other questions myself but I'd like to wait for other reviewers to make sure that their concerns have been properly addressed.

Nice work!

src/cli/proxy_mode.rs
```diff
@@ -32,9 +37,25 @@ pub async fn main(arg0: &str, current_dir: PathBuf, process: &Process) -> Result
...
+        } else {
+            None
+        },
+    );
     run_command_for_dir(cmd, arg0, &cmd_args)
 }
+
+/// Set the `RUSTUP_TOOLCHAIN_SOURCE` environment variable to indicate how the toolchain was
+/// determined.
+fn set_env_source(cmd: &mut Command, reason: Option<ActiveReason>) {
```
djc: Let's inline this single-caller trivial function.

doc/user-guide/src/environment-variables.md
```diff
@@ -72,6 +72,8 @@
 - `RUSTUP_CONCURRENT_DOWNLOADS` *unstable* (default: the number of components to download). Controls the number of
   downloads made concurrently.
 
+- `RUSTUP_TOOLCHAIN_SOURCE` *unstable*. Set by rustup to tell proxied tools how `RUSTUP_TOOLCHAIN` was determined.
```
djc, line 75: Please clean up the commit message for the first commit to remove all the cruft from the development commits you squashed.

src/config.rs
```diff
@@ -92,38 +92,39 @@ impl<T: Into<String>> From<T> for OverrideFile {
...
     CommandLine,
     OverrideDB(PathBuf),
     ToolchainFile(PathBuf),
 }
 
-impl ActiveReason {
-    /// Format `ActiveReason` for setting the `RUSTUP_TOOLCHAIN_SOURCE` environment variable.
-    pub fn to_source(&self) -> &str {
+impl ActiveSource {
+    pub fn to_reason(&self) -> String {
```
djc, line 106: IMO switching around the `Display` impl and `to_source()` is another separate change. I think the commit history should be like this:

- Rename `ActiveReason` to `ActiveSource`
- Move `Display` impl to `to_reason()`
- Implement `RUSTUP_TOOLCHAIN_SOURCE` with new `Display` impl

djc (approved):
LGTM, thanks!

smoelius:
> LGTM, thanks!

And the tests haven't even passed yet! :laughing: Thank you!

smoelius:
Is anything still needed before this PR can be merged?

djc:
Don't think so, thanks again!

doc/user-guide/src/environment-variables.md
weihanglo, line 75: AFAIK, rust-analyzer (and some other tools) overrides `RUSTUP_TOOLCHAIN` environment variable directly based on the result of its probing [^1]. They also documents that user might set that directly [^2].

My questions are:

* Should those cases above all be considered as environment overrides source? What is the source of true of `RUSTUP_TOOLCHAIN_SOURCE`?
* Should tool override and provide the original `RUSTUP_TOOLCHAIN_SOURCE` if they have resolved to the actual toolchain first on their own through?

Note that this is probably not going to affect general user cases, and `RUSTUP_TOOLCHAIN_SOURCE` can just be a best-effort env. Just FYI at $WORK we have a quite complex layer of rustup overrides and I wonder whether we should fully trust this or reset it whenever we override `RUSTUP_TOOLCHAIN`.

[^1]: https://github.com/rust-lang/rust-analyzer/blob/1e20331e42449dfc0b44bce84147a06772d045d7/crates/rust-analyzer/src/flycheck.rs#L639

[^2]: https://github.com/rust-lang/rust-analyzer/blob/1e20331e42449dfc0b44bce84147a06772d045d7/docs/book/src/installation.md?plain=1#L24-L26

smoelius: > AFAIK, rust-analyzer (and some other tools) overrides `RUSTUP_TOOLCHAIN` environment variable directly based on the result of its probing [1](#user-content-fn-1-962a993a67c1fe0f54e9c4f546df8699). They also documents that user might set that directly [2](#user-content-fn-2-962a993a67c1fe0f54e9c4f546df8699).
> 
> My questions are:
> 
> * Should those cases above all be considered as environment overrides source? What is the source of true of `RUSTUP_TOOLCHAIN_SOURCE`?

Could your question be restated as, "What are the definitions of the five `RUSTUP_TOOLCHAIN_SOURCE` categories?"

I would argue they are the same as in the ["Overrides" section](https://rust-lang.github.io/rustup/overrides.html) of the Rustup book.

For the sake of your examples, the toolchain would not be specified on the command line,[^1] so the toolchain should be determined by `RUSTUP_TOOLCHAIN`. In that sense, it seems correct that `RUSTUP_TOOLCHAIN_SOURCE` should be set to `env`.

[^1]: It's also not clear to me what that would mean.

> * Should tool override and provide the original `RUSTUP_TOOLCHAIN_SOURCE` if they have resolved to the actual toolchain first on their own through?
> 
> ...Just FYI at $WORK we have a quite complex layer of rustup overrides and I wonder whether we should fully trust this or reset it whenever we override `RUSTUP_TOOLCHAIN`.

My non-authoritative opinion is "no", tools other that rustup should not set `RUSTUP_TOOLCHAIN_SOURCE`.

I am reminded of a comment @epage made in https://github.com/rust-lang/cargo/issues/15099#issuecomment-2666737150:

> In the Cargo team meeting, someone brought up users manually unsetting `CARGO` but I worry about how error prone that is and how difficult to debug.

If non-rustup tools are allowed to set `RUSTUP_TOOLCHAIN_SOURCE`, I could imagine it leading to a million bug reports down the road.

Should we expand the documentation to say that non-rustup tools should not set `RUSTUP_TOOLCHAIN_SOURCE`?

weihanglo: > the toolchain would not be specified on the command line,[1](https://github.com/rust-lang/rustup/pull/4518#user-content-fn-1-a3dc2a83a6142be8be66ffb464608ea5) so the toolchain should be determined by RUSTUP_TOOLCHAIN. In that sense, it seems correct that RUSTUP_TOOLCHAIN_SOURCE should be set to env.

Given the current definition of `RUSTUP_TOOLCHAIN_SOURCE`, I agree. Although it’s somewhat of an optimization on rust-analyzer’s side, the value r-a sets is still retrieved through the rustup proxy.

I guess my question could be rephrased as: “If a tool caches the rustup-resolved toolchain path for optimization, should it also propagate the original `RUSTUP_TOOLCHAIN_SOURCE`?”

I don’t think we need to overthink this, since rust-analyzer users likely won’t be affected. That said, we should keep in mind that some “rustup” wrappers might produce unexpected or less precise diagnostics if Cargo starts treating `RUSTUP_TOOLCHAIN_SOURCE` as the source of truth. (Those wrappers probably want to preserve the original source instead of being treated as `env`.)

Anyway, I agree that we may want to make this clearer in the documentation, especially which environments are meant for reading and which are for writing.

djc: > I guess my question could be rephrased as: “If a tool caches the rustup-resolved toolchain path for optimization, should it also propagate the original `RUSTUP_TOOLCHAIN_SOURCE`?”

It seems reasonable to answer this with yes.

I wonder if we can test this whole thing before we release rustup with this code? I guess it would need some of the relevant Cargo changes (not sure if anyone is working on those already) and then pulling in both as snapshot builds.

rami3l: IMHO I agreed to include this patch based on the premise that this is a) a private contract between us and cargo; b) a best-effort optimization to communicate with the latter why a toolchain is activated.

As such, I am wondering what will the problems caused by overriding that env var be if you are already bypassing rustup's toolchain resolution. So "yes" to the question above, since the results are purely cargo-oriented.

smoelius: > I guess it would need some of the relevant Cargo changes (not sure if anyone is working on those already)

@djc Just to answer your question: https://github.com/rust-lang/cargo/pull/16131

smoelius:
Can I ask how l long the [Rustup 1.29.0 beta testing](https://blog.rust-lang.org/inside-rust/2025/12/20/rustup-1.29.0-beta-cft/) is expected to take?

rami3l:
> Can I ask how l long the [Rustup 1.29.0 beta testing](https://blog.rust-lang.org/inside-rust/2025/12/20/rustup-1.29.0-beta-cft/) is expected to take?

@smoelius In the past we didn't have a fixed timeline like the rest of the ecosystem, but in general it's about 2-4 weeks depending on the amount of changes. I think it'd be reasonable not to ship during the holiday season so I'd say mid-late Jan 2026. Does this influence the shipping plans on your side?

smoelius:
> Does this influence the shipping plans on your side?

Sort of, but mostly I was asking for planning purposes.

By what you are saying, it will be possible to test the [Cargo changes](https://github.com/rust-lang/cargo/pull/16131) mid-late Jan 2026. So I will have to accelerate my efforts there.

Thanks very much for your response. 🙏

rami3l:
> By what you are saying, it will be possible to test the [Cargo changes](https://github.com/rust-lang/cargo/pull/16131) mid-late Jan 2026.

@smoelius Sure, but I am wondering what particular test(s) could be difficult on your side given that you can opt in rustup's dev environment as [described in the post](https://blog.rust-lang.org/inside-rust/2025/12/20/rustup-1.29.0-beta-cft/#how-to-test). This way we might have an idea to improve the beta testing experience in the long run. Many thanks in advance :)

smoelius:
I have already opted-in locally. The issue is testing the rustup changes elsewhere, e.g., on Cargo's CI.

I am sure the Cargo PR could be adapted to use the opt-in procedure you cited. But if I were a Cargo maintainer, I would be skittish about submitters setting environment variables (e.g., `RUSTUP_UPDATE_ROOT`) in CI.

And one has to ask, since the new version is only a few weeks away, is it better to just wait?

Once the new version of rustup is published, all of this becomes a non-issue.

PR #4519, opened 2025-10-08
feat(cli/self-update): add support for PowerShell on Unix systems

paxbun:
This PR adds PowerShell support on Unix systems. Some may find it unusual to use PowerShell on macOS or Linux, but it is already supported by popular technologies as follows:

- [GitHub Actions](https://docs.github.com/en/actions/tutorials/build-and-test-code/powershell) (`shell: pwsh` is supported in every type of runner)
- Visual Studio Code
- Python virtual environment managers like [Conda](https://docs.conda.io/projects/conda/en/stable/dev-guide/deep-dives/activation.html) and [venv](https://docs.python.org/3/library/venv.html#how-venvs-work).

This is my first contribution to an official Rust project, so please let me know if there's anything else I should do. Thanks!

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -58,7 +58,7 @@ fn cargo_home_str_with_home(home: &str, process: &Process) -> Result<Cow<'static
 }
 
 // TODO: Tcsh (BSD)
-// TODO?: Make a decision on Ion Shell, Power Shell, Nushell
+// TODO?: Make a decision on Ion Shell
 // Cross-platform non-POSIX shells have not been assessed for integration yet
 fn enumerate_shells() -> Vec<Shell> {
     vec![
@@ -68,6 +68,7 @@ fn enumerate_shells() -> Vec<Shell> {
         Box::new(Fish),
         Box::new(Nu),
         Box::new(Tcsh),
+        Box::new(Pwsh),
     ]
 }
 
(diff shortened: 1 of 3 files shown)


paxbun:
Tested using a development build on macOS:

| Installation using a development build | New terminal after installation |
| - | - |
| image | image |

rami3l (approved):
This looks okay to me. Many thanks!

However, since shell support doesn't come with proper testing, I'll let my teammates have another look just to be sure :)

src/cli/self_update/shell.rs
```diff
@@ -368,6 +369,71 @@ impl UnixShell for Tcsh {
...
+            // `~/.config/powershell/profile.ps1` is the "Current User, All Hosts"
+            // profile file. It affects all PowerShell hosts of the current user.
+            paths.push(config_dir.join("profile.ps1"));
+
+            // Some editors like Visual Studio Code or PowerShell ISE use their
+            // own dedicated profile files, whose file names are
+            // `<Host Profile ID>_profile.ps1`. Such customization may use
+            // PowerShell Editor Services for IDE integration.
+            // https://github.com/PowerShell/PowerShellEditorServices
+            for host_profile in config_dir.read_dir().into_iter().flatten().flatten() {
```
djc: I don't know what this is trying to do exactly, but `into_iter().flatten().flatten()` looks pretty convoluted to my eyes.

src/cli/self_update/shell.rs
```diff
@@ -368,6 +369,71 @@ impl UnixShell for Tcsh {
...
+impl UnixShell for Pwsh {
+    fn does_exist(&self, process: &Process) -> bool {
+        matches!(process.var("SHELL"), Ok(sh) if sh.contains("pwsh"))
+            || utils::find_cmd(&["pwsh"], process).is_some()
+    }
+
+    fn rcfiles(&self, process: &Process) -> Vec<PathBuf> {
+        let mut paths = vec![];
+
+        if let Some(mut config_dir) = process.home_dir() {
```
djc: Nit: suggest writing it like this:

```rust
let Some(mut config_dir) = process.home_dir() else {
     return paths;
};
```

to reduce rightward drift.

djc:
Mostly LGTM, some small nits.

src/cli/self_update/shell.rs
paxbun: This is just to use fewer `if`s and `let-else`s; would it be better to write like the following?

```rust
if let Ok(config_dir) = config_dir.read_dir() {
    for host_profile in config_dir {
        let Ok(host_profile) = host_profile else {
            continue;
        };
        let host_profile_path = host_profile.path();
        if !host_profile_path.is_file() {
            continue;
        }
        if host_profile_path.ends_with("_profile.ps1") {
            paths.push(host_profile_path);
        }
    }
}
```

src/cli/self_update/shell.rs
paxbun: Updated as requested!

src/cli/self_update/shell.rs
paxbun: Updated as above.

src/cli/self_update/shell.rs
```diff
@@ -380,25 +380,32 @@ impl UnixShell for Pwsh {
...
+        // `~/.config/powershell/profile.ps1` is the "Current User, All Hosts"
+        // profile file. It affects all PowerShell hosts of the current user.
+        paths.push(config_dir.join("profile.ps1"));
+
+        // Some editors like Visual Studio Code or PowerShell ISE use their
+        // own dedicated profile files, whose file names are
+        // `<Host Profile ID>_profile.ps1`. Such customization may use
+        // PowerShell Editor Services for IDE integration.
+        // https://github.com/PowerShell/PowerShellEditorServices
+        if let Ok(config_dir) = config_dir.read_dir() {
```
djc: Would you mind turning this into a let-else, too, and squashing all the commits? Thanks!

paxbun: Pushed a new commit as requested!

djc (approved):

PR #4520, opened 2025-10-08
refactor: Replace termcolor with anstream

epage:
`anstream` is an existing dependency through clap that allows a simpler, more natural way of writing styled text by making ANSI escape codes the API for styling, rather than a `stream.set_color(_)`.  You write ANSI escape codes unconditionally, and `anstream` will automatically strip them if needed (or convert them to wincon API calls for old Windows 8 systems).  The caller can use whatever styling API they want and don't have to worry about what stream type they are ultimately written to.

This will unblock `clap` output being styled.  Clap can either directly handle writing output to the terminal, write unstyled output to a string, or write styled output to a string.  With Rustup requiring all output to go through custom, termcolor-like streams for testing, this limited Rustup to only getting unstyled output.  With this change, Rustup can get Clap's styled output and then have `anstream` process it as needed.

To generate ANSI escape codes, this uses `anstyle` which is primarily intended to be a low-policy API for ANSI escape codes so they are API stable for theming in APIs like `clap` or `annotate-snippets`. Using the formatting `alternate` though provides a nice way to reset the stream.

To simplify the transition for callers of `ColorableTerminal` through this transition, intermediate steps are taken, including adding the `anstyle-termcolor` adapter and then removing it.

I manually tested what I could of Rustup's output to spot check the styling.  Cargo uses `snapbox` (the core of `trycmd`) to perform snapshot testing of ANSI escape codes by rendering them as SVGs which get rendered in GitHub diffs.  Rustc integrated the lower level `anstyle-svg` package into their existing snapshotting system.  I have considered SVG support for `trycmd` but haven't implemented it yet.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -193,7 +193,8 @@ fn show_channel_updates(
         Ok((pkg, banner, width, color, version, previous_version))
     });
 
-    let mut t = cfg.process.stdout();
+    let t = cfg.process.stdout();
+    let mut t = t.lock();
 
     let data: Vec<_> = data.collect::<Result<_>>()?;
     let max_width = data
@@ -208,17 +209,13 @@ fn show_channel_updates(
             None => Style::new(),
         }
         .bold();
-        let _ = write!(t.lock(), "  {padding}");
-        let _ = t.style(&style);
-        let _ = write!(t.lock(), "{pkg} {banner}");
-        let _ = t.reset();
-        let _ = write!(t.lock(), " - {version}");
+        let _ = write!(t, "  {padding}{style}{pkg} {banner}{style:#} - {version}");
         if let Some(previous_version) = previous_version {
-            let _ = write!(t.lock(), " (from {previous_version})");
+            let _ = write!(t, " (from {previous_version})");
         }
-        let _ = writeln!(t.lock());
+        let _ = writeln!(t);
     }
-    let _ = writeln!(t.lock());
+    let _ = writeln!(t);
 
     Ok(())
 }
@@ -258,15 +255,14 @@ pub(super) fn list_items(
     quiet: bool,
     process: &Process,
 ) -> Result<utils::ExitCode> {
-    let mut t = process.stdout();
+    let t = process.stdout();
+    let mut t = t.lock();
     let bold = Style::new().bold();
     for (name, installed) in items {
         if installed && !installed_only && !quiet {
-            t.style(&bold)?;
-            writeln!(t.lock(), "{name} (installed)")?;
-            t.reset()?;
+            writeln!(t, "{bold}{name} (installed){bold:#}")?;
         } else if installed || !installed_only {
-            writeln!(t.lock(), "{name}")?;
+            writeln!(t, "{name}")?;
         }
     }
     Ok(utils::ExitCode(0))
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
 use std::{cmp, env};
 
-use anstyle::AnsiColor;
+use anstyle::{AnsiColor, Style};
 use anyhow::{Context, Result, anyhow};
 use git_testament::{git_testament, render_testament};
 use tracing::{error, info, warn};
@@ -18,7 +18,7 @@ use crate::{
     dist::{TargetTriple, ToolchainDesc},
     errors::RustupError,
     install::UpdateStatus,
-    process::{Attr, Process},
+    process::Process,
     toolchain::{LocalToolchainName, Toolchain, ToolchainName},
     utils,
 };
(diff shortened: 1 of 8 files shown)


epage:
```
---- suite::cli_self_upd::as_rustup_setup stdout ----
ran: rustup ["set", "default-host", "aarch64-unknown-linux-gnu"]
inprocess: true
status: Some(0)
duration: 0.014s
stdout:
====

====

stderr:
====

====

ran: rustup ["set", "auto-self-update", "disable"]
inprocess: true
status: Some(0)
duration: 0.006s
stdout:
====

====

stderr:
====
info: auto-self-update mode set to 'disable'

====


thread 'suite::cli_self_upd::as_rustup_setup' panicked at /checkout/src/test/clitools.rs:373:18:
Unable to run test command: Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" }


failures:
    suite::cli_self_upd::as_rustup_setup
```

Not all jobs saw this failure and I can't reproduce this locally.  Anyone have insight into this?

rami3l:
> Not all jobs saw this failure and I can't reproduce this locally.  Anyone have insight into this?

@epage This looks like someone lingering problem from the past (before my joining the team). Is that frequent?

epage:
Rebased

src/cli/common.rs
```diff
@@ -203,11 +203,14 @@ fn show_channel_updates(
     for (pkg, banner, width, color, version, previous_version) in data {
         let padding = max_width - width;
         let padding: String = " ".repeat(padding);
-        let _ = write!(t.lock(), "  {padding}");
-        let _ = t.attr(Attr::Bold);
-        if let Some(color) = color {
-            let _ = t.fg(color);
+        let style = if let Some(color) = color {
```
djc: Nit: would prefer a `match` for this:

```rust
let style = match color {
     Some(color) => color.on_default(),
     None => Style::new(),
};
```

src/process/terminal_source.rs
```diff
@@ -60,8 +58,12 @@ impl ColorableTerminal {
             _ => ColorChoice::Never,
         };
         let inner = match stream {
-            StreamSelector::Stdout => TerminalInner::StandardStream(StandardStream::stdout(choice)),
-            StreamSelector::Stderr => TerminalInner::StandardStream(StandardStream::stderr(choice)),
+            StreamSelector::Stdout => {
+                TerminalInner::Stdout(AutoStream::new(std::io::stdout(), choice))
```
djc: Nit: please import `io`.

djc (approved):
Good stuff!

src/process/terminal_source.rs
```diff
@@ -14,7 +14,7 @@ use anstyle::{Reset, Style};
 
 use super::Process;
 #[cfg(feature = "test")]
-use super::file_source::{TestWriter, TestWriterLock};
+use super::file_source::TestWriter;
```
rami3l, line 18: Sanity check: Is `TestWriterLock` used elsewhere in your final version? If not it might be better to remove the dead code.

djc: I was looking at the same thing but figured we could tackle it after this is merged.

rami3l: @djc Thanks! I did a simple LSP-based search and have found no other cases, if that's true then there is no need to do another PR; we could do it directly in this commit even :)

src/process/terminal_source.rs
epage: `use std::io` was already present.  FYI this could be caught automatically with `unused_qualifications = "warn"` in `Cargo.toml`.

djc: > `use std::io` was already present. FYI this could be caught automatically with `unused_qualifications = "warn"` in `Cargo.toml`.

Huh, I thought that was `warn` by default.

src/process/terminal_source.rs
epage: The one remaining use of `TestWriterLock` is relatively trivial: https://github.com/rust-lang/rustup/blob/2bd635656e6e772b7ad3ea59b010018920ffadd6/src/process/file_source.rs#L102

rami3l: @epage In that case, I agree with @djc that it might be worth another PR after this to address it, albeit relatively trivial. As such, I think we can merge this PR with this comment unresolved for now. Thanks again!

rami3l (approved):

src/process/terminal_source.rs
rami3l: Should be addressed by https://github.com/rust-lang/rustup/pull/4538.

xtqqczze:
> ```
> ---- suite::cli_self_upd::as_rustup_setup stdout ----
> ran: rustup ["set", "default-host", "aarch64-unknown-linux-gnu"]
> inprocess: true
> status: Some(0)
> duration: 0.014s
> stdout:
> ====
> 
> ====
> 
> stderr:
> ====
> 
> ====
> 
> ran: rustup ["set", "auto-self-update", "disable"]
> inprocess: true
> status: Some(0)
> duration: 0.006s
> stdout:
> ====
> 
> ====
> 
> stderr:
> ====
> info: auto-self-update mode set to 'disable'
> 
> ====
> 
> 
> thread 'suite::cli_self_upd::as_rustup_setup' panicked at /checkout/src/test/clitools.rs:373:18:
> Unable to run test command: Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" }
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
> 
> failures:
>     suite::cli_self_upd::as_rustup_setup
> ```
> 
> Not all jobs saw this failure and I can't reproduce this locally. Anyone have insight into this?

Tracked by https://github.com/rust-lang/rustup/issues/4779

PR #4521, opened 2025-10-09
Clean up terminal source usage

djc:
cc @epage -- sorry, this will cause some conflicts in #4520, but I believe it will also make that PR easier to review once rebased.


src/cli/common.rs
```diff
@@ -10,17 +10,19 @@ use std::{cmp, env};
 
 use anyhow::{Context, Result, anyhow};
 use git_testament::{git_testament, render_testament};
+use termcolor::Color;
 use tracing::{debug, error, info, trace, warn};
 use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
 
+use crate::process::Attr;
```
rami3l: Maybe just merge this to the section below?

src/cli/self_update.rs
```diff
@@ -47,9 +47,11 @@ use clap::builder::PossibleValue;
 use itertools::Itertools;
 use same_file::Handle;
 use serde::{Deserialize, Serialize};
+use termcolor::Color;
 use tracing::{error, info, trace, warn};
 
 use crate::download::download_file;
+use crate::process::Attr;
```
rami3l: Maybe just merge this to the section below?

rami3l (approved):
LGTM, and thanks a lot :)

src/cli/common.rs
djc: Good call!

src/cli/self_update.rs
djc: Done.

Issue #4423, opened 2025-07-22
1.29.0 Release Planning


rami3l:
@rust-lang/rustup is welcomed to propose outstanding issues that would be urgent enough to address before cutting the upcoming patch release.
Other issues on the current milestone will be postponed to the milestone representing the next release.

## Outstanding Issues/PRs

- [x] #4420
- [x] #4337
- [x] #4428
- [x] #4605
- [x] #4631

## Release Process

- https://github.com/rust-lang/rustup/pull/4513
- #4637
- https://github.com/rust-lang/rustup/pull/4640

## Note

It could be a good opportunity to test out https://github.com/rust-lang/promote-release/pull/84, but it would not be a dealbreaker for cutting this release.

djc:
I'd like to go to 1.29.0 and drop curl and native-tls support. I have one more change I'd like to make before that.

Might be good to also write up a list of (notable) changes that we have on master right now to get a handle on what the release would look like.

rami3l:
> I'd like to go to 1.29.0 and drop curl and native-tls support.

@djc That sounds... risky since there are still remaining problems in #3790, and I would originally expect OpenSSL to last for at least a few more patches.

This doesn't mean we cannot do 1.29.0 here and now; it's just that it requires more attention into justifying/advertising the change and the possible consequences, and I expect that would take longer than a normal patch release... combined with #4420 I'm afraid that some people might never get a working 1.28 release and now they are forced to face 1.29 🙈

djc:
> [@djc](https://github.com/djc) That sounds... risky since there are still remaining problems in [#3790](https://github.com/rust-lang/rustup/issues/3790), and I would originally expect OpenSSL to last for at least a few more patches.

It's not obvious to me what problems/remaining risks you are referring to. As I understand it, the change to rustls by default has been pretty smooth and has not led to a lot of issues, even with the explicit warnings in 1.28.2. The one issue I'm aware of is https://github.com/rust-lang/rustup/issues/4325, which is arguably a pretty specific environment problem, and we now have better error handling for that via rustls-platform-verifier 0.6 and I'm working on a fix that includes the roots required for verifying static.rust-lang.org certificates as part of the rustup binary.

rami3l:
[...]

PR #4523, opened 2025-10-10
ci: add support for aarch64-pc-windows-gnullvm target

### LINKED ISSUE #3297: Support for `aarch64-pc-windows-gnullvm`

Can we get target triples going to support Cygwin-style Rust applications for ARM Windows?

While we do technically have `aarch64-pc-windows-msvc` available, cross-compiling Windows Rust apps from non-Windows hosts is remarkably easier with the `windows-gnu` target triple variants. So a `aarch-pc-windows-gnu` target may be easier for many macOS/Linux/UNIX Rust devs to support than `aarch64-pc-windows-msvc`.

### Solution you'd like

Introduce `aarch-pc-windows-gnu` to the set of available rustup targets, similar to the other `windows-gnu` targets.

### Notes

_No response_

magic-akari:
- Closes: #3297

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,6 +25,9 @@ jobs:
     env:
       RUSTFLAGS: -Ctarget-feature=+crt-static
       RUST_MIN_STACK: 16777216
+      # Below is a workaround in order to generate the right bindings for `aws-lc-sys`.
+      # See https://github.com/aws/aws-lc-rs/issues/919 for more details.
+      AWS_LC_SYS_EFFECTIVE_TARGET_aarch64_pc_windows_gnullvm: aarch64-pc-windows-gnu
     permissions:
       id-token: write
       contents: read
@@ -202,6 +205,9 @@ jobs:
     env:
       RUSTFLAGS: -Ctarget-feature=+crt-static
       RUST_MIN_STACK: 16777216
+      # Below is a workaround in order to generate the right bindings for `aws-lc-sys`.
+      # See https://github.com/aws/aws-lc-rs/issues/919 for more details.
+      AWS_LC_SYS_EFFECTIVE_TARGET_aarch64_pc_windows_gnullvm: aarch64-pc-windows-gnu
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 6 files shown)


rami3l:
Looks like we are blocked by the upstream in a few places; I can take it from there in terms of contacting them. Thanks :)

PS: It'd be nice if `rustup-init.sh` can be modified to detect this new host target as well, similar to `x86_64-pc-windows-gnu`.

rami3l:
@djc So the current situation in terms of the download support of this new target is:

* libcurl is not buildable, but you are dropping it anyway?
* aws-lc is not buildable, but I think I can ask the `aws-lc` people to look into it.

Could it be alright, say, for certain builds to use `reqwest` on `native-tls` exclusively?

djc:
> @djc So the current situation in terms of the download support of this new target is:
> 
> * libcurl is not buildable, but you are dropping it anyway?

Last I remember you [didn't want to go there](https://github.com/rust-lang/rustup/issues/4423#issuecomment-3103304390) just yet?

> * aws-lc is not buildable, but I think I can ask the `aws-lc` people to look into it.

That would probably be useful.

> Could it be alright, say, for certain builds to use `reqwest` on `native-tls` exclusively?

Yeah, I guess we're going to need that workaround for a while given the lack of support from aws-lc-rs for a bunch of these targets.

rami3l:
> > * libcurl is not buildable, but you are dropping it anyway?
> 
> Last I remember you [didn't want to go there](https://github.com/rust-lang/rustup/issues/4423#issuecomment-3103304390) just yet?

@djc Yes, but the thing is we can totally merge this in the next release cycle. As I understand it the need for this target is useful but not essential; current ARM64 Windows users can use the MSVC build of `rustup` as a workaround, except the installation experience might be a bit different.

Given that, it is not necessary for us to address the `libcurl` error here and now.

ChrisDenton:
What is the reason for preferring gnullvm builds? Does it really matter if rustup is built using msvc, gnu or gnullvm?

rami3l:
> What is the reason for preferring gnullvm builds? Does it really matter if rustup is built using msvc, gnu or gnullvm?

@ChrisDenton The past convention is that targets with at least tier 2 with host tools are expected to have a respective rustup build. For Windows I understand that this is a bit pointless since the user would usually download the MSVC build... But we already have a x64 GNU ABI build anyway.

My POV in this is not that we really want to start distributing for this platform and how much ROI we can get from that, but rather anyone who wishes to develop this tool using that host target is supposed to get a perfectly normal build.

rami3l:
@magic-akari Please be patient as this PR might be included after the cURL backend removal, and there is no urgent need to keep it up to date as I'll offer to do that by then. Many thanks again :)

rami3l:
With the upstream aws-lc issue resolved and the libcurl backend removed, this PR is finally ready to be merged.

Thanks again @magic-akari :)

djc (approved):

PR #4526, opened 2025-10-11
Scope notifications to downloads only

djc:
Previously:

- #4510


src/dist/component/package.rs
```diff
@@ -199,8 +201,8 @@ fn unpack_ram(
             }
         }
         None => {
-            if let Some(h) = cx.notify_handler {
-                h(Notification::SetDefaultBufferSize(default_max_unpack_ram))
+            if RAM_NOTICE_SHOWN.set(true).is_ok() {
```
rami3l: It's indeed a good idea to use `RAM_NOTICE_SHOWN` here, however as the boolean flag itself serves no value, did you mean to use a singleton, something like `RAM_NOTICE_SHOWN.set(())`?

src/cli/log.rs
```diff
@@ -56,7 +56,7 @@ where
         _ if process.var("NO_COLOR").is_ok() => false,
         _ => process.stderr().is_a_tty(),
     };
-    let maybe_rustup_log_directives = process.var("RUSTUP_LOG");
+    let maybe_rustup_log_directives = dbg!(process.var("RUSTUP_LOG"));
```
rami3l: This hits https://github.com/rust-lang/rustup/blob/30ee45bd1084dd8526c0decd27c405870a16877c/Cargo.toml#L152

src/dist/manifestation/tests.rs
```diff
@@ -1479,30 +1479,29 @@ fn allow_installation(prefix: &InstallPrefix) {
...
-        .await
-        .unwrap_err();
-    assert!(!reuse_notification_fired.get());
+    prevent_installation(&cx.prefix);
+    cx.update_from_dist(&[], &[], false).await.unwrap_err();
+    assert!(
+        str::from_utf8(&cx.tp.stderr())
+            .unwrap()
+            .lines()
+            .find(|&ln| ln.contains("reusing previously downloaded file"))
```
rami3l: I still like to see this string literal extracted as a binding. Locally in the scope of this test is okay.

src/dist/download.rs
```diff
@@ -21,7 +21,6 @@ use crate::utils;
 
 const UPDATE_HASH_LEN: usize = 20;
 
-#[derive(Clone)]
```
rami3l: IIRC https://github.com/rust-lang/rustup/pull/4471 will depend on this. Should it be reverted during the rebase or should we drop this change here?

rami3l (commented):
I've reviewed all the stuff up to and including 9182627130016faf8d0567039d14b52cd9efbda8 and it looks mostly great to me modulo some minor questions.

Should I continue reviewing or wait for the CI to pass?

Anyway, many thanks for simplifying this much of the code!

djc:
> I've reviewed all the stuff up to and including [9182627](https://github.com/rust-lang/rustup/commit/9182627130016faf8d0567039d14b52cd9efbda8) and it looks mostly great to me modulo some minor questions.

Thanks for the quick response!

> Should I continue reviewing or wait for the CI to pass?

I don't expect to make any major changes -- having all feedback before I address your feedback would be useful, if you have time.

rami3l:
> I don't expect to make any major changes -- having all feedback before I address your feedback would be useful, if you have time.

@djc I just checked and can confirm that I have no more questions with the rest of the commits then.

src/dist/download.rs
djc: I'd like to understand why it's necessary, rather than using references like I've done in the previous commit. So I think we should keep this and figure out the best solution during the rebase.

src/dist/component/package.rs
djc: Yes, done.

src/cli/log.rs
djc: Cleaned it up!

src/dist/manifestation/tests.rs
djc: Extracted this and wrapped it in a little `TestContext` API to reduce duplication.

djc:
* Addressed feedback
* Added a commit, "cli: build Cfg earlier in setup mode"
* Fixed Clippy issues and Windows compile errors

src/dist/download.rs
rami3l: @djc Because we will be sending this across the thread boundary and using references in `tokio` is difficult in that case. I'll take this response as accepting the first proposal then.

cc @FranciscoTGouveia for a potential revert to be added while rebasing

rami3l (approved):

rami3l:
@djc Many thanks for the effort!

@FranciscoTGouveia has promised to take it from here :)

PR #4531, opened 2025-10-13
Download cleanup

djc:
Followup from:

- #4526 

Sorry @FranciscoTGouveia, here are some more changes which IMO make things a bunch cleaner. I haven't done any testing yet (would be great if you want to do some of that) but I think the changes here are relatively straightforward?

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1275,14 +1275,7 @@ pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> Result<Option<Pa
 
     // Download new version
     info!("downloading self-update (new version: {available_version})");
-    download_file(
-        &download_url,
-        &setup_path,
-        None,
-        &dl_cfg.tracker,
-        dl_cfg.process,
-    )
-    .await?;
+    download_file(&download_url, &setup_path, None, None, dl_cfg.process).await?;
 
     // Mark as executable
     utils::make_executable(&setup_path)?;
@@ -1301,14 +1294,7 @@ async fn get_available_rustup_version(dl_cfg: &DownloadCfg<'_>) -> Result<String
     let release_file_url = format!("{update_root}/release-stable.toml");
     let release_file_url = utils::parse_url(&release_file_url)?;
     let release_file = tempdir.path().join("release-stable.toml");
-    download_file(
-        &release_file_url,
-        &release_file,
-        None,
-        &dl_cfg.tracker,
-        dl_cfg.process,
-    )
-    .await?;
+    download_file(&release_file_url, &release_file, None, None, dl_cfg.process).await?;
     let release_toml_str = utils::read_file("rustup release", &release_file)?;
     let release_toml = toml::from_str::<RustupManifest>(&release_toml_str)
         .context("unable to parse rustup release file")?;
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1279,7 +1279,7 @@ pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> Result<Option<Pa
         &download_url,
         &setup_path,
         None,
-        &dl_cfg.notifier,
+        &dl_cfg.tracker,
         dl_cfg.process,
     )
     .await?;
@@ -1305,7 +1305,7 @@ async fn get_available_rustup_version(dl_cfg: &DownloadCfg<'_>) -> Result<String
         &release_file_url,
         &release_file,
         None,
-        &dl_cfg.notifier,
+        &dl_cfg.tracker,
         dl_cfg.process,
     )
     .await?;
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -23,7 +23,7 @@ const UPDATE_HASH_LEN: usize = 20;
 pub struct DownloadCfg<'a> {
     pub tmp_cx: &'a temp::Context,
     pub download_dir: &'a PathBuf,
-    pub(crate) notifier: Notifier,
+    pub(crate) tracker: DownloadTracker,
     pub process: &'a Process,
 }
 
@@ -33,7 +33,7 @@ impl<'a> DownloadCfg<'a> {
         DownloadCfg {
             tmp_cx: &cfg.tmp_cx,
             download_dir: &cfg.download_dir,
-            notifier: Notifier::new(cfg.quiet, cfg.process),
+            tracker: DownloadTracker::new(!cfg.quiet, cfg.process),
             process: cfg.process,
         }
     }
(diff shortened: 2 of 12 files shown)


src/utils/mod.rs
```diff
@@ -445,15 +444,13 @@ pub(crate) fn delete_dir_contents_following_links(dir_path: &Path) {
     }
 }
 
-pub(crate) struct FileReaderWithProgress<'a> {
+pub(crate) struct FileReaderWithProgress {
     fh: io::BufReader<File>,
-    tracker: &'a DownloadTracker,
```
rami3l: Question: If I'm reading this right, you are completely removing the progress bar during installation? Why?

src/dist/download.rs
```diff
@@ -198,7 +198,7 @@ pub(crate) struct DownloadTracker {
     /// the message "retrying download" for at least a second.
```
rami3l: Some of this needs to be moved to the new type's docstring.

rami3l (approved):
LGTM modulo some questions :)

src/cli/self_update.rs
```diff
@@ -1275,14 +1275,7 @@ pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> Result<Option<Pa
...
     info!("downloading self-update (new version: {available_version})");
-    download_file(
-        &download_url,
-        &setup_path,
-        None,
-        &dl_cfg.tracker,
-        dl_cfg.process,
-    )
-    .await?;
+    download_file(&download_url, &setup_path, None, None, dl_cfg.process).await?;
```
rami3l, line 1278: Question: Why changing the download tracker to `None` here?

rami3l:
Please note that since this PR is related to more subtle changes in the user behavior that is not fully covered by the test suite, more attention should be paid to ensure the necessary preservation of original semantics.

src/utils/mod.rs
djc: Before this PR, calling `handle()` on a `Notification` would do nothing unless a `Notification::DownloadingComponent` value had been handled first, which would internally call `create_progress_bar()` and added an entry into the `file_progress_bar` `HashMap`. For `Notification`s with a `None` URL field, calling these methods actually did nothing. Thus, I removed all `DownloadTracker` method calls passing `None` as the `url` argument, which made the `DownloadTracker` unused in a bunch of places.

src/cli/self_update.rs
djc: I believe this is preserving the same semantics: there is no earlier call `handle(Notification::DownloadingComponent(..))` aka `DownloadTracker::create_progress_bar()` for this download, thus there will be no entry for this in the `file_progress_bars` `HashMap`, and therefore any downstream calls to `DownloadTracker` did nothing.

(That the code on this branch makes it much more clear how this happens is IMO a great improvement and should make it easier to add better tracking for other downloads using the same scheme.)

src/dist/download.rs
djc: Good call, done.

src/utils/mod.rs
FranciscoTGouveia: Apologies for joining this conversation uninvited, but I would like to note that I have rebased my current PR on top of this one, and the progress bars are not appearing when downloading components -- the only notification shown is when the installation has already completed.

djc: No need to apologize, any feedback is very welcome!

What is your test scenario? Can you use git bisect to see which commit breaks it?

FranciscoTGouveia: I ran `rustup toolchain install beta` (also testing with different values for `RUSTUP_CONCURRENT_DOWNLOADS`) and consistently observed the same behavior.

I have already performed a bisect, which pointed to commit [b42b6b2b](https://github.com/rust-lang/rustup/pull/4531/commits/b42b6b2b31ba374d81257f539bed46088f63e5d8), the second one of this PR.

After reviewing and testing this commit, I confirmed that the `DownloadTracker` methods are being called correctly.
However, the `ProgressBar`s are hidden (as verified using the `.is_hidden()` method), which explains the lack of visible progress reporting.

I traced this issue to the creation of the `DownloadTracker`, where it appears that the `display_progress` boolean variable is inverted, causing the `ProgressDrawTarget` to be _hidden_.

FranciscoTGouveia: I think I have found it!
Inside the `Notifier` we were calling `!quiet`, indeed flipping the variable (see [here](https://github.com/rust-lang/rustup/pull/4531/files#diff-33afa4cad3e3c6cf900b3d5b69a21fbc1bfbf6adeda7116516764a87ddbf877fL195)).

As such, I think it suffices to add the `!` to `cfg.quiet` in the creation of the `DownloadTracker` (see [here](https://github.com/rust-lang/rustup/pull/4531/files#diff-33afa4cad3e3c6cf900b3d5b69a21fbc1bfbf6adeda7116516764a87ddbf877fR36)).

rami3l: @djc Thanks for the explanation! I think this is okay modulo what @FranciscoTGouveia has observed; in the past the installation phase did show a progress bar (but it happened so fast that it might serve no purpose except for the docs); might be an unexpected regression introduced by previous changes, so thanks for having spotted that if that's the case. We can decide whether that's desirable in the new UX...

djc: @FranciscoTGouveia great work! Fixed in my latest push.

Can't reiterate enough -- please always feel free to give feedback on any rustup PRs.

> might be an unexpected regression introduced by previous changes,

Yeah, I think that's what happened -- anyway, I think these changes also make it easier to reintroduce progress bars for any other downloads we think are worth tracking.

FranciscoTGouveia: I have just fetched your branch and confirmed that the progress bars are still not appearing.
This time the issue was introduced in [2050503](https://github.com/rust-lang/rustup/pull/4531/commits/20505036eebcf235c368d5358dc2051857f1cc23) (not related with `cfg.quiet` this time).

FranciscoTGouveia: After a quick inspection, the `ProgressBar`s are correctly created and added to the `MultiProgress`, but the notifications (e.g. `DownloadContentLengthReceived`) are not actually being sent.

This is due to the fact that `status` is `None` [here](https://github.com/rust-lang/rustup/pull/4531/files#diff-3be5702cefcaef4788ff18b2826c64510a70584af74f20f560c2c3bb0550fe7fR110-R112).

If I am not mistaken, we cannot pass `None` [here](https://github.com/rust-lang/rustup/pull/4531/files#diff-33afa4cad3e3c6cf900b3d5b69a21fbc1bfbf6adeda7116516764a87ddbf877fR79).

djc: Great catch! Pushed another update, and going to do another careful self review as well.

djc: Confirmed that the progress bars show up for me now, although there's a lot of flickering involved.

djc:
> Please note that since this PR is related to more subtle changes in the user behavior that is not fully covered by the test suite, more attention should be paid to ensure the necessary preservation of original semantics.

Do you think there's anything left to do here, or is it okay to merge this?

rami3l:
@djc I think it should be okay to merge this patch now that the concerns have been addressed. If there are further problems we can fix them separately later.

PR #4533, opened 2025-10-14
fix(www): removes www subdomain from all rust-lang.org urls

WilliamCollishaw:
# The issue
All `www.rust-lang.org` requests are being redirected to `rust-lang.org` with no subdomain.

Image 10-14-25 at 3 25 PM

This is resulting in a content security policy violation because due to the strict requirement on `img-src 'self' https://www.rust-lang.org;`. 
Screenshot 2025-10-14 at 3 33 56 PM 1

This violation is preventing logos and favicons from displaying.
Screenshot 2025-10-14 at 3 28 34 PM
Screenshot 2025-10-14 at 3 28 38 PM

# The fix

1. Updates the Content-Security-Policy img-src url, removing the `www` subdomain.
2. Removes all `www` subdomain references in HTML to prevent unnecessarily going through the redirect every load.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -7,9 +7,9 @@
   <meta name="description" content="The Rust toolchain installer">
   <link rel="stylesheet" href="normalize.css">
   <link rel="stylesheet" href="rustup.css">
-  <link rel="icon" type="image/png" sizes="32x32" href="https://www.rust-lang.org/static/images/favicon-32x32.png">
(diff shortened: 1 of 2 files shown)


ehuss:
I have noted this at https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Making.20www.2Erust-lang.2Eorg.20a.20static.20website/with/544864303. I don't think it was known that this would happen.

djc (approved):

rami3l (approved):

PR #4534, opened 2025-10-15
cli: add `rustup self uninstall --no-modify-path`

### LINKED ISSUE #4529: `--no-modify-path` option for `rustup self uninstall`

My dotfiles are carefully tracked and managed by [chezmoi](https://www.chezmoi.io/). It would be nice if there's a way to tell rustup not to mess with it during uninstall, much like the option during install.

### Solution you'd like

Add a `--no-modify-path` option for `rustup self uninstall`.

### Notes

Seems like a rather simple addition. I can submit a PR.

I'm also open to naming the option something else. Bikeshedding welcomed.

cyqsimon:
Closes #4529

## TODO

- [x] Tests

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -543,6 +543,10 @@ enum SelfSubcmd {
         /// Disable confirmation prompt
         #[arg(short = 'y')]
         no_prompt: bool,
+
+        /// Do not clean up the `PATH` environment variable
+        #[arg(long)]
+        no_modify_path: bool,
     },
 
     /// Upgrade the internal data format
@@ -725,7 +729,10 @@ pub async fn main(
         RustupSubcmd::Man { command, toolchain } => man(cfg, &command, toolchain).await,
         RustupSubcmd::Self_ { subcmd } => match subcmd {
             SelfSubcmd::Update => self_update::update(cfg).await,
-            SelfSubcmd::Uninstall { no_prompt } => self_update::uninstall(no_prompt, process),
+            SelfSubcmd::Uninstall {
+                no_prompt,
+                no_modify_path,
+            } => self_update::uninstall(no_prompt, no_modify_path, process),
             SelfSubcmd::UpgradeData => cfg.upgrade_data().map(|_| ExitCode(0)),
         },
         RustupSubcmd::Set { subcmd } => match subcmd {
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -540,6 +540,7 @@ enum SelfSubcmd {
 
     /// Uninstall rustup
     Uninstall {
+        /// Disable confirmation prompt
         #[arg(short = 'y')]
         no_prompt: bool,
     },
--- a/tests/suite/cli_paths.rs
+++ b/tests/suite/cli_paths.rs
@@ -267,6 +267,49 @@ error: could not amend shell profile[..]
     }
 
     #[tokio::test]
+    async fn uninstall_doesnt_modify_rcs_with_no_modify_path() {
+        let cx = CliTestContext::new(Scenario::Empty).await;
+        let rcs = [
+            ".bashrc",
+            ".bash_profile",
+            ".bash_login",
+            ".profile",
+            // This requires zsh to be installed, so we test it on macOS only.
+            #[cfg(target_os = "macos")]
+            ".zshenv",
+        ]
+        .map(|rc| cx.config.homedir.join(rc));
+
+        for rc in &rcs {
+            raw::write_file(rc, FAKE_RC).unwrap();
+        }
+
+        let expected = FAKE_RC.to_owned() + &source(cx.config.cargodir.display(), POSIX_SH);
+
+        cx.config.expect(&INIT_NONE).await.is_ok();
+        // sanity check
+        for rc in &rcs {
+            let new_rc = fs::read_to_string(rc).unwrap();
+            assert_eq!(
+                new_rc, expected,
+                "Rc file {rc:?} does not contain a source after rustup-init",
+            );
+        }
+
+        cx.config
+            .expect(&["rustup", "self", "uninstall", "-y", "--no-modify-path"])
+            .await
+            .is_ok();
+        for rc in &rcs {
+            let new_rc = fs::read_to_string(rc).unwrap();
+            assert_eq!(
+                new_rc, expected,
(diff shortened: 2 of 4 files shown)


src/cli/rustup_mode.rs
```diff
@@ -532,8 +532,13 @@ enum SelfSubcmd {
 
     /// Uninstall rustup
     Uninstall {
+        /// Continue without confirmation
```
cyqsimon: This change is technically unrelated. Let me know if you want it removed from this PR.

cyqsimon:
Not really sure how to properly add the tests that check for environment variable configuration. There doesn't seem to be existing infrastructure for testing this sort of thing.

Also I don't really understand how the current tests are avoiding modifying the user's actual rc files. I imagine the provisions for it must be hidden away somewhere in the test harness but I just couldn't find it.

src/cli/rustup_mode.rs
rami3l: This change is welcomed! However it'd be better if you could split this into a separate commit :)

cyqsimon: Got it. I will rebase and force push if that's okay with you.

rami3l: @cyqsimon Sorry for the back and forth, but it occurred to me that it'd still be better if we mirror the existing text from `rustup-init --help`:

```
rustup-init 1.28.2 :: 1.27.1+1000 (f8dab60b1 2025-10-13)

The installer for rustup

Usage: rustup-init[EXE] [OPTIONS]

Options:
[..]
  -y
          Disable confirmation prompt
[..]
```

rami3l:
> Not really sure how to properly add the tests that check for environment variable configuration. There doesn't seem to be existing infrastructure for testing this sort of thing.
> 
> Also I don't really understand how the current tests are avoiding modifying the user's actual rc files. I imagine the provisions for it must be hidden away somewhere in the test harness but I just couldn't find it.

@cyqsimon That's a completely fine question to ask! I think you can find some inspiration from:

https://github.com/rust-lang/rustup/blob/2bd635656e6e772b7ad3ea59b010018920ffadd6/tests/suite/cli_paths.rs#L239-L267

Basically we use `CliTestContext` to create a sandboxed installation for `rustup` to operate in `TempDir`, so copying-pasting and modifying neighboring tests should do the trick for you.

cyqsimon:
Alright I've added the tests.

When I initially went over the existing tests I was fixated on `tests/suite/cli_self_upd.rs` and found nothing there 😅. But yeah within `cli_paths.rs` it's largely a copy-paste job.

How would you like me to resolve the conflict with #4536? Merge or rebase?

djc:
Rebase please!

cyqsimon:
I have having a very difficult time reproducing the CI error locally (or in fact, making sense of it at all). Will keep trying for the time being.

Edit: seems like there's some shenanigans influencing whether `rustup-init` sets up `.zshenv` or not, which is different between the CI and my local environment. That's all I can say with any degree of confidence. Zsh is not my daily shell so I'm not that familiar with it.

rami3l:
@cyqsimon Many thanks all the same! I think I can have a look at that... I'll let you know if I've found anything 🙏

rami3l:
@cyqsimon That's quite easy actually: in our Ubuntu CI environment there's no `zsh` so the `~/.zshenv` predicate is never triggered. Maybe just remove that from the test case?

cyqsimon:
> @cyqsimon That's quite easy actually: in our Ubuntu CI environment there's no `zsh` so the `~/.zshenv` predicate is never triggered. Maybe just remove that from the test case?

I see. But shouldn't the fix be to install zsh in the CI, so that the relevant code paths are actually tested?

rami3l:
> I see. But shouldn't the fix be to install zsh in the CI, so that the relevant code paths are actually tested?

@cyqsimon I see your point here; however I don't think a typical Linux environment assumes zsh, so this might confuse more users than it should when they are developing rustup on Linux; we want the test suite to have as few surprising external dependencies as possible.

Also I don't think placing a dynamic predicate for both cases is a good choice either, the point being you are almost always testing one specific path in the CI and the opposite path almost never gets tested and could rust over time. This has happened before in this repo's history for a few times, unfortunately...

As a compromise I think you can put a `cfg` for `darwin` on that specific array item. Don't put it for `linux` otherwise you might break CI for other Unices (BSDs, Solaris, ...)!

cyqsimon:
> As a compromise I think you can put a `cfg` for `darwin` on that specific array item. Don't put it for `linux` otherwise you might break CI for other Unices (BSDs, Solaris, ...)!

Yep that sounds reasonable. Will do that.

And what about `uninstall_removes_source_from_rcs` (the test you referenced)? That test is only passing because the lack of an intermediate `assert!` between init and uninstall.

cyqsimon:
Rebased to resolve merge conflicts.

rami3l:
> And what about `uninstall_removes_source_from_rcs` (the test you referenced)? That test is only passing because the lack of an intermediate `assert!` between init and uninstall.

@cyqsimon You can mirror that in a separate commit if you want.

tests/suite/cli_paths.rs
```diff
@@ -266,6 +266,49 @@ error: could not amend shell profile[..]
...
 
+    #[tokio::test]
+    async fn uninstall_doesnt_modify_rcs_with_no_modify_path() {
+        let cx = CliTestContext::new(Scenario::Empty).await;
+        let rcs = [
+            ".bashrc",
+            ".bash_profile",
+            ".bash_login",
+            ".profile",
+            // Zsh is not installed in Linux CI
```
rami3l: Nit: Hmmm as I already said, don't mention Linux here. macOS is the outlier.

cyqsimon: So should I just delete the comment I guess?

rami3l: @cyqsimon I think now that you've added it, it might still make some sense to explain why the exclusion is necessary.

cyqsimon: Can you suggest a wording? I don't think I fully understand what exactly you want.

Currently I got this:

> MacOS has Zsh by default, so we test it there

rami3l: @cyqsimon Yes, `This requires zsh to be installed, so we test it on macOS only.` would be enough. Many thanks in advance!

rami3l (approved):

Issue #4415, opened 2025-07-15
Locally failing test runs


djc:
When I run `cargo test --features test` locally on my macOS laptop, I see a lot of failures:

```
failures:
    suite::cli_self_upd::update_exact
    suite::cli_self_upd::update_no_change
    suite::cli_self_upd::update_precise
    suite::cli_self_upd::updater_leaves_itself_for_later_deletion
    suite::cli_ui::rustup_ui_doc_text_tests
    suite::cli_v1::change_override
    suite::cli_v1::default_existing_toolchain
    suite::cli_v1::expected_bins_exist
    suite::cli_v1::install_override_toolchain_from_archive
    suite::cli_v1::install_override_toolchain_from_channel
    suite::cli_v1::install_override_toolchain_from_version
    suite::cli_v1::install_toolchain_from_archive
    suite::cli_v1::install_toolchain_from_channel
    suite::cli_v1::install_toolchain_from_version
    suite::cli_v1::list_toolchains
    suite::cli_v1::multiple_overrides
    suite::cli_v1::no_update_on_channel_when_date_has_not_changed
    suite::cli_v1::override_overrides_default
    suite::cli_v1::remove_override_no_default
    suite::cli_v1::remove_override_toolchain_err_handling
    suite::cli_v1::remove_override_with_default
    suite::cli_v1::remove_override_with_multiple_overrides
    suite::cli_v1::run_command
    suite::cli_v1::rustc_no_default_toolchain
    suite::cli_v1::update_channel
    suite::cli_v1::update_on_channel_when_date_has_changed
    suite::cli_v2::add_remove_multiple_toolchains
    suite::cli_v2::add_target_v1_toolchain
    suite::cli_v2::bad_manifest
    suite::cli_v2::change_override
    suite::cli_v2::default_existing_toolchain
    suite::cli_v2::expected_bins_exist
    suite::cli_v2::file_override_toolchain_err_handling
    suite::cli_v2::install_allow_downgrade
    suite::cli_v2::install_override_toolchain_from_archive
    suite::cli_v2::install_override_toolchain_from_channel
    suite::cli_v2::install_override_toolchain_from_version
    suite::cli_v2::install_toolchain_from_archive
    suite::cli_v2::install_toolchain_from_channel
    suite::cli_v2::install_toolchain_from_version
    suite::cli_v2::install_with_profile
    suite::cli_v2::list_targets_v1_toolchain
    suite::cli_v2::list_toolchains
[...]

Issue #266, opened 2016-04-05
Clean up output


vks:
Currently the output looks like this:

```
info: syncing channel updates for 'stable'
info: downloading component 'rust'
info: syncing channel updates for 'nightly'
info: downloading component 'rustc'
46.40 MiB / 46.40 MiB (100.00%) 4.57 MiB/s ETA: 0s                
info: downloading component 'rust-std'
58.85 MiB / 58.85 MiB (100.00%) 3.57 MiB/s ETA: 0s                
info: downloading component 'rust-docs'
6.25 MiB / 6.25 MiB (100.00%) 3.00 MiB/s ETA: 0s                
info: downloading component 'cargo'
4.06 MiB / 4.06 MiB (100.00%) 2.37 MiB/s ETA: 0s                
info: installing component 'rustc'
info: installing component 'rust-std'
info: installing component 'rust-docs'
```

The "info" is bold and quite redundant. It actually carries the least information, why is it highlighted?
It would be much nicer to have something like cargo's output, i.e.:

```
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading rust-gmp v0.2.10
 Downloading rustc-serialize v0.3.19
   Compiling rustc-serialize v0.3.19
   Compiling libc v0.2.9
```

Here the verbs are bold and carry the most important information.

brson:
Thanks for the feedback. The `info` is bold because when that word is not `info`, but `error` or `warning` it is styled in bold color. Having all the log-line headers styled specially looks more consistent than just error/warning.

Part of future work is to make the cargo/rustup/rustfmt CLIs more consistent. I do like the way cargo emphasizes a verb, though there are other things about Cargo's UI i dislike.

The thing that bothers me more in your example is the download stats that interrupt the flow of the log.

vks:
Shouldn't errors and warnings be rare? Why not optimize for the common case?

brson:
> Shouldn't errors and warnings be rare? Why not optimize for the common case?

They are rarer than non-errors for sure. You may be right that it would look bettor to not bold 'info:'. It's a point of preference. I took a look at the output without the bolding didn't have a strong opinion either way.

vks:
> It's a point of preference. I took a look at the output without the bolding didn't have a strong opinion either way.

Fair enough. I would probably remove the "info:" entirely and keep the "warning:" and "error:", but I guess this is just bikeshedding.

brson:
[...]

PR #4536, opened 2025-10-15
test: Replace trycmd with snapbox

epage:
This is to prepare for adding styling to help and maybe changing the existing styling (#266).  Maybe this will help with #4415 as well.

The layout I went with is:
- `tests/suite/cli-ui/rustup-init/*.toml` -> `tests/suite/cli_rustup_init_ui.rs` + `tests/suite/cli_rustup_init_ui/*.term.svg`
- `tests/suite/cli-ui/rustup/*.toml` -> `tests/suite/cli_rustup_ui.rs` + `tests/suite/cli_rustup_ui/*.term.svg`


tests/suite/cli_rustup_init_ui.rs
```diff
@@ -0,0 +1,65 @@
...
+use snapbox::Data;
+
+#[track_caller]
+fn test_help(name: &str, args: &[&str]) {
+    let home = Path::new(env!("CARGO_TARGET_TMPDIR")).join("home-ro");
+    create_dir_all(&home).unwrap();
+
+    let rustup_init = cargo_bin!("rustup-init");
+    Command::new(rustup_init)
+        .env("RUSTUP_TERM_COLOR", "always")
```
epage, line 14: I went with `RUSTUP_TERM_COLOR=always`  and `term.svg` tests in prep for adding color to clap's help and errors (blocked on #4520).

tests/suite/cli_rustup_init_ui.rs
```diff
@@ -0,0 +1,65 @@
+use std::fs::create_dir_all;
+use std::path::Path;
+
+use snapbox::cmd::cargo_bin;
+use snapbox::cmd::Command;
+use snapbox::Data;
+
+#[track_caller]
+fn test_help(name: &str, args: &[&str]) {
```
epage, line 8: I consolidated down to 4 helper methods for testing but haven't consolidated down further to give room to see where this goes before trying to see what common threads should be abstracted

rami3l (approved):
This looks fine to me; however we have to recognize the platform differences as well... Do you happen to have a way to overcome the regarding the failed snapshots on Windows? (Some commands are expected to have different output on Windows than on Unix.)

epage:
Sorry, had missed that there were failures.  Hopefully fixed now.

djc:
How do you look at the SVG files locally? What does a test failure look like in a terminal?

rami3l:
> How do you look at the SVG files locally? What does a test failure look like in a terminal?

@djc SVGs are just plain text files so you'll be seeing text file diffs. It's not as good as before in terminals but most of the times we are updating the snapshots anyway. I know from @weihanglo that this has been used in `cargo` for a while and when the styling is more complex this approach will bring more value.

epage:
> How do you look at the SVG files locally?

I use `open` on Linux which uses the default viewer for the file type.  A browser should also work.

> What does a test failure look like in a terminal?

A diff of XML.  If this is limited in its use, then its not bad.  We've not had a problem with it in Cargo.  Unsure how they are handling it with Rustc.

weihanglo:
FWIW: https://github.com/rust-lang/cargo/pull/13461#issuecomment-1953158984

> One pretty minor concern is that GitHub Mobile renders SVG as text. (Who else review PRs on mobile? 🤷🏾‍♂️)

djc:
I definitely review on mobile sometimes. But, let's try it.

tests/suite/cli_rustup_init_ui.rs
```diff
@@ -0,0 +1,65 @@
+use std::fs::create_dir_all;
+use std::path::Path;
+
+use snapbox::Data;
+use snapbox::cmd::Command;
+use snapbox::cmd::cargo_bin;
```
djc: Nit: keep these on the same line?

djc (approved):

rami3l:
> FWIW: [rust-lang/cargo#13461 (comment)](https://github.com/rust-lang/cargo/pull/13461#issuecomment-1953158984)
> 
> > One pretty minor concern is that GitHub Mobile renders SVG as text. (Who else review PRs on mobile? 🤷🏾‍♂️)

@weihanglo Nice remark. Hmmm can we somehow contact the client support for this?

tests/suite/cli_rustup_ui.rs
```diff
@@ -0,0 +1,361 @@
...
+#[cfg(not(windows))] // On windows, we don't have the `man` command
+fn rustup_help_flag() {
+    test_help("rustup_help_flag", &["--help"]);
+}
+
+#[test]
+#[cfg(not(windows))] // On windows, we don't have the `man` command
+fn rustup_only_options() {
+    test_error("rustup_only_options", &["-q"]);
+}
```
epage, line 67: > This looks fine to me; however we have to recognize the platform differences as well... Do you happen to have a way to overcome the regarding the failed snapshots on Windows? (Some commands are expected to have different output on Windows than on Unix.)

epage: So using `...` doesn't work with SVG files because it throws off the `y` variable.

For now, I'm just excluding those tests from Windows.

Alternatives
- We keep only these tests as txt instead of svg

epage:
CI is now passing though unsure how people feel about the approach I took in https://github.com/rust-lang/rustup/pull/4536#discussion_r2436255154

tests/suite/cli_rustup_ui.rs
djc: Seems fine to me.

djc (approved):

djc:
Also, it looks like this solves #4415 so that's pretty nice.

tests/suite/cli_rustup_ui.rs
rami3l: @epage I'm okay with that given that it seems like the best we could do for now.

OTOH what would the recommended approach be if I am using Windows and just want to draft a PR to make the tests pass?

Is there any prior art in that regard? Otherwise I'm thinking about something like uploading the updated snapshots to the CI results for the developer to download...

I can address that in a separate PR, of course.

djc: For that scenario, keeping theese as text instead of SVG would be an improvement. On the other hand, it doesn't feel like we get a lot of external contributors that use Windows, so maybe this is rare enough that it's fine? For those cases, it should also be easy enough for non-Windows team members to contribute an updated SVG.

rami3l: Okay, let's merge it first and see how it goes :)

epage: For Windows developers, you can grab the diff from CI. We have to do this in Cargo for platform-specific tests already.

PR #4538, opened 2025-10-16
process: clean up more terminal stuff

djc:
Following up from

- #4520

cc @epage if you want to take a look.

### DIFF
--- a/src/process/file_source.rs
+++ b/src/process/file_source.rs
@@ -67,39 +67,16 @@ mod test_support {
 
     // ----------------------- test support for writers ------------------
 
-    pub(in super::super) struct TestWriterLock<'a> {
-        inner: MutexGuard<'a, Vec<u8>>,
-    }
-
-    impl Write for TestWriterLock<'_> {
-        fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-            self.inner.write(buf)
-        }
-
-        fn flush(&mut self) -> io::Result<()> {
-            Ok(())
-        }
-    }
-
     pub(in super::super) type TestWriterInner = Arc<Mutex<Vec<u8>>>;
 
     /// A thread-safe test file handle that pretends to be e.g. stdout.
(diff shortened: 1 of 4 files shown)


src/process/file_source.rs
```diff
@@ -67,39 +67,15 @@ mod test_support {
...
-            TestWriterLock {
-                inner: self.0.lock().unwrap_or_else(|e| e.into_inner()),
-            }
-        }
-    }
-
     impl Write for TestWriter {
         fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-            self.lock().write(buf)
+            self.0.lock().unwrap_or_else(|e| e.into_inner()).write(buf)
```
rami3l, line 79: Nit: bring along the deleted comment lines?

src/process/terminal_source.rs
```diff
@@ -23,7 +21,7 @@ pub struct ColorableTerminal {
     // Arc<Mutex<...> which isn't great, but OTOH it is test code. Locking the
```
rami3l: These comments explains that the original code was written like so due to a limitation in `TermColor` which seems to be long gone. Maybe this can be removed as well?

rami3l (approved):
Looks pretty straightforward to review, bravo 🥳

src/process/file_source.rs
djc: Good point, done.

src/process/terminal_source.rs
djc: Yeah, I don't think this makes sense to keep. Removed.

rami3l:
@djc I thought you wanted to wait for @epage?

epage:
Eh, he just CCed me.  I'm good with all of this.

djc:
> @djc I thought you wanted to wait for @epage?

I wanted to let him know and give him the opportunity to give feedback (whether pre- or post-merge), but I don't think we need to block merging on that.

PR #4539, opened 2025-10-16
refactor(log): Single source RUSTUP_TERM_COLOR

epage:
With #4520 merged, we've abstracted away whether ANSI escape codes are supported or not so we don't have to re-decide that for logging.


epage:
>  cli::self_update::tests::default_toolchain_is_stable

Thats strange that its not stripping the ANSI escape codes and only in one test.  Will need to dig into this further.

rami3l:
I think this is clashing with @djc's https://github.com/rust-lang/rustup/pull/4538. Any suggestions?

epage:
Alright, this is now passing CI

djc (approved):

Issue #4446, opened 2025-08-16
No Windows SDK installed when installing Rust on a freshly installed Windows 11


vagreargnatry:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue+is%3Aopen%2Cclosed and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

I just reinstalled Windows 11 Home Edition 24H2. When I installed Rustup, I didn't see the Windows SDK installed. I remember it being installed with Windows SDK 22000. After installing Rust, I tried creating a new project with cargo and then running cargo run. The terminal complained that I was missing link.exe. I then went to Visual Studio 17.14.12 and saw that the only Windows SDKs available were 19041, 22621, and 26100. I chose to install 26100. After the installation, I ran cargo run again, and it worked correctly.

![Image](https://github.com/user-attachments/assets/ab686773-64f8-494d-82e4-c0448e8a1df9)

<img width="2559" height="1527" alt="Image" src="https://github.com/user-attachments/assets/c3d8e2dd-076f-4ba5-9f8b-34d7304341cb" />

<img width="1919" height="1079" alt="Image" src="https://github.com/user-attachments/assets/0ec77ad6-39f3-44e8-9bb8-552938ff93aa" />

### Steps

1. Prepare a Windows 11 system that has not installed Rust and Visudo Studio related content.
2. Double-click rustup-init.exe.
3. Choose to install Visual Studio Community Edition.

### Possible Solution(s)

If I remember correctly, it used to install Windows SDK version 22000. Now it seems it is gone, maybe it's time to choose to install another version of Windows SDK by default.

### Rustup version

```console
rustup 1.28.2 (e4f3ad6f8 2025-04-28)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.89.0 (29483883e 2025-08-04)`
```

### Installed toolchains

```console
Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\shapg\.rustup

installed toolchains
--------------------
stable-x86_64-pc-windows-msvc (active, default)

active toolchain
----------------
name: stable-x86_64-pc-windows-msvc
active because: it's the default toolchain
installed targets:
  x86_64-pc-windows-msvc
```

### OS version

```console
[...]

PR #4542, opened 2025-10-17
Update the default Windows SDK version

ChrisDenton:
This is a temporary workaround for #4446. The problem with the installer is that the version of the SDK we requested was removed due to security issues. Since a new rustup release is near, this PR simply changes to a version that is available as a temporary workaround. It does not fix the underlying problem (that of SDKs potentially being removed) but that is a very unusual scenario so hopefully this will be good enough for the short term.

In any case, Visual Studio 2026 is set for release very soon so I'd rather wait for that before figuring out what to do in the long term.


djc (approved):

djc:
I'm fine with this change but have no particular expertise in Windows SDK, so trusting your judgement!

PR #4543, opened 2025-10-17
Add vs-components.toml

ChrisDenton:
This is just a draft I'm posting to gather feedback and maybe discuss implementation details.

Related to https://github.com/rust-lang/rustup/issues/4446.

The configuration file in this PR tells rustup how to install the needed Visual Studio components. The components are listed per supported host architecture, except that i686 is included with x86_64. The Windows SDK is listed separately because it's independent of architecture and may be skipped if already installed via other means.

I used toml format but this could be json or a simpler `key=value` format but toml seemed more consistent with other files.

This file will need to be deployed to `https://static.rust-lang.org/rustup` independently of the rustup release process.

### DIFF
--- a/vs-components.toml
+++ b/vs-components.toml
@@ -1,0 +1,12 @@
+schema-version = "1"
+installer = "https://aka.ms/vs/17/release/vs_community.exe"
+
+# Component names are listed at
+# https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
+[components]
+x86_64 = ["Microsoft.VisualStudio.Component.VC.Tools.x86.x64"]
+aarch64 = [
+  "Microsoft.VisualStudio.Component.VC.Tools.ARM64",
+  "Microsoft.VisualStudio.Component.VC.Tools.ARM64EC",
+]
+sdk = "Microsoft.VisualStudio.Component.Windows11SDK.26100"


djc (commented):
This makes sense to me.

Nits:

- Name it `vs-components.toml` with a hyphen?
- Maybe `Components` -> `components`?
- How did you come up with the component names? Are there standardized versions of these, either from Rust targets or from the MS ecosystem?

rami3l:
> How did you come up with the component names?

@djc The component names come from https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022 and cannot be changed.

Not sure if that will change with the VS version, but the current format has the flexibility required to reflect such changes if present.

rami3l (approved):

ChrisDenton:
I've updated it to address djc's suggestions and added a link to where component names can be found, for future reference.

rami3l:
@ChrisDenton Thanks! Not sure if `?view=vs-2022` is required though. Will that change someday?

ChrisDenton:
Hm, it isn't strictly necessary. Without it it'll point to the latest stable Visual Studio version which we'd usually want to use. But doing so does require updating the installer url.

ChrisDenton:
I guess the next step is to wait on the new release process and figure out how publishing this will fit in since it needs to be a separate step (like publishing the website would ideally be),

rami3l:
@ChrisDenton Yes, I do hope that pushing https://github.com/rust-lang/promote-release/pull/84 will eventually give us more freedom on that matter; that said, I have the feeling that I might not know who to contact yet if we want to change the structure of the release repository... Do you have any idea?

cc @Mark-Simulacrum

ChrisDenton:
We can probably ask on the infra channel on zulip. But it may be worth focusing on getting the main release process done first.

rami3l:
@ChrisDenton I came across this script today: https://gist.github.com/mmozeiko/7f3162ec2988e81e56d5c4e22cde9977 and it seems to have indicated a way of getting what SDK versions are available out there... Could this be of interest for resolving our problem here?

ChrisDenton:
I'm aware of that method, I believe xwin uses something similar. However, I'm not too keen on parsing channel manifests seeing as how they're not documented anywhere (as far as I'm aware) and are an internal feature of the Visual Studio installer.

My current plan is to use `C:\Program Files (x86)\Microsoft Visual Studio\Installer\setup.exe` to export a list of components. This has documentation <https://learn.microsoft.com/en-us/visualstudio/install/import-export-installation-configurations?view=visualstudio#vsconfig-file-format>. The only downside is that it requires installing Visual Studio first then querying for the missing component. But I think that's a minor issue. If it's an issue we can always attempt to use the known components and only fallback if that fails.

ChrisDenton:
For anyone following along I create a prototype application that installs any missing components if Visual Studio is already installed: https://github.com/ChrisDenton/vssetup/blob/install/crates/install-msvc/src/main.rs

If anyone want to test it you can download a pre-built binary [here](https://github.com/ChrisDenton/vssetup/actions/runs/22241907478/artifacts/5596521615).

PR #4546, opened 2025-10-18
config: clean up override handling

djc:
Trying to clean up override handling a bit, and avoiding reprocessing the active toolchain in proxy mode.


rami3l (approved):

rami3l:
@djc Huge thanks for having spotted all these places to make the codebase even cleaner!

PR #4549, opened 2025-10-19
Expand `RUSTUP_TOOLCHAIN_SOURCE`'s documentation

smoelius:
As suggested by @weihanglo in https://github.com/rust-lang/rustup/pull/4518#discussion_r2442721847.

### DIFF
--- a/doc/user-guide/src/environment-variables.md
+++ b/doc/user-guide/src/environment-variables.md
@@ -72,7 +72,7 @@
 - `RUSTUP_CONCURRENT_DOWNLOADS` *unstable* (default: the number of components to download). Controls the number of
   downloads made concurrently.
 
-- `RUSTUP_TOOLCHAIN_SOURCE` *unstable*. Set by rustup to tell proxied tools how `RUSTUP_TOOLCHAIN` was determined.
+- `RUSTUP_TOOLCHAIN_SOURCE` *unstable*. Set by rustup to tell proxied tools how `RUSTUP_TOOLCHAIN` was determined. Non-rustup tools should not set this environment variable, except insofar as to mirror an earlier invocation from rustup.
 
 [directive syntax]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
 [dc]: https://docs.docker.com/storage/storagedriver/overlayfs-driver/#modifying-files-or-directories


djc:
Should we add a caveat like "except insofar as to mirror an earlier invocation from rustup"?

rami3l (approved):
LGTM modulo https://github.com/rust-lang/rustup/pull/4549#issuecomment-3419556630.

weihanglo (approved):
Thanks!

smoelius:
I've made the change since that seems to be the consensus. But, personally, I think it's a bad idea to include the caveat.

First, there is no demonstrated need. The situation described in https://github.com/rust-lang/rustup/pull/4518#discussion_r2442721847 is hypothetical. I think it is safer to not include caveats until there is a demonstrated need for them.

And second, it's not clear to me how it would work. Currently, the proxy sets `RUSTUP_TOOLCHAIN_SOURCE`. So, for a tool to set `RUSTUP_TOOLCHAIN_SOURCE`, it would have to set it and call the proxied tool directly, not through the rustup proxy. Is that what we expect?

rami3l:
@smoelius Indeed, if recursive calls are in place then this caveat currently serves no purpose; however IIRC what you described (bypassing rustup's toolchain resolution) was exactly the scenario @weihanglo has pointed out.

This gets me thinking if recursive calls have a role to play in this story and, if they do, which one. I personally think it has to do with what cargo wants from this feature (cc @weihanglo).

smoelius:
Please merge if you are happy with the changes.

weihanglo:
> First, there is no demonstrated need. The situation described in https://github.com/rust-lang/rustup/pull/4518#discussion_r2442721847 is hypothetical.

That kind of use cases I believe are mostly and more likely in enterprise environments, rather than in a public space. And I assure it is not hypothetical. 

The only use case right now is <https://github.com/rust-lang/cargo/pull/16131>, which would incur an imprecise warning in `cargo install` if wrapper doesn't propogate `RUSTUP_TOOLCHAIN_SOURCE` (which is not too bad to be honest).

> This gets me thinking if recursive calls have a role to play in this story and, if they do, which one.

It depends on what kind of recursive call, I guess? If they call something like `cargo +nightly build` in a `build.rs`, then there is nothing Cargo can/should do, as the code path will just be handled by normal rustup toolchain resolution. If they are setting `RUSTUP_TOOLCHAIN` directly, then we know they are doing weird thing, so either they should set `RUSTUP_TOOLCHAIN_SOURCE` along with that, or do whatever they want.

weihanglo:
> Currently, the proxy sets RUSTUP_TOOLCHAIN_SOURCE. So, for a tool to set RUSTUP_TOOLCHAIN_SOURCE, it would have to set it and call the proxied tool directly, not through the rustup proxy.

Not exactly, The rust-analyzer case is like

1. Run `rustc --print sysroot`. The `rustc` here is actually a rustup proxy
2. Retrieve the sysroot path and set it in `RUSUTP_TOOLCHAIN` to ensure the same toolchain is called <https://github.com/rust-lang/rust-analyzer/pull/16563>.


rust-analyzer itself doesn't really call proxies directly IIUC. Granted, the way rust-analyzer dealing with toolchain proxies has changed a couple of times and I might already have the outdated understanding.

PR #4551, opened 2025-10-20
feat(cli): Add color to clap help/errors

epage:
This adds styling to clap help and errors to make it easier to scan.  This uses Cargo's styling (via `clap-cargo`) for consistency across the project.
- For consistency, my main priorities are end-user commands like `rustup` and `cargo` and not plumbing commands like `rustfmt` and `rustc`.
- We've not yet decided how to handle this through an official crate, so doing it unofficially for now.

To ensure consistency, the Discussion sections were updated.
- I was limited in my use of color within Discussion to match at least the initial approach we've taken in Cargo. I worry that if we color every literal, placeholder, etc then it will be a sea of colors and lose the benefit.  Instead color is limited to items we want to draw extra attention to or visually big items to help set them apart
- Cargo doesn't have sub-headers, so I had to define that myself for `rustup completions`
- I could have used something like `color-print` to do the styling at compile time but that would require duplicating Cargo's styling which would have been harder to maintain.  Instead, we style at runtime so we can use `clap-cargo`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ otel = [
 ]
 
 # Exports code dependent on private interfaces for the integration test suite
-test = ["dep:snapbox", "dep:walkdir"]
+test = ["dep:snapbox", "dep:walkdir", "clap-cargo/testing_colors"]
 
 # Sorted by alphabetic order
 [dependencies]
@@ -47,6 +47,7 @@ anyhow = "1.0.69"
 cfg-if = "1.0"
 chrono = { version = "0.4", default-features = false, features = ["std"] }
 clap = { version = "4", features = ["derive", "wrap_help"] }
+clap-cargo = "0.18.2"
 clap_complete = "4"
 console = "0.16"
 curl = { version = "0.4.44", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ anstyle = "1.0.11"
 anyhow = "1.0.69"
 cfg-if = "1.0"
 chrono = { version = "0.4", default-features = false, features = ["std"] }
-clap = { version = "4", features = ["derive", "wrap_help"] }
+clap = { version = "4", features = ["derive", "wrap_help", "string"] }
 clap-cargo = "0.18.2"
 clap_complete = "4"
 console = "0.16"
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -1,5 +1,8 @@
-pub(crate) fn rustup_help() -> &'static str {
-    r"Discussion:
+use clap_cargo::style::{HEADER, LITERAL, PLACEHOLDER};
+
+pub(crate) fn rustup_help() -> String {
+    format!(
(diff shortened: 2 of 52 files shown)


Cargo.toml
```diff
@@ -105,6 +105,7 @@ zstd = { version = "0.13", default-features = false }
 # test only (depends on `test` feature)
 snapbox = { version = "0.6.22", features = ["term-svg"], optional = true }
 walkdir = { version = "2", optional = true }
+clap-cargo = "0.18.1"
```
djc: Nit: this is not test only, so let's move it up, between clap and clap_complete?

djc (approved):
Looks great, thanks!

rami3l (approved):
We might indeed want to address https://rust-lang.zulipchat.com/#narrow/channel/490103-t-rustup/topic/More.20rustup.2Fcargo.2Frustc.20consistency/near/545410015 in addition.

Otherwise this looks pretty nice, thanks!

epage:
Everything should now be styled.

Cargo.toml
```diff
@@ -37,7 +37,7 @@ otel = [
 ]
 
 # Exports code dependent on private interfaces for the integration test suite
-test = ["dep:snapbox", "dep:walkdir"]
+test = ["dep:snapbox", "dep:walkdir", "clap-cargo/testing_colors"]
```
epage, line 40: This ensures consistency across platforms for testing because rustc's colors (which Cargo aims to be consistent with) has workarounds for default shells for specific platforms

src/cli/help.rs
```diff
@@ -1,5 +1,8 @@
-pub(crate) fn rustup_help() -> &'static str {
-    r"Discussion:
+use clap_cargo::style::*;
```
djc: Nit: I'd prefer to only import the symbols we need.

djc (approved):

src/cli/help.rs
epage: Posted #4553 to automate this review feedback

src/cli/help.rs
```diff
@@ -1,5 +1,8 @@
+use anstyle::Style;
 use clap_cargo::style::{HEADER, LITERAL, PLACEHOLDER};
 
+const SUBHEADER: Style = Style::new().bold();
```
djc: Nit: sorry, one more that I forgot in the last round: please put this at the bottom.

djc (approved):

src/cli/help.rs
epage: Went ahead and did this but this is one of the rare cases where it feels better at the top to me.

In general, the way I view organizing a source file is it should act like a tree with the top item(s) at the top and branch out from there, so the top item(s) act as a ToC and provide context.  In that case, constants are noise.  However, in this case, this has all of the styling elements together and now we're splitting them up.

djc: I don't find that a very strong argument. Lots of modules where some items are imported while other items are defined near the bottom.

PR #4552, opened 2025-10-20
Handle the progress reporting of installations through progress bars

FranciscoTGouveia:
Cherry-picked from #4471, in response to [#4471 (comment)](https://github.com/rust-lang/rustup/pull/4471#issuecomment-3418470314).

### DIFF
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -194,7 +194,7 @@ impl<'a> DownloadCfg<'a> {
         let progress = ProgressBar::hidden();
         progress.set_style(
             ProgressStyle::with_template(
-                "{msg:>12.bold}  [{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
+                "{msg:>12.bold}  [{bar:30}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
             )
             .unwrap()
             .progress_chars("## "),
@@ -260,7 +260,7 @@ impl DownloadStatus {
         *retry_time = None;
         self.progress.set_style(
             ProgressStyle::with_template(
-                "{msg:>12.bold}  [{bar:40}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
+                "{msg:>12.bold}  [{bar:30}] {bytes}/{total_bytes} ({bytes_per_sec}, ETA: {eta})",
             )
             .unwrap()
             .progress_chars("## "),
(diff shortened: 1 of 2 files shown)


rami3l (approved):
LGTM!

@djc To clarify, this PR introduces two messages, `pending installation` for a toolchain waiting to be installed after the download phase, and a spinner to replace the original `installing component...` line.

src/dist/download.rs
```diff
@@ -269,10 +269,10 @@ impl DownloadStatus {
 
     pub(crate) fn finished(&self) {
         self.progress.set_style(
-            ProgressStyle::with_template("{msg:>12.bold}  downloaded {total_bytes} in {elapsed}")
+            ProgressStyle::with_template("{msg:>12.bold}  pending installation {total_bytes:>10}")
                 .unwrap(),
         );
-        self.progress.finish();
+        self.progress.tick(); // A tick is needed for the new style to appear, as it is static.
```
djc, line 275: Don't we still want to `finish()` after the `tick()`?

djc (commented):
This seems to have a bunch of smaller changes in a single commit. It would be nice to split them apart so that each is individually motivated. (IIRC it was two commits in #4471 in the first place.)

src/dist/download.rs
FranciscoTGouveia: No, the component will still be installed, and its progress will be reported.
Since we want to have only one progress bar per component -- rather than two (one for downloading and another for installing) -- we cannot finish it immediately.

We were calling `finish()` previously because we were not reporting the installs through the progress bars, but rather through [here](https://github.com/rust-lang/rustup/pull/4552/files#diff-aa51ee662316e678d658bb58b30e2de77d486689843c8532419f5f8fd54fe1aeL747-L757).

FranciscoTGouveia:
> This seems to have a bunch of smaller changes in a single commit. It would be nice to split them apart so that each is individually motivated. (IIRC it was two commits in #4471 in the first place.)

The two new progress bar states introduced (`installing()` and `installed()`) are closely related, so I don’t believe they can be separated into distinct commits.

Additionally, the original PR (4471) contained two commits because it introduced a special case for components that were already downloaded -- which I do not think is necessary.
When installing, the size of the component’s download is already available, and for components that are already installed, that value is simply `0 B`.

A screenshot can be seen below to illustrate this:

image

However, I made some changes for the different possible states of the progress bars to be column-aligned and I think those would benefit being in a separate commit.

djc (approved):

djc:
> When installing, the size of the component’s download is already available, and for components that are already installed, that value is simply `0 B`.

I think it might be better to retain the component's size or drop the size, since `0 B` seems kinda wrong.

PR #4553, opened 2025-10-20
style: Remove wildcard imports

epage:
This is based on review feedback from #4551.

I experimented with a `clippy.toml` file for "blessing" some of these, like `errors`, but for some reason I wasn't able to get it to work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,6 +152,7 @@ unused_qualifications = "warn"
 # `dbg!()` and `todo!()` clearly shouldn't make it to production:
 dbg_macro = "warn"
 todo = "warn"
+wildcard_imports = "warn"
 
 [lib]
 name = "rustup"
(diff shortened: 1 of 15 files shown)


djc (approved):
Thanks!

rami3l (approved):
LGTM modulo one question: does this change include common exceptions such as `super::*` in inline tests, or `foo::prelude::*`? I think we still want those to be excluded.

epage:
By default there are exceptions, from https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports

> Wildcard imports are allowed from modules that their name contains prelude. Many crates (including the standard library) provide modules named “prelude” specifically designed for wildcard import.
>
> Wildcard imports reexported through pub use are also allowed.
>
> use super::* is allowed in test modules. This is defined as any module with “test” in the name.
>
> These exceptions can be disabled using the warn-on-all-wildcard-imports configuration flag.

djc:
I guess this needs a rebase? Lot of issues in the merge queue CI.

epage:
Rebased

PR #4554, opened 2025-10-21
feat(cli/help): discuss `rustup toolchain install`

rami3l:
Addresses https://github.com/rust-lang/rustup/pull/4513#discussion_r2405339953.

### DIFF
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -118,6 +118,13 @@ pub(crate) fn toolchain_help() -> String {
     )
 }
 
+pub(crate) fn toolchain_install_help() -> String {
+    format!(
+        r"{HEADER}Discussion:{HEADER:#}
+{TOOLCHAIN_INSTALL_HINT}"
+    )
+}
+
 pub(crate) fn toolchain_link_help() -> String {
     format!(
         r"{HEADER}Discussion:{HEADER:#}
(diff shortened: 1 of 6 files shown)


djc (approved):
This is great! Should we also show this for `update`?

Tiny nit: I would use lowercase after ":".

rami3l:
@djc Thanks for the review! I've addressed the concerns accordingly along with a tiny fix that turns out to be a leftover from #3596.

rami3l:
@epage Remaining concern from https://github.com/rust-lang/rustup/pull/4551: do you think it's a good idea to indent all discussions by 2 spaces instead of 4, or are you sure that `clap` segments are indented by 2 spaces? The current style (see the svg diffs in this PR) is a bit confusing...

epage:
Clap indents sections by 2 spaces but for flags, that two spaces is for where the short flags is placed.

https://github.com/clap-rs/clap/blob/c7c761f988684ad97c8b2c521b05cf7f8192b3eb/clap_builder/src/output/mod.rs#L21

https://github.com/clap-rs/clap/blob/c7c761f988684ad97c8b2c521b05cf7f8192b3eb/clap_builder/src/output/help_template.rs#L518-L520

https://github.com/clap-rs/clap/blob/c7c761f988684ad97c8b2c521b05cf7f8192b3eb/clap_builder/src/output/help_template.rs#L1071-L1072

epage:
Note that that we reduced the padding between clap v3 and v4, see https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#400---2022-09-28

djc (approved):

rami3l:
> Note that that we reduced the padding between clap v3 and v4, see https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#400---2022-09-28

@epage Ah I see why the current indentation is what it is. Maybe that deserves another PR to fix? For this patch in particular let's just stay with the old style then.

PR #4556, opened 2025-10-22
fix: Use HEADER styling in 'rustup show' 

epage:
This is continuing my effort to make end-user commands more similar. This isn't meant to force conformancy where it doesn't make sense. My assumption is to first get things similar and then work on if there is feedback. Cargo is doing that with rust; there are things we want to change but priority is on being the same before we try to change. At least for myself, I don't feel anything is sacred and can be re-evaluated.

image

This is setup to match `cargo info`s style:
image
- Top-level keys are treated as headers
- Parentheticals are "context"

For testing, I forked an existing test rather than using SVGs for all of the existing tests.  Having the output inline with the existing tests makes it easier to reason about and the number of test cases relevant for checknng styling is much smaller than for the main logic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -328,9 +328,9 @@ dependencies = [
 
 [[package]]
 name = "clap-cargo"
-version = "0.18.2"
+version = "0.18.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e9c7f07f2b3c9fc1a21da5b4fa268083ac4d4d145a711b286055fdd70d149fcf"
+checksum = "936551935c8258754bb8216aec040957d261f977303754b9bf1a213518388006"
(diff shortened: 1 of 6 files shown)


rami3l (approved):
Thanks! I think that does look better than before...

PS: I'm not completely sure if people are really parsing the output or not, but probably this is less of a problem from our experience in the v1.28 breakage, especially because it's just a change in the colors... Let's see :)

epage:
If they are parsing, then it isn't a tty and they won't have colors unless the someone has set `RUSTUP_TERM_COLOR`.

The idea I've been toying with to try to improve scanability that would affect parsers is to see if we can align on the `-` between version and platform where present.

djc (approved):

PR #4557, opened 2025-10-23
fix(update): Match 'cargo update's colors

epage:
This is adjusting "channel updates" to use `cargo update` style colors like we did with `rustup show` and `cargo info` in #4556.


rami3l (approved):
LGTM :)

PR #4558, opened 2025-10-23
fix(list): Match show command's styling 

epage:
This changes the parenthetical from `bold` to `CONTEXT` to match what
was done in #4556.

This does not remove the bolding of the item itself.
We could do that. I figured its fine for now since there is a clearer
meaning with `installed` than when you can have `active` and `default`
on separate items.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -251,7 +251,7 @@ pub(super) fn list_items(
     let bold = Style::new().bold();
     for (name, installed) in items {
         if installed && !installed_only && !quiet {
-            writeln!(t, "{bold}{name} (installed){bold:#}")?;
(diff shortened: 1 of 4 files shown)


epage:
Looks like some of the output ordering is non-determinstic which makes testing this more difficult.  snapbox has `.unordered()` but that doesn't work well for SVG.

Is there a reason the output hasn't been made deterministic?  I think in Cargo, the only time our output is non-deterministic is when dealing with threads.

rami3l:
@epage I think we might have some sort of misunderstanding here; it's not the output being nondeterministic, rather it's the output being platform-dependent. While we redact the host triple (which will of course be platform-dependent), the output is still strictly in alphanumerical order.

Can we, say, sort the redacted output before sending it to comparison?

epage:
Alright, here's hoping that this works

rami3l (approved):

PR #4561, opened 2025-10-23
fix(check): Respect RUSTUP_TERM_COLOR  in `rustup check`

epage:
`rustup check` was using `console::style` which will do its own "can I
style" checks, overriding `RUSTUP_TERM_COLOR`.

See https://docs.rs/console/0.16.1/src/console/utils.rs.html#696-702


rami3l (approved):
It turns out I have somehow missed this point from #4388... Many thanks for fixing that!

PR #4562, opened 2025-10-24
refactor(cli/rustup-mode): merge self-update checking logic in `rustup update`

rami3l:


src/cli/rustup_mode.rs
```diff
@@ -925,12 +916,17 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<ExitCode> {
...
-        && self_update_mode == SelfUpdateMode::Enable
-        && !opts.no_self_update;
+    let self_update =
+        CFG_SELF_UPDATE && self_update_mode == SelfUpdateMode::Enable && !opts.no_self_update;
 
-    if self_update && check_rustup_update(&DownloadCfg::new(cfg)).await? {
-        update_available = true;
+    if self_update {
+        // Print the separating newline only if toolchain updates were found.
+        if update_available {
```
rami3l: @djc Is that the right place to add a newline, or doesn't the multi-progress bar subsume a newline at the end?

src/cli/self_update.rs
```diff
@@ -979,12 +979,15 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     Ok(())
 }
 
+/// Whether the self-update functionality is enabled during compilation.
+pub(crate) const CFG_SELF_UPDATE: bool = !cfg!(feature = "no-self-update");
```
djc, line 983: This is redoing something that I undid only recently. I don't see much value in it -- it just obscures the original meaning and saves a few characters.

src/cli/rustup_mode.rs
```diff
@@ -1021,13 +1021,15 @@ async fn update(
...
 
-    if CFG_SELF_UPDATE && self_update_mode == SelfUpdateMode::CheckOnly {
-        check_rustup_update(&dl_cfg).await?;
-    }
-
-    if !CFG_SELF_UPDATE {
-        info!("self-update is disabled for this build of rustup");
-        info!("any updates to rustup will need to be fetched with your system package manager")
+    match self_update_mode {
+        _ if !CFG_SELF_UPDATE => {
```
djc, line 1025: If this is the reason, I think you can `cfg!()` instead of an attribute.

src/cli/self_update.rs
rami3l: @djc I still think the double negation here in `!cfg!("no-*")` is a bit confusing, but you can totally have different opinions.

src/cli/rustup_mode.rs
rami3l: Sorry for not having explained the full background until now, but when looking at https://github.com/rust-lang/rustup/issues/4565 I realized that the self-update cleanup code is very curiously written in two parts, so I want to merge them together first to make the overall logic clearer.

However, since this change is not directly related to #4565, I planned on making two separate PRs...

rami3l: @djc I'm not entirely sure about your proposals, but maybe this could be simplified to just the following?

```rust
    if cfg!(feature = "no-self-update") {
        info!("self-update is disabled for this build of rustup");
        info!("any updates to rustup will need to be fetched with your system package manager")
    } else if self_update_mode == SelfUpdateMode::CheckOnly {
        check_rustup_update(&dl_cfg).await?;
    }
```

... the idea is that this block covers all the cases where `self_update()` is not executed, and when I add the extra `if self_update` block in a second PR, it'll be clearer when each case is handled.

Or you might have different opinions regarding the fix?

rami3l: I've made https://github.com/rust-lang/rustup/pull/4566 to hopefully make my original intentions clearer.

rami3l:
Closing in favor of #4566.

PR #4563, opened 2025-10-24
fix(check): Use Cargo's colors

epage:
With `rustup check`s colors being testable (#4561), this updates it to match Cargo's colors

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -850,13 +850,16 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<ExitCode> {
             MultiProgress::with_draw_target(cfg.process.progress_draw_target());
         let semaphore = Arc::new(Semaphore::new(concurrent_downloads));
         let channels = tokio_stream::iter(channels.into_iter()).map(|(name, distributable)| {
+            let msg = format!("{bold}{name} - {bold:#}");
+            let status = "Checking...";
+            let template = format!("{{msg}}{status} {{spinner:.green}}");
             let pb = multi_progress_bars.add(ProgressBar::new(1));
             pb.set_style(
-                ProgressStyle::with_template("{msg:.bold} - Checking... {spinner:.green}")
+                ProgressStyle::with_template(template.as_str())
                     .unwrap()
                     .tick_chars(r"|/-\ "),
             );
-            pb.set_message(format!("{name}"));
+            pb.set_message(msg.clone());
             pb.enable_steady_tick(Duration::from_millis(100));
 
             let sem = semaphore.clone();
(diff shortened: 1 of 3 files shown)


epage:
Looks like a hang in the build and CI needs to be re-kicked off

epage:
Now a test failure that I can't reproduce
```

thread 'download::tests::reqwest::socks_proxy_request' panicked at src/download/tests.rs:179:13:
  left: 0
 right: 1
```

src/cli/rustup_mode.rs
```diff
@@ -825,23 +825,13 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<ExitCode> {
         Style::new()
     };
 
-    let error = if use_colors {
-        AnsiColor::Red.on_default().bold()
-    } else {
-        Style::new()
-    };
+    let transient = if use_colors { TRANSIENT } else { Style::new() };
```
djc: Suggest doing it more like this:

```rust
let (transient, error, good, warn) = match use_colors {
     true => (TRANSIENT, ERROR, GOOD, WARN),
     false => (Style::new(), Style::new(), Style::new(), Style::new()),
};
```

epage: Went ahead and did this with `if`.  I just now noticed you used a `match`.  Is there a reason you went with that?  Seems odd to me to use that with a straight `bool`

djc (approved):

src/cli/rustup_mode.rs
```diff
@@ -819,20 +819,18 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<ExitCode> {
         Some(_) | None => channels_len,
     };
 
-    let bold = if use_colors {
-        Style::new().bold()
+    let (bold, transient, error, good, warn) = if use_colors {
+        (Style::new().bold(), TRANSIENT, ERROR, GOOD, WARN)
     } else {
-        Style::new()
+        (
```
rami3l: Nit: Actually this can be simplified to `Default::default()` :)

rami3l (approved):
LGTM :)

I'm soooo sorry about missing this one, I seem to have been buried in a whole lot of GitHub notifications 🙇

src/cli/rustup_mode.rs
rami3l: @epage I think that's a matter of personal preference... I also like to use `if-else` in this case despite `match` being one line shorter.

src/cli/rustup_mode.rs
djc: Interesting! Maybe `<(Style, Style, Style, Style, Style)>::default()` instead?

src/cli/rustup_mode.rs
djc: I like that it's one line shorter, and that the arms end up vertically next to each other.

src/cli/rustup_mode.rs
epage: Huh, I had assumed `Default` wasn't implemented for tuples.

PR #4566, opened 2025-10-25
fix(cli/rustup-mode): add missing self-update in `rustup toolchain install`

### LINKED ISSUE #4565: `rustup toolchain install` doesn't consistently trigger self update
When going through `update()`'s logic, it seems that both `rustup update` and `rustup toolchain install foo` will do one of the following:

1. Perform self-update based on runtime settings.
2. Issue a warning when self-update is disabled at compile time.

However this is not true for `rustup toolchain install` with no arguments (corresponding to the branch at L1021), since the first case is impossible:

https://github.com/rust-lang/rustup/blob/59aec45394e945e7f60b348ade84b9b58baf0985/src/cli/rustup_mode.rs#L1017-L1028

This could be considered as an inconsistency.

rami3l:
Closes #4565, supersedes #4562.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -946,13 +946,7 @@ async fn update(
 
     common::warn_if_host_is_emulated(cfg.process);
     let self_update_mode = SelfUpdateMode::from_cfg(cfg)?;
-    // Priority: no-self-update feature > self_update_mode > no-self-update args.
-    // Update only if rustup does **not** have the no-self-update feature,
-    // and auto-self-update is configured to **enable**
-    // and has **no** no-self-update parameter.
-    let self_update = !cfg!(feature = "no-self-update")
-        && self_update_mode == SelfUpdateMode::Enable
-        && !opts.no_self_update;
+    let should_self_update = !opts.no_self_update;
     let force_non_host = opts.force_non_host;
     cfg.profile_override = opts.profile;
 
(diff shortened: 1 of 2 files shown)


src/cli/rustup_mode.rs
```diff
@@ -962,6 +955,29 @@ async fn update(
...
+            SelfUpdateMode::Enable if !opts.no_self_update => {
+                return self_update::self_update(&dl_cfg).await;
+            }
+            SelfUpdateMode::CheckOnly => {
+                check_rustup_update(&dl_cfg).await?;
+            }
+            _ => (),
+        }
+        Ok(ExitCode(0))
+    };
```
djc: This seems nice. Can this now just be an `async fn` that lives in the `self_update` module? (Presumably in another separate commit.)

rami3l: @djc Nice catch. I've folded that logic into `self_update::self_update()` because here is that function's only call site.

src/cli/self_update.rs
```diff
@@ -1125,8 +1125,34 @@ pub(crate) fn self_update_permitted(explicit: bool) -> Result<SelfUpdatePermissi
...
+///
+/// Whether the self-update is executed is based on both compile-time and runtime
+/// configurations, where the priority is as follows:
+/// no-self-update feature > self update mode > no-self-update args.
+///
+/// i.e. update only if rustup does **not** have the no-self-update feature,
+/// and self update mode is configured to **enable**
+/// and has **no** no-self-update parameter.
+pub(crate) async fn self_update(
+    dl_cfg: &DownloadCfg<'_>,
```
djc: Nit: suggest keeping `dl_cfg` last (since it's a less specific argument for this function).

Consider making this a `SelfUpdateMethod` method, maybe `update()` or `run()` or `execute()`?

djc (approved):
Very nice!

src/cli/self_update.rs
```diff
@@ -1126,15 +1125,41 @@ pub(crate) fn self_update_permitted(explicit: bool) -> Result<SelfUpdatePermissi
...
+    mode: SelfUpdateMode,
+    no_self_update: bool,
```
ChrisDenton: I'm not sure about this being a double negative. Mistakes have been made in other code bases due to confusion over such things, especially in combination with boolean operators. E.g. `!no_self_update` should be read as "not not self-update" which isn't ideal.

rami3l: I'm okay with reworking on that.

PS: Interestingly, the first version of this PR (https://github.com/rust-lang/rustup/pull/4562) explicitly works to address some double negatives in a PR and got dropped 🤔

src/cli/self_update.rs
```diff
@@ -280,6 +280,56 @@ impl SelfUpdateMode {
             Self::CheckOnly => "check-only",
         }
     }
+
+    /// Optionally performs a self-update: check policy, download, apply and exit.
```
djc, line 284: This ("move to a method") doesn't really make sense as a separate commit, since the function was added in this PR anyway. Just squash this?

rami3l: @djc Hmmm `self_update()` was an existing method. Otherwise I'd have squashed it since the beginning.

djc: Ah, fair enough.

djc (approved):

Issue #4159, opened 2025-01-17
Threaded FS IO probably should be async now?


rami3l:
> Being able to use async where relevant across the codebase makes it possible to use async constructs where relevant. rustup has many places where that can be useful: downloading channel metadata, distribution content, unpacking that content to disk, are work that could usefully proceed in parallel - and async provides a good abstraction for that.
> 
> We already have a sophisticated disk IO layer that accomodates various OS latency-inducing behaviours, and adapting that to async without any regressions could be very interesting too - but for now, it co-exists nicely with an async core.
_https://github.com/rust-lang/rustup/pull/3367_

We are still using an async-unaware threadpool here, and it's known to produce problems (https://github.com/rust-lang/rustup/issues/3125):

https://github.com/rust-lang/rustup/blob/fa4ae320550847f3b14c20023b8b042c91c5cc3e/src/diskio/threaded.rs#L1-L6

Anyway, using `RUSTUP_IO_THREADS=1` to limit concurrency feels a bit off. Will migrating to `async` help?

djc:
IMO the code we have for this looks very complicated/scary. Rewriting it looks like a project, would probably be hard to know if we bring over all the desired properties. `async` could be helpful but won't make this easy by itself IMO.

rami3l:
Ah, I do remember the days when installing Rust took a long time to finish: https://github.com/rust-lang/rustup/pull/1876

(cc the original author, if I'm not mistaken, @rbtcollins)

rbtcollins:
Yah, so the underlying thing here is that rustup does tasks that look very much like offensive action to heuristic detectors that are wired into the syscall path on windows; and it also does a large number of operations that have data dependencies. And we have to run in very resource constrained environments, but also unpack very large files.

## For the first case, more details:
- we write executable files, which includes HTML for docs
- these get scanned at CloseHandle time by a defender filter driver except when the non-user-settable scan-on-open setting is set. Microsoft have opted rustup *into* that setting, though it might pay to check its still active - I don't install Rust on windows every day :).
- even when inline scanning isn't happening, `CloseHandle` latency is much higher than `close` latency because of the differing IO models in Windows (process owns dirty pages) vs Linux / Un*x (pagecache owns dirty pages).

So for this case what we do is quite easy: we make sure that CloseHandle is in a thread of its own **and continue with other IO while that takes place**. tokio's io threads would be entirely suitable for this.

## Second case:
[...]

PR #4567, opened 2025-10-25
Remove lifetime parameters to allow multi-threaded installations

FranciscoTGouveia:
Cherry-picked from https://github.com/rust-lang/rustup/pull/4471, in response to https://github.com/rust-lang/rustup/pull/4471#issuecomment-3418470314.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1127,7 +1127,7 @@ pub(crate) fn self_update_permitted(explicit: bool) -> Result<SelfUpdatePermissi
 }
 
 /// Performs all of a self-update: check policy, download, apply and exit.
-pub(crate) async fn self_update(dl_cfg: &DownloadCfg<'_>) -> Result<utils::ExitCode> {
+pub(crate) async fn self_update(dl_cfg: &DownloadCfg) -> Result<utils::ExitCode> {
     match self_update_permitted(false)? {
         SelfUpdatePermission::HardFail => {
             error!("Unable to self-update.  STOP");
@@ -1144,7 +1144,7 @@ pub(crate) async fn self_update(dl_cfg: &DownloadCfg<'_>) -> Result<utils::ExitC
         return run_update(setup_path);
     } else {
         // Try again in case we emitted "tool `{}` is already installed" last time.
-        install_proxies(dl_cfg.process)?;
+        install_proxies(&dl_cfg.process)?;
     }
 
     Ok(utils::ExitCode(0))
@@ -1242,7 +1242,7 @@ fn parse_new_rustup_version(version: String) -> String {
     String::from(matched_version)
 }
 
-pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> Result<Option<PathBuf>> {
+pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg) -> Result<Option<PathBuf>> {
     let cargo_home = dl_cfg.process.cargo_home()?;
     let rustup_path = cargo_home.join(format!("bin{MAIN_SEPARATOR}rustup{EXE_SUFFIX}"));
     let setup_path = cargo_home.join(format!("bin{MAIN_SEPARATOR}rustup-init{EXE_SUFFIX}"));
@@ -1265,10 +1265,10 @@ pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> Result<Option<Pa
     // If someone really wants to use another version, they still can enforce
     // that using the environment variable RUSTUP_OVERRIDE_HOST_TRIPLE.
     #[cfg(windows)]
-    let triple = TargetTriple::from_host(dl_cfg.process).unwrap_or(triple);
+    let triple = TargetTriple::from_host(&dl_cfg.process).unwrap_or(triple);
 
     // Get update root.
-    let update_root = update_root(dl_cfg.process);
+    let update_root = update_root(&dl_cfg.process);
 
     // Get current version
     let current_version = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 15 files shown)


rami3l (approved):
LGTM, thanks :)

djc:
Did you both miss my comment from the other PR opposing these changes? Or did I miss some response to that?

rami3l:
> Did you both miss my comment from the other PR opposing these changes? Or did I miss some response to that?

@djc You have been giving very important and insightful review comments, but I believe we might be having some sort of a misunderstanding here. If you meant https://github.com/rust-lang/rustup/pull/4471#pullrequestreview-3353605509:

> So while I did say I was fine with doing spawning/channel approach, after looking at the "remove some lifetime parameters" commit I am no longer sure if that's the best approach.
> 
> Effectively it means that whoever goes and puts in place the right architecture will basically have to undo all of those changes which IMO should not be necessary.

The latest update of #4471 doesn't use channels anymore (it's not finished as the refactor should be squashed back), but removing lifetimes are still required to send the current transaction to a new task, since without spawning (say up to https://github.com/rust-lang/rustup/pull/4471/commits/b5e9e4d7559a53351c6ad7d4bba5caddf204f1fb), the feature shouldn't work on it's own and no performance boost is introduced.

@FranciscoTGouveia has, in fact, asked me the very question upon finishing the first commit wondering why the performance was not improved; we deduced together and concluded that a separate thread for installation is required because most of it is still sync and, as we have previously agreed, the installer itself will stay sync for now (https://github.com/rust-lang/rustup/issues/4159).

I've always wanted to minimize the diff to land the change, so I understand your concern of certain commits being too huge to revert and/or a move in the wrong direction. Of course we could be missing some crucial factors in your envisioned solution if the changes proposed in this PR are really not required. In that case we'd be happy to know.

rami3l:
PS: Also, on a general sense, tokio seems to dislike the lifetime dance so I doubt whether changes like this won't spread in the future despite being a bit unfortunate.

djc:
> The latest update of #4471 doesn't use channels anymore (it's not finished as the refactor should be squashed back), but removing lifetimes are still required to send the current transaction to a new task, since without spawning (say up to [b5e9e4d](https://github.com/rust-lang/rustup/commit/b5e9e4d7559a53351c6ad7d4bba5caddf204f1fb)), the feature shouldn't work on it's own and no performance boost is introduced.

Ah, cool. I was referring also to https://github.com/rust-lang/rustup/pull/4471#discussion_r2385119158:

> If that's the case, I would suggest something more like a FuturesUnordered containing all the downloads, and then pulling from that to run the installations.

So I'm suggesting that downloading happens concurrently by storing a bunch  `ComponentBinary::download()` thingies in a `FuturesUnordered`, then the main thread/task can pull from those to do installations one at a time.

rami3l:
> So I'm suggesting that downloading happens concurrently by storing a bunch `ComponentBinary::download()` thingies in a `FuturesUnordered`

@djc IIRC the latest update of #4471 is doing exactly that.

Are you suggesting something like `task::block_in_place()` should be used instead of `task::spawn_blocking()` to avoid sending the installer to another thread?

I'm asking this because we have failed to move downloads to another thread via a `JoinSet` or similar. It's giving errors like ``` `the trait `std::marker::Sync` is not implemented for `RefCell<Option<&mut CoreWrapper<CtVariableCoreWrapper<..., ..., ...>>>>` ```, not sure what is happening either.

### Update

It looks like the above error comes from here:

https://github.com/rust-lang/rustup/blob/d676c6f00768779bd2ff8f415cb240f16e0abe11/src/download/mod.rs#L92-L101

Either way it seems like it's about doing the remove refcell/lifetime dance either with the downloading or the installing code.

djc:
> Are you suggesting something like `task::block_in_place()` should be used instead of `task::spawn_blocking()` to avoid sending the installer to another thread?

I'm suggesting we don't need to send the installer to another thread at all, because the downloads will continue for a bit in the background even if the main task is not polling them.

rami3l:
> I'm suggesting we don't need to send the installer to another thread at all, because the downloads will continue for a bit in the background even if the main task is not polling them.

@djc Given a sync function call blocking a certain thread, how can other tasks make progress if we don't have other threads to use?

If you are talking about sending downloading tasks to other threads first (instead of sending the installer), then we are on the same page now. However, as explained in https://github.com/rust-lang/rustup/pull/4567#issuecomment-3448580560, it's not immediately possible either since it requires a dance similar to this PR.

Of course I can experiment on that direction and see which approach will result in a smaller diff.

djc:
> @djc Given a sync function call blocking a certain thread, how can other tasks make progress if we don't have other threads to use?

I'm pretty sure you're modelling async I/O wrong in your head. Once we have initiated a download, TCP frames are going to continue to come in (modulo flow control and congestion management). Thus, we don't need to continuously poll the I/O tasks in order for the downloads to make progress, we can execute on an installation and then poll the downloads again.

> If you are talking about sending downloading tasks to other threads first (instead of sending the installer), then we are on the same page now. However, as explained in [#4567 (comment)](https://github.com/rust-lang/rustup/pull/4567#issuecomment-3448580560), it's not immediately possible either since it requires a dance similar to this PR.

No, as I've tried to explain several times now I'm suggesting we don't send any tasks to other threads.

rami3l:
@djc Thanks for your clarification. It is true that the downloads that have been initiated will not be influenced, but the expected behavior also includes the normal initiation of new downloads when a component is being unpacked (which is very important when large components such as `rust-docs` are being installed and small components are being downloaded), as well as download progress reporting in that period.

My confusion is that it seems this cannot be handled using the naive blocking approach, but of course I'd be glad to be corrected.

djc:
> It is true that the downloads that have been initiated will not be influenced, but the expected behavior also includes the normal initiation of new downloads when a component is being unpacked (which is very important when large components such as `rust-docs` are being installed and small components are being downloaded), as well as download progress reporting in that period.

Okay, that makes a little more sense. I submitted #4570 which substantially simplifies the core goal of concurrent downloading & installation. I suggest we use that as a starting point to figure out what the effect is with large installations like `rust-docs` and how best to mitigate those effects.

rami3l:
Closing as superseded by https://github.com/rust-lang/rustup/pull/4605, however this approach has played a very important part in proving the necessity of `spawn_blocking()` uses, so the work did not end up in vain.

cc https://github.com/rust-lang/rustup/pull/4471#issuecomment-3564581077

PR #4570, opened 2025-10-27
Concurrent downloads & installation

djc:
This is what I had in mind. It seems to compile but I've done no testing or benchmarking.

@FranciscoTGouveia maybe that's something you can have a look at?

Alternative to

- #4471 

Obsoletes

- #4567

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -11,7 +11,6 @@ use futures_util::stream::StreamExt;
 use std::sync::Arc;
 use tokio::sync::Semaphore;
 use tracing::{info, warn};
-use url::Url;
 
 use crate::dist::component::{Components, DirectoryPackage, Transaction};
 use crate::dist::config::Config;
@@ -151,8 +150,6 @@ impl Manifestation {
             }
         }
 
-        let altered = tmp_cx.dist_server != DEFAULT_DIST_SERVER;
-
         // Download component packages and validate hashes
         let mut things_to_install = Vec::new();
         let mut things_downloaded = Vec::new();
@@ -188,17 +185,7 @@ impl Manifestation {
             let sem = semaphore.clone();
             async move {
                 let _permit = sem.acquire().await.unwrap();
-                let url = if altered {
-                    utils::parse_url(
-                        &bin.binary
-                            .url
-                            .replace(DEFAULT_DIST_SERVER, tmp_cx.dist_server.as_str()),
-                    )?
-                } else {
-                    utils::parse_url(&bin.binary.url)?
-                };
-
-                bin.download(&url, download_cfg, max_retries, new_manifest)
+                bin.download(download_cfg, max_retries, new_manifest)
                     .await
                     .map(|downloaded| (bin, downloaded))
             }
@@ -697,16 +684,16 @@ struct ComponentBinary<'a> {
 impl<'a> ComponentBinary<'a> {
     async fn download(
         &self,
-        url: &Url,
         download_cfg: &DownloadCfg<'_>,
         max_retries: usize,
         new_manifest: &Manifest,
     ) -> Result<File> {
         use tokio_retry::{RetryIf, strategy::FixedInterval};
 
+        let url = download_cfg.url(&self.binary.url)?;
         let downloaded_file = RetryIf::spawn(
             FixedInterval::from_millis(0).take(max_retries),
-            || download_cfg.download(url, &self.binary.hash, &self.status),
+            || download_cfg.download(&url, &self.binary.hash, &self.status),
             |e: &anyhow::Error| {
                 // retry only known retriable cases
                 match e.downcast_ref::<RustupError>() {
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -160,6 +160,8 @@ impl Manifestation {
                     component,
                     binary,
                     status: download_cfg.status_for(component.short_name(new_manifest)),
+                    manifest: new_manifest,
+                    download_cfg,
                 })
             })
             .collect::<Result<Vec<_>>>()?;
@@ -185,7 +187,7 @@ impl Manifestation {
             let sem = semaphore.clone();
             async move {
                 let _permit = sem.acquire().await.unwrap();
-                bin.download(download_cfg, max_retries, new_manifest).await
+                bin.download(max_retries).await
             }
         });
         if components_len > 0 {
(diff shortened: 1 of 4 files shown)


FranciscoTGouveia:
It compiles, but I believe there is an infinite loop; at least, I could not get any toolchain installation to finish as it seemed to hang on the download of the last component.

Additionally, I would like to note that when a component is being installed, although ongoing downloads continue normally, new component downloads cannot start.
As a result, during an installation (assuming `RUSTUP_CONCURRENT_DOWNLOADS=2`, which is the default), no more than one download is active at a time.

This isn’t an issue when `RUSTUP_CONCURRENT_DOWNLOADS` is configured to maximize concurrency, but otherwise, it can have a noticeable impact on performance (I wasn’t able to benchmark this due to the infinite loop).

Finally, I would like to point out that running both downloads and installations on the same thread prevents the progress bars for other downloads from updating (see [here](https://github.com/rust-lang/rustup/pull/4567#issuecomment-3448708164)).

src/dist/manifestation.rs
```diff
@@ -255,15 +214,40 @@ impl Manifestation {
...
         }
 
-        // Install components
-        for (component_bin, installer_file) in things_to_install {
-            tx = component_bin.install(installer_file, tx, new_manifest, self, download_cfg)?;
+        let mut downloads = FuturesUnordered::new();
+        let mut component_iter = components.iter();
+        let mut cleanup_downloads = vec![];
+        loop {
+            while downloads.len() < concurrent_downloads {
```
FranciscoTGouveia: This loop will run _ad aeternum_.

I believe the issue is that after the penultimate component finishes, `downloads.len()` remains 1.
Since the default value for `concurrent_downloads` is 2, the condition is never satisfied, and the loop never exits.

Even if `RUSTUP_CONCURRENT_DOWNLOADS` is overridden to `1`, the issue persists; the last component will complete its download and installation, but the loop will still not exit.

src/dist/manifestation.rs
```diff
@@ -242,15 +214,40 @@ impl Manifestation {
             tx = self.uninstall_component(component, new_manifest, tx, download_cfg.process)?;
         }
 
-        // Install components
-        for (component_bin, installer_file) in things_to_install {
-            tx = component_bin.install(installer_file, tx, new_manifest, self, download_cfg)?;
+        let mut downloads = FuturesUnordered::new();
+        let mut component_iter = components.iter();
+        let mut cleanup_downloads = vec![];
+        loop {
```
rami3l, line 221: @djc Thanks for the testbed!

After careful comparisons, I think this is equivalent to [`b5e9e4d` (#4471)](https://github.com/rust-lang/rustup/pull/4471/commits/b5e9e4d7559a53351c6ad7d4bba5caddf204f1fb) once we squash [`5800cba` (#4471)](https://github.com/rust-lang/rustup/pull/4471/commits/5800cba76419fcdee5100086b0d3a005b99ec410) back onto it to eliminate the use of channels, modulo some edge cases such as empty todo list and download failures.

The core of this approach includes directly issuing a blocking call to what we have pulled from a `FuturesUnordered`. When that blocking call is running, there will be no progress updates and no new installations being initiated. So this is the same as what we've had after the very first step of #4471.

src/dist/manifestation.rs
djc: Thanks for investigating, will fix this.

src/dist/manifestation.rs
djc: Thanks for looking into. Do you think it's useful to merge (something like) this and proceed from there?

> When that blocking call is running, there will be no progress updates and no new installations being initiated.

If this is the main issue that's left, it feels like there might be easier (and better) ways to solve this.

djc:
I've fixed this up so that it can at least pass basic manual tests. CI is failing because I changed the log output -- happy to fix this up if we decide that we want to move forward with this.

src/dist/manifestation.rs
rami3l: @djc I think that is THE remaining issue, yes.

However I still have concerns regarding the aforementioned edge cases that I and @FranciscoTGouveia have found out when working on #4471 so more careful checks have to be carried out.

I think it'd be nice if both of us can have another look at this PR to ensure the semantics are right, it might take some time though while I try to coordinate 🙇

djc: > However I still have concerns regarding the aforementioned edge cases that I and @FranciscoTGouveia have found out when working on #4471 so more careful checks have to be carried out.
> 
> I think it'd be nice if both of us can have another look at this PR to ensure the semantics are right, it might take some time though while I try to coordinate 🙇

Sounds great!

src/dist/manifestation.rs
```diff
@@ -242,15 +216,41 @@ impl Manifestation {
...
+            if downloads.is_empty() && component_iter.len() == 0 {
+                break;
+            }
+
+            let (bin, downloaded) = match downloads.next().await {
+                Some(Ok((bin, downloaded))) => (bin, downloaded),
+                Some(Err(e)) => return Err(e),
+                None => {
+                    if let Some(bin) = component_iter.next() {
```
rami3l: This seems inconsistent with the `while` section below in the number of tasks being added to the todo list.

djc: Fair point. I've tweaked the code to improve on this.

src/dist/manifestation.rs
```diff
@@ -240,15 +214,38 @@ impl Manifestation {
...
-            tx = component_bin.install(installer_file, tx, self)?;
+        let mut downloads = FuturesUnordered::new();
+        let mut component_iter = components.iter();
+        let mut cleanup_downloads = vec![];
+        loop {
+            if downloads.is_empty() && component_iter.len() == 0 {
+                break;
+            }
+
+            let installable = match downloads.next().await {
```
rami3l: Remark: I think this is the same as `let installable = downloads.next().await.transpose()?;`.

rami3l (approved):
I think this works out pretty well during both code inspection and manual tests, so I consider this to be a somewhat faithful reproduction of the first step of #4471. Given that, I agree with merging this despite the fact that the project will suffer from temporary UX regressions.

I'd like to invite @FranciscoTGouveia do another round of review from his side to make sure everything is retained from the original solution.

src/dist/manifestation.rs
djc: Nice, fixed!

djc:
Here's what I think we could do next to solve the remaining UX problem and even further improve the UX:

- Update the progress indicator during unpacking, using a [`ProgressBarIter`](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBarIter.html) wrapper
- Update the progress indicator during installing, by making installation "async", maybe by calling [`yield_now()`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) every 250ms or so from `DirectoryPackage::install()`

(A more principled approach might be to use `tokio::fs` but because there's a lot of layers in our install process that's probably a bunch of work, if not exactly hard.)

PR #4573, opened 2025-10-28
Simplify unpacking abstractions

djc:


rami3l (approved):
Looks reasonable to me, thanks!

PR #4581, opened 2025-11-03
Change toolchain override message from info to warn

### LINKED ISSUE #4504: Make message about rust-toolchain.toml a warning

In our Github CI for Ratatui, we have a setup which installs rust beta using dtolnay/rust-toolchain, which eventually delegates to rustup. We anticipated that this would cause cargo to run using the beta toolchain, but missed that rust-toolchain overrides this. There is an existing message that highlights this problem in rustup, but I think this should likely be written as a warning level rather than an info level message:

https://github.com/ratatui/ratatui/actions/runs/18035682655/job/51321994095#step:3:84

```
info: note that the toolchain 'stable-x86_64-unknown-linux-gnu' is currently in use (overridden by '/home/runner/work/ratatui/ratatui/rust-toolchain.toml')
```

To be clear, this is about making a problematic configuration simple and obvious, it's not about changing how the rustup / rust-toolchain.toml things work.

### Solution you'd like

Change this log message to an warning to make it more visible in logs

### Notes

_No response_

joshka:
Problem:
rustup currently logs an info-level message when a rust-toolchain.toml file overrides the active toolchain. This message is easy to miss in CI logs, leading to confusion when the expected toolchain (e.g., beta) isn’t being used.

Fix:
Promote the message about toolchain override (rust-toolchain.toml) from info to warn level, making it more visible during builds and CI runs.

Fix https://github.com/rust-lang/rustup/issues/4504

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -790,7 +790,7 @@ async fn default_(
         if let Some((toolchain, source)) = cfg.active_toolchain()?
             && !matches!(source, ActiveSource::Default)
         {
-            info!(
+            warn!(
                 "note that the toolchain '{toolchain}' is currently in use ({})",
                 source.to_reason()
             );
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -324,7 +324,7 @@ async fn default_override() {
         .with_stderr(snapbox::str![[r#"
 info: using existing install for stable-[HOST_TRIPLE]
 info: default toolchain set to stable-[HOST_TRIPLE]
-info: note that the toolchain 'nightly-[HOST_TRIPLE]' is currently in use (directory override for '[..]')
+warn: note that the toolchain 'nightly-[HOST_TRIPLE]' is currently in use (directory override for '[..]')
 
 "#]])
         .is_ok();


djc:
I don't think this PR makes sense. Since the behavior is intended, displaying this as a warning violates a common expectation of what a warning means in most logging.

joshka:
> I don't think this PR makes sense. Since the behavior is intended, displaying this as a warning violates a common expectation of what a warning means in most logging.

I don't think that's the case here. This should be a warning because it's easy to miss that when you set the default toochain, but have an override then setting the default actually has no effect. This message is warning the user user that this configuration is nonsense and won't do what you expect from running commands that alter the default toolchain.

This is a very real problem that came from thinking that we were running some tasks against the beta toolchain, but actually weren't. Warnings in logging are often used to mark places like this that are possibly a problem but may not be, hence this is not an error.

djc:
> This should be a warning because it's easy to miss that when you set the default toochain, but have an override then setting the default actually has no effect. This message is warning the user user that this configuration is nonsense and won't do what you expect from running commands that alter the default toolchain.

From my perspective, this message is informing the user that rustup is executing on its configuration. I don't think rustup can decide whether that configuration is nonsense or not.

joshka:
Strong opinion, weakly held disagreement with your perspective here.
Happy to close as I understand the problem.

djc:
> Strong opinion, weakly held disagreement with your perspective here. Happy to close as I understand the problem.

What part do you disagree with? How would we make the warning not annoying to people who have intentionally set up an override?

joshka:
I believe this is a footgun. It's not an error, but it's an easy misconfiguring that rustup can and should highlight.

The ability to install a default rust toolchain when you're in a context where that default will not be used should be an annoying warning even when this is intentional as using a tool like dtolnay/rust-toolchain

I can see how having any warnings that you can't disable might be also a problem. But it's not an error, and it will never fail a ci process to see this warning. The ability to explicitly opt in to warnings as errors for compilation is something to bear in mind here.

Warnings that help ensure that things are ok are good warnings. Making them info level makes them bad warnings.

As mentioned, I'm happy to defer to your wider context on this though as it's unlikely that I'll personally make this mistake again and it may be me over indexing on prevention of something that happens rarely. (That said, I just submitted a PR at work where I would have hit this exact problem had I not been aware of it).

djc:
@rami3l @ChrisDenton thoughts?

rami3l:
@djc This is a footgun, I agree with @joshka, but given that the normal `rustup` semantics is just `rustup +<active>`, this is indeed the expected behavior, so no warnings should be issued (the original proposal is #1648).

On the other hand, it could be that sometimes people would want `rustup +<default>`. #4504 could be such a case. If it is, then @ChrisDenton could potentially redirect it to his [Zulip topic](https://rust-lang.zulipchat.com/#narrow/channel/490103-t-rustup/topic/Could.20we.20support.20.60.2Bdefault.60.20toolchain.3F/with/542518920). Maybe a new issue can be made, or even an existing issue can be found?

ChrisDenton:
I too agree this is a footgun. I think this is another instance where the needs of CI differ from the needs of local users. I do wonder if we should publish our own GitHub action because I think this sort of setup/configuration is best handled by that, though I do accept that would be a bigger job.

rami3l:
> I too agree this is a footgun. I think this is another instance where the needs of CI differ from the needs of local users. I do wonder if we should publish our own GitHub action because I think this sort of setup/configuration is best handled by that, though I do accept that would be a bigger job.

@ChrisDenton Indeed. In this hypothetical CI job we can always specify the toolchain in the shell-env-var level (or an explicit `none` for the old behavior). Hopefully this will work great with [the current precedence](https://rust-lang.github.io/rustup/overrides.html) since I don't quite believe methods with lower precedence such as `rustup override` make that much sense in CI.

PR #4583, opened 2025-11-03
Take semver-compatible dependencies

djc:
Replaces

- #4582

### DIFF
--- a/tests/suite/cli_rustup_init_ui.rs
+++ b/tests/suite/cli_rustup_init_ui.rs
@@ -2,14 +2,14 @@ use std::fs::create_dir_all;
 use std::path::Path;
 
 use snapbox::Data;
-use snapbox::cmd::{Command, cargo_bin};
+use snapbox::cmd::Command;
 
 #[track_caller]
 fn test_help(name: &str, args: &[&str]) {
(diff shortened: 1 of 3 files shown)


tests/suite/cli_rustup_init_ui.rs
```diff
@@ -2,14 +2,14 @@ use std::fs::create_dir_all;
 use std::path::Path;
 
 use snapbox::Data;
-use snapbox::cmd::{Command, cargo_bin};
+use snapbox::cmd::Command;
```
ChrisDenton, line 5: It seems like a diagnostic bug that the import counts as a "use" of the function even though only the macro version is used. Still, not much else we can do about that here.

ChrisDenton (approved):

tests/suite/cli_rustup_init_ui.rs
djc: I'm guessing that importing prioritizes the function namespace over the macro namespace, or something? It feels a little iffy to expose a macro and function by the same name in the same module. (Although I guess we have a similar pattern for derive macros and traits...)

djc: Filed

- https://github.com/rust-lang/rust/issues/148426

Issue #4442, opened 2025-08-11
rustup won't force-skip unavailable `rust-std` for targets specified through `rust-toolchain.toml`


Pizza989:
### Problem you are trying to solve

I have a no_std project that uses a custom target and the `rust-toolchain.toml` file. I have specified the custom target with the `toolchain.targets` key of the `rust-toolchain.toml` file. I have uninstalled all toolchains at this point so when i run cargo commands rustup tries installing one according to the config file, however this will always fail because `rust-std` doesn't exist for the custom target.
```
error: component 'rust-std' for target '<custom_target>' is unavailable for download for channel 'stable'
If you don't need the component, you could try a minimal installation with:

    rustup toolchain add stable --profile minimal

If you require these components, please install and use the latest successful build version,
which you can find at <https://rust-lang.github.io/rustup-components-history>.

After determining the correct date, install it with a command such as:

    rustup toolchain install nightly-2018-12-27

Then you can use the toolchain with commands such as:

    cargo +nightly-2018-12-27 build

```
The workaround for this is to install the toolchain for the host target first. Then, when running a cargo command, it will now say: `warning: Force-skipping unavailable component 'rust-std-<custom_target>` but proceed with the installation of i.e. extra components anyways. Finally it won't be an issue because it's a `no_std` environment anyways. 

### Solution you'd like

I would like a flag to disable the rust-std component from the `rust-toolchain.toml` file as in this particular use-case, human intervention might not be wanted if it runs as part of CI or similar.

### Notes

If someone implements this enhancement i will suggest updating the hint it gives in the first error message i included where it tells you to set the profile to "minimal" since that will still include `rust-std`

rami3l:
@Pizza989 Thanks for reporting!

I am sorry, but I'm not completely clear about your use case.

What is this custom target to be precise? Why would you want to specify it in `rust-toolchain.toml`?

tage64:
I had a similar usecase where I wanted to use the `sparc-unknown-none-elf` target for a project. It makes sence to specify it in `rust-toolchain.toml` because then the user doesn't have to add the toolchain with rustup manually.

rami3l:
@tage64 Thanks! I'll have a deeper look at this to see if this is a bug in the codebase or an enhancement is required.

rami3l:
@cachebag I think this can be considered a bug.

Could you find a way to port the force-skipping logic to the installation process directed by `rust-toolchain.toml`?

cachebag:
[...]

PR #4585, opened 2025-11-03
fix(install): force-skip behavior for `rust-std` 

cachebag:
Draft for #4442

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -20,21 +20,21 @@ pub(crate) enum UpdateStatus {
     Unchanged,
 }
 
-pub(crate) enum InstallMethod<'a> {
+pub(crate) enum InstallMethod<'cfg, 'a> {
     Copy {
         src: &'a Path,
         dest: &'a CustomToolchainName,
-        cfg: &'a Cfg<'a>,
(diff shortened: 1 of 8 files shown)


cachebag:
@djc 

**As a note**
I did end up attempting to reproduce #4442 on my MacBook using my branch. The environment was configured with a clean `RUSTUP_HOME` and `CARGO_HOME`, and the custom build (--no-default-features) behaved as expected without triggering the original failure.



### stable `rustup` with `skip_std` flag
image

<br>

### with patch applied
image


# 

I hope I understood this correctly. I believe I am still doing something wrong with the propogation as even a `false` flag triggers the new behavior
image
<br>

But I'd like to defer to you guys before I continue.

src/install.rs
```diff
@@ -31,7 +31,7 @@ pub(crate) enum InstallMethod<'a> {
         dest: &'a CustomToolchainName,
         cfg: &'a Cfg<'a>,
     },
-    Dist(DistOptions<'a>),
+    Dist(DistOptions<'a>, bool),
```
djc: Probably better to keep this inside `DistOptions`?

PR #4586, opened 2025-11-04
dist: improve cohesion for downloads/installation

djc:
From reviewing

- #4585

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -1,7 +1,13 @@
 //! Installation from a Rust distribution server
 
 use std::{
-    collections::HashSet, env, fmt, io::Write, ops::Deref, path::Path, str::FromStr, sync::LazyLock,
+    collections::HashSet,
+    env, fmt,
+    io::Write,
+    ops::Deref,
+    path::{Path, PathBuf},
+    str::FromStr,
+    sync::LazyLock,
 };
 
 use anyhow::{Context, Result, anyhow, bail};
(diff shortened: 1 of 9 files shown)


src/dist/mod.rs
```diff
@@ -877,45 +877,68 @@ impl fmt::Display for Profile {
...
+
+impl<'a> DistOptions<'a> {
+    pub(super) fn new(
+        components: &'a [&'a str],
+        targets: &'a [&'a str],
+        toolchain: &'a ToolchainDesc,
+        profile: Profile,
+        force: bool,
+        cfg: &'a Cfg<'_>,
+    ) -> Result<Self> {
```
ChrisDenton, line 913: Hm, I'm slightly nervous about the number of arguments here, especially as the first two have the same type and `force` is just a `bool` (making it not immediately obvious what it means without looking it up). Does the new function pull its weight? Looking at the current uses it seems fine as they're mostly passing arguments on but I did want to query this.

If we had some kind of partial `Default` I'd suggest using that but we don't.

djc: Have a look at the rest of the commits I just pushed. I believe this shows that doing this obviates some intermediate layers that do essentially the same thing (same type and `bool` arguments). I think that means this function does pull its weight. Ideally we'd create the `DistOptions` directly from `clap::Parse` values where possible...

ChrisDenton: Ah, then that seems fine. I would love if there's a better option but there's no reason to block this on that.

ChrisDenton (approved):

PR #4587, opened 2025-11-04
feat(install): `add skip_std` flag to allow skipping `rust-std` installation for custom targets

### LINKED ISSUE #4442: rustup won't force-skip unavailable `rust-std` for targets specified through `rust-toolchain.toml`

cachebag:
Closes #4442 
Revision of #4585

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -780,7 +780,7 @@ async fn default_(
             MaybeResolvableToolchainName::Some(ResolvableToolchainName::Official(toolchain)) => {
                 let desc = toolchain.resolve(&cfg.get_default_host_triple()?)?;
                 let status = cfg
-                    .ensure_installed(&desc, vec![], vec![], None, force_non_host, true)
+                    .ensure_installed(&desc, vec![], vec![], None, force_non_host, true, false)
                     .await?
                     .0;
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -56,6 +56,7 @@ struct ToolchainSection {
     components: Option<Vec<String>>,
     targets: Option<Vec<String>>,
     profile: Option<String>,
+    skip_std: Option<bool>,
 }
 
 impl ToolchainSection {
@@ -64,6 +65,7 @@ impl ToolchainSection {
             && self.components.is_none()
             && self.targets.is_none()
             && self.path.is_none()
+            && self.skip_std.is_none()
     }
 }
 
@@ -137,6 +139,7 @@ enum OverrideCfg {
         components: Vec<String>,
         targets: Vec<String>,
         profile: Option<Profile>,
+        skip_std: bool,
     },
 }
 
@@ -180,6 +183,7 @@ impl OverrideCfg {
             ToolchainName::Official(desc) => {
                 let components = file.toolchain.components.unwrap_or_default();
                 let targets = file.toolchain.targets.unwrap_or_default();
+
                 Self::Official {
                     toolchain: desc,
                     components,
@@ -190,6 +194,7 @@ impl OverrideCfg {
                         .as_deref()
                         .map(Profile::from_str)
                         .transpose()?,
+                    skip_std: file.toolchain.skip_std.unwrap_or(false),
                 }
             }
             ToolchainName::Custom(name) => Self::Custom(name),
@@ -213,6 +218,7 @@ impl From<ToolchainName> for OverrideCfg {
                 components: vec![],
                 targets: vec![],
                 profile: None,
+                skip_std: false,
             },
             ToolchainName::Custom(name) => Self::Custom(name),
         }
@@ -737,6 +743,7 @@ impl<'a> Cfg<'a> {
                 components,
                 targets,
                 profile,
+                skip_std,
             } = override_config
             {
                 self.ensure_installed(
@@ -746,6 +753,7 @@ impl<'a> Cfg<'a> {
                     profile,
                     force_non_host,
                     verbose,
+                    skip_std,
                 )
                 .await?;
             } else {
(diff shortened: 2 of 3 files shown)


src/config.rs
```diff
@@ -180,16 +182,18 @@ impl OverrideCfg {
             ToolchainName::Official(desc) => {
                 let components = file.toolchain.components.unwrap_or_default();
                 let targets = file.toolchain.targets.unwrap_or_default();
+                let profile = file
```
djc: This seems like an unrelated change? Please leave that out.

src/dist/mod.rs
```diff
@@ -909,6 +911,7 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
         toolchain: &'a ToolchainDesc,
         profile: Profile,
         force: bool,
+        skip_std: bool,
```
djc: Suggest to leave this out, and only have callers set it separately after the initialization.

djc (commented):
So I'm not sure yet how/where the `ToolchainSection::skip_std` field makes it into the `DistOptions`?

cachebag:
@djc 
>So I'm not sure yet how/where the ToolchainSection::skip_std field makes it into the DistOptions?

Yeah, sorry about that. I think it should be fine now, the flag flows through the chain where it should be applied as `options.skip_std = skip_std` before being used to skip the component in `try_update_from_dist_()`.

src/cli/rustup_mode.rs
```diff
@@ -775,7 +775,7 @@ async fn default_(
             MaybeResolvableToolchainName::Some(ResolvableToolchainName::Official(toolchain)) => {
                 let desc = toolchain.resolve(&cfg.get_default_host_triple()?)?;
                 let status = cfg
-                    .ensure_installed(&desc, vec![], vec![], None, force_non_host, true)
+                    .ensure_installed(&desc, vec![], vec![], None, force_non_host, true, false)
```
cachebag, line 783: @djc I think I might be confused with this one- how do you suggest I handle this? I don't know if it's best here to add an attribute, ignoring it.

djc: I'm not sure what you mean. Do you have more context on what the problem is/why you're not sure?

cachebag: @djc I'm sorry, I think I get it..So if I understand correctly, I'm supposed to remove the extra parameter completely from `ensure_installed` and only set `skip_std` on `DistOptions` in the caller before installation?

djc: Current usage seems okay -- you need to pass it in somehow.

djc:
(You'll need to rebase and squash before this can be merged.)

src/config.rs
```diff
@@ -800,6 +810,7 @@ impl<'a> Cfg<'a> {
             false,
             self,
         )?;
+        options.skip_std = skip_std;
```
rami3l, line 813: I'm sorry, but this feels very strange. Why are other args passed to  the constructor, but this one updated from the outside?

@djc What do you think?

rami3l:
Thanks a lot for your contribution and sorry for having taken so long to respond, but this seems to have indicated some bigger changes to be done beforehand, made pretty visible by the extra large number of parameters... I'll find some time to have another look.

src/config.rs
djc: I think I basically suggested this in https://github.com/rust-lang/rustup/pull/4587#discussion_r2491261138? It seems nicer to me to not bother all callers with supplying a default value.

rami3l: @djc Oh I'm sorry for commenting on the suggested revision, but due to an web-related problem (website crash) it seems that I cannot properly see what that comment was trying to address 🙇 Closing this thread as resolved now.

rami3l:
As I understand it, the original issue #4442 actually describes an inconsistency between `rustup toolchain install` and `rustup target add`. Although an extra flag is definitely reasonable to be added, I still think it would be better if the user basically doesn't need to do anything and the installation of `rust-std` is automatically skipped, because that brings the overall experience of the former closer to the latter which is what we would like to see.

I'm saying this because rustup can trigger toolchain installations in many places, and adding a flag regarding installation is like playing a round of whack-a-mole here.

cachebag:
> As I understand it, the original issue #4442 actually describes an inconsistency between `rustup toolchain install` and `rustup target add`. Although an extra flag is definitely reasonable to be added, I still think it would be better if the user basically doesn't need to do anything and the installation of `rust-std` is automatically skipped, because that brings the overall experience of the former closer to the latter which is what we would like to see.
> 
> 
> 
> I'm saying this because rustup can trigger toolchain installations in many places, and adding a flag regarding installation is like playing a round of whack-a-mole here.

I did feel awkward writing this patch. So I presume this change actually propagates a bit deeper throughout the codebase than what my stopgap currently does? 

Should I just iterate on this PR or would you like me to open a new one?

rami3l:
@cachebag Given the analysis in https://github.com/rust-lang/rustup/issues/4442#issuecomment-3529753775 I'm not 100% sure of what to do next, but this PR wouldn't make it there anyway.

As such I'm closing it for now, many thanks all the same!

PR #4588, opened 2025-11-07
Be more consistent about aliases for different subcommands

joshtriplett:
In some cases we have the same subcommand in multiple places, but it accepts different aliases in each place. For instance, you can write either `rustup toolchain uninstall` or `rustup toolchain rm`, but you can only write `rustup uninstall`, not `rustup rm`.

Add aliases for consistency:
- `rustup add` for `rustup install`, for consistency with `rustup toolchain install`.
- `rustup rm` (or "remove" or "delete" or "del") for `rustup uninstall`, for consistency with `rustup toolchain uninstall`.
- `rustup self up` for `rustup self update`, for consistency with other `update` commands.
- `rustup toolchain up` for `rustup toolchain update`, for consistency with other `update` commands.

I've also added `rustup status` as an alias for `rustup show`, for consistency with `git status`. :smile:

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -123,14 +123,14 @@ fn plus_toolchain_value_parser(s: &str) -> clap::error::Result<ResolvableToolcha
 #[command(name = "rustup", bin_name = "rustup[EXE]")]
 enum RustupSubcmd {
     /// Install or update the given toolchains, or by default the active toolchain
-    #[command(hide = true, after_help = install_help())]
+    #[command(hide = true, after_help = install_help(), aliases = ["add"])]
     Install {
         #[command(flatten)]
         opts: UpdateOpts,
     },
 
     /// Uninstall the given toolchains
-    #[command(hide = true)]
+    #[command(hide = true, aliases = ["remove", "rm", "delete", "del"])]
     Uninstall {
         #[command(flatten)]
         opts: UninstallOpts,
@@ -332,7 +332,7 @@ enum ToolchainSubcmd {
     },
 
     /// Install or update the given toolchains, or by default the active toolchain
-    #[command(aliases = ["update", "add"], after_help = toolchain_install_help())]
+    #[command(aliases = ["update", "up", "add"], after_help = toolchain_install_help())]
     Install {
         #[command(flatten)]
         opts: UpdateOpts,
@@ -541,6 +541,7 @@ enum OverrideSubcmd {
 )]
 enum SelfSubcmd {
     /// Download and install updates to rustup
+    #[command(alias = "up")]
     Update,
 
     /// Uninstall rustup


rami3l (approved):
I like it, thanks!

@djc @ChrisDenton Does the `show` -> `status` alias seem okay to you as well?

src/cli/rustup_mode.rs
```diff
@@ -123,14 +123,14 @@ fn plus_toolchain_value_parser(s: &str) -> clap::error::Result<ResolvableToolcha
 #[command(name = "rustup", bin_name = "rustup[EXE]")]
 enum RustupSubcmd {
     /// Install or update the given toolchains, or by default the active toolchain
-    #[command(hide = true, after_help = install_help())]
+    #[command(hide = true, after_help = install_help(), aliases = ["add"])]
```
djc, line 126: So this is `hide = true`, I think because the team has wanted to deprecated this in favor of `rustup toolchain install`? If we consider it deprecated I don't think we should add a bunch of aliases for it.

(Same for top-level `uninstall`.)

src/cli/rustup_mode.rs
```diff
@@ -158,7 +158,7 @@ enum RustupSubcmd {
     },
 
     /// Show the active and installed toolchains or profiles
-    #[command(after_help = show_help())]
+    #[command(after_help = show_help(), aliases = ["status"])]
```
djc: The commit message says "makes fingers want" in passive voice. Have you observed/discussed this with anyone else? I'm not convinced the parallel with `status` is strong enough for me to want to add this.

If anything, I think I'd want to call it `list` rather than `show`.

Maybe the solution is to add some support for user-defined aliases?

ChrisDenton: I agree that `rustup show` itself isn't really analogous to `git status`.  I mean, I like the idea of a `status` command but that would basically be `rustup show active-toolchain` (maybe with `--verbose` too).

src/cli/rustup_mode.rs
rami3l: @djc Oh no, not really.

`rustup install` was an old subcommand that didn't have all the features of `rustup toolchain install`, and that is why it was hidden.

Now (since the migration to clap-derive in #3596) that it delegates directly to the latter, I think it's ready to become visible again in a future version.

rami3l (approved):
@joshtriplett Given the overall feedback from the team, I think maybe we should drop the `rustup status` part, but other parts of this PR should be fine.

src/cli/rustup_mode.rs
rami3l: Okay. I think it's better to drop this part of the change until a consensus is made.

joshtriplett:
No objection, that part was a convenience anyway. Will rebase that commit away.

djc (approved):

PR #4592, opened 2025-11-11
rustup: unhide top-level install/uninstall commands

djc:
As suggested in https://github.com/rust-lang/rustup/pull/4588/files#r2504957482.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -123,14 +123,14 @@ fn plus_toolchain_value_parser(s: &str) -> clap::error::Result<ResolvableToolcha
 #[command(name = "rustup", bin_name = "rustup[EXE]")]
 enum RustupSubcmd {
     /// Install or update the given toolchains, or by default the active toolchain
-    #[command(hide = true, after_help = install_help(), aliases = ["add"])]
+    #[command(after_help = install_help(), aliases = ["add"])]
     Install {
         #[command(flatten)]
         opts: UpdateOpts,
     },
 
     /// Uninstall the given toolchains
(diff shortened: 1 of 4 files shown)


rami3l (approved):
Thanks :)

djc:
BTW, how do you feel about changing it so that `add` and `remove` are the primary names for these top-level commands instead of `install` and `uninstall` (we'd keep those as aliases)? Seems nicer to me.

rami3l:
> BTW, how do you feel about changing it so that `add` and `remove` are the primary names for these top-level commands instead of `install` and `uninstall` (we'd keep those as aliases)? Seems nicer to me.

@djc I think the fact that it constitutes an official suggestion rather than an option does have raised a yellow flag for me. If we are changing that, we should also mirror the suggestions in `rustup toolchain`, and suddenly all the past documentations are using an alias?

djc:
> @djc I think the fact that it constitutes an official suggestion rather than an option does have raised a yellow flag for me. If we are changing that, we should also mirror the suggestions in `rustup toolchain`, and suddenly all the past documentations are using an alias?

"suddenly all the past documentations are using an alias" doesn't seem like a problem to me necessarily. It's the way to advance the UX without breaking backwards compatibility. The question is if we think it is a UX improvement.

ChrisDenton:
While that change wouldn't break anything, recommendations do come with the implication that people should switch to the new thing.. My concern therefore would be churn, or at least the perception of churn. As a consequence of being actively developed again, rustup has undergone a lot of changes in the last few releases or so (even if some of those changes were implemented longer ago). So I worry that people's patience for perceived churn will be low atm.

PR #4594, opened 2025-11-11
Unpack progress

djc:
cc @FranciscoTGouveia 

https://github.com/user-attachments/assets/7696dac3-7ee7-424c-9eb1-454ea4c9c6d2

### DIFF
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -7,7 +7,7 @@ use std::sync::Mutex;
 use std::time::{Duration, Instant};
 
 use anyhow::{Context, Result, anyhow};
-use indicatif::{MultiProgress, ProgressBar, ProgressDrawTarget, ProgressStyle};
+use indicatif::{MultiProgress, ProgressBar, ProgressBarIter, ProgressDrawTarget, ProgressStyle};
 use sha2::{Digest, Sha256};
 use tracing::{debug, info, warn};
 use url::Url;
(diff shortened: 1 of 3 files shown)


FranciscoTGouveia:
After testing the changes, it seems that the ticking speed for the unpacking phase is so high that it feels a bit disconcerting.
Would it be possible to slow it down?

Moreover, I would like to ask whether this change is a step toward achieving concurrent downloads & installs (without spawning an additional thread).
If so, I believe the issue where progress bars stop updating during installation (as discussed in #4570) remains, and I do not think that this will help in facilitating that -- I might be overlooking something here.
Even though downloads continue in the background, the blocking `install` call never yields, so the progress bars never get the chance to update during installation.

I think that making `install` *async* is ultimately the only way to avoid the spawning of an additional thread.

rami3l:
@FranciscoTGouveia In the current stable version we actually have progress reporting for unpacking as well, however it's behind a 1-second timeout so no one actually sees it except for the docs component.

Thus, I think your concern is reasonable, and even though it's probably great for @djc to restore that behavior here, it might be worth it to make it closer to the original in terms of lowering the number of ticks indeed.

rami3l (commented):
Just tested locally. The code looks okay but I share @FranciscoTGouveia's concern: it looks like the tick frequency is way too high...

djc:
Changed it to display a progress bar while unpacking. See the movie in the PR description (which also includes #4602).

rami3l (approved):

PR #4595, opened 2025-11-11
refactor(dist/component/package): make `DirectoryPackage::install()` async

rami3l:
> **Don't merge**: this PR is for evaluation purposes only.

From my local testing it seems that `DirectoryPackage::install()` is not a good candidate for the said optimization in https://rust-lang.zulipchat.com/#narrow/channel/490103-t-rustup/topic/Shipping.20fully.20concurrent.20downloads.20of.20rust.20components/near/554900068. On my machine, this has slowed down the installation of the `beta` toolchain with `default` profile by ~2x.

### DIFF
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -88,7 +88,7 @@ impl<P: Deref<Target = Path>> DirectoryPackage<P> {
             }
     }
 
-    pub fn install<'a>(
+    pub async fn install<'a>(
         &self,
         target: &Components,
         name: &str,
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -8,6 +8,7 @@ use std::mem;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::sync::OnceLock;
+use std::time::{Duration, Instant};
 
 use anyhow::{Context, Result, anyhow, bail};
 use tar::EntryType;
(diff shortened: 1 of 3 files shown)


src/dist/component/package.rs
```diff
@@ -108,7 +109,15 @@ impl<P: Deref<Target = Path>> DirectoryPackage<P> {
         let manifest = utils::read_file("package manifest", &root.join("manifest.in"))?;
         let mut builder = target.add(name, tx);
 
+        let yield_timeout = Duration::from_millis(50);
```
djc, line 112: This is pretty short -- have you tried with longer intervals, like 250/500/1000ms?

djc:
(IMO `[DON'T MERGE]` is really not helpful when the PR is draft anyway.)

rami3l:
> (IMO `[DON'T MERGE]` is really not helpful when the PR is draft anyway.)

The intention is that it will never be merged instead of passing through the normal state.

src/dist/component/package.rs
rami3l: No significant changes on my side with either of those values: the installation is always ~2x slower than latest.

djc: Wow, that seems very surprising.

rami3l:
Closing as superseded by https://github.com/rust-lang/rustup/pull/4605, however this approach has played a very important part in proving the necessity of `spawn_blocking()` uses, so the work did not end up in vain.

cc https://github.com/rust-lang/rustup/pull/4471#issuecomment-3564581077

PR #4599, opened 2025-11-14
Clarify installation process dependencies

djc:


src/dist/manifestation.rs
```diff
@@ -149,15 +149,32 @@ impl Manifestation {
...
+                    Ok(p) => p,
+                    Err(e) => return Some(Err(e)),
+                };
+
+                let target_package = match package.get_target(component.target.as_ref()) {
+                    Ok(tp) => tp,
+                    Err(e) => return Some(Err(e)),
+                };
+
+                match target_package.bins.is_empty() {
```
rami3l: Nit: Existing, but when looking at this I realized that maybe this could be matched as `&target_package.bins` against `[]` or `[binary, ..]`? It's true that using `match` rather than `if-else` would help spotting such issues.

src/dist/component/package.rs
```diff
@@ -4,10 +4,10 @@
 
 use std::collections::{HashMap, HashSet};
 use std::io::{self, ErrorKind as IOErrorKind, Read};
-use std::mem;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::sync::OnceLock;
+use std::{env, mem};
```
rami3l: Nit: ~~RA likes to suggest stuff like that. When this happens I just merge the imports altogether.~~ This change seems to be unused?

rami3l (approved):
Looks pretty cool, thanks a lot :)

src/dist/component/package.rs
djc: It is a temporary state within the commits in this PR.

src/dist/manifestation.rs
djc: Added a commit that restructures this a bunch. I used `.first()` for this but also moved stuff around in other ways.

PR #4600, opened 2025-11-14
Manifest names

djc:


rami3l (commented):
I think this indicates a good direction to move forward, thanks!

However, in theory this is a refactoring PR and should not result in changes in the CI status? Or am I missing something here?

djc:
> However, in theory this is a refactoring PR and should not result in changes in the CI status? Or am I missing something here?

You're not wrong. I guess I made some bad change...

rami3l (approved):

PR #4602, opened 2025-11-18
Fix refresh rate for progress bars

djc:
I guess it seemed like a good idea at the time to avoid applying a rate limit to progress bar drawing to a `TermLike`, since a consumer might want to be able to observe every draw. In this case, though, we really want some rate limiting to prevent flickering, so explicitly set the refresh rate to indicatif's default of 20 Hz.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -215,7 +215,7 @@ impl Process {
         };
 
         match term {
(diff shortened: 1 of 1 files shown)


rami3l (approved):
The code looks pretty simple. I entrust you with the manual testing of the fix :)

PR #4604, opened 2025-11-19
dist: align progress bar elements

djc:
- Increase label to accomodate rust-analyzer component
- Drop bytes so lines will fit within 80 columns
- Shorten progress bars so lines will fit within 80 columns
- Add label to download progress bar

Before:

https://github.com/user-attachments/assets/3efaf078-3c13-4a80-bf33-8358ec1730e4

After:

https://github.com/user-attachments/assets/6f70d062-4edf-494b-9b27-ca09d054014a

### DIFF
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -285,7 +285,7 @@ impl<'a> DownloadCfg<'a> {
         let progress = ProgressBar::hidden();
         progress.set_style(
             ProgressStyle::with_template(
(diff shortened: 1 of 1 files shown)


rami3l (approved):
Nice!

PS: Would you mind adding a screenshot or two so we can have a clearer idea of what's changed though?

djc:
Added screen recordings for before and after.

PR #4605, opened 2025-11-19
dist: make installation asynchronous

djc:
cc @FranciscoTGouveia 

https://github.com/user-attachments/assets/763a40ea-af3f-41a6-8863-aa895fe82dc3

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -4,11 +4,17 @@
 #[cfg(test)]
 mod tests;
 
+use std::collections::VecDeque;
 use std::path::Path;
+use std::pin::Pin;
 use std::sync::Arc;
+use std::task::{Context, Poll, ready};
+use std::vec;
 
-use anyhow::{Context, Result, anyhow, bail};
+use anyhow::{Context as _, Result, anyhow, bail};
+use futures_util::Stream;
 use futures_util::stream::{FuturesUnordered, StreamExt};
+use tokio::task::{JoinHandle, spawn_blocking};
 use tracing::{info, warn};
 
 use crate::diskio::{Executor, IO_CHUNK_SIZE, get_executor, unpack_ram};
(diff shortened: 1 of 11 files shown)


src/dist/manifestation.rs
```diff
@@ -445,6 +455,86 @@ impl Manifestation {
...
+            cleanup_downloads: Vec::with_capacity(components.len()),
+            components,
+            install_queue: VecDeque::new(),
+            installing: None,
+            downloads: FuturesUnordered::new(),
+        }
+    }
+
+    fn try_install(&mut self, tx: &mut Option<Transaction>) {
+        if let Some(installable) = self.install_queue.pop_front() {
```
rami3l: Nit: I think we can reduce some indentations here.

rami3l (commented):
These changes look pretty reasonable and definitely an improvement over #4567’s attempt of spilling `Arc` everywhere.

I imagine that if correctly implemented, it would be equivalent to #4471 in terms of performance because as long as we are spawning the installation task, the extra overhead introduced by #4567 should be minimal.

It’s worth another bench nonetheless though.

PS: Would you mind investigating the CI failures? This implementation doesn’t seem to have covered all the edge cases.

djc:
Running `time rustup-init toolchain install beta` (compiled with `--release`) on #4471:

```
4.78s user 14.11s system 101% cpu 18.604 total
```

On this PR:

```
5.12s user 14.54s system 138% cpu 14.214 total
```

So it looks like this PR is a bit faster, by making better use of parallelism.

On `main`:

```
5.21s user 14.69s system 103% cpu 19.154 total
```

src/dist/manifestation.rs
djc: Good point, fixed.

djc:
> PS: Would you mind investigating the CI failures? This implementation doesn’t seem to have covered all the edge cases.

Oops, solved those.

rami3l:
Thanks! I really want to test this with my own machine but I don't have my fiber installed until tomorrow... I hope you won't mind my delaying merging this for a bit just to be sure 🙏

djc:
No problem! Hope the fiber install will work out better for you tomorrow!

rami3l (approved):
@djc Just tested a few times after connecting all my cables. The thing did work which is super cool, but I have some concerns when downloading the `cargo` component, which seems to be stuck all of the time.

However, considering the fact that this implementation shouldn't have set any bias of any kind in that regard, I'm considering this to be a particular environment-related or remote-related problem (the same thing happens for stable `rustup` on my side and not just latest). Anyway, I think we can merge this first and see how it goes.

Many thanks again for all the refactoring work to make the final implementation more clearer and much more elegant than the original :)

Issue #988, opened 2017-03-16
Rustup (including proxies) is not safe for concurrent use


matklad:
(Edited to capture all the details that have emerged over time)

# Recovering from this bug:

Usually just doing a `rustup component remove NAME && rustup component add NAME` will fix things. Sometimes removing the entire toolchain will be needed. In rare cases uninstalling rustup entirely will be needed.

# User model of Rustup

Rustup may be run as three different commands:
* rustup-init to install rustup (& by default a toolchain)
* rustup to explicitly query or modify an installation (including rustup itself and one or more toolchains)
* as a proxy of rustc, cargo etc (& which can implicitly trigger installation, upgrade of modification of a toolchain e.g. through toolchain files)

# Locking in Rustup

Rustup gets run concurrently in two very different contexts: within a single (machine, user), it may be run concurrently by the user, or an IDE, or both, to perform tasks ranging from toolchain installation, component addition, documentation opening. All of these require multiple reads to be made to the rustup data with a consistent view; some of them require writes to be made as well.

Rustup may also be run across machines, where a different machine but a shared rustup install is present - (machineA, userA) + (machineB, userA) - and in this case the same set of operations may take place, with the same requirements. 

Whatever consistency solution we adopt would be best if it can deliver both use cases, and not require manual lock administration as lockdir style solutions do, nor additional running network daemons.

## Proxies

Rustup has one set of proxies shared across all toolchains; the proxies are held open while any process from a toolchain is running - e.g. IDE's hold rls open for extended periods.

We need a lock to prevent concurrent attempts to install new proxies, and we need a notification mechanism back to the running proxies to allow them to be notified to exit when an update is required (because of presumed limitations of in-use-file-replacement on Windows, though recent changes may mean we can avoid this)

## Toolchains

We have many toolchains in one rustup installation; stable, beta, nightly, dated nightly versions, and custom versions. Adding a toolchain adds a directory and a hash file; we need a lock to prevent collisions attempting to move the directory into place. Deleting a toolchain does a recursive rm in-place, which also needs a lock to prevent other rustup invocations presuming that the toolchain is actually installed during the time the deletion takes place (or perhaps we need to rename-then-delete, though that can run into race conditions with virus scanners, especially if the toolchain was just installed). Further, permitting deletions at any point will require notifications to running rls process proxies from that toolchain to cause them to shutdown, or the .exe is likely not deletable on Windows.

## Components

[...]

PR #4606, opened 2025-11-19
Lock download file to avoid races

### LINKED ISSUE #4607: rustup-init fails when invoked in parallel
### Verification

### Problem

When `rustup-init -y` runs in parallel, multiple processes write the same partial file, and then try to move it to the same target location. One of those process succeeds and renames the file, the other processes error due to the missing file:

```
error: component download failed for cargo-x86_64-unknown-linux-gnu: could not rename 'downloaded' file from '/home/konsti/projects/rustup/rustup/downloads/8d17d5a4734e684228adee4ea4a75da94eb7f71993804adf29f344709e65fe09.partial' to '/home/konsti/projects/rustup/rustup/downloads/8d17d5a4734e684228adee4ea4a75da94eb7f71993804adf29f344709e65fe09': No such file or directory (os e
```

After locking those files, the transaction step fails when running in parallel:

```
error: failed to install component: 'cargo-x86_64-unknown-linux-gnu', detected conflict: 'bin/cargo'
```

### Steps

MRE:

```
rm -rf $PWD/cargo $PWD/rustup
CARGO_HOME=$PWD/cargo/ RUSTUP_HOME=$PWD/rustup target/debug/rustup-init -y --no-modify-path --profile minimal & \
  CARGO_HOME=$PWD/cargo/ RUSTUP_HOME=$PWD/rustup target/debug/rustup-init -y --no-modify-path --profile minimal & \
  CARGO_HOME=$PWD/cargo/ RUSTUP_HOME=$PWD/rustup target/debug/rustup-init -y --no-modify-path --profile minimal
wait -n
```

### Possible Solution(s)

There's two popular options:
* Use a unique (temporary) file or directory instead of just `.partial`
* Use file locks

Prototype: https://github.com/rust-lang/rustup/pull/4606

A temp file has wider support (no requirement on OS level locking), file locks actually prevent multiple processes downloading the same, and may be required for fixing parallel transactions. For the unique file name solution, I wouldn't know how to implement cleaning up previous failures, while there's an existing implementation when using locking.

### Notes

_No response_

### Rustup version

```console
rustup-init 1.28.2+424 (4ae201236 2025-11-19)
```

### Installed toolchains

```console
n/a
```

### OS version

```console
Ubuntu 24.04
```

konstin:
See #4607 for details.

This is one of two possible implementations, the easier one is using a temporary directory or a unique filename to ensure processes don't collide. Using locks has less platform support, but OTOH means that we don't create stray temp files that don't get cleaned up (because we never know if one of those files is actually used by another process).

This does not fix #4607 completely. It only fixed the downloaded, parallel `rustup-init` calls still fail at the transaction stage.

This PR introduces a `LockedFile` primitive that's modeled after the uv type of the same name. It's currently very simple and locks forever. We can extend it with a timeout that makes rustup exit after a certain duration and print an error message, instead of an inscrutable waiting without any message.


src/download/mod.rs
```diff
@@ -26,6 +27,96 @@ use crate::{dist::download::DownloadStatus, errors::RustupError, process::Proces
...
+        let path = path.into();
+
+        // If path already exists, return it.
+        if let Ok(file) = OpenOptions::new().read(true).write(true).open(&path) {
+            return Ok(Self { path, file });
+        }
+
+        // Otherwise, create a temporary file with 777 permissions, to handle races between
+        // processes running under different UIDs (e.g., in Docker containers). We must set
+        // permissions _after_ creating the file, to override the `umask`.
```
weihanglo: Interesting. I guess Cargo's file locks also have the same issue?

weihanglo:
> after the uv type of the same name

Would be great if you can a link to it :)

konstin:
The uv type: <https://github.com/astral-sh/uv/blob/fc0cf90795dbfa961f22fc0934a53afc9c13c349/crates/uv-fs/src/lib.rs#L672>. I have a branch that adds async timeouts, but that's in review still.

src/download/mod.rs
```diff
@@ -26,6 +27,96 @@ use crate::{dist::download::DownloadStatus, errors::RustupError, process::Proces
...
+        let file = if let Some(parent) = path.parent() {
+            NamedTempFile::new_in(parent)?
+        } else {
+            NamedTempFile::new()?
+        };
+        if let Err(err) = file
+            .as_file()
+            .set_permissions(std::fs::Permissions::from_mode(0o777))
```
reneleonhardt: Can the permissions be cached in a static variable?

konstin: The method is a no-op constructor.

konstin:
Link to the merged async uv implementation: https://github.com/astral-sh/uv/blob/3e22637c93068418506b0e4f79b47dbe956d2836/crates/uv-fs/src/locked_file.rs#L91-L353

rami3l:
@konstin Many thanks again!

I'm OTOH looking for another way of resolving #988 (currently my 3rd attempt now XD) so I don't want to rush to decisions, currently I'm open to all options 🙏 

It's nice that this avoids lock files or similar per toolchain, but I'm working on the transaction level so I don't have enough info to say at this moment if this relatively local solution is the best way forward, I'm afraid maybe a lock is still required after all 🙇

konstin:
You definitely know better than me, this is built on my naive assumption that we'll need this plus something for the rest of #988, where we lock the install stage too, similar to how we handle it in uv (though for packages we also do streaming install-and-unpack in a temp dir and use atomic replace, that's the other version mentioned in the PR description, but for installing a Python interpreter (our version of a toolchain), we do have a file lock: https://github.com/astral-sh/uv/blob/f7d1215a980c3e332cbe602db0a170a7967e7850/crates/uv-python/src/installation.rs#L286).

Otherwise, we can also splice out the lock file implementation and use it for some other place.

rami3l:
I'm actively pursuing a proper solution of the underlying issue as a part of a project goal:
- https://github.com/rust-lang/goals/issues/736

As such, I'm closing this PR as stale.

Many thanks nonetheless @konstin for your support and all the contributions!

PR #4613, opened 2025-11-21
[STABLE] Prepare for mdbook 0.5 migration

ehuss:
This is the stable-branch version of https://github.com/rust-lang/rustup/pull/4612

Because the pull_request checks are building docs from main and stable, the checks fail due to mdbook 0.5 changing and removing some fields. This change should allow the books to work with both 0.4 and 0.5. In particular:

- `multilingual` -- This was never used, is optional, and is ignored in 0.4. This is rejected in 0.5.
- `curly-quotes` -- This was renamed to `smart-punctuation` a long while ago. Both 0.4 and 0.5 use `smart-punctuation`, and 0.5 has removed the old `curly-quotes`.
- `src` -- This isn't needed and is an artifact of an `mdbook init` bug.

### DIFF
--- a/doc/dev-guide/book.toml
+++ b/doc/dev-guide/book.toml
@@ -1,12 +1,10 @@
 [book]
 authors = ["The Rust Project Developers"]
 language = "en"
-multilingual = false
-src = "src"
 title = "The Rustup developer guide"
 
 [output.html]
-curly-quotes = true
+smart-punctuation = true
 edit-url-template = "https://github.com/rust-lang/rustup/edit/master/doc/dev-guide/{path}"
 git-repository-url = "https://github.com/rust-lang/rustup/tree/master/doc/dev-guide"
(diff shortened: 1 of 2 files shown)


ehuss:
Hm, looks like stable is not in a good space right now 😕

rami3l:
@ehuss That's because the `r-l/r` repo has changed its CI structure, unfortunately.

I think we really need to decouple website releases from the main release... everything is very confusing right now for both us and the users, and some stuff that should've been updated hasn't been modified in months :o

ehuss:
Talked with @rami3l. I'm going to close this for now. This isn't at all urgent, and can wait until the next release that contains https://github.com/rust-lang/rustup/pull/4612. 

https://github.com/rust-lang/rustup/issues/4615 is now open to figure out a longer-term strategy for updating the live site.

PR #4614, opened 2025-11-21
feat(cli/rustup-mode): support `rustup` completion for nushell

### LINKED ISSUE #4603: Add nushell completion support

Enable nushell to enjoy first-class completion support, just like shells such as bash and zsh.

### Solution you'd like

You can also have native rustup command completion support in nushell.

### Notes

We can consider using [clap_complete_nushell](https://docs.rs/clap_complete_nushell/latest/clap_complete_nushell/) crate.
In many projects, this `clap_complete_nushell` crate is used to generate nushell completion scripts, for example:
1. [jj](https://github.com/jj-vcs/jj)
2. [uv](https://github.com/astral-sh/uv)
3. [atuin](https://github.com/atuinsh/atuin)

rami3l:
Mirrors https://github.com/jj-vcs/jj/blob/54511ee3fde4f381a67a38ca3a4a0f1c7412753a/cli/src/commands/util/completion.rs#L80; closes #4603.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1868,13 +1868,12 @@ fn output_completion_script(
             );
         }
         CompletionCommand::Cargo => {
-            if let Shell::Zsh = shell {
-                writeln!(process.stdout().lock(), "#compdef cargo")?;
-            }
-
             let script = match shell {
                 Shell::Bash => "/etc/bash_completion.d/cargo",
-                Shell::Zsh => "/share/zsh/site-functions/_cargo",
+                Shell::Zsh => {
+                    writeln!(process.stdout().lock(), "#compdef cargo")?;
+                    "/share/zsh/site-functions/_cargo"
+                }
                 _ => {
                     return Err(anyhow!(
                         "{command} does not currently support completions for {shell}",
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -281,7 +281,7 @@ enum RustupSubcmd {
     /// Generate tab-completion scripts for your shell
     #[command(after_help = completions_help(), arg_required_else_help = true)]
     Completions {
-        shell: Shell,
+        shell: CompletionShell,
 
         #[arg(default_value = "rustup")]
         command: CompletionCommand,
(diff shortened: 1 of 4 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1853,8 +1853,57 @@ impl fmt::Display for CompletionCommand {
     }
 }
 
+#[derive(clap::ValueEnum, Clone, Copy, Debug)]
+enum CompletionShell {
```
djc, line 1857: It seems pretty unclear from the code why this needs to be a separate enum rather than another variant added to `Shell`? Should at least clarify in a comment/docstring.

rami3l: `Shell` is an external enum so cannot be extended. Do you have some other ideas regarding the extension of such a type? Will nested enums be better in your opinion?

djc: @epage why is this a separate crate? Seems like a pain in the ass for downstreams?

rami3l: PS: When I say this PR mirrors the `jj` one that's because `jj` has also flattened the enum like this.

epage: > why is this a separate crate? Seems like a pain in the ass for downstreams?

There are two distinctions
- In `clap_complete` or not
- In `Shell` or not

Looking back in the past, I wish we had fewer shells in `Shell` then we do and in the next major version would like to trim it down and locking it down.  imo we shouldn't be implicitly extending the compatibility surface of our callers by adding new content to `Shell`. Pulling in esoteric shells like elvish and fig should be explicit choices.  Where the line is for what should be in `clap_complete` or even in `Shell` is not yet determined.

See also
- https://github.com/clap-rs/clap/issues/5329
- https://github.com/clap-rs/clap/discussions/5880

At least with our new completion system, it is easier to add new shells, see https://docs.rs/clap_complete/latest/clap_complete/env/struct.Shells.html

Would be willing to explore something similar for our existing completion system.

djc: Maybe there shouldn't be an enum at all, only a trait?

PR #4618, opened 2025-11-23
fix(utils): downgrade panic to warning in `delete_dir_contents_following_links()`

### LINKED ISSUE #4609: `rustup update` needs a cleaner error message if it fails while cleaning files - and no panic
### Verification

### Problem

`rustup update` needs a cleaner error message when it fails while cleaning files. 

PS C:\Windows\System32> rustup update
info: syncing channel updates for 'stable-x86_64-pc-windows-msvc'

  stable-x86_64-pc-windows-msvc unchanged - rustc 1.91.1 (ed61e7d7e 2025-11-07)

thread 'main' panicked at src\utils\mod.rs:479:13:
Unable to clean up C:\Users\XYZ\.rustup\tmp: Os { code: 5, kind: PermissionDenied, message: "Access is denied." }

### Steps

run: 

rustup update

### Possible Solution(s)

_No response_

### Notes

Yesterday the whole internet went done because of a panic (thanks CloudFlare). Maybe we need to remove these unwrap()? 

### Rustup version

```console
PS C:\Windows\System32> rustup --version
info: The currently active `rustc` version is `rustc 1.91.1 (ed61e7d7e 2025-11-07)`
```

### Installed toolchains

```console
installed toolchains
--------------------

active toolchain
----------------
installed targets:
  wasm32-unknown-unknown
  x86_64-pc-windows-msvc
```

### OS version

```console
Windows 11
```

rami3l:
Closes #4609.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -435,7 +435,7 @@ pub(crate) fn delete_dir_contents_following_links(dir_path: &Path) {
 
     match raw::open_dir_following_links(dir_path).and_then(|mut p| p.remove_dir_contents(None)) {
         Err(e) if e.kind() != io::ErrorKind::NotFound => {
-            panic!("Unable to clean up {}: {:?}", dir_path.display(), e);
+            warn!("unable to clean up {}: {e}", dir_path.display());
         }
         _ => {}
     }


djc:
Are you sure it's okay to skip cleaning up the download directory?

rami3l:
> Are you sure it's okay to skip cleaning up the download directory?

@djc That's a good question. My rationale is this: as per

https://github.com/rust-lang/rustup/blob/a1676f4adf942c22c4a5ae58a8e30b8bb81a2029/src/dist/download.rs#L48-L51

... this directory is supposed to be owned exclusively by rustup, so rustup in theory should have full control over it, and any violation is considered an internal state violation. The same is true for the tmp dir. Given your previous analysis in https://github.com/rust-lang/rustup/issues/4609#issuecomment-3556604707, switching the control flow from divergent to non-divergent is not really harmful here and will only leave some cache in the directory, and either the cache is correct and can be reused, or it isn't and the FS writes will fail. Of course I'd be glad to be corrected.

ChrisDenton:
If it's not OK to skip cleaning then an error would leave rustup in a broken state anyway, no?

djc:
> If it's not OK to skip cleaning then an error would leave rustup in a broken state anyway, no?

That's fair!

src/utils/mod.rs
```diff
@@ -435,7 +435,7 @@ pub(crate) fn delete_dir_contents_following_links(dir_path: &Path) {
 
     match raw::open_dir_following_links(dir_path).and_then(|mut p| p.remove_dir_contents(None)) {
         Err(e) if e.kind() != io::ErrorKind::NotFound => {
-            panic!("Unable to clean up {}: {:?}", dir_path.display(), e);
+            warn!("unable to clean up {}: {e:?}", dir_path.display());
```
ChrisDenton:
```suggestion
            warn!("unable to clean up {}: {e}", dir_path.display());
```

I think it'd be nicer to use the `Display` formatting for the error.

rami3l: I'm not 100% sure about that since it might still be interesting to look at the underlying error in this case, and the `Display` implementation only focuses on the error message...

```rs
fn main() -> std::io::Result<()> { std::fs::File::open("foo.txt").map(drop) }
```

... this gives:

```console
> cargo run
[..]
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

Returning to the example of #4609, `Os { code: 5, kind: PermissionDenied, message: "Access is denied." }` has also made the origin of the error much clearer IMHO. Otherwise it might not be immediately clear where the `Access is denied.` string is coming from?

ChrisDenton: Hm, the difference is:

Before:
```
Os { code: 5, kind: PermissionDenied, message: "Access is denied." }
```

After:
```
Access is denied. (os error 5)
```

I think the latter looks nicer and the only thing it's missing is the the error kind.

rami3l: @ChrisDenton Okay, I'm convinced. Many thanks 🙏

ChrisDenton (approved):

PR #4625, opened 2025-11-27
fix(dist/manifestation): print "downloading component" only on `InstallEvents`

rami3l:
This avoids the following regression:

```console
> rustup component remove rust-analyzer
info: removing component rust-analyzer
info: downloading component(s)
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,7 +6,7 @@ use std::str::FromStr;
 use anyhow::{Context, Result, anyhow, bail};
 use serde::Deserialize;
 use thiserror::Error as ThisError;
-use tracing::{debug, error, info, trace, warn};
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -6,7 +6,7 @@ use std::str::FromStr;
 use anyhow::{Context, Result, anyhow, bail};
 use serde::Deserialize;
 use thiserror::Error as ThisError;
-use tracing::{debug, error, info, trace, warn};
+use tracing::{debug, info, trace, warn};
```
rami3l, line 9: I have no idea why this wasn't linted before. Toolchain bug?

djc (approved):

PR #4623, opened 2025-11-27
feat: Added xonsh shell support - try 1

anki-code:
Resolve #4622

PR #4626, opened 2025-11-28
feat: Added xonsh shell support

### LINKED ISSUE #4622: Xonsh shell support in rustup

Hey! 
It will be cool to have [xonsh shell](https://github.com/xonsh/xonsh/) support.
Thanks!

### Solution you'd like

I would like to see the xonsh shell in the list of shells:

<details>

Image

</details>

### Notes

_No response_

anki-code:
Resolve https://github.com/rust-lang/rustup/issues/4622

Test:

```xsh
docker run --rm -it rust

apt update && apt install -y python3 python3-pip && pip install --break-system-packages xonsh

xonsh

git clone -b xonsh_shell3 https://github.com/anki-code/rustup && cd rustup/

# https://rust-lang.github.io/rustup/dev-guide/
cargo build
mkdir home
$RUSTUP_HOME='home' $CARGO_HOME='home' target/debug/rustup-init --no-modify-path -y
source /rustup/home/env.xsh

$PATH.pop(0)  # remove '/usr/local/cargo/bin' from container
which cargo
# /rustup/home/bin/cargo
cargo --version
# cargo 1.91.1
```

### DIFF
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -69,6 +69,7 @@ fn enumerate_shells() -> Vec<Shell> {
         Box::new(Nu),
         Box::new(Tcsh),
         Box::new(Pwsh),
+        Box::new(Xonsh),
     ]
 }
 
@@ -442,6 +443,61 @@ impl UnixShell for Pwsh {
 
     fn source_string(&self, process: &Process) -> Result<String> {
         Ok(format!(r#". "{}/env.ps1""#, self.cargo_home_str(process)?))
+    }
+}
+
+struct Xonsh;
+
+impl UnixShell for Xonsh {
+    fn does_exist(&self, process: &Process) -> bool {
+        process.var("XONSHRC").is_ok() || utils::find_cmd(&["xonsh"], process).is_some()
+    }
+
+    fn rcfiles(&self, process: &Process) -> Vec<PathBuf> {
+        let mut paths = vec![];
+
+        if let Ok(p) = process.var("XDG_CONFIG_HOME") {
+            let mut p = PathBuf::from(p);
+            p.extend(["xonsh", "rc.xsh"]);
+            paths.push(p);
+        }
+
+        if let Some(mut p) = process.home_dir() {
+            p.extend([".config", "xonsh", "rc.xsh"]);
+            paths.push(p);
+        }
+
+        if let Some(home) = process.home_dir() {
+            paths.push(home.join(".xonshrc"));
+        }
+
+        paths
+    }
+
+    fn update_rcs(&self, process: &Process) -> Vec<PathBuf> {
+        // The first rcfile in XDG_CONFIG_HOME takes precedence.
+        match self.rcfiles(process).into_iter().next() {
+            Some(path) => vec![path],
+            None => vec![],
+        }
+    }
+
+    fn env_script(&self) -> ShellScript {
+        ShellScript {
+            name: "env.xsh",
+            content: include_str!("env.xsh"),
+        }
+    }
+
+    fn source_string(&self, process: &Process) -> Result<String> {
+        Ok(format!(
+            r#"source "{}/env.xsh""#,
+            self.cargo_home_str(process)?
+        ))
+    }
+
+    fn cargo_home_str(&self, process: &Process) -> Result<Cow<'static, str>> {
+        cargo_home_str_with_home("$HOME", process)
     }
 }
 
--- a/doc/user-guide/src/installation/index.md
+++ b/doc/user-guide/src/installation/index.md
@@ -100,3 +100,5 @@ For `zsh`, you must then add the following line in your `~/.zshrc` before
 ```zsh
 fpath+=~/.zfunc
 ```
+
+In Xonsh you can reuse Fish completion by installing [xontrib-fish-completer](https://github.com/xonsh/xontrib-fish-completer).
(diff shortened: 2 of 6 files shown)


djc (approved):
Please stop reopening the PR, which loses all the context.

anki-code:
I believe this PR should work. Thanks for review.
(I've faced first time with that ["Squash and merge" button](https://github.com/rust-lang/rustup/pull/4623#issuecomment-3587232568) is not the solution and "Update branch" button doing something bad. Also there is no clarifications what to do and what to not do in the [dev guide](https://rust-lang.github.io/rustup/dev-guide/) so I stepped on all the rakes here, haha)

src/cli/self_update/shell.rs
```diff
@@ -445,6 +446,62 @@ impl UnixShell for Pwsh {
     }
 }
 
+struct Xonsh;
+
+impl UnixShell for Xonsh {
+    fn does_exist(&self, process: &Process) -> bool {
+        process.var("XONSHRC").is_ok() || utils::find_cmd(&["xonsh"], process).is_some()
```
rami3l, line 453: I believe mirroring `fish`'s implementation would be better here. Is there a particular reason to check for `XONSHRC` instead?

src/cli/self_update/shell.rs
```diff
@@ -445,6 +446,62 @@ impl UnixShell for Pwsh {
...
+        if let Some(home) = process.home_dir() {
+            paths.push(home.join(".xonshrc"));
+        }
+
+        paths
+    }
+
+    fn update_rcs(&self, process: &Process) -> Vec<PathBuf> {
+        for f in self.rcfiles(process) {
+            if f.is_file() {
```
rami3l: Please don't copy from `tcsh` here, it was written in this way because both `.tcshrc` and `.cshrc` are supported as valid `tcsh` RC files.

Copying the implementation from a more regular shell like `Nu` would be fine.

rami3l (changes requested):

src/cli/self_update/shell.rs
anki-code: Because of xonsh is not POSIX compatible it keeps path to bash or zsh in the SHELL env var for compatibility (dependent on OS etc). So we need to detect here that we're in xonsh. Checking for XONSHRC env variable is good enough.

src/cli/self_update/shell.rs
anki-code: @rami3l could you please change this part of code as you prefer to see? The logic looks trivial but I will spend much time on understanding what you want to see than really doing implementation. If you have no time I'm ok to close this until someone will grab this.

djc: @rami3l suggested just copying from Nushell instead of tcsh, which seems straightforward?

anki-code: Updated and [tested](https://github.com/rust-lang/rustup/pull/4626#issue-3674255563).

src/cli/self_update/shell.rs
rami3l: @anki-code It should be noted however that `$SHELL`, which usually signifies the [login shell](https://man.openbsd.org/login.1#ENVIRONMENT), can be set to a non-POSIX shell. This is already the case for other shells such as `fish` in this file. What do you think about that?

PS: I agree that most of the time this won't matter since the second catch-all will usually trigger.

anki-code: @rami3l The big picture:
1. $SHELL can contain any shell and xonsh can be installed as a login shell as well and can set the path to xonsh to $SHELL.
2. But most of tools in the world expect that $SHELL is POSIX. They blindfold grab the value of $SHELL and run POSIX commands there and have errors as result and this can follows to unexpected consequences. This is why it's preferred way to keep POSIX shell in the $SHELL forever.
3. Some shells like zsh and _maybe_ fish has POSIX-compatible mode. They detect POSIX commands and run them mostly without errors. This is why they set itselfs to the $SHELL.
4. So this is why using $SHELL value is not the path for shell checking.

rami3l: > most of tools in the world expect that $SHELL is POSIX

@anki-code I doubt whether that is the case.

I use `fish` as my login shell and it doesn't have a POSIX-compatibility mode, so I think myself can count as a data point. From my experience, shell script writers would trust `sh` as authentically POSIX rather than `$SHELL`.

Also, a quick [GitHub global search](https://github.com/search?q=%22%24SHELL%22+language%3Ashell&type=code&p=1) seems to support this idea.

However, I guess it's okay to merge it while leaving this thread open. We can always come back when either approach is proven suboptimal, and then we can replace the existing with yours if applicable, for example.

rami3l (approved):
@anki-code Many thanks for making this PR!

Could you just squash these commits into one so we can merge it?

src/cli/self_update/shell.rs
anki-code: > I use fish as my login shell 

You're lucky. My [first result](https://github.com/lunixbochs/ActualVim/blob/1f555ce719e49d6584f0e35e9f0db2f216b98fa5/lib/util.py#L41-L55) [from search](https://github.com/search?q=%22env%5B%27SHELL%27%5D%22&type=code) shows that there is no support for non-posix + support fish so I believe this code will fail or have consequences for non-posix.

Try next examples from search and you'll see that people have no thoughts about this could be another shell. So the probability to have fail with non POSIX in $SHELL exists. I read a report where someone catch Linux distro freezing after rebooting with new $SHELL. So if it's work to you you're lucky. Or may be fish as pretty old shell has more support.

anki-code:
@rami3l done

PR #4629, opened 2025-11-29
fix(dist/download): align `total_bytes` fields in progress reporting UI

rami3l:
In #4604, the `total_bytes` fields seem to be inconsistently left-aligned in some places, and right-aligned in others. As such, scenarios like the following will occur from time to time (note the `cargo` line):

t-y

Thus, it is required to make them right-aligned with `:>`, however it seems (from my testing) that specifying the alignment without a width is a no-op, so a width of ~~10~~ 11 is deduced from the longest string from my observation ~~`123.45 MiB`~~ `1023.00 KiB`.

However, since the rest of the codebase assumes a width of 9, I have manually increased every appearance of explicit width by a value of ~~1~~ 2 for the sake of coherence.

This gives something like:

image

### DIFF
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -288,7 +288,7 @@ impl<'a> DownloadCfg<'a> {
         let progress = ProgressBar::hidden();
         progress.set_style(
             ProgressStyle::with_template(
-                "{msg:>13.bold} downloading [{bar:15}] {total_bytes} ({bytes_per_sec}, ETA: {eta})",
+                "{msg:>13.bold} downloading [{bar:15}] {total_bytes:>11} ({bytes_per_sec}, ETA: {eta})",
             )
             .unwrap()
             .progress_chars("## "),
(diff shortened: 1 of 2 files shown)


djc (approved):

rami3l:
@djc Thanks for the quick response! Do you happen to have any info proving my point that the field is no longer than `123.45 MiB`? That's just from my observation so I'm not 100% sure about that :)

djc:
Seems unlikely that we'll have 1GB components, so probably fine for now?

rami3l:
Oh no, just found out myself and that's gross... `indicatif::HumanBytes(1023 * 1024).to_string()` gives `1023.00 KiB` so the maximum width is 11. ~~However this is going to make the correct gap look even bigger than in the above image. I'll see what this will actually look like in the current setting though.~~

Update: Looking at https://github.com/console-rs/indicatif/blob/95088ffd980a6f0cdd6571418fd9c6d6fd1603d1/src/style.rs#L740 it does look like I'll need to bump the width to 11. Sorry!

PR #4631, opened 2025-12-02
download: statically bundle relevant trust anchors

djc:
This seems to compile at least. I still want to give it another look, but happy to get feedback from others, too.

Let me know if/where it needs more comments.


rami3l:
@djc Not directly related, but should we establish a standard for generated files and include `src/dist/triple/known.rs`?

And, since I cannot compile the code (yet) so I'm not able to verify, maybe it's a good idea to disable `rustfmt` and/or `clippy` in the generated files in case anything changes in these two tools?

tests/suite/static_roots.rs
```diff
@@ -0,0 +1,172 @@
...
+        let root_cert = roots
+            .iter()
+            .find_map(|(cert_der, ta)| match ta == &root {
+                true => Some(cert_der),
+                false => None,
+            })
+            .unwrap();
+
+        code.push_str(&format!(
+            "    CertificateDer::from_slice(&{:?}),\n",
```
rami3l: I doubt whether that will give the right formatting. How about disabling formatting and/or lints altogether and just maintaining an OK-ish style here?

src/anchors.rs
```diff
@@ -0,0 +1,92 @@
+// @generated by tests/static_roots.rs
+use rustls::pki_types::CertificateDer;
+
+pub(crate) const RUSTUP_TRUST_ANCHORS: &[CertificateDer<'static>] = &[
+    CertificateDer::from_slice(&[
```
rami3l, line 6: Nit: Since you already have the hostnames, maybe it could be useful to indicate which is which, say with a comment?

tests/suite/static_roots.rs
```diff
@@ -0,0 +1,172 @@
...
+                ServerName::try_from(host).unwrap(),
+                TcpStream::connect((host, 443)).await.unwrap(),
+            )
+            .await
+            .unwrap();
+
+        let root = verifier.root.lock().unwrap().take().unwrap();
+        let root_cert = roots
+            .iter()
+            .find_map(|(cert_der, ta)| match ta == &root {
```
rami3l: Nit: This might be a good place to put `.then_some()` since the `match` isn't really doing too much.

tests/suite/static_roots.rs
```diff
@@ -0,0 +1,172 @@
...
+    for &host in HOSTS {
+        connector
+            .connect(
+                ServerName::try_from(host).unwrap(),
+                TcpStream::connect((host, 443)).await.unwrap(),
+            )
+            .await
+            .unwrap();
+
+        let root = verifier.root.lock().unwrap().take().unwrap();
```
rami3l, line 69: Nit: That's a lot of `.unwrap()` here. Maybe when this test function returns a result, the overall code can be shortened by quite a bit? Not 100% sure, but probably worth _testing_ though (pun intended) :)

djc: I tried that, but `tokio::test` doesn't seem to like it.

tests/suite/static_roots.rs
djc: Matched rustfmt's style without using it directly, seems okay?

src/anchors.rs
djc: Good idea, done.

tests/suite/static_roots.rs
djc: Done.

djc:
> @djc Not directly related, but should we establish a standard for generated files and include `src/dist/triple/known.rs`?

I think it's fine for now, we should re-evaluate if/when we add something else.

tests/suite/static_roots.rs
rami3l: @djc I see what you mean in that probably not all `.unwrap()`s can be eliminated, but they still have created some visual noise during the review...

Optionally, what about strategically eliminating at least a part of those? Something like...

```diff
diff --git a/tests/suite/static_roots.rs b/tests/suite/static_roots.rs
index eaa55f54..a0e3beb4 100644
--- a/tests/suite/static_roots.rs
+++ b/tests/suite/static_roots.rs
@@ -22,20 +22,19 @@ use webpki::{EndEntityCert, anchor_from_trusted_cert};
 use webpki_root_certs::TLS_SERVER_ROOT_CERTS;
 
 #[tokio::test]
-async fn store_static_roots() {
+async fn store_static_roots() -> anyhow::Result<()> {
     let provider = Arc::new(aws_lc_rs::default_provider());
     let mut root_store = RootCertStore::empty();
     let mut roots = Vec::with_capacity(TLS_SERVER_ROOT_CERTS.len());
     for cert_der in TLS_SERVER_ROOT_CERTS {
-        let ta = anchor_from_trusted_cert(cert_der).unwrap();
+        let ta = anchor_from_trusted_cert(cert_der)?;
         roots.push((cert_der, ta.clone()));
         root_store.roots.push(ta);
     }
 
     let root_store = Arc::new(root_store);
     let inner = WebPkiServerVerifier::builder_with_provider(root_store.clone(), provider.clone())
-        .build()
-        .unwrap();
+        .build()?;
 
     let verifier = Arc::new(TrackRootVerifier {
         root: Mutex::default(),
@@ -46,8 +45,7 @@ async fn store_static_roots() {
 
     let config = Arc::new(
         rustls::ClientConfig::builder_with_provider(provider)
-            .with_safe_default_protocol_versions()
-            .unwrap()
+            .with_safe_default_protocol_versions()?
             .dangerous()
             .with_custom_certificate_verifier(verifier.clone())
             .with_no_client_auth(),
@@ -60,11 +58,10 @@ async fn store_static_roots() {
     for &host in HOSTS {
         connector
             .connect(
-                ServerName::try_from(host).unwrap(),
-                TcpStream::connect((host, 443)).await.unwrap(),
+                ServerName::try_from(host)?,
+                TcpStream::connect((host, 443)).await?,
             )
-            .await
-            .unwrap();
+            .await?;
 
         let root = verifier.root.lock().unwrap().take().unwrap();
         let root_cert = roots
@@ -88,11 +85,13 @@ async fn store_static_roots() {
     }
     code.push_str("];\n");
 
-    let old = fs::read_to_string(PATH).unwrap();
+    let old = fs::read_to_string(PATH)?;
     if old != code {
-        fs::write(PATH, &code).unwrap();
+        fs::write(PATH, &code)?;
         panic!("anchors.rs is outdated; updated it");
     }
+
+    Ok(())
 }
 
 const PATH: &str = "src/anchors.rs";
```

rami3l (approved):

rami3l:
Many thanks :)

tests/suite/static_roots.rs
djc: As I said, `tokio::test` doesn't appear to support tests that return a `Result` type.

rami3l: @djc That's very strange given that the patch above works on my machine 🤔

djc: > @djc That's very strange given that the patch above works on my machine 🤔

Huh, weird. Feel free to PR it, it looks like an improvement to me!

PR #4637, opened 2025-12-11
dist(rustup-init/sh): update commit shasum in help string

rami3l:
Following #4513, this is the 2nd PR for the `1.29.0` beta release according to our [release process](https://rust-lang.github.io/rustup/dev-guide/release-process.html#making-a-release).

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -33,7 +33,7 @@ RUSTUP_QUIET=no
 # NOTICE: If you change anything here, please make the same changes in setup_mode.rs
 usage() {
     cat <<EOF
-rustup-init 1.28.2 (d1f31992a 2025-04-28)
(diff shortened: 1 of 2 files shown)


ChrisDenton (approved):

djc (approved):

rami3l:
Oh I forgot that now the snapshots are working differently than before... Will fix that real quick.

ChrisDenton:
Do we need to add that step to the release process docs?

rami3l:
> Do we need to add that step to the release process docs?

@ChrisDenton Good comment! Probably? But I prefer to do that in a subsequent PR in case there are anything else changing in the release process. It's been months so who knows what else need to change in the meantime :)

PR #4645, opened 2025-12-15
cli: add `doc --releases` to open release notes

cuviper:


rami3l (approved):
@cuviper LGTM! I think I can entrust you with testing out the patch locally?

djc (approved):

cuviper:
> I think I can entrust you with testing out the patch locally?

I did try it locally, yes. :)

I'd be happy to add a test as a followup if there's a recommended way of doing that.

rami3l:
@cuviper The actual effects of such changes are subject to the `opener` crate I believe, so there unfortunately aren't very good ways to test that with code. OTOH it should be okay as long as the relative path is exact.

Issue #4059, opened 2024-10-23
Tracking: Limit the range of valid toolchain names, take caution in `rust-toolchain.toml`


rami3l:
As discussed in https://github.com/rust-lang/rustup/issues/4053#issuecomment-2421879443, there are basically two issues to address:

1. Improve current error messages and treat toolchain names coming from `rust-toolchain.toml` as untrusted.
2. Come up with more restrictive rules in respect to legal toolchain names.
    - If the rules allow the prevention of phishing, lift the restriction on displaying toolchain names from `rust-toolchain.toml` if they are legal.

## Tasks

- [x] https://github.com/rust-lang/rustup/issues/4053

cachebag:
@rami3l 
I think we should be pretty conservative for custom toolchain names only (i.e. `[A-Za-z0-9._-]+`) while still rejecting
- none
- +prefix
- official/resolvable toolchain names
- names containing / or \

This IMO keeps common names a user might use and just removes whitepsaces, quotes, shell metacharacters, etc. that seemingly makes error messages or terminal output easier to weaponize. 

Now I guess this is technically a compatibility break for existing linked custom toolchains with names such as spaces or `#`/parentheses, since `ToolchainName::try_from` is also used when listing installed toolchains...

But given the goal of this issue is to make legal toolchain names clearly defined and safer to display, I think the allowlist is preferable to continuing with “anything except a few banned characters.”

What do you think?

rami3l:
@cachebag I think we are pretty much discussing 2 things with the naming restrictions here:
1. What should a "generally acceptable toolchain name" should look like.
    * I agree that indeed `[A-Za-z0-9._-]+` might already be enough for 99% of the cases. This seems to be a very good decision considering `[a-zA-Z_][a-zA-Z_0-9]*` from the [common identifier rule](https://craftinginterpreters.com/scanning.html#regular-languages-and-expressions) and `1.23.0-beta.2` from https://github.com/rust-lang/rustup/pull/3858).
    * However there might be some existing normalization rules about slashes that we might want to keep.
    * This is later categorized into official and custom toolchain names so in principle, as you have noticed above, the two cannot overlap. If you have found any counter-examples existing in our current logic, please don't hesitate to file another issue.
2. What names we want to reserve that may have special meaning in the future. For now we have `none`, but we may want more than that just in case.
[...]

PR #4650, opened 2025-12-17
fix(toolchain): forbid toolchain names starting with `+`

### LINKED ISSUE #4571: Error for `rustup run +stable cargo check` is misleading
### Verification

### Problem

As a user used to `+toolchain`, it can be easy to accidentally do that with `cargo run` but the error you get doesn't communicate that problem:
```console
$ rustup run +stable cargo check
error: toolchain '+stable' is not installed
```

### Steps

1. Run `rustup run +stable cargo -h`

### Possible Solution(s)

If `+` is reserved from being a leading character in toolchains: focus the error around the `+` prefix syntax being unsupported and provide a help message saying to drop it

If `+` is allowed as a leading character in toolchains: Check for a `+` prefix` and add a note saying that the `+` prefix syntax isn't used here and it can be dropped if thats the users intent

### Notes

_No response_

### Rustup version

```console
```

### Installed toolchains

```console
rustup home:  /home/epage/.rustup

installed toolchains
--------------------
beta-x86_64-unknown-linux-gnu
nightly-2023-11-13-x86_64-unknown-linux-gnu
nightly-2024-06-03-x86_64-unknown-linux-gnu
nightly-2024-09-10-x86_64-unknown-linux-gnu
nightly-2024-10-19-x86_64-unknown-linux-gnu
nightly-2025-01-27-x86_64-unknown-linux-gnu
nightly-2025-04-02-x86_64-unknown-linux-gnu
nightly-2025-06-26-x86_64-unknown-linux-gnu
1.0.0-x86_64-unknown-linux-gnu
1.48-x86_64-unknown-linux-gnu
1.48.0-x86_64-unknown-linux-gnu
1.56.0-x86_64-unknown-linux-gnu
1.57.0-x86_64-unknown-linux-gnu
1.58.0-x86_64-unknown-linux-gnu
1.60.0-x86_64-unknown-linux-gnu
1.64-x86_64-unknown-linux-gnu
1.64.0-x86_64-unknown-linux-gnu
1.65-x86_64-unknown-linux-gnu
1.65.0-x86_64-unknown-linux-gnu
1.66-x86_64-unknown-linux-gnu
1.66.0-x86_64-unknown-linux-gnu
1.70-x86_64-unknown-linux-gnu
1.71-x86_64-unknown-linux-gnu
1.71.0-x86_64-unknown-linux-gnu
1.71.1-x86_64-unknown-linux-gnu
1.72-x86_64-unknown-linux-gnu
1.74-x86_64-unknown-linux-gnu
1.75-x86_64-unknown-linux-gnu
1.76-x86_64-unknown-linux-gnu
1.80-x86_64-unknown-linux-gnu
1.80.1-x86_64-unknown-linux-gnu
1.81-x86_64-unknown-linux-gnu
1.83-x86_64-unknown-linux-gnu
1.83.0-x86_64-unknown-linux-gnu
1.84-x86_64-unknown-linux-gnu
1.85-x86_64-unknown-linux-gnu
1.86-x86_64-unknown-linux-gnu
1.86.0-x86_64-unknown-linux-gnu
1.88-x86_64-unknown-linux-gnu
1.89-x86_64-unknown-linux-gnu
1.90-x86_64-unknown-linux-gnu
stage1

active toolchain
----------------
installed targets:
  aarch64-apple-darwin
  i686-unknown-linux-gnu
  i686-unknown-linux-musl
  powerpc-unknown-linux-gnu
  wasm32-unknown-unknown
  x86_64-pc-windows-gnu
  x86_64-pc-windows-msvc
  x86_64-unknown-linux-gnu
  x86_64-unknown-linux-musl
  x86_64-unknown-none
```

### OS version

```console
Linux dross 6.16.3-76061603-generic #202508231538~1758561135~22.04~171c8de SMP PREEMPT_DYNAMIC Mon S x86_64 x86_64 x86
_64 GNU/Linux
```

cachebag:
Users accidentally use `+stable` with `rustup run`; in this PR we now forbid toolchain names starting with `+` globally.

closes #4571

### DIFF
--- a/src/toolchain/names.rs
+++ b/src/toolchain/names.rs
@@ -67,6 +67,8 @@ pub enum InvalidName {
     ToolchainPath(String),
     #[error("invalid toolchain name '{0}'")]
     ToolchainName(String),
+    #[error("invalid toolchain name '+{0}'; valid toolchain names do not start with '+'")]
+    PlusPrefix(String),
 }
 
 macro_rules! from_variant {
@@ -117,6 +119,9 @@ macro_rules! try_from_str {
 
 /// Common validate rules for all sorts of toolchain names
 fn validate(candidate: &str) -> Result<&str, InvalidName> {
+    if let Some(without_plus) = candidate.strip_prefix('+') {
+        return Err(InvalidName::PlusPrefix(without_plus.to_string()));
+    }
     let normalized_name = candidate.trim_end_matches('/');
     if normalized_name.is_empty() {
         Err(InvalidName::ToolchainName(candidate.into()))
@@ -501,8 +506,9 @@ mod tests {
 
     prop_compose! {
         fn arb_custom_name()
-            (s in r"[^\\/]+") -> String {
+            (s in r"[^\\/+][^\\/]*") -> String {
                 // perhaps need to filter 'none' and partial toolchains - but they won't typically be generated anyway.
+                // Also filter '+' prefix as that's reserved for +toolchain syntax.
                 s
         }
     }
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -2223,6 +2223,25 @@ error:[..] relative path toolchain '[..]/toolchains/nightly-[HOST_TRIPLE]'
 }
 
 #[tokio::test]
+async fn run_with_plus_prefix_fails_with_helpful_message() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+    cx.config
+        .expect(["rustup", "default", "stable"])
+        .await
+        .is_ok();
+
+    cx.config
+        .expect(["rustup", "run", "+stable", "cargo", "check"])
+        .await
+        .with_stderr(snapbox::str![[r#"
+...
+error:[..] invalid toolchain name '+stable'; valid toolchain names do not start with '+'
+...
+"#]])
+        .is_err();
+}
+
+#[tokio::test]
 async fn plus_override_abspath_is_supported() {
     let cx = CliTestContext::new(Scenario::SimpleV2).await;
     cx.config


src/toolchain/names.rs
```diff
@@ -67,6 +67,8 @@ pub enum InvalidName {
     ToolchainPath(String),
     #[error("invalid toolchain name '{0}'")]
     ToolchainName(String),
+    #[error("toolchain '{0}' is not installed\nhelp: remove the '+' prefix: `rustup run {1} ...`")]
```
cachebag: I didn't know how verbose or concise this should have been. I'm happy to change it if it's not quite what we want.

djc (approved):
This makes sense to me, thanks!

src/toolchain/names.rs
```diff
@@ -302,6 +304,12 @@ impl ResolvableLocalToolchainName {
 
     /// Validates if the string is a resolvable toolchain, or a path based toolchain.
     fn validate(candidate: &str) -> Result<Self, InvalidName> {
+        if let Some(without_plus) = candidate.strip_prefix('+') {
+            return Err(InvalidName::PlusPrefix(
+                candidate.to_string(),
```
rami3l: Nit: It seems that there's no need to store `+stable` and `stable` in two separate strings, only the latter would be enough?

rami3l (dismissed):

rami3l:
@djc Just confirming, currently it's far from common but we do allow toolchain names with the `+` prefixes right? I don't see anything about that in `rustup toolchain link --help` so I assume it's still valid.

If that is the case, then it's possible that we need to raise this error later in the pipeline, only when a toolchain with the said name is not found?

> As a user used to `+toolchain`, it can be easy to accidentally do that with `cargo run` but the error you get doesn't communicate that problem:
> 
> $ rustup run +stable cargo check
> error: toolchain '+stable' is not installed
_#4571_

djc:
From what I'm seeing `+` prefixes are only specifically supported at the top level (in both rustup and proxy mode). I definitely don't recall any details but I think I'm comfortable with the risk of these changes.

rami3l:
> From what I'm seeing `+` prefixes are only specifically supported at the top level (in both rustup and proxy mode). I definitely don't recall any details but I think I'm comfortable with the risk of these changes.

@djc Sorry for not being clear enough previously, but what I wanted to say is about the edge case of a toolchain name starting with `+`, say `+foo`, so that the calls would be written as `cargo ++foo ...`.

I know this sounds crazy and should be banned in https://github.com/rust-lang/rustup/issues/4059, but IIRC for now such names are still valid 😓

```console
> rustup run +foo cargo
error: toolchain '+foo' is not installed

> rustup ++foo show
[...]
error: toolchain '+foo' is not installed
```

... totally coherent 😱

djc:
Ahh. Let's not worry about it and/or ban it now?

rami3l (changes requested):
@djc Yeah, why not? I'd expect this breakage will be relatively small and nearly no one will actually notice it.

@cachebag How about just banning all toolchain names starting with `+` right now?

Instead of the current changes, put the general part in the `fn validate(candidate: &str) -> Result<&str, InvalidName>;` function (that's where common rules are defined), ~~and put the hint specific to #4571 in `rustup run`~~ and just suggest using the toolchain name without that prefix, without being specific to `rustup run`.

rami3l (approved):
Many thanks! I'm okay with this as long as the name of the commit and the PR is changed to reflect the latest changes, say `` forbid toolchain names starting with `+` ``.

@djc What do you think?

cachebag:
> forbid toolchain names starting with `+`

Does this also steer from the original issue at hand?

rami3l:
Hmmm the original error message was confusing but made sense exactly because `+` has been permitted at the beginning of the toolchain names... So I won't call this a deviation.

Also, given the current error message, it should be easy to deduce that a toolchain name is expected as the positional argument here, rather than the `+`-prefixed form.

Besides, I think this is also an opportunity for us to test out what toolchain names are actually used so at least we'd have some expectation of what can be broken.

Finally, this could allow smarter guesses to happen, since we are sure that the `+` prefix here is not conveying any particular intention.

- For example, when `stable` is installed we can eventually issue something like: "`+stable` is not a valid toolchain name, did you mean `stable`?" which IMHO would be a great follow-up to do.

rami3l (approved):

rami3l:
As I already said, this looks good to me as long as @djc feels the same. However we could potentially use more prudence in merging yet another breaking change between beta and stable... I'd currently prefer playing the safe card and postpone this to the upcoming 2.0 👀

djc (approved):

djc:
> However we could potentially use more prudence in merging yet another breaking change between beta and stable... I'd currently prefer playing the safe card and postpone this to the upcoming 2.0 👀

I mean, do you have a concrete plan for 2.0? When will that be released? I still think the version number is not going to make much difference in how much testing/caution people exercise before/when upgrading.

rami3l:
> I mean, do you have a concrete plan for 2.0? When will that be released? I still think the version number is not going to make much difference in how much testing/caution people exercise before/when upgrading.

@djc Thank you for this constructive question.

I'm sorry but I'm not sure if we are on the same page here; I think my stance with breaking changes will not change significantly whether the next non-patch release will be `1.30` or `2.0`, as it is in theory more dangerous after a beta release has been cut.

Of course, in terms of actual impact this should be considered a minor breakage, so I'm also okay with merging this patch immediately, as long as enough attention has been drawn. Your call!

djc:
Given the amount of response on internals I have no reason to feel the beta has resulted in a substantial amount of testing. Maybe we should do a post to the Inside Rust blog?

rami3l:
> Given the amount of response on internals I have no reason to feel the beta has resulted in a substantial amount of testing. Maybe we should do a post to the Inside Rust blog?

@djc Yes, that's a good idea! I'll take care of it.

Besides, after giving this PR another look, I think I'm probably worrying too much again over a small change that can be easily reverted in the next patch release. My apologies!

PR #4651, opened 2025-12-19
Add common commands section in help text

dingxiangfei2009:
Judging from user feedback coming out of Debian distros, it appears that after installing `rustup` from the package manager it is not immediately clear which command actually installs the toolchain. It is true that `rustup` from `rustup.rs` one-liner makes one-stop installation, Debian packaging actually skips the essential `rustup-init` step as of writing for a good reason. This left a few Debian users, who I talked to recently, feel lost in their first attempt to use the toolchain.

This patch adds a small text so that after installation of `rustup`, which normally does not invoke the toolchain installation for the user, the user can still receive guidance on how to set up the local toolchain immediately.

### DIFF
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -11,7 +11,18 @@ pub(crate) fn rustup_help() -> String {
   for common platforms.
 
   If you are new to Rust consider running `rustup doc --book` to
-  learn Rust."
+  learn Rust.
+
+{HEADER}Common commands:{HEADER:#}
+
+  Update Rust toolchains and rustup
+
+    {LITERAL}$ rustup update{LITERAL:#}
+    
+  Install the current stable release of Rust for your host platform
+
+    {LITERAL}$ rustup toolchain install stable{LITERAL:#}
+  "
     )
 }
 
--- a/tests/suite/cli_rustup_ui/rustup_help_cmd.stdout.term.svg
+++ b/tests/suite/cli_rustup_ui/rustup_help_cmd.stdout.term.svg
@@ -1,4 +1,4 @@
-<svg width="776px" height="830px" xmlns="http://www.w3.org/2000/svg">
+<svg width="776px" height="1010px" xmlns="http://www.w3.org/2000/svg">
   <style>
     .fg { fill: #AAAAAA }
     .bg { fill: #000000 }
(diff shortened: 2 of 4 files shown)


djc:
I think it would be good to try and address this problem, but not sure adding a "common commands" section to the end is the best way to do it. Wonder if we can add this to the top, either as part of `Usage:` or even before that.

rami3l:
> I think it would be good to try and address this problem, but not sure adding a "common commands" section to the end is the best way to do it.

I would like to disagree with @djc on that matter, as when a command is executed, the first thing I see is the last lines of its output, which is different from when reading a normal doc page.

---

OTOH I do think the case for Debian is still somewhat particular, since installing via APT technically still has quite limited support notably because it explicitly bypasses the `rustup-init` stage.

We have tried to improve the situation in both our [docs](https://rust-lang.github.io/rustup/installation/other.html#general-tips) and, in my case, its homebrew packaging:

```console
> brew install rustup
[..]
==> Caveats
To initialize `rustup`, set a default toolchain:
  rustup default stable

If you have `rust` installed, ensure you have "$(brew --prefix rustup)/bin"
before "$(brew --prefix)/bin" in your $PATH:
  https://rust-lang.github.io/rustup/installation/already-installed-rust.html
```

So the policy that I had in the past would be to acknowledge this as the package manager/maintainer's responsibility. Can something similar be done on Debian's side?

src/cli/help.rs
```diff
@@ -11,7 +11,18 @@ pub(crate) fn rustup_help() -> String {
...
-  learn Rust."
+  learn Rust.
+
+{HEADER}Common commands:{HEADER:#}
+
+    {LITERAL}$ rustup update{LITERAL:#}
+  
+  Update Rust toolchains and rustup
+
+    {LITERAL}$ rustup toolchain install{LITERAL:#}
```
rami3l: That's `$ rustup toolchain install stable` since you've mentioned the stable release.

dingxiangfei2009: Applied in the latest patch.

dingxiangfei2009:
I would also prefer to leave the common commands as the last lines of the help text, so that this information does not scroll away in the terminal print out.

As for Debian, I am currently researching a way to update the package description of `rustup` so that one can be guided to initiate the missing `init` step, like how `homebrew` guides the users.

src/cli/help.rs
```diff
@@ -11,7 +11,18 @@ pub(crate) fn rustup_help() -> String {
...
 
   If you are new to Rust consider running `rustup doc --book` to
-  learn Rust."
+  learn Rust.
+
+{HEADER}Common commands:{HEADER:#}
+
+    {LITERAL}$ rustup update{LITERAL:#}
+  
+  Update Rust toolchains and rustup
```
rami3l, line 18: Nit: I've just compared that with `tldr` where we have descriptions followed by commands, and that's more intuitive IMHO:

```console
> tldr cargo
[..]
  Manage Rust projects and their module dependencies (crates).
  Some subcommands such as `build` have their own usage documentation.
  More information: <https://doc.rust-lang.org/stable/cargo/>.

  Search for crates:

      cargo search search_string

  Install a binary crate:

      cargo install crate_name
[..]
```

@epage Does `cargo` have a similar section that we need to align format with?

rami3l (approved):
LGTM modulo one tiny bit :)

src/cli/help.rs
dingxiangfei2009: Yeah it looks much better. Suggestion applied.

rami3l: @dingxiangfei2009 Thanks! Let's wait for the final confirmation to see if there's any alignments to be made and I believe we are done :)

rami3l: @epage has suggested looking at `cargo help add`. I'd say it's more or less in the same vein so I'm in favor of merging this as-is. Many thanks :)

PR #4664, opened 2025-12-30
Upgrade to reqwest 0.13

djc:


ChrisDenton:
Looking at the lockfile it seems we are using two different versions of reqwest after this? Also, that is a lot more dependencies but maybe that can't be helped.

djc:
> Looking at the lockfile it seems we are using two different versions of reqwest after this? Also, that is a lot more dependencies but maybe that can't be helped.

What do you mean "a lot more dependencies"? That's not what I'm seeing in the `Cargo.lock` file.

Yes, some of the opentelemetry crates are still using reqwest 0.12. I was actually working on a PR for that project but the `otel` feature is not enabled by default anyway (I'm still in favor of just removing it).

rami3l (approved):

PR #4665, opened 2025-12-30
Drop release builds from CI

djc:
CI is already super slow, and I don't think these are likely to add value.


djc:
I guess we archive those builds so probably not the best idea. Also we already seem to skip tests for release build workflows.

rami3l:
@djc I think release builds at least are not the bottleneck? What has been consuming a lot of time is the test suite I believe.

PR #4670, opened 2026-01-06
Unify nightly component unavailable messages

### LINKED ISSUE #4667: Unify 2 versions of "sometimes not all components are available"
Before https://github.com/rust-lang/rustup/commit/983e9faacf5fef9636c11313f16cf2fafb96f91e this was actually coherent (cc @djc).

IMHO this is more of a `note:` kind of message (but made probably before the logging convention was fully established), otherwise it doesn't make too much sense to put a newline before it? 

https://github.com/rust-lang/rustup/blob/de2b979bd4569cacfabf501bb58464c9139b0d6d/src/dist/mod.rs#L87-L93

https://github.com/rust-lang/rustup/blob/de2b979bd4569cacfabf501bb58464c9139b0d6d/src/errors.rs#L205-L210

It'd be better if we can think of an appropriate way to unify those.

enthropy7:
fixes https://github.com/rust-lang/rustup/issues/4667

Unified two versions of the "sometimes not all components are available" message for nightly builds. Added `nightly_component_unavailable_help()` to generate a consistent `note/help` format. Removed parentheses from the error string in `component_unavailable_msg()` - error is separate, explanation via `note/help`. Replaced the long text block in `components_missing_msg()` with structured `note/help` lines. Updated tests, all code paths use single message in a consistent format, eliminating duplication.

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -21,7 +21,7 @@ use tracing::{debug, info, warn};
 
 use crate::{
     config::Cfg,
-    errors::RustupError,
+    errors::{NIGHTLY_COMPONENT_NOTE, RustupError},
     process::Process,
     toolchain::{DistributableToolchain, ToolchainName},
     utils,
@@ -87,22 +87,15 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
     if toolchain.starts_with("nightly") {
         let _ = write!(
             buf,
-            "\
-Sometimes not all components are available in any given nightly.
-If you don't need these components, you could try a minimal installation with:
-
-    rustup toolchain add {toolchain} --profile minimal
-
-If you require these components, please install and use the latest successfully built version,
-which you can find at <https://rust-lang.github.io/rustup-components-history>.
-
-After determining the correct date, install it with a command such as:
-
-    rustup toolchain install nightly-2018-12-27
-
-Then you can use the toolchain with commands such as:
-
-    cargo +nightly-2018-12-27 build"
+            "\n{NIGHTLY_COMPONENT_NOTE}\n\
+        help: if you don't need these components, you could try a minimal installation with:\n\
+        help:     rustup toolchain add {toolchain} --profile minimal\n\
+        help: if you require these components, please install and use the latest successfully built version,\n\
+        help: which you can find at <https://rust-lang.github.io/rustup-components-history>\n\
+        help: after determining the correct date, install it with a command such as:\n\
+        help:     rustup toolchain install nightly-2018-12-27\n\
+        help: then you can use the toolchain with commands such as:\n\
+        help:     cargo +nightly-2018-12-27 build"
         );
     } else if ["beta", "stable"].iter().any(|&p| toolchain.starts_with(p)) {
         let _ = write!(
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -181,6 +181,9 @@ fn suggest_message(suggestion: &Option<String>) -> String {
     }
 }
 
+pub(crate) const NIGHTLY_COMPONENT_NOTE: &str =
+    "note: sometimes not all components are available in any given nightly";
+
 /// Returns a error message indicating that certain [`Component`]s are unavailable.
 ///
 /// See also [`components_missing_msg`](../dist/dist/fn.components_missing_msg.html)
@@ -203,10 +206,7 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
             );
 
             if toolchain.starts_with("nightly") {
-                let _ = write!(
-                    buf,
-                    "(sometimes not all components are available in any given nightly)"
-                );
+                let _ = write!(buf, "{NIGHTLY_COMPONENT_NOTE}");
             }
         }
         cs => {
@@ -227,16 +227,13 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
                     .join(", ")
             };
 
-            let _ = write!(
+            let _ = writeln!(
                 buf,
-                "some components are unavailable for download for channel '{toolchain}': {cs_str}"
+                "some components are unavailable for download for channel '{toolchain}': {cs_str}",
             );
 
             if toolchain.starts_with("nightly") {
-                let _ = write!(
-                    buf,
-                    "(sometimes not all components are available in any given nightly)"
-                );
+                let _ = write!(buf, "{NIGHTLY_COMPONENT_NOTE}");
             }
         }
     }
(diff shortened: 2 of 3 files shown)


rami3l (changes requested):
Thanks for your interest and your contribution!

As seen from the test suite, we essentially have 3 different tests for 3 different error messages served for different purposes. And a fix for #4667 shouldn't change that fact.

The styling of these 3 message could indeed use some unification though.

enthropy7:
> Thanks for your interest and your contribution!
> 
> As seen from the test suite, we essentially have 3 different tests for 3 different error messages served for different purposes. And a fix for #4667 shouldn't change that fact.
> 
> The styling of these 3 message could indeed use some unification though.

thanks you for your review! upd: kept distinct messaging per scenario and only unified the nightly disclaimer wording/styling via a shared note/help helper. also updated other related stuff

src/errors.rs
```diff
@@ -203,10 +222,7 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
             );
 
             if toolchain.starts_with("nightly") {
-                let _ = write!(
```
rami3l, line 206: Replacing `write!()` with `writeln!()` is unnecessary, since the original extended notes already included the required newlines.

src/errors.rs
```diff
@@ -227,16 +243,13 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
                     .join(", ")
             };
 
-            let _ = write!(
+            let _ = writeln!(
```
rami3l, line 230: Replacing `write!()` with `writeln!()` is unnecessary, since the original extended notes already included the required newlines.

src/dist/mod.rs
```diff
@@ -87,22 +87,8 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
...
-
-After determining the correct date, install it with a command such as:
-
-    rustup toolchain install nightly-2018-12-27
-
-Then you can use the toolchain with commands such as:
-
-    cargo +nightly-2018-12-27 build"
+            "{}",
+            nightly_component_unavailable_help_toolchain(toolchain)
```
rami3l: This function is called only once, it'd be better to just inline this call.

rami3l (changes requested):

enthropy7:
thanks, fixed :3

tests/suite/cli_v2.rs
```diff
@@ -1972,7 +1972,7 @@ async fn add_missing_component() {
         .with_stderr(snapbox::str![[r#"
 ...
 error: component 'rls' for target '[HOST_TRIPLE]' is unavailable for download for channel 'nightly'
-(sometimes not all components are available in any given nightly)
+
```
rami3l: There should've been a `note: ` line... did the logic break somewhere?

src/dist/mod.rs
```diff
@@ -87,22 +87,17 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
...
+        help: if you don't need these components, you could try a minimal installation with:\n\
+        help:     rustup toolchain add {toolchain} --profile minimal\n\
+        help: if you require these components, please install and use the latest successfully built version,\n\
+        help: which you can find at <https://rust-lang.github.io/rustup-components-history>\n\
+        help: after determining the correct date, install it with a command such as:\n\
+        help:     rustup toolchain install nightly-2018-12-27\n\
+        help: then you can use the toolchain with commands such as:\n\
+        help:     cargo +nightly-2018-12-27 build\n",
+            NIGHTLY_COMPONENT_NOTE = NIGHTLY_COMPONENT_NOTE,
+            toolchain = toolchain
```
rami3l: It seems to me that these two lines can be omitted.

enthropy7:
fixed

src/dist/mod.rs
```diff
@@ -87,22 +87,15 @@ fn components_missing_msg(cs: &[Component], manifest: &ManifestV2, toolchain: &s
...
-    cargo +nightly-2018-12-27 build"
+            "\n{NIGHTLY_COMPONENT_NOTE}\n\
+        help: if you don't need these components, you could try a minimal installation with:\n\
+        help:     rustup toolchain add {toolchain} --profile minimal\n\
+        help: if you require these components, please install and use the latest successfully built version,\n\
+        help: which you can find at <https://rust-lang.github.io/rustup-components-history>\n\
+        help: after determining the correct date, install it with a command such as:\n\
+        help:     rustup toolchain install nightly-2018-12-27\n\
+        help: then you can use the toolchain with commands such as:\n\
+        help:     cargo +nightly-2018-12-27 build\n"
```
rami3l: That's one extra `\n` at the end compared to the original.

rami3l (approved):
A few nits (one of which comes from the previous round) and this seems ready to go. Thanks a lot :)

tests/suite/cli_v2.rs
```diff
@@ -1972,7 +1972,8 @@ async fn add_missing_component() {
...
```
rami3l: There is one extra newline, not sure if it's related to the previous comments...

src/errors.rs
```diff
@@ -203,10 +206,7 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
             );
 
             if toolchain.starts_with("nightly") {
-                let _ = write!(
-                    buf,
-                    "(sometimes not all components are available in any given nightly)"
-                );
+                let _ = write!(buf, "\n{NIGHTLY_COMPONENT_NOTE}");
```
rami3l: No leading `\n` here.

tests/suite/cli_v2.rs
rami3l: See https://github.com/rust-lang/rustup/pull/4670#discussion_r2669096244.

src/errors.rs
```diff
@@ -227,16 +227,13 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
...
-                "some components are unavailable for download for channel '{toolchain}': {cs_str}"
+                "some components are unavailable for download for channel '{toolchain}': {cs_str}",
             );
 
             if toolchain.starts_with("nightly") {
-                let _ = write!(
-                    buf,
-                    "(sometimes not all components are available in any given nightly)"
-                );
+                let _ = write!(buf, "\n{NIGHTLY_COMPONENT_NOTE}");
```
rami3l: No leading `\n` here since you have done a `writeln!()` before.

src/errors.rs
rami3l: Resolved in favor of https://github.com/rust-lang/rustup/pull/4670#discussion_r2669106635.

enthropy7:
glad that i could help :3

PR #1521, opened 2018-10-04
Fix `utils::copy_file` for symlink.

reitermarkus:
When `rustup-init` is a symlink, create a symlink to that symlink instead of resolving it first. This is also in line with [using the `no-self-update` feature](https://github.com/Homebrew/homebrew-core/pull/32701), to allow managing the `rustup` version with Homebrew.

Fixes https://github.com/rust-lang-nursery/rustup.rs/issues/1512.

PR #4671, opened 2026-01-06
fix: `copy_dir` and `copy_file` to preserve symlinks instead of following them

cachebag:
This PR employs symlink preservation in `copy_file` and adds it to `copy_dir`.

PR #1504 added symlink preservation to fix crashes in lldb-preview, which contains internal symlinks (liblldb was loading twice due to broken symlinks). Then, PR #1521 later changed `copy_file` to create symlinks pointing  to the source path (for Homebrew integration), which may have regressed #1504. 

This PR restores #1504's behavior: read the symlink target and preserve it. What I'm curious about is what the intended behavior for #1521 was because shouldn't users be able to run `rustup-init --no-self-update`? Is that not the recommended way to manage `rustup` via external package managers?

If this PR steers out of scope of the goals of `rustup` I'm happy to close it but I feel like the former behavior is preferable for most use cases.

### DIFF
--- a/src/dist/component/tests.rs
+++ b/src/dist/component/tests.rs
@@ -492,3 +492,208 @@ fn rollback_failure_keeps_going() {
 #[test]
 #[ignore]
 fn intermediate_dir_rollback() {}
+
+#[test]
+#[cfg(unix)]
+fn copy_dir_preserves_symlinks() {
+    // copy_dir must preserve symlinks, not follow them
+    use std::os::unix::fs::symlink;
+
+    let cx = DistContext::new(None).unwrap();
+    let mut tx = cx.transaction();
+
+    let src_dir = cx.pkg_dir.path();
+
+    let src_real_file = src_dir.join("real_file.txt");
+    utils::write_file("", &src_real_file, "original content").unwrap();
+
+    let src_subdir = src_dir.join("subdir");
+    fs::create_dir(&src_subdir).unwrap();
+
+    let src_subdir_link_to_file = src_subdir.join("link_to_file.txt");
+    symlink("../real_file.txt", &src_subdir_link_to_file).unwrap();
+
+    let src_real_dir = src_dir.join("real_dir");
+    fs::create_dir(&src_real_dir).unwrap();
+    utils::write_file("", &src_real_dir.join("inner.txt"), "inner content").unwrap();
+    let src_subdir_link_to_dir = src_subdir.join("link_to_dir");
+    symlink("../real_dir", &src_subdir_link_to_dir).unwrap();
+
+    assert!(
+        fs::symlink_metadata(&src_subdir_link_to_file)
+            .unwrap()
+            .file_type()
+            .is_symlink(),
+        "Source file symlink should be a symlink"
+    );
+    assert!(
+        fs::symlink_metadata(&src_subdir_link_to_dir)
+            .unwrap()
+            .file_type()
+            .is_symlink(),
+        "Source dir symlink should be a symlink"
+    );
+
+    tx.copy_dir("test-component", PathBuf::from("dest"), src_dir)
+        .unwrap();
+    tx.commit();
+
+    let dest_file_symlink = cx.prefix.path().join("dest/subdir/link_to_file.txt");
+    let dest_dir_symlink = cx.prefix.path().join("dest/subdir/link_to_dir");
+
+    assert!(
+        fs::symlink_metadata(&dest_file_symlink)
+            .unwrap()
+            .file_type()
+            .is_symlink(),
+        "Destination file symlink should be preserved as a symlink"
+    );
+    assert!(
+        fs::symlink_metadata(&dest_dir_symlink)
+            .unwrap()
+            .file_type()
+            .is_symlink(),
+        "Destination dir symlink should be preserved as a symlink"
+    );
+
+    assert_eq!(
+        fs::read_link(&dest_file_symlink).unwrap().to_str().unwrap(),
+        "../real_file.txt",
+        "File symlink target should be preserved"
+    );
+    assert_eq!(
+        fs::read_link(&dest_dir_symlink).unwrap().to_str().unwrap(),
+        "../real_dir",
+        "Dir symlink target should be preserved"
+    );
+}
+
+/// Test that utils::copy_file preserves symlink targets
+#[test]
+#[cfg(unix)]
+fn copy_file_preserves_symlinks() {
+    use std::os::unix::fs::symlink;
+
+    let tmp = tempfile::tempdir().unwrap();
+    let src_dir = tmp.path().join("src");
+    let dest_dir = tmp.path().join("dest");
+    fs::create_dir_all(&src_dir).unwrap();
+    fs::create_dir_all(&dest_dir).unwrap();
+
+    let src_real_file = src_dir.join("real_file.txt");
+    utils::write_file("", &src_real_file, "content").unwrap();
+
+    let src_link_file = src_dir.join("link.txt");
+    symlink("real_file.txt", &src_link_file).unwrap();
+
+    assert!(
+        fs::symlink_metadata(&src_link_file)
+            .unwrap()
+            .file_type()
+            .is_symlink()
+    );
+    assert_eq!(
+        fs::read_link(&src_link_file).unwrap().to_str().unwrap(),
+        "real_file.txt"
+    );
+
+    // copy_file should preserve the symlink target
+    let dest_link_file = dest_dir.join("link.txt");
+    utils::copy_file(&src_link_file, &dest_link_file).unwrap();
+
+    assert!(
+        fs::symlink_metadata(&dest_link_file)
+            .unwrap()
+            .file_type()
+            .is_symlink(),
+        "copy_file should preserve symlinks"
+    );
+    assert_eq!(
+        fs::read_link(&dest_link_file).unwrap().to_str().unwrap(),
+        "real_file.txt",
+        "copy_file should preserve the original symlink target"
+    );
+}
+
+/// Test that utils::copy_file_symlink_to_source creates a symlink pointing to the source path
+#[test]
+#[cfg(unix)]
+fn copy_file_symlink_to_source_creates_symlink_to_source() {
+    use std::os::unix::fs::symlink;
+
+    let tmp = tempfile::tempdir().unwrap();
+    let src_dir = tmp.path().join("src");
+    let dest_dir = tmp.path().join("dest");
+    fs::create_dir_all(&src_dir).unwrap();
+    fs::create_dir_all(&dest_dir).unwrap();
+
+    let src_real_file = src_dir.join("real_file.txt");
+    utils::write_file("", &src_real_file, "original content").unwrap();
+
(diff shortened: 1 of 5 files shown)


rami3l:
Thanks for making this patch! Reasonable support for external packagers are still expected, and whether a manual setup is required is largely based on the packagers' opinion: https://rust-lang.github.io/rustup/installation/other.html#using-a-package-manager

I have the feeling that we need to somehow support both cases based on whether that's copying the original binary or not...

src/utils/raw.rs
```diff
@@ -253,12 +253,29 @@ pub(crate) fn copy_dir(src: &Path, dest: &Path) -> io::Result<()> {
     for entry in src.read_dir()? {
         let entry = entry?;
         let kind = entry.file_type()?;
-        let src = entry.path();
-        let dest = dest.join(entry.file_name());
-        if kind.is_dir() {
-            copy_dir(&src, &dest)?;
+        let src_path = entry.path();
```
rami3l: Nit: `src` and `dest` bindings in the loop body are preventing the input params from being used. Thus, I don't see a very convincing reason to rename them here.

rami3l (commented):
Do you have a specific use case where rustup is not working as you'd expect, which this PR allows to fix?

If you can find such a case, I agree with using this new implementation except for this:

https://github.com/rust-lang/rustup/blob/3382f6f7bf09e591bbda4bcda7daf2280a0061d9/src/cli/self_update.rs#L845

#1521 is probably too widespread as a fix, and it makes perfect sense to use different  copy implementations during self installation and during toolchain installation.

Suggest refining the test cases accordingly if going this way.

cachebag:
> Do you have a specific use case where rustup is not working as you'd expect, which this PR allows to fix?

I don't, and I only encountered this looking at the code. It's my understanding that copying symlinks breaks their identity and can cause software to load libraries twice (like the crash that was fixed ad-hoc above). It's also just correct behavior afaik; `cp -R` preserves symlinks by default.
 
> https://github.com/rust-lang/rustup/pull/1521 is probably too widespread as a fix, and it makes perfect sense to use different copy implementations during self installation and during toolchain installation.

I would agree with this and will adjust, so long as you think it's a reasonable patch. I figured at the least I could flag this to note the variance in functionality for copy's.

rami3l:
@cachebag Thanks for your clarification. I think although it has no direct influence now, judging by what I'd expect from the function API then indeed, the old behavior feels more correct to me, and leaving it uncorrected might lead to misuse in the long run.

Looking back there is still a slight possibility that we could be missing other cases apart from self installation, so I'd say let's go on polishing this patch by carefully analyzing the call sites. Please don't hesitate to ask me if you don't feel certain about anything!

cachebag:
> @cachebag Thanks for your clarification. I think although it has no direct influence now, judging by what I'd expect from the function API then indeed, the old behavior feels more correct to me, and leaving it uncorrected might lead to misuse in the long run.
> 
> 
> 
> Looking back there is still a slight possibility that we could be missing other cases apart from self installation, so I'd say let's go on polishing this patch by carefully analyzing the call sites. Please don't hesitate to ask me if you don't feel certain about anything!

That makes sense to me, thank you!

src/utils/mod.rs
```diff
@@ -232,7 +232,38 @@ pub(crate) fn copy_file(src: &Path, dest: &Path) -> Result<()> {
...
+                    dest.display()
+                )
+            })
+            .map(|_| ())
+    }
+}
+
+/// Copy a file, preserving symlink targets instead of creating a link to the source.
+/// We use this for component installation where internal symlink structure must be maintained.
+pub(crate) fn copy_file_preserving_symlink(src: &Path, dest: &Path) -> Result<()> {
```
rami3l: 1. If I'm not mistaken, the new function you have introduced is identical to the old function you have modified, which means you are not distinguishing the two approaches.

   To align with the title of your PR, the symlink-preserving `copy_file` should just be named `copy_file()`. Also, it might not be worth it to duplicate code here since the fact that the function is duplicated without any changes is not very obvious in my review already; try to DRY it up a bit, for example by making `copy_file()` and `copy_file_symlink_to_source()` shorthands of `copy_file_impl(preserve_symlink: bool)` calls.

2. Also, the tests are not enough to prove the fact that the both functions are working as intended.

    You might need to add more tests to demonstrate this point.

rami3l (changes requested):

src/utils/mod.rs
cachebag: Hopefully I followed that train of thought correctly. I should have led with the unified path instead, sorry about that.

As far as tests, do you have suggestions on how to approach that, or specific cases I should be covering?

src/utils/mod.rs
```diff
@@ -226,13 +226,29 @@ pub(crate) fn copy_dir(src: &Path, dest: &Path) -> Result<()> {
...
         path: PathBuf::from(src),
     })?;
-    if metadata.file_type().is_symlink() {
-        symlink_file(src, dest).map(|_| ())
+    if preserve_symlink && metadata.file_type().is_symlink() {
+        let target = fs::read_link(src).with_context(|| RustupError::ReadingFile {
+            name: "symlink target for",
+            path: PathBuf::from(src),
+        })?;
+        symlink_file(&target, dest).map(|_| ())
```
rami3l: Judging by the diff, the current version doesn't seem like a refactor? Shouldn't this be something like:

```suggestion
    if metadata.file_type().is_symlink() {
        let target = if preserve_symlink {
            &fs::read_link(src).with_context(|| RustupError::ReadingFile {
                name: "symlink target for",
                path: PathBuf::from(src),
            })?;
        } else {
           src
        };
        symlink_file(&target, dest).map(|_| ())
```

src/utils/mod.rs
rami3l: @cachebag That depends on where you have used the old behavior (`utils::copy_file_symlink_to_source()`). No unit tests needed, but regression tests must make sure that the end-to-end tests behave as before for those cases.

src/utils/mod.rs
cachebag: @rami3l Actually, I don't know. Your suggestion would be creating symlinks that point back to `src` rather than preserving the original target. 

Unless I am misunderstanding something, `fs::copy()` correctly follows the symlink and copy the real file content when `preserve_symlink=false`. Isn't that what we want if we want to re-instate the general behavior of #1521 ?

rami3l: @cachebag I believe there is some misunderstanding going on here, my apologies. #1521 has explicitly stated "creating a symlink to the old symlink" so we wouldn't want to copy the real file content here?

cachebag: > @cachebag I believe there is some misunderstanding going on here, my apologies. #1521 has explicitly stated "creating a symlink to the old symlink" so we wouldn't want to copy the real file content here?

No, I think the misunderstanding was on my part. I misinterpreted what you were suggesting, your restructuring makes sense, sorry about that.

src/utils/mod.rs
```diff
@@ -226,13 +226,33 @@ pub(crate) fn copy_dir(src: &Path, dest: &Path) -> Result<()> {
     })
 }
 
+/// Copy a file, preserving symlink targets instead of following them.
+/// This is the default behavior for component installation.
 pub(crate) fn copy_file(src: &Path, dest: &Path) -> Result<()> {
+    copy_file_impl(src, dest, true)
+}
+
+/// Copy a file, creating a symlink pointing to the source path if src is a symlink.
```
rami3l: Nit: Maybe the wording here can be a bit clearer...

```suggestion
/// Copy a file from `src` to `dst`, or if `src` is a symlink, create a new symlink
/// at `dst` pointing to it.
```

src/utils/mod.rs
```diff
@@ -226,13 +226,33 @@ pub(crate) fn copy_dir(src: &Path, dest: &Path) -> Result<()> {
...
     })?;
     if metadata.file_type().is_symlink() {
-        symlink_file(src, dest).map(|_| ())
+        let target = if preserve_symlink {
+            fs::read_link(src).with_context(|| RustupError::ReadingFile {
+                name: "symlink target for",
+                path: PathBuf::from(src),
+            })?
+        } else {
+            src.to_path_buf()
```
rami3l: Nit: We should be able to save a clone here...

```rs
        let target = if preserve_symlink {
            &fs::read_link(src).with_context(|| RustupError::ReadingFile {
                name: "symlink target for",
                path: PathBuf::from(src),
            })?
        } else {
            src
        };
        symlink_file(target, dest).map(|_| ())
```

src/dist/component/tests.rs
```diff
@@ -492,3 +492,208 @@ fn rollback_failure_keeps_going() {
...
+fn copy_dir_preserves_symlinks() {
+    // copy_dir must preserve symlinks, not follow them
+    use std::os::unix::fs::symlink;
+
+    let cx = DistContext::new(None).unwrap();
+    let mut tx = cx.transaction();
+
+    let src_dir = cx.pkg_dir.path();
+
+    let real_file = src_dir.join("real_file.txt");
```
rami3l: Nit: Since the actual test file hierarchy might take a bit of time to grasp, suggest mirroring the path names using the variable names, for example `src_real_file` instead of `real_file`, `src_subdir_link_to_file` instead of `file_symlink`.

Idem for the rest of the new tests.

src/utils/mod.rs
```diff
@@ -226,13 +226,33 @@ pub(crate) fn copy_dir(src: &Path, dest: &Path) -> Result<()> {
     })
 }
 
+/// Copy a file, preserving symlink targets instead of following them.
```
rami3l: Nit: For the sake of coherence:

```suggestion
/// Copy a file from `src` to `dst`, preserving the symlink target if `src` is a symlink.
```

rami3l (approved):
LGTM modulo a few lints, many thanks!

src/utils/mod.rs
```diff
@@ -226,13 +227,36 @@ pub(crate) fn copy_dir(src: &Path, dest: &Path) -> Result<()> {
...
+}
+
+fn copy_file_impl(src: &Path, dest: &Path, preserve_symlink: bool) -> Result<()> {
     let metadata = fs::symlink_metadata(src).with_context(|| RustupError::ReadingFile {
         name: "metadata for",
         path: PathBuf::from(src),
     })?;
     if metadata.file_type().is_symlink() {
-        symlink_file(src, dest).map(|_| ())
+        let target: Cow<'_, Path> = if preserve_symlink {
```
djc: Don't think we need this type annotation?

djc (approved):

src/utils/mod.rs
rami3l: @djc Exactly, I tried https://github.com/rust-lang/rustup/pull/4671#discussion_r2706213044 locally and seems to work.

src/utils/mod.rs
rami3l: @cachebag Again, I think there is no need to convert to `Cow` here, lifetime extension seems to work normally.

rami3l (approved):

PR #4672, opened 2026-01-07
Support adding `Authorization` HTTP headers through the `RUSTUP_AUTHO…

amejia1:
…RIZATION_HEADER` environment variable.

This partially addresses https://github.com/rust-lang/rustup/issues/1343 by supporting reverse proxies which require authentication that function as mirrors to https://static.rust-lang.org .


djc:
If this is about a reverse proxy I don't think the complexity of doing this is worth it. Instead, it seems like the main issue is that we might leak credentials in logging when using `RUSTUP_DIST_SERVER`, which can be remedied more directly. We should also not unlink #1343 from this, because that issue seems to be about a forward proxy, not a reverse proxy.

amejia1:
> If this is about a reverse proxy I don't think the complexity of doing this is worth it. Instead, it seems like the main issue is that we might leak credentials in logging when using `RUSTUP_DIST_SERVER`, which can be remedied more directly. We should also not unlink #1343 from this, because that issue seems to be about a forward proxy, not a reverse proxy.

This change would also support token based authentication through the Bearer scheme. I had those in mind when I thought to implement these changes. Here are some examples of what I mean https://api.goauthentik.io/authentication/

amejia1:
I have converted this PR to a draft. I need to investigate ways to start a forward proxy which requires authentication.

djc:
I don't think we're going to want a Ferron proxy setup in our source tree.

amejia1:
I just noticed your last comment. I had this in draft mode and thought that would disable notifications. I am working on a good and simple way to test the changes and thought to use containers. Is use of containers for testing fine? If no, what should I do for tests?

amejia1:
I am done working out how to test and verify the new changes using containers. If usage of containers is not acceptable, another idea I have so far is writing a small test web app much like how there is httpbin to test the Python "request" module.

PR #4675, opened 2026-01-10
Update windows-msvc.md

pino1536:
- shrink this step drastically
- use VS Build Tools instead (no need to install an full IDE)
- get the VS Installer via Windows Package Manager (its an offical way)
- added command line that opens the VS Installer and automatically chooses the right components (if not already installed)

### DIFF
--- a/doc/user-guide/src/installation/windows-msvc.md
+++ b/doc/user-guide/src/installation/windows-msvc.md
@@ -3,72 +3,24 @@
 To compile programs into an exe file, Rust requires a linker, libraries and Windows API import libraries.
 For `msvc` targets these can be acquired through Visual Studio.
 
-## Automatic install
+## Automatically through the installer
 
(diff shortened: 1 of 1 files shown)


ChrisDenton:
This should note remove the note about licensing. We have to be transparent about that and shouldn't attempt to conceal it.

Also, again due to licensing, we cannot recommend build tools without caveats.

pino1536:
> This should note remove the note about licensing. We have to be transparent about that and shouldn't attempt to conceal it.
> 
> Also, again due to licensing, we cannot recommend build tools without caveats.

Fixed it. Now it still uses winget but with the regular gui installation, just like downloading the installer from the website (same .exe file). It shows the lizens agreement as usual. After that the 2 needed components are already selected.

This should fix the grayzone completely and remove any responsibility from rust to the user itself because he has to accept every step just like installing it manual.

pino1536:
another pull later

PR #4676, opened 2026-01-11
Visual Studio prerequisites changes

pino1536:
**rustup auto install script:**
- Update Visual Studio Installer download via winget
_winget is very stable and the download is official maintained by microsoft and used official_
- Remove the extra check if Visual Studio Installer is installed
_This is simple not necessary or brings any negative aspect. The download is only 4mb and it will try update the the VS Installer but don't change any installations. Its a indented and the reason why the VS Installer-Installer itself can take setup arguments._

**documentation**
- shrink down the documentation
_maybe it shouldn't even have an extra site for a step you can do in a single command-line._
- update manual download link and info's in the rustup tool
_Was very confusing, reads like you're installing the entire IDE._
- winget command line to install (but gui install because of license agreement)

**To the license thing of the build tools:**
First of all they are indented to be used outside of an whole VS IDE. Official VSCode advertising it for the C#/C++ programming. Just like "Visual Studio Community", where you don't need an account or anything, the "Build Tools for Visual Studio" is also a Visual Studio Product itself where you "get an free license" by agree to it. The winget download runs the installer in gui mode and the user has to accept the license itself.

### DIFF
--- a/doc/user-guide/src/installation/windows-msvc.md
+++ b/doc/user-guide/src/installation/windows-msvc.md
@@ -14,7 +14,7 @@ If you chose the automatic install this winget command will run in the backgroun
 winget install --id Microsoft.VisualStudio.BuildTools --force --interactive  --custom "--focusedUi --addProductLang En-us --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 --add Microsoft.VisualStudio.Component.Windows11SDK.26100"
 ```
 
-But you could [download the Microsoft Build Tools for Visual Studio][vs download] Installer yourself and do the same steps manual.
+But you could [download the Microsoft Build Tools for Visual Studio][vs download] Installer yourself and do the same steps manually.
 
 The installer will start by linking to the [Build Tools license][vs licences] and will install the "Visual Studio Installer", a tool to manage Visual Studio components.
 
--- a/doc/user-guide/src/installation/windows-msvc.md
+++ b/doc/user-guide/src/installation/windows-msvc.md
@@ -5,70 +5,32 @@ For `msvc` targets these can be acquired through Visual Studio.
 
 ## Automatic install
 
-If you don't have Visual Studio already installed then [rustup-init] will offer to automatically install the prerequisites.
-Doing so means you can skip the rest of this page.
-However, it installs Visual Studio Community edition which may not be appropriate for all users.
-It is free for individuals, academic and open source use, but not for other uses, such as in proprietary enterprise software.
-Users should ask their organisation which edition is right for them.
-See [licensing terms][vs licences] for more details.
+If you don't have Visual Studio already installed then the [rustup-init.exe][rustinstall] installer will offer to automatically install the prerequisites. 
 
 ## Manual install
 
-[Download Visual Studio][vs downloads].
-Rust supports Visual Studio 2017 and later but it is recommended that you use the latest version (currently 2022) for new projects.
-You can opt to download only the Build Tools for Visual Studio, which does not include the IDE.
-However this requires you already have a license to the Community, Professional or Enterprise edition.
+If you chose the automatic install this winget command will run in the background:
+```Batchfile
+winget install --id Microsoft.VisualStudio.BuildTools --force --interactive  --custom "--focusedUi --addProductLang En-us --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 --add Microsoft.VisualStudio.Component.Windows11SDK.26100"
+```
 
-Once you've downloaded and started the installer, the easiest way to get everything installed is to select "Desktop Development with C++".
-This will include the necessary components.
-On the "Language Packs" tab, make sure the English language pack is installed in addition to your preferred language.
+But you could [download the Microsoft Build Tools for Visual Studio][vs download] Installer yourself and do the same steps manual.
 
-If you want more details on the installation process or want to further customize the install then follow the walkthrough below.
-Otherwise complete the Visual Studio install and continue with installing Rust.
-
-## Walkthrough: Installing Visual Studio 2022
-
-This walkthrough uses the Community edition of Visual Studio but the Professional, Enterprise and the Build Tools all work the same way.
-
-The installer will start by linking to the [license][vs licences] and for your edition of Visual Studio and then preparing the installer.
+The installer will start by linking to the [Build Tools license][vs licences] and will install the "Visual Studio Installer", a tool to manage Visual Studio components.
 
 ![Accept the license](images/step1.png)
-![Installing the installer](images/step2.png)
 
-Once this finishes, you can then select the components to be installed.
-Here we use the "Workload" tab to select the "Desktop Development with C++" workload.
-This will includes all needed components for Rust:
-![Select the C++ Workload](images/step3.png)
 
-### Installing only the required components (optional)
+Then you need to install the "Individual components":
+* MSVC Build Tools for x64/x86 (Latest)
+* Windows 11 SDK (10.0.26100.XXXX)
 
-If you'd like a more minimal install (and won't be doing C++ development) then you can use the "Individual Components" tab to select just the essentials, which are:
+And the "Language pack":
+* English
 
-* MSVC v143 - VS 2022 C++ x64/x86 build tools (Latest)
-* Windows 11 SDK (10.0.22621.0)
 
-Note that the specific version of the Windows SDK doesn't matter for pure Rust code but if using C++ as well you'll likely want either the latest or whichever version is required by the C++ project (or both).
+Once finished, you can continue on to installing Rust and the installer should detect MSVC.
 
-![Select the latest MSVC component](images/component-msvc.png)
-![Select the Windows 11 SDK component](images/component-sdk.png)
-
-### Adding Language Packs (optional)
-
-After choosing the components, you may also want to select the language packs to install.
-Switch to the "Language Packs" tab and add the languages.
-It is recommended that you add the English language pack in addition to your preferred language.
-This will provide English language error messages, which may help when reporting errors.
-
-![Add the English language](images/step4.png)
-
-### Completing the install
-
-Finally click the install button and wait for everything to be installed.
-
-![Wait for the install to complete](images/step5.png)
-
-Once finished, you can continue on to installing Rust.
-
-[rustup-init]: https://rustup.rs
-[vs downloads]: https://visualstudio.microsoft.com/downloads/
-[vs licences]: https://visualstudio.microsoft.com/license-terms/
+[vs licences]: https://visualstudio.microsoft.com/license-terms/vs2026-ga-diagnostic-buildtools
+[vs download]: https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2026
+[rustinstall]: https://rust-lang.org/tools/install/
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -260,11 +260,12 @@ pub(crate) async fn try_install_msvc(
     opts: &InstallOpts<'_>,
     cfg: &Cfg<'_>,
 ) -> Result<ContinueInstall> {
-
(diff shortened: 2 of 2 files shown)


rami3l:
@pino1536 Oh, I didn't notice the [license change](https://devblogs.microsoft.com/cppblog/updates-to-visual-studio-build-tools-license-for-c-and-cpp-open-source-projects/) until today. Thanks for pointing it out!

@ChrisDenton this should be a good opportunity for us in this regard I guess? We don't really want the user to accept the VS Community license just to fetch `Microsoft.VisualStudio.BuildTools` which would be way too annoying.

~~This may be as an alternative to https://github.com/rust-lang/rustup/pull/4543 even, given that we no longer have to hardcode anything with this approach.~~ I just realized that the version string is moved elsewhere in this approach, my apologies.

doc/user-guide/src/installation/windows-msvc.md
```diff
@@ -5,70 +5,32 @@ For `msvc` targets these can be acquired through Visual Studio.
...
-However this requires you already have a license to the Community, Professional or Enterprise edition.
+If you chose the automatic install this winget command will run in the background:
+```Batchfile
+winget install --id Microsoft.VisualStudio.BuildTools --force --interactive  --custom "--focusedUi --addProductLang En-us --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 --add Microsoft.VisualStudio.Component.Windows11SDK.26100"
+```
 
-Once you've downloaded and started the installer, the easiest way to get everything installed is to select "Desktop Development with C++".
-This will include the necessary components.
-On the "Language Packs" tab, make sure the English language pack is installed in addition to your preferred language.
+But you could [download the Microsoft Build Tools for Visual Studio][vs download] Installer yourself and do the same steps manual.
```
rami3l:
```suggestion
But you could [download the Microsoft Build Tools for Visual Studio][vs download] Installer yourself and do the same steps manually.
```

doc/user-guide/src/installation/windows-msvc.md
```diff
@@ -5,70 +5,32 @@ For `msvc` targets these can be acquired through Visual Studio.
...
 
 ## Manual install
 
-[Download Visual Studio][vs downloads].
-Rust supports Visual Studio 2017 and later but it is recommended that you use the latest version (currently 2022) for new projects.
-You can opt to download only the Build Tools for Visual Studio, which does not include the IDE.
-However this requires you already have a license to the Community, Professional or Enterprise edition.
+If you chose the automatic install this winget command will run in the background:
+```Batchfile
+winget install --id Microsoft.VisualStudio.BuildTools --force --interactive  --custom "--focusedUi --addProductLang En-us --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 --add Microsoft.VisualStudio.Component.Windows11SDK.26100"
```
rami3l, line 14: Question regarding https://github.com/rust-lang/rustup/pull/4543: will the wrong `--add` params causing the whole command to fail? 

The thing is that Microsoft, as it turned out, can remove any of those components from its server, meaning that these versions might no longer be valid a while later. However, if this is not a fatal error (meaning the GUI launches correctly and the user can tick the missing boxes manually) then this would be less of a problem.

src/cli/self_update/windows.rs
```diff
@@ -199,9 +198,8 @@ pub(crate) fn do_msvc_check(opts: &InstallOpts<'_>, process: &Process) -> Option
         // If the user does not have Visual Studio installed and their host
         // machine is i686 or x86_64 then it's OK to try an auto install.
         // Otherwise a manual install will be required.
-        let has_any_vs = windows_registry::find_vs_version().is_ok();
         let is_x86 = host_triple.contains("i686") || host_triple.contains("x86_64");
-        if is_x86 && !has_any_vs {
```
rami3l, line 204: What will happen with your new logic when the user already has VS installed?

src/cli/self_update/windows.rs
```diff
@@ -262,47 +260,31 @@ pub(crate) async fn try_install_msvc(
...
-            "Microsoft.VisualStudio.Component.Windows11SDK.26100",
-        ]);
+        vsi_args.push_str(" --add Microsoft.VisualStudio.Component.Windows11SDK.26100");
     }
-    info!("running the Visual Studio install");
-    info!("rustup will continue once Visual Studio installation is complete\n");
+
+    // Run the installer. Arguments are documented at:
+    // https://docs.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio
+    let mut cmd = Command::new("winget");
```
rami3l, line 277: Is `winget` a component that comes with all supported Windows installations? Please note that we are currently [stating](https://doc.rust-lang.org/nightly/rustc/platform-support/windows-msvc.html) support for Windows 10/2026 or later.

If this is not the case, maybe we should notify the user and provide an installation hint?

ChrisDenton:
> @pino1536 Oh, I didn't notice the [license change](https://devblogs.microsoft.com/cppblog/updates-to-visual-studio-build-tools-license-for-c-and-cpp-open-source-projects/) until today. Thanks for pointing it out!
>
> @ChrisDenton this should be a good opportunity for us in this regard I guess? We don't really want the user to accept the VS Community license just to fetch Microsoft.VisualStudio.BuildTools which would be way too annoying.

The license change does not affect us because we use link.exe to directly develop software. Basically you can substitute "Rust" for "C++" in that blog post. This was investigated at the time and we got an [official response from Microsoft in 2023](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/Visual.20C.2B.2B.20Build.20Tools/near/401860589).

doc/user-guide/src/installation/windows-msvc.md
```diff
@@ -5,70 +5,32 @@ For `msvc` targets these can be acquired through Visual Studio.
...
 
-Once you've downloaded and started the installer, the easiest way to get everything installed is to select "Desktop Development with C++".
-This will include the necessary components.
-On the "Language Packs" tab, make sure the English language pack is installed in addition to your preferred language.
+But you could [download the Microsoft Build Tools for Visual Studio][vs download] Installer yourself and do the same steps manual.
 
-If you want more details on the installation process or want to further customize the install then follow the walkthrough below.
-Otherwise complete the Visual Studio install and continue with installing Rust.
-
-## Walkthrough: Installing Visual Studio 2022
```
ChrisDenton, line 29: I do not think we should remove the GUI instructions, especially as it's the main user interface for the Visual Studio installer. If it's confusing we should improve that. If other installation methods are better then we should add those.

doc/user-guide/src/installation/windows-msvc.md
ChrisDenton: It will appear to succeed but the component will be missing from the UI. There's nothing to indicate to the user that there is a missing component.

Image

src/cli/self_update/windows.rs
pino1536: Tested nearly every case:
- A installed Visual Studio Installer without any Visual Studio Installation
- A installed Visual Studio Installer with Visual Studio Community installed
- A installed Visual Studio Installer with Build Tools but only one component or none of the needed

In any case the 4 mb "Visual Studio Installer"-Installer will try to update or check the already installed "Visual Studio Installer". There is no case where you need an older "VS Installer". This is also intended otherwise the "Visual Studio Installer"-Installer would allow the same arguments as the "installed Installer"

In case of an installation of "Visual Studio Community/..." but no build tools the "Visual Studio Build Tools" (only 90-100 mb) will be downloaded beside "Visual Studio Community/...". But in most cases if you use another Compiler for other languages you already have the Build tools.

doc/user-guide/src/installation/windows-msvc.md
pino1536: but this was also a problem before (the vs installer is the same) and it only affects Windows 11 sdk. the "Microsoft.VisualStudio.Component.VC.Tools.x86.x64" is always the lasted version.
The best way would be to simple update the "Windows 11 SDK" it in the tool periodical.

src/cli/self_update/windows.rs
pino1536: yes it comes with every Windows 10 and 11 pre installed. In many cases Microsoft uses it itself for many things. (you can see it because many auto installed stuff comes with the install source winget) and you can find everything official through winget.

doc/user-guide/src/installation/windows-msvc.md
pino1536: But to be honest if someone cant use the installer and find the, in text form described, components he is on the level of my grandma using a computer 😬

I even would remove the license part... somehow there isn't even a way around other then to accept them.

And somehow I wanted to clarify that it would be better to use the new auto installer because unlike other examples it does the exact same (but guided) as you would do manual.

For me as a complete newcomer to rust (one week), the documentation only to get rustup ready, splatted over 3 documentations was a bit too much for these little requirements.

pino1536:
As I mentioned above, I tested nearly every case:

- No Visual Studio Installer installed
(= Installs the Visual Studio Installer and then Visual Studio Build Tools)
- A installed Visual Studio Installer without any Visual Studio Installation
(= Installs the Visual Studio Build Tools)
- A installed Visual Studio Installer with Build Tools but none component
(= it will install the 2 components.)
- A installed Visual Studio Installer with Build Tools but only one component
(= it will install the missing component.)
- A installed Visual Studio Installer with Visual Studio Community installed
(= Installs the Visual Studio Build Tools beside the Visual Studio Community.)

For the last Test: In most cases the Visual Studio Build Tools are installed instead of the full IDE if you use other dev environments like Flutter etc. Somehow it even better in any case to install the Components as a part of the "Visual Studio Build Tools" instead to add it to "Visual Studio Community".

src/cli/self_update/windows.rs
ChrisDenton: Not with every SKU. E.g. Windows Sandbox does not have it installed by default. I'm not entirely against using winget here but I also don't think it get us anything useful. The aka.ms links always work.

doc/user-guide/src/installation/windows-msvc.md
ChrisDenton: Yes this is a problem currently which is why we're looking into mitigations.

doc/user-guide/src/installation/windows-msvc.md
ChrisDenton: You don't need to read it all. We can add a command line section before the full GUI instructions. The reason it was added in the first place is because people do indeed struggle with it.

ChrisDenton (changes requested):
I've stated this a few times but I do want to be absolutely clear because I consider this a blocker:

For licensing reasons we cannot recommend the build tools nor install them by default unless the user already has Visual Studio installed. We could suggest it as an alternative, so long as we're clear on the licensing requirements. Or we could perhaps allow users to select their Visual Studio edition.

doc/user-guide/src/installation/windows-msvc.md
pino1536: > Yes this is a problem currently which is why we're looking into mitigations.

There are another 2 options:

1. We could get the SDK via winget (its not a part of the Visual Studio licens):
```
winget install --id Microsoft.WindowsSDK.10.0.26100
```

2. We could export a json file with all available components to parse the last version:

```Batch
Microsoft Visual Studio\Installer\setup.exe export --productId Microsoft.VisualStudio.Product.BuildTools --channelId VisualStudio.18.Release --config "%UserProfile%\Desktop\temp.vsconfig" --all --quiet```

ChrisDenton: If we use `winget` then doing a search for the latest would be best (we can use the version field to filter it).

Using the json from the VS installer would be better in a sense because then we can be sure the it will show up labelled as installed when someone is attempting to debug issues. But I do worry that the json may not be stable between Visual Studio versions. That could be mitigated by ensuring we use a defined version.

PR #4678, opened 2026-01-14
fix(utils): "Directory not empty" error when uninstalling multiple toolchains

### LINKED ISSUE #4657: Repeated "could not remove 'install' directory" ... "Directory not empty" errors when uninstalling toolchains
### Verification

### Problem

When uninstalling toolchains en masse using `rustup toolchain list | rg ^nightly-20 | xargs rustup toolchain uninstall` (à la #2530), I keep getting errors about rustup not being able to remove directories due to them not being empty. This only happens after at least one has been successfully removed, sometimes several. E.g.:

```console
❯ rustup toolchain list | rg ^nightly-20 | xargs rustup toolchain uninstall
info: uninstalling toolchain 'nightly-2025-05-12-aarch64-apple-darwin'
info: toolchain 'nightly-2025-05-12-aarch64-apple-darwin' uninstalled
info: uninstalling toolchain 'nightly-2025-05-21-aarch64-apple-darwin'
info: toolchain 'nightly-2025-05-21-aarch64-apple-darwin' uninstalled
info: uninstalling toolchain 'nightly-2025-06-24-aarch64-apple-darwin'
info: toolchain 'nightly-2025-06-24-aarch64-apple-darwin' uninstalled
info: uninstalling toolchain 'nightly-2025-07-04-aarch64-apple-darwin'
error: could not remove 'install' directory: '/Users/noahlev/.rustup/toolchains/nightly-2025-07-04-aarch64-apple-darwin': Directory not empty (os error 66)
```

Note that if I run the command again, then it successfully finishes uninstalling that toolchain and usually a few more, before the error then happens again with a different toolchain.

### Steps

1. Install several nightly toolchains.
2. Run `rustup toolchain list | rg ^nightly-20 | xargs rustup toolchain uninstall`.

### Possible Solution(s)

It seems like some kind of race condition when removing files, but it's strange because `xargs` allows each command to complete before running the next. I wonder if this could be an issue with macOS metadata files like `.DS_Store`, but I'm not actually sure -- just a guess.

### Notes

_No response_

### Rustup version

```console
❯ rustup --version
info: The currently active `rustc` version is `rustc 1.90.0 (1159e78c4 2025-09-14)`
```

### Installed toolchains

```console
Not really relevant because they're in flux as I uninstall, but here it is anyway:

❯ rustup show
rustup home:  /Users/noahlev/.rustup

installed toolchains
--------------------
stable-aarch64-apple-darwin (active, default)
beta-aarch64-apple-darwin
nightly-aarch64-apple-darwin
nightly-2025-07-04-aarch64-apple-darwin
1.50.0-aarch64-apple-darwin
1.65.0-aarch64-apple-darwin
1.79.0-aarch64-apple-darwin
bisector-ci-a77cc64af491a31db224109a76b9b81cd26cd07c-aarch64-apple-darwin
bisector-ci-cfa4ac66c194046f631ce076c75516ecfdeb77ee-aarch64-apple-darwin
bisector-ci-f1ce0e6a00593493a12e0e3662119786c761f375-aarch64-apple-darwin
bisector-nightly-2021-01-06-aarch64-apple-darwin
bisector-nightly-2022-01-06-aarch64-apple-darwin
bisector-nightly-2022-01-07-aarch64-apple-darwin
miri
r1stage0
r1stage1
r1stage2
r2stage0
r2stage1
r2stage2

active toolchain
----------------
name: stable-aarch64-apple-darwin
installed targets:
  aarch64-apple-darwin
  wasm32-unknown-unknown
  x86_64-apple-darwin
  x86_64-unknown-linux-gnu
```

### OS version

```console
macOS Tahoe - Apple Silicon
```

cachebag:
Fixes #4657 - "Directory not empty" errors when uninstalling multiple toolchains sequentially.

The `remove_dir_all` crate with parallel feature can leave directories in an inconsistent state, causing intermittent failures:
```
error: could not remove 'install' directory: Directory not empty (os error 66)
```
This PR adds retry logic with exponential backoff to `remove_dir()`:
- Retries on `DirectoryNotEmpty`, `PermissionDenied`, and `EBUSY` 
- Fibonacci backoff (10ms base) with jitter, up to 10 attempts
- Syncs parent directory after removal

### DIFF
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -7,6 +7,9 @@ use std::io::Write;
 use std::path::Path;
 use std::str;
 
+use retry::delay::{Fibonacci, jitter};
+use retry::{OperationResult, retry};
+
 #[cfg(not(windows))]
 use crate::process::Process;
 
@@ -229,6 +232,15 @@ fn symlink_junction_inner(target: &Path, junction: &Path) -> io::Result<()> {
     }
 }
 
+fn is_retryable_dir_error(e: &io::Error) -> bool {
+    matches!(
+        e.kind(),
+        io::ErrorKind::DirectoryNotEmpty
+            | io::ErrorKind::PermissionDenied
+            | io::ErrorKind::ResourceBusy
+    )
+}
+
 pub fn remove_dir(path: &Path) -> io::Result<()> {
     if fs::symlink_metadata(path)?.file_type().is_symlink() {
         #[cfg(windows)]
@@ -240,11 +252,28 @@ pub fn remove_dir(path: &Path) -> io::Result<()> {
             fs::remove_file(path)
         }
     } else {
-        // Again because remove_dir all doesn't delete write-only files on windows,
-        // this is a custom implementation, more-or-less copied from cargo.
-        // cc rust-lang/rust#31944
-        // cc https://github.com/rust-lang/cargo/blob/master/tests/support/paths.rs#L52
-        remove_dir_all::remove_dir_all(path)
+        let result = retry(Fibonacci::from_millis(10).map(jitter).take(10), || {
+            // Again because remove_dir all doesn't delete write-only files on windows,
+            // this is a custom implementation, more-or-less copied from cargo.
+            // cc rust-lang/rust#31944
+            // cc https://github.com/rust-lang/cargo/blob/master/tests/support/paths.rs#L52
+            match remove_dir_all::remove_dir_all(path) {
+                Ok(()) => OperationResult::Ok(()),
+                Err(e) if is_retryable_dir_error(&e) => OperationResult::Retry(e),
+                Err(e) => OperationResult::Err(e),
+            }
+        });
+
+        // Best-effort sync of parent directory to ensure filesystem metadata consistency
+        // after parallel deletion. Helps prevent "Directory not empty" errors on subsequent
+        // operations. See https://github.com/rust-lang/rustup/issues/4657
+        if result.is_ok()
+            && let Some(parent) = path.parent()
+        {
+            let _ = open_dir_following_links(parent).and_then(|f| f.sync_all());
+        }
+
+        result.map_err(|e| e.error)
     }
 }
 
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -1817,6 +1817,53 @@ no active toolchain
 }
 
 #[tokio::test]
+async fn uninstall_multiple_toolchains_sequentially() {
+    let mut cx = CliTestContext::new(Scenario::None).await;
+
+    {
+        let cx = cx.with_dist_dir(Scenario::SimpleV2);
+        cx.config
+            .expect(["rustup", "toolchain", "install", "stable"])
+            .await
+            .is_ok();
+        cx.config
+            .expect(["rustup", "toolchain", "install", "beta"])
+            .await
+            .is_ok();
+        cx.config
+            .expect(["rustup", "toolchain", "install", "nightly"])
+            .await
+            .is_ok();
+    }
+
+    cx.config
+        .expect(["rustup", "default", "none"])
+        .await
+        .is_ok();
+
+    cx.config
+        .expect([
+            "rustup",
+            "toolchain",
+            "uninstall",
+            "stable",
+            "beta",
+            "nightly",
+        ])
+        .await
+        .is_ok();
+
+    cx.config
+        .expect(["rustup", "toolchain", "list"])
+        .await
+        .with_stdout(snapbox::str![[r#"
+no installed toolchains
+
+"#]])
+        .is_ok();
+}
+
+#[tokio::test]
 async fn proxy_toolchain_shorthand() {
     let cx = CliTestContext::new(Scenario::SimpleV2).await;
     cx.config


rami3l:
@cachebag Many thanks for the investigation! I'll review shortly.

In the meantime I'm wondering if this should be reported upstream as well 🤔

cachebag:
> In the meantime I'm wondering if this should be reported upstream as well 🤔

It's probably worth doing so, though I should note it took me a long time re-produce the bug on my mac.

djc:
> > In the meantime I'm wondering if this should be reported upstream as well 🤔
> 
> It's probably worth doing so, though I should note it took me a long time re-produce the bug on my mac.

I'm honestly not sure if this is the kind of thing libraries should default to -- it could make sense for the logic to live upstream but it seems like it should be opt-in behavior.

src/utils/raw.rs
```diff
@@ -240,11 +259,21 @@ pub fn remove_dir(path: &Path) -> io::Result<()> {
             fs::remove_file(path)
         }
     } else {
-        // Again because remove_dir all doesn't delete write-only files on windows,
-        // this is a custom implementation, more-or-less copied from cargo.
-        // cc rust-lang/rust#31944
-        // cc https://github.com/rust-lang/cargo/blob/master/tests/support/paths.rs#L52
-        remove_dir_all::remove_dir_all(path)
+        let result = retry(Fibonacci::from_millis(10).map(jitter).take(10), || {
+            match remove_dir_all::remove_dir_all(path) {
```
rami3l, line 260: I think it's still worth it to keep the original comments here :)

src/utils/raw.rs
```diff
@@ -240,11 +259,21 @@ pub fn remove_dir(path: &Path) -> io::Result<()> {
...
+                Ok(()) => OperationResult::Ok(()),
+                Err(e) if is_retryable_dir_error(&e) => OperationResult::Retry(e),
+                Err(e) => OperationResult::Err(e),
+            }
+        });
+
+        if result.is_ok()
+            && let Some(parent) = path.parent()
+        {
+            let _ = open_dir_following_links(parent).and_then(|f| f.sync_all());
```
rami3l, line 273: It'd be better to add a short description (maybe including the issue link) so it's more obvious what this step is about :)

tests/suite/cli_rustup.rs
```diff
@@ -1816,6 +1816,54 @@ no active toolchain
...
+    cx.config
+        .expect(["rustup", "toolchain", "uninstall", "stable"])
+        .await
+        .is_ok();
+    cx.config
+        .expect(["rustup", "toolchain", "uninstall", "beta"])
+        .await
+        .is_ok();
+    cx.config
+        .expect(["rustup", "toolchain", "uninstall", "nightly"])
```
rami3l: You should really pass multiple toolchains here to match the original issue description.

src/utils/raw.rs
```diff
@@ -229,6 +232,22 @@ fn symlink_junction_inner(target: &Path, junction: &Path) -> io::Result<()> {
...
+        e.kind(),
+        io::ErrorKind::DirectoryNotEmpty | io::ErrorKind::PermissionDenied
+    ) || {
+        #[cfg(unix)]
+        {
+            e.raw_os_error() == Some(libc::EBUSY)
+        }
+        #[cfg(not(unix))]
+        {
+            false
```
rami3l: Question: Why is this part Unix-specific? I see the same error code can be produced by Windows for example (https://github.com/lovell/sharp/issues/2054) 🤔

ChrisDenton: This could simply be `io::ErrorKind::ResourceBusy`

src/utils/raw.rs
rami3l: @cachebag The comments will still need to match the `remove_dir_all()` call, so move them down by a line would be better :)

src/utils/raw.rs
rami3l: @cachebag I think the solution proposed by @ChrisDenton already makes the conditional compilation part redundant, so the `#[cfg]` dance can be removed.

rami3l (approved):
Many thanks again for tackling this!

rbtcollins:
@rami3l @djc  I'm torn here, the remove_dir_all crates reason for existence is reliable deleting. OTOH the comment about reproducing this on Mac makes me think we don't fully understand whats going on here - Mac doesn't have the same file system deferred delete semantics Windows does, and remove_dir_all uses rayon for scatter-gather IO in parallel mode: there should be no room for such a race to impact Mac.

rami3l:
> It's probably worth doing so, though I should note it took me a long time re-produce the bug on my mac.

@cachebag Would you mind sharing more details? I mean I was never able to reproduce it locally not I guess it won't be super wrong to be defensive about this one in our code so...

cachebag:
Replying here for cleanliness.
@rbtcollins 

> Alternatively, if you can create something to trigger the bug when run in a loop I could make a Mac VM in AWS and look at it myself.

Working on this now, but I don't have my notes from when I initially drafted this PR anymore. I mainly followed @camelid's steps [here](https://github.com/rust-lang/rustup/issues/4657#issue-3752327706). 

> and remove_dir_all uses rayon for scatter-gather IO in parallel mode: there should be no room for such a race to impact Mac.

Unless I'm reading this incorrectly, even though rustup enables the parallel feature, on macOS `default_parallel_mode()` unconditionally returns Serial. Rayon is never involved on macOS at all?

https://github.com/XAMPPRocky/remove_dir_all/blob/b09c40ea38a9f528df3f742a4f7c68c46967ddac/src/_impl.rs#L28-L34

(For ref):
```rust
pub(crate) fn default_parallel_mode() -> ParallelMode {
    #[cfg(all(feature = "parallel", not(target_os = "macos")))]
    return ParallelMode::Parallel;
    #[cfg(any(not(feature = "parallel"), target_os = "macos"))]
    return ParallelMode::Serial;
}
```

[This bug](https://gregoryszorc.com/blog/2018/10/29/global-kernel-locks-in-apfs/) was referenced in `lib.rs`

But regardless, I think this was misguided on my part anyways, due to the fact that It looks like the parallel logic, where it IS used (Linux/Windows), is actually race-free by construction. `rmdir_at` is only called after `remove_dir_contents_recursive` fully returns, and `try_for_each` on `par_bridge` blocks until all rayon tasks finish. So the parent is never `rmdir`'d while children are still being deleted. 

https://github.com/XAMPPRocky/remove_dir_all/blob/b09c40ea38a9f528df3f742a4f7c68c46967ddac/src/_impl.rs#L117-L126
(For ref):
```rust
        #[cfg(feature = "parallel")]
        _ => {
            let iter = fs_at::read_dir(&mut d)?;
            let iter = iter.par_bridge();
            iter.try_for_each(|dir_entry| {
                scan_and_remove_entry_recursively::<I>(debug_root, &dirfd, dir_entry, parallel)
            })?;
        }
    };
```
https://github.com/XAMPPRocky/remove_dir_all/blob/b09c40ea38a9f528df3f742a4f7c68c46967ddac/src/_impl.rs#L189-L199
(And for each directory entry):
```rust
                if is_dir {
                    remove_dir_contents_recursive::<I>(child_file, &dir_debug_root, parallel)?;
                    #[cfg(feature = "log")]
                    log::trace!("rmdir: {}", &dir_debug_root);
                    let opts = fs_at::OpenOptions::default();
                    opts.rmdir_at(dirfd, name).map_err(|e| {
                        #[cfg(feature = "log")]
                        log::debug!("error removing {}", dir_debug_root);
                        e
                    })?;
                }
```

This leads me to believe that if `remove_dir_all` is serial on macOS and the internal parallel logic is race-free everywhere else, then the `ENOTEMPTY` must be coming from somewhere above the crate (maybe rustup issuing concurrent `remove_dir_all` calls for multiple toolchains from async tasks when you pass stable beta nightly to toolchain uninstall all at once?) cc @rami3l @djc.

That also would explain why it's hard to reproduce and why this patch seemingly works; because it buys time for the concurrent sibling deletion to finish.

Curious to know anyone's thoughts if I'm misreading this. I'll let you know if I'm able to reproduce the original issue again.

rami3l:
@cachebag Thanks for the detailed investigation!

> maybe rustup issuing concurrent remove_dir_all calls for multiple toolchains from async tasks when you pass stable beta nightly to toolchain uninstall all at once?

I don't remember anything like that; AFAIK uninstallation is still strictly serial.

PR #4680, opened 2026-01-15
fix(cli/rustup-mode): `check` for self updates for `SelfUpdateMode::CheckOnly`

rami3l:
This PR addresses the following concern of @steffahn:

> with a configuration of rustup set auto-self-update check-only, arguably I would prefer to still be able see the result of the self-update check.
_https://internals.rust-lang.org/t/seeking-beta-testers-for-rustup-v1-29-0/23814/3_

It is quite likely that this section was originally copied over from `rustup update` without careful verification/review. My apologies.


ChrisDenton (approved):
That does make sense to me.

PR #4682, opened 2026-01-16
Remove the mixed singular/plural phrasing as “`component(s)`”

steffahn:
Instead, we can use "components" or "component" depending on the number (which is always known).
In the latter case also add the name of the component for more consistent messaging style with other `info!` outputs about single components.

Screenshot comparing the output for 2 scenarios in action (top half is *before* this PR):

Bildschirmfoto_20260116_030854

---

Also I feel like it could be a reasonable idea [which this PR currently does *not* do (yet?)] to also add the number of components in the middle, i.e. `downloading components` would further become `downloading 6 components` then.

---

I do expect some CI failure as `component(s)` appears in some ui test – but I haven’t yet figured out how to run tests locally correctly, and it’s getting too late today.

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -217,7 +217,15 @@ impl Manifestation {
         }
 
         if !components.is_empty() {
-            info!("downloading component(s)");
+            if components.len() > 2 {
+                info!("downloading {} components", components.len());
+            } else {
+                info!(
+                    "downloading component {}",
+                    components[0].manifest.short_name(&components[0].component),
+                );
+            };
+
             let mut stream = InstallEvents::new(components.into_iter(), Arc::new(self));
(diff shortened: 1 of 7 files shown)


djc (approved):
Thanks!

djc:
> Also I feel like it could be a reasonable idea [which this PR currently does _not_ do (yet?)] to also add the number of components in the middle, i.e. `downloading components` would further become `downloading 6 components` then.

Sounds good to me.

rami3l (approved):
Many thanks!

Maybe let's just go with your later proposal instead? We'll have to update the snapshots anyway so...

rami3l:
@steffahn Thanks for the update! Just squashing your changes into one commit and if the CI is all green we are good to go :)

PR #4694, opened 2026-01-28
fix(cli/rustup-mode): improve exit code of `rustup check`

### LINKED ISSUE #4681: New exit status for `rustup check` can feel confusing to interactive users
(Originally posted as https://internals.rust-lang.org/t/seeking-beta-testers-for-rustup-v1-29-0/23814/3)

Testing the new beta, I found the exit code of `rustup check` a bit confusing.

The previous behavior is such that if everything is up to date, all output is green, and if it isn’t there are yellow “Update available” lines.

So much is unchanged… but now we also get an exit code indicating the overall result…

…but that exit code is just feeling *very* the wrong way around:

* all is green "Up to date" :white_check_mark:
  * overall exit code: 1 **error** :x:
* some yellow "Update available" :warning: [possibly also some green "Up to date" :white_check_mark:]
  * overall exit code: 0 **success** :white_check_mark:

cc @tjkirch @rami3l

---

The change was introduced in #4340, where the main motivation appears to be usage in some custom minimal interface on top of rustup; using this exit code as a way to check if something can be updated before then prompting the user with the question of if they want to update.

For this specific use-case, it seems plausible that success is "updates are available", and anything else ("all up to date" or "some error occurred") is kind-of the same thing, since it means, rustup can’t do any updates right now (either because everything is up to date, or because some other erroneous condition is present – *like what acutally :thinking:? no network, for example, perhaps?*)

OTOH, `rustup check` was [first](https://github.com/rust-lang/rustup/pull/1980) introduced rather for interactive use, so I’d like that not to become any more confusing… as someone who has their terminal promt set up in such a way that nonzero exit codes do display as red :X: marks, I find it a bit weird this way around, as explained in the previous section.

---

I have not yet found good prior art for any comparable commands of similar tools. One thing I have noticed though is that a 3-way distinction of exit codes might be worth considering in any case, so that users can – if they want to – cleanly differentiate two "successful" cases of "`rustup check` has determined whether updates are available" from the case of "any other errors happened". This way, a script that only wants to act on the "updates are available" case, can still pick out that one, even we choose the exit code 0 to be for "everything is up to date". 

*Now I’m also noticing that rustup unconventionally uses exit code `1` instead of `2` for cases of incorrect command line arguments.*

Anyway; one point of comparison for a 3-way distinction that i could find is `ripgrep`, which appears to produce `0` if matches are found, `1` if no matches are found, and `2` for any other error (incorrect command line arguments or wrong regex syntax for sure; but I also tried e.g. having a file without read permissions).

---

Another idea, if "updates are available" stays the success case, could be to change the color coding. For example, "`up to date`" could become bold(?) white text, and "`update available`" bold(?) green.

---

Yet another (probably too “creative” though(?)) idea could be to leave the exit value always at 0 once again, but instead split the output between stderr (for the `up to date` kind of messages) and stdout (for the `update available` kind of messages), so a script could choose to inspect only stdout if it wants to know if and how many updates are available (by counting the lines of stdout).

I do acknowledge that this is slightly further complicated by the new parallelized checking & output, though I would have ideas on how to make that still properly anyway.

rami3l:
Closes #4681.

Following the feedback on #4340 and particularly the previous discussion in https://github.com/rust-lang/rustup/issues/4681#issuecomment-3813523464, this change follows the convention of DNF where `0` is returned when no updates and `100` when updates are available.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -927,8 +927,7 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<ExitCode> {
         update_available = true;
     }
 
-    let exit_status = if update_available { 0 } else { 1 };
-    Ok(ExitCode(exit_status))
+    Ok(ExitCode(if update_available { 100 } else { 0 }))
 }
 
 async fn update(
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -183,13 +183,18 @@ impl Assert {
 
     /// Asserts that the command exited with an ok status.
     pub fn is_ok(&self) -> &Self {
-        assert!(self.output.ok);
-        self
+        self.has_code(0)
     }
 
     /// Asserts that the command exited with an error.
     pub fn is_err(&self) -> &Self {
-        assert!(!self.output.ok);
+        assert_ne!(self.output.status, Some(0));
+        self
+    }
+
+    /// Asserts that the command exited with the specific code.
+    pub fn has_code(&self, code: i32) -> &Self {
+        assert_eq!(self.output.status, Some(code));
         self
     }
 
@@ -350,7 +355,7 @@ impl Config {
 
     pub async fn expect_ok_contains(&self, args: &[&str], stdout: &str, stderr: &str) {
         let out = self.run(args[0], &args[1..], &[]).await;
-        if !out.ok || !out.stdout.contains(stdout) || !out.stderr.contains(stderr) {
+        if out.status != Some(0) || !out.stdout.contains(stdout) || !out.stderr.contains(stderr) {
             print_command(args, &out);
             println!("expected.ok: true");
             print_indented("expected.stdout.contains", stdout);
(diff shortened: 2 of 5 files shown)


appleGun22:
How one is supposed to know which error code is returned when and what it corresponds to? The error code type is just a wrapped around an integer and it is not documented on the type itself either.

rami3l:
@appleGun22 Please feel free to make a PR adding documentation to an appropriate place once this decision has been settled, preferably in the `--help` section of the `rustup check` command.

src/cli/rustup_mode.rs
```diff
@@ -927,8 +927,7 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<ExitCode> {
         update_available = true;
     }
 
-    let exit_status = if update_available { 0 } else { 1 };
-    Ok(ExitCode(exit_status))
+    Ok(ExitCode(if update_available { 100 } else { 0 }))
```
ChrisDenton, line 930: I do think we should start using constants for exit codes (e.g. `EXIT_SUCCESS`, `EXIT_UPDATE_AVAILABLE`, etc). If not now then soon. Especially if we want to make any more commands return special exit codes.

ChrisDenton (approved):
Hm, it doesn't look like there's a strong consensus from package managers around this but I'm happy to do something that seems reasonable and has some precedence.

I admit I'm slightly worried that this could break existing scripts though on balance I think this is worth doing. As noted on the issue, `rustup check` was intended for interactive use and isn't particularly amenable to scripting (hence the motivation for the change).

I do have one nit but I'm not inclined to block on it.

rami3l:
@ChrisDenton Thanks! The exit code differences are a new behavior in this release already so the breakages should be controlled.

src/cli/rustup_mode.rs
rami3l: @ChrisDenton Nice catch! I'd say that might belong to a different patch though.

@FranciscoTGouveia would you be interested in trying this one?

ChrisDenton:
For sure. The behaviour change from current stable to current beta is that if you were relying on it not failing (e.g. in a `set -e -o pipefail` script) then you'll get an immediate error when using beta. With this PR it won't fail immediately but it will at some point in the future once there is another update available. I doubt it will have a big impact but it is something I wanted to at least acknowledge.

djc:
Sorry for being slow to review -- LGTM!

PR #4695, opened 2026-01-31
Preserve partial downloads on network failure

### LINKED ISSUE #4448: Resume partial download on network failure
Download failures always require downloading files from scratch,
which is simply unacceptable in poor network environments

Is it possible to avoid repeatedly downloading data that has already been partially downloaded

```
info: syncing channel updates for 'nightly-aarch64-unknown-linux-gnu'
Total:   1.0 MiB Speed:  81.4 KiB/s Elapsed:  1m 13s
info: latest update on 2025-08-17, rust version 1.91.0-nightly (2e2642e64 2025-08-16)
230.5 KiB /  59.1 MiB (  0 %)  29.6 KiB/s in 57s ETA: 33m 57s
info: retrying download for 'https://rsproxy.cn/dist/2025-08-17/rustc-nightly-aarch64-unknown-linux-gnu.tar.xz'
  9.3 MiB /  59.1 MiB ( 16 %)  28.8 KiB/s in 12m 17s ETA: 29m 32s
info: retrying download for 'https://rsproxy.cn/dist/2025-08-17/rustc-nightly-aarch64-unknown-linux-gnu.tar.xz'
 58.4 MiB /  59.1 MiB ( 99 %)  49.6 KiB/s in 26m 53s ETA: 15s
info: retrying download for 'https://rsproxy.cn/dist/2025-08-17/rustc-nightly-aarch64-unknown-linux-gnu.tar.xz'
 58.5 MiB /  59.1 MiB ( 99 %)  81.6 KiB/s in 26m 14s ETA:  8s
info: retrying download for 'https://rsproxy.cn/dist/2025-08-17/rustc-nightly-aarch64-unknown-linux-gnu.tar.xz'
error: component download failed for rustc-aarch64-unknown-linux-gnu: could not download file from 'https://rsproxy.cn/dist/2025-08-17/rustc-nightly-aarch64-unknown-linux-gnu.tar.xz' to '/home/lrne/.rustup/downloads/3ad6539cfc893725b69f59bb2e72a4ff71c1719186b3ea48eb3c9da56c2a8405.partial': error decoding response body: request or response body error: error reading a body from connection: connection reset
info: syncing channel updates for 'nightly-aarch64-unknown-linux-gnu'
Total:   1.0 MiB Speed:  76.6 KiB/s Elapsed: 14s
info: latest update on 2025-08-17, rust version 1.91.0-nightly (2e2642e64 2025-08-16)
 10.2 MiB /  59.1 MiB ( 17 %)  75.2 KiB/s in  3m 57s ETA: 11m  6s
```

FranciscoTGouveia:
Closes #4448.

If `resume_from_partial` is set to true, it short-circuits, ensuring the `remove_file(path)` is not called.

### DIFF
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -76,6 +76,14 @@ pub(crate) async fn download_file_with_resume(
     }
 }
 
+pub(crate) fn is_network_failure(err: &anyhow::Error) -> bool {
+    match err.downcast_ref::<DownloadError>() {
+        #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
+        Some(DownloadError::Reqwest(e)) => e.is_timeout() || e.is_connect(),
+        _ => false,
+    }
+}
+
 async fn download_file_(
     url: &Url,
     path: &Path,
@@ -262,16 +270,11 @@ impl Backend {
             return Ok(());
         };
 
-        let is_network_failure = match err.downcast_ref::<DownloadError>() {
-            Some(DownloadError::Reqwest(e)) => e.is_timeout() || e.is_connect(),
-            _ => false,
-        };
-
         // TODO: Currently, we only refrain from removing the cached download
         // if there was a network failure from the client side.
         // It may be worth looking for other cases where removal is also not desired.
         Err(
-            if !(resume_from_partial && is_network_failure)
+            if !(resume_from_partial && is_network_failure(&err))
                 && let Err(file_err) = remove_file(path).context("cleaning up cached downloads")
             {
                 file_err.context(err)
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -44,18 +44,19 @@ pub(crate) async fn download_file_with_resume(
     status: Option<&DownloadStatus>,
     process: &Process,
 ) -> anyhow::Result<()> {
-    use crate::download::DownloadError as DEK;
     match download_file_(url, path, hasher, resume_from_partial, status, process).await {
         Ok(_) => Ok(()),
         Err(e) => {
             if e.downcast_ref::<std::io::Error>().is_some() {
                 return Err(e);
             }
-            let is_client_error = match e.downcast_ref::<DEK>() {
+            let is_client_error = match e.downcast_ref::<DownloadError>() {
                 // Specifically treat the bad partial range error as not our
                 // fault in case it was something odd which happened.
-                Some(DEK::HttpStatus(416)) => false,
-                Some(DEK::HttpStatus(400..=499)) | Some(DEK::FileNotFound) => true,
+                Some(DownloadError::HttpStatus(416)) => false,
+                Some(DownloadError::HttpStatus(400..=499)) | Some(DownloadError::FileNotFound) => {
+                    true
+                }
                 _ => false,
             };
             Err(e).with_context(|| {
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -262,9 +262,18 @@ impl Backend {
             return Ok(());
         };
 
-        // TODO: We currently clear up the cached download on any error, should we restrict it to a subset?
+        let is_network_failure = match err.downcast_ref::<DownloadError>() {
+            Some(DownloadError::Reqwest(e)) => e.is_timeout() || e.is_connect(),
+            _ => false,
+        };
+
+        // TODO: Currently, we only refrain from removing the cached download
+        // if there was a network failure from the client side.
+        // It may be worth looking for other cases where removal is also not desired.
         Err(
-            if let Err(file_err) = remove_file(path).context("cleaning up cached downloads") {
+            if !(resume_from_partial && is_network_failure)
+                && let Err(file_err) = remove_file(path).context("cleaning up cached downloads")
+            {
                 file_err.context(err)
             } else {
                 err
@@ -587,7 +596,7 @@ mod reqwest_be {
 
         let mut stream = res.bytes_stream();
         while let Some(item) = stream.next().await {
-            let bytes = item?;
+            let bytes = item.map_err(DownloadError::Reqwest)?;
             callback(Event::DownloadDataReceived(&bytes))?;
         }
         Ok(())
--- a/src/download/tests.rs
+++ b/src/download/tests.rs
@@ -257,6 +257,23 @@ mod reqwest {
         assert_eq!(observed_bytes, vec![b'1', b'2', b'3', b'4', b'5']);
         assert_eq!(std::fs::read_to_string(&target_path).unwrap(), "12345");
     }
+
+    #[tokio::test]
+    async fn network_failure_does_not_delete_partial_file() {
+        let _guard = scrub_env().await;
+        let tmpdir = tmp_dir();
+        let target_path = tmpdir.path().join("downloaded.partial");
+        write_file(&target_path, "123");
+
+        let from_url = "http://240.0.0.0:1080".parse().unwrap();
+        Backend::Reqwest(TlsBackend::NativeTls)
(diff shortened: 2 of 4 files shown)


rami3l:
@FranciscoTGouveia Looks easy enough! I'll give it a test drive when I'm back with my machine :)

src/download/mod.rs
```diff
@@ -263,7 +263,9 @@ impl Backend {
 
         // TODO: We currently clear up the cached download on any error, should we restrict it to a subset?
```
rami3l: I think the comment above has made the situation quite clear: I'd expect the file removal to happen when all of the following applies:

1. Not resuming from partial;
2. `err` downcasts to a certain kind of network failure (otherwise your PR title means nothing).

rami3l (commented):
Many thanks for investigating into this!

OTOH I do think we need to be more careful about filtering the right kind of error since we do want to redownload if the cache is proven to be corrupted for example 🙏

src/download/mod.rs
FranciscoTGouveia: Yes, I should have opened this as a draft; my bad.

I have looked into this, and the error returned by [`download_impl`](https://github.com/rust-lang/rustup/blob/b3fdca3c2a913cad6d7db55f37319eeb14e08117/src/download/mod.rs#L274) could actually be downcast to `DownloadError`, which provides more information.

Should I delete this TODO, or should we keep it?
Is `is_network_timeout` a good variable name?

rami3l: @FranciscoTGouveia It looks more like `is_network_failure` to me.
Also, you'd probably adjust the wording of the TODO accordingly, indicating that it'll be worth looking for other cases where removal is not desired.
Otherwise it seems pretty good at the first glance. Thanks :)

src/download/mod.rs
```diff
@@ -44,7 +47,6 @@ pub(crate) async fn download_file_with_resume(
     status: Option<&DownloadStatus>,
     process: &Process,
 ) -> anyhow::Result<()> {
-    use crate::download::DownloadError as DEK;
```
rami3l, line 47: Aside: `DEK` is a bit confusing I think, is it used elsewhere and can we use a more readable alias? Would just `DownloadError` be okay?

src/download/mod.rs
```diff
@@ -261,9 +263,16 @@ impl Backend {
             return Ok(());
         };
 
+        let is_network_timeout = match err.downcast_ref::<DEK>() {
+            Some(DEK::Reqwest(e)) => e.is_timeout() || e.is_connect(),
```
rami3l: Question: How did you perform the test and/or is there a way to stably demonstrate the effectiveness of this change?

src/download/mod.rs
FranciscoTGouveia: Done!

src/download/mod.rs
FranciscoTGouveia: Yes, I agree.
Now that I look at it more closely, I think we can actually drop this import, since `DownloadError` is defined in `src/download/mod.rs`, provided we agree to use `DownloadError` instead of `DEK`.

src/download/mod.rs
FranciscoTGouveia: I performed it locally: during an installation, I dropped my internet connection, and after `rustup toolchain install` failed, I looked in `.rustup/downloads` for `.partial` files.
Where, before this patch, there were none, they were now left untouched.
To ensure this only happens when we are certain, I whitelisted the situations in which the removal should not be allowed (`e.is_timeout() || e.is_connect()`).

Would it be preferable to add a unit test for this?

rami3l: @FranciscoTGouveia A test for this would be cool if that's fairly easy to do. That said, I'm not completely sure about the feasibility, so feel free to tell me if it turns out too difficult :)

rami3l: @FranciscoTGouveia I have one remaining concern with this change.

When trying to reproduce your scenario above on my machine:

```console
info: rolling back changes
error: component download failed for cargo-aarch64-apple-darwin: partially downloaded file may have been damaged and was removed, please try again: [..] client error (Connect) [..]
```

... the `.partial` is actually saved which is okay. The transaction system is rolling back changes which is also okay. However the `partially downloaded file may have been damaged and was removed` part might be misleading. Have you encountered something similar and, do you have an idea of fixing that, by adjusting the wording or by adding a flag to be passed over? Many thanks again! 🙏

FranciscoTGouveia: I was thinking of extracting the `is_network_failure` to an external function and using it to determine which error message to throw, implying that I will create a new error message.
However, I can also simply delete the `...and was removed` part from the current `BrokenPartialFile` error.

FranciscoTGouveia: Is `IncompletePartialFile` a good error name?

I used a `match` statement [here](https://github.com/rust-lang/rustup/pull/4695/changes#diff-33afa4cad3e3c6cf900b3d5b69a21fbc1bfbf6adeda7116516764a87ddbf877fR98-R102) because I think it is the most readable approach.
Please let me know if you think I should change it.

src/errors.rs
```diff
@@ -27,6 +27,10 @@ pub struct OperationError(pub anyhow::Error);
 pub enum RustupError {
     #[error("partially downloaded file may have been damaged and was removed, please try again")]
     BrokenPartialFile,
+    #[error(
+        "download was only partially completed; the partial file was kept so it can be resumed, please try again"
```
rami3l: I think we can make this even easier to read: "partially downloaded file was kept for resumption, please try again"

rami3l (approved):
LGTM modulo a minor wording change and some nits, nice work :)

When you are done, just squash the first 3 commits into one (they include some back-and-forth) and we are good to go.

src/download/tests.rs
```diff
@@ -257,6 +257,28 @@ mod reqwest {
...
     }
+
+    #[tokio::test]
+    async fn network_failure_does_not_delete_partial_file() {
+        let _guard = scrub_env().await;
+        let tmpdir = tmp_dir();
+        let target_path = tmpdir.path().join("downloaded.partial");
+        write_file(&target_path, "123");
+
+        let listener = TcpListener::bind("127.0.0.1:1080").unwrap();
```
rami3l: Question: What is the point of this listener here? You don't seem to be using it anyway?

If it's just for the address, maybe just construct it directly, and probably use an explicitly invalid address such as `240.0.0.0:1080`?

```diff
- let listener = TcpListener::bind("127.0.0.1:1080").unwrap();
- let addr = listener.local_addr().unwrap();
- drop(listener);
- let from_url = format!("http://{addr}").parse().unwrap();
-
+ let from_url = format!("http://240.0.0.0:1080").parse().unwrap();
```

src/download/mod.rs
```diff
@@ -44,18 +44,19 @@ pub(crate) async fn download_file_with_resume(
...
 ) -> anyhow::Result<()> {
-    use crate::download::DownloadError as DEK;
     match download_file_(url, path, hasher, resume_from_partial, status, process).await {
         Ok(_) => Ok(()),
         Err(e) => {
             if e.downcast_ref::<std::io::Error>().is_some() {
                 return Err(e);
             }
-            let is_client_error = match e.downcast_ref::<DEK>() {
+            let is_client_error = match e.downcast_ref::<DownloadError>() {
```
djc, line 53: Would prefer to have this in a separate commit.

djc (approved):

src/download/mod.rs
rami3l: @FranciscoTGouveia Okay, let's remove the `DEK` alias in the first commit, before all the rest of your work :)

rami3l (approved):
Nice work :)

PR #4696, opened 2026-01-31
feat(cli/rustup-mode): add "Exit status" section to `rustup check --help`

appleGun22:
In continuation to #4694.

### DIFF
--- a/src/cli/help.rs
+++ b/src/cli/help.rs
@@ -216,6 +216,15 @@ pub(crate) fn run_help() -> String {
     )
 }
 
+pub(crate) fn check_help() -> String {
+    format!(
+        r"{HEADER}Exit status:{HEADER:#}
+  {LITERAL}0{LITERAL:#}   Everything is up to date.
+  {LITERAL}100{LITERAL:#} At least one update is available.
+  {LITERAL}1{LITERAL:#}   An error occurred."
+    )
+}
+
 pub(crate) fn doc_help() -> String {
     format!(
         r"{HEADER}Discussion:{HEADER:#}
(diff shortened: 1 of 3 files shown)


src/cli/help.rs
```diff
@@ -216,6 +216,16 @@ pub(crate) fn run_help() -> String {
     )
 }
 
+pub(crate) fn check_help() -> String {
+    format!(
+        r"{HEADER}Exit status:{HEADER:#}
+
+  {LITERAL}0{LITERAL:#}   All toolchains are up to date.
+  {LITERAL}100{LITERAL:#} At least one toolchain has an update available.
```
rami3l: The actual situation is more nuanced than that since the first line of `rustup check --help` says:

```
Check for updates to Rust toolchains and rustup
```

So I'd suggest changing the wording to something like `No updates found`/`At least one update is available` to reduce redundancy.

rami3l (commented):
Please update the test snapshots at the same time with `SNAPSHOTS=overwrite` so that the CI turns green again. Many thanks for your help!

rami3l (approved):
@appleGun22 LGTM!

Please change the commit message and the PR title to better reflect what this patch is doing: clearly it's not about adding `--help` since it's already been implemented. Rather, you are adding a section about exit codes to it.

Apart from that, just squash all commits into one and we are good to go.

Many thanks again :)

djc (approved):

src/cli/help.rs
```diff
@@ -216,6 +216,16 @@ pub(crate) fn run_help() -> String {
     )
 }
 
+pub(crate) fn check_help() -> String {
+    format!(
+        r"{HEADER}Exit status:{HEADER:#}
+
```
rami3l: There is one extra newline which is inconsistent with its surroundings.

rami3l (dismissed):

appleGun22:
@rami3l I see that you already squashed them together, but still thank you for the guidance, it was my first PR to rustup :)

djc (approved):

PR #4700, opened 2026-02-02
cli: introduce semantic exit code constants for rustup check

dharmvr1:
continuation to #4694

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -66,7 +66,7 @@ fn handle_epipe(res: Result<ExitCode>) -> Result<ExitCode> {
             if let Some(io_err) = root.downcast_ref::<io::Error>()
                 && io_err.kind() == io::ErrorKind::BrokenPipe
             {
-                return Ok(ExitCode(0));
+                return Ok(ExitCode::SUCCESS);
             }
             Err(e)
         }
@@ -597,12 +597,12 @@ pub async fn main(
         Ok(matches) => matches,
         Err(err) if err.kind() == DisplayHelp => {
             write!(process.stdout().lock(), "{}", err.render().ansi())?;
-            return Ok(ExitCode(0));
+            return Ok(ExitCode::SUCCESS);
         }
         Err(err) if err.kind() == DisplayVersion => {
             write!(process.stdout().lock(), "{}", err.render().ansi())?;
             display_version(current_dir, process).await?;
-            return Ok(ExitCode(0));
+            return Ok(ExitCode::SUCCESS);
         }
         Err(err) => {
             if [
@@ -616,7 +616,7 @@ pub async fn main(
             } else {
                 write!(process.stderr().lock(), "{}", err.render().ansi())?;
             }
-            return Ok(ExitCode(1));
+            return Ok(ExitCode::FAILURE);
         }
     };
 
@@ -628,7 +628,7 @@ pub async fn main(
     let Some(subcmd) = matches.subcmd else {
         let help = Rustup::command().render_long_help();
         writeln!(process.stderr().lock(), "{}", help.ansi())?;
-        return Ok(ExitCode(1));
+        return Ok(ExitCode::FAILURE);
     };
 
     match subcmd {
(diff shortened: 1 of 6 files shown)


dharmvr1:
Hi @rami3l @ChrisDenton — quick check: is this the right approach, or should exit codes be handled differently? Also, do we want to replace all exit codes with named constants, or only the ones in the update/no-update path? Any advice would be appreciated.
thanks

ChrisDenton:
Yes, it should change all exit codes to use constants.

src/cli/exit_code.rs
```diff
@@ -0,0 +1,9 @@
+//! Semantic exit codes for rustup CLI commands.
```
ChrisDenton: I don't think it's worth adding a new file just for constants

src/cli/exit_code.rs
```diff
@@ -0,0 +1,9 @@
...
+
+/// No updates are available.
+pub const EXIT_NO_UPDATES: i32 = 0;
```
ChrisDenton: This should just be `EXIT_SUCCESS`. There's no need for an alias.

src/cli/exit_code.rs
```diff
@@ -0,0 +1,9 @@
...
+
+/// Updates are available.
+///
+/// Matches the DNF `check-update` convention.
+pub const EXIT_UPDATES_AVAILABLE: i32 = 100;
```
rami3l: These should be associated constants on the `crate::utils::ExitCode` type rather than straight `i32`s.

rami3l (changes requested):

src/cli/exit_code.rs
rami3l: Actually there is a need for an alias if it hasn't been added yet, considering https://github.com/rust-lang/rustup/pull/4700#discussion_r2755364060.

src/cli/rustup_mode.rs
dharmvr1: @ChrisDenton I’m a bit confused here — based on the discussion in #4681 , exit code 1 seems intentionally treated as a generic error. Just wanted to confirm: is it correct to keep this generalized and not add a named constant for it, since multiple kinds of errors map to it?

rami3l: You should name ExitCode(0) a constant ExitCode::SUCCESS, ExitCode(1) a constant ExitCode::FAILURE, and ExitCode(100) a constant ExitCode::NO_UPDATES.

src/utils/mod.rs
```diff
@@ -29,6 +29,17 @@ pub(crate) mod units;
 #[derive(Debug, PartialEq, Eq)]
 pub struct ExitCode(pub i32);
 
+impl ExitCode {
+    /// Successful execution.
+    pub const SUCCESS: i32 = 0;
```
rami3l: Those constants should have `ExitCode` type?

rami3l (changes requested):

src/cli/rustup_mode.rs
```diff
@@ -597,12 +597,12 @@ pub async fn main(
         Ok(matches) => matches,
         Err(err) if err.kind() == DisplayHelp => {
             write!(process.stdout().lock(), "{}", err.render().ansi())?;
-            return Ok(ExitCode(0));
+            return Ok(ExitCode::FAILURE);
```
rami3l: Please double check all replacements and prevent errors like this from happening, thanks :)

src/utils/mod.rs
```diff
@@ -29,6 +29,17 @@ pub(crate) mod units;
 #[derive(Debug, PartialEq, Eq)]
 pub struct ExitCode(pub i32);
 
+impl ExitCode {
+    /// Successful execution.
+    pub const SUCCESS: ExitCode = ExitCode(0);
```
rami3l: Nit: since you are in an implementation block, this should be legal:
```suggestion
    pub const SUCCESS: Self = Self(0);
```

Same thing for all `const`s in this block.

rami3l (approved):
LGTM modulo https://github.com/rust-lang/rustup/pull/4700#discussion_r2759488152, many thanks!

PS: Please squash all commits into one when you are done so that we can merge it.

rami3l (approved):
LGTM modulo https://github.com/rust-lang/rustup/pull/4700#discussion_r2759488152, many thanks!

PS: Please squash all commits into one when you are done so that we can merge it.

rami3l (commented):
Just change the commit message to be the same as the PR title please? "wip" is not very descriptive :]

dharmvr1:
Thanks a lot @rami3l  @ChrisDenton  for helping me get my first PR merged ,really appreciate the guidance and patience, and sorry if I bothered you a bit along the way 😅 Learned a lot from this.
I’ve got some free time right now, so if there’s any task , feel free to ping me anytime. Thanks again!

PR #4701, opened 2026-02-03
Replace `$nu.home-path` with `~` in `env.nu` file

### LINKED ISSUE #4703: startup failing on nushell `v0.110.0` due to breaking change from `$nu.home-path` to `$nu.home-dir`
### Verification

### Problem

https://www.nushell.sh/blog/2026-01-17-nushell_v0_110_0.html#renamed-nu-temp-path-and-nu-home-path-toc

this causes both the `source $"($nu.home-path)/.cargo/env.nu"  # For nushell` printed after installation and `path add $"($nu.home-path)/.cargo/bin"` in `.cargo/env.nu` to fail with the following

```
Error: nu::parser::error

  × Error: nu::shell::column_not_found
  │
  │   × Cannot find column 'home-path'
  │    ╭─[source:1:1]
  │  1 │ nu --login
  │    · ▲
  │    · ╰── value originates here
  │    ╰────
  │     ╭─[/root/.config/nushell/config.nu:22:15]
  │  22 │ source $"($nu.home-path)/.cargo/env.nu"
  │     ·               ────┬────
  │     ·                   ╰── cannot find column 'home-path'
  │     ╰────
  │
    ╭─[/root/.config/nushell/config.nu:22:8]
 22 │ source $"($nu.home-path)/.cargo/env.nu"
    ·        ────────────────┬───────────────
    ·                        ╰── Encountered error during parse-time evaluation
    ╰────
```

### Steps

1. Install rustup using https://rustup.rs/
2. Write source line printed after installation in nushell config if not already present
3. Restart shell to see the issue

### Possible Solution(s)

Change from `home-path` to `home-dir` fixes the issue

### Notes

_No response_

### Rustup version

```console
info: The currently active `rustc` version is `rustc 1.93.0 (254b59607 2026-01-19)`
```

### Installed toolchains

```console
rustup home:  /root/.rustup

installed toolchains
--------------------

active toolchain
----------------
installed targets:
  x86_64-unknown-linux-gnu
```

### OS version

```console
Fedora Linux 43 (Cloud Edition) x86_64
Linux 6.18.7-200.fc43.x86_64
```

VasanthakumarV:
`$nu.home-path` was renamed to `$nu.home-dir` in the latest nushell 0.110 release (https://github.com/nushell/nushell/pull/17129).

This PR replaces `$nu.home-path` with `~`, so that both old and new versions of nushell remain compatible.

`~` is a shorthand for `$nu.home-dir`, as per [nushell's doc](https://www.nushell.sh/book/special_variables.html#nu).

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -454,7 +454,7 @@ the corresponding `env` file under {cargo_home}.
 This is usually done by running one of the following (note the leading DOT):
     . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
     source "{cargo_home}/env.fish"  # For fish
-    source $"{cargo_home_nushell}/env.nu"  # For nushell
+    source "{cargo_home_nushell}/env.nu"  # For nushell
     source "{cargo_home}/env.tcsh"  # For tcsh
     . "{cargo_home}/env.ps1"        # For pwsh
     source "{cargo_home}/env.xsh"   # For xonsh
--- a/src/cli/self_update/env.nu
+++ b/src/cli/self_update/env.nu
@@ -1,2 +1,2 @@
 use std/util "path add"
-path add $"{cargo_bin}"
+path add "{cargo_bin}"
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -317,7 +317,7 @@ impl UnixShell for Nu {
     }
 
     fn cargo_home_str(&self, process: &Process) -> Result<Cow<'static, str>> {
-        cargo_home_str_with_home("($nu.home-path)", process)
+        cargo_home_str_with_home("~", process)
     }
 }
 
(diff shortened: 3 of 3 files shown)


rami3l (approved):
LGTM, but when was this `~` shorthand introduced? I'm curious about how compatible we can be when it comes to this (for fish we have explicitly used old syntax I think, for example).

VasanthakumarV:
> LGTM, but when was this `~` shorthand introduced? I'm curious about how compatible we can be when it comes to this (for fish we have explicitly used old syntax I think, for example).

0.1 version of `nu` was published to crates.io on May-2019.
Good support for substituting ~ with the home-directory came on Dec-2019 (https://github.com/NiklasJonsson/nushell/commit/219b7e64cdaabad9e95067b0e5abcbd4e396b9c1).

djc (approved):

PR #4708, opened 2026-02-08
chore(deps): downgrade `openssl-src` to 300.5.4+3.5.4

### LINKED ISSUE #4707: Build failure when targeting `powerpc64-unknown-linux-gnu`
The regression happened between 20260202 and 20260204:

Image

```
= note: /x-tools/powerpc64-unknown-linux-gnu/lib/gcc/powerpc64-unknown-linux-gnu/8.5.0/../../../../powerpc64-unknown-linux-gnu/bin/ld: /checkout/target/powerpc64-unknown-linux-gnu/debug/deps/libopenssl_sys-05b0abda159f9dc2.rlib(libcrypto-lib-aes-gcm-ppc.o): ABI version 2 is not compatible with ABI version 1 output
          /x-tools/powerpc64-unknown-linux-gnu/lib/gcc/powerpc64-unknown-linux-gnu/8.5.0/../../../../powerpc64-unknown-linux-gnu/bin/ld: failed to merge target specific data of file /checkout/target/powerpc64-unknown-linux-gnu/debug/deps/libopenssl_sys-05b0abda159f9dc2.rlib(libcrypto-lib-aes-gcm-ppc.o)
```
_https://github.com/rust-lang/rustup/actions/runs/21734438192/job/62696277679_

rami3l:
Closes #4707, verified in https://github.com/rust-lang/rustup/actions/runs/21796260865/job/62884244255?pr=4708.

### DIFF
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -554,10 +554,11 @@ mod curl {
 
 #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
 mod reqwest_be {
-    use std::io;
+    #[cfg(feature = "reqwest-rustls-tls")]
+    use std::sync::Arc;
     #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
-    use std::sync::{Arc, OnceLock};
-    use std::time::Duration;
+    use std::sync::OnceLock;
+    use std::{io, time::Duration};
 
     #[cfg(all(feature = "reqwest-rustls-tls", not(target_os = "android")))]
     use crate::anchors::RUSTUP_TRUST_ANCHORS;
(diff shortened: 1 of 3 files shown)


src/download/mod.rs
```diff
@@ -541,10 +541,11 @@ mod curl {
 
 #[cfg(any(feature = "reqwest-rustls-tls", feature = "reqwest-native-tls"))]
 mod reqwest_be {
-    use std::io;
+    #[cfg(feature = "reqwest-rustls-tls")]
+    use std::sync::Arc;
```
rami3l, line 558: When looking at the CI logs, I realized that the old version actually has a redundant `use std::sync::Arc;` on `cfg!(feature = "reqwest-native-tls")`, so I'm taking this opportunity to clean that up.

rami3l:
It looks like cargo has done the same in https://github.com/rust-lang/cargo/pull/16601

djc (approved):

rami3l:
@djc Many thanks for the review!

One concern is that if we don't receive an OpenSSL release the next Monday following this PR's being merged, it might be bumped back to a problematic version. Do you think we should downgrade-pin it instead and merge it now, or don't merge it until the release is ready?

djc:
I think pinning could make sense for this.

PR #4717, opened 2026-02-13
fix(diskio): fall back to single-threaded unpacking 

### LINKED ISSUE #3125: OOM installing rustc-dev with 1GB of memory
### Problem

I tried installing a rust toolchain in a VM I assigned 1GB of ram. (100MB used by the minimal debian install) When installing rustc-dev it got an OOM however.

### Steps

1. Create a VM
2. Assign 1GB of ram
3. Install a minimal debian system in the VM. When running it only uses 100MB of ram.
4. Attempt to install a rust toolchain with the rustc-dev component in the VM.
5. OOM

### Possible Solution(s)

I thought rustup 1.24 enabled installation on systems with little ram. Is this a regression? Would it be possible to stream the component rather than attempt to load it all at once in memory?

### Notes

For now I will be increasing the assigned amount of memory, but IMO it should be possible to install on systems with this little ram, hence this issue.

### Rustup version

```console
```

### Installed toolchains

```console
rustup home:  /home/bjorn/.rustup

no active toolchain
```

I attempted to install nightly-2022-12-13-x86_64-unknown-linux-gnu.

cachebag:
When`ram_budget`< 512MB, fall back to single-threaded unpacking to avoid OOM on memory-constrained systems.

Rustup OOMs on systems with <1GB RAM because it spawns multiple I/O threads for unpacking regardless of available memory. The Threaded executor uses far more memory than its buffer pool budget tracks. This PR adds a 512MB `ram_budget` threshold below which rustup falls back to single-threaded unpacking, which peaks at ~110MB.

Fixes #3125

### DIFF
--- a/src/diskio/mod.rs
+++ b/src/diskio/mod.rs
@@ -62,6 +62,7 @@ use std::{fmt::Debug, fs::OpenOptions};
 use anyhow::Result;
 use tracing::{error, trace, warn};
 
+use crate::process::IoThreadCount;
 use crate::utils::units::Size;
 
 pub(crate) mod immediate;
@@ -445,12 +446,29 @@ pub(crate) fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
 /// Get the executor for disk IO.
 pub(crate) fn get_executor<'a>(
     ram_budget: usize,
-    io_thread_count: usize,
+    thread_count: IoThreadCount,
 ) -> Box<dyn Executor + 'a> {
     // If this gets lots of use, consider exposing via the config file.
-    match io_thread_count {
+    let threads = effective_thread_count(ram_budget, thread_count);
+    match threads {
         0 | 1 => Box::new(immediate::ImmediateUnpacker::new()),
         n => Box::new(threaded::Threaded::new(n, ram_budget)),
+    }
+}
+
+fn effective_thread_count(ram_budget: usize, thread_count: IoThreadCount) -> usize {
+    match thread_count {
+        IoThreadCount::Default(n) if n > 1 && ram_budget < LOW_MEMORY_THRESHOLD => {
+            warn!(
+                "using single-threaded unpacking due to low memory \
+                 (ram budget: {} < {} threshold), \
+                 set RUSTUP_IO_THREADS to override",
+                Size::new(ram_budget),
+                Size::new(LOW_MEMORY_THRESHOLD)
+            );
+            1
+        }
+        IoThreadCount::Default(n) | IoThreadCount::UserSpecified(n) => n,
     }
 }
 
@@ -505,3 +523,11 @@ pub(crate) fn unpack_ram(io_chunk_size: usize, budget: Option<usize>) -> usize {
 }
 
 static RAM_NOTICE_SHOWN: OnceLock<()> = OnceLock::new();
+
+/// The Threaded executor uses substantially more memory than its ram_budget
+/// accounts for (pool overhead, sharded_slab metadata, multiple in-flight
+/// operations, thread stacks, allocator fragmentation). On systems where the
+/// ram_budget is under this threshold, fall back to single-threaded unpacking
+/// which peaks at ~110MB.
+/// See https://github.com/rust-lang/rustup/issues/3125
+const LOW_MEMORY_THRESHOLD: usize = 512 * 1024 * 1024;
--- a/src/diskio/test.rs
+++ b/src/diskio/test.rs
@@ -163,3 +163,46 @@ fn test_complete_file_immediate() {
 fn test_complete_file_threaded() {
     test_complete_file("2").unwrap()
 }
+
+#[test]
+fn test_effective_thread_count() {
+    use super::{LOW_MEMORY_THRESHOLD, effective_thread_count};
+    use crate::process::IoThreadCount::{Default, UserSpecified};
+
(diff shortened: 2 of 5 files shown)


src/process.rs
```diff
@@ -69,23 +82,23 @@ impl Process {
...
-                _ => return Ok(threads),
+                _ => return Ok(IoThreadCount::UserSpecified(threads)),
             }
         };
 
         Ok(match thread::available_parallelism() {
             // Don't spawn more than 8 I/O threads unless the user tells us to.
             // Feel free to increase this value if it improves performance.
-            Ok(threads) => Ord::min(threads.get(), 8),
+            Ok(threads) => IoThreadCount::Default(Ord::min(threads.get(), 8)),
```
djc: Nit: suggest hoisting `IoThreadCount::Default(..)` out of the `match`.

src/process.rs
```diff
@@ -31,6 +31,19 @@ mod file_source;
 mod terminal_source;
 pub use terminal_source::ColorableTerminal;
 
+pub enum IoThreadCount {
```
djc: Nit: we follow top-down style, move this below its user.

src/diskio/mod.rs
```diff
@@ -442,18 +443,45 @@ pub(crate) fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
...
     }
 }
 
+fn effective_thread_count(ram_budget: usize, thread_count: IoThreadCount) -> usize {
+    match thread_count {
+        IoThreadCount::UserSpecified(n) => n,
+        IoThreadCount::Default(n) if n <= 1 => n,
+        IoThreadCount::Default(n) if ram_budget < LOW_MEMORY_THRESHOLD => {
+            warn!(
+                "Using single-threaded unpacking due to low memory \
```
djc: Nits: I think our usual style does not use capitals (and periods).

src/diskio/mod.rs
```diff
@@ -442,18 +443,45 @@ pub(crate) fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
     std::fs::create_dir(path)
 }
 
+/// The Threaded executor uses substantially more memory than its ram_budget
+/// accounts for (pool overhead, sharded_slab metadata, multiple in-flight
+/// operations, thread stacks, allocator fragmentation). On systems where the
+/// ram_budget is under this threshold, fall back to single-threaded unpacking
+/// which peaks at ~110MB.
+/// See https://github.com/rust-lang/rustup/issues/3125
+const LOW_MEMORY_THRESHOLD: usize = 512 * 1024 * 1024;
```
djc: Nit: move this to the bottom of the module, please.

djc (approved):
Nice, this makes sense to me!

src/process.rs
```diff
@@ -235,6 +236,19 @@ impl Process {
     }
 }
 
+pub enum IoThreadCount {
+    Default(usize),
+    UserSpecified(usize),
+}
+
+impl IoThreadCount {
+    pub fn count(&self) -> usize {
```
rami3l: This is a bit confusing to me. Why not just use [From](https://doc.rust-lang.org/std/convert/trait.From.html) and call `.into()` instead?

src/diskio/mod.rs
```diff
@@ -445,15 +446,34 @@ pub(crate) fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
...
         0 | 1 => Box::new(immediate::ImmediateUnpacker::new()),
         n => Box::new(threaded::Threaded::new(n, ram_budget)),
     }
 }
 
+fn effective_thread_count(ram_budget: usize, thread_count: IoThreadCount) -> usize {
+    match thread_count {
+        IoThreadCount::UserSpecified(n) => n,
+        IoThreadCount::Default(n) if n <= 1 => n,
+        IoThreadCount::Default(n) if ram_budget < LOW_MEMORY_THRESHOLD => {
```
rami3l: Nit: I think this predicate will be clearer like this:

```suggestion
        IoThreadCount::Default(n) if n > 1 && ram_budget < LOW_MEMORY_THRESHOLD => {
```

src/process.rs
cachebag: I thought that maybe the explicitness would have been helpful here

src/diskio/test.rs
```diff
@@ -163,3 +163,34 @@ fn test_complete_file_immediate() {
...
     test_complete_file("2").unwrap()
 }
+
+#[test]
+fn test_effective_thread_count() {
+    use super::{LOW_MEMORY_THRESHOLD, effective_thread_count};
+    use crate::process::IoThreadCount::{Default, UserSpecified};
+
+    // Already single-threaded: no change regardless of budget
+    assert_eq!(effective_thread_count(32 * 1024 * 1024, Default(1)), 1);
```
rami3l: Nit: Similar to `LOW_MEMORY_THRESHOLD - 1`, to convey the intent more clearly it'd probably be better to just write `LOW_MEMORY_THRESHOLD / 16`, `LOW_MEMORY_THRESHOLD / 2`, `LOW_MEMORY_THRESHOLD * 2`, etc., so that we don't need to guess what were above or below the constant.

rami3l (approved):
LGTM modulo one final nit (my apologies for the back and forth), and thanks again for your contribution :)

PR #4718, opened 2026-02-13
Ensure the server honors range requests to prevent download corruption

### LINKED ISSUE #4710: Download resume can silently corrupt files if server ignores Range header
When resuming a partial download (`resume_from > 0`), rustup sends a `Range: bytes=N-` header and appends the response to the existing partial file. However, it doesn't verify that the server actually honored the Range request.

If the server ignores the Range header and returns `200 OK` with the full file (instead of `206 Partial Content`), the full file content gets appended to the partial file, producing a corrupted result.

The response code check accepts any 2xx:

https://github.com/rust-lang/rustup/blob/b3fdca3c2a913cad6d7db55f37319eeb14e08117/src/download/mod.rs#L302-L307

```rust
match code {
    0 | 200..=299 => {}
    _ => {
        return Err(DownloadError::HttpStatus(code).into());
    }
};
```

When `resume_from > 0`, this should verify that the response is `206 Partial Content`. If the server returns `200`, the download should either restart from scratch (truncating the partial file) or return an error.

The reqwest backend has the same issue — it checks `res.status().is_success()` without distinguishing 200 from 206.

Found while porting rustup's download resume logic to [elan](https://github.com/leanprover/elan/pull/191).

This issue has been assigned to @FranciscoTGouveia via [this comment](https://github.com/rust-lang/rustup/issues/4710#issuecomment-3887272508).

FranciscoTGouveia:
Fixes #4710.

As discussed in the issue, I am still unsure what position rustup should take: should we simply throw an error, or should we go a step further and restart the download from scratch?

I am more inclined toward the former, since correctly handling the range header should be the server's responsibility.
This would only become an issue if `RUSTUP_DIST_SERVER` is overridden; of course, this is the case for all forks of rustup (e.g., `elan`, or `juliaup`).

On another note, I would appreciate feedback on the added test.
I introduced a small workaround by adding a variable to [`serve_contents`](https://github.com/rust-lang/rustup/blob/4cc92aad109e18ee5ced2a82a7a2fbceaec023cc/src/download/tests.rs#L348-L353) to allow the mock server to ignore the range request. Is there a cleaner way to achieve this?

### DIFF
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -538,14 +538,14 @@ mod curl {
                 })?;
             }
 
-            // If we didn't get a 20x or 0 ("OK" for files) then return an error
+            // If we didn't get a 20x or 0 ("OK" for files) then return an error.
+            // If resuming a download, we need a 206, as a 200 would mean the server ignored
+            // the range header, resulting in corruption.
             let code = handle.response_code()?;
-            match code {
-                0 | 200..=299 => {}
-                _ => {
-                    return Err(DownloadError::HttpStatus(code).into());
-                }
-            };
+            match (resume_from > 0, code) {
+                (_, 0) | (true, 206) | (false, 200..=299) => {}
+                _ => return Err(DownloadError::HttpStatus(code).into()),
+            }
 
             Ok(())
         })
--- a/src/download/mod.rs
+++ b/src/download/mod.rs
@@ -587,9 +587,13 @@ mod reqwest_be {
             .await
             .context("error downloading file")?;
 
-        if !res.status().is_success() {
-            let code: u16 = res.status().into();
-            return Err(anyhow!(DownloadError::HttpStatus(u32::from(code))));
+        // If a download is being resumed, we expect a 206 response;
+        // otherwise, if the server ignored the range header,
+        // an error is thrown preemptively to avoid corruption.
+        let status = res.status().into();
+        match (resume_from > 0, status) {
+            (true, 206) | (false, 200..=299) => {}
+            _ => return Err(DownloadError::HttpStatus(u32::from(status)).into()),
         }
 
         if let Some(len) = res.content_length() {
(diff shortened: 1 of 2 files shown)


src/download/mod.rs
```diff
@@ -538,14 +538,14 @@ mod curl {
...
+            // the range header, resulting in corruption.
             let code = handle.response_code()?;
-            match code {
-                0 | 200..=299 => {}
-                _ => {
-                    return Err(DownloadError::HttpStatus(code).into());
-                }
-            };
+            match (resume_from > 0, code) {
+                (true, 0) | (true, 206) | (false, 0) | (false, 200..=299) => {}
```
rami3l: Nit: This can be simplified to:

```suggestion
                (_, 0) | (true, 206) | (false, 200..=299) => {}
```

src/download/mod.rs
```diff
@@ -587,9 +587,10 @@ mod reqwest_be {
             .await
             .context("error downloading file")?;
 
-        if !res.status().is_success() {
-            let code: u16 = res.status().into();
-            return Err(anyhow!(DownloadError::HttpStatus(u32::from(code))));
+        let status = res.status().into();
```
rami3l, line 593: Nit: Note that the `curl` backend is going away soon, bring and adapting the comments from https://github.com/rust-lang/rustup/pull/4718/changes#r2807888352 would be greatly appreciated :)

rami3l (approved):
LGTM modulo some tiny nits, nice work :)

I do think that we should throw an error directly since that might urge the server side to rectify the protocol conformance issue.

As for the test code changes, it's completely okay this way so don't worry :D

PS: The CI failure for Android is a separate problem (#4720), I'm on it already.

src/download/mod.rs
FranciscoTGouveia: Absolutely! Sorry for not noticing this :/

djc (approved):

Issue #4724, opened 2026-02-19
New parallel download + install can easily leave a partially installed toolchain when downloads are interrupted


steffahn:
I haven’t yet checked if current rustup produces a similar state when aborted during the extraction phase of an install, but in any case, it’s going to be much more common that someone manually interrupts the process during the download phase, because downloads can get stuck and timeouts are slow.

What I did:
```
$ rustup --version
rustup 1.29.0 :: 1.28.2+486 (732af7663 2026-02-02)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: the currently active `rustc` version is `rustc 1.93.1 (01f6ddf75 2026-02-11)`

$ rustup toolchain install nightly-2026-01-01
info: syncing channel updates for nightly-2026-01-01-x86_64-unknown-linux-gnu
info: latest update on 2026-01-01 for version 1.94.0-nightly (8d670b93d 2025-12-31)
info: downloading 6 components
        cargo installed                       10.36 MiB
       clippy installed                        4.41 MiB
    rust-docs downloading [######         ]   20.58 MiB (304.31 KiB/s, ETA: 44s)
     rust-std downloading [#              ]   28.29 MiB (208.89 KiB/s, ETA: 2m)
^C
```
(interrupted via ctrl+c while cargo & clippy was already installed, and the next ones were in the middle of downloading)

then:
```
$ cargo +nightly-2026-01-01 --version 
cargo 1.94.0-nightly (b54051b15 2025-12-30)

$ rustc +nightly-2026-01-01 --version
error: Missing manifest in toolchain 'nightly-2026-01-01-x86_64-unknown-linux-gnu'

$ cargo +nightly-2026-01-01 clippy --version
/home/frank/.rustup/toolchains/nightly-2026-01-01-x86_64-unknown-linux-gnu/bin/cargo-clippy: error while loading shared libraries: librustc_driver-bcf6c5c396c0890f.so: cannot open shared object file: No such file or directory
```
I think this is an undesirable state to leave someone’s install in. Additionally, if re-tried now, it produces a lot of extra warnings:
```
$ rustup toolchain install nightly-2026-01-01
info: syncing channel updates for nightly-2026-01-01-x86_64-unknown-linux-gnu
info: latest update on 2026-01-01 for version 1.94.0-nightly (8d670b93d 2025-12-31)
info: removing previous version of component rustc
warn: component rustc not found during uninstall
info: removing previous version of component cargo
warn: component cargo not found during uninstall
[...]

PR #4725, opened 2026-02-20
Improve logs when recovering from an interrupted installation

FranciscoTGouveia:
Related to #4724.

Currently, when a toolchain installation is retried after being interrupted, an excessive number of warnings are emitted.

This patch detects that situation, skips the redundant warnings, and prints a single, more informative message instead.

I would appreciate any feedback on the wording of the new message.

---
For reference, I leave below the before and after:

**Before**:
```
$ rustup toolchain install 1.90 --no-self-update
info: syncing channel updates for 1.90-x86_64-unknown-linux-gnu
info: latest update on 2025-09-18 for version 1.90.0 (1159e78c4 2025-09-14)
info: removing previous version of component rustc
warn: component rustc not found during uninstall
info: removing previous version of component cargo
warn: component cargo not found during uninstall
info: removing previous version of component rust-std
warn: component rust-std not found during uninstall
info: removing previous version of component rust-docs
warn: component rust-docs not found during uninstall
info: removing previous version of component rustfmt
warn: component rustfmt not found during uninstall
info: removing previous version of component clippy
warn: component clippy not found during uninstall
info: downloading 6 components
        rustc installed                       78.70 MiB
        cargo installed                       10.18 MiB
     rust-std installed                       27.77 MiB
    rust-docs installed                       20.49 MiB
      rustfmt installed                        2.15 MiB
       clippy installed                        4.33 MiB
  1.90-x86_64-unknown-linux-gnu installed - rustc 1.90.0 (1159e78c4 2025-09-14)
  ```

**After**:
```
$ rustup toolchain install 1.90 --no-self-update
info: syncing channel updates for 1.90-x86_64-unknown-linux-gnu
info: latest update on 2025-09-18 for version 1.90.0 (1159e78c4 2025-09-14)
info: recovering from a partially installed toolchain
info: downloading 6 components
        rustc installed                       78.70 MiB
        cargo installed                       10.18 MiB
     rust-std installed                       27.77 MiB
    rust-docs installed                       20.49 MiB
      rustfmt installed                        2.15 MiB
       clippy installed                        4.33 MiB
  1.90-x86_64-unknown-linux-gnu installed - rustc 1.90.0 (1159e78c4 2025-09-14)
```

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -15,7 +15,7 @@ use anyhow::{Context as _, Result, anyhow, bail};
 use futures_util::Stream;
 use futures_util::stream::{FuturesUnordered, StreamExt};
 use tokio::task::{JoinHandle, spawn_blocking};
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 
 use crate::diskio::{Executor, IO_CHUNK_SIZE, get_executor, unpack_ram};
 use crate::dist::component::{Components, DirectoryPackage, Transaction};
@@ -185,6 +185,12 @@ impl Manifestation {
         // to uninstall it first.
         tx = self.maybe_handle_v2_upgrade(&config, tx)?;
 
+        // If there are no components installed, and we are recovering from
+        // a partial installation, we can provide a more informative message.
+        if !update.components_to_uninstall.is_empty() && self.installation.list()?.is_empty() {
+            info!("recovering from a partially installed toolchain");
+        }
+
         // Uninstall components
         for component in update.components_to_uninstall {
             match (implicit_modify, &component.target) {
@@ -329,7 +335,7 @@ impl Manifestation {
         } else if let Some(c) = self.installation.find(short_name)? {
             tx = c.uninstall(tx)?;
         } else {
-            warn!(
+            debug!(
                 "component {} not found during uninstall",
                 manifest.short_name(&component),
             );

### CODE REVIEW

File: src/dist/manifestation.rs [hunk 1]
```diff
@@ -187,6 +187,12 @@ impl Manifestation {
 
         // Uninstall components
         for component in update.components_to_uninstall {
+            // If there are no components installed, we don't need to keep
+            // checking, we can simply exit the loop earlier.
+            if self.installation.list()?.is_empty() {
```
rami3l: I think in principle this fix is correct, thanks!

However, isn't `.list()` doing a FS read every time in this loop? I'd imagine the result of `.installation.list()` to be perfectly predictable (don't care about ToC/ToU for now), so may I ask what do we have in `.components_to_uninstall` and `.installation.list()` here?

In other words, could these be two sets of elements of the same format, so that we can make a _difference_?

FranciscoTGouveia: Yes, indeed, `.list()` implies a FS read, so a reasonable solution may be to enter the loop only if `.list()` is not empty (i.e., by adding a conditional check before the loop).

> so may I ask what do we have in `.components_to_uninstall` and `.installation.list()` here?

In this case, as the toolchain was not installed, `.components_to_uninstall` was set to all the 6 components (see [here](https://github.com/rust-lang/rustup/blob/e0e1f45454291049db50e1ea93455e696674fcd9/src/dist/manifestation.rs#L678)), and `.installation.list()` was empty.


That said, I think a better approach would be to restrict the `components_to_uninstall` vector to only include components that actually exist; this [comment](https://github.com/rust-lang/rustup/blob/e0e1f45454291049db50e1ea93455e696674fcd9/src/dist/manifestation.rs#L645-L646) may be misleading in that regard.
This would eliminate the need for the additional check and further simplify the `uninstall_component` function.

What do you think about this approach?

rami3l: Rethinking about it I found out that I might need to be more careful about the aforementioned approach. Could this loop be written this way to defend against the interrupted in-flight state where the component is partially in-place but not appended to the final list yet?

If that is the case, maybe those checks are not to be skipped since otherwise the installation later on will fail (due to the previous version not being uninstalled)... and we should simply downgrade those warning messages to a lower logging level (say debug) and concatenating the info messages? 🤔

FranciscoTGouveia: Yes, so we can simply change the `warn!` to `debug!`, and merge all the messages at the beginning by moving this new condition before the loop?
This way, we would reduce noisy log messages while keeping the loop defensive as it is.

rami3l: @FranciscoTGouveia Yeah, let's keep the old logic for now. Thanks! 🙏

djc (approved):

src/dist/manifestation.rs
```diff
@@ -185,6 +185,12 @@ impl Manifestation {
         // to uninstall it first.
         tx = self.maybe_handle_v2_upgrade(&config, tx)?;
 
+        // If there are no components installed, we don't need to keep
+        // checking, we can simply exit the loop earlier.
```
rami3l: Nit: Update the comments since there's no "loop" right now...

src/dist/manifestation.rs
```diff
@@ -185,6 +185,12 @@ impl Manifestation {
...
+        if self.installation.list()?.is_empty() && !update.components_to_uninstall.is_empty() {
```
rami3l: Nit: since `.installation.list()` is an FS operation, we should short-circuit it if we can:

```suggestion
        if !update.components_to_uninstall.is_empty() && self.installation.list()?.is_empty() {
```

rami3l (approved):
Many thanks :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4725/e143373287648906f1370b0db40c77c938b11398..e2909f7454e74c82b10f7b3727672519df2ac15a)*

src/dist/manifestation.rs
FranciscoTGouveia: You're right, apologies.
I have already updated the wording.

src/dist/manifestation.rs
FranciscoTGouveia: Sorry for missing that. Updated.

PR #4727, opened 2026-02-23
Add `--override` flag in `rustup toolchain install`

### LINKED ISSUE #3105: Support "--override" flag in `rustup toolchain install`

I am installing rustup in CI and would like to avoid repeating myself. Currently that does not entirely work:
```
      - name: Install Rust
        run: |
          rustup toolchain install ${{ matrix.rust }} --profile minimal -c rust-src
          rustup override set ${{ matrix.rust }}
```
I need to repeat the toolchain name if I want to be sure that the rest of the script actually uses the configured toolchain.

### Solution you'd like

It would be nice if something like the following were to work
```
      - name: Install Rust
        run: |
          rustup toolchain install ${{ matrix.rust }} --profile minimal --override -c rust-src
```
The `--override` flag indicates that an override for the current directory should be set up with the given toolchain (this should happen even if the toolchain was already installed).

### Notes

Many projects use https://github.com/actions-rs/toolchain instead of directly invoking rustup, but that has not been maintained in years and deprecation warnings are starting to show up. https://github.com/dtolnay/rust-toolchain provides an alternative, but if a rustup one-line is sufficient, it would not even be necessary to have a github action for this.

FranciscoTGouveia:
Fixes #3105.

With this flag, instead of running `rustup toolchain install <some-toolchain>` followed by `rustup override set <some-toolchain>`, both steps can be combined into a single command: `rustup toolchain install <some-toolchain> --override`.

Feedback on the wording of the doc comment is welcome.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -402,6 +402,10 @@ struct UpdateOpts {
     /// Install toolchains that require an emulator. See https://github.com/rust-lang/rustup/wiki/Non-host-toolchains
     #[arg(long)]
     force_non_host: bool,
+
+    /// Set the installed toolchain as the override for the current directory
+    #[arg(long)]
+    r#override: bool,
 }
 
 #[derive(Debug, Default, Args)]
@@ -946,6 +950,7 @@ async fn update(
     let self_update_mode = SelfUpdateMode::from_cfg(cfg)?;
     let should_self_update = !opts.no_self_update;
     let force_non_host = opts.force_non_host;
+    let set_override = opts.r#override;
     cfg.profile_override = opts.profile;
 
     let cfg = &cfg;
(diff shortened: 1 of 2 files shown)


djc (approved):

src/cli/rustup_mode.rs
```diff
@@ -402,6 +402,10 @@ struct UpdateOpts {
     /// Install toolchains that require an emulator. See https://github.com/rust-lang/rustup/wiki/Non-host-toolchains
     #[arg(long)]
     force_non_host: bool,
+
+    /// Set the installed toolchain as the current directory override
```
rami3l: Nit: Mirroring `rustup override set --help`:

```suggestion
    /// Set the installed toolchain as the override for the current directory
```

rami3l (approved):
LGTM :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4727/e143373287648906f1370b0db40c77c938b11398..64c7502e94cb589d6cf0e181e590d542025d4371)*

PR #4729, opened 2026-02-26
Implement dark mode for rustup.rs website

### LINKED ISSUE #4209: color-scheme support for rustup.rs

Visiting rustup.rs from a dark-theme device can be blinding as it is an almost entirely plain white page.

### Solution you'd like

A simple `@media (prefers-color-scheme: dark)` based dark-mode style. The main rust-lang.org site seems to lack one as well (but is tempered a bit by the big blocks of background color reducing the overall brightness), but https://blog.rust-lang.org/ has a theme that I assume is consistent with the rest of the project.

### Notes

_No response_

Bogay:
This PR introduces dark mode for rustup.rs website. For simplicity, I haven't added theme-switching logic in javascript. Instead, it relies solely on `@media
  (prefers-color-scheme: dark)` to override css variables for dark mode. The color palette is adapted from the blog's dark mode with a little modification (rust-lang/blog.rust-lang.org#1343).

However, I am not sure whther the I should adjust of the copy button. It currently uses the browser's default styles, which makes the "Copied!" feedback text slightly difficult to read in dark mode.

## Screenshots

before:

image

image

after:

image

image

Close #4209

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -226,7 +226,7 @@
 </p>
 
 <p id="about">
-  <img src="https://rust-lang.org/logos/rust-logo-blk.svg" alt="" />
+  <img class="rust-logo" src="https://rust-lang.org/logos/rust-logo-blk.svg" alt="" />
   rustup is an official Rust project.
   <br/>
   <a href="https://rust-lang.github.io/rustup/installation/other.html">other installation options</a>
--- a/www/rustup.css
+++ b/www/rustup.css
@@ -216,25 +216,19 @@ hr {
 #platform-instructions-win-arm64 button.copy-button,
 #platform-instructions-default button.copy-button,
 #platform-instructions-unknown button.copy-button {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    position: relative;
+    aspect-ratio: 51 / 58;
     height: 58px;
     margin: 0;
-    padding: 0 5px 0 5px;
     border-width: 0px;
     cursor: pointer;
 }
 
-#platform-instructions-unix div.copy-icon,
-#platform-instructions-win32 div.copy-icon,
-#platform-instructions-win64 div.copy-icon,
-#platform-instructions-win-arm64 div.copy-icon,
-#platform-instructions-default div.copy-icon,
-#platform-instructions-unknown div.copy-icon {
-    position: relative;
-    width: fit-content;
-    height: fit-content;
-    top: 27px;
-    left: 23px;
-    transform: translate(-50%, -50%);
+div.copy-icon {
+    transform: translateX(3px);
 }
 
 #platform-instructions-unix div.copy-button-text,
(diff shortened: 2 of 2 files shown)


rami3l (commented):
Thanks a lot! Not using JavaScript is also what I'd prefer with the problem you are trying to resolve. Please feel free to adjust the remaining elements with the light theme background :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4729/a373a2c8c4c0fc547dc200ec0f553b4a66ffdd3c..16b2f18cb9a7d0d419f40270f62579b65ad173da)*

rami3l (approved):

Bogay:
I've moved the "Copied!" text out of the button so that it displays over body background like other link text. I think this make it a little more readable. BTW, I found that `#platform-xxx-yyy` selectors in rustup.css seems to be unused. Should I remove them?

light:

image

dark:

image

rami3l:
> I've moved the "Copied!" text out of the button so that it displays over body background like other link text. I think this make it a little more readable.

@Bogay Good idea! Lemme have a test locally and get back to you later, please be patient 🙏 

> BTW, I found that `#platform-xxx-yyy` selectors in rustup.css seems to be unused. Should I remove them?

I believe they are not "dead code" per se but rather reserved for future use, see below for an usage example :)

https://github.com/rust-lang/rustup/blob/a373a2c8c4c0fc547dc200ec0f553b4a66ffdd3c/www/index.html#L45

... the visibility is controlled either by your host platform, or by typing `test` and clicking `n`.

rami3l (commented):
Can we make the copy button dark mode-aware as well? Something among the lines of...

```css
...
#platform-instructions-unknown button.copy-button {
    ...
    background-color: var(--code-shadow-color);
    color: var(--copy-button-text-color);
}

div.copy-icon {
    ...
}

div.copy-icon svg {
    fill: var(--body-color)
}
```

PS: Sorry if I misused your palette, I'm not super good at this 🙏 

Otherwise, LGTM!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4729/a373a2c8c4c0fc547dc200ec0f553b4a66ffdd3c..c7c6414ea547cc64553558de3ce7f2f73f73ff91)*

Bogay:
> Can we make the copy button dark mode-aware as well?

Sure, but it seems that setting background color remove the hover effect. I've added a simple hover effect on it. The icon become teal when hover. Let me know if you have other concerns.

light:

| | normal | hover |
|---|---|---|
| light | image | image |
| dark | image | image |

> I believe they are not "dead code" per se but rather reserved for future use, see below for an usage example :)

Thanks for the clarification! To clarify, I didn't intend to remove those IDs from `index.html`. My intention was to simplify the CSS selectors. Since the elements with the `platform-instruction-*` IDs already have the `instructions` class, we can just use that class to select them instead.

rami3l:
>  I've added a simple hover effect on it. The icon become teal when hover.

@Bogay Looks cool, thanks!

> My intention was to simplify the CSS selectors. Since the elements with the `platform-instruction-*` IDs already have the `instructions` class, we can just use that class to select them instead.

Sure, that's a good idea :)

Bogay:
I've updated the css selectors. Maybe it's ready to merge now?

Issue #3387, opened 2023-06-21
`rustup run` also searches PATH


jyn514:
### Problem

```
; rustup run 1.69 whoami
jyn
```
This seems ... undesirable. In particular, it means that `rustup which` and `rustup run` differ in behavior:
```
; rustup which whoami
error: not a file: '/Users/jyn/.local/lib/rustup/toolchains/1.69-aarch64-apple-darwin/bin/whoami'
```

As far as I can tell, the only difference is that `rustup run` sets RUSTUP_TOOLCHAIN and DYLD_FALLBACK_LIBRARY_PATH:
```
; rustup run nightly python3 -c $'import os \nfor k,v in os.environ.items(): print(k, v)'  | sort -u | diff - <(python3 -c $'import os \nfor k,v in os.environ.items(): print(k, v)' | sort -u)
8d7
< DYLD_FALLBACK_LIBRARY_PATH /Users/jyn/.local/lib/rustup/toolchains/nightly-aarch64-apple-darwin/lib:/Users/jyn/lib:/usr/local/lib:/usr/lib
29,30d27
< RUSTUP_TOOLCHAIN nightly-aarch64-apple-darwin
< RUST_RECURSION_COUNT 1
49c46
< _ /Users/jyn/.local/lib/cargo/bin/rustup
---
> _ /Applications/Xcode.app/Contents/Developer/usr/bin/python3
```

<details><summary>my original use case, although it's pretty complicated</summary>

i am writing a new rustc_driver tool that goes in the sysroot. i want to be able to run it as `cargo +nightly foo` instead of `PATH=$PATH:$(rustc +nightly --print sysroot)/bin cargo +nightly foo`. to that end, i've put a shell script in ~/.cargo/bin/cargo-foo that emulates a rustup proxy:

```
#!/bin/sh
me=$(basename $0)
exec rustup run "$RUSTUP_TOOLCHAIN" $me "$@"
```

This works fine if the tool is actually present in the sysroot. However, if it's not present, rustup falls back to path and *re-executes this same script*, eventually leading to a recursion error that it's nested too deeply. I would like to instead give a hard error. Today, i have to workaround rustup's behavior with more calls:

```
# Make sure this is actually installed for the given toolchain. `rustup run` falls back to PATH,
# which will recursively invoke this script; that's not what we want.
if ! rustup which $me --toolchain $RUSTUP_TOOLCHAIN 2>/dev/null; then
[...]

PR #4741, opened 2026-03-05
feat(toolchain): include err in rustc_version output

### LINKED ISSUE #3607: The `error reading rustc version` message could be improved
### Problem

M1 macOS 12.4.1

```
> rustup update
info: syncing channel updates for 'stable-aarch64-apple-darwin'
info: syncing channel updates for 'nightly-aarch64-apple-darwin'

      stable-aarch64-apple-darwin unchanged - rustc 1.75.0 (82e1608df 2023-12-21)
  stable-x86_64-unknown-linux-gnu unchanged - (error reading rustc version)
     nightly-aarch64-apple-darwin unchanged - rustc 1.77.0-nightly (89e2160c4 2023-12-27)
```

### Steps

rustup update with stable-x86_64-unknown-linux-gnu toolchain

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
info: The currently active `rustc` version is `rustc 1.75.0 (82e1608df 2023-12-21)`
```
```


### Installed toolchains

```console
rustup home:  /Users/foobar/.rustup

installed toolchains
--------------------

nightly-aarch64-apple-darwin
1.60.0-aarch64-apple-darwin
1.61.0-aarch64-apple-darwin
1.65.0-aarch64-apple-darwin
1.70.0-aarch64-apple-darwin

aarch64-apple-darwin
aarch64-unknown-linux-gnu
x86_64-pc-windows-msvc

active toolchain
----------------

rustc 1.75.0 (82e1608df 2023-12-21)
```

Bogay:
This PR includes error details in `Toolchain::rustc_version` so that it's more clear to end-user what happened.

Close #3607

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -284,6 +284,7 @@ impl<'a> Toolchain<'a> {
     #[tracing::instrument(level = "trace")]
     pub fn rustc_version(&self) -> String {
         match self.create_command("rustc") {
+            Err(e) => format!("(rustc does not exist: {e})"),
             Ok(mut cmd) => {
                 cmd.arg("--version");
                 cmd.stdin(Stdio::null());
@@ -291,40 +292,41 @@ impl<'a> Toolchain<'a> {
                 cmd.stderr(Stdio::piped());
                 self.set_ldpath(&mut cmd);
 
+                let mut child = match cmd.spawn() {
+                    Err(e) => return format!("(error reading rustc version: {e})"),
+                    Ok(child) => child,
+                };
                 // some toolchains are faulty with some combinations of platforms and
                 // may fail to launch but also to timely terminate.
                 // (known cases include Rust 1.3.0 through 1.10.0 in recent macOS Sierra.)
                 // we guard against such cases by enforcing a reasonable timeout to read.
-                let mut line1 = None;
-                if let Ok(mut child) = cmd.spawn() {
-                    let timeout = Duration::new(10, 0);
-                    match child.wait_timeout(timeout) {
-                        Ok(Some(status)) if status.success() => {
-                            let out = child
-                                .stdout
-                                .expect("Child::stdout requested but not present");
-                            let mut line = String::new();
-                            if BufReader::new(out).read_line(&mut line).is_ok() {
-                                let lineend = line.trim_end_matches(&['\r', '\n'][..]).len();
-                                line.truncate(lineend);
-                                line1 = Some(line);
-                            }
-                        }
-                        Ok(None) => {
-                            let _ = child.kill();
-                            return String::from("(timeout reading rustc version)");
-                        }
-                        Ok(Some(_)) | Err(_) => {}
+                let timeout = Duration::new(10, 0);
+                let status = match child.wait_timeout(timeout) {
+                    Ok(None) => {
+                        let _ = child.kill();
+                        return String::from("(timeout reading rustc version)");
                     }
+                    Err(e) => return format!("(error reading rustc version: {e})"),
+                    Ok(Some(status)) => status,
+                };
+
+                if !status.success() {
+                    return format!("(error reading rustc version: {status})");
                 }
 
-                if let Some(line1) = line1 {
-                    line1
-                } else {
-                    String::from("(error reading rustc version)")
+                let out = child
+                    .stdout
+                    .expect("Child::stdout requested but not present");
+                let mut line = String::new();
+                match BufReader::new(out).read_line(&mut line) {
+                    Ok(_) => {
+                        let lineend = line.trim_end_matches(&['\r', '\n'][..]).len();
+                        line.truncate(lineend);
+                        line
+                    }
+                    Err(e) => format!("(error reading rustc version: {e})"),
                 }
             }
-            Err(_) => String::from("(rustc does not exist)"),
         }
     }
 
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -1392,6 +1392,34 @@ nightly-[HOST_TRIPLE] (default)
 }
 
 #[tokio::test]
+async fn show_active_toolchain_rustc_missing() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+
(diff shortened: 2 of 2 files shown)


src/toolchain.rs
```diff
@@ -284,47 +284,55 @@ impl<'a> Toolchain<'a> {
...
-                        Ok(Some(status)) if status.success() => {
-                            let out = child
-                                .stdout
-                                .expect("Child::stdout requested but not present");
-                            let mut line = String::new();
-                            if BufReader::new(out).read_line(&mut line).is_ok() {
-                                let lineend = line.trim_end_matches(&['\r', '\n'][..]).len();
-                                line.truncate(lineend);
-                                line1 = Some(line);
+                match cmd.spawn() {
```
rami3l: Nit: could try to reduce some rightward drift here and in sections below. Something along the lines of:

```rs
let mut child =  match cmd.spawn() {
    Err(e) => return format!("(error reading rustc version: {e})"),
    Ok(child) => child,
};
[..]
```

src/toolchain.rs
```diff
@@ -284,47 +284,55 @@ impl<'a> Toolchain<'a> {
     #[tracing::instrument(level = "trace")]
```
rami3l, line 284: Suggest adding some snapshot tests to better demonstrate this new feature.

rami3l (approved):
Thanks for working on this!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4741/28d1352dbcb436d3111c3594b9e1588e94950464..b8c00f3fcb9f3a06f8aa73d28ac04f36d9b1d2ca)*

src/toolchain.rs
```diff
@@ -284,47 +284,49 @@ impl<'a> Toolchain<'a> {
...
+                    return format!("(error reading rustc version: {status})");
                 }
 
-                if let Some(line1) = line1 {
-                    line1
-                } else {
-                    String::from("(error reading rustc version)")
+                let out = child
+                    .stdout
+                    .expect("Child::stdout requested but not present");
```
Bogay, line 319: Should we return a string instead of panic here? The doc of `rustc_version` says it's infallible function

tests/suite/cli_rustup.rs
```diff
@@ -1391,6 +1391,63 @@ nightly-[HOST_TRIPLE] (default)
...
+    cx.config
+        .expect(&["rustup", "show", "active-toolchain", "--verbose"])
+        .await
+        .with_stdout(snapbox::str![
+            r#"
+...
+compiler: (rustc does not exist: [..])
+...
+        "#
+        ])
```
Bogay, line 1418: This test failed ([example job](https://github.com/rust-lang/rustup/actions/runs/23110377064/job/67126707466?pr=4741)) because it the `Toolchain::create_command` in `Toolchain::rustc_version` does not return `Err` even `rustc` doesn't exist in toolchain. AFAIK it fallback to the `rustc` in `PATH`, which I believe related to #3387 . It's also hard to produce a env that don't have `rustc` in `PATH` because [mock_bin](https://github.com/rust-lang/rustup/blob/78fa26e4e8c1e085762e613d354b5129e24014e4/src/test/clitools.rs#L1204-L1215) use `rustc` to compile the binary. (If we can restrict `rustup` proxy to not use `rustc` inside `PATH`, I think that would be easier.)

tests/suite/cli_rustup.rs
```diff
@@ -1391,6 +1391,63 @@ nightly-[HOST_TRIPLE] (default)
...
+        .is_err();
+}
+
+#[cfg(unix)]
```
Bogay, line 1422: I use shell script to mock the exit code of `rustc` in toolchain, which I think can only be executed on unix. Not sure whether it's acceptable in rustup test suite. :thinking:

Bogay (commented):
I've added some testcases. But I found some new issues I didn't aware of need to be discussed.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4741/78fa26e4e8c1e085762e613d354b5129e24014e4..44ec1341bb606c0f3d8cea8ce18e8f012e7254fa)*

rami3l:
Sorry for getting sidetracked for a while... I'll find some time to give it another look.

src/toolchain.rs
rami3l: No. If the function panics here it means the `Command` setup is wrong and this is a programming error, and we should expose the error instead of covering it up.

tests/suite/cli_rustup.rs
rami3l: Thanks for asking! Actually we have `mock_bin_src.rs` which you can use instead of this dirty hack. You can modify it to make the thing crash in a cross-platform and deterministic way. IIRC the `rustc` we call in other tests is actually this instead of the real `rustc`.

Bogay: Yes. In other tests `mock_bin_src.rs` is used for mock `rustc`. However, I don't know how to control args passed to `rustc` in this testcase. Maybe we can control the exit code using env var? But that's a new pattern which is not used in `mock_bin_src.rs`.

rami3l: @Bogay I do think you can try passing an env variable instead. The rustup proxy should pass the env variables down to the mock binary. Just make sure the env var is not super commonly-named to avoid side effects.

Alternatively you can compile a rust binary yourself which is independent from `mock_bin_src.rs` if that turns out easier for you.

tests/suite/cli_rustup.rs
rami3l: @Bogay Give this a try:

```rs
fs::remove_file(path.join("bin/rustc")).unwrap();
cx.config
  .expect_with_env(&["rustup", "show", "active-toolchain", "--verbose"], [("PATH", ""), ("RUST_RECURSION_COUNT", "1023")])
[..]
```

The `RUST_RECURSION_COUNT` is explicitly large to make sure that `create_command()` bails out.

OTOH if the error string spans multiple lines we might want to capture the first line only.

PR #4752, opened 2026-03-15
Run a parallel pre-check before updating toolchains

### LINKED ISSUE #4747: Improve `rustup update`s performance by starting with a parallel check
As discussed [here](https://rust-lang.zulipchat.com/#narrow/channel/490103-t-rustup/topic/.60rustup.20update.60.20performance/with/578967273), the performance of `rustup update` could be improved compared to `rustup check`.

When no updates are available, `rustup update` is about 3x slower than `rustup check`.
This happens because the channels are checked sequentially for updates, while `rustup check` performs these checks concurrently.

A possible improvement would be to perform a parallel check before starting the update process itself; allowing the performance of a no-op `rustup update` to match that of `rustup check`.

FranciscoTGouveia:
Fixes #4747.

Instead of sequentially checking each toolchain for updates, `rustup update` now first checks all toolchains concurrently and only then performs the updates (if needed).

This brings two advantages:
1. A no-op `rustup update` is now ~**2x** faster (see below);
2. After updating a toolchain, there is no need to wait for the remaining toolchains to be checked for updates; this is especially helpful in cases where only `nightly` needs updating.

Benchmarks were run using `hyperfine` (20 iterations on a 50 Mbps connection) on a setup with 11 toolchains, and no regressions were observed.
For the no-op `rustup update`, this patch was **2.2x** faster; for cases where a single toolchain was updated, there was no measurable difference.

CC @epage

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -10,6 +10,7 @@ use std::{cmp, env};
 use anstyle::Style;
 use anyhow::{Context, Result, anyhow};
 use clap_cargo::style::{CONTEXT, ERROR, UPDATE_ADDED, UPDATE_UNCHANGED, UPDATE_UPGRADED};
+use futures_util::future::join_all;
 use git_testament::{git_testament, render_testament};
 use tracing::{error, info, warn};
 use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
@@ -213,11 +214,26 @@ fn show_channel_updates(
 
 pub(crate) async fn update_all_channels(cfg: &Cfg<'_>, force_update: bool) -> Result<ExitCode> {
     let profile = cfg.get_profile()?;
+    let channels = cfg.list_channels()?;
+
+    // Run a pre-check to determine which channels have updates available.
+    let channels_with_manifests = join_all(channels.into_iter().map(|(desc, d)| async move {
+        // Prefetch errors are intentionally discarded; a failed prefetch falls
+        // back to the normal update path, where errors are properly propagated.
+        let manifest = d.fetch_dist_manifest().await.unwrap_or(None);
+        (desc, d, manifest)
+    }))
+    .await;
+
     let mut toolchains = Vec::new();
-    for (desc, distributable) in cfg.list_channels()? {
-        let options = DistOptions::new(&[], &[], &desc, profile, force_update, cfg)?
-            .for_update(&distributable, false);
-        let result = InstallMethod::Dist(options).install(None).await;
+    for (desc, distributable, manifest) in channels_with_manifests {
+        let result = if force_update || manifest.is_some() {
+            let options = DistOptions::new(&[], &[], &desc, profile, force_update, cfg)?
+                .for_update(&distributable, false);
+            InstallMethod::Dist(options).install(manifest).await
+        } else {
+            Ok(UpdateStatus::Unchanged)
+        };
 
         if let Err(e) = &result {
             error!("{e}");
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -217,7 +217,7 @@ pub(crate) async fn update_all_channels(cfg: &Cfg<'_>, force_update: bool) -> Re
     for (desc, distributable) in cfg.list_channels()? {
         let options = DistOptions::new(&[], &[], &desc, profile, force_update, cfg)?
             .for_update(&distributable, false);
-        let result = InstallMethod::Dist(options).install().await;
+        let result = InstallMethod::Dist(options).install(None).await;
 
         if let Err(e) = &result {
             error!("{e}");
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -28,6 +28,12 @@ use crate::{
     toolchain::DistributableToolchain,
 };
 
+/// A download manifest paired with its respective hash, as returned by the dist server.
+pub struct ManifestWithHash {
+    pub manifest: Manifest,
+    pub hash: String,
+}
+
 /// Used by the `installed_components` function
 pub(crate) struct ComponentStatus {
     pub component: Component,
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -955,7 +955,11 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
     //
     // Returns the manifest's hash if anything changed.
     #[tracing::instrument(level = "trace", err(level = "trace"), skip_all, fields(profile = ?self.profile, prefix = %prefix.path().display()))]
-    pub(crate) async fn install_into(&self, prefix: &InstallPrefix) -> Result<Option<String>> {
+    pub(crate) async fn install_into(
+        &self,
+        prefix: &InstallPrefix,
+        manifest: Option<ManifestWithHash>,
+    ) -> Result<Option<String>> {
         let fresh_install = !prefix.path().exists();
         // fresh_install means the toolchain isn't present, but hash_exists means there is a stray hash file
         if fresh_install && self.update_hash.exists() {
@@ -1013,6 +1017,7 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
         };
 
         let mut toolchain = self.toolchain.clone();
+        let mut prefetched_manifest = manifest;
         let res = loop {
             let result = try_update_from_dist_(
                 &self.dl_cfg,
@@ -1028,6 +1033,7 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
                 self.targets,
                 &mut fetched,
                 self.cfg,
+                prefetched_manifest.take(),
             )
             .await;
 
@@ -1126,27 +1132,32 @@ async fn try_update_from_dist_(
     targets: &[&str],
     fetched: &mut String,
     cfg: &Cfg<'_>,
+    prefetched_manifest: Option<ManifestWithHash>,
 ) -> Result<Option<String>> {
     let toolchain_str = toolchain.to_string();
     let manifestation = Manifestation::open(prefix.clone(), toolchain.target.clone())?;
 
     // TODO: Add a notification about which manifest version is going to be used
     info!("syncing channel updates for {toolchain_str}");
-    match download
-        .dl_v2_manifest(
-            // Even if manifest has not changed, we must continue to install requested components.
-            // So if components or targets is not empty, we skip passing `update_hash` so that
-            // we essentially degenerate to `rustup component add` / `rustup target add`
-            if components.is_empty() && targets.is_empty() {
-                Some(update_hash)
-            } else {
-                None
-            },
-            toolchain,
-            cfg,
-        )
-        .await
-    {
+    let manifest_result = if prefetched_manifest.is_some() {
+        Ok(prefetched_manifest)
+    } else {
+        download
+            .dl_v2_manifest(
+                // Even if manifest has not changed, we must continue to install requested components.
+                // So if components or targets is not empty, we skip passing `update_hash` so that
+                // we essentially degenerate to `rustup component add` / `rustup target add`
+                if components.is_empty() && targets.is_empty() {
+                    Some(update_hash)
+                } else {
+                    None
+                },
+                toolchain,
+                cfg,
+            )
+            .await
+    };
+    match manifest_result {
         Ok(Some(ManifestWithHash { manifest: m, hash })) => {
             match m.get_rust_version() {
                 Ok(version) => info!("latest update on {} for version {version}", m.date),
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -34,7 +34,7 @@ pub mod download;
 use download::DownloadCfg;
 
 pub mod manifest;
-use manifest::{Component, Manifest as ManifestV2};
+use manifest::{Component, Manifest as ManifestV2, ManifestWithHash};
 
 pub mod manifestation;
 use manifestation::{Changes, Manifestation, UpdateStatus};
@@ -1147,7 +1147,7 @@ async fn try_update_from_dist_(
         )
         .await
     {
-        Ok(Some((m, hash))) => {
+        Ok(Some(ManifestWithHash { manifest: m, hash })) => {
             match m.get_rust_version() {
                 Ok(version) => info!("latest update on {} for version {version}", m.date),
                 Err(_) => info!("latest update on {}", m.date),
--- a/src/install.rs
+++ b/src/install.rs
@@ -7,7 +7,7 @@ use tracing::debug;
 
 use crate::{
     config::Cfg,
-    dist::{DistOptions, prefix::InstallPrefix},
+    dist::{DistOptions, manifest::ManifestWithHash, prefix::InstallPrefix},
     errors::RustupError,
     toolchain::{CustomToolchainName, LocalToolchainName, Toolchain},
     utils,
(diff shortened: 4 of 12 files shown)


src/cli/common.rs
```diff
@@ -10,6 +10,7 @@ use std::{cmp, env};
 use anstyle::Style;
 use anyhow::{Context, Result, anyhow};
 use clap_cargo::style::{CONTEXT, ERROR, UPDATE_ADDED, UPDATE_UNCHANGED, UPDATE_UPGRADED};
+use futures_util::future;
```
FranciscoTGouveia: Is it preferable to have this import or to simply call `futures_util::future::join_all()` below?

src/cli/common.rs
```diff
@@ -213,11 +214,29 @@ fn show_channel_updates(
...
+    // This step is skipped when force-updating.
+    let needs_update = if force_update {
+        vec![true; channels.len()]
+    } else {
+        future::join_all(channels.iter().map(|(_, d)| d.show_dist_version()))
+            .await
+            .into_iter()
+            .map(|v| v.map_or(true, |opt| opt.is_some()))
+            .collect()
+    };
```
FranciscoTGouveia: Is there a cleaner way to achieve this?

src/cli/common.rs
```diff
@@ -213,11 +214,29 @@ fn show_channel_updates(
 
 pub(crate) async fn update_all_channels(cfg: &Cfg<'_>, force_update: bool) -> Result<ExitCode> {
     let profile = cfg.get_profile()?;
+    let channels = cfg.list_channels()?;
+
+    // Run a pre-check to determine which channels have updates available.
+    // This step is skipped when force-updating.
```
FranciscoTGouveia: Feedback on the wording of the comment is welcome :)

src/cli/common.rs
rami3l: I'd say just `use futures_util::future::join_all` directly :)

src/cli/common.rs
rami3l: Am I reading it right that now for every toolchain that needs to be updated, the update is checked twice?

FranciscoTGouveia: Yes, I forgot to add a comment about this, sorry.

I see two possible ways to proceed:
1. Accept the cost of performing a double check. In practice, since the first check is done concurrently in a batch, the overhead should be relatively small.
2. Do a small refactor, starting in `DistOptions::install_into()`, to allow skipping the second check.

rami3l: @FranciscoTGouveia I think we should try to avoid double checking if the changes involved are not super disruptive 🙏

src/cli/common.rs
```diff
@@ -213,11 +214,29 @@ fn show_channel_updates(
...
+    let needs_update = if force_update {
+        vec![true; channels.len()]
+    } else {
+        join_all(channels.iter().map(|(_, d)| d.show_dist_version()))
+            .await
+            .into_iter()
+            .map(|v| v.map_or(true, |opt| opt.is_some()))
```
djc: I'm confused by this logic. `show_dist_version()` seems to merely yield the latest available version, but that a `DistributableToolchain` has a manifest with a `rust` component with a version `is_some()` doesn't mean that it's newer than the installed version?

FranciscoTGouveia: Sorry, I do not think I fully understood your comment.
The rationale for using `show_dist_version()` comes from how it's used in `rustup check`.
As I understand it, `show_dist_version()` only returns `Some` when the hashes do not match (i.e., there is an update available), which aligns with what we need here.

In this setup, for toolchains that do not require updating, we would never reach [`.get_rust_version()`](https://github.com/rust-lang/rustup/blob/4b2c0919cfd11db5f79695509a71b217a88bac68/src/toolchain/distributable.rs#L457), since `.dl_v2_manifest()` returned `None`.

Apologies if I've misunderstood or overlooked anything.

djc: > As I understand it, `show_dist_version()` only returns `Some` when the hashes do not match (i.e., there is an update available), which aligns with what we need here.

You're completely right. Might be good to add a docstring for `show_dist_version()` and maybe consider renaming it.

FranciscoTGouveia: Sounds good, I will take care of it. :)

src/cli/common.rs
FranciscoTGouveia: I have been thinking about this, and my initial idea was to add a new `manifest` field (of type `ManifestV2`) to the `DistOptions` struct; however, I believe that would change the semantics quite a bit, since it would store more than just "options".

A direction I am more inclined toward is to reuse the manifest returned by `show_dist_version()`, which is currently discarded since we only call `.is_some()` on it.
Pairing this with a potential new `install_with_manifest()` method in `InstallMethod` (alongside the existing `install()`) would allow us to skip re-downloading the manifest.
What do you think?

djc: > A direction I am more inclined toward is to reuse the manifest returned by `show_dist_version()`, which is currently discarded since we only call `.is_some()` on it.
> Pairing this with a potential new `install_with_manifest()` method in `InstallMethod` (alongside the existing `install()`) would allow us to skip re-downloading the manifest.
> What do you think?

Sounds pretty good to me.

rami3l: > A direction I am more inclined toward is to reuse the manifest returned by `show_dist_version()`, which is currently discarded since we only call `.is_some()` on it. Pairing this with a potential new `install_with_manifest()` method in `InstallMethod` (alongside the existing `install()`) would allow us to skip re-downloading the manifest. What do you think?

@FranciscoTGouveia Yeah I agree with that. It'd be nice if the old installation method can be refactored to use the newly-introduced one by fetching the manifests before the call. I imagine there shouldn't be a lot of dependency issues in doing so?

src/cli/common.rs
rami3l: I think you are not skipping anything anymore.

FranciscoTGouveia: Yes, I have already removed it.

src/dist/mod.rs
```diff
@@ -1028,6 +1033,7 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
                 self.targets,
                 &mut fetched,
                 self.cfg,
+                prefetched_manifest.take(),
```
rami3l, line 1036: Question: How is this different from just passing `manifest` by move?

src/cli/common.rs
rami3l: The `Option<>` semantics still remain unclear. Maybe try `updated_dist_{manifest,version}()`?

rami3l (approved):

src/cli/common.rs
FranciscoTGouveia: Apologies, could you clarify where you were suggesting putting `updated_dist_manifest()`?
I have renamed it to `fetch_dist_manifest`, which I believe is more accurate than the previous `show_dist_version()`.

For reference, [this](https://github.com/rust-lang/rustup/pull/4752/changes#diff-e510b461aeea5b0176806131110b7eea4c53e50067a41bc06382c42b70da6493R224) is the updated code, where `.is_some()` has been replaced with `.unwrap_or(None)`.

src/dist/mod.rs
FranciscoTGouveia: Since the call to `try_update_from_dist` is inside a loop, moving the value directly would only work for the first iteration. By using `.take()`, I’m not moving the `Option` itself, but instead passing its contents (`Some` in the first iteration and `None` in the subsequent ones).

src/toolchain/distributable.rs
```diff
@@ -445,15 +445,18 @@ impl<'a> DistributableToolchain<'a> {
...
                 Some(&self.toolchain.cfg.get_hash_file(&self.desc, false)?),
                 &self.desc,
                 self.toolchain.cfg,
             )
-            .await?
-        {
+            .await
+    }
+
+    pub async fn show_dist_version(&self) -> anyhow::Result<Option<String>> {
```
djc: Since this has a single caller, maybe this should be inlined into its caller instead?

(If not, should do top-down order such that `show_dist_version()` comes before `fetch_dist_manifest()`.)

src/install.rs
```diff
@@ -38,6 +38,13 @@ impl InstallMethod<'_, '_> {
     // Install a toolchain
     #[tracing::instrument(level = "trace", err(level = "trace"), skip_all)]
     pub(crate) async fn install(self) -> Result<UpdateStatus> {
+        self.install_with_manifest(None).await
+    }
+
+    pub(crate) async fn install_with_manifest(
```
djc: Suggest just adding the argument to `install()` instead of adding a wrapper, since calling it with `None` is fairly trivial.

src/dist/mod.rs
```diff
@@ -955,7 +955,11 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
     //
     // Returns the manifest's hash if anything changed.
     #[tracing::instrument(level = "trace", err(level = "trace"), skip_all, fields(profile = ?self.profile, prefix = %prefix.path().display()))]
-    pub(crate) async fn install_into(&self, prefix: &InstallPrefix) -> Result<Option<String>> {
+    pub(crate) async fn install_into(
+        &self,
+        prefix: &InstallPrefix,
+        manifest: Option<(ManifestV2, String)>,
```
djc: I think it's fairly unclear with all the places you're passing around `(ManifestV2, String)` (or `(Manifest, String)`) what the relation is between those or what the `String` contains. Perhaps it's worth wrapping these in `struct` type with named fields?

src/cli/common.rs
```diff
@@ -213,11 +214,27 @@ fn show_channel_updates(
...
     let profile = cfg.get_profile()?;
+    let channels = cfg.list_channels()?;
+
+    // Run a pre-check to determine which channels have updates available.
+    let prefetched_manifests: Vec<Option<(Manifest, String)>> =
+        join_all(channels.iter().map(|(_, d)| d.fetch_dist_manifest()))
+            .await
+            .into_iter()
+            .map(|r| r.unwrap_or(None))
+            .collect();
```
djc: So this is sort of just silencing/ignoring errors? Is that what we did before?

Instead of the type annotation, I would use turbofishing and leaving any inner types implied, as in `collect::<Vec<_>>()`.

src/dist/mod.rs
```diff
@@ -1126,27 +1132,32 @@ async fn try_update_from_dist_(
     targets: &[&str],
     fetched: &mut String,
     cfg: &Cfg<'_>,
+    prefetched_manifest: Option<(ManifestV2, String)>,
```
djc: Does this actually need to take ownership of the manifest, or could it make do with a reference?

src/dist/mod.rs
rami3l: Ah okay, it's a diff problem on my side. Clear!

src/cli/common.rs
rami3l: @FranciscoTGouveia Sorry for not being super clear. I was only referring to the renames. `show_dist_manifest()` may be replaced by `updated_dist_manifest()` since it will only return a value when there is actually an updated dist manifest, right?

For `show_dist_version()` I think it'd be better to just inline it as per https://github.com/rust-lang/rustup/pull/4752#discussion_r2953395714.

src/dist/mod.rs
rami3l: @djc It is worth noting that this is the return value of `.dl_v2_manifest()`. OTOH maybe it _is_ worth it to extract that in a separate commit though.

FranciscoTGouveia: Agreed, would `ManifestWithHash` be a good name for the struct?

src/cli/common.rs
FranciscoTGouveia: Yes, errors are intentionally discarded.
A failed prefetch falls back to the normal update path, where errors are properly propagated.

src/dist/mod.rs
FranciscoTGouveia: I believe it needs to take ownership, as the manifest is then moved into [`manifestation.update()`](https://github.com/rust-lang/rustup/blob/6f8931c636cb20317d85223ac72bfd44f58f5e89/src/dist/mod.rs#L1202) which takes the `Manifest` [by value](https://github.com/rust-lang/rustup/blob/6f8931c636cb20317d85223ac72bfd44f58f5e89/src/dist/manifestation.rs#L117-L125).

src/cli/common.rs
```diff
@@ -217,7 +217,7 @@ pub(crate) async fn update_all_channels(cfg: &Cfg<'_>, force_update: bool) -> Re
     for (desc, distributable) in cfg.list_channels()? {
         let options = DistOptions::new(&[], &[], &desc, profile, force_update, cfg)?
             .for_update(&distributable, false);
-        let result = InstallMethod::Dist(options).install().await;
+        let result = InstallMethod::Dist(options).install(None).await;
```
djc: The commit message for this commit is no longer accurate.

src/dist/mod.rs
djc: At this stage, would it make more sense to make this argument non-optional? In the backtracking loop in this function's caller, could we reuse the same manifest across iterations? (Maybe this should be in yet another commit?)

src/cli/common.rs
```diff
@@ -213,11 +214,24 @@ fn show_channel_updates(
...
+        .into_iter()
+        .map(|r| r.unwrap_or(None))
+        .collect::<Vec<_>>();
+
     let mut toolchains = Vec::new();
-    for (desc, distributable) in cfg.list_channels()? {
-        let options = DistOptions::new(&[], &[], &desc, profile, force_update, cfg)?
-            .for_update(&distributable, false);
-        let result = InstallMethod::Dist(options).install(None).await;
+    for ((desc, distributable), manifest) in channels.into_iter().zip(prefetched_manifests) {
```
djc: The `zip()` here effectively repeats the earlier loop over `channels`. Maybe it would be better to wrap `d.fetch_dist_manifest()` in an `async move` block that can keep the channel and manifest together?

src/dist/manifest.rs
```diff
@@ -28,6 +28,12 @@ use crate::{
     toolchain::DistributableToolchain,
 };
 
+/// A download manifest paired with its respective hash, as returned by the dist server.
+pub struct ManifestWithHash {
```
djc, line 32: Nit: I'd introduce this earlier in the commit history, before further shenanigans propagate its use.

src/dist/mod.rs
rami3l: @djc The argument should stay optional IMHO.

My understanding is that this PR is doing a first-run optimization. When we are trying to backtrack to earlier nightlies, there is NO room for us to reuse the same manifest because the contents of those toolchains will be different from the prefetched ones :)

And no, I don't think we should prefetch all the possible backtracks since this is a very rare case. Fetching on demand like we do right now would be perfectly fine. (Of course we can fetch all possible backtracks once they are proven needed and choose the latest valid one, but the improvements would be marginal and this is slightly out of the scope of this patch.)

PS: This is also why we have `let mut toolchain`: in every iteration of the loop the toolchain is actually different because the contents are different.

src/cli/common.rs
rami3l: @djc Considering the optimization-only nature of this change (see also https://github.com/rust-lang/rustup/pull/4752#discussion_r2996108216) I think it'd be okay to keep it that way. Although @FranciscoTGouveia it'd be nice if we can add a short comment here since it deviates from what is usually done WRT error handling in this project.

src/dist/mod.rs
djc: Fair enough.

rami3l (approved):
LGTM, if @djc is okay with this I think we are ready to go. Thanks a lot :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4752/5c2c7cd8344070a356162c249105f6ff6b2df77b..7f407cf4f76537089d3d99636ef3000d46b72dc6)*

PR #4756, opened 2026-03-16
fix(rustup-init/sh): prevent passing `--default-host` twice

### LINKED ISSUE #4755: rustup-init: --default-host is broken under cygwin
### Verification

### Problem

Attempting to pass a value for `--default-host` to rustup-init.sh will fail with the following error:
```
error: error: the argument '--default-host <DEFAULT_HOST>' cannot be used multiple times
```
It looks like #4497 is the cause; that will prepend a `--default-host=$_arch` unconditionally when running on a windows host.

### Steps

On a cygwin host:
```
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-host=x86_64-pc-windows-msvc
```

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
rustup-init 1.29.0 (28d1352db 2026-03-05)
```

### Installed toolchains

```console
n/a
```

### OS version

```console
CYGWIN_NT-10.0-20348 3.6.6-1.x86_64 x86_64 Cygwin
```

rami3l:
Follow-up of https://github.com/rust-lang/rustup/pull/4497, closes #4755.

My apologies for having misled you @cachebag on this one (I mistakenly thought rustup followed the GNU coreutils conventions, but should we adopt just that in rustup the binary instead?) 🙇 

@abr-egn Please feel free to test the patch locally if you have time 🙏

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -174,6 +174,17 @@ main() {
         exit 1
     fi
 
+    # If a default host has been specified on the command line, we should
+    # revert our own override before calling the installer.
+    for arg in "$@"; do
+        case "$arg" in
+            --default-host|--default-host=*)
+                _default_host_override=
+                break
+                ;;
+        esac
+    done
+
     if [ "$need_tty" = "yes" ] && [ ! -t 0 ]; then
         # The installer is going to want to ask for confirmation by
         # reading stdin.  This script was piped into `sh` though and


ChrisDenton (approved):
Changes look good to me.

I think an argument could be made that rustup should just use the last value passed in. I don't think this is clear cut though and could do with some more discussion. For example, `rustc` only allows one `--target` (is that a mistake? is `--default-host` different? I don't know).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4756/f8c2da5bb70af7d4820535f38214cffa06414d40..2e6338dfd28cdd7b6ac64b643aed5a96ead4a8c5)*

rami3l:
> I think an argument could be made that rustup should just use the last value passed in.

@ChrisDenton Yeah that's how I did the review 😅 clearly confused rustup with something else; but we could easily revert this if we are going that way.

I'd say I'll merge upon @abr-egn's approval.

abr-egn (approved):
This does indeed fix it, thanks for the quick update!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4756/f8c2da5bb70af7d4820535f38214cffa06414d40..2e6338dfd28cdd7b6ac64b643aed5a96ead4a8c5)*

cachebag:
> should we adopt just that in rustup the binary instead?

@rami3l most CLI tools follow this convention of "last flag wins", as @ChrisDenton notes (e.g. curl, docker, cargo, etc.). 

we could just take a 'belt and suspenders' approach to this; `#[arg(long, overrides_with = "default_host")]` in the `RustupInit` struct and then your fix here is perfect- now the script is covered and so is the binary. I can open a PR for this if you agree.

> For example, rustc only allows one --target (is that a mistake? is --default-host different? I don't know).

@ChrisDenton 
It's my understanding that `rustc` disallowing multiple `--target` is arguably a design choice specific to compilation, where "which target?" is a critical single-answer question. I feel that `--default-host` in rustup is a configuration preference, and "last wins" is more user-friendly IMHO.

rami3l:
@cachebag I also like the said behavior, but thinking about it I retreated from my original stance (at least for now) because:

- If it's not clap's default it's possibly not widely used across the official projects;
- The GNU coreutils convention is more friendly to script authors and not necessarily CLI users, whereas rustup should focus mainly on the latter;
- If we are to adopt the conventions it'll be better to do it consistently across all named args.

Issue #4532, opened 2025-10-13
cargo bash-completion ignores `+toolchain` arg


Rudxain:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

If I type `cargo +nightly run -- `<kbd>TAB</kbd> in a project with MSRV, it starts downloading the MSRV, instead of using nightly as I explicitly specified

### Steps

See summary

### Rustup version

```console
rustup 1.28.2 (e4f3ad6f8 2025-04-28)
```

### Installed toolchains

```console
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/rudxain/.rustup

installed toolchains
--------------------
nightly-x86_64-unknown-linux-gnu (active, default)
1.82.0-x86_64-unknown-linux-gnu

active toolchain
----------------
name: nightly-x86_64-unknown-linux-gnu
active because: it's the default toolchain
installed targets:
  x86_64-unknown-linux-gnu
```

### OS version

```console
Debian Testing
```

djc:
By MSRV, do you mean, a `rust-toolchain.toml` file?

rami3l:
Thanks for filing this issue!

Unfortunately, this is a known problem: due to a limitation in `clap`, it is completely unaware of the `+toolchain` syntax out of the box.

Recently there seems to have been a way to teach `clap` of this fact, but it would require some manual analysis and implementation to work correctly. Once https://github.com/rust-lang/rustup/issues/2268 is fixed, this should be solved at the same time.

As such, I'm closing this issue as a duplicate of #2268.

rami3l:
Interestingly, this could be a useful argument against implicit installation in #4211 given that something as small as hitting the tab key can lead to unexpected response time...

At least it can align better with https://github.com/rust-lang/rust-analyzer/issues/20719; not completely sure if `clap` can be taught to do that though.

Issue #2268, opened 2020-03-18
Duplicate auto completions; `clap` unaware of the `+toolchain` syntax


Aloxaf:
**Problem**

#2031 add a special argument `+toolchain`, this breaks the zsh completion script generated by clap.

**Detail**
`+toolchain` is treated as a positional argument by clap, this brings two problems:
1. `rustup \t` will also generate file list because `+toolchain` is treated as a file.
    > - Maintainer's update: Resolved in #4763.
2. the first argument is occupied by `+toolchain`, so the completion script thinks the real subcommand starts from the 2nd argument. This means only `rustup xxx toolchain \t` will generate completion for `toolchain` subcommand.

**Steps**

1. generate the completion script: `rustup completions zsh > a_dir_in_fpath/_rustup`
2. type `rustup \t`, you will see both rustup commands and local files (they shouldn't exist) in candidates.
3. type `rustup toolchain \t`, you will still see rustup commands, not rustup toolchain commands.

**Possible Solution(s)**

1. Throw this issue to clap
2. Use a (semi-)hand-written completion script
3. Do a search-and-replace in clap's output

**Notes**

Output of `rustup --version`: `rustup 1.21.1 (2020-02-23)`
Output of `rustup show`:
```
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/aloxaf/.rustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
nightly-2019-12-20-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)

installed targets for active toolchain
--------------------------------------

wasm32-unknown-unknown
x86_64-pc-windows-gnu
x86_64-unknown-linux-gnu

active toolchain
----------------

nightly-x86_64-unknown-linux-gnu (default)
 rustc 1.43.0-nightly (d3c79346a 2020-02-29)
```

Aloxaf:
A temporary fix for those who need it:

```diff
--- _rustup     2020-03-18 19:08:18.865256496 +0800
+++ _rustup_fix 2020-03-18 18:42:23.837856740 +0800
@@ -23,16 +23,18 @@
 '--help[Prints help information]' \
 '-V[Prints version information]' \
 '--version[Prints version information]' \
-'::+toolchain -- release channel (e.g. +stable) or custom toolchain to set override:_files' \
+'(+beta +nightly)+stable[use the stable toolchain]' \
[...]

PR #4763, opened 2026-03-20
Fix zsh completion showing all PATH entries for +toolchain arg

simonhkswan:
## Problem

When using zsh completions generated by `rustup completions zsh`, typing `rustup <TAB>` shows every executable on `$PATH` alongside the actual rustup subcommands (`install`, `update`, `show`, etc.). This makes the completion menu nearly unusable — thousands of irrelevant entries mixed in with the ~15 real subcommand completions.

## How to reproduce

1. Generate zsh completions: `rustup completions zsh > _rustup`
2. Place `_rustup` on your `$fpath` and reload the shell (`exec zsh`)
3. Type `rustup ` and press `<TAB>`
4. Observe that all commands on `$PATH` appear in the completion list

## Cause

The `+toolchain` optional positional argument in `src/cli/rustup_mode.rs` is defined without a `value_hint`. When `clap_complete` generates zsh completions, arguments with no hint default to `ValueHint::Unknown`, which the zsh generator maps to the `_default` completion action ([source](https://github.com/clap-rs/clap/blob/master/clap_complete/src/aot/shells/zsh.rs#L407)):

```rust
ValueHint::Unknown => "_default",
```

This produces the following in the generated completion script:

```
'::+toolchain -- Release channel (e.g. +stable) or custom toolchain to set override:_default'
```

In zsh, `_default` is a catch-all completer that suggests files and commands on `$PATH`. Since this is an optional positional argument (`::` prefix), zsh offers these completions alongside the subcommand completions every time.

## Fix

This PR adds `value_hint = ValueHint::Other` to the `+toolchain` argument. In the zsh generator, `ValueHint::Other` maps to an empty string — no completion action — which prevents the PATH pollution. Rustup subcommand completions continue to work as expected.

## Future improvement

A more complete solution would be to provide dynamic completions for the `+toolchain` argument using `clap_complete`'s `ArgValueCompleter`, which could call `rustup toolchain list` at completion time to suggest installed toolchains. That would be a larger change involving adoption of `clap_complete`'s dynamic completion infrastructure, so this PR focuses on the minimal fix to eliminate the broken behavior.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -13,7 +13,10 @@ use std::{
 use anstream::ColorChoice;
 use anstyle::Style;
 use anyhow::{Context, Error, Result, anyhow};
-use clap::{Args, CommandFactory, Parser, Subcommand, ValueEnum, builder::PossibleValue};
+use clap::{
+    Args, CommandFactory, Parser, Subcommand, ValueEnum,
+    builder::{PossibleValue, ValueHint},
+};
 use clap_cargo::style::{CONTEXT, ERROR, GOOD, HEADER, TRANSIENT, WARN};
 use clap_complete::Shell;
 use futures_util::stream::StreamExt;
@@ -97,6 +100,7 @@ struct Rustup {
     #[arg(
         name = "+toolchain",
         value_parser = plus_toolchain_value_parser,
+        value_hint = ValueHint::Other,
     )]
     plus_toolchain: Option<ResolvableToolchainName>,
 


rami3l:
@simonhkswan Thanks for investigating! Do you think your fix will address https://github.com/rust-lang/rustup/issues/2268 / #4532 as well, or at least a part of them? The symptoms look pretty similar...

src/cli/rustup_mode.rs
```diff
@@ -13,7 +13,10 @@ use std::{
 use anstream::ColorChoice;
 use anstyle::Style;
 use anyhow::{Context, Error, Result, anyhow};
-use clap::{Args, CommandFactory, Parser, Subcommand, ValueEnum, builder::PossibleValue};
```
rami3l, line 16: Please squash the 2nd commit into the 1st one.

rami3l (approved):
I think this might need more testing by the impacted users, but otherwise LGTM!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4763/bd69b96d9f6cc112d386b58f4d784fb60d1469f5..5c4dfe479fefd7aa4a7a554f742578cffcdc3d10)*

src/cli/rustup_mode.rs
simonhkswan: Done!

simonhkswan:
> @simonhkswan Thanks for investigating! Do you think your fix will address #2268 / #4532 as well, or at least a part of them? The symptoms look pretty similar...

So I believe this PR directly addresses problem 1 from #2268 — `rustup <TAB>` showing file/PATH entries alongside the real subcommands. That's caused by `clap_complete` emitting the `_default` zsh completion action for the `+toolchain` positional arg, which this fix eliminates.

Problem 2 from #2268 (`rustup toolchain <TAB>` showing top-level commands instead of toolchain subcommands) is a deeper issue with how clap handles subcommand parsing when an optional positional arg precedes them. The more involved follow-up I mentioned in the PR description — using `clap_complete`'s `ArgValueCompleter` to dynamically complete `+toolchain` values via `rustup toolchain list` — could potentially address this too, by giving clap proper knowledge of what values `+toolchain` accepts and helping it correctly distinguish the toolchain argument from the subcommand position.

#4532 seems closed as a duplicate but maybe the same approach mentioned above could address this too.

rami3l:
@simonhkswan Sounds good! Did you test this change on your machine?

If it works well I'd love to merge it and update the problem description to leave the second problem above for further discussion.

djc (approved):

simonhkswan:
> @simonhkswan Sounds good! Did you test this change on your machine?
> 
> If it works well I'd love to merge it and update the problem description to leave the second problem above for further discussion.

@rami3l - I did test it (and actually using it permanently now :)

rustup

rami3l:
@simonhkswan Cool, and thanks again for your contribution!

PR #4773, opened 2026-03-24
ci: don't install protoc

djc:
@rami3l don't think we actually need it?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1512,14 +1512,6 @@ jobs:
         uses: actions/checkout@v6
       - name: Install rustup stable
         run: rustup toolchain install stable --profile minimal
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v3
-        env:
-          # renovate: datasource=github-releases depName=protocolbuffers/protobuf versioning=semver
-          PROTOBUF_VERSION: "34.1"
-        with:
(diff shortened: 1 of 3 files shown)


rami3l (approved):
@djc LGTM!

In that case maybe we should remove the description about `protoc` in https://rust-lang.github.io/rustup/dev-guide/tracing.html?highlight=protoc#opentelemetry-tracing as well? I remember having to install `protoc` locally to build the feature but that might have changed since then.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4773/dfc31fe32f67f5f105ab99267027810213fe4e97..ab2bc55677d7e8ffcd33aee540c9dfc1f1b18c72)*

djc:
> In that case maybe we should remove the description about `protoc` in https://rust-lang.github.io/rustup/dev-guide/tracing.html?highlight=protoc#opentelemetry-tracing as well? I remember having to install `protoc` locally to build the feature but that might have changed since then.

Removed that section.

rami3l:
To double check: [no trace](https://github.com/search?q=repo%3Aactions%2Frunner-images+%2Fproto%28c%7Cbuf%29%2F&type=code) of protobuf being implicitly installed on GitHub Actions, so looks safe to remove. Thanks again :)

Issue #4738, opened 2026-03-04
Tracking: Introduce proper trunk/backport workflow


rami3l:
For now every rustup release is directly merged onto trunk and thus we might want to block patch releases from happening once a major/breaking change has landed (or vice versa). This is not very practical.

To keep and even further increase the bandwidth and momentum, an important change we would need to make to the current workflow would be to allow people from both inside and outside of the project to contribute whenever they want, and for the changes to land in the right release in a more flexible timeframe, so that we can introduce major changes even when we are waiting for patch releases. Thus the desire for a backport branch that collects all the bugfixes in this minor cycle, with proper CI and even CD (when the new release process is ready), from which we will force-push our `stable`s.

Surely this needs more bandwidth, so it will probably not happen immediately, but I can see that this might be an important step in ensuring the project's long term health.

## Tasks

- Merge the initial PR
  - #4774
- Test this change thoroughly a set dummy PRs:
  - #4792
  - #4799
  - #4815 
- [x] Make sure everything is good:
  - [x] In both the CI files and git itself, rename `backport/*` to `release/*` as [suggested](https://github.com/rust-lang/rustup/pull/4774#issuecomment-4153376910):
    - [x] https://github.com/rust-lang/rustup/pull/4791 
  - [x] The `conclusion` job is identical to the `main` one.
  - [x] Other pre-enqueue jobs are identical to the `main` ones.
  - [ ] ~~The in-queue jobs are identical to the `main` ones.~~ No queue in the end, see below.
  - [x] The after-merge jobs are identical to the `main` ones modulo the deployment.
    - [x] The deployment to GitHub Actions is kept.
    - [x] All other deployments should not trigger.
- [x] Enable the manual triggering of `schedule`-level CI.
  - #4796 
  - #4799
- [x] Delete the `release/1.29-dummy` branch.
- [x] Inform the infra team to ~~enable merge queue~~ (technically impossible due to GitHub restrictions (https://github.com/rust-lang/team/pull/2372#issuecomment-4182312803)) ~~and~~ add branch protection rules (notably the force-push ban) for `release/*` branches.
    - https://github.com/rust-lang/team/pull/2437
- [x] Update the dev guide to reflect the new workflow, including the manual triggering of CI, changed branching rules, changed `stable` status (can be force pushed instead of fast-forwarded), etc.
    - https://github.com/rust-lang/rustup/pull/4805
[...]

PR #4774, opened 2026-03-25
ci(backport): add support for backporting

rami3l:
Partially addresses #4738. 
It turns out that my previous CI refactoring was so good that this change can stay minimal.

cc @FranciscoTGouveia


ci/actions-templates/test-docs-template.yaml
```diff
@@ -41,7 +41,7 @@ jobs: # skip-all
           mkdir -p ${{ runner.temp }}/book
           mv book ${{ runner.temp }}/book/dev-guide
       - name: Deploy to GitHub
-        if: ${{ !contains('["pull_request", "merge_group"]', github.event_name) }}
+        if: ${{ !contains('["pull_request", "merge_group"]', github.event_name) && !startsWith(github.ref_name, 'backport/') }}
```
rami3l, line 44: There are a lot of docs-deploying jobs already (plus the latest `www` branch rule for www-specific hotfix from @Kobzol), I don't think we should overwrite GitHub Pages in any more places 😅

ChrisDenton (approved):

djc:
One suggestion: I think the `backport/` prefix is a little weird for this. Backporting IMO is the action of copying changes from a branch targeting a future release to branches for previous releases. I think something like `release/` or `rel/` is more obvious.

rami3l:
> One suggestion: I think the `backport/` prefix is a little weird for this. Backporting IMO is the action of copying changes from a branch targeting a future release to branches for previous releases.

@djc That's exactly what I meant though... As I see it, we're always going to include either a port or a backport process, can we do it otherwise actually? How do you usually do it in other projects you maintain? 🤔

rami3l:
I imagine some kind of automation can be involved if we enforce semi-parsable PR/commit titles, if there's a `BREAKING` or `!` indicator in the title we just don't include the change in the backporting PR 🙈

djc:
Sorry, I only meant the branch name prefix -- projects like Hickory and rustls use something like what I proposed.

Backporting probably doesn't make sense to automate until we're at substantially higher throughput.

rami3l:
@djc Okay, I'd take `release/xxx` in that case.

PR #4768, opened 2026-03-22
Fix install message misalignment

Cloud0310:
Fix #4759 with dynamic name width.

PR #4778, opened 2026-03-26
chore: Rename `Component::name_in_manifest` and `short_name_in_manifest` method to `name` and `short_name`

Cloud0310:
These function on `Component` is not related to `Manifest`, so renaming them to distinguish against `Manifest::name()` and `Manifest::short_name()`. 

Founded this problem when I was dealing with #4768. By naming, two functions looks like should be the same, but they're actually different things.

ref: [src/dist/manifest.rs#L55-L62](https://github.com/rust-lang/rustup/tree/main/src/dist/manifest.rs#L55-L62) [src/dist/manifest.rs#L55-L62](https://github.com/rust-lang/rustup/tree/main/src/dist/manifest.rs#L73-L80)

### DIFF
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -317,7 +317,7 @@ impl Manifest {
     }
 
     pub(super) fn binary(&self, component: &Component) -> Result<Option<&HashedBinary>> {
-        let package = self.get_package(component.short_name_in_manifest())?;
+        let package = self.get_package(component.short_name())?;
         let target_package = package.get_target(component.target.as_ref())?;
         // We prefer the first format in the list, since the parsing of the
         // manifest leaves us with the files/hash pairs in preference order.
@@ -456,7 +456,7 @@ impl Manifest {
 
             // Get the component so we can check if it is available
             let component_pkg = self
-                .get_package(component.short_name_in_manifest())
+                .get_package(component.short_name())
                 .unwrap_or_else(|_| {
                     panic!(
                         "manifest should contain component {}",
@@ -538,7 +538,7 @@ impl Component {
         let manifest = distributable.get_manifest()?;
         for component_status in distributable.components()? {
             let component = component_status.component;
-            if name == component.name_in_manifest() || name == manifest.name(&component) {
+            if name == component.name() || name == manifest.name(&component) {
                 return Ok(component);
             }
         }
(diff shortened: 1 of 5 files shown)


djc (approved):
Makes sense to me, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4778/6f8931c636cb20317d85223ac72bfd44f58f5e89..16e04c8f0078035b523cfe0f1d3abd43fd858229)*

src/dist/manifest.rs
```diff
@@ -558,11 +558,11 @@ impl Component {
         }
     }
 
-    pub fn short_name_in_manifest(&self) -> &String {
+    pub fn short_name(&self) -> &String {
```
rami3l, line 567: Nit: Kind of pre-existing, but I'd argue that to make this even less of a confusion going forward, it'd be cool to add doc comments to explicitly compare the `Component::` and `Manifest::` functions in question. It can be really simple.

How about something like the following in the `impl Manifest` block (not sure about the syntax, but the overall idea should be clear):

```rs
/// Returns the [`Component`]'s short name (not including the target) before renaming.
/// For the name after renaming, see [`Component::short_name()`].
```

```rs
/// Returns the [`Component`]'s name (including the target) before renaming.
/// For the name after renaming, see [`Component::name()`].
```

rami3l (approved):
LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4778/6f8931c636cb20317d85223ac72bfd44f58f5e89..16e04c8f0078035b523cfe0f1d3abd43fd858229)*

src/dist/manifest.rs
Cloud0310: Got the idea here, and emphasized the `before` and `after` keywords.
Though the comment should be:
```
/// Returns the [`Component`]'s name (including the target) before renaming.
/// For the name after renaming, see [`Manifest::name()`].
```

Cloud0310:
Added new comments for clarify the usage of the functions in latest commit here.

rami3l (approved):
LGTM, let's merge it :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4778/0a96a3388fe0fe0915fe2e00580a9879bc8ea434..d9c345debb4c96eaef24f775684a5399c36e6c5a)*

src/dist/manifest.rs
```diff
@@ -70,6 +72,8 @@ impl Manifest {
         }
     }
 
+    /// Returns the [`Component`]'s short name (not including the target) **after** renaming.
+    /// For the name before renaming, see [`Component::short_name()`].
     pub(crate) fn short_name<'a>(&'a self, component: &'a Component) -> &'a str {
         if let Some(from) = self.reverse_renames.get(&component.pkg) {
```
rami3l, line 78: Wait... We are getting the name from `self.reverse_renames` so shouldn't this be the short name **before** renaming, instead of **after**?

Cloud0310: Oh, kinda confused here. Thanks for pointing me out. :heart:

PR #4786, opened 2026-03-29
fix: Reduce flickering by using `set_move_cursor`

### LINKED ISSUE #4757: The new progressbar flickers badly
### Verification

### Problem

When there are multiple progressbars being updated at the same time, there is heavy flickering when using Konsole.

### Steps

1. `rustup update`

### Possible Solution(s)

Reduce update interval to at most twice every second and/or use the synchronized update escape codes: https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec

### Notes

_No response_

### Rustup version

```console
```

### Installed toolchains

```console
Not relevant
```

### OS version

```console
Linux
```

ChrisDenton:
Instead of clearing the line then writing the new line.

Fixes #4757 (hopefully)

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -844,6 +844,8 @@ async fn check_updates(cfg: &Cfg<'_>, opts: CheckOpts) -> Result<ExitCode> {
     if channels_len > 0 {
         let multi_progress_bars =
             MultiProgress::with_draw_target(cfg.process.progress_draw_target());
+        // Help avoid flickering by moving the cursor instead of clearing the line.
+        multi_progress_bars.set_move_cursor(true);
         let semaphore = Arc::new(Semaphore::new(concurrent_downloads));
         let channels = tokio_stream::iter(channels.into_iter()).map(|(name, distributable)| {
             let msg = format!("{bold}{name} - {bold:#}");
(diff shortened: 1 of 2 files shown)


ChrisDenton:
cc @bjorn3

rami3l:
I don't think we are adding new progress bars on-the-fly so we should be able to safely ignore the caveats of this method.

Also cc @ehuss

rami3l (approved):

djc (approved):

djc:
Would be great to get some testing for this!

rami3l:
> Would be great to get some testing for this!

@djc Definitely. I'm also curious about the solution @bjorn3 proposed about reducing the tick rate 😶

djc:
> > Would be great to get some testing for this!
> 
> @djc Definitely. I'm also curious about the solution @bjorn3 proposed about reducing the tick rate 😶

I'd rather see if this works first.

ChrisDenton:
Oops Sorry. I wasn't paying attention when operating my git client.

ehuss:
This looks *substantially* better. Zero flicker using the same setup for Windows Terminal as I had in the issue.

rami3l:
Let's merge it then. If anything happens again let's see what we can do with the tick rate.

PR #4795, opened 2026-04-01
Fix documentation link to `no-self-update` feature

FranciscoTGouveia:
While reviewing #3051, I noticed that this link was deprecated.

That said, I would like to point out that whenever `Cargo.toml` is modified (specifically, if lines are added above the current location of `no-self-update`), the link will break again.
I am not sure if there is a reliable way to prevent this, but I wanted to highlight it as a potential issue going forward.

### DIFF
--- a/doc/user-guide/src/basics.md
+++ b/doc/user-guide/src/basics.md
@@ -30,7 +30,7 @@ info: downloading self-update
 
 ## Keeping `rustup` up to date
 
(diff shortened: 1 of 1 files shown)


rami3l:
> whenever `Cargo.toml` is modified (specifically, if lines are added above the current location of `no-self-update`), the link will break again. I am not sure if there is a reliable way to prevent this, but I wanted to highlight it as a potential issue going forward.

@FranciscoTGouveia Fair point, I'd say maybe just forget about the line number...

rami3l (approved):

djc (approved):

PR #4796, opened 2026-04-02
ci(test): add `workflow_dispatch` trigger on par with `schedule`

rami3l:
Part of #4738.

This PR enables manually-triggered "full" CI at the same level as `schedule`.

As per https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow#running-a-workflow, the workflow can be manually triggered via the GitHub CLI as shown below (must be under the `rustup` repo, commands not yet tested):

```fish
# Run on trunk
> gh workflow run ci.yaml

# Run on backport
> gh workflow run ci.yaml --ref release/1.29-dummy
```

This should be helpful for occasionally running the full CI on backport branches, and in the meantime we won't have to wait for the scheduled run on trunk again if the situation is urgent.


djc (approved):

rami3l:
@djc Thanks! The first command mentioned above seems to be working, just so you know :)

I'll backport the change to the `release/1.29-dummy` branch and try the second.

PR #4797, opened 2026-04-04
Make component removal best-effort and preserve single-error behavior

### LINKED ISSUE #4794: Make the removal of components consistent
Currently, the behavior of `rustup component remove` depends on the order of the provided components.

If an invalid component is listed first, the command fails immediately, and no components are removed.
However, if valid components are listed before an invalid one, those valid components are removed before the command returns an error.

This behavior is thus inconsistent, as it is dependent on the argument ordering.
From a usability perspective, I would argue that it is more intuitive for the command to attempt to remove all valid components and report any invalid ones afterward, regardless of their position.

That said, I am interested in hearing others' thoughts on this.

---
Examples of said inconsistency:

```
$ rustup component remove asd rustc-dev llvm-tools
error: toolchain 'stable-x86_64-unknown-linux-gnu' does not contain component 'asd' for target 'x86_64-unknown-linux-gnu'
```

```
$ rustup component remove rustc-dev llvm-tools asd
info: removing component rustc-dev
info: removing component llvm-tools
error: toolchain 'stable-x86_64-unknown-linux-gnu' does not contain component 'asd' for target 'x86_64-unknown-linux-gnu'
```

Inconnu08:
Previously, `rustup component remove` parsed and removed components in a single pass,
which meant the command stopped on the first error after potentially removing earlier
valid components. As a result, behavior depended on argument order.

This change makes component removal best-effort by:
- first collecting all parseable requested components
- then attempting to remove each of them
- deferring error reporting until the end

This makes the behavior consistent regardless of whether an invalid component appears
before or after valid ones as discussed here: #4794 .

Also adds a regression test covering both orderings.

Closes #4794

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1492,12 +1492,37 @@ async fn component_remove(
     let distributable = DistributableToolchain::from_partial(toolchain, cfg).await?;
     let target = get_target(target, &distributable);
 
-    for component in &components {
-        let new_component = Component::try_new(component, &distributable, target.as_ref())?;
-        distributable.remove_component(new_component).await?;
+    let parsed_components = components
+        .iter()
+        .map(|component| Component::try_new(component, &distributable, target.as_ref()))
+        .collect::<Result<Vec<_>>>()?;
+
+    let mut unknown_components = Vec::new();
+
+    for component in parsed_components {
+        let Err(err) = distributable.remove_component(component).await else {
+            continue;
+        };
+
+        if let Some(RustupError::UnknownComponents { components, .. }) =
+            err.downcast_ref::<RustupError>()
+        {
+            unknown_components.extend(components.iter().cloned());
+            continue;
+        }
+
+        return Err(err);
     }
 
-    Ok(ExitCode::SUCCESS)
+    if unknown_components.is_empty() {
+        Ok(ExitCode::SUCCESS)
+    } else {
+        Err(RustupError::UnknownComponents {
+            desc: distributable.desc().clone(),
+            components: unknown_components,
+        }
+        .into())
+    }
 }
 
 async fn toolchain_link(cfg: &Cfg<'_>, dest: &CustomToolchainName, src: &Path) -> Result<ExitCode> {
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::large_enum_variant)]
 
 use std::ffi::OsString;
-use std::fmt::Debug;
+use std::fmt::{Debug, Write as FmtWrite};
 use std::io;
 use std::io::Write;
 use std::path::PathBuf;
@@ -22,6 +22,13 @@ use crate::{
 #[derive(Debug, ThisError)]
 #[error(transparent)]
 pub struct OperationError(pub anyhow::Error);
+
+#[derive(Debug, Clone)]
+pub struct UnknownComponentInfo {
+    pub name: String,
+    pub description: String,
+    pub suggestion: Option<String>,
+}
 
 #[derive(ThisError, Debug)]
 pub enum RustupError {
@@ -124,15 +131,10 @@ pub enum RustupError {
         help: run 'rustup default stable' to download the latest stable release of Rust and set it as your default toolchain."
     )]
     ToolchainNotSelected(String),
-    #[error("toolchain '{}' does not contain component {}{}{}", .desc, .component, suggest_message(.suggestion), if .component.contains("rust-std") {
-        format!("\nnote: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html{}",
-            if desc.channel == Channel::Nightly { "\nhelp: consider using `cargo build -Z build-std` instead" } else { "" }
-        )
-    } else { "".to_string() })]
-    UnknownComponent {
+    #[error("{}", unknown_components_msg(.desc, .components))]
+    UnknownComponents {
         desc: ToolchainDesc,
-        component: String,
-        suggestion: Option<String>,
+        components: Vec<UnknownComponentInfo>,
     },
     #[error(
         "toolchain '{desc}' has no prebuilt artifacts available for target '{platform}'\n\
@@ -241,4 +243,50 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
     }
 
     String::from_utf8(buf).unwrap()
+}
+
+fn unknown_components_msg(desc: &ToolchainDesc, components: &[UnknownComponentInfo]) -> String {
+    let mut buf = String::new();
+
+    match components {
+        [] => panic!("`unknown_components_msg` should not be called with an empty collection"),
+        [component] => {
+            let _ = write!(
+                buf,
+                "toolchain '{desc}' does not contain component {}",
+                component.description,
+            );
+
+            if let Some(suggestion) = &component.suggestion {
+                let _ = write!(buf, "\nhelp: did you mean '{suggestion}'?");
+            }
+
+            if component.description.contains("rust-std") {
+                let _ = write!(
+                    buf,
+                    "\nnote: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html"
+                );
+
+                if desc.channel == Channel::Nightly {
+                    let _ = write!(
+                        buf,
+                        "\nhelp: consider using `cargo build -Z build-std` instead"
+                    );
+                }
+            }
+        }
+        components => {
+            let _ = writeln!(buf, "toolchain '{desc}' does not contain these components:");
+
+            for component in components {
+                let _ = writeln!(buf, "  - '{}'", component.name);
+
+                if let Some(suggestion) = &component.suggestion {
+                    let _ = writeln!(buf, "    help: did you mean '{suggestion}'?");
+                }
+            }
+        }
+    }
+
+    buf
 }
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -1,14 +1,19 @@
 //! Test cases for new rustup UI
 
-use std::fs;
-use std::path::PathBuf;
-use std::{env::consts::EXE_SUFFIX, path::Path};
-
-use rustup::test::{
-    CROSS_ARCH1, CROSS_ARCH2, CliTestContext, MULTI_ARCH1, Scenario, this_host_triple,
-    topical_doc_data,
+use std::{
+    env::consts::EXE_SUFFIX,
+    fs,
+    path::{Path, PathBuf},
 };
-use rustup::utils::raw;
+
+use rustup::{
+    for_host,
+    test::{
+        CROSS_ARCH1, CROSS_ARCH2, CliTestContext, MULTI_ARCH1, Scenario, this_host_triple,
+        topical_doc_data,
+    },
+    utils::raw,
+};
 
 #[tokio::test]
 async fn rustup_stable() {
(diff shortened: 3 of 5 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1481,12 +1481,33 @@ async fn component_remove(
     let distributable = DistributableToolchain::from_partial(toolchain, cfg).await?;
     let target = get_target(target, &distributable);
 
+    let mut parsed_components = Vec::new();
```
rami3l: I think it'd not be user friendly enough to only show the first error? I'd expect to get all wrong names at once and only correct once. You can even add/generalize the existing error type for this.

rami3l (commented):
Many thanks for working on this :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4797/9f7bd9c2780bfed5d08c6d53ce8e196aa7a3a2e5..9bcc18098b766b572d7526890a23e5ef14a95a36)*

src/cli/rustup_mode.rs
Inconnu08: I wanted to keep the change minimal, but I agree 💯 it would be better to report all invalid component names at once. I’ll update the implementation to collect all failures and use a generalized error type for them.

src/cli/rustup_mode.rs
```diff
@@ -1482,11 +1482,12 @@ async fn component_remove(
     let target = get_target(target, &distributable);
 
     let mut parsed_components = Vec::new();
+    let mut unknown_components = Vec::new();
     let mut first_error = None;
 
     for component in &components {
         match Component::try_new(component, &distributable, target.as_ref()) {
-            Ok(new_component) => parsed_components.push(new_component),
+            Ok(new_component) => parsed_components.push((new_component, component.clone())),
```
rami3l: Since `component` borrows from `components`, I don't think this `.clone()` is needed; instead you could delay the clone till the moment of generating the error `RustupError::UnknownComponents`.

src/cli/rustup_mode.rs
```diff
@@ -1495,12 +1496,27 @@ async fn component_remove(
...
+    for (component, original_name) in parsed_components {
+        if let Err(err) = distributable.remove_component(component).await {
+            if let Some(RustupError::UnknownComponent { suggestion, .. }) =
+                err.downcast_ref::<RustupError>()
+            {
+                unknown_components.push((original_name, suggestion.clone()));
+                continue;
+            }
+
+            if first_error.is_none() {
```
rami3l: Nit: This can be more clearly written as `first_error.get_or_insert(err)`.

tests/suite/cli_v2.rs
```diff
@@ -2180,7 +2180,7 @@ async fn remove_component_suggest_best_match() {
         .expect(["rustup", "component", "remove", "rsl"])
         .await
         .with_stderr(snapbox::str![[r#"
-error: toolchain 'nightly-[HOST_TRIPLE]' does not contain component 'rsl' for target '[HOST_TRIPLE]'; did you mean 'rls'?
+error: toolchain 'nightly-[HOST_TRIPLE]' does not contain component 'rsl'; did you mean 'rls'?
```
rami3l: Question: This regression seems to be avoidable?

src/cli/rustup_mode.rs
Inconnu08: Ah, yes ...updated to avoid cloning eagerly and instead clone only when constructing the grouped error

src/cli/rustup_mode.rs
Inconnu08: Wow, that's cleaner, thanks 👍

tests/suite/cli_v2.rs
Inconnu08: Yes ...I’ve now updated the implementation to preserve the existing single-component UnknownComponent behavior and only use grouped reporting when multiple invalid components are present.

src/cli/rustup_mode.rs
```diff
@@ -1495,12 +1496,27 @@ async fn component_remove(
         }
     }
 
-    for component in parsed_components {
-        if let Err(err) = distributable.remove_component(component).await
-            && first_error.is_none()
-        {
-            first_error = Some(err);
+    for (component, original_name) in parsed_components {
+        if let Err(err) = distributable.remove_component(component).await {
```
djc: Nit: rewrite this condition to avoid rightward drift? Like

```rust
let Err(err) = distributable.remove_component(component).await {
     continue;
};
```

src/errors.rs
```diff
@@ -242,3 +247,27 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
 
     String::from_utf8(buf).unwrap()
 }
+
+fn unknown_components_msg(desc: &ToolchainDesc, components: &[(String, Option<String>)]) -> String {
+    let mut buf = vec![];
```
djc: I don't think it really makes sense to initialize this to a `Vec` and then convert it to `String` at the end? Suggest to make it `String` immediately and use `.write_fmt(format_args!(..))` to write into it.

djc (commented):
These commits are really not independent -- would prefer to have them all squashed.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4797/9f7bd9c2780bfed5d08c6d53ce8e196aa7a3a2e5..6c5bb6e06ff05fd74e0379e13fafe992c5df1059)*

src/cli/rustup_mode.rs
Inconnu08: Thanks, this reads much cleaner!

src/errors.rs
Inconnu08: True that ...switched to building the message directly as a String 👍

djc (approved):

rami3l:
Hmmm my apologies, but I think I might need some extra time for a final look at this before merging 🙏

rami3l (commented):
I'm feel okay about it globally but I have a few concerns:

- I would like to confirm with @FranciscoTGouveia whether the original intention of https://github.com/rust-lang/rustup/issues/4794 is to aggregate certain errors (which is what I believe) rather than just showing the first error but with the bail-out delayed. IMHO the behavior as implemented in this patch will be very confusing to the user if multiple errors do happen. In this case I suggest that we either aggregate `UnknownComponent*` separately and early bail out for the rest, or actually aggregate all errors (which might be more difficult in terms of display).

- The current distinction of `UnknownComponent` and `UnknownComponents` doesn't seem super necessary since it doesn't prevent the API user from creating the latter with only one `Vec` member.

- The tuples in `UnknownComponents` might be better expressed via a separate struct, but that's a minor implementation detail.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4797/5100b80375fade4ee4c7c0714c1132fec28319d9..eeb7b06f2369f1b48a80da1281ebd39d80f89d07)*

Inconnu08:
Thanks, that makes sense.

My intent with the current implementation was to:

- preserve the existing UnknownComponent behavior for the single-component case
- aggregate unknown component names only when multiple invalid components are present
- keep the rest of the behavior as close as possible to the existing flow

One of the motivations for introducing `UnknownComponents` was also to avoid repeating the full error message multiple times when several invalid components are provided, which felt a bit noisy UX-wise. The grouped format seemed more concise and easier to read in that case.

That said, I agree the mixed behavior could be confusing if multiple different error types occur. I’m happy to adjust the approach once the intended UX for #4794 is confirmed with @FranciscoTGouveia  whether that means aggregating only `UnknownComponent*` and bailing early for others, or reshaping the error handling more uniformly.

Also, point on UnknownComponent vs UnknownComponents: I introduced the second mainly to avoid regressing the existing single-component behavior, but I’m open to simplifying that if we settle on a cleaner model.

And agreed on the tuple, a small struct would likely make that clearer if we keep the grouped variant.

FranciscoTGouveia:
The original intention of 4794 was indeed to install all possible components and report back _every_ invalid one.

For example, running `rustup component add rust-analyzere rustc-dev llvm-toolse`, should install `rustc-dev` successfully, while returning two separate errors, one for each misspelled component.

On another note, I briefly tested the patch and believe it would be worth adding a test case covering the example above, as this is the current output:
```
$ my-rustup component add rust-analyzere rustc-dev llvm-toolse
error: toolchain '1.90-x86_64-unknown-linux-gnu' does not contain component 'rust-analyzere' for target 'x86_64-unknown-linux-gnu'; did you mean 'rust-analyzer'?
```

rami3l:
> The original intention of 4794 was indeed to install all possible components and report back _every_ invalid one.

@FranciscoTGouveia Thanks for your reply! Given that, my suggestions would be:

- It'd be easier to just aggregate all `UnknownComponent*` errors in the installation process and bail out immediately like before if other kinds of errors do occur.

- On the distinction of `UnknownComponent` vs `UnknownComponents`, I'd say it's more idiomatic to special-case the display format on a unified `UnknownComponents`, rather than doing two `enum` variants. In other words: if the `Vec` is of length 1 we print like the old `UnknownComponent`, otherwise we use the new printing format as introduced here.

- It'd be better if the helper `struct` could be introduced.

@Inconnu08 What do you think?

rami3l:
(Assigned @FranciscoTGouveia for review upon his request.)

src/errors.rs
```diff
@@ -242,3 +247,27 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
...
+fn unknown_components_msg(desc: &ToolchainDesc, components: &[(String, Option<String>)]) -> String {
+    let mut buf = String::new();
+
+    match components {
+        [] => panic!("`unknown_components_msg` should not be called with an empty collection"),
+        [(component, suggestion)] => {
+            let _ = write!(
+                buf,
+                "toolchain '{desc}' does not contain component '{component}'{}",
+                suggest_message(suggestion),
```
FranciscoTGouveia: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4797/5100b80375fade4ee4c7c0714c1132fec28319d9..eeb7b06f2369f1b48a80da1281ebd39d80f89d07)*

What if we insert a newline before calling `suggest_message`?
Otherwise, we could easily exceed 80 chars :/

tests/suite/cli_rustup.rs
```diff
@@ -2136,6 +2136,145 @@ async fn add_remove_multiple_components() {
...
+
+    for file in &files {
+        let path = PathBuf::from(format!(
+            "toolchains/nightly-{}/{}",
+            this_host_triple(),
+            file
+        ));
+        assert!(!cx.config.rustupdir.has(path.parent().unwrap()));
+    }
+}
```
FranciscoTGouveia, line 2181: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4797/5100b80375fade4ee4c7c0714c1132fec28319d9..eeb7b06f2369f1b48a80da1281ebd39d80f89d07)*

There should also be a third case in which a valid component appears between invalid ones, ensuring that errors are correctly reported for both invalid components.

src/errors.rs
Inconnu08: Good point. I will update this.

tests/suite/cli_rustup.rs
Inconnu08: That’s a useful edge case to cover. I will add it to the test.

Inconnu08:
Apologies for the late response. Thanks to both of you for clarifying the intended behavior :) 

My understanding now is:

- we should process all valid components: best-effort
- aggregate all `UnknownComponent*` errors and report them together
- still bail out immediately for other kinds of errors

Using a single `UnknownComponents` variant with display logic that special-cases the length (to preserve the current single-component/best-match UX) makes sense to me and simplifies the current approach.

The earlier split between `UnknownComponent` and `UnknownComponents ` was mainly to avoid regressing the single-component formatting/suggestion behavior, but I’m happy to unify it now that the intended direction is clear.

I’ll update the implementation accordingly, and I’ll also:

- add a test covering the example (`rust-analyzere rustc-dev llvm-toolse`)
- extend the existing tests with a case where a valid component appears between invalid ones
- adjust the error formatting to move suggestions onto a new line for readability

Let me know if that aligns before I push the changes.

src/errors.rs
Inconnu08: I’ve updated the formatting to move the suggestion onto a new `help:` line. This keeps the line length in check and improves readability. I’ve also updated the affected snapshots accordingly.

src/errors.rs
rami3l: I think given that this is not used a lot elsewhere, we can just remove it to save some redundancy? (Please correct me if I'm wrong.)

src/errors.rs
```diff
@@ -242,3 +254,47 @@ fn component_unavailable_msg(cs: &[Component], manifest: &Manifest, toolchain: &
...
+
+            if let Some(suggestion) = &component.suggestion {
+                let _ = write!(buf, "\nhelp: did you mean '{suggestion}'?");
+            }
+
+            if component.description.contains("rust-std") {
+                let _ = write!(
+                    buf,
+                    "\nnote: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html{}",
+                    if desc.channel == Channel::Nightly {
```
rami3l: Nit: Since we are not forced to writing in expression-oriented style any more (we are no longer in an attribute), prefer doing this instead:

```rs
let _ = write!(buf, "\nnote: not all platforms have the standard library pre-compiled: https://doc.rust-lang.org/nightly/rustc/platform-support.html");
if desc.channel == Channel::Nightly {
  let _ = write!(buf, "\nhelp: consider using `cargo build -Z build-std` instead");
}
```

rami3l (approved):
That's much clearer than before, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4797/9b6b629804d6e4e8c6baf83b62ad3a052f4286ee..e626a89062264285545d48eada123d9f0b2217b4)*

src/errors.rs
Inconnu08: since we’ve unified on UnknownComponents, the old UnknownComponent variant is redundant. I’ll remove it and update any remaining call sites if needed

src/errors.rs
rami3l: Looks cool to me :)

rami3l:
Thanks again @Kobzol , your newly-enabled automations have made the tracking of subtle changes across rebases much easier 🙏

rami3l:
@Inconnu08 Great work!

There are some test-related stuff that I'll update for you :)

Kobzol:
> Thanks again @Kobzol , your newly-enabled automations have made the tracking of subtle changes across rebases much easier 🙏 

That was actually implemented by @Urgau :)

PR #4790, opened 2026-03-31
Allow rustup component add to install multiple components in one update

Inconnu08:
This updates `rustup component add` to handle multiple components in a
single operation instead of processing them sequentially.

Closes #4787
Related to #731

PR #4798, opened 2026-04-04
fix(dist/manifestation): fix log format when installing exactly 2 components

rami3l:
Follow-up of https://github.com/rust-lang/rustup/pull/4790.

TLDR: During manual testing a very silly typo has surfaced 🐛

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -244,13 +244,12 @@ impl Manifestation {
         }
 
         if !components.is_empty() {
-            if components.len() > 2 {
-                info!("downloading {} components", components.len());
-            } else {
-                info!(
+            match &components[..] {
+                [c] => info!(
                     "downloading component {}",
-                    components[0].manifest.short_name(&components[0].component),
-                );
+                    c.manifest.short_name(&c.component),
+                ),
+                _ => info!("downloading {} components", components.len()),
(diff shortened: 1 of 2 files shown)


src/dist/manifestation.rs
```diff
@@ -243,14 +243,14 @@ impl Manifestation {
             tx = self.uninstall_component(component, &new_manifest, tx)?;
         }
 
-        if !components.is_empty() {
-            if components.len() > 2 {
-                info!("downloading {} components", components.len());
-            } else {
-                info!(
+        'install: {
+            match &components[..] {
```
rami3l, line 247: The one-line fix is `if components.len() > 1` but since we are matching the `components` anyway, I guess it'd be clearer to group all len-related predicates here.

djc: I don't love this labeled block thing... it seems pretty unidiomatic?

rami3l: @djc I personally find it more readable... But I've found for us a middle ground. What do you think about this newer version?

djc: Yes, I like this better. I appreciate the attempt to do the length check in one place but I think in this case the control flow around it makes this better.

djc (approved):

PR #4805, opened 2026-04-09
docs(dev-guide): update release process with new backporting flow

rami3l:
Part of https://github.com/rust-lang/rustup/issues/4738.

This flow is not verified yet but I think as long as this passes the vibe check, we can merge it first, and then update it in place as a part of the next `CHANGELOG.md` update during the next release if some changes are indeed required.

### DIFF
--- a/doc/dev-guide/src/release-process.md
+++ b/doc/dev-guide/src/release-process.md
@@ -1,5 +1,88 @@
 # Making a release
 
+In rustup, there are two possible release "modes": the beta release and the
+official release. The main difference between the two is that they use
+different values for the `RUSTUP_UPDATE_ROOT` environment variable:
+
+- A beta release is deployed on `https://dev-static.rust-lang.org/rustup`.
+- An official release is deployed on `https://static.rust-lang.org/rustup`.
+
+By switching between those two values, rustup effectively provides two "self
+update channels", making beta testing possible with `rustup self update`.
+
+Currently, rustup does one beta release followed by one official release for
+each version number in the increasing order. In other words, we don't release
+any `1.28.x` once the `1.29.0` beta release is out, and the latter is followed
+by the `1.29.0` stable release, and so on.
+
+## Bumping the version number
+
+The version number is registered in the `Cargo.toml` file of the project.
+
+The general principle for version numbers is that we always increment the
+_minor_ number unless:
+
+- A major incompatibility has been introduced in this release:
+  increment the _major_ number instead.
+- This release is a hotfix because the last one had a defect:
+  increment the _patch_ number instead.
+
+### Minor version bumps
+
+> NOTE: Rustup hasn't been doing major version bumps since a long time ago, but
+> if we ever do, the procedure for it should be similar to that of a minor one.
+
+A minor version bump should be performed immediately after the latest `X.Y.0`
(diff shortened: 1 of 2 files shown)


doc/dev-guide/src/release-process.md
```diff
@@ -1,78 +1,149 @@
...
+
+## Bumping the version number
+
+The version number is registered in the `Cargo.toml` file of the project.
+
+The general principle for version numbers is that we always increment the
+_minor_ number unless:
+
+- A major incompatibility has been introduced in this release:
+  increment the _major_ number instead.
```
ChrisDenton, line 26: Hm, this isn't really true in practice though? As noted further down. I know we've had discussions on whether we should or not and even if semver makes sense for binaries.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4805/5100b80375fade4ee4c7c0714c1132fec28319d9..1a554cc0bf36965056ed5e49b655603ac081d75b)*

doc/dev-guide/src/release-process.md
```diff
@@ -1,78 +1,149 @@
...
-        > - This release is a hotfix because the last one had a defect:
-        >   increment the _patch_ number instead.
-      - Update `Cargo.toml` and `download/Cargo.toml` to have that same new
-        version number, then run `cargo build` and review `Cargo.lock` changes.
-      - If all looks well, make a commit.
-   2. Update `CHANGELOG.md` accordingly if necessary.
-2. [b/r] After merging the PR made in step 1, in a separate PR:
-   1. Update `rustup-init.sh` so that:
+Rustup is distributed. Below is a list of things to be done in order to cut a
+new [b]eta release or an official [r]elease:
```
ChrisDenton: These steps look right to me atm but I usually find the real test is when I actually try following them for real 😉. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4805/5100b80375fade4ee4c7c0714c1132fec28319d9..1a554cc0bf36965056ed5e49b655603ac081d75b)*

doc/dev-guide/src/release-process.md
rami3l: @ChrisDenton This is inherited from the previous docs (it's at least the 3rd version since my arrival), and if you all feel okay for removing/adjusting just (this) line I'd love to do that :)

doc/dev-guide/src/release-process.md
rami3l: @ChrisDenton I'm actually following those steps (in the previous version) but please feel free to suggest if you have found anything that can be improved on the spot for the new version. Otherwise, as I have suggested, we can merge this first and refine the instructions afterwards.

doc/dev-guide/src/release-process.md
ChrisDenton: Thinking about, we should probably leave this to another PR. It might be worth resolving the conversation about semver first.

ChrisDenton (approved):
Ok, I'm happy with this!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4805/5100b80375fade4ee4c7c0714c1132fec28319d9..1a554cc0bf36965056ed5e49b655603ac081d75b)*

Issue #3639, opened 2024-01-16
Populate more `env.*` files under `$CARGO_HOME` regardless of platform


mwisnicki:
### Problem

It looks like `$HOME/.cargo/env` is not created on Windows.
This breaks shell scripts that depend on it.

### Steps

1. Download and run rust-init.exe
2. Check ~/.cargo/env existence

### Possible Solution(s)

Just generate it like you do on Linux. It's useful for people using native bash on Windows.

### Rustup version

```console
rustup 1.26.0 (5af9b9484 2023-04-05)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.75.0 (82e1608df 2023-12-21)`
```

### Installed toolchains

```console
Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\WDAGUtilityAccount\.rustup

stable-x86_64-pc-windows-msvc (default)
rustc 1.75.0 (82e1608df 2023-12-21)
```

rami3l:
@rbtcollins do you think that'll be useful? I personally don't use Windows for development so I guess I can't decide by myself this time, sorry...

bloodgain:
I mostly work from a Git Bash prompt when I have to use Windows. Rustup appears to correctly set the `PATH` variable for Windows, so `$HOME/.cargo/bin` ends up in my path, anyway. I can't say this doesn't break other scripts that assume the env file(s) exist, however, just that it doesn't require extra steps to use it from a Git Bash/Mintty CLI.

That said, all the various checks the init scripts and/or rustup appear to go through to determine whether to create the env files makes the existence of the files unpredictable. I ran into similar issues trying to get Rust installed in a Docker container for a GitLab runner and found several other issues that complain about `$HOME/.cargo/env` not being created. Of course, I can manually add it to the PATH, but the effects of rustup-init and install steps differ more than necessary between environments.

It would be better if the installers had consistent effects, and users can use manual workarounds for special cases. The env files are not large or risky to install. Better to just install them unless explicitly told not to with / as a side-effect of a flag like `--no-modify-path`, or perhaps if the file already exists (unless it's assume it should never be modified, in which case, that's on the user).

There's even an open issue about the fact that [Rustup _unconditionally_ assumes the env file exists](https://github.com/rust-lang/rustup/issues/2840). So even the Rust (meta) toolchain assumes it's there.

rami3l:
[...]

Issue #247, opened 2016-04-01
Tracking: Support for platformdirs (platform-specific system directories) such as XDG Paths


brson:
Many people would prefer cargo/rustup conform to various platform-specific standards for storing configuration. cc https://github.com/rust-lang/cargo/pull/2127

(This issue was originally about a different subject, but is now an XDG etc. thread, some comments may make less sense out of context).

Gyscos:
I may have missed the answer before, but why not use XDG specifications again? (`${XDG_DATA_HOME}/rustup`)

brson:
@Gyscos It's because [cargo doesn't](https://github.com/rust-lang/cargo/pull/2127). That's a change that all Rust tools need to make together.

ollie27:
> That's a change that all Rust tools need to make together.

Why?

The issue with cargo is that it was released using the wrong directories so now it's a lot harder to change. If rustup uses the correct directories from the start then there isn't a problem and seeing as is has to be moved anyway there isn't even any backwards compatibility issues for people using the beta.

On windows for example it should be something like `%LocalAppData%\rustup`.

brson:
> Why?

Consistency is important. If you know where one Rust tool puts its stuff you know where they all do.

ollie27:
Well I reckon changing both cargo and rustup at the same time is going to be harder than just changing cargo.

Is knowing where the .rustup folder is important?

brson:
@ollie27 I think it is yes. Like with .cargo, .git etc sometimes it's convenient to poke at internals. With rustup it seems to be important to know where the toolchains are stored for example.

Gyscos:
Putting things in the XDG folders don't make them opaque and inaccessible. Printing the path to downloaded toolchains (or showing this information in a `rustup show` command or anything) could help further.

Since cargo and rustup use different folders, I'd argue consistency is already lost (it's not like we had a `~/.rust` directory with everything related - nor that this would be a better solution). If we can afford to rename rustup's folder while keeping it separate, then it makes sense to lead by the example and directly take the correct destination.

kamalmarhubi:
On OS X it'd be nice to use `~/Library/Applicacation Support` (eventually?).

critiqjo:
[...]

PR #4806, opened 2026-04-10
Only show post-install instructions for currently installed shells

ryankopf:
When installing Rust on a new system, you get this ugly block of text listing all the possible ways to source the cargo environment variables. However, **we already detect which shells are available** - so the message should show the ones that exist on the users system.

Instead of:
```
To configure your current shell, you need to source the
corresponding `env` file under {cargo_home}.

source "~/.cargo/env.nu"  # For nushell
source "$HOME/.cargo/env.tcsh"  # For tcsh
. "$HOME/.cargo/env.ps1"        # For pwsh
source "$HOME/.cargo/env.xsh"   # For xonsh
```

If you have just bash and pwsh, for example, it will instead just show the relevant lines:

```
To configure your current shell, you need to source the
corresponding `env` file under {cargo_home}.

Consider running the right command for your shell (note the leading DOT):
. "$HOME/.cargo/env"            # For bash
. "$HOME/.cargo/env.ps1"        # For pwsh
```

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -84,9 +84,6 @@ use unix::{delete_rustup_and_cargo_home, do_add_to_path, do_remove_from_path};
 #[cfg(unix)]
 pub(crate) use unix::{run_update, self_replace};
 
-#[cfg(unix)]
-use self::shell::{Nu, UnixShell};
-
 #[cfg(windows)]
 mod windows;
 #[cfg(windows)]
@@ -446,35 +443,19 @@ but will not be added automatically."
 }
 
 #[cfg(not(windows))]
-macro_rules! post_install_msg_unix_source_env {
-    () => {
-        r#"To configure your current shell, you need to source
-the corresponding `env` file under {cargo_home}.
-
-This is usually done by running one of the following (note the leading DOT):
-    . "{cargo_home}/env"            # For sh/bash/zsh/ash/dash/pdksh
-    source "{cargo_home}/env.fish"  # For fish
-    source "{cargo_home_nushell}/env.nu"  # For nushell
-    source "{cargo_home}/env.tcsh"  # For tcsh
-    . "{cargo_home}/env.ps1"        # For pwsh
-    source "{cargo_home}/env.xsh"   # For xonsh
-"#
-    };
-}
-
-#[cfg(not(windows))]
 macro_rules! post_install_msg_unix {
     () => {
-        concat!(
-            r"# Rust is installed now. Great!
+        r"# Rust is installed now. Great!
 
 To get started you may need to restart your current shell.
 This would reload your `PATH` environment variable to include
 Cargo's bin directory ({cargo_home}/bin).
 
-",
-            post_install_msg_unix_source_env!(),
-        )
+To configure your current shell, you need to source the
+corresponding `env` file under {cargo_home}.
+
+Consider running the right command for your shell (note the leading DOT):
+{source_env_lines}"
     };
 }
 
@@ -494,15 +475,16 @@ Cargo's bin directory ({cargo_home}\\bin).
 #[cfg(not(windows))]
 macro_rules! post_install_msg_unix_no_modify_path {
     () => {
-        concat!(
-            r"# Rust is installed now. Great!
+        r"# Rust is installed now. Great!
 
 To get started you need Cargo's bin directory ({cargo_home}/bin) in your `PATH`
 environment variable. This has not been done automatically.
 
-",
-            post_install_msg_unix_source_env!(),
-        )
+To configure your current shell, you need to source
+the corresponding `env` file under {cargo_home}.
+
+Consider running the right command for your shell (note the leading DOT):
+{source_env_lines}"
     };
 }
 
@@ -661,19 +643,19 @@ pub(crate) async fn install(
         format!(post_install_msg_win!(), cargo_home = cargo_home)
     };
     #[cfg(not(windows))]
-    let cargo_home_nushell = Nu.cargo_home_str(cfg.process)?;
+    let source_env_lines = shell::build_source_env_lines(cfg.process);
     #[cfg(not(windows))]
     let msg = if no_modify_path {
         format!(
             post_install_msg_unix_no_modify_path!(),
             cargo_home = cargo_home,
-            cargo_home_nushell = cargo_home_nushell,
+            source_env_lines = source_env_lines,
         )
     } else {
         format!(
             post_install_msg_unix!(),
             cargo_home = cargo_home,
-            cargo_home_nushell = cargo_home_nushell,
+            source_env_lines = source_env_lines,
         )
     };
     md(&mut term, msg);
--- a/src/cli/self_update/shell.rs
+++ b/src/cli/self_update/shell.rs
@@ -73,6 +73,30 @@ fn enumerate_shells() -> Vec<Shell> {
     ]
 }
 
+/// Builds the shell source lines for the post-install message, showing only
+/// shells that are available on the current system. Shells sharing the same
+/// env file are grouped onto one line (e.g. sh/bash/zsh all use `env`).
+pub(crate) fn build_source_env_lines(process: &Process) -> String {
+    let mut groups = Vec::<(_, Vec<_>)>::new();
+    for shell in get_available_shells(process) {
+        let Ok(src) = shell.source_string(process) else {
+            continue;
+        };
+        if let Some(names) = groups
+            .iter_mut()
+            .find_map(|(s, names)| (*s == src).then_some(names))
+        {
+            names.push(shell.name());
+        } else {
+            groups.push((src, vec![shell.name()]));
+        }
+    }
+    groups
+        .into_iter()
+        .map(|(src, names)| format!("    {}  # For {}\n", src, names.join("/")))
+        .collect()
+}
+
 pub(crate) fn get_available_shells(process: &Process) -> impl Iterator<Item = Shell> + '_ {
     enumerate_shells()
         .into_iter()
@@ -83,6 +107,9 @@ pub(crate) trait UnixShell {
     // Detects if a shell "exists". Users have multiple shells, so an "eager"
     // heuristic should be used, assuming shells exist if any traces do.
     fn does_exist(&self, process: &Process) -> bool;
+
+    // Returns the display name of the shell, used in post-install messages.
+    fn name(&self) -> &'static str;
 
     // Gives all rcfiles of a given shell that Rustup is concerned with.
     // Used primarily in checking rcfiles for cleanup.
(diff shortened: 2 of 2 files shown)


src/cli/self_update/shell.rs
```diff
@@ -84,6 +84,9 @@ pub(crate) trait UnixShell {
     // heuristic should be used, assuming shells exist if any traces do.
     fn does_exist(&self, process: &Process) -> bool;
 
+    // Returns the display name of the shell, used in post-install messages.
+    fn name(&self) -> &'static str;
```
djc, line 112: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..1f4590f4368bd12f0886a0e8d53713270f4bd76d)*

Could this be an associated `const`?

src/cli/self_update.rs
```diff
@@ -445,36 +445,20 @@ but will not be added automatically."
     };
 }
 
-#[cfg(not(windows))]
-macro_rules! post_install_msg_unix_source_env {
```
djc, line 449: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..1f4590f4368bd12f0886a0e8d53713270f4bd76d)*

I think this is a good idea, but I wonder if we should also keep some kind of indication that there is support for these other shells.

Maybe there should be a page in the documentation that covers this (if there isn't one already), and we could link to it here?

src/cli/self_update/shell.rs
```diff
@@ -501,6 +536,27 @@ impl UnixShell for Xonsh {
     }
 }
 
+/// Builds the shell source lines for the post-install message, showing only
+/// shells that are available on the current system. Shells sharing the same
+/// env file are grouped onto one line (e.g. sh/bash/zsh all use `env`).
+pub(crate) fn build_source_env_lines(process: &Process) -> String {
```
djc: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..1f4590f4368bd12f0886a0e8d53713270f4bd76d)*

Nit: we follow a mostly top-down style, please move this before `get_available_shells()`.

src/cli/self_update/shell.rs
```diff
@@ -501,6 +536,27 @@ impl UnixShell for Xonsh {
...
+    let mut groups: Vec<(String, Vec<&'static str>)> = Vec::new();
```
djc: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..1f4590f4368bd12f0886a0e8d53713270f4bd76d)*

Style nit: instead of a type annotation, write this as `let mut groups = Vec::<(String, Vec<&'static str>)>::new();`.

src/cli/self_update/shell.rs
ryankopf: > _[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..1f4590f4368bd12f0886a0e8d53713270f4bd76d)_
> 
> Could this be an associated `const`?

I'm not 100% sure how a const works with a trait object, but I think since this code uses trait objects like Box<dyn UnixShell>, I think you can't access associated consts directly through the trait object

src/cli/self_update/shell.rs
ryankopf: Done :)

src/cli/self_update/shell.rs
ryankopf: Done :)

src/cli/self_update.rs
ryankopf: I thought about this after your comment. I looked at the code and there are not many places we put "just in case, look at the docs here: [url]".

I think we already rely very heavily on the "get_supported_shells" function and it seems to have served well over the years. If they have a shell that isn't discovered by "get_supported_shells" for some reason, it's .rc files or whatever it uses will not get updated by rustup in the first place, and we don't show a documentation message related to that issue. Like there's no "just in case we didn't find your shell, look here: [url]" message in what we're already doing.

So I think it's a better strategy to just make sure the "get_supported_shells" does a really good job of detecting whatever shells you have, which I think we have done since there I searched issues for "is:issue state:open get_supported_shells" and "shells missing" and there don't seem to be any complaints about it.

rami3l: @djc @ryankopf Actually a smarter way to do this would be to dump all `env` files in one place and tell the user "just source anything you want in this folder if you install any new shells later".

This is part of the plan (#3639), but for now I'd suggest postponing this because we are making the installation path change (https://github.com/rust-lang/rustup/issues/247) and the former change will be easier afterwards.

Your proposal is going exactly in the direction we originally wanted though, thanks again!

src/cli/self_update/shell.rs
rami3l: @djc Do we really need a complete notation here though? 🤔 The below seems enough:

```rs
let mut groups = Vec::<(_, Vec<_>)>::new();
```

src/cli/self_update/shell.rs
rami3l: @djc Unless we do the `enum_dispatch` trick this seems [impossible](https://doc.rust-lang.org/stable/error_codes/E0038.html#trait-contains-associated-constants), and I don't think it's worth it to put an extra dependency here.

src/cli/self_update/shell.rs
```diff
@@ -73,6 +73,27 @@ fn enumerate_shells() -> Vec<Shell> {
...
+/// shells that are available on the current system. Shells sharing the same
+/// env file are grouped onto one line (e.g. sh/bash/zsh all use `env`).
+pub(crate) fn build_source_env_lines(process: &Process) -> String {
+    let mut groups = Vec::<(String, Vec<&'static str>)>::new();
+    for shell in get_available_shells(process) {
+        let Ok(src) = shell.source_string(process) else {
+            continue;
+        };
+        if let Some(group) = groups.iter_mut().find(|(s, _)| *s == src) {
+            group.1.push(shell.name());
```
rami3l: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..d9073345510d911193193bc6993931064d3b848b)*

Nit: It's probably better to match and filter in one place:

```rs
        if let Some(names) = groups
            .iter_mut()
            .find_map(|(s, names)| (*s == src).then_some(names))
        {
            names.push(shell.name());
```

src/cli/self_update/shell.rs
djc: Okay, wasn't sure if we were using trait objects.

rami3l (commented):
Thanks for working on this!

Since this patch involves no snapshot updates, it is recommended to include a screenshot/terminal output snippet in addition to the existing PR description with the original output scenario (meaning with all possible shells) for demonstration. This should be relatively easy by temporarily modifying the relevant logic.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..d9073345510d911193193bc6993931064d3b848b)*

src/cli/self_update/shell.rs
ryankopf: Done

ryankopf:
> Thanks for working on this!
> 
> Since this patch involves no snapshot updates, it is recommended to include a screenshot/terminal output snippet with the original output scenario (meaning with all possible shells) for demonstration. This should be relatively easy by temporarily modifying the relevant logic.

See below.

rustup-old

rust-new

Perhaps for clarity, the message could be updated, although I am not certain it is necessary. Perhaps:

```

. "$HOME/.cargo/env"  # For sh/ash/dash/pdksh/bash
```

Could say:

```
To configure your current shell, you need to source the
corresponding `env` file under {cargo_home}. Consider running the
corresponding command for your shell (note the leading DOT):
. "$HOME/.cargo/env"  # For sh/ash/dash/pdksh/bash
```

I think this might strike a nice balance, leaving out the textual implication that "we are certain this is your shell" from the message, while providing with probably 99+% chance that it is the right command for your shell.

rami3l (approved):
Yeah, your suggestion on shortening the help message looks good, let's do that!

Although it'd be cooler if we make the lines not super long. Something like:

```
the corresponding env file under {cargo_home}.
Consider running the right command for your shell (note the leading DOT):
. "$HOME/.cargo/env"  # For sh/ash/dash/pdksh/bash
```

Note we cannot wrap the second line because {cargo_home} can be super long.

Please don't forget to update the PR description afterwards, and once you have fixed the compilation error in the CI we are done :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..a35a9bc648ce7265282f3fa7f7d92f39ff1d26cd)*

djc:
I think it would be more readable to keep an empty line between the help text and the potential shell invocations.

ryankopf:
> Yeah, your suggestion on shortening the help message looks good, let's do that!
> 
> Although it'd be cooler if we make the lines not super long. Something like:
> 
> ```
> To configure your current shell, you need to source
> the corresponding env file under {cargo_home}.
> Consider running the right command for your shell (note the leading DOT):
> . "$HOME/.cargo/env"  # For sh/ash/dash/pdksh/bash
> ```
> 
> Note we cannot wrap the second line because {cargo_home} can be super long.
> 
> Please don't forget to update the PR description afterwards, and once you have fixed the compilation error in the CI we are done :)
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4806/5100b80375fade4ee4c7c0714c1132fec28319d9..a35a9bc648ce7265282f3fa7f7d92f39ff1d26cd)_



> I think it would be more readable to keep an empty line between the help text and the potential shell invocations.

Both done. I changed the text to the suggested "Consider running the right command for your shell (note the leading DOT):" and I put the newline back before it, because you're right, it does seem to get too crowded without.

Thanks for the note regarding {cargo_home} possibly being long, it didn't cross my mind at the time.

I have updated the description with the PR update texts.

rami3l:
Thanks again @ryankopf for your work! Let's merge this first and see how it goes...

rami3l (approved):

PR #4807, opened 2026-04-10
Use `cc-rs` to detect the default linker, instead of assuming `cc`

jamesmunns:
On hosts like `illumos`, `cc` is not present in the path, leading to misleading warnings like:

```
warn: no default linker (`cc`) was found in your PATH
warn: many Rust crates require a system C toolchain to build
```

However, there is a linker present: `gcc`. Since this warning at install time is to help ensure the system is prepared to do build.rs work, which typically uses `cc-rs`, and has this additional metadata for targets, this PR uses the `cc-rs` infra directly to determine the correct binary to search for. If `cc-rs` has nothing to say about our particular host, we fall back to the existing "just look for `cc` in PATH" logic.

There's a *little* unfortunate song and dance required, as `cc-rs` doesn't make this information directly available (requiring you to create a `Build`, which then requires some information like opt-level and host/target triples to get the correct tool path), if this isn't palatable, I can follow up with `cc-rs` to see if they want to make this metadata a bit more directly obtainable.

I'm not sure the best way to directly test the rustup install process using this branch, if there are docs somewhere I'm happy to run and verify the warning is skipped now on illumos targets.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -828,20 +828,47 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
 
 #[cfg(unix)]
 fn warn_if_default_linker_missing(process: &Process) {
-    // Search for `cc` in PATH
-    if let Some(path) = process.var_os("PATH") {
-        let cc_binary = format!("cc{}", EXE_SUFFIX);
+    // Search for linker in PATH
+    let Some(path) = process.var_os("PATH") else {
+        warn!("unable to search PATH for a default linker");
+        warn!("many Rust crates require a system C toolchain to build");
+        return;
+    };
 
-        for mut p in env::split_paths(&path) {
-            p.push(&cc_binary);
-            if p.is_file() {
-                return;
-            }
-        }
+    // If we have the host triple, attempt to determine the CC/linker path that
+    // `cc-rs` would use, as this is *usually* why we need a linker to invoke.
+    //
+    // Doing it this way allows us to correctly diagnose quirky systems like
+    // solaris and illumos that use `gcc` rather than `cc` for historical reasons.
+    let cc_tool = TargetTriple::from_host(process).and_then(|triple| {
+        // Fill in some dummy settings for `Build`/`Tool` to be able to properly
+        // give us the metadata we want
+        cc::Build::new()
+            .opt_level(0)
+            .target(&triple)
+            .host(&triple)
+            .try_get_compiler()
+            .ok()
+    });
+
+    let cc_binary = if let Some(cc_tool) = &cc_tool {
+        Cow::Borrowed(cc_tool.path())
+    } else {
+        // If we don't get info from cc-rs, fall back to just looking for the literal `cc`.
+        Cow::Owned(format!("cc{EXE_SUFFIX}").into())
+    };
+
+    // Search the path for the selected binary
+    let found = env::split_paths(&path).any(|mut p| {
+        p.push(&cc_binary);
+        p.is_file()
+    });
+
+    if !found {
+        let bin_disp = cc_binary.display();
+        warn!("no default linker (`{bin_disp}`) was found in your PATH");
+        warn!("many Rust crates require a system C toolchain to build");
     }
-
-    warn!("no default linker (`cc`) was found in your PATH");
-    warn!("many Rust crates require a system C toolchain to build");
 }
 
 fn install_bins(process: &Process) -> Result<()> {
--- a/tests/suite/cli_inst_interactive.rs
+++ b/tests/suite/cli_inst_interactive.rs
@@ -684,7 +684,7 @@ async fn install_warns_if_default_linker_missing() {
         .is_ok()
         .with_stderr(snapbox::str![[r#"
 ...
-warn: no default linker (`cc`) was found in your PATH
+warn: no default linker ([CC_TOOL]) was found in your PATH
 warn: many Rust crates require a system C toolchain to build
 ...
 "#]]);
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ test = ["dep:snapbox", "dep:walkdir", "clap-cargo/testing_colors"]
 anstream = "1"
 anstyle = "1.0.11"
 anyhow = "1.0.69"
+cc = "1"
 cfg-if = "1.0"
 chrono = { version = "0.4", default-features = false, features = ["std"] }
 clap = { version = "4", features = ["derive", "wrap_help", "string"] }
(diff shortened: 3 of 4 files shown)


src/cli/self_update.rs
```diff
@@ -828,19 +828,43 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+        let cc_binary_opt: Option<String> = TargetTriple::from_host(process).and_then(|triple| {
+            let mut build = cc::Build::new();
+
+            // Fill in some dummy settings for `Build`/`Tool` to be able to properly
+            // give us the metadata we want.
+            build.opt_level(0);
+            build.target(&triple);
+            build.host(&triple);
+
+            build
```
djc: Suggest we actually chain these method calls?

src/cli/self_update.rs
```diff
@@ -828,19 +828,43 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+                .try_get_compiler()
+                .ok()
+                .and_then(|tool| tool.path().to_str().map(str::to_string))
+        });
+
+        // If we don't info from cc-rs, fall back to just looking for the literal `cc`.
```
djc: "don't info"?

src/cli/self_update.rs
```diff
@@ -828,19 +828,43 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+    // Search for linker in PATH
     if let Some(path) = process.var_os("PATH") {
-        let cc_binary = format!("cc{}", EXE_SUFFIX);
-
+        // If we have the host triple, attempt to determine the CC/linker path that
+        // `cc-rs` would use, as this is *usually* why we need a linker to invoke.
+        //
+        // Doing it this way allows us to correctly diagnose quirky systems like
+        // solaris and illumos that use `gcc` rather than `cc` for historical reasons.
+        let cc_binary_opt: Option<String> = TargetTriple::from_host(process).and_then(|triple| {
```
djc: Nit: don't think we need the type annotation here?

djc (commented):
Looks okay to me modulo a few nits, thanks!

(Please squash your changes into a single commit, we prefer to rebase clean history.)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4807/5100b80375fade4ee4c7c0714c1132fec28319d9..6f0e432901805cf77082e202e3a2de29e1fa39d4)*

src/cli/self_update.rs
jamesmunns: Fixed

src/cli/self_update.rs
jamesmunns: Corrected comment

src/cli/self_update.rs
jamesmunns: Removed, I put it there while futzing with the other method chaining :)

jamesmunns:
@djc thanks for the review! Fixed nits, looking at how to re-bless the snapshot tests now, feel free to handle it, or I can look at the devguide.

djc:
I think the test failures give a hint of how to update the snapshots?

jamesmunns:
It's not immediately clear to me how to run the test suite directly. `cargo test --all` does not run without errors on MacOS, at least. The error says:

```
[4](https://github.com/rust-lang/rustup/actions/runs/24247005720/job/70795886044#step:14:1145)
thread 'suite::cli_inst_interactive::install_warns_if_default_linker_missing' (122418) panicked at src/test/clitools.rs:230:9:

---- expected: tests/suite/cli_inst_interactive.rs:685:22
++++ actual:   In-memory
   1    1 | ...
   2    2 | warn: no default linker (`cc`) was found in your PATH
   3      - warn: many Rust crates require a system C toolchain to build
        3 + warn: unable to search PATH for a default linker
   4    4 | ...∅

Update with SNAPSHOTS=overwrite
```

So, once I get tests running, I will try setting that env var. Currently trying to run via `bash ci/run.bash`.

edit: the error above is from CI, I get different errors when building tests on macos, unrelated to the failing snapshot tests.

djc:
You need `--features test` I think.

jamesmunns:
@djc CI is now green! Took a bit more time to use redaction to paper over the now More Helpful error that tells you what toolchain we look for.

minor note: in terms of CI optimization, it might be worth cancelling jobs on first failure, or sequencing the more expensive macos/windows ci runners after the cheaper linux ones pass, or at least cancelling the old run after a new commit is pushed. Not sure if rustup gets enough PRs to make the utilization worth caring about, but there are a couple of ~30m jobs run on mac/win on every commit, and they don't stop if some other error is hit.

tests/suite/cli_inst_interactive.rs
```diff
@@ -684,8 +684,7 @@ async fn install_warns_if_default_linker_missing() {
         .is_ok()
         .with_stderr(snapbox::str![[r#"
 ...
-warn: no default linker (`cc`) was found in your PATH
-warn: many Rust crates require a system C toolchain to build
```
rami3l, line 688: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4807/5100b80375fade4ee4c7c0714c1132fec28319d9..efa509f43691cad6e8d1cdece168b7d9c1375f8e)*

This line seems to have been deleted by mistake 🙏

src/cli/self_update.rs
```diff
@@ -828,19 +828,41 @@ fn current_install_opts(opts: &InstallOpts<'_>, process: &Process) -> String {
...
+                .opt_level(0)
+                .target(&triple)
+                .host(&triple)
+                .try_get_compiler()
+                .ok()
+                .and_then(|tool| tool.path().to_str().map(str::to_string))
+        });
+
+        // If we don't get info from cc-rs, fall back to just looking for the literal `cc`.
+        let cc_binary = cc_binary_opt.unwrap_or_else(|| format!("cc{}", EXE_SUFFIX));
```
rami3l: Nit: There's no need to convert `Path` to `String`, you'll need to convert back anyway. I'd suggest the following instead:

```suggestion
        let cc_tool = TargetTriple::from_host(process).and_then(|triple| {
            // Fill in some dummy settings for `Build`/`Tool` to be able to properly
            // give us the metadata we want
            cc::Build::new()
                .opt_level(0)
                .target(&triple)
                .host(&triple)
                .try_get_compiler()
                .ok()
        });

        let cc_binary = if let Some(cc_tool) = &cc_tool {
            Cow::Borrowed(cc_tool.path())
        } else {
            // If we don't get info from cc-rs, fall back to just looking for the literal `cc`.
            Cow::Owned(format!("cc{EXE_SUFFIX}").into())
        };
```

djc: If we're going to restructure like this, there might be value in inverting the initial if and yeeting out consecutive error cases (with log messages) to reduce rightward drift and make things more linear.

djc (approved):

djc:
Thanks, looks good!

PR #4809, opened 2026-04-12
Upgrade to rustls-platform-verifier 0.7

djc:


rami3l:
@djc Thanks!

PS: I've noticed that we are having two copies of the platform verifier in the dependency tree: the one that comes with reqwest in theory remains unused, right?

rami3l (approved):

djc:
> PS: I've noticed that we are having two copies of the platform verifier in the dependency tree: the one that comes with reqwest in theory remains unused, right?

Yup -- it looks like there is no easy/supported way to disable that dependency.

PR #4813, opened 2026-04-15
Add riscv64gc-unknown-linux-musl

eshattow:
riscv6gc-unknown-linux-musl is now a Tier 2 target with host tools. Add support for it in rustup.

Additionally, add some missing riscv64gc-unknown-linux-gnu entries in rustup tree. The glibc dist build target name was renamed from dist-riscv64-linux to dist-riscv64-linux-gnu as part of the addition of musl dist build target.


rami3l:
@eshattow Hello, and sorry for the late reply 🙏 

Are there any updates on this matter? Is this PR waiting for something to happen? I'd love to help if you need anything to land it. Many thanks in advance 🙇

eshattow:
Thanks @rami3l no actionables yet, I have been looking at the powerpc promotion (to Tier 2) PR as an example. Welcome any advice or guidance :-) I have an amd64 desktop to run the build, also StarFive JH7110 (riscv64gc) and SpacemiT K3 (riscv64a23) boards to build or test with. Admittedly I don't know much about rust or the build system previously this is from my viewpoint something that needs done to move along the official rust docker images that much CI/CD depends on in the GitHub-using community; learning as I go.

One point I am unclear about is if we have RISE RISC-V runners or not configured for the GitHub account? Should I be writing this as a cross-compiled build with amd64 or instead natively with the ubuntu-24.04-riscv runner; or is it a blend of both for build and test?

rami3l:
@eshattow Hello, thanks again for the update! I don't think we have riscv runners here at all, but in principle you won't need that to achieve Tier 2. So just adding a cross compilation target on Linux would be enough.

For a very recent example you can see https://github.com/rust-lang/rustup/pull/4908.

eshattow:
Pending rollup of completed MCP 982 implementation.

eshattow:
Patches in upstream Linux for detection and reporting of RVA20 or RVA23 microarchitecture runtime environment (for example https://lore.kernel.org/linux-riscv/20260701-rva23u64-hwprobe-v2-v5-0-2c61f94a695a@gmail.com/ ) have not yet landed, so I remain here in rustup context with the assumption of RVA20 profile that is the most compatible. This will need some changes if/when RVA23 targets are promoted; I would tee this up to be prepared but there's not an agreed upon way to do it yet.

eshattow:
@rami3l writes:
> @eshattow Hello, thanks again for the update! I don't think we have riscv runners here at all, but in principle you won't need that to achieve Tier 2. So just adding a cross compilation target on Linux would be enough.
> 
> For a very recent example you can see #4908.

Thanks for that suggestion! I think I have implemented this as described, also see for reference MCP 982 tracking issue rust-lang/rust#156191 and the merge PR implementing that promotion to "with Tools" rust-lang/rust#158766

The `uname -m` reporting "riscv64" on RVA20 and RVA23 profiles may change in future but it is only proposed (elsewhere) to change the reporting for RVA23 and no changes to RVA20, RVA22. The difference of the profiles is most prominently the vector extensions mandated beginning with RVA23 and so a big performance impact there;  out of scope for this pull request so we'll just assume if the host is reporting machine "riscv64" to fetch the Tier 2 build of glibc or musl target for RVA20 (riscv64gc).

Let me know what you think, thanks again! -E

doc/user-guide/src/installation/other.md
```diff
@@ -163,6 +163,8 @@ You can manually download `rustup-init` for a given target from
   - [sha256 file](https://static.rust-lang.org/rustup/dist/powerpc64le-unknown-linux-gnu/rustup-init.sha256)
 - [powerpc64le-unknown-linux-musl](https://static.rust-lang.org/rustup/dist/powerpc64le-unknown-linux-musl/rustup-init)
   - [sha256 file](https://static.rust-lang.org/rustup/dist/powerpc64le-unknown-linux-musl/rustup-init.sha256)
+- [riscv64gc-unknown-linux-musl](https://static.rust-lang.org/rustup/dist/riscv64gc-unknown-linux-musl/rustup-init)
```
rami3l, line 168: I'm surprised to find the seeming absence of `riscv64gc-unknown-linux-gnu` in this list... Maybe we can take the opportunity to add it here?

ci/actions-templates/linux-builds-template.yaml
```diff
@@ -44,6 +44,7 @@ jobs: # skip-main skip-pr skip-stable
           - armv7-linux-androideabi        # skip-pr skip-main
           - x86_64-linux-android           # skip-pr skip-main
           - riscv64gc-unknown-linux-gnu    # skip-pr skip-main
+          - riscv64gc-unknown-linux-musl   # skip-pr skip-main
```
rami3l, line 47: Please don't forget to run `./ci/actions-templates/gen-workflows.sh` again to sync the actual CI files :)

rami3l (approved):
LGTM modulo some tiny nits, thanks!

As for the remaining concerns esp. regarding `uname -m` I believe we can address them when this target is actually enabled. For now keeping the most conservative/compatible approach sounds reasonable.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4813/10f348abb608c00133ff48be104fbed74cd7daac..7423b04ed7b813fc6428d66cadc0691e1975b802)*

doc/user-guide/src/installation/other.md
eshattow: Will do, thanks!

PR #4819, opened 2026-04-19
Improve error message for incomplete toolchains

Inconnu08:
Improve the error message shown when `rustup` encounters a toolchain that is missing its manifest.

Interrupted installations can leave a partially installed toolchain behind. Rustup already improves the recovery experience when reinstalling such toolchains (#4725), but users can still encounter a confusing "Missing manifest" error when interacting with the incomplete toolchain before retrying installation.

- Added a test that removes the manifest file from an installed toolchain and verifies the improved error message

Related to #4724
Builds on the interrupted-install UX improvements from #4725

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -71,7 +71,11 @@ pub enum RustupError {
     DownloadingFile { url: Url, path: PathBuf },
     #[error("could not download file from '{url}' to '{}'", .path.display())]
     DownloadNotExists { url: Url, path: PathBuf },
-    #[error("Missing manifest in toolchain '{}'", .0)]
+    #[error(
+        "missing manifest in toolchain '{0}'\n\
+     help: this may happen if the toolchain installation was interrupted\n\
+     help: try reinstalling or updating the toolchain"
+    )]
     MissingManifest(ToolchainDesc),
     #[error("server sent a broken manifest: missing package for component {0}")]
     MissingPackageForComponent(String),
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -4144,3 +4144,35 @@ installed targets:
         .with_stderr(snapbox::str![[""]])
         .is_ok();
 }
+
+#[tokio::test]
+async fn missing_manifest_shows_reinstall_help() {
+    let cx = CliTestContext::new(Scenario::SimpleV2).await;
+
+    cx.config
+        .expect(["rustup", "toolchain", "install", "nightly"])
+        .await
+        .is_ok();
+
+    let manifest_path = cx
+        .config
+        .rustupdir
+        .join("toolchains")
+        .join(format!("nightly-{}", this_host_triple()))
+        .join("lib")
+        .join("rustlib")
+        .join("multirust-channel-manifest.toml");
+
+    fs::remove_file(&manifest_path).unwrap();
+
+    cx.config
+        .expect(["rustup", "component", "list", "--toolchain", "nightly"])
+        .await
+        .with_stderr(snapbox::str![[r#"
+error: missing manifest in toolchain 'nightly-[HOST_TRIPLE]'
+help: this may happen if the toolchain installation was interrupted
+help: try reinstalling or updating the toolchain
+
+"#]])
+        .is_err();
+}


tests/suite/cli_rustup.rs
```diff
@@ -4144,3 +4146,35 @@ installed targets:
...
+
+    fs::remove_file(&manifest_path).unwrap();
+
+    cx.config
+        .expect(["rustup", "component", "list", "--toolchain", "nightly"])
+        .await
+        .with_stderr(str![[r#"
+error: toolchain 'nightly-[HOST_TRIPLE]' appears to be incomplete
+help: this may happen if the toolchain installation was interrupted
+help: run `rustup toolchain install nightly-[HOST_TRIPLE]` to reinstall
```
rami3l, line 4176: Suggest adding an extra step to see if the suggestion really works.

rami3l (commented):
Thanks for working on this!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4819/d32f88bba027fda2efad7fc67ba34a03fa481fad..96cbb9e2bb137ae2b32b994498b20d8bf5e87cef)*

tests/suite/cli_rustup.rs
Inconnu08: Edit:  I tried verifying the suggested reinstall step.

In the test environment, `rustup toolchain install nightly` reports the toolchain as unchanged and does not repair the missing-manifest state.

However, when testing manually with a real dist server, I noticed that `rustup component list --toolchain nightly` actually triggers a re-sync and repairs the toolchain automatically.

So recovery currently depends on the command used, and `toolchain install` does not explicitly detect and repair incomplete states.

Given this, I think the current suggestion may be misleading, so I’m considering adjusting the help text to be less prescriptive and soften it.

Inconnu08: So after further testing, I re-ran the reproduction using the exact same dated toolchain (`nightly-2026-01-01`) to avoid channel updates.

In that case, both:
- `rustup component list --toolchain nightly-2026-01-01`
- `rustup toolchain install nightly-2026-01-01`

fail to recover the toolchain, and `toolchain install` still reports it as unchanged.

So it seems the issue is not just command inconsistency, but that `toolchain install` does not detect or repair an incomplete toolchain in this state.

I’ll open a follow-up issue to track this behavior, and keep this PR focused on just improving the diagnostic and softening the help text.

tests/suite/cli_rustup.rs
```diff
@@ -15,6 +15,8 @@ use rustup::{
     utils::raw,
 };
 
+use snapbox::str;
```
djc: Nit: putting this is in a block below the `rustup` imports looks a little wrong. Not sure whether other test modules treat `rustup` as external or internal but depending on that either `snapbox` should be in a block by itself between the `std` and `rustup` blocks or in the same block as `rustup`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4819/247b429ef381550deb13cc3e30cf40cdb9e6e00c..787f6ff2d37cdbbcdf4804fd29d1a17311ff271d)*

djc (approved):

src/errors.rs
```diff
@@ -71,7 +71,11 @@ pub enum RustupError {
     DownloadingFile { url: Url, path: PathBuf },
     #[error("could not download file from '{url}' to '{}'", .path.display())]
     DownloadNotExists { url: Url, path: PathBuf },
-    #[error("Missing manifest in toolchain '{}'", .0)]
+    #[error(
+        "toolchain '{0}' appears to be incomplete\n\
```
djc: "appears to be incomplete" is a lot more vague than "missing manifest". Does that represent actual usage? If this is really about manifests only, I think keeping the precision is probably useful.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4819/247b429ef381550deb13cc3e30cf40cdb9e6e00c..787f6ff2d37cdbbcdf4804fd29d1a17311ff271d)*

tests/suite/cli_rustup.rs
rami3l: @Inconnu08 Ah, I see. In that case it's worth going for https://github.com/rust-lang/rustup/issues/4825. Many thanks for the investigation :)

PS: Rethinking about it, we can indeed have a vague wording to begin with, it's better than having nothing 🙏

tests/suite/cli_rustup.rs
rami3l: @djc Previously you have suggested qualifying all `snapbox` calls like `snapbox::str` so there doesn't seem to be a lot of `snapbox` imports so far.

Otherwise they seem to be in the same group: https://github.com/rust-lang/rustup/blob/622ab2339e57ab9c2b9f2d527607e1de61595a52/tests/suite/cli_rustup_ui.rs#L4-L6

@Inconnu08 My suggestion would be to qualify this call.

src/errors.rs
Inconnu08: Actually, I agree. I’ve updated the message to keep the specific “missing manifest” wording while retaining the added guidance.

rami3l:
Let's merge this first and see how it goes, I guess https://github.com/rust-lang/rustup/issues/4825 can wait until later :)

Issue #4715, opened 2026-02-12
`multirust-channel-manifest.toml` should have reproducible hash


loynoir:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

Try to make `nix` package using `fixed-output-derivation`.

But `multirust-channel-manifest.toml` have different hash, seems caused by `HashMap`, leads to impossible to declare nix FOD `outputHash`.

### Steps

```console
$ rustup toolchain install
```

```toml
...

[pkg.cargo]
version = "0.96.0-nightly (efcd9f586 2026-01-23)"

...

[pkg.rust-analyzer-preview]
version = "0.0.0-nightly"

...
```

```console
$ rustup toolchain install
```

```toml
...

[pkg.rust-analyzer-preview]
version = "0.0.0-nightly"

...

[pkg.cargo]
version = "0.96.0-nightly (efcd9f586 2026-01-23)"

...
```

### Possible Solution(s)

For reproducible `multirust-channel-manifest.toml`, use btree map instead of hash map

### Rustup version

```console
channel = "nightly-2026-01-28"
```

### Installed toolchains

```console
channel = "nightly-2026-01-28"
```

### OS version

```console
archlinux with nix
```

djc:
> For reproducible `multirust-channel-manifest.toml`, use btree map instead of hash map

Seems reasonable. Would you be able to submit a PR?

rami3l:
@djc There is nothing we can do in the rustup repo unfortunately, since the channel manifest is downloaded directly from the release server and previously rustup's workings don't rely on its checksum (except for verification purposes during the download). This doesn't mean it cannot be rectified in the future though, but it has to be done on a per-release server basis.

Interestingly, this is also causing problems for my ongoing proposal, so I hacked around it:

> We cannot use the checksum of the manifest file here, unfortunately, because 1.92.0 and 1.92 can have the same manifest with different checksums, for example. That non-determinism is a reality in our release server and there is nothing we can do to change what has been published in the past.
_https://hackmd.io/@-BxJthnhTQWCxHdTwpNSWw/BkAcWKRIZx#fn2_

[...]

PR #4822, opened 2026-04-20
test(dist/manifest): use the reordered fixture in `manifest_serialized_with_sorted_keys`

cachebag:
The regression test for #4715 was meant to parse two TOML manifests with packages declared in different orders and assert that their serialized forms are identical, proving keys are emitted in deterministic (sorted) order. However, `EXAMPLE_REORDERED` was `include_str!`-ing the same file as `EXAMPLE`, so the assertion held trivially and the test never actually exercised the sort-on-serialize behavior. 

This PR points it at the existing `channel-rust-nightly-example-reordered.toml` fixture so the test can catch any future regression that breaks deterministic manifest serialization.

### DIFF
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -656,7 +656,7 @@ mod tests {
     static EXAMPLE: &str = include_str!("manifest/tests/channel-rust-nightly-example.toml");
     // Same manifest as above, but with the packages in a different order.
     static EXAMPLE_REORDERED: &str =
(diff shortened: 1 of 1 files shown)


rami3l (approved):
Must've been a typo on my end when introducing the new fixture; nice catch!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4822/d093663df81724d759b422e982e71d51f6fc9a68..1fc7ab6c33ae03aaa9d6fe656e5d74ea55c62254)*

Issue #3390, opened 2023-06-22
Nightly installation failures


JoshLind:
### Problem

Hi,

We have a bunch of continuously running CI/CD jobs that install the nightly version of rust (specifically, so that we can use nightly `cargo fmt` across our codebase). However, recently, we've seen these jobs begin to flake when trying to install the nightly builds. For example, our scripts run these commands to install `rust nightly`:

```
rustup toolchain install nightly
rustup component add rustfmt --toolchain nightly
```

But, these have begun flaking recently (e.g., the last few days) with errors:
```
+ rustup toolchain install nightly
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: update not yet available, sorry! try again later
error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installable
```

We have recently upgraded to use the rust `1.70.0` toolchain (which I suspect is when the flakes started). I know there's a couple options we have, e.g., fix our nightly build to a more stable date and then rename the toolchain from `nightly-DATE` to `nightly`, but we thought it was worth calling out the issue and seeing if folks were aware.

### Steps

1. Use the 1.70.0 rust toolchain.
2. Run a script every few hours that calls:
```
rustup toolchain install nightly
rustup component add rustfmt --toolchain nightly
```
3. See that some jobs flake with the above error?

### Possible Solution(s)

Update all our scripts to fix the nightly installation to a specific date/build and use that?

### Notes

In the ideal case we'd have an option in `rustup` to install the most recent nightly that works, e.g., `rustup install nightly --any-recent-version`, which could essentially just loop back through all the builds until one of them works, but I recognize this is likely a niche ask 😄 

### Rustup version

```console
info: This is the version for the rustup toolchain manager, not the rustc compiler.
rustup 1.26.0 (5af9b9484 2023-04-05)
info: The currently active `rustc` version is `rustc 1.70.0 (90c541806 2023-05-31)`
```

### Installed toolchains

```console
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/runner/.rustup
```

rbtcollins:
Not a Rustup problem - see https://github.com/dtolnay/rust-toolchain/issues/88

I'm not sure what rust-lang project to move this onto, but I'm inclined to close this particular report - even though rustup is the visible face of it, it isn't the driving factor.

jdno:
Hi folks,

[...]

PR #4830, opened 2026-04-24
fix(dist): propagate v2 manifest checksum failure

cachebag:
Our nightly CI in a project I am working on started failing intermittently after we bumped MSRV. a bunch of agents were still on the old toolchain even though rustup update stable was reporting unchanged and exiting 0.

On one agent, rustc --version was a full release behind what our Artifactory mirror was serving. See below:

```log
info: syncing channel updates for stable-x86_64-unknown-linux-gnu
warn: checksum failed for 'https://blahblahblah/rust-dist/dist/channel-rust-stable.toml',
      expected: '7f3c91ad024a1cbe1e8915f35aff0a3fbbdf8d293ad48ab8f31e3b0440c581f9',
      calculated: '821ff14e4c4a1cbe1e8915f35aff0a3fbbdf8d293ad48ab8f31e3b0440c581f9'
info: this might indicate an issue with the third-party release server 'https://blahblahblah/rust-dist/dist'

  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.94.0 (4a4ef493e 2026-03-02)
```

This PR drops the arm in `try_update_from_dist_` so the error propagates.

### DIFF
--- a/src/dist/manifestation/tests.rs
+++ b/src/dist/manifestation/tests.rs
@@ -5,6 +5,7 @@
 use std::{
     collections::HashMap,
     env, fs,
+    io::Write,
     path::{Path, PathBuf},
     str::FromStr,
     sync::Arc,
@@ -14,6 +15,7 @@ use anyhow::{Result, anyhow};
 use url::Url;
 
 use crate::{
+    config::Cfg,
     dist::{
         DEFAULT_DIST_SERVER, Profile, TargetTuple, ToolchainDesc,
         download::{DownloadCfg, DownloadTracker},
@@ -26,8 +28,10 @@ use crate::{
     errors::RustupError,
     process::TestProcess,
     test::{
+        Env, RustupHome,
         dist::*,
         mock::{MockComponentBuilder, MockFile, MockInstallerBuilder},
+        test_dir,
     },
     utils::{self, raw as utils_raw},
 };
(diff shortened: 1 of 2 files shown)


djc (approved):
Oh, good catch. Any chance you can write up a test for this so that we don't regress it?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4830/6a80874687c3ff710007721c6520164b5565f7fb..fa00dd13912abf3676bea004ec08a1f178a01e83)*

cachebag:
Reposting for grammar/spelling errors:

Worth noting that this also tightens behaviour against static.rust-lang.org itself, where the v2 swallow was historically there to absorb brief mismatches during releases (cf. https://github.com/rust-lang/rustup/issues/3390). 

 If you prefer, I think it would make sense to gate this on `dl_cfg.tmp_cx.dist_server != DEFAULT_DIST_SERVER`.

rami3l:
@cachebag Thanks for your investigation!

As you have investigated already, I think the original behavior is somewhat intentional in that if the manifest update failed with checksum mismatch, it usually indicates an inconsistent state within the release server itself and in most of the cases it should be considered that the new Rust release is still in progress in the official release server.|

> I think it would make sense to gate this on dl_cfg.tmp_cx.dist_server != DEFAULT_DIST_SERVER.

... and no, I don't want to make official release server that different in this sense; if the release server were to cause rustup to fail, the type of release server shouldn't have mattered.

Swallowing the error in that routine completely is absolutely not ideal, but this does looks like the right thing to do so maybe we can try it first and see how it goes... In the worst of the cases we can downgrade that to a warning 🤔

ChrisDenton:
Could we talk to infra about if such inconsistent states still happen and if they could be mitigated on their side somehow?

rami3l:
> Could we talk to infra about if such inconsistent states still happen and if they could be mitigated on their side somehow?

@ChrisDenton I think we can try (cc @marcoieni) but since the error has been swallowed since a long time ago I doubt we can get any real feedback since.

djc:
Could/should we retry it a few times and really fail if it still fails?

rami3l:
@djc I think we could add retries but if this is mostly not a client-side problem then it should not help very much...

Thinking about it, if each time we get the same "wrong" checksum then we do get more sure that this is a server-side problem, so retrying does have some benefits after all.

cachebag:
@rami3l 
We already have retry logic in `download_file_with_resume` for network failures; I could wrap the `download_and_check` call in a small retry loop?

Also, I just want to mention that retries wouldn't have helped my case. The mirror was consistently serving stale content, so all retries would fail identically.

rami3l:
@cachebag I suggest that we wait for a bit and see how the infra people relies.

IMHO as you said this is most likely a server-side problem so it is hard to say whether issuing multiple retries will actually turns a failure into a success, what I added was just that the user could be more sure of the server-side problem if they happen to get the same shasum every time.

rami3l:
Without any updates from their side I think we should merge it first and test drive on our end. Thanks again for the work @cachebag!

rami3l (approved):

rami3l:
@cachebag `Cfg::from_env()` now has one more argument to pass...

Would you mind resolving the test conflicts? Many thanks in advance 🙏

Issue #4826, opened 2026-04-23
"Triple" should be renamed to "Tuple" to reflect the new terminology


rami3l:
> These are changes I expect would not break anything.
> 
> For context, main Rust repo has started [moving to new terminology](https://github.com/rust-lang/rust/pull/125579/changes/a26450cf81d67d68d3c6157579f8d968349129e7)
>
_originally by @tshepang in https://github.com/rust-lang/rustup/pull/4658#issue-3758532941_

## Task List
- https://github.com/rust-lang/rustup/pull/4658
- https://github.com/rust-lang/rustup/pull/4668
- #4743
- #4827
- https://github.com/rust-lang/rustup/pull/4834
- #4843
- #4849 
- https://github.com/rust-lang/rustup/pull/4852

rami3l:
My sincere appreciation to @tshepang for piloting the project and to @motorailgun @WaterWhisperer for all your efforts in reviving #4668 and pushing it towards the finish line :)

PR #4834, opened 2026-04-26
fix: rename partially "Triple" to "Tuple" to reflect the new terminology (pt.2)

motorailgun:
Partial effort to #4826, and continuous to #4827.

This PR still does not do functionality changes. ~~Opening as draft PR to rely on CI for Windows-only cfgs.~~ <- fixed!


src/dist/mod.rs
```diff
@@ -44,8 +44,8 @@ use prefix::InstallPrefix;
 
 pub mod temp;
 
-pub(crate) mod triple;
-pub(crate) use triple::*;
+pub(crate) mod tuple;
```
rami3l: We need to be a bit more careful here because previously `triple` as an abbreviation for `target_triple` is more or less unambiguous. As discussed in https://github.com/rust-lang/rustup/pull/4668, we would like to rename this module to `target_tuple` instead of just `tuple` to avoid potential confusion with the builtin `tuple` type.

PS: In e.g. method names however, this is less of a problem, so you can do nothing for now. If anyone still finds it too problematic, we can use `_target` to replace `_triple` later on.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4834/445d972da627fe3348a443e24ad28b6984d58e5d..88fe876455d220d498926ba504c175996de58f6c)*

tests/suite/known_tuples.rs
```diff
@@ -50,12 +50,12 @@ fn gen_known_tuples() {
     }
 }
 
-/// Parses the given triple into 3 parts (target architecture, OS and environment).
+/// Parses the given tuple into 3 parts (target architecture, OS and environment).
```
rami3l: Similarly to https://github.com/rust-lang/rustup/pull/4834/changes/b07ea10c40e0ef7aaf504339216bad91d45c5089#r3144041456, this module should be renamed to `known_target_tuples`. The function names can stay with `_tuple`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4834/445d972da627fe3348a443e24ad28b6984d58e5d..88fe876455d220d498926ba504c175996de58f6c)*

src/dist/mod.rs
motorailgun: Thanks! So do you mean we want to keep method names with `_triple` suffix (or whatsoever) as is, while renaming `triple` module to `target_tuple`? (Sorry I'm not really good at English)

rami3l: > Thanks! So do you mean we want to keep method names with `_triple` suffix (or whatsoever) as is, while renaming `triple` module to `target_tuple`?

@motorailgun My apologies for not being super clear previously!

For method names you can perform the switch from `_triple` to `_tuple` as usual, depending on your progress.

If this doesn't feel quite right for you in certain places please let me know.

motorailgun: Ah! Got it now, sorry for my dumbness...

rami3l: @motorailgun It's perfectly fine... There are no dumb questions :)

src/dist/mod.rs
```diff
@@ -655,7 +655,7 @@ impl PartialToolchainDesc {
         })
     }
 
-    pub(crate) fn has_triple(&self) -> bool {
+    pub(crate) fn has_tuple(&self) -> bool {
         self.target.arch.is_some() || self.target.os.is_some() || self.target.env.is_some()
```
rami3l: Nit: Pre-existing, but I think you can extract before this commit a `PartialTargetTriple::is_empty()` helper, and then inline all uses of this function to `!desc.target.is_empty()`. That way the level of abstraction feels more correct and the overall logic is more readable to me... What do you think?

rami3l (approved):
LGTM modulo one tiny nit, nice work :]

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4834/4fed1c8001b0d92b8acd0df7af0de8858e2ec10e..ce09c9a4807e4bac35335da75fdc2f671d03b8fc)*

src/dist/mod.rs
motorailgun: Oh yeah that feels better especially because it is checking/digging `self.target`. Will look into more and fix this.

motorailgun:
Got 'em!

src/dist/triple.rs
```diff
@@ -51,6 +51,10 @@ impl PartialTargetTriple {
             }
         })
     }
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.arch.is_none() || self.env.is_none() || self.os.is_none()
```
rami3l: De Morgan's Law: we should write `&&` here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4834/4fed1c8001b0d92b8acd0df7af0de8858e2ec10e..ed78b525c12effd9c0e3f849b46b34efc275d2a1)*

motorailgun: Oh damn, I am stupid... fixing!

rami3l: @motorailgun No worries :)

PR #4835, opened 2026-04-26
dist: bump `rustup` version to v1.29.1

rami3l:
This is the first step in the v1.29.1 release process using our newly in-test release process (https://rust-lang.github.io/rustup/dev-guide/release-process.html#patch-version-bumps).

It also prepares for the v1.30.0 release process on `main` (https://github.com/rust-lang/rustup/pull/4832).

See also: https://github.com/rust-lang/blog.rust-lang.org/pull/1880

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2230,7 +2230,7 @@ dependencies = [
 
 [[package]]
 name = "rustup"
-version = "1.29.0"
+version = "1.29.1"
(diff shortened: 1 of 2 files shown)


rami3l:
PS: I know this PR doesn't require any approval to merge, but still this way the whole team can get notified.

djc (approved):

PR #4840, opened 2026-04-28
fix(cli/rustup_mode)!: skip auto-installation in some `rustup` commands

### LINKED ISSUE #4836: Auto-installation of the active toolchain gets triggered too eagerly
> [!IMPORTANT]
> See here https://github.com/rust-lang/rustup/issues/4836#issuecomment-4467247352 for a list of subcommands that will be influenced by this change in rustup v1.30.

### Verification

### Problem

Prerequisite for #4454, related to #4211.

When resolving to a toolchain, if `RUSTUP_AUTO_INSTALL` is enabled, rustup will try to resolve the toolchain no matter what, and this will cause the toolchain to be installed.

For example, even `rustup +beta target remove aarch64-apple-darwin` will cause `beta` to be installed when it is not installed yet.

### Steps

As noticed in https://github.com/rust-lang/rustup/pull/4454#discussion_r2395038850, this behavior dates even before v1.27.1:

```console
> docker run --rm -it rust:alpine sh

> rustup --version
info: The currently active `rustc` version is `rustc 1.81.0 (eeb90cda1 2024-09-04)`

> rustup +beta target remove aarch64-apple-darwin
info: syncing channel updates for 'beta-aarch64-unknown-linux-musl'
1007.6 KiB / 1007.6 KiB (100 %) 567.4 KiB/s in  1s ETA:  0s
info: latest update on 2025-09-28, rust version 1.91.0-beta.4 (aa7859c0d 2025-09-27)
^C
```

### Possible Solution(s)

In the future, it might be reasonable to restrict auto installation behavior to a reasonable subset, even with `RUSTUP_AUTO_INSTALL` set to `1`.

Thus, the suggested solution here would be to perform a case-by-case analysis to restrict auto install behavior, and hopefully eventually to the strict minimum.

For example, as the first step, for `rustup component` it might be reasonable to ban auto installation altogether; or we can even ban all cases in `rustup_mode.rs`, etc. It really depends on in what way v1.28.0 was broken for our users.

### Notes

Also related to https://github.com/rust-lang/rustup/issues/4211.

Implicit behaviors like this is one of the very important justifications for not shipping `RUSTUP_AUTO_INSTALL=1`; the test failures in #4454 has also demonstrated how often this setting is consulted which might be surprising even for those who use/support auto installation.

### Rustup version

```console
> rustup --version
rustup 1.30.0 :: 1.28.2+599 (4fed1c800 2026-04-26)
```

### Installed toolchains

```console
N/A
```

### OS version

```console
macOS 26.4.1 (25E253)
```

rami3l:
Closes #4836 by making it possible to skip `should_auto_install` 's dynamic detection entirely.

Whether the skip should be applied can be predetermined during `clap` dispatch (at least for `rustup_mode.rs`), so the only remaining question seems to be where we would want to apply the skip.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -301,6 +301,36 @@ fn update_toolchain_value_parser(s: &str) -> Result<PartialToolchainDesc> {
     })
 }
 
+impl RustupSubcmd {
+    fn allow_auto_install(&self) -> bool {
+        match self {
+            // These subcommands execute or rely on the active toolchain, so auto-installing it when
+            // missing may be reasonable depending on the user's decision.
+            #[cfg(not(windows))]
+            RustupSubcmd::Man { .. } => true,
+            RustupSubcmd::Doc { .. } | RustupSubcmd::Run { .. } => true,
+
+            // These subcommands don't require the active toolchain, so auto-installing it should be
+            // disabled to avoid surprises.
+            RustupSubcmd::Check { .. }
+            | RustupSubcmd::Completions { .. }
+            | RustupSubcmd::Component { .. }
+            | RustupSubcmd::Default { .. }
+            | RustupSubcmd::DumpTestament
+            | RustupSubcmd::Install { .. }
+            | RustupSubcmd::Override { .. }
+            | RustupSubcmd::Self_ { .. }
+            | RustupSubcmd::Set { .. }
+            | RustupSubcmd::Show { .. }
+            | RustupSubcmd::Target { .. }
+            | RustupSubcmd::Toolchain { .. }
+            | RustupSubcmd::Uninstall { .. }
+            | RustupSubcmd::Update { .. }
+            | RustupSubcmd::Which { .. } => false,
+        }
+    }
+}
+
 #[derive(Debug, Subcommand)]
 enum ShowSubcmd {
     /// Show the active toolchain
@@ -631,14 +661,19 @@ pub async fn main(
 
     update_console_filter(process, &console_filter, matches.quiet, matches.verbose);
 
-    let cfg = &mut Cfg::from_env(current_dir, matches.quiet, true, process)?;
-    cfg.toolchain_override = matches.plus_toolchain;
-
     let Some(subcmd) = matches.subcmd else {
         let help = Rustup::command().render_long_help();
         writeln!(process.stderr().lock(), "{}", help.ansi())?;
         return Ok(ExitCode::FAILURE);
     };
+
+    let cfg = &mut Cfg::from_env(
+        current_dir,
+        matches.quiet,
+        subcmd.allow_auto_install(),
+        process,
+    )?;
+    cfg.toolchain_override = matches.plus_toolchain;
 
     match subcmd {
         RustupSubcmd::DumpTestament => common::dump_testament(process),
@@ -1968,8 +2003,8 @@ fn output_completion_script(
 }
 
 async fn display_version(current_dir: PathBuf, process: &Process) -> Result<()> {
-    info!("This is the version for the rustup toolchain manager, not the rustc compiler.");
-    let mut cfg = Cfg::from_env(current_dir, true, true, process)?;
+    info!("this is the version for the rustup toolchain manager, not the rustc compiler");
+    let mut cfg = Cfg::from_env(current_dir, true, false, process)?;
     cfg.toolchain_override = cfg
         .process
         .args()
@@ -1982,6 +2017,12 @@ async fn display_version(current_dir: PathBuf, process: &Process) -> Result<()> 
                 "the currently active `rustc` version is `{}`",
                 tc.rustc_version()
             ),
+            Err(RustupError::ToolchainNotInstalled {
+                name,
+                is_active: true,
+            }) => {
+                info!("the active toolchain `{name}` is not installed");
+            }
             Err(err) => error!("failed to display the current `rustc` version: {err}"),
         },
         Ok(None) => info!("no `rustc` is currently active"),
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -631,7 +631,7 @@ pub async fn main(
 
     update_console_filter(process, &console_filter, matches.quiet, matches.verbose);
 
-    let cfg = &mut Cfg::from_env(current_dir, matches.quiet, process)?;
+    let cfg = &mut Cfg::from_env(current_dir, matches.quiet, true, process)?;
     cfg.toolchain_override = matches.plus_toolchain;
 
     let Some(subcmd) = matches.subcmd else {
@@ -1974,7 +1974,7 @@ fn output_completion_script(
 
 async fn display_version(current_dir: PathBuf, process: &Process) -> Result<()> {
     info!("This is the version for the rustup toolchain manager, not the rustc compiler.");
-    let mut cfg = Cfg::from_env(current_dir, true, process)?;
+    let mut cfg = Cfg::from_env(current_dir, true, true, process)?;
     cfg.toolchain_override = cfg
         .process
         .args()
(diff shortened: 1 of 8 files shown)


rami3l:
@djc ~~At the moment of writing, this PR only covers `rustup component` and `rustup target` subcommands.~~ As your previous comment (https://github.com/rust-lang/rustup/issues/4836#issuecomment-4366294693) has given a clear criterion for whether auto-installation is accepted, I'll propose the following classification to be implemented here:

```console
Commands that should respect auto-installation:
  rustup-init  No, because the notion of toolchain doesn't exist yet in that stage
  install      No, because `rustup install` installs the active toolchain by default
  uninstall    No, similarly to `rustup install`
  toolchain    No
  default      No, because `rustup default foo` installs `foo`
  show         No, because it's query-only
  update       No, similarly to `rustup toolchain`
  check        No, because it's query-only
  target       No, because it operates on an installed toolchain
  component    No, similarly to `rustup target`
  override     No
  run          Yes, because the said toolchain is invoked
  which        No, because it's query-only
  doc          Yes, because the said toolchain is relied upon
  man          Yes, similarly to `rustup doc`
  self         No, similarly to `rustup-init`
  set          No
  completions  No
```

What do you think?

djc:
I like it! Should we review `toolchain` by subcommand?

rami3l:
> I like it! Should we review `toolchain` by subcommand?

@djc We have four subcommands here: `install` and `uninstall` are already discussed above; `list` is query-only and `link` operates on an existing toolchain, so the conclusion would be a hard NO in all cases.

djc:
Sounds good. Are there other commands with subcommands? `target`?

rami3l:
> Sounds good. Are there other commands with subcommands? `target`?

@djc I wrote all the above analysis with all subcommands in mind already. Please feel free to tell me if you think there are any cases you don't agree with.

PS: The best way to enumerate all dispatch points is to go to the code itself:

https://github.com/rust-lang/rustup/blob/6e6816481070d4cd91a2ca7ad12c03c471ad25a7/src/cli/rustup_mode.rs#L643-L770

djc:
> @djc I wrote all the above analysis with all subcommands in mind already. Please feel free to tell me if you think there are any cases you don't agree with.

Okay, just checking!

rami3l:
Almost done. The only concern would be that `rustup show` bails out too early when auto-installation is disabled (pre-existing issue).

I'll try to fix it here soon...

rami3l:
PS: It's very fun in that I'll just need to search for `.await` to see where we can possibly trigger auto-installation in a subcommand XD

rami3l:
Oh, the CI passes now, so I'm marking this first PR as ready. Please feel free to let me know if there are any improvements/clarifications to be made :)

src/cli/rustup_mode.rs
```diff
@@ -631,15 +631,41 @@ pub async fn main(
...
+        | RustupSubcmd::Override { .. }
+        | RustupSubcmd::Self_ { .. }
+        | RustupSubcmd::Set { .. }
+        | RustupSubcmd::Show { .. }
+        | RustupSubcmd::Target { .. }
+        | RustupSubcmd::Toolchain { .. }
+        | RustupSubcmd::Uninstall { .. }
+        | RustupSubcmd::Update { .. }
+        | RustupSubcmd::Which { .. } => false,
+    };
```
djc: Nit: suggest moving this into a method attached to `RustupSubCmd`?

src/cli/rustup_mode.rs
djc: Pre-existing nits to avoid going very deep into rightward drift:

- It's not clear why we're using a nested scope here, but I think we could get rid of it since this is the last expression in the function anyway?
- Yield the `Some` case from `active_toolchain_and_source` out of the match, since the `None` case is pretty trivial

src/cli/rustup_mode.rs
```diff
@@ -1207,13 +1192,14 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<ExitCode> {
 
         match active_toolchain_and_source {
             Some((active_toolchain_name, active_source)) => {
+                writeln!(t.lock(), "name: {active_toolchain_name}")?;
+                writeln!(t.lock(), "active because: {}", active_source.to_reason())?;
+
```
djc: This change looks more or less unrelated, but feels related, maybe split it into a separate commit?

src/cli/rustup_mode.rs
```diff
@@ -1195,39 +1195,48 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<ExitCode> {
...
-                            .components()?
-                            .into_iter()
-                            .filter_map(|c| {
-                                (c.installed && c.component.short_name() == "rust-std").then(|| {
-                                    c.component.target.expect("rust-std should have a target")
-                                })
-                            })
-                            .collect()
+                match Toolchain::new(cfg, active_toolchain_name.clone().into()) {
+                    Ok(active_toolchain) => {
```
djc: Similarly, get this case out of the `match` if possible? By this point line 1218 is 13 levels deep which IMO is not a good state.

djc (commented):
Nice work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4840/6fb0b6f0ad45577f1eda1794822a15298be1a677..42b4bea25c210d7fad999141c7345210bd25e8de)*

src/cli/rustup_mode.rs
rami3l: Did a more ambitious refactoring upfront.

src/cli/rustup_mode.rs
rami3l: Unfortunately I don't think this is possible because we are matching on both `Ok()` and `Err()` cases, and Rust IIRC doesn't have anything other than `match` for this.

I've addressed other concerns though so the total level of nesting should be down already.

src/config.rs
```diff
@@ -246,12 +246,19 @@ pub(crate) struct Cfg<'a> {
     pub quiet: bool,
     pub current_dir: PathBuf,
     pub process: &'a Process,
+
+    /// If this flag is set to `false`, it can stop `rustup` from automatically installing the
```
djc: Nit: I think best practice for docstrings is to summarize meaning in a way that fits on one line, then expand on that in subsequent lines (usually after an empty line). That makes it a bit easier to skim/skip over them.

src/cli/rustup_mode.rs
```diff
@@ -1190,32 +1190,41 @@ async fn show(cfg: &Cfg<'_>, verbose: bool) -> Result<ExitCode> {
     writeln!(t.lock(), "name: {active_toolchain_name}")?;
     writeln!(t.lock(), "active because: {}", active_source.to_reason())?;
 
-    let active_toolchain =
-        Toolchain::with_source(cfg, active_toolchain_name.clone().into(), &active_source)?;
-    if verbose {
-        writeln!(t.lock(), "compiler: {}", active_toolchain.rustc_version())?;
-        writeln!(t.lock(), "path: {}", active_toolchain.path().display())?;
-    }
+    match Toolchain::new(cfg, active_toolchain_name.clone().into()) {
```
djc: I think it might still be worthwhile here to handle the error cases inside the `match` and yield the `Ok()` case out of the `match` to reduce rightward drift a bit more.

djc (approved):
Nice work! That deeply nested code looks a lot better now...

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4840/d51b56c4aa38a159d983bdeca64b10b9e6b79d4f..326e48dcb74e888e028f95280dafe1a4bda7ee77)*

src/config.rs
rami3l: @djc Thanks! Sorry I've messed up the history a bit (not very good at `jj` yet, still learning)... But it should be good now!

src/cli/rustup_mode.rs
rami3l: @djc Did an early-return in all the sad paths. I hope the diff looks better to you now.

djc (approved):

PR #4844, opened 2026-05-03
Simplify downloads

djc:
Following up to

- #3788

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -63,7 +63,7 @@ use crate::{
         DistOptions, PartialToolchainDesc, Profile, TargetTuple, ToolchainDesc,
         download::DownloadCfg,
     },
-    download::download_file,
+    download::Download,
     errors::RustupError,
     install::{InstallMethod, UpdateStatus},
     process::Process,
@@ -1340,7 +1340,9 @@ pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> Result<Option<Pa
 
     // Download new version
     info!("downloading self-update (new version: {available_version})");
-    download_file(&download_url, &setup_path, None, None, dl_cfg.process).await?;
+    Download::new(&download_url, &setup_path, dl_cfg.process)
+        .download()
+        .await?;
 
     // Mark as executable
     utils::make_executable(&setup_path)?;
@@ -1359,7 +1361,10 @@ async fn get_available_rustup_version(dl_cfg: &DownloadCfg<'_>) -> Result<String
     let release_file_url = format!("{update_root}/release-stable.toml");
     let release_file_url = utils::parse_url(&release_file_url)?;
     let release_file = tempdir.path().join("release-stable.toml");
-    download_file(&release_file_url, &release_file, None, None, dl_cfg.process).await?;
+    Download::new(&release_file_url, &release_file, dl_cfg.process)
+        .download()
+        .await?;
+
     let release_toml_str = utils::read_file("rustup release", &release_file)?;
     let release_toml = toml::from_str::<RustupManifest>(&release_toml_str)
         .context("unable to parse rustup release file")?;
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -63,7 +63,7 @@ use crate::{
         DistOptions, PartialToolchainDesc, Profile, TargetTuple, ToolchainDesc,
         download::DownloadCfg,
     },
-    download::Download,
+    download::DownloadOptions,
     errors::RustupError,
     install::{InstallMethod, UpdateStatus},
     process::Process,
@@ -1340,7 +1340,8 @@ pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> Result<Option<Pa
 
     // Download new version
     info!("downloading self-update (new version: {available_version})");
-    Download::new(&download_url, &setup_path, dl_cfg.process)
+    DownloadOptions::try_from(dl_cfg.process)?
+        .start(&download_url, &setup_path)
         .download()
         .await?;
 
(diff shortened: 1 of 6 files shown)


src/cli/self_update.rs
```diff
@@ -51,6 +51,7 @@ use same_file::Handle;
 use serde::{Deserialize, Serialize};
 use tracing::{error, info, trace, warn};
 
+use crate::download::DownloadOptions;
```
rami3l: Nit: Move this to the `use` group below.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4844/57b8fa0850b4a1123f8853ceeb74d9b2c15afe11..851139ec6687f7871cae23a2fc36894240ce968e)*

src/dist/download.rs
```diff
@@ -267,7 +262,14 @@ impl<'a> DownloadCfg<'a> {
         let file = self.tmp_cx.new_file_with_ext("", ext)?;
 
         let mut hasher = Sha256::new();
-        download_file(&url, &file, Some(&mut hasher), status, self.process).await?;
+        let download = Download::new(&url, &file, self.process).with_hasher(&mut hasher);
+
+        let download = match status {
+            Some(status) => download.with_status(status),
```
rami3l, line 272: Nit: A common builder trick is to let `.with_foo()` accept `impl Into<Option<Foo>>` so it can take `foo`, `Some(foo)` and `None` at the same time. Hopefully this way the chain can be more readable here 🙏 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4844/57b8fa0850b4a1123f8853ceeb74d9b2c15afe11..851139ec6687f7871cae23a2fc36894240ce968e)*

src/download/mod.rs
```diff
@@ -285,25 +257,9 @@ impl<'a> Download<'a> {
         };
 
         // TODO: the sync callback will stall the async runtime if IO calls block, which is OS dependent. Rearrange.
```
rami3l: Question: What should we do about this outdated comment?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4844/57b8fa0850b4a1123f8853ceeb74d9b2c15afe11..851139ec6687f7871cae23a2fc36894240ce968e)*

djc: I meant to remove that! I don't think it's still relevant after the last commit here!

src/dist/download.rs
djc: Hmm, not sure I love it. I like the added precision of the builder method exactly taking the thing (not the thing wrapped in `Option`), and I don't mind this boilerplate so much -- while it's more verbose, it still seems still pretty straightforward to understand.

src/cli/self_update.rs
djc: Done.

src/download/mod.rs
djc: Killed it.

src/download/mod.rs
```diff
@@ -163,35 +163,9 @@ impl<'a> Download<'a> {
...
         debug!(url = %self.url, "downloading file");
 
-        // This callback will write the download to disk and optionally
-        // hash the contents, then forward the notification up the stack
-        let callback: &dyn Fn(Event<'_>) -> anyhow::Result<()> = &|msg| {
-            if let Event::DownloadDataReceived(data) = msg
-                && let Some(h) = self.hasher.as_ref()
-            {
-                h.borrow_mut().update(data);
-            }
```
FranciscoTGouveia: Was the `h.borrow_mut().update(data)` call (previously inside the callback) moved elsewhere when the callback was removed, or is it no longer necessary?

The hasher is indeed initialized, via `with_hasher()`, but it does not appear to receive any bytes.
That said, shouldn't hash verification always fail?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4844/57b8fa0850b4a1123f8853ceeb74d9b2c15afe11..cbd4f1e696f3899a135286684751442174f067ee)*

src/download/mod.rs
```diff
@@ -295,106 +243,73 @@ impl Backend {
...
         }
 
-        let res = request(url, resume_from, client)
-            .await
-            .context("error downloading file")?;
+        let mut req = client.get(url.as_str());
+        if resume_from != 0 {
+            req = req.header(header::RANGE, format!("bytes={resume_from}-"));
+        }
+        let res = req.send().await.context("error downloading file")?;
```
FranciscoTGouveia: Shouldn't this be `.map_err(DownloadError::Reqwest)` instead?

Previously, `request()` returned `Result<Response, DownloadError>`, so any `reqwest::Error` was automatically wrapped as `DownloadError::Reqwest` before being propagated.

As it stands, this change breaks the `is_network_failure()` function, and likely the `network_failure_does_not_delete_partial_file` test as well.
Without the wrapping, the error can no longer be downcast to `DownloadError`, causing `is_network_failure` to always return false, and thus, always removing partial files.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4844/57b8fa0850b4a1123f8853ceeb74d9b2c15afe11..cbd4f1e696f3899a135286684751442174f067ee)*

djc: Good suggestions, should be fixed now.

src/download/mod.rs
djc: Dropped this commit for now.

djc:
Dropped the last commit for now, which was a bit too ambitious/too much change at once. Will rework it in a few steps after this is merged.

rami3l (approved):
LGTM, let's see how it goes :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4844/6fb0b6f0ad45577f1eda1794822a15298be1a677..5f303440b292527f24eb383841c2cb73b9b909c1)*

PR #4851, opened 2026-05-08
feat(cli/self-update): refine wording of "already installed Rust" warning

### LINKED ISSUE #3679: when rustup finds an existing rust install it doesn't give the user any hints about what to do next

have an app that I'm building that gives instructions to "install rust", and links to the rustup site.
I ran this as instructed: 
```
curl https://sh.rustup.rs -sSf | sh
info: downloading installer
warning: it looks like you have an existing installation of Rust at:
warning: /usr/bin
warning: It is recommended that rustup be the primary Rust installation.
warning: Otherwise you may have confusion unless you are careful with your PATH
warning: If you are sure that you want both rustup and your already installed Rust
warning: then please reply `y' or `yes' or set RUSTUP_INIT_SKIP_PATH_CHECK to yes
warning: or pass `-y' to ignore all ignorable checks.
```

rustup's idea of what rust is, not installed, because of "some other" rust  and no idea how to fix that above error.

### Solution you'd like

I wanted it to install rust using rustup

after 10 minutes of googling this came up: 
https://www.reddit.com/r/rust/comments/se7ecj/how_to_uninstall_rust_without_rustup/
https://github.com/rust-lang/rls/issues/1399
https://github.com/rust-lang/rustup/issues/953

.. so its not just me..

solution:
emit this message instead:
"Would you like to uninstall the system-provided/older rust and use a newer one?" run 'sudo apt remove rustc' or similar for your platform

### Notes

_No response_

rami3l:
Closes #3679 by mentioning https://rust-lang.github.io/rustup/installation/already-installed-rust.html in the `rustup-init` installer's output.

As discussed in https://github.com/rust-lang/rustup/issues/3679#issuecomment-4387717425, if a `rustup` packager chooses to skip the installer completely, it is the responsibility of the packager to issue this warning in the package manager instead.


mobiuscog:
Would this cause confusion if:
1) Rustup is installed elsewhere (system) but not in the user override
2) Rustup is installed in the user override already
Or would the code checks prevent this text being shown in the case of (2) ?

I like the change though (as a non-reviewer).

rami3l:
> Would this cause confusion if:
> 
> 1. Rustup is installed elsewhere (system) but not in the user override
> 2. Rustup is installed in the user override already
>    Or would the code checks prevent this text being shown in the case of (2) ?

@mobiuscog My apologies but I am not sure if I have fully comprehended the said confusion.

Rustup is designed as a per-user application, but the way it's packaged in say Debian, as long as the package is installed by the sysadmin, `rustc` and `cargo` will be present in your `PATH` immediately; it's the toolchain installation that will happen locally in your user account.

Whether that is the case or you have previously installed `rustup` from https://rustup.rs doesn't matter; in both cases there's no need to run the installer again 👀

Maybe I have misunderstood something here?

mobiuscog:
Apologies @rami3l - perhaps it's my confusion around what happens if rustup is already installed at a system level and I install a toolchain - I would have expected it to be at a system level as well (and therefore collide) but if the system rustup is per-user and would simply add to a local user toolchain install (and not affect the system install) then that's likely fine.

I have only ever installed it as a user, never as root. I have just noticed that rustup and rust are often different packages in a repository, and so there could be confusion around what happens if a user installs rustup via the shell, and then a later system update installs it as a system package. Perhaps I'm overthinking things here and just don't understand the difference between rustup as a distro package and rustup via the website/script, if at all there is one.

src/cli/self_update.rs
```diff
@@ -710,11 +710,13 @@ fn check_existence_of_rustc_or_cargo_in_path(no_prompt: bool, process: &Process)
     if let Err(path) = rustc_or_cargo_exists_in_path(process) {
         warn!("It looks like you have an existing installation of Rust at:");
         warn!("{}", path);
-        warn!("It is recommended that rustup be the primary Rust installation.");
-        warn!("Otherwise you may have confusion unless you are careful with your PATH.");
-        warn!("If you are sure that you want both rustup and your already installed Rust");
-        warn!("then please reply `y' or `yes' or set RUSTUP_INIT_SKIP_PATH_CHECK to yes");
-        warn!("or pass `-y' to ignore all ignorable checks.");
+        warn!("This could mean one of the following:");
+        warn!("- Rustup is already installed, and you can quit the installer right now.");
```
djc: I would canonically spell "rustup" with a lowercase r.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4851/228e4b4cea45565fedb3a332b86a968d3c8fe685..03d41fd00ae336d831aa73479c9189606bcb64e3)*

djc (approved):

src/cli/self_update.rs
rami3l: @djc I'm treating "rustup" as a regular compound word here. That said, from what I've observed, the canonical spelling from https://github.com/rust-lang/rustup/blob/9f5c731051a4d6cb749e93e80288705fd4666f39/README.md is clearly with an initial uppercase, and the [book](https://rust-lang.github.io/rustup) uses an initial lowercase, I don't think there is a clear consensus on this matter 🤷‍♀️

rami3l: PS: I've noticed your preference and I've been avoiding it by putting the name of our project at the middle of the sentence.

rami3l:
> I have only ever installed it as a user, never as root. I have just noticed that rustup and rust are often different packages in a repository, and so there could be confusion around what happens if a user installs rustup via the shell, and then a later system update installs it as a system package. Perhaps I'm overthinking things here and just don't understand the difference between rustup as a distro package and rustup via the website/script, if at all there is one.

@mobiuscog No worries, thanks again for your feedback and your prudence!

PR #4861, opened 2026-05-18
fix: stop deleting $CARGO_HOME during uninstall

### LINKED ISSUE #285: Uninstall deletes entire .cargo directory
Surely it should only delete stuff that it created?

Cloud0310:
fix #285
This PR fix the cargo bin nuke and the path removing possilble toutoc issue.
Here's the implementation details:
- compare cargo bin dir content to rustup binary, if it's a link, remove it
- then, we remove the rustup binary, on success, we then remove the cargo bin from path

This prevents `rustup self uninstall` from nuking out the `$CARGO_HOME/bin` and possible toctou problem of removing up the path.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1148,10 +1148,15 @@ pub(crate) fn uninstall(
 
     info!("removing rustup binaries");
 
-    // Delete rustup. This is tricky because this is *probably*
+    // Delete rustup.
+    #[cfg(unix)]
+    delete_rustup_and_cargo_home(process)?;
+    // NOTE: On windows, this is tricky because this is *probably*
     // the running executable and on Windows can't be unlinked until
     // the process exits.
-    delete_rustup_and_cargo_home(process)?;
+    // see: windows::{complete_windows_uninstall,spawn_uninstall_gc}
+    #[cfg(windows)]
+    windows::spawn_uninstall_gc(process)?;
 
     info!("rustup is uninstalled");
 
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -24,15 +24,16 @@
 //! During uninstall (`rustup self uninstall`):
 //!
 //! * Delete `$RUSTUP_HOME`.
-//! * Delete everything in `$CARGO_HOME`, including
-//!   the rustup binary and its hardlinks
+//! * Delete all entries in `$CARGO_HOME` except `bin`.
+//! * Delete rustup tool links and binary from `$CARGO_HOME/bin`.
+//! * Delete `$CARGO_HOME/bin` if it is empty after uninstall.
+//! * Delete `$CARGO_HOME` if it is empty after uninstall.
 //!
 //! Deleting the running binary during uninstall is tricky
 //! and racy on Windows.
 
 use std::borrow::Cow;
 use std::env::{self, consts::EXE_SUFFIX};
-#[cfg(not(windows))]
 use std::io;
 use std::io::Write;
 use std::path::{Component, MAIN_SEPARATOR, Path, PathBuf};
@@ -47,7 +48,7 @@ use clap::ValueEnum;
 use clap::builder::PossibleValue;
 use clap_cargo::style::{GOOD, WARN};
 use itertools::Itertools;
-use same_file::Handle;
+use same_file::{Handle, is_same_file};
 use serde::{Deserialize, Serialize};
 use tracing::{error, info, trace, warn};
 
@@ -1050,10 +1051,11 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
 
 /// Uninstall process:
 /// 1. Remove rustup home.
-/// 2. Delete everything in CARGO_HOME *except* the except the bin directory
-/// 3. Remove other tools in $CARGO_HOME/bin, but skipping rustup binary file and links to rustup.
-/// 4. If no `--no-modify-path` is passed $CARGO_HOME/bin, clean up $PATH.
-/// 5. Remove $CARGO_HOME
+/// 2. Remove all entries in `$CARGO_HOME` except `bin`.
+/// 3. Remove rustup tool links and binary.
+/// 4. Try to remove $CARGO_HOME/bin directory if it's empty.
+/// 5. Upon successfully removing $CARGO_HOME/bin, clean up $PATH.
+/// 6. Try to remove $CARGO_HOME directory if it's empty.
 pub(crate) fn uninstall(
     no_prompt: bool,
     no_modify_path: bool,
@@ -1096,11 +1098,9 @@ pub(crate) fn uninstall(
         utils::remove_dir("rustup_home", &rustup_dir)?;
     }
 
-    info!("removing rustup binaries");
-
     // Delete rustup.
     #[cfg(unix)]
-    delete_rustup_and_cargo_home(no_modify_path, process)?;
+    clean_cargo_home(no_modify_path, process)?;
     // NOTE: On windows, this is tricky because this is *probably*
     // the running executable and on Windows can't be unlinked until
     // the process exits.
@@ -1113,14 +1113,17 @@ pub(crate) fn uninstall(
     Ok(ExitCode::SUCCESS)
 }
 
-fn delete_rustup_and_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
+/// Remove rustup-owned cargo-home state.
+/// This removes non-`bin` entries in `$CARGO_HOME`, removes rustup tool links and executable from
+/// `$CARGO_HOME/bin`, then removes `$CARGO_HOME/bin` and `$CARGO_HOME` only if they are empty.
+/// Nonempty directories are left in place.
+fn clean_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
     let cargo_home = process.cargo_home()?;
+    let cargo_bin = cargo_home.join("bin");
 
     info!("removing cargo home");
 
-    // Delete everything in CARGO_HOME *except* the rustup bin.
-
-    // First everything except the bin directory.
+    // Delete everything in CARGO_HOME except the bin directory first.
     let diriter = fs::read_dir(&cargo_home).map_err(|e| CliError::ReadDirError {
         p: cargo_home.clone(),
         source: e,
@@ -1139,39 +1142,58 @@ fn delete_rustup_and_cargo_home(no_modify_path: bool, process: &Process) -> Resu
         }
     }
 
-    // Then everything in bin except rustup and tools. These can't be unlinked
-    // until this process exits (on windows).
-    let tools = TOOLS
+    info!("removing rustup tool links and binary");
+
+    let rustup_path = cargo_bin.join(format!("rustup{EXE_SUFFIX}"));
+
+    let proxy_paths = TOOLS
         .iter()
         .chain(DUP_TOOLS.iter())
-        .map(|t| format!("{t}{EXE_SUFFIX}"));
-    let tools: Vec<_> = tools.chain(vec![format!("rustup{EXE_SUFFIX}")]).collect();
-    let bin_dir = cargo_home.join("bin");
-    let diriter = fs::read_dir(&bin_dir).map_err(|e| CliError::ReadDirError {
-        p: bin_dir.clone(),
-        source: e,
-    })?;
-    for dirent in diriter {
-        let dirent = dirent.map_err(|e| CliError::ReadDirError {
-            p: bin_dir.clone(),
-            source: e,
-        })?;
-        let name = dirent.file_name();
-        let file_is_tool = name.to_str().map(|n| tools.iter().any(|t| *t == n));
-        if file_is_tool == Some(false) {
-            if dirent.path().is_dir() {
-                utils::remove_dir("cargo_home", &dirent.path())?;
-            } else {
-                utils::remove_file("cargo_home", &dirent.path())?;
-            }
+        .map(|tool| cargo_bin.join(format!("{tool}{EXE_SUFFIX}")));
+
+    for proxy_path in proxy_paths {
+        if is_same_file(&proxy_path, &rustup_path).unwrap_or(false) {
+            utils::remove_file("rustup tool proxy", &proxy_path)?;
         }
     }
 
-    if !no_modify_path {
-        do_remove_from_path(process)?;
+    utils::remove_file("rustup_bin", &rustup_path)?;
+
+    let cargo_bin_display = cargo_bin.display();
+    info!("removing empty cargo bin directory `{cargo_bin_display}`");
+
+    match fs::remove_dir(&cargo_bin) {
+        Err(e) if e.kind() == io::ErrorKind::DirectoryNotEmpty => {
+            warn!("keeping non-empty cargo bin directory `{cargo_bin_display}`")
+        }
+        Err(e) => {
+            return Err(e).with_context(|| {
+                format!("failed to remove cargo bin directory `{cargo_bin_display}`")
+            });
+        }
+        Ok(()) if !no_modify_path => {
+            info!("removing cargo bin directory `{cargo_bin_display}` from $PATH");
+            do_remove_from_path(process)?;
+        }
+        Ok(()) => {}
     }
 
-    utils::remove_dir("cargo_home", &cargo_home)
+    let cargo_home_display = cargo_home.display();
+    info!("removing empty cargo home directory `{cargo_home_display}`");
+
+    match fs::remove_dir(&cargo_home) {
+        Err(e) if e.kind() == io::ErrorKind::DirectoryNotEmpty => {
+            warn!("keeping non-empty cargo home directory `{cargo_home_display}`");
+        }
+        Err(e) => {
+            return Err(e).with_context(|| {
+                format!("failed to remove cargo home directory `{cargo_home_display}`")
+            });
+        }
+        Ok(()) => {}
+    }
+
+    Ok(())
 }
 
 #[derive(Clone, Copy, Debug)]
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1048,6 +1048,12 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     Ok(())
 }
 
+/// Uninstall process:
+/// 1. Remove rustup home.
+/// 2. Delete everything in CARGO_HOME *except* the except the bin directory
+/// 3. Remove other tools in $CARGO_HOME/bin, but skipping rustup binary file and links to rustup.
+/// 4. If no `--no-modify-path` is passed $CARGO_HOME/bin, clean up $PATH.
+/// 5. Remove $CARGO_HOME
 pub(crate) fn uninstall(
     no_prompt: bool,
     no_modify_path: bool,
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -80,7 +80,7 @@ mod shell;
 #[cfg(unix)]
 mod unix;
 #[cfg(unix)]
-use unix::{delete_rustup_and_cargo_home, do_add_to_path, do_remove_from_path};
+use unix::{do_add_to_path, do_remove_from_path};
 #[cfg(unix)]
 pub(crate) use unix::{run_update, self_replace};
 
@@ -1161,6 +1161,12 @@ pub(crate) fn uninstall(
     info!("rustup is uninstalled");
 
     Ok(ExitCode::SUCCESS)
+}
+
+#[cfg(unix)]
+fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
+    let cargo_home = process.cargo_home()?;
+    utils::remove_dir("cargo_home", &cargo_home)
 }
 
 #[derive(Clone, Copy, Debug)]
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -91,7 +91,7 @@ pub use windows::complete_windows_uninstall;
 #[cfg(all(windows, feature = "test"))]
 pub use windows::{RegistryGuard, RegistryValueId, USER_PATH, get_path};
 #[cfg(windows)]
-use windows::{delete_rustup_and_cargo_home, do_add_to_path, do_remove_from_path};
+use windows::{do_add_to_path, do_remove_from_path};
 #[cfg(windows)]
 pub(crate) use windows::{run_update, self_replace};
 
@@ -1163,7 +1163,6 @@ pub(crate) fn uninstall(
     Ok(ExitCode::SUCCESS)
 }
 
-#[cfg(unix)]
 fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
     let cargo_home = process.cargo_home()?;
     utils::remove_dir("cargo_home", &cargo_home)
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -553,8 +553,8 @@ fn canonical_cargo_home(process: &Process) -> Result<Cow<'static, str>> {
 }
 
 /// Installing is a simple matter of copying the running binary to
-/// `CARGO_HOME`/bin, hard-linking the various Rust tools to it,
-/// and adding `CARGO_HOME`/bin to PATH.
+/// `$CARGO_HOME/bin`, hard-linking the various Rust tools to it,
+/// and adding `$CARGO_HOME/bin` to PATH.
 pub(crate) async fn install(
     no_prompt: bool,
     mut opts: InstallOpts<'_>,
@@ -1096,11 +1096,31 @@ pub(crate) fn uninstall(
         utils::remove_dir("rustup_home", &rustup_dir)?;
     }
 
+    info!("removing rustup binaries");
+
+    // Delete rustup.
+    #[cfg(unix)]
+    delete_rustup_and_cargo_home(no_modify_path, process)?;
+    // NOTE: On windows, this is tricky because this is *probably*
+    // the running executable and on Windows can't be unlinked until
+    // the process exits.
+    // see: windows::{complete_windows_uninstall,spawn_uninstall_gc}
+    #[cfg(windows)]
+    windows::spawn_uninstall_gc(no_modify_path, process)?;
+
+    info!("rustup is uninstalled");
+
+    Ok(ExitCode::SUCCESS)
+}
+
+fn delete_rustup_and_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
+    let cargo_home = process.cargo_home()?;
+
     info!("removing cargo home");
 
-    // Delete everything in CARGO_HOME *except* the rustup bin
+    // Delete everything in CARGO_HOME *except* the rustup bin.
 
-    // First everything except the bin directory
+    // First everything except the bin directory.
     let diriter = fs::read_dir(&cargo_home).map_err(|e| CliError::ReadDirError {
         p: cargo_home.clone(),
         source: e,
@@ -1147,25 +1167,6 @@ pub(crate) fn uninstall(
         }
     }
 
-    info!("removing rustup binaries");
-
-    // Delete rustup.
-    #[cfg(unix)]
-    delete_rustup_and_cargo_home(no_modify_path, process)?;
-    // NOTE: On windows, this is tricky because this is *probably*
-    // the running executable and on Windows can't be unlinked until
-    // the process exits.
-    // see: windows::{complete_windows_uninstall,spawn_uninstall_gc}
-    #[cfg(windows)]
-    windows::spawn_uninstall_gc(no_modify_path, process)?;
-
-    info!("rustup is uninstalled");
-
-    Ok(ExitCode::SUCCESS)
-}
-
-fn delete_rustup_and_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
-    let cargo_home = process.cargo_home()?;
     if !no_modify_path {
         do_remove_from_path(process)?;
     }
@@ -1210,7 +1211,7 @@ pub(crate) fn self_update_permitted(explicit: bool) -> Result<SelfUpdatePermissi
     Ok(SelfUpdatePermission::Permit)
 }
 
-/// Self update downloads rustup-init to `CARGO_HOME`/bin/rustup-init
+/// Self update downloads rustup-init to `$CARGO_HOME/bin/rustup-init`
 /// and runs it.
 ///
 /// It does a few things to accommodate self-delete problems on windows:
@@ -1223,7 +1224,7 @@ pub(crate) fn self_update_permitted(explicit: bool) -> Result<SelfUpdatePermissi
 ///
 /// Because it's again difficult for rustup-init to delete itself
 /// (and on windows this process will not be running to do it),
-/// rustup-init is stored in `CARGO_HOME`/bin, and then deleted next
+/// rustup-init is stored in `$CARGO_HOME/bin`, and then deleted next
 /// time rustup runs.
 pub(crate) async fn update(cfg: &Cfg<'_>) -> Result<ExitCode> {
     common::warn_if_host_is_emulated(cfg.process);
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -1092,11 +1092,6 @@ pub(crate) fn uninstall(
 
     info!("removing cargo home");
 
-    // Remove CARGO_HOME/bin from PATH
-    if !no_modify_path {
-        do_remove_from_path(process)?;
-    }
-
     // Delete everything in CARGO_HOME *except* the rustup bin
 
     // First everything except the bin directory
@@ -1150,21 +1145,25 @@ pub(crate) fn uninstall(
 
     // Delete rustup.
     #[cfg(unix)]
-    delete_rustup_and_cargo_home(process)?;
+    delete_rustup_and_cargo_home(no_modify_path, process)?;
     // NOTE: On windows, this is tricky because this is *probably*
     // the running executable and on Windows can't be unlinked until
     // the process exits.
     // see: windows::{complete_windows_uninstall,spawn_uninstall_gc}
     #[cfg(windows)]
-    windows::spawn_uninstall_gc(process)?;
+    windows::spawn_uninstall_gc(no_modify_path, process)?;
 
     info!("rustup is uninstalled");
 
     Ok(ExitCode::SUCCESS)
 }
 
-fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
+fn delete_rustup_and_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
     let cargo_home = process.cargo_home()?;
+    if !no_modify_path {
+        do_remove_from_path(process)?;
+    }
+
     utils::remove_dir("cargo_home", &cargo_home)
 }
 
--- a/src/cli/self_update/unix.rs
+++ b/src/cli/self_update/unix.rs
@@ -134,7 +134,7 @@ pub(crate) fn run_update(setup_path: &Path) -> Result<utils::ExitCode> {
 }
 
 /// This function is as the final step of a self-upgrade. It replaces
-/// `CARGO_HOME`/bin/rustup with the running exe, and updates the
+/// `$CARGO_HOME/bin/rustup` with the running exe, and updates the
 /// links to it.
 pub(crate) fn self_replace(process: &Process) -> Result<utils::ExitCode> {
     install_bins(process)?;
--- a/src/cli/self_update/unix.rs
+++ b/src/cli/self_update/unix.rs
@@ -47,11 +47,6 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
     Ok(utils::ExitCode(0))
 }
 
-pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
-    let cargo_home = process.cargo_home()?;
-    utils::remove_dir("cargo_home", &cargo_home)
-}
-
 pub(crate) fn do_remove_from_path(process: &Process) -> Result<()> {
     for sh in shell::get_available_shells(process) {
         let source_bytes = format!("{}\n", sh.source_string(process)?).into_bytes();
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -644,10 +644,10 @@ pub(crate) fn self_replace(process: &Process) -> Result<utils::ExitCode> {
     Ok(utils::ExitCode(0))
 }
 
-// The last step of uninstallation is to delete *this binary*,
-// rustup.exe and the CARGO_HOME that contains it. On Unix, this
-// works fine. On Windows you can't delete files while they are open,
-// like when they are running.
+// Spawn a temporary rustup-gc-$random.exe to finish Windows uninstall
+// after the original rustup.exe process exits. On Unix, the running
+// executable can be deleted directly. On Windows you can't delete files
+// while they are open, like when they are running.
 //
 // Here's what we're going to do:
 // - Copy rustup.exe to a temporary file in
@@ -674,7 +674,7 @@ pub(crate) fn self_replace(process: &Process) -> Result<utils::ExitCode> {
 //
 // .. augmented with this SO answer
 // https://stackoverflow.com/questions/10319526/understanding-a-self-deleting-program-in-c
-pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
+pub(crate) fn spawn_uninstall_gc(process: &Process) -> Result<()> {
     use std::io;
     use std::ptr;
     use std::thread;
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -358,7 +358,7 @@ pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode> 
     let no_modify_path = process.var_os(GC_MODIFY_PATH).as_deref() != Some(OsStr::new("1"));
 
     // Now that the parent has exited there are hopefully no more files open in CARGO_HOME.
-    super::delete_rustup_and_cargo_home(no_modify_path, process)?;
+    super::clean_cargo_home(no_modify_path, process)?;
 
     // Now, run a *system* binary to inherit the DELETE_ON_CLOSE
     // handle to *this* process, then exit. The OS will delete the gc
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -355,8 +355,10 @@ pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode> 
 
     wait_for_parent()?;
 
+    let no_modify_path = process.var_os(GC_MODIFY_PATH).as_deref() != Some(OsStr::new("1"));
+
     // Now that the parent has exited there are hopefully no more files open in CARGO_HOME.
-    super::delete_rustup_and_cargo_home(process)?;
+    super::delete_rustup_and_cargo_home(no_modify_path, process)?;
 
     // Now, run a *system* binary to inherit the DELETE_ON_CLOSE
     // handle to *this* process, then exit. The OS will delete the gc
@@ -673,7 +675,7 @@ pub(crate) fn self_replace(process: &Process) -> Result<utils::ExitCode> {
 //
 // .. augmented with this SO answer
 // https://stackoverflow.com/questions/10319526/understanding-a-self-deleting-program-in-c
-pub(crate) fn spawn_uninstall_gc(process: &Process) -> Result<()> {
+pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Result<()> {
     use std::io;
     use std::ptr;
     use std::thread;
@@ -733,6 +735,7 @@ pub(crate) fn spawn_uninstall_gc(process: &Process) -> Result<()> {
     };
 
     Command::new(gc_exe)
+        .env(GC_MODIFY_PATH, if no_modify_path { "0" } else { "1" })
         .spawn()
         .context(CliError::WindowsUninstallMadness)?;
 
@@ -747,6 +750,10 @@ pub(crate) fn spawn_uninstall_gc(process: &Process) -> Result<()> {
 
     Ok(())
 }
+
+// The rustup-gc executable cannot accept normal function call here,
+// so we use env var here, notifying it if we need to remove $CARGO_HOME/bin from $PATH
+const GC_MODIFY_PATH: &str = "RUSTUP_GC_MODIFY_PATH";
 
 #[cfg(any(test, feature = "test"))]
 pub fn get_path() -> Result<Option<Value>> {
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -355,9 +355,8 @@ pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode> 
 
     wait_for_parent()?;
 
-    // Now that the parent has exited there are hopefully no more files open in CARGO_HOME
-    let cargo_home = process.cargo_home()?;
-    utils::remove_dir("cargo_home", &cargo_home)?;
+    // Now that the parent has exited there are hopefully no more files open in CARGO_HOME.
+    super::delete_rustup_and_cargo_home(process)?;
 
     // Now, run a *system* binary to inherit the DELETE_ON_CLOSE
     // handle to *this* process, then exit. The OS will delete the gc
--- a/tests/suite/cli_self_upd.rs
+++ b/tests/suite/cli_self_upd.rs
@@ -347,7 +347,15 @@ async fn uninstall_doesnt_leave_gc_file() {
 fn ensure_empty(dir: &Path) -> Result<(), GcErr> {
     let garbage = fs::read_dir(dir)
         .unwrap()
-        .map(|d| d.unwrap().path().to_string_lossy().to_string())
+        .filter_map(|entry| {
+            let path = entry.unwrap().path();
+            let name = path.file_name()?.to_str()?;
+            // On Windows, this binary is cleaned up on exit
+            if !(name.starts_with("rustup-gc-") && name.ends_with(EXE_SUFFIX)) {
+                return None;
+            }
+            Some(path.to_string_lossy().to_string())
+        })
         .collect::<Vec<_>>();
     match garbage.len() {
         0 => Ok(()),
--- a/tests/suite/cli_self_upd.rs
+++ b/tests/suite/cli_self_upd.rs
@@ -8,6 +8,7 @@ use std::process::Command;
 
 use remove_dir_all::remove_dir_all;
 
+#[cfg(windows)]
 use retry::{
     delay::{Fibonacci, jitter},
     retry,
@@ -368,6 +369,7 @@ async fn uninstall_self_delete_works() {
 // On windows rustup self uninstall temporarily puts a rustup-gc-$randomnumber.exe
 // file in CONFIG.CARGODIR/.. ; check that it doesn't exist.
 #[tokio::test]
+#[cfg(windows)]
 async fn uninstall_doesnt_leave_gc_file() {
     let cx = setup_empty_installed().await;
     cx.config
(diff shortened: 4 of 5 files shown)


tests/suite/cli_self_upd.rs
```diff
@@ -300,7 +290,6 @@ async fn uninstall_self_delete_works() {
 
     assert!(out.status.success());
     assert!(!rustup.exists());
-    assert!(!cx.config.cargodir.exists());
```
FranciscoTGouveia, line 303: Shouldn't this assert be inverted instead of removed?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..07d31c8d31dcc43e6de246150cc7e15033c46e6a)*

Cloud0310: Yeap, could be better

src/cli/self_update/unix.rs
```diff
@@ -47,9 +49,31 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
     Ok(utils::ExitCode(0))
 }
 
-pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
-    let cargo_home = process.cargo_home()?;
-    utils::remove_dir("cargo_home", &cargo_home)
+pub(crate) fn remove_rustup_executable(process: &Process, no_modify_path: bool) -> Result<()> {
```
rami3l: Maybe `delete_rustup_and_bin_dir()` would be a better name for this, wdyt?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..b6195c7eff9a9da65b81e91a323241ea1008bf69)*

src/cli/self_update/unix.rs
```diff
@@ -47,9 +49,31 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
...
+        Ok(()) => {
+            if !no_modify_path {
+                do_remove_from_path(process)?;
+                info!("remove cargo bin from path");
+            }
+
+            info!("removed empty cargo bin");
+        }
+        Err(e) => {
+            let cargo_bin_path = cargo_bin_path.display();
```
rami3l: We now assume that the bindir can be left not empty, so we'd only warn the user here if `e` is of the nonempty kind, and only for the other errors that may happen should we actually throw the error.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..b6195c7eff9a9da65b81e91a323241ea1008bf69)*

tests/suite/cli_self_upd.rs
```diff
@@ -337,17 +326,24 @@ async fn uninstall_doesnt_leave_gc_file() {
...
 }
 
-fn ensure_empty(dir: &Path) -> Result<(), GcErr> {
+fn ensure_no_gc_file(dir: &Path) -> Result<(), GcErr> {
     let garbage = fs::read_dir(dir)
         .unwrap()
-        .map(|d| d.unwrap().path().to_string_lossy().to_string())
+        .filter_map(|entry| {
+            let path = entry.unwrap().path();
+            let is_gc_file = path
```
rami3l: Nit: Prefer direct style here, something like:

```rs
let name = path.file_name()?.to_str()?;
if !(name.starts_with("rustup-gc-") && name.ends_with(EXE_SUFFIX)) { return None; }
Some(path.to_string_lossy().to_string())
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..b6195c7eff9a9da65b81e91a323241ea1008bf69)*

src/cli/self_update.rs
```diff
@@ -1090,71 +1094,30 @@ pub(crate) fn uninstall(
...
-        if file_is_tool == Some(false) {
-            if dirent.path().is_dir() {
-                utils::remove_dir("cargo_home", &dirent.path())?;
-            } else {
-                utils::remove_file("cargo_home", &dirent.path())?;
-            }
+        .map(|tool| cargo_bin_path.join(format!("{tool}{EXE_SUFFIX}")));
+
+    for proxy_path in proxy_paths {
+        if matches!((
```
rami3l: Nit: This is written as `is_same_file(&rustup_path, &proxy_path)` elsewhere in the codebase.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..b6195c7eff9a9da65b81e91a323241ea1008bf69)*

src/cli/self_update/unix.rs
```diff
@@ -47,9 +49,31 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
...
-pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
-    let cargo_home = process.cargo_home()?;
-    utils::remove_dir("cargo_home", &cargo_home)
+pub(crate) fn remove_rustup_executable(process: &Process, no_modify_path: bool) -> Result<()> {
+    let cargo_bin_path = process.cargo_home()?.join("bin");
+    let rustup_path = cargo_bin_path.join(format!("rustup{EXE_SUFFIX}"));
+
+    utils::remove_file("rustup_bin", &rustup_path)?;
+
+    match fs::remove_dir(&cargo_bin_path) {
```
rami3l: Try reducing rightward drift here:

```rs
if let Err(e) = fs::remove_dir(&cargo_bin_path) {
  // sad path; handle the error and bail out
  // must early return before here
}

// happy path
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..b6195c7eff9a9da65b81e91a323241ea1008bf69)*

src/cli/self_update/windows.rs
```diff
@@ -371,6 +378,25 @@ pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode>
         .spawn()
         .context(CliError::WindowsUninstallMadness)?;
 
+    // clean up PATH and dir
```
rami3l: Nit: Maybe extract this as a helper function?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..b6195c7eff9a9da65b81e91a323241ea1008bf69)*

Cloud0310:
I think I may need more tests against this PR, this includes:
- `$CARGO_HOME/bin` with unrelated files
- empty bin cleanup test

Should there be more tests? Please let me know @FranciscoTGouveia @rami3l .

src/cli/self_update.rs
```diff
@@ -1048,6 +1046,11 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     Ok(())
 }
 
+// Uninstall process:
```
rami3l: Nit: Now that we are halfway there, consider turning this into actual doc comments.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..b6195c7eff9a9da65b81e91a323241ea1008bf69)*

src/cli/self_update/unix.rs
Cloud0310: Yeap, should have this one as warn.

src/cli/self_update/unix.rs
Cloud0310: Reasonable to me, I guess we may even need to add extra comments for this function?
It's actually does:
- remove executable
- cleanup `$CARGO_HOME/bin` on empty
- cleanup `$PATH`

So, I propose the new name as `delete_exe_and_cargo_bin`?
And we should add comments for this function for clarification.

tests/suite/cli_self_upd.rs
Cloud0310: Thanks for the tip, I'll fix.

src/cli/self_update/windows.rs
Cloud0310: I think just two dups won't harm anything here?

src/cli/self_update/unix.rs
djc: I think the advice to reduce rightward drift is good but your example case has comments that are inconsistent with the code? The happy path would go inside the `else` block in this case, leaving the remainder of the function to properly handle the error.

src/cli/self_update/windows.rs
```diff
@@ -347,17 +350,21 @@ fn has_windows_sdk_libs(process: &Process) -> bool {
     };
     false
 }
-
-/// Run by rustup-gc-$num.exe to delete CARGO_HOME
+/// Run by rustup-gc-$num.exe to delete rustup bianry
```
djc: We'll want to keep the empty line after the previous function.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..48acf08815e5e4c293534530174e139c7ac8e7da)*

src/cli/self_update/windows.rs
```diff
@@ -371,6 +378,31 @@ pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode>
...
+            warn!("not removing cargo bin `{cargo_bin_path}` because it is not empty");
+        }
+        Err(e) => {
+            let cargo_bin_path = cargo_bin_path.display();
+
+            return Err(e).with_context(|| {
+                format!("failed to remove cargo bin directory `{cargo_bin_path}`")
+            });
+        }
+    }
```
djc: This looks very similar to the code in `unix`. Can we deduplicate?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..48acf08815e5e4c293534530174e139c7ac8e7da)*

src/cli/self_update/windows.rs
```diff
@@ -645,7 +677,7 @@ pub(crate) fn self_replace(process: &Process) -> Result<utils::ExitCode> {
 }
 
 // The last step of uninstallation is to delete *this binary*,
-// rustup.exe and the CARGO_HOME that contains it. On Unix, this
+// rustup.exe. On Unix, this
```
djc: Reflow the comment?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..48acf08815e5e4c293534530174e139c7ac8e7da)*

src/cli/self_update.rs
```diff
@@ -1048,6 +1046,12 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     Ok(())
 }
 
+/// Uninstall process:
+/// Remove rustup home
+/// Clean up rustup tool proxys
+/// Remove rustup binary file
+/// Try to clean up $CARGO_HOME/bin if it's empty
+/// Upon successfully removing the $CARGO_HOME/bin, clean up $PATH
```
djc: If you're going to list bullets, make it an actual Markdown bullet list?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..48acf08815e5e4c293534530174e139c7ac8e7da)*

src/cli/self_update.rs
```diff
@@ -1181,7 +1144,7 @@ pub(crate) fn self_update_permitted(explicit: bool) -> Result<SelfUpdatePermissi
         .tempdir_in(current_exe_dir)
     {
         match e.kind() {
-            io::ErrorKind::PermissionDenied => {
+            std::io::ErrorKind::PermissionDenied => {
```
djc: Why? This seems worse.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..48acf08815e5e4c293534530174e139c7ac8e7da)*

src/cli/self_update/unix.rs
rami3l: @djc I was reading this part of the comments: 

```rs
// Clean up CARGO_HOME/bin
// On success, remove it from $PATH
```

... so when I mentioned the "happy path" I actually meant the next step of the cleanup that should be after the `if let` part:

```rs
do_remove_from_path(process)?;
```

But looking at this new version I think you may be suggesting using let-else instead of if-let, something like this?

```rs
let Err(e) = fs::remove_dir(&cargo_bin_path) else {
  // Happy path, continue with the cleanup
}

if e.kind() == std::io::ErrorKind::DirectoryNotEmpty {
  // Not throwing
  return Ok(());
}

// Throwing?
```

src/cli/self_update/unix.rs
FranciscoTGouveia: I may be overlooking something, but `$CARGO_HOME/bin` will never be empty.
The `rustup` binary will always be there, right?

rami3l: @FranciscoTGouveia The original intention at least is that on Unix we should remove the rustup shims in one go; on Windows we remove all the rustup shims except the potentially current running binary which then gets removed in another process.

When the other review comments are addressed I think we can do a final round of check about this.

Cloud0310:
I've add more tests about the PR here. Please review.

src/cli/self_update/unix.rs
rami3l: > So, I propose the new name as delete_exe_and_cargo_bin?
> And we should add comments for this function for clarification.

@Cloud0310 Looks reasonable to me, although I'd prefer `delete_rustup_exe_and_cargo_bin()` for extra clarity :)

Also the `cargo_bin` part is okay for now, but when you move to XDG you'll need to change that accordingly.

src/cli/self_update/unix.rs
```diff
@@ -47,9 +48,16 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
...
-    utils::remove_dir("cargo_home", &cargo_home)
+pub(crate) fn cleanup_cargo_bin(process: &Process, no_modify_path: bool) -> Result<()> {
+    use super::do_cargo_bin_clean;
+    let cargo_bin_path = process.cargo_home()?.join("bin");
+    let rustup_path = cargo_bin_path.join(format!("rustup{EXE_SUFFIX}"));
+
+    utils::remove_file("rustup_bin", &rustup_path)?;
+
+    // Clean up CARGO_HOME/bin, if it's empty now
+    // On success, remove it from $PATH
```
rami3l: Nit: Suggest extracting this part as doc comments on the newly-extracted `do_cargo_bin_clean()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..905c540e57435e8c2c8642da178695f50a47098e)*

src/cli/self_update.rs
```diff
@@ -1048,6 +1046,39 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     Ok(())
 }
 
+pub(crate) fn do_cargo_bin_clean(
```
rami3l: Nit: For the sake of consistency, maybe rename this to something like `delete_cargo_bin()`?

tests/suite/cli_self_upd.rs
```diff
@@ -259,13 +259,121 @@ async fn uninstall_works_if_rustup_home_doesnt_exist() {
...
+        .is_ok();
+
+    let check = || {
+        if cargo_bin.exists() {
+            Err(format!("cargo bin still exists: {}", cargo_bin.display()))
+        } else {
+            Ok(())
+        }
+    };
+    match retry(Fibonacci::from_millis(1).map(jitter).take(23), check) {
```
rami3l: Nit: Isn't this just `.unwrap()`?

tests/suite/cli_self_upd.rs
```diff
@@ -259,13 +259,13 @@ async fn uninstall_works_if_rustup_home_doesnt_exist() {
 }
 
 #[tokio::test]
-async fn uninstall_deletes_cargo_home() {
```
rami3l, line 262: I think you can just squash all commits into one since all of them seems to be focusing on a single topic.

rami3l (approved):
LGTM modulo some tiny nits, nice work :D

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/5862009d0cbf29699ab5818a95d9ade82e126c61..905c540e57435e8c2c8642da178695f50a47098e)*

src/cli/self_update/unix.rs
rami3l: Finally adjusted to `clean_cargo_bin()` which may call `delete_cargo_bin()`.

src/cli/self_update/unix.rs
```diff
@@ -47,9 +48,16 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
     Ok(utils::ExitCode(0))
 }
 
-pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
-    let cargo_home = process.cargo_home()?;
-    utils::remove_dir("cargo_home", &cargo_home)
+pub(crate) fn clean_cargo_bin(process: &Process, no_modify_path: bool) -> Result<()> {
+    use super::delete_cargo_bin;
```
djc: Nit: keep imports at the top of the module.

src/cli/self_update/windows.rs
```diff
@@ -26,6 +26,8 @@ use crate::download::DownloadOptions;
 use crate::process::{ColorableTerminal, Process};
 use crate::utils;
 
+const GC_NO_MODIFY_PATH: &str = "RUSTUP_GC_NO_MODIFY_PATH";
```
djc: Nit: keep `const`s at the bottom of the module.

src/cli/self_update/windows.rs
```diff
@@ -348,16 +350,26 @@ fn has_windows_sdk_libs(process: &Process) -> bool {
     false
 }
 
-/// Run by rustup-gc-$num.exe to delete CARGO_HOME
+/// Run by rustup-gc-$num.exe to delete rustup binary
 #[tracing::instrument(level = "trace")]
 pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode> {
+    use super::delete_cargo_bin;
```
djc: Nit: move this to the top. (Also, the pre-existing import below.)

src/cli/self_update/windows.rs
```diff
@@ -348,16 +350,26 @@ fn has_windows_sdk_libs(process: &Process) -> bool {
     false
 }
 
-/// Run by rustup-gc-$num.exe to delete CARGO_HOME
+/// Run by rustup-gc-$num.exe to delete rustup binary
```
djc: This seems unrelated, maybe move it into a separate commit?

djc (commented):
I think this PR needs a bunch of work, it seems to have a lot of changes together in a single commit that are pretty messy. It would be good if the commit history made it clear that there are a sequence of smaller changes being made.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/12f54158bcef3c9845f987e6ff72139790139922..61b3e49fc7af4c1617717986ee11d43f9fcfe70e)*

src/cli/self_update/windows.rs
rami3l: @djc I think this is relevant though; the new behavior is to remove traces of rustup in the bindir rather than nuking CARGO_HOME.

Cloud0310:
> I think this PR needs a bunch of work, it seems to have a lot of changes together in a single commit that are pretty messy. It would be good if the commit history made it clear that there are a sequence of smaller changes being made.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/12f54158bcef3c9845f987e6ff72139790139922..61b3e49fc7af4c1617717986ee11d43f9fcfe70e)_

I have re-writed the history here. Are they fine now?

rami3l:
> > I think this PR needs a bunch of work, it seems to have a lot of changes together in a single commit that are pretty messy. It would be good if the commit history made it clear that there are a sequence of smaller changes being made.
> > _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/12f54158bcef3c9845f987e6ff72139790139922..61b3e49fc7af4c1617717986ee11d43f9fcfe70e)_
> 
> I have re-writed the history here. Are they fine now?

@Cloud0310 That is looking better, but still I think we should further split the 2nd commit. An immediate optimization point I can see for example is to make the introduction of the `no_modify_path` param more natural. You can use a first commit to move the actual PATH manipulation in the cleanup function that you've just renamed to `clean_cargo_bin()`, and that would justify the introduction of the param in the same commit, and then you can do similar things with the rest of your changes until each commit feels as boring as possible, WDYT?

cc @djc

Cloud0310:
The splitting is done. Do I need further splitting with tests?
If so, it could be more splitted into a) updated expected values, b) new tests upon keeping `$CARGO_HOME`.

src/cli/self_update.rs
```diff
@@ -1090,69 +1095,58 @@ pub(crate) fn uninstall(
...
+    no_modify_path: bool,
+    process: &Process,
+) -> Result<()> {
+    let Err(e) = fs::remove_dir(&cargo_bin_path) else {
+        if !no_modify_path {
+            do_remove_from_path(process)?;
+            info!("removed cargo bin from $PATH");
+        }
+
+        info!("removed empty cargo bin");
```
rami3l: Nit: The way you refer to the bindir is not consistent both in this function and elsewhere:

- `cargo bin`
- `` cargo bin `{cargo_bin_path}` ``
- `` cargo bin directory `{cargo_bin_path}` ``

I suggest using the final one consistently.

src/cli/self_update.rs
```diff
@@ -1090,69 +1095,58 @@ pub(crate) fn uninstall(
...
+        return Ok(());
+    };
+
+    let cargo_bin_path = cargo_bin_path.display();
+    if e.kind() == io::ErrorKind::DirectoryNotEmpty {
+        warn!("not removing cargo bin `{cargo_bin_path}` because it is not empty");
```
rami3l: Nit: Avoid using double negations here. Try ``` kept non-empty cargo bin directory `{cargo_bin_path}` ```

src/cli/self_update.rs
```diff
@@ -1090,69 +1095,58 @@ pub(crate) fn uninstall(
...
+/// If the directory is not empty, emit a warning and return success.
+pub(crate) fn delete_cargo_bin(
+    cargo_bin_path: PathBuf,
+    no_modify_path: bool,
+    process: &Process,
+) -> Result<()> {
+    let Err(e) = fs::remove_dir(&cargo_bin_path) else {
+        if !no_modify_path {
+            do_remove_from_path(process)?;
+            info!("removed cargo bin from $PATH");
```
rami3l: A very important rule is that the logs must happen before (NOT AFTER) the action, so that when the action fails, the user has a clue about what the application was trying to do.

Thus, prefer using `removing...` instead of `removed ...`.

src/cli/self_update/windows.rs
```diff
@@ -738,6 +746,7 @@ pub(crate) fn clean_cargo_bin(process: &Process, no_modify_path: bool) -> Result
     };
 
     Command::new(gc_exe)
+        .env(GC_NO_MODIFY_PATH, if no_modify_path { "1" } else { "0" })
```
rami3l: Nit: As per our [style guide](https://rust-lang.github.io/rustup/dev-guide/coding-standards.html#prefer-positive-booleans), newly-introduced variable names (of both Rust and environment variables) should not include any negations, so `*NO_*` is not permitted; try `GC_MODIFY_PATH`.

src/cli/self_update/windows.rs
```diff
@@ -348,16 +349,25 @@ fn has_windows_sdk_libs(process: &Process) -> bool {
...
 #[tracing::instrument(level = "trace")]
 pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode> {
     use std::process::Stdio;
 
     wait_for_parent()?;
 
-    // Now that the parent has exited there are hopefully no more files open in CARGO_HOME
-    let cargo_home = process.cargo_home()?;
-    utils::remove_dir("cargo_home", &cargo_home)?;
+    // Now that the parent has exited, rustup.exe is hopefully no longer open.
```
rami3l: This new hunk has the same contents as `unix::clean_cargo_bin()` except the `let no_modify_path` line; consider further polishing the refactoring process by first moving `unix::clean_cargo_bin()` to the parent module and inline your current `delete_cargo_bin()` over there (it should have a single caller by then so the introduction of this new function `delete_cargo_bin()` is unnecessary).

rami3l (commented):
@Cloud0310 It is looking better, but some rework remains to be done. Especially, the 3rd commit (which used to be the 2nd commit in the last version) is still not boring enough.

> The splitting is done. Do I need further splitting with tests?
If so, it could be more splitted into a) updated expected values, b) new tests upon keeping $CARGO_HOME.

Please perform the split you mentioned above as well.

 Many thanks in advance :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..9f17fce2209e2fc14f24d1488ac5a37dd81bcf8c)*

src/cli/self_update.rs
```diff
@@ -1090,69 +1095,58 @@ pub(crate) fn uninstall(
...
-        })?;
-        let name = dirent.file_name();
-        let file_is_tool = name.to_str().map(|n| tools.iter().any(|t| *t == n));
-        if file_is_tool == Some(false) {
-            if dirent.path().is_dir() {
-                utils::remove_dir("cargo_home", &dirent.path())?;
-            } else {
-                utils::remove_file("cargo_home", &dirent.path())?;
-            }
+        .map(|tool| cargo_bin_path.join(format!("{tool}{EXE_SUFFIX}")));
```
rami3l: Nit: This line of diff seems redundant.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..9f17fce2209e2fc14f24d1488ac5a37dd81bcf8c)*

src/cli/self_update/unix.rs
```diff
@@ -47,8 +47,12 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
     Ok(utils::ExitCode(0))
 }
 
-pub(crate) fn clean_cargo_bin(process: &Process) -> Result<()> {
+pub(crate) fn clean_cargo_bin(process: &Process, no_modify_path: bool) -> Result<()> {
```
djc: Nit: change the argument order (for this and other functions with the added argument), to have the more "interesting"/rare/specific argument before the less interesting/more common/less specific `process` argument.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..6e1aa36d22452bdb121fb2bb6abf3d49966f448e)*

src/cli/self_update/unix.rs
```diff
@@ -1,9 +1,11 @@
+use std::env::consts::EXE_SUFFIX;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use anyhow::{Context, Result, bail};
 use tracing::{error, warn};
 
+use super::delete_cargo_bin;
```
djc: Nit: this should be included in the existing `super::` import, below.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..6e1aa36d22452bdb121fb2bb6abf3d49966f448e)*

src/cli/self_update/unix.rs
```diff
@@ -48,12 +50,14 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
...
-    }
+    let cargo_bin_path = process.cargo_home()?.join("bin");
+    let rustup_path = cargo_bin_path.join(format!("rustup{EXE_SUFFIX}"));
+
+    utils::remove_file("rustup_bin", &rustup_path)?;
 
-    utils::remove_dir("cargo_home", &cargo_home)
+    // Clean up CARGO_HOME/bin, if it's empty now
+    // On success, remove it from $PATH
+    delete_cargo_bin(cargo_bin_path, no_modify_path, process)
```
djc: It is unclear to me why there are separate commits for the Unix and Windows paths? It seems to me like these should be in the same commit.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..6e1aa36d22452bdb121fb2bb6abf3d49966f448e)*

src/cli/self_update.rs
```diff
@@ -1104,67 +1102,45 @@ pub(crate) fn uninstall(
...
-            source: e,
-        })?;
-        if dirent.file_name().to_str() != Some("bin") {
-            if dirent.path().is_dir() {
-                utils::remove_dir("cargo_home", &dirent.path())?;
-            } else {
-                utils::remove_file("cargo_home", &dirent.path())?;
-            }
-        }
-    }
```
djc, line 1119: It is not clear from reviewing this (presumably Unix-only) commit why this (presumably platform-independent) logic could be removed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..6e1aa36d22452bdb121fb2bb6abf3d49966f448e)*

src/cli/self_update/windows.rs
```diff
@@ -18,6 +18,7 @@ use windows_sys::Win32::Foundation::{ERROR_FILE_NOT_FOUND, ERROR_INVALID_DATA};
 
 use super::super::errors::CliError;
 use super::common;
+use super::delete_cargo_bin;
```
djc: Nit: use existing import.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..6e1aa36d22452bdb121fb2bb6abf3d49966f448e)*

src/cli/self_update/windows.rs
```diff
@@ -374,6 +384,8 @@ pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode>
     Ok(utils::ExitCode(0))
 }
 
+const GC_MODIFY_PATH: &str = "RUSTUP_GC_MODIFY_PATH";
```
djc: Perhaps this should be introduced independently to make it more clear why it is introduced/what it's purpose is? (It should be defined below all callers.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..6e1aa36d22452bdb121fb2bb6abf3d49966f448e)*

src/cli/self_update.rs
```diff
@@ -1046,6 +1046,12 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     Ok(())
 }
 
+/// Uninstall process:
+/// 1. Remove rustup home.
+/// 2. Clean up rustup tool proxies.
+/// 3. Remove rustup binary file.
+/// 4. Try to clean up $CARGO_HOME/bin if it's empty.
+/// 5. Upon successfully removing $CARGO_HOME/bin, clean up $PATH.
```
djc, line 1057: Suggest introducing this in a separate first commit, and then updating it in each commit (where necessary) as a way of documenting what is changed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/7f71766866193e4aed0873da3ffd79232bc039c3..6e1aa36d22452bdb121fb2bb6abf3d49966f448e)*

src/cli/self_update/unix.rs
rami3l: @djc I'm not quite sure what you are trying to say here, I assume you mean "in a first commit merge those in two groups of imports `super::{...}` and `crate::{...}`, then introduce the new import in the respective group"? Because in this file the imports are not grouped yet.

src/cli/self_update/windows.rs
rami3l: @Cloud0310 Same as https://github.com/rust-lang/rustup/pull/4861#discussion_r3334336722.

src/cli/self_update/unix.rs
djc: I mean that this is adding a `use super::delete_cargo_bin;` when there is already a `use super::some_other_thing;`.

rami3l: @djc Okay, so we should put the single-level imports like this `use super::{delete_cargo_bin, some_other_thing};`

src/cli/self_update/windows.rs
rami3l: @Cloud0310 Please delete the function `delete_cargo_bin`. As previously discussed, `unix::clean_cargo_bin()` should be moved to the parent module that contains logic common to both Windows and Unix (which includes the logic of `delete_cargo_bin` which is not repeated anywhere else); in Windows we use a platform specific wrapper to wrap that function in the `windows` module.

rami3l:
@Cloud0310 Since your PR needs more work, I've marked this PR as draft.

Please address or at least respond to all open threads before marking this PR as "ready for review" again, at which point the team will return for another round of review, thanks 🙏

src/cli/self_update/unix.rs
Cloud0310: Guess this is not related to current PR, but I would like to fix with another PR. Or we prefer fix it here?

src/cli/self_update.rs
```diff
@@ -1096,6 +1097,19 @@ pub(crate) fn uninstall(
         utils::remove_dir("rustup_home", &rustup_dir)?;
     }
 
+    // Clean up rustup tool links
```
rami3l: Nit: Logically this snippet should be placed after the "for dirent in diriter {" block, since the removal of cargo home should start after that.

src/cli/self_update/unix.rs
```diff
@@ -47,15 +47,6 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
     Ok(utils::ExitCode(0))
 }
 
-pub(crate) fn clean_cargo_bin(no_modify_path: bool, process: &Process) -> Result<()> {
```
rami3l: Nit: Could use a more straightforward commit message like ``refactor(cli/self-update): move `unix::clean_cargo_bin()` to parent module``

rami3l (approved):
This is looking much better than before! I have a few final nits; address them in addition to the remaining ones and we should be good to go 🙏 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/e3cdb718b6bef2d4f6603f799987c65b135d25b9..197f611c72503216435b31f91ba89b6ddbd487f8)*

src/cli/self_update.rs
```diff
@@ -1098,22 +1098,40 @@ pub(crate) fn uninstall(
...
     // Optionally remove it from $PATH if successfully removed.
-    #[cfg(windows)]
-    windows::clean_cargo_bin(no_modify_path, process)?;
     #[cfg(unix)]
     clean_cargo_bin(no_modify_path, process)?;
+    // NOTE: On windows, this is *tricky*,
+    // the running executable and on Windows can't be unlinked until the process exits.
+    // see: windows::{complete_windows_uninstall,clean_cargo_bin}
+    #[cfg(windows)]
+    windows::clean_cargo_bin(no_modify_path, process)?;
```
rami3l: I have the impression that we are first adding this line in a previous commit, then removing and re-adding it here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/40dcace8cd16a309fa4079b7f5a06397563c00f3..60c433c0d152f92e9fad77099ca67d85fb6a74fc)*

tests/suite/cli_self_upd.rs
```diff
@@ -331,6 +381,7 @@ fn ensure_empty(dir: &Path) -> Result<(), GcErr> {
         .filter_map(|entry| {
             let path = entry.unwrap().path();
             let name = path.file_name()?.to_str()?;
+            // On Windows, this binary is cleaned up on exit
```
rami3l, line 399: This part should also be absorbed to the commit where the next line is introduced.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/40dcace8cd16a309fa4079b7f5a06397563c00f3..60c433c0d152f92e9fad77099ca67d85fb6a74fc)*

rami3l (approved):

src/cli/self_update.rs
```diff
@@ -1151,16 +1146,25 @@ pub(crate) fn uninstall(
     // Delete rustup. This is tricky because this is *probably*
     // the running executable and on Windows can't be unlinked until
     // the process exits.
-    clean_cargo_bin(process)?;
+    #[cfg(unix)]
+    clean_cargo_bin(no_modify_path, process)?;
+    // NOTE: On windows, this is *tricky*,
+    // the running executable and on Windows can't be unlinked until the process exits.
```
rami3l: It looks like you are repeating this comment twice in the commit "refactor: pass PATH cleanup to `clean_cargo_bin`"?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/40dcace8cd16a309fa4079b7f5a06397563c00f3..ca2bba6484354b3b89ed1c076b1b425477ff8d65)*

src/cli/self_update/windows.rs
```diff
@@ -355,9 +355,10 @@ pub fn complete_windows_uninstall(process: &Process) -> Result<utils::ExitCode>
 
     wait_for_parent()?;
 
-    // Now that the parent has exited there are hopefully no more files open in CARGO_HOME
-    let cargo_home = process.cargo_home()?;
-    utils::remove_dir("cargo_home", &cargo_home)?;
```
rami3l, line 360: You can possibly further refine the history by splitting "[refactor: pass PATH cleanup to clean_cargo_bin](https://github.com/rust-lang/rustup/pull/4861/commits/0f086b30bab777fcd6739d978e4f045238addeb2)" into two commits:
1. In a first commit, replace those two lines with `super::clean_cargo_bin(process)?;` and in the parent module remove `#[cfg(unix)]` on `fn clean_cargo_bin` then add `#[cfg(windows)] windows::clean_cargo_bin(process)?;`
2. In a second commit, introduce `no_modify_path` to `fn clean_cargo_bin` in the parent module.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/40dcace8cd16a309fa4079b7f5a06397563c00f3..ca2bba6484354b3b89ed1c076b1b425477ff8d65)*

src/cli/self_update.rs
```diff
@@ -1048,6 +1048,12 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     Ok(())
 }
 
+/// Uninstall process:
+/// 1. Remove rustup home.
+/// 2. Remove cargo home except $CARGO_HOME/bin
+/// 3. Remove other tools in $CARGO_HOME/bin.
+/// 4. Remove rustup binary file and links to rustup.
+/// 5. Upon successfully removing $CARGO_HOME/bin, clean up $PATH.
```
rami3l, line 1057: You didn't mention the fact that it also removes `$CARGO_HOME` as the final step (in the commit "doc: add uninstall process overall docstring").

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/40dcace8cd16a309fa4079b7f5a06397563c00f3..ca2bba6484354b3b89ed1c076b1b425477ff8d65)*

src/cli/self_update/windows.rs
```diff
@@ -673,7 +673,7 @@ pub(crate) fn self_replace(process: &Process) -> Result<utils::ExitCode> {
 //
 // .. augmented with this SO answer
 // https://stackoverflow.com/questions/10319526/understanding-a-self-deleting-program-in-c
-pub(crate) fn clean_cargo_bin(process: &Process) -> Result<()> {
+pub(crate) fn spawn_uninstall_gc(process: &Process) -> Result<()> {
```
djc: Suggest moving this commit earlier in the series (maybe the very first) to make this more obvious during the changes made earlier.

src/cli/self_update/unix.rs
```diff
@@ -47,7 +47,7 @@ pub(crate) fn do_anti_sudo_check(no_prompt: bool, process: &Process) -> Result<u
     Ok(utils::ExitCode(0))
 }
 
-pub(crate) fn delete_rustup_and_cargo_home(process: &Process) -> Result<()> {
+pub(crate) fn clean_cargo_bin(process: &Process) -> Result<()> {
```
djc: I think this rename should be postponed until the commit where the change actually happens (from home to just bin).

src/cli/self_update.rs
```diff
@@ -1109,16 +1109,31 @@ pub(crate) fn uninstall(
...
+    // Remove rustup tool links
+    let proxy_paths = TOOLS
+        .iter()
+        .chain(DUP_TOOLS.iter())
+        .map(|tool| cargo_bin_path.join(format!("{tool}{EXE_SUFFIX}")));
+    for proxy_path in proxy_paths {
+        if is_same_file(&proxy_path, &rustup_path).unwrap_or(false) {
+            utils::remove_file("rustup tool proxy", &proxy_path)?;
+        }
+    }
```
djc: Looks like this is reintroducing code that was removed in the previous commit?

djc (commented):
Definitely getting better, still some suggested improvements though.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/835392dadb40d7c8c0926304ff52c60c562e0bc0..31432e436bc033f667ebfb748f036690a7310264)*

src/cli/self_update.rs
rami3l: @djc I just checked, and that didn't really happen (the old cleanup logic tried to exclude all rustup proxies, and the new cleanup logic tries to include all of them, plus they happened in different stages of the uninstallation flow, so those are two separate things instead of back-and-forth changes, and there is nothing to be moved around).

However there is one extra [concern](https://github.com/rust-lang/rustup/pull/4861#discussion_r3397339170) that I would like to point out while doing the check, so thanks for the comment anyways :)

src/cli/self_update.rs
```diff
@@ -1050,9 +1049,10 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
 
 /// Uninstall process:
 /// 1. Remove rustup home.
-/// 2. Remove other tools in $CARGO_HOME/bin, but skipping rustup binary file and links to rustup.
-/// 3. If no `--no-modify-path` is passed $CARGO_HOME/bin, clean up $PATH.
-/// 4. Remove $CARGO_HOME
+/// 2. Remove rustup tool links.
```
rami3l: In the commit "fix: preserve CARGO_HOME during uninstall", we are not deleting rustup tool links yet.

The comments should really reflect what the code is doing in that particular commit.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/835392dadb40d7c8c0926304ff52c60c562e0bc0..31432e436bc033f667ebfb748f036690a7310264)*

src/cli/self_update/unix.rs
rami3l: @djc For extra clarity for @Cloud0310, do you mean right before the commit X where the change happens, after X, or within X?

djc: Probably in the same commit?

src/cli/self_update.rs
Cloud0310: Yes, just as @rami3l said, to clarify, here is the behaviors.

Legacy behavior:

1. Attempt to remove everything under `$CARGO_HOME`.
2. Explicitly skip tool links to avoid the Windows file lock issue.
3. After the file lock is released, remove `$CARGO_HOME` (including the tool links) itself.

Current behavior:

1. Wait for the Windows file lock to be released.
2. Remove only the tool links.

So this is only reintroducing **tool link removal after the file lock is released**, not the broader legacy cleanup that removed everything under `$CARGO_HOME`.

src/cli/self_update/unix.rs
Cloud0310: Here, I guess I would like to keep the renaming as the first commit, as on Unix we're just removing the cargo home in the specific code.

So the change would be two of them:
1. Renaming the func as `remove_cargo_home`
2. In the fix up commit, renaming it as `clean_cargo_bin`

This sounds better to me here I guess? @rami3l @djc

src/cli/self_update/windows.rs
Cloud0310: Yes, I guess my order now would be:
1. Rename `delete_rustup_and_cargo_home` to `delete_cargo_home`
2. Renaming this func
3. Upon fix, rename `delete_cargo_home` to `clean_cargo_bin`

And Others.

This would be related to the other renaming issue I guess? 
Please tell me the my understanding of commits' order here is right for less back and forth.

src/cli/self_update/unix.rs
Cloud0310: Or, as now the previous naming is already wrong, how about we just make all the renaming(this one and the Windows one) along with the fix commit.
Would be cleaner in history here.

src/cli/self_update/windows.rs
Cloud0310: So I'd love to mark this one as resolved, and make all the discussion in the other review comment list.

src/cli/self_update.rs
```diff
@@ -1089,66 +1089,15 @@ pub(crate) fn uninstall(
...
-            source: e,
-        })?;
-        if dirent.file_name().to_str() != Some("bin") {
-            if dirent.path().is_dir() {
-                utils::remove_dir("cargo_home", &dirent.path())?;
-            } else {
-                utils::remove_file("cargo_home", &dirent.path())?;
-            }
-        }
-    }
```
rami3l, line 1119: @Cloud0310 Really sorry for the back and forth, what you have implemented here matches the expected semantics in a future version, there's nothing wrong about it; but I think it might cause a downgrade in terms of UX within the scope of this PR, so it would be better if:

- The code that "removes everything except the bin directory" is kept;
- The code that "removes everything in bin except rustup and tools" is removed as you are doing right now.

When we move further into XDG support it is indeed intended to do this more precisely, but for now it might still feel surprising if `self uninstall` is not removing a bunch of cargo-specific stuff.

The comments and the tests should be updated accordingly.

My apologies for not having noticed this in advance 🙏 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/835392dadb40d7c8c0926304ff52c60c562e0bc0..8c89d1b97b8be7fafc89888fdf63e04bd2dbd35e)*

src/cli/self_update.rs
rami3l: Similar to [`6aa45d2` (this PR)](https://github.com/rust-lang/rustup/pull/4861/changes/6aa45d21d3043cbc53773741fd9dce4ec86dce6e#r3401573254), the user might want to see `$CARGO_HOME` being removed if it ends up empty at this point. In most cases this will be a no-op.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/835392dadb40d7c8c0926304ff52c60c562e0bc0..8c89d1b97b8be7fafc89888fdf63e04bd2dbd35e)*

src/cli/self_update/unix.rs
rami3l: @Cloud0310 I think the current suggestion would be perfectly fine, because it is about an existing issue that will be fixed in this PR.

src/cli/self_update.rs
```diff
@@ -1164,6 +1164,9 @@ fn delete_rustup_and_cargo_home(no_modify_path: bool, process: &Process) -> Resu
         do_remove_from_path(process)?;
     }
 
+    #[cfg(windows)]
+    windows::do_remove_from_programs()?;
+
```
rami3l, line 1183: For this to be a refactoring commit, `#[cfg(windows)] windows::do_remove_from_programs()?;` should be in the `if` block.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/f9282c93f70fbcb180fa6b9c7b683f156dc5fabc..2bd69f970f3d53449be0f77b093b442b805a533f)*

tests/suite/cli_self_upd.rs
```diff
@@ -548,12 +548,119 @@ async fn update_overwrites_programs_display_version() {
...
+                 UninstallString={uninstall_string:?}"
+            ));
+        }
+        Ok(())
+    })
+    .unwrap()
+}
+
+#[cfg(windows)]
+struct InstalledProgramsRegistryGuard {
```
rami3l: This can be replaced by three `RegistryGuard`s, each initialized with one ID.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/f9282c93f70fbcb180fa6b9c7b683f156dc5fabc..2bd69f970f3d53449be0f77b093b442b805a533f)*

tests/suite/cli_self_upd.rs
```diff
@@ -351,6 +351,25 @@ async fn complete_uninstall_keeps_non_empty_cargo_bin() {
     assert!(cargodir.exists());
 }
 
+#[cfg(windows)]
+#[tokio::test]
+async fn windows_complete_uninstall_removes_empty_cargo_bin() {
```
rami3l: Suggest reorganizing "[test: cover cargo bin removal during uninstall](https://github.com/rust-lang/rustup/pull/4861/changes/252f57c236b25c58f03289e15b91f4d37631eb8c)", "[test: cover cargo bin PATH cleanup during uninstall](https://github.com/rust-lang/rustup/pull/4861/changes/ab26807a1237f944428687dcaf322365672f1bd3)", "[test: cover Windows cargo bin cleanup during uninstall](https://github.com/rust-lang/rustup/pull/4861/changes/ab0cf62ce42c5143b4c1d162409ca0829e22d09d)" to 2 commits:

In the first commit, use the commit message "test: cover cargo bin removal during uninstall" and keep 2 tests for 2 cases:

1. `complete_uninstall_removes_empty_cargo_bin`
2. `complete_uninstall_keeps_non_empty_cargo_bin`

... and those tests should be based on the existing tests of the same name introduced in "[test: cover cargo bin PATH cleanup during uninstall](https://github.com/rust-lang/rustup/pull/4861/changes/ab26807a1237f944428687dcaf322365672f1bd3)", but should not do `profile` checks, instead only the existence of `$CARGO_HOME/bin` should be checked (you should use `cfg` for platform-specific assertions).

In the second commit, use "test: cover cargo bin PATH cleanup during uninstall" as the commit message. Only do `PATH`-related checks, and put them in `cli_paths.rs`. Note that case 1 has been covered by existing tests, so you will just need to add case 2 for Windows and Unix respectively. You can get some inspirations from `unix::uninstall_removes_source_from_rcs()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/f9282c93f70fbcb180fa6b9c7b683f156dc5fabc..2bd69f970f3d53449be0f77b093b442b805a533f)*

src/cli/self_update.rs
rami3l: Concern moved into https://github.com/rust-lang/rustup/issues/4915.

tests/suite/cli_self_upd.rs
rami3l: Concern moved into https://github.com/rust-lang/rustup/issues/4915.

rami3l (approved):
At last it can be merged, nice work :))

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4861/9773857c7c0698112f4745b6b5b662a28fb7408c..561b6dfa443ba6c278568f0fe7ecb4ec8f901692)*

Issue #4862, opened 2026-05-19
1.29.1 Release Planning


rami3l:
@rust-lang/rustup is welcomed to propose outstanding issues that would be urgent enough to address before cutting the upcoming patch release.

Other issues on the current milestone will be postponed to the milestone representing the next release.

## Outstanding Issues/PRs

- #4808
- https://github.com/rust-lang/rustup/issues/4894
- https://github.com/rust-lang/rustup/pull/4935

## Release Process

TBD

## Note

N/A

rami3l:
Reopening so that it can be closed again with the actual release announcement.

rami3l:
@rust-lang/rustup It has been about 1 month since the beta release.

If anyone has remaining concerns or outstanding issues/PRs that is not yet listed here, please feel free to comment in this thread; otherwise IMHO we can start preparing for the stable release now.

ChrisDenton:
I think we're ready.

rami3l:
Closed by https://github.com/rust-lang/blog.rust-lang.org/pull/1909.

PR #4867, opened 2026-05-23
dist(release/1.29): backport patches, pt. 1

rami3l:
Part of #4862.

@djc wow, with atomic commits, backporting is as simple as dropping the incompatible commits? I'm impressed...


rami3l:
Well, there ARE some rough edges but should be pretty simple to resolve, I'll find some time to look at them later... At least there has been 0 conflicts during the rebase which is quite impressive in itself already :o

rami3l:
So far I think the backporting overhead is surprisingly low.

Next up I guess it's okay to merge this first, and then I'll add a deprecation warning for #4840 in the backport.

djc (approved):

djc:
Probably should wire up the release branch to get merge queue support, as well? I forget where/how that happens.

rami3l:
> Probably should wire up the release branch to get merge queue support, as well? I forget where/how that happens.

@djc That's in the original plan but then the team has found out that it's technically impossible (GitHub restrictions, see https://github.com/rust-lang/team/pull/2372#issuecomment-4182312803).

However auto-merge is enabled plus the contention is almost inexistent on backport targets so that should work as a second best.

djc:
I think it might still be good to have that configuration without wildcards (having to do a PR per minor release branch doesn't seem too bad).

rami3l:
@djc I would argue that the benefits of adding a merge queue is mostly that of auto merge + a bit of contention protection.

Actually, I have already added the same required checks as in trunk in addition to auto merge in https://github.com/rust-lang/team/pull/2437/changes, and given that I believe the CI quality itself should be the same.

As for the contending PR part, we shouldn't expect a lot of (>3) PRs open against the same backport target, so that part is also fine.

PR #4868, opened 2026-05-24
build(cargo): add `cargo` alias for `RUSTUP_FORCE_ARG0='rustup'`

rami3l:
Many thanks to @Cloud0310 for this idea.

Basically this crystalizes the [`RUSTUP_FORCE_ARG0='rustup'` trick](https://rust-lang.github.io/rustup/dev-guide/tips-and-tricks.html#rustup_force_arg0) into an alias subcommand `cargo run-rustup`, which I believe is a quite common action to be taken when playing with the project.

So far the value of `RUSTUP_FORCE_ARG0` is hardcoded to `'rustup'`, but I believe that would cover most use cases already (if there is a way to make it smarter while keeping this simple, please don't hesitate to help me out); if this turns out to be not enough, we can always add more as long as we are careful about choosing subcommand names.

Tested locally without issues:

```console
> cargo run-rustup -- show
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/rustup-init show`
rustup home: [redacted]

installed toolchains
--------------------
stable-aarch64-apple-darwin (active, default)
nightly-aarch64-apple-darwin
1.92.0-aarch64-apple-darwin

active toolchain
----------------
name: stable-aarch64-apple-darwin
active because: overridden by environment variable RUSTUP_TOOLCHAIN
installed targets:
  aarch64-apple-darwin
  x86_64-unknown-freebsd

### DIFF
--- a/doc/dev-guide/src/tips-and-tricks.md
+++ b/doc/dev-guide/src/tips-and-tricks.md
@@ -1,21 +1,40 @@
 # Developer tips and tricks
 
-## `RUSTUP_FORCE_ARG0`
+## `cargo run-rustup`
 
-The environment variable `RUSTUP_FORCE_ARG0` can be used to get rustup to think
-it's a particular binary, rather than e.g. copying it, symlinking it or other
-tricks with exec. This is handy when testing particular code paths from cargo
-run.
+This is the easiest way to build rustup from source and test it against your
+current rustup installation. Compared to `cargo run` which executes
+`rustup-init`, `cargo run-rustup` runs `rustup` directly.
 
-For example, if you want to run `rustup show` with `cargo run`, you may execute:
+For example, if you want to run `rustup show`, you may execute:
 
 ```console
-> cargo run --config env.RUSTUP_FORCE_ARG0=\'rustup\' -- show
+> cargo run-rustup -- show
+```
+
+## `RUSTUP_FORCE_ARG0`
+
+Sometimes, you may want to test how rustup behaves when it's invoked with an
+executable name different from `rustup-init` and `rustup`, such as `rustc` or
+`cargo` when testing the proxy mode. In this case, `RUSTUP_FORCE_ARG0` can be
+used to force rustup into thinking it's being invoked with the given name.
+Similar to [`cargo run-rustup`], it directly runs on your existing rustup
+installation.
+
+For example, if you want to run `rustc --version`, you may execute:
+
+```console
+> cargo run --config env.RUSTUP_FORCE_ARG0=\'rustc\' -- --version
 ```
 
 This command passes the `RUSTUP_FORCE_ARG0` environment variable to the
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: doc/dev-guide/src/tips-and-tricks.md [hunk 1]
```diff
@@ -17,6 +17,15 @@ This command passes the `RUSTUP_FORCE_ARG0` environment variable to the
...
 which is very important since `cargo` could also be a rustup proxy.
 
+This also works in the proxy binary mode: just replace `'rustup'` with
+`'cargo'`, `'rustc'`, etc. to test how rustup redirects the invocation to those
+binaries.
+
+In addition, since `cargo run --config env.RUSTUP_FORCE_ARG0=\'rustup\'` in
+particular is very commonly used in the development of rustup, an alias `cargo
+run-rustup` has been added to `.cargo/config.toml` to make it easier to do the
+same thing.
```
ChrisDenton: Should the discussion of `cargo run-rustup` be moved to before the discussion about `RUSTUP_FORCE_ARG0`? It seems to me that most devs won't need to know the details and will be happy enough to just use the cargo alias. They'd only need to reach for the environment variable if they want to use a different proxy (which would be less common, imho).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4868/12f54158bcef3c9845f987e6ff72139790139922..b3d48d675704c0a8632958f1b68df458923709db)*

djc (approved):
Good idea!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4868/12f54158bcef3c9845f987e6ff72139790139922..b3d48d675704c0a8632958f1b68df458923709db)*

rami3l:
~~This is broken with newer versions of `cargo` (`>= cargo 1.95.0 (f2d3ce0bd 2026-03-21)`, last worked in `cargo 1.94.1 (29ea6fb6a 2026-03-24)`):~~

```console
DEBUG cargo: exit_with_error; err=CliError { error: Some(`rustc -vV` didn't have a line for `host:`, got:
rustup 1.30.0 :: 1.28.2+661 (19abb6c0a 2026-05-25)
), exit_code: 101 }
```

~~... but I think I probably know how to fix: when `cargo` is invoking `rustc -vV` via our proxy it can add `RUSTUP_FORCE_ARG0=rustc`.~~

~~I see it as an independent issue that can be fixed separately so we can merge this PR as-is if you agree with the new wording 🙏~~

---

Update 20260527: ~~Weird, it seems that I can no longer reproduce this anymore. Anyway... 🤷‍♀️~~ Actually this is a Homebrew regression from https://github.com/Homebrew/homebrew-core/commit/cfffe64fa6fc4d56b3b2299ce07f224191098b6a. Time to drop that from my machine, I guess.

PR #4866, opened 2026-05-23
docs(dev-guide/coding-standards): adapt style guide from rustls

rami3l:
Closes #4865.

PR #4869, opened 2026-05-24
Hint that a new Rust release may be available

### LINKED ISSUE #4846: Notify users when a new stable Rust release is available
Users may be running outdated stable toolchains without realizing it, and rustup has not, until now, had the ability to signal to the user that a new release exists unless they explicitly run `rustup update` or `rustup check`.

Since stable releases follow a fixed 6-week cadence, rustup could estimate locally (without a network request) whether a new release is likely due.
When the threshold (6 weeks) is crossed, we can query the release server to confirm that there is indeed a new version, and then hint to the user to update their toolchain.

When to display this hint is open for discussion.
Should it be limited to direct rustup commands, or should it also appear when using proxies?

My inclination is to show it on every `rustup` invocation, proxy or not.
To avoid overwhelming users -- especially those who intentionally choose not to update -- it could be limited to appearing only once, or, at max, once per day.

As prior art, OCaml's package manager, `opam`, does something similar (see [here](https://github.com/ocaml/opam/blob/6770101f1d77731f5c371e0fd94a23693d2756f4/src/state/opamRepositoryState.ml#L434-L443)), notifying the user when their setup hasn't been updated for 21 days.
In contrast to `opam`, this would be exact, not just a "likely" update.

FranciscoTGouveia:
Closes #4846.

After a rustup command completes (excluding proxy commands), rustup checks whether a new Rust release is available.

To avoid introducing any additional overhead, this feature does not communicate with the release server.
Instead, it checks whether the stable toolchain manifest date is more than six weeks old.
Since Rust follows a regular and predictable release cadence, this should provide a sufficiently accurate indication that a new release is available.

When an outdated stable toolchain is detected, rustup prints a hint suggesting that the user may update their stable toolchain; this hint is shown at most once per day.
Users who prefer not to receive these hints can opt out with `rustup set release-hint false`.

I have run some benchmarks (100 iterations) for the `rustup show` command and observed no significant slowdown (1.1x) when we do not show the hint.
However, due to manifest parsing, there is a 1.4x slowdown when the hint is shown (once per day).

Feedback is appreciated on:
1. Whether this approach and its associated overhead are acceptable, or if we should take a different direction;
2. Which rustup commands should display this hint;
3. Whether relying on Rust’s regular release cadence, instead of contacting the release server, is acceptable;

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -44,7 +44,7 @@ use crate::{
     command, component_for_bin,
     config::{ActiveSource, Cfg},
     dist::{
-        AutoInstallMode, DistOptions, PartialToolchainDesc, Profile, TargetTuple,
+        DistOptions, PartialToolchainDesc, Profile, Switch, TargetTuple,
         download::DownloadCfg,
         manifest::{Component, ComponentStatus, ManifestWithHash},
     },
@@ -647,7 +647,13 @@ enum SetSubcmd {
     /// The auto toolchain install mode
     AutoInstall {
         #[arg(value_enum, default_value_t)]
-        auto_install_mode: AutoInstallMode,
+        auto_install_mode: Switch,
+    },
+
+    /// The new stable release hint mode
+    ReleaseHint {
+        #[arg(value_enum, default_value_t)]
+        release_hint_mode: Switch,
     },
 }
 
@@ -704,6 +710,12 @@ pub async fn main(
     cfg.toolchain_override = matches.plus_toolchain;
 
     let should_warn = subcmd.should_warn_empty_setup();
+
+    let should_notify = !matches.quiet
+        && !matches!(
+            subcmd,
+            RustupSubcmd::Update { .. } | RustupSubcmd::Install { .. }
+        );
 
     let exit_code = match subcmd {
         RustupSubcmd::DumpTestament => common::dump_testament(process),
@@ -828,6 +840,9 @@ pub async fn main(
             SetSubcmd::AutoInstall { auto_install_mode } => cfg
                 .set_auto_install(auto_install_mode)
                 .map(|_| ExitCode::SUCCESS),
+            SetSubcmd::ReleaseHint { release_hint_mode } => cfg
+                .set_release_hint(release_hint_mode)
+                .map(|_| ExitCode::SUCCESS),
         },
         RustupSubcmd::Completions { shell, command } => {
             output_completion_script(shell, command, process)
@@ -836,6 +851,10 @@ pub async fn main(
 
     if should_warn && cfg.list_toolchains()?.is_empty() && cfg.get_default()?.is_none() {
         warn!("no toolchain installed and no default toolchain set\n{DEFAULT_STABLE_HINT}");
+    }
+
+    if should_notify && let Err(e) = cfg.notify_release() {
+        warn!("could not check if there is a new Rust release: {e}");
     }
 
     Ok(exit_code)
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,20 +1,20 @@
 use std::fmt::{self, Debug, Display};
 use std::io;
+use std::io::Write;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use anyhow::{Context, Result, anyhow, bail};
+use chrono::{DateTime, NaiveDate};
 use serde::{Deserialize, Serialize};
 use thiserror::Error as ThisError;
 use tracing::{debug, info, trace, warn};
 
 use crate::{
     cli::{common, self_update::SelfUpdateMode},
-    dist::{
-        self, AutoInstallMode, DistOptions, PartialToolchainDesc, Profile, TargetTuple,
-        ToolchainDesc,
-    },
+    dist::{self, DistOptions, PartialToolchainDesc, Profile, Switch, TargetTuple, ToolchainDesc},
     errors::RustupError,
     fallback_settings::FallbackSettings,
     install::{InstallMethod, UpdateStatus},
@@ -417,12 +417,21 @@ impl<'a> Cfg<'a> {
         Ok(())
     }
 
-    pub(crate) fn set_auto_install(&self, mode: AutoInstallMode) -> Result<()> {
+    pub(crate) fn set_auto_install(&self, mode: Switch) -> Result<()> {
         self.settings_file.with_mut(|s| {
             s.auto_install = Some(mode);
             Ok(())
         })?;
         info!("setting auto install mode to {}", mode.as_str());
+        Ok(())
+    }
+
+    pub(crate) fn set_release_hint(&self, mode: Switch) -> Result<()> {
+        self.settings_file.with_mut(|s| {
+            s.release_hint = Some(mode);
+            Ok(())
+        })?;
+        info!("setting release hint mode to {}", mode.as_str());
         Ok(())
     }
 
@@ -435,7 +444,7 @@ impl<'a> Cfg<'a> {
             Ok(mode != "0")
         } else {
             self.settings_file
-                .with(|s| Ok(s.auto_install != Some(AutoInstallMode::Disable)))
+                .with(|s| Ok(s.auto_install != Some(Switch::Disable)))
         }
     }
 
@@ -988,6 +997,64 @@ impl<'a> Cfg<'a> {
             LocalToolchainName::Path(p) => p.to_path_buf(),
         }
     }
+
+    /// Notifies a user with a hint whenever a new Rust release is available.
+    /// This is only shown at max once per day and only if not in proxy mode.
+    pub(crate) fn notify_release(&self) -> Result<()> {
+        if self.settings_file.with(|s| Ok(s.release_hint))? == Some(Switch::Disable) {
+            return Ok(());
+        }
+
+        let time_now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs();
+
+        // Limit notifications to at most once per day.
+        // This is checked before loading the manifest to avoid unnecessary disk I/O.
+        let last_notified = self
+            .state_file
+            .with(|s| Ok(s.last_release_notified_secs.unwrap_or(0)))?;
+        if time_now.saturating_sub(last_notified) < NOTIFY_INTERVAL_SECS {
+            return Ok(());
+        }
+
+        let default_host = self.default_host_tuple()?;
+        let stable_desc = PartialToolchainDesc::from_str("stable")?.resolve(&default_host)?;
+        let distributable = DistributableToolchain::new(self, stable_desc)
+            .map_err(|e| anyhow!("stable toolchain unavailable: {e}"))?;
+        let release_date_str = match distributable.get_manifestation() {
+            Ok(manifestation) => match manifestation.load_manifest() {
+                Ok(Some(manifest)) => manifest.date,
+                Ok(None) | Err(_) => FALLBACK_RELEASE_DATE.to_owned(),
+            },
+            Err(_) => FALLBACK_RELEASE_DATE.to_owned(),
+        };
+
+        let today = DateTime::from_timestamp(time_now as i64, 0)
+            .unwrap_or_default()
+            .date_naive();
+
+        let release_date = NaiveDate::parse_from_str(&release_date_str, "%Y-%m-%d")
+            .map_err(|e| anyhow!("could not parse release date '{}': {e}", release_date_str))?;
+
+        // Skip the hint if fewer than 6 weeks have passed since the last known release.
+        if (today - release_date).num_days() < RELEASE_CYCLE_DAYS {
+            return Ok(());
+        }
+
+        self.state_file.with_mut(|s| {
+            s.last_release_notified_secs = Some(time_now);
+            Ok(())
+        })?;
+
+        writeln!(
+            self.process.stderr().lock(),
+            "hint: a new stable Rust release is available, run `rustup update stable` to install it"
+        )?;
+
+        Ok(())
+    }
 }
 
 /// The root path of the release server, without the `/dist` suffix.
@@ -1137,6 +1204,10 @@ pub(crate) enum InstalledPath<'a> {
     File { name: &'static str, path: PathBuf },
     Dir { path: &'a Path },
 }
+
+const RELEASE_CYCLE_DAYS: i64 = 42;
+const NOTIFY_INTERVAL_SECS: u64 = 24 * 60 * 60;
+const FALLBACK_RELEASE_DATE: &str = "2026-04-17";
 
 #[cfg(test)]
 mod tests {
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,7 +5,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use anyhow::{Context, Result, anyhow, bail};
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use thiserror::Error as ThisError;
 use tracing::{debug, info, trace, warn};
 
@@ -278,6 +278,7 @@ pub(crate) struct Cfg<'a> {
     pub profile_override: Option<Profile>,
     pub rustup_dir: PathBuf,
     pub settings_file: SettingsFile,
+    state_file: StateFile,
     fallback_settings: Option<FallbackSettings>,
     pub toolchains_dir: PathBuf,
     update_hash_dir: PathBuf,
@@ -323,6 +324,8 @@ impl<'a> Cfg<'a> {
             }
         })?;
 
+        let state_file = StateFile::new(rustup_dir.join("state.toml"));
+
         // Centralised file for multi-user systems to provide admin/distributor set initial values.
         #[cfg(unix)]
         let fallback_settings = FallbackSettings::new(
@@ -355,6 +358,7 @@ impl<'a> Cfg<'a> {
             profile_override: None,
             rustup_dir,
             settings_file,
+            state_file,
             fallback_settings,
             toolchains_dir,
             update_hash_dir,
@@ -1012,6 +1016,7 @@ impl Debug for Cfg<'_> {
             profile_override,
             rustup_dir,
             settings_file,
+            state_file,
             fallback_settings,
             toolchains_dir,
             update_hash_dir,
@@ -1030,6 +1035,7 @@ impl Debug for Cfg<'_> {
             .field("profile_override", profile_override)
             .field("rustup_dir", rustup_dir)
             .field("settings_file", settings_file)
+            .field("state_file", state_file)
             .field("fallback_settings", fallback_settings)
             .field("toolchains_dir", toolchains_dir)
             .field("update_hash_dir", update_hash_dir)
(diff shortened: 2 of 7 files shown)


src/cli/common.rs
```diff
@@ -569,3 +574,61 @@ pub(super) fn update_console_filter(
             .expect("error reloading `EnvFilter` for console_logger");
     }
 }
+
+/// Notifies a user with a hint whenever a new Rust release is available.
+/// This is only shown at max once per day and only if not in proxy mode.
+pub(crate) fn notify_release(cfg: &Cfg<'_>) -> Result<()> {
+    if cfg.process.var("RUSTUP_NO_RELEASE_HINT").is_ok() {
```
rami3l: Following the convention of other env vars, this should be called `RUSTUP_RELEASE_HINT` and defaults to truthy; when set to "0" we'd also want to disable this feature.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/12f54158bcef3c9845f987e6ff72139790139922..0679b120f609a2e11bd8d7f3624eabd647dbb389)*

src/cli/common.rs
```diff
@@ -569,3 +574,61 @@ pub(super) fn update_console_filter(
...
+
+    let release_date = chrono::NaiveDate::parse_from_str(&release_date_str, "%Y-%m-%d")
+        .map_err(|e| anyhow!("could not parse release date '{}': {e}", release_date_str))?;
+
+    // Skip the hint if fewer than 6 weeks have passed since the last known release.
+    if (today - release_date).num_days() < RELEASE_CYCLE_DAYS {
+        return Ok(());
+    }
+
+    cfg.settings_file.with_mut(|s| {
```
rami3l: I still think it might be harmful to write this in `settings_file`; I'd suggest if we were to write the timestamp to a file, it should be an individual one so that under XDG we can put it under `~/.local/state/rustup/` whereas the settings file should clearly be under `~/.config/rustup/`.

cc @Cloud0310 for potential interaction/collaboration regarding XDG.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/12f54158bcef3c9845f987e6ff72139790139922..0679b120f609a2e11bd8d7f3624eabd647dbb389)*

src/cli/common.rs
FranciscoTGouveia: Regarding this review comment, would it be a good idea to add a note about this in #4866?

rami3l: > Regarding this review comment, would it be a good idea to add a note about this in #4866?

Good idea! I'll do this tomorrow I think.

rami3l: Updated https://github.com/rust-lang/rustup/pull/4866 to address this. Thanks again :)

src/cli/rustup_mode.rs
```diff
@@ -675,7 +675,13 @@ pub async fn main(
     )?;
     cfg.toolchain_override = matches.plus_toolchain;
 
-    match subcmd {
+    let should_notify = !matches.quiet
+        && !matches!(
+            subcmd,
+            RustupSubcmd::Update { .. } | RustupSubcmd::Install { .. }
+        );
```
djc, line 718: I think we could just inline this into its usage site? Splitting the logic this way is confusing.

djc (commented):
LGTM modulo comments. Please squash the commits?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/12f54158bcef3c9845f987e6ff72139790139922..f9a9d91b7cad51dc126561afa289f877dbac80e8)*

src/cli/rustup_mode.rs
FranciscoTGouveia: Hmm, I thought it would be better to split this out so that the logic for deciding when to notify is explicit through a named binding, which possibly should be closer to its usage site.

Before correcting it, could you please give your thoughts on keeping it, but right before the usage site?

FranciscoTGouveia: Moreover, I just realized that `subcmd` is consumed by the following match statement (i.e., `let exit_code = match subcmd`),  so I can't use it afterward.

src/cli/common.rs
FranciscoTGouveia: Agreed.
Is there anywhere I should look regarding the XDG paths in the code, or should I leave this refactoring (i.e., `settings.toml` -> `XDG`) until after the XDG patch lands?

Happy to collaborate with @Cloud0310 on this!

src/cli/rustup_mode.rs
djc: Fair enough.

src/cli/common.rs
```diff
@@ -17,15 +19,18 @@ use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
...
     process::Process,
-    toolchain::{LocalToolchainName, Toolchain, ToolchainName},
+    toolchain::{DistributableToolchain, LocalToolchainName, Toolchain, ToolchainName},
     utils::{self, ExitCode},
 };
 
 pub(crate) const WARN_COMPLETE_PROFILE: &str = "downloading with complete profile isn't recommended unless you are a developer of the rust language";
+const RELEASE_CYCLE_DAYS: i64 = 42;
+const NOTIFY_INTERVAL_SECS: u64 = 24 * 60 * 60;
+const FALLBACK_RELEASE_DATE: &str = "2026-04-17";
```
djc: For top-down style, these should be at the bottom of the module (current exceptions notwithstanding -- want to fix them up in a separate commit?).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/0d03045abc01c086778dcc4d9b1191e9ba38fed4..bebc2bf77df25b6a36992fc87a501919c0506d2a)*

doc/user-guide/src/environment-variables.md
```diff
@@ -47,6 +47,9 @@
 - `RUSTUP_NO_BACKTRACE`. Disables backtraces on non-panic errors even when
   `RUST_BACKTRACE` is set.
 
+- `RUSTUP_RELEASE_HINT` (default: `1`). When set to `0`, suppresses the hint
```
rami3l: Nit: I think it's a bit weird to start this with the negative case: cf. the documentation for `RUSTUP_AUTO_INSTALL` right below.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/0d03045abc01c086778dcc4d9b1191e9ba38fed4..bebc2bf77df25b6a36992fc87a501919c0506d2a)*

djc: Do we have a strong reason to add this in addition to filesystem-based configuration? I am not a fan of the proliferation of environment variables that change rustup behavior.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/0d03045abc01c086778dcc4d9b1191e9ba38fed4..bebc2bf77df25b6a36992fc87a501919c0506d2a)*

src/cli/common.rs
```diff
@@ -569,3 +574,61 @@ pub(super) fn update_console_filter(
...
+        .map_err(|e| anyhow!("stable toolchain unavailable: {e}"))?;
+    let release_date_str = distributable
+        .get_manifestation()
+        .and_then(|m| m.load_manifest())
+        .ok()
+        .flatten()
+        .map(|m| m.date.clone())
+        .unwrap_or_else(|| FALLBACK_RELEASE_DATE.to_owned());
+
+    let today = chrono::DateTime::from_timestamp(time_now as i64, 0)
```
djc: Nit: suggest ditching the `chrono::` prefixes everywhere.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/0d03045abc01c086778dcc4d9b1191e9ba38fed4..bebc2bf77df25b6a36992fc87a501919c0506d2a)*

src/cli/common.rs
```diff
@@ -569,3 +574,61 @@ pub(super) fn update_console_filter(
...
+    let stable_desc = PartialToolchainDesc::from_str("stable")?.resolve(&default_host)?;
+    let distributable = DistributableToolchain::new(cfg, stable_desc)
+        .map_err(|e| anyhow!("stable toolchain unavailable: {e}"))?;
+    let release_date_str = distributable
+        .get_manifestation()
+        .and_then(|m| m.load_manifest())
+        .ok()
+        .flatten()
+        .map(|m| m.date.clone())
+        .unwrap_or_else(|| FALLBACK_RELEASE_DATE.to_owned());
```
djc: IMO this combinator-heavy style makes it pretty hard to understand what's actually going on. Suggest using more actual `match`es (and avoid `flatten()` in particular).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/0d03045abc01c086778dcc4d9b1191e9ba38fed4..bebc2bf77df25b6a36992fc87a501919c0506d2a)*

src/cli/common.rs
rami3l: @FranciscoTGouveia I don't think it should be in the `settings.toml` in the first place (it's not user configuration).

For the moment I'll say just write it as a new file under `$RUSTUP_HOME` such as `state.toml` and the XDG change should handle the rest.

doc/user-guide/src/environment-variables.md
rami3l: @djc I don't think https://github.com/rust-lang/rustup/pull/4869#discussion_r3295266501 is considered "configuration" because it just notes the last timestamp the routine is run...

src/cli/common.rs
FranciscoTGouveia: I have been working on a patch and wanted to ask whether it would be fine to add a new `src/state.rs`.
The idea is for the file to contain a `StateFile` struct, similar to how `settings.rs` is structured.
What do you think?

rami3l: @FranciscoTGouveia Sound good to me :)

FranciscoTGouveia: I've added a new commit (we can squash it later) related to the new `state.toml` file.

I drew inspiration from the `SettingsFile` in `src/settings.rs`, but refrained from adding the `cache` field.
For now, `StateFile` is only expected to be accessed at most twice per execution (i.e., once for reading and once for a potential write).
Do you think it would be worth adding a `cache` field to the `StateFile` struct?

Regarding tests, `SettingsFile` does have them, but given how simple `StateFile` is, I was not sure they would provide much value.
That said, I'm happy to add them if you think they would be beneficial.

src/cli/common.rs
FranciscoTGouveia: Done!
Just to make sure I've understood it correctly, you are suggesting a separate PR to move all (non-pub) constants to the end of their respective files?

src/state.rs
```diff
@@ -0,0 +1,63 @@
...
+impl StateFile {
+    pub(crate) fn new(path: PathBuf) -> Self {
+        Self { path }
+    }
+
+    fn load(&self) -> Result<State> {
+        if !utils::is_file(&self.path) {
+            return Ok(State::default());
+        }
+        let content = utils::read_file("state", &self.path)?;
```
rami3l: Note: Pre-existing and not directly related to this new feature, but we might need to introduce `File::lock*()` in this pattern, otherwise this file risks getting corrupted and if that happens, rustup will get crazy.

This doesn't happen yet because the update of `settings.toml` is not automatic.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/a6635323661932252e24ae7cd281d3bcbea05f4a..33eae9758beccad8a1f129085d3641a7a8a44e23)*

tests/suite/cli_self_upd.rs
```diff
@@ -650,6 +652,7 @@ async fn update_no_change() {
 "#]])
         .with_stderr(snapbox::str![[r#"
 info: checking for self-update (current version: [CURRENT_VERSION])
+hint: a new stable Rust release is available. Run `rustup update stable` to install it.
```
rami3l: Pre-existing, but the current message style is to avoid forming complete sentences:

```suggestion
hint: a new stable Rust release is available, run `rustup update stable` to install it
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/a6635323661932252e24ae7cd281d3bcbea05f4a..33eae9758beccad8a1f129085d3641a7a8a44e23)*

src/cli/common.rs
rami3l: > I've added a new commit (we can squash it later) related to the new state.toml file.

@FranciscoTGouveia Thanks! Please put this at the beginning of this PR though.

> Do you think it would be worth adding a cache field to the StateFile struct?

I don't think a cache for the state would be necessary for now; however it might be interesting to add a comment explaining this point.

> Regarding tests, SettingsFile does have them, but given how simple StateFile is, I was not sure they would provide much value.

I don't think we should assert the settings file itself, but we should assert this new feature. Also, this feature should not be visible everywhere in the test case; instead we may consider asserting only the happy/sad paths explicitly in the new tests.

src/cli/common.rs
djc: Could be a separate PR, or just a separate commit in this PR -- yes, that would be nice.

FranciscoTGouveia: Ok, I will submit a separate PR for this then.

src/cli/common.rs
FranciscoTGouveia: I've already reordered the commits, added a comment regarding the cache, and added two new tests.

src/state.rs
```diff
@@ -0,0 +1,67 @@
+use std::path::PathBuf;
```
djc: IMO it would be better to stick this in `config.rs` rather than adding another 50-line module.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..58ad1369e98ecf39ce70830c057e0e5e13afed83)*

src/cli/common.rs
```diff
@@ -569,3 +572,65 @@ pub(super) fn update_console_filter(
             .expect("error reloading `EnvFilter` for console_logger");
     }
 }
+
+/// Notifies a user with a hint whenever a new Rust release is available.
+/// This is only shown at max once per day and only if not in proxy mode.
+pub(crate) fn notify_release(cfg: &Cfg<'_>) -> Result<()> {
```
djc: Given that this only takes a `Cfg` as input, I think it should be attached to `Cfg` as a method, and we could keep it (and the entire `State` abstraction) private to the `config` module.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..58ad1369e98ecf39ce70830c057e0e5e13afed83)*

doc/user-guide/src/environment-variables.md
djc: I mean the overall disabling via the environment variable; that feels like it should be configuration rather than environment variable.

FranciscoTGouveia: Something like `rustup set release-hint false`?

djc: Yes.

src/cli/common.rs
FranciscoTGouveia: Yes, apologies for not noting this earlier. Many thanks for pointing it out!

doc/user-guide/src/environment-variables.md
FranciscoTGouveia: I have added it now.
However, I would like to point out that there was some duplication regarding the `ReleaseHintMode`, which I do not appreciate.
Possibly another PR trying to unify these would be a good idea?

src/config.rs
```diff
@@ -274,10 +274,67 @@ impl From<LocalToolchainName> for OverrideCfg {
 #[cfg(unix)]
 pub(crate) const UNIX_FALLBACK_SETTINGS: &str = "/etc/rustup/settings.toml";
 
+/// Contrary to `settings.toml`, this file is not intended to be user-facing
```
djc: Nit: top-down style, this should go below `Cfg`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/3432d809f48686b13592fffad7f66a111f08c922..a3e9eb01324bd610a19027d65d2c1113ecb29012)*

djc (approved):

doc/user-guide/src/environment-variables.md
djc: What duplication do you mean? Perhaps an inline comment would make this clearer?

src/config.rs
FranciscoTGouveia: Done!

src/state.rs
FranciscoTGouveia: Should we create a new issue to track this?

rami3l: > Should we create a new issue to track this?

@FranciscoTGouveia Certainly! :)

src/cli/rustup_mode.rs
```diff
@@ -838,6 +853,10 @@ pub async fn main(
         warn!("no toolchain installed and no default toolchain set\n{DEFAULT_STABLE_HINT}");
     }
 
+    if should_notify {
+        let _ = cfg.notify_release();
```
rami3l: Nit: Sorry I have almost missed this, but can we at least have some error logging here? A simple `if should_notify && let Err(e) = cfg.notify_release() { warn!(...) }` would do 🙏 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4869/3432d809f48686b13592fffad7f66a111f08c922..882b94accff319a6edd5c2397f6a249b1dd7139d)*

FranciscoTGouveia:
I believe all review comments are now addressed.
Should I update the initial PR message to reflect the changes done during the review process (i.e., using `rustup set release-hint` instead of an environment variable)?

rami3l:
@FranciscoTGouveia I think we can merge this first and see how it goes. Many thanks for making this patch :)

PS: I won't backport this patch to 1.29 yet to give it more confidence WRT stability; if we hit https://github.com/rust-lang/rustup/issues/4942 in the process I think it's safe to just remove the state file. The more important thing here is that @Cloud0310 has a new file to migrate for XDG.

doc/user-guide/src/environment-variables.md
rami3l: @FranciscoTGouveia Sorry, but may I ask what you meant by https://github.com/rust-lang/rustup/pull/4869#discussion_r3509092961? Looking at this thread I am afraid that we haven't fully addressed all the concerns but please feel free to dismiss me if I missed something.

FranciscoTGouveia: Whilst changing the approach from an env var, I created a new `ReleaseHintMode` enum for the `rustup set` command. Such an enum is somewhat redundant, since we already have `AutoInstallMode`, which is basically the same, hence my "duplication" comment.

Apologies for not making this clearer from the start, @rami3l.
My idea was to, after this PR, send a simplification patch to address said duplication.

djc: What's the rationale for not just fixing it in this PR?

FranciscoTGouveia: My idea was to not elongate this PR further.

I have already fixed it and merged the `ReleaseHintMode` and `AutoInstallMode`.

rami3l (approved):

PR #4875, opened 2026-05-26
Display a different color for installed components during a toolchain install

### LINKED ISSUE #4870: `rustup toolchain install` should use a different color for installed components
During toolchain installation or updates, the "installed" text shown when a component finishes installing could be displayed in bold green (following Cargo’s color conventions) to improve readability.

Feedback is welcome.

CC @blyxyas

FranciscoTGouveia:
Closes #4870.

I decided to set a prefix for the `ProgressBar` since we were not using one before.
Looking at indicatif's documentation, this seemed like a reasonable way to style a specific portion of the `ProgressBar`.

Is there a more idiomatic approach to achieving this that I might be overlooking?

CC @djc 

---
I have left two screenshots below to illustrate this change.

Before:
image

After:
image

NB: My terminal does not display the last message (regarding the installed toolchain) in bright green as it should.

### DIFF
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -430,9 +430,10 @@ impl DownloadStatus {
     }
 
     pub(crate) fn installed(&self) {
+        self.progress.set_prefix("installed");
         self.progress.set_style(DownloadStatus::progress_style(
             self.name_width,
-            "installed {total_bytes:>31}",
+            "{prefix:.green.bold} {total_bytes:>31}",
         ));
         self.progress.finish();
     }


rami3l (commented):
The `:.green.bold` and the actual test is too far away from each other so I'm not sure if I'd like this change... To me intuitively the prefix should have been the name of the component:

Image

That said I'm not an `indicatif` expert either, but if there's a more straightforward way I'd choose that.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4875/b2a5f45daf9f58979f960041976b920f56c499c0..4a4c8a4e0520f8579e9950619d08246276a6e194)*

djc:
Can you add before and after screenshots in the PR description?

FranciscoTGouveia:
I have already updated the PR description with the screenshots.
Additionally, I have moved the `.set_prefix()` into the `installed` method to reduce the distance between where it is set and where it is used.

> That said I'm not an `indicatif` expert either, but if there's a more straightforward way I'd choose that.

From what I can tell, `indicatif` can only color placeholder fields, not literal text.
The only alternative I can see is to set the the component name as the `{prefix}` and then set the `"installed"` as `{msg}`.

What are your thoughts about this?

rami3l (approved):
> From what I can tell, `indicatif` can only color placeholder fields, not literal text. The only alternative I can see is to set the the component name as the `{prefix}` and then set the `"installed"` as `{msg}`.

@FranciscoTGouveia OK with me then! Thanks :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4875/835392dadb40d7c8c0926304ff52c60c562e0bc0..3f5db6112644ec934f818b0ed929d6b2f869482e)*

djc (approved):

PR #4878, opened 2026-05-27
Remove double buffering when extracting archives

Kobzol:
While reading through the extraction code (https://rust-lang.zulipchat.com/#narrow/channel/490103-t-rustup/topic/Optimise.20rustup.20download.20speed/with/598167289), I noticed that the archive extraction is using unnecessary double buffering. The input file is already buffered, and then the compression reader introduces another buffer on top. That seems unnecessary, and might degrace performance and increase memory usage unnecessarily.

### DIFF
--- a/src/dist/component/package.rs
+++ b/src/dist/component/package.rs
@@ -3,7 +3,7 @@
 //! prefix, represented by a `Components` instance.
 
 use std::collections::{HashMap, HashSet};
-use std::io::{self, ErrorKind as IOErrorKind, Read};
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4878/b2a5f45daf9f58979f960041976b920f56c499c0..006b114fdfdf995fe28081f4a07ef59500067f0d)*

PR #4885, opened 2026-05-28
Display the full names of targets not matching the host target tuple

### LINKED ISSUE #4884: Add target names during rust-std downloads

Currently the downloads for rust-std don't show the individual target names.
```
     rust-std installed                       24.74 MiB
     rust-std installed                       27.12 MiB
     rust-std installed                       11.37 MiB
```

### Solution you'd like

I would like rust-std downloads to show the target names.

### Notes

_No response_

mrmelon54:
Displays the target tuple when it doesn't match the host target tuple.

Example output:

```
info: downloading 7 components
                             cargo installed                       10.64 MiB (29.13 MiB/s, ETA: 0s)                                         
                            clippy installed                        4.58 MiB (27.97 MiB/s, ETA: 0s)                                         
                         rust-docs installed                       21.29 MiB (8.54 MiB/s, ETA: 0s))s)                                       
rust-std-aarch64-unknown-linux-gnu installed                       27.12 MiB (26.71 MiB/s, ETA: 0s)                                         
                          rust-std installed                       28.36 MiB (26.10 MiB/s, ETA: 0s)                                         
                             rustc installed                       76.47 MiB (30.56 MiB/s, ETA: 0s)                                         
                           rustfmt installed                        2.06 MiB                                                                info: default toolchain set to stable-x86_64-unknown-linux-gnu
```

Resolves #4884

### DIFF
--- a/src/dist/manifest.rs
+++ b/src/dist/manifest.rs
@@ -87,6 +87,15 @@ impl Manifest {
             &component.pkg
         }
     }
+
+    pub(crate) fn display_name(&self, component: &Component, host_target: &TargetTuple) -> String {
+        match &component.target {
+            Some(component_target) if *host_target == *component_target => {
+                self.short_name(component).to_owned()
+            }
+            _ => self.name(component),
+        }
+    }
 }
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -167,7 +167,11 @@ impl Manifestation {
             .iter()
             // The same as the component name's length used in `ComponentBinary::new`
             // As it's used for status output
-            .map(|component| new_manifest.short_name(component).len())
+            .map(|component| {
+                new_manifest
+                    .display_name(component, &toolchain.target)
+                    .len()
+            })
             .max()
             .unwrap_or(0)
             .max(DEFAULT_MIN_NAME_WIDTH);
@@ -177,7 +181,13 @@ impl Manifestation {
             .components_to_install
             .into_iter()
             .filter_map(|component| {
-                ComponentBinary::new(component, &new_manifest, download_cfg, max_name_width)
+                ComponentBinary::new(
+                    component,
+                    &new_manifest,
+                    download_cfg,
+                    max_name_width,
+                    &toolchain.target,
+                )
             })
             .collect::<Result<Vec<_>>>()?;
 
@@ -766,6 +776,7 @@ impl<'a> ComponentBinary<'a> {
         manifest: &'a Manifest,
         download_cfg: &'a DownloadCfg<'a>,
         name_width: usize,
+        host_toolchain_target: &TargetTuple,
     ) -> Option<Result<Self>> {
         Some(Ok(ComponentBinary {
             binary: match manifest.binary(&component) {
@@ -773,7 +784,10 @@ impl<'a> ComponentBinary<'a> {
                 Ok(None) => return None,
                 Err(e) => return Some(Err(e)),
             },
-            status: download_cfg.status_for(manifest.short_name(&component).to_owned(), name_width),
+            status: download_cfg.status_for(
+                manifest.display_name(&component, host_toolchain_target),
+                name_width,
+            ),
             component,
             manifest,
             download_cfg,


rami3l (approved):
I just checked and the old version before the regression used to print both short name and target name (but one component per line), so this looks fine.

Actually a width of 100 chars might be okay in this case, given that `aarch64-unknown-linux-gnu` should be [among the long ones](https://doc.rust-lang.org/nightly/rustc/platform-support.html) already... Thanks again!

PS: I have a small nit here, please squash all changes into one before after addressing it we can merge it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4885/4361269b298cedfb30f36d0c334d6b2de27c3d50..213d837b032e53ab777d77bd1f1c2f41a414ac63)*

src/dist/manifest.rs
```diff
@@ -87,6 +87,19 @@ impl Manifest {
             &component.pkg
         }
     }
+
+    pub(crate) fn display_name(
+        &self,
+        component: &Component,
+        host_target: &Option<TargetTuple>,
```
rami3l: [Prefer](https://stackoverflow.com/a/71027236):

```suggestion
        host_target: Option<&TargetTuple>,
```

You can use `.as_ref()` to convert `&Option<T>` into `Option<&T>` at call site.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4885/4361269b298cedfb30f36d0c334d6b2de27c3d50..213d837b032e53ab777d77bd1f1c2f41a414ac63)*

rami3l:
PPS: If there are even longer component names, then we can probably reduce the size of the progress bar itself. The width of the first column has been known since #4768 by @Cloud0310 so I guess it'd be rather easy to do afterwards.

src/dist/manifest.rs
mrmelon54: I agree

mrmelon54:
I made the requested change, formatted my code, and squashed my commits.

rami3l (commented):
> I made the requested change, formatted my code, and squashed my commits.

@mrmelon54 Many thanks again!

I have a minor question though... What do you think will be more intuitive to the user, hiding the target tuple based on the current toolchain's host target tuple, or based on the current host's target tuple?

There is a rare case where e.g. Windows x64 machines can install i686 toolchains, and in this case the former will be i686, while the latter will be x64...

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4885/4361269b298cedfb30f36d0c334d6b2de27c3d50..99b2cb9d495a17b49e78fd5251adcb99c9e85c5a)*

src/dist/manifestation.rs
```diff
@@ -161,13 +161,19 @@ impl Manifestation {
             }
         }
 
+        let host_target = TargetTuple::from_host(download_cfg.process);
```
rami3l: ... if we are going this way, `TargetTuple::from_host_or_build` might be a better choice than this. Same thing for the other call site.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4885/4361269b298cedfb30f36d0c334d6b2de27c3d50..99b2cb9d495a17b49e78fd5251adcb99c9e85c5a)*

mrmelon54:
> I have a minor question though... What do you think will be more intuitive to the user, hiding the target tuple based on the current toolchain's host target tuple, or based on the current host's target tuple?
> 
> There is a rare case where e.g. Windows x64 machines can install i686 toolchains, and in this case the former will be i686, while the latter will be x64...

I thought about that too, the default toolchain might make more sense as a baseline for hiding the target tuple.

src/dist/manifestation.rs
rami3l: Actually, looking at your previous comment in https://github.com/rust-lang/rustup/issues/4884#issuecomment-4568870859 I think our goal should be to avoid (if possible) super long names showing up multiple times and producing a lot of visual noise, so it's probably better to do `let host_target = &toolchain.target;` here.

Sorry for the extra work! If you are in doubt please don't hesitate to ping me 🙏 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4885/4361269b298cedfb30f36d0c334d6b2de27c3d50..99b2cb9d495a17b49e78fd5251adcb99c9e85c5a)*

rami3l: Replaced by https://github.com/rust-lang/rustup/pull/4885/changes#r3326555438.

rami3l: The idea is to pass both `host_target: &TargetTuple` to both the `.map()` call and `ComponentBinary::new(..., host_target, max_name_width)`.

mrmelon54: Yeah I thought of that before seeing your comment.

Loading from `toolchain.target` is a good idea too.

rami3l (approved):
LGTM :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4885/4361269b298cedfb30f36d0c334d6b2de27c3d50..e0fcc58555261268e136c59c986c54d3e7cd4a7f)*

PR #4876, opened 2026-05-26
Update dev guide with cross checking guides

Cloud0310:
Closes #4864

PR #4887, opened 2026-05-31
Add `.cargo/config.windows-cross.toml` to gitignore

Cloud0310:
Add missing gitignore entry for #4876

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,8 @@
 /.settings
 /.idea/
 /.vscode/
-rust-analyzer.toml
+/rust-analyzer.toml
+/.cargo/config.windows-cross.toml
 *~
 /**/target
 /home


.gitignore
```diff
@@ -5,6 +5,7 @@
 /.idea/
 /.vscode/
 rust-analyzer.toml
+.cargo/config.windows-cross.toml
```
rami3l: Here we need a leading slash just like the other entries above it.

Same thing for rust-analyzer.toml.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4887/7f71766866193e4aed0873da3ffd79232bc039c3..cf30a9af5677f1fb88b5ae083a7bba34337dcc85)*

rami3l (approved):

PR #4895, opened 2026-06-05
feat(cli/rustup-mode): warn about auto-installation in some subcommands

### LINKED ISSUE #4894: Warn about auto-installation in some `rustup` subcommands
Follow-up to #4836: a deprecation warning should be added to the v1.29 releases regarding the upcoming breaking changes in https://github.com/rust-lang/rustup/pull/4840.

PS: This deprecation warning is important because I've seen people using e.g. `rustup show` to install the active toolchain, as there was a time when `rustup toolchain install` wasn't available (v1.27 and earlier) and that was the only way to install the active toolchain without calling the proxy.

rami3l:
Adapted from [`c037408` (#4840)](https://github.com/rust-lang/rustup/pull/4840/changes/c0374080555d64b694bbba0a49036cf6f3ee3289); closes #4894.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -142,10 +142,10 @@ impl<T> EnsureInstalled<T> {
 }
 
 impl<T: Display> EnsureInstalled<T> {
-    fn warn_auto_install(&self, process: &Process) {
+    fn warn_auto_install(&self, cfg: &Cfg<'_>) {
         // If we're already in a recursion, or we haven't just installed the active toolchain, then
         // don't print the warning.
-        let recursions = process.var("RUST_RECURSION_COUNT");
+        let recursions = cfg.process.var("RUST_RECURSION_COUNT");
         if recursions.is_ok_and(|it| it != "0") || !matches!(self.status, UpdateStatus::Installed) {
             return;
         }
@@ -562,7 +562,7 @@ impl<'a> Cfg<'a> {
         match self.ensure_active_toolchain(true, false).await {
             Ok(r) => {
                 let (tc, source) = r;
-                tc.warn_auto_install(self.process);
+                tc.warn_auto_install(self);
                 Ok(Some((tc.inner, source)))
             }
             Err(e) => match e.downcast_ref::<RustupError>() {
@@ -768,7 +768,7 @@ impl<'a> Cfg<'a> {
                 let install_if_missing = self.should_auto_install()?;
                 let EnsureInstalled { inner: tc, status } =
                     Toolchain::from_local(tc, install_if_missing, self).await?;
-                EnsureInstalled::new(tc.name(), status).warn_auto_install(self.process);
+                EnsureInstalled::new(tc.name(), status).warn_auto_install(self);
                 Ok((tc, source))
             }
             None => {
--- a/src/config.rs
+++ b/src/config.rs
@@ -154,6 +154,16 @@ impl<T: Display> EnsureInstalled<T> {
             "the missing active toolchain `{}` has been auto-installed",
             self.inner,
         );
+
+        if !cfg.allow_auto_install {
+            // NOTE: Special behavior for rustup v1.29
+            warn!("auto-installation is deprecated for most `rustup` commands");
+            warn!("scripts relying on this behavior in `rustup` may stop working in the future");
+            warn!("to install the active toolchain, use `rustup install` instead");
+            warn!("see <https://github.com/rust-lang/rustup/issues/4836> for more info");
+            return;
+        }
+
         warn!("this might cause rustup commands to take longer time to finish than expected");
         info!("you may opt out with `RUSTUP_AUTO_INSTALL=0` or `rustup set auto-install disable`");
     }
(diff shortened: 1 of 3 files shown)


rami3l:
Hmmm the blast radius is too large.

Lemme think: maybe we should just print the warning as an additional step for `warn_auto_install()`?

src/config.rs
```diff
@@ -156,6 +156,15 @@ impl<T: Display> EnsureInstalled<T> {
         );
         warn!("this might cause rustup commands to take longer time to finish than expected");
         info!("you may opt out with `RUSTUP_AUTO_INSTALL=0` or `rustup set auto-install disable`");
+
+        if cfg.allow_auto_install {
+            return;
+        }
+
+        // NOTE: Special behavior for rustup v1.29
+        warn!("auto-installation of the active toolchain is enabled when running `rustup`");
```
djc: Seems weird to have both the warning above and this warning shown at the same time?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4895/4e5b12c366f9cf575746b5a34e33220e5c4e3303..5f10e420bb936005a118ed520380d0cee1aeb981)*

src/config.rs
```diff
@@ -423,10 +432,6 @@ impl<'a> Cfg<'a> {
     }
 
     pub(crate) fn should_auto_install(&self) -> Result<bool> {
-        if !self.allow_auto_install {
```
djc, line 426: What is this trying to achieve?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4895/4e5b12c366f9cf575746b5a34e33220e5c4e3303..5f10e420bb936005a118ed520380d0cee1aeb981)*

rami3l: @djc Thanks for asking!

In [`c037408` (#4840)](https://github.com/rust-lang/rustup/pull/4840/changes/c0374080555d64b694bbba0a49036cf6f3ee3289) which is what this PR is based on, `self.allow_auto_install` is used to reject auto install from happening which is achieved by passing `false` instead of `true` for some `rustup` subcommands which will cause this function to reject auto installation right away.

However, since it is decided that in 1.29 we should only warn instead of rejecting directly, this commit replaces it with a warning.

It should be noted that the warning should have existed in this function but since this will cause too many false positives (we should only show it when an auto-installation is actually triggered, not whenever we are in an auto-install deprecated subcommand), it is moved to another module.

rami3l: Please don't hesitate to suggest any changes if you think that will make your review easier.

src/config.rs
rami3l: @djc Good catch!

How about showing the existing warning only when the deprecation notice is not shown?

rami3l: @djc I have adjusted the logic accordingly and reused the first line of the existing warning. Please tell me if you think the wording could use more adjustment.

src/config.rs
```diff
@@ -154,6 +154,16 @@ impl<T: Display> EnsureInstalled<T> {
...
             self.inner,
         );
+
+        if !cfg.allow_auto_install {
+            // NOTE: Special behavior for rustup v1.29
+            warn!(
+                "this is deprecated for most `rustup` commands and may not work in a future release"
+            );
+            warn!("see <https://github.com/rust-lang/rustup/issues/4836> for more info");
+            return;
```
djc, line 164: Okay, now I am confused about this (sorry). The way I read this, this will issue a warning for subcommands for which `allow_auto_install()` yields `false` exactly because they don't use the active toolchain anyway.

So am I misunderstanding this or is this pure noise?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4895/4e5b12c366f9cf575746b5a34e33220e5c4e3303..69a3107dd1129ea83c3e6cab2ad17dbc6becd682)*

rami3l: @djc Your understanding looks correct to me.

Looking back I'd argue that `allow_auto_install` might no longer be the best name here anymore because it now stands for "allows auto install in v1.30" or "should not emit a deprecation warning in v1.29" which is making the code harder for you to review. Is that you are trying to say?

I can offer to do a cleanup commit upfront in which I'll rename this variable, if that makes it easier for you 🙏

One downside for that is that it might make future backports more difficult if they happen to touch this part 🤔

djc: I'm suggesting that it does not make sense to warn about this because automatic installation is not actually affecting the running command anyway.

rami3l: @djc It doesn't mean anything in the context of the current command indeed, but I've seen people using this (e.g. `rustup show`) to install the active toolchain, because there was a time when `rustup toolchain install` wasn't available (v1.27 and earlier) and that was the only way to install the active toolchain. Maybe it makes more sense to you now?

rami3l: Hmmm maybe for the sake of clarity I should have added "to install the active toolchain, use `rustup toolchain install` instead" or something like that.

djc: The warning states "this is deprecated", but I don't think "this" is going to be necessarily very clear for uninformed users. Suggest expanding it.

(Otherwise I better understand the objective now, thanks.)

rami3l: @djc I've reworded the warning for a bit. Hope that looks better to you :)

djc (approved):

PR #4816, opened 2026-04-16
ci: powerpc64-unknown-linux-musl is now stable

Gelbpunkt:
As requested in https://github.com/rust-lang/rustup/pull/4688#pullrequestreview-3705544168 (cc @rami3l)

PR #3921, opened 2024-07-03
Add loongarch64-unknown-linux-musl support

heiher:
skip-stable because we're not going to release this for now

PR #4908, opened 2026-06-15
Add aarch64-unknown-freebsd

mrkajetanp:
aarch64-unknown-freebsd is now a Tier 2 target with host tools. Add support for it in rustup.

### DIFF
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -24,6 +24,7 @@ jobs: # skip-main skip-pr skip-stable
           - aarch64-linux-android
           - aarch64-unknown-linux-gnu
           - aarch64-unknown-linux-musl     # skip-pr skip-main
+          - aarch64-unknown-freebsd        # skip-pr skip-main skip-stable
           - powerpc64-unknown-linux-gnu    # skip-pr
           - x86_64-unknown-linux-musl      # skip-pr
           - i686-unknown-linux-gnu         # skip-pr skip-main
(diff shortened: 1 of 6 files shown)


mrkajetanp:
I'm not sure what `freebsd-builds-template.yaml` should look like to make it work for both aarch64 and x86, any pointers would be appreciated.

rami3l:
@mrkajetanp Hello there, and thanks for this patch!

For adding a new target you could refer to https://github.com/rust-lang/rustup/pull/3921 and add all the skips including `skip-stable` here if it's not yet stabilized... When it reaches stable we do something like https://github.com/rust-lang/rustup/pull/4816.

`freebsd-builds-template.yaml` is really for testing FreeBSD as a host which I believe is not yet possible on GitHub today for your new target. As such I suggest editing the Linux ~~target~~ template and add a new cross-compilation target over there.

Many thanks in advance 🙏

mrkajetanp:
> `freebsd-builds-template.yaml` is really for testing FreeBSD as a host which I believe is not yet possible on GitHub today for your new target. As such I suggest editing the Linux target and add a new cross-compilation target over there.

Ah right, that's the context I was missing. Thanks a lot!

mrkajetanp:
Do I understand correctly that rustup only works with stable, so `rustup-init` for a given target can only be available in CI once its builds are available in the stable channel?

rami3l:
> Do I understand correctly that rustup only works with stable, so `rustup-init` for a given target can only be available in CI once its builds are available in the stable channel?

@mrkajetanp Yes; the current CD relies on stable for checks and prebuilt artifacts.

rami3l (approved):
LGTM :)

Let's merge it first (it already provides value for those who want to build it with the nightly toolchains I imagine) and I'll see with you again when it reaches stable 🙏 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4908/0e51b0a3063fa9232ad91047961636574317a912..ceb60562befa454e06b298d15da1c78e16d273a2)*

mrkajetanp:
Thanks a lot, will do!

PR #4911, opened 2026-06-16
fix(download): avoid shared partial cache files

### LINKED ISSUE #4910: component download failed under parallel
### Verification

### Problem

component download failed under parallel

### Steps

component download failed under parallel

```sh
$ rm -rf ~/.rustup && fd Cargo.toml -x env -C {//} cargo build
info: syncing channel updates for ...
info: latest update on ... for version ...
info: downloading ... components
info: rolling back changes
error: component download failed for clippy-x86_64-unknown-linux-gnu: could not rename 'downloaded' file from '~/.rustup/downloads/fea05854f7f99eab2551578ac013c7159d9679ea657e45516bdb0c31906cf90d.partial' to '~/.rustup/downloads/fea05854f7f99eab2551578ac013c7159d9679ea657e45516bdb0c31906cf90d': No such file or directory (os error 2): No such file or directory (os error 2)
info: syncing channel updates for ...
info: latest update on ... for version ...
info: downloading ... components
info: rolling back changes
error: component download failed for clippy-x86_64-unknown-linux-gnu: could not rename 'downloaded' file from '~/.rustup/downloads/fea05854f7f99eab2551578ac013c7159d9679ea657e45516bdb0c31906cf90d.partial' to '~/.rustup/downloads/fea05854f7f99eab2551578ac013c7159d9679ea657e45516bdb0c31906cf90d': No such file or directory (os error 2): No such file or directory (os error 2)
info: syncing channel updates for ...
info: latest update on ... for version ...
info: downloading ... components
info: rolling back changes
error: component download failed for cargo-x86_64-unknown-linux-gnu: could not rename 'downloaded' file from '~/.rustup/downloads/c5421d7fad5e5c3596ba6c0748e4441a87deb237a56e1ef7c1551c0dc93dd3cd.partial' to '~/.rustup/downloads/c5421d7fad5e5c3596ba6c0748e4441a87deb237a56e1ef7c1551c0dc93dd3cd': No such file or directory (os error 2): No such file or directory (os error 2)
info: syncing channel updates for ...
info: latest update on ... for version ...
info: downloading ... components
info: rolling back changes
error: component download failed for clippy-x86_64-unknown-linux-gnu: could not rename 'downloaded' file from '~/.rustup/downloads/fea05854f7f99eab2551578ac013c7159d9679ea657e45516bdb0c31906cf90d.partial' to '~/.rustup/downloads/fea05854f7f99eab2551578ac013c7159d9679ea657e45516bdb0c31906cf90d': No such file or directory (os error 2): No such file or directory (os error 2)
```

### Possible Solution(s)

workaround component download failed by not parallel

```sh
$ rm -rf ~/.rustup && for dir in $(fd Cargo.toml -x echo {//}); do env -C "$dir" cargo build; done
info: syncing channel updates for ...
info: latest update on ... for version ...
info: downloading ... components
    Updating `...` index
  Downloaded ... ... (registry `...`)
...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
```

### Notes

_No response_

### Rustup version

```console
rustup 1.29.0
```

### Installed toolchains

```console
$ rm -rf ~/.rustup
```

### OS version

```console
nixos
```

cachebag:
After going back and forth a bit for a few hours and perusing the history of this issue, I figured it would at the least, make sense to change cached component downloads such that they use unique partial filenames instead of sharing `<hash>.partial` across processes. 

This fixes the specific download-cache race seen in #4910, where parallel auto-installs can fail while renaming a shared partial file. It is an _alternative_ to the locking approach in #4606, using unique partial paths instead of file locks.

This does not attempt to solve the broader concurrency problem in #988, since concurrent toolchain transactions can still race after downloads complete. I don't think I want to pretend to know where to start there, though I think this is a decent start and at the least covers the issue reported in #4910.

Existing legacy `<hash>.partial` files are still claimed for resume, so interrupted downloads remain recoverable.

Partially Closes #4910

### DIFF
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::fs;
-use std::io::Read;
+use std::io::{self, Read};
 use std::ops;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
@@ -46,9 +46,9 @@ impl<'a> DownloadCfg<'a> {
     }
 
     /// Downloads a file and validates its hash. Resumes interrupted downloads.
-    /// Partial downloads are stored in `self.download_dir`, keyed by hash. If the
-    /// target file already exists, then the hash is checked and it is returned
-    /// immediately without re-downloading.
+    /// Partial downloads are stored in `self.download_dir` under unique names.
+    /// If the target file already exists, then the hash is checked and it is
+    /// returned immediately without re-downloading.
     pub(crate) async fn download(
         &self,
         url: &Url,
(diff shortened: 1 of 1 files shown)


src/dist/download.rs
```diff
@@ -115,13 +104,125 @@ impl<'a> DownloadCfg<'a> {
...
+    }
+
+    fn unique_partial_path(target_file: &Path) -> PathBuf {
+        let file_name = target_file
+            .file_name()
+            .map(|s| s.to_str().unwrap_or("_"))
+            .unwrap_or("_");
+        target_file.with_file_name(format!(
+            "{file_name}.{}.partial",
+            utils::raw::random_string(16)
```
rami3l, line 171: @cachebag Thanks for looking into this!

However, I believe you are trying to solve the problem at the wrong level. If I have under stood your proposed solution correctly, even if you have avoided shared partial cache files, the unpacking will still clash because they will unpack to the same positions.

The current logic has essentially prevented rustup from picking up partial downloads (the network failure case was only a special one, but the more generally case would be that it has to recover from interruption or even kill).

The right behavior we would like to implement is that when rustup tries to download a toolchain, it will ensure that all downloading/extracting actions of essentially the same toolchain will be mutually exclusive. I actually am working on a plan for this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4911/bc46fe8f6eddba84aaa42263bf28f11334e73f58..518f7005348e2f3fb8df46c7007245fd52fd6ed6)*

cachebag: Ahh. Okay that makes sense...yeah I guess in retrospect, an ad-hoc change like this isn't very conducive..

Thanks and let me know if you need anything. Happy to help

PR #4912, opened 2026-06-16
Show setup hint when rustup has no default toolchain

### LINKED ISSUE #4901: Print special hints on `rustup` when `rustup-init` is skipped by the packager
Follow-up on https://github.com/rust-lang/rustup/issues/2915 and https://github.com/rust-lang/rustup/pull/4651.

The TLDR is that more and more packagers are skipping our installer `rustup-init` and unpacking `rustup` directly as if it were installed, with APT and Homebrew seemingly becoming the biggest two of them.

The idea here is to use simple heuristics: e.g. when the toolchain list is empty and the default toolchain is not set, we can print some basic commands to get the user started in `rustup_mode` (`proxy_mode` is already covered in https://github.com/rust-lang/rustup/issues/2915), such as a (maybe truncated) version of one of the messages in the above PRs.

The rationale behind this is that we will continue to privilege the user experience for the default flow and consider the `rustup`-direct flow advanced usage, but if minimal work can be done to unblock such users, we can also try to do something about it.

Many thanks to @dingxiangfei2009 and @Cloud0310 for the inspiration.

0xEgao:
It intends to fix #4901 

This adds in `rustup` to print the default stable setup hint when a  rustup command finishes in an empty setup state: no installed toolchains and no configured default toolchain.

The hint reuses `DEFAULT_STABLE_HINT`:

```text
```
Changes:
1. Extracted the default stable hint into `DEFAULT_STABLE_HINT`.
2. Added `RustupSubcmd::should_warn_empty_setup()` to decide which rustup commands may show the empty setup warning.
3. Added the warning after subcommand execution, so rustup checks the final state before printing the hint.
4. Skipped the warning for `DumpTestament`, `Self_`, and `Completion` as those commands are not related to toolchain state.
5. Updated CLI test expectations.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -48,7 +48,7 @@ use crate::{
         download::DownloadCfg,
         manifest::{Component, ComponentStatus, ManifestWithHash},
     },
-    errors::RustupError,
+    errors::{DEFAULT_STABLE_HINT, RustupError},
     install::{InstallMethod, UpdateStatus},
     process::{ColorableTerminal, Process},
     toolchain::{
@@ -327,6 +327,36 @@ impl RustupSubcmd {
             | RustupSubcmd::Uninstall { .. }
             | RustupSubcmd::Update { .. }
             | RustupSubcmd::Which { .. } => false,
+        }
+    }
+
+    /// Returns whether rustup should warn about having no default toolchain and no installed toolchains.
+    fn should_warn_empty_setup(&self) -> bool {
+        match self {
+            // These subcommands are not about toolchains, so the hint would be noise.
+            RustupSubcmd::Completions { .. }
+            | RustupSubcmd::DumpTestament
+            | RustupSubcmd::Self_ { .. } => false,
+
+            // For all other subcommands, the hint may be useful if rustup is still unusable after
+            // the command has completed.
+            RustupSubcmd::Check { .. }
+            | RustupSubcmd::Component { .. }
+            | RustupSubcmd::Default { .. }
+            | RustupSubcmd::Doc { .. }
+            | RustupSubcmd::Install { .. }
+            | RustupSubcmd::Override { .. }
+            | RustupSubcmd::Run { .. }
+            | RustupSubcmd::Set { .. }
+            | RustupSubcmd::Show { .. }
+            | RustupSubcmd::Target { .. }
+            | RustupSubcmd::Toolchain { .. }
+            | RustupSubcmd::Uninstall { .. }
+            | RustupSubcmd::Update { .. }
+            | RustupSubcmd::Which { .. } => true,
+
+            #[cfg(not(windows))]
+            RustupSubcmd::Man { .. } => true,
         }
     }
 }
@@ -675,7 +705,9 @@ pub async fn main(
     )?;
     cfg.toolchain_override = matches.plus_toolchain;
 
-    match subcmd {
+    let should_warn = subcmd.should_warn_empty_setup();
+
+    let exit_code = match subcmd {
         RustupSubcmd::DumpTestament => common::dump_testament(process),
         RustupSubcmd::Install { opts } => update(cfg, opts, true).await,
         RustupSubcmd::Uninstall { opts } => toolchain_remove(cfg, opts).await,
@@ -802,7 +834,13 @@ pub async fn main(
         RustupSubcmd::Completions { shell, command } => {
             output_completion_script(shell, command, process)
         }
+    }?;
+
+    if should_warn && cfg.list_toolchains()?.is_empty() && cfg.get_default()?.is_none() {
+        warn!("no toolchain installed and no default toolchain set\n{DEFAULT_STABLE_HINT}");
     }
+
+    Ok(exit_code)
 }
 
 async fn default_(
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -18,6 +18,8 @@ use crate::{
     toolchain::{PathBasedToolchainName, ToolchainName},
 };
 
+pub(crate) const DEFAULT_STABLE_HINT: &str = "help: run 'rustup default stable' to download the latest stable release of Rust and set it as your default toolchain.";
+
 /// A type erasing thunk for the retry crate to permit use with anyhow. See <https://github.com/dtolnay/anyhow/issues/149>
 #[derive(Debug, ThisError)]
 #[error(transparent)]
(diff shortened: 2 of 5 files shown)


src/cli/common.rs
```diff
@@ -362,6 +365,16 @@ pub(crate) async fn list_toolchains(cfg: &Cfg<'_>, verbose: bool, quiet: bool) -
     Ok(ExitCode::SUCCESS)
 }
 
+pub(crate) fn show_default_toolchain_hint(cfg: &Cfg<'_>) -> Result<()> {
+    let mut t = cfg.process.stdout().lock();
+    writeln!(t)?;
+    writeln!(
+        t,
+        "help: run `rustup default stable` to install Rust and set it as the default toolchain"
```
rami3l: I think we can reuse this message.

Consider rewriting the history, in the first commit extracting it to a const string `DEFAULT_STABLE_HINT` and in the second commit reuse it here:

https://github.com/rust-lang/rustup/blob/bc46fe8f6eddba84aaa42263bf28f11334e73f58/src/errors.rs#L135

src/cli/common.rs
```diff
@@ -362,6 +365,16 @@ pub(crate) async fn list_toolchains(cfg: &Cfg<'_>, verbose: bool, quiet: bool) -
     Ok(ExitCode::SUCCESS)
 }
 
+pub(crate) fn show_default_toolchain_hint(cfg: &Cfg<'_>) -> Result<()> {
```
rami3l: I don't think this function is bringing much additional value. After https://github.com/rust-lang/rustup/pull/4912/changes#r3422986833 just print that const string instead.

src/cli/common.rs
```diff
@@ -362,6 +365,16 @@ pub(crate) async fn list_toolchains(cfg: &Cfg<'_>, verbose: bool, quiet: bool) -
...
+    let mut t = cfg.process.stdout().lock();
```
rami3l: See the pre-existing hints, it'd be much better if this is not directly printed, but as a part of a `warn!()` call. Consider adding a first line explaining what is going wrong. Something like:

```rs
warn!("no toolchain installed and no default toolchain set\n{DEFAULT_STABLE_HINT}")
```

src/cli/common.rs
rami3l: IMHO you don't need to limit yourself to two subcommands to print this warning. What if we just print it in `rustup_mode` anyway after the `cfg` initialization and before the subcommand dispatch?

rami3l (commented):
Thanks a lot for the patch! 

However I don't believe we can merge this patch as-is. I have added some review comments, please don't hesitate to tell me if you have any questions.

Many thanks in advance 🙏 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4912/bc46fe8f6eddba84aaa42263bf28f11334e73f58..8adcd9b85f2ba0679a2e1297b5ce7f1725004253)*

src/cli/common.rs
0xEgao: I restricted to two subcommands only because i thought those 2 are closely related for this semantic and to avoid broader change-set, but yeah it make much sense to adopt what you stated, cool i will patch this.

0xEgao:
@rami3l Thanks for reviewing!!, 
I have patched the changes and history as you suggested.

Happy to contribute :D

src/cli/rustup_mode.rs
```diff
@@ -675,6 +675,28 @@ pub async fn main(
...
+        | RustupSubcmd::Which { .. } => true,
+        RustupSubcmd::Component { subcmd } => matches!(subcmd, ComponentSubcmd::List { .. }),
+        RustupSubcmd::Override { subcmd } => matches!(subcmd, OverrideSubcmd::List),
+        RustupSubcmd::Target { subcmd } => matches!(subcmd, TargetSubcmd::List { .. }),
+        RustupSubcmd::Toolchain { subcmd } => matches!(subcmd, ToolchainSubcmd::List { .. }),
+        #[cfg(not(windows))]
+        RustupSubcmd::Man { .. } => true,
+        RustupSubcmd::Doc { .. } => true,
+        _ => false,
+    };
```
rami3l: What is the intention of this `should_warn` variable? Are there any cases that you explicitly want to exclude?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4912/bc46fe8f6eddba84aaa42263bf28f11334e73f58..f0ad2cfcf5e8e6be115bbb0346842ba9a3cd464e)*

0xEgao: Yeah I don't think we should have this in cases like `Uninstall`, `Install`, `Update`, `Set`, `Self`,... .
As this behaviour is not really required in them.
So that's why kept a match arm.
If you think we can keep it in all cases I will do it.

rami3l: @0xEgao Thanks for the reply! My apologies but I'm still not quite sure what the rationale was when you excluded those subcommands.

I assume you mean "if the user wants to install a toolchain at the empty initial state, maybe they already know what they are doing so thus this warning will be a spam"? If that is the case, actually it makes a lot of sense, but looking back we can probably do it better.

1. Instead of warning before actually dispatching on the `subcmd`, maybe we should do it afterwards, so that if the user _ends up_ with an empty state they might need some help?
2. Also, this could be rewritten to a method call, like in the following example: https://github.com/rust-lang/rustup/blob/bc46fe8f6eddba84aaa42263bf28f11334e73f58/src/cli/rustup_mode.rs#L304-L332
3. Additionally, when doing the match, write each top-level match cases clearly without any wildcards (like in the link above) to avoid confusion when someone adds a subcommand later.
4. Finally, with that new logic we can probably think about the cases in which this warning is allowed and why. I'll say allow nearly everything and just ban `DumpTestament`, `Self_` and `Completions` because they are not about toolchains at all, what do you think?

0xEgao: Yeah i agree, warning after dispatch is better here because then we can map it to most of the subcommands without the fear of spam. 

In case of `DumpTestament`, `Self_` and `Completetions` this warning is not required as they are not about toolchain state at all. 

Cool i will add the suggested changes.

0xEgao: I have added the changes.

src/cli/rustup_mode.rs
```diff
@@ -329,6 +329,35 @@ impl RustupSubcmd {
             | RustupSubcmd::Which { .. } => false,
         }
     }
+
+    fn should_warn_about_missing_default_toolchain(&self) -> bool {
```
```suggestion

    /// Returns whether rustup should warn about having no default toolchain and no installed toolchains.
    fn should_warn_empty_setup(&self) -> bool {
```

Nit: Suggest giving this situation a name so that we can shorten this identifier name.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4912/bc46fe8f6eddba84aaa42263bf28f11334e73f58..e3c5a4db9068f91ad22b9ec7d7bf27722dfd9133)*

rami3l (approved):
@0xEgao Thanks for the patch! I've left some small nits but looks pretty good overall :)

Also please don't forget to update the PR description to reflect the latest changes.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4912/bc46fe8f6eddba84aaa42263bf28f11334e73f58..e3c5a4db9068f91ad22b9ec7d7bf27722dfd9133)*

rami3l:
> update the PR description to reflect the latest changes.


@0xEgao LGTM :) Once you've done that I can merge it.

0xEgao:
> > update the PR description to reflect the latest changes.
> 
> @0xEgao LGTM :) Once you've done that I can merge it.

Done i have update the pr description
Happy to contribute :)

PR #4913, opened 2026-06-19
Use the `prefix` placeholder instead of `msg` for components' names

FranciscoTGouveia:
Follow up on [#4875 (comment)](https://github.com/rust-lang/rustup/pull/4875#issuecomment-4699936560).

### DIFF
--- a/src/dist/download.rs
+++ b/src/dist/download.rs
@@ -304,7 +304,7 @@ impl<'a> DownloadCfg<'a> {
             )
             .progress_chars("## "),
         );
-        progress.set_message(component_name);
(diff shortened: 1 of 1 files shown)


djc:
Do you have before/after screenshots?

FranciscoTGouveia:
No, I do not.
This change does not involve any modifications to the UI.

rami3l (approved):

rami3l:
Yup I think that looks more reasonable to me to use `prefix` to hold component names and `msg` for the actual message... Even if it only has differences in naming.

djc (approved):

PR #4918, opened 2026-06-21
fix(test): test delete Windows registry accidently

### LINKED ISSUE #4915: Current tests run on Windows will accidently clears installed registry values by normal installation of rustup
### Verification

### Problem

This bug was originally discovered by @Cloud0310 , but due to operating system limitations, it could not be reproduced independently; I was able to reproduce the issue.

Currently, some tests incorrectly clear the properly installed Windows RustUp registry values while running. like:

```sh
cargo test --features test --test test_bonanza suite::cli_self_upd::uninstall_deletes_bins -- --exact --nocapture
```

This is a clear violation of the rules and will result in the properly installed version of Rust on the developers' computers being accidentally “uninstalled.”

### Steps

1. For test, backup `HKCU\Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup` in anywhere
2. Ran `cargo test --features test --test test_bonanza suite::cli_self_upd::uninstall_deletes_bins -- --exact --nocapture`
3. Values in the registry may be accidentally deleted. You can use `Get-ItemProperty -Path <mentioned key before>` to verify this.

### Possible Solution(s)

Decoupling File Deletion from Registry Cleanup

### Notes

This is because the method for removing directories is tightly coupled with the method for removing files.

https://github.com/rust-lang/rustup/blob/9773857c7c0698112f4745b6b5b662a28fb7408c/src/cli/self_update/windows.rs#L567-L571

### Rustup version

```console
```

### Installed toolchains

```console
installed toolchains
--------------------
stable-x86_64-pc-windows-gnu
1.78-x86_64-pc-windows-msvc
1.80-x86_64-pc-windows-msvc
1.86-x86_64-pc-windows-msvc
1.90.0-x86_64-pc-windows-msvc
1.93-x86_64-pc-windows-msvc

active toolchain
----------------
installed targets:
  x86_64-pc-windows-msvc
  x86_64-unknown-linux-gnu
```

### OS version

```console
Windows 25H2(26220.8690)
```

baka-gourd:
fix #4915

This PR is intended to fix the issue mentioned in the ticket regarding the accidental deletion of existing registry entries. However, while investigating the problem, we also discovered a concurrency issue in RegistryGuard, so the scope of the fix is quite extensive.

- Add per-test Windows registry isolation using a unique test UUID.
- Route Windows registry helpers through test-specific registry subkeys when a UUID is present.
- Update RegistryGuard to support multiple registry values and clean up the per-test registry subtree.
- Forward the registry UUID through the test Config and CliTestContext.
- Update Windows unit and integration tests to use UUID-scoped registry guards.
- Keep PATH registry checks and Programs entry checks isolated in tests.
- Treat missing registry keys or values as non-fatal during registry cleanup.
- No new UUID dependency is introduced; UUIDs are generated from rand.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -93,7 +93,7 @@ mod windows;
 #[cfg(windows)]
 pub use windows::complete_windows_uninstall;
 #[cfg(all(windows, feature = "test"))]
-pub use windows::{RegistryGuard, RegistryValueId, USER_PATH, get_path};
+pub use windows::{RUSTUP_REGISTRY_TEST_ID, RegistryValueId, USER_PATH, get_path};
 #[cfg(windows)]
 use windows::{do_add_to_path, do_add_to_programs, do_remove_from_path, do_remove_from_programs};
 #[cfg(windows)]
@@ -539,7 +539,7 @@ impl SelfUpdateMode {
         let setup_path = prepare_update(dl_cfg).await?;
 
         if let Some(setup_path) = &setup_path {
-            return run_update(setup_path);
+            return run_update(setup_path, dl_cfg.process);
         } else {
             // Try again in case we emitted "tool `{}` is already installed" last time.
             install_proxies(dl_cfg.process)?;
@@ -1030,7 +1030,7 @@ fn clean_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
     utils::remove_file("rustup_bin", &rustup_path)?;
 
     #[cfg(windows)]
-    do_remove_from_programs()?;
+    do_remove_from_programs(process)?;
 
     let cargo_bin_display = cargo_bin.display();
     info!("removing empty cargo bin directory `{cargo_bin_display}`");
@@ -1157,7 +1157,7 @@ pub(crate) async fn update(cfg: &Cfg<'_>) -> Result<ExitCode> {
                 PackageUpdate::Rustup,
                 Ok(UpdateStatus::Updated(version)),
             );
-            return run_update(&setup_path);
+            return run_update(&setup_path, cfg.process);
         }
         None => {
             let _ = common::show_channel_update(
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -95,7 +95,7 @@ pub use windows::complete_windows_uninstall;
 #[cfg(all(windows, feature = "test"))]
 pub use windows::{RegistryGuard, RegistryValueId, USER_PATH, get_path};
 #[cfg(windows)]
-use windows::{do_add_to_path, do_remove_from_path, do_remove_from_programs};
+use windows::{do_add_to_path, do_add_to_programs, do_remove_from_path, do_remove_from_programs};
 #[cfg(windows)]
 pub(crate) use windows::{run_update, self_replace};
 
@@ -248,6 +248,8 @@ impl InstallOpts<'_> {
 
         if !self.no_modify_path {
             do_add_to_path(process)?;
+            #[cfg(windows)]
+            do_add_to_programs(process)?;
         }
 
         // If RUSTUP_HOME is not set, make sure it exists
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -95,7 +95,10 @@ pub use windows::complete_windows_uninstall;
 #[cfg(all(windows, feature = "test"))]
 pub use windows::{RUSTUP_REGISTRY_TEST_ID, RegistryValueId, USER_PATH, get_path};
 #[cfg(windows)]
-use windows::{do_add_to_path, do_add_to_programs, do_remove_from_path, do_remove_from_programs};
+use windows::{
+    add_uninstall_registry_entry, do_add_to_path, do_remove_from_path,
+    remove_uninstall_registry_entry,
+};
 #[cfg(windows)]
 pub(crate) use windows::{run_update, self_replace};
 
@@ -251,7 +254,7 @@ impl InstallOpts<'_> {
         }
 
         #[cfg(windows)]
-        do_add_to_programs(process)?;
+        add_uninstall_registry_entry(process)?;
 
         // If RUSTUP_HOME is not set, make sure it exists
         if process.var_os("RUSTUP_HOME").is_none() {
@@ -1030,7 +1033,7 @@ fn clean_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
     utils::remove_file("rustup_bin", &rustup_path)?;
 
     #[cfg(windows)]
-    do_remove_from_programs(process)?;
+    remove_uninstall_registry_entry(process)?;
 
     let cargo_bin_display = cargo_bin.display();
     info!("removing empty cargo bin directory `{cargo_bin_display}`");
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -95,7 +95,7 @@ pub use windows::complete_windows_uninstall;
 #[cfg(all(windows, feature = "test"))]
 pub use windows::{RegistryGuard, RegistryValueId, USER_PATH, get_path};
 #[cfg(windows)]
-use windows::{do_add_to_path, do_remove_from_path};
+use windows::{do_add_to_path, do_remove_from_path, do_remove_from_programs};
 #[cfg(windows)]
 pub(crate) use windows::{run_update, self_replace};
 
@@ -1041,6 +1041,8 @@ fn clean_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
         Ok(()) if !no_modify_path => {
             info!("removing cargo bin directory `{cargo_bin_display}` from $PATH");
             do_remove_from_path(process)?;
+            #[cfg(windows)]
+            do_remove_from_programs()?;
         }
         Ok(()) => {}
     }
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -248,9 +248,10 @@ impl InstallOpts<'_> {
 
         if !self.no_modify_path {
             do_add_to_path(process)?;
-            #[cfg(windows)]
-            do_add_to_programs(process)?;
         }
+
+        #[cfg(windows)]
+        do_add_to_programs(process)?;
 
         // If RUSTUP_HOME is not set, make sure it exists
         if process.var_os("RUSTUP_HOME").is_none() {
@@ -1028,6 +1029,9 @@ fn clean_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
 
     utils::remove_file("rustup_bin", &rustup_path)?;
 
+    #[cfg(windows)]
+    do_remove_from_programs()?;
+
     let cargo_bin_display = cargo_bin.display();
     info!("removing empty cargo bin directory `{cargo_bin_display}`");
 
@@ -1043,8 +1047,6 @@ fn clean_cargo_home(no_modify_path: bool, process: &Process) -> Result<()> {
         Ok(()) if !no_modify_path => {
             info!("removing cargo bin directory `{cargo_bin_display}` from $PATH");
             do_remove_from_path(process)?;
-            #[cfg(windows)]
-            do_remove_from_programs()?;
         }
         Ok(()) => {}
     }
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -567,8 +567,7 @@ where
 
 pub(crate) fn do_remove_from_path(process: &Process) -> Result<()> {
     let new_path = _with_path_cargo_home_bin(_remove_from_path, process)?;
-    _apply_new_path(new_path)?;
-    do_remove_from_programs()
+    _apply_new_path(new_path)
 }
 
 const RUSTUP_UNINSTALL_ENTRY: &str = r"Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup";
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::env::{consts::EXE_SUFFIX, split_paths};
 use std::ffi::{OsStr, OsString};
 use std::fmt;
@@ -5,8 +6,6 @@ use std::io::Write;
 use std::os::windows::ffi::OsStrExt;
 use std::path::Path;
 use std::process::Command;
-#[cfg(any(test, feature = "test"))]
-use std::sync::{LockResult, Mutex, MutexGuard};
 
 use anyhow::{Context, Result, anyhow};
 use tracing::{info, warn};
@@ -450,10 +449,10 @@ pub(crate) fn wait_for_parent() -> Result<()> {
 
 pub(crate) fn do_add_to_path(process: &Process) -> Result<()> {
     let new_path = _with_path_cargo_home_bin(_add_to_path, process)?;
-    _apply_new_path(new_path)
+    _apply_new_path(new_path, process)
 }
 
-fn _apply_new_path(new_path: Option<HSTRING>) -> Result<()> {
+fn _apply_new_path(new_path: Option<HSTRING>, process: &Process) -> Result<()> {
     use std::ptr;
     use windows_sys::Win32::Foundation::{LPARAM, WPARAM};
     use windows_sys::Win32::UI::WindowsAndMessaging::{
@@ -464,7 +463,7 @@ fn _apply_new_path(new_path: Option<HSTRING>) -> Result<()> {
         return Ok(()); // No need to set the path
     };
 
-    let environment = CURRENT_USER.create("Environment")?;
+    let environment = process.registry_environment_key()?;
 
     if new_path.is_empty() {
         environment.remove_value("PATH")?;
@@ -492,9 +491,9 @@ fn _apply_new_path(new_path: Option<HSTRING>) -> Result<()> {
 // Get the windows PATH variable out of the registry as a String. If
 // this returns None then the PATH variable is not a string and we
 // should not mess with it.
-fn get_windows_path_var() -> Result<Option<HSTRING>> {
-    let environment = CURRENT_USER
-        .create("Environment")
+fn get_windows_path_var(process: &Process) -> Result<Option<HSTRING>> {
+    let environment = process
+        .registry_environment_key()
         .context("Failed opening Environment key")?;
 
     let reg_value = environment.get_hstring("PATH");
@@ -558,7 +557,7 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
 where
     F: FnOnce(HSTRING, HSTRING) -> Option<HSTRING>,
 {
-    let windows_path = get_windows_path_var()?;
+    let windows_path = get_windows_path_var(process)?;
     let mut path_str = process.cargo_home()?;
     path_str.push("bin");
     Ok(windows_path.and_then(|old_path| f(old_path, HSTRING::from(path_str.as_path()))))
@@ -566,19 +565,46 @@ where
 
 pub(crate) fn do_remove_from_path(process: &Process) -> Result<()> {
     let new_path = _with_path_cargo_home_bin(_remove_from_path, process)?;
-    _apply_new_path(new_path)
+    _apply_new_path(new_path, process)
 }
 
 const RUSTUP_UNINSTALL_ENTRY: &str = r"Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup";
 
-fn rustup_uninstall_reg_key() -> Result<Key> {
-    CURRENT_USER
-        .create(RUSTUP_UNINSTALL_ENTRY)
+impl Process {
+    fn registry_sub_key_path<'a>(&self, sub_key: &'a str) -> Cow<'a, str> {
+        #[cfg(any(test, feature = "test"))]
+        if let Ok(test_id) = self.var(RUSTUP_REGISTRY_TEST_ID) {
+            return Cow::Owned(format!(r"RustupTest-{test_id}\{sub_key}"));
+        }
+
+        Cow::Borrowed(sub_key)
+    }
+
+    fn registry_key(&self, sub_key: &str, parent: &Key) -> windows_registry::Result<Key> {
+        let mut options = parent.options();
+        options.read().write().create();
+
+        #[cfg(any(test, feature = "test"))]
+        if self.var(RUSTUP_REGISTRY_TEST_ID).is_ok() {
+            options.volatile();
+        }
+
+        options.open(self.registry_sub_key_path(sub_key))
+    }
+
+    fn registry_environment_key(&self) -> windows_registry::Result<Key> {
+        self.registry_key("Environment", CURRENT_USER)
+    }
+}
+
+fn rustup_uninstall_reg_key(process: &Process) -> Result<Key> {
+    process
+        .registry_key(RUSTUP_UNINSTALL_ENTRY, CURRENT_USER)
         .context("Failed creating uninstall key")
 }
 
-pub(crate) fn do_update_programs_display_version(version: &str) -> Result<()> {
-    rustup_uninstall_reg_key()?
+pub(crate) fn do_update_programs_display_version(version: &str, process: &Process) -> Result<()> {
+    rustup_uninstall_reg_key(process)?
         .set_string("DisplayVersion", version)
         .context("Failed to set `DisplayVersion`")
 }
@@ -586,7 +612,7 @@ pub(crate) fn do_update_programs_display_version(version: &str) -> Result<()> {
 pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
     use std::path::PathBuf;
 
-    let key = rustup_uninstall_reg_key()?;
+    let key = rustup_uninstall_reg_key(process)?;
 
     // Don't overwrite registry if Rustup is already installed
     let prev = key.get_hstring("UninstallString");
@@ -608,20 +634,20 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
         .context("Failed to set `UninstallString`")?;
     key.set_string("DisplayName", "Rustup: the Rust toolchain installer")
         .context("Failed to set `DisplayName`")?;
-    do_update_programs_display_version(env!("CARGO_PKG_VERSION"))?;
+    do_update_programs_display_version(env!("CARGO_PKG_VERSION"), process)?;
 
     Ok(())
 }
 
-pub(crate) fn do_remove_from_programs() -> Result<()> {
-    match CURRENT_USER.remove_tree(RUSTUP_UNINSTALL_ENTRY) {
+pub(crate) fn do_remove_from_programs(process: &Process) -> Result<()> {
+    match CURRENT_USER.remove_tree(process.registry_sub_key_path(RUSTUP_UNINSTALL_ENTRY)) {
         Ok(()) => Ok(()),
         Err(e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => Ok(()),
         Err(e) => Err(anyhow!(e)),
     }
 }
 
-pub(crate) fn run_update(setup_path: &Path) -> Result<utils::ExitCode> {
+pub(crate) fn run_update(setup_path: &Path, process: &Process) -> Result<utils::ExitCode> {
     Command::new(setup_path)
         .arg("--self-replace")
         .spawn()
@@ -631,7 +657,7 @@ pub(crate) fn run_update(setup_path: &Path) -> Result<utils::ExitCode> {
         warn!("failed to get the new rustup version in order to update `DisplayVersion`");
         return Ok(utils::ExitCode(1));
     };
-    do_update_programs_display_version(&version)?;
+    do_update_programs_display_version(&version, process)?;
 
     Ok(utils::ExitCode(0))
 }
@@ -753,38 +779,14 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
 // so we use env var here, notifying it if we need to remove $CARGO_HOME/bin from $PATH
 const GC_MODIFY_PATH: &str = "RUSTUP_GC_MODIFY_PATH";
 
+/// Environment variable carrying the per-test registry ID.
 #[cfg(any(test, feature = "test"))]
-pub fn get_path() -> Result<Option<Value>> {
-    USER_PATH.get()
-}
+pub const RUSTUP_REGISTRY_TEST_ID: &str = "RUSTUP_REGISTRY_TEST_ID";
 
 #[cfg(any(test, feature = "test"))]
-pub struct RegistryGuard<'a> {
-    _locked: LockResult<MutexGuard<'a, ()>>,
-    id: &'static RegistryValueId,
-    prev: Option<Value>,
+pub fn get_path(test_id: &str) -> Result<Option<Value>> {
+    USER_PATH.get(test_id, CURRENT_USER)
 }
-
-#[cfg(any(test, feature = "test"))]
-impl RegistryGuard<'_> {
-    pub fn new(id: &'static RegistryValueId) -> Result<Self> {
-        Ok(Self {
-            _locked: REGISTRY_LOCK.lock(),
-            id,
-            prev: id.get()?,
-        })
-    }
-}
-
-#[cfg(any(test, feature = "test"))]
-impl Drop for RegistryGuard<'_> {
-    fn drop(&mut self) {
-        self.id.set(self.prev.as_ref()).unwrap();
-    }
-}
-
-#[cfg(any(test, feature = "test"))]
-static REGISTRY_LOCK: Mutex<()> = Mutex::new(());
 
 #[cfg(any(test, feature = "test"))]
 pub const USER_PATH: RegistryValueId = RegistryValueId {
@@ -800,8 +802,10 @@ pub struct RegistryValueId {
 
 #[cfg(any(test, feature = "test"))]
 impl RegistryValueId {
-    pub fn get(&self) -> Result<Option<Value>> {
-        let sub_key = CURRENT_USER.create(self.sub_key)?;
+    pub fn get(&self, test_id: &str, parent: &Key) -> Result<Option<Value>> {
+        let mut options = parent.options();
+        options.read().write().create().volatile();
+        let sub_key = options.open(format!(r"RustupTest-{test_id}\{}", self.sub_key))?;
         match sub_key.get_value(self.value_name) {
             Ok(val) => Ok(Some(val)),
             Err(e) if e.code() == HRESULT::from_win32(ERROR_FILE_NOT_FOUND) => Ok(None),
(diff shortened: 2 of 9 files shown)


src/cli/self_update/windows.rs
```diff
@@ -755,38 +751,89 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
+        }
+        Ok(Self { uuid, saved })
     }
 }
 
 #[cfg(any(test, feature = "test"))]
-static REGISTRY_LOCK: Mutex<()> = Mutex::new(());
+impl Drop for RegistryGuard {
+    fn drop(&mut self) {
+        for (id, prev) in self.saved.iter().rev() {
```
Cloud0310: I guess since we're trying to remove the whole test subtree all together, no need for settting them empty again here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/957b0c392a7e67817104ade2baeb03d4f7e874a1..9be439921a25858588eb9c4efad06487610da051)*

Cloud0310:
Another thing here is, can we make code format coming along with commits instead of doing formatting in the end commit here.

src/cli/self_update/windows.rs
```diff
@@ -754,39 +750,79 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
 
 #[cfg(any(test, feature = "test"))]
-impl Drop for RegistryGuard<'_> {
+impl Drop for RegistryGuard {
     fn drop(&mut self) {
-        self.id.set(self.prev.as_ref()).unwrap();
+        let root = format!(
+            r"Software\Microsoft\Windows\CurrentVersion\Uninstall\RustupTest-{}",
+            self.uuid
+        );
```
Cloud0310: Nit: use new style inline `format!`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/3432d809f48686b13592fffad7f66a111f08c922..360e048cb5d101066ac95a78bb53cc8b0099b4ce)*

src/cli/self_update.rs
rami3l, line 1052: Moving `do_remove_from_programs()` out of this `match{}` is a separate change and has nothing to do with this refactoring commit.

Consider spitting this into a refactoring commit (that puts `do_remove_from_programs()` directly below this line) first, and a breaking change commit (moving the above call out of the `match{}` block) after that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/cli/self_update.rs
```diff
@@ -239,6 +239,9 @@ impl InstallOpts<'_> {
             do_add_to_path(cfg.process)?;
         }
 
+        #[cfg(windows)]
+        do_add_to_programs(cfg.process)?;
```
rami3l: Same problem as https://github.com/rust-lang/rustup/pull/4918/changes/7ae30ea30c69d9ad18907dac2d3c76e2659337f2#r3720674968.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/cli/self_update/windows.rs
```diff
@@ -752,38 +750,93 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
+/// the identity function.
 #[cfg(any(test, feature = "test"))]
-impl RegistryGuard<'_> {
-    pub fn new(id: &'static RegistryValueId) -> Result<Self> {
-        Ok(Self {
-            _locked: REGISTRY_LOCK.lock(),
-            id,
-            prev: id.get()?,
-        })
+fn registry_sub_key_path(sub_key: &str, process: &Process) -> String {
```
rami3l: Nit: This should be converted into a Windows-only method of the `Process` type.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/cli/self_update/windows.rs
```diff
@@ -752,38 +750,93 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
+    match process.var_opt(RUSTUP_TEST_REGISTRY_UUID) {
+        Ok(Some(uuid)) if !uuid.is_empty() => map_sub_key(sub_key, &uuid),
+        _ => sub_key.to_owned(),
     }
 }
 
+#[cfg(not(any(test, feature = "test")))]
+fn registry_sub_key_path(sub_key: &str, _process: &Process) -> String {
+    sub_key.to_owned()
```
rami3l: There doesn't seem to be a need for this function.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/cli/self_update/windows.rs
```diff
@@ -799,8 +852,15 @@ pub struct RegistryValueId {
 
 #[cfg(any(test, feature = "test"))]
 impl RegistryValueId {
-    pub fn get(&self) -> Result<Option<Value>> {
-        let sub_key = CURRENT_USER.create(self.sub_key)?;
+    fn resolved_sub_key(&self, uuid: Option<&str>) -> windows_registry::Result<Key> {
```
rami3l: Suggest inlining this function.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

rami3l: It seems that in every test case involving registry access, a UUID is needed anyways, if that is indeed true, I don't see a reason to pass `Option<>` here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/cli/self_update/windows.rs
```diff
@@ -752,38 +750,93 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
+fn random_uuid() -> String {
+    format!("{:032x}", rand::random::<u128>())
+}
+
+/// Guards a set of registry values for the duration of a test.
+#[cfg(any(test, feature = "test"))]
+pub struct RegistryGuard {
+    uuid: String,
+    saved: Vec<(&'static RegistryValueId, Option<Value>)>,
+}
```
rami3l: If the analysis in https://github.com/rust-lang/rustup/pull/4918/changes/d303f39310372f729547229e88c2cf3e463ca7be#r3721131249 is correct, then we don't need a registry guard at all.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/cli/self_update/windows.rs
rami3l: When we are running tests, new registry keys and all its children should be created as volatile, so that we won't leave unintended garbage entries in the developer's machine (thanks, @Cloud0310).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/test/clitools.rs
```diff
@@ -64,6 +64,9 @@ pub struct Config {
     pub workdir: RefCell<PathBuf>,
     /// This is the test root for keeping stuff together
     test_root_dir: PathBuf,
+    /// Per-test Windows registry UUID.
+    #[cfg(windows)]
+    pub test_registry_uuid: Option<String>,
```
rami3l: It feels like this (01182b78ba788718505d7eb4266e0595f2ef3948) should be squashed into one of the previous commits; your goal should be that each commit would still be able to pass all the tests.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

rami3l (commented):
@baka-gourd I think this looks quite interesting but would require more work before it can be merged.

Three general principles of our committing style:
1. Each commit should do one single thing, whether that is refactoring, introducing a breaking change, or adding a new feature.
2. Each commit should ensure that all test cases are passing. If running all tests is not practical in your local environment, you should have run at least all related tests and have made sure that they are passing.
3. When a commit indicates a behavioral change that might cause some tests to fail, you should update those influenced tests in the same commit so that they pass again.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..6366e7f8f207fabf785431a364fbd0c1d9a0bfb1)*

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,60 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
+    _apply_new_path(new_path, process)
 }
 
 const RUSTUP_UNINSTALL_ENTRY: &str = r"Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup";
 
-fn rustup_uninstall_reg_key() -> Result<Key> {
-    CURRENT_USER
-        .create(RUSTUP_UNINSTALL_ENTRY)
+impl Process {
+    fn registry_key(&self, sub_key: &str) -> windows_registry::Result<Key> {
```
rami3l: I cannot see the point of `registry_key()` being bundled with `CURRENT_USER` here. Also, the use of this function across your changes are not consistent either.

The idea of introducing such a method looks good, but I would expect the syntax to look more like `process.registry_key(sub_key, CURRENT_USER)`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..78a2fde6716c93ff6b619965cedaf1d61d9b1aeb)*

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,60 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
+        if self.var_os(RUSTUP_TEST_REGISTRY_UUID).is_some() {
+            options.volatile();
+        }
+
+        options.open(self.registry_sub_key_path(sub_key))
+    }
+
+    fn registry_sub_key_path(&self, sub_key: &str) -> String {
+        #[cfg(any(test, feature = "test"))]
+        if let Ok(Some(uuid)) = self.var_opt(RUSTUP_TEST_REGISTRY_UUID)
```
rami3l: You can use `var_os()` directly here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..78a2fde6716c93ff6b619965cedaf1d61d9b1aeb)*

src/cli/self_update/windows.rs
```diff
@@ -752,38 +775,32 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
-    fn drop(&mut self) {
-        self.id.set(self.prev.as_ref()).unwrap();
-    }
+pub fn get_path(uuid: &str) -> Result<Option<Value>> {
+    USER_PATH.get(uuid)
 }
 
 #[cfg(any(test, feature = "test"))]
-static REGISTRY_LOCK: Mutex<()> = Mutex::new(());
+pub(crate) fn random_uuid() -> String {
```
rami3l: This really shouldn't be called UUID, maybe `test_id`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/23b2cb0af6418325cbeb88283289946b379e2c5e..78a2fde6716c93ff6b619965cedaf1d61d9b1aeb)*

rami3l (changes requested):

src/cli/self_update/windows.rs
```diff
@@ -752,38 +778,32 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
-        self.id.set(self.prev.as_ref()).unwrap();
-    }
+pub fn get_path(test_id: &str) -> Result<Option<Value>> {
+    USER_PATH.get(test_id, CURRENT_USER)
 }
 
 #[cfg(any(test, feature = "test"))]
-static REGISTRY_LOCK: Mutex<()> = Mutex::new(());
+pub(crate) fn test_id() -> String {
+    format!("{:032x}", rand::random::<u128>())
```
rami3l: We could reuse the existing utility `utils::raw::random_string()` here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..25f755314fb70ff210c4a8c997f979289eaa00e0)*

tests/suite/cli_paths.rs
```diff
@@ -565,10 +573,10 @@ mod windows {
     #[tokio::test]
     async fn uninstall_doesnt_affect_path_with_no_modify_path() {
         let cx = CliTestContext::new(Scenario::Empty).await;
-        let _guard = RegistryGuard::new(&USER_PATH).unwrap();
+        let test_id = cx.config.test_registry_uuid.clone();
```
rami3l: The renaming is not complete: this should be named `cx.config.test_id`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..25f755314fb70ff210c4a8c997f979289eaa00e0)*

tests/suite/cli_paths.rs
```diff
@@ -591,21 +599,19 @@ mod windows {
     async fn install_uninstall_affect_path_with_non_unicode() {
         use std::os::windows::ffi::OsStrExt;
 
-        use windows_registry::{CURRENT_USER, Type};
+        use windows_registry::Type;
 
         let cx = CliTestContext::new(Scenario::Empty).await;
-        let _guard = RegistryGuard::new(&USER_PATH).unwrap();
+        let test_id = cx.config.test_registry_uuid.clone();
```
rami3l: Nit: Taking a reference of the ID is possibly enough, there doesn't seem to be a reason to clone here. Same issue for many other places in your code.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..25f755314fb70ff210c4a8c997f979289eaa00e0)*

src/test/clitools.rs
```diff
@@ -324,6 +327,12 @@ impl Config {
         if let Some(root) = self.rustup_update_root.as_ref() {
             cmd.env("RUSTUP_UPDATE_ROOT", root);
         }
+
+        #[cfg(windows)]
+        cmd.env(
+            crate::cli::self_update::RUSTUP_TEST_REGISTRY_UUID,
+            &self.test_registry_uuid,
```
rami3l: Also rename here...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..25f755314fb70ff210c4a8c997f979289eaa00e0)*

src/test/clitools.rs
```diff
@@ -64,6 +64,9 @@ pub struct Config {
     pub workdir: RefCell<PathBuf>,
     /// This is the test root for keeping stuff together
     test_root_dir: PathBuf,
+    /// Per-test Windows registry UUID.
+    #[cfg(windows)]
+    pub test_registry_uuid: String,
```
rami3l: Also rename here...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..25f755314fb70ff210c4a8c997f979289eaa00e0)*

src/cli/self_update/windows.rs
```diff
@@ -752,38 +778,32 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
-    fn drop(&mut self) {
-        self.id.set(self.prev.as_ref()).unwrap();
-    }
+pub fn get_path(test_id: &str) -> Result<Option<Value>> {
+    USER_PATH.get(test_id, CURRENT_USER)
 }
 
 #[cfg(any(test, feature = "test"))]
-static REGISTRY_LOCK: Mutex<()> = Mutex::new(());
+pub(crate) fn test_id() -> String {
```
rami3l: I think it's probably better to put this into `src/test.rs`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..25f755314fb70ff210c4a8c997f979289eaa00e0)*

rami3l (changes requested):

src/cli/self_update/windows.rs
```diff
@@ -754,38 +778,32 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
-            prev: id.get()?,
-        })
+fn map_sub_key(sub_key: &str, test_id: &str) -> String {
+    let root = format!(r"Software\Microsoft\Windows\CurrentVersion\Uninstall\RustupTest-{test_id}");
+    if sub_key == RUSTUP_UNINSTALL_ENTRY {
+        format!(r"{root}\Programs")
+    } else if sub_key == "Environment" {
+        format!(r"{root}\Environment")
+    } else {
+        format!(r"{root}\{sub_key}")
```
rami3l: Why is this function implemented like this? Can (and should) it be a simple concatenation of `root` and `subkey`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..25f755314fb70ff210c4a8c997f979289eaa00e0)*

src/cli/self_update/windows.rs
```diff
@@ -799,32 +812,124 @@ pub struct RegistryValueId {
...
+        let vars: HashMap<String, String> = [
+            ("HOME".to_string(), "/unused".to_string()),
+            (RUSTUP_TEST_REGISTRY_ID.to_string(), test_id.to_owned()),
+        ]
+        .into_iter()
+        .collect();
+        TestProcess::with_vars(vars)
+    }
+
+    fn test_environment_key(process: &Process) -> Key {
```
Cloud0310: @rami3l Do you think this should be a method of `TestProcess`? Guess would be better.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/07e34ea5823a93960334944616b7b9760d39cfba..91589c27aac074a6c1778cd4eb5d395a17029535)*

src/cli/self_update/windows.rs
```diff
@@ -799,32 +812,124 @@ pub struct RegistryValueId {
...
+    use std::collections::HashMap;
     use std::os::windows::ffi::OsStringExt;
 
     use windows_registry::Type;
 
     use super::*;
     use crate::process::TestProcess;
+    use crate::test::test_id;
+
+    fn test_process(test_id: &str) -> TestProcess {
```
Cloud0310, line 845: Integrate this into `TestProcess::new`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/07e34ea5823a93960334944616b7b9760d39cfba..91589c27aac074a6c1778cd4eb5d395a17029535)*

Cloud0310: Might test id needed for all the tests in the future?
If so, I guess we can move this and `test_id` into `TestProcess::new`.

@rami3l how do you think of this?

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,63 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
+
+        options.open(self.registry_sub_key_path(sub_key))
+    }
+
+    fn remove_registry_tree(&self, sub_key: &str, parent: &Key) -> windows_registry::Result<()> {
+        parent.remove_tree(self.registry_sub_key_path(sub_key))
+    }
+}
+
+fn rustup_uninstall_reg_key(process: &Process) -> Result<Key> {
```
Cloud0310: Since we're already here, hope you don't bother create another commit renaming old legacy function namings.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/07e34ea5823a93960334944616b7b9760d39cfba..91589c27aac074a6c1778cd4eb5d395a17029535)*

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,63 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
+    process
+        .registry_key(RUSTUP_UNINSTALL_ENTRY, CURRENT_USER)
         .context("Failed creating uninstall key")
 }
 
-pub(crate) fn do_update_programs_display_version(version: &str) -> Result<()> {
-    rustup_uninstall_reg_key()?
+pub(crate) fn do_update_programs_display_version(version: &str, process: &Process) -> Result<()> {
```
Cloud0310: Rename: https://github.com/rust-lang/rustup/pull/4918/changes/91589c27aac074a6c1778cd4eb5d395a17029535#r3730701127

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/07e34ea5823a93960334944616b7b9760d39cfba..91589c27aac074a6c1778cd4eb5d395a17029535)*

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,63 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
+    rustup_uninstall_reg_key(process)?
         .set_string("DisplayVersion", version)
         .context("Failed to set `DisplayVersion`")
 }
 
 pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
```
Cloud0310: Rename: https://github.com/rust-lang/rustup/pull/4918/changes/91589c27aac074a6c1778cd4eb5d395a17029535#r3730701127

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/07e34ea5823a93960334944616b7b9760d39cfba..91589c27aac074a6c1778cd4eb5d395a17029535)*

src/cli/self_update/windows.rs
```diff
@@ -607,20 +633,20 @@ pub(crate) fn do_add_to_programs(process: &Process) -> Result<()> {
         .context("Failed to set `UninstallString`")?;
     key.set_string("DisplayName", "Rustup: the Rust toolchain installer")
         .context("Failed to set `DisplayName`")?;
-    do_update_programs_display_version(env!("CARGO_PKG_VERSION"))?;
+    do_update_programs_display_version(env!("CARGO_PKG_VERSION"), process)?;
 
     Ok(())
 }
 
-pub(crate) fn do_remove_from_programs() -> Result<()> {
```
Cloud0310, line 618: Rename: https://github.com/rust-lang/rustup/pull/4918/changes/91589c27aac074a6c1778cd4eb5d395a17029535#r3730701127

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/07e34ea5823a93960334944616b7b9760d39cfba..91589c27aac074a6c1778cd4eb5d395a17029535)*

baka-gourd:
The CI error isn't my fault; CI pulled the wrong commit😓

Cloud0310:
Lmao, a re-run seems don't fix up the problem with buggy github actions. :laughing:

rami3l:
@baka-gourd I think a rebase would help though.

src/cli/self_update/windows.rs
rami3l: There is no need to create this directly in TestProcess since most tests won't use this variable.

src/cli/self_update/windows.rs
rami3l: Suggest moving this to `Process::registry_environment_key()` without the `.unwrap()`.

src/cli/self_update/windows.rs
```diff
@@ -491,9 +489,9 @@ fn _apply_new_path(new_path: Option<HSTRING>) -> Result<()> {
 // Get the windows PATH variable out of the registry as a String. If
 // this returns None then the PATH variable is not a string and we
 // should not mess with it.
-fn get_windows_path_var() -> Result<Option<HSTRING>> {
-    let environment = CURRENT_USER
-        .create("Environment")
+fn get_windows_path_var(process: &Process) -> Result<Option<HSTRING>> {
+    let environment = process
+        .registry_key("Environment", CURRENT_USER)
```
rami3l: ... and now you can use the extracted `Process::registry_environment_key()` here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,63 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
 const RUSTUP_UNINSTALL_ENTRY: &str = r"Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup";
 
-fn rustup_uninstall_reg_key() -> Result<Key> {
-    CURRENT_USER
-        .create(RUSTUP_UNINSTALL_ENTRY)
+impl Process {
+    fn registry_sub_key_path(&self, sub_key: &str) -> String {
+        #[cfg(any(test, feature = "test"))]
+        if let Some(test_id) = self.var_os(RUSTUP_TEST_REGISTRY_ID) {
+            let test_id = test_id.to_string_lossy();
```
rami3l: Since you are already using `.to_string_lossy()` here, suggest replacing `self.var_os()` with `self.var()`, that's easier.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,63 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
+            return map_sub_key(sub_key, &test_id);
+        }
+
+        sub_key.to_owned()
```
rami3l: Use `Cow` to avoid cloning here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

src/cli/self_update/windows.rs
```diff
@@ -557,35 +555,63 @@ fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> Result<Option<HSTRIN
...
+        if self.var_os(RUSTUP_TEST_REGISTRY_ID).is_some() {
+            options.volatile();
+        }
+
+        options.open(self.registry_sub_key_path(sub_key))
+    }
+
+    fn remove_registry_tree(&self, sub_key: &str, parent: &Key) -> windows_registry::Result<()> {
+        parent.remove_tree(self.registry_sub_key_path(sub_key))
+    }
```
rami3l, line 597: This is a single-use function and thus should be inlined.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

src/cli/self_update/windows.rs
rami3l: Otherwise, since you are operating on a `sub_key` based on the original ROOT (not this `root`). This will cause the following path which is clearly unnecessary:

```
ROOT\Software\Microsoft\Windows\CurrentVersion\Uninstall\RustupTest-TEST_ID\SUBKEY
```

This function should just return `format!("RustupTest-{test_id}\{sub_key}")`, so that we can end up operating on the following tree:

```
ROOT\RustupTest-TEST_ID\SUBKEY
```

Finally, I think this function is too simple and can be inlined.

src/cli/self_update/windows.rs
```diff
@@ -752,39 +778,26 @@ pub(crate) fn spawn_uninstall_gc(no_modify_path: bool, process: &Process) -> Res
...
-    USER_PATH.get()
-}
-
-#[cfg(any(test, feature = "test"))]
-pub struct RegistryGuard<'a> {
-    _locked: LockResult<MutexGuard<'a, ()>>,
-    id: &'static RegistryValueId,
-    prev: Option<Value>,
-}
+pub const RUSTUP_TEST_REGISTRY_ID: &str = "RUSTUP_TEST_REGISTRY_ID";
```
rami3l:
```rs
pub const RUSTUP_REGISTRY_TEST_ID: &str = "RUSTUP_REGISTRY_TEST_ID";
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

src/test/clitools.rs
```diff
@@ -324,6 +327,12 @@ impl Config {
         if let Some(root) = self.rustup_update_root.as_ref() {
             cmd.env("RUSTUP_UPDATE_ROOT", root);
         }
+
+        #[cfg(windows)]
+        cmd.env(
+            crate::cli::self_update::RUSTUP_TEST_REGISTRY_ID,
```
rami3l: Consider importing this under `#[cfg(windows)]` at the top.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

src/test/clitools.rs
```diff
@@ -826,6 +835,8 @@ async fn setup_test_state(test_dist_dir: TempDir) -> (TempDir, Config) {
         rustup_update_root: None,
         workdir: RefCell::new(workdir),
         test_root_dir: test_dir.path().to_path_buf(),
+        #[cfg(windows)]
+        test_registry_id: crate::test::test_id(),
```
rami3l: Consider importing this under `#[cfg(windows)]` at the top.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

tests/suite/cli_paths.rs
```diff
@@ -504,46 +504,54 @@ error: could not amend shell profile[..]
...
         let cx = CliTestContext::new(Scenario::Empty).await;
-        let _guard = RegistryGuard::new(&USER_PATH).unwrap();
+        let test_id = &cx.config.test_registry_id;
         let cfg_path = cx.config.cargodir.join("bin").display().to_string();
-        let get_path_ = || {
-            HSTRING::try_from(get_path().unwrap().unwrap())
-                .unwrap()
-                .to_string()
+        let read_path = |test_id: &str| -> Option<String> {
+            retry(
```
rami3l, line 523: Did you observe any related test failures (in which case I think it's okay to leave it as it is, but it'd be better to add some comments explaining the failure mode)?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..e40d5d651c50a7a629f48ee9c1199e308037f8bf)*

rami3l (changes requested):

tests/suite/cli_paths.rs
baka-gourd: When testing on my machine, read_path sometimes fails; this is essentially a timing issue

rami3l (approved):
LGTM and thank you so much :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..71b575ada4fd42fb1e92f727a54b228d028d20aa)*

baka-gourd:
```diff
-let environment = process.registry_environment_key()?;
+let environment = process.registry_key("Environment", CURRENT_USER)?;
```

I don't think that's a good idea. Even though it only has one function, using inline code will lead to a lot of duplication, which makes it harder to modify later.

rami3l:
> ```diff
> 
> -let environment = process.registry_environment_key()?;
> 
> +let environment = process.registry_key("Environment", CURRENT_USER)?;
> 
> ```
> 
> 
> 
> I don't think that's a good idea. Even though it only has one function, using inline code will lead to a lot of duplication, which makes it harder to modify later.

Sorry there might be some GitHub availability issues going on; I'm okay with both the inlined and the non-inlined versions of this call, as long as it's consistent.

ChrisDenton (approved):
This approach seems fine to me. There's still the slight risk that, if something goes badly wrong, there will be left over registry entries. But the impact of that is limited and in any case they can be cleaned up on the next test run.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4918/79766cfa9a13f26b7485a870cce9b4b3c9ada481..71b575ada4fd42fb1e92f727a54b228d028d20aa)*

rami3l:
> This approach seems fine to me. There's still the slight risk that, if something goes badly wrong, there will be left over registry entries. 

@ChrisDenton Indeed; but in this regard I'd like to add that this PR is using the volatile option for registry operations during tests for that very reason, so in the worst case the temp keys created here shouldn't survive a system reboot.

ChrisDenton:
> the temp keys created here shouldn't survive a system reboot

Yep, though for some of us that can be a long time 😉

PR #4931, opened 2026-06-30
Clean up lints

djc:


src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![warn(clippy::use_self)]
```
rami3l: Nit: This lint is unconditional so suggest putting it to `Cargo.toml` and be applied to the whole repo like the rest of the clippy lints :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4931/a6635323661932252e24ae7cd281d3bcbea05f4a..20aee1510a5a3f0b49df919bd9a766103d64e088)*

rami3l (approved):

src/lib.rs
djc: Yup, moved it already.

PR #4932, opened 2026-06-30
fix(toolchain)!: restrict named toolchain characters

cachebag:
Partially addresses #4059 by restricting what characters a named toolchain may contain.

Named toolchains are now validated against the UTS [#39](https://www.unicode.org/reports/tr39/) general security profile.

In essence; `unicode_security::GeneralSecurityProfile::identifier_allowed()` vs this hand-rolled pattern we had before. Letters and digits in any script stay legal, so `合法的` still works, as do `.`, `_` and `-`. Whitespace, most punctuation, emoji, and invisible or direction-altering characters are rejected. I also exclude `:` (NTFS alternate data streams) and `'` (shell quoting), plus `.` and `..`, which the profile allows.

This is breaking for unusual names like `foo#bar` or names with spaces. However it doesn't touch the `rust-toolchain.toml` diagnostics. Confusables are also still unresolved (i.e.`μ` vs `µ`, precomposed vs decomposed `é`).

### DIFF
--- a/doc/user-guide/src/concepts/toolchains.md
+++ b/doc/user-guide/src/concepts/toolchains.md
@@ -57,6 +57,14 @@ local builds of the Rust toolchain. To teach `rustup` about your build, run:
 $ rustup toolchain link my-toolchain path/to/my/toolchain/sysroot
 ```
 
+Custom toolchain names may contain any character that [UTS #39] allows in an
+identifier, except for `:` and `'`. In practice that means letters and digits
+in any script, along with `.`, `_`, and `-`; whitespace, most punctuation,
+emoji, and invisible or direction-altering characters are rejected. A name may
+not start with `-`, which would be read as an argument, nor be `.` or `..`.
+
+[UTS #39]: https://www.unicode.org/reports/tr39/#Identifier_Status_and_Type
+
 For example, on Ubuntu you might clone `rust-lang/rust` into `~/rust`, build
 it, and then run:
 
--- a/src/toolchain/names.rs
+++ b/src/toolchain/names.rs
@@ -71,19 +71,6 @@ pub enum InvalidName {
     PlusPrefix(String),
 }
 
-/// Common validate rules for all sorts of toolchain names
-fn validate(candidate: &str) -> Result<&str, InvalidName> {
-    if let Some(without_plus) = candidate.strip_prefix('+') {
-        return Err(InvalidName::PlusPrefix(without_plus.to_string()));
-    }
-    let normalized_name = candidate.trim_end_matches('/');
-    if normalized_name.is_empty() {
-        Err(InvalidName::ToolchainName(candidate.into()))
-    } else {
-        Ok(normalized_name)
-    }
-}
-
 /// A toolchain name from user input.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub(crate) enum ResolvableToolchainName {
@@ -103,7 +90,7 @@ impl ResolvableToolchainName {
     // If candidate could be resolved, return a ready to resolve version of it.
     // Otherwise error.
     fn validate(candidate: &str) -> Result<Self, InvalidName> {
-        let candidate = validate(candidate)?;
+        let candidate = normalize_name(candidate)?;
         if let Ok(desc) = PartialToolchainDesc::from_str(candidate) {
             return Ok(Self::Official(desc));
         }
@@ -151,7 +138,7 @@ impl MaybeResolvableToolchainName {
     // If candidate could be resolved, return a ready to resolve version of it.
     // Otherwise error.
     fn validate(candidate: &str) -> Result<Self, InvalidName> {
-        Ok(match validate(candidate)? {
+        Ok(match normalize_name(candidate)? {
             "none" => Self::None,
             candidate => Self::Some(ResolvableToolchainName::validate(candidate)?),
         })
@@ -185,7 +172,7 @@ pub(crate) enum MaybeOfficialToolchainName {
 
 impl MaybeOfficialToolchainName {
     fn validate(candidate: &str) -> Result<Self, InvalidName> {
-        Ok(match validate(candidate)? {
+        Ok(match normalize_name(candidate)? {
             "none" => Self::None,
             candidate => Self::Some(
                 PartialToolchainDesc::from_str(candidate)
@@ -224,7 +211,7 @@ pub enum ToolchainName {
 impl ToolchainName {
     /// If the string is already resolved, allow direct conversion
     fn validate(candidate: &str) -> Result<Self, InvalidName> {
-        let candidate = validate(candidate)?;
+        let candidate = normalize_name(candidate)?;
         if let Ok(desc) = ToolchainDesc::from_str(candidate) {
             return Ok(Self::Official(desc));
         }
@@ -285,7 +272,7 @@ impl ResolvableLocalToolchainName {
 
     /// Validates if the string is a resolvable toolchain, or a path based toolchain.
     fn validate(candidate: &str) -> Result<Self, InvalidName> {
-        let candidate = validate(candidate)?;
+        let candidate = normalize_name(candidate)?;
         if let Ok(name) = ResolvableToolchainName::from_str(candidate) {
             return Ok(Self::Named(name));
         }
@@ -373,7 +360,7 @@ pub struct CustomToolchainName(String);
 
 impl CustomToolchainName {
     fn validate(candidate: &str) -> Result<Self, InvalidName> {
-        let candidate = validate(candidate)?;
+        let candidate = normalize_name(candidate)?;
         if candidate.parse::<PartialToolchainDesc>().is_ok()
             || candidate == "none"
             || candidate.contains('/')
@@ -460,6 +447,23 @@ impl Deref for PathBasedToolchainName {
 
     fn deref(&self) -> &PathBuf {
         &self.0
+    }
+}
+
+/// Normalization shared by all sorts of toolchain names.
+///
+/// Strips the trailing slashes a shell may have completed onto a toolchain
+/// directory, and rejects the `+toolchain` argument form along with names that
+/// are empty once normalized.
+fn normalize_name(candidate: &str) -> Result<&str, InvalidName> {
+    if let Some(without_plus) = candidate.strip_prefix('+') {
+        return Err(InvalidName::PlusPrefix(without_plus.to_string()));
+    }
+    let normalized_name = candidate.trim_end_matches('/');
+    if normalized_name.is_empty() {
+        Err(InvalidName::ToolchainName(candidate.into()))
+    } else {
+        Ok(normalized_name)
     }
 }
 
--- a/src/toolchain/names.rs
+++ b/src/toolchain/names.rs
@@ -49,6 +49,7 @@ use std::{
 };
 
 use thiserror::Error;
+use unicode_security::GeneralSecurityProfile;
 
 use crate::dist::{PartialToolchainDesc, TargetTuple, ToolchainDesc};
 
@@ -69,6 +70,8 @@ pub enum InvalidName {
     ToolchainName(String),
     #[error("invalid toolchain name '+{0}'; valid toolchain names do not start with '+'")]
     PlusPrefix(String),
+    #[error("invalid toolchain name '-{0}'; valid toolchain names do not start with '-'")]
+    DashPrefix(String),
 }
 
 /// A toolchain name from user input.
@@ -94,7 +97,7 @@ impl FromStr for ResolvableToolchainName {
     // If value could be resolved, return a ready to resolve version of it.
     // Otherwise error.
     fn from_str(value: &str) -> Result<Self, Self::Err> {
-        let candidate = normalize_name(value)?;
+        let candidate = validate_name(value)?;
         if let Ok(desc) = PartialToolchainDesc::from_str(candidate) {
             return Ok(Self::Official(desc));
         }
@@ -164,7 +167,7 @@ impl FromStr for MaybeOfficialToolchainName {
     type Err = InvalidName;
 
     fn from_str(value: &str) -> Result<Self, Self::Err> {
-        Ok(match normalize_name(value)? {
+        Ok(match validate_name(value)? {
             "none" => Self::None,
             candidate => Self::Some(
                 PartialToolchainDesc::from_str(candidate)
@@ -209,7 +212,7 @@ impl FromStr for ToolchainName {
 
     /// If the string is already resolved, allow direct conversion
     fn from_str(value: &str) -> Result<Self, Self::Err> {
-        let candidate = normalize_name(value)?;
+        let candidate = validate_name(value)?;
         if let Ok(desc) = ToolchainDesc::from_str(candidate) {
             return Ok(Self::Official(desc));
         }
@@ -259,9 +262,13 @@ impl FromStr for ResolvableLocalToolchainName {
             return Ok(Self::Named(name));
         }
 
-        Ok(Self::Path(PathBasedToolchainName::try_from(
-            &PathBuf::from(candidate) as &Path,
-        )?))
+        if candidate.contains('/') || candidate.contains('\\') {
+            let path = PathBuf::from(candidate);
+            let path = PathBasedToolchainName::try_from(&path as &Path)?;
+            return Ok(Self::Path(path));
+        }
+
+        Err(InvalidName::ToolchainName(candidate.into()))
     }
 }
 
@@ -344,12 +351,8 @@ impl FromStr for CustomToolchainName {
     type Err = InvalidName;
 
     fn from_str(value: &str) -> Result<Self, Self::Err> {
-        let candidate = normalize_name(value)?;
-        if candidate.parse::<PartialToolchainDesc>().is_ok()
-            || candidate == "none"
-            || candidate.contains('/')
-            || candidate.contains('\\')
-        {
+        let candidate = validate_name(value).map_err(|_| InvalidName::CustomName(value.into()))?;
+        if candidate.parse::<PartialToolchainDesc>().is_ok() || candidate == "none" {
             Err(InvalidName::CustomName(candidate.into()))
         } else {
             Ok(Self(candidate.into()))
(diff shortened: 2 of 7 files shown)


cachebag:
Note to myself that this will probably need some corrections once #4930 is merged

src/toolchain/names.rs
```diff
@@ -117,19 +117,6 @@ macro_rules! try_from_str {
     };
 }
 
-/// Common validate rules for all sorts of toolchain names
```
djc, line 74: Don't move code in the same commit as changing it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/a6635323661932252e24ae7cd281d3bcbea05f4a..9118fccee7389d9ea8508f9f341d56164a3016a7)*

src/toolchain/names.rs
```diff
@@ -473,6 +466,36 @@ impl Deref for PathBasedToolchainName {
...
+    } else {
+        Ok(normalized_name)
+    }
+}
+
+fn is_legal_named_toolchain(candidate: &str) -> bool {
+    !matches!(candidate, "." | "..") && candidate.chars().all(is_legal_named_toolchain_char)
+}
+
+fn is_legal_named_toolchain_char(c: char) -> bool {
```
djc: Nit: suggest inlining the single-caller function.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/a6635323661932252e24ae7cd281d3bcbea05f4a..9118fccee7389d9ea8508f9f341d56164a3016a7)*

src/toolchain/names.rs
cachebag: Sure. But doesn't rustup follow a top-down code style?

djc: Yes, and I am in favor of moving it. But prefer to isolate moving it in a separate commit; see

https://rust-lang.github.io/rustup/dev-guide/coding-standards.html#atomic-commits

rami3l:
@cachebag Thanks for this PR!

I think given your comment in https://github.com/rust-lang/rustup/pull/4932#issuecomment-4843704923 I'll probably review https://github.com/rust-lang/rustup/pull/4930 first and come back to this one later, many thanks for your understanding 🙏

src/toolchain/names.rs
```diff
@@ -151,8 +167,8 @@ impl ResolvableToolchainName {
 
     // If candidate could be resolved, return a ready to resolve version of it.
     // Otherwise error.
-    fn validate(candidate: &str) -> Result<Self, InvalidName> {
-        let candidate = validate(candidate)?;
+    fn validate(candidate: &str) -> Result<ResolvableToolchainName, InvalidName> {
```
djc: Nit: should keep the `Self` here.

(I noticed this happens in the next commit, but it should instead not be changed in this commit.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/47139f6d64daa76230f427ac5c970c41175ea9e1..955c8cbf00c817044019d73936a49414ce0a3326)*

src/toolchain/names.rs
```diff
@@ -307,8 +323,13 @@ impl ResolvableLocalToolchainName {
         ResolvableToolchainName::try_from(candidate)
             .map(ResolvableLocalToolchainName::Named)
             .or_else(|_| {
-                PathBasedToolchainName::try_from(&PathBuf::from(candidate) as &Path)
-                    .map(ResolvableLocalToolchainName::Path)
+                let path = PathBuf::from(candidate);
+                if candidate.contains('/') || candidate.contains('\\') {
```
djc: Should we use `std::path::MAIN_SEPARATOR` instead to make this platform-specific, or would that be too confusing? Otherwise, should we clarify that this is intended to cover platform separators for all common platforms?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/47139f6d64daa76230f427ac5c970c41175ea9e1..955c8cbf00c817044019d73936a49414ce0a3326)*

src/toolchain/names.rs
```diff
@@ -117,35 +117,6 @@ macro_rules! try_from_str {
...
-    let candidate = validate(candidate)?;
-    if is_legal_named_toolchain(candidate) {
-        Ok(candidate)
-    } else {
-        Err(InvalidName::ToolchainName(candidate.into()))
-    }
-}
-
-/// Common validate rules for all sorts of toolchain names
-fn validate(candidate: &str) -> Result<&str, InvalidName> {
```
djc, line 75: IMO it would be nicer to move this commit (moving `validate()` down) before the other commit introducing new related helpers, so that there's fewer changes between commits in this PR -> less work for reviewers.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/47139f6d64daa76230f427ac5c970c41175ea9e1..955c8cbf00c817044019d73936a49414ce0a3326)*

src/toolchain/names.rs
cachebag: Yes it seems a bit odd to do that here if it's meant to target all platforms.

> Otherwise, should we clarify that this is intended to cover platform separators for all common platforms?

Agreed.

src/toolchain/names.rs
```diff
@@ -131,15 +131,15 @@ impl ResolvableToolchainName {
     // If candidate could be resolved, return a ready to resolve version of it.
```
rami3l: Nit: Now that you are introducing an incompatible change, please use the breaking marker `!` (`fix(toolchain)!: restrict named toolchain characters`) to mark this commit as not backportable (I believe the previous commits can stay as-is?). Thanks :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..a8ed506a949adb88dea2b0af87c59e48b487a7e6)*

src/toolchain/names.rs
```diff
@@ -131,15 +131,15 @@ impl ResolvableToolchainName {
...
-        let candidate = validate(candidate)?;
-        if let Ok(desc) = PartialToolchainDesc::from_str(candidate) {
-            return Ok(Self::Official(desc));
-        }
-
-        match CustomToolchainName::try_from(candidate) {
-            Ok(custom) => Ok(Self::Custom(custom)),
-            Err(_) => Err(InvalidName::ToolchainName(candidate.into())),
-        }
+        let candidate = validate_named_toolchain(candidate)?;
```
rami3l: Nit: Given that the previous change #4930 has focused on rewriting the existing validation code from functional to imperative, suggest keeping the imperative style here to minimize the diff. Same with the other functions that you may or may not have changed in this patch.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..a8ed506a949adb88dea2b0af87c59e48b487a7e6)*

src/toolchain/names.rs
```diff
@@ -447,6 +453,15 @@ impl Deref for PathBasedToolchainName {
     }
 }
 
+fn validate_named_toolchain(candidate: &str) -> Result<&str, InvalidName> {
```
rami3l: Question: what is the rationale for having two distinct functions `validate()` and `validate_named_toolchain()` in this module?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..a8ed506a949adb88dea2b0af87c59e48b487a7e6)*

src/toolchain/names.rs
```diff
@@ -460,6 +475,13 @@ fn validate(candidate: &str) -> Result<&str, InvalidName> {
     }
 }
 
+fn is_legal_named_toolchain(candidate: &str) -> bool {
```
rami3l: Question: How is it/why should it be different from `validate()`/`validate_named_toolchain()`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..a8ed506a949adb88dea2b0af87c59e48b487a7e6)*

rami3l: Nit: If we decide to keep it, it looks like it should be called `is_legal_toolchain_name()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..a8ed506a949adb88dea2b0af87c59e48b487a7e6)*

cachebag: `validate()` only does shared normalization/basic checks: reject `+`, trim trailing `/`, reject empty. It still allows paths and old broad custom names.

`validate_named_toolchain()` builds on that and rejects illegal named-toolchain characters, returning a `ToolchainName` error.

It feels to me like it is the reusable predicate for just the allowlist. I wrote it because `CustomToolchainName::validate()` needs the same allowlist but must return `CustomName`, not `ToolchainName`.

That being said the names are not great. We could probably have them be `normalize_toolchain_name()`, `is_valid_toolchain_name` and `validate_toolchain_name()`?

src/toolchain/names.rs
cachebag: As touched on in https://github.com/rust-lang/rustup/pull/4932#discussion_r3505850186 - `validate()` is a common baseline used by both named and path-capable inputs. The latter is stricter for the purpose of the introduced allowlist

src/toolchain/names.rs
cachebag: See above responses. 

I do like `is_legal_toolchain_name()` more than `is_valid_...()`

djc: Why? We have a bunch of `validate_*` functions to I like that `is_valid_` references that more clearly than introducing a seemingly separate concept of "legality".

cachebag: "legal" seemed to reflect the new allowlist/policy. i'm not hard pressed on the naming convention, `is_valid_` is better for your reasons, that makes sense to me from a code standpoint

src/toolchain/names.rs
```diff
@@ -460,6 +465,22 @@ fn validate(candidate: &str) -> Result<&str, InvalidName> {
     }
 }
 
+fn validate_named_toolchain(candidate: &str) -> Result<&str, InvalidName> {
```
djc: Nit: since this calls `validate()`, it should be defined before it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..c1a8e40051677c440d0614d5cb598bd9eff0290d)*

src/toolchain/names.rs
```diff
@@ -460,6 +465,22 @@ fn validate(candidate: &str) -> Result<&str, InvalidName> {
...
+    let candidate = validate(candidate)?;
+    if is_legal_named_toolchain(candidate) {
+        Ok(candidate)
+    } else {
+        Err(InvalidName::ToolchainName(candidate.into()))
+    }
```
djc: Nit: suggest writing this as a `match`:

```rust
match is_legal_named_toolchain(candidate) {
     true => Ok(candidate),
     false => Err(InvalidName::ToolchainName(candidate.to_owned()),
}
```

(Also using `to_owned()` instead of `into()` which is weaker and therefore makes it clearer what is going on.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/4b1423bf74aa6102068b34ab52ccc28de5ff2aa9..c1a8e40051677c440d0614d5cb598bd9eff0290d)*

djc (approved):

src/toolchain/names.rs
cachebag: Done.

Curious- what do you mean by weaker? Do you mean semantically? It was my understanding that `.into()` was generally more flexible but had no runtime difference.

> and therefore makes it clearer what is going on.

This makes sense to me

djc: I mean "weaker" in the sense of "less expressive" -- it can do fewer things and thus as a reviewer I have less to think about in evaluating it.

src/toolchain/names.rs
rami3l: Sorry for not having expressed my intentions super clearly. My question is rather: after your changes, do the `validate()` function still have any reasons to exist? Are there any cases where we want the normalization without the legality check/validation?

If not or if it becomes single-used, then maybe it's better just to keep the actual "validation" function to avoid confusions.

src/toolchain/names.rs
rami3l: Looking at your proposed name it looks like `is_valid_toolchain_name(name)` can be inlined as `validate_toolchain_name(name).is_ok()`. Of course this is not the case currently but I can't help thinking about moving towards that direction.

Looking at the docstring of `validate()`:

```rs
/// Common validate rules for all sorts of toolchain names
```

It looks like we should just do all possible verification passes in one place? Would that be possible? Something like the following, for example:

```diff
impl CustomToolchainName {
    fn validate(candidate: &str) -> Result<Self, InvalidName> {
-       let candidate = validate(candidate)?;
+       // At this point there is already no need to check for `/` and `\`.
+       let candidate = validate_toolchain_name(candidate)?;
        if candidate.parse::<PartialToolchainDesc>().is_ok()
            || candidate == "none"
-           || candidate.contains('/')
-           || candidate.contains('\\')
        {
            return Err(InvalidName::CustomName(candidate.into()))
        }
        Ok(Self(candidate.into()))
    }
}
```

What I am trying to say is that in that case `is_valid_toolchain_name(name)` is not needed anymore.

cachebag: Makes sense. Done.

src/toolchain/names.rs
cachebag: Right- so `validate()` is still needed because `ResolvableLocalToolchainName::validate` uses it to handle both named toolchains and paths, so it needs the basic normalization without the character allowlist.

https://github.com/rust-lang/rustup/blob/3432d809f48686b13592fffad7f66a111f08c922/src/toolchain/names.rs#L293-L305

More concretely, it's used in three places: `ResolvableLocalToolchainName::validate`, `MaybeResolvableToolchainName::validate` and `validate_named_toolchain` itself.

`validate_named_toolchain()` builds on top of it for the stricter named-only callers.

I've applied your suggestion for `CustomToolchainName`. It now uses `validate_named_toolchain()` and `is_legal_named_toolchain` has been inlined.

Is there something I am misunderstanding? Are you asking if we should inline `validate()`'s logic? If so, that repeats code more than we would be by just keeping these two functions separate.

Though maybe I am just misunderstanding your question completely, and if so then pardon me.

rami3l:
@cachebag Sorry for the delay in reviewing! I have to see if the current code can be further streamlined and what we should do about https://github.com/rust-lang/rustup/issues/4059#issuecomment-4867181775. Please stay tuned...

src/toolchain/names.rs
```diff
@@ -475,8 +443,12 @@ fn validate_named_toolchain(candidate: &str) -> Result<&str, InvalidName> {
...
 }
 
-/// Common validate rules for all sorts of toolchain names
-fn validate(candidate: &str) -> Result<&str, InvalidName> {
+/// Normalization shared by all sorts of toolchain names.
+///
+/// Strips the trailing slashes a shell may have completed onto a toolchain
+/// directory, and rejects the `+toolchain` argument form along with names that
+/// are empty once normalized.
+fn normalize_name(candidate: &str) -> Result<&str, InvalidName> {
```
rami3l, line 458: This is a good idea! In this case can we use `validate_name()` instead of `validate_named_toolchain()` for the previous function?

src/toolchain/names.rs
```diff
@@ -475,8 +443,12 @@ fn validate_named_toolchain(candidate: &str) -> Result<&str, InvalidName> {
     }
 }
 
-/// Common validate rules for all sorts of toolchain names
-fn validate(candidate: &str) -> Result<&str, InvalidName> {
+/// Normalization shared by all sorts of toolchain names.
```
rami3l: Nit: Suggest putting the renaming commit(s) in the beginning of this PR.

doc/user-guide/src/concepts/toolchains.md
```diff
@@ -57,8 +57,13 @@ local builds of the Rust toolchain. To teach `rustup` about your build, run:
 $ rustup toolchain link my-toolchain path/to/my/toolchain/sysroot
 ```
 
-Custom toolchain names may contain ASCII letters, ASCII digits, `.`, `_`, and
-`-`.
+Custom toolchain names may contain any character that [UTS #39] allows in an
```
rami3l, line 60: Nit: I think this commit can be squashed into previous ones to keep the overall diff minimized, as we are not really going to adopt the ASCII-only string pattern for toolchain names.

src/toolchain/names.rs
```diff
@@ -463,6 +420,52 @@ impl Deref for PathBasedToolchainName {
...
+    } else {
+        Ok(normalized_name)
+    }
+}
+
+/// Characters that UTS #39 permits in identifiers but that we reject anyway,
+/// because a named toolchain also has to work as a directory name under
+/// `.rustup/toolchains`: `:` would open an NTFS alternate data stream, and `'`
+/// needs quoting in enough shells to be a nuisance.
+const EXCLUDED: &[char] = &[':', '\''];
```
rami3l, line 474: Very carefully designed, nice work :)

rami3l (commented):
LGTM modulo some tiny nits, nice work! 🙏 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/62df1e988b995b4410acbf36fde56f43111f72ad..028f723a35969d4424d7d2fd2771a08270c091bd)*

src/toolchain/names.rs
```diff
@@ -463,6 +418,23 @@ impl Deref for PathBasedToolchainName {
     }
 }
 
+/// Normalization shared by all sorts of toolchain names.
+///
+/// Strips the trailing slashes a shell may have completed onto a toolchain
+/// directory, and rejects the `+toolchain` argument form along with names that
+/// are empty once normalized.
+fn normalize_name(candidate: &str) -> Result<&str, InvalidName> {
```
rami3l, line 458: Nit: Renaming `validate()` to `normalize_name()` should be in a separate commit like before.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/23b2cb0af6418325cbeb88283289946b379e2c5e..a92447802c7a001d64476318720850e4315ad9d8)*

cachebag: Whoops. Missed this one

src/toolchain/names.rs
```diff
@@ -463,6 +450,23 @@ impl Deref for PathBasedToolchainName {
     }
 }
 
+/// Normalization shared by all sorts of toolchain names.
```
rami3l: Nit: Pre-existing, but according to the `std` docstring style this first line should be a single line starting with a verb. Maybe something like "Normalizes candidate toolchain name using generic rules."?

doc/user-guide/src/concepts/toolchains.md
```diff
@@ -57,6 +57,14 @@ local builds of the Rust toolchain. To teach `rustup` about your build, run:
 $ rustup toolchain link my-toolchain path/to/my/toolchain/sysroot
 ```
 
+Custom toolchain names may contain any character that [UTS #39] allows in an
+identifier, except for `:` and `'`. In practice that means letters and digits
+in any script, along with `.`, `_`, and `-`; whitespace, most punctuation,
+emoji, and invisible or direction-altering characters are rejected. The names
+`.` and `..` are not accepted either.
```
rami3l: @ChrisDenton Does this new pattern look good to you?

rami3l (approved):
LGTM, thanks a whole lot for the work!

I'll ask @ChrisDenton for one additional round just to make sure that this new pattern looks good to everyone :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4932/23b2cb0af6418325cbeb88283289946b379e2c5e..bfdbeb49fb34ad9b0249ca2b43cc1da83037e30f)*

doc/user-guide/src/concepts/toolchains.md
ChrisDenton: Yes it looks good except maybe we should also disallow starting with `-` to avoid confusing toolchain names with arguments.

cachebag: This makes sense. Thanks for that

PR #4935, opened 2026-07-02
Remove Windows i686 special case from `can_run`

ChrisDenton:
While it is possible to run a 32-bit toolchain on a 64-bit host, it is strongly discouraged. If needed, the user can use `--force-non-host` to install it anyway. Note that Windows is the only target that has this particular special case.

Demoting 32-bit host to tier 2 is [currently being discussed](https://rust-lang.zulipchat.com/#narrow/channel/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202026-07-02/near/607854580) and so I think rustup should be updated before that change.

This does not affect installing and using `i686` as a *target*. Rust fully supports cross-compiling from a 64-bit Windows host to an `i686-windows` target and will likely always do so.

### DIFF
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -579,13 +579,8 @@ impl TargetTuple {
         let ret = if partial_self.os != partial_other.os {
             false
         } else if partial_self.os.as_deref() == Some("pc-windows") {
-            // Windows is a special case here: we can run gnu and msvc on the same system,
-            // x86_64 can run i686, and aarch64 can run i686 through emulation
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
@rami3l maybe this is worth backporting if it's not too much work?

rami3l (approved):

rami3l:
@djc Yeah it's okay for a 3rd round of backport. I'll utilize the lack of contention and do it as a part of the changelog update if you agree.

PR #4941, opened 2026-07-07
Platforms 4

djc:
Also assorted semver-compatible dependency updates, including a fix for [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204.html).


rami3l (approved):

rami3l:
@djc Thanks a lot for taking care of this!

BTW I don't think this should be a blocker for the ongoing beta release since we have multiple version bumps before the stable is out.

djc:
> BTW I don't think this should be a blocker for the ongoing beta release since we have multiple version bumps before the stable is out.

Fine with me.

PR #4946, opened 2026-07-09
fix: lock state and settings file updates

### LINKED ISSUE #4942: Use `File::lock*()` when reading/writing state and settings files
When reading `state.rs`'s internal state file (and potentially `settings.toml`), we don't currently use file locking, which may cause files to become corrupted, as per https://github.com/rust-lang/rustup/pull/4869#discussion_r3501509485.

ychampion:
## Summary

- Add locked read/write helpers for file contents.
- Use them for `state.toml` and `settings.toml` reads and writes.
- Cover locked write truncation with a unit test.

## Why

Fixes #4942. These files can be touched by concurrent rustup processes, so writes now take an exclusive lock before truncating and reads take a shared lock.

## Validation

- `cargo test --features test locked_file_round_trip_truncates_previous_contents` — passed
- `cargo test --features test settings::tests` — passed
- `cargo test --features test --test test_bonanza notify_release_hint` — passed
- `cargo check --all-targets --features test` — passed
- `cargo clippy --all-targets --features test -- -D warnings` — passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed

### DIFF
--- a/src/utils/raw.rs
+++ b/src/utils/raw.rs
@@ -3,7 +3,7 @@ use std::env;
 use std::fs;
 use std::fs::File;
 use std::io;
-use std::io::Write;
+use std::io::{Read, Seek, SeekFrom, Write};
 use std::path::Path;
 use std::str;
 
@@ -60,6 +60,21 @@ pub fn path_exists<P: AsRef<Path>>(path: P) -> bool {
     fs::metadata(path).is_ok()
 }
 
+pub(crate) fn read_locked_file(path: &Path) -> io::Result<String> {
+    let mut file = File::open(path)?;
+    file.lock_shared()?;
+
+    let size = file
+        .metadata()
+        .map(|metadata| usize::try_from(metadata.len()).unwrap_or(usize::MAX))
+        .ok();
+    let mut contents = String::new();
+    contents.try_reserve_exact(size.unwrap_or(0))?;
+    file.read_to_string(&mut contents)?;
+
+    Ok(contents)
+}
+
 pub(crate) fn random_string(length: usize) -> String {
     const CHARSET: &[u8] = b"abcdefghijklmnopqrstuvwxyz0123456789_";
     let mut rng = rand::rng();
@@ -77,6 +92,24 @@ pub fn write_file(path: &Path, contents: &str) -> io::Result<()> {
 
     Write::write_all(&mut file, contents.as_bytes())?;
 
+    file.sync_data()?;
+
+    Ok(())
+}
+
+pub(crate) fn write_locked_file(path: &Path, contents: &str) -> io::Result<()> {
+    let mut file = fs::OpenOptions::new()
+        .read(true)
+        .write(true)
+        .create(true)
+        // Truncation must happen after the exclusive lock is held.
+        .truncate(false)
+        .open(path)?;
+
+    file.lock()?;
+    file.set_len(0)?;
+    file.seek(SeekFrom::Start(0))?;
+    Write::write_all(&mut file, contents.as_bytes())?;
     file.sync_data()?;
 
     Ok(())
@@ -348,5 +381,22 @@ pub(crate) mod windows {
             Ok(maybe_result)
         }
         inner(s.as_ref())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn locked_file_round_trip_truncates_previous_contents() -> io::Result<()> {
+        let temp_dir = tempfile::tempdir()?;
+        let path = temp_dir.path().join("state.toml");
+
+        write_locked_file(&path, "last_release_notified_secs = 123456\n")?;
+        write_locked_file(&path, "version = \"12\"\n")?;
+
+        assert_eq!(read_locked_file(&path)?, "version = \"12\"\n");
+        Ok(())
     }
 }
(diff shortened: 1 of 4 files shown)


src/utils/raw.rs
```diff
@@ -60,6 +60,16 @@ pub fn path_exists<P: AsRef<Path>>(path: P) -> bool {
     fs::metadata(path).is_ok()
 }
 
+pub(crate) fn read_locked_file(path: &Path) -> io::Result<String> {
+    let mut file = File::open(path)?;
+    file.lock_shared()?;
+
+    let mut contents = String::new();
```
rami3l, line 71: Nit: Following the source code of https://doc.rust-lang.org/std/fs/fn.read_to_string.html:

```rs
let size = file.metadata().map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX)).ok();
let mut string = String::new();
string.try_reserve_exact(size.unwrap_or(0))?;
```

... it might be better if we can reserve some space here in a similar fashion.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4946/d84e63b55ee411695f210695994e8366aeef0fb8..9dea8b37ddc10e9c6bf1962896149995cf570104)*

ychampion: Done in 6c613d62. I reserve from the file metadata after taking the shared lock, then read into that buffer.

Validation:
- `CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 CARGO_BUILD_JOBS=1 cargo test --features test locked_file_round_trip_truncates_previous_contents`
- `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 CARGO_BUILD_JOBS=1 cargo check --features test -p rustup --lib`
- `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 CARGO_BUILD_JOBS=1 cargo clippy --features test -p rustup --lib -- -D warnings`
- `git diff --check`

rami3l (approved):
LGTM, thanks! Please squash the two commits into one and I'll merge it shortly after.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4946/d84e63b55ee411695f210695994e8366aeef0fb8..6c613d62e8b3d99d1ab78d928d2a30efce835021)*

ychampion:
Squashed into one commit now. I reran the locked-file regression plus fmt/diff checks after the rewrite.

rami3l:
@ychampion The CI is failing but it's probably not your fault; it might be due to an upstream update of Rust.

Since we are at it, would you mind fixing the clippy warnings in **a separate commit**? Many thanks in advance 🙏

djc (commented):
Linting issues might be from Rust 1.97 release that just dropped.

Personally I don't like splitting this code between `utils/mod.rs` and `utils/raw.rs`; maybe just keep the "raw" variants just below the `utils` functions?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4946/d84e63b55ee411695f210695994e8366aeef0fb8..8ba412fee972eba80079fa8db7303a385a0e4287)*

rami3l:
> Personally I don't like splitting this code between `utils/mod.rs` and `utils/raw.rs`; maybe just keep the "raw" variants just below the `utils` functions?


@djc I see your point but refactoring PRs can always follow this one... I would prefer not putting too much pressure on first-time contributors in fixing our pre-existing issues WRT code organization, if you agree 🙏

djc:
> @djc I see your point but refactoring PRs can always follow this one... I would prefer not putting too much pressure on first-time contributors in fixing our pre-existing issues WRT code organization, if you agree 🙏

I might agree but I guess I put less value on seemingly drive-by contributions that seem clearly LLM-assisted.

ychampion:
Done in 42a5b3e5 as a separate commit.

Validation:
- `cargo clippy --all-targets --all-features -- -D warnings` on Rust 1.97.0
- `cargo test --features test --test test_bonanza add_target -- --nocapture`

ychampion:
@djc That's fair. I can move the raw helpers back beside the existing `utils` functions if you'd prefer that in this PR; I kept the latest push to the requested Rust 1.97 Clippy-only cleanup.

djc:
> @djc That's fair. I can move the raw helpers back beside the existing `utils` functions if you'd prefer that in this PR; I kept the latest push to the requested Rust 1.97 Clippy-only cleanup.

Okay, let's keep it as is for now. Would be nice if you can clean up single caller `raw` variants in a next PR.

Please rebase, since the Clippy fixes are already on `main`.

ychampion:
Rebased onto current `main`. The separate Clippy cleanup dropped out because it is already present there; the locking change itself is unchanged.

rami3l:
> > That's fair. I can move the raw helpers back beside the existing `utils` functions if you'd prefer that in this PR; I kept the latest push to the requested Rust 1.97 Clippy-only cleanup.
> 
> Okay, let's keep it as is for now. Would be nice if you can clean up single caller `raw` variants in a next PR.

@ychampion Would you mind evaluating the proposed clean-up of the `raw` module and, if feasible, make a follow-up PR? Many thanks in advance 🙏

PR #4947, opened 2026-07-09
feat(config)!: use unqualified names for `default_toolchain`

### LINKED ISSUE #3651: `rustup set default-host` does not change the default toolchain

### LINKED ISSUE #4945: Should `settings.toml` allow unqualified toolchain names?
Split from https://github.com/rust-lang/rustup/issues/4944#issuecomment-4913146836.

For now, `rustup default <toolchain>` resolves the full toolchain name from `<toolchain>`, which could be a function of `default_host_tuple`, and then stores that in `settings.toml`.

I can see two downsides of the current approach:

- If the user changes `default_host_tuple` after setting `default_toolchain` to a generic name such as `stable`, the default toolchain would still be in the original host triple since it has been resolved beforehand.
- This has made the config file hostile to cross-platform dotfile management, since one may want to use the `stable` toolchain specific to the default host tuple of that platform.

Maybe rustup can start storing the unresolved name (or if it is breaking too many things, under a new subcommand/option) for `default_toolchain` instead? The idea is that one can always store the fully qualified toolchain name to go back to the old behavior.

This way:

- In the first point, after changing `default_host_tuple`, the default toolchain will be resolved in the new host triple.
- In the second point, you will be able to use `stable` in your dotfiles as `default_toolchain` without problems.

Update: The same logic also applies to `rustup override set`, as discovered in #5044.

rami3l:
Stacked on https://github.com/rust-lang/rustup/pull/4950 and https://github.com/rust-lang/rustup/pull/4991.

Closes #3651.
Closes #4945.

### Problem Description

The original problem as seen in both #3651 and #4945 is the following:

> For now, `rustup default <toolchain>` resolves the full toolchain name from `<toolchain>`, which could be a function of `default_host_tuple`, and then stores that in `settings.toml`.
_#4945_

... this is notably problematic because the value of `default_toolchain` is inappropriately coupled with the value of the default host at the wrong moment: not the moment of running any `rustup` command, but the last execution of `rustup default` (whether manually by running that very command, or implicitly upon `rustup-init`):

> This has made the config file hostile to cross-platform dotfile management, since one may want to use the `stable` toolchain specific to the default host tuple of that platform.
_https://github.com/rust-lang/rustup/issues/4945#issue-4836038407_

> i have a stable-x86_64-pc-windows-gnu toolchain as my default [..] i want it to be `-msvc` instead so i can use windbg (i don't have gdb installed). i ran `rustup set default-host x86_64-pc-windows-msvc`, which succeeded, but did not install any toolchains or change the default:
_https://github.com/rust-lang/rustup/issues/3651#issue-2103220226_

### Proposed Solution

This PR makes rustup use unqualified names such as `stable` instead of fully qualified names such as `stable-aarch64-apple-darwin` for the `default_toolchain` field in the `settings.toml` file.

#### Compatibility

This change is breaking exactly because the determination of the default toolchain will change to reflect the change of the default host tuple.

However, the user can always get the old behavior by specifying the fully qualified name when running `rustup default` (such as `rustup default stable-aarch64-apple-darwin`), in the rare case that they would like to decouple the default toolchain from the default host tuple.

For users with an existing `settings.toml`, manually running `rustup default stable` or similar commands should be necessary for them to get the new behavior.

#### Performance

As discussed in https://github.com/rust-lang/rustup/issues/4945#issuecomment-4956813676, since the default host tuple is resolved at the beginning of `rustup`'s execution, this change has introduced no extra I/O operations, and thus the performance impact is negligible.

### Validation

Validated through:

- Local testing.
- A specific test case that reflects @jyn514's original use case as described in https://github.com/rust-lang/rustup/issues/3651.
- Another specific test case that explicitly asserts backwards compatibility with existing `settings.toml` format with a fully qualified `default_toolchain` name.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -272,7 +272,7 @@ impl InstallOpts<'_> {
 
             check_proxy_sanity(cfg.process, components, &desc)?;
 
-            cfg.set_default(Some(&desc.clone().into()))?;
+            cfg.set_default(Some(&partial_desc.into()))?;
             writeln!(cfg.process.stdout().lock())?;
             common::show_channel_update(cfg, PackageUpdate::Toolchain(desc), Ok(status))?;
         }
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -69,7 +69,6 @@ use crate::{
     process::Process,
     toolchain::{
         DistributableToolchain, MaybeOfficialToolchainName, ResolvableToolchainName, Toolchain,
-        ToolchainName,
     },
     utils::{self, ExitCode},
 };
@@ -252,7 +251,8 @@ impl InstallOpts<'_> {
 
         let (components, targets) = (self.components, self.targets);
         let toolchain = self.select_toolchain(cfg)?;
-        if let Some(desc) = toolchain {
+        if let Some(partial_desc) = toolchain {
+            let desc = partial_desc.clone().resolve(&cfg.default_host_tuple()?)?;
             let options =
                 DistOptions::new(components, targets, &desc, cfg.get_profile()?, true, cfg)?;
             let status = if Toolchain::exists(cfg, &desc.clone().into())? {
@@ -284,7 +284,7 @@ impl InstallOpts<'_> {
     /// This function first initializes the default profile and default host tuple in the
     /// configuration, then returns the toolchain that should be installed, or `None` if none is
     /// specified by the user.
-    fn select_toolchain(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
+    fn select_toolchain(self, cfg: &mut Cfg<'_>) -> Result<Option<PartialToolchainDesc>> {
         let Self {
             default_host_tuple,
             default_toolchain,
@@ -342,18 +342,14 @@ impl InstallOpts<'_> {
                         MaybeOfficialToolchainName::None => unreachable!(),
                         MaybeOfficialToolchainName::Some(n) => n,
                     };
-                    Some(toolchain_name.resolve(&cfg.default_host_tuple()?)?)
+                    Some(toolchain_name)
                 }
-                None => match cfg.get_default()? {
+                None => match cfg.get_default_resolvable()? {
                     // Default is installable
-                    Some(ToolchainName::Official(t)) => Some(t),
+                    Some(ResolvableToolchainName::Official(t)) => Some(t),
                     // Default is custom, presumably from a prior install. Do nothing.
-                    Some(ToolchainName::Custom(_)) => None,
-                    None => Some(
-                        "stable"
-                            .parse::<PartialToolchainDesc>()?
-                            .resolve(&cfg.default_host_tuple()?)?,
-                    ),
+                    Some(ResolvableToolchainName::Custom(_)) => None,
+                    None => Some(PartialToolchainDesc::from_str("stable")?),
                 },
             })
         } else {
@@ -1378,11 +1374,7 @@ mod tests {
             };
 
             assert_eq!(
-                "stable"
-                    .parse::<PartialToolchainDesc>()
-                    .unwrap()
-                    .resolve(&cfg.default_host_tuple().unwrap())
-                    .unwrap(),
+                "stable".parse::<PartialToolchainDesc>().unwrap(),
                 opts.select_toolchain(&mut cfg)
                     .unwrap() // result
                     .unwrap() // option
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -70,7 +70,6 @@ use crate::{
     settings::SettingsFile,
     toolchain::{
         DistributableToolchain, MaybeOfficialToolchainName, ResolvableToolchainName, Toolchain,
-        ToolchainName,
     },
     utils::{self, ExitCode},
 };
@@ -270,7 +269,8 @@ impl InstallOpts<'_> {
 
         let (components, targets) = (self.components, self.targets);
         let toolchain = self.select_toolchain(&mut cfg)?;
-        if let Some(desc) = toolchain {
+        if let Some(partial_desc) = toolchain {
+            let desc = partial_desc.clone().resolve(&cfg.default_host_tuple()?)?;
             let options =
                 DistOptions::new(components, targets, &desc, cfg.get_profile()?, true, &cfg)?;
             let status = if Toolchain::exists(&cfg, &desc.clone().into())? {
@@ -302,7 +302,7 @@ impl InstallOpts<'_> {
     /// This function first initializes the default profile and default host tuple in the
     /// configuration, then returns the toolchain that should be installed, or `None` if none is
     /// specified by the user.
-    fn select_toolchain(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
+    fn select_toolchain(self, cfg: &mut Cfg<'_>) -> Result<Option<PartialToolchainDesc>> {
         let Self {
             default_host_tuple,
             default_toolchain,
@@ -360,18 +360,14 @@ impl InstallOpts<'_> {
                         MaybeOfficialToolchainName::None => unreachable!(),
                         MaybeOfficialToolchainName::Some(n) => n,
                     };
-                    Some(toolchain_name.resolve(&cfg.default_host_tuple()?)?)
+                    Some(toolchain_name)
                 }
-                None => match cfg.get_default()? {
+                None => match cfg.get_default_resolvable()? {
                     // Default is installable
-                    Some(ToolchainName::Official(t)) => Some(t),
+                    Some(ResolvableToolchainName::Official(t)) => Some(t),
                     // Default is custom, presumably from a prior install. Do nothing.
-                    Some(ToolchainName::Custom(_)) => None,
-                    None => Some(
-                        "stable"
-                            .parse::<PartialToolchainDesc>()?
-                            .resolve(&cfg.default_host_tuple()?)?,
-                    ),
+                    Some(ResolvableToolchainName::Custom(_)) => None,
+                    None => Some(PartialToolchainDesc::from_str("stable")?),
                 },
             })
         } else {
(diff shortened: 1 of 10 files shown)


src/cli/self_update.rs
```diff
@@ -1022,7 +1017,8 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
 
     let (components, targets) = (opts.components, opts.targets);
     let toolchain = opts.install(cfg)?;
-    if let Some(desc) = &toolchain {
+    if let Some(partial_desc) = toolchain {
+        let desc = &partial_desc.clone().resolve(&cfg.default_host_tuple()?)?;
```
rami3l: Nit: Pre-existing, but later on we have `desc.into()` which is converting `&ToolchainDesc` to something owned. Looks like [`d494459` (#4930)](https://github.com/rust-lang/rustup/pull/4930/changes/d4944592194f82d5dd9181227597e24a10ad3221) is not complete? cc @djc 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/d84e63b55ee411695f210695994e8366aeef0fb8..6b11c035dc2a0d79ce8c32517d5eb03b80808de2)*

src/cli/self_update.rs
```diff
@@ -107,7 +106,7 @@ pub(crate) struct InstallOpts<'a> {
 }
 
 impl InstallOpts<'_> {
-    fn install(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
+    fn install(self, cfg: &mut Cfg<'_>) -> Result<Option<PartialToolchainDesc>> {
```
rami3l: Nit: Pre-existing but this naming is quite confusing, how can `InstallOpts::install()` be actually not installing anything?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/d84e63b55ee411695f210695994e8366aeef0fb8..6b11c035dc2a0d79ce8c32517d5eb03b80808de2)*

src/cli/self_update.rs
djc: That commit just fixed some instances of it, it didn't claim to fix all of it.

rami3l: Fixed in https://github.com/rust-lang/rustup/pull/4950.

src/cli/self_update.rs
rami3l: Fixed in https://github.com/rust-lang/rustup/pull/4950.

src/cli/self_update.rs
```diff
@@ -342,18 +342,14 @@ impl InstallOpts<'_> {
...
+                    Some(ResolvableToolchainName::Official(t)) => Some(t),
                     // Default is custom, presumably from a prior install. Do nothing.
-                    Some(ToolchainName::Custom(_)) => None,
-                    None => Some(
-                        "stable"
-                            .parse::<PartialToolchainDesc>()?
-                            .resolve(&cfg.default_host_tuple()?)?,
-                    ),
+                    Some(ResolvableToolchainName::Custom(_)) => None,
+                    None => Some(PartialToolchainDesc::from_str("stable")?),
```
rami3l, line 352: Note: Add a test case for https://github.com/rust-lang/rustup/issues/3651#issuecomment-5058868560

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/10f348abb608c00133ff48be104fbed74cd7daac..bb5bf44baaac49ea831a7617fd5951a02514f47e)*

src/cli/self_update.rs
```diff
@@ -272,7 +272,7 @@ impl InstallOpts<'_> {
 
             check_proxy_sanity(cfg.process, components, &desc)?;
 
-            cfg.set_default(Some(&desc.clone().into()))?;
+            cfg.set_default(Some(&partial_desc.into()))?;
```
rami3l, line 275: Note: Add a test case for https://github.com/rust-lang/rustup/issues/3651#issuecomment-5058663951.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/10f348abb608c00133ff48be104fbed74cd7daac..bb5bf44baaac49ea831a7617fd5951a02514f47e)*

src/config.rs
```diff
@@ -904,6 +904,16 @@ impl<'a> Cfg<'a> {
     /// If none is configured, returns None
     /// If a bad toolchain name is configured, errors.
     pub(crate) fn get_default(&self) -> Result<Option<ToolchainName>> {
+        let Some(toolchain) = self.get_default_resolvable()? else {
+            return Ok(None);
+        };
+        Ok(Some(toolchain.resolve(&self.default_host_tuple()?)?))
+    }
+
+    /// Gets the configured default toolchain.
```
djc: Nit: documentation comment style.

Also, duplicating the documentation comment from `get_default()` seems pretty unhelpful. Ideally these documentation comments should tell a clear story about the difference and their interaction.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/720e6b862df1309c0a5b2aea7f81be1f975af41f..5616ba497d7ea0fb6841204887d4663ea4cc3213)*

src/toolchain/names.rs
```diff
@@ -304,6 +304,17 @@ impl ResolvableLocalToolchainName {
...
+from_variant!(
+    PartialToolchainDesc,
+    ResolvableToolchainName,
+    ResolvableToolchainName::Official
+);
+from_variant!(
+    CustomToolchainName,
+    ResolvableToolchainName,
+    ResolvableToolchainName::Custom
+);
```
djc: I'm not a fan of these macros and have looked several times at cleaning them up. How about just writing them out (could inline macros with Rust-Analyzer I think)?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/720e6b862df1309c0a5b2aea7f81be1f975af41f..5616ba497d7ea0fb6841204887d4663ea4cc3213)*

src/cli/self_update.rs
```diff
@@ -342,18 +342,14 @@ impl InstallOpts<'_> {
                         MaybeOfficialToolchainName::None => unreachable!(),
                         MaybeOfficialToolchainName::Some(n) => n,
                     };
-                    Some(toolchain_name.resolve(&cfg.default_host_tuple()?)?)
+                    Some(toolchain_name)
                 }
-                None => match cfg.get_default()? {
+                None => match cfg.get_default_resolvable()? {
```
djc, line 347: Does `get_default()` go unused? Where is it still used after this?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/720e6b862df1309c0a5b2aea7f81be1f975af41f..5616ba497d7ea0fb6841204887d4663ea4cc3213)*

rami3l: It still has 10+ callers after this change. That is expected: many places still expect fully resolved default toolchains.

src/config.rs
rami3l: @djc I think the new version should be better now?

src/toolchain/names.rs
rami3l: @djc Done; also in that case I can make a separate PR to remove them entirely.

rami3l: Made https://github.com/rust-lang/rustup/pull/4991.

rami3l:
@djc Would you mind giving this PR another look please? 🙇

djc:
> Stacked on #4950 and #4991, closes #4945.
> 
> The breakage should be minimal anyway because the default host tuple almost never changes.

I think the PR description should explain (a) what you're trying to achieve here and (b) how you're trying to achieve it, and probably (c) the effects on backwards compatibility, instead of just relying on issue/PR references.

rami3l:
> I think the PR description should explain (a) what you're trying to achieve here and (b) how you're trying to achieve it, and probably (c) the effects on backwards compatibility, instead of just relying on issue/PR references.

@djc Thank you so much for this suggestion! I have completely rewritten the PR description to include the 3 points mentioned above. Please don't hesitate to tell me if you happen to have any more concerns, as I will try my best to address them.

djc:
> This change is breaking exactly because the determination of the default toolchain will change to reflect the change of the default host tuple.
> 
> However, the user can always get the old behavior by specifying the fully qualified name when running `rustup default` (such as `rustup default stable-aarch64-apple-darwin`), in the rare case that they would like to decouple the default toolchain from the default host tuple.
> 
> For users with an existing `settings.toml`, manually running `rustup default stable` or similar commands should be necessary for them to get the new behavior.

In the code, it was not clear to me how/where old configurations with fully qualified names where still kept supported. I think it would be good to add some comments that make it obvious where/how that happens and that it must not regress.

rami3l:
> In the code, it was not clear to me how/where old configurations with fully qualified names where still kept supported. I think it would be good to add some comments that make it obvious where/how that happens and that it must not regress.

@djc Sure! I have just added one regression test specifically for the case you've mentioned, asserting that `settings.toml` can contain a fully qualified toolchain name which can be generated by `rustup default` followed by that name, both before and after this breaking change.

In addition, I've added two more assertions against the exact shape of `settings.toml` in the originally introduced test case.

djc (approved):
Thanks for the clarifications!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4947/8e6dbd213acfaa457a0bae0390688719d1f38ac3..b24bd2b9b15c180f537dc273fb76b0952eefe601)*

PR #4950, opened 2026-07-10
refactor(cli/self-update)!: reorganize `InstallOpts` methods

rami3l:
Addresses https://github.com/rust-lang/rustup/pull/4947#discussion_r3551051461 and https://github.com/rust-lang/rustup/pull/4947#discussion_r3551059053.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -109,7 +109,12 @@ pub(crate) struct InstallOpts<'a> {
 }
 
 impl InstallOpts<'_> {
-    fn install(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
+    /// Selects the toolchain to install based on the user's intent.
+    ///
+    /// This function first initializes the default profile and default host tuple in the
+    /// configuration, then returns the toolchain that should be installed, or `None` if none is
+    /// specified by the user.
+    fn select_toolchain(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
         let Self {
             default_host_tuple,
             default_toolchain,
@@ -869,7 +874,7 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
     }
 
     let (components, targets) = (opts.components, opts.targets);
-    let toolchain = opts.install(cfg)?;
+    let toolchain = opts.select_toolchain(cfg)?;
     if let Some(desc) = toolchain {
         let options = DistOptions::new(components, targets, &desc, cfg.get_profile()?, true, cfg)?;
         let status = if Toolchain::exists(cfg, &desc.clone().into())? {
@@ -1377,7 +1382,7 @@ mod tests {
                     .unwrap()
                     .resolve(&cfg.default_host_tuple().unwrap())
                     .unwrap(),
-                opts.install(&mut cfg)
+                opts.select_toolchain(&mut cfg)
                     .unwrap() // result
                     .unwrap() // option
             );
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -870,9 +870,9 @@ async fn maybe_install_rust(opts: InstallOpts<'_>, cfg: &mut Cfg<'_>) -> Result<
 
     let (components, targets) = (opts.components, opts.targets);
     let toolchain = opts.install(cfg)?;
-    if let Some(desc) = &toolchain {
-        let options = DistOptions::new(components, targets, desc, cfg.get_profile()?, true, cfg)?;
-        let status = if Toolchain::exists(cfg, &desc.into())? {
+    if let Some(desc) = toolchain {
+        let options = DistOptions::new(components, targets, &desc, cfg.get_profile()?, true, cfg)?;
+        let status = if Toolchain::exists(cfg, &desc.clone().into())? {
             warn!("Updating existing toolchain, profile choice will be ignored");
             // If we have a partial install we might not be able to read content here. We could:
             // - fail and folk have to delete the partially present toolchain to recover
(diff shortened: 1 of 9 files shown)


src/cli/self_update.rs
```diff
@@ -96,6 +96,63 @@ use windows::{do_add_to_path, do_remove_from_path};
 #[cfg(windows)]
 pub(crate) use windows::{run_update, self_replace};
 
+#[cfg(not(windows))]
+macro_rules! post_install_msg_unix {
+    () => {
```
rami3l: Rust macros follow the buttom-up order so it contradicts with our top-down style guide. Should I move those macros up, or move the renamed method down to its original position?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4950/cc40f2be73e27c88e8aeed74e4918454bc754b41..767e015fc89c61c1907fa62b391297eb6b43a3da)*

rami3l: Finally I decided to move those messages templates in a new file to minimize the diff.

rami3l:
Just to clarify, this change doesn't depend on https://github.com/rust-lang/rustup/pull/4947; instead it's just a cleanup of what is existing in the codebase.

So I think this PR can be merged without breaking anything in terms of existing functionality.

cc @djc Would you mind having a look at this please? 🙏

src/cli/self_update.rs
```diff
@@ -109,7 +109,10 @@ pub(crate) struct InstallOpts<'a> {
 }
 
 impl InstallOpts<'_> {
-    fn install(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
+    /// Prepares rustup's configuration for installation, and determines whether a toolchain should
+    /// be installed based on the user's intent. Returns the toolchain to install, or `None` if none
+    /// should be installed.
```
djc: The first sentence for a documentation comment should fit on one line.

See https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#summary-sentence.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4950/c950062148fd649833012c668f0d6fd2b2ccf727..a74234d5cf0362234f0514a3d116c71274c32125)*

src/cli/self_update.rs
```diff
@@ -109,7 +109,10 @@ pub(crate) struct InstallOpts<'a> {
...
+    fn prepare(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
```
djc: I wonder if this is the best name? It is kind of generic -- if there's a concise way to more precisely describe what the function tries to do, that might be better.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4950/c950062148fd649833012c668f0d6fd2b2ccf727..a74234d5cf0362234f0514a3d116c71274c32125)*

src/cli/self_update.rs
```diff
@@ -109,6 +109,122 @@ pub(crate) struct InstallOpts<'a> {
 }
 
 impl InstallOpts<'_> {
+    /// Installing is a simple matter of copying the running binary to
```
djc: Pre-existing -- also improve the documentation comment here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4950/c950062148fd649833012c668f0d6fd2b2ccf727..a74234d5cf0362234f0514a3d116c71274c32125)*

src/cli/self_update.rs
rami3l: @djc It turns out that the original name was pretty clear, not sure why it has been changed to the current version:

https://github.com/rust-lang/rustup/blob/bbb9276d24519f8684cd616b2e46b7596f7dd891/src/cli/self_update.rs#L886

I've chosen `.select_toolchain()` to be closer to the original. Does that sound okay to you?

djc: Sounds okay.

rami3l:
@djc I see that this patch is still not approved. Do you have any more concerns or can I merge it now? Thanks :)

src/cli/self_update.rs
```diff
@@ -109,7 +109,12 @@ pub(crate) struct InstallOpts<'a> {
 }
 
 impl InstallOpts<'_> {
-    fn install(self, cfg: &mut Cfg<'_>) -> Result<Option<ToolchainDesc>> {
+    /// Selects the toolchain to install based on the user's intent.
+    ///
+    /// This function first resets the default profile and default host tuple in the configuration,
```
djc: What do you mean by "resets"? That seems a strange thing for this function to do, I guess...

djc (approved):
Good stuff!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4950/c950062148fd649833012c668f0d6fd2b2ccf727..64fda73c07a3b500f7bab4ed8c07c73b26691ba1)*

src/cli/self_update.rs
rami3l: @djc It literally does this before returning a value, and it's been like this since 2023:

https://github.com/rust-lang/rustup/blob/bbb9276d24519f8684cd616b2e46b7596f7dd891/src/cli/self_update.rs#L886-L903

There is even a test function asserting this behavior:

https://github.com/rust-lang/rustup/blob/bde2b13605ed8989b6011fae22a1a95788e57640/src/cli/self_update.rs#L1522-L1539

I've changed this to "initialize" from "reset" but should I rather extract that part out of this function?

PS: "Set host triple now as it will affect resolution of toolchain_str" will no longer apply after #4947 but I must keep it here.

djc: I guess it's fine? Just seems like weird design to do that here.

rami3l: @djc Yup I agree...

PR #4951, opened 2026-07-10
refactor: replace `cfg_if!{}` with `cfg_select!{}`

rami3l:
This gets us one fewer external dependency.

There might more `cfg()`s here that can be written this way for better readability, but for now I'm just doing the direct replacements.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2047,7 +2047,6 @@ dependencies = [
  "anstyle",
  "anyhow",
  "cc",
- "cfg-if 1.0.4",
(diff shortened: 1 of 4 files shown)


djc (approved):
Nice!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4951/cc40f2be73e27c88e8aeed74e4918454bc754b41..23aacf6f67409014486afc6fc669452ce6fe4e0a)*

PR #4953, opened 2026-07-11
fix: reject unknown toolchain file keys

### LINKED ISSUE #2876: Unknown keys in rust-toolchain.toml are silently ignored

ychampion:
## Summary

- reject unknown top-level and `[toolchain]` keys instead of silently ignoring them
- document the rustup release that introduced each supported setting

## Why

Misspelled or unsupported keys can otherwise leave rustup using a different toolchain configuration than the file appears to request.

Closes #2876

## Validation

- `cargo test --features=test parse_toml_unknown --lib`
- `cargo test --features=test --test test_bonanza rust_toolchain_toml`
- `cargo clippy --all-targets --features=test -- -D warnings`
- `git diff --check`

### DIFF
--- a/doc/user-guide/src/overrides.md
+++ b/doc/user-guide/src/overrides.md
@@ -114,6 +114,14 @@ issues with dependencies.
 
 ### Toolchain file settings
 
+| Setting | Supported since |
+|---------|-----------------|
+| `channel` | rustup 1.23.0 |
+| `components` | rustup 1.23.0 |
+| `targets` | rustup 1.23.0 |
+| `path` | rustup 1.24.0 |
+| `profile` | rustup 1.24.0 |
+
 #### channel
 
 The `channel` setting specifies which [toolchain] to use. The value is a
--- a/src/config.rs
+++ b/src/config.rs
@@ -40,6 +40,7 @@ enum OverrideFileConfigError {
 }
 
 #[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 struct OverrideFile {
     toolchain: ToolchainSection,
 }
@@ -51,6 +52,7 @@ impl OverrideFile {
 }
 
 #[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 struct ToolchainSection {
     channel: Option<String>,
     path: Option<PathBuf>,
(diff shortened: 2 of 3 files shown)


rami3l:
@ychampion Thanks for working on this!

I'm glad to see that this is more in line with `cargo`'s policy of rejecting `Cargo.toml` containing property keys from a newer schema. We definitely want to inform user of misspelled property keys, and as per your analysis, we have already added some keys in a backwards-compatible fashion.

However I have a design question to ask here: switching to a hard error immediately is a breaking change, and I wonder if this behavior is already (ab)used in the wild. Do you have any evidence in that regard, in that if it has already been (somewhat widely) abused, we may have to issue a warning first before going straight rejecting?

ychampion:
I found enough public use that a warning-first transition looks safer.

At the time of the search, GitHub code search found 13 `rust-toolchain.toml` files with an exact `component =` assignment and 73 with `target =`. I inspected all 13 component hits and the first 20 target hits; they were actual keys under `[toolchain]`, not prose. For example:

- [JasonShin/sqlx-ts](https://github.com/JasonShin/sqlx-ts/blob/f6baee4b003b6b1c38e7acea082862c77a5f1bd2/rust-toolchain.toml) uses both singular keys.
- [jito-labs/memento](https://github.com/jito-labs/memento/blob/f3ed2d4fa56171965902a880b4db279a9a8fc8b0/rust-toolchain.toml) uses `component`.
- [oxidecomputer/opte](https://github.com/oxidecomputer/opte/blob/dfcf0663cfb90ea75e8ee5591b033d68ba6d9fcc/xde/rust-toolchain.toml) uses `target`.

Those settings are currently ignored, so an immediate error would break files whose channel selection still works today. The original `channnel` typo had no indexed hits, and code search is not exhaustive, but the singular-key cases are concrete compatibility evidence. I would emit warnings first, then move to hard errors in a later release.

rami3l:
@ychampion Thanks for the investigation! In that case I suggest leaving this PR as-is or convert it to draft, and instead use another PR to issue a warning first (that should also allow closing the original issue as resolved).

ychampion:
Opened #4969 for the warning-first step. It preserves current parsing, warns once per file/key (including nested paths), and covers the misspelled-only case before the existing error. I'll leave this PR as-is for the later hard-error step.

PR #4954, opened 2026-07-12
fix: repair toolchains missing their installed manifest

### LINKED ISSUE #4825: `rustup toolchain install <toolchain>` reports `unchanged` for toolchains missing their manifest
### Verification

### Problem

If an installed toolchain is missing its manifest, running:

    rustup toolchain install <toolchain>

does not repair the toolchain.

Instead, rustup reports the toolchain as unchanged, and commands that rely on the manifest continue to fail.

This leaves the toolchain in an unusable state without any clear recovery path.

### Steps

Using a dated nightly toolchain to avoid channel updates:

1. Install the toolchain:

    `rustup toolchain install nightly-2026-01-01`

2. Remove the manifest file from the installed toolchain:

    `rm ~/.rustup/toolchains/nightly-2026-01-01-*/lib/rustlib/multirust-channel-manifest.toml`

3. Verify the toolchain is broken:

   ` rustup component list --toolchain nightly-2026-01-01`

   Output:

    `error: Missing manifest in toolchain 'nightly-2026-01-01-x86_64-unknown-linux-gnu'
`
4. Attempt to recover:

    `rustup toolchain install nightly-2026-01-01`

   Output:

    `nightly-2026-01-01-x86_64-unknown-linux-gnu unchanged - rustc ...`

5. Check again:

    `rustup component list --toolchain nightly-2026-01-01`

   The **same** `Missing manifest` error is still reported.

### Expected behavior

`rustup toolchain install <toolchain>` should detect that the existing toolchain is incomplete and reinstall or repair it.

---

### Actual behavior

Rustup reports the toolchain as unchanged and leaves it in a broken state.

---

### Possible Solution(s)

_No response_

### Notes

- This state can occur in practice if installation is interrupted (see #4724).

---

### Context

Discovered while working on improving error messaging for incomplete toolchains.

Related to #4724.

### Rustup version

```console
```

### Installed toolchains

```console

installed toolchains
--------------------
nightly-2026-01-01-x86_64-unknown-linux-gnu

active toolchain
----------------
installed targets:
  x86_64-unknown-linux-gnu
```

### OS version

```console
Pop!_OS 24.04
```

ychampion:
## Summary

- fetch the channel manifest when the installed v2 manifest is missing
- repair damaged toolchains through both explicit install and all-channel update paths

## Why

A matching update hash could skip the channel manifest download even after the installed manifest was deleted, leaving the toolchain broken while rustup reported it as unchanged.

Fixes #4825.

## Validation

- `cargo test --features=test --test test_bonanza suite::cli_rustup::` - 130 passed
- `cargo test --features=test --lib v2_manifest_checksum_mismatch_surfaces_error` - passed
- `cargo clippy --features=test --all-targets -- -D warnings` - passed
- `cargo fmt --all -- --check` - passed

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -29,14 +29,13 @@ use crate::{
         config::Config,
         download::{DownloadCfg, DownloadStatus, File},
         manifest::{Component, CompressionKind, HashedBinary, Manifest},
-        prefix::InstallPrefix,
+        prefix::{DIST_MANIFEST, InstallPrefix},
         temp,
     },
     errors::RustupError,
     utils,
 };
 
-pub(crate) const DIST_MANIFEST: &str = "multirust-channel-manifest.toml";
 pub(crate) const CONFIG_FILE: &str = "multirust-config.toml";
 
 #[derive(Debug)]
@@ -396,8 +395,7 @@ impl Manifestation {
     #[tracing::instrument(level = "trace")]
     pub fn load_manifest(&self) -> Result<Option<Manifest>> {
         let prefix = self.installation.prefix();
-        let old_manifest_path = prefix.manifest_file(DIST_MANIFEST);
-        if utils::path_exists(&old_manifest_path) {
+        if let Some(old_manifest_path) = prefix.dist_manifest() {
             let manifest_str = utils::read_file("installed manifest", &old_manifest_path)?;
             Ok(Some(Manifest::parse(&manifest_str).with_context(|| {
                 RustupError::ParsingFile {
--- a/src/dist/mod.rs
+++ b/src/dist/mod.rs
@@ -1125,10 +1125,9 @@ async fn try_update_from_dist_(
     } else {
         download
             .dl_v2_manifest(
-                // Even if manifest has not changed, we must continue to install requested components.
-                // So if components or targets is not empty, we skip passing `update_hash` so that
-                // we essentially degenerate to `rustup component add` / `rustup target add`
-                if components.is_empty() && targets.is_empty() {
+                // Skip the update hash when the installed manifest is missing or when components
+                // or targets were requested, since either case still requires the channel manifest.
+                if prefix.dist_manifest().is_some() && components.is_empty() && targets.is_empty() {
                     Some(update_hash)
                 } else {
                     None
--- a/src/toolchain/distributable.rs
+++ b/src/toolchain/distributable.rs
@@ -439,12 +439,14 @@ impl<'a> DistributableToolchain<'a> {
     }
 
     pub async fn fetch_dist_manifest(&self) -> anyhow::Result<Option<ManifestWithHash>> {
+        let prefix = InstallPrefix::from(self.toolchain.path());
+        let update_hash = if prefix.dist_manifest().is_some() {
+            Some(self.toolchain.cfg.get_hash_file(&self.desc, false)?)
+        } else {
+            None
+        };
         DownloadCfg::new(self.toolchain.cfg)
-            .dl_v2_manifest(
-                Some(&self.toolchain.cfg.get_hash_file(&self.desc, false)?),
-                &self.desc,
-                self.toolchain.cfg,
-            )
+            .dl_v2_manifest(update_hash.as_deref(), &self.desc, self.toolchain.cfg)
             .await
     }
 
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -4201,14 +4201,7 @@ async fn missing_manifest_shows_reinstall_help() {
         .await
         .is_ok();
 
-    let manifest_path = cx
-        .config
-        .rustupdir
-        .join("toolchains")
-        .join(format!("nightly-{}", this_host_tuple()))
-        .join("lib")
-        .join("rustlib")
-        .join("multirust-channel-manifest.toml");
+    let manifest_path = nightly_manifest_path(&cx);
 
     fs::remove_file(&manifest_path).unwrap();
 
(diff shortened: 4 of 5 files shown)


src/toolchain/distributable.rs
```diff
@@ -439,9 +440,12 @@ impl<'a> DistributableToolchain<'a> {
     }
 
     pub async fn fetch_dist_manifest(&self) -> anyhow::Result<Option<ManifestWithHash>> {
+        let update_hash = self.toolchain.cfg.get_hash_file(&self.desc, false)?;
+        let prefix = InstallPrefix::from(self.toolchain.path());
         DownloadCfg::new(self.toolchain.cfg)
             .dl_v2_manifest(
-                Some(&self.toolchain.cfg.get_hash_file(&self.desc, false)?),
+                utils::path_exists(prefix.manifest_file(DIST_MANIFEST))
+                    .then_some(update_hash.as_path()),
```
rami3l: Nit: The evaluation of `update_hash` happens too soon.

```suggestion
        let prefix = InstallPrefix::from(self.toolchain.path());
DownloadCfg::new(self.toolchain.cfg)
.dl_v2_manifest(
                if prefix.has_manifest() { Some(&self.toolchain.cfg.get_hash_file(&self.desc, false)?) } else { None },
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4954/4add58e5a9bf09822b821d4918f5eac5ff029a75..46a54f07598852221ddbcb5dfde61be032d8b94f)*

src/dist/mod.rs
```diff
@@ -1125,10 +1127,9 @@ async fn try_update_from_dist_(
     } else {
         download
             .dl_v2_manifest(
-                // Even if manifest has not changed, we must continue to install requested components.
-                // So if components or targets is not empty, we skip passing `update_hash` so that
-                // we essentially degenerate to `rustup component add` / `rustup target add`
-                if components.is_empty() && targets.is_empty() {
+                // Skip the update hash when the installed manifest is missing or when components
+                // or targets were requested, since either case still requires the channel manifest.
+                if has_installed_v2_manifest && components.is_empty() && targets.is_empty() {
```
rami3l:
```suggestion
                if prefix.has_manifest() && components.is_empty() && targets.is_empty() {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4954/4add58e5a9bf09822b821d4918f5eac5ff029a75..46a54f07598852221ddbcb5dfde61be032d8b94f)*

src/dist/mod.rs
```diff
@@ -1114,6 +1115,7 @@ async fn try_update_from_dist_(
     fetched: &mut String,
     cfg: &Cfg<'_>,
     prefetched_manifest: Option<ManifestWithHash>,
+    has_installed_v2_manifest: bool,
```
rami3l: Please don't pass this as a parameter, instead introduce a first commit in this PR where you extract a method `prefix.has_manifest()` that replaces the first half of `Manifestation::load_manifest()`.

PS: This variable name also makes no sense, because this manifest could be of v1 or v2, they are located at the same place.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4954/4add58e5a9bf09822b821d4918f5eac5ff029a75..46a54f07598852221ddbcb5dfde61be032d8b94f)*

rami3l (changes requested):

src/toolchain/distributable.rs
ychampion: Moved the hash lookup after the prefix check, so it is only evaluated when an installed manifest exists.

src/dist/mod.rs
ychampion: Split this into two commits. The first adds `InstallPrefix::has_manifest()` and uses it in `load_manifest`; the repair now uses that helper without passing a version-specific flag.

src/dist/mod.rs
ychampion: Updated this guard to use `prefix.has_manifest()`.

src/dist/manifestation.rs
```diff
@@ -397,7 +396,7 @@ impl Manifestation {
     pub fn load_manifest(&self) -> Result<Option<Manifest>> {
         let prefix = self.installation.prefix();
         let old_manifest_path = prefix.manifest_file(DIST_MANIFEST);
-        if utils::path_exists(&old_manifest_path) {
+        if prefix.has_manifest() {
```
rami3l: My apologies 🙏 

Judging from the access pattern here it seems that it'd be better to change `.has_manifest()` to `.dist_manifest()` which returns `Option<PathBuf>`. This way we will be able to use:

```suggestion
        if let Some(old_manifest_path) = prefix.dist_manifest() {
```

... and that it'd be clearer that the manifest path is actually not used in the sad path.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4954/4add58e5a9bf09822b821d4918f5eac5ff029a75..3a17c151bc602bb482f5f440c7aa1e3e5b2c5b81)*

src/dist/mod.rs
```diff
@@ -1125,10 +1125,9 @@ async fn try_update_from_dist_(
     } else {
         download
             .dl_v2_manifest(
-                // Even if manifest has not changed, we must continue to install requested components.
-                // So if components or targets is not empty, we skip passing `update_hash` so that
-                // we essentially degenerate to `rustup component add` / `rustup target add`
-                if components.is_empty() && targets.is_empty() {
+                // Skip the update hash when the installed manifest is missing or when components
+                // or targets were requested, since either case still requires the channel manifest.
+                if prefix.has_manifest() && components.is_empty() && targets.is_empty() {
```
rami3l: Following https://github.com/rust-lang/rustup/pull/4954/changes#r3567067397, this should be changed to:


```suggestion
                if prefix.dist_manifest().is_some() && components.is_empty() && targets.is_empty() {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4954/4add58e5a9bf09822b821d4918f5eac5ff029a75..3a17c151bc602bb482f5f440c7aa1e3e5b2c5b81)*

tests/suite/cli_rustup.rs
```diff
@@ -4192,6 +4192,16 @@ installed targets:
         .is_ok();
 }
 
+fn nightly_manifest_path(cx: &CliTestContext) -> PathBuf {
```
djc: Nit: callers before callees. See the style guide:

https://rust-lang.github.io/rustup/dev-guide/coding-standards.html#coding-style

djc (commented):
Please drop the slop from the commit messages.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4954/4add58e5a9bf09822b821d4918f5eac5ff029a75..3a17c151bc602bb482f5f440c7aa1e3e5b2c5b81)*

ychampion:
Ok looking

ychampion:
Done. I rewrote both commit messages as concise subjects and verified that the resulting tree is unchanged.

djc:
> Done. I rewrote both commit messages as concise subjects and verified that the resulting tree is unchanged.

Did you miss the earlier feedback from @rami3l?

src/dist/manifestation.rs
ychampion: Done. `dist_manifest()` now returns `Option<PathBuf>`, and `load_manifest()` uses the returned path directly.

src/dist/mod.rs
ychampion: Updated this call site to use `prefix.dist_manifest().is_some()`.

tests/suite/cli_rustup.rs
ychampion: Moved the helper below its callers to match the style guide.

ychampion:
> > Done. I rewrote both commit messages as concise subjects and verified that the resulting tree is unchanged.
> 
> 
> 
> Did you miss the earlier feedback from @rami3l?

Yeah maybe, could you reference it?

djc:
> > > Done. I rewrote both commit messages as concise subjects and verified that the resulting tree is unchanged.
> > 
> > 
> > Did you miss the earlier feedback from @rami3l?
> 
> Yeah maybe, could you reference it?

I meant the commit you added, but the new commit should be squashed into the first one.

ychampion:
Done. I folded the follow-up into the first commit, so the branch is back to two commits.

rami3l (approved):
Thank you :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4954/434858b3e6ceb66b776527ed57dd70cf9cd5e64f..edf566440f0daf1aa976569396149dc0c4141daa)*

PR #4956, opened 2026-07-13
Reject unknown rust-toolchain.toml keys

### LINKED ISSUE #2876: Unknown keys in rust-toolchain.toml are silently ignored

Bortlesboat:
Reject unknown keys at both levels of `rust-toolchain.toml` deserialization.

This prevents misspellings such as `channnel = "nightly"` from being silently ignored and also rejects unsupported top-level sections. The TOML parser reports the unknown key and lists the supported fields, while valid and legacy toolchain files retain their existing behavior.

Fixes #2876.

Validation:

- RED/GREEN tests for an unknown `[toolchain]` key and an unknown top-level section
- all 13 `config::tests` parser tests pass with the `test` feature
- `cargo clippy --all-targets --features test -- -D warnings`
- `git diff --check`

I also attempted the broader `--all-features` Clippy command. On this Windows host its optional vendored-OpenSSL lane requires a `perl` executable and stopped while configuring OpenSSL; the repository's normal Windows feature set above is green.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -40,6 +40,7 @@ enum OverrideFileConfigError {
 }
 
 #[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 struct OverrideFile {
     toolchain: ToolchainSection,
 }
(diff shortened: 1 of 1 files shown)


djc:
There already was a PR for this before you claimed it:

- https://github.com/rust-lang/rustup/pull/4953

PR #4958, opened 2026-07-13
ci(dist): ensure pushing to `dev-static` on `stable` update

rami3l:
Part of https://github.com/rust-lang/rustup/issues/4738.

During our release process, @Mark-Simulacrum [wasn't able to find](https://rust-lang.zulipchat.com/#narrow/channel/241545-t-release/topic/Rustup.201.2E29.2E1.20-.20testing.2Frelease/near/609737359) the required `dev-static` artifacts from the commit 69d75452157eed4f71ac1dab1cf2ad5a2f4de679 which is probably caused by our new release process disengaging `stable` from `main`.

This looks like a pre-existing bug, since the branch used for artifact uploading is semantically incorrect. However, this issue was unfortunately never exposed before since it used to be that `stable` was tracking `main`, thus with every `stable` we would want to release, the artifacts have been uploaded by a previous CI step on `main`.

This fix ensures that the `dev-static` uploads happen when `stable` is updated ~~instead of~~ in addition to `main`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -202,13 +202,13 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-west-1
       - name: Configure AWS credentials
-        if: github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.mode == 'release'
+        if: github.event_name == 'push' && matrix.mode == 'release' && (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/main')
         uses: aws-actions/configure-aws-credentials@v6
         with:
           role-to-assume: arn:aws:iam::890664054962:role/ci--rust-lang--rustup
           aws-region: us-east-1
       - name: Deploy build to rustup-builds bucket for release team
-        if: github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.mode == 'release'
+        if: github.event_name == 'push' && matrix.mode == 'release' && (github.ref == 'refs/heads/stable' || github.ref == 'refs/heads/main')
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -202,13 +202,13 @@ jobs:
...
       - name: Configure AWS credentials
-        if: github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.mode == 'release'
+        if: github.event_name == 'push' && github.ref == 'refs/heads/stable' && matrix.mode == 'release'
         uses: aws-actions/configure-aws-credentials@v6
         with:
           role-to-assume: arn:aws:iam::890664054962:role/ci--rust-lang--rustup
           aws-region: us-east-1
       - name: Deploy build to rustup-builds bucket for release team
-        if: github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.mode == 'release'
+        if: github.event_name == 'push' && github.ref == 'refs/heads/stable' && matrix.mode == 'release'
```
Mark-Simulacrum: I think we want the condition here to be an OR. You can see how Rust's CI does this here: https://github.com/rust-lang/rust/blob/5503df87342a73d0c29126a7e08dc9c1255c46ad/.github/workflows/ci.yml#L82 -- note that all branches with approved commits upload.

The goal is that we have artifacts for every merged PR, which means that we can build e.g. a dev-static nightly at any time, bisect, etc. I think I'd want to achieve the same shape for rustup.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4958/4add58e5a9bf09822b821d4918f5eac5ff029a75..2f833d1a5788ef9b97ee604f9768cea48b909a87)*

rami3l: @Mark-Simulacrum Thanks for the review! I think this predicate already exists in previous steps so I've copied it over.

djc (approved):

Mark-Simulacrum (approved):

PR #4962, opened 2026-07-15
Add `--default` flag in `rustup toolchain install`

FranciscoTGouveia:
Based on [#t-rustup > Reducing the need for setup-rust actions @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/490103-t-rustup/topic/Reducing.20the.20need.20for.20setup-rust.20actions/near/610697184).

Feedback on the wording of the doc comment for the new flag is welcome.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1059,7 +1059,6 @@ async fn update(
     let self_update_mode = SelfUpdateMode::from_cfg(cfg)?;
     let should_self_update = !opts.no_self_update;
     let force_non_host = opts.force_non_host;
-    let set_override = opts.r#override;
     cfg.profile_override = opts.profile;
 
     let cfg = &cfg;
@@ -1118,7 +1117,7 @@ async fn update(
                 Ok(status.clone()),
             )?;
 
-            if set_override {
+            if opts.r#override {
                 cfg.make_override(&cfg.current_dir, &desc.clone().into())?;
             }
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -468,6 +468,10 @@ struct UpdateOpts {
     /// Set the installed toolchain as the override for the current directory
     #[arg(long)]
     r#override: bool,
+
+    /// Set the installed toolchain as the default toolchain
+    #[arg(long)]
+    default: bool,
 }
 
 #[derive(Debug, Default, Args)]
@@ -1121,7 +1125,9 @@ async fn update(
                 cfg.make_override(&cfg.current_dir, &desc.clone().into())?;
             }
 
-            if cfg.get_default()?.is_none() && matches!(status, UpdateStatus::Installed) {
+            if opts.default
+                || (cfg.get_default()?.is_none() && matches!(status, UpdateStatus::Installed))
+            {
                 cfg.set_default(Some(&desc.into()))?;
             }
         }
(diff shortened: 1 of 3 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1060,6 +1064,7 @@ async fn update(
     let should_self_update = !opts.no_self_update;
     let force_non_host = opts.force_non_host;
     let set_override = opts.r#override;
+    let set_default = opts.default;
```
djc: Nit: if not for the prevailing style, do we actually need this or can we just inline `opts.set_default` in the condition below?

djc (approved):
Should this get some tests?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4962/fadde36bb081a159e9c37fcef69746caaf64eae6..a9aa51621a3ee7bc2413e47baa70946ce59197c5)*

src/cli/rustup_mode.rs
FranciscoTGouveia: The idea was to follow the current style, yes.

I've just inlined it, and I've also inlined the above `set_override` in a different commit (which had only one use).

FranciscoTGouveia:
> Should this get some tests?

I would argue that this change is somewhat trivial, but I wouldn't oppose it.
Should I add a new test?

src/cli/rustup_mode.rs
```diff
@@ -468,6 +468,10 @@ struct UpdateOpts {
     /// Set the installed toolchain as the override for the current directory
     #[arg(long)]
     r#override: bool,
+
+    /// Set the installed toolchain as the default
```
ChrisDenton: Considering the command is `rustup set default-host` maybe it would be best to use the term "default host" so the connection is clear:
```suggestion
    /// Set the installed toolchain as the default host
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4962/0fbd6f8b1b200236aaf20e579d25f776b20bafc7..e40e49f4ce8bd773fa25cd053bf019ca701fec97)*

FranciscoTGouveia: Agreed.
Can the flag be kept as is, or should it also be changed to `--default-host` instead of the current `--default`?

ChrisDenton: I think keeping `--default` is fine in this context and being consistent with `--override` makes sense to me.

rami3l: @ChrisDenton @FranciscoTGouveia Default host (a target tuple) and default toolchain (a toolchain) are not the same thing and we are referring to the latter here, I believe :)

rami3l (approved):
LGTM modulo the one remaining concern, thanks :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4962/0fbd6f8b1b200236aaf20e579d25f776b20bafc7..87dc74bd21a15ba3de48e4e1c5cbbf5eab135d19)*

src/cli/rustup_mode.rs
ChrisDenton: In that case `host` -> `toolchain`

rami3l:
> > Should this get some tests?
> 
> I would argue that this change is somewhat trivial, but I wouldn't oppose it. Should I add a new test?

@FranciscoTGouveia Looking back I'd second @djc that it would be useful to add a tiny test for it; probably also for `--override` if that hasn't been done previously.

djc (approved):

Issue #4400, opened 2025-07-03
Interrupted transaction could get rustup stuck in loop


ShaunSHamilton:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue+is%3Aopen%2Cclosed and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

`rustup`, `cargo`, and `rustc` are unusable after clean install.

### Steps

1. Install 1.28.2
2. Run any command from toolchain (e.g. `rustup update`, `rustc --version`, `cargo check`)
3. See error

### Notes

Clean install was done, because of https://github.com/rust-lang/rustup/issues/4213

<details>
   <summary>Terminal</summary>

```powershell
PS C:\Users\shauh\exam-creator> rustup toolchain install nightly-aarch64-pc-windows-msvc
info: syncing channel updates for 'nightly-aarch64-pc-windows-msvc'
  1.0 MiB /   1.0 MiB (100 %)  75.2 KiB/s in 16s
info: latest update on 2025-07-03, rust version 1.90.0-nightly (667787527 2025-07-02)
info: downloading component 'cargo'
  4.0 MiB /   8.7 MiB ( 46 %) 400.0 KiB/s in 30s ETA: 12s
info: retrying download for 'https://static.rust-lang.org/dist/2025-07-03/cargo-nightly-aarch64-pc-windows-msvc.tar.xz'
  5.6 MiB /   8.7 MiB ( 65 %)  91.8 KiB/s in 30s ETA: 34s
info: retrying download for 'https://static.rust-lang.org/dist/2025-07-03/cargo-nightly-aarch64-pc-windows-msvc.tar.xz'
351.8 KiB /   8.7 MiB (  4 %) 143.8 KiB/s in  4s ETA: 59s
PS C:\Users\shauh\exam-creator> rustup update
info: syncing channel updates for 'stable-aarch64-pc-windows-msvc'
840.6 KiB / 855.0 KiB ( 98 %)  28.8 KiB/s in 30s
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml' to 'C:\Users\shauh\.rustup\tmp\1xi17qo53_6z50tk_file.toml'
info: checking for self-update
info: downloading self-update
[...]

PR #4963, opened 2026-07-16
test: add deterministic crash checkpoints for installs

cachebag:
Under the guidance of @rami3l, this is the first unit of work toward #4400, with #4724 as a related failure mode.

Ref: [toolchain pool proposal](https://hackmd.io/@cachebag/Byd2SENNMg)

The plan is to establish deterministic infrastructure for testing hard interruptions without relying on debugger automation. This PR adds a test-only, environment-selected checkpoint mechanism. A child rustup process signals that it has reached the selected checkpoint and blocks; the test driver then kills and reaps it before retrying the operation in a fresh process.

This was my solution as opposed to something like a a debugger automation..but if it proves to be weak we can go that route instead. Would love feedback on this part.

The initial test interrupts installation after component payloads are installed but before toolchain metadata is written, then verifies that retrying converges to a working toolchain. This PR intentionally adds test infrastructure rather than changing recovery behavior. 

(Additional named interruption points can be added as needed without changing the parent-process test harness, I figured a smaller scoped PR was appropriate first)

Follow-up work, generally speaking, will be as follows:

- Add checkpoints around additional filesystem mutations, metadata writes, and download failures.
- Reproduce and characterize the interrupted states from #4400 and #4724.
- Add staging and atomic publication for fresh toolchain installations.
- Extend the model to updates of existing toolchains and concurrent writers.
- Address reference lifetime, abandoned-object cleanup, and eventual garbage collection separately.

### DIFF
--- a/src/dist/manifestation.rs
+++ b/src/dist/manifestation.rs
@@ -21,6 +21,8 @@ use futures_util::{
 use tokio::task::{JoinHandle, spawn_blocking};
 use tracing::{debug, info, warn};
 
+#[cfg(feature = "test")]
+use crate::test::checkpoint;
 use crate::{
     diskio::{Executor, IO_CHUNK_SIZE, get_executor, unpack_ram},
     dist::{
@@ -290,6 +292,9 @@ impl Manifestation {
             download_cfg.clean(&stream.cleanup_downloads)?;
             drop(stream);
         }
+
+        #[cfg(feature = "test")]
+        checkpoint(download_cfg.process, "manifestation-update-before-metadata");
 
         // Install new distribution manifest
         let new_manifest_str = new_manifest.clone().stringify()?;
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -12,10 +12,11 @@ use std::{
     mem,
     ops::{Deref, DerefMut},
     path::{Path, PathBuf},
-    process::Command,
+    process::{Command, ExitStatus},
     string::FromUtf8Error,
     sync::{Arc, LazyLock, RwLock, RwLockWriteGuard},
-    time::{Instant, SystemTime, UNIX_EPOCH},
+    thread,
+    time::{Duration as StdDuration, Instant, SystemTime, UNIX_EPOCH},
 };
 
 use chrono::{DateTime, Duration};
@@ -33,7 +34,7 @@ use crate::test::this_host_tuple;
 use crate::utils;
 
 use super::{
-    CROSS_ARCH1, CROSS_ARCH2, MULTI_ARCH1,
+    CHECKPOINT_ENV, CROSS_ARCH1, CROSS_ARCH2, MULTI_ARCH1, checkpoint_path,
     dist::{MockDistServer, MockManifestVersion, Release, RlsStatus, change_channel_date},
     mock::MockFile,
 };
@@ -487,7 +488,7 @@ impl Config {
                         && e.raw_os_error() == Some(26)
                     {
                         // This is an ETXTBSY situation
-                        std::thread::sleep(std::time::Duration::from_millis(250));
+                        thread::sleep(StdDuration::from_millis(250));
                     } else {
                         panic!("Unable to run test command: {e:?}");
                     }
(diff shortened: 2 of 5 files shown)


src/test/checkpoint.rs
```diff
@@ -0,0 +1,24 @@
...
+        .expect("selected test checkpoint requires a marker file");
+    fs::write(&marker, name).expect("failed to write test checkpoint marker");
+
+    loop {
+        thread::park();
+    }
+}
+
+pub(super) const CHECKPOINT_ENV: &str = "RUSTUP_TEST_CHECKPOINT";
+pub(super) const CHECKPOINT_FILE_ENV: &str = "RUSTUP_TEST_CHECKPOINT_FILE";
```
rami3l: I think we'd be more careful when adding more env vars.

`RUSTUP_TEST_CHECKPOINT=manifestation-update-before-metadata` might be just enough to derive the checkpoint file's path from `self.config.test_root_dir`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4963/1bda00775d8b9f13537e89732fc02251e77af098..c6d83d76014a4d71271a5c4944cfd331992a1e99)*

cachebag: The current test setup does place `RUSTUP_HOME` directly under `test_root_dir`, so the child could derive the marker root from `RUSTUP_HOME.parent()`...

My only concern is that this makes the checkpoint mechanism depend on that layout and would not work for commands that override `RUSTUP_HOME`; I thought that maybe the separate variable keeps the control path explicit and not so tightly coupled.

rami3l: > The current test setup does place `RUSTUP_HOME` directly under `test_root_dir`, so the child could derive the marker root from `RUSTUP_HOME.parent()`...
> 
> My only concern is that this makes the checkpoint mechanism depend on that layout and would not work for commands that override `RUSTUP_HOME`; I thought that maybe the separate variable keeps the control path explicit and not so tightly coupled.

@cachebag I think `test_root_dir` currently relies on `test_dir()` so that's already a tempdir, which should mean that we can reuse it.

src/dist/manifestation.rs
```diff
@@ -291,6 +291,9 @@ impl Manifestation {
             drop(stream);
         }
 
+        #[cfg(feature = "test")]
+        crate::test::checkpoint::wait(download_cfg.process, "manifestation-update-before-metadata");
```
djc: Nit: suggest importing `checkpoint` at the top of the module.

src/test/checkpoint.rs
```diff
@@ -0,0 +1,24 @@
+use std::{fs, path::PathBuf, thread};
```
djc: IMO adding such small modules makes the source tree harder to navigate.

src/test/clitools.rs
```diff
@@ -972,6 +973,48 @@ impl CliTestContext {
...
+                .expect("failed to query checkpoint test command")
+            {
+                panic!("command exited before reaching checkpoint {checkpoint:?}: {status}");
+            }
+            if Instant::now() >= deadline {
+                let _ = child.kill();
+                let _ = child.wait();
+                panic!("timed out waiting for checkpoint {checkpoint:?}");
+            }
+            std::thread::sleep(StdDuration::from_millis(10));
```
djc: Nit: import `thread`?

djc (approved):
This looks okay to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4963/1bda00775d8b9f13537e89732fc02251e77af098..c6d83d76014a4d71271a5c4944cfd331992a1e99)*

rami3l (approved):
LGTM! Make sure that you have addressed all remaining concerns and I think we can merge this.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4963/1bda00775d8b9f13537e89732fc02251e77af098..c7f9e73401ddfa3668f64bc07ecf6b585183402d)*

cachebag:
Everything should be addressed? Unless I am mistaken.

rami3l:
> Everything should be addressed? Unless I am mistaken. 

@cachebag Sorry for the confusion, I was referring to the remaining open threads but since you have pushed a fix, please feel free to resolve them directly from your side :)

Issue #4663, opened 2025-12-30
Deduplicate toolchains, components and/or artifacts


rami3l:
### Problem you are trying to solve

It happens from time to time that the user gets confused about why `rustup toolchain install stable`, `rustup toolchain install 1.92` and `rustup toolchain install 1.92.0` require separate and duplicate/redundant downloads when they are essentially the same thing (given that the current stable is `1.92.0`).

Such reports include:
- #555
- https://github.com/rust-lang/rustup/issues/1974
- https://github.com/rust-lang/rustup/issues/1366
- https://github.com/rust-lang/rustup/issues/2617

The status quo is very well explained in the quote below:

> Firstly, 1.70 and 1.70.0 are not the same toolchain. They might be on a given date, but once 1.70.1 releases, they are not.
> 
> Secondly, the installed components are currently defined as separate for each toolchain; adding a component that isn't expected might be strange and conceivable a bug for some developers - and turning minor versions into symlinks to concrete versions will have that effect, as you would no longer have separate component lists.
_https://github.com/rust-lang/rustup/issues/1366#issuecomment-1633065070_

In short, `1.92.0` is a "fixed target" that will always be pointing to that very version of the toolchain, but the other two are "moving target"s: `1.92` will be upgradable to any later `1.92.*`, and `stable` to any later stable. These toolchains are logically separated and I have no intentions to change that, but physically this separation is not an absolute must.

### Solution you'd like

A potential solution is to borrow from [PNPM](https://pnpm.io/faq#does-it-work-on-windows) and use hard links or junctions that reference a user-global artifact pool. This might have some interesting interactions with multiple proposals, e.g.
- #3937 
- #731 
- #4636

Another choice could be providing a user-global component (or even toolchain) pool instead of an artifact pool, depending on the flexibility we'd like to achieve.

I personally think choosing the toolchain pool is a bad idea since it turns out sometimes multiple toolchains share the same component esp. when they are not updated.

daniel-pfeiffer:
I have also been annoyed that a project with a `rust-toolchain.toml` reinstalled the same one under a different name. It’s a waste of network bandwidth (which some people pay per volume,) download time and disk space!

[...]

PR #4965, opened 2026-07-20
feat(dist): stage fresh toolchains before atomic publication

cachebag:
This PR implements the second step of the plan discussed in #4663 (details in my [design notes](https://hackmd.io/H9-IA2oFTgmcjVbreOFSuw)), addressing #4400 and #4724 for fresh installs.

The idea is that a fresh toolchain now builds in its own "staging" directory and is then published to `toolchains/<name>` with a same-filesystem rename, with the update hash only being written after publication. In the case of an interrupted install, this means no partial toolchain is left behind and a retry always works. The stage itself lives at a deterministic per-toolchain path guarded by an OS-released file lock, so an abandoned stage gets reclaimed by the next install instead of accumulating, and a concurrent install of the same toolchain fails fast instead of racing.

Updates still modify toolchains in place, and blocking instead of failing is deferred to the locking work in #988.

cc @rami3l

### DIFF
--- a/src/install.rs
+++ b/src/install.rs
@@ -1,9 +1,13 @@
 //! Installation and upgrade of both distribution-managed and local
 //! toolchains
-use std::path::{Path, PathBuf};
+use std::{
+    ffi::OsString,
+    fs,
+    path::{Path, PathBuf},
+};
 
-use anyhow::Result;
-use tracing::debug;
+use anyhow::{Context, Result, bail};
+use tracing::{debug, warn};
 
 use crate::{
     config::Cfg,
@@ -53,8 +57,9 @@ impl InstallMethod<'_, '_> {
             _ => debug!("updating existing install for '{}'", self.dest_basename()),
         }
 
-        debug!("toolchain directory: {}", self.dest_path().display());
-        let updated = self.run(&self.dest_path(), manifest).await?;
+        let destination = self.dest_path();
+        debug!("toolchain directory: {}", destination.display());
+        let updated = self.run(&destination, manifest).await?;
 
         let status = match updated {
             false => {
@@ -102,16 +107,50 @@ impl InstallMethod<'_, '_> {
                 utils::symlink_dir(src, path)?;
                 Ok(true)
             }
-            InstallMethod::Dist(opts) => {
+            // Updates modify the published toolchain in place.
+            InstallMethod::Dist(opts) if opts.exists => {
                 let prefix = &InstallPrefix::from(path.to_owned());
-                let maybe_new_hash = opts.install_into(prefix, manifest).await?;
-
-                if let Some(hash) = maybe_new_hash {
-                    utils::write_file("update hash", &opts.update_hash, &hash)?;
-                    Ok(true)
-                } else {
-                    Ok(false)
+                match opts
+                    .install_into(prefix, &opts.update_hash, manifest)
+                    .await?
+                {
+                    Some(hash) => {
+                        utils::write_file("update hash", &opts.update_hash, &hash)?;
+                        Ok(true)
+                    }
+                    None => Ok(false),
                 }
+            }
+            // A fresh toolchain is constructed under a private stage and only
+            // published once complete, so that an interruption can never
+            // leave a partial toolchain behind at a selectable path.
+            InstallMethod::Dist(opts) => {
+                let staging = StagedToolchain::new(path)?;
+                let prefix = InstallPrefix::from(staging.prefix.to_owned());
+                // A stray alias-scoped update hash (e.g. left behind by an
+                // interrupted uninstall) is deliberately ignored rather than
+                // removed: the staged install only ever reads the
+                // staging-local hash, and the alias-scoped one is rewritten
+                // after publication.
+                let staging_hash = staging.update_hash();
+                let Some(hash) = opts.install_into(&prefix, &staging_hash, manifest).await? else {
+                    return Ok(false);
+                };
+
+                #[cfg(feature = "test")]
+                opts.cfg
+                    .process
+                    .checkpoint(CHECKPOINT_INSTALL_BEFORE_PUBLISH);
+
+                staging.publish(path)?;
+
+                #[cfg(feature = "test")]
+                opts.cfg
+                    .process
+                    .checkpoint(CHECKPOINT_INSTALL_AFTER_PUBLISH);
+
+                utils::write_file("update hash", &opts.update_hash, &hash)?;
+                Ok(true)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


tests/suite/cli_crash.rs
```diff
@@ -1,19 +1,178 @@
...
+    let command = cx.config.cmd("rustup", ["update", "nightly"]);
+    let status = cx.kill_at_checkpoint(command, BEFORE_METADATA);
+    assert!(!status.success());
+
+    assert_completes_successfully(cx.config.cmd("rustup", ["update", "nightly"]));
+
+    assert_nightly_is_complete(&cx).await;
 }
+
+const BEFORE_METADATA: &str = "manifestation-update-before-metadata";
```
rami3l: Nit: It looks like those constants are not fully reused. Try moving them to the specific module instead of just under `./tests` (there is an existing example already).

src/install.rs
```diff
@@ -13,6 +17,61 @@ use crate::{
     utils,
 };
 
+#[cfg(feature = "test")]
+use crate::test::checkpoint;
+
+impl StagedToolchain {
```
rami3l, line 30: Nit: Logically `impl` blocks should follow the type declaration.

src/install.rs
```diff
@@ -116,6 +183,43 @@ impl InstallMethod<'_, '_> {
         }
     }
 
+    async fn run_staged_dist(
```
rami3l: Maybe inline this into `.run()`?

tests/suite/cli_crash.rs
```diff
@@ -163,16 +213,13 @@ async fn interrupted_update_can_be_retried() {
...
+        ["rustup", "update", "nightly"],
+    );
     assert!(!status.success());
 
     assert_completes_successfully(cx.config.cmd("rustup", ["update", "nightly"]));
 
     assert_nightly_is_complete(&cx).await;
 }
-
-const BEFORE_METADATA: &str = "manifestation-update-before-metadata";
```
rami3l: Prevent back-and-forth in a single PR; reorganize the history so that each new change are as close to the place at the end of this PR as possible.

src/install.rs
```diff
@@ -13,6 +16,102 @@ use crate::{
...
 
+impl StagedToolchain {
+    fn new(destination: &Path) -> Result<Self> {
+        let parent = destination
+            .parent()
+            .expect("toolchain destination must have a parent");
+        let name = destination
+            .file_name()
+            .and_then(|name| name.to_str())
+            .expect("toolchain destination must have a UTF-8 base name");
```
rami3l: I think here `name` being a path segment can safely stay as `OsStr[ing]`, there seems to be no need to convert it to a native string?.

rami3l (commented):
Thanks!

I have a question though: If I understand it correctly, we are performing the final stage of the installation in an isolated path under `toolchain/` in this case (so no indirection WRT the actual toolchain), how can this evolve into the A/B partitioning scheme as we have previously discussed?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4965/ac3ea6c48573b8cd1ee1ba0cd91669ec2608f2dd..5553723d464814c1c15ea1df47ee93c788296e27)*

cachebag:
> If I understand it correctly, we are performing the final stage of the installation in an isolated path under toolchain/ in this case (so no indirection WRT the actual toolchain), how can this evolve into the A/B partitioning scheme as we have previously discussed?

@rami3l 
For a fresh install there is no "A" side to keep alive, so atomically renaming the finished directory to `toolchains/<name>` is observably the same as flipping a reference there: you go from nothing to a complete toolchain in one step. This is the special case you pointed out in #4724.

IMO, this PR builds the half of A/B that will be invariable for us in the near future: the complete "B" side is constructed in an isolated stage, alias-scoped metadata isn't touched until the commit point, there is exactly one atomic publication operation, and the stage is owned through a crash-released lock so an abandoned B side gets reclaimed. The only thing that changes when we introduce real A/B is the body of `StagedToolchain::publish()`: instead of `rename(stage -> toolchains/<name>)` it becomes `rename(stage -> heap/<object>)` followed by an atomic reference flip at `toolchains/<name>`. 

Does that make sense?

rami3l:
> IMO, this PR builds the half of A/B that will be invariable for us in the near future: the complete "B" side is constructed in an isolated stage, alias-scoped metadata isn't touched until the commit point, there is exactly one atomic publication operation, and the stage is owned through a crash-released lock so an abandoned B side gets reclaimed. The only thing that changes when we introduce real A/B is the body of `StagedToolchain::publish()`: instead of `rename(stage -> toolchains/<name>)` it becomes `rename(stage -> heap/<object>)` followed by an atomic reference flip at `toolchains/<name>`.
> 
> Does that make sense?

@cachebag Yup, that makes perfect sense, thanks :)

I largely agree with your direction here. I'll just need to evaluate whether it fits into the overall plan to see if we can finally merge this, many thanks in advance for your patience 🙏

src/install.rs
```diff
@@ -13,6 +17,114 @@ use crate::{
     utils,
 };
 
+/// A fresh toolchain being constructed in a private staging directory,
+/// invisible to toolchain enumeration until it is published.
```
djc: Nit: docstrings should start with a single-line sentence, as explained in the style guide (as part of the dev guide).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4965/46035eb95a3db74ed0ac6d065be749a363923a9d..18db95f4ca5226cea75bcb4eb94a3e25f161a908)*

src/install.rs
```diff
@@ -13,6 +17,114 @@ use crate::{
...
+struct StagedToolchain {
+    root: PathBuf,
+    prefix: PathBuf,
+    /// Exclusive lock marking the stage as owned by a live process; `None`
+    /// only while dropping.
```
djc: Nit: same here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4965/46035eb95a3db74ed0ac6d065be749a363923a9d..18db95f4ca5226cea75bcb4eb94a3e25f161a908)*

src/install.rs
```diff
@@ -13,6 +17,114 @@ use crate::{
     utils,
 };
 
+/// A fresh toolchain being constructed in a private staging directory,
+/// invisible to toolchain enumeration until it is published.
+struct StagedToolchain {
```
djc: Per the style guide, this type should appear below its users.

Please read the style guide again and try to keep it in mind/self-review.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4965/46035eb95a3db74ed0ac6d065be749a363923a9d..18db95f4ca5226cea75bcb4eb94a3e25f161a908)*

src/install.rs
```diff
@@ -13,6 +17,114 @@ use crate::{
...
+    fn prefix(&self) -> &Path {
+        &self.prefix
+    }
+
+    /// Path of the update hash tracked next to the stage. The hash under
+    /// `$RUSTUP_HOME/update-hashes` is alias-scoped metadata and must not be
+    /// touched before the staged toolchain is published.
+    fn update_hash(&self) -> PathBuf {
+        self.root.join(STAGING_HASH_FILE)
+    }
```
djc: Nit: this small helper function should appear below `publish()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4965/46035eb95a3db74ed0ac6d065be749a363923a9d..18db95f4ca5226cea75bcb4eb94a3e25f161a908)*

src/install.rs
```diff
@@ -13,6 +17,114 @@ use crate::{
...
+        Ok(Self {
+            root,
+            prefix,
+            lock: Some(lock),
+        })
+    }
+
+    fn prefix(&self) -> &Path {
+        &self.prefix
+    }
```
djc: Nit: this helper seems superfluous.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4965/46035eb95a3db74ed0ac6d065be749a363923a9d..18db95f4ca5226cea75bcb4eb94a3e25f161a908)*

src/install.rs
```diff
@@ -102,16 +215,49 @@ impl InstallMethod<'_, '_> {
...
+            // Updates modify the published toolchain in place.
+            InstallMethod::Dist(opts) if opts.exists => {
                 let prefix = &InstallPrefix::from(path.to_owned());
-                let maybe_new_hash = opts.install_into(prefix, manifest).await?;
+                let Some(hash) = opts
+                    .install_into(prefix, &opts.update_hash, manifest)
+                    .await?
+                else {
+                    return Ok(false);
+                };
```
djc: Nit: avoid stylistic changes in a larger commit with a bunch of functional changes.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4965/46035eb95a3db74ed0ac6d065be749a363923a9d..18db95f4ca5226cea75bcb4eb94a3e25f161a908)*

rami3l:
@cachebag Thanks for the work! I'll take care of the merging of your commits in my in-repo PoC from now on, so switching this to draft to prevent merging accidentally 🙏

djc:
> @cachebag Thanks for the work! I'll take care of the merging of your commits in my in-repo PoC from now on, so switching this to draft to prevent merging accidentally 🙏

In the spirit of digesting changes in smaller chunks, might it still make sense to get this merged separately from your PoC?

rami3l:
> > @cachebag Thanks for the work! I'll take care of the merging of your commits in my in-repo PoC from now on, so switching this to draft to prevent merging accidentally 🙏
> 
> In the spirit of digesting changes in smaller chunks, might it still make sense to get this merged separately from your PoC?

@djc Yes, but this PR has already made one design decision, namely the concrete directory of the on-disk staging area. Later on this staging area will be reused by all kinds of other transactions, so IMHO it's still a bit early to fix it right now. Also, there is the problem of the API design regarding the whole transaction: for me right now it is hard to say whether this is the right shape that we can safely extend other features on, although that'd be much clearer shortly.

Finally, as you have mentioned, this is a very digestible piece of change, so I'm expecting to base my future changes on top of this patch (maybe do a bit of small changes) and ship in small batches. Otherwise there is no point in reviewing this PR in the first place 🙏

PR #4966, opened 2026-07-20
refactor(test/clitools)!: refine semantics of `CliTestContext::kill_at()`

rami3l:
Refines #4963 before going further with e.g. #4965.

cc @cachebag

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -8,6 +8,7 @@ use std::str::FromStr;
 #[cfg(feature = "test")]
 use std::{
     collections::HashMap,
+    fs,
     io::Cursor,
     path::Path,
     sync::{Arc, Mutex},
@@ -25,7 +26,10 @@ use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
 
 #[cfg(feature = "test")]
-use crate::cli::log;
+use crate::{
+    cli::log,
+    test::{CHECKPOINT_ENV, checkpoint_path},
+};
 
 mod file_source;
 mod terminal_source;
@@ -233,6 +237,29 @@ impl Process {
     pub fn concurrent_downloads(&self) -> Option<usize> {
         let s = self.var("RUSTUP_CONCURRENT_DOWNLOADS").ok()?;
         Some(NonZero::from_str(&s).ok()?.get())
+    }
+
+    /// Registers a testing checkpoint with the given name and parks the current thread.
+    ///
+    /// Usually, the current process will be killed by the test driver.
+    #[cfg(feature = "test")]
+    pub(crate) fn checkpoint(&self, name: &str) {
+        if self.var(CHECKPOINT_ENV).as_deref() != Ok(name) {
+            return;
+        }
+
+        let rustup_home = self
+            .rustup_home()
+            .expect("selected test checkpoint requires RUSTUP_HOME");
+        let test_root = rustup_home
+            .parent()
+            .expect("test RUSTUP_HOME must be inside the test root");
+        fs::write(checkpoint_path(test_root, name), name)
+            .expect("failed to write test checkpoint marker");
+
+        loop {
+            thread::park();
+        }
     }
 }
 
--- a/src/process.rs
+++ b/src/process.rs
@@ -12,6 +12,7 @@ use std::{
     io::Cursor,
     path::Path,
     sync::{Arc, Mutex},
+    time::{Duration, Instant},
 };
 use std::{env, thread};
 
@@ -257,9 +258,14 @@ impl Process {
         fs::write(checkpoint_path(test_root, name), name)
             .expect("failed to write test checkpoint marker");
 
-        loop {
-            thread::park();
+        let start_time = Instant::now();
+        let max_wait = Duration::from_mins(5);
+        while start_time.elapsed() < max_wait {
+            thread::sleep(Duration::from_secs(10));
         }
+        panic!(
+            "test checkpoint '{name}' timed out after {max_wait:?} without being killed by the test driver",
+        );
     }
 }
 
--- a/src/test.rs
+++ b/src/test.rs
@@ -28,7 +28,7 @@ pub use crate::cli::self_update::{RegistryGuard, RegistryValueId, USER_PATH, get
 
 mod clitools;
 pub use clitools::{
-    Assert, CliTestContext, Config, SanitizedOutput, Scenario, SelfUpdateTestContext,
+    Assert, CliTestContext, Config, ParkedChild, SanitizedOutput, Scenario, SelfUpdateTestContext,
     output_release_file, print_command, print_indented,
 };
 pub(super) mod dist;
--- a/src/test.rs
+++ b/src/test.rs
@@ -21,7 +21,7 @@ use anyhow::Result;
 use sha2::{Digest, Sha256};
 
 use crate::dist::TargetTuple;
-use crate::process::{Process, TestProcess};
+use crate::process::TestProcess;
 
 #[cfg(windows)]
 pub use crate::cli::self_update::{RegistryGuard, RegistryValueId, USER_PATH, get_path};
@@ -36,28 +36,7 @@ pub use dist::DistContext;
 pub(super) mod mock;
 pub use mock::{MockComponentBuilder, MockFile, MockInstallerBuilder};
 
-/// Signal that a selected test checkpoint has been reached, then wait for the
-/// test driver to terminate this process.
-pub(crate) fn checkpoint(process: &Process, name: &str) {
-    if process.var(CHECKPOINT_ENV).as_deref() != Ok(name) {
-        return;
-    }
-
-    let rustup_home = process
-        .rustup_home()
-        .expect("selected test checkpoint requires RUSTUP_HOME");
-    let test_root = rustup_home
-        .parent()
-        .expect("test RUSTUP_HOME must be inside the test root");
-    fs::write(checkpoint_path(test_root, name), name)
-        .expect("failed to write test checkpoint marker");
-
-    loop {
-        std::thread::park();
-    }
-}
-
-fn checkpoint_path(test_root: &Path, name: &str) -> PathBuf {
+pub fn checkpoint_path(test_root: &Path, name: &str) -> PathBuf {
     test_root.join(format!("rustup-checkpoint-{name}"))
 }
 
@@ -353,4 +332,4 @@ pub static MULTI_ARCH1: &str = "i686-unknown-linux-gnu";
 #[cfg(not(target_pointer_width = "64"))]
 static MULTI_ARCH1: &str = "x86_64-unknown-linux-gnu";
 
-const CHECKPOINT_ENV: &str = "RUSTUP_TEST_CHECKPOINT";
+pub const CHECKPOINT_ENV: &str = "RUSTUP_TEST_CHECKPOINT";
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -973,9 +973,8 @@ impl CliTestContext {
         Self { config, _test_dir }
     }
 
-    /// Run a rustup command until it reaches `checkpoint`, then terminate it
-    /// without giving destructors an opportunity to run.
-    pub fn kill_at_checkpoint(&self, mut command: Command, checkpoint: &str) -> ExitStatus {
+    /// Run a rustup command until it reaches `checkpoint` and terminate it.
+    pub fn kill_at<S: AsRef<OsStr>>(&self, checkpoint: &str, args: impl AsRef<[S]>) -> ExitStatus {
         let marker = checkpoint_path(&self.config.test_root_dir, checkpoint);
         if let Err(error) = fs::remove_file(&marker)
             && error.kind() != io::ErrorKind::NotFound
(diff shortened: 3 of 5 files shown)


src/test/clitools.rs
```diff
@@ -975,16 +975,33 @@ impl CliTestContext {
...
+        let mut cmd = self.config.cmd(
+            program
+                .as_ref()
+                .to_str()
+                .expect("invalid UTF-8 in program name"),
+            args,
+        );
+        cmd.env(CHECKPOINT_ENV, checkpoint);
+
+        let _lock = CMD_LOCK.read().unwrap();
```
cachebag: Should this be dropped right after `spawn()`? Similar to as it already hapens here:

https://github.com/rust-lang/rustup/blob/09b779d757cb5336780228624e65bae891e1d9f1/src/test/clitools.rs#L479-L481

cachebag (commented):
I reworked #4965 a bit to follow the patterns you have here. I will plan to rebase this onto it once it is merged. One thing maybe to note for myself it to drop my local copy of the metadata checkpoint name in favor of`CHECKPOINT_UPDATE_BEFORE_METADATA`.

One question if you don't mind: for testing concurrent installs of the same toolchain, I'll need to keep a child parked at a checkpoint while a second command runs, then kill it explicitly. Would you be open to a `spawn_at(checkpoint, args)`-style primitive returning a guard (child killed on drop), with `kill_at` becoming a thin wrapper over it? I can propose that when I rebase.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/c950062148fd649833012c668f0d6fd2b2ccf727..09b779d757cb5336780228624e65bae891e1d9f1)*

src/test/clitools.rs
rami3l: @cachebag The lock should be dropped after `wait()`, not `spawn()`. See the std implementation of the `output()` function:

```rs
pub fn output(cmd: &mut Command) -> crate::io::Result<(ExitStatus, Vec<u8>, Vec<u8>)> {
    let (mut process, mut pipes) = cmd.spawn(Stdio::MakePipe, false)?;

    drop(pipes.stdin.take());
    let (mut stdout, mut stderr) = (Vec::new(), Vec::new());
    match (pipes.stdout.take(), pipes.stderr.take()) {
        (None, None) => {}
        (Some(out), None) => {
            let res = out.read_to_end(&mut stdout);
            res.unwrap();
        }
        (None, Some(err)) => {
            let res = err.read_to_end(&mut stderr);
            res.unwrap();
        }
        (Some(out), Some(err)) => {
            let res = read_output(out, &mut stdout, err, &mut stderr);
            res.unwrap();
        }
    }

    let status = process.wait()?;
    Ok((status, stdout, stderr))
}
```

rami3l:
> One question if you don't mind: for testing concurrent installs of the same toolchain, I'll need to keep a child parked at a checkpoint while a second command runs, then kill it explicitly. Would you be open to a `spawn_at(checkpoint, args)`-style primitive returning a guard (child killed on drop), with `kill_at` becoming a thin wrapper over it? I can propose that when I rebase.

@cachebag Thanks for the feedback! Please feel free to suggest a patch based on the changes here and I'll add you as a coauthor here.

cachebag:
@rami3l 

<details><summary>patch</summary>

```patch
From 91705e92efcc5fdd5adcb4b29f15dd624f5ad45f Mon Sep 17 00:00:00 2001
From: akrm al-hakimi <alhakimiakrmj@gmail.com>
Date: Fri, 24 Jul 2026 11:09:26 -0400
Subject: [PATCH] refactor(test/clitools): extract `spawn_at()` and a
 kill-on-drop guard

`kill_at()` becomes a thin wrapper over `spawn_at()`, which parks the
command at its checkpoint and returns a `ParkedChild` guard that kills
it on drop, covering the timeout and early-exit panic paths as well.
The `CMD_LOCK` read guard is scoped to the spawn itself: it cannot stay
held while the child is parked, since a writer queued behind it would
block every command spawned in the meantime, including the ones a test
wants to run against the parked process.
---
 src/test.rs          |  2 +-
 src/test/clitools.rs | 79 +++++++++++++++++++++++++++++++++-----------
 2 files changed, 61 insertions(+), 20 deletions(-)

diff --git a/src/test.rs b/src/test.rs
index 314b4624..7c96bff6 100644
--- a/src/test.rs
+++ b/src/test.rs
@@ -28,7 +28,7 @@ pub use crate::cli::self_update::{RegistryGuard, RegistryValueId, USER_PATH, get
 
 mod clitools;
 pub use clitools::{
-    Assert, CliTestContext, Config, SanitizedOutput, Scenario, SelfUpdateTestContext,
+    Assert, CliTestContext, Config, ParkedChild, SanitizedOutput, Scenario, SelfUpdateTestContext,
     output_release_file, print_command, print_indented,
 };
 pub(super) mod dist;
diff --git a/src/test/clitools.rs b/src/test/clitools.rs
index a8cfe9fb..6c563059 100644
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -12,7 +12,7 @@ use std::{
     mem,
     ops::{Deref, DerefMut},
     path::{Path, PathBuf},
-    process::{Command, ExitStatus},
+    process::{Child, Command, ExitStatus},
     string::FromUtf8Error,
     sync::{Arc, LazyLock, RwLock, RwLockWriteGuard},
     thread,
@@ -973,13 +973,14 @@ impl CliTestContext {
         Self { config, _test_dir }
     }
 
-    /// Run a rustup command until it reaches `checkpoint`, then terminate it
-    /// without giving destructors an opportunity to run.
-    pub fn kill_at<S: AsRef<OsStr> + Clone + Debug>(
+    /// Run a rustup command until it reaches `checkpoint` and keep it parked
+    /// there, e.g. to observe or race the paused operation from another
+    /// process. The returned guard kills the command when dropped.
+    pub fn spawn_at<S: AsRef<OsStr> + Clone + Debug>(
         &self,
         checkpoint: &str,
         args: impl AsRef<[S]>,
-    ) -> ExitStatus {
+    ) -> ParkedChild {
         let marker = checkpoint_path(&self.config.test_root_dir, checkpoint);
         if let Err(error) = fs::remove_file(&marker)
             && error.kind() != io::ErrorKind::NotFound
@@ -1000,36 +1001,46 @@ impl CliTestContext {
         );
         cmd.env(CHECKPOINT_ENV, checkpoint);
 
-        let _lock = CMD_LOCK.read().unwrap();
-        let mut child = cmd
-            .spawn()
-            .expect("failed to start command for checkpoint test");
+        // The lock covers the spawn itself. It must not be held while the
+        // child stays parked: a writer queued behind it would then block
+        // every command spawned in the meantime, including the ones a test
+        // wants to run against the parked process.
+        let child = {
+            let _lock = CMD_LOCK.read().unwrap();
+            cmd.spawn()
+                .expect("failed to start command for checkpoint test")
+        };
+        let mut parked = ParkedChild { child: Some(child) };
 
         let deadline = Instant::now() + StdDuration::from_secs(10);
         loop {
             if matches!(fs::read_to_string(&marker), Ok(contents) if contents == checkpoint) {
-                break;
+                break parked;
             }
-            if let Some(status) = child
+            if let Some(status) = parked
+                .child
+                .as_mut()
+                .unwrap()
                 .try_wait()
                 .expect("failed to query checkpoint test command")
             {
                 panic!("command exited before reaching checkpoint {checkpoint:?}: {status}");
             }
             if Instant::now() >= deadline {
-                let _ = child.kill();
-                let _ = child.wait();
                 panic!("timed out waiting for checkpoint {checkpoint:?}");
             }
             thread::sleep(StdDuration::from_millis(10));
         }
+    }
 
-        child
-            .kill()
-            .expect("failed to terminate command at checkpoint");
-        child
-            .wait()
-            .expect("failed to reap command after checkpoint")
+    /// Run a rustup command until it reaches `checkpoint`, then terminate it
+    /// without giving destructors an opportunity to run.
+    pub fn kill_at<S: AsRef<OsStr> + Clone + Debug>(
+        &self,
+        checkpoint: &str,
+        args: impl AsRef<[S]>,
+    ) -> ExitStatus {
+        self.spawn_at(checkpoint, args).kill()
     }
 
     /// Move the dist server to the specified scenario and restore it
@@ -1108,6 +1119,36 @@ impl Drop for WorkDirGuard<'_> {
     }
 }
 
+/// A rustup command spawned by [`CliTestContext::spawn_at`], parked at a
+/// checkpoint. Killed on drop unless [`ParkedChild::kill`] was called first.
+#[must_use]
+pub struct ParkedChild {
+    child: Option<Child>,
+}
+
+impl ParkedChild {
+    /// Terminate the parked command without giving destructors an opportunity
+    /// to run, then reap it.
+    pub fn kill(mut self) -> ExitStatus {
+        let mut child = self.child.take().unwrap();
+        child
+            .kill()
+            .expect("failed to terminate command at checkpoint");
+        child
+            .wait()
+            .expect("failed to reap command after checkpoint")
+    }
+}
+
+impl Drop for ParkedChild {
+    fn drop(&mut self) {
+        if let Some(mut child) = self.child.take() {
+            let _ = child.kill();
+            let _ = child.wait();
+        }
+    }
+}
+
 #[must_use]
 pub struct DistDirGuard<'a> {
     inner: &'a mut CliTestContext,
-- 
2.39.5 (Apple Git-154)
```

</details>

src/test/clitools.rs
rami3l: Ah I see what you mean in https://github.com/rust-lang/rustup/pull/4966#issuecomment-5071427749. My apologies!

src/test.rs
```diff
@@ -36,24 +36,26 @@ pub use dist::DistContext;
 pub(super) mod mock;
 pub use mock::{MockComponentBuilder, MockFile, MockInstallerBuilder};
 
-/// Signal that a selected test checkpoint has been reached, then wait for the
-/// test driver to terminate this process.
-pub(crate) fn checkpoint(process: &Process, name: &str) {
-    if process.var(CHECKPOINT_ENV).as_deref() != Ok(name) {
-        return;
-    }
+impl Process {
```
djc: I think we should just move this into the existing `Process` impl.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/720e6b862df1309c0a5b2aea7f81be1f975af41f..492864368a2a9cd71aebe2c557a1366a770fdba6)*

src/test/clitools.rs
```diff
@@ -980,6 +980,17 @@ impl CliTestContext {
         checkpoint: &str,
         args: impl AsRef<[S]>,
     ) -> ExitStatus {
+        self.spawn_at(checkpoint, args).kill()
+    }
+
+    /// Run a rustup command until it reaches `checkpoint` and keep it parked
```
djc: Nit: documentation comments should start with a single-line short sentence.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/720e6b862df1309c0a5b2aea7f81be1f975af41f..492864368a2a9cd71aebe2c557a1366a770fdba6)*

src/test/clitools.rs
```diff
@@ -1009,32 +1020,27 @@ impl CliTestContext {
             cmd.spawn()
                 .expect("failed to start command for checkpoint test")
         };
+        let mut parked = ParkedChild { child: Some(child) };
```
djc: This seems like weird style -- why don't we wrap `child` in `ParkedChild` in the earlier block?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/720e6b862df1309c0a5b2aea7f81be1f975af41f..492864368a2a9cd71aebe2c557a1366a770fdba6)*

src/test/clitools.rs
```diff
@@ -1113,6 +1119,37 @@ impl Drop for WorkDirGuard<'_> {
     }
 }
 
+/// A rustup command spawned by [`CliTestContext::spawn_at`], parked at a
```
djc: Nit: documentation comments.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/720e6b862df1309c0a5b2aea7f81be1f975af41f..492864368a2a9cd71aebe2c557a1366a770fdba6)*

src/test/clitools.rs
```diff
@@ -975,18 +975,40 @@ impl CliTestContext {
 
     /// Run a rustup command until it reaches `checkpoint`, then terminate it
     /// without giving destructors an opportunity to run.
-    pub fn kill_at_checkpoint(&self, mut command: Command, checkpoint: &str) -> ExitStatus {
+    pub fn kill_at<S: AsRef<OsStr> + Clone + Debug>(
```
djc: Nit: is there an actual use case for the generics here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/382311277d0b4752b262bcb636e2f7368740111f..c91caa9e28123ecf6f1b43a3b2d0a9cd978e8fc0)*

src/test/clitools.rs
```diff
@@ -1000,41 +1011,37 @@ impl CliTestContext {
...
 
-        let mut child = {
-            // The lock covers the spawn itself. It must not be held while the
-            // child stays parked: a writer queued behind it would then block
-            // every command spawned in the meantime, including the ones a test
-            // wants to run against the parked process.
-            let _lock = CMD_LOCK.read().unwrap();
-            cmd.spawn()
-                .expect("failed to start command for checkpoint test")
+        let mut parked = ParkedChild {
```
djc: This version also looks pretty weird. Can't we do something more like the previous version, but wrapping `ParkedChild` around the result of `cmd.spawn().expect()`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/382311277d0b4752b262bcb636e2f7368740111f..c91caa9e28123ecf6f1b43a3b2d0a9cd978e8fc0)*

src/process.rs
```diff
@@ -257,9 +258,14 @@ impl Process {
         fs::write(checkpoint_path(test_root, name), name)
             .expect("failed to write test checkpoint marker");
 
-        loop {
-            thread::park();
+        let start_time = time::Instant::now();
+        let max_wait = time::Duration::from_mins(5);
+        while start_time.elapsed() < max_wait {
+            thread::sleep(time::Duration::from_secs(10));
```
djc: Suggest importing `Instant` and `Duration` directly here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4966/382311277d0b4752b262bcb636e2f7368740111f..c91caa9e28123ecf6f1b43a3b2d0a9cd978e8fc0)*

src/test/clitools.rs
rami3l: I would argue here that `Clone + Debug` is not used so far so I can remove that first.

Apart from that I think we are just forwarding the bounds of [`Command::new()`](https://doc.rust-lang.org/std/process/struct.Command.html#method.new) here. Previously this was in the constructor.

src/test/clitools.rs
rami3l: I have updated the code accordingly. Do you like this new version better? If you still find it not satisfactory, would you mind providing a small hint (a code snippet) about where I should possibly go? Thanks 🙏

djc (approved):

PR #4969, opened 2026-07-21
fix: warn about unknown toolchain file keys

### LINKED ISSUE #2876: Unknown keys in rust-toolchain.toml are silently ignored

ychampion:
## Summary

- warn with the exact ignored key path when a toolchain file contains unknown settings
- preserve current parsing behavior while avoiding duplicate warnings from repeated reads
- document the warning-first transition

## Why

Unknown settings can be misspellings, but rejecting them immediately would break toolchain files that currently rely on ignored keys. This stages the stricter behavior requested in #4953 without changing current configuration semantics.

Closes #2876

## Validation

- `cargo test --features=test --test test_bonanza rust_toolchain_toml_warns_on_unknown_keys`
- `cargo test --features=test --lib`
- `cargo clippy --all-targets --features=test -- -D warnings`
- `mdbook build`
- `git diff --check`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,9 +1,11 @@
+use std::collections::HashSet;
 use std::fmt::{self, Debug, Display};
 use std::io;
 use std::io::Write;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::sync::Mutex;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use anyhow::{Context, Result, anyhow, bail};
@@ -290,6 +292,7 @@ pub(crate) struct Cfg<'a> {
     pub quiet: bool,
     pub current_dir: PathBuf,
     pub process: &'a Process,
+    warned_unknown_toolchain_keys: Mutex<HashSet<(PathBuf, String)>>,
 
     /// Allows the auto-installation of the active toolchain when it is missing.
     ///
@@ -370,6 +373,7 @@ impl<'a> Cfg<'a> {
             quiet,
             current_dir,
             process,
+            warned_unknown_toolchain_keys: Mutex::new(HashSet::new()),
             allow_auto_install,
         };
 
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -744,10 +772,23 @@ impl<'a> Cfg<'a> {
         Ok(None)
     }
 
+    #[cfg(test)]
```
rami3l: Your LLM is being too defensive again... I don't see a reason to introduce a new API. Just break the existing API by changing the return value of the existing function.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4969/c950062148fd649833012c668f0d6fd2b2ccf727..58d8556160cc582fd891e81351e44cae7f9d6519)*

src/config.rs
```diff
@@ -688,13 +692,37 @@ impl<'a> Cfg<'a> {
...
+                    contents, parse_mode,
+                )
+                .with_context(|| RustupError::ParsingFile {
+                    name: "override",
+                    path: toolchain_file.clone(),
+                })?;
+                {
+                    let mut warned_unknown_toolchain_keys = self
+                        .warned_unknown_toolchain_keys
+                        .lock()
```
rami3l: I understand that here you don't want to spam the user with too many unknown keys. However, as per your previous analysis in https://github.com/rust-lang/rustup/pull/4953#issuecomment-4952373469 this mechanism doesn't sound necessary.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4969/c950062148fd649833012c668f0d6fd2b2ccf727..58d8556160cc582fd891e81351e44cae7f9d6519)*

src/config.rs
ychampion: Changed `parse_override_file` itself to return the parsed config and unknown keys, and removed the parallel helper. The caller still emits warnings before rejecting an empty config. Covered by the parser tests and CLI regression in e480ed8e.

src/config.rs
ychampion: Removed the `HashSet`/`Mutex` deduplication state. Unknown keys are now warned directly from each parse result in e480ed8e.

djc:
Please clean up the PR description and commit messages in line with our AI policy:

https://rust-lang.github.io/rustup/dev-guide/#ai-policy

src/config.rs
```diff
@@ -758,18 +770,20 @@ impl<'a> Cfg<'a> {
...
+                        unknown_keys.push(path.to_string());
+                    })
                     .context(OverrideFileConfigError::Parsing)?;
 
-                if override_file.is_empty() {
-                    Err(anyhow!(OverrideFileConfigError::Invalid))
-                } else {
-                    Ok(override_file)
-                }
+                Ok((override_file, unknown_keys))
```
rami3l, line 786: You have distorted the original semantics of this function.

Also, it might be worth it to add an additional test case if this distortion has not been exposed in our test cases.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4969/c950062148fd649833012c668f0d6fd2b2ccf727..e480ed8eadf83ea5246c031378d0cb8837d63cf8)*

src/config.rs
```diff
@@ -758,18 +770,20 @@ impl<'a> Cfg<'a> {
...
             }
```
rami3l, line 787: Would it be better to just warn about the unknown keys right here instead of altering the return value?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4969/c950062148fd649833012c668f0d6fd2b2ccf727..e480ed8eadf83ea5246c031378d0cb8837d63cf8)*

rami3l (changes requested):

PR #4973, opened 2026-07-25
fix(config): refine error handling of `Cfg::notify_release()`

rami3l:
When notifying a release, we should not assume that the `stable` toolchain is already installed, otherwise we would spam the user with those warnings:

```smalltalk
warn: could not check if there is a new Rust release: stable toolchain unavailable: toolchain 'stable-aarch64-apple-darwin' is not installed
help: run `rustup toolchain install stable-aarch64-apple-darwin` to install it
```

This PR makes this warning disappear when `stable` is not installed.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1021,9 +1021,13 @@ impl<'a> Cfg<'a> {
 
         let default_host = self.default_host_tuple()?;
         let stable_desc = PartialToolchainDesc::from_str("stable")?.resolve(&default_host)?;
-        let distributable = DistributableToolchain::new(self, stable_desc)
-            .map_err(|e| anyhow!("stable toolchain unavailable: {e}"))?;
-        let release_date_str = match distributable.get_manifestation() {
+        let stable = match DistributableToolchain::new(self, stable_desc) {
+            Ok(stable) => stable,
+            // If the `stable` toolchain is not installed, we don't notify the user.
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -1021,9 +1021,13 @@ impl<'a> Cfg<'a> {
 
         let default_host = self.default_host_tuple()?;
         let stable_desc = PartialToolchainDesc::from_str("stable")?.resolve(&default_host)?;
-        let distributable = DistributableToolchain::new(self, stable_desc)
-            .map_err(|e| anyhow!("stable toolchain unavailable: {e}"))?;
-        let release_date_str = match distributable.get_manifestation() {
+        let stable = match DistributableToolchain::new(self, stable_desc) {
+            Ok(stable) => stable,
+            // With no `stable` toolchain installed, we currently don't notify the user.
```
FranciscoTGouveia:
```suggestion
            // If the `stable` toolchain is not installed, we don't notify the user.
```

FranciscoTGouveia (approved):
LGTM! Thanks for working on this.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4973/0878429a3d0868928cb57085332854e183f9fcfe..dc3e4ec4c1bc812e5fe34714b5e2b5e5a2ae9ccb)*

PR #4981, opened 2026-07-27
www: align copy icon

geocine:
## Summary

This has been bothering me everytime I visit the rustup.rs website 

image

## Why
It's misaligned

## Validation
Check screenshot

### DIFF
--- a/www/rustup.css
+++ b/www/rustup.css
@@ -214,11 +214,12 @@ hr {
 }
 
 div.copy-icon {
-    transform: translateX(3px);
+    transform: translateX(2px);
 }
 
 div.copy-icon svg {
     fill: var(--body-color);
+    display: block;
 }
 
 button.copy-button:hover svg {


ChrisDenton:
If the issue is extract vertical spacing messing up the centering, would using `div.copy-icon svg { display:bock }` have the same effect without needing to manually specifying the offset?

rami3l:
> If the issue is extract vertical spacing messing up the centering, would using `div.copy-icon svg { display:bock }` have the same effect without needing to manually specifying the offset?

@ChrisDenton I think the thing with our icon is that IIRC its visual center is not the actual center so that's why manual adjustments have to be made 😓 

I'd argue that in "proper" Web frontend an asset library is needed but probably not in this case: so far we are keeping a very minimal Vanilla JS setup.

rami3l (approved):

ChrisDenton:
For the horizontal center that's true (because there's a bit to one side) but for the vertical center it shouldn't be

rami3l:
> For the horizontal center that's true (because there's a bit to one side) but for the vertical center it shouldn't be

@ChrisDenton I get it, it's a typo XDD (block, not bock)

ChrisDenton:
Oops, sorry!

www/rustup.css
rami3l, line 221:
```suggestion
    fill: var(--body-color);
    display: block;
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4981/720e6b862df1309c0a5b2aea7f81be1f975af41f..c7632115876bdd7de212213553a2b6c3b14bd37f)*

www/rustup.css
```diff
@@ -214,7 +214,7 @@ hr {
 }
 
 div.copy-icon {
-    transform: translateX(3px);
+    transform: translateX(2px) translateY(2px);
```
rami3l:
```suggestion
    transform: translateX(2px);
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4981/720e6b862df1309c0a5b2aea7f81be1f975af41f..c7632115876bdd7de212213553a2b6c3b14bd37f)*

www/rustup.css
rami3l: @ChrisDenton Is it more like this?

ChrisDenton: Yep! Does it make a difference in your browser?

rami3l:
> Oops, sorry!

@ChrisDenton I just applied your suggestions and now it looks like so:

image

... I think we can merge this now.

ChrisDenton:
Great, thanks for testing!

PR #4970, opened 2026-07-22
doc: add AI policy to the dev guide

djc:

PR #4985, opened 2026-07-29
docs(dev-guide): reapply abandoned changes from #4970

rami3l:
In GHMQ, any operation on the topic branch after enqueuing is considered a no-op, so what we have on our `main` branch is an older version of that document.

image

 This PR has thus reapplied what has been abandoned in #4970.


djc (approved):

djc:
Oops!

PR #4986, opened 2026-07-29
Add `rustup doc --serve` to serve docs over local HTTP

### LINKED ISSUE #4975: Improve `rustup doc` UX with sandboxed default browser

`rustup doc` currently opens the local Rust documentation using the system's default browser.

On Ubuntu, if the default browser is a sandboxed browser (such as the Snap packages of Brave or Firefox), the browser may be unable to access the documentation stored under `~/.rustup` via a `file://` URL. From the user's perspective, `rustup doc` appears to fail and the browser only displays an "Access denied" or similar error.

This appears to be caused by the browser's sandbox rather than a bug in `rustup`. However, the current experience is confusing because users receive no indication that the failure is caused by browser restrictions or what they can do instead.

### Solution you'd like

Ideally, `rustup doc` could provide a more helpful experience when the local documentation cannot be opened.

Some possible approaches (not prescriptive) are:

- Detect when opening the documentation fails and print a message explaining that sandboxed browsers (such as Snap or Flatpak packages) may not be able to access documentation stored under `~/.rustup`.
- Suggest a workaround, such as using the online documentation or serving the local documentation over HTTP.
- If feasible, consider providing an option to serve the documentation locally instead of relying on `file://` URLs.

I don't have a specific implementation in mind. The main goal is to make it easier for users to understand why `rustup doc` did not work and how they can access the documentation.

## Environment

- Ubuntu
- Rust installed via `rustup`
- Reproduced with the Snap versions of both Brave (default browser) and Firefox

## Reproduction

```sh
rustup doc
```

The browser fails to open the local documentation, reporting an access denied error.

## Investigation

The documentation exists and is readable. Copying it outside `~/.rustup` allows it to open normally, which suggests the issue is due to browser sandbox restrictions on `file://` access rather than the documentation itself.

## Current workaround

Serve the documentation over a local HTTP server:

```sh
cd ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/rust/html
python3 -m http.server 8000
```

Then open:

```
http://localhost:8000
```

Aneesh-382005:
## The Problem
`rustup doc` currently opens the local Rust documentation as `file://` URLs using the system's default browser, but sandbox browsers are unable to access the documentation. The browser opens, just to display an "Access denied" error

## This PR
Adds an opt-in `--serve` flag. So instead of building a `file://` URL, `rustup doc --serve` opens an `http://127.0.0.1:<port>/...`
URL instead, which can be accessed by a sandboxed browser. Default `rustup doc` behavior is completely unchanged.

## Design decisions
- Bind only to `127.0.0.1` on an ephemeral port.
- Serve the  documentation directory as root so it keeps the relative links/CSS/JS/search working
- Reject path traversal attempts (`..` or absolute paths) to keep requests confined to the doc root.
- Keep the server running until the user exits with `Ctrl+C`. Did not go with any idle timeouts because a timeout would be arbitrary and could interrupt reading.
- Promote `hyper` (`server` + `http1`) from a dev-dependency to a dependency. Although `reqwest` already depends on `hyper`, that only exposes its client-side functionality internally, so a direct dependency is required to use the server APIs. `Cargo.lock` adds no new packages, and the release binary grows from **14,001,792** to **14,204,752** bytes (**+198.2 KiB**, **+1.45%**).
- I did think of automatic fallback but `rustup` can tell whether the browser was launched, but not whether it successfully rendered the documentation.

## Testing
- `cargo build` and `cargo clippy --all-targets --features test` pass cleanly.
- `cargo test --features test download::tests`: 5/5 tests pass.
- Manual: Verified `rustup doc --serve` and `rustup doc --serve std::vec`. Docs, assets, and relative links load correctly, invalid paths return 404, and Ctrl-C stops the server and releases the port.

Fixes #4975

### DIFF
--- a/src/cli/docs.rs
+++ b/src/cli/docs.rs
@@ -1,19 +1,41 @@
 //! `rustup doc` and `rustup man`: opening toolchain documentation and man pages.
+//!
+//! `rustup doc --serve` serves the documentation over a local HTTP server
+//! instead of opening it directly as a `file://` URL. Some browsers (e.g.
+//! Snap/Flatpak builds of Firefox or Brave) run in a sandbox that can't
+//! access `file://` URLs under `~/.rustup`; serving the same static files
+//! over `http://127.0.0.1` sidesteps that restriction entirely.
+//!
+//! The server binds to `127.0.0.1` only (never exposed on the network) and
+//! rejects any request path that would resolve outside the served directory.
 
 use std::{
     borrow::Cow,
+    convert::Infallible,
     io::Write as _,
-    path::{Path, PathBuf},
+    path::{Component, Path, PathBuf},
+    sync::Arc,
 };
 
 use anyhow::{Context, Result, anyhow};
 use clap::Args;
+use http_body_util::Full;
+use hyper::{
+    Request, Response, StatusCode,
+    body::{Bytes, Incoming},
+    header::{CONTENT_LENGTH, CONTENT_TYPE},
+    server::conn::http1,
+    service::service_fn,
+};
+use hyper_util::rt::TokioIo;
+use tokio::net::TcpListener;
 use tracing::info;
 
 use super::topical_doc;
 use crate::{
     config::{ActiveSource, Cfg},
     dist::{PartialToolchainDesc, manifest::ComponentStatus},
+    process::Process,
     toolchain::DistributableToolchain,
     utils::{self, ExitCode},
 };
@@ -162,6 +184,7 @@ impl DocPage {
 pub(crate) async fn doc(
     cfg: &Cfg<'_>,
     path_only: bool,
+    serve: bool,
     toolchain: Option<PartialToolchainDesc>,
     mut topic: Option<&str>,
     doc_page: &DocPage,
@@ -212,6 +235,12 @@ pub(crate) async fn doc(
         return Ok(ExitCode::SUCCESS);
     }
 
+    if serve {
+        let root = toolchain.doc_path("")?;
+        serve_and_open(root, &doc_path, fragment, cfg.process).await?;
+        return Ok(ExitCode::SUCCESS);
+    }
+
     if let Some(name) = topic {
         writeln!(
             cfg.process.stderr().lock(),
@@ -246,4 +275,107 @@ pub(crate) async fn man(
         .status()
         .expect("failed to open man page");
     Ok(ExitCode::SUCCESS)
+}
+
+/// Blocks forever, accepting and serving connections until the process is
+/// killed by Ctrl-C.
+async fn serve_and_open(
+    root: PathBuf,
+    initial_path: &Path,
+    fragment: Option<&str>,
+    process: &Process,
+) -> Result<()> {
+    let listener = TcpListener::bind(("127.0.0.1", 0))
+        .await
+        .context("failed to bind local documentation server")?;
+    let addr = listener
+        .local_addr()
+        .context("failed to read local address")?;
+    let root = Arc::<Path>::from(root);
+
+    let mut url = format!(
+        "http://{addr}/{}",
+        initial_path.to_string_lossy().replace('\\', "/")
+    );
+    if let Some(fragment) = fragment {
+        url.push('#');
+        url.push_str(fragment);
+    }
+
+    writeln!(
+        process.stderr().lock(),
+        "Serving docs at {url} (press Ctrl-C to stop)"
+    )?;
+    utils::open_browser(&url)?;
+
+    loop {
+        let (stream, _) = match listener.accept().await {
+            Ok(accepted) => accepted,
+            Err(err) => {
+                tracing::warn!("doc server: failed to accept connection: {err}");
+                continue;
+            }
+        };
+        let io = TokioIo::new(stream);
+        let root = root.clone();
+        let svc = service_fn(move |req| serve(req, root.clone()));
+
+        tokio::spawn(async move {
+            if let Err(err) = http1::Builder::new().serve_connection(io, svc).await {
+                tracing::warn!("doc server: connection error: {err}");
+            }
+        });
+    }
+}
+
+async fn serve(
+    req: Request<Incoming>,
+    root: Arc<Path>,
+) -> Result<Response<Full<Bytes>>, Infallible> {
+    let request_path = req.uri().path().trim_start_matches('/');
+    let mut path = root.to_path_buf();
+    for segment in Path::new(request_path).components() {
+        match segment {
+            Component::Normal(part) => path.push(part),
+            _ => return Ok(not_found()),
+        }
+    }
+    if !path.starts_with(&root) {
+        return Ok(not_found());
+    }
+    if path.is_dir() {
+        path.push("index.html");
+    }
+
+    let Ok(contents) = tokio::fs::read(&path).await else {
+        return Ok(not_found());
+    };
+
+    let response = Response::builder()
+        .status(StatusCode::OK)
+        .header(
+            CONTENT_TYPE,
+            match path.extension().and_then(|ext| ext.to_str()) {
+                Some("html") => "text/html; charset=utf-8",
+                Some("css") => "text/css",
+                Some("js") => "text/javascript",
+                Some("svg") => "image/svg+xml",
+                Some("png") => "image/png",
+                Some("jpg" | "jpeg") => "image/jpeg",
+                Some("woff2") => "font/woff2",
+                Some("txt") => "text/plain; charset=utf-8",
+                _ => "application/octet-stream",
+            },
+        )
+        .header(CONTENT_LENGTH, contents.len())
+        .body(Full::new(Bytes::from(contents)))
+        .expect("building a static response can't fail");
+    Ok(response)
+}
+
+fn not_found() -> Response<Full<Bytes>> {
+    Response::builder()
+        .status(StatusCode::NOT_FOUND)
+        .body(Full::new(Bytes::from_static(b"404 not found")))
+        .expect("building a static response can't fail")
 }
--- a/src/cli/docs.rs
+++ b/src/cli/docs.rs
@@ -1,0 +1,249 @@
+//! `rustup doc` and `rustup man`: opening toolchain documentation and man pages.
+
+use std::{
+    borrow::Cow,
+    io::Write as _,
+    path::{Path, PathBuf},
+};
+
+use anyhow::{Context, Result, anyhow};
+use clap::Args;
+use tracing::info;
+
+use super::topical_doc;
+use crate::{
+    config::{ActiveSource, Cfg},
+    dist::{PartialToolchainDesc, manifest::ComponentStatus},
+    toolchain::DistributableToolchain,
+    utils::{self, ExitCode},
+};
+
+macro_rules! docs_data {
+    (
+        $(
+            $( #[$meta:meta] )*
+            ($ident:ident, $help:expr, $path:expr $(,)?)
+        ),+ $(,)?
+    ) => {
+        #[derive(Debug, Args)]
+        pub(crate) struct DocPage {
+            $(
+                #[doc = $help]
+                #[arg(long, group = "page")]
+                $( #[$meta] )*
+                $ident: bool,
+            )+
+        }
+
+        impl DocPage {
+            fn path_str(&self) -> Option<&'static str> {
+                $( if self.$ident { return Some($path); } )+
+                None
+            }
+        }
+    };
+}
+
+docs_data![
+    // flags can be used to open specific documents, e.g. `rustup doc --nomicon`
+    // tuple elements: document name used as flag, help message, document index path
+    (
+        alloc,
+        "The Rust core allocation and collections library",
+        "alloc/index.html"
+    ),
+    (
+        book,
+        "The Rust Programming Language book",
+        "book/index.html"
+    ),
+    (cargo, "The Cargo Book", "cargo/index.html"),
+    (clippy, "The Clippy Documentation", "clippy/index.html"),
+    (core, "The Rust Core Library", "core/index.html"),
+    (
+        edition_guide,
+        "The Rust Edition Guide",
+        "edition-guide/index.html"
+    ),
+    (
+        embedded_book,
+        "The Embedded Rust Book",
+        "embedded-book/index.html"
+    ),
+    (
+        error_codes,
+        "The Rust Error Codes Index",
+        "error_codes/index.html"
+    ),
+    (
+        nomicon,
+        "The Dark Arts of Advanced and Unsafe Rust Programming",
+        "nomicon/index.html"
+    ),
+    #[arg(long = "proc_macro")]
+    (
+        proc_macro,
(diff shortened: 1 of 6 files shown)


src/cli/rustup_mode.rs
```diff
@@ -252,6 +253,13 @@ enum RustupSubcmd {
         #[arg(long)]
         path: bool,
 
+        /// Serve the documentation over a local HTTP server instead of
+        /// opening it directly as a `file://` URL. Useful when your browser
+        /// is sandboxed (e.g. installed via Snap or Flatpak) and can't
+        /// access files under `~/.rustup`
```
djc: Let's drop the second sentence here? Once someone has found the option, I think they can decide whether it's useful or not.

src/cli/doc_server.rs
```diff
@@ -0,0 +1,144 @@
...
+    match path.extension().and_then(|ext| ext.to_str()) {
+        Some("html") => "text/html; charset=utf-8",
+        Some("css") => "text/css",
+        Some("js") => "text/javascript",
+        Some("json") => "application/json",
+        Some("svg") => "image/svg+xml",
+        Some("png") => "image/png",
+        Some("woff") => "font/woff",
+        Some("woff2") => "font/woff2",
+        _ => "application/octet-stream",
```
djc: If we're going to do this, should clean this up to match actually used extensions:

```
61852 html
3119 js
 234 woff2
 179 css
  73 svg
  52 png
  48 txt
  20 nojekyll
   2 toml
   2 md
   1 rs
   1 lock
   1 jpg
   1 jpeg
```

So we don't need `json` or `woff`, but we do need `jpg` and `jpeg` and `txt`.

src/cli/doc_server.rs
```diff
@@ -0,0 +1,144 @@
...
+
+async fn serve(
+    req: Request<hyper::body::Incoming>,
+    root: PathBuf,
+) -> Result<Response<Full<Bytes>>, Infallible> {
+    let Some(path) = resolve_within_root(&root, req.uri().path()) else {
+        return Ok(not_found());
+    };
+
+    match tokio::fs::read(&path).await {
```
djc: Nit: do a `let Ok(contents)` for this to reduce rightward drift.

src/cli/doc_server.rs
```diff
@@ -0,0 +1,144 @@
...
+        Ok(contents) => {
+            let content_type = content_type_for(&path);
+            let response = Response::builder()
+                .status(StatusCode::OK)
+                .header(hyper::header::CONTENT_TYPE, content_type)
+                .header(hyper::header::CONTENT_LENGTH, contents.len())
```
djc: Nit: import the header names directly.

src/cli/doc_server.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    writeln!(
+        process.stderr().lock(),
+        "Serving docs at {url} (press Ctrl-C to stop)"
+    )?;
+    utils::open_browser(&url)?;
+
+    // Ctrl-C kills the process via the OS's default `SIGINT` behavior; we
+    // just need to block forever rather than returning immediately.
+    std::future::pending().await
```
djc: I think we can just inline the `loop` from `accept_loop()` here, instead?

src/cli/doc_server.rs
```diff
@@ -0,0 +1,144 @@
...
+                .expect("building a static response can't fail");
+            Ok(response)
+        }
+        Err(_) => Ok(not_found()),
+    }
+}
+
+/// Maps a request path onto a file under `root`, rejecting any path with a
+/// `..` or absolute component so requests can't escape `root`.
+fn resolve_within_root(root: &Path, request_path: &str) -> Option<PathBuf> {
```
djc: Nit: inline this single-caller function.

src/cli/doc_server.rs
```diff
@@ -0,0 +1,144 @@
+//! A minimal local HTTP server for `rustup doc --serve`.
```
djc: Instead of adding a separate module just for this little server, suggest to:

- Squash all the commits you've got so far into a single commit
- Add a new commit before this that moves the existing `doc()` and `man()` functions from `rustup_mode` into a new `docs` module
- Then, change this commit to append the server code to that module

djc (commented):
Thanks, this looks like a good start.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..44aea2ee92dd54dd03580de5822bd2a1b4e1ccc7)*

src/cli/doc_server.rs
Aneesh-382005: Commit 1 (233fea8) moves `doc()` and `man()` into the new `docs` module. No behavior changes.
Squashed the previous commits and there are only two commits at top now.

src/cli/rustup_mode.rs
Aneesh-382005: Done

src/cli/doc_server.rs
Aneesh-382005: Done

src/cli/doc_server.rs
Aneesh-382005: Done

src/cli/doc_server.rs
Aneesh-382005: Done

src/cli/doc_server.rs
Aneesh-382005: Done

src/cli/doc_server.rs
Aneesh-382005: Done

src/cli/docs.rs
```diff
@@ -245,3 +270,112 @@ pub(crate) async fn man(
...
+    process: &Process,
+) -> Result<()> {
+    let listener = TcpListener::bind(("127.0.0.1", 0))
+        .await
+        .context("failed to bind local documentation server")?;
+    let addr = listener
+        .local_addr()
+        .context("failed to read local address")?;
+
+    tokio::spawn(async move {
```
djc: My question is, why do we need to `spawn()` this? Seems to me like we could just end the function with this loop without spawning, in which case we would no longer need the `pending()` call.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..226afb2adc3e95b1120cab7a4b65367c37fc887c)*

src/cli/docs.rs
```diff
@@ -245,3 +270,112 @@ pub(crate) async fn man(
...
+        loop {
```
djc: Nit: suggest to wrap `root` in `Arc` before this loop.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..226afb2adc3e95b1120cab7a4b65367c37fc887c)*

src/cli/docs.rs
```diff
@@ -245,3 +270,112 @@ pub(crate) async fn man(
...
+    }
+
+    let Ok(contents) = tokio::fs::read(&path).await else {
+        return Ok(not_found());
+    };
+
+    let content_type = content_type_for(&path);
+    let response = Response::builder()
+        .status(StatusCode::OK)
+        .header(CONTENT_TYPE, content_type)
```
djc: Nit: I think we can just inline `content_type_for()` in this `header()` call.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..226afb2adc3e95b1120cab7a4b65367c37fc887c)*

src/cli/docs.rs
Aneesh-382005: You're right, there's nothing that needs to run concurrently inside this function. The spawn+pending() removal is in commit a49b9ffd.  The loop is the last thing the function does now.

src/cli/docs.rs
Aneesh-382005: Done

src/cli/docs.rs
Aneesh-382005: Done. Now we clone the pointer per request instead of the entire path.

src/cli/docs.rs
```diff
@@ -0,0 +1,376 @@
...
+//! over `http://127.0.0.1` sidesteps that restriction entirely.
+//!
+//! The server binds to `127.0.0.1` only (never exposed on the network) and
+//! rejects any request path that would resolve outside the served directory.
+
+use std::borrow::Cow;
+use std::convert::Infallible;
+use std::io::Write as _;
+use std::path::{Component, Path, PathBuf};
+use std::sync::Arc;
```
djc: Nits: forgot about this, can you group these into a single statement per top-level item? So `use std::{};` (starting in the first commit).

(And same for the other blocks.)

src/cli/docs.rs
```diff
@@ -0,0 +1,376 @@
...
+use tokio::net::TcpListener;
+use tracing::info;
+
+use crate::config::{ActiveSource, Cfg};
+use crate::dist::PartialToolchainDesc;
+use crate::dist::manifest::ComponentStatus;
+use crate::process::Process;
+use crate::toolchain::DistributableToolchain;
+use crate::utils::{self, ExitCode};
+
```
djc: Nit: drop this empty line to keep the `super` import in the same block as the `crate` imports.

src/cli/docs.rs
```diff
@@ -0,0 +1,376 @@
...
+    loop {
+        let (stream, _) = match listener.accept().await {
+            Ok(accepted) => accepted,
+            Err(err) => {
+                tracing::warn!("doc server: failed to accept connection: {err}");
+                continue;
+            }
+        };
+        let io = hyper_util::rt::TokioIo::new(stream);
+        let root = Arc::clone(&root);
```
djc: Nit: just write `root.clone()` everywhere.

src/cli/docs.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    loop {
+        let (stream, _) = match listener.accept().await {
+            Ok(accepted) => accepted,
+            Err(err) => {
+                tracing::warn!("doc server: failed to accept connection: {err}");
+                continue;
+            }
+        };
+        let io = hyper_util::rt::TokioIo::new(stream);
```
djc: Nit: import `TokioIo` at the module level.

src/cli/docs.rs
```diff
@@ -0,0 +1,376 @@
...
+        tokio::spawn(async move {
+            if let Err(err) = http1::Builder::new().serve_connection(io, svc).await {
+                tracing::warn!("doc server: connection error: {err}");
+            }
+        });
+    }
+}
+
+async fn serve(
+    req: Request<hyper::body::Incoming>,
```
djc: Nit: import `Incoming` at the module level.

djc (approved):
LGTM modulo a few more nits.

Please squash changes addressing review feedback back into the originating commit -- there should be two commits at the end.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..a49b9ffdc43f5c990f1adf844c579e017f81ea97)*

src/cli/docs.rs
Aneesh-382005: Done

src/cli/docs.rs
Aneesh-382005: Done

src/cli/docs.rs
Aneesh-382005: Done

src/cli/docs.rs
Aneesh-382005: Done

src/cli/docs.rs
Aneesh-382005: Done

Aneesh-382005:
Thanks for the review rounds, appreciate the feedback. Nits addressed, and changes have been squashed into two commits.

src/cli/docs.rs
```diff
@@ -0,0 +1,247 @@
...
+
+use anyhow::{Context, Result, anyhow};
+use clap::Args;
+use tracing::info;
+
+use crate::config::{ActiveSource, Cfg};
+use crate::dist::PartialToolchainDesc;
+use crate::dist::manifest::ComponentStatus;
+use crate::toolchain::DistributableToolchain;
+use crate::utils::{self, ExitCode};
```
rami3l: Nit: Since you are already here, please group all the imports (even in the first commit) from the same root path like so: https://rust-lang.github.io/rustup/dev-guide/coding-standards.html#imports

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..6d84ea3de7391b5d671fddcd2aa048ec26a3420e)*

src/cli/docs.rs
```diff
@@ -245,3 +276,106 @@ pub(crate) async fn man(
...
+        let (stream, _) = match listener.accept().await {
+            Ok(accepted) => accepted,
+            Err(err) => {
+                tracing::warn!("doc server: failed to accept connection: {err}");
+                continue;
+            }
+        };
+        let io = TokioIo::new(stream);
+        let root = root.clone();
+        let svc = service_fn(move |req| serve(req, root.clone()));
```
rami3l, line 321: Is this `.clone()` here necessary?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..6d84ea3de7391b5d671fddcd2aa048ec26a3420e)*

src/cli/docs.rs
```diff
@@ -245,3 +276,106 @@ pub(crate) async fn man(
...
+            Component::Normal(part) => path.push(part),
+            _ => return Ok(not_found()),
+        }
+    }
+    if path.is_dir() {
+        path.push("index.html");
+    }
+    if !path.starts_with(&root) {
+        return Ok(not_found());
+    }
```
rami3l, line 345: Consider moving this early exiting branch before the `path.push()` statement.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4986/382311277d0b4752b262bcb636e2f7368740111f..6d84ea3de7391b5d671fddcd2aa048ec26a3420e)*

rami3l (approved):

src/cli/docs.rs
Aneesh-382005: Done. Grouped in the first commit as well now. I've amended the commit

src/cli/docs.rs
Aneesh-382005: Done

src/cli/docs.rs
Aneesh-382005: Yes, both are necessary, the inner clone is needed because `service_fn`'s closure can be called multiple times.

PR #4991, opened 2026-08-03
refactor(toolchain/names)!: remove some conversion macros

rami3l:
Addresses https://github.com/rust-lang/rustup/pull/4947#discussion_r3668513015.

### DIFF
--- a/src/cli/proxy_mode.rs
+++ b/src/cli/proxy_mode.rs
@@ -1,4 +1,4 @@
-use std::{path::PathBuf, process::ExitStatus};
+use std::{path::PathBuf, process::ExitStatus, str::FromStr};
 
 use anyhow::Result;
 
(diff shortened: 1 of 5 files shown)


djc (approved):
Nice! I suggest keeping these impls with the other impls for the same type, rather than in their current location, to align with the style guide (perhaps in a separate commit).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4991/e96e2182ad9d8ad3d5f30161f45ea6ef4332e6ec..7c2271416012678ce6cd3dd75e430331af7608bc)*

rami3l:
> Nice! I suggest keeping these impls with the other impls for the same type, rather than in their current location, to align with the style guide (perhaps in a separate commit).

@djc I just checked and they seem to be in the right place (next to other impls of the same type) already; do you have any specific concern in that regard?

djc:
> > Nice! I suggest keeping these impls with the other impls for the same type, rather than in their current location, to align with the style guide (perhaps in a separate commit).
> 
> 
> 
> @djc I just checked and they seem to be in the right place (next to other impls of the same type) already; do you have any specific concern in that regard?

Nope, that sounds good!

PR #4992, opened 2026-08-03
feat(cli): add --check flag to rustup update for semantic exit codes

### LINKED ISSUE #4987: Option to indicate update via exit code

I'd like to run `rustup update stable` and get an indication of whether stable actually updated, without parsing `rustup`'s human-readable output.

### Solution you'd like

I'd like to have an option to have `rustup update` set an exit code based on whether any of the requested toolchains had an update. (This shouldn't be the default, since that might cause workflows to fail when a no-op update occurs.)

### Notes

_No response_

gkamathe:
## Summary

This adds a `--check` flag to `rustup update` that causes the command to set its exit code based on whether any toolchains were actually updated:

- `0`   — no updates were applied; everything was already up to date
- `100` — at least one toolchain was updated
- `1`   — an error occurred

Without `--check`, the behavior is unchanged (exit 0 on success regardless of whether updates occurred). This preserves backward compatibility for existing workflows.

The exit codes align with `rustup check` (#4681).

Closes #4987

## Changes

- Added `--check` flag to the `rustup update` subcommand
- Modified `update()` to track whether any toolchain's `UpdateStatus` was
  `Installed` or `Updated`, and return `ExitCode::UPDATES_AVAILABLE` (100)
  when `--check` is set and updates occurred
- Modified `update_all_channels()` to return whether any channel was updated
  alongside the existing exit code
- Added exit status documentation to `update_help()` in `help.rs`
- Added 4 integration tests covering: no updates (exit 0), with updates
  (exit 100), without --check (always exit 0), and all-channels update

## Test plan

- [ ] `cargo fmt --all --check` passes
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` passes
- [ ] `cargo test --features=test` passes
- [ ] New tests `update_check_no_updates`, `update_check_with_updates`,
      `update_without_check_always_succeeds`, and `update_check_all_channels`
      all pass

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -212,7 +212,10 @@ fn show_channel_updates(
     Ok(())
 }
 
-pub(crate) async fn update_all_channels(cfg: &Cfg<'_>, force_update: bool) -> Result<ExitCode> {
+pub(crate) async fn update_all_channels(
+    cfg: &Cfg<'_>,
+    force_update: bool,
+) -> Result<(ExitCode, bool)> {
     let profile = cfg.get_profile()?;
     let channels = cfg.list_channels()?;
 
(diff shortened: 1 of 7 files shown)


rami3l:
Hello!

I have reasons to believe that this PR is low-effort and has violated our contributing guidelines (LLM usage is explicitly banned for `E-easy` unless agreed otherwise):

https://github.com/rust-lang/rustup/blob/60ae6e6989c2ff6f700d758eb263e39cb07c2571/doc/dev-guide/src/index.md#L64-L70

As such, I'm closing it now.

PR #4996, opened 2026-08-04
fix(cli/self-update): postpone initialization of `Cfg` in `setup_mode`

rami3l:
Fixes #4744 by partially reverting https://github.com/rust-lang/rustup/commit/eabe9857d3c7479f236864731060de07205dcc84.

PR #4998, opened 2026-08-06
refactor(cli/self-update): move `process` out of `InstallOpts`

rami3l:
Follow-up of https://github.com/rust-lang/rustup/pull/4996.

Initially I thought moving `process` into `InstallOpts` is more convenient since many method calls rely on this value. However since certain operations also rely on `Cfg` which also contains `process`, this might cause confusions regarding which `process` to use. In our particular case, they are fortunately always the same value.

With this change, `process` is systematically injected when calling methods of `InstallOpts`, whether directly or through `Cfg`.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -159,7 +159,7 @@ impl InstallOpts<'_> {
             md(&mut term, msg);
             let mut customized_install = false;
             loop {
-                md(&mut term, current_install_opts(&self, process));
+                md(&mut term, self.display(process));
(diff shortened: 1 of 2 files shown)


ChrisDenton (approved):
I admit looking at the diff I don't feel too strongly about this but it makes sense on a conceptual level.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/4998/07e34ea5823a93960334944616b7b9760d39cfba..564ba7ca59c1abc4d48bd904d8b3e153eae777d5)*

PR #5000, opened 2026-08-08
Extend complete profile warning with a profile-switch suggestion

borngraced:
after adding a component (e.g. miri) on one toolchain, rustup try installing it on every other toolchain, failing with no explanation when unavailable there `rustup` error will now explain the cause and points to rustup set profile minimal.

### DIFF
--- a/src/cli/common.rs
+++ b/src/cli/common.rs
@@ -25,7 +25,9 @@ use crate::{
     utils::{self, ExitCode},
 };
 
-pub(crate) const WARN_COMPLETE_PROFILE: &str = "downloading with complete profile isn't recommended unless you are a developer of the rust language";
+pub(crate) const WARN_COMPLETE_PROFILE: &str = "\
+downloading with the `complete` profile is deprecated
+help: consider switching to the `default` profile with `rustup set profile default`";
 
 pub(crate) fn confirm(question: &str, default: bool, process: &Process) -> Result<bool> {
     write!(process.stdout().lock(), "{question} ")?;
--- a/tests/suite/cli_v2.rs
+++ b/tests/suite/cli_v2.rs
@@ -2467,7 +2467,8 @@ async fn test_warn_if_complete_profile_is_used() {
         .await
         .with_stderr(snapbox::str![[r#"
 ...
-warn: downloading with complete profile isn't recommended unless you are a developer of the rust language
+warn: downloading with the `complete` profile is deprecated
+help: consider switching to the `default` profile with `rustup set profile default`
 ...
 "#]])
         .is_err();
--- a/tests/suite/cli_inst_interactive.rs
+++ b/tests/suite/cli_inst_interactive.rs
@@ -498,7 +498,8 @@ async fn test_warn_if_complete_profile_is_used() {
         .await
         .with_stderr(snapbox::str![[r#"
 ...
-warn: downloading with complete profile isn't recommended unless you are a developer of the rust language
+warn: downloading with the `complete` profile is deprecated
+help: consider switching to the `default` profile with `rustup set profile default`
 ...
 "#]])
         .is_ok();


rami3l:
@borngraced I'm sorry, but I'm not sure to have completely understood what you are trying to address with this PR, would you mind elaborating?

borngraced:
> @borngraced I'm sorry, but I'm not sure to have completely understood what you are trying to address with this PR, would you mind elaborating?



> @borngraced I'm sorry, but I'm not sure to have completely understood what you are trying to address with this PR, would you mind elaborating?

@rami3l so after working on a codebase that requires/uses the nightly compiler(clippy codebase in my case) and switching to another codebase that depends on stable rustc, rustup tries to download the components that were previously used for the nightly toolchain hence failing with a not so clear error on what to do. e.g the current error output is this: 
```bash
error: some components are unavailable for download for channel '1.97-x86_64-unknown-linux-gnu': 'miri' for target 'x86_64-unknown-linux-gnu', 'rustc-codegen-cranelift' for target 'x86_64-unknown-linux-gnu'
```
after this PR it should be this: 
```bash
error: some components are unavailable for download for channel '1.97-x86_64-unknown-linux-gnu': 'miri' for target 'x86_64-unknown-linux-gnu', 'rustc-codegen-cranelift' for target 'x86_64-unknown-linux-gnu'

note: these were added as default components on another toolchain
help: run `rustup set profile minimal` to stop new toolchains from inheriting extra components
```

 I'm not sure if I had explicitly set my rustup profile to `complete` but yeah the error message isn't helpful at all. 
 
 I will take a proper look at the PR this week and fix the edge cases I missed out. thanks to CI

rami3l:
@borngraced Thanks for your clarification!

I think this is indeed caused by your setting the `complete` profile as the default in the override, but In our case this profile is retained for compatibility purposes only and is not supposed to be used anymore (https://github.com/rust-lang/rustup/issues/2136). Thus, it is expected that the installation on `stable` will never succeed on modern Rust toolchains.

In other words, your new error message isn't quite precise because your reported behavior is not caused by your repo-local override.

Rather, I think it'd be better if we can add deprecation warnings when the user tries to use it. Thanks again for your comprehension 🙏

rami3l:
@borngraced It seems that we already have this warning in place. Did you see it in your case, or do you think it could use more refinements?

https://github.com/rust-lang/rustup/blob/46035eb95a3db74ed0ac6d065be749a363923a9d/src/cli/rustup_mode.rs#L1075-L1077

borngraced:
> @borngraced It seems that we already have this warning in place. Did you see it in your case, or do you think it could use more refinements?
> 
> https://github.com/rust-lang/rustup/blob/46035eb95a3db74ed0ac6d065be749a363923a9d/src/cli/rustup_mode.rs#L1075-L1077



> @borngraced It seems that we already have this warning in place. Did you see it in your case, or do you think it could use more refinements?
> 
> https://github.com/rust-lang/rustup/blob/46035eb95a3db74ed0ac6d065be749a363923a9d/src/cli/rustup_mode.rs#L1075-L1077

Yeah, it does warn actually but doesn't really imply anything 
```bash
[I] borngraced@0x ~/w/tdev (main) [1]> RUSTUP_LOG=warn ~/dev/rustup/target/debug/rustup toolchain install 1.97
2026-08-10T13:17:25.615271Z  WARN rustup::cli::rustup_mode: downloading with complete profile isn't recommended unless you are a developer of the rust language
2026-08-10T13:17:30.163548Z ERROR rustup::cli::common: some components are unavailable for download for channel '1.97-x86_64-unknown-linux-gnu': 'miri' for target 'x86_64-unknown-linux-gnu', 'rustc-codegen-cranelift' for target 'x86_64-unknown-linux-gnu'
```

Maybe we could extend the warning instead. what do you think @rami3l ? 

I don't think this is bad either 
```bash
[I] borngraced@0x ~/w/tdev (main) [1]> RUSTUP_LOG=warn ~/dev/rustup/target/debug/rustup toolchain install 1.97
2026-08-10T13:17:25.615271Z  WARN rustup::cli::rustup_mode: downloading with complete profile isn't recommended unless you are a developer of the rust language
2026-08-10T13:17:30.163548Z ERROR rustup::cli::common: some components are unavailable for download for channel '1.97-x86_64-unknown-linux-gnu': 'miri' for target 'x86_64-unknown-linux-gnu', 'rustc-codegen-cranelift' for target 'x86_64-unknown-linux-gnu'

help: run `rustup set profile minimal` to stop new toolchains from inheriting extra components
```

rami3l:
@borngraced I think extending the warning to the second line suggesting `rustup set profile default` or `rustup set profile minimal` makes perfect sense. Would you mind repurposing this PR for that?

borngraced:
> @borngraced I think extending the warning to the second line suggesting `rustup set profile default` or ``rustup set profile minimal` makes perfect sense. Would you mind repurposing this PR for that?

agreed. yeah for sure

src/cli/common.rs
```diff
@@ -25,7 +25,7 @@ use crate::{
     utils::{self, ExitCode},
 };
 
-pub(crate) const WARN_COMPLETE_PROFILE: &str = "downloading with complete profile isn't recommended unless you are a developer of the rust language";
+pub(crate) const WARN_COMPLETE_PROFILE: &str = "downloading with complete profile isn't recommended unless you are a developer of the rust language. consider switching to the default profile with `rustup set profile default` or the minimal profile with `rustup set profile minimal`";
```
rami3l: Nit: I think we can be clearer and more concise about the warning. Something like:

```suggestion
pub(crate) const WARN_COMPLETE_PROFILE: &str = "downloading with the `complete` profile has been deprecated
help: consider switching to the `default` profile with `rustup set profile default`";
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5000/46035eb95a3db74ed0ac6d065be749a363923a9d..4c5bce2315354b51ce7e9f16d550396f40d9a306)*

tests/suite/cli_v2.rs
```diff
@@ -2467,7 +2467,9 @@ async fn test_warn_if_complete_profile_is_used() {
         .await
         .with_stderr(snapbox::str![[r#"
 ...
-warn: downloading with complete profile isn't recommended unless you are a developer of the rust language
+warn: downloading with the `complete` profile has been deprecated
+
```
rami3l: Nit: Remove this newline to align with the existing message style.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5000/46035eb95a3db74ed0ac6d065be749a363923a9d..851b0a49e5bf19f90378d5af52b6e478f0ee90c5)*

rami3l (approved):

djc (approved):

PR #5008, opened 2026-08-11
Drop FreeBSD dependency pin only needed for curl

djc:


rami3l:
~~https://github.com/madler/zlib/issues/1168 remains open, so I'm not sure if this can be done. Am I missing anything here?~~

Ah I see, I misread this in the scope of the backport. Since the tip libcurl dependency has been removed, then of course we can remove this.

A note to myself: this patch SHOULD NOT be backported to v1.29.

rami3l (approved):

Issue #4979, opened 2026-07-27
Broken `x86_64-linux-android` build caused by OpenSSL


rami3l:
https://github.com/rust-lang/rustup/commit/4add58e5a9bf09822b821d4918f5eac5ff029a75 has caused the following regression (first failing CI detected at https://github.com/rust-lang/rustup/actions/runs/29216493279/job/86713334132):

```
crypto/sm3/sm3-x86_64.S:218:2: error: invalid instruction mnemonic 'vsm3rnds2'
crypto/sm4/sm4-x86_64.S:56:2: error: invalid instruction mnemonic 'vsm4key4'
...
Error building OpenSSL: 'make' reported failure with exit status: 2
```

PR #4978, opened 2026-07-27
ci(docker/android): stop building OpenSSL

rami3l:
Closes #4979 by not building OpenSSL for `*-linux-android` targets, as `reqwest-rustls-tls` should be enough for the task.

Fix verified in https://github.com/rust-lang/rustup/actions/runs/31479426119/job/93740694089?pr=4978.

PR #5010, opened 2026-08-11
feat(openssl): strip unnessary openssl features

### LINKED ISSUE #5007: Trace and strip ssl algo/feature into the minimal features that rustup use.

We shouldn't build a full feature/algo set with ssl libraries, as we're just using a limited set of algo and features of openssl.
So maybe we can trace down to the minimal used set of openssl.

This would have these benefits:
1. Better binary size 
2. Less build time
3. More control over the build system, making sure that issues like https://github.com/rust-lang/rustup/issues/4979 happen less.

### Solution you'd like

Trace the API usage of openssl, finding the minimal features that rustup is using, then we apply the minimal feature set to openssl build, making this possible.

### Notes

_No response_

Cloud0310:
This fixes previous android ci/cd error, and further strips some of the openssl parts to improve build speed and size.

Also fixes #5007, allowing for future further reductions about openssl build.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -641,6 +641,7 @@ jobs:
           - armv7-unknown-linux-gnueabihf
           - aarch64-linux-android
           - aarch64-unknown-linux-gnu
+          - x86_64-linux-android           # skip-main
         include:
           - target: x86_64-unknown-linux-gnu
             run_tests: YES
@@ -1001,7 +1002,7 @@ jobs:
           - sparcv9-sun-solaris            # skip-pr skip-main
           - arm-linux-androideabi          # skip-pr skip-main
           - armv7-linux-androideabi        # skip-pr skip-main
-          - x86_64-linux-android           # skip-pr skip-main
+          - x86_64-linux-android           # skip-main
           - riscv64gc-unknown-linux-gnu    # skip-pr skip-main
           - riscv64gc-unknown-linux-musl   # skip-pr skip-main
           - loongarch64-unknown-linux-gnu  # skip-pr skip-main
--- a/ci/actions-templates/linux-builds-template.yaml
+++ b/ci/actions-templates/linux-builds-template.yaml
@@ -42,7 +42,7 @@ jobs: # skip-main skip-pr skip-stable
           - sparcv9-sun-solaris            # skip-pr skip-main
           - arm-linux-androideabi          # skip-pr skip-main
           - armv7-linux-androideabi        # skip-pr skip-main
-          - x86_64-linux-android           # skip-pr skip-main
+          - x86_64-linux-android           # skip-main
           - riscv64gc-unknown-linux-gnu    # skip-pr skip-main
           - riscv64gc-unknown-linux-musl   # skip-pr skip-main
           - loongarch64-unknown-linux-gnu  # skip-pr skip-main
--- a/ci/openssl-configure.sh
+++ b/ci/openssl-configure.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+configure=$1
+shift
+# `openssl-src` has no API for extra Configure arguments.
+exec perl "$configure" no-dtls no-sm2 no-sm3 no-sm4 no-srtp "$@"
--- a/ci/run.bash
+++ b/ci/run.bash
@@ -19,16 +19,10 @@ if [ -n "$INSTALL_BINDGEN" ]; then
   export PATH="$CARGO_HOME/bin/bindgen-cli:$PATH"
 fi
 
-FEATURES=('--no-default-features')
-case "$TARGET" in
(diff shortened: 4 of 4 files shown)


djc:
#4979 was already fixed, and this seems to add a bunch of complexity. Not sure it's worth it?

rami3l:
@djc I just discussed with @Cloud0310 and we agree that:

- https://github.com/rust-lang/rustup/pull/4978 is kind of a special-case hack for Android.
- Actually the root cause of #4979 is the newly-added SM3/SM4 support in OpenSSL which is not actually used anywhere in this project, however our CI is carrying the burden of that update.
  - This might thus indicate that there is a set of features we can disable permanently in the vendored OpenSSL build and, if the OpenSSL binding is updated further to build even more stuff than we care about, we can disable those new features in a similar fashion.

Cloud0310:
> #4979 was already fixed, and this seems to add a bunch of complexity. Not sure it's worth it?

To me this is better in two ways:
1. Linux artifact size

Comparing the binary size main run with the PR run, excluding Android targets:

| Linux target | Main | PR | Reduction |
|---|---:|---:|---:|
| `aarch64-unknown-linux-gnu` | 7.64 MB | 7.10 MB | **0.54 MB (7.1%)** |
| `armv7-unknown-linux-gnueabihf` | 6.40 MB | 6.36 MB | **0.04 MB (0.6%)** |
| `x86_64-unknown-linux-gnu` | 7.44 MB | 7.41 MB | **0.03 MB (0.4%)** |

The largest reduction is for `aarch64-unknown-linux-gnu`.

Android build is also fixed with this PR.

2. Making openssl upgrade more error prone.

> this seems to add a bunch of complexity

It's due to my local config problem, I used format on save, causing the change bloat, actually it's just less then 10 lines.


And as for the potential problem of stripping out wrong binary section, I've verified the `x86_64-unknown-linux-gnu` binary works after the patch manualy within a fresh debian container locally.
Another point is if openssl build fails in the future, we can have a direct way of controling build configure to fix this directly on our side.

djc:
I don't consider a 7%/0.5MB win for adding shell script calling into OpenSSL configuration worth it.

djc:
> 2\. Making openssl upgrade more error prone.

Did you mean less error prone? How/why would it be less error prone?

Cloud0310:
> > 2. Making openssl upgrade more error prone.
> 
> Did you mean less error prone? How/why would it be less error prone?

With a hack like this, we can even strip down further to the minimal stable feature set of openssl, as long as the specific feature stays buildable, we can make sure later versions won't affect or ci build.

djc:
> > > 2. Making openssl upgrade more error prone.
> > 
> > Did you mean less error prone? How/why would it be less error prone?
> 
> With a hack like this, we can even strip down further to the minimal stable feature set of openssl, as long as the specific feature stays buildable, we can make sure later versions won't affect or ci build.

You're adding a hack. We have to maintain that hack potentially forever. It doesn't seem like a good trade-off to me. In my mind we want to maintain OpenSSL support as little as possible.

rami3l:
> In my mind we want to maintain OpenSSL support as little as possible.

@djc I 100% agree with this stance, but according to my experience maintaining the CI, it looks like:

- Either we will have to remove OpenSSL builds entirely, or at least from our official builds, at which point both this PR and #4978 will lose their meaning;
- Or issues like this will reoccur even if we lower the tier of support to "guaranteed to build only", and it's always been a whack-a-mole game since a few years ago for me: each time our OpenSSL version is bumped and it breaks something, I have to either downgrade/pin the OpenSSL version in order to make the release pipeline pass, or add a hack to achieve the same result. I am not aware of the best way to lower our maintenance burden in that case, maybe we can just stick to the old ways...

In any way, I think hacking on this repo is not working on the right abstraction level; as such, I think this PR can be closed now.

djc:
@cuviper do you know people who could help keep OpenSSL support working in rustup?

cuviper:
I can probably help when it comes to linux-gnu targets, at least, although I haven't seen what's been challenging for you. For this case with android, I don't have much expertise to add, but don't we also build openssl-src for cargo in rust-lang/rust dist CI?

rami3l:
> I can probably help when it comes to linux-gnu targets, at least, although I haven't seen what's been challenging for you. For this case with android, I don't have much expertise to add, but don't we also build openssl-src for cargo in rust-lang/rust dist CI?

@cuviper I believe this Android target is a bit particular in that it is a leftover of a future host that never got promoted; in that case I would prefer just disabling it in a later version.

However when it comes to `cargo` it happens quite often that @weihanglo and I are applying the same downgrading/pinning actions etc. on both sides 😓

rami3l:
Closing: we are trying to land a fix too far from the upstream, this fix may be a better fit in OpenSSL or its Rust bindings.

PR #5013, opened 2026-08-13
style: use `oxfmt` as the frontend formatter

### LINKED ISSUE #4091: Consider introducing an HTML and CSS formatter for the Rustup website

As dicsusued in https://github.com/rust-lang/rustup/pull/4080#issuecomment-2481060171 and https://github.com/rust-lang/rustup/pull/4080#issuecomment-2485247199. 

I think it would be a good idea to introduce a formatter for this. It would help us avoid similar situations in the future. I’d recommend using Prettier, but it does require adding a new toolchain, such as npm, pnpm, or yarn, to install it. What are your thoughts on this?

### Solution you'd like

1. Install the prettier
2. Format the project using the default configuration.

### Notes

_No response_

rami3l:
Closes #4091.

## Open Questions

Should we use this for:

- [x] JSON files?
- [x] Markdown files?
- [ ] ~~TOML files?~~ (To be discussed separately)


.github/workflows/ci.yaml
```diff
@@ -1551,6 +1551,9 @@ jobs:
         with:
           tool: taplo
           fallback: none
+      - name: Run frontend formatting checks
+        run: |
+          npx oxfmt@latest .
```
djc: Is there a way to do this that doesn't involve npm tooling?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5013/82a0191173cc43c9732baa28c2c017c06d510466..41806dd42cc8127f6172413afca806905ec5c897)*

.prettierignore
```diff
@@ -0,0 +1,3 @@
+.github/workflows/ci.yaml
```
djc: Seems (a) weird this is called `.prettierignore` -- doesn't Oxfmt have its own ignore file name thingy? -- and (b) you're ignoring `ci.yaml` here but there are a bunch of `ci.yaml` changes.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5013/82a0191173cc43c9732baa28c2c017c06d510466..41806dd42cc8127f6172413afca806905ec5c897)*

rami3l: As for the point b) this is because the CI file is already generated and there should be no point in reformatting a generated file.

OTOH the template files are themselves reformatted so it's natural for the generated file to differ in terms of whitespace :)

rami3l: @djc As for the point a) I've replaced it with `.oxfmtrc.json`.

PS: `oxfmt` only supports `.gitignore` or `.prettierignore` as ignore files. Alternatively a pattern list can be configured directly in `.oxfmtrc.json`. My original rationale of using `.prettierignore` is that if we are not using the TOML support, we have essentially the same output as prettier, just faster.

.github/workflows/ci.yaml
rami3l: @djc After some initial investigation, using `npx` seems to be the fastest (and shortest?) way without extracting from the prebuilt artifacts from [here](https://github.com/oxc-project/oxc/releases#release-apps_v1.78.0).

After all, we are talking about frontend stuff here, so not using `npm` is a bit unconventional!

However I can offer to mention `oxfmt` in https://github.com/taiki-e/install-action so it can be installed directly alongside `taplo`.

rami3l: Made https://github.com/taiki-e/install-action/pull/1967.

rami3l: Merged the above PR and using the action right now. The final CI wall time overhead is less than +1sec.

rami3l:
@djc I guess we can merge it now?

README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build Status][actions-default-branch]][actions-default-branch-status]
 
-*Rustup* installs [The Rust Programming Language][rustlang] from the official
+_Rustup_ installs [The Rust Programming Language][rustlang] from the official
```
ChrisDenton, line 5: This is merely a question, not a blocker, but why `_` instead if `*`? Is that an arbitrary choice or is there a reason?

ChrisDenton (approved):
It does look prettier.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5013/fa5fb274d498005be244fe3b6c141dd56968472f..7613144cd710f85cfbcb8e9d5cbdd3a67315c790)*

README.md
rami3l: @ChrisDenton The rationale can be seen [here](https://github.com/prettier/prettier/discussions/18741#discussioncomment-15773557):

> underscores have stricter flanking rules (they don't work mid-word like foo_bar_baz), so they're "safer" as the default. prettier's printer source only switches to * when _ would break (adjacent words, nested emphasis, etc.).

I think it's been `prettier`'s style since the first day it supported Markdown (https://github.com/prettier/prettier/pull/2943) and inherited by `oxfmt` as a drop-in replacement.

PR #5016, opened 2026-08-15
tests: remove `XDG_CONFIG_HOME` from env

poliorcetics:
A test calling `setup_test_state` inherits the environment of `cargo test`, which is the environment of the calling developer's shell.

Most variables are either already caught by the setup function or ignored by rustup alltogether, `XDG_CONFIG_HOME` was not, leading to a leaked file after running an affected test:

```bash
$ cat /Users/alexis/.config/fish/conf.d/rustup.fish
source "/Users/alexis/.local/cache/target-dirs/cargo/rustup/tests/running-test-d8M8Jq/rustup-cargoMfCGsJ/ch/env.fish"
```

so when opening a new shell after that:

```
source: Error encountered while sourcing file “/Users/alexis/.local/cache/target-dirs/cargo/rustup/tests/running-test-d8M8Jq/rustup-cargoMfCGsJ/ch/env.fish”:
source: No such file or directory
```

### DIFF
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -798,6 +798,8 @@ async fn setup_test_state(test_dist_dir: TempDir) -> (TempDir, Config) {
         // trait-controllable, but it does honour the terminal. To avoid testing
         // claps code, lie about whatever terminal this process was started under.
         env::set_var("TERM", "dumb");
+        // Removed to avoid leaking the developer's environment into the test
+        env::remove_var("XDG_CONFIG_HOME");
 
         match env::var("RUSTUP_BACKTRACE") {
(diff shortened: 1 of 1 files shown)


rami3l (approved):
Thanks for pointing out the loophole and submitting the patch!

@Cloud0310 You may want to redirect this override to another value when we add XDG support (or you can ignore it if you want to go with `RUSTUP_CONFIG_HOME` for the general case).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5016/82a0191173cc43c9732baa28c2c017c06d510466..7f6cc6678730cb4f909598b934d664caefa24251)*

PR #5018, opened 2026-08-17
Fix has_local to not search PATH for external 'local' command

### LINKED ISSUE #5009: Flaky has_local in rustup-init.sh
### Verification

### Problem

On a system without `local`, it would continue searching in `$PATH`. If a such a command were found, that would wrongly be used throughout.

### Steps

1. Login to where `/bin/sh` has no builtin `local`
2. Create an executable `local` in `$PATH`

### Possible Solution(s)

I don’t know how portable an empty path is, nor whether prepending it will affect searching for the following command on such a Shell. This should probably work anywhere:
```sh
has_local() {
    # shellcheck disable=SC2034  # deliberately unused
    PATH= local _has_local
}
```

### Notes

_No response_

### Rustup version

```console
Reading latest script on GitHub.
```

### Installed toolchains

```console
Irrelevant, since I don’t have a Shell with no builtin `local`
```

### OS version

```console
Irrelevant, since I don’t have a Shell with no builtin `local`
```

harriiinnii:
### Problem

`has_local()` in `rustup-init.sh` is used to detect whether the shell has a builtin `local` keyword. On shells that lack it, the function falls through and the script aliases `local` to `typeset`.

However, if the shell has no builtin `local` *and* an executable named `local` happens to exist somewhere in `$PATH`, that external command would be found and used instead — making `has_local` return success incorrectly. Any subsequent `local` usage would then invoke that external command rather than the `typeset` alias.

### Fix

Set `PATH=` before the `local` call inside `has_local`. This prevents the shell from searching `$PATH` for a `local` executable: on shells with a builtin `local` the call succeeds as before, and on shells without it the call correctly fails even if an external `local` exists in `$PATH`.

```sh
has_local() {
    # shellcheck disable=SC2034  # deliberately unused
    PATH= local _has_local
}
```

Fixes #5009

### DIFF
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -14,6 +14,7 @@
 # mksh has this alias by default.
 has_local() {
     # shellcheck disable=SC2034  # deliberately unused
+    # shellcheck disable=SC1007  # PATH= is an intentional command prefix, not a bare assignment
     PATH= local _has_local
 }
 
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -14,7 +14,7 @@
 # mksh has this alias by default.
 has_local() {
     # shellcheck disable=SC2034  # deliberately unused
-    local _has_local
+    PATH= local _has_local
 }
 
 has_local 2>/dev/null || alias local=typeset


rami3l:
@harriiinnii Oops, it seems like your change is not a pure refactoring step? Would you mind doing some more digging?

rustup-init.sh
```diff
@@ -14,7 +14,7 @@
 # mksh has this alias by default.
 has_local() {
     # shellcheck disable=SC2034  # deliberately unused
-    local _has_local
+    PATH= local _has_local
```
rami3l, line 18: Maybe try `(PATH= local _has_local)`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5018/ba7844fae0b0b670adbbaa2f6ce8e81da06602aa..73ac8e1044335d343188defc70a1c3a44980b50d)*

harriiinnii:
Thanks for the review, @rami3l. A couple of things I want to address:

**The `PATH=` change is intentional, not cosmetic.** On shells without a builtin `local` (e.g. early Busybox `ash`), if an executable named `local` exists somewhere in `$PATH`, the current code would find it via PATH lookup and incorrectly return success from `has_local`. Setting `PATH=` before the call prevents that false positive — the function correctly fails, and the `alias local=typeset` fallback fires as intended. So yes, it does change behaviour, but only for the pathological case where `local` is a PATH-shadowing external command. That's exactly the bug described in #5009.

**Accidental file mode change**: I noticed this commit also changed `rustup-init.sh` from mode 100755 to 100644 — that was unintentional (likely a git config issue on my end). I can't fix the mode via the GitHub web API, but the easiest remediation is:

```sh
git checkout --theirs rustup-init.sh   # restore mode from main
git add rustup-init.sh
git commit --amend --no-edit
git push --force-with-lease
```

Or I can re-open the PR from a clean branch if you prefer. Let me know how you'd like to proceed.

rami3l:
@harriiinnii You are not supposed to be copy-pasting from an LLM chat which is in direct conflict with our contribution guidelines.

daniel-pfeiffer:
Conflicting with the guidelines is one thing. Offering a correct fix is the other. IMHO the latter weighs more!

Not clear why [rami3l suggests](https://github.com/rust-lang/rustup/pull/5018#discussion_r3822706732) suggests adding parens, without any motivation. Is it just a misunderstanding of this Shell syntax? Or are there Shells that have problems with this common syntax?

rami3l:
> Not clear why [rami3l suggests](https://github.com/rust-lang/rustup/pull/5018#discussion_r3822706732) suggests adding parens, without any motivation. Is it just a misunderstanding of this Shell syntax? Or are there Shells that have problems with this common syntax?

@daniel-pfeiffer The motivation is that by constraining the path change in a subshell, the command availability check will continue to work in `dash`. You can try the below example by adding parentheses on the line in question and it changes the output:

```sh
#!/bin/sh

PATH=/usr/bin:/bin
export PATH

has_local() {
    # TODO: Add parens to the following line
    PATH= local _has_local
}

has_local 2>/dev/null || alias local=typeset

echo "PATH=$PATH"
echo "curl=$(command -v curl 2>/dev/null || echo NOT_FOUND)"

check_cmd() { command -v "$1" >/dev/null 2>&1; }
check_cmd curl || check_cmd wget || echo "error: need 'curl or wget' (command not found)"
```

https://tio.run/##XY5BTwIxEIXv8yvGshH2QBo4YpbExBhO4AHPm9qOlthtSdsFjfDbV1rMsnEOc3jzvvdGiaC7bnTH33aWBw3w8rhdVbwNPimLtIC@9s5HTBcALUJtnBRmUuIP4GUygVnDuj/DeeDF@ZIrOnDbGoOnEwqzE@GKVPF7T4EiAEntkOW0Im32p8jWm6qYSNc0wiqcHjAp/yOzd73Z1s@b1/VTyQCkJvlZy0alT3GAs2LGcEDPl/ezBzzfgGvBJfSmHD8o9jWMvHd@gZZI4Tibnc@WMfZ/Whfx3bVWlazrfgE

daniel-pfeiffer:
My dash version does not have this bug. But the man-page suggests that older versions would in some ways not have been conformant:

> The current version of dash is in the process of being changed to conform with the POSIX 1003.2 and 1003.2a specifications for the shell.

Anyway, except for making this minimally slower, the parens should not hurt anywhere. Fine with me, to adapt this! OTOH, there might be Shells, that in a sub-Shell forget about being inside a function, making a normal `local` illegal. To be on the safe side, instead make a copy of `"$PATH"` and restore it afterwards!

So will this be reopened?

rami3l:
@daniel-pfeiffer Please feel free to make another PR and I'll review it.

rami3l:
> OTOH, there might be Shells, that in a sub-Shell forget about being inside a function, making a normal `local` illegal. To be on the safe side, instead make a copy of `"$PATH"` and restore it afterwards!

@daniel-pfeiffer I don't think that would be necessary because this script has already considered the case where `local` is illegal, no need to optimize for cases we don't support. Or can you think of a counterexample?

PR #5020, opened 2026-08-18
test(cli/rustup): test sad path about `RUST_RECURSION_COUNT`

### LINKED ISSUE #4983: Adding test for `Toolchain::set_env()`'s behavior during recursive `rustup` calls

The `Toolchain::set_env()` function is a providing state sync upon cargo plugin calling rustup recursively.

However, when looking at the tests related to this mechanism, I found no direct test that sets `RUST_RECURSION_COUNT` to 20/21 and asserts either guard’s error. Thus normal nested recursion has integration coverage; the maximum-count termination behavior does not appear to have explicit coverage.

After I discussed with @rami3l, I think this should be tracked here.

### Solution you'd like

Adding tests specifically triggering the maxium recursion crash, preventing further problems.

### Notes

_No response_

rami3l:
Closes #4983 via adding an extra case to `recursive_cargo()`.

Also cc @Cloud0310 for potential review.

### DIFF
--- a/tests/suite/cli_rustup.rs
+++ b/tests/suite/cli_rustup.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use rustup::{
+    env_var::RUST_RECURSION_COUNT_MAX,
     for_host,
     test::{
         CROSS_ARCH1, CROSS_ARCH2, CliTestContext, MULTI_ARCH1, Scenario, this_host_tuple,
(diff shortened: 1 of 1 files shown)


tests/suite/cli_rustup.rs
```diff
@@ -710,6 +711,23 @@ async fn recursive_cargo() {
 
 "#]])
         .is_ok();
+
+    // On the other hand, if we have set the current recursion count to the maximum while doing
```
ChrisDenton:
```suggestion
    // If we have set the current recursion count to the maximum while doing
```

The "on the other hand" comment is a bit confusing since the previous comment is talking about why we use `cargo-foo` in the test.

ChrisDenton (commented):
Seems worth testing for.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5020/ba7844fae0b0b670adbbaa2f6ce8e81da06602aa..8653190aecfb7ce6480b1d9dbda3bb2e478178ec)*

ChrisDenton (approved):

Cloud0310:
lgtm? Thanks @rami3l for helping me out.

Issue #5011, opened 2026-08-11
Provide a built-in way for rustup to adapt to a CI environment


ChrisDenton:
Discussed at the rust all-hands, it was suggested that it'd be good if rustup had better support for CI scenarios so it's less necessary for people to rely on third parties to provide them. This issue is intended to explore the existing options and see what we can adopt into rustup itself and how we can do so.

The most current github actions for rustup are [rust-toolchain](https://github.com/dtolnay/rust-toolchain/blob/master/action.yml) and [setup-rust-toolchain](https://github.com/actions-rust-lang/setup-rust-toolchain/blob/main/action.yml).

## What the actions do

- sets environment variables by default such as:
  - `CARGO_UNSTABLE_SPARSE_REGISTRY=true`
  - `CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse` (or `=git` for older cargo)
  - `CARGO_INCREMENTAL=0`
  - `CARGO_PROFILE_DEV_DEBUG=0`
  - `RUST_BACKTRACE=short`
  - `CARGO_TERM_COLOR=always`
- runs `rustup toolchain install` to install the current default toolchain
- Uses `--profile minimal` and `--no-self-update` when installing specified toolchains
- provides [problem matchers](https://github.com/actions-rust-lang/setup-rust-toolchain/blob/main/rust.json) for cargo and rustfmt issues

They can also provide caching but that is deemed out of scope for rustup.

ChrisDenton:
I guess the first big question is, if we want to do this, how do we want to go about setting environment variables? We could set them whenever `cargo` is called (but only if they're not set in the environment). Or we could have a command that outputs them to stdout and let the user redirect that as needed (e.g. github actions uses `$GITHUB_ENV`).

ChrisDenton:
> Uses `--profile minimal` and `--no-self-update` when installing specified toolchains

We've already disabled self-update in CI for a long time now: https://github.com/rust-lang/rustup/blob/8e6dbd213acfaa457a0bae0390688719d1f38ac3/src/cli/self_update.rs#L482-L486 So the only thing to do on this point would be to default to `--profile minimal`. Though that could technically break people (I'm unsure what the actual impact would be).

djc:
I would guess that's going to be fairly breaking.

ChrisDenton:
Yeah. We could provide an opt-in command like `rustup setup CI` or something like that so we can make such breaking changes.

rami3l:
[...]

PR #5021, opened 2026-08-18
feat(cli/rustup-mode): add `rustup ci` subcommand

rami3l:
Part of #5011.

This first version has implemented a few main subcommands:

- ~~`rustup ci env`: prints a hardcoded `base.env` taken from https://github.com/rust-lang/rustup/issues/5011#issue-5122912890.~~
    - Retreated due to @epage's [concerns](https://github.com/rust-lang/rustup/issues/5011#issuecomment-5418097548).

- `rustup ci problem-matcher github`: prints a hardcoded problem matcher configuration for GitHub[^1].

[^1]: Taken from @r7kamura's solution (https://github.com/r7kamura/rust-problem-matchers/blob/7e46a19e8d2f3a44f97b551aee14a7b3cc18fcd0/.github/matchers.json) since it seems to be the most comprehensive of all.

### DIFF
--- a/src/cli/ci/matcher/github.json
+++ b/src/cli/ci/matcher/github.json
@@ -1,0 +1,44 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "rust-compiler",
+      "pattern": [
+        {
+          "regexp": "^(?:\\x1B\\[[0-9;]*[a-zA-Z])*(warning|warn|error)(\\[(\\S*)\\])?(?:\\x1B\\[[0-9;]*[a-zA-Z])*: (.*?)(?:\\x1B\\[[0-9;]*[a-zA-Z])*$",
+          "severity": 1,
+          "message": 4,
+          "code": 3
+        },
+        {
+          "regexp": "^(?:\\x1B\\[[0-9;]*[a-zA-Z])*\\s+(?:\\x1B\\[[0-9;]*[a-zA-Z])*-->\\s(?:\\x1B\\[[0-9;]*[a-zA-Z])*(\\S+):(\\d+):(\\d+)(?:\\x1B\\[[0-9;]*[a-zA-Z])*$",
+          "file": 1,
+          "line": 2,
+          "column": 3
+        }
+      ]
+    },
+    {
+      "owner": "rust-formatter",
+      "pattern": [
+        {
+          "regexp": "^(Diff in (\\S+)) at line (\\d+):",
+          "message": 1,
+          "file": 2,
+          "line": 3
+        }
+      ]
+    },
+    {
+      "owner": "rust-panic",
+      "pattern": [
+        {
+          "regexp": "^.*panicked\\s+at\\s+'(.*)',\\s+(.*):(\\d+):(\\d+)$",
+          "message": 1,
+          "file": 2,
+          "line": 3,
+          "column": 4
+        }
+      ]
+    }
+  ]
+}
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -32,6 +32,7 @@ use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
 
 use crate::{
     cli::{
+        ci,
         common::{self, PackageUpdate, update_console_filter},
         docs,
         errors::CliError,
@@ -304,6 +305,13 @@ enum RustupSubcmd {
         #[arg(default_value = "rustup")]
         command: CompletionCommand,
     },
+
+    /// Generate CI-specific configurations
+    #[command(hide = true)]
+    Ci {
+        #[command(subcommand)]
+        subcmd: CiSubcmd,
+    },
 }
 
 fn update_toolchain_value_parser(s: &str) -> Result<PartialToolchainDesc> {
@@ -326,6 +334,7 @@ impl RustupSubcmd {
             // These subcommands don't require the active toolchain, so auto-installing it should be
             // disabled to avoid surprises.
             Self::Check { .. }
+            | Self::Ci { .. }
             | Self::Completions { .. }
             | Self::Component { .. }
             | Self::Default { .. }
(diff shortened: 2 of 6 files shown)


src/cli/ci/base.env
```diff
@@ -0,0 +1,6 @@
+CARGO_INCREMENTAL=0
+CARGO_PROFILE_DEV_DEBUG=0
+CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+CARGO_TERM_COLOR=always
+CARGO_UNSTABLE_SPARSE_REGISTRY=true
```
rami3l: As per https://github.com/actions-rust-lang/setup-rust-toolchain/blob/8439c15d249cd2e2a5c80ad3da200c3d8c553ae4/action.yml#L145-L149 this will be a no-op on cargo versions that don't support this flag.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5021/ba7844fae0b0b670adbbaa2f6ce8e81da06602aa..e6806036fddee5707378ec28fe9986f5e79c3275)*

src/cli/ci/base.env
```diff
@@ -0,0 +1,6 @@
+CARGO_INCREMENTAL=0
+CARGO_PROFILE_DEV_DEBUG=0
+CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
```
rami3l: @weihanglo Is it safe to always set this flag similarly to https://github.com/rust-lang/rustup/pull/5021#discussion_r3804702576?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5021/ba7844fae0b0b670adbbaa2f6ce8e81da06602aa..e6806036fddee5707378ec28fe9986f5e79c3275)*

r7kamura:
Feel free to use my solution for whatever you need👍 
It actually came from the issue and PR below, so check 'em out if you're interested:
- https://github.com/r7kamura/rust-problem-matchers/issues/8
- https://github.com/r7kamura/rust-problem-matchers/pull/11

rami3l:
@r7kamura Thanks a lot for the extra context! I've made you the co-author for this commit as a sign of appreciation.

src/cli/rustup_mode.rs
```diff
@@ -344,6 +353,7 @@ impl RustupSubcmd {
             // For all other subcommands, the hint may be useful if rustup is still unusable after
             // the command has completed.
             Self::Check { .. }
+            | Self::Ci { .. }
```
FranciscoTGouveia: I believe that this new sub-command should belong to the `false` arm instead.

If I am not overlooking anything, it seems that one can run `rustup ci` without any installed toolchain.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5021/531b7442447329ab64e3c0ffa8d8b72f33cc4af8..5917940d285c2c0d38d2a7a8a466981ffe3701b3)*

FranciscoTGouveia (approved):

src/cli/ci/base.env
weihanglo: For incremental I think so. Cargo shipped a similar thing recently: https://github.com/rust-lang/cargo/pull/17220

For profile dev debug, Cargo will have a new debug profile and we planned to turn off debug for dev profile:
<https://github.com/rust-lang/cargo/pull/17214>. So, the direction looks pretty aligned.


For registry protocol, we may want to test if there is any version treating it as a hard error if not recognized. I believe no but worth a double check.

weihanglo: probably could test pre-1.61 1.61 to 1.68, and post 1.68

src/cli/ci/base.env
```diff
@@ -0,0 +1,6 @@
...
+CARGO_TERM_COLOR=always
```
weihanglo: Will this command be used outside GitHub Actions? I know that GitHub Actions is good with ANSI color.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5021/531b7442447329ab64e3c0ffa8d8b72f33cc4af8..5917940d285c2c0d38d2a7a8a466981ffe3701b3)*

src/cli/rustup_mode.rs
rami3l: @FranciscoTGouveia Good catch!

Actually, however, it depends on how https://github.com/rust-lang/rustup/pull/5021/changes#r3843602435 should be addressed so it's still a bit hard to say... If it involves determining the active `cargo` version then everything changes.

src/cli/ci/base.env
rami3l: @weihanglo So far our first step is to support GitHub.

However I've also seen requests/examples regarding:
- GitLab: https://github.com/rust-lang/cargo/issues/16969
- Gitea (including Codeberg etc.): https://git.raphaelforment.fr/BuboBubo/Cagire/src/commit/539aa6a9f7b2532896801797443200adba5ac9a5/.gitea/workflows/build-cross.yml#L7-L8

So I believe this is quite universal.

weihanglo:
FWIW, I created this thread for t-cargo to take a look, in case others have some ideas/concerns for the cargo part.

[#t-cargo > &#96;rustup ci&#96; environment variables](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/.60rustup.20ci.60.20environment.20variables/with/619142714)

src/cli/ci/matcher/github.json
```diff
@@ -0,0 +1,44 @@
+{
```
rami3l, line 1: We should add some mock tests to ensure that the matchers are more or less correct here.

> [r7kamura/rust-problem-matchers](https://github.com/r7kamura/rust-problem-matchers) doesn't have a test suite but I'm looking at something like what GitHub does in [setup-go](https://github.com/actions/setup-go/blob/fba3fb4ead8c8e09a06f1d32ae5436e5818eff15/__tests__/setup-go.test.ts#L659-L681) where you basically write a small interpreter and verify at least the regex declarations are correct.
_https://github.com/rust-lang/rustup/issues/5011#issuecomment-5426322275_

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5021/54cbe77c2949d421573796414922722fced8c254..30e6d552c0cdc28da05a29400580c5b382aedb26)*

epage: > We should add some mock tests

How do we ensure these still work as Cargo evolves?

PR #5023, opened 2026-08-19
fix(dist): fix unexpected downgrades without `--allow-downgrade`

### LINKED ISSUE #5003: `rustup update` should not downgrade toolchains
### Verification

### Problem

`rustup up` does not validate remote version against local version.

I found out this because the fastest mirror (RUSTUP_DIST_SERVER) was temporarily down and I switched to another one, which unfortunately had an outdated version 1.96.

When I ran `rustup up`, it *unconditionally* removed my local version 1.97.1 and installed the old versin 1.96.

### Steps

N/A

### Possible Solution(s)

_No response_

### Notes

_No response_

### Rustup version

```console
```

### Installed toolchains

```console
Default host: i686-pc-windows-gnu
rustup home:  C:\Programs\rust\rustup

installed toolchains
--------------------
stable-i686-pc-windows-gnu (active, default)
stable-x86_64-pc-windows-gnu

active toolchain
----------------
name: stable-i686-pc-windows-gnu
installed targets:
  i686-pc-windows-gnu
```

### OS version

```console
Windows 11 25H2 version 10.0.26200.8973
```

rami3l:
Closes #5003.

Normally, when a toolchain is updated, the manifest date of the toolchain to be installed is compared to the current manifest date, so that `--allow-downgrade` can reject or allow potential downgrades accordingly.

However, there is a bug in the existing logic in that the  comparison is performed only _after_ the target toolchain is installed, which has effectively blocked all effects of `--allow-downgrade` _until the backtracking[^1] has started_, which is unfortunately never the case except for certain `nightly` builds. This seems to be the root cause of #5003.

[^1]: Rustup backtracks to older manifests to meet all component requirements. This is necessary because certain nightlies are allowed to not have the complete set of components.

This PR fixes this issue by comparing the manifest dates before the target toolchain is installed.

### DIFF
--- a/src/dist/manifestation/tests.rs
+++ b/src/dist/manifestation/tests.rs
@@ -17,7 +17,7 @@ use url::Url;
 use crate::{
     config::Cfg,
     dist::{
-        DEFAULT_DIST_SERVER, Profile, TargetTuple, ToolchainDesc,
+        DEFAULT_DIST_SERVER, DistOptions, Profile, TargetTuple, ToolchainDesc,
         download::{DownloadCfg, DownloadTracker},
         manifest::{Component, Manifest},
         manifestation::{Changes, Manifestation, UpdateStatus},
@@ -1534,25 +1534,13 @@ async fn v2_manifest_checksum_mismatch_surfaces_error() {
 
     let tp = TestProcess::new(env::current_dir().unwrap(), &["rustup"], vars, "");
     let cfg = Cfg::from_env(tp.process.current_dir().unwrap(), false, true, &tp.process).unwrap();
-    let dl_cfg = DownloadCfg::new(&cfg);
-    let update_hash = cfg.get_hash_file(&cx.toolchain, true).unwrap();
-    let mut fetched = String::new();
 
-    let err = super::super::try_update_from_dist_(
-        &dl_cfg,
-        &update_hash,
-        &cx.toolchain,
-        Some(Profile::Default),
-        &cx.prefix,
-        false,
-        &[],
-        &[],
-        &mut fetched,
-        &cfg,
-        None,
-    )
-    .await
-    .unwrap_err();
+    let mut fetched = String::new();
+    let err = DistOptions::new(&[], &[], &cx.toolchain, Profile::Default, false, &cfg)
+        .unwrap()
+        .try_update(None, &cx.prefix, &mut fetched, None)
+        .await
+        .unwrap_err();
 
     match err.downcast_ref::<RustupError>() {
         Some(RustupError::ChecksumFailed { .. }) => {}
--- a/src/dist/manifestation/tests.rs
+++ b/src/dist/manifestation/tests.rs
@@ -1536,9 +1536,13 @@ async fn v2_manifest_checksum_mismatch_surfaces_error() {
     let cfg = Cfg::from_env(tp.process.current_dir().unwrap(), false, true, &tp.process).unwrap();
 
     let mut fetched = String::new();
-    let err = DistOptions::new(&[], &[], &cx.toolchain, Profile::Default, false, &cfg)
-        .unwrap()
-        .try_update(None, &cx.prefix, &mut fetched, None)
+    let dist_opts =
+        DistOptions::new(&[], &[], &cx.toolchain, Profile::Default, false, &cfg).unwrap();
+    let manifest_result = dist_opts
+        .dl_v2_manifest(&cx.prefix, dist_opts.toolchain)
+        .await;
+    let err = dist_opts
+        .try_update(None, &cx.prefix, &mut fetched, manifest_result)
         .await
         .unwrap_err();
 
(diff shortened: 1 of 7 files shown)


src/dist/mod.rs
```diff
@@ -1004,6 +1004,21 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
 
         let mut toolchain = self.toolchain.clone();
         let res = loop {
+            let toolchain_date = toolchain.date.as_ref();
```
rami3l: This won't work completely yet because when we do `rustup install nightly` this `nightly` still doesn't have a date. This will be stored in `fetched` for comparison, plus we cannot store the date directly in `mut toolchain` because it won't play with `stable` etc. I have to rearrange the relevant logic to make this work.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5023/fa5fb274d498005be244fe3b6c141dd56968472f..d32943e7536baa4767131421891f93c63c13656a)*

src/dist/manifestation/tests.rs
```diff
@@ -17,7 +17,7 @@ use url::Url;
 use crate::{
```
rami3l, line 17: Note: As a follow-up, we could allow passing `--allow-downgrade` to `rustup update` (currently the flag is always false).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5023/fa5fb274d498005be244fe3b6c141dd56968472f..7bd32518e890b3fffc34a3e0fe3ccc2385ee57fe)*

src/dist/mod.rs
```diff
@@ -1265,6 +1253,25 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
 
         result
     }
+
+    async fn dl_v2_manifest(
```
rami3l: This could be merged with `DistributableToolchain` under a common trait.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5023/fa5fb274d498005be244fe3b6c141dd56968472f..7bd32518e890b3fffc34a3e0fe3ccc2385ee57fe)*

rami3l:
@djc Would you mind having a look at this? Many thanks in advance 🙏

tests/suite/cli_v2.rs
```diff
@@ -2619,6 +2619,58 @@ error: component 'rls' for target '[HOST_TUPLE]' is unavailable for download for
     cx.config.expect_component_executable("rls").await;
 }
 
+#[tokio::test]
+async fn update_allow_downgrade() {
```
djc, line 2623: Does that mean this test fail after this commit? If so, suggest putting `#[ignore]` on this (and removing that later in the fixing commit).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5023/9b191c419fa83adab2bed4665797808edccb3711..6fddede55ddcc830e261fa7bf34cb8c0f40a2042)*

src/dist/mod.rs
```diff
@@ -1006,6 +1006,30 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
...
+            } else if let Ok(Some(ManifestWithHash { manifest, .. })) = &manifest_result {
+                Some(&manifest.date)
+            } else {
+                None
+            };
+            let try_date = try_date_str
+                .and_then(|s| date_from_manifest_date(s))
+                // Parsing error can only be a problem if we enter backtracking, so we delay the
+                // error until we actually need to try another date.
+                .ok_or_else(|| format!("malformed manifest date: {try_date_str:?}"));
```
djc: Suggest this will read better as a `match` expression. Also suggest to use vertical whitespace to separate these blocks...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5023/9b191c419fa83adab2bed4665797808edccb3711..6fddede55ddcc830e261fa7bf34cb8c0f40a2042)*

src/dist/mod.rs
```diff
@@ -1006,6 +1006,30 @@ impl<'cfg, 'a> DistOptions<'cfg, 'a> {
...
 
+            // Skip installation when the toolchain to be installed comes with a valid manifest date
+            // that is older than the oldest acceptable date.
+            let try_date_str = if let Some(date) = &toolchain.date {
+                Some(date)
+            } else if let Ok(Some(ManifestWithHash { manifest, .. })) = &manifest_result {
+                Some(&manifest.date)
+            } else {
+                None
+            };
```
djc, line 1012: I think this will be easier to follow as a `match`:

```rust
let try_date_str = match (&toolchain.date, &manifest_result) {
     (Some(date), _) => Some(date),
     (_, Ok(Some(..) => ..
};
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5023/9b191c419fa83adab2bed4665797808edccb3711..6fddede55ddcc830e261fa7bf34cb8c0f40a2042)*

djc (approved):

tests/suite/cli_v2.rs
rami3l: TLDR: The fix caused a snapshot change rather than a test failure.

In this case the existing bug has caused an unexpected downgrade even if the downgrade flag is set to false, so the toolchain version in the current snapshot isn't the expected one.

In the "fix" commit, the snapshot will contain the correct toolchain version.

Issue #5024, opened 2026-08-19
[MSYS2] Allow default host triple to be overridden by an environment variable


pgalbraith:
### Problem you are trying to solve

MSYS2 on Windows provides multiple development environments which can share the same `$HOME` and therefore the same rustup configuration, but which require different Rust host ABIs. For example:

```text
native Windows → x86_64-pc-windows-msvc
UCRT64         → x86_64-pc-windows-gnu
CLANG64        → x86_64-pc-windows-gnullvm
```

The appropriate host is therefore a property of the current development environment rather than of the user, machine, or project.

At the same time, a project may want to pin its Rust version without pinning its host ABI:

```toml
[toolchain]
channel = "1.97"
```

Ideally, the same project would then resolve that partial toolchain specification according to the active environment:

```text
native Windows → 1.97-x86_64-pc-windows-msvc
UCRT64         → 1.97-x86_64-pc-windows-gnu
CLANG64        → 1.97-x86_64-pc-windows-gnullvm
```

Rustup already supports this distinction conceptually: an unqualified toolchain such as `1.97` is completed using rustup's default host. However, the default host is persisted in rustup's configuration, so environments sharing the same rustup home cannot have different defaults.

`RUSTUP_TOOLCHAIN` does not solve the problem because it selects the complete toolchain and overrides `rust-toolchain.toml`. Using it to select the ABI would therefore also override the project's choice of Rust version.

Separate `RUSTUP_HOME`s can provide separate persisted default hosts, but that makes the host choice a property of separate rustup installations. For environments such as MSYS2, the host choice is naturally a property of the currently active shell environment, while the environments otherwise share the same home directory, projects, and installed tooling.

### Solution you'd like

Add an environment variable, perhaps `RUSTUP_DEFAULT_HOST`, which overrides rustup's configured default host when resolving a toolchain specification that does not contain an explicit host.

For example:

```bash
export RUSTUP_DEFAULT_HOST=x86_64-pc-windows-gnullvm
```

combined with:

```toml
[toolchain]
channel = "1.97"
```

would resolve to:

```text
1.97-x86_64-pc-windows-gnullvm
```

This would allow environments such as MSYS2 UCRT64 and CLANG64 to set their appropriate default host during environment initialization while continuing to share the same `RUSTUP_HOME`.

The environment variable would affect only resolution of the omitted host component. It would not select or override the channel/version as `RUSTUP_TOOLCHAIN` does.

Conceptually, host resolution could follow:

```text
[...]

PR #5026, opened 2026-08-20
feat(config): allow the default host to be overridden with `RUSTUP_DE…

pgalbraith:
Delivers the suggested override in #5024.

Allow overriding the default host from env var RUSTUP_DEFAULT_HOST.  Allows for env-based global toolchain (msvc, gnu, gnullvm) selection while still respecting project-level channel selection.  Preferable to setting RUSTUP_TOOLCHAIN at the env level because that would pin channel/version across all projects.

This also documents the new env var and also updates 'rustup show' to unambiguously explain how the current default host has been derived.


pgalbraith:
Withdrawing per discussion on linked issue: this may not be an issue at all.

Issue #3254, opened 2023-03-06
Error running musl rustup on Ubuntu


fenhl:
### Problem

```
$ rustc +stable-x86_64-unknown-linux-musl --version
Error loading shared library libgcc_s.so.1: No such file or directory (needed by /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so)
Error loading shared library libgcc_s.so.1: No such file or directory (needed by /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/libstd-86aefecbddda356d.so)
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_GetRegionStart: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_GetTextRelBase: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_RaiseException: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_SetIP: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_Resume: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_DeleteException: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_Resume_or_Rethrow: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_GetIPInfo: symbol not found
Error relocating /home/fenhl/.rustup/toolchains/stable-x86_64-unknown-linux-musl/lib/librustc_driver-a19ffab541701154.so: _Unwind_GetDataRelBase: symbol not found
[...]

PR #5029, opened 2026-08-22
docs(user-guide/cross-compilation): revamp descriptions

rami3l:
It looks like many newcomers to Rust are still having difficulties understanding the distinction between the host platform and the compilation targets. Examples:

- You SHOULDN'T install a `x86_64-unknown-linux-musl` toolchain on a glibc host (like Ubuntu): https://github.com/rust-lang/rustup/issues/3254
- You DON'T NEED TO install a `x86_64-pc-windows-gnullvm` toolchain on a x64 Windows host to build for that target: https://github.com/rust-lang/rustup/issues/5024

This patch aims to further clarify that point.

cc @pgalbraith for review.

### DIFF
--- a/doc/user-guide/src/cross-compilation.md
+++ b/doc/user-guide/src/cross-compilation.md
@@ -1,16 +1,44 @@
-# Cross-compilation
+# Targets and cross-compilation
 
-Rust [supports a great number of platforms][p]. For many of these platforms
-The Rust Project publishes binary releases of the standard library, and for
+Rust [supports a great number of platforms][p]. For many of these platforms,
+the Rust Project publishes binary releases of the standard library, and for
 some the full compiler. `rustup` gives easy access to all of them.
 
 [p]: https://doc.rust-lang.org/nightly/rustc/platform-support.html
 
-When you first install a toolchain, `rustup` installs only the standard
-library for your _host_ platform - that is, the architecture and operating
-system you are presently running. To compile to other platforms you must
-install other _target_ platforms. This is done with the `rustup target add`
-command. For example, to add the Android target:
+## Targets
+
+Before going further, it is important to know the distinction between a
+toolchain's _host platform_ and the _compilation targets_ the toolchain supports.
+
+For example, if you develop on Linux and have installed the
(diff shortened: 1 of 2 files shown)


FranciscoTGouveia (approved):
Thanks, this looks great!

While looking through this, I decided to test whether the completions would work for adding a new target -- which I think is very useful -- but unfortunately it was not possible.

Could it have something to do with my setup, or do we currently not support completions for target names?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5029/831470a7345dd1cb5d77b2efb48e5e4c13c527f4..8e9bc1d5580093e9a5196681dec4661126a2693e)*

rami3l:
> Thanks, this looks great!
> 
> While looking through this, I decided to test whether the completions would work for adding a new target -- which I think is very useful -- but unfortunately it was not possible.
> 
> Could it have something to do with my setup, or do we currently not support completions for target names?

@FranciscoTGouveia No I don't think dynamic completions are supported, like at all. It's a relatively new clap feature and no one had the time to integrate it, probably until now. It would be nice to make an issue and, start working on it a bit if you have the time, or you can leave the implementation to me if you have other priorities :)

doc/user-guide/src/cross-compilation.md
```diff
@@ -1,43 +1,78 @@
...
+toolchain's _host platform_ and the _compilation targets_ the toolchain supports.
+
+For example, if you develop on Linux and have installed the
+`stable-x86_64-unknown-linux-gnu` toolchain, `x86_64-unknown-linux-gnu` is that
+toolchain's _host tuple_. When this toolchain is first installed, it can
+already compile to that same target, because it comes with the _target_ support
+for the host tuple preinstalled.
+
+It should be noted, however, that using _the same toolchain_, you can also
+compile for various other _compilation targets_, such as:
```
rami3l: compile to various other _targets_

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5029/831470a7345dd1cb5d77b2efb48e5e4c13c527f4..8e9bc1d5580093e9a5196681dec4661126a2693e)*

PR #5033, opened 2026-08-24
Use clap dynamic completion for `+toolchain`

Cloud0310:
Related to #2268.

Verified manually (see https://github.com/rust-lang/rustup/pull/5033#discussion_r3845227817).

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -683,7 +683,7 @@ pub async fn main(
         }
         Err(err) if err.kind() == DisplayVersion => {
             write!(process.stdout().lock(), "{}", err.render().ansi())?;
-            display_version(current_dir, process).await?;
+            display_version(&mut Cfg::from_env(current_dir, true, false, process)?).await?;
             return Ok(ExitCode::SUCCESS);
         }
         Err(err) => {
@@ -1835,9 +1835,8 @@ fn output_completion_script(
     Ok(ExitCode::SUCCESS)
 }
 
-async fn display_version(current_dir: PathBuf, process: &Process) -> Result<()> {
+async fn display_version(cfg: &mut Cfg<'_>) -> Result<()> {
     info!("this is the version for the rustup toolchain manager, not the rustc compiler");
-    let mut cfg = Cfg::from_env(current_dir, true, false, process)?;
     cfg.toolchain_override = cfg
         .process
         .args()
@@ -1845,7 +1844,7 @@ async fn display_version(current_dir: PathBuf, process: &Process) -> Result<()> 
         .transpose()?;
 
     match cfg.maybe_ensure_active_toolchain(None).await {
-        Ok(Some((name, _))) => match Toolchain::new(&cfg, name) {
+        Ok(Some((name, _))) => match Toolchain::new(cfg, name) {
             Ok(tc) => info!(
                 "the currently active `rustc` version is `{}`",
                 tc.rustc_version()
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1,6 +1,7 @@
 use std::{
     borrow::Cow,
     env::consts::EXE_SUFFIX,
+    ffi::OsStr,
     fmt,
     io::{self, Write},
     path::{Path, PathBuf},
@@ -18,7 +19,10 @@ use clap::{
     builder::{PossibleValue, ValueHint},
 };
 use clap_cargo::style::{CONTEXT, ERROR, GOOD, HEADER, TRANSIENT, WARN};
-use clap_complete::Shell;
+use clap_complete::{
+    Shell,
+    engine::{ArgValueCompleter, CompletionCandidate},
+};
 use futures_util::stream::StreamExt;
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use itertools::Itertools;
@@ -673,6 +677,11 @@ pub async fn main(
     console_filter: Handle<EnvFilter, Registry>,
 ) -> Result<ExitCode> {
     let cfg = &mut Cfg::from_env(current_dir, true, false, process)?;
+    clap_complete::CompleteEnv::with_factory(|| completion_command(cfg))
+        .var("RUSTUP_COMPLETE")
+        .bin("rustup")
+        .complete();
+
     self_update::cleanup_self_updater(process)?;
 
     use clap::error::ErrorKind::*;
@@ -865,6 +874,26 @@ pub async fn main(
     }
 
     Ok(exit_code)
+}
+
+fn completion_command(cfg: &Cfg<'_>) -> clap::Command {
+    let toolchains = cfg.list_toolchains().unwrap_or_default();
+    Rustup::command().mut_arg("+toolchain", move |arg| {
+        arg.add(ArgValueCompleter::new(move |current: &OsStr| {
+            let Some(prefix) = current.to_str() else {
+                return Vec::new();
+            };
+            toolchains
+                .iter()
+                .filter_map(|toolchain| {
+                    let candidate = format!("+{toolchain}");
+                    candidate
+                        .starts_with(prefix)
+                        .then(|| CompletionCandidate::new(candidate))
+                })
+                .collect()
+        }))
+    })
 }
 
 async fn default_(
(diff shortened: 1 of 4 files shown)


src/bin/rustup-init.rs
```diff
@@ -37,6 +37,14 @@ fn main() -> Result<ExitCode> {
...
 
     let process = Process::os();
+    if process.name().as_deref() == Some("rustup") {
+        let completion_process = process.clone();
+        clap_complete::CompleteEnv::with_factory(move || {
+            rustup_mode::completion_command(completion_process.clone())
+        })
+        .bin("rustup")
+        .complete();
+    }
```
rami3l: It looks like this snippet should belong to `rustup_mode::main()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..90d4fe8f924e53aab1c1d3f92e54e777de1ffbc9)*

src/cli/rustup_mode.rs
```diff
@@ -1799,12 +1837,22 @@ fn output_completion_script(
...
     match command {
         CompletionCommand::Rustup => {
-            clap_complete::generate(
-                shell,
-                &mut Rustup::command(),
+            let shells = clap_complete::env::Shells::builtins();
+            let shell_name = shell.to_string();
+            let completer = shells.completer(&shell_name).ok_or_else(|| {
+                anyhow!(
+                    "{} does not currently support completions for {}",
```
rami3l: Nit: Use new-style format.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..90d4fe8f924e53aab1c1d3f92e54e777de1ffbc9)*

src/cli/rustup_mode.rs
```diff
@@ -123,6 +126,41 @@ fn plus_toolchain_value_parser(s: &str) -> clap::error::Result<ResolvableToolcha
     }
 }
 
+pub fn completion_command(process: Process) -> clap::Command {
```
rami3l: Question: What is the expected ownership model here? I have the feeling that repeatedly cloning `Process` is definitely not the best solution here. Also, move this below the new caller in https://github.com/rust-lang/rustup/pull/5033/changes#r3844899779.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..30aaa6fa796cec178eaef2cefea71ba113a92173)*

tests/suite/cli_misc.rs
```diff
@@ -1245,10 +1245,13 @@ async fn nightly_backtrack_skips_missing() {
 #[tokio::test]
 async fn completion_rustup() {
```
rami3l, line 1246: Besides, since this is a problem that you have personally encountered, did you test the generated script on your own machine?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..30aaa6fa796cec178eaef2cefea71ba113a92173)*

src/cli/rustup_mode.rs
```diff
@@ -123,6 +126,41 @@ fn plus_toolchain_value_parser(s: &str) -> clap::error::Result<ResolvableToolcha
...
+            let Some(prefix) = current
+                .to_str()
+                .and_then(|current| current.strip_prefix('+'))
+            else {
+                return Vec::new();
+            };
+            let Some(toolchains) = process
+                .current_dir()
+                .ok()
+                .and_then(|current_dir| Cfg::from_env(current_dir, true, false, &process).ok())
```
rami3l: It looks like this callback should rely on an existing `Cfg` and use its `.process`, instead of creating a new `Cfg` from an existing process.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..d5093d43a723a31681ed0bff79d54e35b07fc304)*

src/cli/rustup_mode.rs
```diff
@@ -1864,3 +1915,64 @@ async fn display_version(current_dir: PathBuf, process: &Process) -> Result<()>
...
+            (&["rustup", "component", ""][..], 2),
+            (&["rustup", "+custom", "component", ""][..], 3),
+        ] {
+            let candidates = complete(&process, args, index, rustup_home.path());
+            for expected in ["list", "add", "remove"] {
+                assert!(candidates.iter().any(|candidate| candidate == expected));
+            }
+            for root_only in ["install", "toolchain", "show"] {
+                assert!(!candidates.iter().any(|candidate| candidate == root_only));
+            }
```
rami3l: Use `HashSet<>` operations with appropriate error messages for better readability.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..d5093d43a723a31681ed0bff79d54e35b07fc304)*

tests/suite/cli_misc.rs
Cloud0310: Host zsh test:
`rustup +<Tab>`:
Image
`rustup component <Tab>`:
Image
Fixed, tested out with alpine + zsh within podman:
Image
`rustup +cus<Tab>` is also working properly.

src/cli/rustup_mode.rs
```diff
@@ -1796,12 +1824,18 @@ fn output_completion_script(
...
+                anyhow!("{command} does not currently support completions for {shell}")
+            })?;
+            completer.write_registration(
+                "COMPLETE",
+                "rustup",
+                "rustup",
                 "rustup",
                 &mut process.stdout().lock(),
-            );
+            )?;
```
rami3l: This works technically but I think this usage is advised against since it's probably not reliable across upgrades. Instead a fixed string for each supported string is recommended, as shown in the official clap_complete docs: https://docs.rs/clap_complete/4.6.9/clap_complete/env:

> We recommend re-sourcing your completions on upgrade. These completions work by generating shell code that calls into your_program while completing. That interface is unstable and a mismatch between the shell code and your_program may result in either invalid completions or no completions being generated. For this reason, we recommend generating the shell code anew on shell startup so that it is “self-correcting” on shell launch, rather than writing the generated completions to a file.

Let's take a step back and think about what we should output here.

rami3l (commented):
Great work!

I have one remaining concern and I guess we can merge this real quick once that is resolved.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..d36fb178f660c69d5b1401623d4af7fe63757ee8)*

src/cli/rustup_mode.rs
```diff
@@ -672,6 +676,11 @@ pub async fn main(
     process: &Process,
     console_filter: Handle<EnvFilter, Registry>,
 ) -> Result<ExitCode> {
+    let cfg = &mut Cfg::from_env(current_dir, true, false, process)?;
+    clap_complete::CompleteEnv::with_factory(|| completion_command(cfg))
+        .bin("rustup")
```
rami3l, line 682: I think following `cargo`'s example, we should name the env var `RUSTUP_COMPLETE` here:

https://github.com/rust-lang/cargo/blob/77a6fbc7cea1e3c4e7506c55de112bf9bade11bb/src/bin/cargo/main.rs#L33-L37

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..d36fb178f660c69d5b1401623d4af7fe63757ee8)*

src/cli/rustup_mode.rs
rami3l: Looking at https://github.com/rust-lang/cargo/blob/77a6fbc7cea1e3c4e7506c55de112bf9bade11bb/doc/book/src/reference/unstable.md#native-completions I think we shouldn't change anything here actually.

In other words, the two mechanisms should coexist: the `completions` subcommand should output the old script.

It is possible that when the dynamic completion is stable enough, we will be able to deprecate the `completions` subcommand.

Cloud0310: One concern is PowerShell startup performance: regenerating and piping the completion script on every launch, instead of loading a cached copy from disk, may add more than 500 ms because pipelines are relatively expensive in `pwsh`. And I've experienced this problem myself on Windows.
This trade-off may be worth mentioning in the documentation.

rami3l: @Cloud0310 Good call!

rami3l (commented):
@Cloud0310 I think we can leave the rest (what to change in `rustup completions`, what to change in the docs etc.) to future PRs.

Let's merge this first and unblock @FranciscoTGouveia!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..7102660ae61bcf694f4e71adae281024fb53c588)*

rami3l (approved):

Cloud0310:
> @Cloud0310 I think we can leave the rest (what to change in `rustup completions`, what to change in the docs etc.) to future PRs.
> 
> Let's merge this first and unblock @FranciscoTGouveia!
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5033/67485a240ef24ed5189810dd6b832c91910e69fe..7102660ae61bcf694f4e71adae281024fb53c588)_

Yes, but I guess the reasonable choice on Windows dynamic completion is to opt in finally, for unix I guess it's okay for defaulting to use dynamic completion by default. And this should be documented as well. Till' Windows has a solution for their poor pipe performance under `pwsh` (this is even slower then loading from the disk), guess defaulting to dynamic shouldn't be a valid option.

rami3l:
@Cloud0310 Sure. OTOH cargo's support for dynamic completions is also unstable so I think there is no rush in resolving this...

Cloud0310:
Sorry for the misclick.

rami3l:
> Sorry for the misclick.

@Cloud0310 No worries! I've re-enqueued the PR, it should be good now.

Issue #5035, opened 2026-08-26
(Windows) "The system cannot find the file specified. (os error 2)" when running `rustup doc` in a non-ANSI path


SnipUndercover:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

I installed `rustup` to learn the language. When running it standalone, I noticed the following remark:
> If you are new to Rust consider running `rustup doc --book` to learn Rust.

I tried it out, however it did not work. Checking other commands I found that no `rustup doc` commands work...
```
Opening docs named `book` in your browser
error: couldn't open browser: IO error: The system cannot find the file specified. (os error 2)
```

### Steps

1. Have a user with a non-ANSI character in their username (for example, `ł`)
1. *Probably important:* Change the locale encoding to UTF-8
   - Open the region control panel applet
     - Open the Run menu (<kbd>Win</kbd>+<kbd>R</kbd>)
     - Enter `intl.cpl`
     - Press <kbd>Enter</kbd>
   - Open the "Administrative" tab
   - Click "Change system locale..."
   - Check the "Beta: Use Unicode UTF-8 for worldwide language support" checkbox
   - Restart the PC
1. Download `rustup-init` and install `rustup` with default settings
1. Run `rustup doc --book`
1. Get really sad

### Possible Solution(s)

The path seems to be URL-encoded as UTF-8 and passed to `ShellExecuteW`, which doesn't understand UTF-8; only UCS-2, or `wchar_t`.

### Notes

Investigating events under Process Monitor, I found that the path `rustup` attempts to read is malformed. The expected path should contain `Michał`, not `MichaÅ`. *(The `\x82` byte is invisible under Process Monitor.)*

<img width="1920" height="245" alt="A screenshot of SysInternals' Process Monitor, showcasing various filesystem events. The path is corrupted due to an encoding mismatch." src="https://github.com/user-attachments/assets/d52a0057-db45-43eb-8054-f1cc6bab091e" />

I attempted to get `rustup` to spit out some logs, but `$env:RUSTUP_LOG=trace` did not produce anything useful. I then decided to clone `rust-lang/rustup` and poke around, getting it to spit out the path it's trying to access to see where the encoding mismatch happens.

[...]

PR #5036, opened 2026-08-27
Only create a URL when opening docs with URI fragments

ChrisDenton:
`ShellExecuteW` does not handle urls very well so we try to avoid it as much as possible.

This is only a partial workaround for #5035 but it does allow new users basic access to the docs. I can implement a fuller fix but I'd want more time to test and evaluate the trade-offs so I think this is good to go in the meantime and I think this might be a good idea even if we do have a full fix (no point making a url if we don't have to).

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -523,11 +523,21 @@ impl<'a> Toolchain<'a> {
         fragment: Option<&str>,
     ) -> anyhow::Result<()> {
         let relative = relative.as_ref();
-        let mut doc_url = Url::from_file_path(self.doc_path(relative)?)
-            .ok()
-            .with_context(|| anyhow!("invalid doc file absolute path `{}`", relative.display()))?;
-        doc_url.set_fragment(fragment);
-        utils::open_browser(doc_url.to_string())
+        let doc_path = self.doc_path(relative)?;
+        // HACK: If a fragment is needed then turn the path into a url
+        // We avoid doing this if a fragment isn't needed because
+        // `open_browser` can call `ShellExecuteW` on Windows which doesn't
+        // handle urls very well. See #5035.
+        if fragment.is_some() {
+            let mut doc_url = Url::from_file_path(&doc_path).ok().with_context(|| {
+                anyhow!("invalid doc file absolute path `{}`", relative.display())
+            })?;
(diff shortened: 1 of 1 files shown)


src/toolchain.rs
```diff
@@ -523,11 +523,21 @@ impl<'a> Toolchain<'a> {
...
-        let mut doc_url = Url::from_file_path(self.doc_path(relative)?)
-            .ok()
-            .with_context(|| anyhow!("invalid doc file absolute path `{}`", relative.display()))?;
-        doc_url.set_fragment(fragment);
-        utils::open_browser(doc_url.to_string())
+        let doc_path = self.doc_path(relative)?;
+        // If a fragment is needed then turn the path into a url
+        // We avoid doing this if a fragment isn't needed because
+        // `open_browser` can call `ShellExecuteW` which doesn't
+        // handle urls very well. See #5035.
```
rami3l: Nit: It could be better to probably use the `HACK:` prefix (like elsewhere in this codebase) for the comment string so it's clearer that it is intended to be a hack/workaround.

src/toolchain.rs
```diff
@@ -523,11 +523,21 @@ impl<'a> Toolchain<'a> {
...
         let relative = relative.as_ref();
-        let mut doc_url = Url::from_file_path(self.doc_path(relative)?)
-            .ok()
-            .with_context(|| anyhow!("invalid doc file absolute path `{}`", relative.display()))?;
-        doc_url.set_fragment(fragment);
-        utils::open_browser(doc_url.to_string())
+        let doc_path = self.doc_path(relative)?;
+        // If a fragment is needed then turn the path into a url
+        // We avoid doing this if a fragment isn't needed because
+        // `open_browser` can call `ShellExecuteW` which doesn't
```
rami3l: Also probably "call `ShellExecuteW`" -> "call `ShellExecuteW` on Windows" so we can get a more complete idea without opening the issue link.

rami3l (approved):
LGTM modulo some minor points, thanks a lot :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5036/9b191c419fa83adab2bed4665797808edccb3711..246f518b5476e08452fc88aab379e36c3180e08a)*

PR #5038, opened 2026-08-30
Support setting an "Authorization" and a "Proxy-Authorization" HTTP request headers.

### LINKED ISSUE #1343: Supports HTTPS proxy authorization

amejia1:
The "Authorization" request header is used to authenticate to rustup distribution mirrors that require authentication. The "Proxy-Authorization" request header is used to authenticate to proxies.

This change introduces the new "rustup-mock-server" and "rustup-mock-proxy" test binaries, which are used to test the proper functioning of the new support for the "Authorization" and "Proxy-Authorization" HTTP request headers. A "test-rustup-init.sh" script has also been added to test the proper functioning of the "rustup-init.sh" script with the newly supported request headers.

As part of this change, "rustup-init.sh" was also modified to support distribution servers and proxies running on the localhost using HTTP. This not only provides the benefit of being able to use the "rustup-mock-server" and "rustup-mock-proxy" to run tests, it also provides additional support for custom servers and proxies which run on the localhost and forward requests to corporate servers that may require more elaborate forms of authentication, such as the usage of cookies or the usage of mTLS.

It doesn't seem like putting a disclaimer that AI agents were used is necessary. In any case, at least for the benefit of those that may have been using AI agents for much longer than I have, these code changes were made with the assistance of AI agents. I used a local LLM running this [Qwen3.8-27B model](https://huggingface.co/ggml-org/Qwen3.8-27B-GGUF).

This addresses https://github.com/rust-lang/rustup/issues/1343 .

I had a previous PR to offer similar functionality (see https://github.com/rust-lang/rustup/pull/4672 ). I decided to redo the code changes entirely, using a locally built mock server and proxy for the tests instead of relying on containers. I also decided to redo the changes with the aid of an AI agent.

Resolves https://github.com/rust-lang/rustup/issues/1343

### DIFF
--- a/doc/user-guide/src/environment-variables.md
+++ b/doc/user-guide/src/environment-variables.md
@@ -72,6 +72,14 @@
 - `RUSTUP_CONCURRENT_DOWNLOADS` _unstable_ (default: 2). Controls the number of
   downloads made concurrently.
 
+- `RUSTUP_AUTHORIZATION_HEADER` (default: none). Sets the `Authorization` HTTP request header
+  that will be included in all downloads from the rustup distribution server. Useful for
+  authenticated downloads when using a private package index.
+
+- `RUSTUP_PROXY_AUTHORIZATION_HEADER` (default: none). Sets the `Proxy-Authorization` HTTP request header
+  that will be included in all downloads from the rustup distribution server that are made
+  through a proxy. Useful for authenticated proxy connections.
+
 - `RUSTUP_TOOLCHAIN_SOURCE` _unstable_. Set by rustup to tell proxied tools how `RUSTUP_TOOLCHAIN` was determined. Non-rustup tools should not set this environment variable, except insofar as to mirror an earlier invocation from rustup.
 
 [directive syntax]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
--- a/rustup-init.sh
+++ b/rustup-init.sh
@@ -657,8 +657,51 @@ ignore() {
     "$@"
 }
 
+# Succeeds if $1 is an http:// URL that points at the local machine:
+# the host must be `localhost`, an IPv4 loopback address (127.0.0.0/8), or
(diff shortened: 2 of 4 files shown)


djc:
I am not excited about these changes. I think they add a lot of complexity solely for the benefit of testing pretty niche behavior. There's also no clean commit history.

amejia1:
> I am not excited about these changes. I think they add a lot of complexity solely for the benefit of testing pretty niche behavior. There's also no clean commit history.

I can start just using a single clean commit for these changes going forward which is what I was originally doing.

I did need some way to manually run the rustup-init program and rustup-init.sh script with these authorization http headers locally and verify they work as expected. With my old PR at https://github.com/rust-lang/rustup/pull/4672 I was proposing the use of certain tools inside of containers to serve a mock server and proxy and run some tests against those. In the end I thought those were even more complicated and opted to come up with some simple mock server and proxy that can be implemented and maintained along with rustup.

Also, I believe the rust dist server started rejecting any requests with authorization or proxy-authorization headers in the requests sometime while I was developing this feature several months ago. I think that I may have caused the need for that change with the dist server since early on, I would just run my modified build of rustup-init which would send these new headers. Since then, I thought it best to use a local mock server and proxy for offline testing.

amejia1:
@djc I moved out all that code for the implementation of the mock server and mock proxy. What is left is the support for the new headers and updates to rustup-init.sh so it can use servers and proxies running on the local host.

I still need some way to test these new changes without using the official rust distribution server so I have the mock server and proxy implemented [here](https://github.com/amejia1/rustup-mocks/).

PR #5041, opened 2026-08-31
refactor(config): postpone qualification of toolchain in overrides

rami3l:
Prepares for https://github.com/rust-lang/rustup/issues/2868.

This should in itself be a no-op, because any override requires qualification to be finally used.

However, by allowing unqualified toolchain names in `OverrideCfg`, it'll be easier to choose whether to qualify the toolchain when serializing it.

PR #5042, opened 2026-08-31
feat(config): add `rustup toolchain pin` subcommand

### LINKED ISSUE #2868: Command to create rust-toolchain.toml in the current directory
Apologies if this has already been requested, but I couldn't seem to find any issues describing something like it, nor any feature already built-in to `rustup` to do this. If there is, please feel free to close this issue.

Frequently, when I create a new binary project, I want to pin a specific toolchain immediately after initializing the project, e.g.
```sh
cargo new --bin some-project
cat >some-project/rust-toolchain.toml <<EOF
[toolchain]
channel = "stable"
EOF
```

However, I frequently forget the precise format of the toolchain file, and I have to search in `rustup` documentation to find an example and copy-paste that into my project. It would be nice if there was a simple way to generate a file with reasonable defaults that I could then customize, similar to the generated `Cargo.toml` created by `cargo new` or `cargo init`.

Add one more more new rustup commands to create a `rust-toolchain.toml` in the current directory. I think this could work similarly to `rustup override set` (take a toolchain argument), or just use the default / currently-set toolchain. The generated file could also include comments that briefly describe some of the fields and link to the documentation.

**Notes**

Example of one possible workflow:

```sh
cargo new --bin some-project
cd some-project
rustup init nightly . # or whatever the syntax of the command looks like
```
After which the contents of `some-project/rust-toolchain.toml` would look like this:
```toml
# See more keys and their definitions at https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file
[toolchain]
channel = "nightly"
components = []
targets = []
profile = "default"
```

Some possible command names:
 * `rustup new` (for parity with `cargo new`)
 * `rustup init` (possibly confusing with `rustup-init`)
 * `rustup generate-toolchain-file` or similar (very wordy)
 * `rustup pin`
 * A subcommand or flag to pass to `rustup override set` ?

rami3l:
Based on https://github.com/rust-lang/rustup/pull/5041 and #5044; closes #2868.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -13,7 +13,7 @@ use std::{
 
 use anstream::ColorChoice;
 use anstyle::Style;
-use anyhow::{Error, anyhow};
+use anyhow::{Context as _, Error, anyhow};
 use clap::{
     Args, CommandFactory, Parser, Subcommand, ValueEnum,
     builder::{PossibleValue, ValueHint},
@@ -26,6 +26,7 @@ use clap_complete::{
 use futures_util::stream::StreamExt;
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use itertools::Itertools;
+use serde::Serialize;
 use tokio::sync::Semaphore;
 use tracing::{error, info, warn};
 use tracing_subscriber::{EnvFilter, Registry, reload::Handle};
@@ -46,7 +47,7 @@ use crate::{
         self_update::{self, SelfUpdateMode, check_rustup_update},
     },
     command, component_for_bin,
-    config::{ActiveSource, Cfg},
+    config::{ActiveSource, Cfg, OverrideCfg, OverrideFile},
     dist::{
         DistOptions, PartialToolchainDesc, Profile, Switch, TargetTuple,
         download::DownloadCfg,
@@ -430,6 +431,13 @@ enum ToolchainSubcmd {
         /// Path to the directory
         path: PathBuf,
     },
+
+    /// Write the active toolchain as an override file
+    Pin {
+        /// Write the toolchain name with host tuple
+        #[arg(long)]
+        qualified: bool,
+    },
 }
 
 #[derive(Debug, Default, Args)]
@@ -781,6 +789,7 @@ pub async fn main(
                 toolchain_link(cfg, &toolchain, &path).await
             }
             ToolchainSubcmd::Uninstall { opts } => toolchain_remove(cfg, opts).await,
+            ToolchainSubcmd::Pin { qualified } => pin_active_toolchain(qualified, cfg),
         },
         RustupSubcmd::Check { opts } => check_updates(cfg, opts).await,
         RustupSubcmd::Default {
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -976,7 +976,11 @@ impl<'a> Cfg<'a> {
     }
 
     /// Create an override for a toolchain
-    pub(crate) fn make_override(&self, path: &Path, toolchain: &ToolchainName) -> Result<()> {
+    pub(crate) fn make_override(
```
rami3l, line 1025: We should also add extra test cases for this change like we did in #4947.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5042/92d7f059f1fd5cd767d0133948764d2423030928..00126b6a02bf41bc451cb63de71e9d192669a640)*

src/config.rs
```diff
@@ -39,23 +42,60 @@ enum OverrideFileConfigError {
...
 }
 
-#[derive(Debug, Default, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
 struct OverrideFile {
     toolchain: ToolchainSection,
 }
 
 impl OverrideFile {
+    fn from_partial_desc(
```
djc: I suggest to instead have a `From` impl for `ResolvableToolchainName`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5042/924059437670f4d583dd557e2afd971282dfe515..c9cf68f81a29e8d45cabc95b15b6c9db43b73e85)*

src/config.rs
```diff
@@ -766,6 +834,49 @@ impl<'a> Cfg<'a> {
         }
     }
 
+    pub fn pin_active_toolchain(&self, qualified: bool) -> Result<()> {
```
djc: I think this code should live somewhere in `cli`, not in `config`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5042/924059437670f4d583dd557e2afd971282dfe515..c9cf68f81a29e8d45cabc95b15b6c9db43b73e85)*

src/config.rs
```diff
@@ -39,23 +42,60 @@ enum OverrideFileConfigError {
...
 struct OverrideFile {
     toolchain: ToolchainSection,
 }
 
 impl OverrideFile {
+    fn from_partial_desc(
+        desc: PartialToolchainDesc,
+        qualified: bool,
+        cfg: &Cfg<'_>,
+    ) -> Result<Self> {
```
djc: FWIW (lots of pre-existing instances), I'm not a fan of bare `Result<Self>` especially when `Result` is not a local alias. IMO it would be better to convert bare single-argument `Result` to `anyhow::Result` everywhere.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5042/924059437670f4d583dd557e2afd971282dfe515..c9cf68f81a29e8d45cabc95b15b6c9db43b73e85)*

rami3l: @djc I think this change is out of the scope of the current PR. I could offer to do a cleanup afterwards though.

djc: I was just suggesting to not introduce new instances of (what I consider to be an) anti-pattern.

But if you want to clean them up separately that's fine too!

rami3l: @djc Using `anyhow::Result<>` here will make clippy complain because it has found a shorter form of the path which is.... `Result<>`. So it's not possible here, unfortunately.

rami3l: Update: I have made https://github.com/rust-lang/rustup/pull/5052. Maybe this PR can be rebased on top of that.

src/cli/rustup_mode.rs
```diff
@@ -1714,6 +1723,70 @@ async fn toolchain_remove(cfg: &Cfg<'_>, opts: UninstallOpts) -> Result<ExitCode
     Ok(ExitCode::SUCCESS)
 }
 
+fn pin_active_toolchain(qualified: bool, cfg: &Cfg<'_>) -> Result<ExitCode> {
+    let default_host = cfg.default_host_tuple()?;
+
+    let (mut override_, backfill_components) = match cfg.find_override_config()? {
```
djc: Nit: why the trailing `_` in `override_`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5042/924059437670f4d583dd557e2afd971282dfe515..6c0883a17edda6c604d96acf5f277cecc07f4b3b)*

src/cli/rustup_mode.rs
```diff
@@ -1714,6 +1723,70 @@ async fn toolchain_remove(cfg: &Cfg<'_>, opts: UninstallOpts) -> Result<ExitCode
...
+        Some((o, _)) => (o, None),
+        None => {
+            let default_ = cfg
```
djc: Same here, why the trailing `_`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5042/924059437670f4d583dd557e2afd971282dfe515..6c0883a17edda6c604d96acf5f277cecc07f4b3b)*

src/cli/rustup_mode.rs
```diff
@@ -1714,6 +1723,70 @@ async fn toolchain_remove(cfg: &Cfg<'_>, opts: UninstallOpts) -> Result<ExitCode
...
+                    Some(
+                        tc.components()?
+                            .into_iter()
+                            .filter_map(|c| {
+                                (c.available && c.installed)
+                                    .then(|| manifest.display_name(&c.component, &default_host))
+                            })
+                            .collect(),
+                    )
+                }
```
djc: Should this move into some method? It doesn't seem to be very specific to pinning a particular toolchain?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5042/924059437670f4d583dd557e2afd971282dfe515..6c0883a17edda6c604d96acf5f277cecc07f4b3b)*

src/cli/rustup_mode.rs
rami3l: Because `override` and `default` are Rust keywords... Or should I name them differently?

djc: Ah yes... maybe just `r#override` and `r#default`?

rami3l: @djc I've got no objections, let's do that then :)

src/cli/rustup_mode.rs
rami3l: I'm a bit confused here. My thinking process is as follows:

I think we do have a very specific logic in this case: we cannot use the generic component listing logic for `ResolvableToolchainName` because custom toolchains are also supported now, and providing the list over there makes no sense in pinning.

This narrows down the research to generalizing the method for `DistributableToolchain`. For that I've searched across many uses of `.components()` iteration and have found nothing like what we have now.

Note that in our particular case here. it should be a list of display names so as to be as friendly as possible to the user. This means we should keep the names:
- Without the default host tuple where applicable.
- "Before renaming" for easy-to-read names, e.g. `miri` instead of `miri-preview`.

Would you mind providing some more detailed suggestions in that regard?

Also, would extracting a single-use method be considered early optimization here?

Many thanks 🙏

djc (approved):

src/cli/rustup_mode.rs
djc: Fair enough, let's keep it like this for now.

rami3l:
Cancelling auto-merge because I think it'd be better to stack this on top of https://github.com/rust-lang/rustup/pull/5052.

Issue #4198, opened 2025-02-27
Failed to execute `rustup update` in the Rust `devcontainer`


rami3l:
> I get the same problem calling `rustup update` in a devcontainer. The devcontainer just installs 1.84 and I need 1.85. I get
> 
> ```
> vscode ➜ /workspaces/my_project $ rustup update
> info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
> info: latest update on 2025-02-20, rust version 1.85.0 (4d91de4e4 2025-02-17)
> info: downloading component 'clippy'
> info: downloading component 'rustfmt'
> info: downloading component 'rust-src'
> info: downloading component 'rust-analysis'
> info: downloading component 'rls'
> info: downloading component 'cargo'
> info: downloading component 'rust-std'
> info: downloading component 'rustc'
>  69.5 MiB /  69.5 MiB (100 %)  27.2 MiB/s in  2s ETA:  0s
> info: removing previous version of component 'clippy'
> info: rolling back changes
> error: could not rename component file from '/usr/local/rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/doc/clippy' to '/usr/local/rustup/tmp/58z480gzx9kr9vug_dir/bk'
> info: checking for self-update
> 
>   stable-x86_64-unknown-linux-gnu update failed - rustc 1.84.1 (e71f9a9a9 2025-01-27)
> 
> info: cleaning up downloads & tmp directories
> ``` 

 _Originally posted by @Psy-Kai in [#4181](https://github.com/rust-lang/rustup/issues/4181#issuecomment-2687638391)_

rami3l:
@psy-kai I cannot seem to be able to reproduce this on my machine. Here's what I've tried:

```console
> uname -moprsv
Darwin 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000 arm64 arm
> git clone https://github.com/microsoft/vscode-remote-try-rust && cd vscode-remote-try-rust
[..]
> devcontainer up --workspace-folder .
[..]
> devcontainer exec --workspace-folder . bash
```

Now we have entered the devcontainer.

```console
$ rustup show active-toolchain
1.83.0-aarch64-unknown-linux-gnu (default)
```

Having noticed that the default toolchain is `1.83.0`, let's pretend it's `stable` to trigger the update.

```console
[...]

Issue #2949, opened 2022-02-17
Invalid cross-device link (os error 18) when building inside pods due to OverlayFS restrictions


huguesBouvier:
### Problem

Hello!

I  developing my application on kubernetes. For development, I use minikube, I **mount my host repo** inside a pod and I build/run from inside it.
However, I have the following issue when building:
   Compiling libc v0.2.118
error: failed to write /home/azureuser/..../target/debug/deps/libunicode_xid-8bb8457dab823b6f.rmeta: Invalid cross-device link (os error 18)

error: could not compile `unicode-xid` due to previous error.

Thanks!

### Steps

Install and start minikube.
mount any rust repo:
minikube mount /repo:/repo

Create a deployment to mount the repo inside a pod.
Do cargo build inside the pod

### Rustup version

```console
rustup 1.24.3 (ce5817a94 2021-05-31)
```

### Installed toolchains

```console
rustc 1.57.0 (f1edd0429 2021-11-29)
```

rbtcollins:
I suspect that this isn't a rustup problem, sauce you are getting this error from the compiler.

What's the full path in the error?

g00nix:
I can confirm the issue. I am running rustup inside a podman container. Code to the full setup can be found [here](https://codeberg.org/gunix/nvim-rust-ide/).

```sh
~ # ❯❯❯ rustup update
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: latest update on 2022-02-24, rust version 1.59.0 (9d1b2106e 2022-02-23)
info: downloading component 'rust-std' for 'wasm32-unknown-unknown'
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
 53.8 MiB /  53.8 MiB (100 %)  30.8 MiB/s in  1s ETA:  0s
info: downloading component 'rustfmt'
info: removing previous version of component 'rust-std' for 'wasm32-unknown-unknown'
info: removing previous version of component 'cargo'
info: rolling back changes
error: could not rename component file from '/root/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/share/zsh/site-functions' to '/root/.rustup/tmp/4jpq5_2ziff3ygx8_dir/bk'
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
[...]

Issue #4291, opened 2025-04-16
Broken Cargo binaries on Windows Docker images due to symlinks


amyspark:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue+is%3Aopen%2Cclosed and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

Since 1.28.0, we have been finding subtle yet impossible to debug issues in our CI, where we get nonsensical results like this when trying to call Cargo:

```
Traceback (most recent call last):
  File "<string>", line 24, in <module>
  File "C:\Users\Administrator\runner\builds\amyspark\cerbero\cerbero\main.py", line 224, in main
    Main(sys.argv[1:])
  File "C:\Users\Administrator\runner\builds\amyspark\cerbero\cerbero\main.py", line 52, in __init__
    self.run_command()
  File "C:\Users\Administrator\runner\builds\amyspark\cerbero\cerbero\main.py", line 192, in run_command
    res = commands.run(command, self.config, self.args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Administrator\runner\builds\amyspark\cerbero\cerbero\commands\__init__.py", line 79, in run
    return _commands[command].run(config, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Administrator\runner\builds\amyspark\cerbero\cerbero\commands\bootstrap.py", line 179, in run
    run_until_complete(build_tools_task)
  File "C:\Users\Administrator\runner\builds\amyspark\cerbero\cerbero\utils\__init__.py", line 763, in run_until_complete
    result = loop.run_until_complete(tasks)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\asyncio\base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\Users\Administrator\runner\builds\amyspark\cerbero\cerbero\bootstrap\build_tools.py", line 145, in fetch_recipes
    await Fetch.fetch(self.cookbook, self.recipes, False, False, False, False, jobs)
[...]

PR #5045, opened 2026-09-01
hack(process/linux): enable `RUSTUP_PERMIT_COPY_RENAME` by default on CI

rami3l:
Addresses the concern raised in https://github.com/rust-lang/rustup/issues/5011#user-content-fn-1-d6ab2ff576a7d006325502f4a047667b. Mitigates:

- https://github.com/rust-lang/rustup/issues/4198
- https://github.com/rust-lang/rustup/issues/2949

## Background

One of the many cases where OverlayFS can break rustup's transactional semantics is when rustup is used in a CI environment conveniently placed in Docker containers, where `RUSTUP_PERMIT_COPY_RENAME=1` has been proven to be a viable workaround (https://github.com/rust-lang/rustup/issues/2949#issuecomment-2584968659, https://github.com/dtolnay/rust-toolchain/pull/177).

## Proposed Solution

This PR automatically enables `RUSTUP_PERMIT_COPY_RENAME` when `CI=1`.

The rationale is that Linux CI machines are quite common and wasting some disk spaces there should be more or less acceptable as long as more CI workflows pass. Hopefully, by limiting the hack to CI machines, this won't have a surprisingly large burst ratio.

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -475,9 +475,7 @@ pub enum SelfUpdateMode {
 
 impl SelfUpdateMode {
     pub(crate) fn from_cfg(cfg: &Cfg<'_>) -> Result<Self> {
-        if cfg.process.var("CI").is_ok() && cfg.process.var("RUSTUP_CI").is_err() {
-            // If we're in CI (but not rustup's own CI, which wants to test this stuff!),
-            // disable automatic self updates.
+        if cfg.process.is_ci() {
             return Ok(Self::Disable);
         }
 
--- a/src/process.rs
+++ b/src/process.rs
@@ -147,6 +147,11 @@ impl Process {
     #[cfg(target_os = "linux")]
     pub fn permit_copy_rename(&self) -> bool {
         match self {
+            // HACK: On Linux CI machines, rustup is sometimes installed in a Docker image and we
+            // may hit OverlayFS restrictions that prevent renaming. Thus, we allow falling back to
+            // copying to avoid this issue.
+            // See: <https://github.com/dtolnay/rust-toolchain/pull/177>
+            _ if self.is_ci() => true,
             Self::OsProcess(_) => env::var_os("RUSTUP_PERMIT_COPY_RENAME").is_some(),
             #[cfg(feature = "test")]
             Self::TestProcess(p) => p.vars.contains_key("RUSTUP_PERMIT_COPY_RENAME"),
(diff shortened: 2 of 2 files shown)


src/cli/self_update.rs
```diff
@@ -475,7 +475,7 @@ pub enum SelfUpdateMode {
 
 impl SelfUpdateMode {
     pub(crate) fn from_cfg(cfg: &Cfg<'_>) -> Result<Self> {
-        if cfg.process.var("CI").is_ok() && cfg.process.var("RUSTUP_CI").is_err() {
+        if cfg.process.is_ci() {
             // If we're in CI (but not rustup's own CI, which wants to test this stuff!),
```
ChrisDenton: This comment could be updated too. The fact that `is_ci()` ignores rustup's CI for testing is documented on the function and doesn't need repeating here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5045/54cbe77c2949d421573796414922722fced8c254..f9871e54a601e274f7a5e7a2f3a1c78e8a92585a)*

src/process.rs
```diff
@@ -134,6 +147,7 @@ impl Process {
     #[cfg(target_os = "linux")]
     pub fn permit_copy_rename(&self) -> bool {
         match self {
+            _ if self.is_ci() => true,
```
ChrisDenton, line 154: This should add a "HACK" comment to explain why we're doing this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5045/54cbe77c2949d421573796414922722fced8c254..f9871e54a601e274f7a5e7a2f3a1c78e8a92585a)*

ChrisDenton:
We should also consider doing a similar thing for `RUSTUP_HARDLINK_PROXIES` on Windows. See #4291.

rami3l:
> We should also consider doing a similar thing for `RUSTUP_HARDLINK_PROXIES` on Windows. See #4291.

@ChrisDenton Do you want me to do this in the same PR or in a subsequent one? It should be a good fit either way.

ChrisDenton (approved):
I'm fine with leaving it to a follow up.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5045/54cbe77c2949d421573796414922722fced8c254..1707932041c18454662ca2302b4a69a0ca0e5cd5)*

PR #5047, opened 2026-09-02
feat: support .env files in rustup home

intellild:
Users can add rustup only envs in `${RUSTUP_HOME}/.env` without a global shell envs.

For example, users who need rustup to access the network through an HTTP proxy can add the following to `${RUSTUP_HOME}/.env`:
```dotenv
http_proxy=http://127.0.0.1:7890
```

> ## Summary
> - Load an optional `${RUSTUP_HOME}/.env` before initializing the Tokio runtime and tracing.
> - Preserve inherited process environment variables as higher-precedence one-off overrides.
> - Ignore a missing `.env` file while reporting read and parse errors with the file path.
> - Document the configuration and test loading, precedence, and invalid-file behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -515,6 +515,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dotenvy"
+version = "0.15.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
+
+[[package]]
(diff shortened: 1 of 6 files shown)


djc:
Read the AI policy first:

https://rust-lang.github.io/rustup/dev-guide/#ai-policy

djc:
Not sure it makes sense to me to implement a feature like this on a per-tool basis? Why do you need different environment variables for rustup compared to other tooling? Have you surveyed existing tools that allow you to set per-command environment variables?

PR #5051, opened 2026-09-03
test(clitools): refine error messages of exit code assertions

rami3l:
From interactions with new contributors here I have realized that the source position in E2E tests are incorrect (from a DX perspective), and a plain `assert_eq!()` output without any explanation can be confusing at times.

So this simple patch aims to improve that experience by adding simple error messages and using the right positions:

Before:

```smalltalk
thread 'suite::cli_exact::update_once' (25194046) panicked at src/test/clitools.rs:202:9:
  left: Some(1)
 right: Some(0)
```

After:

```smalltalk
thread 'suite::cli_exact::update_once' (25191171) panicked at tests/suite/cli_exact.rs:15:10:
assertion `left == right` failed: unexpected command failure
  left: Some(1)
 right: Some(0)
```

The position anchors like `tests/suite/cli_exact.rs:15:10` should be sufficient in guiding the developer to the failed assertion with a simple click on most IDEs.

This patch is a no-op when the tests pass, but IMHO it's a good DX improvement nonetheless.

### DIFF
--- a/src/test/clitools.rs
+++ b/src/test/clitools.rs
@@ -187,19 +187,22 @@ impl Assert {
     }
 
     /// Asserts that the command exited with an ok status.
+    #[track_caller]
     pub fn is_ok(&self) -> &Self {
         self.has_code(0)
     }
 
     /// Asserts that the command exited with an error.
+    #[track_caller]
     pub fn is_err(&self) -> &Self {
-        assert_ne!(self.output.status, Some(0));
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5051/9b1cd6c92f37ff0af2a3251e9a72135b70d76bcc..89273c4671fa9f0feaca4557dae0cd075899f00c)*

PR #5053, opened 2026-09-04
ci: clean cargo caches with plain bash

rami3l:
Looks like `cargo-cache` is no longer doing any good for us because its build suddenly failed (https://github.com/rust-lang/rustup/actions/runs/33809636224/job/100828189334?pr=5052).

I've checked the implementation and found out that this is the only part we need:

```rs
        CargoCacheCommands::AutoClean { dry_run } => {
            // clean the registry sources and git checkouts
            let reg_srcs = &cargo_cache.registry_sources;
            let git_checkouts = &cargo_cache.git_checkouts;

            // depending on the size of the cache and the system (SSD, HDD...) this can take a few seconds.
            println!("Clearing cache...\n");

            for dir in &[reg_srcs, git_checkouts] {
                let size = cumulative_dir_size(dir);
                if dir.is_dir() {
                    remove_file(
                        dir,
                        dry_run,
                        &mut size_changed,
                        None,
                        &DryRunMessage::Default,
                        Some(size.dir_size),
                    );
                }
            }
```

https://github.com/matthiaskrgr/cargo-cache/blob/2f4497beeab989becd2ecdae8a42601956e822b0/src/main.rs#L355-L375

So I figured that we might as well use plain bash for this: it's just easier to audit, and should save some CI time for us.


djc (approved):
Nice!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5053/924059437670f4d583dd557e2afd971282dfe515..0fa774c690d6b5c7df392e1c5cb1bfd4cb0dc90e)*

PR #5054, opened 2026-09-04
refactor(cli/rustup-mode): avoid redundant allocs in `target_add()`

rami3l:
Follow-up on a minor pre-existing issue discovered when reviewing #5042.

This patch avoids an extra `.to_string()` allocation and a redundant `into_iter() -> collect() -> into_iter()` round trip.

### DIFF
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1478,17 +1478,17 @@ async fn target_add(
     )
     .await?;
 
-    let components = distributable.components()?;
-    if targets.contains(&"all".to_string()) {
-        if targets.len() != 1 {
-            return Err(anyhow!(
-                "`rustup target add {}` includes `all`",
-                targets.join(" ")
-            ));
-        }
+    let all = targets.iter().any(|it| it == "all");
+    if all && targets.len() != 1 {
+        return Err(anyhow!(
+            "`rustup target add {}` includes `all`",
+            targets.join(" ")
+        ));
+    }
 
+    if all {
         targets.clear();
-        for component in components {
+        for component in distributable.components()? {
             if component.component.short_name() == "rust-std"
                 && component.available
                 && !component.installed
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1504,18 +1504,13 @@ async fn target_add(
     }
 
     distributable
-        .add_components(
-            targets
-                .into_iter()
-                .map(|target| {
-                    Component::new(
-                        "rust-std".to_string(),
-                        Some(TargetTuple::new(target)),
-                        false,
-                    )
-                })
-                .collect(),
-        )
+        .add_components(targets.into_iter().map(|target| {
+            Component::new(
+                "rust-std".to_string(),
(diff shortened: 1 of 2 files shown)


src/cli/rustup_mode.rs
```diff
@@ -1468,43 +1468,26 @@ async fn target_add(
...
+                    return Err(anyhow!(
+                        "`rustup target add {}` includes `all`",
+                        targets.join(" "),
+                    ));
+                }
+                false => Either::Right(targets.into_iter().map(|t| {
+                    Component::new("rust-std".to_string(), Some(TargetTuple::new(t)), false)
+                })),
+            }
+            .collect(),
```
djc: I'd do something like this up front:

```
    let all = targets.iter().any(|it| it == "all");
    if all && targets.len() > 1 {
        return Err(anyhow!(
            "`rustup target add {}` includes `all`",
            targets.join(" "),
        ));
    }
```

Also, it looks like `add_components()` could take an `impl ExactSizeIterator` instead of a `Vec<_>`.

(You know this is just code golf and not a meaningful optimization, right?)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5054/4b08719243451d05af8090caff8e04a54e4b05e7..cb9dc23eaa26cd2b196e16836737d3ed62ed22cb)*

rami3l: Yeah I just have found the trunk code quite unreadable, and as I tried to simplify I just found out more issues...

rami3l: @djc I've minimized the diff, it should look better now?

src/toolchain/distributable.rs
```diff
@@ -65,7 +65,10 @@ impl<'a> DistributableToolchain<'a> {
         &self.desc
     }
 
-    pub(crate) async fn add_components(&self, components: Vec<Component>) -> anyhow::Result<()> {
+    pub(crate) async fn add_components(
+        &self,
+        components: impl Iterator<Item = Component>,
```
djc, line 70: Why not `ExactSizeIterator`?

djc (approved):
Nice, this seems much better.

The one other thing I played with was adding a `Component::std(target)` constructor wrapping `Component::new("rust-std".to_string(), Some(target), false)`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5054/6196da198fd5d3eca5b4ec41f05d907727a33de6..81044d61d888b50724ee56a170e18cc9d9cf0450)*

src/toolchain/distributable.rs
rami3l: Because in one of the branches (pre-existing logic) a `.filter_map()` has broken the `ExactSize` assumption.

djc: Makes sense!

rami3l:
> The one other thing I played with was adding a `Component::std(target)` constructor wrapping `Component::new("rust-std".to_string(), Some(target), false)`.

@djc Sounds pretty much like a systematic replacement. I can do this as a follow-up.

PR #5062, opened 2026-09-07
refactor(dist/manifest): extract `Component::std()`

rami3l:
> [!NOTE]
> 
> This PR is generated by GitHub Copilot on GPT 5.6 Luna, with manual review and minor adjustments. 

Addresses the concern in https://github.com/rust-lang/rustup/pull/5054#pullrequestreview-5130129288.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,9 +4,7 @@ version = "1.30.0"
 edition = "2024"
 license = "MIT OR Apache-2.0"
 description = "Manage multiple rust installations with ease"
-homepage = "https://github.com/rust-lang/rustup"
 keywords = ["rustup", "multirust", "install", "proxy"]
-readme = "README.md"
 repository = "https://github.com/rust-lang/rustup"
 build = "build.rs"
 
--- a/src/cli/rustup_mode.rs
+++ b/src/cli/rustup_mode.rs
@@ -1498,13 +1498,11 @@ async fn target_add(
     }
 
     distributable
-        .add_components(targets.into_iter().map(|target| {
-            Component::new(
-                "rust-std".to_string(),
-                Some(TargetTuple::new(target)),
-                false,
-            )
-        }))
+        .add_components(
+            targets
+                .into_iter()
+                .map(|target| Component::std(TargetTuple::new(target))),
+        )
         .await?;
 
     Ok(ExitCode::SUCCESS)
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -4,9 +4,7 @@ version = "1.30.0"
 edition = "2024"
 license = "MIT OR Apache-2.0"
 description = "Manage multiple rust installations with ease"
-homepage = "https://github.com/rust-lang/rustup"
 keywords = ["rustup", "multirust", "install", "proxy"]
-readme = "README.md"
```
rami3l, line 9: `cargo` now says that it can deduce those values and thus suggests removing them altogether.

The same thing has happened to another project of mine: https://github.com/rami3l/pacaptr/commit/19d2cd667ce9939a025dab4794978136c90aa9a9

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5062/142063987b346d75b5f2908624a9c1640118023b..e7b63dfdde724d666e2ae307e496596ce943b027)*

src/cli/rustup_mode.rs
```diff
@@ -1543,7 +1541,7 @@ async fn target_remove(
         if has_at_most_one_target {
             warn!("removing the last target; no build targets will be available");
         }
-        let new_component = Component::new("rust-std".to_string(), Some(target), false);
+        let new_component = Component::std(target);
```
djc: Suggest additionally inlining this into the call below.

Alternatively, it looks like all other calls construct their own `TargetTuple` -- so maybe this should still use `Component::new()` such that all other callers can have the `TargetTuple` constructed for them?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5062/142063987b346d75b5f2908624a9c1640118023b..e7b63dfdde724d666e2ae307e496596ce943b027)*

djc (approved):

src/cli/rustup_mode.rs
rami3l: @djc Thanks for pointing that out; implementation-wise `TargetTuple` is just a transparent newtype of `String`, but I still think they are conceptually distinct and I prefer having a `TargetTuple` rather than taking a `String` directly.

Or, balancing verbosity and clarity, maybe we can `impl From<String> for TargetTuple` so `Component::std()` can accept either `tuple` or `tuple_str.into()`.

rami3l: I've tried the alternative option I've proposed above and it feels like implicit conversions coming back (noting that in tests it's mostly `&str` rather than `String` so a conversion is needed anyway)... Maybe it's better to keep the implementation as-is.

PR #5067, opened 2026-09-09
fix(windows): match complete PATH entries when adding and removing

Cloud0310:
Fix Windows potentially adding/removing wrong PATH entry.

### DIFF
--- a/src/cli/self_update/windows.rs
+++ b/src/cli/self_update/windows.rs
@@ -10,6 +10,7 @@ use std::{
 };
 
 use anyhow::{Context, anyhow};
+use itertools::Itertools;
 use tracing::{info, warn};
 #[cfg(any(test, feature = "test"))]
 use windows_registry::Value;
@@ -531,7 +532,7 @@ fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
     if old_path.is_empty() {
         Some(path_str)
     } else if old_path
-        .windows(path_str.len())
+        .split(|&c| c == PATH_SEPARATOR)
         .any(|path| *path == *path_str)
     {
         None
@@ -543,29 +544,22 @@ fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
     }
 }
 
-// Returns None if the existing old_path does not need changing
+// Returns the updated PATH string with empty entries and all entries equal to
+// path_str removed, or None if the contents are unchanged.
 fn _remove_from_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
-    let idx = old_path
-        .windows(path_str.len())
-        .position(|path| *path == *path_str)?;
-    // If there's a trailing semicolon (likely, since we probably added one
-    // during install), include that in the substring to remove. We don't search
-    // for that to find the string, because if it's the last string in the path,
-    // there may not be.
-    let mut len = path_str.len();
-    if old_path.get(idx + path_str.len()) == Some(&(b';' as u16)) {
-        len += 1;
-    }
-
-    let mut new_path = old_path[..idx].to_owned();
-    new_path.extend_from_slice(&old_path[idx + len..]);
-    // Don't leave a trailing ; though, we don't want an empty string in the
-    // path.
-    if new_path.last() == Some(&(b';' as u16)) {
-        new_path.pop();
-    }
-    Some(HSTRING::from_wide(&new_path))
+    let new_path = Itertools::intersperse(
+        old_path
+            .split(|&c| c == PATH_SEPARATOR)
+            .filter(|path| !path.is_empty() && **path != *path_str),
+        &[PATH_SEPARATOR],
+    )
+    .flatten()
+    .copied()
+    .collect::<Vec<_>>();
+    (new_path != *old_path).then(|| HSTRING::from_wide(&new_path))
 }
+
+const PATH_SEPARATOR: u16 = b';' as u16;
 
 fn _with_path_cargo_home_bin<F>(f: F, process: &Process) -> anyhow::Result<Option<HSTRING>>
 where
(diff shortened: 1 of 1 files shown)


src/cli/self_update/windows.rs
```diff
@@ -545,23 +547,18 @@ fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
...
+    // precede an exact match. Remove only the first exact match.
+    let mut paths: Vec<&[u16]> = old_path.split(|&c| c == b';' as u16).collect();
+    let idx = paths.iter().position(|path| **path == *path_str)?;
+    paths.remove(idx);
+
+    // Joining removes the separator after the deleted entry, or the one before
+    // it if it was last. Preserve the existing cleanup of one trailing semicolon
+    // in the former case; in the latter, that separator is already gone.
+    let mut new_path = paths.join(&(b';' as u16));
+    if idx < paths.len() && new_path.ends_with(&[b';' as u16]) {
```
rami3l: How can the new path end with ';' here? If it's because the previous split may generate empty substrings, you should probably filter over there. Also probably add a test case for this as well.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5067/bf4f258102f0021f6c48370cc60e319dd5783dc4..b8b18915773522d5df63b2378c61970990f3ef28)*

src/cli/self_update/windows.rs
```diff
@@ -545,23 +547,18 @@ fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
...
+    // leaves C:\tools\bin-extra and XC:\tools\bin intact, even when they
+    // precede an exact match. Remove only the first exact match.
+    let mut paths: Vec<&[u16]> = old_path.split(|&c| c == b';' as u16).collect();
+    let idx = paths.iter().position(|path| **path == *path_str)?;
+    paths.remove(idx);
+
+    // Joining removes the separator after the deleted entry, or the one before
+    // it if it was last. Preserve the existing cleanup of one trailing semicolon
+    // in the former case; in the latter, that separator is already gone.
+    let mut new_path = paths.join(&(b';' as u16));
```
rami3l: Should `b';' as u16` be extracted as a new constant?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5067/bf4f258102f0021f6c48370cc60e319dd5783dc4..b8b18915773522d5df63b2378c61970990f3ef28)*

rami3l (approved):

src/cli/self_update/windows.rs
Cloud0310: With further investigation, it turns out that Windows allows for empty PATH entries. Yes, your concern is true.

src/cli/self_update/windows.rs
djc: This seems like a confusing way to do what needs to be done?

Could it be more like this:

```rust
let new = old_path
     .split(|&c| c == b';' as u16)
     .filter(|path| **path != &path_str)
     .join(&(b';' as u16));
```

I don't think the second comment about what the deal is with semi-colons with joining is very clear, either.

(And edge cases are better handled with tests than by adding comment coverage.)

djc (commented):
(Also it doesn't really appear like the first two commits are really independent?)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5067/bf4f258102f0021f6c48370cc60e319dd5783dc4..2acbbee13190c4df812dbd5f4129e7fe17e5d12f)*

src/cli/self_update/windows.rs
```diff
@@ -543,30 +543,19 @@ fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
...
-    if new_path.last() == Some(&(b';' as u16)) {
-        new_path.pop();
-    }
-    Some(HSTRING::from_wide(&new_path))
+    // Remove redundant empty entries and target path_str.
+    let paths: Vec<&[u16]> = old_path
+        .split(|&c| c == PATH_SEPARATOR)
+        .filter(|path| !path.is_empty() && **path != *path_str)
+        .collect();
+    let new_path = paths.join(&PATH_SEPARATOR);
```
djc: Why do we need to `collect()` before `join()`ing?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5067/bf4f258102f0021f6c48370cc60e319dd5783dc4..70c4ce7fc3bb7e681c176dbc83e932b0f3355c5d)*

src/cli/self_update/windows.rs
Cloud0310: sry for this, I was trying to use `std::env::{join_path,split_path}` and `OsString` here. And this turns out is causing extra allocations, and not aligned up with current code base's style.

src/cli/self_update/windows.rs
```diff
@@ -543,30 +543,19 @@ fn _add_to_path(old_path: HSTRING, path_str: HSTRING) -> Option<HSTRING> {
     }
 }
 
-// Returns None if the existing old_path does not need changing
+// Removes all matching entries and empty entries, returning None if unchanged.
```
rami3l: This comment needs refinement: It feels like this function is returning a collection of entries, however it is not.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5067/bf4f258102f0021f6c48370cc60e319dd5783dc4..70c4ce7fc3bb7e681c176dbc83e932b0f3355c5d)*

src/cli/self_update/windows.rs
rami3l: @djc Because `Itertools::join()` doesn't apply here (signature mismatch, can only collect to a string), and we only have `slice::join()` from std which seems like a waste in terms of allocation.

Cloud0310: Switched to [itertools::intersperse]( https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.intersperse). Is this better?

rami3l: @Cloud0310 In the codebase we use `itertools::*()` and `: Vec<u16>` rather rarely. How about:

```rs
    let new_path = old_path
        .split(|&c| c == PATH_SEPARATOR)
        .filter(|path| !path.is_empty() && **path != *path_str);
        .intersperse(&[PATH_SEPARATOR])
        .flatten()
        .copied()
        .collect::<Vec<_>>();
```

Cloud0310: Fixed, but std has nightly `intersperse` api would cause potential naming collision, so used fuller qualification here.

rami3l (approved):

PR #5068, opened 2026-09-11
fix(config): improve error handling in `Cfg::list_toolchains()`

rami3l:
This is another problem that I have noticed when working on https://github.com/rust-lang/goals/issues/736.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -983,23 +983,23 @@ impl<'a> Cfg<'a> {
     /// Currently no notification of incorrect names or entry type is done.
     #[tracing::instrument(level = "trace", skip_all)]
     pub(crate) fn list_toolchains(&self) -> anyhow::Result<Vec<ToolchainName>> {
-        if utils::is_directory(&self.toolchains_dir) {
-            let mut toolchains: Vec<_> = utils::read_dir("toolchains", &self.toolchains_dir)?
-                // TODO: this discards errors reading the directory, is that
-                // correct? could we get a short-read and report less toolchains
-                // than exist?
-                .filter_map(io::Result::ok)
-                .filter(|e| e.file_type().map(|f| !f.is_file()).unwrap_or(false))
-                .filter_map(|e| e.file_name().into_string().ok())
-                .filter_map(|n| ToolchainName::from_str(&n).ok())
-                .collect();
-
-            toolchains.sort();
-
-            Ok(toolchains)
-        } else {
-            Ok(Vec::new())
+        if !utils::is_directory(&self.toolchains_dir) {
+            return Ok(vec![]);
         }
+
+        let mut toolchains: Vec<_> = utils::read_dir("toolchains", &self.toolchains_dir)?
+            // TODO: this discards errors reading the directory, is that
+            // correct? could we get a short-read and report less toolchains
+            // than exist?
+            .filter_map(io::Result::ok)
+            .filter(|e| e.file_type().map(|f| !f.is_file()).unwrap_or(false))
+            .filter_map(|e| e.file_name().into_string().ok())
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -983,23 +982,36 @@ impl<'a> Cfg<'a> {
...
+            };
+            match ToolchainName::from_str(&name) {
+                Ok(tc) => toolchains.push(tc),
+                Err(e) => {
+                    warn!("ignoring invalid toolchain name '{name}' in toolchains dir: {e}",)
+                }
+            }
+        }
+
+        toolchains.sort();
```
rami3l: It is possible that this `sort()` could be delayed to the display logic.

Previously in 9207137dfbd45b422cb01f76eac7a27aa0ad869e the sorting logic has been moved into in this function, but that was because back then the sorting was not done with `.sort()` which means the API consumer may be quite easily sorting it wrong.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5068/7df2ba07037827dcc7e0bf7bab4bb8b8fb7b3b26..79265c213f0c77aa04498b9630b8d5fa51fe68e0)*

src/config.rs
```diff
@@ -1026,30 +1026,26 @@ impl<'a> Cfg<'a> {
             }
         }
 
-        toolchains.sort();
```
djc: If we're no longer sorting here, we could return an iterator directly?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5068/7df2ba07037827dcc7e0bf7bab4bb8b8fb7b3b26..784946fcc2658419779f9a09988d7d846e8da639)*

src/config.rs
```diff
@@ -1026,30 +1026,26 @@ impl<'a> Cfg<'a> {
...
+                _ => None,
             })
-            .filter(ToolchainDesc::is_tracking)
-            .map(|n| {
-                DistributableToolchain::new(self, n.clone())
-                    .map_err(Into::into)
-                    .map(|t| (n.clone(), t))
-            })
-            .collect::<anyhow::Result<Vec<_>>>()
+            .map(|n| Ok((n.clone(), DistributableToolchain::new(self, n)?)))
```
djc: Suggest moving this `map()` logic into the previous `filter_map()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5068/7df2ba07037827dcc7e0bf7bab4bb8b8fb7b3b26..784946fcc2658419779f9a09988d7d846e8da639)*

src/config.rs
```diff
@@ -986,13 +984,16 @@ impl<'a> Cfg<'a> {
     /// This function returns every valid toolchain name that has a corresponding non-file entry in
     /// the toolchains directory. These names may be returned in any order.
     #[tracing::instrument(level = "trace", skip_all)]
-    pub(crate) fn list_toolchains(&self, quiet: bool) -> anyhow::Result<Vec<ToolchainName>> {
+    pub(crate) fn list_toolchains(
```
djc: In the end, this doesn't feel like an improvement maybe?

(I also have a possibly irrational dislike of `Either`.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5068/5fc78f7aec6480485d06667453023ebeaa55c7fb..e09a4c9ae4bd9e77cc56073b85df753c00db0e62)*

djc (approved):

src/config.rs
rami3l: No problem, I'll revert just the final commit.

PR #5070, opened 2026-09-11
Fix Windows CI's Powershell not propagate error by default

Cloud0310:
Previous Windows GNU and MSVC target on CI/CD has different report. This is what I found when dealing with CI/CD errors on https://github.com/rust-lang/rustup/pull/5056.
This is due to the default shell for running GNU target on Windows is `pwsh`, and its not exiting upon first error, causing checking failures being ignored.

Note: bug finding assisted by codex GPT-6.

### DIFF
--- a/ci/actions-templates/windows-builds-template.yaml
+++ b/ci/actions-templates/windows-builds-template.yaml
@@ -152,17 +152,23 @@ jobs: # skip-main skip-pr skip-stable
         run: bash ci/run.bash
       - name: Run cargo check
         if: matrix.mode != 'release'
+        # `bash` on Windows uses Git's Perl, which breaks the OpenSSL build.
+        shell: pwsh
         env:
           TARGET: ${{ matrix.target }}
         # os-specific code leads to lints escaping if we only run this in one target
         run: |
+          # Abort on the first nonzero exit code so a failure isn't masked by the next command.
+          $PSNativeCommandUseErrorActionPreference = $true
           cargo check --all --all-targets --features test
           git ls-files -- '*.rs' | xargs touch
       - name: Run cargo clippy
         if: matrix.mode != 'release' && matrix.mingwdir == ''
+        shell: pwsh
         env:
           TARGET: ${{ matrix.target }}
         run: |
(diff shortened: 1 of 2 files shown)


rami3l (approved):

ci/actions-templates/windows-builds-template.yaml
```diff
@@ -161,7 +161,9 @@ jobs: # skip-main skip-pr skip-stable
           # Abort on the first nonzero exit code so a failure isn't masked by the next command.
           $PSNativeCommandUseErrorActionPreference = $true
           cargo check --all --all-targets --features test
-          git ls-files -- '*.rs' | xargs touch
+          git -c core.quotePath=false ls-files -- '*.rs' | ForEach-Object {
```
djc, line 164: Uh, why do we even need to update the last write time (aka `touch`) files here? Seems like some crappy workaround that is probably no longer needed?

rami3l: @djc Historically, this is to make sure that `cargo clippy` doesn't rely on the results of `cargo check` before it, which prevents the CI from giving the correct behavior, as reported in issues like https://github.com/rust-lang/rust-clippy/issues/2218.

rami3l: This is also why these two steps have been taking so much wall time on CI: the clippy analysis is always performed from scratch.

rami3l: Although, at this point, I'd probably prefer running the `clippy` step only, if that makes sense.

djc: Yes, those issues were a thing, but I'm pretty sure they've been fixed for years at this point.

(I do agree that we can can collapse `check` and `clippy`.)

Cloud0310: I'd like to make another PR for removing `cargo check` then.

Cloud0310: > I'd like to make another PR for removing `cargo check` then.

https://github.com/rust-lang/rustup/pull/5074

Issue #5025, opened 2026-08-20
Proposal: special toolchain names referring to particular override levels


rami3l:
Part of https://github.com/rust-lang/rustup/issues/4059.

Since https://github.com/rust-lang/rustup/pull/4932 it has been decided that only characters satisfying `GeneralSecurityProfile::identifier_allowed()` should be allowed in rustup toolchain names.

This has opened up new possibilities for us to assemble our special/reserved toolchain names as needed with special characters such as `@` or `=`.

Following up the above efforts, noting the [five levels of rustup toolchain overrides](https://rust-lang.github.io/rustup/overrides.html):

> 1. The `+<toolchain>` override.
> 2. The `RUSTUP_TOOLCHAIN` environment variable.
> 3. The `rustup override set` override.
> 4. The `rust-toolchain.toml` override.
> 5. The default toolchain.

... aiming to address https://github.com/rust-lang/rustup/issues/4391, I propose the following changes:

- Make `env` and `default` new reserved words in addition to the existing `none`.
- Introduce the following toolchain syntax:
    - `env`: Refers to the toolchain that should be activated by the env override.
        - This will not be useful per se but sometime people want to be explicit, e.g. `rustup run env cargo` instead of `cargo`.
    - `dir=<dir>`: Refers to the toolchain that should be activated by the dir override on `<dir>` or its fallback dirs.
        - e.g. `cargo +dir=./crates/download`
    - `file=<file>`: Refers to the toolchain that should be activated by the TOML override in `<file>` which follows the `rust-toolchain.toml` schema. 
        - e.g. `cargo +file=windows.toml`, for a Windows-specific `rust-toolchain.toml` setup.
    - `default`: Refers to the default toolchain. Bails out if none is set.
        - e.g. `cargo +default`

## Concerns

- How should we handle the error for `dir=` and `file=` when the dir/TOML override doesn't exist? Should we bail out or warn and fall back to the next level?
    - My guess is that we should warn and fall back.
- Consider the relation between this proposal and #4945. Should we resolve the toolchain name when the user issues `rustup default file=windows.toml`? If so, should we fully resolve or partially resolve it?
    - My guess is that we should partially resolve to an official or a custom toolchain name (that isn't a reserved word and doesn't have special characters like `=`).
[...]

PR #5071, opened 2026-09-12
feat(toolchain): support default toolchain aliases

14elias:
Related to #5025.
Related discussion: #2793.
this PR implements the **default** special toolcahin name from the proposal in #5025

this change makes **default**  a reserved toolchain alias that always resolves to the currently configured global default toolchain.
It works consistently across all supported override mechanisms:

- Command line: **+default**
- Environment variable: **RUSTUP_TOOLCHAIN=default**
- Directory overrides
- **rust-toolchain.toml** / **rust-toolchain** files

For example, when the configured default is **stable**:
- **cargo +default ...**
- **RUSTUP_TOOLCHAIN=default cargo ...**
- a toolchain file containing **channel = "default"**
all resolves to the **stable**

**Why we need this feature**
The global default has the lowest priority in rustup’s override order.  A project’s **rust-toolchain.toml** therefore completely hides it.  
This change gives users a way to say ignore the project pin and use my personal default right now with out having to hard code the actual name of the default. this change is especially useful in scripts and tools that want to follow the user’s current preference even if that preference later changes.  for example if a script is written like  **cargo +default test**  the script uses what the user's default is with out editing the script.

**Tests**
- +default resolution
- RUSTUP_TOOLCHAIN=default
- rust-toolchain.toml using channel = "default"


src/cli/rustup_mode.rs
```diff
@@ -929,7 +929,17 @@ async fn default_(
     common::warn_if_host_is_emulated(cfg.process);
 
     if let Some(toolchain) = toolchain {
-        match toolchain.to_owned() {
+        let toolchain = match toolchain.to_owned() {
+            MaybeResolvableToolchainName::Some(ResolvableToolchainName::Default) => {
+                MaybeResolvableToolchainName::Some(
+                    cfg.get_default_resolvable()?
```
rami3l: You are repeating this logic too many times to the point where it looks like a more systematic solution is required.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/7df2ba07037827dcc7e0bf7bab4bb8b8fb7b3b26..7b83b06c9c11ed5e7a57ea705a928be32a6ba3b8)*

src/cli/rustup_mode.rs
```diff
@@ -950,6 +960,9 @@ async fn default_(
 
                 common::show_channel_update(cfg, PackageUpdate::Toolchain(desc), Ok(status))?;
             }
+            MaybeResolvableToolchainName::Some(ResolvableToolchainName::Default) => {
+                unreachable!("default should have been resolved above")
```
rami3l: We'd like to prevent cases like this if possible, because it means for this type, certain values are no longer valid starting from a certain point in time.

Ideally since we are introducing a new resolution hierarchy, would it be possible to introduce a new type rather than adding a variant to the existing ones?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/7df2ba07037827dcc7e0bf7bab4bb8b8fb7b3b26..7b83b06c9c11ed5e7a57ea705a928be32a6ba3b8)*

rami3l (changes requested):

src/cli/rustup_mode.rs
14elias: i created new enum type for the alias  and a wrapper to avoid from repeating   the logic                                                                                                           let toolchain = match toolchain.to_owned() {
            MaybeResolvableToolchainName::Some(ResolvableToolchainName::Default) => {
                MaybeResolvableToolchainName::Some(
                    cfg.get_default_resolvable()? 
the wrapper enum has two variants Alias and Explicit  and it has resolve method, so the checking and toolchain dispatching happens in the resolve method.

src/cli/self_update.rs
```diff
@@ -367,6 +367,7 @@ impl InstallOpts<'_> {
                     Some(ResolvableToolchainName::Official(t)) => Some(t),
                     // Default is custom, presumably from a prior install. Do nothing.
                     Some(ResolvableToolchainName::Custom(_)) => None,
+                    // The configured default cannot itself be the default alias.
```
rami3l: I think this change is redundant now.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/toolchain/names.rs
```diff
@@ -74,6 +74,67 @@ pub enum InvalidName {
...
+        if value == "default" {
+            Ok(Self::Default)
+        } else {
+            Err(InvalidName::ToolchainName(value.into()))
+        }
+    }
+}
+
+impl Display for ToolchainAlias {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
rami3l, line 96: Suggest importing `std::fmt`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/toolchain/names.rs
```diff
@@ -74,6 +74,67 @@ pub enum InvalidName {
...
+
+impl FromStr for ToolchainAlias {
+    type Err = InvalidName;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        if value == "default" {
+            Ok(Self::Default)
+        } else {
+            Err(InvalidName::ToolchainName(value.into()))
+        }
```
rami3l: Suggest using `match` instead of `if` here so that we can easily add more variants (with exhaustiveness check).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/toolchain/names.rs
```diff
@@ -74,6 +74,67 @@ pub enum InvalidName {
     DashPrefix(String),
 }
 
+/// An alias for a toolchain name.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ToolchainAlias {
+    Default,
```
rami3l, line 81: Suggest adding a docstring explaining the precise semantics of each variant.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/toolchain/names.rs
```diff
@@ -74,6 +74,67 @@ pub enum InvalidName {
...
+}
+
+impl<T: FromStr> FromStr for ToolchainOverride<T>
+where
+    T::Err: Into<InvalidName>,
+{
+    type Err = InvalidName;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        if let Ok(candidate) = validate_name(value)
```
rami3l: This check makes no sense because it overlaps with the latter check. Note that your aliases are a predetermined set of strings.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/config.rs
```diff
@@ -669,7 +713,11 @@ impl<'a> Cfg<'a> {
                 None
             };
 
-        Ok(override_config)
+        let Some((override_cfg, source)) = override_config else {
+            return Ok(None);
+        };
+
+        Ok(Some((override_cfg, source)))
```
rami3l: Why is this change necessary?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/config.rs
```diff
@@ -735,44 +785,49 @@ impl<'a> Cfg<'a> {
...
-                            .list_toolchains(true)?
-                            .iter()
-                            .any(|s| s == resolved_name)
+                        .map_err(|_| {
+                            anyhow!(
+                                "invalid toolchain name detected in override file '{}'",
+                                toolchain_file.display()
+                            )
+                        })?;
+                    if let ToolchainOverride::Explicit(toolchain_name) = toolchain_override {
```
rami3l: I think we may not need to specialize for `Explicit` here, just resolve it as usual (note that `find_override_from_dir_walk` may apply to more than just the project repo; it can totally be a user-local configuration option).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/config.rs
```diff
@@ -25,12 +25,52 @@ use crate::{
...
+            Self::Alias(ToolchainAlias::Default) => {
+                let default = cfg
+                    .get_default_resolvable()?
+                    .ok_or_else(|| anyhow::anyhow!("no default toolchain is configured"))?;
+                Ok(MaybeResolvableToolchainName::Some(default))
+            }
+            Self::Explicit(r) => Ok(r),
+        }
+    }
+}
```
rami3l: It should be able to replace all this with a single, generic impl block with an appropriate use of the `From` trait. Suggest adding the necessary `From` implementations if needed.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/toolchain/names.rs
```diff
@@ -74,6 +74,67 @@ pub enum InvalidName {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Default => write!(f, "default"),
+        }
+    }
+}
+
+/// A wrapper for types that can be overridden by an alias.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ToolchainOverride<T> {
```
rami3l: Since the uses of this type are always of the form `ToolchainOverride<xxxToolchain>`, suggest renaming this to just `Override`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/toolchain/names.rs
```diff
@@ -74,6 +74,67 @@ pub enum InvalidName {
...
+    Alias(ToolchainAlias),
```
rami3l: Suggest using the same part of speech (preferably adjective) for all variants. In this case I'd suggest `Aliased` to mirror `Explicit`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/cli/rustup_mode.rs
```diff
@@ -924,13 +926,15 @@ fn completion_command(cfg: &Cfg<'_>) -> clap::Command {
...
+    toolchain: Option<ToolchainOverride<MaybeResolvableToolchainName>>,
     force_non_host: bool,
 ) -> anyhow::Result<ExitCode> {
     common::warn_if_host_is_emulated(cfg.process);
 
     if let Some(toolchain) = toolchain {
-        match toolchain.to_owned() {
+        let toolchain = toolchain.resolve(cfg)?;
+
+        match toolchain {
```
rami3l: Suggest inlining this temporary variable `toolchain`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/cli/proxy_mode.rs
```diff
@@ -37,10 +37,15 @@ pub async fn main(
     let cfg = Cfg::from_env(current_dir, false, true, process)?;
     let (toolchain, source) = cfg
         .local_toolchain(match toolchain {
-            Some(name) => Some((
-                name.resolve(&cfg.default_host_tuple()?)?,
-                ActiveSource::CommandLine,
-            )),
+            Some(name) => {
+                let name = name.resolve(&cfg)?;
```
rami3l: Suggest inlining this temporary variable `name`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

rami3l (changes requested):

src/cli/proxy_mode.rs
```diff
@@ -5,7 +5,7 @@ use crate::{
     command::run_command_for_dir,
```
rami3l, line 5: The category specified in the commit message is wrong. It should look more like `feat(toolchain):` or similar. Same thing for the title of this PR.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

src/toolchain/names.rs
14elias: ///Refers to rustup's configured default toolchain.
///Parsed from the literal string `"default"`.           is this docstring good at representing the variant

rami3l: @14elias I think just keeping the first line would be fine. Also could you write your comments, PR descriptions etc in proper formatted markdown? It's not super easy to read to me in its current state.

src/cli/proxy_mode.rs
rami3l, line 2: Besides, if you have used LLM/agents to draft this PR, please properly disclose the fact in the PR description according to our contribution guidelines. Otherwise, I may close this PR without any prior warnings.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/454ff04cdefebc8f38f47f64b3904866f9e0660f..de81ab2cc212db181d019eaf2c3af2de3a371671)*

14elias: sorry for the confusion or inconvenience caused, but i did not used AI agent or LLM to write the code, or draft this PR description. i used LLM to brainstorm how to solve the problem specifically after you suggested me to create new data type instead of mixing the alias Default variant to the real tool chain variants and also i used github copilot to check if i applied  ToolchainOverride<T> consistently across the relevant functions and return types and i tested everything by myself, but i will add the tools i used for help  when i update the PR description. thanks for checking.

14elias: i also used github copilot to help me with tests i mean the role is not like writing the tests it is just like discussing the edge cases and checking what i write and telling me to update the parts i have to modify and what i have to cover. that is how i use AI

src/toolchain/names.rs
14elias: i kept validate_name() in ToolchainOverride::from_str() because it allow input +default/  as +stable/ . but if aliases should only accept exact names then checking if the value is alias is enough.

src/toolchain/names.rs
14elias: the existing code uses this pattern that's why i used that but i can import std::fmt and change in all the places if you want.

src/toolchain/names.rs
rami3l: Actually you need a space between `///` and the first word.

src/toolchain/names.rs
rami3l: Yes, please do that. You can import `std::fmt` in a separate commit before this one, and then stack your current changes on top of that one.

src/cli/rustup_mode.rs
```diff
@@ -1794,10 +1796,13 @@ fn pin_active_toolchain(qualified: bool, cfg: &Cfg<'_>) -> anyhow::Result<ExitCo
...
     cfg: &Cfg<'_>,
-    toolchain: ResolvableToolchainName,
+    toolchain: Override<ResolvableToolchainName>,
     path: Option<&Path>,
 ) -> anyhow::Result<ExitCode> {
-    let toolchain_name = toolchain.clone().resolve(&cfg.default_host_tuple()?)?;
+    let resolved_toolchain = toolchain.clone().resolve(cfg)?;
+    let toolchain_name = resolved_toolchain
+        .clone()
+        .resolve(&cfg.default_host_tuple()?)?;
```
rami3l: Could you please inline the temporary variable `resolved_toolchain`, or is there anything that prevents you from doing so? Also the `resolved_toolchain.clone()` below may not be necessary.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/7464de1b6121e5ee0fceee3bdcd695ded379cc12..802fa8ef1cf7c6ef35878b171322537e44892b09)*

djc:
Why would people want/need this? What is the use case?

And why add all this complexity with more type system layering, what problem does it solve?

src/config.rs
```diff
@@ -25,12 +25,30 @@ use crate::{
...
+                let default = cfg
+                    .get_default_resolvable()?
+                    .ok_or_else(|| no_toolchain_error(cfg.process))?;
+                Ok(T::from(default))
+            }
+            Self::Explicit(value) => Ok(value),
+        }
+    }
+}
+
```
rami3l: Nit: We'd generally want the `impl` block to live right below the type declaration, so this should be put into `src/toolchain/names.rs` below the declaration of `Override`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/7464de1b6121e5ee0fceee3bdcd695ded379cc12..802fa8ef1cf7c6ef35878b171322537e44892b09)*

src/config.rs
```diff
@@ -668,7 +688,6 @@ impl<'a> Cfg<'a> {
             else {
                 None
             };
-
```
rami3l, line 671: This whitespace change looks redundant.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/7464de1b6121e5ee0fceee3bdcd695ded379cc12..802fa8ef1cf7c6ef35878b171322537e44892b09)*

rami3l:
> Why would people want/need this? What is the use case?
> 
> And why add all this complexity with more type system layering, what problem does it solve?

@djc This PR implements the `default` layer of https://github.com/rust-lang/rustup/issues/5025. My current plan would be to add only `default` and `file=xxx` layers first, starting with the easiest one (this one, and @14elias happens to have offered to do it), and leave the rest for discussion.

@14elias In the PR description you should have made that clear, otherwise the context may be lost quite easily.

djc:
Still think it's good to explain in this PR when/why people would want this, specifically.

rami3l (approved):
LGTM, thanks a lot :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5071/7464de1b6121e5ee0fceee3bdcd695ded379cc12..90a931710ec868b5439ca9da5861a4ea227bb175)*

djc:
> And why add all this complexity with more type system layering, what problem does it solve?

I'm surprised this was merged without more of a response to these questions.

rami3l:
@djc The author did update the PR description to add more details though. What do you think we should do right now?

14elias:
> > And why add all this complexity with more type system layering, what problem does it solve?
> 
> I'm surprised this was merged without more of a response to these questions.

i have updated the PR description

djc:
My comment contained two questions: one suggesting a clarification of the context of the PR, and one asking about the complexity of the code in the PR. The first one was adequately handled, but it doesn't feel to me like the second one was considered much.

14elias:
> My comment contained two questions: one suggesting a clarification of the context of the PR, and one asking about the complexity of the code in the PR. The first one was adequately handled, but it doesn't feel to me like the second one was considered much.

it can be solved simply by checking if the input is default and dispatching it to the configured toolchain but there will be repeating the same codes multiple times specifically the dispatching logic . but this changes follows DRY policy like i write the dispatching in the resolve() method once and i will call it in all the places i want  and also this change allows another aliases to be added easily in the future.  i hope i answered you r question.

djc:
> it can be solved simply by checking if the input is default and dispatching it to the configured toolchain but there will be repeating the same codes multiple times specifically the dispatching logic . but this changes follows DRY policy like i write the dispatching in the resolve() method once and i will call it in all the places i want and also this change allows another aliases to be added easily in the future. i hope i answered you r question.

To be clear, my feedback isn't for you so much as it is for @rami3l.

rami3l:
@djc My apologies for moving too fast here.

That said, I think the wrapper type is useful because override strings are expected to show up in two main cases, `ResolvableToolchainName` and `ResolvableLocalToolchainName`, since different override layers would accept different value types (the difference between the two is because, notably, `RUSTUP_TOOLCHAIN` would accept path-based toolchains).

That said, I dislike the fact that the `none` reserved name is handled in `MaybeOfficialToolchainName` and I'd love to be able to handle all the logic around reserved names in a single place, but I currently don't have a concrete idea for improvement regarding that part. I'd love to work on cleanup PRs if I happen to have found something.

djc:
It feels to me like we have a lot of different "transition" and a whole bunch of supporting code, but we should use a small number of toolchain name types throughout the codebase and proactively resolve to these early/upfront.

It would be good to summarize the design of these and see if some of it can be flattened.

Issue #4777, opened 2026-03-26
unable to run updater on `rustup self update`


Thomasdezeeuw:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

I'm trying to update my rustc versions and the following error occurred:

```
$ rustup update

info: syncing channel updates for 'stable-aarch64-apple-darwin'
info: latest update on 2026-03-26, rust version 1.94.1 (e408947bf 2026-03-25)
info: downloading component 'rustfmt'
info: downloading component 'cargo'
info: downloading component 'rust-std'
info: downloading component 'rustc'
 61.3 MiB /  61.3 MiB (100 %)  40.8 MiB/s in  1s
info: removing previous version of component 'rustfmt'
info: removing previous version of component 'cargo'
info: removing previous version of component 'rust-std'
info: removing previous version of component 'rustc'
info: installing component 'rustfmt'
info: installing component 'cargo'
info: installing component 'rust-std'
info: installing component 'rustc'
 61.3 MiB /  61.3 MiB (100 %)  29.1 MiB/s in  2s
info: syncing channel updates for 'nightly-aarch64-apple-darwin'
info: latest update on 2026-03-26, rust version 1.96.0-nightly (80d0e4be6 2026-03-25)
info: downloading component 'rustc-dev'
 94.6 MiB /  94.6 MiB (100 %)   4.3 MiB/s in 21s
info: downloading component 'rust-std' for 'x86_64-unknown-freebsd'
 27.6 MiB /  27.6 MiB (100 %)   4.2 MiB/s in  6s
info: downloading component 'rustfmt'
info: downloading component 'miri'
info: downloading component 'rust-src'
info: downloading component 'rust-docs'
info: downloading component 'clippy'
info: downloading component 'cargo'
info: downloading component 'rust-std'
info: downloading component 'rustc'
 46.8 MiB /  46.8 MiB (100 %)  39.9 MiB/s in  1s
info: removing previous version of component 'rustc-dev'
info: removing previous version of component 'rust-std' for 'x86_64-unknown-freebsd'
info: removing previous version of component 'rustfmt'
info: removing previous version of component 'miri'
info: removing previous version of component 'rust-src'
info: removing previous version of component 'rust-docs'
info: removing previous version of component 'clippy'
[...]

Issue #4648, opened 2025-12-15
failed to set permissions on `rustup self update`


kornelski:
### Verification

- [x] I searched for recent similar issues at https://github.com/rust-lang/rustup/issues?q=is%3Aissue and found no duplicates.
- [x] I am on the latest version of Rustup according to https://github.com/rust-lang/rustup/tags and am still able to reproduce my issue.

### Problem

I wanted to try out the new dev version, but the older version failed self-update:

```text
$ RUSTUP_UPDATE_ROOT=https://dev-static.rust-lang.org/rustup rustup update
info: syncing channel updates for 'stable-aarch64-apple-darwin'
info: syncing channel updates for 'nightly-aarch64-apple-darwin'
info: checking for self-update
info: downloading self-update
error: failed to set permissions for '/Users/me/.cargo/bin/rustup-init'

Caused by:
    No such file or directory (os error 2)

$ RUSTUP_UPDATE_ROOT=https://dev-static.rust-lang.org/rustup rustup --version
rustup 1.28.2 (77863418e 2025-04-13)
```

I have `~/.cargo/bin/rustup`, but not `rustup-init`.

### Steps

`RUSTUP_UPDATE_ROOT=https://dev-static.rust-lang.org/rustup rustup update`

### Rustup version

```console
rustup 1.28.2 (77863418e 2025-04-13)
```

### Installed toolchains

```console
Default host: aarch64-apple-darwin
rustup home:  /Users/me/.rustup

installed toolchains
--------------------
stable-aarch64-apple-darwin
nightly-aarch64-apple-darwin (active, default)
nightly-2023-06-14-aarch64-apple-darwin
nightly-2024-06-24-aarch64-apple-darwin
nightly-2025-02-06-aarch64-apple-darwin
1.68.0-aarch64-apple-darwin
1.70.0-aarch64-apple-darwin
1.71.0-aarch64-apple-darwin
1.73.0-aarch64-apple-darwin
1.75.0-aarch64-apple-darwin
1.76.0-aarch64-apple-darwin
1.77.0-aarch64-apple-darwin
1.83.0-aarch64-apple-darwin
stage1
stage2
stage3

active toolchain
----------------
name: nightly-aarch64-apple-darwin
active because: it's the default toolchain
installed targets:
  aarch64-apple-darwin
  aarch64-apple-ios
  thumbv7em-none-eabihf
  wasm32-unknown-unknown
  x86_64-apple-darwin
  x86_64-unknown-linux-musl
[...]

PR #5077, opened 2026-09-13
fix(self-update): make replacement atomic and race-safe

### LINKED ISSUE #1864: Self update could interfere with itself
We don't seem to download the self-update to a unique pathname.  As such two self-updates could interfere with one another.

It probably makes sense to alter the self-update pathways to use unique names and potentially to use a lockfile during the actual update process after download.

### LINKED ISSUE #5076: Concurrent proxy startup can delete the active self-updater and remove rustup
### Problem

Rustup stages every self-update at the shared path `$CARGO_HOME/bin/rustup-init`.

Normal `rustup` and proxy invocations such as `rustc` run updater cleanup on startup. If a proxy starts while self-update is replacing rustup, it can delete the running updater's path.

On Unix, replacement then:

1. removes the installed `rustup`;
2. tries to copy the updater from its now-deleted path;
3. fails with `No such file or directory`;
4. leaves `rustup` missing.

This seems to be related to #1864, #4648, and #4777.

### Reproduction

The reproduction uses isolated test directories and does not affect the user's real rustup installation. I have included a branch with the accompanying changes to demonstrate the bug.

```console
git clone https://github.com/cachebag/rustup.git
cd rustup
git switch demo/self-update-race-before

cargo test \
  --test test_bonanza \
  --no-default-features \
  --features reqwest-rustls-tls,test \
  suite::cli_self_upd::self_update_replacement_survives_proxy_cleanup \
  -- --exact --nocapture
```

The test deterministically pauses self-update immediately before replacement, starts a real `rustc` proxy invocation, then resumes replacement.

The vulnerable branch fails with:

```text
error: could not read metadata for file: '.../bin/rustup-init':
No such file or directory
concurrent proxy removed the installed rustup during self-update
```

For comparison, the same test passes on my branch (which is the PR linked to this issue):

```console
git switch demo/self-update-race-after
cargo test \
  --test test_bonanza \
  --no-default-features \
  --features reqwest-rustls-tls,test \
  suite::cli_self_upd::self_update_replacement_survives_proxy_cleanup \
  -- --exact --nocapture
```

### Expected behavior

Concurrent rustup or proxy invocations must not delete another process's self-update artifact. A failed replacement must leave the previously installed `rustup` runnable.

cachebag:
Self-update currently stages every updater at the shared `$CARGO_HOME/bin/rustup-init` path. Any rustup proxy started while replacement is in progress can unlink that path during startup cleanup. On Unix, the replacement process then removes the installed `rustup`, fails to reopen its own deleted path, and leaves rustup missing. This appears to explain the failures reported in #1864, #4648, and #4777.

The first commit adds a deterministic test for that race. It fails on [`demo/self-update-race-before`](https://github.com/cachebag/rustup/tree/demo/self-update-race-before) and the identical test passes with this branch.

This PR uses a single managed updater under `$RUSTUP_HOME`, guarded by an OS-released global self-update lock. It fully prepares and syncs a sibling binary before publishing it with rename on Unix or ReplaceFileW on Windows. Completed, failed, and abandoned artifacts are cleaned without racing an active update.

The core result of this fix is that now, a failure before publication will leave the existing rustup untouched.

Fixes #5076
Fixes #1864


tests/suite/cli_self_upd.rs
```diff
@@ -23,6 +23,13 @@ use rustup::{
 #[cfg(windows)]
 use windows_registry::{CURRENT_USER, Value};
 
+#[cfg(windows)]
+const USER_RUSTUP_VERSION: RegistryValueId = RegistryValueId {
+    sub_key: r"Software\Microsoft\Windows\CurrentVersion\Uninstall\Rustup",
+    value_name: "DisplayVersion",
+};
+const TEST_VERSION: &str = "1.1.1";
```
rami3l: Nit: The current style guide suggests placing the constants at the bottom. Also, moving them there could be in a separate commit before this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/17f3d120587c6f365d4040b69ba029481ced5f97..a60db3a93bdc613e943a07f63009ec29b21a032c)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,396 @@
...
+            .context("failed to open self-update replacement lock")?;
+        file.lock()
+            .context("failed to lock self-update replacement")?;
+
+        Ok(Self { _file: file })
+    }
+}
+
+pub(super) fn configure_replacer(command: &mut Command, updater_path: &Path) {
+    if let Some(stage) = updater_path.parent() {
```
rami3l: Maybe this function should be returning a configured command based on `updater_path()`? What do you think?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/17f3d120587c6f365d4040b69ba029481ced5f97..a60db3a93bdc613e943a07f63009ec29b21a032c)*

rami3l: In the case where `.parent()` returns `None`, is this supposed to be an internal error?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/17f3d120587c6f365d4040b69ba029481ced5f97..a60db3a93bdc613e943a07f63009ec29b21a032c)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,396 @@
...
+        return Ok(());
+    }
+
+    utils::write_file("self-update status marker", &stage.join(marker), "")
+}
+
+fn cleanup_at(process: &Process, now: SystemTime) -> anyhow::Result<()> {
+    cleanup_stages(process, now)?;
+    cleanup_pending_binaries(process, now)?;
+    cleanup_legacy_updater(process, now)
```
rami3l: Nit: Suggest inlining those single-use functions. If function naming is used for segmenting the procedure, you can use comment lines to achieve the same goal.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/17f3d120587c6f365d4040b69ba029481ced5f97..a60db3a93bdc613e943a07f63009ec29b21a032c)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,396 @@
...
+            directory,
+            updater_path,
+        })
+    }
+
+    pub(super) fn updater_path(&self) -> &Path {
+        &self.updater_path
+    }
+
+    pub(super) fn persist(self) -> PathBuf {
```
rami3l: Question: What problem is randomizing the name of the updater trying to solve?

Would it be simpler to just not tolerate multiple updaters at the same time and thus a global self-update lock may be enough (if to concurrent instances of `run_update()` clash with each other then so be it)?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/17f3d120587c6f365d4040b69ba029481ced5f97..a60db3a93bdc613e943a07f63009ec29b21a032c)*

src/cli/self_update.rs
```diff
@@ -773,21 +774,51 @@ fn warn_if_default_linker_missing(process: &Process) {
 }
 
 fn install_bins(process: &Process) -> anyhow::Result<()> {
-    let bin_path = process.cargo_home()?.join("bin");
+    let _replacement_lock = stage::ReplacementLock::acquire(process)?;
+    install_bins_unlocked(process)
+}
+
+pub(super) fn install_bins_unlocked(process: &Process) -> anyhow::Result<()> {
```
rami3l: Would it be easier if we just unify those two APIs, changing the signature to `fn install_bins(&ReplacementLock, &Process) -> anyhow::Result<()>` so that the lock acts as a witness object? Or, better yet, should it be a method of `ReplacementLock` to make this even clearer?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/17f3d120587c6f365d4040b69ba029481ced5f97..a60db3a93bdc613e943a07f63009ec29b21a032c)*

src/cli/self_update/stage.rs
cachebag: Yeah it was essentially to isolate overlapping downloads..but I guess parallel self-updates aren’t behavior we need to preserve. A single managed updater guarded by a global self-update lock makes sense to me. Thank you!

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,421 @@
...
+                    .file_name()
+                    .to_string_lossy()
+                    .starts_with(PENDING_BINARY_PREFIX)
+                    && is_stale(&path, now)
+                {
+                    remove_file_best_effort("pending rustup binary", &path);
+                }
+            }
+        }
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
```
rami3l: Maybe we should just `use std::io:` here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/7464de1b6121e5ee0fceee3bdcd695ded379cc12..3b4cc2bd76f391eacde429bece84dce1fa566c83)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,421 @@
...
+
+use crate::{process::Process, utils};
+
+pub(super) struct PreparedUpdate {
+    updater_path: PathBuf,
+    _lock: SelfUpdateLock,
+}
+
+impl PreparedUpdate {
+    pub(super) fn updater_path(&self) -> &Path {
```
djc: Nit: move the trivial getter below `replacer_command()`?

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,421 @@
...
+        .is_some_and(|age| age >= ABANDONED_UPDATE_AGE)
+}
+
+fn stage_root(process: &Process) -> anyhow::Result<PathBuf> {
+    Ok(process.rustup_home()?.join(SELF_UPDATE_DIRECTORY))
+}
+
+const SELF_UPDATE_DIRECTORY: &str = "self-update";
+const COMPLETE_MARKER: &str = "complete";
+const FAILED_MARKER: &str = "failed";
```
djc: Suggest making these markers into an `enum` with an `as_str() -> &'static str` method.

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,421 @@
...
+    process::Command,
+    time::{Duration, SystemTime},
+};
+
+use anyhow::Context;
+use tracing::{debug, warn};
+
+use crate::{process::Process, utils};
+
+pub(super) struct PreparedUpdate {
```
djc: Reorder items (functions, types) in top-down order, so caller before callee.

src/cli/self_update/unix.rs
```diff
@@ -140,13 +145,32 @@ pub(crate) fn run_update(setup_path: &Path, _process: &Process) -> anyhow::Resul
...
-    install_bins(process)?;
+    let result = self_update_lock.install_bins(process);
+    stage::mark_result(process, result.is_ok());
+    result?;
 
     Ok(utils::ExitCode(0))
 }
 
+pub(super) fn replace_rustup_binary(replacement: &Path, rustup: &Path) -> anyhow::Result<()> {
+    std::fs::rename(replacement, rustup).with_context(|| {
```
djc: Suggest importing `fs` and `File` directly.

src/cli/self_update/windows.rs
```diff
@@ -658,28 +657,71 @@ pub(crate) fn remove_uninstall_registry_entry(process: &Process) -> anyhow::Resu
...
+                )
+            }),
+        Err(error) => Err(error).context("failed to inspect the installed rustup binary"),
+    }
+}
+
+fn replace_existing_rustup(replacement: &Path, rustup: &Path) -> anyhow::Result<()> {
+    use std::ptr::null;
+
+    use windows_sys::Win32::Storage::FileSystem::ReplaceFileW;
```
djc: Move imports to the top of the module.

src/cli/self_update/windows.rs
```diff
@@ -658,28 +657,71 @@ pub(crate) fn remove_uninstall_registry_entry(process: &Process) -> anyhow::Resu
...
+
+    let rustup = utils::raw::windows::to_u16s(rustup)?;
+    let replacement = utils::raw::windows::to_u16s(replacement)?;
```
djc: Don't use fully-qualified paths here.

src/cli/self_update.rs
```diff
@@ -773,19 +772,49 @@ fn warn_if_default_linker_missing(process: &Process) {
 }
 
 fn install_bins(process: &Process) -> anyhow::Result<()> {
-    let bin_path = process.cargo_home()?.join("bin");
-    let this_exe_path = utils::current_exe()?;
-    let rustup_path = bin_path.join(format!("rustup{EXE_SUFFIX}"));
+    stage::SelfUpdateLock::acquire(process)?.install_bins(process)
```
djc: Nit: import `SelfUpdateLock` directly.

src/cli/self_update.rs
```diff
@@ -773,19 +772,49 @@ fn warn_if_default_linker_missing(process: &Process) {
...
+}
+
+impl stage::SelfUpdateLock {
+    pub(super) fn install_bins(&self, process: &Process) -> anyhow::Result<()> {
```
djc: I'm not a fan of splitting inherent impls over multiple files, just keep it with the type definition.

src/cli/self_update.rs
```diff
@@ -1214,18 +1233,14 @@ fn parse_new_rustup_version(version: String) -> String {
     String::from(matched_version)
 }
 
-pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> anyhow::Result<Option<PathBuf>> {
+async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> anyhow::Result<Option<stage::PreparedUpdate>> {
```
djc: Nit: import `PreparedUpdate` directly.

djc (commented):
Here's an initial round of feedback.

IMO the last commit is still pretty messy, and looks like there could be more intermediate commits to clean things up and provide more structure.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/7464de1b6121e5ee0fceee3bdcd695ded379cc12..3b4cc2bd76f391eacde429bece84dce1fa566c83)*

src/cli/self_update.rs
cachebag: This originally came from @rami3l's suggestion to make `install_bins` a method on the lock (which I've kept), but I agree that where I put the impl is not ideal.

src/cli/self_update/windows.rs
cachebag, line 1: cc @ChrisDenton 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/23ebafd2c623968c1e0122fe84747284e7b90870..af1b0393e072eeeacc13e24a4f51da72d4bfbb22)*

ChrisDenton: Sorry, catching up. Why `ReplaceFileW` instead of `std::fs::rename`?

cachebag: I want the replaced file's ACL and attributes to be preserved. I also figured it was MS' blessed "replace a file" API since it can cope with an in-use target

ChrisDenton: Could you add a comment to that effect?

cachebag: In retrospect, I noticed I pass a null backup name, so `ReplaceFileW` still tries to delete the old file and fails on a running exe. And it's not a single rename, so a failure can leave `rustup.exe` renamed away, which is what this PR is trying to prevent. 

I'll switch to `std::fs::rename`.

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,543 @@
...
+    // Legacy updaters have no result marker, and an older rustup process may
+    // still own the shared path.
+    if is_stale(&updater, now) {
+        remove_file_best_effort("legacy self-updater", &updater);
+    }
+
+    Ok(())
+}
+
+fn remove_file_best_effort(name: &str, path: &Path) -> bool {
```
ChrisDenton: I need to think more about this but I feel like we have a lot of utilities for removing files in different ways or with different strategies. In any case, it might be useful to keep this with the others in utils in case something else needs it. At least then more of the `remove_file` wrappers will be in the same place.

ChrisDenton (commented):
I'll want to look at this more tomorrow when I'm more rested but looks good to me so far.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..c1d76d84d69bd45480c5528d7cf49ae833c1405a)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,412 @@
...
+        for marker in [Marker::Complete, Marker::Failed] {
+            utils::ensure_file_removed(
+                "self-update status marker",
+                &self.directory.join(marker.as_str()),
+            )?;
+        }
+        Ok(PreparedUpdate {
+            updater_path,
+            _lock: self,
+        })
```
rami3l: Nit: `.prepare_updater()` returning `PreparedUpdate` is slightly surprising to me here...

How about renaming this to `PreparedUpdater { path, _lock }`?
Or, if you prefer, probably rename this method to `.prepare_update()` instead?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..c1d76d84d69bd45480c5528d7cf49ae833c1405a)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    fn open(process: &Process) -> anyhow::Result<Self> {
+        let directory = stage_root(process)?;
+        utils::ensure_dir_exists("self-update", &directory)?;
+        let file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .create(true)
+            .truncate(false)
```
rami3l, line 46: A quick question: is the truncating flag important here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..c1d76d84d69bd45480c5528d7cf49ae833c1405a)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,543 @@
...
+        );
+        return Ok(());
+    }
+
+    utils::write_file(
+        "self-update status marker",
+        &stage.join(marker.as_str()),
+        "",
+    )
+}
```
rami3l: Nit: It looks like this could be a method of `Marker`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..c1d76d84d69bd45480c5528d7cf49ae833c1405a)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,543 @@
...
+fn replace_rustup_binary(replacement: &Path, rustup: &Path) -> anyhow::Result<()> {
+    // `rename` replaces an existing destination in one step on every platform,
+    // so a failure here leaves the installed rustup untouched.
+    fs::rename(replacement, rustup).with_context(|| {
+        format!(
+            "failed to replace rustup binary '{}' with '{}'",
+            rustup.display(),
+            replacement.display()
+        )
+    })?;
```
rami3l: Nit: I think we have `utils::rename()` for that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..c1d76d84d69bd45480c5528d7cf49ae833c1405a)*

src/cli/self_update/unix.rs
```diff
@@ -121,13 +118,21 @@ pub(crate) fn do_write_env_files(process: &Process) -> anyhow::Result<()> {
...
+    _process: &Process,
+) -> anyhow::Result<utils::ExitCode> {
+    let setup_path = prepared_update.updater_path().to_owned();
+    let mut updater = prepared_update
+        .replacer_command()?
         .arg("--self-replace")
-        .status()
+        .spawn()
         .context(format!("unable to run updater ({})", setup_path.display()))?;
+    drop(prepared_update);
```
rami3l: Question: Judging from the usage, should `.replacer_command()` take the ownership of `self` directly? It looks like you are always dropping the instance immediately after 👀 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..c1d76d84d69bd45480c5528d7cf49ae833c1405a)*

src/cli/self_update/windows.rs
```diff
@@ -658,24 +656,28 @@ pub(crate) fn remove_uninstall_registry_entry(process: &Process) -> anyhow::Resu
...
 
     Ok(utils::ExitCode(0))
 }
 
 pub(crate) fn self_replace(process: &Process) -> anyhow::Result<utils::ExitCode> {
     wait_for_parent()?;
-    install_bins(process)?;
+    let self_update_lock = SelfUpdateLock::acquire(process)?;
+    let result = self_update_lock.install_bins(process).and_then(|()| {
+        update_uninstall_registry_display_version(env!("CARGO_PKG_VERSION"), process)
```
rami3l, line 673: I thought we are supposed to get the version string from the cmdline, not from `CARGO_PKG_VERSION` (which is from `Cargo.toml` IIRC and may not be 100% accurate)?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..c1d76d84d69bd45480c5528d7cf49ae833c1405a)*

src/cli/self_update/unix.rs
cachebag: The lock has to stay held until `spawn()` returns. Otherwise a concurrent `rustup self update` can acquire the lock, run `prepare_updater`, delete and re-download the updater, and the first process execs a half-written file.

What if we replace `replacer_command()` with `spawn_replacer(self) -> anyhow::Result<Child>` that adds `--self-replace`, spawns, then drops the lock. Unix then can just wait on the child?

src/cli/self_update/stage.rs
cachebag: clippy's `suspicious_open_options` lint requires an explicit truncate choice. the file is only opened to take the lock. we do something similar here already

https://github.com/rust-lang/rustup/blob/638f330db9a2400af12389949f0cc9b372d94454/src/utils/raw.rs#L104-L112

src/cli/self_update/windows.rs
cachebag: the only reason we shelled out to `rustup-init --version` was because the parent is the old binary and can't know the new version. the replacer is the new binary, so its own `CARGO_PKG_VERSION` is the authoritative value. 

also, `--version` renders from `git-testament`, which itself embeds `CARGO_PKG_VERSION`, and the parent then regex-parsed that string back out.

src/cli/self_update/stage.rs
```diff
@@ -130,6 +130,48 @@ pub(super) fn cleanup(process: &Process) -> anyhow::Result<()> {
...
+    if !fs::symlink_metadata(&pending)?.file_type().is_symlink() {
+        OpenOptions::new()
+            .write(true)
+            .open(&pending)
+            .and_then(|file| file.sync_all())
+            .context("failed to sync the pending rustup binary")?;
+    }
+
+    Ok(pending)
+}
```
rami3l: Hmmm I've repeated this for quite a few times, but if this is a single-use function, just inline it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/638f330db9a2400af12389949f0cc9b372d94454..1dcce2901e70cadf413d1e1ae26d6f821f0df5df)*

src/cli/self_update/windows.rs
rami3l: @cachebag Thanks! I just double checked. Indeed the parent fetches the full string like `1.30.0 :: 1.29.0+361 (02a69d73f 2026-09-17)` but only the first section is used, so I guess we are good here.

rami3l (changes requested):
In terms of the end result this looks good. However as @djc has mentioned in https://github.com/rust-lang/rustup/pull/5077#pullrequestreview-5207256902, I have the feeling that the two main commits here have been way too juicy to review properly.

Also, this is introducing an API surface that is way too wide without a clear reason. For example, helpers like `is_finished()` have introduced no clear benefits and has slowed down the review process since the reviewer has to find the function and do the jump.

My perspective as the reviewer regarding atomic commits is that every should be as dumb as possible: the less juicy the meat of the feature actually is (and the fewer jumps I have to make when reading the diff file), the better.

Most commits in your PR should contain mostly tasteless changes with a clear single focus: renaming, extraction, inlining actions etc., that sandwich the actual feature changes, and when a PR contains too many such changes upfront, you should consider splitting some of them into one preparation PR or more.

A more concrete example here: WRT the staged updater, the cleanup of the stale file can be completely separated from the main feature, so probably you can split the change into two.

I totally understand that you might feel uncomfortable "inventing" intermediate states of your history, but as with every paper or code publication there is, some polishing to the "how we got there" is required for easier digestion.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/638f330db9a2400af12389949f0cc9b372d94454..1dcce2901e70cadf413d1e1ae26d6f821f0df5df)*

src/cli/self_update/unix.rs
rami3l: @cachebag That looks good to me :)

src/cli/self_update/stage.rs
cachebag: Sorry about that- this has now been inlined along with `is_finished()` as it was of similar shape.

cachebag:
> I totally understand that you might feel uncomfortable "inventing" intermediate states of your history, but as with every paper or code publication there is, some polishing to the "how we got there" is required for easier digestion.

@rami3l @djc 
This is reasonable. And I appreciate the feedback.

What do you think of the new structure? Let me know if there are any changes you'd want me to make.

src/cli/self_update/unix.rs
```diff
@@ -121,13 +118,15 @@ pub(crate) fn do_write_env_files(process: &Process) -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Tell the upgrader to replace the rustup bins, then delete
-/// itself.
```
rami3l, line 125: Nit: I think this docstring still applies?

src/cli/self_update/unix.rs
```diff
@@ -121,13 +118,15 @@ pub(crate) fn do_write_env_files(process: &Process) -> anyhow::Result<()> {
...
-        .context(format!("unable to run updater ({})", setup_path.display()))?;
+pub(super) fn run_update(
+    prepared_updater: PreparedUpdater,
+    _process: &Process,
+) -> anyhow::Result<utils::ExitCode> {
+    let setup_path = prepared_updater.path().to_owned();
+    let status = prepared_updater.spawn_replacer()?.wait().context(format!(
+        "unable to wait for updater ({})",
+        setup_path.display()
+    ))?;
```
rami3l: Nit: Pre-existing, but possibly we can use `.with_context()` instead of `.context()` to make the evaluation a bit lazier?

src/cli/self_update.rs
```diff
@@ -1266,18 +1253,19 @@ pub(crate) async fn prepare_update(dl_cfg: &DownloadCfg<'_>) -> anyhow::Result<O
 
     // Get download path
     let download_url = utils::parse_url(&url)?;
+    let prepared_updater = self_update_lock.prepare_updater(dl_cfg.process)?;
```
rami3l, line 1246: Nit: Adding `let setup_path = prepared_updater.path();` here could save some space in the diff.

src/cli/self_update/stage.rs
```diff
@@ -87,23 +111,118 @@ impl PreparedUpdater {
     }
 }
 
+pub(super) fn mark_result(process: &Process, succeeded: bool) {
```
rami3l: Nit: According to our style guide, you should place the most interesting param upfront, so `succeeded` in `mark_result()` first, and then `process`.

src/cli/self_update/stage.rs
```diff
@@ -87,23 +111,118 @@ impl PreparedUpdater {
...
+    };
+    if let Err(error) = marker.record(process, &stage) {
+        warn!("could not record self-update result: {error}");
+    }
+}
+
+pub(super) fn cleanup(process: &Process) -> anyhow::Result<()> {
+    if let Some(lock) = SelfUpdateLock::try_acquire(process)? {
+        let updater = lock.updater_path();
+        // A finished update has recorded its outcome; an abandoned one never will.
```
rami3l: Is this claim a bit inaccurate?

If I have understood it correctly, it looks like "having markers" implies that an update has finished, but not having markers may or may not indicate a completed update, because the updater could have crashed exactly when trying to write the markers?

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,137 @@
...
+use super::install_proxies;
+use crate::{process::Process, utils};
+
+/// Exclusive right to download the updater or replace the installed rustup.
+pub(super) struct SelfUpdateLock {
+    _file: File,
+}
+
+impl SelfUpdateLock {
+    pub(super) fn acquire(process: &Process) -> anyhow::Result<Self> {
```
rami3l, line 23: Nit: Following the naming convention, this series of methods are better named `.lock()` and `.try_lock()`.

rami3l (commented):
@cachebag Thanks for the update! I think it is indeed looking a bit better now.

I really want to get this right before merging, so I'd like to invite the rest of @rust-lang/rustup to review as well just in case I've missed anything and/or they have some opinions regarding this patch.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/02a69d73f69dbc42dfa26667ab836d88e7d80140..38048e7846956d7a64a69e6d19b2f41ba60e437c)*

src/cli/self_update/stage.rs
```diff
@@ -87,23 +111,118 @@ impl PreparedUpdater {
...
+            .iter()
+            .any(|marker| marker.path(&lock.directory).is_file());
+        if finished && utils::remove_file_best_effort("self-updater", &updater) {
+            for marker in markers {
+                utils::remove_file_best_effort(
+                    "self-update status marker",
+                    &marker.path(&lock.directory),
+                );
+            }
+        }
```
rami3l, line 190: Question: Would you mind elaborating why the marker is required in addition to the lock and, if it is indeed needed, why do we have two of them?

The rationale here is not very clear to me at least through your docstrings and/or commit messages.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/02a69d73f69dbc42dfa26667ab836d88e7d80140..38048e7846956d7a64a69e6d19b2f41ba60e437c)*

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl SelfUpdateLock {
+    pub(super) fn acquire(process: &Process) -> anyhow::Result<Self> {
+        let lock = Self::open(process)?;
+        lock._file.lock().context("failed to lock self-update")?;
+        Ok(lock)
+    }
+
+    fn open(process: &Process) -> anyhow::Result<Self> {
```
djc, line 38: This seems to be single-caller, so could we just inline it?

src/cli/self_update/stage.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /// Removes any leftover updater and reserves its path for the download.
+    pub(super) fn prepare_updater(self, process: &Process) -> anyhow::Result<PreparedUpdater> {
+        let path = process
+            .cargo_home()?
+            .join(format!("bin/rustup-init{EXE_SUFFIX}"));
+        utils::ensure_file_removed("self-updater", &path)?;
+        Ok(PreparedUpdater { path, _lock: self })
+    }
```
djc: So I think this is effectively a type state thing where we're taking the `SelfUpdateLock` and wrapping the contained lock into a `PreparedUpdater`, right? I think it might be simpler to redefine `SelfUpdateLock` as a type with a generic type parameter somewhat like `Mutex`, with a `Deref` impl to replace `PreparedUpdater::path()`.

src/cli/self_update/stage.rs
```diff
@@ -87,23 +111,118 @@ impl PreparedUpdater {
...
+#[derive(Clone, Copy)]
+enum Marker {
+    Complete,
+    Failed,
+}
+
+impl Marker {
+    /// Records this outcome in `stage`, ignoring stages outside the managed directory.
+    fn record(self, process: &Process, stage: &Path) -> anyhow::Result<()> {
+        if stage != stage_root(process)? {
```
djc, line 245: Nit: since this uses `stage_root()`, it should be defined before it.

src/utils/mod.rs
```diff
@@ -423,6 +423,34 @@ pub(crate) fn format_path_for_display(path: &str) -> String {
     }
 }
 
+/// Removes `path` if possible, without failing.
+///
+/// Unlike [`remove_file`], a busy file is left alone instead of retried, since
+/// callers use this for cleanup that another process may legitimately still be
+/// using. Returns whether `path` is gone afterwards.
+pub(crate) fn remove_file_best_effort(name: &str, path: &Path) -> bool {
```
djc, line 431: Unless/until this is used for other callers, suggest keeping it where it is used.

djc (commented):
I think the first few commits are nice and clean, and we could probably merge them proactively via a separate commit?

Later commits seem to not be logically independent, so I'm not sure keeping them all separate really makes sense.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/02a69d73f69dbc42dfa26667ab836d88e7d80140..38048e7846956d7a64a69e6d19b2f41ba60e437c)*

src/cli/self_update/stage.rs
cachebag: The lock only says whether an update is running right now. After the parent exits and before the replacer takes the lock, cleanup sees a free lock and an updater on disk and can’t tell the difference between being finished or queued to run. Deleting it in the second case is the bug. 

The marker is the “done” signal. Without it the only safe option is age, and it doesn't seem conducive to then let every updated sit around for a day.

> why do we have two of them?

I guess cleanup treats them the same. `failed` only helps the test fail fast and leaves a trace on Windows.

src/cli/self_update/stage.rs
cachebag: It’s single-caller only in this commit. `try_lock` in the next commit is the second caller. Inlining it here would just mean re-extracting it one commit later.

src/cli/self_update/stage.rs
cachebag: My hesitation with this is the payload is only ever a `PathBuf`, so it’d be one type parameter for one call site, and `Deref` to a path hides what’s being handed around.

src/utils/mod.rs
cachebag: I agreed with what @ChrisDenton said here https://github.com/rust-lang/rustup/pull/5077#discussion_r4021568703.

wdyt?

cachebag:
> Later commits seem to not be logically independent, so I'm not sure keeping them all separate really makes sense.

@djc 
I split these at @rami3l's request in https://github.com/rust-lang/rustup/pull/5077#pullrequestreview-5233569083. I do not mind either way; could either of you propose the preferred structure?

Re: a separate PR; I suppose that is fine...ab0d89ad and 1ca3033d can be by themselves, though what should I do with fc00f5ff? The whole point of that test is that it fails without the fix.

src/cli/self_update/stage.rs
cachebag: I will withhold it until https://github.com/rust-lang/rustup/pull/5077#discussion_r4045275736 is settled, since that may reshape these methods anyway.

rami3l:
> what should I do with [fc00f5f](https://github.com/rust-lang/rustup/commit/fc00f5ffbb45a4bef99b9b786c1f467161edcb35)? The whole point of that test is that it fails without the fix.

@cachebag Sorry for the late reply!

As per our current contribution guidelines, each commit is supposed to pass the test suite on its own, so we encourage adding a test upfront demonstrating the current behavior (even the current behavior is an error being thrown) upfront, and when the fix arrives, the test case is updated accordingly.

cachebag:
@rami3l understood, thank you for the feedback that makes a lot of sense. this is how i've decided to split it out then:

- c26231c9 is now a separate harness-only commit.
- df596700 adds the test upfront asserting the current behavior (replacer fails, `rustup` is gone).
- a6f8d544 flips the test to assert the replacement succeeds.

i ran the race test at each of those commits locally to confirm it passes at every step.

no rush of course: but i've left some threads open with questions/comments on your review(s).

src/utils/mod.rs
djc: Can live with it.

djc:
> @djc
> I split these at @rami3l's request in [#5077 (review)](https://github.com/rust-lang/rustup/pull/5077#pullrequestreview-5233569083). I do not mind either way; could either of you propose the preferred structure?

I don't really disagree with anything @rami3l wrote there, but I think in the execution "logically independent commits" can be up to interpretation. Here are some guidelines that I like to keep in mind: specifically when you're mainly adding code, until you've actually achieved a desirable outcome the intermediate outcomes (that don't achieve a desirable outcome yet) aren't as valuable, because the state of the code really only makes sense when you achieve the desirable outcome. For adding code, splitting up the additions into multiple code often doesn't really make it more reviewable (which is very different from when you're adding code and changing existing code in the same commit).

So in that light do you think the current commit history could be simplified somewhat? @rami3l thoughts?

rami3l:
@djc I agree with you that if one single change is adding a large chunk of stuff then there is generally no use in splitting it into multiple commits.

However, my previous point was just that since the previous version of that change (`fix(self-update): publish the rustup binary atomically`) when I was making the commit was proposing alternatives to quite a few existing APIs in a non-mechanical way, I needed to do a lot of back-and-forth jumps to make sure that the code stayed equivalent before and after the change.

@cachebag The new split looks pretty good to me, so I'd consider that concern to have been addressed. Nice work on that matter!

Frankly speaking, I think the current state of this PR looks pretty polished already (probably modulo some minor concerns). I'd love to do another round once you have done with the rebase and I'm looking forward to merging it soon :)

src/cli/self_update/windows.rs
```diff
@@ -651,13 +654,14 @@ pub(crate) fn remove_uninstall_registry_entry(process: &Process) -> anyhow::Resu
...
-    Command::new(setup_path)
-        .arg("--self-replace")
-        .spawn()
-        .context("unable to run updater")?;
+pub(super) fn run_update(
+    prepared_updater: PreparedUpdater,
+    process: &Process,
+) -> anyhow::Result<utils::ExitCode> {
+    let updater_path = prepared_updater.path().to_owned();
+    prepared_updater.spawn_replacer()?;
```
rami3l: Nit: It feels like `spawn_replacer()` should return both a child and its path?

tests/suite/cli_self_upd.rs
```diff
@@ -1236,6 +1255,26 @@ async fn install_minimal_profile() {
...
+    })
+    .unwrap();
+}
+
+fn managed_updater(rustup_home: &Path) -> PathBuf {
+    rustup_home
+        .join("self-update")
+        .join(format!("rustup-init{EXE_SUFFIX}"))
+}
+
```
rami3l: Nit: `mod self_update` is public so I guess you can reuse some constants/helpers from over there.

src/cli/self_update/stage.rs
rami3l: > The lock only says whether an update is running right now. After the parent exits and before the replacer takes the lock, cleanup sees a free lock and an updater on disk and can’t tell the difference between being finished or queued to run. Deleting it in the second case is the bug.

Please correct me if I have misread anything, but it seems to me that this analysis didn't tell the full story. In the following snippet that you have written:

```rs
    let result = process.cargo_home().and_then(|cargo_home| {
        self_update_lock.install_bins(&cargo_home.join("bin"), super::force_hard_links(process))
    });
    stage::mark_result(result.is_ok(), process);
    result?;
```

If the process crashes in `let result = ...`, then the cleanup never happens, because the proxy still thinks that the updater is queued to run, so not deleting the replacer in this case is still a bug?

src/cli/self_update/windows.rs
```diff
@@ -656,26 +656,24 @@ pub(crate) fn remove_uninstall_registry_entry(process: &Process) -> anyhow::Resu
...
-        self_update_lock.install_bins(&cargo_home.join("bin"), super::force_hard_links(process))
-    });
+    let result = process
+        .cargo_home()
+        .and_then(|cargo_home| {
+            self_update_lock.install_bins(&cargo_home.join("bin"), super::force_hard_links(process))
+        })
+        .and_then(|()| {
+            update_uninstall_registry_display_version(env!("CARGO_PKG_VERSION"), process)
+        });
```
rami3l: Nit: Maybe we can do both steps in one callback?

rami3l (approved):
LGTM, thanks a whole lot!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/2919b80da905d44cc731c01add014c2eb3648aed..83ae15fe60a06d036be2ed1e9cf6e63a93ce8f77)*

src/cli/self_update/stage.rs
rami3l: I like the `Deref` idea in https://github.com/rust-lang/rustup/pull/5077#discussion_r4051977445. Could you explore a bit more how to make the relevant APIs more idiomatic? Although I think this may not necessarily be a generic container: `impl Deref for SelfUpdateLock { type Target = PathBuf; }` seems enough.

src/cli/self_update/stage.rs
rami3l: I think APIs like `MutexGuard::map(...)` would be an overkill here; just holding a lock in the `_lock` field looks fine to me. The `path` is not the same as `*self._lock` so there shouldn't be much confusion here...

src/cli/self_update/stage.rs
cachebag: Correct. So if the replacer dies before `mark_result`, no marker gets written and cleanup in this commit can't tell that updater apart from one that is queued to run, so it leaves it alone. The next commit uses age as the tiebreaker, since a real self-update never takes anywhere near a day.

Now, it is worth noting that while the replacer is alive, its lock keeps cleanup away entirely (`try_acquire` returns `None`), and the next `rustup self update` clears the leftover updater and markers in `prepare_updater` before downloading, regardless of age. So the only cost of a crash there is one stale file under `$RUSTUP_HOME/self-update` until either of those happens.

The comment at the top of `cleanup_at` says this in short form.

src/cli/self_update/stage.rs
cachebag: The `Deref` concept will go in via https://github.com/rust-lang/rustup/pull/5077#discussion_r4084278252 so `.path()` goes away.

src/cli/self_update/stage.rs
cachebag: What I ended up doing is `Deref<Target = Path>` on `PreparedUpdater` rather than on `SelfUpdateLock`. 

The lock only knows the stage directory; the prepared updater is what carries the path callers actually need. `Path` over `PathBuf` so callers get the borrowed form, the same way `PathBuf` itself derefs. `path()` is gone and the methods are `lock()` / `try_lock()` now.

cachebag: I think this makes the _most sense_ but I am open to hearing what you think.

rami3l: @cachebag Looks good to me as well :)

rami3l (approved):
Let's merge this and see how it goes online. Nice work!

@cachebag Would you mind reminding me when we ship this patch to the next beta so that our users can test the self updater and/or the uninstaller more? Unfortunately I don't use it at all in my daily-driving build powered by homebrew :(

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/bf98b8ca3b488f7a1a1daaf0b4febc4da6a21d94..779a067c766f862083c8f9541a4c95175c937499)*

cachebag:
> Let's merge this and see how it goes online. Nice work!
> 
> @cachebag Would you mind reminding me when we ship this patch to the next beta so that our users can test the self updater and/or the uninstaller more? Unfortunately I don't use it at all in my daily-driving build powered by homebrew :(
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5077/bf98b8ca3b488f7a1a1daaf0b4febc4da6a21d94..779a067c766f862083c8f9541a4c95175c937499)_

Absolutely. 

I use it quite often at work for our cross-compile toolchains so I can test it out myself pretty thoroughly as well.

PR #5080, opened 2026-09-15
refactor(toolchain)!: remove multiple components at once

rami3l:
Part of https://github.com/rust-lang/goals/issues/736.

Currently, a new transaction is set up when removing every component, which is clearly unnecessary.

This PR addresses that issue by removing multiple components at once.


src/cli/rustup_mode.rs
```diff
@@ -1520,15 +1520,18 @@ async fn target_remove(
         cfg,
     )
     .await?;
+    let targets = targets
+        .into_iter()
+        .map(TargetTuple::new)
+        .collect::<Vec<_>>();
```
rami3l: This is very cheap because `TargetTuple` is a newtype on top of `String`. Under `-C opt-level=3` this is guaranteed to have minimal overhead.

Verified by https://godbolt.org that the following snippets have the same output:

```rs
#[unsafe(no_mangle)]
pub fn should_be_id(targets: Vec<String>) -> Vec<TargetTuple> {
    targets.into_iter().map(TargetTuple::new).collect()
}

pub struct TargetTuple(String);
impl TargetTuple {
    pub fn new(name: String) -> Self { Self(name) }
}
```

```rs
#[unsafe(no_mangle)]
pub fn id(targets: Vec<String>) -> Vec<String> { targets }
```

djc: I'm not worried about the performance of 1 allocation for a CLI tool invcation (but it could well be that I'm not especially consistent in how I review for performance).

djc (commented):
I think this direction is okay? But I'd like smaller commits, and I'd probably move `DistributableToolchain::remove_components()` out of `DistributableToolchain` into `rustup_mode` as a first step.

If we're going to pass an `IntoIterator<Item = Component>` into it, then it doesn't really make sense to me in terms of program flow (as opposed to performance) to allocate a `Vec` for it upfront, then iterate over it to check some stuff, then transform the `Vec` into an iterator again.

So I'd like to challenge you to rethink this code a bit more than you're currently doing.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5080/7464de1b6121e5ee0fceee3bdcd695ded379cc12..57d1d7db460bdde6c6762cfcaca11edd7735ebad)*

rami3l:
> I'd probably move `DistributableToolchain::remove_components()` out of `DistributableToolchain` into `rustup_mode` as a first step.

@djc I'm not sure to have fully understood your suggestion here. As it currently stands, `DistributableToolchain::remove_components()` is really a distributable-level abstraction that powers both `rustup component remove` and `rustup target remove`, both of which have a dispatch function in `rustup_mode.rs` that contains the specific validation and user-facing displaying logic (which involves converting `Vec<>` from clap to real component lists). That looks quite reasonable to me as the function organization pattern?

What I'm doing in the current patch is really making the underlying implementation accept multiple components at once, so that the dispatching functions don't need to call it in a loop.

If you agree with that, then in terms of argument parsing and history rewriting of course I can reevaluate my decisions.

rami3l:
> If we're going to pass an `IntoIterator<Item = Component>` into it, then it doesn't really make sense to me in terms of program flow (as opposed to performance) to allocate a `Vec` for it upfront, then iterate over it to check some stuff, then transform the `Vec` into an iterator again.
> 
> So I'd like to challenge you to rethink this code a bit more than you're currently doing.

@djc But looking back, what you have pointed out here makes a lot of sense actually. I should have tucked the allocation into the `clap` layer so the flow will be cleaner on our side. I'll try to do that instead.

src/cli/rustup_mode.rs
```diff
@@ -551,7 +551,7 @@ enum TargetSubcmd {
     Remove {
         /// List of targets to uninstall
         #[arg(required = true, num_args = 1..)]
-        target: Vec<String>,
+        target: Vec<TargetTuple>,
```
rami3l, line 553: Note: I failed to do the same thing for `rustup component rm` because the parsing logic that generates `Vec<Component>` relies on `Cfg` to the point that it can't be put into `value_parser` easily, so I have to keep `Vec<String>` first and parse them later in the dispatch function. I'm not sure if we should put part of the validation logic in the latter part of the pipeline: the need to bail out fast here seems to be quite legit.

The same problem exists for `rustup component add`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5080/2fc296f17a1059b4c6e4156a7882fd2d8d82bd1f..50cdc5ee5e0c3a369d29a94596b49cba81373677)*

rami3l: @djc Any thoughts?

Alternatively, I could make `distributable.(add|remove)_components` accept a `IntoIterator<Item = anyhow::Result<Component>>` so that the error handing is further delayed, if that makes sense to you.

rami3l: Update: I tried the above idea and the history does look cleaner.

src/dist/mod.rs
```diff
@@ -678,6 +678,14 @@ impl TargetTuple {
     }
 }
 
+impl FromStr for TargetTuple {
+    type Err = Infallible;
+
+    fn from_str(name: &str) -> Result<Self, Self::Err> {
+        Ok(Self(name.to_string()))
+    }
+}
```
djc, line 685: Could clap deal with `From<String> for TargetTuple`, or is there another way to slightly rewrite this that saves the unnecessary allocation?

djc (approved):
LGTM, nice!

For the future, suggest doing the rightward drift reduction before the changes to do multi-component removal because it likely makes the latter change easier to review.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5080/02a69d73f69dbc42dfa26667ab836d88e7d80140..9020f9b64de48b8429a191808a635e6da89f430d)*

src/dist/mod.rs
rami3l: @djc I think clap can detect that as well. Let me just switch to that.

src/cli/rustup_mode.rs
```diff
@@ -1523,33 +1522,21 @@ async fn target_remove(
...
-        if has_at_most_one_target {
-            warn!("removing the last target; no build targets will be available");
-        }
-        distributable
-            .remove_component(Component::std(target))
-            .await?;
+    if targets.contains(&cfg.default_host_tuple()?) {
+        warn!(
+            "removing the default host target; proc-macros and build scripts might no longer build"
+        );
```
FranciscoTGouveia, line 1528: Nit: Shouldn't we print this after confirming the removal succeeded? I know the phrasing is "removing" and not "removed", but still, a command like `rustup target rm <host> non-existent` would give the warning and fail without removing anything.

Maybe we could move this to after the actual removal?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5080/02a69d73f69dbc42dfa26667ab836d88e7d80140..367c94afa800d5eb11d6169711a17a5986b88c6c)*

FranciscoTGouveia: Oops, I was looking into the code and did not see that this was already queued. Apologies :)

djc: Feedback is still useful!

rami3l: > Nit: Shouldn't we print this after confirming the removal succeeded? I know the phrasing is "removing" and not "removed", but still, a command like `rustup target rm <host> non-existent` would give the warning and fail without removing anything.

@FranciscoTGouveia Thanks for the feedback!

I'd like to clarify that since https://github.com/rust-lang/rustup/pull/4797, `rustup target rm <host> non-existent` will cause `<host>` to be removed, only AFTER which the warning about `non-existent` will surface (same thing even if you put `non-existent` before `<host>`, if you are thinking about that), so we are quite unlikely to encounter false positives here merely due to user input. If the command execution really fails in the middle, I would still like to warn the user at first.

PR #5081, opened 2026-09-15
refactor(install): simplify `InstallMethod` methods

rami3l:
Part of https://github.com/rust-lang/goals/issues/736.

Cherry-picked from https://github.com/rust-lang/rustup/pull/5080 to be more digestible. This patch doesn't involve any complex changes, just some simple inlining and deduplication.

PR #5083, opened 2026-09-15
refactor(install): simplify `InstallMethod` methods, pt. 2

rami3l:
Part of https://github.com/rust-lang/goals/issues/736.

This is a simple follow-up of #5081. As usual, only inlining and deduplication are involved.

However, I might not further simplify this API from here, because there will be more stuff to be returned from `.install()` later on, notably the stale object to be garbage collected...


djc:
image

Here's one reason I'm not a big fan of your commit message style.

rami3l:
@djc I agree I could have been more succinct in composing my commit messages. I've just shortened them to be below 70 letters each. If you would like to suggest a general style and/or a length limit for the first line (for example, you may want to keep the most important messages upfront as you have done with the function parameters so that it may be easier for you to review on the particular client you are using), please feel free to let me know :)

That said, I'm still surprised by the bad presentation here. It looks like this on my mobile:

image

djc:
> @djc I agree I could have been more succinct in composing my commit messages. I've just shortened them to be below 70 letters each. If you would like to suggest a general style and/or a length limit for the first line (for example, you may want to keep the most important messages upfront as you have done with the function parameters so that it may be easier for you to review on the particular client you are using), please feel free to let me know :)

I think if you're going to stick with "Conventional Commits"-style messages, you should never use a two-segment scope style. Personally, I don't like the conventional commits style, here's a recent blog post arguing against it:

https://sumnerevans.com/posts/software-engineering/stop-using-conventional-commits/

djc (approved):

src/install.rs
```diff
@@ -40,7 +40,13 @@ impl InstallMethod<'_, '_> {
         let _ = rayon::ThreadPoolBuilder::new()
             .num_threads(self.cfg().process.io_thread_count()?.into())
             .build_global();
-        let local_name = self.local_name();
+
+        let local_name = match &self {
+            Self::Link { dest, .. } => (*dest).clone().into(),
```
djc, line 45: FWIW, I don't like the repeated usage of `(*dest).clone().into()` very much because it's very unclear what it's doing and has quite a lot of line noise. I think `LocalToolchainName::from(dest.clone())` would be better.

I also think of the two names `dest` is much worse than `toolchain`, because it carries very little context.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5083/62c4ab1c849e9ac8d18db2f9c94a5cc6f2cb2bbc..325f6834281bcfff2d6b8a4e98c52125fe4df2d6)*

rami3l: I agree. Actually that is one of the reasons I tried to inline those methods: the naming system is just confusing. However I believe this is not in the scope of the current patch. Should I add something on top of this?

rami3l:
> Personally, I don't like the conventional commits style

@djc Thanks for voicing your opinions on this. For now I don't think it's practical to enforce a commit message style or format, but I will try my best to avoid long prefixes and keep the message length to 50 chars or so.

> here's a recent blog post arguing against it:
> 
> [sumnerevans.com/posts/software-engineering/stop-using-conventional-commits](https://sumnerevans.com/posts/software-engineering/stop-using-conventional-commits/)

I have already read that blog post so I see your point, but I still have my reservations on this topic. From my personal experience, this is sometimes helpful with changelogs and backports (especially when I want to distinguish unexpected breakages from refactoring patches and intended breakages from features).

src/install.rs
djc: It could definitely be in a separate PR -- otherwise I wouldn't have approved this already.

(And to be clear, I do think the changes in this PR are nice improvements!)

djc:
> @djc Thanks for voicing your opinions on this. For now I don't think it's practical to enforce a commit message style or format, but I will try my best to avoid long prefixes and keep the message length to 50 chars or so.

I think keeping the message length to 50 chars is actually an anti-pattern -- it's good to make use of the full 70 or so chars for the first line, but not by filling it with repetitive boilerplate.

> I have already read that blog post so I see your point, but I still have my reservations on this topic. From my personal experience, this is sometimes helpful with changelogs and backports (especially when I want to distinguish unexpected breakages from refactoring patches and intended breakages from features).

I understand that part, but IIRC the convention allows the simple use of `!` to denote incompatible changes -- so those don't actually need to take up a lot of space?

rami3l:
> but IIRC the convention allows the simple use of `!` to denote incompatible changes -- so those don't actually need to take up a lot of space?

No it doesn't, I agree as I've said above that my previous prefix was too long.

> I think keeping the message length to 50 chars is actually an anti-pattern -- it's good to make use of the full 70 or so chars for the first line, but not by filling it with repetitive boilerplate.

I am thinking about like around 50 without the prefix and in total it should be around 70. So I guess we are on the same page here.

src/install.rs
rami3l: Closed by #5086.

PR #5086, opened 2026-09-16
refactor(install): simplify `InstallMethod` methods, pt. 3

rami3l:
Part of https://github.com/rust-lang/goals/issues/736; continuation of #5083.

Specifically, this patch hopes to address the concern raised by @djc in https://github.com/rust-lang/rustup/pull/5083#discussion_r4024091192.


djc (approved):

djc:
Good stuff! Still wondering if we can improve on the need to deref-then-clone some kinds of toolchain names, which looks pretty ugly to my eyes.

rami3l:
> Good stuff! Still wondering if we can improve on the need to deref-then-clone some kinds of toolchain names, which looks pretty ugly to my eyes.

@djc I tried, but that would involve the introduction of the `ref` keyword which will make you hate it even more, which is why I didn't do it.

PR #5087, opened 2026-09-16
print empty toolchain message to  stderr instead of  stdout

### LINKED ISSUE #5046: Disambiguate "no toolchains installed"
### rustup to provide a way to list toolchains without human "hints"

Greetings from [fish-shell](https://github.com/fish-shell/fish-shell/). The fish command line shell ships tab-completions for rustup. A user might type `rustup --toolchain <TAB>` and expect the toolchains to be offered as completions; fish-shell invokes `rustup toolchain list` to discover these options.

But [this recent rustup change](https://github.com/rust-lang/rustup/pull/4912) causes rustup to print a "setup hint" to stdout if no toolchains are installed. This setup hint emits the text "no installed toolchains," which is difficult to disambiguate from a real toolchain (in practice our shoddy regex computes "no" as a toolchain).

We discovered this through test failures that [we have worked around](https://github.com/fish-shell/fish-shell/commit/7e577c8a0ef187893aae2a4de208b7e7d2e636d2). We don't expect this to affect many users, since rustup users overwhelmingly have at least one rust toolchain installed. Still I think this is worth surfacing, and fixing.

### Solution you'd like

This request is for a machine-friendly way to determine installed toolchains without unwanted "hint" stdout intended for humans. I think any of these would suffice (priority order):

- Emit the hint to stderr, not stdout, or
- Respect the `-q/--quiet` option to suppress this hint if there are no toolchains installed, or
- Exit with a non-zero status code if there are no toolchains installed

### Notes

Now that I've written this up, my gut says stderr is the best option for this hint. I might write up a PR for that.

Sreehari425:
closes #5046 

`rustup toolchain list` prints `no installed toolchains` to stdout , when no toolchains are installed in the system, this can cause problem for some tools which parses the stdout as machine readable.

for this fix , i moved the message from stdout -> stderr , so stdout stay clean for the program that depend on it.


src/cli/common.rs
```diff
@@ -323,7 +323,7 @@ pub(crate) async fn list_toolchains(
 ) -> anyhow::Result<ExitCode> {
     let mut toolchains = cfg.list_toolchains(quiet)?;
     if toolchains.is_empty() {
-        writeln!(cfg.process.stdout().lock(), "no installed toolchains")?;
+        writeln!(cfg.process.stderr().lock(), "no installed toolchains")?;
```
rami3l: As discussed in https://github.com/rust-lang/rustup/issues/5046#issuecomment-5506244717, I would prefer to use the `info!()` macro here instead of writing to `stderr` directly.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5087/2fc296f17a1059b4c6e4156a7882fd2d8d82bd1f..1957a8c685c98c63d3530b82fce451354ac6ac36)*

rami3l (approved):
LGTM and thanks for the contribution! Please squash your changes into a single commit and then I can merge it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5087/2fc296f17a1059b4c6e4156a7882fd2d8d82bd1f..1b3ba2e1fd5be82bed4c8c4c003acb21a89f568f)*

src/cli/common.rs
Sreehari425: thanks for reviewing, switched to `info!()` 👍

PR #5089, opened 2026-09-16
fix(windows/uninstall-gc): cleanup gc program even if uninstall failed

Cloud0310:
Proceeding of #5088.
This fixed cleanup `rustup-gc.exe` even if uninstallation fails, also keeping the error information for better tracing.


rami3l:
@Cloud0310 I think you can close this and drop the last commit over to #5088. Just deleting/pushing your branches on your fork should do the job.

PR #5090, opened 2026-09-16
refactor(windows/uninstall-gc): Run Windows uninstall GC from Alternate data streams

Cloud0310:
Proceeding of #5088 
This PR does:
- Run Windows uninstall GC from ADS and ultilize the POSIX semantics for better gc experience.
- Use `tempfile` to manage the GC file and copy the running executable’s contents into it. This creates an independent executable instead of reproducing a source symlink and avoids relying on the installed rustup copy.
- Use `fs::copy` and `utils::current_exe` for copying gc program, allowing for potential copying dead symlink and being less dependent upon fixed binary path (`~/.cargo/bin/rustup.exe`).


rami3l:
@Cloud0310 When stacking a PR on top of another, you should keep the top PRs as draft and wait for the prerequisites to be merged first.

PR #5091, opened 2026-09-16
fix(uninstall): reuse resolved cargo home during cleanup

Cloud0310:
Reuse cargo home resolution result in APIs.
Part of #5056.


ChrisDenton (approved):

rami3l:
@Cloud0310 What is the rationale of this PR? As I see it, it doesn't always remove the implicit dependency on `process.cargo_home()`, because functions like `do_remove_from_path()` still rely on it?

Cloud0310:
Here, the purpose is for later allowing passing down the actual bin path for clean up later in future PR. But as for `do_remove_path` the final view is kinda different from others. It's also carrying the duty of resolving home and the location of legacy rc files via `shell::legacy_paths(process)`.
So this part isn't included as this would require further modification.

I would re-check on this, then make another fix.

rami3l:
@Cloud0310 I see. In that case I think it'd be beneficial to resolve all data dependencies upfront, rather than passing both `process` and `cargo_home` which might be a bit confusing because I have the feeling that the latter subsumes the former. You can do this in a subsequent PR.

PR #5094, opened 2026-09-18
Suggest alternative toolchains for unavailable targets

### LINKED ISSUE #4164: Suggest another toolchain when removing a target unavailable to the current toolchain

1. `rustup update` fails (unrelated to issue)
Image
2. `rustup target remove wasm32-wasi` fails
Image
3. `rustup +target remove wasm32-wasi` succeeds, as the default toolchain was nightly as specified in `.rustup/settings.toml`

### Solution you'd like

Rustup could search and suggest targets on other versions when the default does not have the target

### Notes

_No response_

Sreehari425:
Closes #4164

`rustup target remove <target>` now suggests another installed toolchain when the target isn't available in the current one. If no other toolchain has the target, the existing error is unchanged.


src/toolchain/distributable.rs
```diff
@@ -412,20 +436,35 @@ impl<'a> DistributableToolchain<'a> {
...
+                let target = component
+                    .target
+                    .as_ref()
+                    .expect("component target should be known");
+                let toolchain_suggestion =
+                    self.find_toolchain_with_target(target).map(|toolchain| {
+                        TargetSuggestion::Toolchain(format!(
+                            "rustup {} target remove {}",
+                            toolchain, target
+                        ))
```
Cloud0310: Use new style format (inline this two var)?

src/toolchain/distributable.rs
```diff
@@ -412,20 +436,35 @@ impl<'a> DistributableToolchain<'a> {
...
                     desc: Box::new(self.desc.clone()),
-                    target: component.target.expect("component target should be known"),
-                    suggestion,
+                    target: target.clone(),
+                    suggestion: final_suggestion,
                 }
                 .into());
             }
+
+            let suggestion = self.get_component_suggestion(&component, &config, &manifest, true);
```
Cloud0310, line 478: Seems suggestion is used on both inside the missing branch and the outer branch?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/615e345a3fb1e067b43f043029f3bdb0b57bc189..6e345bfd283e49a33bca52b7ed8ed10f300c8121)*

src/errors.rs
```diff
@@ -192,6 +198,17 @@ fn suggest_message(suggestion: &Option<String>) -> String {
     }
 }
 
+pub(crate) fn target_not_installed_message(suggestion: &Option<TargetSuggestion>) -> String {
```
Cloud0310: Do we use this helper func elsewhere? If not, guess maybe inline this func.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/615e345a3fb1e067b43f043029f3bdb0b57bc189..6e345bfd283e49a33bca52b7ed8ed10f300c8121)*

src/toolchain/distributable.rs
```diff
@@ -385,6 +385,30 @@ impl<'a> DistributableToolchain<'a> {
         }
     }
 
+    fn find_toolchain_with_target(&self, target: &TargetTuple) -> Option<String> {
```
Cloud0310: Do we use this elsewhere? If not, the same as https://github.com/rust-lang/rustup/pull/5094/changes#r4052997668.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/615e345a3fb1e067b43f043029f3bdb0b57bc189..6e345bfd283e49a33bca52b7ed8ed10f300c8121)*

src/toolchain/distributable.rs
```diff
@@ -412,20 +436,35 @@ impl<'a> DistributableToolchain<'a> {
...
                 .any(|c| c.target() == component.target())
             {
+                let target = component
+                    .target
+                    .as_ref()
+                    .expect("component target should be known");
+                let toolchain_suggestion =
+                    self.find_toolchain_with_target(target).map(|toolchain| {
+                        TargetSuggestion::Toolchain(format!(
+                            "rustup {} target remove {}",
```
Cloud0310: Use new style format and inline these two vars? 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/615e345a3fb1e067b43f043029f3bdb0b57bc189..6e345bfd283e49a33bca52b7ed8ed10f300c8121)*

src/toolchain/distributable.rs
Sreehari425: I kept this as a separate helper even though it’s only used here, since I felt moving the toolchain lookup out of remove_components made the function a bit easier to read.

Cloud0310: This func sounds like a method of `TargetTuple`?

Sreehari425: I don’t think this belongs on `TargetTuple`, since it only represents the target and has no knowledge of installed toolchains or rustup configuration

I kept it on `DistributableToolchain`, which already has that context. Moving it to `TargetTuple` would require passing that context in, making the API more complicated than necessary.

src/toolchain/distributable.rs
```diff
@@ -413,19 +434,31 @@ impl<'a> DistributableToolchain<'a> {
...
                 .any(|c| c.target() == component.target())
             {
+                let target = component
+                    .target
+                    .as_ref()
+                    .expect("component target should be known");
+                let suggestion = self
+                    .find_toolchain_with_target(target)
+                    .map(|toolchain| {
+                        TargetSuggestion::Toolchain(format!(
```
djc: I don't think it makes sense to format here? We could just store `toolchain` and `target` values in the `TargetSuggestion::Toolchain` variant.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..52dd32cbeb953ed28af480d62ff651b77d183f8a)*

src/toolchain/distributable.rs
```diff
@@ -413,19 +434,31 @@ impl<'a> DistributableToolchain<'a> {
             }
 
             let suggestion = self.get_component_suggestion(&component, &config, &manifest, true);
```
djc: If/where we're not using this `suggestion`, we shouldn't derive it in the first place.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..52dd32cbeb953ed28af480d62ff651b77d183f8a)*

src/toolchain/distributable.rs
```diff
@@ -385,6 +385,27 @@ impl<'a> DistributableToolchain<'a> {
         }
     }
 
+    fn find_toolchain_with_target(&self, target: &TargetTuple) -> Option<String> {
```
djc: As a single-caller function, I'm not sure having this as a separate function really makes sense? Perhaps there should be a `TargetSuggestion` constructor (or two) that can take all the relevant context and comes up with a value.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..52dd32cbeb953ed28af480d62ff651b77d183f8a)*

src/errors.rs
```diff
@@ -22,6 +22,21 @@ use crate::{
 
 pub(crate) const DEFAULT_STABLE_HINT: &str = "help: run 'rustup default stable' to download the latest stable release of Rust and set it as your default toolchain.";
 
+#[derive(Debug, Clone)]
+pub enum TargetSuggestion {
+    Toolchain(String),
+    Component(String),
+}
+
+impl std::fmt::Display for TargetSuggestion {
```
djc: Nit: import `fmt` at the module level.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..52dd32cbeb953ed28af480d62ff651b77d183f8a)*

djc:
Uh, this seems to have a bunch of unrelated changes, and lost the useful commit history split in the process.

Sreehari425:
> Uh, this seems to have a bunch of unrelated changes, and lost the useful commit history split in the process.

yeah you are  right sorry. I was trying to make the target changes into one clean commit and accidentally pulled the unrelated changes into it too. I messed up the history, fixing it now rn

Sreehari425:
@djc  thanks for reviewing this earlier :), i have  addressed the issues you pointed out and cleaned up the changes would appreciate another look when you get a chance.

src/toolchain/distributable.rs
```diff
@@ -384,7 +384,51 @@ impl<'a> DistributableToolchain<'a> {
             ))
         }
     }
+}
+
+impl TargetSuggestion {
```
djc, line 389: The impl should be next to the type definition IMO.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..128fe320671399752fd0ca6d932e650269690e3d)*

src/toolchain/distributable.rs
```diff
@@ -384,7 +384,51 @@ impl<'a> DistributableToolchain<'a> {
...
+    fn from_target(
+        distributable: &DistributableToolchain<'_>,
+        target: &TargetTuple,
+        component: &Component,
+        config: &Config,
+        manifest: &Manifest,
+    ) -> Option<Self> {
+        let toolchain = distributable
```
djc, line 397: I think isolating all this is an improvement, but the logic here is still pretty convoluted. I think there can be a number of early returns that make this easier to follow.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..128fe320671399752fd0ca6d932e650269690e3d)*

src/toolchain/distributable.rs
```diff
@@ -384,7 +384,51 @@ impl<'a> DistributableToolchain<'a> {
...
+        toolchain
+            .map(|name| Self::Toolchain {
+                name: name.to_string(),
+                target: target.clone(),
+            })
+            .or_else(|| {
+                distributable
+                    .get_component_suggestion(component, config, manifest, true)
+                    .map(Self::Component)
+            })
```
djc, line 427: Instead of `map().or_else()`, it's better to just use `match`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..128fe320671399752fd0ca6d932e650269690e3d)*

src/errors.rs
```diff
@@ -22,6 +22,23 @@ use crate::{
...
+    Component(String),
+}
+
+impl fmt::Display for TargetSuggestion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Toolchain { name, target } => {
+                write!(f, "; try `rustup +{name} target remove {target}`")
+            }
+            Self::Component(name) => write!(f, "; did you mean '{name}'?"),
```
rami3l, line 37: Nit: I think this has a tendency to become too long a line. One trick we can use here would be prepending the string with `\nhelp:` like so:

https://github.com/rust-lang/rustup/blob/2919b80da905d44cc731c01add014c2eb3648aed/src/cli/rustup_mode.rs#L1256-L1258

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..128fe320671399752fd0ca6d932e650269690e3d)*

src/errors.rs
```diff
@@ -162,11 +179,11 @@ pub enum RustupError {
         suggestion: Option<String>,
     },
     #[error("toolchain '{}' does not have target '{}' installed{}\n", .desc, .target,
```
rami3l, line 181: Probably related to https://github.com/rust-lang/rustup/pull/5094/changes/128fe320671399752fd0ca6d932e650269690e3d#r4080162242, but those error messages don't usually have a trailing `\n`. Is this an existing bug that we can use the opportunity to fix here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..128fe320671399752fd0ca6d932e650269690e3d)*

src/toolchain/distributable.rs
```diff
@@ -384,7 +384,51 @@ impl<'a> DistributableToolchain<'a> {
...
+                        return None;
+                    }
+
+                    let toolchain =
+                        Toolchain::new(distributable.toolchain.cfg, toolchain_name.clone().into())
+                            .ok()?;
+                    let installed_targets = toolchain.installed_targets().ok()?;
+                    installed_targets.contains(target).then_some(toolchain_name)
+                })
+            });
```
rami3l, line 416: We prefer the imperative style instead of the functional one here. You may use streaming methods from time to time for certain intermediate expressions, but the callbacks should stay concise and clear, and the overall flow should be imperative. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5094/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..128fe320671399752fd0ca6d932e650269690e3d)*

src/toolchain/distributable.rs
rami3l: Actually, the same suggestion is still used both here and in https://github.com/rust-lang/rustup/pull/5094/changes#r4052993357.

Sreehari425:
i have addressed the feedback and pushed the changes. 
i would appreciate another look when you get a chance :)

src/errors.rs
```diff
@@ -13,15 +13,142 @@ use thiserror::Error as ThisError;
...
+        Self::from_component(desc, component, config, manifest, true).map(Self::Component)
+    }
+
+    pub(crate) fn from_component(
+        desc: &ToolchainDesc,
+        component: &Component,
+        config: &Config,
+        manifest: &Manifest,
+        only_installed: bool,
+    ) -> Option<String> {
```
rami3l: I can't see how this is related to the `Self` type here... isn't this a suggestion about the component name, rather than the target name?

Sreehari425: yeah good point i thought grouping these together made sense while refactoring, but I agree it makes more sense as a separate helper.

src/errors.rs
```diff
@@ -13,15 +13,142 @@ use thiserror::Error as ThisError;
...
+impl fmt::Display for TargetSuggestion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Toolchain { name, target } => {
+                write!(f, "\nhelp: try `rustup +{name} target remove {target}`")
+            }
+            Self::Component(name) => write!(f, "\nhelp: did you mean '{name}'?"),
+        }
+    }
+}
```
djc: Nit: move the `Display` impl below the inherent impl.

src/errors.rs
```diff
@@ -13,15 +13,142 @@ use thiserror::Error as ThisError;
...
+
+impl TargetSuggestion {
+    pub(crate) fn from_target(
+        cfg: &Cfg<'_>,
```
djc: Nit: move the `cfg` argument last, since it's kind of boring/less specific.

src/errors.rs
```diff
@@ -13,15 +13,142 @@ use thiserror::Error as ThisError;
...
+}
+
+pub(crate) fn component_suggestion(
+    desc: &ToolchainDesc,
+    component: &Component,
+    config: &Config,
+    manifest: &Manifest,
+    only_installed: bool,
+) -> Option<String> {
+    use strsim::damerau_levenshtein;
```
djc: Nit: hoist the import up to the module level.

src/errors.rs
```diff
@@ -13,15 +13,142 @@ use thiserror::Error as ThisError;
...
+                    target: target.clone(),
+                });
+            }
+        }
+
+        component_suggestion(desc, component, config, manifest, true).map(Self::Component)
+    }
+}
+
+pub(crate) fn component_suggestion(
```
djc: Nit: moving this function should be in a separate commit.

djc:
The logic is looking a lot better, thanks for the iterations!

Sreehari425:
Thanks again for the review. Fixed all the suggestions you pointed out :)

djc (approved):

src/errors.rs
```diff
@@ -3,25 +3,151 @@
...
 use platforms::Platform;
+use strsim::damerau_levenshtein;
 use thiserror::Error as ThisError;
 use url::Url;
 
 use crate::{
+    config::Cfg,
     dist::{
         Channel, TargetTuple, ToolchainDesc,
+        config::Config,
```
rami3l: Nit: Could we import this somewhat differently, say `as DistConfig`? I'm afraid `config: &Config, .., cfg: &Cfg<'_>` is not looking very good below.

@djc WDYT? What is the convention in other repos you maintain?

rami3l (approved):

src/errors.rs
djc: Aliasing as `DistConfig` seems reasonable. Or just keep it as `dist::Config`?

Sreehari425: Done, aliased it as DistConfig 👍

PR #5096, opened 2026-09-21
refactor(install): decouple helpers from process then re-use the resolution result

Cloud0310:
#5091 was not a complete decouple of `Process`'s state from the related install/uninstall helpers.
So, this PR does:
- renaming relevant fucntions for better expressing what they're doing
- removing unnessary `Cow` usage of `cargo_home_str_with_home`, `cargo_home_str`, as the caller immediately constructs `String`
- decouple `Process`' states, e.g. `cargo_home` and other, allowing for re-using the resolution result.
- dedup `remove_legacy_source_command`, as the same logic is already done in `do_remove_from_path`

In the meantime, this the first of a series of refactors around uninstall/install/self-replacing logic.


djc (commented):
Most of this looks fine to me.

Decoupling helpers from `Process` seems pretty messy for ostensibly very limited benefit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5096/d15206e556507586c36c6737fdf47b177e71bff4..ce17bc77392dd2326f2dc2f4553627f1ca364adc)*

rami3l:
@djc I would like to provide some extra context regarding this part of the changes: it has been a follow-up of my request https://github.com/rust-lang/rustup/pull/5091#issuecomment-5711168981 in a previous attempt of decoupling the rcfile-related logic from `&Process`.

This is needed because the new uninstallation logic will attempt to remove the rcfile entries from both the `CARGO_HOME/bin` and `CARGO_BIN_HOME` installations in #5056, and thus these paths can no longer be injected from `&Process` anymore.

djc (approved):

src/cli/self_update.rs
```diff
@@ -1065,7 +1092,30 @@ fn clean_cargo_home(
...
+                    &shell::legacy_paths(process).collect::<Vec<_>>(),
+                )?;
+            }
+            #[cfg(windows)]
+            {
+                let environment = process
+                    .registry_environment_key()
+                    .context("Failed opening Environment key")?;
+                do_remove_from_path(cargo_home, &environment)?;
+            }
```
rami3l: Nit: You can use `cfg_select!{}` for this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5096/d15206e556507586c36c6737fdf47b177e71bff4..a672820135b0402a3d6a767bdc8d734a2da83f08)*

src/cli/self_update.rs
```diff
@@ -961,7 +986,7 @@ pub(crate) fn uninstall(
         } else {
             format!(
                 pre_uninstall_msg!(),
-                cargo_home = display_cargo_home(process)?
+                cargo_home = display_cargo_home(&cargo_home, process.home_dir().as_deref())
```
rami3l: Question: Judging from the immediate result, it seems like this call begs to be a new type declaration like the following...

```rs
struct CargoHomeDisplay { cargo_home: ..., home_dir: ... }
impl Display for CargoHomeDisplay { ... }
```

However I am wondering how you are planning to fit the current refactoring change into the final implementation of #5056. Ideally the APIs should stay as close as possible to the final shape. Would you mind elaborating a bit on that point (as well as whether you think the above suggestion makes sense WRT your final goal)? 🙏 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5096/d15206e556507586c36c6737fdf47b177e71bff4..a672820135b0402a3d6a767bdc8d734a2da83f08)*

src/cli/self_update.rs
Cloud0310: After narrowed the refactor scope into only home related things, this is no longer needed.

src/cli/self_update.rs
Cloud0310: > Question: Judging from the immediate result, it seems like this call begs to be a new type declaration like the following...
> 
> ```rust
> struct CargoHomeDisplay { cargo_home: ..., home_dir: ... }
> impl Display for CargoHomeDisplay { ... }
> ```

I though introducing a custom struct for this purpose is not worth for  two call sites. A better alternative is to unify upon the install message side. And this also aligned up with my xdg PR's goal.

> However I am wondering how you are planning to fit the current refactoring change into the final implementation of #5056. Ideally the APIs should stay as close as possible to the final shape. Would you mind elaborating a bit on that point (as well as whether you think the above suggestion makes sense WRT your final goal)? 🙏 

Yes, after re-thinking here, I think hiding platform things under the hood, and keep `home` related things here. This enables us for not messing up execution flow, also keeps the reusing of resolution here.

rami3l: > I though introducing a custom struct for this purpose is not worth for two call sites. A better alternative is to unify upon the install message side. And this also aligned up with my xdg PR's goal.

However that is not just 2 call sites: this also applies to the whole https://github.com/rust-lang/rustup/pull/5096/changes/645a1853cff58f6f031dab89e3da7be33236d89c as far as I am concerned. So it really seems to me that we would need a struct for this, or do you have other plans?

That said, since in your new patch `CARGO_HOME/<shell>.env` might no longer apply, passing `cargo_home` systematically probably isn't a good idea. Thus, I think we should also plan ahead for a bit if that goes with your plan. For example, extracting `cargo_home_bin` in the message template is pretty nice. Should we just rename it to `cargo_bin_dir` though so it makes things easier in the next stage?

Cloud0310: Maybe the best idea is to leave this helper along for now, as in #5056  would just remove this func. And it's kinda.
As for other parts, it's fixed now.

Cloud0310:
@rami3l seems there's an error upon  CI/CD, and it's not due my code. Could you re-run it?

And leaving alone the part of `Display`, I've refactored the git history for cleanup.

src/cli/self_update.rs
rami3l: > as in https://github.com/rust-lang/rustup/pull/5016 would just remove this func. And it's kinda.

@Cloud0310 What do you mean?

Cloud0310: Sorry for the typing issue here.
After final thought, I guess we still need this upon rustup bin home, and all solutions (https://github.com/rust-lang/rustup/pull/5056#issuecomment-5557550676) still share the home prefix: `%USERPROFILE%`, so it would be better to keep this abbr and use a struct and a `Display` trait for this.

src/cli/self_update.rs
```diff
@@ -1379,6 +1403,41 @@ mod tests {
...
+            let expected = if suffix.is_empty() {
+                home_dir_str.to_owned()
+            } else {
+                Path::new(home_dir_str).join(suffix).display().to_string()
+            };
+            assert_eq!(
+                BinHomeDisplay::new(&bin_home, Some(home)).to_string(),
+                expected
+            );
+        }
```
rami3l: Nit: Unroll this loop, otherwise it's not immediately clear what the `expected` variable would evaluate to.

src/cli/self_update.rs
```diff
@@ -204,7 +204,11 @@ impl InstallOpts<'_> {
             format!(post_install_msg_win!(), cargo_home = cargo_home)
         };
         #[cfg(not(windows))]
-        let source_env_lines = shell::build_source_env_lines(process);
+        let source_env_lines = {
+            let env_dir = process.cargo_home()?;
+            let home_dir = process.home_dir();
```
rami3l: Nit: Maybe inline those two temporaries?

src/cli/self_update.rs
```diff
@@ -584,17 +585,39 @@ fn update_root(process: &Process) -> String {
...
-fn canonical_cargo_home(cargo_home: &Path, home_dir: Option<&Path>) -> Cow<'static, str> {
-    let default_cargo_home = home_dir.unwrap_or_else(|| Path::new(".")).join(".cargo");
-    if default_cargo_home == cargo_home {
-        cfg_select! {
-            windows => r"%USERPROFILE%\.cargo".into(),
-            _ => "$HOME/.cargo".into(),
+/// Displays an installation path with a platform-specific home abbreviation.
+struct BinHomeDisplay<'a> {
+    bin_home: &'a Path,
+    home_dir_str: Option<&'static str>,
```
rami3l: Nit: It shouldn't hurt if you generalize this a bit and write `&'a str` here, because `&'static T: &'a T` always holds.

src/cli/self_update.rs
```diff
@@ -191,38 +191,26 @@ impl InstallOpts<'_> {
...
+            + &format!(
                 post_install_msg_unix!(),
-                cargo_home = cargo_home,
-                source_env_lines = source_env_lines,
-            )
-        };
+                env_dir = env_dir,
+                source_env_lines =
+                    shell::build_source_env_lines(process, &cargo_home, home_dir.as_deref()),
+            );
```
rami3l: Nit: I would prefer calling `md(&mut term)` two times rather than conditionally concating two strings like this.

src/cli/self_update.rs
```diff
@@ -193,8 +194,8 @@ impl InstallOpts<'_> {
 
         let cargo_home = process.cargo_home()?;
         let home_dir = process.home_dir();
-        let env_dir = canonical_cargo_home(&cargo_home, home_dir.as_deref());
-        let cargo_bin_dir = format!("{env_dir}{MAIN_SEPARATOR}bin");
+        let bin_home = cargo_home.join("bin");
```
rami3l: Nit: Maybe inline this new temporary?

src/cli/self_update.rs
```diff
@@ -584,17 +585,39 @@ fn update_root(process: &Process) -> String {
...
-    } else {
-        cargo_home.to_string_lossy().into_owned().into()
+    }
+}
+
+impl fmt::Display for BinHomeDisplay<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(home_dir_str) = self.home_dir_str {
+            f.write_str(home_dir_str)?;
+            if !self.bin_home.as_os_str().is_empty() {
```
rami3l: Nit: Just use `!self.bin_home.is_empty()`:

```rs
pub fn is_empty(&self) -> bool { self.as_os_str().is_empty() }
```

rami3l (approved):
LGTM modulo some tiny nits. Thanks again :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5096/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..f9994e1850a829523cdf694b140a190d98927cd1)*

Cloud0310:
Fixed all of them @rami3l .

rami3l (approved):

src/cli/self_update/msg.rs
rami3l, line 4: LGTM!

One final bonus: We don't seem to have snapshot tests for `(pre|post)_install_msg*` (particularly the latter).

To make sure that this refactoring action didn't change the output in a substantial way, it'd be cool to add snapshot tests for them. This would also benefit #5056 later on.

If this is easily feasible, you can add the tests right here. If not, it can also be a separate PR or part of #5056.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustup/5096/96f5c1c92b2ab0eeb6fe0ed3de61483a3d9f3d46..2468895cc04e2f2fc5099e0528820f5608089431)*

rami3l (approved):

